PR #701, opened 2016-06-19
delete include omp statement

keonkim:
`#inlcude <omp.h>` seems to be not needed. and it breaks build on mac.
`size_t` is not supported in omp for windows.


rcurtin:
If you want to include omp.h in the future, you should add it to `mlpack/core.hpp` with an ifdef (since OpenMP is not required to build mlpack), like `#ifdef OMP_FOUND` (or something like this, I am not sure, maybe there is some macro that is available when OpenMP is in use).

PR #696, opened 2016-06-14
edge_boxes: feature extraction updated, tests added

nilayjain:

PR #523, opened 2016-02-24
add train test split

stereomatchingkiss:
Add function to split train and test data

PR #703, opened 2016-06-20
Added the discretize function. Design changes as discussed.

nilayjain:
Some small changes pointed out are still to be made and comments need to be written.


nilayjain:
Only works with arma 7.2 right now.

src/mlpack/core/arma_extend/fn_index_min_max.hpp
```diff
@@ -0,0 +1,323 @@
...
+typename arma_not_cx<typename T1::elem_type>::result
+op_min::min_with_index(const Proxy<T1>& P, uword& index_of_min_val)
+  {
+  arma_extra_debug_sigprint();
+  
+  typedef typename T1::elem_type eT;
+  
+  const uword n_elem = P.get_n_elem();
+  
+  if(n_elem == 0)
```
stereomatchingkiss: I think this format is a little bit weird

nilayjain: Yeah it is not working. I willl change it and write my own function

src/mlpack/methods/edge_boxes/edge_boxes_main.cpp
```diff
@@ -0,0 +1,83 @@
...
+  split: options include 'gini', 'entropy' and 'twoing'
+  discretize: optional function mapping structured to class labels
+
+  stride: stride at which to compute edges
+  sharpen: sharpening amount (can only decrease after training)
+  n_tree_eval: number of trees to evaluate per location
+  nms: if true apply non-maximum suppression to edges
+  */
+
+  FeatureParameters params = FeatureParameters();
```
stereomatchingkiss: Thanks for the update :+1: 
In c++, when you declare an object, it will call the default constructor, so you can just write(Maybe you already know this)

`FeatureParameters params;`

In theory, initialize the object by using constructor is better than using copy assignment, but compiler should be able to optimize it, so it is just a preference of taste now.

src/mlpack/methods/edge_boxes/feature_extraction.hpp
```diff
@@ -0,0 +1,99 @@
...
+
+template <typename MatType = arma::mat, typename CubeType = arma::cube>
+class StructuredForests
+{
+ private:
+  FeatureParameters params;
+
+ public:
+
+  static constexpr double eps = 1e-20;
```
stereomatchingkiss: same as before, if there are no special, could you set it as private data member?

The lesser you let the users access you data directly, the more flexible you can refine your codes in the future without changing the api

src/mlpack/methods/edge_boxes/feature_extraction_impl.hpp
```diff
@@ -0,0 +1,1039 @@
...
+ * of feature_parameters class.
+ */
+template<typename MatType, typename CubeType>
+StructuredForests<MatType, CubeType>::
+StructuredForests(FeatureParameters F) 
+{
+  params = F; 
+}
+
+/*
```
stereomatchingkiss: If we do not need the codes again, I think we can remove it, git will record the codes we removed before

src/mlpack/methods/edge_boxes/feature_extraction_impl.hpp
```diff
@@ -0,0 +1,1039 @@
...
+namespace mlpack {
+namespace structured_tree {
+
+/**
+ * Constructor: stores all the parameters in an object
+ * of feature_parameters class.
+ */
+template<typename MatType, typename CubeType>
+StructuredForests<MatType, CubeType>::
+StructuredForests(FeatureParameters F)
```
stereomatchingkiss: Although this is not std::map, but using initialization list to initialize still a good practice

This is initialize

`StructuredForests(FeatureParameters F)  : params(F) {}`

This is assignment

```
{
  params = F;
}
```

src/mlpack/methods/edge_boxes/feature_extraction_impl.hpp
```diff
@@ -0,0 +1,1039 @@
...
+  const size_t shrink = this->params.Shrink();
+  const size_t pSize = this->params.PSize();
+  const size_t numCell = this->params.NumCell();
+  const size_t numOrient = this->params.NumOrient();
+  
+  size_t nColorCh;
+  if (this->params.RGBD() == 0)
+    nColorCh = 3;
+  else
+    nColorCh = 4;
```
stereomatchingkiss: Another solution(a problem of taste, pick the one you like more :) ) : 
`const size_t nColorCh = params.RGBD() == 0 ? 3 : 4;`

src/mlpack/methods/edge_boxes/feature_extraction.hpp
```diff
@@ -0,0 +1,102 @@
...
+
+  void Gradient(const CubeType& InImage, 
+         MatType& Magnitude,
+         MatType& Orientation);
+
+  void MaxAndLoc(CubeType &mag, arma::umat &Location, MatType& MaxVal) const;
+
+  void Histogram(const MatType& Magnitude,
+            const MatType& Orientation, 
+            size_t downscale, size_t interp,
```
stereomatchingkiss: Do you forgot const?const size_t downscale, const size_t interp?

src/mlpack/methods/edge_boxes/feature_extraction.hpp
```diff
@@ -0,0 +1,102 @@
...
+                 CubeType& OutImage);
+  
+  void GetShrunkChannels(const CubeType& InImage, CubeType &reg_ch,\
+                  CubeType &ss_ch, const arma::vec& table);
+  
+  void RGB2LUV(const CubeType& InImage, CubeType& OutImage,\
+                   const arma::vec& table);
+  
+  void BilinearInterpolation(const MatType& src,
+                          size_t height, size_t width,
```
stereomatchingkiss: Do you forgot const?const size_t height, const size_t width?

src/mlpack/methods/edge_boxes/feature_extraction_impl.hpp
```diff
@@ -0,0 +1,1052 @@
...
+ * Constructor: stores all the parameters in an object
+ * of feature_parameters class.
+ */
+template<typename MatType, typename CubeType>
+StructuredForests<MatType, CubeType>::
+StructuredForests(FeatureParameters F) : params(F) {}
+
+/*
+template<typename MatType, typename CubeType>
+MatType StructuredForests<MatType, CubeType>::
```
stereomatchingkiss: I guess we can remove useless codes, let git help us memorize the codes we written

src/mlpack/methods/edge_boxes/feature_extraction_impl.hpp
```diff
@@ -0,0 +1,1052 @@
...
+/**
+ * Computes distance transform of 1D vector f.
+ * @param f input vector whose distance transform is to be found.
+ * @param n size of the Output vector to be made.
+ * @param inf a large double value.
+ * @param d Output vector which stores distance transform of f.
+ */
+template<typename MatType, typename CubeType>
+void StructuredForests<MatType, CubeType>::
+DistanceTransform1D(const arma::vec& f, const size_t n, const double inf,
```
stereomatchingkiss: May I ask what is the algorithm you are using? Thanks

src/mlpack/methods/edge_boxes/feature_extraction_impl.hpp
```diff
@@ -0,0 +1,1052 @@
...
+RGB2LUV(const CubeType& InImage, CubeType& OutImage,\
+        const arma::vec& table)
+{
+  //assert type is double or float.
+
+
+  MatType rgb2xyz;
+  rgb2xyz << 0.430574 << 0.222015 << 0.020183 << arma::endr
+          << 0.341550 << 0.706655 << 0.129553 << arma::endr
+          << 0.178325 << 0.071330 << 0.939180;
```
stereomatchingkiss: Do not know what is this for?I guess you want to do vectorization, right?

src/mlpack/methods/edge_boxes/feature_extraction_impl.hpp
```diff
@@ -0,0 +1,1052 @@
...
+  for (size_t i = 0; i < InImage.slice(0).n_elem; ++i)
+  {
+    double r = InImage.slice(0)(i);
+    double g = InImage.slice(1)(i);
+    double b = InImage.slice(2)(i);
+    
+    xyz.slice(0)(i) = 0.430574 * r + 0.341550 * g + 0.178325 * b;
+    xyz.slice(1)(i) = 0.222015 * r + 0.706655 * g + 0.071330 * b;
+    xyz.slice(2)(i) = 0.020183 * r + 0.129553 * g + 0.939180 * b;
+  }
```
stereomatchingkiss: Why not just
```
xyz.slice(0) = 0.430574 * InImage.slice(0) + 0.341550 * InImage.slice(1) + 0.178325 * InImage.slice(2);
xyz.slice(1) = 0.222015 * InImage.slice(0) + 0.706655 * InImage.slice(1) + 0.071330 * InImage.slice(2);
xyz.slice(2) = 0.020183 * InImage.slice(0) + 0.129553 * InImage.slice(1) + 0.939180 * InImage.slice(2);
```

src/mlpack/methods/edge_boxes/feature_extraction_impl.hpp
```diff
@@ -0,0 +1,1052 @@
...
+ * @param radius Decides the size of kernel to be applied on Image.
+ */
+template<typename MatType, typename CubeType>
+void StructuredForests<MatType, CubeType>::
+ConvTriangle(CubeType &InImage, const size_t radius)
+{
+  if (radius == 0)
+  {
+    //nothing to do
+  }
```
stereomatchingkiss: A little bit weird to see empty if, since radius is size_t, you can
```
if(radius != 0){
  if(radius == 1){
  //.....
  }else{
  //....
 }
}
```

src/mlpack/methods/edge_boxes/feature_extraction_impl.hpp
```diff
@@ -0,0 +1,1052 @@
...
+        HistArr(r, c, o1) += Magnitude(i, j);
+      }
+    }
+  }
+
+  HistArr = HistArr / downscale;
+
+  for (size_t i = 0; i < HistArr.n_slices; ++i)
+    HistArr.slice(i) = arma::square(HistArr.slice(i));
+}
```
stereomatchingkiss: arma::square works for Cube<T> too, following should work

`HistArr = arma::square(HistArr);`

src/mlpack/methods/edge_boxes/feature_extraction_impl.hpp
```diff
@@ -0,0 +1,1052 @@
...
+
+    // do same for dy.
+    dy.slice(i).row(0) = InImage.slice(i).row(1) - InImage.slice(i).row(0);
+    dy.slice(i).row(InImage.n_rows - 1) = InImage.slice(i).row(InImage.n_rows - 1)
+                                        - InImage.slice(i).row(InImage.n_rows - 2);
+
+    for (size_t j = 1; j < InImage.n_rows-1; j++)
+      dy.slice(i).row(j) = 0.5 * ( InImage.slice(i).row(j+1) - InImage.slice(i).row(j) );
+  }  
+
```
stereomatchingkiss: I think by usual, we would apply convolution on the matrix directly, cnn of mlpack implement convolution algo, maybe you can use it to find out the gradient.

zoq: Hello @stereomatchingkiss, to save you some time, I made a bunch of similar comments here: https://github.com/mlpack/mlpack/pull/696

src/mlpack/methods/edge_boxes/feature_extraction_impl.hpp
```diff
@@ -0,0 +1,1052 @@
...
+  MaxVal = MatType(Location.n_rows, Location.n_cols);
+  for(size_t i = 0; i < mag.n_rows; ++i)
+  {
+    for(size_t j = 0; j < mag.n_cols; ++j)
+    {
+      /*can use -infinity here*/
+      double max =  std::numeric_limits<double>::min();
+      for(size_t k = 0; k < mag.n_slices; ++k)
+      {
+        if(mag(i, j, k) > max)
```
zoq: I figured out why it doesn't work on my machine. The check: 

``if(mag(i, j, k) > max)``

doesn't work if mag is [0 0 0] because 0 < double max =  std::numeric_limits<double>::min();.  Since the the location is undefined, we could probably run into an index out of bounds error.

As pointed out here: https://github.com/mlpack/mlpack/pull/696#discussion_r67073948 we should use -DBL_MAX or just use the first value of mag as starting point.

src/mlpack/methods/edge_boxes/feature_extraction_impl.hpp
nilayjain: I am not sure what you mean by this here. We are  populating the rgb2xyz matrix with the corresponding values.

zoq: I think what @stereomatchingkiss meant was, that ``rgb2xyz`` is unused, so we could delete the parameter.

src/mlpack/methods/edge_boxes/feature_extraction_impl.hpp
```diff
@@ -0,0 +1,1052 @@
...
+  size_t ind = 0;
+  arma::uvec k = arma::find(zs > 0);
+  if ( k.n_elem == 0)
+  {
+    DiscreteLabels.ones();
+  }
+  else
+  {
+    //find most representative label (closest to mean)
+    arma::vec k = arma::sum(arma::abs(zs), 0);
```
zoq: As pointed out here: http://arma.sourceforge.net/docs.html#sum ``arma::sum(..., 0)`` returns a ``rowvec``, arma::vec is a ``colvec``.

src/mlpack/methods/edge_boxes/feature_extraction.hpp
```diff
@@ -0,0 +1,102 @@
...
+                 CubeType& HistArr);
+ 
+  void ViewAsWindows(const CubeType& channels, const arma::umat& loc,
+                     CubeType& features);
+
+  void GetRegFtr(const CubeType& channels, const arma::umat& loc,
+                 CubeType& RegFtr);
+
+  void GetSSFtr(const CubeType& channels, const arma::umat& loc,
+                 CubeType SSFtr);
```
zoq: I was wondering, why I don't see any similarity features, turns out we missed a reference here.

src/mlpack/methods/edge_boxes/feature_extraction.hpp
```diff
@@ -19,53 +19,133 @@ class StructuredForests
...
 
- 
+  /**
+   * Constructor: stores all the parameters in an object
+   * of feature_parameters class.
+   * @param F FeatureParameters object which stores necessary parameters.
+   */ 
   StructuredForests(FeatureParameters F);
 
   void PrepareData(const MatType& Images, const MatType& Boundaries,\
```
stereomatchingkiss: Could you remove the "\"?

nilayjain:
do you think i should rebase/upstream and make these changes in a new PR ? We seem to use test_tools.hpp instead of old_boost_test_definitions.hpp .

zoq:
I think there is no need to open a new PR, you could update your branch instead. What do you think does this sounds reasonable?

src/mlpack/methods/edge_boxes/feature_extraction_impl.hpp
```diff
@@ -0,0 +1,916 @@
...
+  const size_t nColorCh = params.RGBD() == 0 ? 3 : 4;
+  const size_t nCh = nColorCh + 2 * (1 + numOrient);
+  FtrDim[0] = std::pow((pSize / shrink) , 2) * nCh;
+  FtrDim[1] = std::pow(numCell , 2) * (std::pow (numCell, 2) - 1) / 2 * nCh;
+}
+
+
+template<typename MatType, typename CubeType>
+void StructuredForests<MatType, CubeType>::
+DistanceTransform1D(const arma::vec& f, const size_t n, const double inf,
```
zoq: Is this function from the "Distance Transforms of Sampled Functions" paper? If that's the case, we should cite the paper, as we did here: https://github.com/mlpack/mlpack/blob/master/src/mlpack/methods/ann/layer/dropout_layer.hpp. Also do you think, we could define inf inside the function and probably use DBL_MAX and -DBL_MAX instead of 999999.99?

src/mlpack/methods/edge_boxes/feature_extraction.hpp
```diff
@@ -0,0 +1,262 @@
...
+                            MatType& dst);
+  
+  /**
+   * Applies a separable linear filter to an Image
+   * @param InOutImage Input/Output Contains the input Image, The final filtered Image is
+   *          stored in this param.
+   * @param kernel Input Kernel vector to be applied on Image.
+   * @param radius amount, the Image should be padded before applying filter.
+   */  
+  void Convolution(CubeType &InOutImage, const MatType& Filter, const size_t radius);
```
zoq: Another picky comment, please use prefix const, also I know it's a matter of style, but I prefer to use type& parameter instead of type &parameter. For more details: https://github.com/mlpack/mlpack/pull/523/files/c92dd2fc6561b1cf40efb744dcc9233bc3675c91#r60094845

src/mlpack/methods/edge_boxes/feature_extraction_impl.hpp
```diff
@@ -0,0 +1,916 @@
...
+  size_t k = 0;
+  v[0] = 0.0;
+  z[0] = -inf;
+  z[1] = inf;
+  for (size_t q = 1; q <= n - 1; ++q)
+  {
+    float s  = ( (f[q] + q * q)-( f[v[k]] + v[k] * v[k]) ) / (2 * q - 2 * v[k]);
+    while (s <= z[k])
+    {
+      --k;
```
zoq: I'm not sure, but if we set k = 0 at the beginning, how can we decrease k (``size_t``), and still get the right index?

src/mlpack/methods/edge_boxes/feature_extraction_impl.hpp
```diff
@@ -0,0 +1,916 @@
...
+    Im.col(x) = d;
+  }
+
+  // transform along rows
+  for (size_t y = 0; y < Im.n_rows; y++)
+  {
+    f.subvec(0, Im.n_cols - 1) = Im.row(y).t();
+    arma::vec d;
+    this->DistanceTransform1D(f, Im.n_cols, inf, d);
+    Im.row(y) = d.t();
```
zoq: It looks like you could pass im.row(y) instead of passing a copy of the transposed row. In this case you have to adjust the DistanceTransform1D function header:

```
 arma::vec DistanceTransform1D(arma::row& f, const size_t n, ...) 
```

src/mlpack/tests/edge_boxes_test.cpp
```diff
@@ -0,0 +1,206 @@
...
+
+/**
+ * This test checks the feature extraction functions 
+ * mentioned in feature_extraction.hpp
+ */
+
+void Test(arma::mat m1, arma::mat m2)
+{
+  for (size_t i = 0; i < m1.n_cols; ++i)
+    BOOST_REQUIRE_CLOSE(m1(i), m2(i), 1e-2);
```
zoq: That only compares the first (n_cols) elements. I think you like to use n_elem. Also you could use a template parameter for the mat type and just write a single function for cube and mat.

src/mlpack/tests/edge_boxes_test.cpp
```diff
@@ -0,0 +1,206 @@
...
+}
+
+void ConvTriangleTest(arma::cube& input, int radius,
+    arma::cube& output, StructuredForests<arma::mat, arma::cube>& SF)
+{
+  SF.ConvTriangle(input, radius);
+  Test(input, output);
+}
+
+BOOST_AUTO_TEST_CASE(FeatureExtractionTest)
```
zoq: Can you split the FeatureExtractionTest test case into smaller test cases? Something like:

```
BOOST_AUTO_TEST_CASE(DistanceTransformTest)
BOOST_AUTO_TEST_CASE(CopyMakeBorderTest)
```

src/mlpack/methods/edge_boxes/feature_extraction_impl.hpp
```diff
@@ -0,0 +1,916 @@
...
+    
+    CubeType Hist;
+    this->Histogram(Magnitude, Orientation,
+                    downscale, 0, Hist);
+    BilinearInterpolation( Magnitude, rsize, csize, channels.slice(slice_idx));
+    slice_idx++;
+    for(size_t i = 0; i < InImage.n_slices; ++i)
+      BilinearInterpolation( Magnitude, rsize, csize,
+                      channels.slice(i + slice_idx));    
+    slice_idx += 3;
```
zoq: I would recommend to use slice_idx += n_slicesinstead of slice_idx += 3; there is no guarantee that we call the function with a 3 channel image.

src/mlpack/methods/edge_boxes/feature_extraction_impl.hpp
```diff
@@ -0,0 +1,916 @@
...
+                                channels.slice(slice_idx));
+  double scale = 1.0;
+  
+  while(scale >= 0.5)
+  {
+    CubeType Img( (luv.n_rows * scale),
+                   (luv.n_cols * scale),
+                   luv.n_slices );
+
+    for( slice_idx = 0; slice_idx < luv.n_slices; ++slice_idx)
```
zoq: This overwrites the ``slice_idx`` parameter, we use below.

src/mlpack/methods/edge_boxes/feature_extraction_impl.hpp
```diff
@@ -0,0 +1,916 @@
...
+      // we take a random permutation of posLoc and negLoc.
+      posLoc = arma::shuffle(posLoc);
+      negLoc = arma::shuffle(negLoc);
+
+      size_t lenLoc = std::min((int) negLoc.n_elem, std::min((int) nPatchesPerGt,
+                              (int) posLoc.n_elem));
+      arma::umat loc(lenLoc * 2, 2);
+
+      for(size_t i = 0; i < lenLoc; ++i)
+        loc.row(i) = arma::ind2sub(arma::size(dis.n_rows, dis.n_cols), posLoc(i) ).t();
```
zoq: Do you think, we could store the location in the columns, in this case we don't have to transpose the locations vector.

src/mlpack/methods/edge_boxes/feature_extraction.hpp
```diff
@@ -0,0 +1,262 @@
...
+ public:
+
+  /**
+   * Constructor: stores all the parameters in an object
+   * of feature_parameters class.
+   * @param F FeatureParameters object which stores necessary parameters.
+   */ 
+  StructuredForests(FeatureParameters F);
+
+  void PrepareData(const MatType& Images, const MatType& Boundaries,
```
zoq: Can you also comment the ``PrepareData``function?

src/mlpack/methods/edge_boxes/feature_extraction.hpp
```diff
@@ -0,0 +1,262 @@
...
+   */
+  void DistanceTransform1D(const arma::vec& f, const size_t n,
+                           const double inf, arma::vec& d);
+  
+  /**
+   * Computes distance transform of a 2D array
+   * @param Im input array whose distance transform is to be found.
+   * @param inf a large double value.
+   */
+  void DistanceTransform2D(MatType &Im, const double inf);
```
zoq: Note that Variable, reference, and pointer names begin with a lower case letter. Take a look at: https://github.com/mlpack/mlpack/wiki/DesignGuidelines#naming-conventions for more informations.

src/mlpack/methods/edge_boxes/feature_extraction.hpp
```diff
@@ -0,0 +1,262 @@
...
+
+  /**
+   * euclidean distance transform of binary Image using squared distance
+   * This is the discription of the paper which discribes the approach 
+   * for this algorithm : Distance Transforms of Sampled Functions,
+   * P. Felzenszwalb, D. Huttenlocher
+   * Theory of Computing, Vol. 8, No. 19, September 2012
```
zoq: Can you use the doxygen style to cite the paper. Here is an example: https://github.com/mlpack/mlpack/blob/master/src/mlpack/methods/ann/layer/dropout_layer.hpp

src/mlpack/methods/edge_boxes/feature_extraction.hpp
```diff
@@ -0,0 +1,262 @@
...
+   * Refer to the following paper for details:
+   * Fast edge detection using structured forests
+   * Authors: Piotr Dollar, Larry Zitnick
+   * Published In: ICCV
+   * @param InImage Input Image.
+   * @param regCh Channels used in calculating Regular Features.
+   * @param ssCh Channels used in calculating Self Similarity Features.
+   * @param table a vector which helps in converting image from RGB to LUV.
+   */
+  void GetShrunkChannels(const CubeType& InImage, CubeType &reg_ch,
```
zoq: Can you use the doxygen style to cite the paper. Here is an example: https://github.com/mlpack/mlpack/blob/master/src/mlpack/methods/ann/layer/dropout_layer.hpp

src/mlpack/methods/edge_boxes/feature_extraction.hpp
```diff
@@ -0,0 +1,262 @@
...
+   * Sketch Tokens: A Learned Mid-level Representation for Contour 
+   * and Object Detection.
+   * Published in: Computer Vision and Pattern Recognition (CVPR),
+   * 2013 IEEE Conference 
+   * Author(s) : Joseph J. Lim ; C. Lawrence Zitnick ; Piotr Dollár
+   * @param channels input candidate regular features
+   * @param loc input locations
+   * @param RegFtr output Regular Features at that location.
+   */
+  void GetRegFtr(const CubeType& channels, const arma::umat& loc,
```
zoq: Can you use the doxygen style to cite the paper. Here is an example: https://github.com/mlpack/mlpack/blob/master/src/mlpack/methods/ann/layer/dropout_layer.hpp

src/mlpack/methods/edge_boxes/feature_extraction.hpp
```diff
@@ -0,0 +1,262 @@
...
+  /**
+   * Compute the pairwise differences between n-dimensional points in a way
+   * specified in the paper "Structured Forests for Fast Edge Detection".
+   * Note: Indeed this is not a valid distance measurement (asymmetry).
+   * find nC2 differences, for locations in the grid_pos.
+   * @param features n-dimensional points
+   * @param grid_pos locations
+   * @param Output stores differences between features at grid_pos locations
+   */
+  void PDist(const CubeType& features, const arma::uvec& grid_pos,
```
zoq: Please do not use an underscore for parameter names.

src/mlpack/methods/edge_boxes/feature_extraction.hpp
```diff
@@ -0,0 +1,262 @@
...
+             CubeType& Output);
+
+  /**
+   * Find the index of minimum element in a vector
+   * @param k Vector in which we want to find element
+   * return value index of minimum element
+   */
+  size_t IndexMin(arma::vec& k);
+
+  size_t Discretize(const MatType& labels, const size_t nClass,
```
zoq: Can you add a comment that describes what this function does.

src/mlpack/methods/edge_boxes/feature_extraction_impl.hpp
```diff
@@ -0,0 +1,916 @@
...
+    features.slices(channel, channel + 12) = patch;
+    channel += 13;
+  }
+}
+
+template<typename MatType, typename CubeType>
+void StructuredForests<MatType, CubeType>::
+Rearrange(const CubeType& channels, CubeType& ch)
+{
+  //we do (16,16,13*1000) to 256, 1000, 13, in vectorized code.
```
zoq: Can you explain why we use 256, I think we could also use a value > 256.

src/mlpack/methods/edge_boxes/feature_extraction_impl.hpp
```diff
@@ -0,0 +1,916 @@
...
+  this->ViewAsWindows(channels, loc, wind);
+  this->Rearrange(wind, RegFtr);
+}
+
+template<typename MatType, typename CubeType>
+void StructuredForests<MatType, CubeType>::
+PDist(const CubeType& features, const arma::uvec& grid_pos,
+      CubeType& Output)
+{
+  // size of DestArr:
```
zoq: I guess, we can remove the debug output here.

src/mlpack/methods/edge_boxes/feature_extraction_impl.hpp
```diff
@@ -0,0 +1,916 @@
...
+  CubeType ss_ch = CubeType(rsize, csize, numChannels);
+
+  this->GetShrunkChannels(InImage, reg_ch, ss_ch, table);
+
+  loc /= shrink;
+  this->GetRegFtr(reg_ch, loc, RegFtr);
+  this->GetSSFtr(ss_ch, loc, SSFtr);
+}
+
+/**
```
zoq: I hink we can remove the comment here.

src/mlpack/methods/edge_boxes/feature_parameters.hpp
```diff
@@ -0,0 +1,128 @@
...
+//! This class holds all the fields for the FeatureExtraction class.
+class FeatureParameters
+{
+ public:
+
+  //! getter and setter methods for all the fields in class.
+  void NumImages(size_t value) { numImages = value; }
+  size_t NumImages() const { return numImages; }
+
+  void RowSize(size_t value) { rowSize = value; }
```
zoq: Can you add a comment for each function? You could use the comments from ``edge_boxes_main.cpp``.

src/mlpack/tests/edge_boxes_test.cpp
```diff
@@ -0,0 +1,206 @@
...
+
+#include <boost/test/unit_test.hpp>
+#include "old_boost_test_definitions.hpp"
+
+using namespace mlpack;
+using namespace mlpack::structured_tree;
+/*
+  
+ 
+  //void GetShrunkChannels(CubeType& InImage, CubeType& reg_ch, CubeType& ss_ch);
```
zoq: Looks like we can remove all comments here.

src/mlpack/tests/edge_boxes_test.cpp
```diff
@@ -0,0 +1,206 @@
...
+             << 1 << 1 << 0 << 1 << 1 << arma::endr
+             << 0 << 0 << 1 << 2 << 2 << arma::endr
+             << 0 << 0 << 1 << 2 << 2;
+  arma::cube out_b(out_border.n_rows, out_border.n_cols, 1);
+  out_b.slice(0) = out_border;
+  CopyMakeBorderTest(in1, out_b, SF);
+
+  arma::mat out_conv;
+
+  out_conv << 1.20987 << 1.25925 << 1.30864 << arma::endr
```
zoq: Where do our values come from?

rcurtin:
Closing for inactivity.

PR #708, opened 2016-06-27
Vantage point tree

lozhnikov:
I implemented the vantage point tree. I didn't think about tests yet. Right now the tree passes TreeTest/VantagePointTreeTest (the test is similar to the ball tree test). BallBound is not suitable for the vantage point tree. I have to fix that or implement another bound type. I am not sure that I understand the paper correctly. Especially I am not sure in the median calculation method. And I am not sure that there is no confusion of the second moment and variance in the article.


rcurtin:
I will read through the paper tomorrow and look through the code and let you know what I think. I agree that the existing BallBound is unsuitable for the VP tree.

src/mlpack/core/tree/binary_space_tree/vantage_point_split_impl.hpp
```diff
@@ -0,0 +1,289 @@
...
+}
+
+template<typename BoundType, typename MatType>
+void VantagePointSplit<BoundType, MatType>::
+GetDistinctSamples(arma::uvec& distinctSamples, const size_t numSamples,
+    const size_t begin, const size_t upperBound)
+{
+  if (upperBound > numSamples)
+  {
+    arma::Col<size_t> samples;
```
lozhnikov: I think it is better to move all temporal vectors and matrices outside (i.e. define as class members). That allows to avoid unnecessary memory allocations.

lozhnikov:
I moved the vantage point tree into a separate class and implemented `HollowBallBound`. Right now, the vantage point tree passes all tests in `src/mlpack/tests/vantage_point_tree_test.cpp` if `TreeTraits::FirstPointIsCentroid = false`.

If `TreeTraits::FirstPointIsCentroid = true` the tree don't pass VantagePointTreeTest/DualTreeTraverserTest. It seems that there is an error in `VantagePointTree::DualTreeTraverser` but I could not find that. Moreover, I am not sure that the first point is the centroid since the first point of the parent node is the centroid of its children.

rcurtin:
`FirstPointIsCentroid` is maybe a bit misleading of a name. When I wrote that originally I intended that `FirstPointIsCentroid` would be true if the first point in the node was the centroid of the region represented by the bound, not the centroid of the descendant points held in the node. So for ball bounds this is the center of the bound. But maybe a better term is `FirstPointIsCenter`. Either way for the VP tree it should be true. I'll look through the code and see if I can figure out why the traverser is failing.

rcurtin:
Ah hang on, what I wrote is not quite right. The left node holds the centroid point but the right node does not. So we  can leave `FirstPointIsCentroid` set to false and get a little bit of calculation duplication or we can try to do something crazy and set `FirstPointIsCentroid` to `true` for the left children and `false` for the right children.

But since `FirstPointIsCentroid` is a static compile time constant, we would need the types of the left and right children to be different. Let me think about how it might be possible to get the desired behavior...

rcurtin:
I spent a long time thinking about it.  First, I thought, how can we make right children have `FirstPointIsCentroid = false` and left children have `FirstPointIsCentroid = true` but have both children be of the same type?  But I don't think that is possible.

The second option is to have left and right children have different type.  This could work because all of the classes only hold pointers to the root of the tree, which in this case would have the type of the left children.  But there would need to be some refactoring of the rules, so that the `Score()` and `Rescore()` methods could take two different types of trees, and so that the `TreeType` template parameter was at the method level, not the class level.  I think that is possible and not too difficult, though there might be a gotcha or two with the `TraversalInfoType`.  Whatever we changed, we would have to document in the TreeType API page.

The third option is to find a way so that `FirstPointIsCentroid` is a static compile-time constant for some trees, but is a runtime constant for others.  I have not thought about this much so I am not sure if it is completely feasible, but I think it is possible.

The fourth option, from earlier, is to just leave `FirstPointIsCentroid = false` for vantage point trees.

What do you think is the best approach here?

lozhnikov:
I think the second option will lead to enormous duplication of the code. Right now, `VantagePointTree` has almost the same code as `BinarySpaceTree`. And if I am not mistaken we have to add the third similar class in order to implement that.

I guess it is better to add a method in each `TreeType`: `IsFirstPointCentroid()`. That should not lead to any slowdown since this check is used rarely. (There are two options: we can implement this method in the class definition, so the method will be inlined or we can add two tree traits `AllFirstPointsAreCentroids` and `SomeFirstPointsAreCentroids` and invoke this method when it is needed. I think the first option is better.)

Maybe the fourth option is better. We can try to implement the third approach and test performance in order to choose between the third and the fourth options.

So, I guess we should choose either the third or the fourth option.

rcurtin:
There shouldn't be too much duplication in the second option.  We would say that the left child would have type `VantagePointTree<BallBound>` and the right child would have type `VantagePointTree<HollowBallBound>`, then we could just specialize `TreeTraits` for each of those classes with different values for the left and right children.  But, the real difficulty there, which I did not think about earlier, is that what is returned when I call `Child(i)` is unclear, and there is no way to preserve the type fully.  I can think of some ways, maybe, but it would be a bit backhanded.

The checks for `FirstPointIsCentroid` are actually used pretty often; at least once in every `Score()` call.  So it is of some importance that those remain static compile-time constants.  If we were to make those supported as runtime checks in some cases, we would need to

 * change the boolean constants to constexpr functions that take TreeTypes as a parameter
 * for VP-trees, the function can't be constexpr, but instead looks at whether the TreeType parameter is true or false

We already have the fourth option implemented, so it is easy to test. :)

lozhnikov:
I tested the traversers on a number of datasets and I did not find errors. But I am still not sure that it works. It seems I do not understand something. If I am not mistaken I have to avoid base cases that are already calculated. But `RangeSearchRules::Score()` each time calculates a base case and `NeighborSearchRules::Score(queryIndex, referenceNode)` does not calculate a base case. Right now, the traversers calculate each base case that is not pruned. But it seems it is wrong.

src/mlpack/core/tree/hollow_ball_bound.hpp
```diff
@@ -0,0 +1,222 @@
...
+  //! The underlying data type.
+  typedef typename VecType::elem_type ElemType;
+  //! A public version of the vector type.
+  typedef VecType Vec;
+
+ private:
+  //! The radius of the inner ball bound.
+  ElemType innerRadius;
+  //! The radius of the outer ball bound.
+  ElemType outerRadius;
```
rcurtin: It might be reasonable to use `math::Range` here.

src/mlpack/core/tree/hollow_ball_bound.hpp
```diff
@@ -0,0 +1,222 @@
...
+   *
+   * @param dimension Dimensionality of ball bound.
+   */
+  HollowBallBound(const size_t dimension);
+
+  /**
+   * Create the ball bound with the specified radius and center.
+   *
+   * @param innerRradius Inner radius of ball bound.
+   * @param outerRradius Outer radius of ball bound.
```
rcurtin: Very minor typos in the parameter names here.

src/mlpack/core/tree/hollow_ball_bound_impl.hpp
```diff
@@ -0,0 +1,384 @@
+/**
+ * @file hollow_ball_bound_impl.hpp
+ *
+ * Bounds that are useful for binary space partitioning trees.
+ * Implementation of HollowBallBound ball bound metric policy class.
+ *
+ * @experimental
```
rcurtin: The @experimental mark doesn't need to be here... I think hrectbound_impl.hpp has it too, it's probably not needed there either.

src/mlpack/core/tree/rectangle_tree/rectangle_tree.hpp
```diff
@@ -492,6 +492,10 @@ class RectangleTree
   //! Returns false: this tree type does not have self children.
   static bool HasSelfChildren() { return false; }
 
+  //! Returns false: The first point of this node is not the centroid
+  //! of its bound.
+  static constexpr bool IsFirstPointCentroid() { return false; }
```
rcurtin: The idea I had here for `IsFirstPointCentroid()` was to leave it a part of `TreeTraits`, but just to make it a function:

```
template<typename TreeType>
struct TreeTraits
{
  static constexpr bool FirstPointIsCentroid(TreeType*) { return false; }
  ...
}
```

and then for the VP tree we can specialize and make it non-constexpr:

```
template<>
struct TreeTraits<VPTree>
{
  static bool FirstPointIsCentroid(VPTree* v) { ... }
}
```

It changes the syntax of all of the traits, but it keeps them out of the definition of each tree itself, and we are guaranteed that there is a value for these even if the person who wrote the tree did not specify them.  Does that seem reasonable to you?

src/mlpack/core/tree/vantage_point_tree/dual_tree_traverser_impl.hpp
```diff
@@ -0,0 +1,567 @@
+/**
+ * @file dual_tree_traverser_impl.hpp
```
rcurtin: Is there any difference between this traverser and the BinarySpaceTree traverser?  If not, maybe we could make that traverser generic (accepting any tree type) and use it inside each class.  The same applies for the single-tree traverser.

src/mlpack/core/tree/vantage_point_tree/typedef.hpp
```diff
@@ -0,0 +1,26 @@
...
+ */
+#ifndef MLPACK_CORE_TREE_VANTAGE_POINT_TREE_TYPEDEF_HPP
+#define MLPACK_CORE_TREE_VANTAGE_POINT_TREE_TYPEDEF_HPP
+
+// In case it hasn't been included yet.
+#include "../vantage_point_tree.hpp"
+
+namespace mlpack {
+namespace tree {
+
```
rcurtin: We should add a quick comment here on what the tree is, so that users will find it in the documentation: http://mlpack.org/docs/mlpack-git/doxygen.php?doc=namespacemlpack_1_1tree.html#typedef-members (I wish that page was less ugly but doxygen is not the best).

src/mlpack/core/tree/vantage_point_tree/vantage_point_split.hpp
```diff
@@ -0,0 +1,197 @@
...
+                        const size_t begin,
+                        const size_t count,
+                        size_t& splitCol,
+                        std::vector<size_t>& oldFromNew);
+ private:
+  /**
+   * The maximum number of samples used for vantage point estimation and for
+   * estimation of the median.
+   */
+  static const size_t maxNumSamples = 100;
```
rcurtin: It might make sense to have this be a template parameter defaulting to 100, but I don't have a strong opinion either way.

src/mlpack/core/tree/vantage_point_tree/vantage_point_split.hpp
```diff
@@ -0,0 +1,197 @@
...
+    size_t point;
+    ElemType dist;
+  };
+
+  template<typename StructElemType>
+  static bool StructComp(const SortStruct<StructElemType>& s1,
+      const SortStruct<StructElemType>& s2)
+  {
+    return (s1.dist < s2.dist);
+  };
```
rcurtin: Should we use `std::pair<size_t, ElemType>` here too, like in the RectangleTree code?

src/mlpack/core/tree/vantage_point_tree/vantage_point_split_impl.hpp
```diff
@@ -0,0 +1,298 @@
...
+
+namespace mlpack {
+namespace tree {
+
+template<typename BoundType, typename MatType>
+bool VantagePointSplit<BoundType, MatType>::
+SplitNode(const BoundType& bound, MatType& data, const size_t begin,
+    const size_t count, size_t& splitCol)
+{
+  typename BoundType::ElemType mu;
```
rcurtin: Couldn't you just use `ElemType` here?  We should be guaranteed that `BoundType::ElemType == MatType::elem_type`.

src/mlpack/core/tree/vantage_point_tree/vantage_point_split.hpp
```diff
@@ -0,0 +1,197 @@
...
+   * random points that belong to the node. Each sample belongs to
+   * the interval [begin, begin + upperBound)
+   *
+   * @param distinctSamples The vector of samples indices.
+   * @param numSamples Maximum number of samples.
+   * @param begin The least index.
+   * @param upperBound The upper bound of indices.
+   */
+  static void GetDistinctSamples(arma::uvec& distinctSamples,
+      const size_t numSamples, const size_t begin, const size_t upperBound);
```
rcurtin: This has a lot of similarities to `ObtainDistinctSamples()` from `mlpack/methods/rann/ra_util.hpp`.  Do you think we should take that function, combine it with this one (I think your version is more flexible), put it in `core/math/`, and refactor both pieces to use that code?

src/mlpack/core/tree/vantage_point_tree/vantage_point_split.hpp
```diff
@@ -0,0 +1,197 @@
...
+   * number of samples.
+   *
+   * @param bound The bound used for this node.
+   * @param data The dataset used by the binary space tree.
+   * @param samples The indices of random samples.
+   * @param vantagePoint The vantage point.
+   * @param mu The median value.
+   */
+  static void GetMedian(const BoundType& bound, const MatType& data,
+      const arma::uvec& samples,  const size_t vantagePoint, ElemType& mu);
```
rcurtin: I think here and in `GetSecondMoment()`, it would be better to pass a `MetricType` instead of a `BoundType`.

src/mlpack/core/tree/vantage_point_tree/vantage_point_split_impl.hpp
```diff
@@ -0,0 +1,298 @@
...
+    sorted[i].dist = bound.Metric().Evaluate(data.col(vantagePoint),
+        data.col(samples[i]));
+  }
+
+  //  Sort samples according to the distance to the vantage point
+  std::sort(sorted.begin(), sorted.end(), StructComp<ElemType>);
+
+  //  Get the midian value
+  mu = bound.Metric().Evaluate(data.col(vantagePoint),
+      data.col(sorted[sorted.size() / 2].point));
```
rcurtin: Could you instead use Armadillo's `median()` function here?  You don't need the index of the median, just the median value itself, so I think Armadillo could work fine, and I think it would be less code (and possibly faster too).

src/mlpack/core/tree/vantage_point_tree/vantage_point_split_impl.hpp
```diff
@@ -0,0 +1,298 @@
...
+    const ElemType spread = GetSecondMoment(bound, data, samples,
+        vantagePointCandidates[i]);
+
+    if (spread > bestSpread)
+    {
+      bestSpread = spread;
+      vantagePoint = vantagePointCandidates[i];
+      //  Calculate the median value of the distance from the vantage point candidate
+      //  to these samples
+      GetMedian(bound, data, samples, vantagePoint, mu);
```
rcurtin: This call to `GetMedian()` will duplicate all the distance calculations that are done in `GetSecondMoment()`.  To be honest, since `GetSecondMoment()` and `GetMedian()` are only called once, I would just inline them into this function, which will allow you to more easily avoid the duplicate calculations.  Note also that if you already have an Armadillo object with the distances of each point to the vantage point, you can just call `arma::sum(distances % distances)` to get the sum of squared distances.

src/mlpack/core/tree/vantage_point_tree/vantage_point_split_impl.hpp
```diff
@@ -0,0 +1,298 @@
...
+
+  // Get no more than max(maxNumSamples, count) vantage point candidates
+  GetDistinctSamples(vantagePointCandidates, maxNumSamples, begin, count);
+
+  ElemType bestSpread = 0;
+
+  //  Evaluate eache candidate
+  for (size_t i = 0; i < vantagePointCandidates.n_rows; i++)
+  {
+    arma::uvec samples;
```
rcurtin: (I am not sure if compiler optimizations will take care of this.)  If you declare this outside the loop, then every time you call `GetDistinctSamples()`, memory won't need to be allocated.

src/mlpack/core/tree/vantage_point_tree/vantage_point_split_impl.hpp
```diff
@@ -0,0 +1,298 @@
...
+  if (upperBound > numSamples)
+  {
+    arma::Col<size_t> samples;
+
+    samples.zeros(upperBound);
+
+    for (size_t i = 0; i < numSamples; i++)
+      samples [ (size_t) math::RandInt(upperBound) ]++;
+
+    distinctSamples = arma::find(samples > 0);
```
rcurtin: I think this does not guarantee that `distinctSamples.n_elem == numSamples`.  Correct me if I have misunderstood.

src/mlpack/core/tree/vantage_point_tree/vantage_point_tree.hpp
```diff
@@ -0,0 +1,513 @@
...
+#include "vantage_point_split.hpp"
+
+namespace mlpack {
+namespace tree /** Trees and tree-building procedures. */ {
+
+/**
+ * The vantage point tree is a variant of a binary space tree. The difference
+ * from BinarySpaceTree is a presence of points in intermediate nodes.
+ * If an intermediate node holds a point, this point is the centroid of the
+ * bound.
```
rcurtin: We should probably reference Peter Yianilos's paper here (and possibly Jeffrey Uhlmann's too), and point out that the vantage point tree and the metric tree are the same thing.

src/mlpack/core/tree/vantage_point_tree/vantage_point_tree_impl.hpp
```diff
@@ -0,0 +1,921 @@
...
+         template<typename BoundMetricType, typename...> class BoundType,
+         template<typename SplitBoundType, typename SplitMatType>
+             class SplitType>
+inline
+typename VantagePointTree<MetricType, StatisticType, MatType, BoundType,
+    SplitType>::ElemType
+VantagePointTree<MetricType, StatisticType, MatType, BoundType,
+    SplitType>::MinimumBoundDistance() const
+{
+  return bound.MinWidth() / 2.0;
```
rcurtin: In this case I think this will just be `bound.OuterRadius()`. (That's the same as this, I think, just without a division.)

src/mlpack/core/tree/vantage_point_tree/vantage_point_tree_impl.hpp
```diff
@@ -0,0 +1,921 @@
...
+         template<typename SplitBoundType, typename SplitMatType>
+             class SplitType>
+inline size_t VantagePointTree<MetricType, StatisticType, MatType, BoundType,
+                              SplitType>::NumPoints() const
+{
+  // Each intermediate node contains exactly one point.
+  if (left && parent)
+    return 1;
+  else if(!parent)
+	return 0;
```
rcurtin: This means that the root does not have a vantage point?

src/mlpack/core/tree/vantage_point_tree/vantage_point_tree_impl.hpp
```diff
@@ -0,0 +1,921 @@
...
+  left = new VantagePointTree(this, splitBegin, splitCol - splitBegin, splitter,
+      maxLeafSize, true);
+  right = new VantagePointTree(this, splitCol, splitBegin + splitCount - splitCol,
+      splitter, maxLeafSize, false);
+
+  // Calculate parent distances for those two nodes.
+  arma::vec center, leftCenter, rightCenter;
+  Center(center);
+  left->Center(leftCenter);
+  right->Center(rightCenter);
```
rcurtin: Shouldn't the parent distance be identical for both children?  (I think it will just be the distance between this node's vantage point and the left child's vantage point)

rcurtin:
Ok, it looks pretty good but I think there are a lot of optimizations we can make.  Aside from the comments I've made in the diffs, it looks like there is a test failure for `RSModelTest`, probably for the VP tree.  I know why this is, but it is a subtle detail that I am not sure is documented:

The mlpack dual-tree algorithms all assume that all of the query points are in the leaf nodes of the query tree.  This is true for the kd tree, ball tree, cover tree, and rectangle tree (... I think?  Correct me if I am wrong), but not the VP tree as implemented here, since we are not duplicating points inside the tree.  As a result, what happens is that if I have two two-level binary trees like this:

```
    N_q0
   |   |
N_q1   N_q2
```

```
   N_r0
  |   |
N_r1  N_r2
```

then we will visit node combinations like this (if nothing is pruned):

```
(N_q0, N_r0)
(N_q1, N_r1)
(N_q1, N_r2)
(N_q2, N_r1)
(N_q2, N_r2)
```

and at each visit, we will perform base cases between the points held in each node.  But note that we never visit `(N_q0, N_r1)` or `(N_q0, N_r2)`, so base cases will not be performed between any points in `N_q0` and `N_r1` or `N_r2`, and the results end up being wrong.

Therefore, we need to, in our traverser, ensure that base cases between high-level query points and low-level reference points still happen.  We could do this if we forced a single-tree traversal with a vantage point.  So if we say that `N_q0` contains a single point `p_q0` that neither of its children do, then our results would be correct if we performed a single-tree traversal with the point-node combinations `(p_q0, N_r1)` and `(p_q0, N_r2)`.  (This generalizes to the case where there are more than three nodes in the tree.)

Similarly, the converse case is true---`N_q1` and `N_q2` will not have BaseCase() called with their points and whatever point is held by `N_r0`.  So for any given query/reference node combination, we need to

 * fire off single-tree recursions for the query vantage point and the reference nodes
 * perform `BaseCase()` between all points in the query node and all of the parent vantage points of the reference nodes

Hopefully this makes sense.  We should do the single-tree recursions and the extra base cases before we score the child nodes, so that some of the more interesting terms in the bounding functions will be used (like the `NeighborSearchRules` bounds that @MarcosPividori worked on).

So I guess what I am saying is, the dual-tree recursion structure should look like this:

```
Traverse(VPTree& q, VPTree& r):
  - Perform base cases as usual.
  - Fire off single-tree reference recursions if this is a left child: Traverse(q.Point(0), r.Left()), Traverse(q.Point(0), r.Right());
  - Do extra base cases: BaseCase(q.Point(i), r.Parent()->Point(0)), BaseCase(q.Point(i), r.Parent()->Parent()->Point(0)), etc.
 - Do usual traversal into (q.Left(), r.Left()), (q.Left(), r.Right()), (q.Right(), r.Left()), (q.Right(), r.Right())
```

Let me know if I can clarify anything about what I have written here.  I think maybe it is a bit confusing.

My other concern is that the implementation seems to not prune very well in either single-tree or dual-tree mode.  If I run with the corel dataset (37749x32, I can supply it if you like) with the kd-tree, I get these statistics:

```
$ bin/mlpack_knn -q ~/datasets/corel.csv -r ~/datasets/corel.csv -v -k 3 -t kd --single | grep "were"
[INFO ] 16220040 node combinations were scored.
[INFO ] 13861878 base cases were calculated.
```

```
$ bin/mlpack_knn -q ~/datasets/corel.csv -r ~/datasets/corel.csv -v -k 3 -t vp --single | grep "were"
[INFO ] 194931712 node combinations were scored.
[INFO ] 1214753327 base cases were calculated.
```

Maybe there is an error somewhere, where the bounds are way looser than they need to be?

src/mlpack/core/tree/vantage_point_tree/dual_tree_traverser_impl.hpp
rcurtin: My comment here is invalid, so you can ignore it.

src/mlpack/core/tree/vantage_point_tree/dual_tree_traverser_impl.hpp
```diff
@@ -0,0 +1,567 @@
...
+  else if (((!queryNode.IsLeaf()) && referenceNode.IsLeaf()) ||
+           (queryNode.NumDescendants() > 3 * referenceNode.NumDescendants() &&
+            !queryNode.IsLeaf() && !referenceNode.IsLeaf()))
+  {
+    // We have to recurse down the query node.  In this case the recursion order
+    // does not matter.
+
+    // If the first point of the query node is the centroid, the query node
+    // contains a point. In this case we should run the single tree traverser.
+    if (queryNode.IsFirstPointCentroid())
```
lozhnikov: If the query node contains a point we have to use the single tree traverser.

src/mlpack/core/tree/vantage_point_tree/dual_tree_traverser_impl.hpp
```diff
@@ -0,0 +1,567 @@
...
+    if (rightScore != DBL_MAX)
+      Traverse(*queryNode.Right(), referenceNode);
+    else
+      ++numPrunes;
+  }
+  else if (queryNode.IsLeaf() && (!referenceNode.IsLeaf()))
+  {
+    // If the reference node contains a point we should calculate all
+    // base cases with this point.
+    if (referenceNode.IsFirstPointCentroid())
```
lozhnikov: If the reference node contains a point we have to calculate all base cases.

src/mlpack/core/tree/vantage_point_tree/dual_tree_traverser_impl.hpp
```diff
@@ -0,0 +1,567 @@
...
+        else
+          ++numPrunes;
+      }
+    }
+  }
+  else
+  {
+    // If the reference node contains a point we should calculate all
+    // base cases with this point.
+    if (referenceNode.IsFirstPointCentroid())
```
lozhnikov: I think this is inefficient. Is it possible to implement the score algorithm for a reference point and a query node?

src/mlpack/core/tree/vantage_point_tree/dual_tree_traverser_impl.hpp
```diff
@@ -0,0 +1,567 @@
...
+    // query descent order does not matter, we will go to the left query child
+    // first.  Before recursing, we have to set the traversal information
+    // correctly.
+
+    double leftScore;
+    typename RuleType::TraversalInfoType leftInfo;
+    double rightScore;
+    typename RuleType::TraversalInfoType rightInfo;
+
+    if (queryNode.IsFirstPointCentroid())
```
lozhnikov: Use the single tree traverser if the query node contains a point.

src/mlpack/core/tree/vantage_point_tree/dual_tree_traverser_impl.hpp
```diff
@@ -0,0 +1,567 @@
...
+
+template<typename MetricType,
+         typename StatisticType,
+         typename MatType,
+         template<typename BoundMetricType, typename...> class BoundType,
+         template<typename SplitBoundType, typename SplitMatType>
+             class SplitType>
+template<typename RuleType>
+void VantagePointTree<MetricType, StatisticType, MatType, BoundType, SplitType>::
+DualTreeTraverser<RuleType>::Traverse(
```
lozhnikov: Single tree traverser.

src/mlpack/core/tree/vantage_point_tree/single_tree_traverser_impl.hpp
```diff
@@ -0,0 +1,115 @@
...
+  if (referenceNode.IsLeaf())
+  {
+    const size_t refEnd = referenceNode.Begin() + referenceNode.Count();
+    for (size_t i = referenceNode.Begin(); i < refEnd; ++i)
+      rule.BaseCase(queryIndex, i);
+    return;
+  }
+
+  // If the reference node contains a point we should calculate the base case.
+  if (referenceNode.IsFirstPointCentroid())
```
lozhnikov: If the reference node contains a point we have to calculate the base case.

src/mlpack/core/tree/vantage_point_tree/dual_tree_traverser_impl.hpp
lozhnikov: I pointed out all differences.

src/mlpack/core/tree/rectangle_tree/rectangle_tree.hpp
lozhnikov: Yes, I agree. That should simplify the TreeType API.

src/mlpack/core/tree/vantage_point_tree/vantage_point_split.hpp
lozhnikov: Oh, I forget that `VantagePointTreeSplit` has `SortStruct`.

src/mlpack/core/tree/vantage_point_tree/vantage_point_split_impl.hpp
lozhnikov: It seems I wrote that before `ElemType` was introduced and I forgot to replace.

src/mlpack/core/tree/vantage_point_tree/vantage_point_split.hpp
lozhnikov: I agree, that will be better. Moreover, the random projection tree uses the same function.

src/mlpack/core/tree/vantage_point_tree/vantage_point_split_impl.hpp
lozhnikov: I'll fix that.

src/mlpack/core/tree/vantage_point_tree/vantage_point_split_impl.hpp
lozhnikov: I guess it is better to define all vectors as private members of the class. It is possible since we use an instantiated object. But we have to make all methods non-static. How do you think?

src/mlpack/core/tree/vantage_point_tree/vantage_point_tree_impl.hpp
lozhnikov: Yeah, I'll fix the comment.

lozhnikov:
> But note that we never visit (N_q0, N_r1) or (N_q0, N_r2), so base cases will not be performed between any points in N_q0 and N_r1 or N_r2, and the results end up being wrong.

That was my mistake, I did not point out the difference in the traversers. I use the single tree traverser for a query point and a reference node. But if a reference node contains a point I fire off all base cases `BaseCase(queryNode.Descendant(i), referenceNode.Point(0))` since we have not got the score algorithm for a query node and a reference point. Is it possible to implement the score algorithm for a query node and a reference point?

> Maybe there is an error somewhere, where the bounds are way looser than they need to be?

Could that happen due to a huge overlap?

src/mlpack/core/tree/vantage_point_tree/dual_tree_traverser_impl.hpp
rcurtin: I agree that this is inefficient.  I think that the better solution is to do this: during recursion, for all query points in a node, perform a base case with all parent vantage points.  Something like this:

```
for (size_t i = 0; i < queryNode.NumPoints(); ++i)
{
  TreeType* p = referenceNode.Parent();
  do
  {
    if (p->NumPoints() > 0) // We are holding a vantage point.
      BaseCase(queryNode.Point(i), p->Point(0)); // p->NumPoints() should never be greater than 1.
  } while ((p = p->Parent()) != NULL);
}
```

That would have to come after the base case section but before any recursions, so I guess up by line 50 or so would be the right place.

rcurtin: Ah hang on, but we could end up with duplicated base cases like this.  Let me think for a while if there might be a better way to avoid duplications but preserve this type of technique.  Maybe you are right that traversing with a query node and a reference point could work, but that would require a lot of refactoring of all the dual-tree algorithms so I'd prefer to avoid that...

src/mlpack/core/tree/vantage_point_tree/dual_tree_traverser_impl.hpp
```diff
@@ -0,0 +1,567 @@
...
+    const size_t queryIndex,
+    VantagePointTree<MetricType, StatisticType, MatType, BoundType, SplitType>&
+        referenceNode)
+{
```
rcurtin: Would it be possible to use the `SingleTreeTraverser` here, or possibly to merge both the `DualTreeTraverser` and `SingleTreeTraverser` into a single class?

rcurtin:
Oops, it seems like I did not look closely enough at the traversers to see how they were different, so it seems like most of my comment was not necessary... sorry about that.

I thought of something else, that may allow us to avoid needing a `Score(queryNode, referenceIndex)` function in each rules class.  What if each node in the vantage-point tree had three children: a zero-radius child that only contained the vantage point, the inner child (that does not contain the vantage point), and the outer child?  The `MinDistance()` and `MaxDistance()` functions would be correct for the vantage point child, so there would be no need for a `Score(queryNode, referenceIndex)`.  I think also that might simplify traversal greatly, because the vantage point child (the zero-radius child) will be a leaf node, and thus will be kept during the traversal.  Clever ordering of which child to visit first would result in, I think, an identical traversal to what you've already implemented.  Let me know what you think of that idea---like most of my ideas, I have not thought through the full details yet in order to try and provide a quick response, so it's possible there are fatal flaws I have not encountered in my thinking.

I haven't thought completely about why the pruning might not be doing a good job.  I'll try and think about this in the next few days, but I suspect that the `NeighborSearchRules` as implemented aren't actually doing the same thing (or even a similar thing) to the algorithm given in the paper.  I'll let you know what I think of.

lozhnikov:
I think I fixed all issues discussed above except Uhlmann's paper (Do you mean "Implementing Metric Trees to Satisfy General Proximity/Similarity Queries"?).

I like the idea with a zero-radius child. I'll try to implement that. In this case I am not sure that we need the `TreeTraits:FirstPointIsCentroid()` method since the zero-radius child is the only node that has this property.

src/mlpack/core/tree/vantage_point_tree/dual_tree_traverser_impl.hpp
lozhnikov: `SingleTreeTraverser` does not take care of `traversalInfo`. If I am not mistaken `Score(queryIndex, referenceNode)` also does not modify `traversalInfo`. So I think in order to do that we should simply add a `SingleTreeTraverser` object to `DualTreeTraverser`. But `numBaseCases` will be wrong... I should think about that.

Is it a good idea to use an inherited class here?.

rcurtin: I believe you are right that `traversalInfo` is not modified by a single-tree traversal.

I don't think it's necessary to add a `SingleTreeTraverser` object to `DualTreeTraverser`, but instead simply implement both in a single class:

```
template<typename RuleType>
class Traverser
{
  void Traverse(const size_t queryIndex, TreeType& referenceNode);
  void Traverse(TreeType& queryNode, TreeType& referenceNode); 
}
```

Then inside the `VantagePointTree` class you could define this class as both the single-tree and dual-tree traverser type.  I don't think there would be an issue with `numBaseCases` or `numPrunes` or `numScores` that way (or whichever of those statistics are tracked).  Maybe I have overlooked something here?

rcurtin:
I actually meant this Uhlmann paper:
http://www.sciencedirect.com/science/article/pii/002001909190074R

Metric trees are essentially the same thing as vantage point trees, but they were invented independently by Yianilos and Uhlmann around the same time in the early 90s.  I think it's important to just point out in the comments that the VP tree is the same as the metric tree, for people who are looking for the metric tree who might not know it is the same as the VP tree.

rcurtin:
I think you're right that `FirstPointIsCentroid()` will be no longer applicable.  The whole idea of that is to avoid extra calculations of the center-to-center distances between nodes.  If we have another child for only the vantage points, we can still avoid center-to-center distance calculations, but it's less clear precisely how to do that.  The extent of my thinking so far is that we can define some trait like `FirstSiblingFirstPointIsCentroid` or something like this, that indicates that `Parent()->Child(0).Point(0)` is the center of the node.  But it's less clear how we would, e.g., cache the center-to-center calculation for that sibling.  I guess it is possible to use `traversalInfo` for that, if we can assume that the first sibling is always recursed into before the other siblings.  (That last bit might be another tree trait we can add, although really that's a traversal trait, so maybe this implies a new traits class?  Or alternately we can just check in the `traversalInfo` object to see if that sibling was the previous base case.)

lozhnikov:
I did the refactoring of `VantagePointTree`. Right now, each vantage point is contained in a separate leaf node.
I added `TreeTraits::FirstSiblingFirstPointIsCentroid`. But the traverser has to use `traversalInfo` in a different way since the traversal information should be equal for all siblings (since they have equal centroids).

rcurtin:
Performance is better now:
```
$ bin/mlpack_knn -r ~/datasets/corel.csv -v -k 3 -t vp --single
[INFO ] 188080778 node combinations were scored.
[INFO ] 965686362 base cases were calculated.
```

That's about 75% as many base cases.  Unfortunately when I use dual-tree search, there are some duplicates in the results (this is probably why the test is failing for range search, too).

But I still think we have a long way to go---vp-trees should be competitive with kd trees (or at least R trees!).

src/mlpack/core/tree/vantage_point_tree/traits.hpp
```diff
@@ -0,0 +1,66 @@
...
+
+  /**
+   * Points are rearranged during building of the tree.
+   */
+  static const bool RearrangesDataset = true;
+
+  /**
+   * This is always a binary tree.
+   */
+  static const bool BinaryTree = true;
```
rcurtin: I think this is no longer true.

src/mlpack/core/tree/vantage_point_tree/vantage_point_split.hpp
rcurtin: It seems like we can remove this declaration now, then.

rcurtin:
This might be a clue to the performance issue.  I modified the code to print information on the bounds and distances for single-tree search.  Here's what I see on the first node of the tree, for query point 0 (with the corel dataset, but you could reproduce these results with probably any dataset):

```
[DEBUG] MinDistance() with inner radius 0, outer radius 8.48376 and center distance 0 (innerDistance 0, outerDistance 0).
[INFO ] Score query index 0 and reference node 1c20746 (child 1): 0 vs. bound 1.79769e+308.
[DEBUG] MinDistance() with inner radius 0, outer radius 13.8071 and center distance 0 (innerDistance 0, outerDistance 0).
[INFO ] Score query index 0 and reference node 20747c17002 (child 2): 0 vs. bound 1.79769e+308.
```

It's expected that neither the first inner child nor the first outer child should be pruned.  But what's not expected is that my code in `HollowBallBound` (the `[DEBUG]` message) is reporting that the inner radius of the outer child is 0!  Maybe there is some bug where that value is calculated?

lozhnikov:
The inner radius of the outer child may be equal to zero due to the following code
```
  VantagePointTree* tree = this;

  while (tree->Parent() != NULL)
  {
    tree->Parent()->Bound() |= tree->Bound();
    tree->Parent()->furthestDescendantDistance = 0.5 *
			tree->Parent()->Bound().Diameter();
    tree = tree->Parent();
  }
```
(we have to ensure that each parent contains its children), there is no errors.

It seems I do not use `FirstSiblingFirstPointIsCentroid` correctly. I have to think about it.

rcurtin:
I see, you are right that if we expand the bound to contain all of the child bounds, we end up with this situation.  Perhaps that behavior is useful.  But at the same time, for `MaxDistance()` (single-tree), I realized that we only require the maximum possible distance between the given point and any descendant points of the node.  So I commented out the code you referenced and re-ran single tree search and got the following results:

```
$ bin/mlpack_knn -r ~/datasets/corel.csv -v -k 3 -t vp -d d-vp.csv -n n-vp.csv --single
[INFO ] 53753034 node combinations were scored.
[INFO ] 301691470 base cases were calculated.
```

The results are still correct, and it runs faster, but we are still an order of magnitude off of kd-trees with the base cases (kd-trees do about 16.3M base cases with single-tree search).  It seems like this does not affect any test results either, so I think this is a valid change.  It might be worth noting somewhere, maybe as something new in TreeTraits or something, that the vantage point tree is not guaranteed to be such that the child HollowBallBounds are entirely contained within the parent HollowBallBounds---but it is always true that all descendant points of a node are contained within the bound (that is a required property of space trees in general).

Do you think this is a valid change to make?  And maybe there is somewhere I overlooked where it is possible to make the balls even tighter than just the bit I commented out?

lozhnikov:
I think the change is valid. Unfortunately it is not possible to make balls even tighter since each ball of `HollowBallBound` always intersects a point. So, the volume of each bound depends only on the vantage point. I think the number of base cases is huge since outer balls cover a lot of empty space.

I think it is possible to make `HollowBallBound` tighter. I suggest to consider an intersection of `HollowBallBound` with the minimum bounding ball. I guess it is easy to compute `MinDistance` and `MaxDistance`. And I think that should reduce the volume of the bound significantly. How do you think?

rcurtin:
Do you mean the minimum bounding ball of the parent bound?  If so, I think that is perfectly reasonable.  All we need to do is provide a `MinDistance()` and `MaxDistance()` function that calculate tight lower and upper bounds on the distance between a query point and all a node's descendant points, or one node's descendant points and another node's descendant points.  So any clever optimizations we can do inside of the vantage point tree are fair game.  (As long as we document it, so that when someone goes to modify it later they can understand what is going on! :))

lozhnikov:
No, I mean the minimum bounding ball of points that are contained in the node.

Thus, `HollowBallBound` is an intersection of a hollow ball and the minimum bounding ball.
I think it is correct to define `MinDistance` in such a way
```
dist1 = minBoundBall.MinDistance();
dist2 = hollowBall.MinDistance();
return max(dist1, dist2);
```
And `MaxDistance` may be the minimum of maximal distances to the ball and the hollow ball.
These estimates are rough, but I think they are close to reality.

rcurtin:
I am not sure this will get us much improvement except in very specific circumstances; if points are distributed on both sides of the vantage point, then the minimum bounding ball will have radius about equal to the outer radius of the hollow bound.

I think that we can do better by inspecting the parent bound... it is not hard to detect, e.g., when an outer child of an outer parent have intersecting hollow balls (in which case, MinDistance() = 0).  Bounding what that minimum distance is when the hollow balls do not intersect, though, is not math I have yet derived, but it should not be _too_ difficult.

rcurtin:
At this point, we've both spent quite some time thinking about tighter bounds for `MinDistance()` and `MaxDistance()`, but I think the code here is fine for now.  So, I can merge this now (after making a final check over the code), and we can open a ticket for tighter bounds and consider that in some other PR once we come up with better bounds.  What do you think?

lozhnikov:
I fixed a tiny error that sometimes leads to duplicated base cases. So, I think the PR may be merged now.

rcurtin:
I checked with a number of datasets, and everything seems to work fine.

PR #709, opened 2016-06-27
Removed RectangleTree::Points()

lozhnikov:
I removed the `RectangleTree::Points()` method. Unfortunately, we can not remove `RectangleTree::Children()` so easy since sometimes we have to modify the pointers. We can use friend classes instead i.e. use `tree.children[i]` instead of `tree.Children()[i]`.

### DIFF
--- a/src/mlpack/core/tree/rectangle_tree/rectangle_tree.hpp
+++ b/src/mlpack/core/tree/rectangle_tree/rectangle_tree.hpp
@@ -328,11 +328,6 @@ class RectangleTree
   //! Modify the dataset which the tree is built on.  Be careful!
   MatType& Dataset() { return const_cast<MatType&>(*dataset); }
 
-  //! Get the points vector for this node.
-  const std::vector<size_t>& Points() const { return points; }
-  //! Modify the points vector for this node.  Be careful!
-  std::vector<size_t>& Points() { return points; }
-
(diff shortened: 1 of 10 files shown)


src/mlpack/core/tree/rectangle_tree/rectangle_tree.hpp
```diff
@@ -424,7 +419,10 @@ class RectangleTree
    *
    * @param index Index of point for which a dataset index is wanted.
    */
-  size_t Point(const size_t index) const;
+  const size_t& Point(const size_t index) const { return points[index]; }
```
rcurtin: I don't think there's any reason to return a reference here, since that is the same size as a size_t, so I'll go ahead and change this back after merge unless there is a reason I shouldn't that I am not aware of.

rcurtin:
If you think using friend classes is a good way to do it, I am fine with that.  I think ideally only `DescentType`, `SplitType`, and `AuxiliaryInformationType` should need to be friend classes of `RectangleTree`.  It's low-priority though, so only do that if you want to. :)

src/mlpack/core/tree/rectangle_tree/rectangle_tree.hpp
lozhnikov: You're right, the reference is not needed here.

PR #710, opened 2016-06-28
Hilbert R tree fixes

lozhnikov:
I found some weird errors in the Hilbert R tree implementation. I fixed them in 05fbd27.
Moreover, I added the Hilbert R tree to (NS, RS, RA)Model and (A)KNNTest, RangeSearchTest and KRANNTest.


rcurtin:
Yeah, I am not sure why that compiled, but thanks for the fixes. :)

PR #711, opened 2016-06-29
RectangleTree::NumDescendants() optimization and some RectangleTree fixes

lozhnikov:
I removed `RectangleTree::Children()`. Thus, `SplitType` and `AuxiliaryInformationType` become friend classes of the `RectangleTree` class.
Moreover, I made some RectangleTree::NumDescendants() optimizations proposed by Marcos Pividori.


rcurtin:
Thanks, this is a nice improvement.  `NumDescendants()` isn't used everywhere---I can only see its use in some situations for range search, tree-based k-means, and rank-approximate nearest neighbor search.  Still, it is good for `NumDescendants()` to be quick. :)

PR #715, opened 2016-07-04
file closed after performing operation

ashu01:
In C++, whenever a file is opened for any operation, 
then we are suppose to close that very file.


rcurtin:
Hi there,

Thanks for the contribution. But this is already done---stream objects call `close()` in the destructor, so there is no need to explicitly call it:
http://www.cplusplus.com/reference/fstream/fstream/close/

PR #716, opened 2016-07-04
Refactor PCA class: able to use different decomposition techniques (exact, randomized, QUIC SVD).

zoq:
As Nilay pointed out in his blog post: http://mlpack.org/gsocblog/we-need-to-go-deeper-googlenet-week-5-highlights.html

the discretize function that uses the PCA class, wasn't what we expected runtime wise, so we decided to use the another PCA approach: Randomized SVD as proposed in:

```
* @article{Halko2011,
*   author  = {Halko, N. and Martinsson, P. G. and Tropp, J. A.},
*   title   = {Finding Structure with Randomness: Probabilistic Algorithms for
                Constructing Approximate Matrix Decompositions},
*   journal = {SIAM Rev.},
*   volume  = {53},
*   year    = {2011},
* }
*
* @article{Szlam2014,
*   author  = {Arthur Szlam Yuval Kluger and Mark Tygert},
*   title   = {An implementation of a randomized algorithm for principal
            component analysis},
*   journal = {CoRR},
*   volume  = {abs/1412.3510},
*   year    = {2014},
* }
```
I refactored the PCA class, that is now able to use different decomposition methods including the already existing QUIC-SVD technique.


src/mlpack/methods/pca/decomposition_policies/exact_svd_method.hpp
```diff
@@ -0,0 +1,72 @@
...
+
+/**
+ * Implementation of the exact SVD policy.
+ */
+class ExactSVDPolicy
+{
+  public:
+  /**
+   * Apply Principal Component Analysis to the provided data set using the
+   * randomized SVD.
```
rcurtin: I think this comment is inaccurate, doesn't this class use exact SVD from Armadillo instead of randomized SVD?

src/mlpack/methods/pca/pca_main.cpp
```diff
@@ -57,26 +95,23 @@ int main(int argc, char** argv)
...
-  else
+  else if(decompositionMethod == "randomized")
   {
-    varRetained = p.Apply(dataset, newDimension);
+    RunPCA<RandomizedSVDPolicy>(dataset, newDimension, scale, varToRetain);
+  }
+  else if(decompositionMethod == "quic")
+  {
+    RunPCA<QUICSVDPolicy>(dataset, newDimension, scale, varToRetain);
   }
```
rcurtin: I think if a user picks an invalid decomposition policy, no error is issued? I am not sure if I am reading this right, but if so I think we should add an else to catch the error. :)

rcurtin:
This is nice, I like the changes. My only thought is, this breaks reverse compatibility---now any code that calls the `PCA` class must be changed to `PCA<>`. When I refactored the `Range` class to accept different template types, I instead renamed it to `RangeType<>` and then typedef'ed `Range` to `RangeType<double>` to avoid breaking reverse compatibility, and made a note in the documentation and in HISTORY.txt about the change and that in mlpack 3.0.0 we would change the name of `RangeType` to `Range` and remove the typedef. Maybe we could do the same here?

src/mlpack/methods/pca/decomposition_policies/quic_svd_method.hpp
```diff
@@ -0,0 +1,92 @@
...
+
+  /**
+   * Apply Principal Component Analysis to the provided data set using the
+   * QUIC-SVD method.
+   *
+   * @param data Data matrix.
+   * @param data Centered data matrix.
+   * @param transformedData Matrix to put results of PCA into.
+   * @param eigVal Vector to put eigenvalues into.
+   * @param eigvec Matrix to put eigenvectors (loadings) into.
```
stereomatchingkiss: The parameter name are different, one called eigvec, and another called coeff.
Besides, The Apply function of PCA, the comments are written for the same functions and different with each other.

pca.hpp

```
 /**
   * Apply Principal Component Analysis to the provided data set.  It is safe to
   * pass the same matrix reference for both data and transformedData.
   *
   * @param data Data matrix.
   * @param transformedData Matrix to put results of PCA into.
   * @param eigval Vector to put eigenvalues into.
   * @param eigvec Matrix to put eigenvectors (loadings) into.
   */
  void Apply(const arma::mat& data,
             arma::mat& transformedData,
             arma::vec& eigval,
arma::mat& eigvec);

```

pca_impl.hpp

```
/**
 * Apply Principal Component Analysis to the provided data set.
 *
 * @param data - Data matrix
 * @param transformedData - Data with PCA applied
 * @param eigVal - contains eigen values in a column vector
 * @param coeff - PCA Loadings/Coeffs/EigenVectors
 */
template<typename DecompositionPolicy>
void PCA<DecompositionPolicy>::Apply(const arma::mat& data,
                arma::mat& transformedData,
                arma::vec& eigVal,
arma::mat& coeff)
```

zoq: Nice catch, thanks.

src/mlpack/methods/pca/pca_main.cpp
zoq: Ah, right, changed in 1127e61.

src/mlpack/methods/pca/decomposition_policies/exact_svd_method.hpp
zoq: You are right, it should be exact SVD from Armadillo instead of randomized SVD.

rcurtin:
Looks good to me.  I might consider adding a "this class will be changed to have the name PCA in mlpack 3.0.0" comment to the PCAType documentation, but that's your choice whether to do that or not.

src/mlpack/methods/pca/pca_impl.hpp
```diff
@@ -1,19 +1,29 @@
...
  */
-#include "pca.hpp"
+
+#ifndef MLPACK_METHODS_PCA_PCA_IMPL_HPP
+#define MLPACK_METHODS_PCA_PCA_IMPL_HPP
+
 #include <mlpack/core.hpp>
+#include "pca.hpp"
 
 using namespace std;
```
stereomatchingkiss: Do you forgot to removed using namespace std from header file?

src/mlpack/methods/pca/pca_impl.hpp
```diff
@@ -1,19 +1,29 @@
 /**
  * @file pca.cpp
```
stereomatchingkiss: I think it should be pca_impl.hpp

src/mlpack/methods/pca/pca_impl.hpp
```diff
@@ -33,10 +35,10 @@ PCA<DecompositionPolicy>::PCA(const bool scaleData,
...
  */
 template<typename DecompositionPolicy>
-void PCA<DecompositionPolicy>::Apply(const arma::mat& data,
-                arma::mat& transformedData,
-                arma::vec& eigVal,
-                arma::mat& coeff)
+void PCAType<DecompositionPolicy>::Apply(const arma::mat& data,
+                                         arma::mat& transformedData,
+                                         arma::vec& eigVal,
+                                         arma::mat& coeff)
```
stereomatchingkiss: This parameter should call "eigvec" or "coeff", they are different in pca.hpp and pca_impl.hpp

src/mlpack/methods/pca/pca_impl.hpp
zoq: It's used for:

``
Log::Fatal << "PCA::Apply(): varRetained (" << varRetained << ") must be "
        << "greater than or equal to 0." << endl;
``

src/mlpack/methods/pca/pca_impl.hpp
zoq: Nice catch, thanks!

zoq:
Merged in cfbd604.

Issue #712, opened 2016-06-29
Check if SortStruct provides any benefit.


rcurtin:
This comes from the discussion of #699: we are not sure if using `SortStruct` instead of `pair<double, size_t>` provides any speedup. So we should check, and then pick the better of the two approaches.

lozhnikov:
I wrote a simple test. With `-O3` `std::pair` is faster then `SortStruct`.
If `n=100M`:
```
SortStruct results:
sorted[0].n = 26200464
fill duration = 1620ms
sort duration = 17090ms
std::pair results:
sorted[0].n = 26200464
fill duration = 1597ms
sort duration = 12287ms
```
[main.cpp.zip](https://github.com/mlpack/mlpack/files/340894/main.cpp.zip)

lozhnikov:
I realized that we do not sort huge amounts of data. I corrected the test.
With `-O3` `std::pair` is a little bit faster.
If `n = 10`, `numIterations = 100M`:
```
SortStruct results:
unnecessaryCodeRemovalPreventer = 1.09091e+08
sort duration = 33621ms
std::pair results:
unnecessaryCodeRemovalPreventer = 1.03127e+08
sort duration = 29536ms
```
[main2.cpp.zip](https://github.com/mlpack/mlpack/files/340950/main2.cpp.zip)

rcurtin:
Results on my system with gcc:

```
$ ./main2
SortStruct results:
unnecessaryCodeRemovalPreventer = 1.09092e+08
sort duration = 21890ms
std::pair results:
unnecessaryCodeRemovalPreventer = 1.20977e+08
sort duration = 18936ms
```

and with clang:

```
SortStruct results:
unnecessaryCodeRemovalPreventer = 1.0909e+08
sort duration = 81754ms
std::pair results:
unnecessaryCodeRemovalPreventer = 1.01206e+08
sort duration = 80806ms
```

clang is much slower!  I also had one more test I wanted to do---using the default comparator for `std::pair`.  So I removed the third argument from line 74 of main2.cpp and switched the order of the pair to hold double first then size_t.

gcc:

```
SortStruct results:
unnecessaryCodeRemovalPreventer = 1.09092e+08
sort duration = 21970ms
std::pair results:
unnecessaryCodeRemovalPreventer = 1.26484e+08
sort duration = 19171ms
```

clang:

```

```
SortStruct results:
unnecessaryCodeRemovalPreventer = 1.09091e+08
sort duration = 80945ms
[...]

PR #721, opened 2016-07-12
Replaced SortStruct by std::pair

lozhnikov:
I replaced `SortStruct` by `std::pair` in the R* tree and the X tree and fixed some compiler warnings about uninitialized variables. I did some tests, `std::pair` provides some speedups in `mlpack_knn` for the R* tree and the X tree.

### DIFF
--- a/src/mlpack/core/tree/rectangle_tree/x_tree_split_impl.hpp
+++ b/src/mlpack/core/tree/rectangle_tree/x_tree_split_impl.hpp
@@ -75,12 +75,7 @@ void XTreeSplit::SplitLeafNode(TreeType *tree,std::vector<bool>& relevels)
        sorted[i].second = i;
     }
 
-    std::sort(sorted.begin(), sorted.end(),
-        [] (const std::pair<ElemType, size_t>& p1,
-            const std::pair<ElemType, size_t>& p2)
-        {
-          return p1.first < p2.first;
-        });
+    std::sort(sorted.begin(), sorted.end(), PairComp<ElemType>);
     std::vector<size_t> pointIndices(p);
 
     for (size_t i = 0; i < p; i++)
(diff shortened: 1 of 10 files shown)


src/mlpack/methods/ann/layer/one_hot_layer.hpp
```diff
@@ -58,7 +58,7 @@ class OneHotLayer
     output = inputActivations;
     output.zeros();
 
-    arma::uword maxIndex;
+    arma::uword maxIndex = 0;
```
lozhnikov: I looked through the armadillo code, `op_min::direct_min` equates `index_of_min_val` to zero by default.

src/mlpack/core/tree/rectangle_tree/x_tree_split_impl.hpp
```diff
@@ -645,25 +601,30 @@ bool XTreeSplit::SplitNonLeafNode(TreeType *tree,std::vector<bool>& relevels)
...
     }
   }
 
-  std::sort(sorted.begin(), sorted.end(), structComp<ElemType>);
+  std::sort(sorted.begin(), sorted.end(),
+      [] (const std::pair<ElemType, size_t>& p1,
+          const std::pair<ElemType, size_t>& p2)
+      {
+        return p1.first < p2.first;
+      });
```
rcurtin: Do you think it would be better to define this comparator function in some header file in the rectangle_tree directory instead of writing the lambda many times?  If not, I'll merge as-is, but it might be a good idea to make the code a bit shorter.

src/mlpack/methods/ann/layer/one_hot_layer.hpp
rcurtin: Yeah; I agree that the -Wuninitialized is erroneous in this case, but I am not sure there is any other way to stop that warning from appearing here other than initializing the index to 0.  Do you know any other ways?

lozhnikov: No, I don't. I think this is the easiest way.

rcurtin:
Looks good to me.  This fixes #721.

lozhnikov:
Do you mean #712? :)

rcurtin:
Yep, I must have mistyped... :)

lozhnikov:
I mean to say that if the code looks good, did you forget to merge the PR?

rcurtin:
Oh, yes, so I not only mistyped, I also hit the wrong button!

PR #724, opened 2016-07-16
Fixed a segfault in RPlusTreeSplit. Fixed traits for the R+/R++ tree.

lozhnikov:
If we are going to split an overflowed node we have to ensure that the resulting nodes have the desired capacity.
The R+/R++ tree has non-overlapping children, so we should fix the traits.

### DIFF
--- a/src/mlpack/core/tree/rectangle_tree/traits.hpp
+++ b/src/mlpack/core/tree/rectangle_tree/traits.hpp
@@ -56,6 +56,54 @@ class TreeTraits<RectangleTree<MetricType, StatisticType, MatType, SplitType,
   static const bool BinaryTree = false;
 };
 
+/**
+ * Since the R+/R++ tree can not have overlapping children, we should define
+ * traits for the R+/R++ tree.
+ */
+template<typename MetricType,
+         typename StatisticType,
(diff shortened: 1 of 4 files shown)


src/mlpack/core/tree/rectangle_tree/traits.hpp
```diff
@@ -56,6 +56,54 @@ class TreeTraits<RectangleTree<MetricType, StatisticType, MatType, SplitType,
...
+    RPlusTreeSplit<SplitPolicyType,
+                   SweepType>,
+    DescentType,
+    AuxiliaryInformationType>>
+{
+ public:
+  /**
+   * The R+/R++ tree can't have overlapping children.
+   */
+  static const bool HasOverlappingChildren = false;
```
rcurtin: I thought that there was a degenerate case that could occur with the R+ tree where nodes could end up overlapping.  I don't remember the exact details; is that still true?  Or did your code work around that situation?

lozhnikov: No, if we can not enlarge any node in order to insert a point, we just add an intermediate node.

PR #726, opened 2016-07-18
Random projection trees

lozhnikov:
I added an implementation of random projection trees. The RPTreeMax variant is implemented slightly different since the technique for the calculation of a random deviation suggested in the paper is weird. And I use a number of random dataset points instead of all points in order to calculate the median value and the average distance between points.

Since the norm in the paper always denotes Euclidean distance I wrote `arma::dot` in order to calculate the scalar product of two vectors.

Right now, the tree use `HRectBound`.


src/mlpack/core/tree/binary_space_tree/rp_tree_max_split_impl.hpp
```diff
@@ -0,0 +1,156 @@
...
+    arma::Col<ElemType>& direction)
+{
+  arma::Col<ElemType> origin;
+
+  origin.zeros(direction.n_rows);
+
+  for (size_t k = 0; k < direction.n_rows; k++)
+    direction[k] = math::Random(-1.0, 1.0);
+
+  ElemType length = metric::EuclideanDistance::Evaluate(origin, direction);
```
rcurtin: I think you could simplify this:

```
direction.randu(); // Fill with [0, 1].
direction -= 0.5; // Shift to [-0.5, 0.5].
const double norm = arma::norm(direction); // Get length for normalization.
if (norm == 0)
{
  const size_t k = math::RandInt(direction.n_rows);
  direction[k] = 1.0;
  length = 1.0;
}

// Normalize.
direction /= length;
```

src/mlpack/core/tree/binary_space_tree/rp_tree_max_split_impl.hpp
```diff
@@ -0,0 +1,156 @@
...
+
+  samples.zeros(count);
+
+  for (size_t i = 0; i < numSamples; i++)
+    samples [ (size_t) math::RandInt(count) ]++;
+
+  distinctSamples = arma::find(samples > 0);
+
+  distinctSamples += begin;
+}
```
rcurtin: I think we talked in another PR or somewhere else about creating a general `GetDistinctSamples` function and putting it in `src/mlpack/core/` somewhere.  Or was that @mentekid I talked about that with?  I can't remember :(

mentekid: Doesn't ring a bell to me :/

rcurtin: Ok, no worries, maybe I need to take better notes on my thoughts. :)

MarcosPividori:
Woow! a minute! This pull request seems to implement something very similar to what I have implemented for spill trees: https://github.com/mlpack/mlpack/pull/747/commits/9588a6dd63154481f0442e087bde0e640d02ec7f
When splitting the nodes according to the values in a given projection. I think we should merge both implementations.
In my commit I implement an abstraction: ProjectionVector and Hyperplane. And use SpaceSplit class to split the nodes according to their projection.

src/mlpack/core/tree/binary_space_tree/rp_tree_mean_split_impl.hpp
```diff
@@ -0,0 +1,191 @@
...
+    // If the vector is equal to 0, choose an arbitrary dimension.
+    size_t k = math::RandInt(direction.n_rows);
+
+    direction[k] = 1.0;
+
+    length = metric::EuclideanDistance::Evaluate(origin, direction);
+
+    direction[k] /= length;
+  }
+}
```
rcurtin: I think this also can be replaced with the shorter function idea I sketched in another comment.  Another thought is, since you are using it twice, maybe mark the function static in `RPTreeMaxSplit` and then call it from `RPTreeMeanSplit` instead of implementing it again here.  I am not sure this is a great function to add to the core mlpack code, because of its odd behavior when a vector of length 0 is selected.  Is that behavior the best thing to do here, or maybe is it better to simply draw a random vector again?

src/mlpack/core/tree/binary_space_tree/rp_tree_mean_split_impl.hpp
```diff
@@ -0,0 +1,191 @@
...
+    const arma::uvec& samples,
+    const arma::Col<ElemType>& direction,
+    ElemType& splitVal)
+{
+  std::vector<ElemType> values(samples.n_elem);
+
+  for (size_t k = 0; k < samples.n_elem; k++)
+    values[k] = arma::dot(data.col(samples[k]), direction);
+
+  std::sort(values.begin(), values.end());
```
rcurtin: It might be better to use Armadillo objects here, since you know how many elements you'll have.  Armadillo does have a `sort` function too.

src/mlpack/core/tree/binary_space_tree/rp_tree_max_split_impl.hpp
lozhnikov: Yes, I remember. That branch has not merged yet. I think it is better to correct that after merging.

src/mlpack/core/tree/binary_space_tree/rp_tree_mean_split_impl.hpp
```diff
@@ -0,0 +1,191 @@
...
+}
+
+template<typename BoundType, typename MatType>
+bool RPTreeMeanSplit<BoundType, MatType>::GetMeanMedian(
+    const MatType& data,
+    const arma::uvec& samples,
+    arma::Col<ElemType>& mean,
+    ElemType& splitVal)
+{
+  std::vector<ElemType> values(samples.n_elem);
```
rcurtin: Same here, I think the Armadillo object will be a better choice.

src/mlpack/core/tree/binary_space_tree/rp_tree_mean_split_impl.hpp
```diff
@@ -0,0 +1,191 @@
...
+
+  mean.zeros(data.n_rows);
+
+  for (size_t k = 0; k < samples.n_elem; k++)
+    mean += data.col(samples[k]);
+
+  mean /= samples.n_elem;
```
rcurtin: This is a low-priority suggestion, so if you don't have time to look into it, I don't think it will make much difference.  So, I am wondering if you could use non-contiguous matrix views here to accomplish the same thing.  You could do something like...

```
arma::mean(data.cols(samples), 1);
```

But the question is whether or not the template metaprogramming inside Armadillo will produce something equivalently fast to what you have written above, or if it will instead do something less smart, like instantiating a whole new matrix to hold the subview, and then calculating the mean for that matrix.  I think you could determine this either by inspecting the Armadillo codebase or by just writing a test and comparing.

src/mlpack/core/tree/binary_space_tree/traits.hpp
```diff
@@ -56,6 +56,78 @@ class TreeTraits<BinarySpaceTree<MetricType, StatisticType, MatType, BoundType,
...
+         typename MatType,
+         template<typename BoundMetricType, typename...> class BoundType>
+class TreeTraits<BinarySpaceTree<MetricType, StatisticType, MatType, BoundType,
+                                 RPTreeMaxSplit>>
+{
+ public:
+  /**
+   * Each binary space tree node has two children which represent
+   * non-overlapping subsets of the space which the node represents.  Therefore,
+   * children are not overlapping.
```
rcurtin: I think this comment is inaccurate for the random projection tree.

lozhnikov:
@MarcosPividori I looked through your code, indeed, it is better to rewrite the RP tree split algorithm in terms of `ProjectionVector` and `Hyperplane`. I'll do that when your branch will be merged.

src/mlpack/core/tree/binary_space_tree/traits.hpp
```diff
@@ -56,6 +56,78 @@ class TreeTraits<BinarySpaceTree<MetricType, StatisticType, MatType, BoundType,
...
+         typename MatType,
+         template<typename BoundMetricType, typename...> class BoundType>
+class TreeTraits<BinarySpaceTree<MetricType, StatisticType, MatType, BoundType,
+                                 RPTreeMeanSplit>>
+{
+ public:
+  /**
+   * Each binary space tree node has two children which represent
+   * non-overlapping subsets of the space which the node represents.  Therefore,
+   * children are not overlapping.
```
rcurtin: Same issue with the comment here.

src/mlpack/core/tree/binary_space_tree/typedef.hpp
```diff
@@ -135,6 +135,20 @@ using MeanSplitBallTree = BinarySpaceTree<MetricType,
...
+                                  MatType,
+                                  bound::HRectBound,
+                                  RPTreeMaxSplit>;
+
+template<typename MetricType, typename StatisticType, typename MatType>
+using RPTreeMean = BinarySpaceTree<MetricType,
+                                  StatisticType,
+                                  MatType,
+                                  bound::HRectBound,
+                                  RPTreeMeanSplit>;
```
rcurtin: I think for these, `MaxSplitRPTree` and `MeanSplitRPTree` might be more natural names.  Which do you think is the "default"?  If the max split is the more common one, then I would just call that `RPTree`.  This is kind of what I've done with the other types of trees---only specify the split type if it's not the "usual" split type.

src/mlpack/methods/neighbor_search/kfn_main.cpp
```diff
@@ -194,10 +195,14 @@ int main(int argc, char *argv[])
       tree = KFNModel::R_PLUS_TREE;
     else if (treeType == "r-plus-plus")
       tree = KFNModel::R_PLUS_PLUS_TREE;
+    else if (treeType == "rp-tree-max")
+      tree = KFNModel::RP_TREE_MAX;
+    else if (treeType == "rp-tree-mean")
+      tree = KFNModel::RP_TREE_MEAN;
```
rcurtin: Same here, I'd call one of these just the `rp-tree` and the other the `<split>-rp-tree`.

src/mlpack/core/tree/binary_space_tree/rp_tree_mean_split_impl.hpp
lozhnikov: Yes, I agree, we can move that function to the core, but in that case we can not use `arma::norm`.

src/mlpack/core/tree/binary_space_tree/rp_tree_mean_split_impl.hpp
lozhnikov: okay, I'll look through the armadillo code.

rcurtin:
Ok, I think that this is a good start.  I like that you were able to reuse `BinarySpaceTree`.  In addition to the comments I made inline, I think there are three main things left to be done:

 * The `HRectBound` will not be a tight bound.  Instead the true bound is a random polytope.  But I am not sure it is feasible right now to figure out how to precisely bound that polytope, so we can open an issue and leave that for later.

 * I need to run some timing tests to ensure that performance is reasonable with the RP tree.  I think that it should be the same order of magnitude as the kd-tree search time for `mlpack_knn`.

 * Can you think of any more tests?  Right now the tests just ensure that the dataset is ordered reasonably.  But I think it makes sense to test the `RPTreeMaxSplit` and `RPTreeMeanSplit` classes individually, and then maybe run a test where we ensure that there does exist some splitting hyperplane between each child.  (I am not sure how easy that test would be to construct.)

src/mlpack/core/tree/binary_space_tree/rp_tree_max_split_impl.hpp
rcurtin: Ah, right, that is part of #708.  I left a comment there recently asking if you think it is ready to merge for now, so if you agree it's ready for now then I'll go ahead and merge it, and you can refactor the code here.

src/mlpack/core/tree/binary_space_tree/typedef.hpp
lozhnikov: Maybe you are right. These names are used in Dasgupta's paper "Random projection trees and low dimensional manifolds". I guess the mean split is the "default" since I saw that in another paper.

MarcosPividori:
Continuing with my previous comment:

I think we have 3 independent decisions to make, when building a binary space tree:
 + **Which kind of splitting Hyperplane:** *Axis-Orthogonal*, or *General* (not necessarily Axis-Orthogonal).
 + **How to choose the hyperplane:** *Random* or *Widest range* (the one with widest range of projection values).
 + **Which Splitting value we wil consider:** *MidPoint value*, *Mean value*, *Median Value*.

Every combination is valid, for example:
 + *Axis-Orthogonal - Random - MidPoint*.
 + *General - Widest range - Mean Value*.
 + *Axis-Orthogonal - Widest range - Median Value*.
 + etc.

Now, for Spill Trees, I conside all the combinations between Hyperplane type and Splitting Value, as 2 template parameters. We could add a third template parameter that represents the criterion to choose the hyperplane, between Random and Widest range.

lozhnikov:
I looked through a series of papers. The polytope distance problem requires a lot of arithmetic operations. All methods that I came across are based on Hilbert's algorithm (i.e. they are iterative).

I thought that in order to test the split algorithm I have to remember the normal to the hyperplane. Maybe I am mistaken.

lozhnikov:
@MarcosPividori The RP tree mean split algorithm sometimes split the space according to the distance to a particular point. So, I am not sure that it is good idea to add this functionality to the `SpaceSplit` class since `SpaceSplit` always splits data by a hyperplane.

MarcosPividori:
@lozhnikov you are right. I supposed that binary space trees always consider a splitting hyperplane, but it looks that is not true....
Well it is a bit weird... when considering RPTree-Mean, points can be splitted based on a distance to a mean point, not using a splitting hyperplane... I don't think this case would be exactly a BinarySpaceTree, it would be something different.... woudn't it?
I have quickly read the paper and I can see RPTree-Mean is defined as you implemented it. But I think there is some kind of contradiction in the paper. In a paragraph it is said:

*"On the other hand, an RPTree chooses a direction uniformly at random from the unit sphere ... and splits the data into two roughly equal-sized sets using a hyperplane orthogonal to this direction. We describe two variants, which we call RPTree-Max and RPTree-Mean. Both are adaptive to intrinsic dimension, although the proofs are in different models and use different techniques."*

So, one would expect both variants to consider a hyperplane orthogonal to this direction...

lozhnikov:
@MarcosPividori I reread the paper. I think there are no errors. I guess the bound for the mean split random projection tree should contain a hollow like `HollowBallBound` for the vantage point tree.

Moreover, I found another paper "Random projection trees for vector quantization". The paper describes the same mean split algorithm.

Why do you think that 'RPTreeMean' differs from `BinarySpaceTree`?. As well as `BinarySpaceTree`, `RPTreeMean` divides the space into two parts. In contrast to `KDTree`, children of a `RPTreeMean` node may overlap. Thus, I updated traits for random projection trees.

lozhnikov:
I've fixed issues and I've thought out tests in order to ensure that a splitting hyperplane exists.

Two points are situated on opposite sides of a hyperplane if the hyperplane expression takes different signs on these points. Thus I can find the hyperplane as the solution of a system of inequalities. I use a simple iterative method in order to find the solution of the system. I thought it shouldn't work, but I tried a number of different seeds and it seems it works.

MarcosPividori:
@lozhnikov yes! I agree that there is no error.  I meant to say that it was a bit confusing not using a hyperplane there.... because Random Projection Trees seems to be based on splitting hyperplanes. But, I am sure they have their reasons!
It becomes a bit difficult to generalize some concepts when we implement many differents flavours! :)
So, I think we can consider both PR separately, and once they are merged with the master branch we can see if it is possible to unify both implementations of the splitting algorithms, to avoid duplicated code. Would you agree?
Thanks!

lozhnikov:
@MarcosPividori yeah, sure. I think `RPTreeMax`should be definitely rewritten in terms of `SpaceSplit` and `RPTreeMean` should use the notion `Hyperplane`.

src/mlpack/core/tree/binary_space_tree/rp_tree_max_split.hpp
```diff
@@ -0,0 +1,109 @@
...
+   *
+   * @param point The point that is being assigned.
+   * @param splitInfo An information about the split.
+   */
+  template<typename VecType>
+  static bool AssignToLeftNode(
+    const VecType& point,
+    const SplitInfo& splitInfo)
+  {
+    return (arma::dot(point, splitInfo.direction) <= splitInfo.splitVal);
```
rcurtin: I spent a while thinking about this.  The inner product is defined for L2 space but only for L2 space---in other spaces like L1, L3, L_inf, etc., it is not defined.  I think this is why the paper restricts itself to considering L2.  But I don't think that using the dot product like this in other spaces is a problem.  Let me know if you think otherwise.

src/mlpack/core/tree/binary_space_tree/rp_tree_mean_split_impl.hpp
```diff
@@ -0,0 +1,167 @@
...
+          data.col(samples[j]));
+
+  dist /= (samples.n_elem * (samples.n_elem - 1) / 2);
+
+  return dist;
+}
+
+template<typename BoundType, typename MatType>
+void RPTreeMeanSplit<BoundType, MatType>::GetRandomDirection(
+    arma::Col<ElemType>& direction)
```
rcurtin: I don't know if I have already written this somewhere else, but would it be better to move this to `core/math/` as a standalone function?

rcurtin: Ah actually, the function is already there!  `math::RandVector()`.  Maybe it would be useful to add an overload with a parameter that allows you to select the number of dimensions.  Note that selecting a random vector by uniformly sampling and then normalizing as you've done here actually causes bias, because you want to uniformly sample from the surface of a sphere.  I believe the implementation Nishant wrote many years ago uses the Box-Muller transform:

https://en.wikipedia.org/wiki/Box%E2%80%93Muller_transform

src/mlpack/core/tree/binary_space_tree/rp_tree_max_split_impl.hpp
```diff
@@ -0,0 +1,114 @@
...
+
+  return true;
+}
+
+template<typename BoundType, typename MatType>
+typename MatType::elem_type RPTreeMaxSplit<BoundType, MatType>::
+GetRandomDeviation(const MatType& data,
+                   const size_t begin,
+                   const size_t count,
+                   const arma::Col<ElemType>& direction)
```
rcurtin: I am confused, it seems like this function is not used anywhere.

src/mlpack/core/tree/binary_space_tree/rp_tree_max_split_impl.hpp
```diff
@@ -0,0 +1,114 @@
...
+  if (minimum == maximum)
+    return false;
+
+  splitVal = arma::median(values);
+
+  // Add a random deviation to the median.
+  // This algorithm differs from the method suggested in the
+  // random projection tree paper.
+  splitVal += math::Random((minimum - splitVal) * 0.75,
+      (maximum - splitVal) * 0.75);
```
rcurtin: Other than the clear speed improvement, why did you choose to go with this simpler approach?  I am not sure it will still satisfy the theoretical guarantees given by the paper (I have not spent a huge amount of time thinking about the necessary conditions for that).

src/mlpack/core/tree/binary_space_tree/rp_tree_mean_split_impl.hpp
```diff
@@ -0,0 +1,167 @@
...
+{
+  const size_t maxNumSamples = 100;
+  const size_t numSamples = std::min(maxNumSamples, count);
+  arma::uvec samples;
+
+  // Get no more than numSamples distinct samples.
+  math::ObtainDistinctSamples(begin, begin + count, numSamples, samples);
+
+  // Find the average distance between points.
+  ElemType averageDistanceSq = GetAveragePointDistance(data, samples);
```
rcurtin: Continuing with the discussion about whether the theory is still valid with your modifications, I think that you are right to do sampling to estimate statistics of the dataset.  An exact approach would be quadratic time!  This would yield a total tree building time of something like O(N^2 log N).  So I think that implementing the tree exactly as given in the paper is not really feasible.

It should be possible to re-derive the theorems, taking into account the sampling you are doing here, without affecting the results significantly (only the success probability).  Don't feel obligated to do that---I am mostly just thinking out loud here about the implications. :)

src/mlpack/methods/neighbor_search/knn_main.cpp
```diff
@@ -63,11 +63,12 @@ PARAM_INT_IN("k", "Number of nearest neighbors to find.", "k", 0);
...
 // The user may specify the type of tree to use, and a few parameters for tree
 // building.
-PARAM_STRING_IN("tree_type", "Type of tree to use: 'kd', 'vp', 'cover', 'r', "
-    "'r-star', 'x', 'ball', 'hilbert-r', 'r-plus', 'r-plus-plus'.", "t", "kd");
-PARAM_INT_IN("leaf_size", "Leaf size for tree building (used for kd-trees, vp "
-    "trees, R trees, R* trees, X trees, Hilbert R trees, R+ trees and R++ "
-    "trees).", "l", 20);
+PARAM_STRING_IN("tree_type", "Type of tree to use: 'kd', 'vp', 'rp-tree', "
+    "'max-split-rp-tree', 'cover', 'r', 'r-star', 'x', 'ball', 'hilbert-r', "
+    "'r-plus', 'r-plus-plus'.", "t", "kd");
```
rcurtin: We should call it `rp` and `max-rp` to be consistent with the rest of the tree names.

src/mlpack/methods/neighbor_search/ns_model_impl.hpp
```diff
@@ -484,6 +492,10 @@ std::string NSModel<SortPolicy>::TreeName() const
       return "R++ tree";
     case VP_TREE:
       return "Vantage point tree";
+    case RP_TREE:
+      return "Random projection tree (mean split)";
```
rcurtin: No need for a capital r here, lowercase fits better with how these strings are used (I guess that applies to the VP tree too).

rcurtin:
Ok, things look pretty good to me.  As for the refactoring, we can do that after spill trees are merged.

I ran timing simulations, and as expected, the RP tree is slower than the kd-tree (after all it has looser bounds in all cases).  I only have one remaining question and then a few observations.

 * Should we use RPMax as the default?  Nearly all of my simulations indicate that it performs better than RPMean.

 * The random projection tree appears to do almost as well as the kd-tree in higher dimensions; for instance, performance is almost the same on the MNIST dataset.  I wonder if there exist datasets where it may outperform the kd-tree.

 * The results of one of Pari's papers (http://papers.nips.cc/paper/5121-which-space-partitioning-tree-to-use-for-search) compares random projection trees, kd-trees, and some other types of trees for defeatist search.  Pari's results indicate that the rp-tree is generally the worst-performing of the tree types that he tried.  I wonder if we will see the same results in our own surveys for defeatist search.

I am not sure, in the end, how effective the random projection tree will be, but I think that it is good to have it implemented.  It seems there are not many studies of its empirical effectiveness (if you know of any, send some links!  I would be interested to read).

src/mlpack/core/tree/binary_space_tree/rp_tree_max_split.hpp
lozhnikov: I should reread the paper. I am not sure that other spaces preserve the theoretical guarantees.

src/mlpack/core/tree/binary_space_tree/rp_tree_max_split_impl.hpp
lozhnikov: Yeah, I substituted that by 
`math::Random((minimum - splitVal) * 0.75, + (maximum - splitVal) * 0.75);`
since I am not sure that the original algorithm guarantees that both children will not be empty.

src/mlpack/core/tree/binary_space_tree/rp_tree_max_split_impl.hpp
lozhnikov: I am not sure that both children will not be empty. Maybe it is better to use the original algorithm with restricted upper and lower bounds.

lozhnikov:
I fixed the issues. Indeed, the number of base cases with the mean-split rp tree is greater than the number of base cases with the max split rp-tree. On the other hand, the total number of scores is less. I tried a number of datasets and noticed that the total time with the mean-split rp tree is less. Indeed, the mean-split rp tree is more balanced since there is no shift from the median split value.

src/mlpack/core/tree/binary_space_tree/rp_tree_max_split.hpp
rcurtin: It's true, I think the theory is only valid for L2. I just wanted to point out the issue, that's all. I don't think there is anything we can easily do about it.

src/mlpack/core/tree/binary_space_tree/rp_tree_max_split_impl.hpp
rcurtin: Ok, I can agree with that.  But I will remove this function when I merge and add a comment why we are not using the paper's technique.

rcurtin:
The AppVeyor build seems to be because there is no `/bigobj`, but #747 solves that, so I'll just ignore the error for now and it will be fixed when that is merged (very soon).  Based on what you've said, I think staying with mean-split rp-tree as a default is fine.  The numbers I saw were quite close anyway.

rcurtin:
I made a few simplifications and style fixes in 0f4b25a; let me know if I broke anything and I will fix it. :)

PR #727, opened 2016-07-19
Fix TreeTrait for BallTree.

MarcosPividori:
I update TreeTraits for Ball Trees, to set HasOverlappingChildren = true.
It shouldn't make any difference in current implementation because HasOverlappingChildren is not used in any method.
I thought it could be useful in the future. :)


rcurtin:
I agree, thanks for the fix.

PR #729, opened 2016-07-20
Adds a GammaDistribution object for estimating gamma parameters

mentekid:
This is the first step for the LSH Model. Since it is unrelated to the rest of the code and can be merged by itself, I opened a separate PR.


src/mlpack/core/dists/gamma_distribution.cpp
```diff
@@ -0,0 +1,76 @@
+/**
+ * @file gamma_distribution.cpp
+ * @author Yannis Mentekidis
+ *
+ * Implementation of the methods of GammaDistribution.
+ */
+#include "gamma_distribution.hpp"
+
+#include <boost/math/special_functions/digamma.hpp>
+#include <boost/math/special_functions/trigamma.hpp>
```
mentekid: The travis build can't find this function. Why? Isn't boost::math a dependency of mlpack?

Or is it a new function added to a later version of boost?

rcurtin: Maybe it is from a newer version of Boost.  You can download the different versions to pinpoint when it became available.  If we have to upgrade our required Boost version, we can probably do that, or possibly backport the digamma and trigamma distributions if needed.

mentekid: I replaced trigamma(.) with polygamma(2, .) to see if that works. According to [this](http://mathworld.wolfram.com/PolygammaFunction.html), I could avoid including it completely and compute it using digamma. I'm trying that while travis attempts to build the new version.

rcurtin: Sounds good.  Either way, it is probably worth checking that the functions you are using are available in our earliest-supported version of Boost (1.49).

mentekid: Yep, my mistake, it was added on Boost 1.58. I misunderstood the article I linked before, when scanning it I thought there was a relationship so you could compute trigamma(x) = f(digamma(x)) but instead it is trigamma(x) = f(trigamma(x-1)).

I think the best option is to backport the source code from boost... I'll start examining how to do that, hopefully it will be as straightforward as copying a file...

rcurtin: Take a look in `src/mlpack/core/boost_backport/`, it should be helpful.

mentekid:
Let's see if it was as simple as that...

Pending: Document the backport in [core/boost_backport/README.md](https://github.com/mlpack/mlpack/blob/master/src/mlpack/core/boost_backport/README.md)

mentekid:
So far I have backported these files:
```
bernoulli.hpp
math_fwd.hpp
policy.hpp
polygamma.hpp
trigamma.hpp
detail/bernoulli_details.hpp
detail/polygamma.hpp
detail/unchecked_bernoulli.hpp
```

In some of these (`bernoulli.hpp`, `detail/unchecked_bernoulli.hpp`, `trigamma.hpp`) the file `math_fwd.hpp` was included, which contains forward declarations. This was supposed to stop the "xyz is not a member of boost::math::..." errors, but in many cases it didn't. I don't understand why, since that file did contain all the forward declarations these files are unable to see.

Anyway, I started hard-coding the declarations by hand at the top of files. This resolved a number of errors, I'll keep doing it to see if I can end up with a passing build.

If, though, someone has any insight as to why this is happening, I would be very happy to implement a simpler solution, both because my fix is ugly and because I'm not sure it will ever work.

mentekid:
So, I've solved almost all the problems except for `‘digits_base10’ is not a member of ‘boost::math::policies’` in `boost_backport/detail/polygamma.hpp`.

I can't do the same thing for this function (declare it on the top of the file) because I get the error `BOOST_MATH_CONSTEXPR does not name a type` (the definition of `digits_base10` specifies that as the return type).

I thought of moving the backported files to the top of my includes so that, if files like boost's `policy.hpp` are included they will be ignored because of the ifndef shield. That obviously didn't work.

I'm not sure how to fix that (or if there are other bugs after that)

mentekid:
Seems that simply removing `BOOST_MATH_CONSTEXPR` made travis build this. I'm not sure if the Appveyor error is related to my changes

mentekid:
This fixed it for version 1.58.0 and newer. If it didn't mess travis up again, I'd say this is ready to merge from a technical standpoint (I'll look at style issues now)

src/mlpack/core/dists/gamma_distribution.hpp
```diff
@@ -0,0 +1,90 @@
...
+    /* Get alpha parameters of each dimension */
+    arma::Col<double>& Alpha(void) { return alpha; };
+    /* Get beta parameters of each dimension */
+    arma::Col<double>& Beta(void) { return beta; };
+
+  private:
+    arma::Col<double> alpha; // Array of fitted alphas.
+    arma::Col<double> beta; // Array of fitted betas.
+    arma::mat referenceSet; // Matrix of reference set.
+    double tol; // Convergence tolerance.
```
rcurtin: I think both `referenceSet` and `tol` should be parameters to `Train()`, because they're only used for training.  I think it makes more sense if the class itself only holds the arrays of alphas and betas.

src/mlpack/core/dists/gamma_distribution.hpp
```diff
@@ -0,0 +1,90 @@
...
+
+#ifndef _MLPACK_CORE_DISTRIBUTIONS_GAMMA_DISTRIBUTION_HPP 
+#define _MLPACK_CORE_DISTRIBUTIONS_GAMMA_DISTRIBUTION_HPP
+
+#include <mlpack/core.hpp>
+namespace mlpack{
+namespace distribution{
+
+/**
+ * Class for fitting the Gamma Distribution to a dataset.
```
rcurtin: We should point out here that this is a multidimensional Gamma distribution, but it assumes that each dimension is fully independent.

src/mlpack/core/dists/gamma_distribution.cpp
```diff
@@ -0,0 +1,76 @@
...
+  alpha.set_size(fittingSet.n_rows);
+  beta.set_size(fittingSet.n_rows);
+
+  // Treat each dimension (i.e. row) independently.
+  for (size_t row = 0; row < fittingSet.n_rows; ++row)
+  {
+    // Calculate log(mean(x)) and mean(log(x)) required for fitting process.
+    const double meanLogx = arma::mean(arma::log(fittingSet.row(row)));
+    const double meanx = arma::mean(fittingSet.row(row));
+    const double logMeanx = std::log(arma::mean(meanx));
```
rcurtin: Hm, if you are willing to pay a little more memory to calculate the means of each dimension all at once, instead of one row at a time, I think that you can get some speedup here, since row-wise access is suboptimal (but accessing the full matrix to calculate the mean should be fine).  Then inside this loop where you estimate parameters for each dimension, you could just use the relevant mean/log-mean/mean-log from the vector that was calculated.

rcurtin:
Thanks for taking the time to figure out the Boost backporting situation.  The tests look good and the implementation seems good too (I think maybe there are a few small speedups we can do).

Do you have time to implement the other functions used by other distributions?  Specifically it would be `Probability(arma::vec&)`, `Random()`, `Train(observations, probabilities)` (that's the hard one, since we have to factor in different weights for each point; at that point we'd be deriving an original algorithm for estimation, but I think it might be a fairly straightforward modification), `Serialize()`, and `LogProbability()` (plus tests).  If you don't, that's fine, but then I would ask if you can instead just open an issue mentioning that each of these methods needs to be implemented for `GammaDistribution` to be a true distribution that can be used in the rest of mlpack. :)

src/mlpack/core/dists/gamma_distribution.hpp
mentekid: Ok, so the object shouldn't own the dataset. That makes sense I think, you're right

mentekid:
>Do you have time to implement the other functions used by other distributions? Specifically it would be Probability(arma::vec&), Random(), Train(observations, probabilities) (that's the hard one, since we have to factor in different weights for each point; at that point we'd be deriving an original algorithm for estimation, but I think it might be a fairly straightforward modification), Serialize(), and LogProbability() (plus tests). If you don't, that's fine, but then I would ask if you can instead just open an issue mentioning that each of these methods needs to be implemented for GammaDistribution to be a true distribution that can be used in the rest of mlpack. :)

I think I prefer to move to the rest of the LSH model for now and come back to do these once that's done, so I'll go with opening an issue if that's ok (I'll assign myself).

PR #730, opened 2016-07-20
Fixing vector classes to comply with method calls.

sbrodehl:
Changed vector classes of 'assignments' to comply with the corresponding method calls.
Fixes compile errors for those samples.


rcurtin:
Thank you, this is very helpful!  If you see any more fixes, do not hesitate to open another PR.  It is very difficult to keep documentation perfectly correct (since we currently don't have an easy way to compile and test any of our examples).

PR #731, opened 2016-07-20
Fix memory leaf when dual tree search and treeOwner == true

MarcosPividori:


rcurtin:
Nice catch.  I also applied the fix to RangeSearch and RASearch in 9dc7de5.

PR #732, opened 2016-07-22
Heaps for mlpack!

MarcosPividori:
Hi,
I realized in many parts of the code, we keep track of  the best k candidates visited.
In this situation, it is not necessary to keep a sorted list of the k elements. We only need to know the value of the worst candidate, so we can compare it with new elements and know if we must insert them to this list.
So, instead of maintaining a sorted list, a better approach is to use a priority queue.
When we finish looking for the best k candidates, we can easily get a sorted list from the priority queue, taking the top elements and saving them in reverse order.
If we use a greater value of k, the difference in performance is greater and we have a considerably faster algorithm.
Thanks!
Marcos


src/mlpack/methods/cf/cf.hpp
```diff
@@ -258,22 +258,24 @@ class CF
...
+    Candidate(double value, size_t item) :
+        value(value),
+        item(item)
+    {};
+    //! Compare the value of two candidates.
+    friend bool operator>(const Candidate& l, const Candidate& r)
+    {
+      return l.value > r.value;
+    };
+  };
```
rcurtin: Is there a compelling reason to avoid `std::pair<>` here?

rcurtin:
This looks like a nice change and should unify the code in a bunch of places.  Have you done any timing simulations to see if this speeds anything up?  Typical values for k for nearest neighbor search should be between 1 and 10... I don't think it's a common case to use k = 1000 or anything like that.

src/mlpack/methods/cf/cf.hpp
MarcosPividori: @rcurtin  , yes I have considered std::pair<> but I thought it would be clearer if we have specific structure for this (maybe value/item is clearer than first/second...). Of course, I can use std::pair<> if you think it would be better :)
However, I can see a problem with the relational operators of std::pair.
They compare the first element and, if these elements are equal, they compare the second element. So, I think we should set "first" to represent the value, and "second" to represent the index.
When we decide if inserting a new element to the list of candidates, we make a comparison with the top element: "if (c > pqueue.top())"
If we use a pair, it will insert c in the case where:

c.first == pqueue.top().first && c.second > pqueue.top().second

This is an unnecesary overhead, because we don't care on the order of the indexes, so we are inserting a new element with the same value but a greater index...

We could easily fix this by explicitly comparing the first element: "if (c.first > pqueue.top().first)"

However, I can see another possible disadvantage. Inside the implementation of the priority queue, if we use the std::pair relational operator, elements will be not only sorted by value but also by index. This can cause a bit more overhead when shifting down/up through the heap.

Maybe I became a bit obsessed with efficiency and this doesn't make a real difference...

rcurtin: Yep, your concerns are 100% valid.  Take a look at #712 where @lozhnikov and I researched different solutions to this same problem.  `std::priority_queue` has a third template parameter---`Compare`---which you could use to specify either a lambda or a custom function for comparison, which would only compare the first element in the pair.

MarcosPividori: Thanks for sharing that information. I can see you found a real difference in that issue!
Yeah, I use the `Compare` template parameter when deciding between a max heap and min heap.
So, what I don't understand is:  Do you think we can have a real difference between using std::pair with lambda comparison,  or using the Candidate struct with the custom operator< ? Even if the code seems to be the same? Is there any special optimization for std::pair?

MarcosPividori:
Hi @rcurtin , I have run some tests with *mlpack_knn* and the *"Isolet"* dataset. It looks like it slightly faster when k is small, and the difference is greater for bigger values of k.
Maybe, we should benchmark with bigger datasets like *"tinyImages100k"*. Also, it would be interesting to see the effect in other methods, like: *mlpack_cf*, *mlpack_lsh*, *mlpack_fastmks*, *mlpack_krann*, etc.


# Naive:
 + **No-heap:**
  - k=3: total_time: 32.790772s
  - k=10: total_time: 32.946072s
  - k=20: total_time: 32.869260s
  - k=100: total_time: 33.801901s
  - k=1000: total_time: 52.723884s

 + **Heap:**
  - k=3: total_time: 31.996061s
  - k=10: total_time: 32.623913s
  - k=20: total_time: 32.756599s
  - k=100: total_time: 33.620033s
  - k=1000: total_time: 44.275435s 

# Single-tree search:
 + **No-heap:**
  - k=3: total_time: 19.419358s
  - k=10: total_time: 21.696434s
  - k=20: total_time: 23.501047s
  - k=100: total_time: 29.048157s
  - k=1000: total_time: 57.329319s

 + **Heap:**
  - k=3: total_time: 19.386155s
  - k=10: total_time: 21.643011s
  - k=20: total_time: 23.270618s
  - k=100: total_time: 29.031573s
  - k=1000: total_time: 50.519661s

# Dual-tree search:
 + **No-heap:**
  - k=3: total_time: 15.207993s
  - k=10: total_time: 17.129599s
  - k=20: total_time: 18.325870s
  - k=100: total_time: 22.422410s
  - k=1000: total_time: 48.341919s

 + **Heap:**
  - k=3: total_time: 15.203970s
  - k=10: total_time: 16.998384s
  - k=20: total_time: 18.240207s
  - k=100: total_time: 22.401873s
  - k=1000: total_time: 41.206338s

rcurtin:
Nice!  Speedup in all conditions makes a decision easy.  I don't think it's necessary to test with other datasets, because since you've both tried many values of k (except k=1... what happens there?) and also in situations where more candidates are inserted (naive search) and where fewer candidates are inserted (dual-tree search).

MarcosPividori:
@rcurtin Ok great!
I added the case when k=1, it seems to show the same difference.

MarcosPividori:
`RAModelTest` fails in Travis, but I have run it many times in my ubuntu machine and I don't see any problem here... Maybe, it is failing because of an external situation...

rcurtin:
Great to hear that it's faster for k=1, too.

It's funny, this is kind of things coming full circle... a very long time ago, the NeighborSearch code (or AllkNN as the class was called) used to use `std::pair<>` to represent candidates:

https://github.com/mlpack/mlpack/blob/43ac2c8a0f165aecb91ef5a5e8f9e0e7e0b3bb55/fastlib/trunk/mlpack/allknn/allknn.h#L375
(it's crazy how different the code was then... I forget how far things have come.  It's fun to browse around that repository and contrast with now)

But it used `std::vector<std::pair<>>` and then used `std::sort()` to sort them and remove the last element, instead of the much more reasonable priority queue.  When I replaced that (back in 2011 sometime) with individual vectors for the neighbors and distances, I achieved an order-of-magnitude speedup or more, especially for large k.  Now we are back to using `std::pair<>` (or similar), but this time with the right data structure---a heap. :)

rcurtin:
Also, the RAModelTest failure is not your fault.  It seems like the rank-approximate nearest neighbor search is not satisfying guarantees for every type of tree.  I should probably double the tolerance and open a ticket about it... I will do that momentarily...

MarcosPividori:
Yes, that code looks really different to the actual version!

src/mlpack/methods/cf/cf.hpp
MarcosPividori: @rcurtin , I have tested the modification to use std::pair in https://github.com/MarcosPividori/mlpack/tree/pair-for-heaps  and I get the same runtime than when using Candidate structure.

rcurtin:
Should I go ahead and merge from the `pair-for-heaps` branch?

MarcosPividori:
Mmm well... as we get the same runtime with std::pair than with Candidate struct, I think it's better to use Candidate struct, would you agree? So, I would prefer `heap-for-knn` branch.

src/mlpack/methods/cf/cf.hpp
MarcosPividori: @rcurtin I think you see a difference in runtime between std::pair and SortStruct in: #712  , because you use a pointer to function comparison parameter instead of a functor or an operator, as I can see in: https://github.com/mlpack/mlpack/pull/721/commits/640f3a7859ea8d400bc6fed6b4395c2291ba0259#diff-3a0c62f44f497b9857aed01fe81dc391L54
I have read some documentation that says that functors can be inlined while pointer to function generally are not.
For Candidate struct, I define an operator, that probably will be inlined by the compiler.

rcurtin:
I thought that `std::pair<>` would be better to use here since it is a more well-known class, and allows us to avoid declaring the `Candidate` class many different places in the code.  Do you think there is a compelling reason to keep the `Candidate` class?  If the thought is that the `Candidate` name makes it clear, we can just add a comment like this:

```
// Holds an (index, score) pair.
std::pair<size_t, double> candidate;
```

MarcosPividori:
@rcurtin  ok! I will complete the `pair-for-heaps` branch to replace the Candidate struct in other methods, and I will rebase it into this branch.

src/mlpack/methods/cf/cf.hpp
rcurtin: That makes sense---there is not much reason to believe that `std::pair<>` is intrinsically faster than a handwritten class that implements the same functionality.  My leaning towards `pair` is primarily a function of minimizing code, but it is not a huge preference. :)

MarcosPividori:
Hi @rcurtin ,
I replaced Candidate with std::pair.
Also, I modified SortPolicy::IsBetter() to return false when the values are equal. This can improve the efficiency, because more nodes can be prunned, and makes code simpler :)
Thanks

rcurtin:
Ah, unfortunately, we can't reliably modify `IsBetter()` in that manner: https://github.com/mlpack/mlpack/commit/3781ccbe93bb4cb886978b4c865ff9b42f5dbc56

MarcosPividori:
@rcurtin
Yes, you are right, that happens because of B_2 bound. Sorry for that confusion.
I have removed that commits.

src/mlpack/methods/fastmks/fastmks_rules_impl.hpp
```diff
@@ -46,6 +45,40 @@ FastMKSRules<KernelType, TreeType>::FastMKSRules(
...
+  indices.set_size(k, querySet.n_cols);
+  products.set_size(k, querySet.n_cols);
+
+  for (size_t i = 0; i < querySet.n_cols; i++)
+  {
+    CandidateList& pqueue = candidates[i];
+    typedef typename CandidateList::iterator Iterator;
+
+    for (Iterator end = pqueue.end(); end != pqueue.begin(); --end)
+      std::pop_heap(pqueue.begin(), end, CandidateCmp());
```
rcurtin: Would it be better to use `sort_heap()` here, since we don't need the heap properties anymore?

src/mlpack/methods/fastmks/fastmks_rules.hpp
```diff
@@ -101,10 +124,28 @@ class FastMKSRules
...
+    {
+      return c1.first > c2.first;
+    };
+  };
+
+  //! Use a min heap to represent the list of candidate points.
+  //! We will use a vector and the std functions: push_heap() pop_heap().
+  //! We can not use a priority queue because we need to iterate over all the
+  //! candidates and std::priority_queue doesn't provide that interface.
+  typedef std::vector<Candidate> CandidateList;
```
rcurtin: Do you think it would be better to use a `boost::priority_queue` here?  We use it in the `CosineTree` code for this exact reason.
http://www.boost.org/doc/libs/1_55_0/doc/html/heap.html

src/mlpack/methods/fastmks/fastmks_rules_impl.hpp
MarcosPividori: Ok, I agree. It will be clearer.

src/mlpack/methods/fastmks/fastmks_rules.hpp
MarcosPividori: @rcurtin Yes, I agree!  When updating fastmks's code, I searched in the documentation for a iterable priority queue and I didn't realize boost::heap::priority_queue provides that interface. I have updated the code in https://github.com/mlpack/mlpack/pull/732/commits/8e1fbf0f413d9cae3a9d88b275afe6d20def2fca
Thanks!

MarcosPividori:
"appveyor error" seems to be an external error.

rcurtin:
Yeah:

```
System.Net.WebException: The underlying connection was closed: A connection that was expected to be kept alive was closed by the server. ---> System.IO.IOException: Unable to read data from the transport connection: An existing connection was forcibly closed by the remote host. ---> System.Net.Sockets.SocketException: An existing connection was forcibly closed by the remote host
```

I seem to see that frequently; I'll restart the build.

MarcosPividori:
Hi @rcurtin , I have run some tests with the last version. (*mlpack_knn* and the *"Isolet"* dataset)

# Naive:
+ **No-heap:**
  - k=1: total_time: 31.567688s
  - k=3: total_time: 31.635909s
  - k=10: total_time: 31.425726s
  - k=100: total_time: 32.611729s
  - k=1000: total_time: 52.003092s

+ **Heap:**
  - k=1: total_time: 31.115321s
  - k=3: total_time: 31.344317s
  - k=10: total_time: 31.384044s
  - k=100: total_time: 32.195979s
  - k=1000: total_time: 42.065338s

# Single-tree search:
+ **No-heap:**
  - k=1: total_time: 15.103463s
  - k=3: total_time: 18.415424s
  - k=10: total_time: 20.861882s
  - k=100: total_time: 27.718015s
  - k=1000: total_time: 56.549890s

+ **Heap:**
  - k=1: total_time: 14.992397s
  - k=3: total_time: 18.360664s
  - k=10: total_time: 20.692259s
  - k=100: total_time: 27.684020s
  - k=1000: total_time: 49.233336s

# Dual-tree search:
+ **No-heap:**
  - k=1:
    + total_time: 12.826351s
    + total_time: 12.843755s
    + total_time: 12.890680s
  - k=3:
    + total_time: 15.096998s
    + total_time: 15.118167s
    + total_time: 15.162257s
  - k=10:
    + total_time: 17.026362s
    + total_time: 16.981066s
    + total_time: 16.955594s
  - k=100:
    + total_time: 22.345687s
    + total_time: 22.457659s
    + total_time: 22.316181s
  - k=1000:
    + total_time: 47.140635s
    + total_time: 47.342563s
    + total_time: 48.636944s

+ **Heap:**
  - k=1:
    + total_time: 12.815268s
    + total_time: 12.835969s
    + total_time: 12.806319s
  - k=3: 
    + total_time: 15.067807s
    + total_time: 15.079179s
    + total_time: 15.083814s
  - k=10:
    + total_time: 16.871578s
    + total_time: 16.909372s
    + total_time: 16.875759s
  - k=100:
    + total_time: 22.247824s
    + total_time: 22.259244s
    + total_time: 22.243977s
  - k=1000:
    + total_time: 41.227029s
    + total_time: 41.253371s
    + total_time: 41.720898s

rcurtin:
Thanks!  Some of the includes were unnecessary; I removed them in b12e151.

MarcosPividori:
@rcurtin Ok, Thanks!

PR #735, opened 2016-07-24
Optimize load csv 01

stereomatchingkiss:
use fast csv parser to parse csv file(without mapping).

Original implementation : around 7220ms
fast csv : around 600ms

Edit : 
The file have 1 million rows and 7 cols


rcurtin:
Great, this is a nice prototype.  One thing to think about is, how adaptable will the fast CSV loader be to other formats, like ARFF, txt, matrix market, and anything else we might choose to support in the future?  Is there a way we can adapt the code and make it flexible?

Another important thing to consider is this: right now, we depend on Armadillo for reading CSVs, and as a result don't really test the CSV parser in great detail.  i.e. we assume that the Armadillo CSV reader doesn't have bugs.  If we switch to boost::spirit we need to test that our usage of boost::spirit is correct (but we can assume the internals of boost::spirit are tested); if we switch to the fast CSV reader here and modify it by including the source, we can no longer assume that, so we either need to write our own tests for everything, since the fast CSV library doesn't have any tests I can see.

What do you think?

I know we are trying many different things here but I think we can figure something out in the end and it will be a big improvement.  Thanks again for the time you are putting into this adventure. :)

stereomatchingkiss:
>how adaptable will the fast CSV loader be to other formats, like ARFF, txt, matrix market, and anything else we might choose to support in the future?

They could be supported in the future.

>so we either need to write our own tests for everything, since the fast CSV library doesn't have any tests I can see.

I am planning on writing some tests for csv parser too, maybe on this week.

This is a prototype, showing one of the possible solution to optimize loading time. There are still many things to do before it really work.

stereomatchingkiss:
Same error keep pop up, do anyone have ideas what is going on?Thanks

```
/home/travis/build/mlpack/mlpack/src/mlpack/../mlpack/core/data/file_reader/parser.hpp:118:9: error: ‘struct mlpack::io::ThrowOnOverflow’ has no member named ‘OnOverFlow’

         OverFlowPolicy().template OnOverFlow<T>(x);
```

rcurtin:
If this is ready for review, let me know and I will take a look through it.  It seems like it builds correctly now.

stereomatchingkiss:
Not yet, I am studying the format of ARFF and checking the coding style of
the codes.
Original coding style do not fit to mlpack

2016-09-07 2:39 GMT+08:00 Ryan Curtin <notifications@github.com>:

stereomatchingkiss:
I think this pull request is ready for review.

stereomatchingkiss:
After discussion, we prefer implementation based on boost spirit

PR #736, opened 2016-07-24
added functionality to pool with stride

nilayjain:


zoq:
Thanks for the contribution, great to see that we are now able to use different stride factors. I made some minor changes in 496f1e48475e5.

PR #737, opened 2016-07-24
fixes: convolve with correct weight slice, pad input in forward pass

nilayjain:


src/mlpack/tests/conv_layer_test.cpp
```diff
@@ -0,0 +1,78 @@
+ * @author Nilay Jain
```
zoq: Can you add ``/*`` so that we can automatically test the code? For more information here is the build log: https://travis-ci.org/mlpack/mlpack/builds/150478977

src/mlpack/tests/conv_layer_test.cpp
```diff
@@ -0,0 +1,78 @@
...
+    {
+      input.slice(i).row(j) = a.subvec(vec_idx, vec_idx + 4).t();
+      vec_idx += 5;
+    }
+  }
+  ConvLayer<> conv3(4, 1, 3, 3, 1, 1, 1, 1);
+  arma::mat id3 = arma::zeros<arma::mat>(3, 3);
+  id3(1, 1) = 1;  
+  arma::cube conv3_w(3, 3, 4 * 1);
+  for (size_t i = 0; i < conv3_w.n_slices; ++i)
```
zoq: We can write ``arma::cube conv3_w = arma::zeros<arma::cube>(3, 3, 4);`` instead of iterating over the slices. Also case you use camel casing for all names. Take a look at: https://github.com/mlpack/mlpack/wiki/DesignGuidelines#naming-conventions for more informations.

src/mlpack/tests/conv_layer_test.cpp
```diff
@@ -0,0 +1,78 @@
...
+  ConvLayer<> conv3(4, 1, 3, 3, 1, 1, 1, 1);
+  arma::mat id3 = arma::zeros<arma::mat>(3, 3);
+  id3(1, 1) = 1;  
+  arma::cube conv3_w(3, 3, 4 * 1);
+  for (size_t i = 0; i < conv3_w.n_slices; ++i) 
+    conv3_w.slice(i) = id3;
+  conv3.Weights() = conv3_w;
+  
+  arma::cube output(5, 5, 1);
+  output.slice(0) = input.slice(0) * 4;
```
zoq: When specifying the size with n_rows and n_cols, n_slices (5, 5, 1), by default the memory is uninitialised. So, I'm not sure the test does the right thing here.

src/mlpack/tests/conv_layer_test.cpp
zoq: I think, we should use a nonzero kernel for the test.

src/mlpack/tests/conv_layer_test.cpp
```diff
@@ -0,0 +1,78 @@
...
+  Test(conv3.Delta(), arma::ones(5, 5, 4));
+
+  arma::cube delta = arma::zeros(5, 5, 1);
+  delta(2, 2, 0) = 1;
+  conv3.Gradient(conv3.InputParameter(), delta, conv3.Gradient());
+  arma::cube gradOut(3, 3, 4);
+  arma::mat grad_w;
+  grad_w << 7 << 8 << 9 << arma::endr
+        << 12 << 13 << 14 << arma::endr
+        << 17 << 18 << 19;
```
zoq: The rest looks good, thanks!

src/mlpack/methods/ann/layer/conv_layer.hpp
```diff
@@ -125,8 +131,10 @@ class ConvLayer
         arma::Mat<eT> output;
         BackwardConvolutionRule::Convolution(gy.slice(inMap), rotatedFilter,
             output);
-
-        g.slice(outMap) += output;
+        g.slice(outMap) += output.submat(rotatedFilter.n_rows / 2,
```
zoq: We only have to use submat if ``(wPad != 0 || hPad != 0)``.

src/mlpack/tests/conv_layer_test.cpp
nilayjain: Right now we are using identity kernel for the test. Do you still think we should use some other kernel?

src/mlpack/tests/conv_layer_test.cpp
nilayjain: in the next line(51) we are initializing output.slice(0) to input.slice(0) * 4, and then testing whether we get this output.

src/mlpack/methods/ann/layer/conv_layer.hpp
nilayjain: I didn't put an if there because if we don't put a padding the submat would take the whole matrix. But putting if makes that point explicit. Thanks.

src/mlpack/methods/ann/layer/conv_layer.hpp
```diff
@@ -126,7 +132,11 @@ class ConvLayer
         BackwardConvolutionRule::Convolution(gy.slice(inMap), rotatedFilter,
             output);
 
-        g.slice(outMap) += output;
+        if (wPad != 0 || hPad != 0)
```
zoq: The gradient isn't set if ``(wPad != 0 || hPad != 0)`` isn't true. In case ``(wPad != 0 || hPad != 0)``is false the gradient is the output value.

zoq: The gradient isn't set if ``(wPad != 0 || hPad != 0)`` isn't true. In case ``(wPad != 0 || hPad != 0)``is false the gradient is the output value.

src/mlpack/tests/conv_layer_test.cpp
zoq: Ah, right, thanks for the clarification.

src/mlpack/tests/conv_layer_test.cpp
zoq: I think, it's reasonable to use the identity kernel, if you like you can test it against another kernel.

nilayjain:
I am back now. Will finish the things promised in the next days. Thank you for letting me do so.

nilayjain:
I will add the Pad() functions in the network_util file and remove them from the conv_layer.hpp in the PR for googlenet. That is still a few days away though.

zoq:
Hello @nilayjain glad to see you back, I've opened a PR #831 that updates the nn structure which hopefully makes things easier to use. Would be great if you could take a look.

nilayjain:
Hi @zoq , Looks like a lot of changes made. Great Job. Will look at it.

zoq:
I think all issues were fixed in 7811b7ce7b7d5a46e44076, feel free to comment on the PR if you think I missed something.

PR #739, opened 2016-07-25
Fixes off-by-one in boost_backport.hpp for boost 1.55.0

mentekid:
I think that was the mistake in the previous PR.
Sorry!


rcurtin:
Yep, this was wrong all along, but now that you have fixed the boost backports to define `BOOST_VERSION`, the bug shows up, and your patch here fixes the issue.

PR #741, opened 2016-07-27
Fixed an error in MidpointSplit that may lead to a segfault

lozhnikov:
I tried to compare the vantage point tree with `BallTree` and came across a segfault.

Since `BallBound[d].Mid()` is equal to the center, one child may be empty. In this case the following loop may lead to a segfault (if `right = 0` then `right-1 = SIZE_MAX`)
```
 while ((data(splitDimension, right) >= splitVal) && (left <= right))
   right--;
```

### DIFF
--- a/src/mlpack/core/tree/binary_space_tree/midpoint_split_impl.hpp
+++ b/src/mlpack/core/tree/binary_space_tree/midpoint_split_impl.hpp
@@ -78,9 +78,7 @@ bool MidpointSplit<BoundType, MatType>::SplitNode(const BoundType& bound,
     delete[] ranges;
   }
 
-  assert(splitVal != DBL_MAX);
-
-  if (maxWidth == 0) // All these points are the same.  We can't split.
+  if (maxWidth <= 0) // All these points are the same.  We can't split.
     return false;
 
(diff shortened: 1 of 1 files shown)


src/mlpack/core/tree/binary_space_tree/midpoint_split_impl.hpp
```diff
@@ -65,18 +69,20 @@ bool MidpointSplit<BoundType, MatType>::SplitNode(const BoundType& bound,
...
       }
     }
 
     delete[] ranges;
   }
 
+  assert(splitVal != DBL_MAX);
+
   if (maxWidth == 0) // All these points are the same.  We can't split.
     return false;
```
rcurtin: Hm, I think we could avoid the assert here (or handle it better) if we just made the conditional `if (maxWidth <= 0)`; what do you think?  Then no split would be performed.

lozhnikov: You're right, that is much better.

rcurtin:
Thanks.  I checked that this is not a problem in `MeanSplit` too.

PR #742, opened 2016-07-27
Descriptive Statistics command-line program

keonkim:
I originally built a [class](https://github.com/keonkim/mlpack/commit/c2f5c5c2e6cbce084992629e192023519873e4cb) that calculates descriptive statistics. But after a few discussion, I ended up shrinking all of the functions down to minimum to provide maximum performance and maintainability.
I also merged all commits to one to discard unnecessary commits.

Sample output on "iris.csv" would be:
```
[INFO ] dim     var     mean    std     median  min     max     range   skew    kurt    SE      
[INFO ] 0       0.6857  5.8433  0.8281  5.8000  4.3000  7.9000  3.6000  0.3149  -0.5521 0.0676  
[INFO ] 1       0.1880  3.0540  0.4336  3.0000  2.0000  4.4000  2.4000  0.3341  0.2908  0.0354  
[INFO ] 2       3.1132  3.7587  1.7644  4.3500  1.0000  6.9000  5.9000  -0.2745 -1.4019 0.1441  
[INFO ] 3       0.5824  1.1987  0.7632  1.3000  0.1000  2.5000  2.4000  -0.1050 -1.3398 0.0623  
```
Users can control the width and precision using -w and -p flag.
I tested the output using excel and they match perfectly.


src/mlpack/methods/preprocess/preprocess_describe_main.cpp
```diff
@@ -0,0 +1,231 @@
...
+double SumNthPowerDeviations(const arma::rowvec& input,
+    const double& rowMean,
+    const size_t Nth) // Degree of Power
+{
+  double sum = 0;
+  for (size_t i = 0; i < input.n_elem; ++i)
+  {
+    sum += pow(input(i) - rowMean, Nth);
+  }
+  return sum;
```
stereomatchingkiss: This one could be vectorize, which performance is better need to measure, but it should make the codes shorter and easier to read.

```
return arma::sum(arma::pow(input - rowMean,
                             static_cast<double>(Nth)));
```

src/mlpack/methods/preprocess/preprocess_describe_main.cpp
```diff
@@ -0,0 +1,231 @@
...
+    const bool population)
+{
+  double skewness = 0;
+  double S3 = pow(rowStd, 3);
+  double M3 = SumNthPowerDeviations(input, rowMean, 3);
+  double n = input.n_elem;
+  if (population)
+  {
+    // Calculate Population Skewness
+    skewness = n * M3 / (n * n * S3);
```
stereomatchingkiss: why not just use

`skewness = M3/(n*S3);`

stereomatchingkiss: Why not just write

`skewness = M3 / (n*S3);`

Besides, according to the [formula ](http://www.macroption.com/skewness-formula/)at this page, I think the equation of population skewness should be

std::sqrt(n) * M3 / S3

stereomatchingkiss: Why not just write

`skewness = M3 / (n*S3);`

Besides, according to the formula at this [page](http://www.macroption.com/skewness-formula/), I think the equation of population skewness should be

`std::sqrt(n) * M3 / (n*S3);`

src/mlpack/methods/preprocess/preprocess_describe_main.cpp
```diff
@@ -0,0 +1,231 @@
...
+ * @return Skewness of the given vector.
+ */
+double Skewness(const arma::rowvec& input,
+    const double& rowStd,
+    const double& rowMean,
+    const bool population)
+{
+  double skewness = 0;
+  double S3 = pow(rowStd, 3);
+  double M3 = SumNthPowerDeviations(input, rowMean, 3);
```
stereomatchingkiss: I guess M3 mean moment 3?If so, I think write it as
`M3/n` maybe can express your intention better

src/mlpack/methods/preprocess/preprocess_describe_main.cpp
```diff
@@ -0,0 +1,231 @@
...
+  double n = input.n_elem;
+  if (population)
+  {
+    // Calculate Population Skewness
+    skewness = n * M3 / (n * n * S3);
+  }
+  else
+  {
+    // Calculate Sample Skewness
+    skewness = n * M3 / ((n-1) * (n-2) * S3);
```
stereomatchingkiss: As the `http://www.macroption.com/skewness-formula/`mention, the equation should be 

`(n*std::sqrt(n-1)/std::sqrt(n-2)) * std::sqrt(n) * M3 / (n*S3)`

Please correct me if I am wrong

src/mlpack/methods/preprocess/preprocess_describe_main.cpp
```diff
@@ -0,0 +1,231 @@
...
+  return kurtosis;
+}
+/**
+ * Calculates standard error of standard deviation.
+ *
+ * @param input Vector that captures a dimension of a dataset
+ * @param rowStd Standard Deviation of the given vector.
+ * @return Standard error of the stanrdard devation of the given vector.
+ */
+double StandardError(const arma::rowvec& input, const double rowStd)
```
stereomatchingkiss: I think we could replace rowvec by size_t, because this could reduce dependency on specific type.
In medium or large size project, reduce dependency of your api can make your codes easier to maintain.

ps : This is not a big problem in this small CLI program

src/mlpack/methods/preprocess/preprocess_describe_main.cpp
```diff
@@ -0,0 +1,231 @@
...
+ * @param rowMean Mean of the given vector.
+ * @return Skewness of the given vector.
+ */
+double Skewness(const arma::rowvec& input,
+    const double& rowStd,
+    const double& rowMean,
+    const bool population)
+{
+  double skewness = 0;
+  double S3 = pow(rowStd, 3);
```
stereomatchingkiss: good candidate for const

src/mlpack/methods/preprocess/preprocess_describe_main.cpp
```diff
@@ -0,0 +1,231 @@
...
+          % rowMean
+          % rowStd
+          % arma::median(row)
+          % rowMin
+          % rowMax
+          % (rowMax - rowMin) // range
+          % Skewness(row, rowStd, rowMean, population)
+          % Kurtosis(row, rowStd, rowMean, population)
+          % StandardError(row, rowStd) << endl;
+    }
```
stereomatchingkiss: We can reduce some duplicate codes

```
if(CLI::HasParam("dimension")){
    printStatResults(dimension);
}else{
    for(size_t i = 0; i < data.n_rows; ++i){
	    printStatResults(i);
	}
}
```

stereomatchingkiss:
What if users want to measure non transpose data?

stereomatchingkiss:
This CLI assume every column equal to one sample, what if the data is every row equal to one sample?

src/mlpack/methods/preprocess/preprocess_describe_main.cpp
```diff
@@ -0,0 +1,231 @@
...
+/**
+* Calculates the sum of deviations to the Nth Power
+*
+* @param input Vector that captures a dimension of a dataset
+* @param rowMean Mean of the given vector.
+* @return sum of nth power deviations
+*/
+double SumNthPowerDeviations(const arma::rowvec& input,
+    const double& rowMean,
+    const size_t Nth) // Degree of Power
```
rcurtin: For consistency with the style guide, I'd name this parameter `n` (lowercase) instead of `Nth`.

src/mlpack/methods/preprocess/preprocess_describe_main.cpp
```diff
@@ -0,0 +1,231 @@
...
+        % arma::var(row, population)
+        % rowMean
+        % rowStd
+        % arma::median(row)
+        % rowMin
+        % rowMax
+        % (rowMax - rowMin) // range
+        % Skewness(row, rowStd, rowMean, population)
+        % Kurtosis(row, rowStd, rowMean, population)
+        % StandardError(row, rowStd) << endl;
```
rcurtin: Everywhere else in the mlpack code we use just regular iostreams instead of boost::format... is it possible to use iostreams and iomanips here?  Like `std::setw`, etc.

src/mlpack/methods/preprocess/preprocess_describe_main.cpp
```diff
@@ -0,0 +1,231 @@
...
+  // Load the data
+  arma::mat data;
+  data::Load(inputFile, data, false, true /*transpose*/);
+
+  // Generate boost format recipe.
+  const string widthPrecision("%-"+
+    to_string(width)+ "." +
+    to_string(precision));
+  const string widthOnly("%-"+
+    to_string(width)+ ".");
```
rcurtin: Be sure to follow the style guide---wrapped lines should be indented twice (not once), and try to place spaces between operators (like `a + b` not `a+b`) for readability. :)

src/mlpack/methods/preprocess/preprocess_describe_main.cpp
keonkim: @rcurtin sorry for late response.
I tried with `std::setw` first, but I found that `std::setprecision` is not compatible with Log::Info.
So the `boost::format` alternative was chosen.  It is not fast, but I thought this would not cause performance issue in this case.

rcurtin: No problem, I think your solution is perfectly reasonable.  Would you like to open a ticket about the `Log::Info` and `std::setw` incompatibility?  We don't need to solve it now (or soon), but it would maybe be a good task for someone who is looking for ways to contribute, since it's a pretty self-contained and easy-to-replicate problem. :)

src/mlpack/methods/preprocess/preprocess_describe_main.cpp
keonkim: @stereomatchingkiss I imitated the result from [excel](https://support.office.com/en-US/article/SKEW-P-function-76530A5C-99B9-48A1-8392-26632D542FCB)
I guess this is equivalent to "direct skewness formula" in the [page](http://www.macroption.com/skewness-formula/)

src/mlpack/methods/preprocess/preprocess_describe_main.cpp
keonkim: @stereomatchingkiss I imitated the result from [excel](https://support.office.com/en-US/article/SKEW-function-BDF49D86-B1EF-4804-A046-28EAEA69C9FA).
This formula appears near the bottom of this [page](http://www.macroption.com/skewness-formula/)

src/mlpack/methods/preprocess/preprocess_describe_main.cpp
keonkim: updated

src/mlpack/methods/preprocess/preprocess_describe_main.cpp
keonkim: updated

src/mlpack/methods/preprocess/preprocess_describe_main.cpp
```diff
@@ -0,0 +1,231 @@
...
+          % rowStd
+          % arma::median(row)
+          % rowMin
+          % rowMax
+          % (rowMax - rowMin) // range
+          % Skewness(row, rowStd, rowMean, population)
+          % Kurtosis(row, rowStd, rowMean, population)
+          % StandardError(row, rowStd) << endl;
+    }
+  }
```
keonkim: updated

src/mlpack/methods/preprocess/preprocess_describe_main.cpp
keonkim: updated

src/mlpack/methods/preprocess/preprocess_describe_main.cpp
keonkim: updated

src/mlpack/methods/preprocess/preprocess_describe_main.cpp
keonkim: updated

src/mlpack/methods/preprocess/preprocess_describe_main.cpp
keonkim: updated

src/mlpack/methods/preprocess/preprocess_describe_main.cpp
stereomatchingkiss: My fault, sorry for misunderstanding, yes, they are the same thing as you mentioned.

src/mlpack/methods/preprocess/preprocess_describe_main.cpp
stereomatchingkiss: @keonkim  I think you are right, my mistakes, thanks for your correction.

src/mlpack/methods/preprocess/preprocess_describe_main.cpp
```diff
@@ -0,0 +1,226 @@
...
+    const double& fMean,
+    const bool population)
+{
+  double kurtosis = 0;
+  const double M4 = SumNthPowerDeviations(input, fMean, 4);
+  const double n = input.n_elem;
+  if (population)
+  {
+    // Calculate Population Excess Kurtosis
+    double M2 = SumNthPowerDeviations(input, fMean, 2);
```
stereomatchingkiss: I think this is a good candidate of const too, as other variables in this function. 
Sorry of being picky on const correctness.

After this is fixed, I think the codes are ready to merge

keonkim: updated

src/mlpack/methods/preprocess/preprocess_describe_main.cpp
```diff
@@ -0,0 +1,228 @@
...
+/**
+* Calculates the sum of deviations to the Nth Power.
+*
+* @param input Vector that captures a dimension of a dataset.
+* @param rowMean Mean of the given vector.
+* @param n Degree of power.
+* @return sum of nth power deviations.
+*/
+double SumNthPowerDeviations(const arma::rowvec& input,
+    const double& fMean,
```
stereomatchingkiss: Discover a small problem before merged, could you align the paramters?like load.hpp did

keonkim: updated

PR #743, opened 2016-07-28
Use move semantics to set a given reference tree in NeighborSearch class.

MarcosPividori:
Hi,

I have modified NeighborSearch class to take a rvalue reference instead of pointer, when setting a specific referenceTree, for the constructor and Train() method.
I think this is more intuitive. We use these methods when we want to build the reference tree with a special configuration, for example to set the leafSize.
Every time we use these methods, we want to transfer ownership of this tree to the neighbor search class, so I think a rvalue reference is more appropiate.
These changes are not completed, we should add a move constructor for all rtrees's flavours. I wanted to be sure you are interested in this, before spending more time.

Another option is to continue using the pointer to referenceTree and incorporate a boolean parameter like:  "setOwnership".
Personally, I prefer rvalues references for this case.

Thanks!

### DIFF
--- a/HISTORY.md
+++ b/HISTORY.md
@@ -8,6 +8,9 @@
   * Added mlpack_preprocess_describe program, which can be used to print
     statistics on a given dataset (#742).
 
+  * Fix prioritized recursion for k-furthest-neighbor search (mlpack_kfn and the
+    KFN class), leading to orders-of-magnitude speedups in some cases.
+
 ### mlpack 2.0.3
 ###### 2016-07-21
   * Added multiprobe LSH (#691).  The parameter 'T' to LSHSearch::Search() can


rcurtin:
I thought about this for a while.  Trees have generally been passed as pointers, in part because internally trees must hold pointers to their children.  But, I suppose, it is not necessarily reasonable to require that trees be passed as pointers.  So, I can't justify any reason why we shouldn't make this change.  But, there are a couple things we have to consider:

 * What if the user prefers to pass a tree that should be copied?  Maybe we should provide overloads with just a reference instead of an rvalue reference?

 * We have to maintain reverse compatibility, so unfortunately we need to retain the overloads that take `TreeType*`.  We can just have those call the reference/rvalue reference overloads, though.  I think that in general they should call the rvalue-reference overloads since those functions generally took ownership of the tree.

src/mlpack/core/tree/binary_space_tree/binary_space_tree_impl.hpp
```diff
@@ -407,6 +407,12 @@ BinarySpaceTree(BinarySpaceTree&& other) :
   other.furthestDescendantDistance = 0.0;
   other.minimumBoundDistance = 0.0;
   other.dataset = NULL;
+
+  //Set new parent.
+  if (left)
+    left->parent = this;
+  if (right)
+    right->parent = this;
```
rcurtin: Nice catch, thanks.

src/mlpack/core/tree/binary_space_tree/traits.hpp
```diff
@@ -73,7 +73,6 @@ class TreeTraits<BinarySpaceTree<MetricType, StatisticType, MatType,
 {
  public:
   static const bool HasOverlappingChildren = true;
-  static const bool HasDuplicatedPoints = false;
```
rcurtin: I think it would be useful to keep this trait, even if it is not used right now.

MarcosPividori:
Hi @rcurtin, thanks for your comments.
I have update the code to include your suggestions:
+ I have added more overloads for the constructor and Train() methods, to consider lvalue and rvalue references (copy and move semantics).
All of them are very simple because they just create a new tree and then call the overloads that take `TreeType*`.
+ I didn't remove the overlads that take `TreeType*` to maintain reverse compatibility, as you suggested. If you agree, we could add a comment that these methods will be removed in the future release (if we decide that).
In my opinion we should remove that overloads in the next release, and only maintain the 2 overloads with references. If we do so, we will sure that `NeighborSearch` class always has ownership of the reference tree (because it was copied or moved).

Would you agree?

Thanks!

rcurtin:
I agree completely except for one point (below). I think GCC provides some symbols to mark methods deprecated, but I am not sure how portable they are. I don't currently have the ability to look into that, but I will when I have a chance unless you beat me to it. :)

Removing the old overloads would force a major version bump, and I'd prefer to be somewhat conservative with major version bumps. I am currently working on generic bindings to other languages, which would be a good reason for a major version bump (the completion of the ANN code would be another good reason).  So I think that either (or both) of those things can be done probably by the end of the year, and at that time we can remove all our collected reversed compatibility code and call it mlpack 3.0.0.

MarcosPividori:
Hi @rcurtin, yes! I completely agree. I meant to say a major version, sorry for the confusion.
Ok, I will investigate possible GCC symbols to mark deprecated methods!
Thanks!

PR #746, opened 2016-08-01
Universal B tree implementation

lozhnikov:
I implemented the UB tree. Despite the tree passes all tests from `ub_tree_test.cpp` including the test that verifies `MinDistance()` and `MaxDistance()`, `KNNTest/KNNModelTest` sometimes fails, I am trying to fix that.

The bound of the UB tree consists of a number of hyperrectangles in such a way that children of the UB tree do not overlap each other. However, I realized that the calculation of the distance between two nodes requires a lot of operations. So, I had to add some limitations on the number of hyperrectangles.

The tree sorts the dataset before the split process since the tree introduces an ordering in multidimensional space and all points are inserted into the tree according to that ordering. Probably it is better to choose a number of distinct samples and to split according the median.


src/mlpack/core/tree/binary_space_tree/ub_tree_split_impl.hpp
```diff
@@ -0,0 +1,326 @@
...
+                                                const size_t count,
+                                                size_t& splitCol,
+                                                std::vector<size_t>& oldFromNew)
+{
+  constexpr size_t order = sizeof(AddressElemType) * CHAR_BIT;
+  if (begin == 0 && count == data.n_cols)
+  {
+    InitializeAddresses(data);
+
+    std::sort(addresses.begin(), addresses.end(),ComparePair);
```
rcurtin: This is a minor issue; there is a space missing between `addresses.end(),` and `ComparePair`. :)

rcurtin:
I need to review the paper before I can make more comments, but I am surprised that this is so much slower than the kd-tree.  I am seeing performance about the same as the rp-tree.  To me this seems odd.  I can understand why the vp-tree is slow and I can understand why the rp-tree (and variants) are slow, and I know that the R tree and its variants will also be slow as compared to the kd-tree, but I had thought that the ub-tree, since it was so similar with axis-aligned pretty-tight hyperrectangles, would be much faster.  Even with low-dimensional data, it often takes an order of magnitude more base cases.

When I am able to review the paper, maybe I will be able to reason if something is wrong.  A cursory inspection of the code to calculate the minimum and maximum distances seems fine, but the only hesitation I have is about whether or not the actual strategy of splitting points into bounds is correct.  (But again I have to refer to the paper before I can comment more there.)

Outside of that, the code quality seems great.

lozhnikov:
The UB tree has two problems.

The bound of the tree is not the tightest since the tightest bound requires a lot of computations in `MinDistance()`. That can explain that the number of base cases with the kd tree is less than the number of base cases with the UB tree.

And the second problem is the performance of `MinDistance()`. The function calculates the distance to a number of hyperrectangles. I believe that many of them may be pruned but I do not know exactly how.

rcurtin:
I see what you mean about the problems of the UB tree, now that I have read the paper.  In essence the idea of mapping the space to addresses, and each address representing a subregion, means that if we have only two points that are adjacent in some dimension d, but the difference in the addresses is high, then the node enclosing both of those encloses a huge amount of empty space!  So, even if we do optimize this tree like crazy, I am not sure we will be able to produce something that is competitive with the kd-tree, at least for the task of nearest neighbor search.  Perhaps it will be more useful for something like range search or other tasks.

src/mlpack/core/tree/binary_space_tree/ub_tree_split.hpp
```diff
@@ -0,0 +1,90 @@
...
+   *
+   * @param data The dataset used by the binary space tree.
+   * @param count Number of points in this node.
+   */
+  void PerformSplit(MatType& data,
+                       const size_t count);
+
+  void PerformSplit(MatType& data,
+                       const size_t count,
+                       std::vector<size_t>& oldFromNew);
```
rcurtin: The spacing seems incorrect in these two function definitions.

src/mlpack/core/tree/binary_space_tree/ub_tree_split_impl.hpp
```diff
@@ -0,0 +1,326 @@
...
+  {
+    // Calculate all addresses.
+    InitializeAddresses(data);
+
+    // Probably this is not a good idea. Maybe it is better to get
+    // a number of distinct samples and find the median.
+    std::sort(addresses.begin(), addresses.end(), ComparePair);
+
+    // Rearrange dataset.
+    PerformSplit(data, count);
```
rcurtin: I think that the code could be simplified significantly if you instead used `arma::Mat<AddressElemType>` in order to store the addresses, and then used `sort_index()` to get the sorted order, which you could then apply to the data points and the addresses, and use to calculate `oldFromNew` if needed.  I think that would remove the need for `PerformSplit` altogether.

src/mlpack/core/tree/cellbound_impl.hpp
```diff
@@ -0,0 +1,844 @@
...
+    for (size_t d = 0; d < dim; d++)
+    {
+      lower = loBound(d, i) - point[d];
+      higher = point[d] - hiBound(d, i);
+
+      // Since only one of 'lower' or 'higher' is negative, if we add each's
+      // absolute value to itself and then sum those two, our result is the
+      // nonnegative half of the equation times two; then we raise to power Power.
+      sum += pow((lower + fabs(lower)) + (higher + fabs(higher)),
+          (ElemType) MetricType::Power);
```
rcurtin: You can save a little bit of computation here if you terminate early:

```
if (sum >= minSum)
  continue;
```

src/mlpack/core/tree/cellbound_impl.hpp
```diff
@@ -0,0 +1,844 @@
...
+    }
+
+    if (sum < minSum)
+      minSum = sum;
+  }
+
```
rcurtin: So you have two different bounding structures, if I understand right: the minimum bounding hyperrectangle, and also the series of bounding rectangles defined by the minimum and maximum addresses.  In some cases that minimum bounding hyperrectangle will be much tighter than the series of bounding rectangles, so I wonder: you might get better performance if this function took the max of the minimum distances to the series of bounding rectangles and to the minimum bounding rectangle.  (The same applies to the other bounding functions too.)

lozhnikov: This is useless since I bound each rectangle in the series by the minimum bounded rectangle (see `AddBound()`).
```
loBound(k, numBounds) = std::max(loCorner[k], bounds[k].Lo());
hiBound(k, numBounds) = std::min(bounds[k].Hi(), hiCorner[k]);
```

src/mlpack/core/tree/binary_space_tree/ub_tree_split_impl.hpp
lozhnikov: I've just verified, `sort_index(X)` provides the sorted order of the elements of `X` instead of the columns. It seems, the function is not applicable in this case,

rcurtin: Oh, I think I understand now.  An address is actually a vector, not just a single element, so yeah, `sort_index()` will not be helpful.  In this case, could you please add some comments to the `PerformSplit` function?  I had a hard time understanding exactly what was going on and why the two auxiliary vectors were necessary.

src/mlpack/core/tree/cellbound_impl.hpp
rcurtin: I see.  Do you think it would make more sense, then, to construct the minimum bounding rectangle on the points held in the node and not on the other rectangles in the node?

lozhnikov: Hmm... I construct the minimum bounding rectangle on the points held in the node:
```
  arma::Col<ElemType> mins(min(data, 1));
  arma::Col<ElemType> maxs(max(data, 1));

  minWidth = std::numeric_limits<ElemType>::max();
  for (size_t i = 0; i < dim; i++)
  {
    bounds[i] |= math::RangeType<ElemType>(mins[i], maxs[i]);
```
`bounds` is the minimum bounding rectangle.

src/mlpack/core/tree/binary_space_tree/binary_space_tree.hpp
```diff
@@ -506,13 +507,55 @@ class BinarySpaceTree
...
+   * @param splitInfo The information about the split.
+   * @param oldFromNew Vector which will be filled with the old positions for
+   *    each new point.
+   */
+  size_t PerformSplit(MatType& data,
+    const size_t begin,
+    const size_t count,
+    const typename UBTreeSplit<BoundType<MetricType>,
+                               MatType>::SplitInfo& splitInfo,
+    std::vector<size_t>& oldFromNew);
```
rcurtin: The spacing seems odd on this line.

src/mlpack/core/tree/binary_space_tree/binary_space_tree_impl.hpp
```diff
@@ -877,6 +879,37 @@ template<typename MetricType,
          template<typename BoundMetricType, typename...> class BoundType,
          template<typename SplitBoundType, typename SplitMatType>
              class SplitType>
+size_t BinarySpaceTree<MetricType, StatisticType, MatType, BoundType,
+    SplitType>::PerformSplit(MatType& data,
+    const size_t begin, const size_t count,
+    const typename UBTreeSplit<BoundType<MetricType>, MatType>::SplitInfo& splitInfo)
```
rcurtin: Ah, can you split each parameter onto its own line here?  That may necessitate putting `MatType& data` on its own line too so it lines up with the other parameters.  I know it is a minor comment but it is for consistency. :)

src/mlpack/core/tree/cellbound.hpp
```diff
@@ -0,0 +1,269 @@
...
+#include <mlpack/core.hpp>
+#include <mlpack/core/math/range.hpp>
+#include <mlpack/core/metrics/lmetric.hpp>
+#include "bound_traits.hpp"
+#include "address.hpp"
+
+namespace mlpack {
+namespace bound {
+
+template<typename MetricType = metric::LMetric<2, true>,
```
rcurtin: Could you add some documentation here for the class itself?  You can reuse what you've written at the top of the file.  It might also be useful to point out that you are caching the minimum bounding rectangle of the points here.

src/mlpack/core/tree/cellbound_impl.hpp
```diff
@@ -0,0 +1,970 @@
...
+  assert(loCorner.n_elem == dim);
+  assert(hiCorner.n_elem == dim);
+
+  // If the subrectangle is not contained entirely in the outer rectangle,
+  // we shrink it.
+  for (size_t k = 0; k < dim; k++)
+  {
+    loBound(k, numBounds) = std::max(loCorner[k], bounds[k].Lo());
+
+    hiBound(k, numBounds) = std::min(bounds[k].Hi(), hiCorner[k]);
```
rcurtin: I believe what you are doing here is, if possible, shrinking one of the UB hyperrectangles down using the minimum point-enclosing hyperrectangle. But I believe you could get a tighter bound if you instead iterated over all points, using only those that fall into this UB hyperrectangle, in order to shrink the UB hyperrectangle.  I guess it is easier to demonstrate what I mean with code...

```
math::Range r(loCorner[k], hiCorner[k]);
for (size_t j = 0; j < data.n_cols; ++j)
{
  if (hyperrectangle contains data.col(j))
  {
    r.Lo() = std::max(r.Lo(), data(k, j));
    r.Hi() = std::min(r.Hi(), data(k, j));
  }
}
```

The handling will become a little more tricky for operator|=(), but I believe this should help make the bounds tighter and decrease runtime. Let me know if I can clarify what I mean.

src/mlpack/core/tree/cellbound_impl.hpp
rcurtin: OK, I see, thanks for the clarification.

src/mlpack/core/tree/cellbound_impl.hpp
```diff
@@ -0,0 +1,970 @@
...
+        else if (MetricType::Power == 2)
+        {
+          loSum += vLo * vLo;
+          hiSum += vHi * vHi;
+        }
+        else
+        {
+          loSum += std::pow(vLo, (ElemType) MetricType::Power);
+          hiSum += std::pow(vHi, (ElemType) MetricType::Power);
+        }
```
rcurtin: You could probably do early termination here too but the condition would be a but stricter.

src/mlpack/core/tree/cellbound_impl.hpp
lozhnikov: I think I understand. Some of the hyperrectangles may be even empty.

src/mlpack/core/tree/cellbound_impl.hpp
lozhnikov: I thought about that. I can reduce the computation of `loSum`. But I have to look through all combinations in order to calculate `hiSum`. That means that I do only two floating-point operations per iteration. I am not sure that will be efficient. But maybe I am mistaken and it is worth to try.

rcurtin: You're right, I did not think my comment through thoroughly before I posted it.  We can't shortcut any max-distance calculation.

src/mlpack/core/tree/cellbound_impl.hpp
rcurtin: That's true, and I guess in those cases you can remove those hyperrectangles.  I thought about it more... if all of the hyperrectangles are tight in this manner, then is there any reason to store the minimum bounding hyperrectangle?  I don't think it would be useful anymore.  But maybe I have overlooked something.

src/mlpack/core/tree/binary_space_tree/binary_space_tree.hpp
```diff
@@ -548,12 +549,13 @@ class BinarySpaceTree
...
-    const typename UBTreeSplit<BoundType<MetricType>,
-                               MatType>::SplitInfo& splitInfo,
-    std::vector<size_t>& oldFromNew);
+  size_t PerformSplit(
+      MatType& data,
+      const size_t begin,
+      const size_t count,
+      const typename UBTreeSplit<BoundType<MetricType>,
+                                 MatType>::SplitInfo& splitInfo,
+      std::vector<size_t>& oldFromNew);
```
lozhnikov: I fixed spaces. But I am not sure that you do not mean the 557th line.

src/mlpack/core/tree/cellbound_impl.hpp
```diff
@@ -169,34 +169,55 @@ inline void CellBound<MetricType, ElemType>::Center(
...
-  // we shrink it.
   for (size_t k = 0; k < dim; k++)
   {
-    loBound(k, numBounds) = std::max(loCorner[k], bounds[k].Lo());
+    loBound(k, numBounds) = std::numeric_limits<ElemType>::max();
+    hiBound(k, numBounds) = std::numeric_limits<ElemType>::lowest();
+  }
 
-    hiBound(k, numBounds) = std::min(bounds[k].Hi(), hiCorner[k]);
+  for (size_t i = 0; i < data.n_cols; i++)
```
lozhnikov: Indeed, that provides some benefits. If I am not mistaken, the minimum bounding rectangle is used in the `Contains()` function but I can replace that by the hyperrectangles.

src/mlpack/core/tree/binary_space_tree/binary_space_tree.hpp
rcurtin: It's fine as-is now, thanks for the fix. :)

src/mlpack/core/tree/address.hpp
```diff
@@ -0,0 +1,249 @@
...
+ * Compare two addresses. The function returns 1 if the first address is greater
+ * than the second one, -1 if the first address is less than the second one,
+ * otherwise the function returns 0.
+ */
+template<typename AddressType1, typename AddressType2>
+int CompareAddresses(const AddressType1& addr1, const AddressType2& addr2)
+{
+  static_assert(sizeof(typename AddressType1::elem_type) ==
+      sizeof(typename AddressType2::elem_type), "We aren't able to compare "
+      "adresses of distinct sizes");
```
rcurtin: I think that you meant to check here if the number of elements in each address was the same, but this only checks that the size of each type of element is the same.  Should we require that the type of element (i.e. double, size_t, whatever) is exactly the same, instead of simply having the same size?

lozhnikov: The type of an address is either `uint32_t` or `uint64_t`
```
typedef typename std::conditional<sizeof(ElemType) * CHAR_BIT <= 32,
                                    uint32_t,
                                    uint64_t>::type AddressElemType;
```
So, there is no difference. But maybe you are right and it is better to check types instead of sizes.

src/mlpack/core/tree/address.hpp
```diff
@@ -0,0 +1,249 @@
...
+/**
+ * Calculate the address of a point. Be careful, the point and the address
+ * variables should be equal-sized and the type of the address should correspond
+ * to the type of the vector.
+ *
+ * @param address The resulting address.
+ * @param point The point that is being translated to the address.
+ */
+template<typename AddressType, typename VecType>
+void PointToAddress(AddressType& address, const VecType& point)
```
rcurtin: If you don't mind, I'd like to add some documentation on what exactly is going on in this function.  I spent a long time reading it, and it appears that really all we are doing (from a high level) is reordering the bits in the floating point number.  So in the one-dimensional case, we have a number of the form

```
[[ mantissa ][ exponent ]]
```

but your code transforms it roughly to

```
[[ exponent ][ mantissa ]]
```

(but not exactly, since some modification of the mantissa may be necessary).  In the multi-dimensional case, after we transform the representation, we have to interleave the bits of the new representation across all of the elements in the address vector.  Is that correct?  If so I can add those comments (or you can add them if you like, maybe it is better if you add them since you are more familiar with the code).

src/mlpack/core/tree/address.hpp
rcurtin: Ah, okay, I forgot that `AddressElemType` is only ever `uint32_t` or `uint64_t`.  It's your call whether or not you want to change it.  I think we should keep the two different `AddressType1` and `AddressType2` types, otherwise a user can't call `CompareAddresses` with a subview (like `matrix.col(i)`) and a point, or other weird derived Armadillo types like that.

src/mlpack/core/tree/address.hpp
lozhnikov: yeah, it's correct. In one dimensional case this transform preserves the ordering (lower addresses correspond lower points). The function looks like `DiscreteHilbertValue::CalculateValue()`. I introduced that since it is faster than recursive calculation (indeed, `RecursiveHilbertValue` was slower than `DiscreteHilbertValue`).

src/mlpack/core/tree/binary_space_tree/binary_space_tree_impl.hpp
```diff
@@ -877,7 +879,43 @@ template<typename MetricType,
...
+    SplitType>::PerformSplit(
+    MatType& data,
+    const size_t begin,
+    const size_t count,
+    const typename UBTreeSplit<BoundType<MetricType>,
+                               MatType>::SplitInfo& splitInfo)
+{
+  return SplitType<BoundType<MetricType>, MatType>::PerformSplit(data, begin,
+      count, splitInfo);
+}
```
rcurtin: Instead of having a specific overload for `UBTreeSplit`, what do you think of this approach instead:

 * Move `BinarySpaceTree::PerformSplit()` (the non-specific version) into the `math` namespace somewhere as an auxiliary standalone function.
 * In `BinarySpaceTree::SplitNode()`, don't call `BinarySpaceTree::PerformSplit()` but instead call `SplitType::PerformSplit()`.
 * In each `SplitType`, `PerformSplit()` can just be a call to the function in the `math` namespace, or in the case of `UBTreeSplit`, it can be the custom strategy you are using there.

What do you think?  If you like it, we can do it.  If you do like it, I can do the refactoring or you can, it's your choice. :)

src/mlpack/core/tree/address.hpp
rcurtin: Ok, thanks.  Sorry my intuition was so wrong with the recursive Hilbert value calculation, I think my ideas wasted a lot of your time there. :(

src/mlpack/core/tree/binary_space_tree/ub_tree_split.hpp
```diff
@@ -0,0 +1,114 @@
...
+  typedef typename std::conditional<sizeof(typename MatType::elem_type) * CHAR_BIT <= 32,
+                                    uint32_t,
+                                    uint64_t>::type AddressElemType;
+
+  /**
+   * This class performs the actual splitting i.e. it reorders the dataset.
+   * This variable should be equal to true if the class does not perform the
+   * actual splitting.
+   */
+  static constexpr bool NeedRearrangeDataset = false;
```
rcurtin: I don't see this used anywhere; should we remove it?

src/mlpack/core/tree/binary_space_tree/ub_tree_split_impl.hpp
```diff
@@ -0,0 +1,244 @@
...
+      newFromOld[index] = i;
+      newFromOld[oldI] = tmp;
+
+      tmp = newToOld[i];
+      newToOld[i] = newToOld[newIndex];
+      newToOld[newIndex] = tmp;
+
+      tmp = oldFromNew[i];
+      oldFromNew[i] = oldFromNew[newIndex];
+      oldFromNew[newIndex] = tmp;
```
rcurtin: I believe this rearranging is only done in the first level of building the tree (at which point `oldFromNew` is initialized in the same way `newToOld` is initialized here), so does it make sense to simply remove the `newToOld` member used here and instead use `oldFromNew`?  If so, maybe it makes sense to have the other overload of `PerformSplit` call this one after initializing the `oldFromNew` vector.  I suspect I may have overlooked something here, so let me know if I have. :)

src/mlpack/core/tree/binary_space_tree/binary_space_tree_impl.hpp
rcurtin: Sure, that sounds good to me.  I figured it might be ok in `math` since maybe somewhere else might use it someday, but I agree, maybe we shouldn't move it there until multiple places use it.

src/mlpack/core/tree/address.hpp
lozhnikov: no problem:) I was not sure of `DiscreteHilbertValue` and that was an interesting experiment.

src/mlpack/core/tree/binary_space_tree/ub_tree_split.hpp
lozhnikov: Oh.. I thought that could solve the problem with the overload, I've forgotten to remove that.

src/mlpack/core/tree/binary_space_tree/ub_tree_split_impl.hpp
```diff
@@ -0,0 +1,244 @@
...
+{
+  constexpr size_t order = sizeof(AddressElemType) * CHAR_BIT;
+  if (begin == 0 && count == data.n_cols)
+  {
+    // Calculate all addresses.
+    InitializeAddresses(data);
+
+    // Probably this is not a good idea. Maybe it is better to get
+    // a number of distinct samples and find the median.
+    std::sort(addresses.begin(), addresses.end(), ComparePair);
```
rcurtin: I think this is an okay approach for now.  Generally tree construction time is pretty minimal compared to the actual search time, so I don't know if there's much need to actually do the sampling to calculate the median.

src/mlpack/core/tree/binary_space_tree/ub_tree_split_impl.hpp
lozhnikov: Yeah; First time I had used `oldFromNew` instead of `newToOld`. But then I had thought that the dataset can be rearranged before the tree construction (that never happens right now).

src/mlpack/core/tree/cellbound_impl.hpp
```diff
@@ -0,0 +1,994 @@
...
+
+    // This hyperrectangle is not contained entirely in the bound.
+    // So, the number of hyperrectangles should be increased.
+    if (tmpHiAddress[row] & ((AddressElemType) 1 << bit))
+      numCorners++;
+
+    // We ran out of the limit of hyperrectangles. In that case we enlare
+    // the last hyperrectangle.
+    if (numCorners >= maxNumBounds / 2)
+      tmpHiAddress[row] |= ((AddressElemType) 1 << bit);
```
rcurtin: What happens when `maxNumBounds` isn't even?  Do we end up with `maxNumBounds - 1` bounds?  That's what it seems like to me.  I don't know if there is an easy fix.

src/mlpack/core/tree/cellbound.hpp
```diff
@@ -0,0 +1,306 @@
...
+
+ private:
+  //! The precision of the tree element type.
+  static constexpr size_t order = sizeof(AddressElemType) * CHAR_BIT;
+  //! Maximum number of subrectangles.
+  const size_t maxNumBounds = 10;
+  //! The dimensionality of the bound.
+  size_t dim;
+  //! The bounds for each dimension.
+  math::RangeType<ElemType>* bounds;
```
rcurtin: It seems like we can't remove this because of the `Center()` calculation and the `Diameter()`.  Removing it would mean that for `Center()` we would have to calculate the mean of all subrectangle centers, and it would be harder to cache something for `Diameter()`.  My idea in removing it was to make `sizeof(CellBound)` smaller but it seems that's not really feasible.  But I thought it would be useful to write down the notes anyway in case it was ever revisited later. :)

rcurtin:
Looks great, I think this is ready for merge, after the `PerformSplit()` refactoring and `NeedRearrangeDataset` removal.  If you'd like me to do either of those, let me know and I will.  I think I still have yet to find any bugs in any of your code. :+1:

While going over this, I thought that the ordering of dimensions makes a difference for the performance of the tree, because of how the addresses are calculated.  So I wondered if ordering the dimensions in the address according to the variance of each dimension (or simply scaling the variances to 1) might help performance.  But I think that the UB tree is so much slower than the kd tree that it's not really worth investigating.  I thought it was an interesting idea to write down anyway though. :)

src/mlpack/core/tree/cellbound_impl.hpp
lozhnikov: yeah, I think if `maxNumBounds` is odd then we end up with `maxNumBounds - 1`. I think we can replace
`if (numCorners >= maxNumBounds / 2)`
by
`if (numCorners >= maxNumBounds - numBounds)` in `InitLowerBound()`
In that case we should document that `InitLowerBound()` should be called after
`InitHighBound()`. But in that case the functions are non-symmetric.

Another way is to document that `maxNumBounds` should be even.

And another way is to make an extra pass and calculate `numCorners` for the lower bound and for the high bound. What do you think about these variants?

lozhnikov:
The idea with the ordering of dimensions is really interesting, I didn't think about that. 

I'll do the refactoring. I edited my comment about `oldFromNew`, I am not sure that github sends the notification.

> P.S. Oh.. I overlooked that `oldFromNew` is initialized in the constructor of `BinarySpaceTree`. You are right.

> P.P.S. Do you mean a function like `SplitType::RearrangeDataset()`? (i.e. a function that rearranges the dataset before the split, the function do nothing in other binary space trees). We could add a check to `PerformSplit()` if we needn't invoke that (e.g. `static bool SplitType::NeedPerformSplit`). This is another solution of the problem with the overload. What do you think about that?

What do you think about this idea? Maybe it is better than calling `SplitType::PerformSplit()`.

rcurtin:
Ah, yeah, Github doesn't seem to send notifications when comments are updated.

I think either strategy for the `PerformSplit()` refactoring can work, but what I wanted to avoid was a specialization for `UBTreeSplit`.  You can definitely add a static boolean and use that to determine whether a function like `PerformSplit()` is called.  Then basically the "default" behavior is that the split is performed if `SplitType::NeedsPerformSplit` is true; otherwise the `SplitType` itself is responsible for rearranging the dataset.  As long as we document how it works, I think either way is fine, so pick whichever seems best to you.  If I did it, I'd probably lean toward the first idea of the standalone function that most `SplitType`s call, but my preference is pretty minimal.

src/mlpack/core/tree/cellbound_impl.hpp
rcurtin: I think the first strategy is best here.  I don't think it's a problem if the functions are not symmetric, and I think the performance impact will be negligible.  Definitely the first strategy is best, but if you want to make an extra pass that's fine too.  I just want to avoid a situation where a user says "but if I set maxNumBounds to 8 and 9 the performance is the same, and that doesn't make sense!"

lozhnikov:
I finished the refactoring (except of ordering the dimensions in the address, I should think a little more about that). It seems I haven't overlooked anything. Since the PR is almost ready to merge, I added the UB tree to `RSModel`.

rcurtin:
I don't think the ordering of dimensions is going to make too much difference, it would only be a minor research idea anyway.  (Plus how do you make sure you're using the same ordering across the whole tree?  That will be a minor refactoring challenge.)  Anyway, I think this is ready to merge, so I can do that when you are ready (I guess you can do it to, so if you want to, go ahead).  I guess it should be added to `RAModel` too, but if you like I can do that after merge.

lozhnikov:
Yeah, I am not sure how to preserve the same ordering across the whole tree. I'll add the tree to `RAModel`.

lozhnikov:
I tried a number of different seeds, it seems the tree works with `RAModel`. So, if you think that the PR is ready to merge, I can do that (in the morning; I'll look through the whole PR, maybe I'll find typos) .

PR #747, opened 2016-08-01
Spill trees

MarcosPividori:
Hi @sumedhghaisas @rcurtin,

I have implemented Spill trees with axis-parallel splitting hyperplanes. I have made an effort to avoid duplicating existing code for neighbor search.

I created a new class `SpillSearch` that provides an interface similar to `NeighborSearch` but with an extension to properly set the tau parameter. It encapsulates an instance of `NeighborSearch`.

Also, I have implemented a new version of `NeighborSearchRules` specialized for Spill Trees, because I needed to modify the methods:
 + `Score()` to consider splitting hyperplanes for overlapping nodes.
 + `CalculateBounds()` to ignore B_2 bound (we can not use B_2 bound for Spill Trees).

## Single Tree Search:

The `SingleTreeTraverser` is similar to the implementation for `BinarySpaceTree`.
The difference is in the implementation of `NeighborSearchRules` for SpillTrees.
When calculating the score of a query point and a reference node I consider 2 cases:
  + If the reference node is non-overlapping, I calculate the score the same than before.
  + If the reference node is overlapping, I analyze the reference node's half space. If it contains the given query point, I return 0 (best score). Else, I return DBL_MAX (prune).

## Dual Tree Search:

The Query tree is built without overlapping.

When calculating the score of a query node and a reference node, I consider 2 cases:
  + If the reference node is a non-overlapping node, I calculate the score the same as before.
  + If the reference node is a overlapping node, I analyze query node's bounding box. If it intersects the reference node's half space, I return 0 (best score). Else, I return DBL_MAX (prune).

The `DualTreeTraverser` is slightly different to the implementation for `BinarySpaceTree`. When referenceNode is a overlapping node and we can't decide which child node to traverse, this means that queryNode is at both sides of the splitting hyperplane, we analyse the queryNode:
 + If queryNode is a non-leaf node, I recurse down the query node.
 + If queryNode is a leaf node, I do single tree search for each point in the query node.

The `DualTreeTraverser` is faster than the `SingleTreeTraverser`. Specially when the value of tau increases, because we will have more non-overlapping nodes which implies more time involved in backtracking.

The extension was incorporated to existing *mlpack_knn*. With actual implementation, we can use *"-t spill"* to consider spill trees and *"--tau 0.1"* to set different values for the overlapping size (default value is tau=0).

Every feedback is welcome! :)

Thanks


MarcosPividori:
Hi @sumedhghaisas! I have implemented generalized Spill Trees, to consider general splitting hyperplanes, not necessarily axis-orthogonal.

I added a new template parameter for Spill Trees: `HyperplaneType`, and I implemented two candidate classes: `AxisOrthogonalHyperplane` and `Hyperplane`.
+ `AxisOrthogonalHyperplane` consider an axis-parallel projection vector. So, we can project any vector in O(1) time, considering the appropiate dimension.
+ `Hyperplane` consider a general projection vector. So we can project any vector in O(d) time, through a dot product.

Inside the SpillTrees, I consider BallBound for non-axis-orthogonal hyperplanes, and HrectBound for axis-orthogonal hyperplanes.
Also, I replaced the methods: SplitValue() and SplitDimension() by HalfSpaceContains() and HalfSpaceIntersects(), which are more general.

Considering this abstraction, I managed to significantly simplify the Split methods: MeanSplit and Midpoint Split. Now, they share most of the code.

I think we could use this abstraction for BinarySpaceTrees too.

By default, SpillSearch considers `AxisOrthogonalHyperplanes` because they seem to be faster in many cases, but this is not always true. I think we should benchmark both methods and see which is faster and which is more accurate, I mean, which has the best relation between running time and relative approximation error.

I will add a new command line flag "--get_real_error" to compare the approximate neighbor search against the naive method and print the real relative error, so we can benchmark with different values of tau and different spill tree flavours.

Thanks,

Marcos

src/mlpack/core/tree/spill_tree/spill_tree.hpp
```diff
@@ -0,0 +1,432 @@
...
+  SpillTree(MatType&& data,
+            const double tau = 0,
+            const size_t maxLeafSize = 20,
+            const double rho = 0.7);
+
+  /**
+   * Construct this node as a child of the given parent, starting at column
+   * begin and using count points.  The ordering of that subset of points in the
+   * parent's data matrix will be modified!  This is used for recursive
+   * tree-building by the other constructors which don't specify point indices.
```
rcurtin: I think this comment is not accurate---"starting at column begin and using count points" is something that applies to the `BinarySpaceTree`, but not here.

src/mlpack/core/tree/spill_tree/spill_tree.hpp
```diff
@@ -0,0 +1,432 @@
...
+   * Create a hybrid spill tree by copying the other tree.  Be careful!  This
+   * can take a long time and use a lot of memory.
+   *
+   * @param other tree to be replicated.
+   */
+  SpillTree(const SpillTree& other);
+
+  /**
+   * Move constructor for a SpillTree; possess all the members of the
+   * given tree.
```
rcurtin: This looks like it's shorter than 80 characters, so I think you can compress it onto a single line.  I know, trivial comment :)

src/mlpack/core/tree/spill_tree/spill_tree.hpp
```diff
@@ -0,0 +1,432 @@
...
+
+  //! Gets the left child of this node.
+  SpillTree* Left() const { return left; }
+  //! Modify the left child of this node.
+  SpillTree*& Left() { return left; }
+
+  //! Gets the right child of this node.
+  SpillTree* Right() const { return right; }
+  //! Modify the right child of this node.
+  SpillTree*& Right() { return right; }
```
rcurtin: If you prefer, I am happy to leave out the `Left()` and `Right()` functions, and have users access the left and right children as `Child(0)` and `Child(1)`.  I think all of the algorithms that use trees should be depending on `Child()` anyway.

MarcosPividori:
AppVeyor failure is driving me mad! I have tried many options but I can't fix it.
Could someone with a windows machine help me? @stereomatchingkiss ? Sorry if you are busy I can continue thinking about this on my own.

src/mlpack/core/tree/spill_tree/spill_tree_impl.hpp
```diff
@@ -0,0 +1,731 @@
...
+    hyperplane(),
+    bound(data.n_rows),
+    parentDistance(0), // Parent distance for the root is 0: it has no parent.
+    dataset(&data),
+    localDataset(false)
+{
+  std::vector<size_t> points;
+  points.reserve(dataset->n_cols);
+  for (size_t i = 0; i < dataset->n_cols; i++)
+    points.push_back(i);
```
rcurtin: I think that it is better to try and use Armadillo objects here.  In general I have found they are faster (even if you are reserving space for the vector).

src/mlpack/core/tree/spill_tree/spill_tree_impl.hpp
```diff
@@ -0,0 +1,731 @@
...
+             class SplitType>
+void SpillTree<MetricType, StatisticType, MatType, HyperplaneType, SplitType>::
+    SplitNode(std::vector<size_t>& points,
+              const size_t maxLeafSize,
+              const double tau,
+              const double rho)
+{
+  // We need to expand the bounds of this node properly.
+  for (size_t i = 0; i < points.size(); i++)
+    bound |= dataset->cols(points[i], points[i]);
```
rcurtin: Why not just `col(points[i])`?

src/mlpack/core/tree/spill_tree/spill_tree_impl.hpp
```diff
@@ -0,0 +1,731 @@
...
+  // We don't need the information in points, so lets clean it.
+  std::vector<size_t>().swap(points);
+
+  // Now we will recursively split the children by calling their constructors
+  // (which perform this splitting process).
+  left = new SpillTree(this, leftPoints, tau, maxLeafSize, rho);
+  right = new SpillTree(this, rightPoints, tau, maxLeafSize, rho);
+
+  // Update count number, to represent the number of descendant points.
+  count = left->NumDescendants() + right->NumDescendants();
```
rcurtin: I have a concern about this.  This means that `NumDescendants()` of the root can actually be greater than `dataset.n_cols`, but I think this can cause problems.  If we refactor so that `NumDescendants()` of the root returns `dataset.n_cols`, then `Descendant()` will be a little more complex: you'll just have to rearrange the point indices so that the first `n` elements are the non-shared elements, and possibly hold another member to indicate how many non-shared elements the node has.

src/mlpack/core/tree/spill_tree/spill_tree_impl.hpp
```diff
@@ -0,0 +1,731 @@
...
+         typename MatType,
+         template<typename HyperplaneMetricType> class HyperplaneType,
+         template<typename SplitMetricType, typename SplitMatType>
+             class SplitType>
+bool SpillTree<MetricType, StatisticType, MatType, HyperplaneType, SplitType>::
+    SplitPoints(const double tau,
+                const double rho,
+                const std::vector<size_t>& points,
+                std::vector<size_t>& leftPoints,
+                std::vector<size_t>& rightPoints)
```
rcurtin: I have a comment about this function.  I think that this function can be pretty slow because of the large number of allocations that may be incurred as `std::vector` resizes itself when `push_back()` is called.  Preventing many allocations was a big concern when I implemented sparse matrices for Armadillo, and the typical strategy looked like this:

 - take a first pass over the data, and determine how large the results need to be
 - take a second pass over the data and fill the results

So in this case, in the first pass you could calculate `leftSize`, `leftFrontierSize`, `rightSize`, and `rightFrontierSize`, and cache the values of the projections for each point.  In the second pass, you could just use those cached values to assign points to the correct side.  Let me know what you think.  I think for big datasets, it will make a difference.

rcurtin:
I have no idea about the AppVeyor failure; maybe it is an MSVC bug?  It seems like maybe there is some issue with template aliases.

I have a concern
I have no idea about the AppVeyor failure; maybe it is an MSVC bug?  It seems like maybe there is some issue with template aliases.

I like the way that you have managed to reuse so much code.  To this point I have only really reviewed the `SpillTree` class and `SpillSearch` class, so I still have much to look at.  But there is a big issue I think we should address:

If we use mlpack's existing single-tree and dual-tree traversers, then the process at each node looks like this (invariably):

 1. Score each node.
 2. Recurse into those nodes that have not been pruned.

So this means for defeatist search, even with spill trees, we are doing the following:

 1. Calculate left score (0 or DBL_MAX).
 2. Calculate right score (0 or DBL_MAX).
 3. Recurse into the direction whose score is 0.

Since calculations 1 and 2 are separate, we are projecting the query point (or query node) onto the splitting hyperplane's normal vector twice.  But that's twice as much work as we need to do.  Instead, we would be better off if we implemented some function `GetNearestChild()` (or something similar), and then did not use the traversers at all for defeatist search, but instead inside of `Search()`, something more like calling `GetNearestChild()` and then proceeding in the direction of that child only, until the current node is a leaf.  Dual-tree search might be a bit more tricky but still definitely possible.

What do you think?  I think that could accelerate the runtime of defeatist search significantly.  If you agree with the idea, then we can keep discussing details, but there is no need until we get to that point, I think. :)

src/mlpack/core/tree/spill_tree/spill_tree.hpp
MarcosPividori: Done in https://github.com/mlpack/mlpack/pull/747/commits/54614baee1591b6e7529c7da61fd7fa390ebb5e3

src/mlpack/core/tree/spill_tree/spill_tree.hpp
MarcosPividori: Done in https://github.com/mlpack/mlpack/pull/747/commits/54614baee1591b6e7529c7da61fd7fa390ebb5e3

src/mlpack/core/tree/spill_tree/spill_tree_impl.hpp
sumedhghaisas: I agree. I think we should use colvec here.

src/mlpack/core/tree/spill_tree/spill_tree.hpp
```diff
@@ -0,0 +1,431 @@
...
+   * @param rho Balance threshold.
+   */
+  SpillTree(SpillTree* parent,
+            std::vector<size_t>& points,
+            const double tau = 0,
+            const size_t maxLeafSize = 20,
+            const double rho = 0.7);
+
+  /**
+   * Create a hybrid spill tree by copying the other tree.  Be careful!  This
```
sumedhghaisas: I like this warning! :)

src/mlpack/core/tree/spill_tree/spill_tree.hpp
sumedhghaisas: I think Left() and Right() functions are more intuitive. In this specific case of course.

sumedhghaisas:
Do you think we should break the file space_split.hpp into a folder space_split_types and have individual headers dedicated to the different splits?

src/mlpack/core/tree/spill_tree/spill_tree_impl.hpp
MarcosPividori: Ok, I have done it in: https://github.com/mlpack/mlpack/pull/747/commits/071588457415094e99883774af541075e3858c22

src/mlpack/core/tree/spill_tree/spill_tree_impl.hpp
MarcosPividori: I have simplified it in: https://github.com/mlpack/mlpack/pull/747/commits/071588457415094e99883774af541075e3858c22

src/mlpack/core/tree/spill_tree/spill_tree_impl.hpp
MarcosPividori: @rcurtin  Yes, I agree. You are right, this can make a difference for big datasets.
I have updated the code in: https://github.com/mlpack/mlpack/pull/747/commits/071588457415094e99883774af541075e3858c22
Thanks!

src/mlpack/core/tree/spill_tree/spill_tree.hpp
MarcosPividori: I agree with @sumedhghaisas .  I think it is more intuitive to use `Right()` and `Left()` in the implementation of the tree's traversers (`SingleTreeTraverser` and `DualTreeTraverser`).
But of course, if you prefer, I can update them to use `Child(0)` and `Child(1)`.

MarcosPividori:
@sumedhghaisas I decided to include them in the same header file because they represent only a few lines of code, so I thought it would be better to maintain a single file for this.
If you prefer, I can create separated files. :)

MarcosPividori:
@rcurtin  Thanks for your feedback! 

I implemented Hybrid Spill Search. It does:
 + backtracking in non-overlapping node.
 + defeatist search on overlapping nodes.

So, it is different to pure defeatist search.

Because of that, I think we should continue using the Tree Traversers for the implementation.

I agree that code could be improved to avoid calculating the score of both child nodes (which means calculating the projection 2 times). It can make a difference, especially with non axis-orthogonal splitting hyperplanes, because the projection takes O(d) time.

So, I can implement a new method as you suggested:  "GetNearestChild()" and use it when working with a overlapping node in the `SingleTreeTraverser` and `DualTreeTraverser`.

That method is not strictly necessary. We could avoid including a new method, implementing it in a simply way:
If I have a overlapping node I calculate the score of one child, and decide based on that score without calculating the score of the second child:
- If the score of the first child is `DBL_MAX`, I have to traverse the second child (its score is `0`).
- If the score of the first child is `0`, I have to traverse the first child (the score of the second child is `DBL_MAX`).

Also, we could implement a pure `DefeatistTreeTraverser`, for every kind of tree (not only spill trees) that always considers the score of all the child nodes, chose the one with the lowest score and doesn't do backtracking.

Thanks!

src/mlpack/core/tree/spill_tree/spill_tree_impl.hpp
MarcosPividori: @rcurtin Thanks for your comments.
Why do you think this can cause problems?
This implementation is different to the implementation of `BinarySpaceTrees`. I don't provide methods such as: `Begin()` and `Count()`.

rcurtin: Sometimes you want to sample descendant points from a node.  Rank-approximate nearest neighbor search (`src/mlpack/methods/rann/`) does this.  So you would just sample uniformly from `i` in [0, `NumDescendants()`) and then take `Descendant(i)` as your random point.  But if descendants are not unique (that is, if they are double-counted), then you get a biased random sample.  In this case, points in the spill region are twice as likely to be sampled.  Let me know if I can clarify further.

MarcosPividori: Yeah, I understand! Ok, so we want `Descendant(i)` to represent a unique points from the dataset. I will think about this!

MarcosPividori: Hi,
I think this is a complicated problem. I have been thinking about it for a while.
The problem is this:
Suppose we have a overlapping node N with 2 children: N<sub>1</sub> and N<sub>2</sub>.
As N<sub>1</sub> and N<sub>2</sub> share some data points, we know that:

   N<sub>1</sub>.NumDescendants() + N<sub>2</sub>.NumDescendants() > N.NumDescendants()

We want to avoid keeping a list of indexes for each non-leaf node. So, we want to access the:
`[0, N.NumDescendants())` set of descendants points of the node N, based on the methods: N<sub>1</sub>.Descendant() and N<sub>2</sub>.Descendant(), setting proper indexes.

For example, we could sort the points so shared points appear at the end of the list (with the greatest indexes), and we record the index `overlapIndex` where the set of duplicated points starts.
Let's call `Shared1` the set of points shared by N<sub>1</sub> and N<sub>2</sub>:

  N<sub>1</sub>:   | ................ | ..... `Shared1` ..... |

  N<sub>2</sub>:   | ................ | ..... `Shared1` ..... |

So, if we want to access to `N.Descendant(i)`, we will do:

```
size_t Descendant(const size_t index) const
{
  if (IsLeaf())
    return (*pointsIndex)[index];
  size_t num = left->overlapIndex;
  if (index < num)
    return left->Descendant(index);
  if (right)
    return right->Descendant(index - num);
}
```

The problem is that this can not be maintained recursively.
Suppose we split N<sub>1</sub> according to another splitting hyperplane, resulting in: N<sub>11</sub> and N<sub>12</sub>.
Let's call `Shared2` the set of points shared by N<sub>11</sub> and N<sub>12</sub>:

  N<sub>11</sub>:   | ................ | ..... `Shared2` ..... | ..... A subset of `Shared1`: S<sub>1</sub>..... |

  N<sub>12</sub>:   | ................ | ..... `Shared2` ..... | ..... A subset of `Shared1`: S<sub>2</sub>..... |

Which value do we record as `overlapIndex`? the index where S<sub>1</sub> starts or the index where `Shared2` starts?

If we consider the index where S<sub>1</sub> starts. We will manage to properly access the descendants points of the node N. But, we won't be able to access to all the descendants points of the node N<sub>1</sub>, because points in S<sub>1</sub> won't be considered.

If we consider the index where `Shared2` starts, we will manage to properly access the descendants points of the node N. But, we won't be able to access to all the descendants points of the node N<sub>1</sub>, because points in S<sub>1</sub> won't be considered.

So, I don't think this can be easily implemented.

I can think of 2 alternative solutions:
+ Don't change anything, and add a note that `NumDescendants()` includes duplicated points. Also we should mencion that Spill Tree shouldn't be considered for `RangeSearch`.
+ Hold a list of indexes in each node, including non-leaf nodes. This would require O(n  logn) space, comparing to current implementation which requires O(n) space.

rcurtin: I think there is an easier way to do this.  We only hold a list of point indices in the leaves, so we only have to worry about the arrangement of points at the leaf level.  Each leaf node only has one sibling, so given two sibling leaf nodes N_l and N_r, let's define a couple sets.  I've just chosen arbitrary set names:

 - S: all the points held in N_l and N_r
 - T: points held in N_r but _not_ in N_l
 - U: points held in both N_l and N_r
 - V: points held in N_l

Now we can arrange the point sets for each child like this:

 - N_l: [V] (order does not matter here)
 - N_r: [T U] (order does matter: all points in T come before all points in U)

If we have set the number of descendants in the parent properly to be |S| (this should be easy), then we can do the following:

```
size_t Descendant(const size_t i)
{
  // Are we a leaf?
  if (IsLeaf())
    return (*pointsIndex)[index];

  if (i > left->NumDescendants())
    return left->Descendant(i);
  else
    return right->Descendant(i - left->NumDescendants());
}
```

The key to this approach is determining the sets T and U, but I think that is straightforward to do either before you recurse into building the children (you can "pre-organize" the set you pass to the right child), or afterwards (you can access the sets of the left and right child and organize them).

When you set `numDescendants` when building a node, that would simply be equal to the number of points held in that node.  This means for a given node, it is possible (and okay) if `node->NumDescendants() < node->Left()->NumDescendants() + node->Right()->NumDescendants()`.

Let me know if I have overlooked anything here.

MarcosPividori: Hi @rcurtin , thanks for your comments.
Yes, I have thought of that approach that you mention, and many similar options. I tried to explain the problem in the previous post, but it is a bit difficult to formulate.
Suppose I implement your approach. And we have a tree of 2 levels of overlapping nodes:
```
 N
 |''''''''''\
N_1          N_2
 |  \         |   \
N_11  N_12   N_21  N_22
```
The problem is how to organize the points in the second level, in the nodes `N_21` and nodes `N_22`.
Suppose we want to get the `i-th` descendant point of the node N, included in N_2, but not included in N_1.
This means when `i >= N_1.NumDescendants()`.

Let U: points held in both N_1 and N_2.

As,  N_2 is splitted in N_21 and N_22 by another different hyperplane, some points of U will be included in N_21 and some points of U will be included in N_22.

If `i < N_1.NumDescendants() + N_21.NumDescendants()`, we will call: `N_21.Descendant( i - N_1.NumDescendants())`

As N_21 includes some points from U, we can repeat a point that was included in the node N_1.

rcurtin: I see what you mean; I did not understand what you wrote the first time.  Sorry about that.  I don't think holding a list of unique points in each node is a good solution, and I can't think of any other easy solution to this problem.  Any way I can think of to organize the points and still get access to unique descendants from arbitrary levels requires caching potentially very much.  So I think unfortunately we should go with your second solution---disable this type of tree for rank-approximate nearest neighbor search or other algorithms that depend on `NumDescendants()` being correct.  Let's do that like this:

 - add a `TreeTraits` element indicating whether `NumDescendants()` returns the number of unique descendants
 - add a `static_assert` in `RASearch` to ensure that with the given tree type, the number of unique descendants is correctly returned by `NumDescendants()`, using that new `TreeTraits` element

What do you think?  I think that's the best we can do here, unfortunately.

rcurtin:
Yes, I see now how defeatist search and hybrid search are different.  I spent a while thinking about how to do hybrid search with other types of trees, and it doesn't really make sense---hybrid search is specific to spill trees only.

I agree that a pure `DefeatistTreeTraverser` would be a nice thing to add.  The literature I have seen---including the extended version of the spill tree paper (which I think is a draft)---does not really comprehensively survey whether or not simple defeatist search can produce reasonable results.  Since defeatist search can be implemented much more efficiently (there is no need for queues or scoring every node), it's possible that with spill trees, you can simply increase the overlap factor tau, and still get fast results.  I can also see how hybrid search can improve results significantly, but I am not sure how much the overhead will affect things.  What do you think?  If you're willing to take the time to implement the `DefeatistTreeTraverser` and some method like `GetNearestChild()`, we can do some testing to see which is the best approach.  (Probably that should be in a different PR, if you decide to go that route.)

src/mlpack/core/tree/spill_tree/spill_tree_impl.hpp
MarcosPividori: Hi @rcurtin ,
Yes, I agree with your proposal. I think it is the best option that we have, unfortunately.
I have implemented that in https://github.com/mlpack/mlpack/pull/747/commits/3b15fdff5817e124b1a33c1dd390da71b2fce8c0.
Thanks!

MarcosPividori:
@sumedhghaisas  I have separated the implementation of different split methods in separate files, and moved all of them to a new directory, so they can be used by other tree types, in: https://github.com/mlpack/mlpack/pull/747/commits/8f3215645f335bf1a96e3ae401c0e66d51517c8b

MarcosPividori:
Hi @rcurtin,

Thanks for your comments.

We can do *"pure defeatist search"* with `SpillTrees`, without implementing a new tree traversal `DefeatistTreeTraversel`.

If we set: `rho = 0.99999`, all nodes will be *overlapping nodes*, and the spill tree's traversers will do defeatist search when visiting each of them.

The problem with that approach, which *hybrid spill tree* aims to solve, is that the tree can be really big. We can't be sure that the tree is reduced by a constant factor.

The default value for *rho* is: `0.7`. If we consider higher values, we will spend more time building the tree, but less time searching for the neighbors. The opposite for lower values of *rho*.

So, I think it only make sense to create a new `DefeatistTreeTraverser` if we want to do defeatist search with other tree types, different to `SpillTrees`. Would you agree?

In that case I don't think we need to implement a new method: `GetNearestChild()` for each Tree Type. We could continue considering the `Score()` method, evaluating each child and choosing the one with lowest score. (We could modify the `DefeatistTreeTraverser` class for `SpillTrees` to avoid calculating the projections of both nodes, calculate the score of one child, and decide based on that score without calculating the score of the second child, as I mentioned before).
The only problem is that Score() would include the unnecessary overhead of CalculateBounds() which will be ignored by *pure defeatist search*. We should think how to avoid that.

What do you think? If you agree I can work on this once we finish with this PR.

Thanks!

rcurtin:
> So, I think it only make sense to create a new DefeatistTreeTraverser if we want to do defeatist search with other tree types, different to SpillTrees. Would you agree?

Yes, I agree, I think that is the right approach.  We can do that in a different PR, like you suggested.  I think single-tree defeatist search is fine, unless you want to try and implement dual-tree defeatist search.  I don't think anyone has derived a way to do that, though I think it wouldn't be too difficult.

> In that case I don't think we need to implement a new method: GetNearestChild() for each Tree Type. We could continue considering the Score() method, evaluating each child and choosing the one with lowest score.

You are right that that would work, but in many cases (like vantage-point trees and random projection trees) we will be able to calculate which node a query point is closer to in less time.  I think it is possible to make the calculation more quickly for kd-trees too but I have not thought about that too in-depth.  Another (ugly) option is to use template metaprogramming along with TreeTraits in order to mark when a tree has a `GetNearestChild()` method available and use it, otherwise use the approach of scoring all the children.  But I am fine with adding another method to the `TreeType` API since I think that will be useful for more than just defeatist search in the future.

src/mlpack/methods/neighbor_search/spill_search_impl.hpp
```diff
@@ -0,0 +1,251 @@
...
+       arma::Mat<size_t>& neighbors,
+       arma::mat& distances)
+{
+  if (Naive() || SingleMode())
+    neighborSearch.Search(querySet, k, neighbors, distances);
+  else
+  {
+    // For Dual Tree Search on SpillTrees, the queryTree must be built with non
+    // overlapping (tau = 0).
+    Tree queryTree(querySet, 0 /* tau */, leafSize, rho);
```
rcurtin: It seems that the entire necessity for the `SpillSearch` class arises from this line; is that correct, or have I overlooked some other difference between `SpillSearch` and `NeighborSearch`?

src/mlpack/core/tree/spill_tree/single_tree_traverser_impl.hpp
```diff
@@ -0,0 +1,138 @@
...
+{
+  // If we are a leaf, run the base case as necessary.
+  if (referenceNode.IsLeaf())
+  {
+    for (size_t i = 0; i < referenceNode.NumPoints(); ++i)
+      rule.BaseCase(queryIndex, referenceNode.Point(i));
+  }
+  else
+  {
+    if(referenceNode.Overlap())
```
rcurtin: This is a minor comment but there is a missing space here after the `if`. :)

src/mlpack/core/tree/spill_tree/single_tree_traverser.hpp
```diff
@@ -0,0 +1,64 @@
...
+
+template<typename MetricType,
+         typename StatisticType,
+         typename MatType,
+         template<typename HyperplaneMetricType> class HyperplaneType,
+         template<typename SplitMetricType, typename SplitMatType>
+             class SplitType>
+template<typename RuleType>
+class SpillTree<MetricType, StatisticType, MatType, HyperplaneType, SplitType>::
+    SingleTreeTraverser
```
rcurtin: The mlpack tree traversers (single-tree and dual-tree) implement the concepts defined as "pruning single-tree traversals" and "pruning dual-tree traversals" in the tree-independent dual-tree algorithms paper.  But in this case the traversals you have here don't actually satisfy that definition because they have defeatist search hard-coded into them.  So if the Rules class doesn't explicitly handle that case, then the traversal is technically invalid.

In the scheme of things this is a really minor point, but I think that it's possible that this may confuse someone way down the line who just says "okay, I want to test my Rules class with all the trees and their default traversers" and suddenly it doesn't work for spill trees and their traversers.  So, I can think of a couple quick solutions:

 * Add a boolean template parameter which, if true, will do the defeatist search as you have implemented it, but it is documented that setting that template parameter to true will make the traverser technically violate the definition.

 * Add documentation stating that the traverser technically violates the definition, providing no way to get a "rigorously correct" traversal.

I am not sure if either of those is the best solution, but I do think it is important that the defeatist search part of this traversal is implemented, otherwise it will be doing two scores when it only needs to do one in those cases.  Maybe you have a better idea for how to handle this?

src/mlpack/core/tree/space_split/hyperplane.hpp
```diff
@@ -0,0 +1,152 @@
...
+
+/**
+ * AxisOrthogonalHyperplane represents a hyperplane orthogonal to an axis.
+ */
+template<typename MetricType>
+using AxisOrthogonalHyperplane = HyperplaneBase<bound::HRectBound<MetricType>,
+    AxisParallelProjVector>;
+
+/**
+ * AxisOrthogonalHyperplane represents a general hyperplane (not necessarily
```
rcurtin: I think the comment is inaccurate, it should probably say just `Hyperplane`. :)

src/mlpack/core/tree/space_split/midpoint_space_split.hpp
```diff
@@ -0,0 +1,45 @@
...
+   * @param points Vector of indexes of points to be considered.
+   * @param hyp Resulting splitting hyperplane.
+   * @return Flag to determine if split is possible.
+   */
+  template<typename HyperplaneType>
+  static bool SplitSpace(
+      const typename HyperplaneType::BoundType& bound,
+      const MatType& data,
+      const arma::Col<size_t>& points,
+      HyperplaneType& hyp);
```
rcurtin: I think you mentioned in #726 that @lozhnikov could use this class for splitting the random projection tree.  I agree, that could be done, but maybe it is worth refactoring this as well as the `MeanSplit` and `MedianSplit` classes in order to have the same API.  The key here would be to make `MatType` a template parameter to the `SplitSpace()` method and then you could pass either an Armadillo contiguous subview (i.e. `mat.cols(begin, begin + count - 1)`) or a non-contiguous subview (i.e. `mat.cols(points)` where `points` is an `arma::uvec` with indices of points).

As you wrote in that ticket, it seems reasonable to do some refactoring and have three template parameters: the type of splitting hyperplane, how to choose the hyperplane, and which splitting value we will consider.  What do you think is the best way to do this refactoring?  Should we make a class that is used by `BinarySpaceTree` that has these three parameters, or should we make `BinarySpaceTree` have those three parameters directly and then make sure we can still make the vantage point tree using those parameters?

Whatever we do will be reverse-incompatible for the "internal API", so it would mean the next version is at least a minor version bump (I'm just fine with that).

src/mlpack/core/tree/space_split/midpoint_space_split_impl.hpp
```diff
@@ -0,0 +1,40 @@
...
+    const typename HyperplaneType::BoundType& bound,
+    const MatType& data,
+    const arma::Col<size_t>& points,
+    HyperplaneType& hyp)
+{
+  typename HyperplaneType::ProjVectorType projVector;
+  double midValue;
+
+  if (!SpaceSplit<MetricType, MatType>::GetProjVector(bound, data, points,
+    projVector, midValue))
```
rcurtin: This should have one more indent. :)

src/mlpack/core/tree/space_split/projection_vector.hpp
```diff
@@ -0,0 +1,148 @@
...
+    return point[dim];
+  };
+
+  /**
+   * Project the given hrect bound on the projection vector.
+   *
+   * @param bound Bound to be projected.
+   * @return Range of projected values.
+   */
+  template<typename MetricType>
```
rcurtin: `HRectBound` also has an `ElemType` template parameter, so either we should add `ElemType` as a template parameter to this function, or to this class.  Otherwise this will not work for trees built with different `ElemType`s.

src/mlpack/core/tree/spill_tree/spill_tree.hpp
```diff
@@ -0,0 +1,429 @@
...
+ private:
+  //! The left child node.
+  SpillTree* left;
+  //! The right child node.
+  SpillTree* right;
+  //! The parent node (NULL if this is the root of the tree).
+  SpillTree* parent;
+  //! The number of points of the dataset contained in this node (and its
+  //! children).
+  size_t count;
```
rcurtin: Could we avoid this member and just use `pointsIndex.n_elem`?

src/mlpack/core/tree/spill_tree/spill_tree.hpp
rcurtin: Nah, `Right()` and `Left()` is just fine.

src/mlpack/core/tree/spill_tree/spill_tree_impl.hpp
```diff
@@ -0,0 +1,761 @@
...
+    overlappingNode(false),
+    hyperplane(),
+    bound(data.n_rows),
+    parentDistance(0), // Parent distance for the root is 0: it has no parent.
+    dataset(&data),
+    localDataset(false)
+{
+  arma::Col<size_t> points(dataset->n_cols);
+  for (size_t i = 0; i < points.n_elem; i++)
+    points[i] = i;
```
rcurtin: You could use `arma::linspace()` here to fill this vector.

src/mlpack/core/tree/spill_tree/spill_tree_impl.hpp
rcurtin: Ok, that looks good to me.  We do also have to handle `RangeSearchRules`, which will prune in some situations and add all descendant points to the results.  We can either add some handling in `RangeSearchRules::AddResult()` to handle when `UniqueNumDescendants` is false and avoid adding any duplicate points, or we can add a static assert to `RangeSearch` like you did for `RASearch`.  Your call, I think either way is easy enough to do.

src/mlpack/core/tree/spill_tree/spill_tree_impl.hpp
```diff
@@ -0,0 +1,761 @@
...
+    if (projections[i] <= 0)
+    {
+      leftPoints[lc] = points[i];
+      lc++;
+    }
+    else
+    {
+      rightPoints[rc] = points[i];
+      rc++;
+    }
```
rcurtin: If you wanted, you could compress this code a lot:

```
if (projections[i] <= 0)
  leftPoints[lc++] = points[i];
else
  rightPoints[rc++] = points[i];
```

Whether or not you want to do that is your choice. :)

src/mlpack/core/tree/spill_tree/typedef.hpp
```diff
@@ -0,0 +1,119 @@
...
+ * This template typedef satisfies the TreeType policy API.
+ *
+ * @see @ref trees, SpillTree, MeanSPTree
+ */
+template<typename MetricType, typename StatisticType, typename MatType>
+using SPTree = SpillTree<MetricType,
+                         StatisticType,
+                         MatType,
+                         AxisOrthogonalHyperplane,
+                         MidpointSpaceSplit>;
```
rcurtin: I'm not sure if `SPTree` is the best name here.  But the name `SpillTree` is already taken.  Maybe you might consider the same type of thing I did for the cover tree, where the class is called `CoverTree` and the standard template alias is called `StandardCoverTree`?  I'm open to ideas here, and maybe there isn't anything better; it just seems counterintuitive to me to compress `Spill` as `SP`.

src/mlpack/methods/neighbor_search/knn_main.cpp
```diff
@@ -78,6 +82,9 @@ PARAM_FLAG("single_mode", "If true, single-tree search is used (as opposed to "
     "dual-tree search).", "S");
 PARAM_DOUBLE_IN("epsilon", "If specified, will do approximate nearest neighbor "
     "search with given relative error.", "e", 0);
+PARAM_STRING_IN("effective_error", "If specified, will compare the results "
+    "against the provided distances file, and will print the average relative "
+    "error.", "E", "");
```
rcurtin: I think we should name this similarly to the parameter to do this for LSH:

```
// For testing recall.
PARAM_STRING_IN("true_neighbors_file", "File of true neighbors to compute "
    "recall with (the recall is printed when -v is specified).", "t", "");
```

In this case you're calculating and printing a slightly different comment, so you'll have to update the comments a bit.  I think you'll need `true_distances_file` instead, and the short option will need to be `-T` since `-t` is already taken.

src/mlpack/methods/neighbor_search/knn_main.cpp
```diff
@@ -64,10 +64,14 @@ PARAM_INT_IN("k", "Number of nearest neighbors to find.", "k", 0);
...
-    "'r-star', 'x', 'ball', 'hilbert-r', 'r-plus', 'r-plus-plus'.", "t", "kd");
+    "'r-star', 'x', 'ball', 'hilbert-r', 'r-plus', 'r-plus-plus', 'spill'.",
+    "t", "kd");
 PARAM_INT_IN("leaf_size", "Leaf size for tree building (used for kd-trees, vp "
-    "trees, R trees, R* trees, X trees, Hilbert R trees, R+ trees and R++ "
-    "trees).", "l", 20);
+    "trees, R trees, R* trees, X trees, Hilbert R trees, R+ trees, R++ trees "
+    "and spill trees).", "l", 20);
+PARAM_DOUBLE_IN("tau", "Overlapping size (for spill trees).", "u", 0);
+PARAM_DOUBLE_IN("rho", "Balance threshold (for spill trees).", "b", 0.7);
```
rcurtin: To emphasize where these are valid, I might consider writing `only valid for spill trees` instead of just `for spill trees`.

src/mlpack/methods/neighbor_search/knn_main.cpp
```diff
@@ -209,12 +245,6 @@ int main(int argc, char *argv[])
     Log::Info << "Loaded kNN model from '" << inputModelFile << "' (trained on "
         << knn.Dataset().n_rows << "x" << knn.Dataset().n_cols << " dataset)."
         << endl;
-
-    // Adjust singleMode and naive if necessary.
-    knn.SingleMode() = CLI::HasParam("single_mode");
-    knn.Naive() = CLI::HasParam("naive");
-    knn.LeafSize() = size_t(lsInt);
-    knn.Epsilon() = epsilon;
```
rcurtin: Why was this removed?  It seems now users don't have a way to load a model and then specify some different search parameters.

src/mlpack/methods/neighbor_search/knn_main.cpp
```diff
@@ -262,6 +292,34 @@ int main(int argc, char *argv[])
...
+          effectiveError += (distances(i) - distancesExact(i)) /
+              distancesExact(i);
+          cases++;
+        }
+      }
+      if (cases)
+        effectiveError /= cases;
+
+      Log::Info << "Effective error: " << effectiveError << endl;
+    }
```
rcurtin: Do you think it would be better to move the effective relative error calculation into the `NeighborSearch` class, perhaps as a static method?  You could pass in the two distance matrices as input and get the effective relative error as output.  That way people could reuse this functionality elsewhere.

src/mlpack/methods/neighbor_search/neighbor_search_rules_impl.hpp
```diff
@@ -375,6 +375,9 @@ inline double NeighborSearchRules<SortPolicy, MetricType, TreeType>::Rescore(
   if (oldScore == DBL_MAX)
     return oldScore;
 
+  if (oldScore == SortPolicy::BestDistance())
+    return oldScore;
```
rcurtin: `SortPolicy::BestDistance()` is `DBL_MAX`, if `SortPolicy = FurthestNeighborSort`, so this statement would end up being meaningless.  I think that what you are meaning to do here is avoid calculating the bound if there is no possibility of pruning, but in that case the score should be `0.0`, so we should probably instead write this:

```
if (oldScore == DBL_MAX || oldScore == 0.0)
  return oldScore;
```

Two small notes:

1. I realized, while writing this comment, that the furthest neighbor search actually did not invert the distances before they were returned!  This means that the runtime of furthest neighbor search was far slower than it needed to be, since the traversal would always recurse into the worst node combination first!  I added some methods to `SortPolicy` to convert distances to and from scores in 1092b1c, so now furthest-neighbor-search is much faster.

2. Technically I think it is possible to prune if the score is 0, so this short-circuit may sometimes overlook those situations: for instance, if you have a nearest neighbor with distance 0, you don't need to look at any other nearest neighbors that could have a distance of 0---you already have a candidate with that distance.  But for our implementation here, bounds like B_2() mean that we cannot prune in those situations.  So my little footnote here is inapplicable to our situation, but I still wanted to point it out. :)

src/mlpack/methods/neighbor_search/ns_model_impl.hpp
```diff
@@ -249,6 +318,9 @@ void NSModel<SortPolicy>::Serialize(Archive& ar,
                                     const unsigned int /* version */)
 {
   ar & data::CreateNVP(treeType, "treeType");
+  ar & data::CreateNVP(leafSize, "leafSize");
+  ar & data::CreateNVP(tau, "tau");
+  ar & data::CreateNVP(rho, "rho");
```
rcurtin: I hate to always be talking about reverse compatibility, but this is reverse-incompatible, since tau, rho, and leafSize aren't part of the previous saved models.  So the thing to do is to uncomment the `version` parameter, and use `version == 0` to assume that leafSize, tau, and rho aren't available (and set them to defaults), and `version == 1` to assume that leafSize, tau, and rho are available.

You'll need to then set the serialization version.  For `LSHSearch`, that's done like this:

```

//! Set the serialization version of the LSHSearch class.
BOOST_TEMPLATE_CLASS_VERSION(template<typename SortPolicy>,
    mlpack::neighbor::LSHSearch<SortPolicy>, 1);
```

src/mlpack/methods/neighbor_search/spill_search_rules_impl.hpp
```diff
@@ -0,0 +1,440 @@
...
+  // Could the existing bounds be better?
+  if (SortPolicy::IsBetter(queryNode.Stat().FirstBound(), worstDistance))
+    worstDistance = queryNode.Stat().FirstBound();
+
+  // Cache bounds for later.
+  queryNode.Stat().FirstBound() = worstDistance;
+
+  worstDistance = SortPolicy::Relax(worstDistance, epsilon);
+
+  return worstDistance;
```
rcurtin: I believe that `CalculateBound()` is the only function that needs to be changed in the rules class; is that right?  Maybe it is better to use some standalone static method to calculate the bound for spill trees, and then something like this in `NeighborSearchRules::CalculateBound()`:

```
if (IsSpillTree<TreeType>::value)
  return CalculateSpillBound(queryNode);
```

If there are other things that differ in this class, too, it might be necessary to use this same trick there.  I see this kind of as a variant of the `#ifdef` style of making things compatible, so it's kind of hackish, but I think maybe it is better than replicating the entire class in order to change only a little functionality.  What do you think?

src/mlpack/tests/aknn_test.cpp
```diff
@@ -242,6 +242,47 @@ BOOST_AUTO_TEST_CASE(DualBallTreeTest)
...
+  dataset.randu(50, 300); // 50 dimensional, 300 points.
+
+  const size_t k = 3;
+
+  KNN exact(dataset);
+  arma::Mat<size_t> neighborsExact;
+  arma::mat distancesExact;
+  exact.Search(dataset, k, neighborsExact, distancesExact);
+
+  double max_dist = 0;
```
rcurtin: I know it is picky but we should name this `maxDist` not `max_dist` for consistency.

src/mlpack/tests/hyperplane_test.cpp
```diff
@@ -0,0 +1,128 @@
...
+  BOOST_REQUIRE(!h1.Left(b1));
+
+  b1.Center() = arma::vec("0 0");
+  b1.Radius() = 1.41;
+  BOOST_REQUIRE(!h1.Right(b1));
+  BOOST_REQUIRE(!h1.Left(b1));
+
+  // AxisParallel hyperplane.
+  AxisParallelProjVector projVect2(1);
+  AxisOrthogonalHyperplane<EuclideanDistance> h2(projVect2, 1);
```
rcurtin: Maybe we should split this into two separate tests.

src/mlpack/tests/knn_test.cpp
```diff
@@ -889,6 +889,123 @@ BOOST_AUTO_TEST_CASE(DualBallTreeTest)
...
+  dataset.randu(50, 300); // 50 dimensional, 300 points.
+
+  const size_t k = 3;
+
+  KNN naive(dataset);
+  arma::Mat<size_t> neighborsNaive;
+  arma::mat distancesNaive;
+  naive.Search(dataset, k, neighborsNaive, distancesNaive);
+
+  double max_dist = 0;
```
rcurtin: The same comment about `max_dist` applies here, too.

src/mlpack/tests/knn_test.cpp
```diff
@@ -889,6 +889,123 @@ BOOST_AUTO_TEST_CASE(DualBallTreeTest)
...
+  {
+    switch (test)
+    {
+      case 0:
+        tau = 0;
+        break;
+      case 1:
+        tau = 0.1;
+        break;
+    }
```
rcurtin: `tau = 0.1 * test`?

src/mlpack/tests/knn_test.cpp
```diff
@@ -889,6 +889,123 @@ BOOST_AUTO_TEST_CASE(DualBallTreeTest)
...
+      {
+        case 0: // Single Tree Search.
+          spTreeSearch.Naive() = false;
+          spTreeSearch.SingleMode() = true;
+          break;
+        case 1: // Dual Tree Search.
+          spTreeSearch.Naive() = false;
+          spTreeSearch.SingleMode() = false;
+          break;
+      }
```
rcurtin: You could simplify:

```
spTreeSearch.SingleMode() = (mode == 0);
```

I don't think setting `Naive()` is necessary.

src/mlpack/tests/spill_tree_test.cpp
```diff
@@ -0,0 +1,247 @@
...
+      case 0:
+        tau = 0;
+        break;
+      case 1:
+        tau = 0.05;
+        break;
+      case 2:
+        tau = 0.1;
+        break;
+    }
```
rcurtin: `tau = 0.05 * cases`?

src/mlpack/tests/spill_tree_test.cpp
```diff
@@ -0,0 +1,247 @@
...
+  arma::mat dataset = arma::randu<arma::mat>(3, 1000);
+  typedef SPTree<EuclideanDistance, EmptyStatistic, arma::mat> TreeType;
+
+  TreeType tree(std::move(dataset));
+
+  BOOST_REQUIRE_EQUAL(dataset.n_elem, 0);
+  BOOST_REQUIRE_EQUAL(tree.Dataset().n_rows, 3);
+  BOOST_REQUIRE_EQUAL(tree.Dataset().n_cols, 1000);
+}
+
```
rcurtin: Should we also test the copy constructor?  You could do this:

```
TreeType t = new TreeType(dataset);
TreeType t2(t);
delete t;

// Now check that everything is still valid.
```

rcurtin:
Ok, no more comments from my side (I think!).  This is a really nice contribution and I am looking forward to when it is merged in. :)

MarcosPividori:
@rcurtin, Thanks for your feedback!

In response to the comment: https://github.com/mlpack/mlpack/pull/747#issuecomment-239602549
A `defeatist single tree traverser` would perform a `greedy search` that chooses the child with the best score and doesn't do backtracking. We could also call it: `SingleTreeGreedyTraverser`
The problem of using `GetNearestChild()` inside the traverser instead of a `RuleType` template parameter, is that that traverser will only be useful for *neighbor search*.
If we continue using the `Score()` function from the `RuleType` class, we can use it for other purposes such as: *Defeatist Farthest Neighbor search*, or any other algorithm that uses the single tree traverser and we want to provide a greedy alternative.

I think we could solve this problem this way:
+ Add a new method to the `RuleType` class:
```
TreeType& GetChildWithBestScore(
    const size_t queryIndex,
    const TreeType& referenceNode);
```
+ `SingleTreeGreedyTraverser` will always call `GetChildWithBestScore()` and continue with that child until it reaches a leaf node.
+ We could implement `NeighborSearchRules::GetChildWithBestScore()` as :
```
TreeType& NeighborSearchRules::GetChildWithBestScore(
    const size_t queryIndex,
    const TreeType& referenceNode)
{
  SortPolicy::GetBestChild(dataset.col(queryIndex), referenceNode);
}
```
```
TreeType& NearestNeighborSort::GetBestChild(
    const VecType& queryPoint,
    const TreeType& referenceNode)
{
   return referenceNode.GetNearestChild(queryPoint);
}

TreeType& FurthestNeighborSort::GetFurthestChild(
    const VecType& queryPoint,
    const TreeType& referenceNode)
{
   return referenceNode.GetFurthestChild(queryPoint);
}
```

And the trivial implementation of that method for other RuleTypes would be:
```
TreeType& SomeRuleType::GetChildWithBestScore(
    const size_t queryPoint,
    const TreeType& referenceNode)
{
  size_t best_child = size_t() - 1;
  size_t best_score = DBL_MAX;
  for (size_t i = 0; i < referenceNode.NumChildren(); i++)
  {
    double score = Score(queryPoint, referenceNode.Child(i));
    if (score < best_score)
    {
      best_score = score;
      best_child = i;
    }
  }
  return referenceNode.Child(best_child);
}
```

What do you think? This way we can take advantage of different tree types when it is possible to calculate which node a query point is closer to in less time. And, at the same time, we can use the `SingleTreeGreedyTraverser` for other purposes.

rcurtin:
You're right, we will need to encapsulate the function that gets the best child in the `RulesType` class.  Perhaps, in this case, it is best to create a standalone rules class for defeatist nearest neighbor (or furthest neighbor) search, in order to keep the complexity of `NeighborSearchRules` down.  What do you think?

src/mlpack/core/tree/spill_tree/single_tree_traverser_impl.hpp
MarcosPividori: Yes, you are right. I updated many files with the same situation in: https://github.com/mlpack/mlpack/pull/747/commits/e06cfc8f950430060506bc895a94f69b1f3fb29d

src/mlpack/core/tree/space_split/hyperplane.hpp
MarcosPividori: Yes, thanks. Done in: https://github.com/mlpack/mlpack/pull/747/commits/722d2115d5e43fb0f56c12b97b79649ca8b9f45d

src/mlpack/core/tree/space_split/midpoint_space_split_impl.hpp
MarcosPividori: Done in : https://github.com/mlpack/mlpack/pull/747/commits/89725f4f03de713674caf93ff52b033e81f1756b

src/mlpack/core/tree/spill_tree/spill_tree_impl.hpp
MarcosPividori: Done in: https://github.com/mlpack/mlpack/pull/747/commits/075795541bf05a687f18d55083f58e80d48fa79a

src/mlpack/core/tree/spill_tree/spill_tree_impl.hpp
MarcosPividori: Yes, I agree. Done in: https://github.com/mlpack/mlpack/pull/747/commits/dce67ec82adedd7be4c113eea7dec95385f755aa

src/mlpack/methods/neighbor_search/knn_main.cpp
MarcosPividori: @rcurtin , yes I agree. I added 2 new command line options in https://github.com/mlpack/mlpack/pull/747/commits/92155154bc8721055702ae3d22c33542f2f07a61 and https://github.com/mlpack/mlpack/pull/747/commits/3e79d273545cb5d5694f6c8e7370a0dbbd95e5e6:
```
PARAM_STRING_IN("true_distances_file", "File of true distances to compute "
    "the effective error (average relative error) (it is printed when -v is "
    "specified).", "D", "");
PARAM_STRING_IN("true_neighbors_file", "File of true neighbors to compute the "
    "recall (it is printed when -v is specified).", "T", "");
```
So "-T" will print the *Recall*, similar to "-t" for *LSH*.

src/mlpack/methods/neighbor_search/knn_main.cpp
MarcosPividori: @rcurtin , I have implemented this in https://github.com/mlpack/mlpack/pull/747/commits/92155154bc8721055702ae3d22c33542f2f07a61 and https://github.com/mlpack/mlpack/pull/747/commits/3e79d273545cb5d5694f6c8e7370a0dbbd95e5e6.

src/mlpack/methods/neighbor_search/spill_search_impl.hpp
MarcosPividori: @rcurtin Yes, that is true.

src/mlpack/methods/neighbor_search/knn_main.cpp
MarcosPividori: I agree. Done in: https://github.com/mlpack/mlpack/pull/747/commits/99de521027ecb6d99ae533096667be591fd7f62f

src/mlpack/tests/aknn_test.cpp
MarcosPividori: Done in: https://github.com/mlpack/mlpack/pull/747/commits/267265b0679f892be50ee0ba98948552ef1784ac

src/mlpack/tests/knn_test.cpp
MarcosPividori: Done in: https://github.com/mlpack/mlpack/pull/747/commits/ac5f836691b619c679b23c9fe37eecc0c7b00038

src/mlpack/tests/hyperplane_test.cpp
MarcosPividori: I agree. Done in: https://github.com/mlpack/mlpack/pull/747/commits/207c1d428e3ca1028144369b98d15e5c7cb9abd4

src/mlpack/tests/knn_test.cpp
MarcosPividori: Done in: https://github.com/mlpack/mlpack/pull/747/commits/471bb24dc2d96620e27a7542cbba8dc8dae7ca29

src/mlpack/tests/spill_tree_test.cpp
MarcosPividori: Ok. Done in: https://github.com/mlpack/mlpack/pull/747/commits/f1253d37af50ab7df14a2c841c0a7345cec140e1

src/mlpack/tests/knn_test.cpp
MarcosPividori: Done in https://github.com/mlpack/mlpack/pull/747/commits/f989164147b33ba87192c311fff19622cc59f714

src/mlpack/tests/spill_tree_test.cpp
MarcosPividori: Done in https://github.com/mlpack/mlpack/pull/747/commits/f989164147b33ba87192c311fff19622cc59f714

src/mlpack/methods/neighbor_search/spill_search_rules_impl.hpp
MarcosPividori: @rcurtin , The `Score()` calculation is different too, because I consider the special case of overlapping nodes. As there are 3 different methods, I thought it would preferable to have a different implementation, but I agree this means some duplicated code. I was not sure what was preferable, some duplicated code with the advantage of not modifying the original search rules, or more compact code but a bit more complex.

As I have to access the `Overlap()` method of Spill Trees and it is not included in all tree types, I can't use the `#ifdef` approach, because we will have a compiler error when instantiating with a different tree type. I think I should consider template specialization for the Score() methods.

src/mlpack/methods/neighbor_search/spill_search_impl.hpp
rcurtin: Ok, in this case, it is a lot of work to duplicate the entire class.  Instead I think we should use template specialization, but we can't specialize only a single member of a template class, so instead you'll need to write an auxiliary method kinda like:

```
template<typename TreeType>
TreeType* BuildQueryTree(arma::mat& dataset, boost::disable_if<IsSpillTree<TreeType>>::value*)
{
  return new TreeType(dataset);
}

template<typename TreeType>
TreeType* BuildQueryTree(const arma::mat& dataset, boost::enable_if<IsSpillTree<TreeType>>::value)
{
  return new TreeType(dataset, 0, leafSize, rho);
}
```

I've omitted a lot of details and there are probably syntax errors but the basic idea is to just use SFINAE in a standalone helper function.  I guess maybe it would have been easier for me to just say "check out `BuildTree()` at the top of neighbor_search_impl.hpp"...! :)

src/mlpack/core/tree/space_split/hyperplane.hpp
rcurtin: Nice, that fixed a lot of other places that had the same issue too.  Thanks! :)

src/mlpack/methods/neighbor_search/spill_search_rules_impl.hpp
rcurtin: Ah, I didn't mean use `#ifdef` directly, I just meant that the idea of using standalone functions with template specializations feels kinda hackish, in the same way that `#ifdef`s feel kinda hackish.

My thought here is that someone who is new to this code and wants to understand it will come along and get overwhelmed the more files they see (or, really, the more complexity that they perceive).  So in that sense, sometimes I think it is better to call out to simple standalone functions.  In this case, using template specializations for those would help identify specifically what in the calculation is different.  What do you think?

src/mlpack/core/tree/spill_tree/spill_tree_impl.hpp
```diff
@@ -0,0 +1,753 @@
...
+    bound(data.n_rows),
+    parentDistance(0), // Parent distance for the root is 0: it has no parent.
+    dataset(&data),
+    localDataset(false)
+{
+  arma::Col<size_t> points;
+  if (dataset->n_cols > 0)
+    // Fill points with all possible indexes: 0 .. (dataset->n_cols - 1).
+    points = arma::linspace<arma::Col<size_t>>(0, dataset->n_cols - 1,
+        dataset->n_cols);
```
rcurtin: Maybe just using `regspace()` is the better thing to do here, but I'm indifferent either way.

MarcosPividori:
Hi @rcurtin,
I have modified the SpillTree's Traversers. Now they take a template boolean parameter `Defeatist`, that determines if the traverser must consider defeatist search on overlapping nodes. In the commits:
https://github.com/mlpack/mlpack/pull/747/commits/0267acb46caef51370e737cf253a1664ea7be263
https://github.com/mlpack/mlpack/pull/747/commits/70fbeab1e0be29113f0e21c688601133c23d4751
https://github.com/mlpack/mlpack/pull/747/commits/b334674a12817b3e4b3f37ac6f99eb52bfd3e44c

src/mlpack/methods/neighbor_search/spill_search_impl.hpp
MarcosPividori: @rcurtin Ok, I see what you mean. But, in the last commits, I also define different single and dual tree traversers for defeatist search. So, we can't consider the default traversers for defeatist search, as `NeighborSearch` does.
Because of that I think we should continue considering a different class for `SpillSearch`.
So the main difference with `NeighborSearch` class is:
 + It includes the parameters `LeafSize`, `Tau` and `Rho`, which are specific for SpillTrees.
 + It ensures we build the query tree with non overlappin (tau = 0).
 + It ensures we use the `DefeatistSingleTreeTraverser` and `DefeatistDualTreeTraverser`.

If we want to include all of this in the `NeighborSearch` class, I think it can become a bit complex ....
What do you think?

src/mlpack/methods/neighbor_search/spill_search_rules_impl.hpp
MarcosPividori: @rcurtin I agree. I have removed the specialization of `NeighborSearchRules` for `SpillTrees` in: https://github.com/mlpack/mlpack/pull/747/commits/e4ce9be6795817f7ceebafc26dd2b24c38a356f9 and https://github.com/mlpack/mlpack/pull/747/commits/2bad753ca4859b6b2067d60b4323dcab5438959e.
+ I moved the specific `Score()` calculation on overlapping nodes to the `Defeatist Traversers`.
+ I defined a simple struct `IsSpillTree`, and use it inside `CalculateBound()`, to avoid considering *B_2* bound for `SpillTrees`.

src/mlpack/core/tree/spill_tree/single_tree_traverser.hpp
MarcosPividori: Hi @rcurtin,
Thanks for your comments, I agree!
I have modified the SpillTree's Traversers. Now they take a template boolean parameter `Defeatist`, that determines if the traverser must consider defeatist search on overlapping nodes. In the commits:
https://github.com/mlpack/mlpack/pull/747/commits/0267acb46caef51370e737cf253a1664ea7be263
https://github.com/mlpack/mlpack/pull/747/commits/70fbeab1e0be29113f0e21c688601133c23d4751
https://github.com/mlpack/mlpack/pull/747/commits/b334674a12817b3e4b3f37ac6f99eb52bfd3e44c

So, for `SpillTrees`, I provide 4 tree traversers:
+`SingleTreeTraverser`   implements a normal single tree traverser.
+`DualTreeTraverser`   implements a normal dual tree traverser.
+`DefeatistSingleTreeTraverser`   does defeatist search on overlapping nodes.
+`DefeatistDualTreeTraverser`   does defeatist serach on overlapping node.

src/mlpack/methods/neighbor_search/spill_search_impl.hpp
MarcosPividori: I have removed the specialization of `NeighborSearchRules` for `SpillTrees`, so we don't have much duplicated code now.

src/mlpack/core/tree/spill_tree/spill_tree_impl.hpp
MarcosPividori: @rcurtin,
I am facing an error: error: `‘regspace’ is not a member of ‘arma'`
Maybe it was included in a very recent version of armadillo...

src/mlpack/core/tree/spill_tree/typedef.hpp
MarcosPividori: Hi @rcurtin , I considered `SPTree` because in the original paper [[1]](http://www.cs.cmu.edu/~agray/approxnn.pdf), "spill-tree"  is abbreviated to "sp-tree".
Mmm I am not really good with names.

src/mlpack/core/tree/space_split/projection_vector.hpp
MarcosPividori: Done in: https://github.com/mlpack/mlpack/pull/747/commits/b55534d084fb56f1f4390e36bece99a82ac3529f

src/mlpack/methods/neighbor_search/knn_main.cpp
MarcosPividori: @rcurtin
I removed all that options because they are incorrect:
+ For example if we save a model in `SingleMode`, and we reload that model with no extra option, it will set:
```
  knn.SingleMode() = CLI::HasParam("single_mode") ; // false
  knn.Naive() = CLI::HasParam("naive")  ;  // false
```
  It will do `Dual Tree Search` instead of `Single Tree Search`.
+ Also, if we set the `LeafSize()` to a different value, we need to call `BuildModel` again, to rebuild the reference tree.
+ Also, in the parameter specifications, it is mentioned that "--leaf_size" "--naive" and "--single_mode" would be ignored when providing an input model file. So, I wanted to be consistent with that comments.

So, it is a bit difficult to properly update a given input model.
For example, if users provide an input model file, without the options "--naive" and "--single_mode", does it means they don't want to modify the model or that they want to update it to do "dual tree search"?

MarcosPividori: A possible approach is this:
"When providing an input model file, all the extra options that are provided will be updated in the given model."
With the only limitation that it is not possible to modify a single tree model to a dual tree model, because we don't have a "--dual_mode" command line option.

MarcosPividori: Also, another comment, I think we should provide a command line option:
```
  --mode (-m) [string]
```
So, we can specify different search modes:  *naive*,  *single_mode*, *dual_mode*, *greedy*

Then, when an input model file is provided, we could check if the option "--mode" was specified. In that case, we update the input model file to the given mode.

src/mlpack/methods/neighbor_search/ns_model_impl.hpp
MarcosPividori: I agree. Done in: https://github.com/mlpack/mlpack/pull/747/commits/2698225eabde8bf2022d62209b965509d3207f73

src/mlpack/core/tree/spill_tree/typedef.hpp
rcurtin: Ah, I see, I guess in that case I disagree with the naming in the paper too, since "SP" is not really an abbreviation.  But since it's what they do in the paper, I guess I am fine with either way, so if you want to change it, feel free, but if not, don't worry about it. :)

src/mlpack/methods/neighbor_search/knn_main.cpp
rcurtin: Great, sounds good to me.  I guess something worth doing might be printing a warning if the user specifies either `-T` or `-D` but is doing a search that gives exact results.  If you like, I can write the code to do that, since you have already made very many changes. :)

src/mlpack/methods/neighbor_search/knn_main.cpp
rcurtin: I see what you mean.  There are really two separate concerns in the comments you made; I'll address them in reverse order:

I agree that a `-m` option would be a nice idea, since the number of options for these neighbor searching programs we have are exploding as we add so many types of trees and types of search!  Unfortunately we will have some reverse compatibility to contend with, but it should not be a huge issue, just annoying.

Onto the other issue, allowing the user to specify some options at query time can provide functionality we don't otherwise have: for instance, the user can specify a query tree with a different leaf size, and more importantly, a user can train a model (i.e. build a reference tree) once, and then use that saved model to do both single-tree and dual-tree search.  This can save a lot of time, for instance if the tree took a long time to build (which is often the case with cover trees).  This is also true of the epsilon parameter---I can do nearest neighbor search with a lot of different epsilon levels.

So, I guess, the documentation was wrong about what was actually happening before.  A better way to put it would be this:  If `--single_mode`, `--naive`, `--leaf_size`, or `--epsilon` are specified along with `--input_model_file`, then these options will be taken into account for the search.  For `--leaf_size`, the option will only apply to the reference tree.  The new model preferences will not be saved unless `--output_model_file` is specified.

Really there are two different types of options there; one can be encapsulated by the `--mode` parameter you suggested; the other (`--leaf_size`) is a specific parameter for the query tree.  I am willing to believe that the use-case for different-leaf-size query trees is so niche that nobody is going to do it, and we can force the user to say that if they are going to build a query tree, then it will be built with the same options as the reference tree, and if they don't like that, then they will need to write C++ and build their trees themselves.  But we can't remove the option to allow the user to specify a different search mode.

If you prefer, for the second part, we can leave that for a separate issue, leave the code in currently to set `LeafSize()` and the other parameters, and then open another Github issue and we can work that out separately.

src/mlpack/core/tree/spill_tree/spill_tree_impl.hpp
rcurtin: I see, `regspace()` was introduced in 6.700, so let's not worry about that now.  Maybe someday when the minimum supported Armadillo version is that or higher we can make all the changes, but it is too much work to backport it for such a minor reason.

src/mlpack/methods/neighbor_search/spill_search_rules_impl.hpp
rcurtin: Thanks!  The changes look really nice.  It seems that moving the `Score()` calculation into the defeatist traversers ties those traversers to the nearest neighbor search problem.  But, it also seems that the other PR that is open for greedy traversers introduces some new functions into the Rules classes to handle that, and once we merge the other PR we can easily refactor the defeatist traverser here and it will be abstract enough to handle any rules case again.  (...as long as that rules case can support greedy search)  What do you think?  Is that a reasonable plan?

src/mlpack/methods/neighbor_search/neighbor_search_rules_impl.hpp
MarcosPividori: @rcurtin 
Yes, I completely agree. I added this modification to avoid calculating the bounds as there is no possibility of prunning when score is `0`. I agree that it would be possible to prune when score is 0, but we can't because of `B_2()`.
Updated the code in : https://github.com/mlpack/mlpack/pull/747/commits/e4cf1fd252b710809654515dccc30a299d0380e2

src/mlpack/methods/neighbor_search/knn_main.cpp
MarcosPividori: @rcurtin
Ok, I understand.
I updated the code to consider `--single_mode`, `--naive`, `--leaf_size`, and  `--epsilon`, and update that values in the input model (https://github.com/mlpack/mlpack/pull/747/commits/ecaf1bf80976b83044cf9d90b65aa9a02e4efba1).
The leaf size specified will be only considered for que query tree. I added a new comment about this (because it doesn't make sense to rebuild the input model file). Would you agree?
If you agree, I will add the `--mode` option once the PR of Greedy Search is merged.
Thanks!

src/mlpack/methods/neighbor_search/knn_main.cpp
MarcosPividori: Sure! Done in: https://github.com/mlpack/mlpack/pull/747/commits/c5fe485b25e93f35b14f2c9fbc6b69ddbe59a837

src/mlpack/methods/neighbor_search/spill_search_rules_impl.hpp
MarcosPividori: @rcurtin Yes, I was thinking of doing exactly what you suggested. :)

src/mlpack/methods/neighbor_search/spill_search_impl.hpp
rcurtin: When I originally designed the `NeighborSearch` class, my intention was to make it as tree-independent as possible (or, well, that's what happened in the end).  So it doesn't hold any `leafSize` parameter despite the fact that the `BinarySpaceTree` uses it.  The policy has always been, `NeighborSearch` will build your trees with the default parameters, and if you don't want the default parameters, you must pass the trees.  I think that we should continue that policy here, so I don't think it's advisable to have `leafSize`, `rho`, and `tau` stored in the class.  If the user wants to specify those, they can build the tree themselves.  (This means that `NSModel` will need to store those parameters, if they can be specified from the command-line.)

For the traversers, an idea is to add two more template parameters to `NeighborSearch` that specify the desired single-tree and dual-tree traversers (and default to the defaults of the TreeType).

Then we could typedef `SpillTreeKNN = NeighborSearch<EuclideanDistance, arma::mat, SPTree, SPTree::DefeatistSingleTreeTraverser, SPTree::DefeatistDualTreeTraverser>` (or whatever the names and syntax may be).  This, plus the template specialization I suggested for building the query tree for spill tree search, should be sufficient to avoid the need for the `SpillSearch` class at all, and keep the number of files and classes down.  What do you think?  Have I overlooked a detail?

src/mlpack/methods/neighbor_search/knn_main.cpp
rcurtin: Looks great to me.  Thanks for making the changes!

src/mlpack/methods/neighbor_search/ns_model_impl.hpp
rcurtin: Looks good.  I noticed that the previous revision did not serialize `leafSize`, so this fixes a bug too!

PR #748, opened 2016-08-02
Adds a Train() function that only needs dataset statistics

mentekid:
Instead of using the dataset itself, this Train() function only needs the dataset statistics. This function works for 1-dimensional distributions only (you can only provide 1 number for each statistic).

This is useful for my code in LSHModel, where instead of using data to fit the distribution, we create a regression function that can predict the arithmetic and geometric mean of squared distances given the size of the dataset. Since we don't have the actual distances - we only have an estimation of the distances' statistics - we can't use the previous Train() function in GammaDistribution.

I modified my code so that the Train() that accepts the dataset simply calls this function for each row after computing the statistics, to avoid code reuse.

I also added a test to make sure this produces the same result as giving the Train() function the dataset.

### DIFF
--- a/src/mlpack/core/dists/gamma_distribution.cpp
+++ b/src/mlpack/core/dists/gamma_distribution.cpp
@@ -43,9 +43,11 @@ void GammaDistribution::Train(const arma::mat& rdata, const double tol)
   using boost::math::trigamma;
   using std::log;
 
-  // Allocate space for alphas and betas (Assume independent rows).
-  alpha.set_size(rdata.n_rows);
-  beta.set_size(rdata.n_rows);
+  // Allocate temporary space for alphas and betas (Assume independent rows).
+  // We need temporary vectors since we are calling Train(logMeanx, meanLogx,
+  // meanx) which modifies the object's own alphas and betas.
+  arma::vec tempAlpha(rdata.n_rows);
+  arma::vec tempBeta(rdata.n_rows);
 
   // Calculate log(mean(x)) and mean(log(x)) of each dataset row.
   const arma::vec meanLogxVec = arma::mean(arma::log(rdata), 1);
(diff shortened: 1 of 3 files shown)


src/mlpack/core/dists/gamma_distribution.cpp
```diff
@@ -55,36 +57,63 @@ void GammaDistribution::Train(const arma::mat& rdata, const double tol)
   // Treat each dimension (i.e. row) independently.
   for (size_t row = 0; row < rdata.n_rows; ++row)
   {
-    // Statistics for this row.
+    // Statistics for this Vrow.
```
rcurtin: Is this an accidental V? :)

src/mlpack/core/dists/gamma_distribution.cpp
```diff
@@ -55,36 +57,63 @@ void GammaDistribution::Train(const arma::mat& rdata, const double tol)
...
+  }
+  alpha = tempAlpha;
+  beta = tempBeta;
+
+}
+
+// Fits an alpha and beta parameter to each dimension of the data.
+void GammaDistribution::Train(const double logMeanx, const double meanLogx,
+                              const double meanx,
+                              const double tol)
```
rcurtin: Hmm, do you think that instead of taking `const double`s here, we should take `const arma::vec&`, and fit all dimensions at once? That would remove the need to call this overload multiple times in the `Train()` overload that takes data, and `tempAlpha` and `tempBeta` would be unnecessary in that method.

rcurtin:
Looks good to me; once we address the comments I'll merge it in.

src/mlpack/core/dists/gamma_distribution.cpp
mentekid: You're right - I was concerned with my design using temps and having two functions changing the vectors continuously. What you propose is cleaner. Thanks :)

src/mlpack/core/dists/gamma_distribution.cpp
mentekid: Woops

PR #749, opened 2016-08-04
Modeling LSH For Performance Tuning

mentekid:
This is in no way ready to merge, but I've done a bit of progress on the LSH tuning class, so I'm opening a PR to track how things are going.

With regards to implementation (in order for the algorithm to be complete), I need to
- Write the code that fits Gamma distributions to distances.
- Define the objective functions used for integration to calculate recall and precision.
- Write the code that integrates the objective functions.
- Implement the binary search that identifies useful parameter sets.

With regards to functionality (in order for this to be usable), I also need to
- Identify any hard-coded things that should become input parameters.
- Handle input and output.
- Write meaningful tests.
- Potentially backport boost integration functionality.

Then finally there's the step of profiling & optimization of the code.


src/mlpack/methods/CMakeLists.txt
```diff
@@ -36,6 +36,7 @@ set(DIRS
   local_coordinate_coding
   logistic_regression
   lsh
+  lsh_model
```
rcurtin: Do you think we should just put the LSH modeling code in the `lsh` directory?  I don't see a problem with keeping them together.

src/mlpack/methods/lsh_model/lshmodel.hpp
```diff
@@ -0,0 +1,373 @@
...
+ *  pages = {669},
+ *  url = {http://portal.acm.org/citation.cfm?doid=1458082.1458172},
+ *  year = {2008}
+ * }
+ * @endcode
+ *
+ * We use a different method to fit Gamma Distributions to pairwise distances.
+ * Instead of the MLE method proposed in the paper above, we use the mlpack
+ * class GammaDistribution, which implements fitting according to Thomas Minka's
+ * work.
```
rcurtin: I think that Minka's method actually does return the MLE (as your convergence criterion tends to 0), so we are still estimating the MLE, just using a clever algorithm.  Correct me if I'm wrong about that.

src/mlpack/methods/lsh_model/lshmodel.hpp
```diff
@@ -0,0 +1,373 @@
...
+template <
+    typename SortPolicy = NearestNeighborSort,
+    typename ObjectiveFunction = DefaultObjectiveFunction
+      >
+class LSHModel
+{
+ public:
+
+   //! Empty Constructor. Do nothing
+   LSHModel(){ referenceSet = NULL; };
```
rcurtin: Maybe consider using initialization lists instead:

```
LSHModel() : referenceSet(NULL) { }
```

src/mlpack/methods/lsh_model/lshmodel.hpp
```diff
@@ -0,0 +1,373 @@
...
+       * @param inputSize A vector of input sizes. The first input variable of 
+       *     the regression.
+       * @param kValues A vector of k values. The second input variable of the
+       *     regression.
+       * @param statistic A vector of responses - the value of the statistic for
+       *     each given inputSize.
+       */
+      double Train(const arma::Col<size_t>& inputSize, 
+                 const arma::Col<size_t>& kValues,
+                 const arma::mat& statistic);
```
rcurtin: The spacing is off a little bit here.

src/mlpack/methods/lsh_model/lshmodel.hpp
```diff
@@ -0,0 +1,373 @@
...
+                                 const arma::mat& Ek, 
+                                 const arma::mat& Gk);
+
+   /**
+    * This is a helper class that uses the function a * k^b * N^c for some
+    * parameters a, b, c that have been fit to either predict the arithmetic or
+    * geometric mean of the squared distance of a point to its k-nearest
+    * neighbor, given some dataset size N and its k-nearest neighbor.
+    */
+   class DistanceStatisticPredictor
```
rcurtin: If you like, you can make this file smaller by moving `DistanceStatisticPredictor` into a different file and making it not a nested class of `LSHModel`.  I'm ambivalent either way.

src/mlpack/methods/lsh_model/lshmodel.hpp
```diff
@@ -0,0 +1,373 @@
...
+       *     statistic.
+       */
+      double Predict(size_t N, size_t k) 
+      { return alpha * std::pow(k, beta) * std::pow(N, gamma); };
+
+      //! Set the alpha parameter.
+      void Alpha(double a) { alpha = a; };
+
+      //! Get the alpha parameter.
+      double Alpha(void) { return alpha; };
```
rcurtin: No need to mark `void` here, that's implicit.  For the other functions, I might consider marking the parameters const (like `const size_t N` instead of `size_t N`) even though it does nothing at all.  At least in mlpack, input options are generally denoted with a const specifier.  You can go either way here, since like I said it does nothing at all.

src/mlpack/methods/lsh_model/lshmodel_impl.hpp
```diff
@@ -0,0 +1,505 @@
...
+// Constructor sets variables and trains the object.
+template <typename SortPolicy, typename ObjectiveFunction>
+LSHModel<SortPolicy, ObjectiveFunction>::
+LSHModel(const arma::mat &referenceSet,
+         const double sampleSize,
+         const size_t k)
+{
+  // We don't own the set - we just point to it.
+  ownsSet = false;
+  this->referenceSet = &referenceSet;
```
rcurtin: Initialization lists might be nicer to use here, too.

src/mlpack/methods/lsh_model/lshmodel_impl.hpp
```diff
@@ -0,0 +1,505 @@
...
+  // Sanity check - sample rate must be in (0, 1].
+  if (sampleRate > 1 || sampleRate <= 0)
+    Log::Fatal << "Sampling rate must be floating point number in (0, 1]"
+        << std::endl;
+
+  // Update the object's max K value information.
+  maxKValue = k;
+
+  // Save pointer to training set.
+  this->referenceSet = &referenceSet;
```
rcurtin: Don't forget to update `ownsSet` here, and possibly delete the old reference set first.

src/mlpack/methods/lsh_model/lshmodel_impl.hpp
```diff
@@ -0,0 +1,505 @@
...
+  // Step 2. Compute all-vs-all distances of points in the sample.
+  // The distance matrix is symmetric, so we only compute elements above the
+  // diagonal. There are (N * (N - 1)) / 2 such elements.
+  Timer::Start("pairwise_distances");
+  arma::vec distances(numSamples * (numSamples - 1) / 2);
+  size_t d = 0; // Index of where to store next.
+  for (size_t i = 0; i < numSamples; ++i)
+    for (size_t j = i + 1; j < numSamples; ++j)
+      distances(d++) = metric::EuclideanDistance::Evaluate(
+          sampleSet.unsafe_col(i), sampleSet.unsafe_col(j));
```
rcurtin: Nice, I like the optimization to only compute less than half the elements in the distance matrix here.

src/mlpack/methods/lsh_model/lshmodel_impl.hpp
```diff
@@ -0,0 +1,505 @@
...
+
+  // Number of samples to create for modeling the Gamma Distributions
+  size_t regressionExamples = 50; // TODO: parameter?
+
+  // Number of points to use as queries.
+  size_t numAnchors = (size_t) std::round(0.1 * numSamples);
+  arma::mat queryMat = sampleSet.cols(0, numAnchors - 1);
+  // Evenly spaced sample sizes.
+  arma::Col<size_t> referenceSizes = arma::conv_to< arma::Col<size_t> >::from(
+    arma::linspace(numAnchors, numSamples - numAnchors - 1,
```
rcurtin: Can't you call linspace with `arma::Col<size_t>` as a template parameter so you can avoid the `conv_to`?  I think that is possible.

src/mlpack/methods/lsh_model/lshmodel_impl.hpp
```diff
@@ -0,0 +1,505 @@
...
+
+
+// Predict parameters for LSH that will have acceptable recall.
+template <typename SortPolicy, typename ObjectiveFunction>
+void LSHModel<SortPolicy, ObjectiveFunction>::Predict(const size_t datasetSize,
+                                                      const size_t k,
+                                                      const double minRecall)
+{
+  // Sanity check. Recall can't be greater/equal to 1, or negative.
+  if (minRecall < 0 || minRecall >=1)
```
rcurtin: Missing space after the `>=` :)

src/mlpack/methods/lsh_model/lshmodel.hpp
```diff
@@ -0,0 +1,373 @@
...
+  bool PerturbationExpand(std::vector<bool>& A) const;
+
+  /**
+   * Return true if perturbation set A is valid. A perturbation set is invalid if
+   * it contains two (or more) actions for the same dimension or dimensions that
+   * are larger than the queryCode's dimensions.
+   * @param A perturbation set to validate.
+   * @param numProj The number of projections for the sequence under validation.
+  */
+  bool PerturbationValid(const std::vector<bool>& A, size_t numProj) const;
```
rcurtin: I believe these Perturbation* functions are exactly the same as in `LSHSearch`, so maybe we should mark those methods static and use them here, or create some auxiliary class for multi-probe LSH searches and use that.

src/mlpack/methods/lsh_model/lshmodel.hpp
```diff
@@ -0,0 +1,373 @@
...
+   void GenerateTemplateSequence(size_t numProj, 
+                                 double hashWidth, 
+                                 size_t numProbes);
+
+   /** Matrix that stores, in each column, the "direction" of the perturbation:
+    * 0 means no perturbation on that dimension, -1 means reduce dimension value
+    * by 1, and +1 means increase dimension value by 1.
+    */
+   
+   arma::Mat<short int> templateSequence;
```
rcurtin: Consider `std::vector<std::vector<bool>>` here, it might be more space-efficient and possibly quicker.  Also a possibility is `boost::dynamic_bitset` instead of `std::vector<bool>`, but I am not sure that gets you anything over `std::vector<bool>` here.

Yet another thought is that generally these sequences will be pretty sparse, so it might be more reasonable to encode each perturbation sequence as a 2-dimensional matrix: one dimension represents the nonzero coordinate and the other dimension represents whether the value is positive or negative.  For that matter you could be tricky and encode both the coordinate and the direction as a single number (i.e. direction*coordinate), but then you could only have number of coordinates equal to half the possible number of coordinates of that data type (but I don't think that is a problem).

src/mlpack/methods/lsh_model/lshmodel_impl.hpp
```diff
@@ -0,0 +1,505 @@
...
+
+  if (hashWidthOut == 0.0)
+    hashWidthOut = this->hashWidth;
+
+  LSHSearch<> lsh(*referenceSet, numProjOut, numTablesOut, hashWidthOut,
+      secondHashSize, bucketSize);
+
+  lshObjectVector.push_back(lsh);
+
+  return lshObjectVector[lshObjectVector.size() - 1];
```
rcurtin: Why are you preserving the LSH objects you create?  It seems to me that they aren't used anywhere else.

mentekid:
@rcurtin: This is a first "working" version, in the sense that the algorithm is now complete.

I didn't implement the optimization over the parameter space, I want to be able to test it module-by-module. This current version has, instead, a Predict() function that accepts specific LSH parameters and predicts a recall and selectivity value.

I also didn't use boost's integration functions yet, though I have my code set up for them to be plugged in (see the IntegralObjective class inside LSHModel, for example). Instead I use a simple rectangle method to calculate the integrals.

From the few test runs I did, it looks like it's hyper-confident, always giving good recall and low selectivity. I'll test it more extensively, there's not guarantee I didn't mess something up :)

Let me know if you have suggestions / comments, or if you find any bugs!

rcurtin:
Ok, looks good.  If you can add tests, I think this is close to ready to merge.  When I have a chance today or tomorrow I will build it and run some simulations with the code.

src/mlpack/methods/lsh/lshmodel_impl.hpp
```diff
@@ -0,0 +1,607 @@
...
+  // sample.
+  arma::vec sampleHelper(referenceSet.n_cols, arma::fill::randu);
+
+  // Keep a sample of the dataset: We have uniformly random numbers in [0, 1],
+  // so we expect about N*sampleRate of them to be in [0, sampleRate).
+  arma::mat sampleSet = referenceSet.cols(
+        arma::find(sampleHelper < sampleRate));
+  // Shuffle to be impartial (in case dataset is sorted in some way).
+  sampleSet = arma::shuffle(sampleSet);
+  const size_t numSamples = sampleSet.n_cols; // Points in sampled set.
```
rcurtin: Are you sampling with or without replacement?  If you're sampling without replacement (I don't think that's the case based on the code here) you can use `math::ObtainDistinctSamples()` from somewhere in `core/math/`.  Otherwise it might be better to simply keep a list of indices of samples, and don't actually extract it from the original matrix.  Then later you can use that vector of indices to create a non-contiguous matrix subview, like this:

```
extern arma::uvec indices; // This has already been filled with stuff.
extern arma::mat dataset; // This is our dataset.
dataset.cols(indices); // Returns all the columns we're interested in.
```

This is a pretty low-priority comment, though, so don't worry too heavily about it, only if you want to.  I'd say testing is higher priority. :)

src/mlpack/methods/lsh/objectivefunction.hpp
```diff
@@ -0,0 +1,171 @@
...
+     * @param kData Vector of k - the kth nearest neighbor for which we
+     *    calculated the statistic.
+     * @param yData Matrix of y, one for each (x, k) value.
+     */
+    DefaultObjectiveFunction(const arma::Col<size_t>& xData, 
+                             const arma::Col<size_t>& kData, 
+                             const arma::mat& yData)
+      : xData(&xData), kData(&kData), yData(&yData)  { };
+
+    //! Return the number of functions
```
rcurtin: If you are only using L-BFGS, you should be able to remove this function. It should only be needed by SGD-like optimizers.

src/mlpack/methods/lsh/objectivefunction.hpp
```diff
@@ -0,0 +1,171 @@
...
+ * beta, gamma).
+ * This is the mean squared error for the current parameters or coordinates.
+ */
+double DefaultObjectiveFunction::Evaluate(const arma::mat& coordinates) const
+{
+  // Use extra variables to make code readable.
+  double alpha = coordinates(0, 0);
+  double beta = coordinates(1, 0);
+  double gamma = coordinates(2, 0);
+  double M = (double) NumFunctions();
```
rcurtin: You can mark these const.

src/mlpack/methods/lsh/objectivefunction.hpp
```diff
@@ -0,0 +1,171 @@
...
+    gradient(1, 0) += error * betaChain;
+    gradient(2, 0) += error * gammaChain;
+  }
+
+  // Return the average of each gradient after the summation is complete.
+  gradient(0, 0) /= ((double) M);
+  gradient(1, 0) /= ((double) M);
+  gradient(2, 0) /= ((double) M);
+}
+
```
rcurtin: These calculations all look correct to me, but you said that L-BFGS is still converging to NaN sometimes? I can try to help debug if you have some initial conditions and data that lead to divergence.

src/mlpack/methods/lsh/lshmodel_impl.hpp
mentekid: I think it's without replacement: I generate uniform numbers in [0, 1] and then threshold at the sample rate, getting a vector of booleans. I keep only columns (so, points) that have "true" in the corresponding vector position:
In matlab/pseudocode it would be:
```MATLAB
sampleRate = 0.3;
referenceSet = [
1 3 5 7;
2 4 6 8;
]
sampleHelper = [0.1 0.3 0.7 0.05];
sampleHelper = sampleHelper > sampleRate;
% So here sampleHelper = [0 0 1 0]
sampleSet = referenceSet.cols(sampleHelper);
% and therefore sampleSet = [5; 6] - only column 3
```

Is there something I don't see here?

I didn't know about `ObtainDistinctSamples()`, I think that will make the code cleaner so I'll refactor it to use that instead.

rcurtin: I think I wrote my comment without thoroughly looking at the code; I see now that it's sampling without replacement.  Thanks for the explanation.  In that case `ObtainDistinctSamples()` should do what you need, I think.

src/mlpack/methods/lsh/objectivefunction.hpp
mentekid: In `objectivefunction`, I initialize the point randomly. I'm not sure if that also plays a role, since I don't seed the random generator.

Try using
```
$ bin/mlpack_lshmodel -r iris.csv -p 0.5 -v
```
and
```
bin/mlpack_lshmodel -r iris.csv -p 0.6 -v
```
The first should converge to some real values, while the second should "converge" to NaN after 10000 iterations (which I imagine is the default maximum number of iterations).
Both the cost function and the gradients include logarithms, so I suspect we pass negative values somewhere in there. I'll investigate that and let you know.

I think that's also a bug in L_BFGS - shouldn't the iterations stop if objective is NaN?

mentekid: I've found the problem. I'll add it as a line note.

src/mlpack/methods/lsh/lshmodel_impl.hpp
```diff
@@ -0,0 +1,607 @@
...
+    arma::mat kNNDistances; // What we need.
+    KNN naive(refMat, true); // true: train and use naive kNN.
+    naive.Search(queryMat, k, neighbors, kNNDistances);
+
+    // Store the squared distances (what we need).
+    kNNDistances = arma::pow(kNNDistances, 2);
+
+    // Compute Arithmetic and Geometric mean of the distances.
+    Ek.row(i) = arma::mean(kNNDistances.t());
+    Gk.row(i) = arma::exp(arma::mean(arma::log(kNNDistances.t()), 0));
```
mentekid: Here's the cause of the L_BFGS -NaN values:
I compute the logarithm of the kNN distances, always assuming that there's no points that have distance 0. In the case of duplicate points, that is not a good assumption to make.
The iris.csv datset that's included in mlpack has some duplicates:
```
5.8,2.7,5.1,1.9 # repeated twice
4.9,3.1,1.5,0.1 # repeated three times
```
running `$sort iris.csv | uniq -c | awk '{print $1}' | sort | uniq` will print `1 2 3` meaning that's all the duplicates.

I think the correct approach here is to simply disregard 0-distances completely, by resizing the kNNDistances matrix to only hold positive entries.

rcurtin: I took a look at the most current code, and I see you are doing `find(kNNDistances > 0)`, but I don't think this will adequately filter duplicates.  Tomorrow I'll try and think about a good way to filter duplicate points; probably the best time to do that is during the calculation of the kNN distances matrix.  (i.e. if we encounter a zero distance, clear the row/column of the matrix and skip to the next point)  I need to think a little bit more about it...

rcurtin: Shouldn't `Gk` be the geometric mean?  What's implemented here doesn't appear to be that.  I think you can calculate the geometric mean as

```
Gk.row(i) = arma::pow(arma::prod(kNNDistances.t()), 1.0 / kNNDistances.n_cols);
```

After making that change, L-BFGS to fit the geometric distribution converges, instead of resulting in a NaN.

src/mlpack/methods/lsh/lshmodel_impl.hpp
```diff
@@ -0,0 +1,610 @@
...
+          sampleSet.unsafe_col(i), sampleSet.unsafe_col(j));
+  Log::Info << "Computed " << d << " pointwise distances." << std::endl;
+  Timer::Stop("pairwise_distances");
+
+  // Step 3. Estimate statistics of these distances: log(mean(d)), mean(log(d)),
+  // mean(d).
+  distances = arma::pow(distances, 2);
+  this->meanDist = arma::mean(distances);
+  this->logMeanDist = std::log(meanDist);
+  this->meanLogDist = arma::mean(arma::log(distances));
```
rcurtin: Even if you replace the geometric mean calculation below, this still causes problems in `Predict()`, because any zero values passed to the Gamma distribution training will cause it to fail, which makes sense because the PDF of a Gamma distribution (with any parameters) is always 0 at x = 0.

So basically what we have come upon here is a flaw in the paper: this entire strategy of modeling the squared distances as a gamma distribution is only applicable if there are no duplicate points in the dataset.  If we wanted this LSH tuning model to be able to handle duplicate points, we would need to pick a new distribution that handled zero distances and rederive everything in the paper.  But... that's not really what I want to do (if you want to feel free)...

So instead it seems like our best option is just to find a hack.  Here is the idea I went with to make my little test program work: take the smallest nonzero distance from `distances`, and use some small factor of that to replace all the 0s:

```
// Find the smallest nonzero distance.
double smallest = DBL_MAX;
for (size_t i = 0; i < distances.n_elem; ++i)
  if (distances[i] > 0.0 && distances[i] < DBL_MAX)
    smallest = distances[i];

// Replace any zero distances, since they break the gamma distribution.
for (size_t i = 0; i < distances.n_elem; ++i)
  if (distances[i] == 0.0)
    distances[i] = 1e-5 * smallest; // 1e-5 selected arbitrarily.
```

Now the distances are all nonzero, though we have probably affected the trained distribution somewhat.  It might be useful to output a warning if there are any duplicate points in the distances, noting that this affects the correctness of the predictions.  (I'm not sure how it affects the correctness.)

src/mlpack/methods/lsh/lshmodel_impl.hpp
mentekid: Yes, I'm using an alternative way to compute it:
```
pow(prod(x_i)), 1/N) = 
exp(log(pow(prod(x_i)), 1/N))) = 
exp(1/N * [ log(prod(x_i))])  = 
exp(1/N * sum( log(x_i))) = 
exp(mean(log(x_i)))
```

Computing the geometric mean this way allows me to use arma::mean(), which can calculate the row-wise, column-wise or overall mean. In this case we have a `queryMat.n_rows` x `k` matrix, and we want the average of each column - or at least, that's my understanding of the paper.

src/mlpack/methods/lsh/lshmodel_impl.hpp
mentekid: This works. I'll have to test how it affects correctness once the other details are ironed out.

I wonder if my way of calculating the geometric mean creates the problem:

I use the exp(log( prod(...) ))  to compute the geometric mean. That is correct assuming the products (i.e. distances) are strictly positive numbers, since the logarithm is undefined for <= 0. The geometric mean, though, is defined for 0s as well, but it is simply equal to 0. 
The authors haven't specified this, and I would expect the gamma distribution with geometric mean equal to 0 to be undefined, but do you think simply setting geometric means to 0 in this case be sufficient?

src/mlpack/methods/lsh/lshmodel_impl.hpp
rcurtin: Ah, ok!  I did not think it through fully. :)

src/mlpack/methods/lsh/lshmodel_impl.hpp
rcurtin: I think for the geometric mean calculation, you should do the same hack as the one I suggested for the full distances distribution: "fake" a small nonzero value.  The gamma distribution doesn't have any probability mass at 0, so really it's not even the right distribution to use when there are zero distances.  But I'm not sure of what other alternatives might be, so I think just using this hack to make it work for now should be okay...

If you have a better idea for how to modify the distances away from 0, do it---the idea I wrote down was just the first thing I could think of that seemed to work.

mentekid:
I think we should close this PR. I will probably start working on the algorithm again, but given the time that has passed, I think it's better to start from scratch if you don't mind.

rcurtin:
Sure, that sounds good to me.  It has been on my list for months to finish this up, but unfortunately it has always been sufficiently low that by the end of the week I never have any time to handle it. :(

mentekid:
And now his watch has ended.

Issue #733, opened 2016-07-22
GammaDistribution Implementation is Incomplete


mentekid:
My GammaDistribution implementation (https://github.com/mlpack/mlpack/pull/729) is incomplete. Here's a summary of what's missing:
- `Probability(arma::vec&)`: A function that gives the probability of a given observation
- `LogProbability(arma::vec&)`: A function that gives the log probability of a given observation.
- `Random()`: Return a gamma-distribution random datapoint.
- `Train(observations, probabilities)`: A function that fits the observations with given probabilities (sort of like the job done by Train(observations), but weighted).

I implemented this class mainly to have a way to fit a Gamma to observation data, but this should be completed to be usable as an mlpack distribution.

I think the Train(observations, probabilities) could be quite straightforward - the three main statistics needed for the Train(observations) function to work is log(mean(observations)), mean(observations) and mean(log(observations)). If we replace mean(.) with weightedMean(., probabilities), it should work similarly well. I might be wrong on this though.

rcurtin:
This is nearly done, but the `Train(observations, probabilities)` function isn't done yet.  Here is some discussion on how that might be done:

https://github.com/mlpack/mlpack/pull/751#discussion_r73780816

aditya1315:
@rcurtin I can take this up. I will try to implement it and get back to you in a few days?

rcurtin:
@aditya1315: you are more than welcome to do this.  If you open a PR, be sure that it follows the style guidelines (found on the wiki under the page 'DesignGuidelines') and that it has some tests to go along with it.  Thanks!

aditya1315:
@rcurtin, I saw discussion on the #751 thread and spent some time on it. Your solution seems to work. So should I base the code your solution and go ahead with testing it?

Regarding the tests - mantekid suggests a way but it gives NaN sometimes. So how should I proceed in this?

Thanks in advance

rcurtin:
@aditya1315: yes, I agree that your approach is reasonable.  You can base the code on what I wrote, and then write some tests for it.  The testing will be the hard part though, I suspect. :)

I wonder if the NaN values for the test are the result of zero values being passed to the training of the distribution.  The Gamma distribution can't handle zero-valued observations (its PDF is 0 at x=0).

yashu-seth:
@mentekid @rcurtin What is the status of this issue?

rcurtin:
[...]

PR #751, opened 2016-08-05
GammaDistribution: Adds functionality to solve #733

mentekid:
While implementing LSHModel I realized I would need the `GammaDistribution::Probability(x)` function. I remembered I mentioned in https://github.com/mlpack/mlpack/issues/733 I said I'd do it at some point, so I completed it now.

I also added a `Probability(x, dimension)`, for evaluating a single dimension at a time. This is useful for my implementation, and might prove useful to others since the gammas are completely independent here.

There's only one thing that bugs me: I'm not sure how to test the `Train(observations, probability)` function, so I haven't written a test for that yet.

### DIFF
--- a/src/mlpack/core/dists/gamma_distribution.cpp
+++ b/src/mlpack/core/dists/gamma_distribution.cpp
@@ -206,11 +206,11 @@ arma::vec GammaDistribution::Random() const
 {
   arma::vec randVec(alpha.n_elem);
 
-  std::default_random_engine generator;
   for (size_t d = 0; d < alpha.n_elem; ++d)
   {
     std::gamma_distribution<double> dist(alpha(d), beta(d));
-    randVec(d) = dist(generator);
+    // Use the mlpack random object.
+    randVec(d) = dist(mlpack::math::randGen);
   }
 
   return randVec;
--- a/src/mlpack/core/dists/gamma_distribution.cpp
+++ b/src/mlpack/core/dists/gamma_distribution.cpp
@@ -41,7 +41,6 @@ inline bool GammaDistribution::Converged(const double aOld,
   return (std::abs(aNew - aOld) / aNew) < tol;
 }
 
-
 // Fits an alpha and beta parameter to each dimension of the data.
 void GammaDistribution::Train(const arma::mat& rdata, const double tol)
 {
@@ -156,10 +155,10 @@ void GammaDistribution::Probability(const arma::mat& observations,
     for (size_t d = 0; d < observations.n_rows; ++d)
     {
       // Compute probability using Multiplication Law.
-      probabilities(i) *= 
-        std::pow(observations(d, i), alpha(d) - 1) *
-        std::exp(-observations(d, i) / beta(d)) /
-        denominators(d);
+      double factor = std::exp(-observations(d, i) / beta(d));
+      double numerator = std::pow(observations(d, i), alpha(d) - 1);
+      
+      probabilities(i) *= factor * numerator / denominators(d);
     }
   }
 }
@@ -194,10 +193,10 @@ void GammaDistribution::LogProbability(const arma::mat& observations,
     {
       // Compute probability using Multiplication Law and Logarithm addition
       // property.
-      LogProbabilities(i) += std::log( 
-        std::pow(observations(d, i), alpha(d) - 1) 
-        * std::exp(-observations(d, i) / beta(d)) 
-        / denominators(d));
+      double factor = std::exp(-observations(d, i) / beta(d));
+      double numerator = std::pow(observations(d, i), alpha(d) - 1);
+
+      LogProbabilities(i) += std::log( numerator * factor / denominators(d));
     }
   }
 }
--- a/src/mlpack/core/dists/gamma_distribution.cpp
+++ b/src/mlpack/core/dists/gamma_distribution.cpp
@@ -23,6 +23,16 @@ GammaDistribution::GammaDistribution(const arma::mat& data,
   Train(data, tol);
 }
 
+GammaDistribution::GammaDistribution(const arma::vec& alpha, 
+                                     const arma::vec& beta)
+{
+  if (beta.n_elem != alpha.n_elem)
+    throw std::runtime_error("Alpha and beta vector dimensions mismatch.");
+
+  this->alpha = alpha;
+  this->beta = beta;
+}
+
 // Returns true if computation converged.
 inline bool GammaDistribution::Converged(const double aOld,
                                          const double aNew,
@@ -30,6 +40,7 @@ inline bool GammaDistribution::Converged(const double aOld,
 {
   return (std::abs(aNew - aOld) / aNew) < tol;
 }
+
 
 // Fits an alpha and beta parameter to each dimension of the data.
 void GammaDistribution::Train(const arma::mat& rdata, const double tol)
@@ -51,6 +62,13 @@ void GammaDistribution::Train(const arma::mat& rdata, const double tol)
   // Call the statistics-only GammaDistribution::Train() function to fit the
   // parameters. That function does all the work so we're done.
   Train(logMeanxVec, meanLogxVec, meanxVec, tol);
+}
+
+// Fits an alpha and beta parameter according to observation probabilities.
+void GammaDistribution::Train(const arma::mat& observations, 
+                              const arma::vec& probabilities,
+                              const double tol)
+{
 }
 
 // Fits an alpha and beta parameter to each dimension of the data.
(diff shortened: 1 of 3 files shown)


src/mlpack/core/dists/gamma_distribution.cpp
```diff
@@ -31,6 +41,7 @@ inline bool GammaDistribution::Converged(const double aOld,
   return (std::abs(aNew - aOld) / aNew) < tol;
 }
 
+
```
rcurtin: I think this extra line is unnecessary. :)

src/mlpack/core/dists/gamma_distribution.cpp
```diff
@@ -117,3 +135,84 @@ void GammaDistribution::Train(const arma::vec& logMeanxVec,
...
+  }
+}
+
+// Returns the probability of one observation (x) for one of the Gamma's
+// dimensions.
+double GammaDistribution::Probability(double x, size_t dim) const
+{
+  return 
+    std::pow(x, alpha(dim) - 1) * std::exp(-x / beta(dim)) / 
+    (std::tgamma(alpha(dim)) * std::pow(beta(dim), alpha(dim)));
```
rcurtin: I know I'm being picky but I think the indentation needs to be one more here.  Another option is to split this into multiple lines for readability, but that's up to you.  Realistically, to the compiler, it shouldn't make a difference if you compute a bunch of doubles then multiply them together, or compute the result all at once.

src/mlpack/core/dists/gamma_distribution.cpp
```diff
@@ -117,3 +135,84 @@ void GammaDistribution::Train(const arma::vec& logMeanxVec,
...
+  for (size_t i = 0; i < numObs; ++i)
+  {
+    for (size_t d = 0; d < observations.n_rows; ++d)
+    {
+      // Compute probability using Multiplication Law and Logarithm addition
+      // property.
+      LogProbabilities(i) += std::log( 
+        std::pow(observations(d, i), alpha(d) - 1) 
+        * std::exp(-observations(d, i) / beta(d)) 
+        / denominators(d));
```
rcurtin: Same comment about indentation and readability applies here.

src/mlpack/core/dists/gamma_distribution.cpp
```diff
@@ -117,3 +135,84 @@ void GammaDistribution::Train(const arma::vec& logMeanxVec,
...
+    }
+  }
+}
+
+// Returns a gamma-random d-dimensional vector.
+arma::vec GammaDistribution::Random() const
+{
+  arma::vec randVec(alpha.n_elem);
+
+  std::default_random_engine generator;
```
rcurtin: I think you should use `mlpack::math::randGen` here instead, so that if the user sets the random seed, the results when they call `GammaDistribution::Random()` are deterministic.  (`randGen` is defined in `mlpack/core/math/random.hpp`.)

src/mlpack/core/dists/gamma_distribution.hpp
```diff
@@ -78,6 +86,19 @@ class GammaDistribution
...
+     * @param observations The reference data, one observation per column
+     * @param probabilities The probability of each observation. One value per
+     *     column of the observations matrix.
+     * @param tol Convergence tolerance. This is *not* an absolute measure:
+     *    It will stop the approximation once the *change* in the value is 
+     *    smaller than tol.
+     */
+    void Train(const arma::mat& observations, 
+                                  const arma::vec& probabilities,
+                                  const double tol = 1e-8);
```
rcurtin: Indentation seems odd here.

src/mlpack/core/dists/gamma_distribution.hpp
```diff
@@ -94,6 +115,58 @@ class GammaDistribution
...
+     *
+     * for one dimension. This implementation assumes each dimension is
+     * independent, so the product rule is used.
+     *
+     * @param observations Matrix of observations, one per column.
+     * @param logProbabilities column vector of log probabilities, one per 
+     *     observation.
+     */
+    void LogProbability(const arma::mat& observations, 
+                     arma::vec& LogProbabilities) const;
```
rcurtin: Odd indentation here too.  I promise I will comment on more than just style :)

src/mlpack/tests/distribution_test.cpp
```diff
@@ -539,4 +539,74 @@ BOOST_AUTO_TEST_CASE(GammaDistributionTrainStatisticsTest)
...
+  arma::vec prob3;
+
+  // Expect that the 2-dimensional distribution returns the product of the
+  // 1-dimensional distributions (evaluated at wolfram|alpha).
+  GammaDistribution d3(a3, b3);
+  d3.LogProbability(x3, prob3);
+  BOOST_REQUIRE_CLOSE(prob3(0), std::log(0.04408), 1e-3);
+  BOOST_REQUIRE_CLOSE(prob3(1), std::log(0.026165), 1e-3);
+}
+
```
rcurtin: Can you add a test for the `Random()` function?  Here's some idea:

```
BOOST_AUTO_TEST_CASE(RandomTest)
{
  arma::vec alphas("1.0 2.0 3.0");
  arma::vec beta("1.5 2.5 3.5");

  GammaDistribution d(alphas, betas);

  arma::mat observations(1000, 3);
  for (size_t i = 0; i < 1000; ++i)
    observations.col(i) = d.Random();

  // Train a second distribution, then ensure it is similar to the first.
  GammaDistribution d2(observations);

  // Check the values, with a fairly large tolerance (3%).
  BOOST_REQUIRE_CLOSE(d.Alpha(0), d2.Alpha(0), 3.0);
  BOOST_REQUIRE_CLOSE(d.Alpha(1), d2.Alpha(1), 3.0);
  BOOST_REQUIRE_CLOSE(d.Alpha(2), d2.Alpha(2), 3.0);
  BOOST_REQUIRE_CLOSE(d.Beta(0), d2.Beta(0), 3.0);
  BOOST_REQUIRE_CLOSE(d.Beta(1), d2.Beta(1), 3.0);
  BOOST_REQUIRE_CLOSE(d.Beta(2), d2.Beta(2), 3.0);
  // Should these all be Alphas()[i] and Betas()[i]?
}
```

To test the tolerance, add a `math::RandomSeed(std::time(NULL))` at the top, recompile, and run this: `while(true); do bin/mlpack_test -t GammaDistributionTest/RandomTest; sleep 1; done` and make sure that over a decent period of time, you don't see any failures.  If so, 5% or maybe even 10% are reasonable tolerances.  We just want to check that it's giving us stuff that's "reasonably close" to a Gamma distribution; checking exact precision is quite a large amount of work and here I am not sure that effort is warranted.

src/mlpack/core/dists/gamma_distribution.cpp
```diff
@@ -53,6 +64,13 @@ void GammaDistribution::Train(const arma::mat& rdata, const double tol)
   Train(logMeanxVec, meanLogxVec, meanxVec, tol);
 }
 
+// Fits an alpha and beta parameter according to observation probabilities.
+void GammaDistribution::Train(const arma::mat& observations, 
+                              const arma::vec& probabilities,
+                              const double tol)
+{
```
rcurtin: Here's an idea for this.  The idea is that we are training the distribution, but we do not know for certain that the points we are training on came from this distribution---we only have an estimate between 0 and 1 of how likely we think it is that the point came from this distribution.  So we can train using the probabilities as weights.

We can use each of the probabilities to construct a "weighted" version of `meanLogxVec`, `meanxVec`, and `logMeanxVec`.  This means we can't use Armadillo's nice syntax though since we have to implement this ourselves.

```
arma::vec meanLogxVec(rdata.n_rows, arma::fill::zeros);
arma::vec meanxVec(rdata.n_rows, arma::fill::zeros);
arma::vec logMeanxVec(rdata.n_rows, arma::fill::zeros);

for (size_t i = 0; i < rdata.n_cols; ++i)
{
  meanLogxVec += probabilities(i) * std::log(rdata.col(i));
  meanxVec += probabilities(i) * rdata.col(i);
}
meanLogxVec /= arma::accu(probabilities);
meanxVec /= arma::accu(probabilities);
logMeanxVec = arma::log(meanxVec);

Train(logMeanxVec, meanLogxVec, meanxVec, tol);
```

What do you think?

Testing that could be done by generating points from a Gamma distribution, assigning them a random weight between, e.g., 0.9 and 1, and then generating points from a uniform distribution, and assigning those points a random weight between, e.g, 0.0 and 0.02.  Then, call Train() with probabilities, and ensure the resulting trained Gamma distribution is close to the distribution that was used to generate points.

What do you think?  I've only outlined ideas here, but I can implement them if you need.

mentekid: I implemented most of the changes. There's a bug in my code for the Train(observations, probabilities) function. I'll push that and the relevant test once it's fixed.

> Testing that could be done by generating points from a Gamma distribution, assigning them a random weight between, e.g., 0.9 and 1, and then generating points from a uniform distribution, and assigning those points a random weight between, e.g, 0.0 and 0.02. Then, call Train() with probabilities, and ensure the resulting trained Gamma distribution is close to the distribution that was used to generate points.

My idea was to get a few points on the x axis and  use `GammaDistribution::Probability(x)` to estimate the pdf at those points, then use the points as observations and their probabilities as the probabilities vector.
The problem is, that generates nan values during the approximation (the bug I'm speaking of). I'm not sure if my idea is bad or my code doesn't work. I'll figure it out :)

rcurtin: Hm, maybe that will work; I haven't thought about it too much.  I think the only issue might be that if you take unlikely points to have low probability, then when you train, this biases the training points towards the high-PDF parts of the distribution, possibly giving (I think) a trained distribution with different properties than the original.  If it was a one-dimensional Gaussian we were training, this would result in a lower variance, but I don't have the intuition to say what it will do the Gamma distribution, only that I think you'll end up with a different Gamma distribution than the one you are taking random samples from.  Hopefully what I've written here is at least somewhat coherent. :)

rcurtin:
I removed the unfinished Train() overload in 76eb541.  We can address that later.

PR #752, opened 2016-08-05
NEAT algorithm

BangLiu:
Implemented NEAT algorithm, and multiple test cases.


src/mlpack/methods/ne/neat.hpp
```diff
@@ -0,0 +1,976 @@
...
+#include "genome.hpp"
+#include "species.hpp"
+#include "population.hpp"
+#include "tasks.hpp"
+#include "parameters.hpp"
+
+namespace mlpack {
+namespace ne {
+
+struct LinkInnovation {
```
zoq: Can you add a comment that describes why this struct is useful.

src/mlpack/methods/ne/neat.hpp
```diff
@@ -0,0 +1,976 @@
...
+  int fromNeuronId;
+  int toNeuronId;
+  int newLinkInnovId;
+};
+
+struct NeuronInnovation {
```
zoq: Can you add a comment that describes why this struct is useful.

src/mlpack/methods/ne/neat.hpp
BangLiu: Added.

src/mlpack/methods/ne/neat.hpp
BangLiu: Added.

zoq:
I've written a command line client and server to run the NEAT method, actually any mlpack classifier, against the OpenAI Gym toolkit. Which enables us to easily test the NEAT method against a variety of environments.

I've tried to reproduce the result for the Mario task, but it looks like the method doesn't converge as it did the last time I tested it. I'm not sure if the last change that effected the Evaluation process could be a reason for that. Maybe @BangLiu  can provide some insights?

BangLiu:
Hi Marcus,

I am so sorry for reply so late! I am in China and didn't check this email
quite often, and also cannot receive the notifications from it due to the
Internet condition. Sorry for that! I check the bang3@ualberta.ca emails
more frequently.

The last change I made to the NEAT algorithm is mainly the changes of
activation calculation in the Genome class. There are some bugs previously.
I test all the testing cases except Mario. I will try to test it when I
have got a free time, as these days I am quite busy with a lot stuffs.

Best wishes,
Bang

2016-11-08 23:55 GMT+08:00 Marcus Edel <notifications@github.com>:

sourabhvarshney111:
@zoq I think this can also be closed because there is another PR for the same by @kartik-nighania.

zoq:
I'd like to keep this open for some references, once NEAT is implemented we can close the PR.

sourabhvarshney111:
Ok got it. Once the issue in the PR #1105 is solved, you would close this PR. Right @zoq.

PR #753, opened 2016-08-05
CNE algorithm

BangLiu:
Implemented CNE algorithm, and XOR test case.


src/mlpack/methods/ne/genome.hpp
```diff
@@ -0,0 +1,334 @@
...
+class Genome {
+ public:
+  // Neurons.
+  std::vector<NeuronGene> aNeuronGenes;
+
+  // Links.
+  std::vector<LinkGene> aLinkGenes;
+
+  // Default constructor.
+  Genome() {
```
zoq: Let's use the constructor initialization lists here.

src/mlpack/methods/ne/genome.hpp
```diff
@@ -0,0 +1,334 @@
...
+    aId = -1;
+    aNumInput = 0;
+    aNumOutput = 0;
+    aFitness = DBL_MAX;
+  }
+  
+  // Parametric constructor.
+  Genome(int id,
```
zoq: Some style issue here, we should:

- To be consistent with the rest of the codebase use the same naming conventions for the parameter and function names e.g. numInput instead of aNumInput. For more informations take a look at: https://github.com/mlpack/mlpack/wiki/DesignGuidelines#naming-conventions. Also, we tend to be very picky about coding standards and conventions, so if you find inconsistencies in existing code or things that are unclear and should be clarified, please feel free to point them out.

- We use doxygen to generate code documentation, so it would be great if you could adapt the style accordingly e.g.

```
/*
 * Construct the Genome object with the given parameters.
 *
 * @param id Genome identifier.
 * @param neuronGenes List of genes to construct the genome.
 * @param linkGenes List of links to construct the genome.
 * @param numInput Number of input genes.
 * @param numOutput Number of output genes.
 * @param fitness Initial fitness.
 */
Genome(int id,
       const std::vector<NeuronGene>& neuronGenes,
       const std::vector<LinkGene>& linkGenes,
       int numInput,
       int numOutput,
       double fitness):
```

src/mlpack/methods/ne/genome.hpp
```diff
@@ -0,0 +1,334 @@
...
+  	     const std::vector<NeuronGene>& neuronGenes,
+         const std::vector<LinkGene>& linkGenes,
+         int numInput,
```
zoq: It might be useful to use ``const`` for ``numInput, numOutput, fitness``.

src/mlpack/methods/ne/genome.hpp
```diff
@@ -0,0 +1,334 @@
...
+    aId = genome.aId;
+    aNeuronGenes = genome.aNeuronGenes;
+    aLinkGenes = genome.aLinkGenes;
+    aNumInput = genome.aNumInput;
+    aNumOutput = genome.aNumOutput;
+    aFitness = genome.aFitness;
+  }
+
+  // Destructor.
+  ~Genome() {}
```
zoq: I think there is no need to implement a default destructor.

src/mlpack/methods/ne/genome.hpp
```diff
@@ -0,0 +1,334 @@
...
+    }
+
+    return *this;
+  }
+
+  // Get genome id.
+  int Id() const { return aId; }
+
+  // Set genome id.
+  void Id(int id) { aId = id; }
```
zoq: Generally we use a setter function that takes the input as an argument if it's necessary to validate what the user is setting the value to. Otherwise, we use a function that allows a direct access e.g.:

```
//! Get the genome id.
int Id() const { return id; }
//! Modify the genome id.
int& Id() { return id; }
```

src/mlpack/methods/ne/genome.hpp
```diff
@@ -0,0 +1,334 @@
...
+    return aNeuronGenes.size();
+  }
+  
+  // Get link number.
+  int NumLink() const {
+    return aLinkGenes.size();
+  }
+
+  // Get input length.
+  int GetNumInput() {
```
zoq: Do you think this function might be better named ``NumInput()``?.

src/mlpack/methods/ne/genome.hpp
```diff
@@ -0,0 +1,334 @@
...
+  void NumOutput(int numOutput) { aNumOutput = numOutput; }
+
+  // Set fitness.
+  void Fitness(double fitness) { aFitness = fitness; }
+
+  // Get fitness.
+  double Fitness() const { return aFitness; }
+
+  // Get neuron number.
+  int NumNeuron() const {
```
zoq: This is a very picky minor style comment: can you put the closing brace on a new line, take a look at: https://github.com/mlpack/mlpack/wiki/DesignGuidelines#brace-placement for more informations.

src/mlpack/methods/ne/genome.hpp
```diff
@@ -0,0 +1,334 @@
...
+          aNeuronGenes[toNeuronIdx].CalcActivation();
+        } else if (GetNeuronIndex(aLinkGenes[i + 1].ToNeuronId()) != toNeuronIdx) {
+          aNeuronGenes[toNeuronIdx].CalcActivation();
+        }
+      }
+    }
+  }
+
+  // Get output vector.
+  void Output(std::vector<double>& output) {
```
zoq: I suggest to use arma::vec for the output, in this case, we can use a bunch of mlpack methods that can directly work with the output, what do you think? The same applies to the input. E.g. we could directly use PCA to preprocess the input. Also, we could allocate the memory once, and overwrite the results in every iteration instead of clearing the vector.

src/mlpack/methods/ne/genome.hpp
```diff
@@ -0,0 +1,334 @@
...
+
+  // Add neuron.
+  void AddHiddenNeuron(NeuronGene& neuronGene) {
+    if (neuronGene.Type() == HIDDEN) {
+      aNeuronGenes.push_back(neuronGene);
+    }
+  }
+
+  // Show genome structure.
+  void Show() {
```
zoq: I think we can remove this unused function for now.

src/mlpack/methods/ne/link_gene.hpp
```diff
@@ -0,0 +1,111 @@
...
+namespace mlpack {
+namespace ne {
+
+/**
+ * This class defines a link gene.
+ */
+class LinkGene {
+ public:
+  // Default constructor.
+  LinkGene() {}
```
zoq: Do we have to specify the default constructor?

src/mlpack/methods/ne/genome.hpp
```diff
@@ -0,0 +1,334 @@
+/**
+ * @file genome.hpp
+ * @author Bang Liu
+ *
+ * Definition of the Genome class.
+ */
+#ifndef MLPACK_METHODS_NE_GENOME_HPP
+#define MLPACK_METHODS_NE_GENOME_HPP
+
+#include <cstddef>
```
zoq: I think we don't need this header file anymore.

src/mlpack/methods/ne/cne.hpp
```diff
@@ -0,0 +1,186 @@
...
+    aSeedGenome = seedGenome;
+    aSpeciesSize = params.aSpeciesSize;
+    aMaxGeneration = params.aMaxGeneration;
+    aMutateRate = params.aMutateRate;
+    aMutateSize = params.aMutateSize;
+    aElitePercentage = params.aElitePercentage;
+  }
+
+  // Destructor.
+  ~CNE() {}
```
zoq: I think there is no need to implement a default destructor.

src/mlpack/methods/ne/cne.hpp
```diff
@@ -0,0 +1,186 @@
...
+
+/**
+ * This class implements Conventional Neuro-evolution (CNE): weight
+ * evolution on topologically fixed neural networks.
+ */
+template<typename TaskType>
+class CNE {
+ public:
+  // Parametric constructor.
+  CNE(TaskType task, Genome& seedGenome, Parameters& params) {
```
zoq: It would be great if you could provide a Constructor that allows a user to directly set the parameters, instead of passing a parameter object.

zoq: It might be useful to use ``const`` for all parameters.

src/mlpack/methods/ne/cne.hpp
```diff
@@ -0,0 +1,186 @@
...
+    aMutateSize = params.aMutateSize;
+    aElitePercentage = params.aElitePercentage;
+  }
+
+  // Destructor.
+  ~CNE() {}
+
+  // Soft mutation: add a random value chosen from
+  // initialization prob distribution with probability p.
+  // TODO: here we use uniform distribution. Can we use exponential distribution?
```
zoq: I guess for now it's reasonable to use a uniform distribution only.

src/mlpack/methods/ne/cne.hpp
```diff
@@ -0,0 +1,186 @@
...
+        double oldW = genome.aLinkGenes[i].Weight();
+        genome.aLinkGenes[i].Weight(oldW + deltaW);
+      }
+    }
+  }
+
+  // Hard mutation: replace with a random value chosen from
+  // initialization prob distribution with probability p.
+  // TODO: here we use uniform distribution. Can we use exponential distribution?
+  static void MutateWeightsUnbiased(Genome& genome, double mutateProb, double mutateSize) {
```
zoq: Do you think, we could use a boolean to distinguish between soft and hard mutation. In this case we could reuse the basic function.

src/mlpack/methods/ne/cne.hpp
```diff
@@ -0,0 +1,186 @@
...
+    while (generation < aMaxGeneration) {
+    	// Evaluate all genomes in the species.
+      for (int i=0; i<aSpecies.SpeciesSize(); ++i) {
+        double fitness = aTask.EvalFitness(aSpecies.aGenomes[i]);
+        aSpecies.aGenomes[i].Fitness(fitness);
+      }
+      aSpecies.SetBestFitnessAndGenome();
+
+    	// Output some information.
+      printf("Generation: %zu\tBest fitness: %f\n", generation, aSpecies.BestFitness());
```
zoq: I think it would be better to use Log::Info here, in that case a user can disable any output e.g:

```
Log::Info << "Generation: " << generation << " best fitness: " <<  species.BestFitness() << std::endl
```

or something like that, the important thing is a user can disable the output.

src/mlpack/methods/ne/cne.hpp
```diff
@@ -0,0 +1,186 @@
...
+      if (aTask.Success()) {
+        printf("Task succeed in %zu iterations.\n", generation);
+        exit(0);
```
zoq: That would terminate the complete process, I would go with ``return true``.

src/mlpack/methods/ne/species.hpp
```diff
@@ -0,0 +1,141 @@
...
+  // Get species size.
+  int SpeciesSize() const { return aGenomes.size(); }
+
+  // Set best fitness to be the minimum of all genomes' fitness.
+  void SetBestFitnessAndGenome() {
+    if (aGenomes.size() == 0) 
+      return;
+
+    aBestFitness = aGenomes[0].Fitness();
+    for (int i=0; i<aGenomes.size(); ++i) {
```
zoq: We can start with ``i = 1``.

src/mlpack/methods/ne/species.hpp
```diff
@@ -0,0 +1,141 @@
...
+      if (aGenomes[i].Fitness() < aBestFitness) {
+        aBestFitness = aGenomes[i].Fitness();
+        aBestGenome = aGenomes[i];
+      }
+    }
+  }
+
+  // Sort genomes by fitness. First is best.
+  static bool CompareGenome(Genome lg, Genome rg) {
```
zoq: We should use a reference here, to avoid a copy, also it might be useful to use const here.

src/mlpack/methods/ne/utils.hpp
```diff
@@ -0,0 +1,88 @@
...
+  boost::random::uniform_real_distribution<> dist(x, y);
+  return dist(rng);
+}
+
+/**
+ * Definitions of other functions.
+ */
+
+// Return the sign of a number.
+template <typename T> int sgn(T val) {
```
zoq: What about using ``arma::sign(val)`` take a look at: http://arma.sourceforge.net/docs.html#misc_fns for more informations.

src/mlpack/methods/ne/utils.hpp
```diff
@@ -0,0 +1,88 @@
...
+
+// Returns a random integer between [x, y]
+// in case of ( 0 .. 1 ) returns 0
+int RandInt(int x, int y) {
+	boost::random::uniform_int_distribution<> dist(x, y);
+    return dist(rng);
+}
+
+// Return a random float between [0, 1]
+double RandFloat() {
```
zoq: It looks like this function is unused, so maybe we can remove it?

src/mlpack/tests/ne_test.cpp
```diff
@@ -0,0 +1,691 @@
+/**
+ * @file ne_test.cpp
```
zoq: Can you split this test into smaller parts. Like ``link_gene_test.cpp, neuron_gene_test.cpp, genome_test.cpp, cne_test.cpp, neat_test.cpp``.

src/mlpack/methods/ne/genome.hpp
BangLiu: Revised.

src/mlpack/methods/ne/genome.hpp
BangLiu: Yeah, I will revise accordingly.

src/mlpack/methods/ne/genome.hpp
BangLiu: I think it shall make no difference. Besides, it will make a lot of consts.
http://stackoverflow.com/questions/117293/use-of-const-for-function-parameters

src/mlpack/methods/ne/genome.hpp
BangLiu: Yeah, just a habit for potential usage.

src/mlpack/methods/ne/genome.hpp
BangLiu: We already have NumInput() which is used to return the value of aNumInput.
This GetNumInput() is get the number of numInput by loop, rather than look at the aNumInput. I implement this function as I found that previously the real numInput sometimes are not the same with aNumInput.

src/mlpack/methods/ne/genome.hpp
BangLiu: Personally, I prefer the current brace style, which I think makes code shorter in height. But if this is required by the mlpack project, I will revise them before merge.

src/mlpack/methods/ne/genome.hpp
BangLiu: Yeah, no problem

BangLiu: I am wondering that how to use the serialization property of mlpack to save an object, such as genome?

src/mlpack/methods/ne/link_gene.hpp
BangLiu: Yeah, maybe useless. I can remove it.

src/mlpack/methods/ne/genome.hpp
BangLiu: Yeah

src/mlpack/methods/ne/cne.hpp
BangLiu: I think for algorithms the parameters are a lot. Users can set parameter by construct an Parameters object without put them into a constructor function in a specific sequence. And this also makes the algorithm constructor be concise. If we have more parameters in the future, we just add new member for Parameters class, no need to change the function headers of constructors. Besides, different NE algorithm will have similar functions for constructor.

src/mlpack/methods/ne/cne.hpp
BangLiu: Yeah, I agree.

src/mlpack/methods/ne/cne.hpp
BangLiu: Yeah, we can do this.

src/mlpack/methods/ne/genome.hpp
rcurtin: Yeah, we use new lines for opening braces.  In this case I think you could put it all in one line:

```
int NumNeuron() const { return aNeuronGenes.size(); }
```

(That should be fewer than 80 characters.)

BangLiu: Yeah, no problem. I will revise accordingly.

src/mlpack/methods/ne/cne.hpp
BangLiu: Good idea, I will revise.

src/mlpack/methods/ne/cne.hpp
BangLiu: Revised for both algorithm: cne and neat.

src/mlpack/methods/ne/species.hpp
BangLiu: Revised.

src/mlpack/tests/ne_test.cpp
BangLiu: I think it would be better if we keep the tests of NE module in the same file.

src/mlpack/methods/ne/utils.hpp
BangLiu: Yeah, no problem.

src/mlpack/methods/ne/species.hpp
BangLiu: If the species only contains one genome, start from i=1 will cause error. So I think we keep start from 0.

src/mlpack/methods/ne/genome.hpp
zoq: Can you provide an example where the real numInput doesn't match with aNumInput?

src/mlpack/methods/ne/genome.hpp
zoq: Every class you like to serialize should implement the Serialize, in case of the Genome I guess the important parameter you like to serialize is the list of genes and links and some other parameter like fitness:
```
template<typename Archive>
void Serialize(Archive& ar, const unsigned int /* version */)
{
  ar & data::CreateNVP(neuronGenes, "neuronGenes");
  ar & data::CreateNVP(linkGenes, "linkGenes");
  ar & data::CreateNVP(fitness, "fitness");
}
```

You can save a model (Genome) using:

```
Genome genome;
data::Save("output_model_file.xml", "genome_model", genome);
```

Loading is also simple:

```
Genome genome;
data::Load("input_model_file.xml"), "genome_model", genome);
```

Also you might find: https://github.com/mlpack/mlpack/wiki/DesignGuidelines#requirements-of-each-class helpful

src/mlpack/methods/ne/cne.hpp
zoq: I agree, that for some methods you have to specify a bunch of parameters, so I think the parameter class comes in handy. On the other side, I like to provide a constructor that works right out of the box and starts with some reasonable parameters. I think in the case of the CNE method, we can specify such parameters. Maybe we can provide an constructor for such a situation, what do you think?

src/mlpack/methods/ne/species.hpp
zoq: In that case aGenomes.size() is 1 and we don't enter the for loop, right? Also, what do you think about the unset aBestGenome problem if genome with index 0 has the best fitness?

src/mlpack/tests/ne_test.cpp
zoq: You mean all NE tests? I'm not sure, if you implement another NE algorithm or like to test an NE methods on another task, the test file gets longer and longer. If you split the test file, and someone is only interested in the NEAT method, he could just take a look at the NEAT test file. let me know what you think. If you like the idea of a single test file, we can go with that solution.

src/mlpack/methods/ne/cne.hpp
```diff
@@ -0,0 +1,257 @@
...
+    	// Evaluate all genomes in the species.
+      for (int i = 0; i < species.SpeciesSize(); ++i)
+      {
+        double fitness = task.EvalFitness(species.genomes[i]);
+        species.genomes[i].Fitness(fitness);
+      }
+      species.SetBestFitnessAndGenome();
+
+    	// Output some information.
+      printf("Generation: %zu\tBest fitness: %f\n", generation, species.BestFitness());
```
zoq: Can you use ``Log::Info << "message" << std::endl;`` instead of ``printf``? That way a user can disable the output if needed or adjust the output according to his own preference e.g. only debug outputs (``Log::Debug``).

src/mlpack/tests/ne_test.cpp
```diff
@@ -0,0 +1,238 @@
...
+
+  // Construct seed genome for xor task.
+  int id = 0;
+  int numInput = 3;
+  int numOutput = 1;
+  double fitness = -1;
+  std::vector<NeuronGene> neuronGenes;
+  std::vector<LinkGene> linkGenes;
+
+  NeuronGene inputGene1(0, INPUT, LINEAR, 0, std::vector<double>(), 0, 0);
```
zoq: Do you think, we could set a default argument for the  ``coordinate``parameter. That way we don't have to specify ``std::vector<double>()`` every time we create a new NeuronGene.

src/mlpack/tests/ne_test.cpp
```diff
@@ -0,0 +1,238 @@
...
+
+  Genome seedGenome = Genome(0, 
+                             neuronGenes,
+                             linkGenes,
+                             numInput,
+                             numOutput,
+                             fitness);
+
+  // Test seed genome.
+  std::vector<std::vector<double>> inputs;  // TODO: use arma::mat for input.
```
zoq: It would be great if we could to that change before we merge the code in. If you like I can do that for you.

src/mlpack/methods/ne/utils.hpp
```diff
@@ -0,0 +1,105 @@
...
+double RandFloat(double x, double y)
+{
+  boost::random::uniform_real_distribution<> dist(x, y);
+  return dist(rng);
+}
+
+/**
+ * Return the sign of a number.
+ */
+template <typename T> int sgn(T val)
```
zoq: If we use armadillo's ``sgn`` function, we could remove the utils file, or do we need another function from the header file?

sourabhvarshney111:
Can @zoq @rcurtin verify what is the issue with this?

zoq:
We can close this one, CNE is implemented see: https://github.com/mlpack/mlpack/tree/master/src/mlpack/core/optimizers/cne

sourabhvarshney111:
Thanks @zoq :)

PR #757, opened 2016-08-07
Inception layer

nilayjain:


src/mlpack/methods/ann/layer/concat_layer.hpp
```diff
@@ -0,0 +1,183 @@
...
+  ForwardTail(std::tuple<Tp...>& layers, DataType& output)
+  {
+    /* Nothing to do. */
+  }
+
+  template<size_t I = 0, typename DataType, typename... Tp>
+  typename std::enable_if<I < sizeof...(Tp), void>::type
+  ForwardTail(std::tuple<Tp...>& layers, DataType& output)
+  {
+    output = arma::join_slices(output, std::get<I>(layers).OutputParameter());
```
zoq: The problem I see here is, that it only works if we use cubes as output type. We could use template specializations to distinguish between arma::mat and arma::cube what do you think?

src/mlpack/methods/ann/layer/concat_layer.hpp
```diff
@@ -0,0 +1,183 @@
...
+#include <mlpack/core.hpp>
+#include <mlpack/methods/ann/activation_functions/rectifier_function.hpp>
+#include <mlpack/methods/ann/layer/layer_traits.hpp>
+#include <mlpack/methods/ann/layer/one_hot_layer.hpp>
+#include <mlpack/methods/ann/layer/conv_layer.hpp>
+#include <mlpack/methods/ann/layer/pooling_layer.hpp>
+#include <mlpack/methods/ann/layer/softmax_layer.hpp>
+#include <mlpack/methods/ann/layer/bias_layer.hpp>
+#include <mlpack/methods/ann/layer/linear_layer.hpp>
+#include <mlpack/methods/ann/layer/base_layer.hpp>
```
zoq: It looks like, we can remove all header files except:

```
#include <mlpack/core.hpp>
#include <mlpack/methods/ann/layer/layer_traits.hpp>
```

src/mlpack/methods/ann/layer/inception_layer.hpp
```diff
@@ -0,0 +1,290 @@
...
+    basePool.InputParameter() = biasPool.OutputParameter();
+    basePool.Forward(convPool.OutputParameter(), basePool.OutputParameter());
+
+    // concatenate outputs of all the paths.
+    output = arma::join_slices( 
+              arma::join_slices(
+                arma::join_slices( 
+                  base1.OutputParameter(), base3.OutputParameter() ), 
+                  base5.OutputParameter() ), basePool.OutputParameter());
+
```
zoq: Since the network class initializes the weights, you have to define the size e.g. using:

```
weights.set_size(rows, cols);
```

Since the inception layer combines different layer, we have to calculate the overall weight dimension accordingly. You can get the weight dimension of each layer by calling the weight function:

```
conv1.Weights().n_rows
conv1.Weights().n_cols
```

src/mlpack/methods/ann/layer/concat_layer.hpp
```diff
@@ -0,0 +1,183 @@
...
+template <typename JoinLayers,
+          typename InputDataType = arma::cube,
+          typename OutputDataType = arma::cube>
+class ConcatLayer
+{
+ public:
+  ConcatLayer(const size_t numLayers,
+              JoinLayers &&layers) :
+  numLayers(numLayers),
+  layers(std::forward<JoinLayers>(layers))
```
stereomatchingkiss: Your solution tell the compiler "JoinLayers must be rvalue", that means

```
SomeLayer layers;
ConcatLayer<SomeLayer> ConcatLayer(layers);
```
Do not work, because layers is an lvalue in this case, but the constructor only accept rvalue.

If you want to perform perfect forwarding by std::forward(which means, if it is rvalue, cast it to rvalue, if it is lvalue, cast it lvalue), you need to write it as

```
template<typename JoinLayersType>
ConcatLayer(const size_t numLayers,
                    JoinLayersType &&layers) ;
```
In this case, the compiler will treat JoinLayers&& as "[universal reference](https://isocpp.org/blog/2012/11/universal-references-in-c11-scott-meyers)", so perfect forwarding work under this case.

You may ask, but I need the JoinLayersType == JoinLayers, use can use static_assert to add this restriction work

```
using RawJLT = typename std::decay<JoinLayersType >::type;
static_assert(std::is_same<RawJLT , JoinLayers>::value, "type of JoinLayerType must equal to JoinLayers");
```

src/mlpack/methods/ann/layer/concat_layer.hpp
```diff
@@ -0,0 +1,183 @@
...
+  ForwardTail(std::tuple<Tp...>& layers, DataType& output)
+  {
+    output = arma::join_slices(output, std::get<I>(layers).OutputParameter());
+    ForwardTail<I + 1, DataType, Tp...>(layers, output);
+  }
+
+  template<typename eT>
+  void Backward(arma::Cube<eT>&, arma::Cube<eT>& error, arma::Cube<eT>& )
+  {
+    size_t slice_idx = 0;
```
stereomatchingkiss: To follow the the naming condition of mlpack, maybe we should change slide_idx to slideIDX or slideIndex

src/mlpack/methods/ann/layer/inception_layer.hpp
```diff
@@ -0,0 +1,290 @@
...
+*/
+
+template<typename InputDataType = arma::cube,
+         typename OutputDataType = arma::cube>
+class InceptionLayer
+{
+ public:
+  //! Locally-stored number of input maps.
+  size_t inMaps;
```
stereomatchingkiss: Any reasons to put these variables as public?

src/mlpack/methods/ann/network_util_impl.hpp
```diff
@@ -275,7 +275,164 @@ LayerWeights(InitializationRuleType& /* initializeRule */,
...
+        wPad + input.n_rows - 1,
+        hPad + input.n_cols - 1) = input;
+}
+
+template<typename eT>
+void Pad(const arma::Cube<eT>& input, size_t wPad, size_t hPad, arma::Cube<eT>& output)
+{
+  output = arma::zeros(input.n_rows + wPad * 2, input.n_cols + hPad * 2, input.n_slices);
+  for (size_t i = 0; i < input.n_slices; ++i)
+    Pad<double>(input.slice(i), wPad, hPad, output.slice(i));
```
stereomatchingkiss: Why Pad<double> but not Pad<eT>?

src/mlpack/methods/ann/layer/conv_layer.hpp
```diff
@@ -251,6 +269,26 @@ class ConvLayer
     return std::floor(size + p * 2 - k) / s + 1;
   }
 
+  template<typename eT>
+  void Pad(const arma::Mat<eT>& input, size_t wPad, size_t hPad, arma::Mat<eT>& output)
```
stereomatchingkiss: I guess we can use the Pad functions of network_utility to replace them

src/mlpack/tests/layer_test.cpp
```diff
@@ -0,0 +1,277 @@
...
+
+BOOST_AUTO_TEST_SUITE(InceptionLayerTest);
+
+void Test(arma::mat m1, arma::mat m2)
+{
+  for (size_t i = 0; i < m1.n_cols; ++i)
+    BOOST_REQUIRE_CLOSE(m1(i), m2(i), 1e-2);  
+}
+
+void Test(arma::cube m1, arma::cube m2)
```
stereomatchingkiss: const& should be able to speed things up

src/mlpack/tests/layer_test.cpp
```diff
@@ -0,0 +1,277 @@
...
+using namespace mlpack;
+using namespace mlpack::ann;
+using namespace mlpack::optimization;
+
+
+BOOST_AUTO_TEST_SUITE(InceptionLayerTest);
+
+void Test(arma::mat m1, arma::mat m2)
+{
+  for (size_t i = 0; i < m1.n_cols; ++i)
```
stereomatchingkiss: you want to compare all of the elements of m1 and m2, or just compare the first column of m1 and m2?If it is the first case, I think it should be `m1.n_elem`

src/mlpack/tests/layer_test.cpp
```diff
@@ -0,0 +1,277 @@
...
+  //! todo: Gradient update test for ConcatLayer.
+}
+
+
+BOOST_AUTO_TEST_CASE(SampleInceptionLayerTest)
+{
+  SampleInceptionNetwork();
+  SmallNetworkTest();
+  ConvLayerTest();
+  ConcatLayerTest();
```
stereomatchingkiss: Maybe you can consider put these test cases in different test case, like

```
BOOST_AUTO_TEST_CASE(SampleInceptionNetworkTest)
{
}
```

This way the error messages of test case should be better

src/mlpack/methods/ann/layer/concat_layer.hpp
nilayjain: I couldn't write 2 conditions using std::enable_if (one for I < sizeof...(Tp) and one for output::value = arma::cube), but i can write 2 overloads for output type arma::cube and arma::mat. Is there a better way ?

src/mlpack/tests/layer_test.cpp
nilayjain: I am comparing all columns of m1 with all columns of m2. Is this incorrect?

stereomatchingkiss: >I am comparing all columns of m1 with all columns of m2. Is this incorrect?

Nope, because m1(i) access single data of the arma::mat but not "a column of the mat".
If you want to make sure every columns have the same value, I think it is same as make sure all of the data as equal, change it to n_elem should do the tricks.

nilayjain: Ok, Thanks for clarifying.

src/mlpack/methods/ann/layer/concat_layer.hpp
stereomatchingkiss: >but i can write 2 overloads for output type arma::cube and arma::mat. 

I think this is a good solution.

>Is there a better way ?

I do not know a better way, but it is possible to specialize your function by std::enable_if, you can write something like

```
//pseudo codes
typename std::enable_if<I < sizeof...(Tp) & std::is_arma_mat<DataType>::value>::type

```
Armadillo has a header file define something like std::is_arma_mat, but I forgot where is it. However, this solution is less elegant compare with overload.

src/mlpack/methods/ann/layer/inception_layer.hpp
nilayjain: When writing tests I access the layers from the InceptionLayer object. I will put the variables as private and let layers remain as public.

src/mlpack/methods/ann/layer/inception_layer.hpp
nilayjain: we have to set the weights of all the layers that constitute the inception layer in the constructor? Also inception layer in itself does not do any operation, so we need not define the overall weight dimension.

zoq: The main class (e.g. CNN) has to initialize the weights somehow, to do that it needs to know the parameter size (weights). So, what the main class basically does is to create a parameter vector, initialize that parameter vector and call the Forward, Backward and Gradient function with the initialized parameters.

nilayjain: but inception layer does not do any operation in itself, it just calls its constituent layers, so if inception layer does not need to have any weights, what should we initialize it to?

zoq: As you already pointed out the inception layer is basically a container for a couple of convolution and pooling layer. Since they are abstracted through the inception layer, the main network has now knowledge about the containing layer. So, the inception layer has to forward any weights to any capsulated layer.

PR #760, opened 2016-08-12
Vantage point tree and HRectBound improvements

lozhnikov:
I removed central nodes. Thus, the VP tree may be implemented in terms of `BinarySpaceTree`. So, I decided to remove the `VantagePointTree` class. Since a point may be a vantage point several times, `TreeTraits::FirstSiblingFirstPointIsCentroid` has become useless. So, I removed that.

Moreover, I tried to improve distance calculations in `HRectBound`. Right now, the kd tree works much faster on my system.


rcurtin:
Empirically the performance improvement seems great; both trees tend to be accelerated.  I have seen in one dataset (`phy.csv`, http://www.ratml.org/datasets/phy.csv ) that the VP tree performance in single-tree mode is actually slower after this commit.  I haven't played with enough datasets to find another dataset that exhibits that characteristic, though.

src/mlpack/core/tree/hrectbound_impl.hpp
```diff
@@ -173,18 +173,40 @@ inline ElemType HRectBound<MetricType, ElemType>::MinDistance(
...
   // that was introduced earlier.  The compiler should optimize out the if
   // statement entirely.
-  if (MetricType::TakeRoot)
-    return (ElemType) pow((double) sum, 1.0 / (double) MetricType::Power) / 2.0;
+  if (MetricType::Power == 1)
+    return sum * 0.5;
+  else if (MetricType::Power == 2)
+  {
+    if (MetricType::TakeRoot)
+      return std::sqrt(sum) * 0.5;
```
lozhnikov: I've forgotten to ask about this piece of code:
` pow((double) sum, 1.0 / (double) MetricType::Power)`

Do you use the conversion in order to increase precision or you just take care of integral data types?

Should I replace `std::sqrt(sum)` by `(ElemType) std::sqrt((double) sum)`?

rcurtin: I believe the casts to double were done because `std::pow(double, int)` was not available on all systems. It looks like an overload of `std::sqrt()` is available that takes any integral type as input, but returns `double`, so you would have to cast the result back to `ElemType`. It appears that the same comments now apply about `std::pow` since C++11.

src/mlpack/core/tree/hollow_ball_bound.hpp
```diff
@@ -35,6 +36,8 @@ class HollowBallBound
   math::RangeType<ElemType> radii;
   //! The center of the ball bound.
   arma::Col<ElemType> center;
+  //! The center of the hollow.
+  arma::Col<ElemType> hollowCenter;
```
rcurtin: I don't understand; what is the purpose of the hollow center? When would it be different than the normal center? Maybe I have overlooked something.

src/mlpack/core/tree/hollow_ball_bound.hpp
```diff
@@ -213,6 +221,21 @@ struct BoundTraits<HollowBallBound<MetricType, ElemType>>
...
+{
+  static const bool value = false;
+};
+
+template<typename MetricType, typename ElemType>
+struct IsHollowBallBound<HollowBallBound<MetricType, ElemType>>
+{
+  static const bool value = true;
+};
+
```
rcurtin: It seems to me like these structs aren't used anywhere.

rcurtin:
Everything looks good to me (aside from the few minor comments I made) and the speed improvement is really nice.  An observation I have is that we end up duplicating vantage points because we store separate vectors in the bounds. But I don't see an easy solution for that and I am not sure it would give huge speedup anyway.

src/mlpack/core/tree/hollow_ball_bound.hpp
lozhnikov: The normal center is the center of the minimal bounding ball (i.e. the outer ball). And `hollowCenter` is the center of the hole (i.e. the inner ball).

src/mlpack/core/tree/hollow_ball_bound.hpp
lozhnikov: yeah, I agree. I thought it will be useful, but then I realized that I needn't it.

lozhnikov:
I think it's ok since the bound is not connected with the dataset and a point may become a vantage point several times.

src/mlpack/core/tree/hollow_ball_bound.hpp
rcurtin: I thought to myself, 'surely, holding two centers, which can double the number of distance computations or more for a single score, can't be as fast as only holding one, even if the bounds end up being a little looser'.  But after I tried implementing it, and removing the `hollowCenter` member, the runtimes were always about 10% slower.  So it seems that `hollowCenter` is the right way to go here.

Issue #761, opened 2016-08-15
General Defeatist Search Traverser.


MarcosPividori:
Hi,
We want to provide a general *Defeatist Search Traverser* for every *TreeType*, as we do with `SingleTreeTraverser` and `DualtreeTraversers`.
This discussion started in the PR https://github.com/mlpack/mlpack/pull/747:
+ https://github.com/mlpack/mlpack/pull/747#issuecomment-237901505
+ https://github.com/mlpack/mlpack/pull/747#issuecomment-238137161
+ https://github.com/mlpack/mlpack/pull/747#issuecomment-238985341
+ https://github.com/mlpack/mlpack/pull/747#issuecomment-239206014
+ https://github.com/mlpack/mlpack/pull/747#issuecomment-239602549
+ https://github.com/mlpack/mlpack/pull/747#issuecomment-239870358
+ https://github.com/mlpack/mlpack/pull/747#issuecomment-239872121

If you agree, I prefer the name: `SingleTreeGreedyTraverser`, instead of `SingleTreeDefeatistTraverser`.

rcurtin:
I agree, I think `SingleTreeGreedyTraverser` is just fine for a name.  I think we can write a completely general traverser; I don't think it needs to be tree-specific.

MarcosPividori:
Closed after  PR https://github.com/mlpack/mlpack/pull/762 was merged!

Issue #750, opened 2016-08-04
Use Enum Type to define different possible search modes, in neighbor search class.


MarcosPividori:
About the implementation of NeighborSearch class, I think we should define a enum type like:

```
enum SearchMode
{
    NAIVE_MODE,
    SINGLE_MODE,
    DUAL_MODE
};
```

This way, constructors will be clearer, instead of using combinations of boolean values.

For example, we could have:

```
 NeighborSearch(const MatType& referenceSet, SearchMode mode, double epsilon, MetricType metric);
```

Instead of:

```
 NeighborSearch(const MatType& referenceSet, bool naive, bool singleMode, double epsilon, MetricType metric);
```

Also, the implementation will be clearer, because we could use `switch` statement when analysing the 3 possible situations.

rcurtin:
I agree with this idea.  If we want to do it before mlpack 3.0.0, that's fine, we just have to remember reverse compatibility.

Seems like there is a lot of reverse compatibility going on, maybe a 3.0.0 release is not so far away...

rcurtin:
This actually won't be completed until mlpack 3.0.0 when #768 is merged, but I guess, since we already have the code ready to go, maybe there is no reason to have this ticket open anymore and I should close it again...!

PR #762, opened 2016-08-16
General greedy traverser.

MarcosPividori:
Hi @rcurtin 
I have implemented the general: `GreedySingleTreeTraverser`, as discussed in https://github.com/mlpack/mlpack/issues/761.
I needed to modify the NeighborSearch class, to consider a enum type, as was proposed in: https://github.com/mlpack/mlpack/issues/750
Also, I implemented a new command line option: "--greedy"
I think it is ready to be merged!
Thanks!
Marcos

### DIFF
--- a/src/mlpack/core/tree/binary_space_tree/binary_space_tree.hpp
+++ b/src/mlpack/core/tree/binary_space_tree/binary_space_tree.hpp
@@ -339,7 +339,7 @@ class BinarySpaceTree
   template<typename VecType>
   BinarySpaceTree& GetNearestChild(
       const VecType& point,
-      typename boost::enable_if<IsVector<VecType> >::type* = 0) const;
+      typename boost::enable_if<IsVector<VecType> >::type* = 0);
 
   /**
    * Return the furthest child node to the given query point.  If this is a leaf
@@ -348,7 +348,7 @@ class BinarySpaceTree
   template<typename VecType>
   BinarySpaceTree& GetFurthestChild(
       const VecType& point,
-      typename boost::enable_if<IsVector<VecType> >::type* = 0) const;
+      typename boost::enable_if<IsVector<VecType> >::type* = 0);
 
   /**
    * Return the furthest distance to a point held in this node.  If this is not
--- a/src/mlpack/core/tree/binary_space_tree/binary_space_tree.hpp
+++ b/src/mlpack/core/tree/binary_space_tree/binary_space_tree.hpp
@@ -351,6 +351,18 @@ class BinarySpaceTree
       typename boost::enable_if<IsVector<VecType> >::type* = 0);
 
   /**
+   * Return the nearest child node to the given query node.  If it can't decide
+   * it will return a null pointer.
+   */
+  BinarySpaceTree* GetNearestChild(const BinarySpaceTree& queryNode);
+
+  /**
+   * Return the furthest child node to the given query node.  If it can't decide
+   * it will return a null pointer.
+   */
+  BinarySpaceTree* GetFurthestChild(const BinarySpaceTree& queryNode);
+
+  /**
    * Return the furthest distance to a point held in this node.  If this is not
    * a leaf node, then the distance is 0 because the node holds no points.
    */
--- a/src/mlpack/core/tree/binary_space_tree/binary_space_tree.hpp
+++ b/src/mlpack/core/tree/binary_space_tree/binary_space_tree.hpp
@@ -333,6 +333,24 @@ class BinarySpaceTree
   size_t NumChildren() const;
 
   /**
+   * Return the nearest child node to the given query point.  If this is a leaf
+   * node, it will return a reference to itself.
+   */
+  template<typename VecType>
+  BinarySpaceTree& GetNearestChild(
+      const VecType& point,
+      typename boost::enable_if<IsVector<VecType> >::type* = 0) const;
+
+  /**
+   * Return the furthest child node to the given query point.  If this is a leaf
+   * node, it will return a reference to itself.
+   */
+  template<typename VecType>
+  BinarySpaceTree& GetFurthestChild(
+      const VecType& point,
+      typename boost::enable_if<IsVector<VecType> >::type* = 0) const;
+
+  /**
    * Return the furthest distance to a point held in this node.  If this is not
    * a leaf node, then the distance is 0 because the node holds no points.
    */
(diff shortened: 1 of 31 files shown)


src/mlpack/core/tree/greedy_single_tree_traverser_impl.hpp
```diff
@@ -0,0 +1,49 @@
...
+void GreedySingleTreeTraverser<TreeType, RuleType>::Traverse(
+    const size_t queryIndex,
+    TreeType& referenceNode)
+{
+  // If we have reached a leaf node, run the base case as necessary.
+  if (referenceNode.IsLeaf())
+  {
+    for (size_t i = 0; i < referenceNode.NumPoints(); ++i)
+      rule.BaseCase(queryIndex, referenceNode.Point(i));
+  }
```
rcurtin: I think we should do all base cases regardless of whether or not we are a leaf, since some nodes may hold points in non-leaf nodes. Of the trees we have implemented now I believe that only applies to the cover tree, but more may be implemented in the future that also have that property.

rcurtin:
Aside from my single comment on the code, this looks great and I agree it is ready to merge. I think maybe I want to get 3.0.0 done soon since we are creating a lot of ugly reverse compatibility code! :)

One concern I do have is that right now the `BinarySpaceTree::GetNearestChild()` and `BinarySpaceTree::GetFurthestChild()` functions loop over both children and perform two distance calculations. We can optimize this by calculating which half space the point falls into. Should we wait until the refactoring for random projection trees is done until doing that? If so I am happy to merge now, I just want to make sure that happens eventually.

src/mlpack/core/tree/greedy_single_tree_traverser_impl.hpp
MarcosPividori: Yeah, you are right! Done in: https://github.com/mlpack/mlpack/pull/762/commits/acb134606a30ce2409cf56806e5625030e2ca66a

MarcosPividori:
@rcurtin , Yeah, I think we have to wait until the refactoring of `SplitSpace`.
If we modify the `BinarySpaceTree` to hold a `Hyperplane` instance, we will be able to efficiently estimate the nearest and furthest child as I do with SpillTrees in: https://github.com/MarcosPividori/mlpack/blob/greedy-traverser/src/mlpack/core/tree/spill_tree/spill_tree_impl.hpp#L333
I think this is ready to be merged.

src/mlpack/methods/neighbor_search/neighbor_search.hpp
```diff
@@ -309,14 +412,21 @@ class NeighborSearch
   //! Return the number of node combination scores during the last search.
   size_t Scores() const { return scores; }
 
+  //! Access the search mode.
+  NeighborSearchMode SearchMode() const { return searchMode; }
+  //! Modify the search mode.
+  void SetSearchMode(const NeighborSearchMode mode);
```
rcurtin: I think I'd prefer this to be `NeighborSearchMode& SearchMode()` instead of the way it is now, so it can operate the same way the other functions did.  I realize that this can cause some difficulty (this is similar to why you need `UpdateSearchMode()`), but I think we can resolve any actual problems from this by saying that calling `SingleMode()` or `Naive()` (until mlpack 3.0.0 when they are gone) will override whatever `SearchMode()` is set to.

Alternately, we remove `SearchMode()` and `SetSearchMode()` and add a `greedy` member and do the same thing that you are doing with `naive` and `singleMode`.  But I like that solution less.

MarcosPividori: @rcurtin  Yes, I agree that it is preferable to have: `NeighborSearchMode& SearchMode()`
Suppose I add that method, the problem it this:
Until mlpack 3.0.0, we have the modifiers: `Naive()` and `SingleMode()`.
So, when a method is called, and I see a difference between the booleans values: `naive` and `singleMode` and the member `searchMode` (for example: `naive == true` , `singleMode == false` , `searchMode == DUAL_TREE_MODE` ) we don't know which fields were modified, the boolean flags `naive` and `singleMode`, or the `searchMode` member.  So, I don't know if the user want to do naive search or dual tree search.
Because of that I decided to include a  member `SetSearchMode()`.
One solution I can think, I am not sure if this is what you suggested, is to include the modifier `SearchMode()` and a comment like:
"Untile mlpack 3.0.0, if you modify the SearchMode, you also have to modify the flags Naive() and SingleMode()"
What do you think?
Thanks!

src/mlpack/core/tree/binary_space_tree/binary_space_tree.hpp
```diff
@@ -333,6 +333,36 @@ class BinarySpaceTree
...
+   * Return the nearest child node to the given query node.  If it can't decide
+   * it will return a null pointer.
+   */
+  BinarySpaceTree* GetNearestChild(const BinarySpaceTree& queryNode);
+
+  /**
+   * Return the furthest child node to the given query node.  If it can't decide
+   * it will return a null pointer.
+   */
+  BinarySpaceTree* GetFurthestChild(const BinarySpaceTree& queryNode);
```
rcurtin: I don't think you're using these overloads anywhere. Can I suggest returning a `size_t` instead of a reference or pointer? You can return `NumChildren()` to indicate an invalid child. This avoids the issues with returning `*this` and there's never a need to return `NULL`. The user can check on their own whether or not the results are valid and we also manage to keep our TreeType API simpler. What do you think?

MarcosPividori: @rcurtin  The overloads of spill trees are used for `DefeatistDualTreeTraverser`. Also, all of these overloads will be used when implementing a greedy dual tree traverser.
I agree that it is best to return an index. I have updated the code in: https://github.com/mlpack/mlpack/pull/762/commits/8e305125edbc85f6f037fbb01b081509153641ff
Thanks!

src/mlpack/methods/neighbor_search/neighbor_search.hpp
rcurtin: You're right; I thought there was a way to work around this problem but really there is not a good one.  What I really want to avoid is making an API change that we'll regret later (like `SetSearchMode()`).  So if we are going to break reverse compatibility in 3.0.0, then I would say maybe the easiest thing to do here is just add a `greedy` boolean parameter that works the same way as `singleMode` and `naive`, and use that for now.  Then, when we release 3.0.0, we'll remove all three of those and _only_ have `searchMode` in the way you've written here.  This would allow us to avoid the weird stuff we have to do for reverse compatibility.

I guess, if you agree with that, then we should save your commits for refactoring by opening another reverse-incompatible "3.0.0 PR" that will be merged in when we decide it is time for a release there.

PR #763, opened 2016-08-17
Update README.md

Kirizaki:
Add info about building mlpack under windows


rcurtin:
Thanks!  I reordered the list in b067ee4 since I think it is more important to link to the usual build instructions first; I think most of our users are on Linux/Unix-like systems.

PR #765, opened 2016-08-18
Use move semantics to set a given reference tree in NeighborSearch class.

MarcosPividori:
@rcurtin 
I moved into a new branch, and I didn't realize this would close the PR. The previous PR was https://github.com/mlpack/mlpack/pull/743.
At this point everything is ready to merge. The only problem is that many warnings are shown when compiling because many methods use the deprecated constructors, passing a pointer to the reference tree.
It would take some time to update all that parts of the code.
Thanks

### DIFF
--- a/src/mlpack/core/tree/cover_tree/traits.hpp
+++ b/src/mlpack/core/tree/cover_tree/traits.hpp
@@ -33,6 +33,12 @@ class TreeTraits<CoverTree<MetricType, StatisticType, MatType, RootPointPolicy>>
   static const bool HasOverlappingChildren = true;
 
   /**
+   * Cover trees do have self-children, so points can be included in more than
+   * one node.
+   */
+  static const bool HasDuplicatedPoints = true;
+
+  /**
    * Each cover tree node contains only one point, and that point is its
    * centroid.
    */
--- a/src/mlpack/core/util/deprecated.hpp
+++ b/src/mlpack/core/util/deprecated.hpp
@@ -1,0 +1,19 @@
+/**
+ * @file deprecated.hpp
+ * @author Marcos Pividori.
+ *
+ * Definition of the DEPRECATED macro.
+ */
+#ifndef MLPACK_CORE_UTIL_DEPRECATED_HPP
+#define MLPACK_CORE_UTIL_DEPRECATED_HPP
+
+#ifdef __GNUG__
+#define DEPRECATED __attribute__((deprecated))
+#elif defined(_MSC_VER)
+#define DEPRECATED __declspec(deprecated)
+#else
+#pragma message("WARNING: You need to implement DEPRECATED for this compiler")
+#define DEPRECATED
+#endif
+
+#endif
--- a/src/mlpack/core/util/deprecated.hpp
+++ b/src/mlpack/core/util/deprecated.hpp
@@ -2,18 +2,19 @@
  * @file deprecated.hpp
  * @author Marcos Pividori.
  *
- * Definition of the DEPRECATED macro.
+ * Definition of the mlpack_deprecated macro.
  */
 #ifndef MLPACK_CORE_UTIL_DEPRECATED_HPP
 #define MLPACK_CORE_UTIL_DEPRECATED_HPP
 
 #ifdef __GNUG__
-#define DEPRECATED __attribute__((deprecated))
+#define mlpack_deprecated __attribute__((deprecated))
 #elif defined(_MSC_VER)
-#define DEPRECATED __declspec(deprecated)
+#define mlpack_deprecated __declspec(deprecated)
 #else
-#pragma message("WARNING: You need to implement DEPRECATED for this compiler")
-#define DEPRECATED
+#pragma message("WARNING: You need to implement mlpack_deprecated for this "
+    "compiler")
+#define mlpack_deprecated
 #endif
 
 #endif
--- a/src/mlpack/methods/neighbor_search/neighbor_search.hpp
+++ b/src/mlpack/methods/neighbor_search/neighbor_search.hpp
@@ -134,9 +134,12 @@ class NeighborSearch
    * Additionally, an instantiated distance metric can be given, for cases where
    * the distance metric holds data.
    *
-   * There is no copying of the data matrices in this constructor (because
-   * tree-building is not necessary), so this is the constructor to use when
-   * copies absolutely must be avoided.
+   * This method is deprecated and will be removed in mlpack 3.0.0! Constructors
+   * taking a reference to the reference tree are prefered.
+   *
+   * This method won't take ownership of the given tree. There is no copying of
+   * the data matrices in this constructor (because  tree-building is not
+   * necessary).
    *
    * @note
    * Mapping the points of the matrix back to their original indices is not done
@@ -150,10 +153,10 @@ class NeighborSearch
    * @param epsilon Relative approximate error (non-negative).
    * @param metric Instantiated distance metric.
    */
-  NeighborSearch(Tree* referenceTree,
-                 const NeighborSearchMode mode = DUAL_TREE_MODE,
-                 const double epsilon = 0,
-                 const MetricType metric = MetricType());
+  DEPRECATED NeighborSearch(Tree* referenceTree,
+                            const NeighborSearchMode mode = DUAL_TREE_MODE,
+                            const double epsilon = 0,
+                            const MetricType metric = MetricType());
 
   /**
    * Create a NeighborSearch object without any reference data.  If Search() is
@@ -281,10 +284,10 @@ class NeighborSearch
    * @param epsilon Relative approximate error (non-negative).
    * @param metric Instantiated distance metric.
    */
-  NeighborSearch(Tree* referenceTree,
-                 const bool singleMode,
-                 const double epsilon = 0,
-                 const MetricType metric = MetricType());
+  DEPRECATED NeighborSearch(Tree* referenceTree,
+                            const bool singleMode,
+                            const double epsilon = 0,
+                            const MetricType metric = MetricType());
 
   /**
    * Initialize the NeighborSearch object with the given pre-constructed
@@ -364,9 +367,12 @@ class NeighborSearch
   /**
    * Set the reference tree to a new reference tree.
    *
+   * This method is deprecated and will be removed in mlpack 3.0.0! Train()
+   * methods taking a reference to the reference tree are prefered.
+   *
    * @param referenceTree Pre-built tree for reference points.
    */
-  void Train(Tree* referenceTree);
+  DEPRECATED void Train(Tree* referenceTree);
 
   /**
    * Set the reference tree as a copy of the given reference tree.
--- a/src/mlpack/methods/neighbor_search/neighbor_search.hpp
+++ b/src/mlpack/methods/neighbor_search/neighbor_search.hpp
@@ -127,39 +127,6 @@ class NeighborSearch
                  const MetricType metric = MetricType());
 
   /**
-   * Initialize the NeighborSearch object with the given pre-constructed
-   * reference tree (this is the tree built on the points that will be
-   * searched).  Optionally, perform the computation in a different mode.
-   * Naive mode is not available as an option for this constructor.
-   * Additionally, an instantiated distance metric can be given, for cases where
-   * the distance metric holds data.
-   *
-   * This method is deprecated and will be removed in mlpack 3.0.0! Constructors
-   * taking a reference to the reference tree are prefered.
-   *
-   * This method won't take ownership of the given tree. There is no copying of
-   * the data matrices in this constructor (because  tree-building is not
-   * necessary).
-   *
-   * @note
-   * Mapping the points of the matrix back to their original indices is not done
-   * when this constructor is used, so if the tree type you are using maps
-   * points (like BinarySpaceTree), then you will have to perform the re-mapping
-   * manually.
-   * @endnote
-   *
-   * @param referenceTree Pre-built tree for reference points.
-   * @param mode Neighbor search mode.
-   * @param epsilon Relative approximate error (non-negative).
-   * @param metric Instantiated distance metric.
-   */
-  mlpack_deprecated NeighborSearch(
-      Tree* referenceTree,
-      const NeighborSearchMode mode = DUAL_TREE_MODE,
-      const double epsilon = 0,
-      const MetricType metric = MetricType());
-
-  /**
    * Initialize the NeighborSearch object with a copy of the given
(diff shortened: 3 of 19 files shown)


rcurtin:
When my brain is fried tomorrow or later today, I'll go ahead and fix all the warnings. :)

MarcosPividori:
@rcurtin  thanks! I realized that we have to fix the copy constructor of Rectangle trees, and implement a move constructor.

src/mlpack/core/util/deprecated.hpp
```diff
@@ -0,0 +1,19 @@
...
+#ifndef MLPACK_CORE_UTIL_DEPRECATED_HPP
+#define MLPACK_CORE_UTIL_DEPRECATED_HPP
+
+#ifdef __GNUG__
+#define DEPRECATED __attribute__((deprecated))
+#elif defined(_MSC_VER)
+#define DEPRECATED __declspec(deprecated)
+#else
+#pragma message("WARNING: You need to implement DEPRECATED for this compiler")
+#define DEPRECATED
```
rcurtin: Do you think we could call this `mlpack_deprecated` so it has the same format (lowercase) as the few other mlpack macros?

MarcosPividori: Yes, I agree! Done in: https://github.com/mlpack/mlpack/pull/765/commits/84195797da5a8ade52a54891add590d3be7d606e

rcurtin:
I looked through this on the plane today and I think the biggest issue is providing a way for users to pass `oldFromNew` to the `NeighborSearch` class.  I disagree with this... a long explanation follows...

I believe that there are two ways one can use the `NeighborSearch` class: either you can pass in a dataset (possibly by reference where it may be copied, possibly by move constructor where it won't), and then the `NeighborSearch` class is responsible for the tree and everything associated with it.  Otherwise, you pass in a pre-constructed tree and the `NeighborSearch` class is not responsible for it.  One of the key aspects of "responsibility" here is that if the `NeighborSearch` class is responsible, then it must handle the `oldFromNew` mappings that _might_ exist if `RearrangesDataset == true`.  But `RearrangesDataset` is not always true, so there are only some cases where we need it.

Thus, if we allow the user to make `NeighborSearch` take control of the tree, this necessitates that *sometimes*, they will also have to pass `oldFromNew` in _with_ the tree; but not always.  We have no way to make a clean API that only has the right overload available depending on the `TreeType`.  (Yes, we can do SFINAE so only one is accepted, but an API with lots of `enable_if`s that is presented to the end user can be very confusing.  Armadillo handles this by basically hiding all of its internal details, but I don't like that strategy.)

So there is another option: we can hold the mappings inside of the `TreeType` and do away with `RearrangesDataset` entirely.  But this is problematic also: it means that every time we call `Point()`, we must apply the mapping... this will cause unacceptable slowdowns for anything that uses `BinarySpaceTree`.  So that's not really an option we can pursue.

There is a final option: do some crazy template overloading in order to hold `oldFromNew` inside of `BinarySpaceTree` but avoid de-mapping every time `Point()` is called.  I don't think that's a great idea though (and I'm not even sure it would work).  In addition it would increase the size of the `BinarySpaceTree` class which is something I'd like to try to avoid if possible.

The policy I have dealt with so far has worked like this: if the user passes in a dataset, then `NeighborSearch` manages the tree and the mappings, and the results you get are what you expect.  But if the user passes in the tree, then it is up to them to reverse the mappings.  This is potentially useful because it offers some speedup in the situation where the user has their own tree, but, say, aren't concerned with the ordering of their points and when their points are remapped, they don't care about the new ordering.

In my view, we can't know whether or not the user truly wants the points to be unmapped when they pass a tree, and I want to avoid the extra complexity of an API with a bunch of `Train()` overloads where some take a mappings vector and some don't.  So in my opinion, we can solve this by not allowing an overload like `Train(TreeType&&)` and instead only allowing `Train(TreeType&)` (implying that the `NeighborSearch` object does not own the mappings or tree and is not responsible for unmapping the results when `Search()` is called).  If we only allow `Train(TreeType&)`, we don't own the tree, so even if the tree does map points, we do not need to do the unmapping ourselves.

I hope my very long comment makes some amount of sense.  I am open to the concept of an eventual redesign, but I believe we must strongly pursue keeping the API simple enough for the average user but flexible enough for an advanced user to be able to get what they need done.  In this case having fewer overloads of `Train()` keeps it simple, but if an advanced user wants to do something tricky with a custom-built tree that maps points, then they can also figure out how to map the points back to their original indices.

rcurtin:
Hah, I wrote "I looked through this on the plane today" but that was actually yesterday... it took me two days to write the comment... :)

MarcosPividori:
@rcurtin thanks for your comments!

The main problem that I am trying to solve here is this:
When we set a reference tree, that tree will *become part of the state* of `NeighborSearch` class, because `NeighborSearch` class will *hold a pointer* to that tree. This is the key point.
So, I think the usage of pointer or references in this case is not the best approach.
For example, we could face problems like:
```
  KNN knn;

  arma::mat dataset = arma::randu<arma::mat>(5, 100);
  std::vector<size_t> oldFromNewReferences;
  KNN::Tree* tree = new KNN::Tree(dataset, oldFromNewReferences);

  knn.Train(tree);

  delete tree;

  arma::Mat<size_t> neighbors;
  arma::mat distances;
  knn.Search(10, neighbors, distances); // Segmentation Fault.
```

So, I think that if we want to set a given reference tree in the `NeighborSearch` class, as it will become part of its state, we should provide 2 options:
 + Pass a lvalue reference, so the tree will be copied.
```
void Train(Tree& referenceTree); // Copy the reference tree.
```
 + Pass a rvalue reference, so the tree will be moved.
```
void Train(Tree&& referenceTree); // Move the reference tree.
```
In both cases the `NeighborSearch` class will take ownership of the tree that it is holding a pointer to.
It would be similar to the approach of the c++ standard library for containers.
For example, for `std::vector`:
```
void push_back (const value_type& val);  // copies the value.
void push_back (value_type&& val); // moves the value.
```

As you suggests, we could assume that in both cases, as the tree was provided by the user, the `NeighborSearch` class shouldn't take care of the mappings.

Another optional approach, that I think is not necessary, is to provide a new method:
```
void SetMappings(std::vector<size_t>& oldFromNew); // The mappings will be copied
void SetMappings(std::vector<size_t>&& oldFromNew); // The mappings will be moved.
```
So, if users want the points to be mapped, they can set the mapping.

But I think this is not necessary, assuming that if users created their own reference tree, they should care of the mappings on their own.


So, I suggest to continue with this approach:
+ Provide the methods.
```
void Train(Tree& referenceTree); // Copy the reference tree.
void Train(Tree&& referenceTree); // Move the reference tree.
```
+ Remove the boolean flag `ownsTree`, because it will be always true.
+ Add a boolean flag `considerMappings`. When the `NeighborSearch` class created the tree from a given dataset, it will set: `considerMappings = true`.
When a reference tree was provided by the user, it will set: `considerMappings = false`.

What do you think?

Thanks!

MarcosPividori:
Travis succeeded except for `RAModelTest`, which I think will succeed if we restart it.

MarcosPividori:
I have forced a rebuild.

MarcosPividori:
Finally it succeeded!

MarcosPividori:
Hi @rcurtin ,
I have rebased on top of master. Also, I removed the `oldFromNew` parameter. Now, I adopt exactly the same approach than with previous method: `void Train(Tree* referenceTree)`  "when providing a reference tree, the points won't be mapped".
So, I added two `Train()` methods:
```
void Train(Tree& referenceTree); // Copy the reference tree.
void Train(Tree&& referenceTree); // Move the reference tree.
```
What do you think?
Thanks!

rcurtin:
Ok, I like the approach you have proposed.  We avoid exposing the mappings problem to the user, and instead if they pass a tree that has been mapped, it is their responsibility to apply the mappings.  My only suggestion is that we make it `Train(const TreeType& tree)`, so that the user knows their tree will not be modified but instead the tree will be copied.

Long ago, before we had C++11 which gave us rvalue references, this would have been a much trickier question, but now, you are right, we can just use `Train(TreeType&& tree)` so that the user can allow the `NeighborSearch` object to take ownership of the tree.

But, there is still one use case I want to consider that this approach doesn't address... what if the user wants to create their own tree that `NeighborSearch` can use, and they don't want to copy the tree when they call `Train()`, but they then want to use that tree for something else?  Trees can be very large, so, if e.g. a user wanted to use a tree with `KNN` then with `KFN`, they should be able to do this without needing to copy the tree.

I guess it is possible to provide

``
TreeType&& ReferenceTree() { return std::move(*referenceTree); }
```

and then a user can do

```
TreeType getMyTreeBack = std::move(knn.ReferenceTree());
```

What do you think?  Is there a better way to do this?  It seems a bit odd to make the user go to weird lengths like this:

```
// Build custom tree and train the model.
TreeType myTree(myData);
KNN knn;
knn.Train(std::move(myTree));

arma::mat distances;
arma::Mat<size_t> neighbors;
knn.Search(3, neighbors, distances);

// Create a KFN model too and reuse the tree.
KFN kfn;
kfn.Train(std::move(knn.ReferenceTree()));
// I'm ignoring the fact that we have to reset the statistic for now...
arma::mat kfnDistances;
arma::Mat<size_t> kfnNeighbors;
kfn.Search(3, kfnNeighbors, kfnDistances);
```

That seems a bit awkward, but it's consistent with the rest of the code, so I am not sure if there is any better way to do it... let me know if you have any other ideas.

I'll try and find time to do a more detailed review of the code shortly.

**Comment by rcurtin** [comment]
Here's a patch for the warnings, and it also fixes some behavior with dual-tree k-means.
http://www.ratml.org/misc/warnings.patch

**Comment by MarcosPividori** [comment]
@rcurtin  Thanks for your comments!
I agree that we should provide access to the reference tree, through a method like ReferenceTree(), so users can get it back in case this is necessary. I have updated the code including your suggestions.
However, with current implementation, we can't do what you proposes. I mean, use the same tree for KNN and KFN, because that trees consider different statistic types (NeighborSearchStat includes a template parameter: SortPolicy), so the compiler rejects that code.

**Comment by rcurtin** [comment]
Yep, you are right, that specific code does not work.  But I have a plan that will make something of that ilk possible... :)

Let me take a pass through in the next few hours and then I think this is ready to merge, if you agree.

File: src/mlpack/core/tree/cover_tree/traits.hpp [hunk 2]
```diff
@@ -33,6 +33,12 @@ class TreeTraits<CoverTree<MetricType, StatisticType, MatType, RootPointPolicy>>
   static const bool HasOverlappingChildren = true;
 
   /**
+   * Cover trees do have self-children, so points can be included in more than
+   * one node.
+   */
+  static const bool HasDuplicatedPoints = true;
```
rcurtin: I think that `HasDuplicatedPoints` was meant to encode whether or not `NumDescendants()` accounted for each point uniquely.  In the case of the cover tree, even though points are duplicated, the properties of the self-child (and the fact that it only holds one point) mean that we can correctly calculate the number of unique descendants and return them with `Descendant()`.  So maybe our name `HasDuplicatedPoints` should be changed somehow, but either way, I think that this value should be `false`.

src/mlpack/methods/neighbor_search/neighbor_search.hpp
```diff
@@ -150,10 +153,70 @@ class NeighborSearch
...
    */
-  NeighborSearch(Tree* referenceTree,
-                 const NeighborSearchMode mode = DUAL_TREE_MODE,
-                 const double epsilon = 0,
-                 const MetricType metric = MetricType());
+  mlpack_deprecated NeighborSearch(
+      Tree* referenceTree,
+      const NeighborSearchMode mode = DUAL_TREE_MODE,
+      const double epsilon = 0,
+      const MetricType metric = MetricType());
```
rcurtin: There was never a release of this constructor in a non-deprecated form, so I think we can simply remove this one.

src/mlpack/methods/neighbor_search/neighbor_search.hpp
```diff
@@ -413,18 +500,21 @@ class NeighborSearch
   size_t Scores() const { return scores; }
 
   //! Access whether or not search is done in naive linear scan mode.
+  //! Deprecated. Will be removed in mlpack 3.0.0.
```
rcurtin: Hm, I dunno, maybe we should update this comment.  Someone might see it and say "but what are we supposed to use once these methods are gone?", but we don't have any advice for them there.  I know that we can't add `SearchMode()` until 3.0.0 though.  So I guess possibly we could just point out that `SearchMode()` will be available in 3.0.0 and will be the replacement for these functions.

rcurtin:
Other than the comments I made, I think everything is good here and it's ready for merge.  We should at some point roll out the same changes to `RASearch`, `RangeSearch`, `FastMKS`, and `DualTreeBoruvka`, but we can open a ticket that says that and just merge these changes now---your call.

src/mlpack/core/tree/cover_tree/traits.hpp
MarcosPividori: Hi @rcurtin ,
In fact we use another flag: `UniqueNumDescendants`,  to recognize the case where `NumDescendants()` includes repeated points.
The flag `HasDuplicatedPoints` is not used anywhere, in fact I orginally removed it but you suggested in https://github.com/mlpack/mlpack/pull/743#discussion_r73456217 that you would like to keep it for the future.
Would you prefer to remove it or keeping it for future use?

src/mlpack/methods/neighbor_search/neighbor_search.hpp
MarcosPividori: Yes, I agree with you. Done in https://github.com/mlpack/mlpack/pull/765/commits/db16a10803e3107a2fd96ae0d196f8f8685cf076

src/mlpack/core/tree/cover_tree/traits.hpp
rcurtin: A classic example of me talking before sufficiently analyzing what I am looking at.  Let's leave it as-is. :)

src/mlpack/methods/neighbor_search/neighbor_search.hpp
MarcosPividori: Yes, I agree that it can be confussing. I have added some comments in: https://github.com/mlpack/mlpack/pull/765/commits/0d071af207002e9c2054bd60f85765942994e0ff

MarcosPividori:
@rcurtin  Yes, I agree. I have created a new issue: https://github.com/mlpack/mlpack/issues/773
I plan to work updating that code in the next days. Also, to include the changes to use `boost variant`.

PR #766, opened 2016-08-19
Fix furthest neighbor sort policy.

MarcosPividori:
@rcurtin 
Was this change https://github.com/mlpack/mlpack/commit/b067ee4bb6cff16be8efb1ad4923052bc12a13c5#diff-1aaf3b3b22ad7dee835d64635b303703L155
 intensional? 
It breaks some tests in my computer...


rcurtin:
Yep, it was an accidental `git commit -a`.  Thanks!

PR #767, opened 2016-08-19
Fixed copy constructor of RectangleTree and added move constructor.

lozhnikov:
I've fixed the copy constructor of `RectangleTree` and implemented move constructor for PR #765 .


rcurtin:
Looks good, thanks.

PR #768, opened 2016-08-20
Search mode 3.0.0

MarcosPividori:
@rcurtin 
This PR removes the reverse compatibility code inside `NeighborSearch` class.
This is intended to be merged for version 3.0.0.
The only difference with the PR https://github.com/mlpack/mlpack/pull/762  is the last commit.
I will rebase on top of master when https://github.com/mlpack/mlpack/pull/762 is merged, so it will be clearer.
Thanks!
Marcos


rcurtin:
Merged.  Thanks again Marcos, these are nice changes!

PR #770, opened 2016-08-26
Adding option to set the root in coverage script.

sumedhghaisas:


rcurtin:
Looks good to me.  Just today our build server moved from big.mlpack.org to masterblaster.mlpack.org, so I transitioned the jobs.  But there may be some packages I still need to set up, so I will try to take care of that.  I also need to set up Github authentication for login to Jenkins, but if you need access before I can get to that, send me a message on IRC and I'll set up a temporary local account.

PR #775, opened 2016-09-02
Automatic bindings step 2: matrix input/output parameters

rcurtin:
Here I have completely refactored the CLI code in order to support matrix-type parameters.  So now I can write

```
PARAM_MATRIX("input", "Input matrix.", "i");
```

and then in my code,

```
// Either works:
arma::mat& matrix = CLI::GetParam<arma::mat>("input");
arma::mat matrixmove = std::move(CLI::GetParam<arma::mat>("input"));
```

And the user specifies the matrix on the command line as `--input_file` (and the documentation says that too).

There are a couple variations, like `PARAM_UMATRIX_IN` (which is `arma::Mat<size_t>`), and `PARAM_TMATRIX_IN` (which is a non-transposed-on-load matrix), and `PARAM_MATRIX_OUT` (an output matrix).  Output matrices are saved when the program terminates, and input matrices are loaded when `GetParam<>()` is called for the first time with that argument name.

The PR also contains a few bugfixes for CLI and vastly expands the test suite that we had for CLI.

In most (but not all cases) I have been able to adapt `PARAM_STRING_IN` parameters to `PARAM_MATRIX_IN` parameters and preserve reverse compatibility, but there are a few cases where I've broken reverse compatibility so technically this means this patch can't happen until mlpack 3.0.0.  I think maybe it is worth having an mlpack-3.x branch, and then switch master to that branch when 3.0.0 is released.  Alternately we release mlpack 2.1.0 very soon (like next week) with the features we have now, and then merge in the mlpack 3 features and break reverse compatibility in the git master branch (I am fine with that).

Lastly, the long-term: now that CLI loads matrices automatically, I should be next able to modify the command-line programs in such a way that they can be generated as bindings for other languages too.  Since I can extract the parameters directly from CLI, the programs can be agnostic to how they are being used.  So I will probably first play with generating a Python binding for PCA (since it does not save any models); then go from there.

I'll let this PR sit for a few weeks (since I am not in a hurry here) and then if there are no comments I'll go ahead and merge it and choose arbitrarily whether to make a new branch or to just release 2.1.0 and move on. :)


src/mlpack/core/util/cli.cpp
```diff
@@ -86,165 +207,6 @@ CLI::~CLI()
   // may not have wanted it there (i.e. in Boost unit tests).
```
zoq: One of this picky comments: Wasn't there a "rule" that words at the beginning of a sentences are capitalised?

src/mlpack/core/util/cli.cpp
```diff
@@ -400,15 +320,42 @@ void CLI::ParseCommandLine(int argc, char** line)
...
+        if (bpo.options[i].string_key == bpo.options[j].string_key)
+        {
+          // If a duplicate is found, check to see if either one has a value.
+          if (bpo.options[i].value.size() == 0 &&
+              bpo.options[j].value.size() == 0)
+          {
+            // If neither has a value, consider it a duplicate flag and remove
+            // the duplicate. It's important to not break out of this loop
+            // because there might be another duplicate later on in the vector.
+            bpo.options.erase(bpo.options.begin() +  j);
```
zoq: I think, If we don't break, and we remove another duplicate later on in the vector, we could run into an index out of bounds error.

rcurtin: Nice catch, thanks.  Fixed in 139ad60.

src/mlpack/core/util/cli.cpp
rcurtin: Fixed in 3f27740.

PR #782, opened 2016-09-13
pkg-config support

rcurtin:
This adds a target, 'pkgconfig', which produces the file mlpack.pc.  This can be used with pkgconfig to get the list of include directories and linker arguments that are needed to compile against mlpack.

I think is a small enough change that I'll merge it in after a few days and after the test builds pass and I manually test on OS X.  @eddelbuettel: if you'd like to check this before I merge, it would be appreciated, but you could also test it after merge and file any bugs if it doesn't work the way you need it to. :)


rcurtin:
Tested on OS X, so I'll go ahead and merge.

PR #786, opened 2016-09-15
Use correct name for linear regression program.

MHenderson:


rcurtin:
Nice catch, thank you!

Issue #787, opened 2016-09-15
Replace `MinDistance(TreeType*)` with `MinDistance(TreeType&)`


rcurtin:
Right now, the TreeType API requires the following functions to have this signature:

```
ElemType MinDistance(const TreeType* other);
ElemType MaxDistance(const TreeType* other);
math::RangeType<ElemType> RangeDistance(const TreeType* other);
```

But we can easily use references here.  We should refactor the trees (`mlpack/core/tree/*/`) to instead use the signature

```
ElemType MinDistance(const TreeType& other);
ElemType MaxDistance(const TreeType& other);
math::RangeType<ElemType> RangeDistance(const TreeType& other);
```

That will also force some changes in the dual-tree algorithm code, but it should be relatively simple refactoring in the end.  This would be a good starter bug for someone looking to get involved---but I want it to be done before the 2.1.0 release, which is soon, so I may end up doing this myself in the next two weeks if nobody beats me to a PR. :)

chvsp:
Hi @rcurtin I would like to solve this bug. Seems pretty straightforward. I would be starting work after 22 September as I have exams till then. Just letting you know.

rcurtin:
Fixed by #788.  Thanks @akanuraj200!

PR #790, opened 2016-09-24
Octree

rcurtin:
@mentekid @MarcosPividori @lozhnikov: I wanted to implement one more tree type (the octree) before implementing SLEEC for automatic nearest neighbor search selection.  Here's that implementation, maybe it is of interest.  Octrees are very good for low dimensions (like 2 or 3) and can outperform kd-trees there.  This implementation does:

```
◈ ryan@humungus ☃ build-nodebug ◈ $ bin/mlpack_knn -r ~/datasets/birch3.txt -k 3 -v -t octree | grep 'were\|computing_neighbors\|tree_building'
[INFO ] 1351544 node combinations were scored.
[INFO ] 2178671 base cases were calculated.
[INFO ]   computing_neighbors: 0.171068s
[INFO ]   tree_building: 0.365612s
◈ ryan@humungus ☃ build-nodebug ◈ $ bin/mlpack_knn -r ~/datasets/birch3.txt -k 3 -v | grep 'were\|computing_neighbors\|tree_building'
[INFO ] 1113714 node combinations were scored.
[INFO ] 2714362 base cases were calculated.
[INFO ]   computing_neighbors: 0.167564s
[INFO ]   tree_building: 0.040383s
```

The only real question I have, is, the command-line programs take the parameter name `-t octree` for octrees, but the other trees are `-t kd`, `-t r`, and so forth.  Maybe should I have gone with `-t oct` or `-t oc` to make things similar across all trees?

If anyone spots any bugs, problems, or opportunities for improvement, I'll get them fixed quickly, otherwise I will probably let this sit for a week and then merge it in. :)  My goal is for this to be part of mlpack 2.1.0.


akanuraj200:
Minor refactoring needed to be compatible with #787 
(`src/mlpack/core/tree/octree/octree.hpp` and `src/mlpack/core/tree/octree/octree_impl.hpp`)

src/mlpack/core/tree/octree/octree_impl.hpp
```diff
@@ -0,0 +1,854 @@
...
+//! Serialize the tree.
+template<typename MetricType, typename StatisticType, typename MatType>
+template<typename Archive>
+void Octree<MetricType, StatisticType, MatType>::Serialize(
+    Archive& ar,
+    const unsigned int /* version */)
+{
+  using data::CreateNVP;
+
+  //
```
lozhnikov: It seems you have forgotten to remove `//`:)

src/mlpack/core/tree/octree/octree_impl.hpp
```diff
@@ -0,0 +1,854 @@
...
+  // No need to split if we have fewer than the maximum number of points in this
+  // node.
+  if (count <= maxLeafSize)
+    return;
+
+  // We must split the dataset by sequentially creating each of the children.
+  // We do this in two steps: first we make a pass to count the number of points
+  // that will fall into each child; then in the second pass we rearrange the
+  // points and create the children.
+  arma::Col<size_t> childCounts(std::pow(2, dataset->n_rows),
```
lozhnikov: I think that `((size_t) 2) << dataset->n_rows` is better than `std::pow()`, how do you think?

lozhnikov: Hm.. I am mistaken, of course I mean `((size_t) 1) << dataset->n_rows`.

rcurtin: I tested it out with both clang and gcc in a simple test program, and the assembly produced is way shorter with your solution.  I'll change the code accordingly.

src/mlpack/core/tree/octree/octree_impl.hpp
```diff
@@ -0,0 +1,907 @@
...
+size_t Octree<MetricType, StatisticType, MatType>::GetNearestChild(
+    const Octree& queryNode) const
+{
+  // It's possible that this could be improved by caching which children we have
+  // and which we don't, but for now this is just a brute force search.
+  ElemType bestDistance = DBL_MAX;
+  size_t bestIndex = NumChildren();
+  for (size_t i = 0; i < NumChildren(); ++i)
+  {
+    const double dist = children[i]->MaxDistance(queryNode);
```
lozhnikov: I think that should be `MinDistance()`, how do you think?

lozhnikov:
It seems you forgot to update `PARAM_STRING_IN("tree_type")` at `knn_main.cpp:73`.

rcurtin:
Ok, I've checked over all the comments and also reviewed the code myself (now that I didn't just write it) and I think this is ready to merge, so if there are no more comments, I'll merge it tomorrow.

mentekid:
Sorry, this took me a while to look at.

I agree with going with `oct` instead of `octree` as a parameter option. I gave the code a glance but I'm not as familiar with mlpack's trees as Marcos and Mikhail. Everything looks good to me :)

PR #792, opened 2016-09-27
Implemeted gradient descent optimizer.

sumedhghaisas:


src/mlpack/core/optimizers/gradient_descent/gradient_descent_impl.hpp
```diff
@@ -0,0 +1,80 @@
...
+template<typename FunctionType>
+double GradientDescent<FunctionType>::Optimize(
+    arma::mat& iterate)
+{
+  // To keep track of where we are and how things are going.
+  double overallObjective = function.Evaluate(iterate);
+  double lastObjective = DBL_MAX;
+
+  // Now iterate!
+  arma::vec gradient(iterate.n_cols);
```
rcurtin: This should be `arma::mat` not `arma::vec`, and you should use both the number of rows and columns to initialize it.  Otherwise it will fail if we use an objective function that uses a 2-d matrix instead of a 1-d vector.

src/mlpack/core/optimizers/gradient_descent/gradient_descent_impl.hpp
```diff
@@ -0,0 +1,80 @@
...
+  {
+    // Output current objective function.
+    Log::Info << "Gradient Descent: iteration " << i << ", objective " 
+      << overallObjective << "." << std::endl;
+
+    if (std::isnan(overallObjective) || std::isinf(overallObjective))
+    {
+      Log::Warn << "Gradient Descent: converged to " << overallObjective 
+        << "; terminating" << " with failure.  Try a smaller step size?" 
+        << std::endl;
```
rcurtin: Continued lines should be intended two tabs from the previous line, not one :)

rcurtin:
Once you fix the constructor to `arma::mat` so it compiles I think this is ready to merge. :)

rcurtin:
Ah I forgot, we should also add something to `HISTORY.md` to point out that we now have a gradient descent optimizer. :)

PR #795, opened 2016-10-04
spelling

barak:
Gbp-Pq: Name 0003-spelling.patch


rcurtin:
Dang we are not very good at spelling.  Thanks!

barak:
Haint ne'er bin my strenth aither. Got meself an elictronic secretery.

Issue #773, opened 2016-08-29
User rvalue references to set a given reference tree in RASearch, RangeSearch, FastMKS, and DualTreeBoruvka


MarcosPividori:
Update the code to include changes similar to the ones included in the PR: https://github.com/mlpack/mlpack/pull/765

itabhiyanta:
Hi Marcos

I would like to take up this issue. I am reading #765 to understand the issue. I am completely new to mlpack and machine learning. I have gone through the getting involved page.
http://www.mlpack.org/involved.html
So I am at the point where it is suggested to jump in and contribute.
Do you have any update/alternate way in which I should approach this (other than going through #765)?
Kindly advise.
Rohit

akanuraj200:
@itabhiyanta 
Hello Rohit

I already have a PR (#799) in motion for the `FastMKS` and `DualTreeBoruvka` part of this issue. I had planned on updating `RASearch` and `RangeSearch` too as soon as i get some time, but you are free to try your hand at these two. 

#765 is a good place to start. You might want to take a look at all the discussion there. They are helpful in understanding what is being implemented. 

All the best :)

aneesh2312:
@akanuraj200 I would like to try RASearch and RangeSearch as I am new here, and would like to start with something labeled easy. Are you working on it? Or can I?

itabhiyanta:
Hi

I haven't started so you may pick it up.
Rohit

akanuraj200:
> @akanuraj200 I would like to try RASearch and RangeSearch as I am new here, and would like to start with something labeled easy. Are you working on it? Or can I

Hello @aneesh2312 
I am not working on `RASearch` and `RangeSearch` currently. You are free to try your hand at them.
All the best. :)

abhinavrai44:
Hi,
I would like to take this issue. Can I also try to solve the problem or is it allotted ?

aneesh2312:
@abhinavrai I am working on it..

Lakshya1605:
@aneesh2312 are you still working on it ?If no i would like to work on this issue .Thank You!!

aneesh2312:
@Lakshya1605 I have done RAsearch, but I never got time to finish RangeSearch.. You can take RangeSearch...

Lakshya1605:
@aneesh2312 Could you please give an overview of what is expected to be done .
Thank You !!!

Issue #756, opened 2016-08-06
Refactor `RSModel` and `RAModel` to parallel `NSModel` and use `boost::variant`


rcurtin:
In 86a9852, `NSModel` was refactored to clean up a lot of the duplicated code and instead use `boost::variant` plus the visitor paradigm.  We should also apply these changes to `RSModel` and `RAModel`.

A good place to start here would be to look at the changes in 86a9852, and apply them to the very similar classes `RSModel` for range searching and `RAModel` for rank-approximate nearest neighbor search.

`NSModel` can be found in `src/mlpack/methods/neighbor_search/ns_model.hpp` and `src/mlpack/methods/neighbor_search/ns_model_impl.hpp`.
`RSModel` can be found in `src/mlpack/methods/range_search/rs_model.hpp` and `src/mlpack/methods/range_search/rs_model_impl.hpp`.
`RAModel` can be found in `src/mlpack/methods/rann/ra_model.hpp` and `src/mlpack/methods/rann/ra_model_impl.hpp`.

This could be a good issue for someone new who is looking to contribute their first PR.

ferbncode:
Hi, i would like to work on this issue. Thanks :).

rishabhkumar296:
Hi, I would like to work on this issue.

rcurtin:
RSModel is done with #866, now only RAModel remains.

Trion129:
Can I work on the RAModel part?

rcurtin:
@Trion129: #889 is already open, so you may want to find something else so that you don't duplicate work.

shardulparab97:
Hi @rcurtin , I wanted to ask which part of the issue is yet pending, since RSModel is complete and issue #889 works on RAModel. I am interested in contributing to MLPack, can you please help me out. Thanks! :)

rcurtin:
@shardulparab97: sorry for the slow response.  You're right that both are done, so I think we can close this issue.  Thanks for pointing it out! :+1:  There are lots of other issues we've labeled `help wanted` or `good first issue`, so maybe you could find one of those that interests you and contribute to that issue?

Hope this helps!

PR #799, opened 2016-10-05
using rvalue references to set a given reference tree in FastMKS and DualTreeBoruvka

akanuraj200:
I have refactored `FastMKS` and `DualTreeBoruvka` to include #773 

I didn't include `RSModel` and `RAModel` in this because i think they can be updated together with #756 
I will do the needful if nobody else takes it up in the next couple of weeks. :)


src/mlpack/methods/fastmks/fastmks_impl.hpp
```diff
@@ -185,10 +223,60 @@ void FastMKS<KernelType, MatType, TreeType>::Train(Tree* tree)
...
+void FastMKS<KernelType, MatType, TreeType>::Train(Tree& tree)
+{
+  if (naive)
+    throw std::invalid_argument("cannot call FastMKS::Train() with a tree when "
+        "in naive search mode");
+
+  if (setOwner)
+    delete this->referenceSet;
+
+  this->referenceSet = &this->tree->Dataset();
```
rcurtin: What is `&this->tree`?  This probably does not compile... consider updating `Train(Tree*)` to simply call this method so that when you test `Train(Tree*)`, this is called too.  Also, the heavy use of `this->` in this function is not necessary.

src/mlpack/methods/fastmks/fastmks_impl.hpp
```diff
@@ -185,10 +223,60 @@ void FastMKS<KernelType, MatType, TreeType>::Train(Tree* tree)
...
+  if (setOwner)
+    delete this->referenceSet;
+
+  this->referenceTree = new Tree(std::move(tree));
+  this->referenceSet = &this->referenceTree->Dataset();
+  this->metric = metric::IPMetric<KernelType>(tree.Metric().Kernel());
+  this->setOwner = false;
+
+  if (treeOwner && this->referenceTree)
+    delete this->referenceTree;
```
rcurtin: This will delete the tree after it is moved!

akanuraj200:
@rcurtin thank you for your review :) 
I will fix it soon. I hope there is no urgency as this won't be a part of mlpack 2.1.0 (according to the milestone).

rcurtin:
Yeah, it is not a huge deal so no need to hurry with the changes.  It would be nice to include it in mlpack 2.1.0 but not completely necessary I don't think, so if it's ready I'll merge it and if not I won't wait. :)

rcurtin:
Hi Anuraj,

I saw your message in IRC:

> 12:37 < topology> why do travis and my local system give different results for the mlpack test suite?

Some tests have a random aspect, so it may be possible to get different results like that.  In the case of this PR, when I check the Travis output, it looks like there is a memory issue causing a segfault; often this will be a result of using some uninitialized pointer or something.  This may not occur on your local machine because the uninitialized value may just happen to be some value that does not cause a problem (like for instance, a pointer that is uninitialized has some chance of being NULL, which may not cause a crash).

Have you tried using valgrind to hunt down the source of the error?  Note that you can run just a single test with `mlpack_test -t FastMKSTest/FastMKSModelLinearTest`.

Hope this is helpful. :)

akanuraj200:
Hi Ryan,
I already tried running the `FastMKSModelLinearTest` exclusively on my system and it passed without any errors. So i guess you are right about the uninitialized value being treated as NULL on my local machine.
I haven't used Valgrind yet. I will try it out.

src/mlpack/methods/fastmks/fastmks_impl.hpp
```diff
@@ -174,21 +212,54 @@ template<typename KernelType,
                   typename TreeMatType> class TreeType>
 void FastMKS<KernelType, MatType, TreeType>::Train(Tree* tree)
 {
+  train(*tree);
```
akanuraj200: This should be Train(*tree)

src/mlpack/methods/fastmks/fastmks_impl.hpp
```diff
@@ -174,21 +212,54 @@ template<typename KernelType,
...
-  this->metric = metric::IPMetric<KernelType>(tree->Metric().Kernel());
-  this->setOwner = false;
+    delete referenceSet;
 
   if (treeOwner && referenceTree)
     delete referenceTree;
+    
+  referenceTree = new Tree(tree);    
+  referenceSet = &tree->Dataset();
+  metric = metric::IPMetric<KernelType>(tree->Metric().Kernel());
```
akanuraj200: `metric = metric::IPMetric<KernelType>(tree.Metric().Kernel());`

akanuraj200:
Hi @rcurtin 
I am unable to figure out a fix for the memory leak. The newly added `FastMKS<>::Train(Tree&)` and `FastMKS<>::Train(Tree&&)` functions are causing the trouble.

This is snippet of the Valgrind output for `FastMKSModelLinearTest`: 
```
==7259== 1 bytes in 1 blocks are definitely lost in loss record 1 of 5
==7259==    at 0x4C2B1C7: operator new(unsigned long) (in /usr/lib/valgrind/vgpreload_memcheck-amd64-linux.so)
==7259==    by 0xBF01BB: mlpack::fastmks::FastMKS<mlpack::kernel::LinearKernel, arma::Mat<double>, mlpack::tree::StandardCoverTree>::FastMKS(bool, bool) (in /home/kanodia/mlpack/build/bin/mlpack_test)
==7259==    by 0xBF1BCF: void mlpack::fastmks::FastMKSModel::BuildModel<mlpack::kernel::LinearKernel>(arma::Mat<double> const&, mlpack::kernel::LinearKernel&, bool, bool, double) (in /home/kanodia/mlpack/build/bin/mlpack_test)
==7259==    by 0xBBFBBC: FastMKSTest::FastMKSModelLinearTest::test_method() (in /home/kanodia/mlpack/build/bin/mlpack_test)
==7259==    by 0xBC1B0D: FastMKSTest::FastMKSModelLinearTest_invoker() (in /home/kanodia/mlpack/build/bin/mlpack_test)
==7259==    by 0xA05C56: boost::unit_test::ut_detail::callback0_impl_t<boost::unit_test::ut_detail::unused, void (*)()>::invoke() (in /home/kanodia/mlpack/build/bin/mlpack_test)
==7259==    by 0x58CC520: ??? (in /usr/lib/x86_64-linux-gnu/libboost_unit_test_framework.so.1.55.0)
==7259==    by 0x58A863D: boost::execution_monitor::catch_signals(boost::unit_test::callback0<int> const&) (in /usr/lib/x86_64-linux-gnu/libboost_unit_test_framework.so.1.55.0)
==7259==    by 0x58A8F3A: boost::execution_monitor::execute(boost::unit_test::callback0<int> const&) (in /usr/lib/x86_64-linux-gnu/libboost_unit_test_framework.so.1.55.0)
==7259==    by 0x58CC693: boost::unit_test::unit_test_monitor_t::execute_and_translate(boost::unit_test::test_case const&) (in /usr/lib/x86_64-linux-gnu/libboost_unit_test_framework.so.1.55.0)
==7259==    by 0x58B7600: boost::unit_test::framework_impl::visit(boost::unit_test::test_case const&) (in /usr/lib/x86_64-linux-gnu/libboost_unit_test_framework.so.1.55.0)
==7259==    by 0x58E9E82: boost::unit_test::traverse_test_tree(boost::unit_test::test_suite const&, boost::unit_test::test_tree_visitor&) (in /usr/lib/x86_64-linux-gnu/libboost_unit_test_framework.so.1.55.0)
```

```
==7259== LEAK SUMMARY:
==7259==    definitely lost: 3 bytes in 3 blocks
==7259==    indirectly lost: 0 bytes in 0 blocks
==7259==      possibly lost: 0 bytes in 0 blocks
==7259==    still reachable: 72,712 bytes in 2 blocks
==7259==         suppressed: 0 bytes in 0 blocks
==7259== Reachable blocks (those to which a pointer was found) are not shown.
==7259== To see them, rerun with: --leak-check=full --show-reachable=yes
==7259== 
==7259== For counts of detected and suppressed errors, rerun with: -v
==7259== ERROR SUMMARY: 3 errors from 3 contexts (suppressed: 2 from 2)
```
Do take a look whenever you find time.
Thank you

src/mlpack/methods/fastmks/fastmks_impl.hpp
```diff
@@ -174,21 +212,54 @@ template<typename KernelType,
...
-  this->referenceSet = &tree->Dataset();
-  this->metric = metric::IPMetric<KernelType>(tree->Metric().Kernel());
-  this->setOwner = false;
+    delete referenceSet;
 
   if (treeOwner && referenceTree)
     delete referenceTree;
+    
+  referenceTree = new Tree(tree);    
+  referenceSet = &tree->Dataset();
```
rcurtin: I bet the problem is in this line.  Shouldn't this be `referenceSet = &referenceTree->Dataset()`?  Otherwise you are setting the pointer to the dataset of the tree that the user passed in, which isn't guaranteed to be a valid pointer outside the scope of this function.

src/mlpack/methods/fastmks/fastmks_model_impl.hpp
```diff
@@ -34,8 +34,7 @@ void BuildFastMKSModel(FastMKS<KernelType>& f,
     // Create the tree with the specified base.
     Timer::Start("tree_building");
     metric::IPMetric<KernelType> metric(k);
-    typename FastMKS<KernelType>::Tree* tree =
-        new typename FastMKS<KernelType>::Tree(referenceData, metric, base);
+    typename FastMKS<KernelType>::Tree tree(referenceData, metric, base);
     Timer::Stop("tree_building");
 
     f.Train(tree);
```
rcurtin: We should use `std::move(tree)` here to avoid copying the entire tree---that will be a big speed improvement.

src/mlpack/methods/fastmks/fastmks.hpp
```diff
@@ -141,6 +184,29 @@ class FastMKS
    * @param tree Tree to use as reference data.
```
rcurtin: To mark a method deprecated, also add `mlpack_deprecated` to the function definition please.  i.e. `mlpack_deprecated void Train(...)`.

src/mlpack/methods/fastmks/fastmks.hpp
```diff
@@ -143,9 +187,35 @@ class FastMKS
...
+   * Train the FastMKS model on the given reference tree.  This takes ownership
+   * of the tree, so you do not need to delete it!  This will throw an exception
+   * if the model is searching in naive mode (i.e. if Naive() == true).
+   *
+   * This method will copy the given tree. You can avoid this copy by using the
+   * Train() method that takes a rvalue reference to the tree.
+   *
+   * @param tree Tree to use as reference data.
+   */
+  void Train(Tree& referenceTree);
```
rcurtin: We can make this `const Tree&` since we are copying the tree.  Also, the comment here is inaccurate---this method does not take ownership of the tree, it copies it.  This comment also applies to other places with arguments like `Tree&` that can be changed to `const Tree&`.

src/mlpack/methods/fastmks/fastmks.hpp
```diff
@@ -200,6 +273,33 @@ class FastMKS
...
+   * Be aware that if your tree modifies the original input matrix, the results
+   * here are with respect to the modified input matrix (that is,
+   * queryTree->Dataset()).
+   *
+   * @param queryTree Tree built on query points.
+   * @param k The number of maximum kernels to find.
+   * @param indices Matrix to store resulting indices of max-kernel search in.
+   * @param kernels Matrix to store resulting max-kernel values in.
+   */
+  void Search(Tree& querySet,
```
rcurtin: Parameter should be called `queryTree` not `querySet`.

src/mlpack/methods/fastmks/fastmks_impl.hpp
```diff
@@ -301,6 +372,20 @@ void FastMKS<KernelType, MatType, TreeType>::Search(
     arma::Mat<size_t>& indices,
     arma::mat& kernels)
 {
+	Search(*queryTree, k, indices, kernels);
```
rcurtin: The spacing is off here, the call to `Search()` should be indented only two spaces from the opening brackets.

src/mlpack/methods/fastmks/fastmks_impl.hpp
```diff
@@ -420,7 +505,7 @@ void FastMKS<KernelType, MatType, TreeType>::Search(
   // Dual-tree implementation.
   Timer::Stop("computing_products");
 
-  Search(referenceTree, k, indices, kernels);
+  Search(std::move(referenceTree), k, indices, kernels);
```
rcurtin: There's no need for a move here, this will destroy the model's tree after the call.  Passing `referenceTree` as a `const Tree&` is just fine.

akanuraj200:
> To mark a method deprecated, also add mlpack_deprecated to the function definition please. i.e. mlpack_deprecated void Train(...).

Hi Ryan,
Adding `mlpack_deprecated` to `FastMKS<>::Search()` and `DualTreeBoruvka<>::CleanupHelper()` results in too many warning. Do you want to keep the flag anyway?

rcurtin:
Yes, we should use the `mlpack_deprecated` flag because the functions are deprecated.  You will have to update the rest of the codebase to use the non-deprecated functions to prevent the warnings.

src/mlpack/methods/emst/dtb_impl.hpp
```diff
@@ -103,6 +103,56 @@ template<
...
+DualTreeBoruvka<MetricType, MatType, TreeType>::DualTreeBoruvka(
+    Tree& tree,
+    const MetricType metric) :
+    tree(new Tree(tree)),
+    data(&this->tree->Dataset()),
+    ownTree(true),
+    naive(false),
+    connections(data.n_cols),
+    totalDist(0.0),
+    metric(metric)
```
akanuraj200: Hi @rcurtin ,

```
error: invalid initialization of reference of type ‘const arma::Mat<double>&’ from expression of type ‘arma::Mat<double>*’
     metric(metric)
                  ^
```
The `NeighborSearch` class is implemented in a similar way, but that runs without an error.  Any thoughts?

rcurtin: The problem, I believe, is the line `data(&this->tree->Dataset())`.  `this->tree->Dataset()` is already a reference, so there's no need to take the address of it.

rcurtin:
The issue in `dtb_impl.hpp` highlights that there is actually no test for that method.  I wrote up a quick test and ran it in master.  You'll have to adapt the test slightly to your changes, but can you add it to `emst_test.cpp`?

```
/**
 * Test passing a pre-created tree and make sure we get the same results.
 */
BOOST_AUTO_TEST_CASE(CopyTreeTest)
{
  arma::mat dataset = arma::randu<arma::mat>(5, 500);

  DualTreeBoruvka<> copyDataMST(dataset);

  // Build a tree by hand.  The tree-building process rearranges the dataset, so
  // we'll build another DTB object on the modified dataset.
  DualTreeBoruvka<>::Tree tree(std::move(dataset));
  DualTreeBoruvka<> copyTreeMST(&tree);

  DualTreeBoruvka<> copyDataMST(tree->Dataset());

  // Now get the results.
  arma::mat copyDataResults, copyTreeResults;
  copyDataMST.ComputeMST(copyDataResults);
  copyTreeMST.ComputeMST(copyTreeResults);

  BOOST_REQUIRE_EQUAL(copyDataResults.n_rows, copyTreeResults.n_rows);
  BOOST_REQUIRE_EQUAL(copyDataResults.n_cols, copyTreeResults.n_cols);
  for (size_t i = 0; i < copyDataResults.n_cols; ++i)
  {
    BOOST_REQUIRE_EQUAL(copyDataResults(0, i), copyTreeResults(0, i));
    BOOST_REQUIRE_EQUAL(copyDataResults(1, i), copyTreeResults(1, i));
    BOOST_REQUIRE_CLOSE(copyDataResults(2, i), copyTreeResults(2, i), 1e-5);
  }
}
```

I see that there are still a few warnings in the travis build from using deprecated constructors.  Can you take a look at the travis build and fix those please?

I realize also now, that this PR adds a handful of methods, but they are not tested:

 * `DualTreeBoruvka::DualTreeBoruvka(const Tree&)`
 * `DualTreeBoruvka::DualTreeBoruvka(Tree&&)`
 * `FastMKS::FastMKS(const Tree&)`
 * `FastMKS::FastMKS(Tree&&)`

We should definitely add tests for those.  Are you willing to implement those?  There are other tests in the codebase that are similar, and I've already provided what can work for the first of those tests, which can be pretty readily adapted to test `FastMKS::FastMKS(const Tree&)`.  If you like, I can implement the rest of the tests (I'm pretty busy so it may be a little while), or you can take a stab at it.  Let me know what you think.

akanuraj200:
Hi Ryan, 
I definitely want to implement the required tests for the new functions added. This will only help me get more comfortable with the codebase. 
I will start working on it in a week's time(I am on a holiday right now). 

I will go through your suggestions in detail once I get back. 

Thank you!

rcurtin:
No hurry, enjoy your holiday. :)

rcurtin:
I'm going to close this for inactivity but if you (or someone else) has the chance to take care of the tests then we can reopen and merge it.

abernauer:
Hey @rcurtin I am open to writing the tests, so the changes can get merged in the code base.

rcurtin:
@abernauer sure, feel free.  You'll probably have to adapt the code a bit; this was last modified about 3 years ago, so it may be out of date. :)

abernauer:
I could open a new PR and just use this as a reference if that approach would be preferred.

rcurtin:
Yeah, that's probably the easiest thing to do here. :+1:

PR #524, opened 2016-02-25
moving cliDeleter instance to log.cpp

theSundayProgrammer:
The sequence of deletion has to be different to the reverse of sequence of creation. Hence cliDeleter is necessary.

PR #800, opened 2016-10-11
Put all singletons in libmlpack.so

rcurtin:
This is a solution for #524.  It actually functions about the same as that solution but creates a separate file for singletons (`src/mlpack/core/singletons.cpp`).  As a result of this change it is possible to create a program that has `#include <mlpack/core.hpp>` without linking against libmlpack.so, if no mlpack symbols are used.  This was not possible before.

I'll leave this for a couple of days before I merge it.

@theSundayProgrammer: I know this took a very long time for me to get around to, but here is (finally) a solution.  Like I said it is essentially equivalent to yours, just arranged differently.


theSundayProgrammer:
Nice! Thank You Ray.

PR #802, opened 2016-10-18
Density Estimation Tree made sparse-enabled

thejonan:
DET class can now work with Arma's sparse matrices. Also:
- The class is templated, so the matrix type becomes template argument.
- Couple of optimizations regarding matrix copying, incl. in dense matrix case.
- OpenMP-ification of tree building routine.
- Added few test cases with `arma::SpMat` data.


rcurtin:
Thanks for your hard work and contribution.  I'll try and give this a good review in the next day or two.

Do you want to add yourself to the list of authors in `src/mlpack/core.hpp` and `COPYRIGHT.txt`?

thejonan:
Yes (for contribution list), is it Ok to do it in this same pull request?
And, yes - I'll check why is this integration test happening... (I did build and run everything before submitting).

src/mlpack/methods/det/dtree_impl.hpp
```diff
@@ -248,31 +362,32 @@ bool DTree::FindSplit(const arma::mat& data,
...
-      // estimate.
-      minError = actualMinDimError;
-      splitDim = dim;
-      splitValue = dimSplitValue;
-      leftError = std::log(dimLeftError) - 2 * std::log((double) data.n_cols)
-          - volumeWithoutDim;
-      rightError = std::log(dimRightError) - 2 * std::log((double) data.n_cols)
-          - volumeWithoutDim;
-      splitFound = true;
+      {
```
rcurtin: Why the extra braces?

src/mlpack/methods/det/dtree_impl.hpp
```diff
@@ -197,44 +308,47 @@ bool DTree::FindSplit(const arma::mat& data,
...
+    //   dimVec = data.row(dim).subvec(start, end - 1);
+    //   dimVec = arma::sort(dimVec);
+    // could be quite inefficient for sparse matrices, due to copy operations (3).
+    // This one has custom implementation for dense and sparse matrices.
+
+    std::vector<SplitItem> splitVec = details::ExtractSplits(data,
+                                                             dim,
+                                                             start,
+                                                             end,
+                                                             minLeafSize);
```
rcurtin: As far as I can tell the reason for the `ExtractSplits` function is because the `sort()` method is not available for sparse matrices.  Suppose that `sort()` did exist for sparse matrices (e.g. suppose I sat down and wrote it, which I might need to do shortly!).  Then we could do this...

```
typename MatType::row_type dimVec = data.row(dim).subvec(start, end - 1);
dimVec = arma::sort(dimVec);

// Iterate over all possible values.
typename MatType::row_type::const_row_col_iterator it;
for (it = dimVec.begin_row_col(); ++it; it != dimVec.end_row_col())
{
  // Check the split to the left side of the point that *it represents, if it exists.
  if (it->col() > 0)
  {
    // do checking for split between dimVec[it->col() - 1] and dimVec[it->col()]...
  }

  // If we are in the next-to-last position, check the split to the right, if applicable.
  // There's probably a cleaner way to write this code.
  typename MatType::row_type::const_row_col_iterator it2 = it;
  if ((++it2) == dimVec.end_row_col())
  {
    // do checking for split between dimVec[it->col()] and dimVec[it->col() + 1]...
  }
}
```

Note that the `row_col_iterator` will only "stop" at points that are actually represented in memory.  So for sparse matrices it will skip over zero elements.  I think that the `row_col_iterator` is not actually documented by Armadillo... I should submit a patch for that...

What do you think?  Would this approach work?  If so I will implement `SpMat::sort()` (it should be pretty straightforward, I think I have a good idea).  That would allow us to avoid having specific code for both the dense and sparse case (I like to push specific code like that to Armadillo wherever possible).

src/mlpack/methods/det/CMakeLists.txt
```diff
@@ -4,11 +4,7 @@
 set(SOURCES
   # the DET class
   dtree.hpp
-  dtree.cpp
-
-  # the util file
-  dt_utils.hpp
-  dt_utils.cpp
+  dtree_impl.hpp
```
rcurtin: Shouldn't we include `dt_utils.hpp` and `dt_utils_impl.hpp` too?  I overlooked this in my review :)

thejonan: I've decided to remove them, because I'm not sure they should actually be part of the library itself. Aren't they mostly related to the command-line tool?

thejonan:
One general comment - as can be seen - I'm making some performance tests with real data and it turns the sparse matrix performance is _tremendously_ worse - specifically in extracting the values to be sorted and then checked. I'm continuing my experiments but, I guess this should not stop the PR merge.

thejonan:
I can't see here your sparse matrix sorting comment. But I see it in my email :-) So here is the answer:
- I did already implemented SpMat sort - it is in another branch, but I haven't made tests, so I didn't make pull request.
- The code in ExtractSplits is pretty close to that sorting, but with removed parts that we're not interested in.

Generally, sparse matrices will have large number of zeroes, which we're not interested in for our splits - we just need to one whether there is none or any. Also arma's sort is not in-place, which is a performance penalty (I've checked the arma's dense matrix sort - it is still based on STD's one :-). And it is perfectly fine that row_col_iterator() jumps over the zeroes, because (1) we don't specifically need them, and (2) even in a sorted version of a sparse row, they will be all together.

rcurtin:
> One general comment - as can be seen - I'm making some performance tests with real data and it turns the sparse matrix performance is tremendously worse - specifically in extracting the values to be sorted and then checked. I'm continuing my experiments but, I guess this should not stop the PR merge.

I am not surprised... the CSC sparse matrix format turns out to only be useful for linear algebra operations in extremely sparse matrices.  In the future I think Armadillo should replace its implementation of CSC with coordinate list format sparse matrices... since I wrote the CSC support, this task should fall to me, but I have not done it yet... :)

src/mlpack/methods/det/dt_utils_impl.hpp
```diff
@@ -148,30 +150,27 @@ DTree* mlpack::det::Trainer(arma::mat& dataset,
...
     prunedSequence.push_back(treeSeq);
     oldAlpha = alpha;
     alpha = dtree.PruneAndUpdate(oldAlpha, dataset.n_cols, useVolumeReg);
 
     // Some sanity checks.  It seems that on some datasets, the error does not
     // increase as the tree is pruned but instead stays the same---hence the
     // "<=" in the final assert.
-    Log::Assert((alpha < std::numeric_limits<double>::max()) ||
-        (dtree.SubtreeLeaves() == 1));
+    Log::Assert((alpha < std::numeric_limits<double>::max()) || (dtree.SubtreeLeaves() == 1));
```
rcurtin: Yeah---the reason for 80-character lines is in part because it makes much easier to print, or include in a paper, or something like this.  Helpful especially for manuals or PDFs generated by Doxygen, etc..

rcurtin:
> - I did already implemented SpMat sort - it is in another branch, but I haven't made tests, so I didn't make pull request.
> - The code in ExtractSplits is pretty close to that sorting, but with removed parts that we're not interested in.

That's really nice, would you like to put it in arma_extend.hpp?  I can help with that and also help write tests for it and then it can be submitted upstream.  For the code in `ExtractSplits()`, is it noticeably faster than the sparse matrix splitting code you wrote?  If the difference is very minimal I think it would be better to go with a simple strategy that does not need special code to handle sparse matrices---the `row_col_iterator` approach I proposed is nice for this reason: it is generic.

thejonan:
The problem with CSC in this task is that we want in-row iteration and we _need_ it to be of complexity O(1), and with CSC it is not. In the same time we need _swap_cols_ operation, which would be much slower (I guess) if we work on the transposed matrix. If we have dual indices, so we can "walk" on both direction fast - that would be fine. And, generally - the matrix that I'm trying with is pretty sparse as well - approx. 99.995% is empty. 

The benefit of `ExtractSplit` is that for dense matrices it makes in-place sorting, which is not the case with `arma::sort`. Another thing is that even with `row_col_iterator` you'll need to know you're in sparse matrix - because if the first entry you get is not in the beginning you need to _fake_ a previous value of zero, to get a split between them. Also if you happen to jump over indices (with sorted sparse rows/cols it is good that all emptiness is in a single slot, so you have one jump at most), etc. So split-attempts won't be generic. That's why I've chosen to have split extraction points custom, and then iteration over them - generic. I'm a big enemy of duplicated code too! :-) Also, I've tried avoiding `submat` call and directly iterating over the given region to extract splits - it turned to be _way_ much slower for sparse matrices. So now the most expensive step is `submat`.

For sparse sorting - isn't it best if you make another branch here, in the master repo and I make a pull request for `SpMat` sort to that branch, rather then to `master`? Or, you can get it directly from my repo - it is in `feature/sparse_sort` branch. Whichever is easier for you.

And finally - the Travis CI is failing on some other tests - not DET-related :-)

thejonan:
Btw, a _very_ important step would be sparse matrix file reading... I've made some workaround for my tests, and even have an idea how to gracefully plug into existing loading paradigm - with variants :-) `boost::variant` or `std::variant`. But that's a quite an effort, and - more importantly - a change on many places, so I've avoided undertaking the task.

rcurtin:
Don't worry about the Travis failure, that seems to be some other test where the tolerance needs adjusting.

The impression I am getting---and correct me if I am wrong---is that you agree it would be nice to have only one implementation of `ExtractSplits()` that had, e.g., `MatType` as a template parameter, but this is not possible due to the current drawbacks of the Armadillo sparse matrix implementation.  I agree that there are definitely problems with the Armadillo CSC implementation so for now I think it's reasonable to have two versions of `ExtractSplits()`, but I think in the long run we should aim in the following directions:

 * reimplement `SpMat` as coordinate list, because really I have not found any cases where CSC gives significant speedup.  Coordinate lists also mean that a lot of non-linear-algebraic operations like sorting, iterating, and so forth become a million times easier.  This task is something I will do, I am just not sure when I'll have the time.

 * Get better sparse matrix file reading in place (and submitted upstream to Armadillo when possible).  I know that Armadillo has coordinate list file loading support, but there isn't a way to use that currently with any of the mlpack command-line programs.  Maybe that is worth looking into too.  What formats are you thinking of?  The more I think of it the following *should* work:

```
arma::sp_mat matrix;
data::Load("coordinate_list.csv", matrix);
```

 * Take your `sort()` implementation and submit it upstream to Armadillo.  Would you like to do this?  I can help.  Basically we'll just need to write tests for it then send it to Conrad and he's likely to take a look over and then include it.

Are you finished with this PR otherwise?  If so I'll go ahead and merge it.  Thanks again for your time on this, it is a definite improvement.

thejonan:
Yes, I'm done with the PR - feel free to merge it, whenever is ok for you.

I wish I could agree on the `ExtractSort` issue, but there are sparse-specific problems, in short - `row_col_iterator` will not report the zero values, but the split decision algorithm does need them - to make a split attempt between a zero and the adjacent non-zero element. The all-universal implementation with sparse enabled `arma::sort` and then index-based iteration will be correct, but terribly slow because of all the zeroes, that will be reported.

My implementation on `arma::sort` is as part of MLPack's arma extensions. It is a less-than-a-page source code, which I can even paste into a comment, I guess :-) However, if it's really better to go the PR to Armadillo way - I can do it, I just don't know exactly when.

Regarding the coordinate list sparse matrix implementation - isn't the CSC storage faster when it comes to column-based iteration? Anything that is fast on both rows and columns directed iteration will be fine.

Regarding the format loading - I'd say that the format to be supported is Matrix Market, which is a coordinate list with the header. Such format will also save one swipe through the input file, which is happening now - for obtaining the matrix size, and which technically is not even quite correct, because if you have zeroes at the last columns/rows you will not have entries for these in the coordinate list and will deduce a false matrix size.

rcurtin:
Ok, great---I am merging a few other things that needed manual intervention today, but I will get to this one when I am done with those.

> I wish I could agree on the ExtractSort issue, but there are sparse-specific problems, in short - row_col_iterator will not report the zero values, but the split decision algorithm does need them - to make a split attempt between a zero and the adjacent non-zero element. The all-universal implementation with sparse enabled arma::sort and then index-based iteration will be correct, but terribly slow because of all the zeroes, that will be reported.

Ah, you are right, I had overlooked that.  I think it is possible to get _pretty_ close to completely generic but I agree now, it is not completely possible.

> My implementation on arma::sort is as part of MLPack's arma extensions. It is a less-than-a-page source code, which I can even paste into a comment, I guess :-) However, if it's really better to go the PR to Armadillo way - I can do it, I just don't know exactly when.

Hm, here is an idea, what do you think?  You can submit a PR to add it to `arma_extend.hpp` and then add tests for it in `src/mlpack/tests/`, and then I'll extract the code and submit it upstream to Armadillo and CC you on the email (since Armadillo is not on Github).  Then when that gets accepted I'll put `#if` guards around the `arma_extend` implementation to make sure that the method is only included when the Armadillo version doesn't already have it.

> Regarding the coordinate list sparse matrix implementation - isn't the CSC storage faster when it comes to column-based iteration? Anything that is fast on both rows and columns directed iteration will be fine.

For iteration, it's a toss-up.  For CSC you have to maintain a variable holding which column you are in as well as which index you are at in the row indices and values vectors, and when you switch columns it can take a few iterations to get the correct new column ID.  For coordinate lists you always have the column available and only need to maintain the index that you are at.

As for the coordinate list implementation, it may be many weeks before I actually have that ready to go.  My current priorities aren't totally pointed towards this, but it is definitely a priority in the long run to get better, faster, and more usable sparse matrix support for mlpack (which means contributions to Armadillo).

> Regarding the format loading - I'd say that the format to be supported is Matrix Market, which is a coordinate list with the header. Such format will also save one swipe through the input file, which is happening now - for obtaining the matrix size, and which technically is not even quite correct, because if you have zeroes at the last columns/rows you will not have entries for these in the coordinate list and will deduce a false matrix size.

Ah, that should not be hard.  You can take `diskio::load_coord_ascii()` and `diskio::save_coord_ascii()` and adapt them easily to produce methods like `diskio::load_mm()` and `diskio::save_mm()` and we could take the same approach I outlined for the `sort()` method to include it in mlpack and then also push it upstream to Armadillo.  Let me know what you think.

thejonan:
Ok, I'll make a PR for sparse's sort as it is now - I'll try to add some tests shortly and you can choose when to get it.

Just to comment on sparse matrix loading - I'm sure it's easy to extend the Arma's `diskio` functions (`save_coord_ascii` can be used directly, but `load_coord_ascii` will need splitting - the size deduction part from actual reading). 

The more interesting question, I'd say, is how MLPack handles different input via `data::Load`, depending on the filename - pretty much the way it does not - but with the ability of initializing a sparse matrix, when that type of file is provided. This runtime flexibility is not very template-friendly, but my idea was to use variants. Something like `typedef boost::variant<arma::mat, arma::spmat> Dataset`, i.e. an encapsulation of different matrix types. There is a runtime mechanism to query what exactly type is the matrix read. I haven't implemented this, however - just sharing the ideas.

rcurtin:
Merge done, thanks for the contribution!  I like the idea about making it easier to use either dense or sparse types depending on the given filename.  I think we will need to think about it additionally though, so maybe it is worth opening a ticket for or something.  For instance, if I pass in "coordinate_list.csv", I can either load it as a 3xN dense matrix or an RxC sparse matrix... I can't autodetect which is right.  In the context of a C++ program, it is probably better to let the user select what they want by using either `data::Load<arma::mat>()` or `data::Load<arma::sp_mat>()`.  But for the command-line programs, it would be nice if the user had a way to specify whether their matrix was sparse or dense and use the correct type then (`boost::variant` could be used there, or just some other template solution).  I wonder if maybe it is easier to allow the user to specify file type by appending `:sparse` or `:dense` to their filename, like `-i coordinate_list.csv:sparse`.  Anyway, maybe this is discussion for another place.  I am open to ideas as long as they are clean, as always. :)

thejonan:
Ok, thanks! The loading discussion definitely deserves a dedicated ticket. I've made a PR for my sparse sorting implementation - #808.

PR #808, opened 2016-11-01
Arma sorting work with SpMat.

thejonan:
A sparse matrix sort capability, added as part of MLPack's extensions to Armadillo lib.

**Note**: There are not tests added, also - custom type of sorting, etc. Just the core sorting function.

### DIFF
--- a/src/mlpack/core/arma_extend/CMakeLists.txt
+++ b/src/mlpack/core/arma_extend/CMakeLists.txt
@@ -4,6 +4,7 @@ set(SOURCES
   arma_extend.hpp
   fn_ccov.hpp
   fn_ind2sub.hpp
+  fn_sort_sparse.hpp
   glue_ccov_meat.hpp
   glue_ccov_proto.hpp
   hdf5_misc.hpp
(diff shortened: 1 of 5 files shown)


rcurtin:
Okay, I've opened a PR on your branch (https://github.com/thejonan/mlpack/pull/1) to add some tests to this.  Sorry for the slowness on this, it took a while to find the time.

Before merging this upstream with Armadillo, could I ask you to add a version of sort() that takes in an input matrix and a dimension on which to sort (either column-wise or row-wise), just like arma::sort() for dense matrices?

http://arma.sourceforge.net/docs.html#sort

I already have some simple tests written for that, so you can just uncomment those.  After that this will be ready to merge and then I can make the minor changes necessary to submit a patch upstream (I'll CC you on the email when that gets done).

Thanks!

thejonan:
I knew, that per-dimension sorting will come into discussion :-) I'll take a look at this, but don't promise that I'll be fast.

rcurtin:
No hurry, whenever you have time is fine. :)

PR #814, opened 2016-11-20
switch last (!) remaining std::cout to destination

eddelbuettel:
so I looked into why R was complaining that `std::cout` was referenced when linking against mlpack -- and it turns out that the (sole !!) remaining use of `std::cout` appears to be an instance that ought to have been converted to `destination`

this PR does that

no Travis build on my side; I just used to have a clone rather than fork and did not have Travis set up to build when I just committed this but it should work out just fine


rcurtin:
I agree, this is an error, thanks for the patch. :)

PR #816, opened 2016-11-23
Update knn_main.cpp

goalslee:
the type of "query" is arma::mat not string! it cause the runtime error


rcurtin:
Hi there,

Thank you very much for pointing this out!  I will release mlpack 2.1.1 as a bugfix immediately.
If you would like, I can add your name to the list of contributors.  Let me know.

Thanks,

Ryan

rcurtin:
Oops, sorry, I misspoke---I thought this was a bug in mlpack 2.1.0 but this code is only in master, not in 2.1.0, so there is no need for a sudden release.

PR #820, opened 2016-11-30
Modify Softmax regression to have numClasses - 1 weights

sumedhghaisas:


rcurtin:
Just for the sake of annotation, the issue here is that this breaks the softmax regression tests for some reason:

```
$ bin/mlpack_test -t SoftmaxRegressionTest
Running 2 test cases...
/home/ryan/src/mlpack-sumedh/src/mlpack/tests/softmax_regression_test.cpp(252): fatal error in "SoftmaxRegressionTwoClasses": difference{40.8%} between acc{59.200000000000003} and 100.0{100} exceeds 0.5%

```

If we can resolve that problem this is a nice improvement for the softmax regression implementation.

I haven't had a chance to really dig in to what is wrong here yet, just to reproduce the problem.

rcurtin:
I think both of us have run out of time for this so I'll go ahead and close it (also since it's got some conflicts).  @sumedhghaisas if you ever have time to look into this one again we can do that. :)

Issue #817, opened 2016-11-23
Implement DTree copy constructor, move constructor, and copy/move assignment operators


rcurtin:
The `mlpack::det::DTree` class doesn't implement a copy constructor, move constructor, or copy/move assignment operators, meaning that if I do this:

```
DTree a = std::move(b); // b is another DTree
```

the child pointers of `b` aren't properly updated.  So, these constructors and assignment operators need to be added.

`DTree` can be found in `src/mlpack/methods/det/dtree.hpp`.

rcurtin:
Fixed in #826 and #824.

PR #824, opened 2016-12-03
Copy, move constructors and copy assignment operator added

piyush-jaiswal:
For issue #817 
Function declarations are in dtree.hpp
Function bodies are in dtree_impl.hpp

Awaiting your response and suggestions. :)

### DIFF
--- a/src/mlpack/methods/det/dtree.hpp
+++ b/src/mlpack/methods/det/dtree.hpp
@@ -52,7 +52,7 @@ class DTree
   typedef typename MatType::elem_type     ElemType;
   typedef typename MatType::vec_type      VecType;
   typedef typename arma::Col<ElemType>    StatType;
-  
+
   /**
    * Create an empty density estimation tree.
    */
@@ -66,7 +66,7 @@ class DTree
    /**
    * Move a density estimation tree into another.
    */
-  DTree(const DTree&& toBeMoved);
+  DTree(DTree&& toBeMoved);
 
   /**
    * Move a density estimation tree into another using copy assignment operator.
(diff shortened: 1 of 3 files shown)


rcurtin:
Hi there Piyush,

Thanks for the contribution.  Before we merge this, could I ask you to add some tests?  Take a look at the tests I wrote for the LSHSearch class's copy/move constructors and assignment operators:

https://github.com/rcurtin/mlpack/blob/bindings/src/mlpack/tests/lsh_test.cpp#L839

Thanks,

Ryan

rcurtin:
I think there is still a bug here.  Can you change your tests so that you construct the DTree on the heap with the 'new' operator?  Then after you make the copy with the copy constructor, delete the first DTree. After that, check the values of the copied DTree.  This will ensure that we are not using memory that the first DTree allocated, and that we have a safe copy.  Let me know if I can explain this further.

piyush-jaiswal:
Hey Ryan,
I built the DTree using the 'new' operator on the heap as you asked. Kindly provide any more changes if needed.

rcurtin:
There are some issues here---the tests do not actually catch the fact that the tree is not properly copied or moved.  There is another PR open for this that is closer to working, I think maybe the right solution is to use that and then merge the tests from this PR too.

piyush-jaiswal:
Sure, no problem.

rcurtin:
Hi there Piyush, I merged some of your changes in the tests after merging #826 (specifically in 47a0a057 and a66a604c).  Thanks for your work on this PR.  I can add your name to the list of contributors---let me know if you want me to do that.

PR #825, opened 2016-12-06
Automatic bindings step 3: model parameters

rcurtin:
This pull request is a lot of work that revamps the CLI system pretty thoroughly.  It's been the majority of what I've been working on recently, which is probably why it seems like my activity has been low. :)  General highlights:

 * Now we can use `PARAM_MODEL_IN()` and `PARAM_MODEL_OUT()` instead of `PARAM_STRING_IN("input_model_file", ...)` and `PARAM_STRING_OUT("output_model_file", ...)`.
 * `CLI::Destroy()` is now required at the end of programs that save model-type parameters.  This won't be a problem because in the long run the call to `CLI::Destroy()` will probably become automatically generated.
 * The way parameters of different types are dealt with by `CLI` is greatly simplified and split into many different files with standalone functions using SFINAE to achieve different behavior for different types.
 * All command-line programs except `mlpack_rmva` now use model parameters where applicable.  I would have done `mlpack_rmva` but @zoq suggested that it's about to be heavily revamped anyway.
 * No command-line programs except `mlpack_rmva` and the preprocessing programs use string options to represent files, meaning that the options at the top of a program are now basically abstracted away from files.  That will allow me to start writing code to turn these definitions into coherent bindings to other languages.

There is just a little more to do:

 * Fix failing CLI test.
 * Add tests for HoeffdingTreeModel.
 * Rebase on current master.

But the PR is still reviewable as-is if anyone would like to.  I'll let this sit for a week or two after I push those last few changes then merge it in and continue onwards.  My goal (or rather my company's goal for me) is completed bindings for at least Python by January 1st, so... I'm doing my best to make that happen!


src/mlpack/core/util/default_param_impl.hpp
```diff
@@ -0,0 +1,108 @@
...
+  // Get the filename and return it, or return an empty string.
+  const std::string& filename = boost::any_cast<std::string>(data.value);
+  return "'" + filename + "'";
+}
+
+/**
+ * Return the default value of a string option.
+ */
+template<typename T>
+std::string DefaultParamImpl(
```
zoq: I'm curious do you have to implemented a version for HasSerialize, arma::mat and std::is_same<T, std::string>, it looks like they all do the same.

src/mlpack/core/util/output_param_impl.hpp
```diff
@@ -0,0 +1,94 @@
...
+//! Output a mapped dataset.
+template<typename T>
+void OutputParamImpl(
+    const ParamData& data,
+    const typename boost::enable_if<std::is_same<T,
+        std::tuple<data::DatasetInfo, arma::mat>>>::type* /* junk */)
+{
+  // Output the matrix with the mappings.
+//  const data::DatasetInfo& info =
+//      std::get<0>(*boost::any_cast<T>(&data.mappedValue));
```
zoq: Looks like unused code.

src/mlpack/core/util/param_data_impl.hpp
```diff
@@ -0,0 +1,137 @@
...
+    typename ParameterType<T>::type& value,
+    ParamData& d,
+    const typename boost::enable_if<data::HasSerialize<T>>::type* /* junk */)
+{
+  // If the model is an input model, we have to load it from file.  'value'
+  // contains the filename.
+  T& model = *boost::any_cast<T>(&d.mappedValue);
+  if (d.input && !d.loaded)
+  {
+    data::Load(value, "model", model, true);
```
zoq: Not sure, do we have to set ``d.loaded = true;``?

zoq:
I made just some minor comments, besides that I think it would be really nice to automatically call CLI::Destroy() at the end of programs that save model-type parameters. Really excited to review the next part of the automatic bindings project.

src/mlpack/core/util/default_param_impl.hpp
rcurtin: You're right, I wrote the separate overloads because I'd done that for all of the other similar functions, but I can combine these into one.  Fixed in 1ab89807.

src/mlpack/core/util/output_param_impl.hpp
rcurtin: Removed in 1061b0bb.  One issue here is that when we save a matrix with a DatasetInfo, the mappings aren't currently being taken into account since there's no Save() overload that takes a DatasetInfo.  But this isn't a problem---for now---because we don't have any programs or bindings that have a matrix with DatasetInfo as an output parameter.  I'm not sure we ever will either.

src/mlpack/core/util/param_data_impl.hpp
rcurtin: Nice catch, thanks.  Fixed in 8afd83cc.  I also simplified some of the overloads here.

rcurtin:
Thanks for the comments.  Should I go ahead and merge this now?

zoq:
Looks good from my perspective.

PR #826, opened 2016-12-11
Added copy constructor, move constructor and copy assignment operator.

s1998:
I tried to add copy constructor, move constructor and copy assignment operator to DTree.
Function declarations are present in dtree.hpp and function definitions are added in dtree_impl.hpp
Token number #817 .
Waiting for review and suggestions to fix the issue.

### DIFF
--- a/src/mlpack/methods/det/dtree_impl.hpp
+++ b/src/mlpack/methods/det/dtree_impl.hpp
@@ -214,8 +214,8 @@ DTree<MatType, TagType>& DTree<MatType, TagType>::operator=(const DTree<MatType,
 template <typename MatType, typename TagType>
 DTree<MatType, TagType>::DTree(DTree&& obj) 
 {   
-	//Moving the values from object
-	start = std::move(obj.start);
+    //Moving the values from object
+    start = std::move(obj.start);
     end = std::move(obj.end);
     maxVals = std::move(obj.maxVals);
     minVals = std::move(obj.minVals);
@@ -235,7 +235,7 @@ DTree<MatType, TagType>::DTree(DTree&& obj)
     delete right;
     
     //Moving the children 	
-	left = std::move(obj.left);
+    left = std::move(obj.left);
     right =  std::move(obj.right);
 
     //Set obj to default values
@@ -257,11 +257,10 @@ DTree<MatType, TagType>::DTree(DTree&& obj)
 }
 
 
-// Root node initializers
 template <typename MatType, typename TagType>
 DTree<MatType, TagType>& DTree<MatType, TagType>::operator=(DTree<MatType, TagType>&& obj)
 {
-	//Moving the values from obj 
+    //Moving the values from obj 
     start = std::move(obj.start);
     end = std::move(obj.end);
     maxVals = std::move(obj.maxVals);
@@ -304,7 +303,7 @@ DTree<MatType, TagType>& DTree<MatType, TagType>::operator=(DTree<MatType, TagTy
 }
 
 
-
+// Root node initializers
 template <typename MatType, typename TagType>
 DTree<MatType, TagType>::DTree(const StatType& maxVals,
                                const StatType& minVals,
--- a/src/mlpack/methods/det/dtree_impl.hpp
+++ b/src/mlpack/methods/det/dtree_impl.hpp
@@ -165,7 +165,85 @@ DTree<MatType, TagType>::DTree() :
     right(NULL)
 { /* Nothing to do. */ }
 
+template <typename MatType, typename TagType>
+DTree<MatType, TagType>::DTree(const DTree& obj) :
+    start(obj.start),
+    end(obj.end),
+    maxVals(obj.maxVals),
+    minVals(obj.minVals),
+    splitDim(obj.splitDim),
+    splitValue(obj.splitValue),
+    logNegError(obj.logNegError),
+    subtreeLeavesLogNegError(obj.subtreeLeavesLogNegError),
+    subtreeLeaves(obj.subtreeLeaves),
+    root(obj.root),
+    ratio(obj.ratio),
+    logVolume(obj.logVolume),
+    bucketTag(obj.bucketTag),
+    alphaUpper(obj.alphaUpper)
+{   
+    //Allocate space to pointers
+    left = new DTree;
+    right = new DTree;
 
+    // Copy the children 
+    *left = *obj.left;
+    *right = *obj.right;
+}
+
+template <typename MatType, typename TagType>
+const DTree<MatType, TagType>& DTree<MatType, TagType>::operator=(const DTree<MatType, TagType>& obj)
+{
+    if (this != &obj){
+      DTree(obj);
+    }
+
+    return *this;
+}
+
+template <typename MatType, typename TagType>
+DTree<MatType, TagType>::DTree(DTree&& obj) :
+    start(obj.start),
+    end(obj.end),
+    maxVals(obj.maxVals),
+    minVals(obj.minVals),
+    splitDim(obj.splitDim),
+    splitValue(obj.splitValue),
+    logNegError(obj.logNegError),
+    subtreeLeavesLogNegError(obj.subtreeLeavesLogNegError),
+    subtreeLeaves(obj.subtreeLeaves),
+    root(obj.root),
+    ratio(obj.ratio),
+    logVolume(obj.logVolume),
+    bucketTag(obj.bucketTag),
+    alphaUpper(obj.alphaUpper)
+{   
+    //Allocate space to pointers
+    left = new DTree;
+    right = new DTree;
+
+    // Copy the children 
+    *left = *obj.left;
+    *right = *obj.right;
+
+    //Set obj to default values
+    obj.start = 0;
+    obj.end = 0;
+    obj.splitDim = size_t(-1);
+    obj.splitValue = DBL_MAX;
+    obj.logNegError = -DBL_MAX;
+    obj.subtreeLeavesLogNegError = -DBL_MAX;
+    obj.subtreeLeaves = 0;
+    obj.root = true;
+    obj.ratio = 1.0;
+    obj.logVolume = -DBL_MAX;
+    obj.bucketTag = -1;
+    obj.alphaUpper = 0.0;
+    obj.left = NULL;
+    obj.right = NULL;
+    obj.maxVals = arma::vec();
+    obj.minVals = arma::vec();
+}
 // Root node initializers
 
 template <typename MatType, typename TagType>
--- a/src/mlpack/methods/det/dtree_impl.hpp
+++ b/src/mlpack/methods/det/dtree_impl.hpp
@@ -203,17 +203,57 @@ DTree<MatType, TagType>& DTree<MatType, TagType>::operator=(const DTree<MatType,
     logVolume = obj.logVolume;
     bucketTag = obj.bucketTag;
     alphaUpper = obj.alphaUpper;
-    
+
     //Copying the children 
     left = new DTree(*obj.left);
     right = new DTree(*obj.right);
-
+    
     return *this;
 }
 
 template <typename MatType, typename TagType>
-DTree<MatType, TagType>::DTree(DTree&& obj) 
+DTree<MatType, TagType>::DTree(DTree&& obj)     
 {   
+    //Moving the values from object
+    start = std::move(obj.start);
+    end = std::move(obj.end);
+    maxVals = std::move(obj.maxVals);
+    minVals = std::move(obj.minVals);
+    splitDim = std::move(obj.splitDim);
+    splitValue = std::move(obj.splitValue);
+    logNegError = std::move(obj.logNegError);
+    subtreeLeavesLogNegError = std::move(obj.subtreeLeavesLogNegError);
+    subtreeLeaves = std::move(obj.subtreeLeaves);
+    root = std::move(obj.root);
+    ratio = std::move(obj.ratio);
+    logVolume = std::move(obj.logVolume);
+    bucketTag = std::move(obj.bucketTag);
+    alphaUpper = std::move(obj.alphaUpper);
+    left = std::move(obj.left);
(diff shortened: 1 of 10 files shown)


rcurtin:
Hi there,

Thanks for the contribution.  Could you please add some tests for these?  Take a look at some tests I recently wrote in a different branch: https://github.com/rcurtin/mlpack/blob/bindings/src/mlpack/tests/lsh_test.cpp#L840

Once there are working tests I'll give it a full review.

s1998:
I added the tests.

src/mlpack/methods/det/dtree_impl.hpp
```diff
@@ -165,10 +165,154 @@ DTree<MatType, TagType>::DTree() :
...
+    delete left;
+    delete right;
+
+    //Allocate space
+    left = new DTree;
+    right = new DTree;
+
+    //Copying the children 
+    left = obj.left;
+    right = obj.right;
```
rcurtin: Couldn't you simplify this to `left = new DTree(*obj.left);` and the same for the right?

src/mlpack/methods/det/dtree_impl.hpp
```diff
@@ -165,10 +165,154 @@ DTree<MatType, TagType>::DTree() :
...
+    logNegError(obj.logNegError),
+    subtreeLeavesLogNegError(obj.subtreeLeavesLogNegError),
+    subtreeLeaves(obj.subtreeLeaves),
+    root(obj.root),
+    ratio(obj.ratio),
+    logVolume(obj.logVolume),
+    bucketTag(obj.bucketTag),
+    alphaUpper(obj.alphaUpper),
+    left(obj.left),
+    right(obj.right)
```
rcurtin: There are some unnecessary copies here, please use `std::move()` when applicable.

src/mlpack/methods/det/dtree_impl.hpp
```diff
@@ -165,10 +165,154 @@ DTree<MatType, TagType>::DTree() :
...
+    splitDim = obj.splitDim;
+    splitValue = obj.splitValue;
+    logNegError = obj.logNegError;
+    subtreeLeavesLogNegError = obj.subtreeLeavesLogNegError;
+    subtreeLeaves = obj.subtreeLeaves;
+    root = obj.root;
+    ratio = obj.ratio;
+    logVolume = obj.logVolume;
+    bucketTag = obj.bucketTag;
+    alphaUpper = obj.alphaUpper;
```
rcurtin: There are also some unnecessary copies here.

src/mlpack/tests/det_test.cpp
```diff
@@ -494,4 +494,124 @@ BOOST_AUTO_TEST_CASE(TestPrintLeafMembership)
...
+  BOOST_REQUIRE_EQUAL(tree3.maxVals[1], 7);
+  BOOST_REQUIRE_EQUAL(tree3.minVals[1], 0);
+  BOOST_REQUIRE_EQUAL(tree3.maxVals[2], 8);
+  BOOST_REQUIRE_EQUAL(tree3.minVals[2], 1);
+
+  //Checking children were safely copied
+  BOOST_REQUIRE(tree2.Left()==leftChild);
+  BOOST_REQUIRE(tree2.Right()==rightChild);
+  BOOST_REQUIRE(tree3.Left()==leftChild);
+  BOOST_REQUIRE(tree3.Right()==rightChild);
```
rcurtin: At this point `leftChild` and `rightChild` should be invalid pointers; I bet valgrind will call this an invalid memory access and in some cases this test will fail.

src/mlpack/tests/det_test.cpp
```diff
@@ -494,4 +494,124 @@ BOOST_AUTO_TEST_CASE(TestPrintLeafMembership)
...
+  BOOST_REQUIRE_EQUAL(tree2.maxVals[0], 7);
+  BOOST_REQUIRE_EQUAL(tree2.minVals[0], 3);
+  BOOST_REQUIRE_EQUAL(tree2.maxVals[1], 7);
+  BOOST_REQUIRE_EQUAL(tree2.minVals[1], 0);
+  BOOST_REQUIRE_EQUAL(tree2.maxVals[2], 8);
+  BOOST_REQUIRE_EQUAL(tree2.minVals[2], 1);
+
+  //Checking children were safely moved
+  BOOST_REQUIRE(tree2.Left()==leftChild);
+  BOOST_REQUIRE(tree2.Right()==rightChild);
```
rcurtin: Same issue with `leftChild` and `rightChild` here.

src/mlpack/methods/det/dtree_impl.hpp
```diff
@@ -165,10 +165,154 @@ DTree<MatType, TagType>::DTree() :
...
+    bucketTag(obj.bucketTag),
+    alphaUpper(obj.alphaUpper),
+    left(obj.left),
+    right(obj.right)
+{ /* Nothing to do. */ }
 
+template <typename MatType, typename TagType>
+DTree<MatType, TagType>& DTree<MatType, TagType>::operator=(const DTree<MatType, TagType>& obj)
+{
+    if (this != &obj){
```
rcurtin: I don't think these checks are necessary.  If that is true, the user is doing something pretty stupid and we shouldn't need to handle it.

s1998:
@rcurtin 
I tried to fix the issues. The checks are still failing. Travis-ci says that RAModel test failed (I was unable to find how dtree and knn is related. If you could tell me about that, I would try to fix the issue).

src/mlpack/methods/det/dtree_impl.hpp
```diff
@@ -165,9 +165,138 @@ DTree<MatType, TagType>::DTree() :
...
+    subtreeLeavesLogNegError(obj.subtreeLeavesLogNegError),
+    subtreeLeaves(obj.subtreeLeaves),
+    root(obj.root),
+    ratio(obj.ratio),
+    logVolume(obj.logVolume),
+    bucketTag(obj.bucketTag),
+    alphaUpper(obj.alphaUpper),
+    left(obj.left),
+    right(obj.right)
+{ /* Nothing to do. */ }
```
rcurtin: This doesn't actually copy the left and right child nodes, but your copy operator does.  Can you fix that please?

src/mlpack/methods/det/dtree_impl.hpp
```diff
@@ -165,9 +165,138 @@ DTree<MatType, TagType>::DTree() :
...
+    obj.logNegError = -DBL_MAX;
+    obj.subtreeLeavesLogNegError = -DBL_MAX;
+    obj.subtreeLeaves = 0;
+    obj.root = true;
+    obj.ratio = 1.0;
+    obj.logVolume = -DBL_MAX;
+    obj.bucketTag = -1;
+    obj.alphaUpper = 0.0;
+    obj.left = NULL;
+    obj.right = NULL;
```
rcurtin: You should be able to do this as an initialization list.

src/mlpack/methods/det/dtree_impl.hpp
```diff
@@ -165,9 +165,138 @@ DTree<MatType, TagType>::DTree() :
...
 
+}
+
+template <typename MatType, typename TagType>
+DTree<MatType, TagType>& DTree<MatType, TagType>::operator=(DTree<MatType, TagType>&& obj)
+{
+    if (this != &obj){
```
rcurtin: I think the test to see if this is the same as the object being copied is unnecessary.

src/mlpack/methods/det/dtree_impl.hpp
```diff
@@ -165,9 +165,138 @@ DTree<MatType, TagType>::DTree() :
...
+    splitDim = std::move(obj.splitDim);
+    splitValue = std::move(obj.splitValue);
+    logNegError = std::move(obj.logNegError);
+    subtreeLeavesLogNegError = std::move(obj.subtreeLeavesLogNegError);
+    subtreeLeaves = std::move(obj.subtreeLeaves);
+    root = std::move(obj.root);
+    ratio = std::move(obj.ratio);
+    logVolume = std::move(obj.logVolume);
+    bucketTag = std::move(obj.bucketTag);
+    alphaUpper = std::move(obj.alphaUpper);
```
rcurtin: I wouldn't bother with `std::move()` for primitive types like `int`, `double`, etc., since move will degrade to copies for primitive types.  As a result, using `move()` there can actually make the code more confusing, because we can't assume that the rvalues will be set to default values or something.  (That is my current understanding of the spec.)

src/mlpack/methods/det/dtree_impl.hpp
s1998: I actually had seen these links before creating the PR and had used initialization list in copy constructor. I saw them again now. What confused me was that at line 249 you commented that one can use initialization list for the same . But there we are trying to set the default values of the field of object passed (obj - the one on RHS ) and not of the object created.
What I understood from the link was that initialization list can be used to initialize the fields of the object that is being created (the one for which the constructor has been called - on LHS).
I think I haven't still understood what exactly was said here. I have tried to fix all other bugs.

src/mlpack/methods/det/dtree_impl.hpp
s1998: @rcurtin The child node (data) needs to be copied or the pointer to the child node?

rcurtin: This is the copy constructor, so you need to copy the data. Take a look at the copy constructor test you wrote---if you just copy the pointer value, then when you delete the tree you copied from, suddenly the left and right pointers point to invalid memory. I bet that if you run valgrind to check for memory leaks on that test, like

```
$ bin/mlpack_test -t DETTest/CopyConstructorAndOperatorTest
```

then valgrind will throw invalid memory access errors everywhere, even if the test itself does not fail.

src/mlpack/methods/det/dtree_impl.hpp
rcurtin: Yeah---you can use an initialization list in this method to set the values of the object being constructed. You are right that you can't do that with the RHS, but that isn't what I meant. Sorry for the confusion.

s1998:
@rcurtin  
I tried to fix the issues and changed the tests to ensure children along with their data are properly copied. 
I don't know why the appveyor still fails.
`./bin/mlpack_test -p` 
`bin/mlpack_test -t DETTest/CopyConstructorAndOperatorTest`
`bin/mlpack_test -t DETTest/MoveOperatorTest`
`bin/mlpack_test -t DETTest/MoveConstructorTest`give positive results on my system.

src/mlpack/methods/det/dtree_impl.hpp
```diff
@@ -165,10 +165,150 @@ DTree<MatType, TagType>::DTree() :
...
+{   
+    //Copying the children
+    if(obj.left != NULL)
+      left = new DTree(*obj.left);
+    else
+      left = NULL;
+    if(obj.right != NULL)
+      right = new DTree(*obj.right);
+    else
+      right = NULL;
```
rcurtin: You can write these in the initialization list cleverly: `left((obj.left == NULL) ? NULL : new DTree(*obj.left))`, for instance.

src/mlpack/methods/det/dtree_impl.hpp
```diff
@@ -165,10 +165,150 @@ DTree<MatType, TagType>::DTree() :
...
+    
+    //Copying the children   
+    if(obj.left != NULL)
+      left = new DTree(*obj.left);
+    else
+      left = NULL;
+    if(obj.right != NULL)
+      right = new DTree(*obj.right);
+    else
+      right = NULL;
```
rcurtin: What happens when `left` and `right` are not `NULL` when this method is called?

src/mlpack/methods/det/dtree_impl.hpp
```diff
@@ -165,10 +165,150 @@ DTree<MatType, TagType>::DTree() :
...
+    splitValue = std::move(obj.splitValue);
+    bucketTag = std::move(obj.bucketTag);
+    
+    //Free the space allocated
+    delete left;
+    delete right;
+    
+    //Moving children
+    left = std::move(obj.left);
+    right = std::move(obj.right);
```
rcurtin: Since you're just moving pointers here, you don't actually need to use `std::move()`, you can just do `left = obj.left` and `right = obj.right`.

rcurtin:
The AppVeyor failure is a compilation error; if you dig through the log and search for "error " you can see what the issue is:

```
C:\projects\mlpack\src\mlpack\tests\det_test.cpp(516): error C2248: 'mlpack::det::DTree<arma::mat,int>::maxVals': cannot access private member declared in class 'mlpack::det::DTree<arma::mat,int>' [C:\projects\mlpack\build\src\mlpack\tests\mlpack_test.vcxp
```

So although gcc lets you get away with accessing a private member of `DTree`, MSVC doesn't.  It seems like you'll have to find a different way to test that.

s1998:
@rcurtin  I tried to fix the issues.

src/mlpack/methods/det/dtree_impl.hpp
s1998: I think they are supposed to be freed.

s1998:
@rcurtin  I tried to fix the said issues.

rcurtin:
Thanks for taking care of all the issues.  Did you have anything else to modify or should I merge this?

s1998:
I can't think of anything else to modify.

rcurtin:
Ok, merged in.  I made some style fixes in 547e2cb3.  Thanks for your hard work on this PR.  Would you like to add your name and email to the list of contributors in `src/mlpack/core.hpp` and `COPYRIGHT.txt`?

s1998:
@rcurtin  I would like to add my name and email to the list, would I need a different PR for that ?

rcurtin:
Yeah---this one is already merged so you'll need to open another one for that.

s1998:
Okay .

PR #832, opened 2016-12-15
Fixes #755 Log compatibility with std::fixed and std::setprecision

### LINKED ISSUE #755: Log::Info has incompatibility issue with some iomanip functions
Log::Info works fine with std::setw, but I found that it has some issues with some functions in `<iomanip>`.
`std::fixed` and `std::setprecision` function is ignored when used together with Log::<any> functions.

shikharbhardwaj:
Changes /src/mlpack/core/util/prefixedoutstream_impl.hpp

### DIFF
--- a/COPYRIGHT.txt
+++ b/COPYRIGHT.txt
@@ -62,6 +62,7 @@ Copyright:
   Copyright 2016, Peter Lehner <peter.lehner@dlr.de>
   Copyright 2016, Anuraj Kanodia <akanuraj200@gmail.com>
   Copyright 2016, Ivan Georgiev <ivan@jonan.info>
+  Copyright 2016, Shikhar Bhardwaj <shikharbhardwaj68@gmail.com>
 License: BSD-3-clause
   All rights reserved.
   .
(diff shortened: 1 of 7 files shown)


rcurtin:
Nice, thanks!  Can you add this test to `cli_test.cpp` and see if it works?

```
BOOST_AUTO_TEST_CASE(TestFormattedOutput)
{ 
  std::stringstream ss;
  PrefixedOutStream pss(ss, BASH_GREEN "[INFO ]" BASH_CLEAR);
  
  const double pi = std::acos(-1.0);
  pss << std::precision(10) << pi;
  
  BOOST_REQUIRE_EQUAL(ss.str(),
      BASH_GREEN "[INFO ]" BASH_CLEAR "3.141592654");
}
```

After building `mlpack_test` you can test with `mlpack_test -t CLITest/TestFormattedOutput`.

rcurtin:
Looks good, the test passes, but it seems some other problem has been introduced.  I'm trying to get to the bottom of it with gdb.

shikharbhardwaj:
Don't know why, but it passed when I moved the test up.

rcurtin:
Yeah, I think it has to do with the test fixture.  I solved it by simply splitting out the PrefixedOutStream tests and Timer tests into different files, which is probably what I'll do here after merge (unless you want to do it).  Would you like to add your name to the list of contributors in `src/mlpack/core.hpp` and `COPYRIGHT.txt`?  Then I'll merge.

shikharbhardwaj:
Sure, I'll split the tests, add my name and commit ASAP.

rcurtin:
Great, thanks so much for the contribution, this is a nice fix. :)

PR #834, opened 2016-12-18
adds GammaDistribution::Train(observations, probabilities)

### LINKED ISSUE #733: GammaDistribution Implementation is Incomplete

yashu-seth:
fixes #733

### DIFF
--- a/src/mlpack/tests/distribution_test.cpp
+++ b/src/mlpack/tests/distribution_test.cpp
@@ -444,6 +444,148 @@ BOOST_AUTO_TEST_CASE(GammaDistributionTrainTest)
 
 /**
  * This test verifies that the fitting procedure for GammaDistribution works
+ * properly when probabilities for each sample is given.
+ */
+BOOST_AUTO_TEST_CASE(GammaDistributionTrainWithProbabilitiesTest)
+{
+  double alphaReal = 5.4;
+  double betaReal = 6.7;
+
+  // Create a gamma distribution random generator
+  std::default_random_engine generator;
+  std::gamma_distribution<double> dist(alphaReal, betaReal);
+
+  size_t N = 50000;
+  size_t d = 2;
+  arma::mat rdata(d, N);
+
+  for(size_t j = 0; j < d; j++)
+    for(size_t i = 0; i < N; i++)
+      rdata(j, i) = dist(generator);
+
+  // create a uniform distribution random generator
+  std::uniform_real_distribution<double> prob(0, 1);
+  arma::vec probabilities(N);
+
+  for(size_t i = 0; i < N; i++)
+    probabilities(i) = prob(generator);
+
+  // fit results with probabilities and data
+  GammaDistribution gDist;
+  gDist.Train(rdata, probabilities);
+
+  // fit results with only data
+  GammaDistribution gDist2;
+  gDist2.Train(rdata);
+
+  BOOST_REQUIRE_CLOSE(gDist2.Alpha(0), gDist.Alpha(0), 1);
+  BOOST_REQUIRE_CLOSE(gDist2.Beta(0), gDist.Beta(0), 1);
+
+  BOOST_REQUIRE_CLOSE(gDist2.Alpha(1), gDist.Alpha(1), 1);
+  BOOST_REQUIRE_CLOSE(gDist2.Beta(1), gDist.Beta(1), 1);
+
+  BOOST_REQUIRE_CLOSE(alphaReal, gDist.Alpha(0), 1);
+  BOOST_REQUIRE_CLOSE(betaReal, gDist.Beta(0), 1);
+
+  BOOST_REQUIRE_CLOSE(alphaReal, gDist.Alpha(1), 1);
+  BOOST_REQUIRE_CLOSE(betaReal, gDist.Beta(1), 1);
+}
+
+
+// This test ensures that the same result is obtained when
+// trained with probabilities all set to 1 and with no probabilities at all.
+BOOST_AUTO_TEST_CASE(GammaDistributionTrainALLProbabilities1Test)
+{
+  double alphaReal = 5.4;
+  double betaReal = 6.7;
+
+  // Create a gamma distribution random generator
+  std::default_random_engine generator;
+  std::gamma_distribution<double> dist(alphaReal, betaReal);
+
+  size_t N = 1000;
+  size_t d = 2;
+  arma::mat rdata(d, N);
+
+  for(size_t j = 0; j < d; j++)
+    for(size_t i = 0; i < N; i++)
+      rdata(j, i) = dist(generator);
+
+  // fit results with only data
+  GammaDistribution gDist;
+  gDist.Train(rdata);
+
+  // fit results with data and each probability as 1
+  GammaDistribution gDist2;
+  arma::vec allProbabilities1(N, arma::fill::ones);
+  gDist2.Train(rdata, allProbabilities1);
+
+  BOOST_REQUIRE_CLOSE(gDist2.Alpha(0), gDist.Alpha(0), 1e-5);
+  BOOST_REQUIRE_CLOSE(gDist2.Beta(0), gDist.Beta(0), 1e-5);
+
+  BOOST_REQUIRE_CLOSE(gDist2.Alpha(1), gDist.Alpha(1), 1e-5);
+  BOOST_REQUIRE_CLOSE(gDist2.Beta(1), gDist.Beta(1), 1e-5);
+}
+
+ /** This test draws points from two different gamma distributions,
+  *  sets the probabilities for the points from the first distribution
+  *  to something small and the probabilities for the second to something large. 
+  *  It ensures that the gamma distribution recovered has the
+  *  same parameters as the second gamma distribution with high probabilities.
+  */
+BOOST_AUTO_TEST_CASE(GammaDistributionTrainTwoDistProbabilities1Test)
+{
+  double alphaReal = 5.4;
+  double betaReal = 6.7;
+
+  double alphaReal2 = 1.9;
+  double betaReal2 = 8.4;
+
+  // Create two gamma distribution random generators
+  std::default_random_engine generator;
+  std::gamma_distribution<double> dist(alphaReal, betaReal);
+  std::gamma_distribution<double> dist2(alphaReal2, betaReal2);
+
+  std::uniform_real_distribution<double> lowProb(0, 0.02);
+  std::uniform_real_distribution<double> highProb(0.98, 1);
+
+  size_t N = 50000;
+  size_t d = 2;
+  arma::mat rdata(d, N);
+  arma::vec probabilities(N);
+
+  // draws points alternately from the two different distributions.
+  for(size_t j = 0; j < d; j++)
+  {
+    for(size_t i = 0; i < N; i++)
+    {
+      if(i % 2 == 0)
+        rdata(j, i) = dist(generator);
+      else
+        rdata(j, i) = dist2(generator);
+    }
+  }
+
+  for(size_t i = 0; i<N; i++)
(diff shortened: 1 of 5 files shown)


yashu-seth:
@rcurtin For testing this method, I have decided to generate some n points from a fixed gamma distribution and n points between 0 and 1 from a uniform distribution. Then use the points from the uniform distribution as probabilities and points from the original gamma distribution as observations to get an another gamma distribution. Then finally compare the parameters of the two gamma distributions.

yashu-seth:
@rcurtin Should I proceed with the tests or do you have any suggestions? I cannot understand why travis is failing but appveyor passes.

rcurtin:
@yashu-seth: thanks for the contribution, sorry it took me a little while to get around to looking at this.  Don't worry about the test failure, that is for the Nystroem method, it doesn't have to do with your code.  Some of the mlpack tests are probabilistic and although we try to keep the failure probability very low, in some cases the probability isn't low enough.

The testing method seems reasonable---I think the gamma distribution that you'll fit with the uniform distribution for weights should have approximately the same parameters as if you trained it without weights.  Here are another couple ideas for simple tests:

 * Draw points from two different gamma distributions.  Set the probabilities for the points from the first distribution to something small (0.01, 0.001, something like that) and the probabilities for the second to something large (0.99, 0.999, something like that).  The gamma distribution that you recover should have the same parameters as the second gamma distribution you drew from.

 * Train with probabilities all set to 1, and ensure that this gives the same result as when you train with no probabilities at all.

Hope this helps---let me know if I can clarify anything.  I glanced at the implementation, I think it looks right; just needs tests. :)

mlpack-mailing-lists:
I've actually done some work on Gamma EM (a generalization of the Gamma
distribution following the idea of Gaussian EM...)  You can obtain a decent
starting value for your Newton-Raphson by noting that the expected value of
the empirical covariance of X,log X is, in fact, beta.

Another alternative is to use the method of moments to obtain initial
estimates.

yashu-seth:
@rcurtin I have added the requisite tests. Please have a look.

src/mlpack/tests/distribution_test.cpp
```diff
@@ -442,6 +442,97 @@ BOOST_AUTO_TEST_CASE(GammaDistributionTrainTest)
...
+
+  // fit results with only data
+  GammaDistribution gDist2;
+  gDist2.Train(rdata);
+
+  BOOST_REQUIRE_CLOSE(gDist2.Alpha(0), gDist.Alpha(0), 10);
+  BOOST_REQUIRE_CLOSE(gDist2.Beta(0), gDist.Beta(0), 10);
+
+  BOOST_REQUIRE_CLOSE(alphaReal, gDist.Alpha(0), 10);
+  BOOST_REQUIRE_CLOSE(betaReal, gDist.Beta(0), 10);
```
rcurtin: We can probably check these with much closer tolerances.  I'd suggest 1e-5 instead of 10.  Shouldn't we also check `gDist.Alpha(1)` and `gDist.Beta(1)`?

src/mlpack/tests/distribution_test.cpp
```diff
@@ -442,6 +442,97 @@ BOOST_AUTO_TEST_CASE(GammaDistributionTrainTest)
...
+    }
+  }
+
+  for(size_t i = 0; i<N; i++)
+  {
+    if(i%2 == 0)
+      probabilities(i) = low_prob(generator);
+    else
+      probabilities(i) = high_prob(generator);
+  }
```
rcurtin: Another possibility to simplify the creation here is just to put the first N/2 points in the first part of `rdata` from `low_prob` and then the second N/2 points in the second part of `rdata` from `high_prob`, then call `arma::shuffle()` to shuffle the dataset.

yashu-seth: This would require ensuring shuffling rdata matrix and probabilities vector in the same order. How to use shuffle for this?

src/mlpack/tests/distribution_test.cpp
yashu-seth: @rcurtin I just checked closer tolerances are not working. A tolerance less than 10 does not pass all the tests.
I forgot to add ```gDist.Alpha(1)``` and ```gDist.Beta(1)```. I will add the checks.

yashu-seth:
@rcurtin I will split these into multiple tests and also fix the style guide violations.
I would love to get my name added in the contributors list.

src/mlpack/tests/distribution_test.cpp
rcurtin: We can use a larger tolerance between the values of `alphaReal` and `betaReal` and the estimated values, due to the noisy sampling from the gamma distribution.  I still think 10% tolerance is a bit large.  But `gDist` and `gDist2` should have virtually identical values so the tolerance should be very small and 1e-5 should be fine.  If that is causing the tests to fail, then I think the implementation is incorrect.

yashu-seth: @rcurtin I increased the sample size from 500 till 50000 and it achieves a tolerance of 1% but it can't get 1e-5.
Besides, for the low and high probability two distribution test, it achieves a tolerance of < 2%.

yashu-seth: If I sample probabilities from (0.5, 1) instead of (0, 1), then ```gDist``` and gDist2``` have a difference of <1e-5 even for a sample size of 500.

yashu-seth: @rcurtin I just confirmed that tolerance for ```gDist``` and ```gDist2``` is <1e-5 for a sample size of 1000. Let me push the latest changes and you can have a look.

rcurtin: Hm, ok, a larger tolerance like 5% (or even 10% if need be) is okay for the difference between alphaReal and gDist.Alpha(0), but the tolerance between gDist and gDist2 should be 1e-5.  The gDist/gDist2 tolerance is really the one I'm much more concerned with.

One way you can test a tolerance is by adding `math::RandomSeed(std::time(NULL))` (although you are using different RNGs so you'll have to set them accordingly) and then running the test over and over again to see how often it fails.

yashu-seth: @rcurtin I checked the implementation and could not find anything incorrect. 

But I think, it is not necessary gDist and gDist2 should always have a difference less than 1e-5 when we are sampling probabilities from a unifrom distribution between 0 and 1. For example, there would be some data points that would get assigned a very low probability and would be treated as if they are not present at all. So when we train gDist and gDist2 they don't get the exact same data to be trained on.

Something in favour of my argument, can be the fact that when I train the distribution with all probabilities equal to 1, I get the exact same distribution when I train it with just the data.

rcurtin:
Ah, a comment to whoever replied via email, the method used for training here is this one:
http://131.107.65.14/en-us/um/people/minka/papers/minka-gamma.pdf
It seems to work quite well!

src/mlpack/tests/distribution_test.cpp
rcurtin: Ok, I think that we were getting confused because we were talking about probabilities for different tests. :)  I looked through the most recent code, and I think the tolerances are fine now.  When I was saying that the tolerance should be small like 1e-5, I was only referencing the test with probabilities all equal to one.  Thanks for the updates.

src/mlpack/tests/distribution_test.cpp
```diff
@@ -444,6 +444,148 @@ BOOST_AUTO_TEST_CASE(GammaDistributionTrainTest)
...
+  // fit results with data and each probability as 1
+  GammaDistribution gDist2;
+  arma::vec allProbabilities1(N, arma::fill::ones);
+  gDist2.Train(rdata, allProbabilities1);
+
+  BOOST_REQUIRE_EQUAL(gDist2.Alpha(0), gDist.Alpha(0));
+  BOOST_REQUIRE_EQUAL(gDist2.Beta(0), gDist.Beta(0));
+
+  BOOST_REQUIRE_EQUAL(gDist2.Alpha(1), gDist.Alpha(1));
+  BOOST_REQUIRE_EQUAL(gDist2.Beta(1), gDist.Beta(1));
```
rcurtin: We should make this `BOOST_REQUIRE_CLOSE()` with a tolerance of `1e-5` or so, because floating-point errors in some implementations could provide slightly different results.  It looks like your current implementation appears to produce the exact same result, but that may not be true in the future if the code is modified (but still valid), so we should be careful to try and make our test future-proof.

rcurtin:
Things look good to me minus the one comment about `BOOST_REQUIRE_CLOSE`.  I'll probably refactor the tests very slightly after merge to try and use mlpack's existing RNGs instead of custom ones.  You can add your name to `src/mlpack/core.hpp` and `COPYRIGHT.txt`.  Let me know when you think this is ready to merge and I'll hit the button.  Thanks again for taking the time to sort out the issues here.

yashu-seth:
@rcurtin It was nice discussing things with you and getting suggestions. I have made the suggested changes. Please have a look.

rcurtin:
Thanks!  I made some style changes in c739839 and ec65745 and changed the random number generators to use the mlpack random number generator object in 1ef2a02 (some of the tests I changed there were not yours, but still this is an improvement because you can now set the random number seed with `mlpack::math::RandomSeed()`).

Then, for each test, I added a line at the top: `math::RandomSeed(std::time(NULL))` and recompiled.  I ran a bash script like this:

```
while(true); do bin/mlpack_test -t GammaDistributionTrainWithProbabilitiesTest; sleep 1; done
```

This will give output like this:

```
Running 1 test case...

*** No errors detected
Running 1 test case...

*** No errors detected
Running 1 test case...

*** No errors detected
Running 1 test case...
/home/ryan/src/mlpack/src/mlpack/tests/distribution_test.cpp(483): fatal error in "GammaDistributionTrainWithProbabilitiesTest": difference{2.03727%} between betaReal{6.7000000000000002} and gDist.Beta(0){6.5662282671947327} exceeds 2%

Running 1 test case...

*** No errors detected
Running 1 test case...

*** No errors detected
Running 1 test case...

*** No errors detected
Running 1 test case...

*** No errors detected
Running 1 test case...

*** No errors detected
```

Using this strategy I found that some of the tolerances needed to be modified to prevent failures, in the test `GammaDistributionTrainWithProbabilitiesTest`: the comparison between `gDist` and `gDist2` parameters occasionally failed, so I loosened the tolerance to 1.5% from 1%.  The comparisons between `alphaReal` and `betaReal` and the trained distributions often failed, so I loosened the tolerance from 1% to 2.5%.  I think this is an ok change to make and I think these are reasonable bounds for the expected noise.

This strategy is important to test with, because each compilation will (effectively) have a different random seed associated with it, so to ensure that tests pass on users' systems, we have to check with many different random seeds.

Anyway, about the Perceptron code, definitely, if you would like to add tests, it would be much appreciated!  You can create an issue if you like or just open a PR, whatever you prefer is fine.

PR #835, opened 2016-12-19
Add Copy, Move Constructor and Assignment Operator

abhinavrai44:
For issue #817
I have tried to make the required changes and implement copy, move constructors and assignment operators
Looking forward to your response and suggestions.

Thanks


rcurtin:
Hi there Abhinav,

This doesn't compile.  Please make sure that the code compiles.  Also, there are already two other PRs for this issue open, so maybe it is better to find a different issue to work on instead of duplicating work...

Thanks,
Ryan

rcurtin:
This PR is no longer relevant---the original issue has been solved.

PR #836, opened 2016-12-19
cover tree optimization

arirang-pust:
made distance calculations faster in cover trees by not computing the full distance if it is not needed

### DIFF
--- a/src/mlpack/core/tree/cover_tree/cover_tree_impl.hpp
+++ b/src/mlpack/core/tree/cover_tree/cover_tree_impl.hpp
@@ -701,7 +701,7 @@ size_t CoverTree<MetricType, StatisticType, MatType, RootPointPolicy>::
   size_t bestIndex = 0;
   for (size_t i = 0; i < children.size(); ++i)
   {
-    ElemType distance = children[i]->MaxDistance(point);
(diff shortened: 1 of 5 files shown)


src/mlpack/core/tree/cover_tree/cover_tree_impl.hpp
```diff
@@ -876,6 +893,23 @@ CoverTree<MetricType, StatisticType, MatType, RootPointPolicy>::
...
+template<
+    typename MetricType,
+    typename StatisticType,
+    typename MatType,
+    typename RootPointPolicy
+>
+typename CoverTree<MetricType, StatisticType, MatType,
+    RootPointPolicy>::ElemType
+CoverTree<MetricType, StatisticType, MatType, RootPointPolicy>::
+    MaxDistance(const arma::vec& other, const ElemType bestDistance, bool forOverload)
```
rcurtin: I think the signature is wrong here, should this be `MaxDistanceForFast`?  (same with the other `MaxDistance` overload)

rcurtin:
Thanks for the contribution---as soon as we can get this to compile I can test to see what the runtime improvement is.

rcurtin:
Closing for inactivity---I'd love to see this one work, but I think there is still a good amount of work to be done.

PR #840, opened 2016-12-22
Initialise "success" variable to fix warnings

rishabhkumar296:


rishabhkumar296:
Any suggestions why the travis-ci build fails.

rcurtin:
Don't worry about the travis failure, it is not related to your fix here.

src/mlpack/core/tree/rectangle_tree/r_plus_tree_descent_heuristic_impl.hpp
```diff
@@ -43,8 +43,6 @@ size_t RPlusTreeDescentHeuristic::ChooseDescentNode(TreeType* node,
         node->Child(bestIndex).Bound();
     bound |= node->Dataset().col(point);
 
-    success = true;
-
```
rcurtin: The change that you've made means that during this loop, `success` is not reinitialized to `true` each iteration.  So this introduces a bug.

rcurtin:
I think you force pushed, so I did not see that there was a change.  This looks better, thanks for the contribution.  If you'd like to add your name to the list of contributors (even though this is a small contribution!) to `src/mlpack/core.hpp` and `COPYRIGHT.txt`, I'll merge that in too.

PR #841, opened 2016-12-22
faster build times

arirang-pust:
I replaced all instances of `#include <mlpack/core.hpp>` with `#include <mlpack/prereqs.hpp>`; this simplifies the dependency graph and reduces compile times.  For example, running `make mlpack_knn` after having modified `mlpack/core/metrics/lmetric.hpp` now only requires compiling 4 files instead of 30.

### DIFF
--- a/src/mlpack/core/metrics/lmetric.hpp
+++ b/src/mlpack/core/metrics/lmetric.hpp
@@ -83,6 +83,12 @@ class LMetric
   static typename VecTypeA::elem_type Evaluate(const VecTypeA& a,
                                                const VecTypeB& b);
 
+ 
+  template<typename VecTypeA, typename VecTypeB>
+  static typename VecTypeA::elem_type Evaluate(const VecTypeA& a,
+                                               const VecTypeB& b,
+					       typename VecTypeA::elem_type bound);
+
   //! Serialize the metric (nothing to do).
   template<typename Archive>
   void Serialize(Archive& /* ar */, const unsigned int /* version */) { }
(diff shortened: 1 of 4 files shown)


arirang-pust:
The project doesn't compile anymore after merging.  There's a bunch of boost errors about `BOOST_MATH_CONSTEXPR` not being defined inside of `mlpack/core/boost_backport/policy.hpp`.  I've tried searching for the right file to include but can't find any information on the internet about it.

rcurtin:
I merged from master with

```
git remote add upstream https://github.com/mlpack/mlpack.git
git fetch upstream
git merge upstream/master
```

and to resolve the conflict with `spherical_kernel.hpp` I changed the includes on that file to just `#include <mlpack/prereqs.hpp>` and this works for a compile.  The boost header isn't necessary anymore---when I went through a few weeks back to try and accelerate compile time a little, I was able to just use `std::tgamma()` instead, so that header is no longer necessary.

Maybe the problem you are seeing is a result of old compile artifacts or something and the build directory needs to be cleaned?  If you can reproduce it, let me know what version of Boost so I can try and reproduce also.

When I do some timing comparisons, making just the core library takes 4m55s of user time instead of 6m21s, a nice improvement. :)

The real speed killer, I think, is all of the boost headers.  I have a dream of managing to un-include any boost headers (including boost serialization) from any file that isn't directly using boost functionality, and I've made some steps in that direction, but it's difficult.  Anyway, this is a nice step.

arirang-pust:
I accidentally overwrote the old commits, with new ones.  Whoops.  Ignore the commit in this pr.

PR #842, opened 2016-12-23
Faster build times

arirang-pust:
Replaced all instances of `#include <mlpack/core.hpp>` with `#include <mlpack/prereqs.hpp>`.  This doesn't affect build times when building the whole project, but it greatly reduces build times after modifying only a few files.

### DIFF
--- a/src/mlpack/core/boost_backport/policy.hpp
+++ b/src/mlpack/core/boost_backport/policy.hpp
@@ -6,6 +6,7 @@
 #ifndef BOOST_MATH_POLICY_HPP
 #define BOOST_MATH_POLICY_HPP
 
+#include <mlpack/core.hpp>
 #include <boost/mpl/list.hpp>
 #include <boost/mpl/contains.hpp>
 #include <boost/mpl/if.hpp>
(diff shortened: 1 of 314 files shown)


mikeizbicki:
This is a better version of #841.  It compiles fine now on my machine (g++ 4.9.2), I'm not sure why it doesn't compile on the appveyor machine.

rcurtin:
```
C:\projects\mlpack\src\mlpack\core\math\lin_alg.cpp(233): error C2065: 'M_SQRT2': undeclared identifier [C:\projects\mlpack\build\src\mlpack\mlpack.vcxproj]
```

That's our offending line, but it's strange because `<mlpack/prereqs.hpp>` includes `<cmath>`.

http://stackoverflow.com/questions/6563810/m-pi-works-with-math-h-but-not-with-cmath-in-visual-studio

So, worth a shot is the explanation that something else is including `<cmath>` before `_USE_MATH_DEFINES` is defined.  A "shot in the dark" idea is to push that `<cmath>` include and `_USE_MATH_DEFINES` definition in `prereqs.hpp` up to be the first thing included.  If you like, push a change like that and let's see what AppVeyor thinks... :)

rcurtin:
I guess the shot in the dark idea worked. :)  Thanks!  This is a nice improvement.

As far as credit goes, should we add the names of the students who did this fix (plus yours) to the list of contributors?

mikeizbicki:
This was work that I did in order to make it easier for the students.

rcurtin:
Ah, ok.  I was surprised not to see your name on the list of contributors, I thought you'd done some other fixes too.  Do you want me to add it? :)

mikeizbicki:
Yeah, thanks.

I think I submitted a couple of other pull requests on the documentation,
but not anything code related.

rcurtin:
Ok, added in d4b2c4b.

PR #847, opened 2017-01-02
More tests for perceptron

yashu-seth:


yashu-seth:
I also noticed that giving the -n option raises Segmentation Fault but --max_iterations runs fine.

```
yashu-seth@ubuntu:~/work/temp$ mlpack_perceptron -t tdata.csv -l label.csv -M prediction.xml -n 500
Segmentation fault (core dumped)
yashu-seth@ubuntu:~/work/temp$ mlpack_perceptron -t tdata.csv -l label.csv -M prediction.xml --max_iterations 500
```

I could not figure out the exact reason behind this.

src/mlpack/methods/perceptron/perceptron_main.cpp
```diff
@@ -49,13 +49,13 @@ PROGRAM_INFO("Perceptron",
...
     "\n\n"
-    "$ perceptron -t training_data.csv -l training_labels.csv -m perceptron.csv"
+    "$ perceptron -t training_data.csv -l training_labels.csv -M perceptron.xml"
     "\n\n"
     "Then, this model can be re-used for classification on 'test_data.csv'.  "
     "The example below does precisely that, saving the predicted classes to "
     "'predictions.csv'."
     "\n\n"
-    "$ perceptron -i perceptron.xml -T test_data.csv -o predictions.csv"
+    "$ perceptron -m perceptron.xml -T test_data.csv -o predictions.csv"
```
rcurtin: Nice catch, thanks for spotting this one.

src/mlpack/tests/perceptron_test.cpp
```diff
@@ -23,6 +24,61 @@ using namespace mlpack::distribution;
...
+  BOOST_CHECK_EQUAL(weights(1, 0), 0);
+  BOOST_CHECK_EQUAL(weights(2, 0), 1);
+  BOOST_CHECK_EQUAL(weights(3, 0), 2);
+  BOOST_CHECK_EQUAL(weights(4, 0), 3);
+
+  BOOST_CHECK_EQUAL(weights(0, 2), 7);
+  BOOST_CHECK_EQUAL(weights(1, 2), 8);
+  BOOST_CHECK_EQUAL(weights(2, 2), 9);
+  BOOST_CHECK_EQUAL(weights(3, 2), 10);
+  BOOST_CHECK_EQUAL(weights(4, 2), 11);
```
rcurtin: Why not check column 1 also?  That should be `[0 1 2 3 4]`, right?

src/mlpack/tests/perceptron_test.cpp
```diff
@@ -23,6 +24,61 @@ using namespace mlpack::distribution;
...
+
+  BOOST_CHECK_EQUAL(biases(0), 1);
+  BOOST_CHECK_EQUAL(biases(2), 8);
+
+  double instanceWeight = 3.0;
```
rcurtin: I know that it causes code duplication, but could we split this into two separate tests?  It makes it easier to see what is wrong when tests fail: if the first (and possibly) test fails, then something is wrong in general with `SimpleWeightUpdate::UpdateWeights()`; if the second test fails, then we know that something is wrong with the instance weights, but that it works otherwise.  This can help save some time in debugging.

src/mlpack/tests/perceptron_test.cpp
rcurtin: Ack, sorry, ignore this.  I forgot that SimpleWeightUpdate only updates one positive and negative class at a time.  (I wonder if that should be updated, since there can be more than one incorrect class.)

src/mlpack/tests/perceptron_test.cpp
```diff
@@ -23,6 +24,61 @@ using namespace mlpack::distribution;
...
+
+  vec trainingPoint("1 2 3 4 5");
+  mat weights("0 1 6;"
+              "2 3 6;"
+              "4 5 6;"
+              "6 7 6;"
+              "8 9 6");
+  vec biases("2 5 7");
+  size_t incorrectClass = 0;
+  size_t correcClass = 2;
```
rcurtin: Do you mind adding a comment somewhere in here about what `UpdateWeights()` is supposed to do?  (In this case, add the point to the correct class weights and subtract the point from the incorrect class weights.)

rcurtin:
Hey Yashu,

Thanks for the PR, this is a nice test addition.  I made just a few comments; if you can address those please then I'll go ahead and merge.

I can't reproduce your segfault; check and make sure that you don't have a stray libmlpack.so of an incorrect version in your `LD_LIBRARY_PATH` or something.

src/mlpack/tests/perceptron_test.cpp
yashu-seth: For a particular iteration, there can only be one incorrect class - the class the networks preditcs according to the weights but it is not the same as the training label.

yashu-seth:
Hey Ryan,

Thanks for the review. I have split the test into two separate tests. I have also added some comments. Please have a look.

rcurtin:
Great, thanks so much for the PR.  Adding more tests is always a good thing. 👍

Issue #844, opened 2016-12-24
Error while using mlpack::data::Load


Lakshya1605:
May be i am doing something wrong but whenever i am trying to load a ".csv" file into a vector i get error like below
 "error: Mat::init(): requested size is not compatible with column vector layout
terminate called after throwing an instance of 'std::logic_error'
  what():  Mat::init(): requested size is not compatible with column vector layout
Aborted (core dumped)".
But armadillo load  (ie=> x.load("***.csv") is working fine.
Is this a specified behaviour or a bug.
Thank You

Lakshya1605:
If it is a bug i would like to work on it please can someone help me out( i

mohit-raj:
Are you executing load through C++ Api or mlpack-CLI(command line interface) ?

mohit-raj:
In case you are load armadillo then you need to transpose the matrix after load, if the coordinate list is in row major format. It might be the cause as the error says requested size is not compatible with column vector layout.
Commands to run :
Let x be the matrix name.
arma :: sp_mat x; // only if in case of loading sparse matrix (this is optional)
x.load("***.csv")  //armadilo load 
x = x.t()  // Here is your command to transpose the matrix.
Try this if it works out fine.

Lakshya1605:
@mohit-raj  sorry my bad i missed the phrase "using mlpack::data::Load" .However i have edited it now.And yes using armadillo load is working fine  but 
Thank You

Lakshya1605:
@mohit-raj  sorry my bad i missed the phrase "using mlpack::data::Load" .However i have edited it now.And yes using armadillo load is working fine but loading using mlpack load isn't.
Thank You

Lakshya1605:
@mohit-raj @rcurtin  could you please confirm if it is a bug .If yes i would like to work on it ,please can you help me get started.

rcurtin:
`data::Load()` was not intended to load column or row vectors.  You are free to help extend the function so that it does, but please be certain before you open a PR that the code compiles, is tested, and follows the mlpack style guidelines.

Lakshya1605:
@rcurtin can you please tell me on how to get started ,where can i find the code for data::Load , and how to compile it after making changes and coding standards of mlpack ( i have tried finding the solution to the  above in the main website and also in the github page but could not find it).Sorry if i have overlooked something.
Thank You

rcurtin:
[...]

PR #848, opened 2017-01-07
Improved Load to load column vectors #844

Lakshya1605:
#844

### DIFF
--- a/src/mlpack/core/data/load_impl.hpp
+++ b/src/mlpack/core/data/load_impl.hpp
@@ -91,6 +91,30 @@ bool inline inplace_transpose(arma::Mat<eT>& X)
 
 template<typename eT>
 bool Load(const std::string& filename,
+          arma::Col<eT>& vec,
+          const bool fatal)
+{
+  arma::mat matrix(vec);
+  Load(filename, matrix, fatal, false);
+  //Check if the returned matrix is vector if yes then 
+  //copy this to our vec else return error
+  if( matrix.is_vec() )
+    vec = matrix;
+  else
+    if( fatal)
+      Log::Fatal << "Loading '" << filename << "'failed "<<
+          ", the data you are tying to load is not a column vector "<< 
+          std::endl; 
+    else
+      Log::Warn << "Loading '" << filename << "'failed " <<
+          ", the data you are trying to load is not a column vector " <<
+          std::endl;
+    return false;
+  return true;
+}
+
+template<typename eT>
+bool Load(const std::string& filename,
           arma::Mat<eT>& matrix,
           const bool fatal,
           const bool transpose)
--- a/src/mlpack/core/data/load_impl.hpp
+++ b/src/mlpack/core/data/load_impl.hpp
@@ -16,6 +16,7 @@
 #include "load.hpp"
 #include "extension.hpp"
 
+#include<exception>
 #include <algorithm>
 #include <mlpack/core/util/timers.hpp>
 
@@ -94,22 +95,16 @@ bool Load(const std::string& filename,
           arma::Col<eT>& vec,
           const bool fatal)
 {
-  arma::mat matrix(vec);
-  Load(filename, matrix, fatal, false);
-  //Check if the returned matrix is vector if yes then 
-  //copy this to our vec else return error
-  if( matrix.is_vec() )
-    vec = matrix;
-  else
-    if( fatal)
-      Log::Fatal << "Loading '" << filename << "'failed "<<
-          ", the data you are tying to load is not a column vector "<< 
-          std::endl; 
-    else
-      Log::Warn << "Loading '" << filename << "'failed " <<
-          ", the data you are trying to load is not a column vector " <<
-          std::endl;
-    return false;
+  Load(filename, vec, fatal, false);
+  return true;
+}
+
+template<typename eT>
+bool Load(const std::string& filename,
+          arma::Row<eT>& rowvec,
+          const bool fatal)
+{
+  Load(filename, rowvec, fatal, false);
   return true;
 }
 
@@ -325,7 +320,9 @@ bool Load(const std::string& filename,
   // We can't use the stream if the type is HDF5.
   bool success;
   if (loadType != arma::hdf5_binary)
-    success = matrix.load(stream, loadType);
+  {
+      success = matrix.load(stream, loadType);
+  }
   else
     success = matrix.load(filename, loadType);
 
--- a/src/mlpack/core/data/load_impl.hpp
+++ b/src/mlpack/core/data/load_impl.hpp
@@ -320,9 +320,7 @@ bool Load(const std::string& filename,
   // We can't use the stream if the type is HDF5.
   bool success;
   if (loadType != arma::hdf5_binary)
-  {
-      success = matrix.load(stream, loadType);
-  }
+    success = matrix.load(stream, loadType);
   else
     success = matrix.load(filename, loadType);
 
--- a/src/mlpack/tests/load_save_test.cpp
+++ b/src/mlpack/tests/load_save_test.cpp
@@ -184,6 +184,84 @@ BOOST_AUTO_TEST_CASE(LoadTransposedCSVTest)
   // Remove the file.
   remove("test_file.csv");
 }
+/**
+  *Make sure ColVec can be loaded
+**/
+BOOST_AUTO_TEST_CASE(LoadColVecCSVTest)
+{
+  fstream f;
+  f.open("test_file.csv", fstream::out);
+
+  for( int i = 0; i < 8; ++i)
+    f << i << endl;
+
+  f.close();
+
+  arma::vec test;
+  BOOST_REQUIRE(data::Load("test_file.csv", test, false) == true);
+
+  BOOST_REQUIRE_EQUAL(test.n_cols, 1);
+  BOOST_REQUIRE_EQUAL(test.n_rows, 8);
+
+  for(size_t i = 0; i < 8; ++i)
+    BOOST_REQUIRE_CLOSE(test[i], (double) (i), 1e-5);
+
+  //Remove the file
+  remove("test_file.csv");
+}
+
+/**
+  *Make Sure Load throws Exception when trying
+  *to load a Matrix in ColVec 
+  *and RowVec
+**/
+BOOST_AUTO_TEST_CASE(LoadMatinColVec)
+{
+  fstream f;
+  f.open("test_file.csv", fstream::out);
+
+  f << "1,2" << endl;
+  f << "3,4" << endl;
+
+  f.close();
+
+  arma::colvec coltest;
+  BOOST_WARN_THROW(data::Load("test_file.csv", coltest, false),
+      std::exception);
+
+  arma::rowvec rowtest;
+  BOOST_WARN_THROW(data::Load("test_file.csv", rowtest, false),
+      std::exception);
+
+  remove("test_file.csv");
+}
+
+/**
+  *Make sure RowVec can be loaded
+**/
+BOOST_AUTO_TEST_CASE(LoadRowVecCSVTest)
+{
+  fstream f;
+  f.open("test_file.csv", fstream::out);
+
+  for( int i = 0 ; i < 7;++i)
+    f << i << ",";
+  f << "7";
+  f << endl;
+
+  f.close();
+
+  arma::rowvec test;
+  BOOST_REQUIRE(data::Load("test_file.csv", test, false) == true);
+
+  BOOST_REQUIRE_EQUAL(test.n_cols, 8);
+  BOOST_REQUIRE_EQUAL(test.n_rows, 1);
+
+  for( size_t i = 0; i < 8 ; ++i)
+    BOOST_REQUIRE_CLOSE(test[i], (double) (i) , 1e-5);
+
+  remove("test_file.csv");
+}
 
 /**
  * Make sure TSVs can be loaded in transposed form.
--- a/src/mlpack/tests/load_save_test.cpp
+++ b/src/mlpack/tests/load_save_test.cpp
@@ -215,6 +215,7 @@ BOOST_AUTO_TEST_CASE(LoadColVecCSVTest)
   *to load a Matrix in ColVec 
   *and RowVec
 **/
+#ifdef DEBUG
 BOOST_AUTO_TEST_CASE(LoadMatinColVec)
 {
   fstream f;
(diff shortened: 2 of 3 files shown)


zoq:
Depending on the data size is probably worth to integrate the vector into the existing ```Load``` function. That way we could directly use the vector instead of using a matrix first. What do you think?

Lakshya1605:
Sorry i did not catch you but if you meant implementing vector load with using existing matrix load then
I have not directly implemented it because the code would be same as the one already existing(matrix load) just the changes would to use vector instead of matrices .I thought this would not be a good idea . But if implementing without using matrix is recommended way i will try that .Thank You !!!

stereomatchingkiss:
If you are interesting about improve loading time, you can check about pull
request 735.
I rewrite the part of csv loading, speed improve a lot

Original implementation : around 7220ms
fast csv : around 600ms

Edit :
The file have 1 million rows and 7 cols


2017-01-09 0:59 GMT+08:00 Lakshya <notifications@github.com>:

zoq:
You are right we should avoid unnecessary code duplication. But I think, we could modify the existing code without any duplication by using a template for the Load function.

```
bool success;
if (loadType != arma::hdf5_binary)
  success = matrix.load(stream, loadType);
else
  success = matrix.load(filename, loadType);
```

Works for vectors too, so 

```
Log::Info << "Size is " << (transpose ? matrix.n_cols : matrix.n_rows)
    << " x " << (transpose ? matrix.n_rows : matrix.n_cols) << ".\n";
```

is probably the only part that needs to be adjusted. What do you think?

Lakshya1605:
@zoq Yes i think this is a better idea(using template for Load function), also i have tried it before trying the given solution but it didn't work .I am not so good at using templates,i have tried googling around on how to fit template in this case but could not find a right answer.It would be of great help if you can help me out  .Thank You !!!

zoq:
A solution would be to use:

```
template<typename MatType>
bool Load(const std::string& filename,
          MatType& input,
          const bool fatal = false,
          const bool transpose = true);
```

instead of:

```
template<typename eT>
bool Load(const std::string& filename,
          arma::Mat<eT>& matrix,
          DatasetInfo& info,
          const bool fatal = false,
          const bool transpose = true);
```

in the header and implementation file. That way ``Load` matches for arma::Mat<eT>, arma::Col<eT>, arma::Row<eT> etc. As mentioned above most of the Load function should work for matrix and the vector type. You can write a specialized function for the matrix and vector type for the parts where you can't reuse the existing code. I hope this makes sense, if not don't hesitate to ask.

Lakshya1605:
@zoq It turns out that we don't have to modify the code for loading vec or rowvec using "data::Load" function .If we keep the option "transpose" as "false" we can load vec as well as rowvec using the existing "Load" function.
Something like this is working 
`arma::vec data;
data::Load("data.csv",data,false,false);`

here data.csv contain elements only in single column.

something like this is also working 
`arma::rowvec data;
data::Load("data_1.csv",data,false,false);`

here data_1.csv contains elements in a single row.

rcurtin:
Unfortunately we can't set `transpose` to `false` as the default option because we need to transpose the vast majority of matrices that are loaded.  So I think that we should have two different overloads of `Load()`:

 * One that operates just like now, on Armadillo matrix objects
 * One that doesn't have a transpose option and operates on Armadillo vector objects (Row or Col)

It seems like a simple way forward might be to create an overload of `Load()` for vector objects that calls the existing method with `transpose = false`.

Lakshya1605:
@rcurtin would it fine if i just removed the line stating declaration of matrix from my pull request and rather use "Load" on the vec directly .Presently i have something like below in my code.
`arma::mat matrix(vec);
Load(filename,matrix,fatal,false;
if(matrix.isvec())
 vec = matrix`

Rather i was think if i could just use the vec directly now for loading like below.
`Load(filename,vec,fatal,false);`

Also the same thing can be done with  rowvec.

rcurtin:
I think that could work, but I don't think that the error checking in the later part of the function will be necessary since `Load()` will already throw an error if you try to load incorrect-size data.

Also, can you write some tests for this new functionality?  We should test that we can load into `vec` and `rowvec` and `Col<size_t>` and `Row<size_t>`, and also test that trying to load incorrectly sized data (like a matrix) into a vector causes a load failure.

Lakshya1605:
@rcurtin i have modified the code as mentioned and also added test cases .But i am unware on how to run test cases can you please help me out .Thank You !!!

rcurtin:
You can use `BOOST_REQUIRE_THROW()` to ensure that exceptions are thrown.

The way to run the test cases is to build `mlpack_test`, then run it.  If you want to run only a specific test suite, like "LoadSaveTest", you can run `mlpack_test -t LoadSaveTest`.

Lakshya1605:
@rcurtin .Sorry if this may seem to be a very basic question but i am not understanding why 'BOOST_REQUIRE_THROW()` is not working .
When ever I run a program in which a matrix is being tried to load in a colvec i get error like below.
`error: Mat::init(): requested size is not compatible with row vector layout
  what():  Mat::init(): requested size is not compatible with row vector layout
Aborted (core `dumped).`

So i tried using `BOOST_REQUIRE_THROW(data::Load("test_file.csv", test), std::logic_error)` 
, `BOOST_REQUIRE_THROW(data::Load("test_file.csv", test), arma::Mat:init)` , `BOOST_REQUIRE_THROW(data::Load("test_file.csv", test), Mat::init)`  
But none of them seem to work .
Here "test_file.csv" is a matrix and "test" is a "colvec".

also i have tried something like below
`try{data::Load("test_file.csv",test)};
catch(std::exception& e) {
std::cout << typeid(e).name() << std::endl;
}`

The output of the above program is 
`error:Mat::init()` : requested size if not compatible with col vec layout;
St11logic_error`

so i tried `BOOST_REQUIRE_THROW(data::Load("test_file.csv", test), St11logic_error` , `BOOST_REQUIRE_THROW(data::Load("test_file.csv", test), std::St11logic_error`
but this also did not work out.

Can you please help me out .
Thank You!!!.

Lakshya1605:
@rcurtin @zoq .Sorry if this may seem to be a very basic question but i am not understanding why `BOOST_REQUIRE_THROW()`  is not working .
When ever I run a program in which a matrix is being tried to load in a colvec i get error like below.
`error: Mat::init(): requested size is not compatible with row vector layout
  what():  Mat::init(): requested size is not compatible with row vector layout
Aborted (core `dumped).`

So i tried using `BOOST_REQUIRE_THROW(data::Load("test_file.csv", test), std::logic_error)` 
i get something like this `/home/lakshya/mlpack/src/mlpack/tests/load_save_test.cpp(229): fatal error in "LoadMatinColVec": exception std::logic_error is expected`.

Then i tried
, `BOOST_REQUIRE_THROW(data::Load("test_file.csv", test), arma::Mat:init)` , `BOOST_REQUIRE_THROW(data::Load("test_file.csv", test), Mat::init)`  
By using the above "make mlpack_test" gives error .
Here "test_file.csv" is a matrix and "test" is a "colvec".

also i have tried something like below
`try{data::Load("test_file.csv",test)};
catch(std::exception& e) {
std::cout << typeid(e).name() << std::endl;
}`

The output of the above program is 
`error:Mat::init() : requested size if not compatible with col vec layout;
St11logic_error`

so i tried `BOOST_REQUIRE_THROW(data::Load("test_file.csv", test), St11logic_error)` , `BOOST_REQUIRE_THROW(data::Load("test_file.csv", test), std::St11logic_error)`
but this also did not work out.

Can you please help me out .
Thank You!!!.

rcurtin:
Hi @Lakshya1605,

There's no need to send multiple email responses.  People work on mlpack typically in their free time and may not be available to send immediate responses.  For instance, in my case, I was gone over the weekend and unable to respond.

As for your issue, take a look through the Boost Unit Test Framework documentation.  The error message `exception std::logic_error is expected` means that the exception was expected by the test, but the code never threw the exception.

Lakshya1605:
@rcurtin Sorry i will make sure that this will not repeat again .Also regarding the issue i have modified the code but i have mistakenly merged the upstream with my pull request branch rather than my master branch can you tell me how to revert this.

rcurtin:
I'm not sure exactly what you did with git, so you'll have to look at the git documentation or try to find some tutorials online that can help you out.

I looked through the code, it looks like it does the right thing.  But the tests should be using `BOOST_REQUIRE_THROW()` not `BOOST_WARN_THROW()`.  If you use `BOOST_WARN_THROW()` then it does not fail when the required behavior (throwing an exception) does not happen, and the test is pointless.

Are you sure that you are calling `data::Load()` in such a way that the exception will even be thrown if there is a problem?  Take a look through the `data::Load()` sources.

Lakshya1605:
@rcurtin  I have changed the line regarding matrix.Load to something like below 
` if (loadType != arma::hdf5_binary)
  {
     try
     {
         success = matrix.load(stream, loadType);
     }
     catch(std::logic_error& e)
     {
          throw std::logic_error"UnsupportedLoad";
     }
}` 

When i run it on a test program its throws error like below.
`error: Mat::init(): requested size is not compatible with row vector layout terminate called after throwing an instance of 'std::logic_error' what(): UnsupportedLoad `

So i thought since the code is throwing std::logic_error.I can use something like below.

But i don't understand why `BOOST_REQUIRE_THROW(data::Load("test.csv", rowvec, true), std::logic_error) ` doesn't work .

Can you please tell what type of error Mat::init() ( from armadillo library)usually throws .Thank You!!!

rcurtin:
Hi @Lakshya1605, I'm sorry for the slow response.  It has been hard to find time to look into this, but I did look this morning.  It turns out that something a little bit tricky is happening here---the `BOOST_REQUIRE_THROW()` is successfully catching the `std::logic_error` that Armadillo is throwing (and it prints the "requested size is not compatible with column vector layout" message), but then another exception is thrown:

```
unknown location(0): fatal error: in "LoadSaveTest/LoadMatinColVec": std::runtime_error: Timer::Start(): timer 'loading_data' has already been started
/home/ryan/src/mlpack-lakshya/src/mlpack/tests/load_save_test.cpp(234): last checkpoint
```

This second exception is thrown by `data::Load()` when it tries to start the `loading_data` timer (see the first line of `data::Load()`).  So I think it's fine in this case to add the line `Timer::Stop("loading_data")` just after the `BOOST_REQUIRE_THROW()`, and this should cause the test to pass.

Let me know if I can clarify anything that I wrote.

Lakshya1605:
@rcurtin  I have tried as you have mentioned but it doesn't work .

rcurtin:
Unfortunately, I can't help you debug the problem with so little information.

Lakshya1605:
@rcurtin @zoq It turns out that In the test file "load_save_test.cpp" `arma::colvec` and `arma::rowvec`
are treated as `arma::mat`. So even `arma::colvec` is able to load matrices rather than just column vector.

I have tried out something like this
`arma::mat A(5,5, fill::randu);`
`arma::colvec coltest(A)`

It should throw an `std::logic_error` while doing this.
When I write a separate code for this let's say in a file "X.cpp"(without using "BOOST"), it throws an error as mentioned .But writing this in "load_save_test.cpp" gives no error .In fact I also tried printing "coltest" and it gets printed same as "A".

Also I tried to change some of the `arma::mat` to `arma::colvec` in "load_save_test.cpp" and to my shock it complied and worked well ,when I expected it to give error.

Is this some specified behaviour of  boost ?
I have tried Googling around on how to fix this error but nothing seems to work.
It would be of great help if you can suggest me on how to resolve this .
Thank You !!!

Lakshya1605:
@rcurtin  The travis build failed due to another file 'gmm_test.cpp' which is one of the other test cases which I haven't edited in this PR. Kindly suggest me what I should do to rectify the same.
Thank you.

src/mlpack/core/data/load_impl.hpp
```diff
@@ -301,7 +320,9 @@ bool Load(const std::string& filename,
   // We can't use the stream if the type is HDF5.
   bool success;
   if (loadType != arma::hdf5_binary)
-    success = matrix.load(stream, loadType);
+  {
+      success = matrix.load(stream, loadType);
+  }
```
rcurtin: Seems like there is no reason to add braces here.

src/mlpack/tests/load_save_test.cpp
```diff
@@ -184,6 +184,87 @@ BOOST_AUTO_TEST_CASE(LoadTransposedCSVTest)
   // Remove the file.
   remove("test_file.csv");
 }
+/**
+  *Make sure ColVec can be loaded
+**/
```
rcurtin: The spacing is off in the comment here (and other comments); I am not sure if that is a function of spaces or tabs or what. Can you please fix it?

src/mlpack/tests/load_save_test.cpp
```diff
@@ -184,6 +184,87 @@ BOOST_AUTO_TEST_CASE(LoadTransposedCSVTest)
...
+BOOST_AUTO_TEST_CASE(LoadColVecCSVTest)
+{
+  fstream f;
+  f.open("test_file.csv", fstream::out);
+
+  for( int i = 0; i < 8; ++i)
```
rcurtin: Minor syntax issue: should be `for (int i` not `for( int i`.

src/mlpack/tests/load_save_test.cpp
```diff
@@ -184,6 +184,87 @@ BOOST_AUTO_TEST_CASE(LoadTransposedCSVTest)
...
+  //Remove the file
+  remove("test_file.csv");
+}
+
+/**
+  *Make Sure Load throws Exception when trying
+  *to load a Matrix in ColVec 
+  *and RowVec
+**/
+#ifdef DEBUG
```
rcurtin: It would be nice to add a comment here pointing out why this test only works under debug mode.

rcurtin:
Hi Lakshya, thanks for the fixes. I think this is about ready to be merged.  If you can address the last couple comments I will go ahead and merge it.

Don't worry about the random test failures.  Some of the mlpack tests are probabilistic, so they may randomly fail. We try to keep the failure probabilities low, but sometimes our tolerances are still too small. So those failures you observed have nothing to do with your code.

Lakshya1605:
@rcurtin Please review .

rcurtin:
Thanks for the contribution.  In f3fcc20b9 I made some minor style fixes and in 41fa0b5e I added some forward declarations of the new overloads with documentation.  Now there is another issue that we can solve to improve this further, I'll go ahead and open that in a few minutes. :)

PR #858, opened 2017-01-24
Add decision tree implementation

rcurtin:
This is an implementation of a generalized decision tree learner.  It's made with a `NoRecursion` parameter so that eventually it can replace the decision stump implementation.  It has a command-line program that can be used to train and predict with decision trees.  The implementation here closely resembles C4.5, but it isn't exactly that.

@lozhnikov, @MarcosPividori, @mentekid: this is another step towards the automatic algorithm selection for nearest neighbors that we talked about so many months ago.  Unfortunately, other work has gotten in the way and I have not been able to get to this until now.  There are a couple pieces remaining: a random forest implementation (I am moving to work on that next), then an implementation of the optimizer SMAC (which needs random forests), a finished implementation of LSH tuning, and then some glue to put it all together.  It may not happen by ICML (end of February) but maybe this is fine in a journal anyway.  Feel free to review this if you like, but don't feel obligated---I tagged you mostly just to keep in touch about the automatic algorithm selection idea.

I'll let this sit for a week or so before merging it to see if anyone has comments (or wants more time to review it).

There are some things that aren't implemented here that I'll open bugs for.  Some of these bugs will be good for prospective GSoC students:

 * an implementation of weighted learning, so that this can be used with AdaBoost and DecisionStump can be fully replaced and removed
 * a gain penalty for the number of children being created (this is a form of normalization, and it's what C4.5 does)
 * more categorical and numeric split strategies so that we can easily use exact types of other trees
 * support for categorical datasets from the command-line interface
 * benchmarking scripts vs. other libraries
 * a numeric split strategy based on sampling, which could find the split value much faster

Some basic timing information from a quick simulation on my desktop (a Xeon E5-2630v3 with 128GB).  Building a tree on the full covertype dataset with a minimum leaf size of 5 takes 110s (8s to load the data) and yields 93.8% performance on the training set.  The GPL implementation of C5.0, as a comparison, takes 65.8s and yields 98.4% accuracy.  This isn't a completely fair comparison since C5.0 has some extra functionality I think.  So there is some improvement yet to do, but this is a good start.


src/mlpack/methods/decision_tree/decision_tree.hpp
```diff
@@ -0,0 +1,235 @@
...
+  //! Allow access to the numeric split type.
+  typedef NumericSplitType<FitnessFunction> NumericSplit;
+  //! Allow access to the categorical split type.
+  typedef CategoricalSplitType<FitnessFunction> CategoricalSplit;
+
+  /**
+   * Construct the decision tree on the given data and labels.
+   */
+  template<typename MatType>
+  DecisionTree(const MatType& data,
```
zoq: Should we add some more informations about the parameter?

rcurtin: Ah, yes, let me go through and update all of the documentation to make sure it is comprehensive.

Issue #839, opened 2016-12-20
Add tests for GaussianDistribution:Train(observations,  probabilities)


yashu-seth:
Tests exist for GaussianDistribution::Train(observations) but not for GaussianDistribution:Train(observations,  probabilities).

Lakshya1605:
I would like to work on this please could you elaborate on what is expected .Thank You!!!

yashu-seth:
@Lakshya1605 Currently Gaussian distribution can be trained by providing observations and the corresponding probability of those observations. But tests have not been added for this. You can have a look at #733 and #751 which deals with ```GammaDistribtuion::Train(observations, probabilities)``` to get some idea.

Lakshya1605:
@yashu-seth I have gone through #733 and #751  as mentioned and i was thinking if the same method as @rcurtin has mentioned there would work fine.Thank You !!!.

Lakshya1605:
@yashu-seth  I have gone through #733 and #751  as mentioned and i was thinking if the same method as @rcurtin has mentioned there would work fine.Thank You !!!!.

rcurtin:
Hi @Lakshya1605, if you are going to implement tests for this, think about the tests you are writing and ensure that they would be good tests for the Gaussian distribution when it is trained with probabilities for each observation.  Certainly an approach similar to the Gamma distribution tests could work, but it will probably require some adaptation.

Lakshya1605:
@rcurtin @yashu-seth I have been trying the test case as mentioned by "rcurtin" that is `Gaussian::Train(rdata, probabilties)`and `Gaussian::Train(rdata)` should give nearly same parameters( mean and covariance) when all probabilities are 1 . But it turns out that Means are coming the same but not Covariance's .

`BOOST_REQUIRE_CLOSE` in case of covariance is throwing error when "tolerance" is "1e-15" and is only working with "1e-2" .
With Mean it works fine though.Should I proceed with the tests or do you have any suggestions?
Thank You !!!

rcurtin:
1e-2 is 1% tolerance---I think that's fine for the covariance.  Covariance estimation is typically a noisy process.  You could also probably reduce the covariance error by adding more points to the dataset.  One useful thing to check is that the test succeeds with lots of different random seeds.  So you can test that like this: add the line `math::RandomSeed(std::time(NULL))` to the start of the test.  That will set the random seed to the current time.  Then rebuild the package, and run the test many times:

```
[...]

PR #860, opened 2017-01-27
Added TestCase for GaussianDistribution::Train

Lakshya1605:
#839


src/mlpack/tests/distribution_test.cpp
```diff
@@ -390,10 +390,145 @@ BOOST_AUTO_TEST_CASE(GaussianDistributionTrainTest)
...
+{
+  double mean = 5.0;
+  double stddeviation = 2.0;
+
+  //Creates a normal distribution generator.
+  std::default_random_engine generator;
+  generator.seed(std::time(NULL));
+  std::normal_distribution<double> dist(mean, stddeviation);
+  
+  size_t N = 50000;
```
rcurtin: I think only 5k samples should be fine for this test---no need to use more samples than are needed.

src/mlpack/tests/distribution_test.cpp
```diff
@@ -390,10 +390,145 @@ BOOST_AUTO_TEST_CASE(GaussianDistributionTrainTest)
...
+**/
+BOOST_AUTO_TEST_CASE(GaussianDistributionTrainWithProbabilitiesTest)
+{
+  double mean = 5.0;
+  double stddeviation = 2.0;
+
+  //Creates a normal distribution generator.
+  std::default_random_engine generator;
+  generator.seed(std::time(NULL));
+  std::normal_distribution<double> dist(mean, stddeviation);
```
rcurtin: Please don't set the random seed for anything other than testing.  It's important that failing tests be reproducible, and if the failure depends on the random seed, and the seed is set in the test, then it's not easily reproducible.

src/mlpack/tests/distribution_test.cpp
```diff
@@ -390,10 +390,145 @@ BOOST_AUTO_TEST_CASE(GaussianDistributionTrainTest)
...
+  
+  size_t N = 50000;
+  size_t d =1;
+  
+  arma::mat rdata(d, N);
+  
+  for (size_t i = 0; i < d; i++)
+    for (size_t j = 0; j < N; j++)
+      rdata(i,j) = dist(generator);
```
rcurtin: You can just use `GaussianDistribution::Random()` here.

src/mlpack/tests/distribution_test.cpp
```diff
@@ -390,10 +390,145 @@ BOOST_AUTO_TEST_CASE(GaussianDistributionTrainTest)
...
+
+  //Creates a uniform distribution generator
+  std::uniform_real_distribution<double> prob(0, 1);
+  arma::vec probabilities(N);
+  
+  for (size_t i = 0; i < N; i++)
+    probabilities(i) = prob(generator);
```
rcurtin: You can just use `math::Random()` here.

src/mlpack/tests/distribution_test.cpp
```diff
@@ -390,10 +390,145 @@ BOOST_AUTO_TEST_CASE(GaussianDistributionTrainTest)
...
+  //fits data with only data
+  GaussianDistribution guDist;
+  guDist.Train(rdata);
+
+  //fits result with data and each probability as 1
+  GaussianDistribution guDist2;
+  guDist2.Train(rdata, probabilities);
+
+  BOOST_REQUIRE_CLOSE(guDist.Mean()[0], guDist2.Mean()[0], 1e-15);
+  BOOST_REQUIRE_CLOSE(guDist.Covariance()[0], guDist2.Covariance()[0], 1e-2);
```
rcurtin: I looked into why a larger tolerance is necessary here.  It has to do with how the covariance is normalized in the case with and without probabilities.  So 1e-2 is fine.

Lakshya1605:
@rcurtin  I have modified the code as specified .I have also tried to see to that my code matches with the Style Guide ,if something is wrong please let me know.
Thank You !!!

rcurtin:
Ok, looks better.  There are still some style issues but I will fix them after merge.  Would you like to add your name to the list of contributors in `src/mlpack/core.hpp` and `COPYRIGHT.txt`?

Lakshya1605:
Yes i would love to add my name.Thank You !!!

PR #864, opened 2017-02-02
Test travis trusty image

zoq:
This might improve the travis build time, let's see.


zoq:
Improved build time: Elapsed time **36 min 27 sec** using the **trusty image** vs Elapsed time **49 min 5 sec** using the travis **precise image**.

PR #867, opened 2017-02-10
Add Softplus activation function

ivmarkp:
All tests pass locally. Please review.

Now, working on adding ELU layer. Will open a PR ASAP.

### DIFF
--- a/src/mlpack/methods/ann/activation_functions/softplus_function.hpp
+++ b/src/mlpack/methods/ann/activation_functions/softplus_function.hpp
@@ -13,8 +13,7 @@
  *   Nadeau, Claude and Garcia, Rene},
  *   title = {Incorporating Second-Order Functional Knowledge for Better Option
  *   Pricing},
- *   year = {2001},
- *   url = {http://www.iro.umontreal.ca/~lisa/pointeurs/nips00-convex.ps}
+ *   year = {2001}
  * }
  * @endcode
  *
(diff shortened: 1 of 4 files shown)


src/mlpack/methods/ann/activation_functions/softplus_function.hpp
```diff
@@ -0,0 +1,114 @@
+/**
+ * @file softplus_function.hpp
+ * @author Vivek Pal
+ *
+ * Definition and implementation of the softplus function.
+ *
```
zoq: Can you cite the "Incorporating Second-Order Functional Knowledge for Better Option Pricing" by C. Dugas et al. paper that introduced the softplus function? Take a look at the softsign function header for an example.

ivmarkp: Sure. I found the citation [here](http://www.iro.umontreal.ca/~lisa/publications2/index.php/export/publication/67/bibtex). Is that format correct? I think may be "abstract" field can be dropped from it for our purposes.

zoq: Sounds good, you can also drop url, topics and cat.

ivmarkp: Okay, pushing the commit in a minute.

zoq: The required fields of INPROCEEDINGS are: author, title, booktitle, year so, we should make sure there is a booktitle.

ivmarkp: Yes, sorry I missed that. Booktitle is "Advances in Neural Information Processing Systems".

zoq:
Thanks again for the great contribution, looking forward to take a look at the other PR as well.

ivmarkp:
Thanks, will address your review comments on the other PR as soon as I can.

PR #537, opened 2016-03-02
Removed trailing underscores in header guards. Resolved issue #533

Falit:
Deletion of (leading)underscores in macros header guards as they being [deprecated now](https://www.securecoding.cert.org/confluence/display/cplusplus/DCL51-CPP.+Do+not+declare+or+define+a+reserved+identifier#DCL51-CPP.Donotdeclareordefineareservedidentifier-NoncompliantCodeExample%28HeaderGuard%29)
#533

PR #871, opened 2017-02-14
Add DBSCAN implementation

rcurtin:
One of my coworkers wanted to do DBSCAN, so I implemented it in C++ using mlpack.  Then it turned out I had misunderstood his problem and DBSCAN wasn't useful, but the implementation was still there.  Seems like it would be useful to just go ahead and merge it into mlpack. :)

This has a few improvements from @s1998 also.


src/mlpack/methods/dbscan/dbscan.hpp
```diff
@@ -0,0 +1,126 @@
+/**
+ * @file dbscan.hpp
+ * @author Ryan Curtin
+ *
+ * An implementation of the DBSCAN clustering method, which is flexible enough
+ * to support other algorithms for finding nearest neighbors.
+ */
+#ifndef __MLPACK_METHODS_DBSCAN_DBSCAN_HPP
```
zoq: We should remove the underscore here, because leading or trailing underscores are deprecated? Here is the PR https://github.com/mlpack/mlpack/pull/537

src/mlpack/methods/dbscan/dbscan.hpp
```diff
@@ -0,0 +1,126 @@
...
+class DBSCAN
+{
+ public:
+  /**
+   * Construct the DBSCAN object with the given parameters.
+   *
+   * @param epsilon Size of range query.
+   * @param minPoints Minimum number of points for each cluster.
+   */
+  DBSCAN(const double epsilon,
```
zoq: Should we add a parameter description for ``rangeSearch`` and ``pointSelector``?

src/mlpack/methods/dbscan/dbscan.hpp
```diff
@@ -0,0 +1,126 @@
...
+   * 
+   *
+   * @param MatType Type of matrix (arma::mat or arma::sp_mat).
+   * @param data Dataset to cluster.
+   * @param unvisited Remembers if a point has been visited.
+   * @param index Index of point to be visited now.
+   * @param assignments Vector to store cluster assignments.
+   * @param currentCluster Index of cluster which will be  
+   *                       assigned to points in current cluster.
+   * @param neighbor Matrix containing list of neighbors for each point
```
zoq: I guess it should be ``@param neighbors Matrix containing list of neighbors for each point...``

src/mlpack/methods/dbscan/dbscan_impl.hpp
```diff
@@ -0,0 +1,183 @@
...
+  //and their distances.
+  std::vector<std::vector<size_t>> neighbors;
+  std::vector<std::vector<double>> distances;
+  Log::Info << "Performing range search." << std::endl;
+  rangeSearch.Train(data);
+  rangeSearch.Search(data, math::Range(0.0, epsilon), neighbors, distances);
+  Log::Info << "Range search complete." << std::endl;
+
+  // Initialize to all true; false means it's been visited.
+  boost::dynamic_bitset<> unvisited(data.n_cols);
```
zoq: Nice idea using boost::dynamic_bitset.

rcurtin:
Thanks, it's really helpful to have a second set of eyes to catch the little things. :)

If you don't have any more comments I'll go ahead and merge after the build succeeds.

zoq:
I've seen some really minor style issue in the form you fixed in a5f92bbf900, besides that nothing else from me.

PR #872, opened 2017-02-15
Implementation of PerametricReLU activation function in mlpack

prasanna08:
This PR implements PReLU activation function in mlpack ann layer.

### DIFF
--- a/src/mlpack/methods/ann/layer/perametric_relu.hpp
+++ b/src/mlpack/methods/ann/layer/perametric_relu.hpp
@@ -1,0 +1,212 @@
+/**
+ * @file perametric_relu.hpp
+ * @author Prasanna Patil
+ *
+ * Definition of PerametricReLU layer first introduced in the,
+ * Kaiming He, Xiangyu Zhang, Shaoqing, Ren Jian Sun,
+ * "Delving Deep into Rectifiers:
+ * Surpassing Human-Level Performance on ImageNet Classification", 2014
+ *
+ * mlpack is free software; you may redistribute it and/or modify it under the
+ * terms of the 3-clause BSD license.  You should have received a copy of the
+ * 3-clause BSD license along with mlpack.  If not, see
+ * http://www.opensource.org/licenses/BSD-3-Clause for more information.
+ */
+#ifndef MLPACK_METHODS_ANN_LAYER_PERAMETRICRELU_HPP
+#define MLPACK_METHODS_ANN_LAYER_PERAMETRICRELU_HPP
+
+#include <mlpack/prereqs.hpp>
+
+namespace mlpack {
+namespace ann /** Artificial Neural Network. */ {
+
+/**
+ * The PerametricReLU activation function, defined by (where alpha is trainable)
+ *
+ * @f{eqnarray*}{
+ * f(x) &=& \max(x, alpha*x) \\
+ * f'(x) &=& \left\{
+ *   \begin{array}{lr}
+ *     1 & : x > 0 \\
+ *     alpha & : x \le 0
+ *   \end{array}
+ * \right.
+ * @f}
+ *
+ * @tparam InputDataType Type of the input data (arma::colvec, arma::mat,
+ *         arma::sp_mat or arma::cube).
+ * @tparam OutputDataType Type of the output data (arma::colvec, arma::mat,
+ *         arma::sp_mat or arma::cube).
+ */
+template <
+    typename InputDataType = arma::mat,
+    typename OutputDataType = arma::mat
+>
+class PerametricReLU
+{
+ public:
+  /**
+   * Create the PerametricReLU object using the specified parameters.
+   * The non zero gradient can be adjusted by specifying tha parameter
+   * alpha in the range 0 to 1. Default (alpha = 0.03). This parameter
+   * is trainable.
+   *
+   * @param alpha Non zero gradient
+   */
+  PerametricReLU(const double user_alpha = 0.03);
+
+  /*
+   * Reset the layer parameter.
+   */
+  void Reset();
+
+  /**
+   * Ordinary feed forward pass of a neural network, evaluating the function
+   * f(x) by propagating the activity forward through f.
+   *
+   * @param input Input data used for evaluating the specified function.
+   * @param output Resulting output activation.
+   */
+  template<typename InputType, typename OutputType>
+  void Forward(const InputType&& input, OutputType&& output);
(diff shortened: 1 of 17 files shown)


src/mlpack/methods/ann/layer/perametric_relu.hpp
```diff
@@ -0,0 +1,212 @@
...
+ * @tparam InputDataType Type of the input data (arma::colvec, arma::mat,
+ *         arma::sp_mat or arma::cube).
+ * @tparam OutputDataType Type of the output data (arma::colvec, arma::mat,
+ *         arma::sp_mat or arma::cube).
+ */
+template <
+    typename InputDataType = arma::mat,
+    typename OutputDataType = arma::mat
+>
+class PerametricReLU
```
zoq: Do you think PReLU might be a better class name?

src/mlpack/methods/ann/layer/perametric_relu_impl.hpp
```diff
@@ -0,0 +1,83 @@
...
+#include "perametric_relu.hpp"
+
+namespace mlpack {
+namespace ann /** Artificial Neural Network. */ {
+
+template<typename InputDataType, typename OutputDataType>
+PerametricReLU<InputDataType, OutputDataType>::PerametricReLU(
+    const double user_alpha) : user_alpha(user_alpha)
+{
+  alpha.set_size(1,1);
```
zoq: Instead of using a arma::mat with size 1x1 we could also use double.

src/mlpack/methods/ann/layer/perametric_relu.hpp
```diff
@@ -0,0 +1,212 @@
...
+ public:
+  /**
+   * Create the PerametricReLU object using the specified parameters.
+   * The non zero gradient can be adjusted by specifying tha parameter
+   * alpha in the range 0 to 1. Default (alpha = 0.03). This parameter
+   * is trainable.
+   *
+   * @param alpha Non zero gradient
+   */
+  PerametricReLU(const double user_alpha = 0.03);
```
zoq: I would just go with alpha and use the constructor initialization list to initialize alpha.

src/mlpack/methods/ann/layer/perametric_relu.hpp
```diff
@@ -0,0 +1,212 @@
...
+  /**
+   * Computes the Perametric ReLU function using a dense matrix as input.
+   *
+   * @param x Input data.
+   * @param y The resulting output activation.
+   */
+  template<typename eT>
+  void Fn(const arma::Mat<eT>& x, arma::Mat<eT>& y)
+  {
+    y = arma::max(x, alpha(0) * x);
```
zoq: We could avoid the computation of ``alpha * x`` if  ``x > 0``that might slightly improve the performance, what do you think?

src/mlpack/methods/ann/layer/perametric_relu_impl.hpp
```diff
@@ -0,0 +1,83 @@
...
+void PerametricReLU<InputDataType, OutputDataType>::Gradient(
+    const arma::Mat<eT>&& input, arma::Mat<eT>&& error,
+    arma::Mat<eT>&& gradient)
+{
+  if (gradient.n_elem == 0) {
+    gradient = arma::zeros<arma::mat>(1, 1);
+  }
+
+  arma::mat zeros = arma::zeros<arma::mat>(input.n_rows, input.n_cols);
+  gradient(0) = arma::accu(error % arma::min(zeros, input)) / input.n_cols;
```
zoq: Can you make sure that the activation function also works with e.g. input = arma::mat("0.2 0.3 -0.5").

src/mlpack/methods/ann/layer/perametric_relu.hpp
prasanna08: Yeah no problem.

src/mlpack/methods/ann/layer/perametric_relu_impl.hpp
prasanna08: Alpha is a trainable. So it is passed when Parameters() method is called and WeightSetVisitor, WeightSizeVisitor etc.. Parameters() to be an arma::mat. So how to pass double rather than arma::mat?

We can create some other matrix (i.e. weigh) then create a double pointer (or reference) which references to first element of weight but I don't know if that would work or not. What do you propose?

src/mlpack/methods/ann/layer/perametric_relu.hpp
prasanna08: Yeah that is possible. I would change implementation accordingly.

src/mlpack/methods/ann/layer/perametric_relu_impl.hpp
prasanna08: I have tested for Forward, Backward and Gradient functions in test file. What test are you suggesting me here ?

src/mlpack/methods/ann/layer/perametric_relu_impl.hpp
zoq: You are right, in that case I would go with your solution.

src/mlpack/methods/ann/layer/perametric_relu_impl.hpp
zoq: Right now you test the layer with an input of size 1x1 but it would be great if you could test it with an input nxm where n is > 1. The JacobianHardTanHLayerTest test case in ann_layer_test.cpp is one test that you could adapt.

zoq: Ah, I missed the PerametricReLUFunctionTest test case, nevermind.

prasanna08: Alright. I will push new commit with suggested changes.

nish21:
Also, it should be `parametric` and not `perametric` throughout.

prasanna08:
Hi @nish21,

Yes I spelled it incorrectly. Thanks for pointing it.

zoq:
Thanks for looking into the issue. Would you like to add your name to the list of contributors in src/mlpack/core.hpp and COPYRIGHT.txt before I merge the code?

prasanna08:
@zoq done. Thanks.

zoq:
Merged, thanks again for the great contribution.

PR #873, opened 2017-02-15
Implement maxout layer in mlpack.

prasanna08:
This PR adds maxout layer in mlpack ann module.

### DIFF
--- a/src/mlpack/methods/ann/layer/maxout_impl.hpp
+++ b/src/mlpack/methods/ann/layer/maxout_impl.hpp
@@ -2,8 +2,10 @@
  * @file maxout_impl.hpp
  * @author Prasanna Patil
  *
- * Implementation of the Maxout layer class also known as fully-connected layer
- * or affine transformation.
+ * Definition of Maxout layer first introduced in the Maxout Networks,
+ * Ian J. Goodfellow, David Warde-Farley, Mehdi Mirza, Aaron Courville,
+ * Yoshua Bengio
+ * "Maxout Networks", 2013
  *
  * mlpack is free software; you may redistribute it and/or modify it under the
  * terms of the 3-clause BSD license.  You should have received a copy of the
--- a/src/mlpack/methods/ann/layer/maxout_impl.hpp
+++ b/src/mlpack/methods/ann/layer/maxout_impl.hpp
@@ -53,20 +53,13 @@ template<typename eT>
 void Maxout<InputDataType, OutputDataType>::Forward(
     const arma::Mat<eT>&& input, arma::Mat<eT>&& output)
 {
-  if (hiddenParameter.n_elem == 0) {
-    hiddenParameter = arma::zeros<arma::mat>(outSize * hiddenSize,
-        input.n_cols);
-  }
-
-  if (output.n_elem == 0) {
+  if (output.n_elem == 0)
     output = arma::zeros<arma::mat>(outSize, input.n_cols);
-  }
 
   hiddenParameter = (weight * input) + bias;
 
-  for (int i = 0, j = 0; j < outSize; i+=hiddenSize, j++) {
+  for (int i = 0, j = 0; j < outSize; i += hiddenSize, j++)
     output.row(j) = arma::max(hiddenParameter.rows(i, i + hiddenSize - 1), 0);
-  }
 }
 
 template<typename InputDataType, typename OutputDataType>
@@ -78,10 +71,12 @@ void Maxout<InputDataType, OutputDataType>::Backward(
       hiddenParameter.n_cols);
   arma::mat hiddenParameterSubset;
 
-  for (int i = 0, j = 0; j < outSize; i+=hiddenSize, j++) {
+  for (int i = 0, j = 0; j < outSize; i += hiddenSize, j++)
+  {
     hiddenParameterSubset = hiddenParameter.rows(i, i + hiddenSize - 1);
     
-    for (int k = 0; k < hiddenParameterSubset.n_cols; k++) {
+    for (int k = 0; k < hiddenParameterSubset.n_cols; k++)
+    {
       int ind = arma::as_scalar(arma::find(arma::max(
           hiddenParameterSubset.col(k)) == hiddenParameterSubset.col(k)));
       hiddenGradient(i + ind, k) = gy(j, k);
@@ -102,10 +97,12 @@ void Maxout<InputDataType, OutputDataType>::Gradient(
       hiddenParameter.n_cols);
   arma::mat hiddenParameterSubset;
 
-  for (int i = 0, j = 0; j < outSize; i+=hiddenSize, j++) {
+  for (int i = 0, j = 0; j < outSize; i += hiddenSize, j++)
+  {
     hiddenParameterSubset = hiddenParameter.rows(i, i + hiddenSize - 1);
     
-    for (int k = 0; k < hiddenParameterSubset.n_cols; k++) {
+    for (int k = 0; k < hiddenParameterSubset.n_cols; k++)
+    {
       int ind = arma::as_scalar(arma::find(arma::max(
           hiddenParameterSubset.col(k)) == hiddenParameterSubset.col(k)));
       hiddenGradient(i + ind, k) = error(j, k);
--- a/src/mlpack/methods/ann/layer/maxout_impl.hpp
+++ b/src/mlpack/methods/ann/layer/maxout_impl.hpp
@@ -56,10 +56,21 @@ void Maxout<InputDataType, OutputDataType>::Forward(
   if (output.n_elem == 0)
     output = arma::zeros<arma::mat>(outSize, input.n_cols);
 
-  hiddenParameter = (weight * input) + bias;
+  if (maxoutIndices.n_elem == 0)
+    maxoutIndices = arma::zeros<arma::mat> (outSize, input.n_cols);
 
-  for (int i = 0, j = 0; j < outSize; i += hiddenSize, j++)
-    output.row(j) = arma::max(hiddenParameter.rows(i, i + hiddenSize - 1), 0);
+  arma::mat hiddenParameter = (weight * input) + bias;
+  arma::mat hiddenParameterSubset;
+
+  for (int i = 0, j = 0; j < outSize; i += hiddenSize, j++) {
+    hiddenParameterSubset = hiddenParameter.submat(i, 0, i + hiddenSize - 1, 0);
+    output.row(j) = arma::max(hiddenParameterSubset, 0);
+
+    for (int k = 0; k < output.n_cols; k++) {
+      maxoutIndices(j, k) = arma::as_scalar(
+          arma::find(hiddenParameterSubset == output(j, k)));
+    }
+  }
 }
 
 template<typename InputDataType, typename OutputDataType>
@@ -67,18 +78,14 @@ template<typename eT>
 void Maxout<InputDataType, OutputDataType>::Backward(
     const arma::Mat<eT>&& /* input */, arma::Mat<eT>&& gy, arma::Mat<eT>&& g)
 {
-  arma::mat hiddenGradient = arma::zeros<arma::mat>(hiddenParameter.n_rows,
-      hiddenParameter.n_cols);
-  arma::mat hiddenParameterSubset;
+  arma::mat hiddenGradient = arma::zeros<arma::mat>(outSize * hiddenSize,
+      maxoutIndices.n_cols);
 
   for (int i = 0, j = 0; j < outSize; i += hiddenSize, j++)
   {
-    hiddenParameterSubset = hiddenParameter.rows(i, i + hiddenSize - 1);
-    
-    for (int k = 0; k < hiddenParameterSubset.n_cols; k++)
+    for (int k = 0; k < maxoutIndices.n_cols; k++)
     {
-      int ind = arma::as_scalar(arma::find(arma::max(
-          hiddenParameterSubset.col(k)) == hiddenParameterSubset.col(k)));
+      int ind = maxoutIndices(j, k);
       hiddenGradient(i + ind, k) = gy(j, k);
     }
   }
@@ -93,18 +100,14 @@ void Maxout<InputDataType, OutputDataType>::Gradient(
     arma::Mat<eT>&& error,
     arma::Mat<eT>&& gradient)
 {
-  arma::mat hiddenGradient = arma::zeros<arma::mat>(hiddenParameter.n_rows,
-      hiddenParameter.n_cols);
-  arma::mat hiddenParameterSubset;
+  arma::mat hiddenGradient = arma::zeros<arma::mat>(outSize * hiddenSize,
+      maxoutIndices.n_cols);
 
   for (int i = 0, j = 0; j < outSize; i += hiddenSize, j++)
-  {
-    hiddenParameterSubset = hiddenParameter.rows(i, i + hiddenSize - 1);
-    
-    for (int k = 0; k < hiddenParameterSubset.n_cols; k++)
+  { 
+    for (int k = 0; k < maxoutIndices.n_cols; k++)
     {
-      int ind = arma::as_scalar(arma::find(arma::max(
-          hiddenParameterSubset.col(k)) == hiddenParameterSubset.col(k)));
+      int ind = maxoutIndices(j, k);
       hiddenGradient(i + ind, k) = error(j, k);
     }
   }
--- a/src/mlpack/methods/ann/layer/maxout_impl.hpp
+++ b/src/mlpack/methods/ann/layer/maxout_impl.hpp
@@ -1,0 +1,135 @@
+/**
+ * @file maxout_impl.hpp
+ * @author Prasanna Patil
+ *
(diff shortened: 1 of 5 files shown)


src/mlpack/methods/ann/layer/maxout_impl.hpp
```diff
@@ -0,0 +1,137 @@
...
+  bias = arma::mat(weights.memptr() + weight.n_elem,
+      outSize * hiddenSize, 1, false, false);
+}
+
+template<typename InputDataType, typename OutputDataType>
+template<typename eT>
+void Maxout<InputDataType, OutputDataType>::Forward(
+    const arma::Mat<eT>&& input, arma::Mat<eT>&& output)
+{
+  if (hiddenParameter.n_elem == 0) {
```
zoq: Minor style issue please put the opening bracket on a new line; For more information: https://github.com/mlpack/mlpack/wiki/DesignGuidelines#brace-placement

I tend to be very  picky about code but I am not mean :)

src/mlpack/methods/ann/layer/maxout_impl.hpp
```diff
@@ -0,0 +1,137 @@
...
+    hiddenParameter = arma::zeros<arma::mat>(outSize * hiddenSize,
+        input.n_cols);
+  }
+
+  if (output.n_elem == 0) {
+    output = arma::zeros<arma::mat>(outSize, input.n_cols);
+  }
+
+  hiddenParameter = (weight * input) + bias;
```
zoq: Do we have to initialize the hiddenParameter in line 57 if we just assign the parameter here?

src/mlpack/methods/ann/layer/maxout_impl.hpp
```diff
@@ -0,0 +1,137 @@
...
+
+  for (int i = 0, j = 0; j < outSize; i+=hiddenSize, j++) {
```
zoq: We usually use size_t as index type, also another picky comment can you put a space between i and i+= hiddenSize.

src/mlpack/tests/feedforward_network_test.cpp
```diff
@@ -410,4 +410,134 @@ BOOST_AUTO_TEST_CASE(DropConnectNetworkTest)
...
+  for (size_t i = 0; i < dataset.n_cols; ++i)
+    dataset.col(i) /= norm(dataset.col(i), 2);
+
+  arma::mat labels = arma::zeros(1, dataset.n_cols);
+  labels.submat(0, labels.n_cols / 2, 0, labels.n_cols - 1).fill(1);
+  labels += 1;
+
+  // Maxout neural net with logistic activation function.
+  BuildMaxoutNetwork<>
+      (dataset, labels, dataset, labels, 2, 5, 10, 50, 0.2);
```
zoq: I didn't expected an error of 0.2 using 50 iterations, nice.

src/mlpack/tests/feedforward_network_test.cpp
```diff
@@ -410,4 +410,134 @@ BOOST_AUTO_TEST_CASE(DropConnectNetworkTest)
...
+                        const size_t maxEpochs,
+                        const double classificationErrorThreshold)
+{
+ /*
+  *  Construct a feed forward network with trainData.n_rows input nodes,
+  *  maxoutUnitSize hidden nodes and trainLabels.n_rows output nodes. The
+  *  network struct that looks like:
+  *
+  *  Input         Maxout         Output
+  *  Layer         Layer          Layer
```
zoq: Shouldn't this be input -> maxout -> linear -> output?

src/mlpack/methods/ann/layer/maxout_impl.hpp
prasanna08: No that is not necessary. We can do direct assignment.

src/mlpack/tests/feedforward_network_test.cpp
prasanna08: No. Actually I transfer output from maxout layer to a linear layer and this linear layer's output is output probability (on which log-likelihood with softmax is applied). So the final linear layer is actually output layer.

src/mlpack/methods/ann/layer/maxout_impl.hpp
prasanna08: Sure. Design guidelines are important. Otherwise code looks messy.

src/mlpack/methods/ann/layer/maxout_impl.hpp
```diff
@@ -0,0 +1,134 @@
...
+void Maxout<InputDataType, OutputDataType>::Forward(
+    const arma::Mat<eT>&& input, arma::Mat<eT>&& output)
+{
+  if (output.n_elem == 0)
+    output = arma::zeros<arma::mat>(outSize, input.n_cols);
+
+  hiddenParameter = (weight * input) + bias;
+
+  for (int i = 0, j = 0; j < outSize; i += hiddenSize, j++)
+    output.row(j) = arma::max(hiddenParameter.rows(i, i + hiddenSize - 1), 0);
```
zoq: Do you think using submat (arma::mat(hiddenParameter.submat(), 0) is faster as iterating over the rows?

src/mlpack/methods/ann/layer/maxout_impl.hpp
```diff
@@ -0,0 +1,134 @@
...
+      hiddenParameter.n_cols);
+  arma::mat hiddenParameterSubset;
+
+  for (int i = 0, j = 0; j < outSize; i += hiddenSize, j++)
+  {
+    hiddenParameterSubset = hiddenParameter.rows(i, i + hiddenSize - 1);
+    
+    for (int k = 0; k < hiddenParameterSubset.n_cols; k++)
+    {
+      int ind = arma::as_scalar(arma::find(arma::max(
```
zoq: We should specify the number of indices find returns.

src/mlpack/methods/ann/layer/maxout_impl.hpp
prasanna08: I can't say. I don't have much experience with arma. If you think that's better way then I will change implementation.

src/mlpack/methods/ann/layer/maxout_impl.hpp
prasanna08: It returns only one index. Basically what I am trying is finding the hidden unit that had maximum activation because the error should be backpropagated to this hidden unit only. So I am finding index of this maximum activation in for all test cases and backpropagating error for them.

I will add comment describing this. Also would you confirm about this backpropagation? I am not sure whether I wrote it correctly or not.

src/mlpack/methods/ann/layer/maxout_impl.hpp
zoq: I'm not sure, it's faster, do you think we can write a simple benchmark to test that out? Something like that:

```
arma::mat input = arma::randu(20000, 1);
arma::mat output;

auto start = chrono::steady_clock::now();

layer.Forward(std::move(input), std::move(output));

auto end = chrono::steady_clock::now();
auto diff = end - start;
cout << "maxout: " << chrono::duration <double, milli> (diff).count() << " ms" << endl;
```

src/mlpack/methods/ann/layer/maxout_impl.hpp
zoq: It looks right, it's basically what you do for MaxPooling, where I also used the same approach:

```
class MaxPoolingRule
{
 public:
  /*
   * Return the maximum value within the receptive block.
   *
   * @param input Input used to perform the pooling operation.
   */
  template<typename MatType>
  size_t Pooling(const MatType& input)
  {
    return arma::as_scalar(arma::find(input.max() == input, 1));
  }
};
```

But I think I can construct a case where max isn't just a single element e.g. ``arma::max({0, 0, 0})`` or ``arma::max({0.5, 0.5, 0})`` here max returns more than one element, which isn't a problem, we can just use the first max. Besides, specifying the number would probably result in a speedup since you can break once you found the first occurrences of the maximum. Let me know what you think.

src/mlpack/methods/ann/layer/maxout_impl.hpp
prasanna08: Yeah sure. I will test it and report back.

src/mlpack/methods/ann/layer/maxout_impl.hpp
prasanna08: Yup. You are correct. Specifying number directly would result in speedup. Also, here we don't actually want the maximum value. The point of storing hiddenParameter is to get index of max element at the time of backprop and gradient. I think we can optimize this by directly storing indices of max element during forward pass (where we have to find max element for output). Then later in backprop or gradient we can directly utilize these indices. What do you think ?

src/mlpack/methods/ann/layer/maxout_impl.hpp
prasanna08: I tested and both are performing equally well.
Data I observed:

<table>
<tr>
<th>method</th>
<th>in_size</th>
<th>hidden_size</th>
<th>out_size</th>
<th>run</th>
<th>time</th>
</tr>
<tr>
<td>rows</td>
<td>20000</td>
<td>7</td>
<td>5</td>
<td>1st</td>
<td>1.54378 ms</td>
</tr>
</tr>

<tr>
<td>rows</td>
<td>20000</td>
<td>7</td>
<td>5</td>
<td>2nd</td>
<td>0.783643 ms</td>
</tr>


<tr>
<td>submat</td>
<td>20000</td>
<td>7</td>
<td>5</td>
<td>1st</td>
<td>0.772132 ms</td>
</tr>

<tr>
<td>submat</td>
<td>20000</td>
<td>7</td>
<td>5</td>
<td>2nd</td>
<td>0.884035 ms</td>
</tr>

<tr>
<td>rows</td>
<td>20000</td>
<td>100</td>
<td>20</td>
<td>1st</td>
<td>33.3094 ms</td>
</tr>

<tr>
<td>rows</td>
<td>20000</td>
<td>100</td>
<td>20</td>
<td>2nd</td>
<td>33.1371 ms</td>
</tr>

<tr>
<td>submat</td>
<td>20000</td>
<td>100</td>
<td>20</td>
<td>1st</td>
<td>36.081 ms</td>
</tr>

<tr>
<td>submat</td>
<td>20000</td>
<td>100</td>
<td>20</td>
<td>2nd</td>
<td>32.0741 ms</td>
</tr>
</table>

So according to my observation its a tie. So its your call.

src/mlpack/methods/ann/layer/maxout_impl.hpp
prasanna08: Hi @zoq I bench marked the maxout code with 3 different implementations which are:
**maxoutIndices**: Storing index of maximum hidden units and then using it.
**maxoutValues**: Storing value of maximum hidden unit and then using it (basically storing output of maxout layer).
**hiddenValues**: current approach I have implemented.

I tested it by training end to end model for all 3 setups (since this changes affected all forward backward and gradient passes). 

<table>
<tr>
<th>method</th>
<th>iterations</th>
<th>time</th>
</tr>

<tr>
<td>maxoutIndices</td>
<td>70</td>
<td>55100 ms </td>
</tr>
<tr>
<td>maxoutIndices</td>
<td>7</td>
<td>4552.45 ms</td>
</tr>
<tr>
<td>maxoutValues</td>
<td>70</td>
<td>57945.5 ms</td>
</tr>
<tr>
<td>maxoutValues</td>
<td>7</td>
<td>4346.81 ms</td>
</tr>
<tr>
<td>hiddenParameters</td>
<td>70</td>
<td>52324.9 ms</td>
</tr>
<tr>
<td>hiddenParameters</td>
<td>7</td>
<td>4373.22 ms</td>
</tr>
</table>

Honestly I expected slight performance boost from maxoutIndices and maxoutValues method but from observations it seems that hiddenParameters outperforms both of them. I am confused by this result. So I leave this decision on you, too.

zoq: I also thought that reusing the indices from the Forward pass would result in an speedup, can you share the code for the other two options you tested out?

zoq: Btw. I like the way you approach this, it's easier to make a decision based on real numbers.

src/mlpack/methods/ann/layer/maxout_impl.hpp
zoq: I agree, I think in that case we can go with the solution you like.

src/mlpack/methods/ann/layer/maxout_impl.hpp
prasanna08: I used following code in **maxoutIndices**

in forward pass I used following loop.
```
for (int i = 0, j = 0; j < outSize; i += hiddenSize, j++) {
    hiddenParameterSubset = hiddenParameter.submat(i, 0, i + hiddenSize - 1, 0);
    output.row(j) = arma::max(hiddenParameterSubset, 0);

    for (int k = 0; k < output.n_cols; k++) {
      maxoutIndices(j, k) = arma::as_scalar(
          arma::find(hiddenParameterSubset == output(j, k)));
    }
```

Calculating index in backward and gradient method.
```
int ind = maxoutIndices(j, k);
```

Similarly in **maxoutValue**

forward:
```
for (int i = 0, j = 0; j < outSize; i += hiddenSize, j++) {
    hiddenParameterSubset = hiddenParameter.submat(i, 0, i + hiddenSize - 1, 0);
    output.row(j) = arma::max(hiddenParameterSubset, 0);
    maxoutValues.row(j) = arma::max(hiddenParameterSubset, 0);
}
```

backward and gradient:
```
int ind = arma::as_scalar(arma::find(maxoutValues(j, k) == hiddenParameterSubset.col(k)));
```

Well initially I was doing timing calculations to measure how much boost I am able to get from these methods but then I did not get any improvement so I ran tests again and again for 3 times and I still got same results. Therefore I posted them here because I am confused from these results.

Also I used thyroid dataset for bench marking. I don't think that these test results are influenced by dataset, are they ?

src/mlpack/methods/ann/layer/maxout_impl.hpp
prasanna08: I would go with submat solution then.

src/mlpack/methods/ann/layer/maxout_impl.hpp
prasanna08: So basically **maxoutIndices** and **maxoutValues** are similar except where index value is getting calculated. maxoutIndices calculates indices in forward pass (and stores it) and maxoutValues calculates them in Backward and Gradient pass (latter calculates twice but former calculates for once).

prasanna08: Hi @zoq, have you decided which implementation to go with? Its the final remaining thing so that this PR can be merged.

Thanks

zoq: Sorry for the slow response, I would go with calculating the maxoutIndices values once in the Forward.

prasanna08: No worries. I understand that you have a buys schedule. I will push new commit soon.

Thanks

src/mlpack/tests/feedforward_network_test.cpp
```diff
@@ -410,4 +410,134 @@ BOOST_AUTO_TEST_CASE(DropConnectNetworkTest)
...
+  }
+
+  double classificationError = 1 - double(error) / testData.n_cols;
+  BOOST_REQUIRE_LE(classificationError, classificationErrorThreshold);
+}
+
+/**
+ * Train the maxout network on a larger dataset.
+ */
+BOOST_AUTO_TEST_CASE(MaxoutNetworkTest)
```
zoq: Can you tell us something about the runtime of this test? ``time bin/mlpack_test -t FeedForwardNetworkTest/MaxoutNetworkTest``

prasanna08: The output of time was following:
```
real	0m45.613s
user	0m54.128s
sys	1m57.040s
```

Also I was wondering if there is a way to run only selected tests. Thanks for the info.

prasanna08: Hi @zoq, is there anything remaining here?

zoq: Do you think we could slightly decrease the number of training iterations?

prasanna08: Hi @zoq I decresed number of iterations by fraction of 10 (7 and 5 respectively) and test is successful with less iterations. Here is output of `time`.
```
real	0m10.143s
user	0m11.111s
sys	0m24.168s
```

zoq: That looks great, can you push the changes?

prasanna08: @zoq done.

zoq:
I'd like to test the layer with a couple models before I'll merge it in, but besides that it looks good. I'll see if I can test the layer with the models in the next days.

PR #878, opened 2017-02-21
Gaussian_Init

kris-singh:
This add gaussian initialisation. To the weights. I am not sure which test should be added so any suggestions on that would be great.

### DIFF
--- a/src/mlpack/methods/ann/init_rules/gaussian_init.hpp
+++ b/src/mlpack/methods/ann/init_rules/gaussian_init.hpp
@@ -15,6 +15,9 @@
 #define MLPACK_METHODS_ANN_INIT_RULES_GAUSSIAN_INIT_HPP
 
 #include <mlpack/prereqs.hpp>
+#include <mlpack/core/math/random.hpp>
+
+using namespace mlpack::math;
 
 namespace mlpack {
 namespace ann /** Artificial Neural Network. */ {
@@ -55,20 +58,14 @@ class GaussianInitialization
    */
   void Initialize(arma::mat& W,
                   const size_t rows,
-                  const size_t cols,
-                  const size_t seed =21)
+                  const size_t cols)
   {
-    std::random_device rd;
-    std::mt19937 gen(rd());
-    std::srand(seed);
     W = arma::mat(rows, cols);
     double m;
     double v;
     m = mean;
     v = variance;
-    std::normal_distribution<> d(m, v);
-    //distribution::GaussianDistribution dist(m, v);
-    W.imbue( [&]() { return arma::as_scalar(d(gen)); } );
+    W.imbue( [&]() { return arma::as_scalar(RandNormal(m, v)); } );
   }
 
   /**
--- a/src/mlpack/methods/ann/init_rules/gaussian_init.hpp
+++ b/src/mlpack/methods/ann/init_rules/gaussian_init.hpp
@@ -55,11 +55,12 @@ class GaussianInitialization
    */
   void Initialize(arma::mat& W,
                   const size_t rows,
-                  const size_t cols)
+                  const size_t cols,
+                  const size_t seed =21)
   {
     std::random_device rd;
     std::mt19937 gen(rd());
-    std::srand(std::time(0));
+    std::srand(seed);
     W = arma::mat(rows, cols);
     double m;
     double v;
--- a/src/mlpack/methods/ann/init_rules/gaussian_init.hpp
+++ b/src/mlpack/methods/ann/init_rules/gaussian_init.hpp
@@ -33,23 +33,21 @@ class GaussianInitialization
    * @param variance The variance of the gaussian.
    */
   GaussianInitialization():
-  mean(arma::zeros<arma::vec>(1)), covariance(arma::eye<arma::mat>(1, 1))
+  mean(0), variance(1)
   {}
 
   /**
-   * Initialize the random initialization rule with the given bound.
-   * Using the negative of the bound as lower bound and the positive bound as
-   * upper bound.
+   * Initialize the gaussian with the given mean and variance.
    *
-   * @param bound The number used as lower bound
+   * @param mean Mean of the gaussian
+   * @param variance Variance of the gaussian
    */
-  GaussianInitialization(arma::mat& W, const size_t rows, const size_t cols):
-  mean(arma::zeros<arma::vec>(rows)), covariance(arma::eye<arma::mat>(rows, rows))
+  GaussianInitialization(const double mean=0, const double variance=1):
+  mean(mean), variance(variance)
   {}
 
   /**
-   * Initialize the elements weight matrix using a Gaussian Distribution
-   * of given mean and variance.
+   * Initialize the elements weight matrix using a Gaussian Distribution.
    *
    * @param W Weight matrix to initialize.
    * @param rows Number of rows.
@@ -60,16 +58,13 @@ class GaussianInitialization
                   const size_t cols)
   {
     W = arma::mat(rows, cols);
-    distribution::GaussianDistribution dist(mean, covariance);
-    if (mean.n_elem == 1 && covariance.n_elem == 1)
-    {
-      W = arma::randn(rows, cols);
-    }
-    else
-    {
-      for (size_t i = 0; i < W.n_rows; i++)
-        W.col(i) = dist.Random();
-    }
+    arma::vec m(1);
+    arma::mat v(1,1);
+    m = {this->mean};
+    v = {this->variance};
+    distribution::GaussianDistribution dist(m, v);
+    for (size_t i = 0; i < W.n_rows; i++)
+      W.col(i) = dist.Random();
   }
 
   /**
@@ -92,10 +87,10 @@ class GaussianInitialization
 
  private:
   //! Mean of the gaussian.
-  const arma::vec mean;
+  const double mean;
 
   //! Variance of the gaussian.
-  const arma::mat covariance;
+  const double variance;
 }; // class GaussianInitialization
 } // namespace ann
 } // namespace mlpack
--- a/src/mlpack/methods/ann/init_rules/gaussian_init.hpp
+++ b/src/mlpack/methods/ann/init_rules/gaussian_init.hpp
@@ -60,11 +60,10 @@ class GaussianInitialization
     W = arma::mat(rows, cols);
     arma::vec m(1);
     arma::mat v(1,1);
-    m = {this->mean};
-    v = {this->variance};
+    m = mean;
+    v = variance;
     distribution::GaussianDistribution dist(m, v);
-    for (size_t i = 0; i < W.n_rows; i++)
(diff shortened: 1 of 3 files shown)


kris-singh:
The code builds locally. Also arma::randn(size(X)) is correct i don't understand why Travis gives this as a error(possibly version issue)

rcurtin:
mlpack supports Armadillo 4.200 and later, so if `arma::randn(size(X))` is not available in that version, you shouldn't use it.  Can you update that code to use something in Armadillo 4.200?  Thanks.

kris-singh:
@rcurtin i can't find the documentation for Armadillo 4.200. Can you help ?

rcurtin:
I just look here to get information on when things were added to Armadillo:

http://arma.sourceforge.net/docs.html#api_additions

kris-singh:
I got documentation for 4.200 from [here](https://src.fedoraproject.org/repo/extras/armadillo/armadillo-4.200.0.tar.gz/0ed282a5b32b39d5458b06c08e2cb15d/). They use randn(row, col) in 4.200.

zoq:
So, if you use ``W = arma::randn(W.n_rows, W.n_cols);`` instead of ``W = arma::randn(size(W));``it should work.

kris-singh:
Yes, i think this can be reviewed. This is my first PR so there might be silly mistakes.

src/mlpack/methods/ann/init_rules/gaussian_init.hpp
```diff
@@ -0,0 +1,103 @@
...
+{
+ public:
+  /**
+   * Initialize the gaussian initialization rule with mean=0 and
+   * variance = 1.
+   *
+   * @param mean The mean of the gaussian.
+   * @param variance The variance of the gaussian.
+   */
+  GaussianInitialization():
```
zoq: To be able to specify mean and covariance, you should also implement the constructor, something like:

```
GaussianInitialization(const MatType& mean, const MatType& covariance) : ...
```

src/mlpack/methods/ann/init_rules/gaussian_init.hpp
```diff
@@ -0,0 +1,103 @@
...
+   * @param rows Number of rows.
+   * @param cols Number of columns.
+   */
+  void Initialize(arma::mat& W,
+                  const size_t rows,
+                  const size_t cols)
+  {
+    W = arma::mat(rows, cols);
+    distribution::GaussianDistribution dist(mean, covariance);
+    if (mean.n_elem == 1 && covariance.n_elem == 1)
```
zoq: You could use ``is_empty`` to check whether object is empty, no need to initialize mean and covariance.

src/mlpack/methods/ann/init_rules/gaussian_init.hpp
```diff
@@ -0,0 +1,103 @@
...
+   *
+   * @param W Weight matrix to initialize.
+   * @param rows Number of rows.
+   * @param cols Number of columns.
+   */
+  void Initialize(arma::mat& W,
+                  const size_t rows,
+                  const size_t cols)
+  {
+    W = arma::mat(rows, cols);
```
zoq: We only need the GaussianDistribution object if the user specified mean and covariance, so we could put

```
W = arma::mat(rows, cols);
distribution::GaussianDistribution dist(mean, covariance);
```

into the else case.

kris-singh:
@zoq Changes Made please review.

kris-singh:
Any idea on how i can write the tests.

src/mlpack/methods/ann/init_rules/gaussian_init.hpp
```diff
@@ -0,0 +1,102 @@
...
+    std::mt19937 gen(rd());
+    std::srand(seed);
+    W = arma::mat(rows, cols);
+    double m;
+    double v;
+    m = mean;
+    v = variance;
+    std::normal_distribution<> d(m, v);
+    //distribution::GaussianDistribution dist(m, v);
+    W.imbue( [&]() { return arma::as_scalar(d(gen)); } );
```
zoq: I think we could reuse the ``RandNormal`` function (``src/mlpack/core/math/random.hpp``).

src/mlpack/methods/ann/init_rules/gaussian_init.hpp
```diff
@@ -0,0 +1,102 @@
...
+   * Initialize the elements weight matrix using a Gaussian Distribution.
+   *
+   * @param W Weight matrix to initialize.
+   * @param rows Number of rows.
+   * @param cols Number of columns.
+   */
+  void Initialize(arma::mat& W,
+                  const size_t rows,
+                  const size_t cols,
+                  const size_t seed =21)
```
zoq: We if use the ``RandNormal``there is no need to specify the seed here.

src/mlpack/methods/ann/init_rules/gaussian_init.hpp
```diff
@@ -0,0 +1,102 @@
...
+  {
+    std::random_device rd;
```
zoq: I think ``randu``is probably faster if zero mean and unit variance. What do you think?

src/mlpack/tests/init_rules_test.cpp
```diff
@@ -123,4 +126,38 @@ BOOST_AUTO_TEST_CASE(OivsInitTest)
...
+  double var,var3d;
+  mean=mean3d=1;
+  var=var3d=1;
+  GaussianInitialization t(0, 0.2);
+
+  for(size_t i =0; i<10; i++)
+  {
+    t.Initialize(weights, row, col);
+    t.Initialize(weights3d, row, col, slice);
+    distribution::GaussianDistribution Dist;
```
zoq: Do we need ``distribution::GaussianDistribution Dist;`` here?

src/mlpack/tests/init_rules_test.cpp
```diff
@@ -123,4 +126,38 @@ BOOST_AUTO_TEST_CASE(OivsInitTest)
...
+    mean += arma::accu(weights) / weights.n_elem;
+    var += arma::accu(pow((weights.t() - mean), 2)) / weights.n_elem -1 ;
+    mean3d += arma::accu(weights3d.slice(0)) / weights3d.slice(0).n_elem;
+    var3d += arma::accu(pow((weights3d.slice(0) - mean), 2)) / weights3d.slice(0).n_elem -1;
+  }
+  mean = mean/10;
+  var = var/10;
+  mean3d /= 10;
+  var3d /= 10;
+  BOOST_REQUIRE_SMALL(mean-0, 0.5);
```
zoq: Maybe we can choose a better threshold here, since what we do here is to test if e.g. ``mean < 0.5``.

src/mlpack/tests/init_rules_test.cpp
```diff
@@ -10,19 +10,22 @@
...
 #include <mlpack/methods/ann/init_rules/random_init.hpp>
 #include <mlpack/methods/ann/init_rules/zero_init.hpp>
+#include <mlpack/methods/ann/init_rules/gaussian_init.hpp>
 
 #include <boost/test/unit_test.hpp>
 #include "test_tools.hpp"
 
 using namespace mlpack;
 using namespace mlpack::ann;
+using namespace mlpack::distribution;
```
zoq: I think we can remove the namespace here.

kris-singh:
@zoq Ready for merge. I have the changes that you asked for. Thanks

src/mlpack/tests/init_rules_test.cpp
```diff
@@ -123,4 +127,37 @@ BOOST_AUTO_TEST_CASE(OivsInitTest)
...
+  arma::cube weights3d;
+  const size_t row = 7;
+  const size_t col = 7;
+  const size_t slice = 2;
+  double mean, mean3d;
+  double var,var3d;
+  mean=mean3d=1;
+  var=var3d=1;
+  GaussianInitialization t(0, 0.2);
+  RandomSeed(21);
```
zoq: It should work with any random seed right? So I'd like to remove the seed here.

src/mlpack/tests/init_rules_test.cpp
```diff
@@ -10,13 +10,17 @@
  * http://www.opensource.org/licenses/BSD-3-Clause for more information.
  */
 #include <mlpack/core.hpp>
+#include <mlpack/core/math/random.hpp>
+#include <mlpack/prereqs.hpp>
```
zoq: ``#include <mlpack/core.hpp>`` includes ``#include <mlpack/prereqs.hpp>``, so I think we can remove the header file.

src/mlpack/methods/ann/init_rules/gaussian_init.hpp
```diff
@@ -0,0 +1,99 @@
...
+   * @param cols Number of columns.
+   */
+  void Initialize(arma::mat& W,
+                  const size_t rows,
+                  const size_t cols)
+  {
+    W = arma::mat(rows, cols);
+    double m;
+    double v;
+    m = mean;
```
zoq: I think this is unnecessary, we could just use ``mean`` and ``variance``right?

src/mlpack/tests/init_rules_test.cpp
kris-singh: No, this won't. I did try it without a seed. But the problem is until and unless we increase the number of iterations to a very high value we can guarantee that the mean would be arbitrarily close to epsilon value. Thats why i have used a seed here.

zoq: So this test "only" works with this particular seed?

zoq: I think we also could do here is to test against the GaussianDistribution.

kris-singh: The test would work for most seeds but not for all. I don't understand "test  againt the GaussianDistribution"

zoq: Nevermind you can't since GaussianDistribution provides the option to use a custom covariance matrix. So Do you think we could remove the random seed and let the test run for e.g. 5 trails and if it passes one from the 5 trails it's fine?

kris-singh: Okay we could use counter to see if the mean and variance values are within certain bounds. And then use the Boost_test_equal to check that counter value if > 1. Yes i think we could do that.

kris-singh:
@zoq all changes are made. Please have a look and let me know.

kris-singh:
@zoq Could you please merge this. So i can start the work on Xavier from master as i we discussed this yesterday.

zoq:
Thanks for the contribution I made some minor style fixes in 8edaea380a1faff43d10833f36b6fd76c3b5b72e.

zoq:
Just a recommendation, it would be great if you could use some more descriptive commit messages.

kris-singh:
@zoq Thanks i will take care of these in things in the future.

PR #879, opened 2017-02-21
Readme idea

zoq:
This PR is kind of an follow up to this commit: https://github.com/zoq/mlpack/commit/d5841b1f7204434f39f4887b342335e195a009f7#diff-04c6e90faac2675aa89e2176d2eec7d8; appveyor decided to change the badge size again. Note this is just an idea, let me know what you all think.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,5 +1,5 @@
 <h1 align="center">
-  <a href="http://mlpack.org"><img src="http://kurg.org/media/mlpack-icon-white-version-8.svg" height=280 alt="mlpack: a scalable C++ machine learning library"></a>
+  <a href="http://mlpack.org"><img src="http://kurg.org/media/mlpack-icon-white-version-9.svg" height=280 alt="mlpack: a scalable C++ machine learning library"></a>
   <br>
   a scalable C++ machine learning library
   <br>
--- a/README.md
+++ b/README.md
@@ -1,5 +1,5 @@
 <h1 align="center">
-  <a href="http://mlpack.org"><img src="http://kurg.org/media/mlpack-icon-white-version-12.svg" height=280 alt="mlpack: a scalable C++ machine learning library"></a>
+  <a href="http://mlpack.org"><img src="http://kurg.org/media/mlpack-icon-white-version-12.svg" style="background-color:rgba(0,0,0,0);" height=280 alt="mlpack: a scalable C++ machine learning library"></a>
   <br>
   a scalable C++ machine learning library
   <br>
--- a/README.md
+++ b/README.md
@@ -1,5 +1,5 @@
 <h1 align="center">
-  <a href="http://mlpack.org"><img src="http://kurg.org/media/mlpack-icon-white.svg" height=280 alt="mlpack: a scalable C++ machine learning library"></a>
+  <a href="http://mlpack.org"><img src="http://kurg.org/media/mlpack-icon-white-version-3.svg" height=280 alt="mlpack: a scalable C++ machine learning library"></a>
   <br>
   a scalable C++ machine learning library
   <br>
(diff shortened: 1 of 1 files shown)


zoq:
That might help to see the updated version: https://github.com/zoq/mlpack/tree/readme

rcurtin:
I can use the current mlpack.org logo if you like, and I can modify it to be black-on-white instead of white-on-black.  Overall I think the idea is great though---Github is kind of the "homepage" for mlpack so it will be what most people see, so we should make the README display nicely. :)

zoq:
Sounds like an interesting idea, let's see how it looks with the modified logo.

rcurtin:
http://www.mlpack.org/mlpack-icon-white.png

zoq:
I agree it looks a little bit weird at the beginning, but aren't this the mlpack colors? Also, I should have never started this ... it bothers me, that it's not a vector graphic, so I guess I'll go and create one, do you remember what font this is? If not I guess I'll train some classifier on a font dataset and figure it out.

zoq:
Updated version: https://github.com/zoq/mlpack/tree/readme

zoq:
Maybe the lines are a little bit thin, but I just wanted to see the whole page to get an impression how it looks.

zoq:
I mean edges not lines.

rcurtin:
I agree, maybe we can thicken the edges somehow?

I think the font that Ryan Birmingham used a long time ago when he made the logo was the same one on the website:
https://www.fontsquirrel.com/fonts/maven-pro
But I'm not sure if that has an SVG font variant with it, so maybe something else will have to be used.  I think anything that's relatively similar and sans-serif should fit ok.

I revisited the original versions of the logo I had, there was never an SVG; even the original FASTLab logo that the mlpack logo was built from was only .png.

zoq:
Okay, thanks, thicken the edges isn't a problem at all, since it's an SVG :)

zoq:
I'm not really happy with the different colors.

zoq:
I will use this comment to link to all versions:

Version A: https://github.com/zoq/mlpack/blob/3f22212fb4434ae75a320bb9001988e23b446e60/README.md
Version B: https://github.com/zoq/mlpack/blob/688cfc348e05a4b73fe57816a9955df20747373a/README.md
Version C: https://github.com/zoq/mlpack/blob/3b712accea178c735e6cc0743e503f8b50c20ac8/README.md

zoq:
The Color scheme G was created by http://colormind.io/ - "Colormind is a deep learning AI that knows color theory."

I like the idea of:

Hey, how did you came up with such nice colors ... We didn't some machine learning algorithm did it for us.

rcurtin:
I like (G) and telling people that we used machine learning to select the colors. :)  What do you think?

Do you think we should change to a sans serif font to match the rest of the mlpack website (and Github too)?

zoq:
Previously I used a font that matched with the existing logo font, but I think using the same font as for the content on the website makes sense. Version I uses the color scheme from Version G and maven pro as font.

rcurtin:
It seems like version (I) doesn't render correctly for me: http://ratml.org/misc_img/mlpack_logo_render.png

zoq:
Do you have maven-pro as system font? I guess you can include the font src into the svg.

rcurtin:
Nope, I don't think I have it as a system font.  I would imagine most don't either, since it's not a standard font.  If it's easy to include the font source in the SVG, that seems like the right thing to do.

zoq:
Turns out you can just create another vector just for the text.

zoq:
Just checked the filesize, 5KB (SVG) the logo on the website is 11KB, I guess it's difficult to beat that without any quality loss.

rcurtin:
I think it looks great, if you want to merge go ahead.

zoq:
I made a new commit: bd453ac66a9fa7fea672b20c.

PR #887, opened 2017-02-23
Implementation of AdaMax optimizer.

ivmarkp:
`SimpleAdaMaxTestFunction` test is currently failing with values of `coordinates` vector as `[-nan; -45.59; -nan]` after optimization. For our context, the original values were `[6; -45.6; 6.2]` to begin with.

So as it appears, Optimize function is reducing positive values to `-nan` which might be the case due to `0 / 0` (I'm suspecting somewhere during the iterations atleast one of the elements of `u` and `m` at the same index become zero) and minimal change seen on negative values.

### DIFF
--- a/src/mlpack/core/optimizers/adam/adam_impl.hpp
+++ b/src/mlpack/core/optimizers/adam/adam_impl.hpp
@@ -3,8 +3,9 @@
  * @author Ryan Curtin
  * @author Vasanth Kalingeri
  * @author Marcus Edel
+ * @author Vivek Pal
  *
- * Implementation of the Adam optimizer.
+ * Implementation of the Adam and AdaMax optimizer.
  *
  * mlpack is free software; you may redistribute it and/or modify it under the
  * terms of the 3-clause BSD license.  You should have received a copy of the
@@ -28,7 +29,8 @@ Adam<DecomposableFunctionType>::Adam(DecomposableFunctionType& function,
                                      const double eps,
                                      const size_t maxIterations,
                                      const double tolerance,
-                                     const bool shuffle) :
+                                     const bool shuffle,
+                                     const bool adaMax) :
     function(function),
     stepSize(stepSize),
     beta1(beta1),
@@ -36,7 +38,8 @@ Adam<DecomposableFunctionType>::Adam(DecomposableFunctionType& function,
     eps(eps),
     maxIterations(maxIterations),
     tolerance(tolerance),
-    shuffle(shuffle)
+    shuffle(shuffle),
+    adaMax(adaMax)
 { /* Nothing to do. */ }
 
 //! Optimize the function (minimize).
@@ -66,6 +69,9 @@ double Adam<DecomposableFunctionType>::Optimize(arma::mat& iterate)
 
   // Exponential moving average of gradient values.
   arma::mat m = arma::zeros<arma::mat>(iterate.n_rows, iterate.n_cols);
+
+  // Initialize the exponentially weighted infinity norm for AdaMax optimizer.
+  arma::mat u = arma::zeros<arma::mat>(iterate.n_rows, iterate.n_cols);
 
   // Exponential moving average of squared gradient values.
   arma::mat v = arma::zeros<arma::mat>(iterate.n_rows, iterate.n_cols);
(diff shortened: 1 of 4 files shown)


src/mlpack/core/optimizers/adam/adam_impl.hpp
```diff
@@ -113,19 +119,44 @@ double Adam<DecomposableFunctionType>::Optimize(arma::mat& iterate)
...
-     * It should be noted that the term, m / (arma::sqrt(v) + eps), in the
-     * following expression is an approximation of the following actual term;
-     * m / (arma::sqrt(v) + (arma::sqrt(biasCorrection2) * eps).
-     */
-    iterate -= (stepSize * std::sqrt(biasCorrection2) / biasCorrection1) *
-                m / (arma::sqrt(v) + eps);
+    if (adaMax)
+    {
+      if (biasCorrection1 != 0.0)
+        iterate -= ((stepSize / biasCorrection1) * (m / u));
```
zoq: I looked over the code, and it looks good; probably using m / (u + eps) might help?

ivmarkp: Yes, you are right. That resolves the issue. Pushing the change right away. Thanks for your suggestion. I was actually hesitant to use eps as it is not used in AdaMax algorithm in the paper.

src/mlpack/core/optimizers/adam/adam_impl.hpp
```diff
@@ -113,19 +119,44 @@ double Adam<DecomposableFunctionType>::Optimize(arma::mat& iterate)
...
+        {
+          if (u.at(row, col) < absGradient.at(row, col))
+            u.at(row, col) = absGradient.at(row, col);
+        }
+      }
+    }
+    else
+    {
+      v *= beta2;
+      v += (1 - beta2) * (gradient % gradient);
```
ivmarkp: It is not related to AdaMax but since it concerns the same code so asking here itself -- according to the paper, v should calculated as `v += (1 - beta2) * arma::pow(gradient, 2)` but here we take the modulo instead of multiplying it with itself. I can't tell but is there an advantage of doing that in some way?

zoq: In this case ``% ``isn't modulo it's the Schur product (element-wise multiplication), if you use ``*`` you get the matrix multiplication.

src/mlpack/core/optimizers/adam/adam_impl.hpp
```diff
@@ -113,19 +119,44 @@ double Adam<DecomposableFunctionType>::Optimize(arma::mat& iterate)
...
     m += (1 - beta1) * gradient;
 
-    v *= beta2;
-    v += (1 - beta2) * (gradient % gradient);
+    if (adaMax)
+    {
+      u *= beta2;
+      arma::mat absGradient = arma::abs(gradient);
+      // Update the exponentially weighted infinity norm.
+      for (size_t row = 0; row != iterate.n_rows; ++row)
```
zoq: Would be interesting to test if ``arma::max`` is faster as iterating through all elements.

src/mlpack/core/optimizers/adam/adam_impl.hpp
ivmarkp: Okay, now I get it.

src/mlpack/core/optimizers/adam/adam_impl.hpp
ivmarkp: I tested for matrices of size 10000 x 10000 and `arma::max` consistently performs better by more than 35%. Here's my [test program](https://gist.github.com/ivmarkp/d549f9af02c769d5496987bf9aef12c2) and I should mention that it was a very enlightening experience :)

Pushing the changes in a couple minutes.

src/mlpack/core/optimizers/adam/adam_impl.hpp
zoq: Such details are often omitted e.g. to simplify the proof.

src/mlpack/core/optimizers/adam/adam_impl.hpp
zoq: There are some side effects like branch predictions, memory usage, etc. that has an influence on the runtime of the code that should theoretically be faster.

But it sounds like you had fun to explore another approach, and if it turns out to be faster even better.

src/mlpack/core/optimizers/adam/adam_impl.hpp
```diff
@@ -67,6 +70,9 @@ double Adam<DecomposableFunctionType>::Optimize(arma::mat& iterate)
   // Exponential moving average of gradient values.
   arma::mat m = arma::zeros<arma::mat>(iterate.n_rows, iterate.n_cols);
 
+  // Initialize the exponentially weighted infinity norm for AdaMax optimizer.
+  arma::mat u = arma::zeros<arma::mat>(iterate.n_rows, iterate.n_cols);
```
zoq: What do you think, should we initialize u and v only at the specified cases, for adamax and standard adam?

ivmarkp: Yes, I think may be we can move the initialization of u and v inside the `if ... else` at line 122. What do you suggest?

zoq: I would do it right at line 74, 77 that way we don't have to check if it's already initialized in every iteration:

```
arma::mat u;
if (adamax)
{
  u = ...
}
```

ivmarkp: Yes, I agree. Should we extend that approach to look something like the following piece of code?
```cpp
arma::mat u, v;
  if (adaMax)
    u = arma::zeros<arma::mat>(iterate.n_rows, iterate.n_cols);
  else
    v = arma::zeros<arma::mat>(iterate.n_rows, iterate.n_cols);
```
 I think it'd more intuitive if preceeded by a comment on what u and v are, and used whether for Adam or Adamax e.g. 

> Initialize, either the exponentially weighted infinity norm for AdaMax optimizer (u) or exponential moving average of squared gradient values for Adam optimizer (v).

What do you suggest should be done?

zoq: Looks good, don't forget the brackets :) Also, adding more comments is always a great idea.

ivmarkp: Okay thanks, pushing the changes.

zoq:
Thanks for the contribution, I think it's a great addition. Maybe we can make a nice comparison between the different methods.

Minor note for the future, it would be great if you could use a more descriptive commit message.

ivmarkp:
Yes, definitely. I think since we now have a number of optimizers in mlpack, we can run some experiments using these optimizers on different network architectures for a simple problem e.g. MNIST classification to draw some comparisons.

 > Minor note for the future, it would be great if you could use a more descriptive commit message.

Sorry about those fixup commit messages. I'll write more descriptive commit messages from now onwards -- good commit messages are always valuable especially when there are multiple contributors to a project.

PR #889, opened 2017-02-23
Refactored RASearch model to include boost variant and visitor paradigm

dineshraj01:


src/mlpack/methods/rann/ra_model.hpp
```diff
@@ -18,13 +18,275 @@
...
+                  typename TreeMatType> class TreeType>
+using RAType = RASearch<SortPolicy,
+                        metric::EuclideanDistance,
+                        arma::mat,
+                        TreeType>;
+
+template<typename SortPolicy>
+struct RAModelName
+{
+  static const std::string Name() { return "rank_approx_search_model";};
```
rcurtin: I know that this is a really picky comment about style, but can you please add a space between the semicolon and closing brace?  I also think the final semicolon is unnecessary so we can just do `; }` at the end of the line.

src/mlpack/methods/rann/ra_model_impl.hpp
```diff
@@ -19,21 +19,258 @@
 namespace mlpack {
 namespace neighbor {
 
+//! Monochromatic search for the given RAType instance.
+template<typename RAType>
+void MonoSearchVisitor::operator()(RAType* ra) const
+{
+  if(ra)
+    return ra->Search(k, neighbors, distances);
+  throw std::runtime_error("no rank-apporximate model initialized");
```
rcurtin: Looks like there is a very minor misspelling here, and it should also be `if (ra)`.  Again very minor picky comments. :)

src/mlpack/methods/rann/ra_model_impl.hpp
```diff
@@ -781,79 +519,42 @@ void RAModel<SortPolicy>::BuildModel(arma::mat&& referenceSet,
...
-        typename RAType<tree::KDTree>::Tree* kdTree =
-            new typename RAType<tree::KDTree>::Tree(std::move(referenceSet),
-            oldFromNewReferences, leafSize);
-        kdTreeRA = new RAType<tree::KDTree>(kdTree, singleMode);
-
-        // Give the model ownership of the tree.
-        kdTreeRA->treeOwner = true;
-        kdTreeRA->oldFromNewReferences = oldFromNewReferences;
-      }
+      raSearch = new RAType<SortPolicy, tree::KDTree> (naive, singleMode);
```
rcurtin: No need for a space between the closing `>` and the opening `(`.

src/mlpack/methods/rann/ra_search_impl.hpp
```diff
@@ -306,6 +306,31 @@ void RASearch<SortPolicy, MetricType, MatType, TreeType>::Train(
...
 
+//! Set the reference tree to a new reference tree.
+template<typename SortPolicy,
+         typename MetricType,
+         typename MatType,
+         template<typename TreeMetricType,
+                  typename TreeStatType,
+                  typename TreeMatType> class TreeType>
+void RASearch<SortPolicy, MetricType, MatType, TreeType>::Train(
+  Tree* referenceTree)
```
rcurtin: I think this will fit all on one line in 80 characters, but if not then the indentation should be two tabs not one tabs on a wrapped line. :)

dineshraj01:
Hi, I think I'm gonna need a little help, not sure whats causing the bug :( SerializeVisitor looks fine to me.

rcurtin:
Unfortunately I don't really have the time to dig too deep here---I'm working with a lot of other PRs and then I have my own mlpack contributions I am putting time into. :)  I would suggest reproducing the issue, seeing when it occurs, manually serializing the objects with and without the visitor, and other types of debugging to try and isolate what the issue might be.

rcurtin:
Hey @dineshraj01, sorry for the long response on this one.  It turns out that the serialize issue exists because the `SerializeVisitor` modifies a `RASearch*` (just a pointer), which is only a local change.  So I changed the signature to accept an `RASearch*&` instead, and now the problem is fixed in e2dca6e; I also fixed the style issues in 95f9487.

Since this PR sat open and approved for so long, I didn't think it was necessary to wait a few more days, so I just went ahead and merged it.

Thanks again for the hard work!  Sorry it took so long to see it through...

dineshraj01:
Hi,
Thanks,  I could have worked more harder to detect the issue. This issue so discouraged me to work. Looks like I can continue now. :D 
Thanks again.

rcurtin:
No worries, it's all just a matter of finding the time. :)  I know the discouraging feeling, sometimes I have spent weeks digging after a single issue and it is hard to have the motivation to keep going.  If you want to contribute more in the future we will gladly review and merge it.

PR #890, opened 2017-02-25
minor documentation update

eddelbuettel:
It's a pretty trivial fix, but something I came across this week.

There seems to be a bit more around it -- some references to 'matrix' when the input at this level (ie `main()`) -- really is a file.  I can poke some more if you want.

### DIFF
--- a/src/mlpack/methods/naive_bayes/nbc_main.cpp
+++ b/src/mlpack/methods/naive_bayes/nbc_main.cpp
@@ -28,7 +28,7 @@ PROGRAM_INFO("Parametric Naive Bayes Classifier",
     "training set, or loads a model from the given model file, and then may use"
     " that trained model to classify the points in a given test set."
     "\n\n"
-    "Labels are expected to be the last row of the training set (--train_file),"
+    "Labels are expected to be the last row of the training set (--training_file),"
     " but labels can also be passed in separately as their own file "
     "(--labels_file).  If training is not desired, a pre-existing model can be "
     "loaded with the --input_model_file (-m) option."


rcurtin:
Thanks! There is a bit more complexity to the documentation here though. Soon (next few weeks?) I will merge in the automatic bindings for Python.  These use the PROGRAM_INFO(), PARAM_*(), etc. macros in order to generate fully functioning Python bindings (and other languages to be added later, possibly including R---but I'll send you an email or open an issue on RcppMLPACK separately about that some time from now).

What this means is that what is written in the PROGRAM_INFO() macro has to be agnostic; in Python, you pass a matrix; for the CLI programs, you pass a file containing a matrix. I don't have the exact solution quite yet---and I want to avoid handwritten documentation for every possible binding language---but somewhat strange phrases like that will have to be gone over at that point.

So, I guess, long story short: it'll get handled, but in a weird and not straightforward way. :)

eddelbuettel:
Perfect, and thanks for the update. I suspected something like that when I saw the LOVELY UPPERCASE MACROS flying around there.

I guess you follow my very slow progress on what we now call 'RcppMLPACK2' (using MLPACK 2.1.*) in the RcppMLPACK org?  If not, come on over.

Also, two more trivial questions:
- is a standard doxygen somewhere?  you're css makes it kinda ... hard.  I may just to our org unless you'd rather not have us do that...
- is it now 'mlpack' instead of 'MLPACK' ?

rcurtin:
Yeah, I have seen the issues briefly but I don't know if there is anything I can do to be helpful there.

 * I guess you can remove the CSS if you like and it should configure like other Doxygen.  Just clear the `HTML_FOOTER` and `HTML_EXTRA_STYLESHEET` from the Doxyfile and I think it will be fine.  I tried to make it so it would generate something that looks more like the mlpack website.  Doxygen output is kind of crap for templates though... maybe someone will write an alternative or improve it someday?

 * Yeah, that's how I try to refer to it for consistency but realistically people call it MLPACK, Mlpack, MLpack, MlPaCk, whatever they like... I guess there is no way to win this one... :)

eddelbuettel:
I did actually run just `doxygen`, and it didn't seem themed at all.  See [here for a temp location](http://dirk.eddelbuettel.com/tmp/mlpack-doxygen/).  

I may just push that into the RcppMLPACK org as well, possibly after turning that one toggle on I found helpful for Rcpp (a local, client-side, Javascript-based search box).    

I have the feeling the ALLUPPERCASE version was there at the start and the cooler 'alllowercase' came later :)

eddelbuettel:
For what it is worth, that works if you flip the very last line in your `Doxyfile`.   I now have

```
#---------------------------------------------------------------------------
# Configuration::additions related to the search engine
#---------------------------------------------------------------------------
#SEARCHENGINE           = NO
SEARCHENGINE           = YES
```

and if you refresh [the link from above](http://dirk.eddelbuettel.com/tmp/mlpack-doxygen/) you see the search box.

rcurtin:
Hm, ok, yeah, I see now that I forgot that all of the styling on mlpack.org is there and not in the extra stylesheet file I mentioned.  That file is actually some smoothing for inline formula displays.

I integrated the search box into the default config with 56090dc47, it's a nice change.

And yeah, the lowercase name came after the original.  In the long ago it was actually even worse: the core library was called "FASTLib" (with the weird camelcase), so the whole product was "FASTLib/MLPACK".  When I started maintaining it I changed it to only "MLPACK", but then later realized it would be strange to distribute `libMLPACK.so` so then I switched to lowercase.  But by that time, it was too late, the paper was already out... so now the capitalization that gets used is pretty arbitrary.

eddelbuettel:
Great.  Saves me another one-line PR :)

PR #895, opened 2017-02-28
Fix support for printing arma objects with custom precision 

shikharbhardwaj:
Relevant issue : #755 

A few issues : 
 * After the user sets up custom flags, `arma` objects are printed with the set precision, but the stream does not align matrix elements.
 * During build, a `pragma` message appears

> In file included from /home/shikhar/Documents/dev/mlpack-fork/src/mlpack/../mlpack/core.hpp:217:0,
                 from /home/shikhar/Documents/dev/mlpack-fork/src/mlpack/tests/test_tools.hpp:15,
                 from /home/shikhar/Documents/dev/mlpack-fork/src/mlpack/tests/mlpack_test.cpp:25:
/home/shikhar/Documents/dev/mlpack-fork/src/mlpack/../mlpack/prereqs.hpp:20:17: note: #pragma message: Armadillo was included before mlpack; this can sometimes cause prob
lems.  It should only be necessary to include <mlpack/core.hpp> and not <armadillo>.
 #pragma message "Armadillo was included before mlpack; this can sometimes cause\

because of the need to include `arma` in `prefixedoutstream.hpp`

### DIFF
--- a/src/mlpack/core/util/prefixedoutstream.hpp
+++ b/src/mlpack/core/util/prefixedoutstream.hpp
@@ -13,13 +13,7 @@
 #ifndef MLPACK_CORE_UTIL_PREFIXEDOUTSTREAM_HPP
 #define MLPACK_CORE_UTIL_PREFIXEDOUTSTREAM_HPP
 
-#include <iostream>
-#include <iomanip>
-#include <string>
-#include <streambuf>
-#include <stdexcept>
-
-#include <mlpack/core/arma_extend/arma_extend.hpp> // Includes Armadillo
+#include <mlpack/prereqs.hpp>
 
 namespace mlpack {
 namespace util {
--- a/src/mlpack/core/util/prefixedoutstream.hpp
+++ b/src/mlpack/core/util/prefixedoutstream.hpp
@@ -19,7 +19,7 @@
 #include <streambuf>
 #include <stdexcept>
 
-#include <mlpack/core/util/sfinae_utility.hpp>
+#include <mlpack/core/arma_extend/arma_extend.hpp> // Includes Armadillo
 
 namespace mlpack {
 namespace util {
@@ -124,11 +124,19 @@ class PrefixedOutStream
    * Conducts the base logic required in all the operator << overloads.  Mostly
    * just a good idea to reduce copy-pasta.
    *
+   * Calls 2 different overloads to be able to output arma objects with custom
+   * precision
+   *
    * @tparam T The type of the data to output.
    * @param val The The data to be output.
    */
   template<typename T>
-  void BaseLogic(const T& val);
+  typename std::enable_if<!arma::is_arma_type<T>::value, void>::type
+  BaseLogic(const T& val);
+
+  template<typename T>
+  typename std::enable_if<arma::is_arma_type<T>::value, void>::type
+  BaseLogic(const T& val);
 
   /**
    * Output the prefix, but only if we need to and if we are allowed to.
--- a/src/mlpack/core/util/prefixedoutstream.hpp
+++ b/src/mlpack/core/util/prefixedoutstream.hpp
@@ -125,11 +125,11 @@ class PrefixedOutStream
    * @param val The The data to be output.
    */
   template<typename T>
-  typename std::enable_if<!arma::is_arma_type<T>::value, void>::type
+  typename std::enable_if<!arma::is_arma_type<T>::value>::type
   BaseLogic(const T& val);
 
   template<typename T>
-  typename std::enable_if<arma::is_arma_type<T>::value, void>::type
+  typename std::enable_if<arma::is_arma_type<T>::value>::type
   BaseLogic(const T& val);
 
   /**
--- a/src/mlpack/tests/prefixedoutstream_test.cpp
+++ b/src/mlpack/tests/prefixedoutstream_test.cpp
@@ -81,12 +81,12 @@ BOOST_AUTO_TEST_CASE(TestArmadilloPrefixedOutStream)
       BASH_GREEN "[INFO ] " BASH_CLEAR "   3.5000\n"
       BASH_GREEN "[INFO ] " BASH_CLEAR "   4.0000\n");
 
-  // TODO: Make Op types printable on logs?
-  //ss.str("");
-  //pss << trans(test);
-  //// This should result in there being stuff on the line.
-  //BOOST_REQUIRE_EQUAL(ss.str(), BASH_GREEN "[INFO ] " BASH_CLEAR
-      //"   1.0000   1.5000   2.0000   2.5000   3.0000   3.5000   4.0000\n");
+  ss.str("");
+  pss << trans(test);
+
+  // This should result in there being stuff on the line.
+  BOOST_REQUIRE_EQUAL(ss.str(), BASH_GREEN "[INFO ] " BASH_CLEAR
+      "   1.0000   1.5000   2.0000   2.5000   3.0000   3.5000   4.0000\n");
 
   arma::mat test2("1.0 1.5 2.0; 2.5 3.0 3.5; 4.0 4.5 4.99999");
   ss.str("");
@@ -113,7 +113,7 @@ BOOST_AUTO_TEST_CASE(TestArmadilloPrefixedOutStream)
 
   pss << test;
 
-  BOOST_REQUIRE_EQUAL(ss.str(), 
+  BOOST_REQUIRE_EQUAL(ss.str(),
       BASH_GREEN "[INFO ] " BASH_CLEAR "   1.000000\n"
       BASH_GREEN "[INFO ] " BASH_CLEAR "   1.500000\n"
       BASH_GREEN "[INFO ] " BASH_CLEAR "   2.000000\n"
@@ -122,6 +122,15 @@ BOOST_AUTO_TEST_CASE(TestArmadilloPrefixedOutStream)
       BASH_GREEN "[INFO ] " BASH_CLEAR "   3.500000\n"
       BASH_GREEN "[INFO ] " BASH_CLEAR "   4.000000\n");
 
+  ss.str("");
+
+  pss << trans(test);
+
+  BOOST_REQUIRE_EQUAL(ss.str(),
+      BASH_GREEN "[INFO ] " BASH_CLEAR
+      "   1.000000   1.500000   2.000000   2.500000"
+      "   3.000000   3.500000   4.000000\n");
+
   // Try printing a matrix, with higher precision
   ss << std::setprecision(8);
   ss.str("");
@@ -129,25 +138,25 @@ BOOST_AUTO_TEST_CASE(TestArmadilloPrefixedOutStream)
   pss << test2;
 
   BOOST_REQUIRE_EQUAL(ss.str(),
-      BASH_GREEN "[INFO ] " BASH_CLEAR 
+      BASH_GREEN "[INFO ] " BASH_CLEAR
       "   1.00000000   1.50000000   2.00000000\n"
-      BASH_GREEN "[INFO ] " BASH_CLEAR 
+      BASH_GREEN "[INFO ] " BASH_CLEAR
       "   2.50000000   3.00000000   3.50000000\n"
-      BASH_GREEN "[INFO ] " BASH_CLEAR 
+      BASH_GREEN "[INFO ] " BASH_CLEAR
       "   4.00000000   4.50000000   4.99999000\n");
 
   // Try alignment with larger values
   test2.at(2) = 40;
   ss.str("");
-  pss << test2;
+  pss << trans(test2);
 
   BOOST_REQUIRE_EQUAL(ss.str(),
       BASH_GREEN "[INFO ] " BASH_CLEAR 
-      "    1.00000000    1.50000000    2.00000000\n"
+      "    1.00000000    2.50000000   40.00000000\n"
       BASH_GREEN "[INFO ] " BASH_CLEAR 
-      "    2.50000000    3.00000000    3.50000000\n"
+      "    1.50000000    3.00000000    4.50000000\n"
       BASH_GREEN "[INFO ] " BASH_CLEAR 
-      "   40.00000000    4.50000000    4.99999000\n");
+      "    2.00000000    3.50000000    4.99999000\n");
 
   // Test stream after reset
   test2.at(2) = 4;
--- a/src/mlpack/tests/prefixedoutstream_test.cpp
+++ b/src/mlpack/tests/prefixedoutstream_test.cpp
@@ -104,6 +104,43 @@ BOOST_AUTO_TEST_CASE(TestArmadilloPrefixedOutStream)
       BASH_GREEN "[INFO ] " BASH_CLEAR "hello   1.0000   1.5000   2.0000\n"
       BASH_GREEN "[INFO ] " BASH_CLEAR "   2.5000   3.0000   3.5000\n"
       BASH_GREEN "[INFO ] " BASH_CLEAR "   4.0000   4.5000   5.0000\n");
+
+  // Try to print armadillo objects with custom precision
+  ss << std::fixed;
+  ss << std::setprecision(6);
+  ss.str("");
+
+  pss << test;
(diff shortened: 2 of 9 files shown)


src/mlpack/core/util/prefixedoutstream.hpp
```diff
@@ -19,7 +19,7 @@
 #include <streambuf>
 #include <stdexcept>
 
-#include <mlpack/core/util/sfinae_utility.hpp>
+#include <mlpack/core/arma_extend/arma_extend.hpp> // Includes Armadillo
```
shikharbhardwaj: Includes `arma` to use `is_arma_type`, but causes the `pragma` message. Including `prereqs.hpp` or `core.hpp` causes the build to fail (I could not fully comprehend the reason, but it looked like a cyclic dependency).

rcurtin:
Yes so we may need to change some ordering of includes in `src/mlpack/prereqs.hpp` so that Armadillo is included before `prefixedoutstream.hpp`.  Basically, the ordering of includes is important because we graft some additional functionality onto Armadillo.

For the spacing issue, can't you set `std::cellwidth` (I think that is what it's called, I can't remember)? If that doesn't work then maybe some modification to Armadillo will be required, but that can be submitted upstream as a patch when done.

shikharbhardwaj:
For solving the spacing issue `arma::print` looks at the ranges of the elements to be printed, and figures out the type of stream flags to use from 4 predefined configurations. 

In our use case, we try to print the elements with custom precision, we'd need to find the element which would take the maximum space in its integer part, and output each element with `std::setw(max_int_width + padding)`, where padding is the required spacing. This requires accessing elements from the object.

rcurtin:
It's your call on how to proceed here---remember that basically anything you do will be better than what we already have. :)

I wouldn't worry about doing operations on the input matrix/Armadillo object; if someone is printing a matrix, then the matrix is almost certainly going to be small (otherwise they will quickly realize that they, in fact, did *not* want to print the matrix!).  So any computational overhead to, e.g., find the max will be negligible and probably outweighed by the overhead of printing itself.

shikharbhardwaj:
I changed the code to use `arma::arma_ostream::print`, with the modify parameter as `false`, setting a custom width for the stream, calculating the cell width from the maximum value from the absolute values of the elements.

It worked as intended, but fails the test for printing the transpose of a `Col`, because an overload for `arma::arma_ostream::print` exists only for matrices, cubes and columns, not for `arma::Op` types.

Perhaps we can push a commit upstream to Armadillo with one such overload, which would basically be a wrapper unpacking the underlying type and calling the original function on it.

Here is the build error while compiling the test : 

> In file included from /home/shikhar/Documents/dev/mlpack-fork/src/mlpack/../mlpack/core/util/prefixedoutstream.hpp:162:0,
                 from /home/shikhar/Documents/dev/mlpack-fork/src/mlpack/../mlpack/core/util/log.hpp:18,
                 from /home/shikhar/Documents/dev/mlpack-fork/src/mlpack/../mlpack/prereqs.hpp:98,
                 from /home/shikhar/Documents/dev/mlpack-fork/src/mlpack/../mlpack/core.hpp:217,
                 from /home/shikhar/Documents/dev/mlpack-fork/src/mlpack/tests/prefixedoutstream_test.cpp:15:
/home/shikhar/Documents/dev/mlpack-fork/src/mlpack/../mlpack/core/util/prefixedoutstream_impl.hpp: In instantiation of ‘typename std::enable_if<arma::is_arma_type<T1>::value, void>::type mlpack::util::PrefixedOutStream::BaseLogic(const T&) [with T = arma::Op<arma::Col<double>, arma::op_htrans>; typename std::enable_if<arma::is_arma_type<T1>::value, void>::type = void]’:
/home/shikhar/Documents/dev/mlpack-fork/src/mlpack/../mlpack/core/util/prefixedoutstream_impl.hpp:32:15:   required from ‘mlpack::util::PrefixedOutStream& mlpack::util::PrefixedOutStream::operator<<(const T&) [with T = arma::Op<arma::Col<double>, arma::op_htrans>]’
/home/shikhar/Documents/dev/mlpack-fork/src/mlpack/tests/prefixedoutstream_test.cpp:85:20:   required from here
/home/shikhar/Documents/dev/mlpack-fork/src/mlpack/../mlpack/core/util/prefixedoutstream_impl.hpp:179:30: error: no matching function for call to ‘arma::arma_ostream::print(std::ostringstream&, const arma::Op<arma::Col<double>, arma::op_htrans>&, bool)’
     arma::arma_ostream::print(convert, val, false);
                              ^
In file included from /usr/include/armadillo:361:0,
                 from /home/shikhar/Documents/dev/mlpack-fork/src/mlpack/../mlpack/core/arma_extend/arma_extend.hpp:52,
                 from /home/shikhar/Documents/dev/mlpack-fork/src/mlpack/../mlpack/prereqs.hpp:90,
                 from /home/shikhar/Documents/dev/mlpack-fork/src/mlpack/../mlpack/core.hpp:217,
                 from /home/shikhar/Documents/dev/mlpack-fork/src/mlpack/tests/prefixedoutstream_test.cpp:15:
/usr/include/armadillo_bits/arma_ostream_meat.hpp:406:1: note: candidate: template<class eT> static void arma::arma_ostream::print(std::ostream&, const arma::Mat<eT>&, bool)
 arma_ostream::print(std::ostream& o, const Mat<eT>& m, const bool modify)
 ^
/usr/include/armadillo_bits/arma_ostream_meat.hpp:406:1: note:   template argument deduction/substitution failed:
In file included from /home/shikhar/Documents/dev/mlpack-fork/src/mlpack/../mlpack/core/util/prefixedoutstream.hpp:162:0,
                 from /home/shikhar/Documents/dev/mlpack-fork/src/mlpack/../mlpack/core/util/log.hpp:18,
                 from /home/shikhar/Documents/dev/mlpack-fork/src/mlpack/../mlpack/prereqs.hpp:98,
                 from /home/shikhar/Documents/dev/mlpack-fork/src/mlpack/../mlpack/core.hpp:217,
                 from /home/shikhar/Documents/dev/mlpack-fork/src/mlpack/tests/prefixedoutstream_test.cpp:15:
/home/shikhar/Documents/dev/mlpack-fork/src/mlpack/../mlpack/core/util/prefixedoutstream_impl.hpp:179:30: note:   ‘const arma::Op<arma::Col<double>, arma::op_htrans>’ is not derived from ‘const arma::Mat<eT>’
     arma::arma_ostream::print(convert, val, false);

Thanks

rcurtin:
Ah, thanks for testing that rigorously.  I think I know what we can do here.

You can use a little bit of template metaprogramming to catch when you have an Armadillo type that's an operation.  You can use the `arma::Proxy<>` class for this, I think it will work pretty much everywhere.  Take a look at `armadillo_bits/Proxy.hpp` to see what I mean.  I think you could do this...

```
arma::Proxy<T> p;
// These hold printable objects inside... I think "Q" is always available.
p.Q.raw_print(...);
```

shikharbhardwaj:
Thanks. That looks perfect.

shikharbhardwaj:
I ended up using `arma::unwrap`, as `Proxy` was not behaving well with `arma::Op<arma::Mat<double>, arma::op_htrans>`, when I tried to print a transposed matrix. I still understand only some of the template magic `arma` works on.

rcurtin:
Yeah, unwrap is fine, I was thinking that was the other way to do it.

Armadillo is crazy; what it does is very cool but figuring out what is actually going on is like a step into insanity thanks to the templates.

When this builds I'll review it and let you know.

shikharbhardwaj:
This builds fine on my machine, with Armadillo 6.500.5. Some changes do exist in `unwrap.hpp`  between the version used in Travis.

src/mlpack/core/util/prefixedoutstream.hpp
rcurtin: If there's no warning produced by this now, then this line is probably superfluous---Armadillo is being included somewhere earlier than this, so this line isn't needed.

src/mlpack/core/util/prefixedoutstream.hpp
```diff
@@ -124,11 +124,19 @@ class PrefixedOutStream
...
    *
+   * Calls 2 different overloads to be able to output arma objects with custom
+   * precision
+   *
    * @tparam T The type of the data to output.
    * @param val The The data to be output.
    */
   template<typename T>
-  void BaseLogic(const T& val);
+  typename std::enable_if<!arma::is_arma_type<T>::value, void>::type
```
rcurtin: No need for the second template parameter to `enable_if`---it's already default `void`.

src/mlpack/tests/prefixedoutstream_test.cpp
```diff
@@ -104,6 +105,70 @@ BOOST_AUTO_TEST_CASE(TestArmadilloPrefixedOutStream)
...
+  test2.at(2) = 4;
+  ss << std::setprecision(6);
+  ss.unsetf(std::ios::floatfield);
+  ss.str("");
+
+  pss << test2;
+  BOOST_REQUIRE_EQUAL(ss.str(),
+      BASH_GREEN "[INFO ] " BASH_CLEAR "   1.0000   1.5000   2.0000\n"
+      BASH_GREEN "[INFO ] " BASH_CLEAR "   2.5000   3.0000   3.5000\n"
+      BASH_GREEN "[INFO ] " BASH_CLEAR "   4.0000   4.5000   5.0000\n");
```
rcurtin: These tests look great.  If you don't mind, maybe you can split them into different tests?  This causes some duplication since you have to recreate the Armadillo object each time, but it can make it easier to debug when something is going wrong, since you can quickly see that either one test fails or they all fail.

shikharbhardwaj:
Thanks for looking into this.

We probably can't remove the include from `prefixedoutstream.hpp`, as it is required by `log.hpp`, which is included `prereqs.hpp`. Probably the reason why `std` headers are also left intact in this file. The build fails after removing the include. We could perhaps change the include to just `#include <armadillo>`, but I am not too sure about that as you mentioned some grafting happened with macros to add serialisation functionality to Armadillo objects.

I'll split the tests, make the other changes and commit ASAP.

I am already a contributor to `mlpack`! :D

rcurtin:
Oh, I see, the grep command I ran on core.hpp was faulty then. I thought I remembered your contributions from before, so I thought it was strange that the grep produced nothing, but I think I forgot the case insensitive flag or something.  Should have checked harder. :)

There could be strange subtle bugs introduced by including Armadillo at the wrong time, so I think that including `arma_extend.hpp` outside of `prereqs.hpp` (which has special setup before that include) could be problematic.  It seems to me that by the time `log.hpp` is included in `prereqs.hpp`, `arma_extend.hpp` has already been included, so Armadillo symbols will already be available.  Maybe the issue is that some other file included by `prereqs.hpp` is including `log.hpp` before `arma_extend.hpp` is included.  I can look at it when I have a chance, but it may be some days before I have time, so if you'd like to dig in please feel free---I don't think it'll be difficult, just tedious.

shikharbhardwaj:
@rcurtin I have made the changes and the PR is now merge ready. Thanks!

rcurtin:
Thanks.  I made some minor style fixes in f463b88bc.

conradsnicta:
I've noticed that the code uses a whole bunch of internal Armadillo structures and functions being used.  This is not a good idea.

As per the [Armadillo API Version Policy](http://arma.sourceforge.net/docs.html#api_additions), _the underlying internal implementation details may change across consecutive minor versions_. Only the functions explicitly documented in http://arma.sourceforge.net/docs.html are guaranteed to be stable.

stereomatchingkiss:
>I've noticed that the code uses a whole bunch of Armadillo internals. This is not a good idea.

I agree it is not a good idea to use implementation details may change in the future, could you help us point out which part should not be used? I could open a pull request to fill up the gap, implement missing features. Thanks

conradsnicta:
things like unwrap, Proxy, is_arma_type should not be used (directly or indirectly).  Essentially, if the functionality isn't described in http://arma.sourceforge.net/docs.html then it shouldn't be used.

Specific examples -- in the file https://github.com/mlpack/mlpack/blob/master/src/mlpack/core/util/prefixedoutstream_impl.hpp
line 38: typename std::enable_if<!arma::is_arma_type<T>::value>::type
line 146; typename std::enable_if<arma::is_arma_type<T>::value>::type
line 150: auto printVal = arma::unwrap<T>(val).M;
line 166: arma::arma_ostream::print(convert, printVal, true);
line 175: auto absVal = arma::abs(printVal).P.Q;
line 185:  arma::arma_ostream::print(convert, printVal, false);

The code around line 146 can replaced with something along these lines:
    template<typename T>
    void
    PrefixedOutStream::BaseLogic(const arma::Base<typename T::elem_type, T>& val)

rcurtin:
To some extent we can't get away from using the undocumented internal API.  For what it's worth, even `arma::Base<>` is not actually documented in docs.html, so technically that would also be the internal API too.  (Though I agree, `arma::Base<>` is certainly far less likely to change and I think it's a better solution.)

I think the `auto` usage is fine here (`M` will not be `Op`/`Gen`/anything that might confuse the compiler).  If we don't use `auto`, we'll have to build some template metaprogramming struct to get the right type, because it could be `Mat`, `Col`, or `Row` depending.  But, I agree, Armadillo+`auto` can be dangerous in a lot of situations.  Do you have an alternative for what we are doing here?

I don't see a good way to escape using `unwrap<>` either; we need some way to take what we might have (which might be an expression of some form) and convert it to some block of memory that we can actually print.  I can do something like `Mat<T::elem_type> m = val` but this incurs a copy; I'd like to avoid copies in general.

I also don't think we can escape using `arma_ostream::print()` for the purposes of this method, and I'm not excited about reimplementing that ourselves (which is the only realistic way I could see to avoid using that function).

In either case, I know that we are breaking the warranty and we get to pick up the pieces if something breaks or changes in later releases. :)

conradsnicta:
If I understand the rules for the `auto` keyword, the line `auto printVal = arma::unwrap<T>(val).M;` does a copy anyway. The line `auto absVal = arma::abs(printVal).P.Q` can lead to random crashes if printVal is an expression. At the very least I suggest using `const auto&` instead.

A betters solution would be using `const Mat<typename T::elem_type>& printVal(val)`, which should avoid a copy if the `val` expression is a matrix. If `val` is an expression, the compiler will automatically evaluate the expression and stick it into a temporary matrix.  Code along these lines will lead to the avoidance of using the interal `unwrap` and `Proxy` classes.

As for using `arma_ostream::print()`, I believe using `Mat::print(outputstream)` or `Mat::raw_print(outputstream)` would be preferable. This also avoids the use of internal functions.

Anyway, I'm not sure it's worth worrying about potential extra copies, especially if this comes at the cost of breaking things later and an increased maintenance burden (ie. technical debt). The most time consuming part of many algorithms is matrix multiplication, not copies.

rcurtin:
Ah, yeah, I agree, `const auto&` would be better, and the `const Mat<typename T::elem_type>&` is a better solution still.  I think we should go with that one.

I think, but am not sure, that `raw_print()` could work for our situation here also; we could print that to a `ostringstream` and then send the `ostringstream`'s contents to the Log object (which for whatever reason is not an `ostream` itself).

In this particular case, you are exactly right too, the copy doesn't matter.  If a user is printing a matrix that's so large that a copy matters, well, they're already making a bad choice.  However, I like to push for efficient code where possible regardless, so that new contributors or people reading the code can pick up good habits for fast code.  It's easier to internalize "always avoid copies" instead of "avoid copies where it matters, and where it matters is something you have to think about", and that mindset also makes the maintenance burden easier for me when I evaluate PRs.

@shikharbhardwaj: are you interested in making these changes, or should I open another issue?

shikharbhardwaj:
I tried to use `const auto&` to bind to the member from temporary from `unwrap` first, but as it turns out, doing so does not lead to a lifetime extension for the temporary object created, as noted [here.](https://herbsutter.com/2008/01/01/gotw-88-a-candidate-for-the-most-important-const/) Probably why the code segfaulted when I did use the const ref. Which means we'd need to bind to the temporary and access members afterwards.

@rcurtin I'll make the changes.

Issue #893, opened 2017-02-28
More optimizer methods


zoq:
An optimizer is one of basically two parts to evaluate/optimize a method. mlpack already implements some optimizers like Stochastic Gradient Descent or RmsProp but there are still some interesting approaches out there that could be of interest. In particular, I would really like to see an implementation of the SMORMS3[1], the Nesterov's Accelerated Gradient[2] and AdaGrad[3] method.

1.  Simon Funk's SMORMS3 - http://sifter.org/~simon/journal/20150420.html
2. Advances in Optimizing Recurrent Networks, Yoshua Bengio et al. - https://arxiv.org/abs/1212.0901
3. Adaptive Subgradient Methods for Online Learning and Stochastic Optimization, John Duchi et al. - http://jmlr.org/papers/v12/duchi11a.html

The goal of this ticket is to implement the mentioned methods and to write tests for each method like we already did for the existing optimizer methods. Realistically you can look at some of the other optimizers in ``src/mlpack/core/optimizers`` for examples; The corresponding test suites are in ``src/mlpack/tests``. Note, if you write a suite called ``TestSuite`` ``(BOOST_AUTO_TEST_SUITE(TestSuite))``, and then build ``mlpack_test`` (``make mlpack_test``), you can run only the tests in that test suite with ``bin/mlpack_test -t TestSuite``.  And a specific test case called ``TestCase (BOOST_AUTO_TEST_CASE(TestCase))`` could be run with 'bin/mlpack_test -t TestSuite/TestCase'.

ivmarkp:
Hi @zoq, I would like to work on this task. Will open separate PRs for each.

zoq:
Hello @ivmarkp sounds good. Do you mind if we split the work, maybe someone else is also interested in implementing one of mentioned optimization algorithms? So maybe can you start with one, and if there is an algorithm left, once you finished you take the next one. What do you think? We could use this issue as an entrance coordination point, so if you agree just tell me the algorithm you like to start with and I modify the issue description accordingly.

nish21:
If @ivmarkp agrees, I'd like to work on this as well.

hxidkd:
if there is any optimizer methods left, I'd like to work on this as well.

ivmarkp:
Hi @zoq, you are right. Splitting the work sounds like a good idea as its meant to be an entrance task. I've actually started working on the first optimizer i.e. SMORMS3 so I'd like to continue working on it.

[...]

PR #896, opened 2017-03-01
SGD with momentum update.

arunreddy:
Implemented SGD with the momentum update towards  #893.

1. Implemented based on the sgd momentum algorithm from Deep learning book (http://www.deeplearningbook.org/) - Section 8.3.2
2. Added a hyperparamter `momentum` and set the default value to 0.5.
3. Added a test case similar to other optimizer test cases.

### DIFF
--- a/src/mlpack/core/optimizers/momentum_sgd/CMakeLists.txt
+++ b/src/mlpack/core/optimizers/momentum_sgd/CMakeLists.txt
@@ -1,0 +1,11 @@
+set(SOURCES
+  momentum_sgd.hpp
+  momentum_sgd_impl.hpp
+)
+
+set(DIR_SRCS)
+foreach(file ${SOURCES})
+  set(DIR_SRCS ${DIR_SRCS} ${CMAKE_CURRENT_SOURCE_DIR}/${file})
+endforeach()
+
+set(MLPACK_SRCS ${MLPACK_SRCS} ${DIR_SRCS} PARENT_SCOPE)
--- a/src/mlpack/core/optimizers/momentum_sgd/momentum_sgd.hpp
+++ b/src/mlpack/core/optimizers/momentum_sgd/momentum_sgd.hpp
@@ -1,0 +1,178 @@
+/**
+ * @file momentum_sgd.hpp
+ * @author Ryan Curtin
+ * @author Arun Reddy
+ *
+ * Stochastic Gradient Descent(SGD) with Momentum update (MomentumSGD) is an approach
+ * that enjoys accelerated convergence rates compared to vanilla SGD.
+ *
+ * mlpack is free software; you may redistribute it and/or modify it under the
+ * terms of the 3-clause BSD license.  You should have received a copy of the
+ * 3-clause BSD license along with mlpack.  If not, see
+ * http://www.opensource.org/licenses/BSD-3-Clause for more information.
+ */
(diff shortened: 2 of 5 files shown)


rcurtin:
There's a lot of code duplication here with the regular SGD optimizer.  I think we can reduce that by refactoring the SGD class to include a template parameter that either represents whether or not momentum should be used, or a template class that defines the momentum to be applied to each step.  Let me know what you think of that idea.

arunreddy:
@rcurtin I completely agree that lot of code is duplicated. MomentumSGD is same as SGD with 5 new lines of code. How about adding a momentum parameter, when the `momentum=0`  it falls back to vanilla SGD.

rcurtin:
The only problem with a parameter like that is that it will be evaluated at runtime. :)

If we make it a compile-time parameter (i.e. a template parameter), then the compiled code we get doesn't check the value of momentum at every iteration.  Since SGD is kind of an inner loop to a lot of calculations, I think it's important here to make it as fast as possible.

Now, you could use a template parameter that's a boolean, and then check the value of that inside a function, and that will be optimized away at compile-time.  That'll work for this momentum strategy, but if someone wants to implement some other variant of momentum later, it will not work there.  That's why I suggested a policy class (or template class parameter) there, but if someone wanted another variant of momentum later, they could also do the refactoring to a template class parameter at that time.

arunreddy:
Yeah, it is going to evaluated every time. If I understand your approach right.. 

```

**Comment by arunreddy** [comment]
Yeah, it is going to evaluated every time. If I understand your approach right.. we can have something like `momentum_update` as template class parameter along with `DecomposableFunctionType`..

``` c++
template<typename DecomposableFunctionType,bool momentum_update>
class SGD
{ 
...
}
```

The object can be instantiated as:
``` c++
// No momentum update
SGD<LogisticRegressionFunction<>,false> sgd(lrf);  

// With momentum update
SGD<LogisticRegressionFunction<>,true> sgd(lrf);  
```

and in the iteration loop we will have
``` c++
    // And update the iterate.
   if(momentum_update){
      v = momentum*v - stepSize * gradient;
      iterate += v;
   }else{
      iterate -= stepSize * gradient;
   }
```
that optimizes the code in the compile time.

arunreddy:
I get the whole point of policy class now. So can I create `MomentumPolicy` type that is passed to `SGD` and create a two classes a default `NoMomentumPolicy` with flag set to `false` and a `SimpleMomentumPolicy` with the flag set to `true` and the hyperparameter?

arunreddy:
Can I follow something similar to https://github.com/mlpack/mlpack/blob/master/src/mlpack/methods/amf/amf.hpp?

arunreddy:
@rcurtin: Please ignore my last two comments. 

After discussing with @zoq I have introduced a template param `UpdatePolicyType`, which defaults to EmptyUpdate (vanilla SGD). The updates with momentum can reside in MomentumUpdate. 

Here is the work in progress: https://github.com/arunreddy/mlpack/tree/sgd_momentum_policy/src/mlpack/core/optimizers/sgd

rcurtin:
Yep, that is the basic idea that I was thinking.  Sorry that I was not able to respond yesterday.  This is nice because it allows us to make many different variants of SGD that can be dropped in place.

arunreddy:
Thanks, working in that direction.

rcurtin:
Closing in favor of #925.

PR #897, opened 2017-03-01
AdaGrad optimizer

abhinavmoudgil95:
Implementation of AdaGrad optimizer. All tests passed locally. #893


abhinavmoudgil95:
@zoq After my 4th commit [#60333b](https://github.com/mlpack/mlpack/pull/897/commits/603333be3e1ec1dea6af9a3e4404d403abcb355d) in this PR, travis build was passing. I just updated authors and email address by making changes in the comments of code, now travis is showing red signal. Could you please review and let me know what is really happening?

abhinavmoudgil95:
@zoq After my 4th commit [#603333b](https://github.com/mlpack/mlpack/pull/897/commits/603333be3e1ec1dea6af9a3e4404d403abcb355d) in this PR, travis build was passing. I just updated authors and email address by making changes in the comments of code, now travis is showing red signal. Could you please review and let me know what is really happening?

abhinavmoudgil95:
The pull request is ready to be merged. Kindly review. 
Thanks!

abhinavmoudgil95:
@zoq The pull request is ready to be merged. Kindly review.
Thanks!

src/mlpack/core/optimizers/adagrad/ada_grad.hpp
```diff
@@ -0,0 +1,155 @@
...
+ * chooses learning rate dynamically by adapting to the data. Hence Adagrad
+ * eliminates the need to manually tune the learning rate.
+ *
+ * mlpack is free software; you may redistribute it and/or modify it under the
+ * terms of the 3-clause BSD license.  You should have received a copy of the
+ * 3-clause BSD license along with mlpack.  If not, see
+ * http://www.opensource.org/licenses/BSD-3-Clause for more information.
+ */
+#ifndef __MLPACK_CORE_OPTIMIZERS_ADAGRAD_ADA_GRAD_HPP
+#define __MLPACK_CORE_OPTIMIZERS_ADAGRAD_ADA_GRAD_HPP
```
rcurtin: We shouldn't use `__` as a prefix to the macros, just `MLPACK_CORE_OPTIMIZERS_ADAGRAD_ADA_GRAD_HPP` will work fine.  (That applies to the other files you've added too.)

rcurtin: As an additional thought, the directory is named `adagrad` but the file is named `ada_grad.hpp`; I think it would be better if we settled on one or the other.  I don't have an opinion on which is better, I'm mostly concerned that we're consistent, is all.

src/mlpack/core/optimizers/adagrad/ada_grad.hpp
```diff
@@ -0,0 +1,155 @@
...
+ * parameters.
+ *
+ * For more information, see the following.
+ *
+ * @code
+ * @article{duchi2011adaptive,
+ *   author    = {Duchi, John and Hazan, Elad and Singer, Yoram},
+ *   title     = {Adaptive subgradient methods for online learning and stochastic optimization},
+ *   journal   = {Journal of Machine Learning Research},
+ *   year      = {2011}
```
rcurtin: This is really pedantic... :)  You can add these too the reference also to make it more complete:

```
  volume={12},
  number={Jul},
  pages={2121--2159},
```

src/mlpack/core/optimizers/adagrad/ada_grad.hpp
```diff
@@ -0,0 +1,155 @@
...
+   * @param eps Value used to initialise the squared gradient parameter.
+   * @param maxIterations Maximum number of iterations allowed (0 means no
+   *        limit).
+   * @param tolerance Maximum absolute tolerance to terminate algorithm.
+   * @param shuffle If true, the function order is shuffled; otherwise, each
+   *        function is visited in linear order.
+   */
+  AdaGrad(DecomposableFunctionType& function,
+      const double stepSize = 0.01,
+      const double eps = 1e-8,
```
rcurtin: The name `eps` may confuse users, who might think that `eps` is the tolerance before termination.  I might consider a different name here.

src/mlpack/core/optimizers/adagrad/ada_grad.hpp
abhinavmoudgil95: Actually, I adopted this nomenclature of macros and directory name from adadelta implementation. To make things consistent, should I update adadelta implementation macros and directory name too with a separate PR?

src/mlpack/core/optimizers/adagrad/ada_grad.hpp
abhinavmoudgil95: Done.

src/mlpack/core/optimizers/adagrad/ada_grad.hpp
abhinavmoudgil95: Yes, updated `eps` to `epsilon`.

abhinavmoudgil95:
@rcurtin Thanks a lot Ryan for the detailed review. :) I updated the PR with the changes you suggested.

I agree that AdaGrad is a modified version of SGD but I am not sure if we should update SGD implementation with AdaGrad too. There is already work going on in updating SGD with momentum and further populating it with AdaGrad would make things confusing for the user. In all the other toolkits I have used, there is always a separate implementation for AdaGrad, hence I feel it is a bit unconventional to add AdaGrad to SGD as most of the users look for AdaGrad optimizer directly.  Also, AdaGrad contains an extra fuzz factor `epsilon` which would be unused in SGD. 

These are just my views, I am open to making changes with your suggestions.

src/mlpack/core/optimizers/adagrad/ada_grad.hpp
abhinavmoudgil95: Actually, I adopted this nomenclature of macros and directory from adadelta implementation. To make things consistent, should I update adadelta implementation macros and directory name too with a separate PR?

abhinavmoudgil95:
@rcurtin Thanks a lot for the detailed review. :) I updated PR with the changes you suggested.

I agree that AdaGrad is a modified version of SGD but I am not sure if we should update SGD implementation with AdaGrad too. There is already work going on in updating SGD with momentum and further populating it with AdaGrad would make things confusing for the user. In all the other toolkits I have used, there is always a separate implementation for AdaGrad, hence I feel it is a bit unconventional to add AdaGrad to SGD as most of the users look for AdaGrad optimizer directly.  Also, AdaGrad contains an extra fuzz factor `epsilon` which would be unused in SGD. 

These are just my views, I am open to making changes with your suggestions.

src/mlpack/core/optimizers/adagrad/ada_grad.hpp
abhinavmoudgil95: Actually, I adopted this nomenclature of macros and directory from AdaDelta implementation. To make things consistent, should I update AdaDelta implementation macros and directory name too with a separate PR?

abhinavmoudgil95:
I have a proposal to resolve code duplication across all the optimizers. Rather than having a seperate classes for each optimizer, we can just have one base optimizer class which implements the following: 

### Base Optimizer Class
```c++
// default optimizer is SGD
template<typename DecomposableFunctionType,
	         typename OptimizerType = SGD>
class Optimizer
{

Optimizer(DecomposableFunctionType& function,
	      OptimizerType optimizerType,
	      const double stepSize = 0.01,
	      const double tolerance = 1e-5,
	      const bool shuffle = true);

double Optimize(arma::mat& iterate);
double StepSize() const { return stepSize; }
bool Shuffle() const { return shuffle; }
double Tolerance() const { return tolerance; }
size_t MaxIterations() const { return maxIterations; }
};
```
Implementation of `Optimize` function in the base `Optimizer` class would almost remain the same except with few minor changes as below: 
```c++
template<typename DecomposableFunctionType, typename OptimizerType>
double Optimizer<DecomposableFunctionType, OptimizerType>::Optimize(arma::mat& iterate)
{
      // same old code for finding number of functions and calculating objective function

      // initialize optimizer
      OptimizerType.initialize(iterate.n_rows,iterate.n_cols);
      
      // Iteration loop
      arma::mat gradient(iterate.n_rows, iterate.n_cols);
      for (size_t i = 1; i != maxIterations; ++i, ++currentFunction)
      {
              // same old code till evalution of gradient

              // take optimizer step
              iterate -= OptimizerType.TakeStep(stepSize, gradient);
      }
	
      // same old code, 
      return overallObjective;
};
```
### Optimizer Classes

For each optimizer, we can create a template class like the following: 
```c++
template<class OptimizerType>
class AdaGrad {

public: 

AdaGrad(const double epsilon = 1e-8); 
void Initialize(const size_t n_rows = 0, const size_t n_cols = 0)
{
	// initialize squaredGradient
}
arma::mat TakeStep(const double stepSize, arma::mat& gradient)
{
	// AdaGrad step code goes here
       squaredGradient += (gradient % gradient);
       return (stepSize * gradient) / (arma::sqrt(squaredGradient) + epsilon);
}

private: 
	double epsilon;
	arma::mat squaredGradient;
}
```
Similarly other optimizers for example, RMSProp can be implemented as follows:
```c++
template<class OptimizerType>
class RMSProp {

public: 

RMSProp(const double epsilon = 1e-8, alpha = 0.99); 
void Initialize(const size_t n_rows = 0, const size_t n_cols = 0)
{
	// initialize meanSquaredGradient
}
arma::mat TakeStep(const double stepSize, arma::mat& gradient)
{
	meanSquaredGradient *= alpha;
	meanSquaredGradient += (1 - alpha) * (gradient % gradient);
	return stepSize * gradient / (arma::sqrt(meanSquaredGradient) + epsilon);

}

private: 
	double epsilon;
	double alpha; 
	arma::mat meanSquaredGradient;
};
```
Please let me know how this sounds.

abhinavmoudgil95:
I have a proposal to resolve code duplication across all the optimizers. 

### Base Optimizer Class
Rather than having a seperate classes for each optimizer, we can just have one base optimizer class which implements the following: 
```c++
// default optimizer is SGD
template<typename DecomposableFunctionType,
	         typename OptimizerType = SGD>
class Optimizer
{

Optimizer(DecomposableFunctionType& function,
	      OptimizerType optimizerType,
	      const double stepSize = 0.01,
	      const double tolerance = 1e-5,
	      const bool shuffle = true);

double Optimize(arma::mat& iterate);
double StepSize() const { return stepSize; }
bool Shuffle() const { return shuffle; }
double Tolerance() const { return tolerance; }
size_t MaxIterations() const { return maxIterations; }
};
```
Implementation of `Optimize` function in the base `Optimizer` class would almost remain the same except with few minor changes as below: 
```c++
template<typename DecomposableFunctionType, typename OptimizerType>
double Optimizer<DecomposableFunctionType, OptimizerType>::Optimize(arma::mat& iterate)
{
      // same old code for finding number of functions and calculating objective function

      // initialize optimizer
      OptimizerType.initialize(iterate.n_rows,iterate.n_cols);
      
      // Iteration loop
      arma::mat gradient(iterate.n_rows, iterate.n_cols);
      for (size_t i = 1; i != maxIterations; ++i, ++currentFunction)
      {
              // same old code till evalution of gradient

              // take optimizer step
              iterate -= OptimizerType.TakeStep(stepSize, gradient);
      }
	
      // same old code, 
      return overallObjective;
};
```
### Optimizer Classes

For each optimizer, we can create a template class like the following: 
```c++
template<class OptimizerType>
class AdaGrad {

public: 

AdaGrad(const double epsilon = 1e-8); 
void Initialize(const size_t n_rows = 0, const size_t n_cols = 0)
{
	// initialize squaredGradient
}
arma::mat TakeStep(const double stepSize, arma::mat& gradient)
{
	// AdaGrad step code goes here
       squaredGradient += (gradient % gradient);
       return (stepSize * gradient) / (arma::sqrt(squaredGradient) + epsilon);
}

private: 
	double epsilon;
	arma::mat squaredGradient;
}
```
Similarly other optimizers for example, RMSProp can be implemented as follows:
```c++
template<class OptimizerType>
class RMSProp {

public: 

RMSProp(const double epsilon = 1e-8, alpha = 0.99); 
void Initialize(const size_t n_rows = 0, const size_t n_cols = 0)
{
	// initialize meanSquaredGradient
}
arma::mat TakeStep(const double stepSize, arma::mat& gradient)
{
	meanSquaredGradient *= alpha;
	meanSquaredGradient += (1 - alpha) * (gradient % gradient);
	return stepSize * gradient / (arma::sqrt(meanSquaredGradient) + epsilon);

}

private: 
	double epsilon;
	double alpha; 
	arma::mat meanSquaredGradient;
};
```
This is just a basic prototype, please let me know how this sounds.

rcurtin:
Hi Abhinav,

To clarify, you would not need to refer to the `AdaGrad` class as `SGD`.  You could instead use a template alias to make it clear:

```
using AdaGrad = SGD<DecomposableFunctionType, AdaGradStep>;
```

To your proposal, I'm having a difficult time seeing what the advantage of this over a template parameter to `SGD` is.  As a side note, not all optimizers behave like `SGD` so calling this proposed class `Optimizer` is way overly general.

One big issue here---and also with the template based approach---is that to match the rest of the mlpack codebase we should not have an `Initialize()` function, and therefore the question of how to cleanly pass optimizer-specific parameters like `epsilon` is unsolved.

I think, perhaps, it may be better for now to allow the code duplication, and then when we have more optimizers revisit the problem and consider refactoring at that time.  @zoq: any opinion on that plan?  If you already had a design in mind then we can go with that instead.  I would need to spend some more time with the problem to come up with something.

zoq:
@rcurtin I agree I have to think about a good solution for the problem, so I would also go with the duplication for now.

abhinavmoudgil95:
@rcurtin Basically, I had an idea to have a general base class for all optimizers and then we can implement several optimizers on top of it. I observed that in all the optimizers, `Optimize` function was almost same except the update rule. But yes, I agree with your view points. Templating `Optimizer` class like that is way overly general and it may not match rest of the mlpack codebase. 

I worked on your reviews in this PR. Please suggest further edits, if any.

src/mlpack/core/optimizers/ada_grad/ada_grad.hpp
```diff
@@ -0,0 +1,158 @@
...
+ *   title     = {Adaptive subgradient methods for online learning and stochastic optimization},
+ *   journal   = {Journal of Machine Learning Research},
+ *   volume    = {12},
+ *   number    = {Jul},
+ *   pages     = {2121--2159},
+ *   year      = {2011}
+ * }
+ * @endcode
+ *
+
```
rcurtin: Seems like there is an empty line here we could remove.

abhinavmoudgil95:
@rcurtin Can you please elaborate this? 
>  I think that that templatization could be perfect for the different way that AdaGrad works, and then we could supply a template typedef for AdaGrad

I am a bit confused. 
Do you mean that we should templatize AdaGrad for implementing different update policies in it like SGD but keeping its implementation separate from SGD?  

OR 

We should implement AdaGrad policy update rule in SGD?

rcurtin:
I mean the latter---implementing the AdaGrad policy update rule as a template parameter to SGD, then providing a template typedef to make clear that it is AdaGrad:

```
using AdaGrad = SGD<AdaGradUpdate>;
```

or something similar to this.

zoq:
In this solution how do we pass optimizer-specific parameters like epsilon via the Policy constructor? Another concern is how do we provide meaningful default values for the optimizer?

rcurtin:
You're right, I overlooked that.  Maybe the better idea then is a wrapper class, something like:

```
class AdaGrad
{
 public:
  AdaGrad(DecomposableFunctionType& function, const double stepSize, const double epsilon, ...);

  double Optimize(const arma::mat& parameters)
  {
    return optimizer.Optimize(parameters);
  }

 private:
  SGD<AdaGradUpdate> optimizer;
};
```

So we have avoided code duplication by using `SGD<AdaGradUpdate>` internally in the `AdaGrad` class, but we can also provide optimizer-specific parameters directly in the constructor.  The constructor could then create the `SGD<AdaGradUpdate>` object with the desired parameters.  What do you think?

zoq:
I agree, a wrapper class would solve the mentioned problems and would minimize code duplication. I'm inclined to do the change now instead of doing it later. @abhinavmoudgil95 @ivmarkp what do you think?

ivmarkp:
@rcurtin @zoq I'm happy with that solution. I think it's in our best interest to minimize code duplication wherever we can. I'll restructure the SMORMS3 implementation accordingly.

ivmarkp:
@rcurtin Small doubt in the following `Optimize` function in wrapper class `AdaGrad` here:

```cpp

double Optimize(const arma::mat& parameters)
{
  return optimizer.Optimize(parameters);
}

```
I think may be we should use `iterate` instead of `parameters` to be consistent with current implementations of other optimizers?

zoq:
@ivmarkp I agree, that was just a draft, unfortunately we still have to provide the same function set as for SGD but we can add new function as well e.g. the Epsilon function.

abhinavmoudgil95:
@zoq Sure, it's a nice solution. I'll refactor my PR and get back soon.

abhinavmoudgil95:
@zoq I guess we should wait for #925 to get merged and then add our new updates to SGD?

zoq:
@abhinavmoudgil95 Your decision, you can pull Arun's code and locally work with that, once it's merged we can restart the CI jobs, but you can also wait until we merged the code.

abhinavmoudgil95:
@rcurtin @zoq I have created a separate pull request #942 which implements AdaGrad update policy class which follows the same format as SGD momentum update class. Similar tests have been added for AdaGrad. Kindly review.

abhinavmoudgil95:
@rcurtin @zoq I have created a separate pull request #942 which implements AdaGrad update policy class. It follows the same format as SGD momentum update class and similar tests have been added for AdaGrad. Kindly review.

abhinavmoudgil95:
Closing this for now. #942 contains all the new updates as per the discussion here.

PR #899, opened 2017-03-01
Implementation of SMORMS3 optimizer

ivmarkp:
This PR implements the SMORMS3 optimizer, addressing ticket #893. All tests pass locally. Please review. Thanks!


src/mlpack/core/optimizers/smorms3/smorms3_impl.hpp
```diff
@@ -0,0 +1,144 @@
...
+
+    g = (1 - r) % g;
+    g += r % gradient;
+
+    g2 = (1 - r) % g2;
+    g2 += r % (gradient % gradient);
+
+    arma::mat x = (g % g) / (g2 + eps);
+
+    arma::mat lRateMat(x.n_rows, x.n_cols);
```
ivmarkp: > It's odd that we have to create another matrix of size x

I think we could do with size of iterate as well, as both are really the same size matrices (x and iterate).
> I think we should avoid the recreation of the learning matrix for every iteration, instead we could create it once and update it every iteration. What do you think?

You are right. In fact, I see that we don't need to update or fill `lRateMat` with `lRate` in every iteration as `lRate` doesn't change during the iterations. So, both the statements in line 116 and 117 can be moved outside the for loop. What do you suggest?

P.S. Sorry for the delayed response. Have been busy with preparing for exams starting from tomorrow.

ivmarkp: Hi @zoq, I was wondering if you had a chance to take a second look. Waiting on your input, thanks.

src/mlpack/core/optimizers/smorms3/smorms3.hpp
```diff
@@ -0,0 +1,153 @@
...
+   */
+  double Optimize(arma::mat& iterate);
+
+  //! Get the instantiated function to be optimized.
+  const DecomposableFunctionType& Function() const { return function; }
+  //! Modify the instantiated function.
+  DecomposableFunctionType& Function() { return function; }
+
+  //! Get the learning rate.
+  double LRate() const { return lRate; }
```
zoq: Do you mind to call this function LearningRate I guess, StepSize could also work?

src/mlpack/core/optimizers/smorms3/smorms3.hpp
```diff
@@ -0,0 +1,153 @@
...
+   * @param eps Value used to initialise the mean squared gradient parameter.
+   * @param maxIterations Maximum number of iterations allowed (0 means no
+   *        limit).
+   * @param tolerance Maximum absolute tolerance to terminate algorithm.
+   * @param shuffle If true, the function order is shuffled; otherwise, each
+   *        function is visited in linear order.
+   */
+  SMORMS3(DecomposableFunctionType& function,
+      const double lRate = 0.001,
+      const double eps = 1e-16,
```
zoq: Do you think we should rename the eps parameter to epsilon, since later you use ``double Epsilon() const { return eps; }``to return the current value.

src/mlpack/core/optimizers/smorms3/smorms3.hpp
```diff
@@ -0,0 +1,153 @@
...
+ * SMORMS3 i.e. squared mean over root mean squared cubed optimizer. It is a
+ * hybrid of RMSprop, which estimates a safe and optimal distance based on
+ * curvature and Yann LeCun’s method in "No more pesky learning rates".
+ *
+ * mlpack is free software; you may redistribute it and/or modify it under the
+ * terms of the 3-clause BSD license.  You should have received a copy of the
+ * 3-clause BSD license along with mlpack.  If not, see
+ * http://www.opensource.org/licenses/BSD-3-Clause for more information.
+ */
+#ifndef __MLPACK_CORE_OPTIMIZERS_SMORMS3_SMORMS3_HPP
```
zoq: We shouldn't use __ as a prefix to the macros, just MLPACK_CORE_OPTIMIZERS_SMORMS3_SMORMS3_HPP will work fine. (That applies to the other files you've added too.)

src/mlpack/core/optimizers/smorms3/smorms3_impl.hpp
zoq: Moving anything  we can outside the for loop is good.

ivmarkp: Okay, got it. I'll update the PR accordingly.

src/mlpack/core/optimizers/smorms3/smorms3.hpp
ivmarkp: Yes, sure. Will rename it as epsilon.

src/mlpack/core/optimizers/smorms3/smorms3.hpp
ivmarkp: Okay, may be let's go with LearningRate.

ivmarkp:
@zoq Updated the PR with SMORMS3 update policy class as discussed previously. Please review, thanks!

src/mlpack/core/optimizers/smorms3/smorms3.hpp
```diff
@@ -0,0 +1,157 @@
...
+    return optimizer.Optimize(iterate);
+  }
+
+  //! Get the instantiated function to be optimized.
+  const DecomposableFunctionType& Function() const { return function; }
+  //! Modify the instantiated function.
+  DecomposableFunctionType& Function() { return function; }
+
+  //! Get the step size.
+  double StepSize() const { return stepSize; }
```
zoq: Should return optimizer.StepSize() to reflect changes in the optimizer instantiation. The same applies to all parameters.

src/mlpack/core/optimizers/sgd/update_policies/smorms3_update.hpp
```diff
@@ -0,0 +1,108 @@
...
+ * @code
+ * @misc{Funk2015,
+ *   author    = {Simon Funk},
+ *   title     = {RMSprop loses to SMORMS3 - Beware the Epsilon!},
+ *   year      = {2015}
+ * }
+ * @endcode
+ */
+
+class SMORMS3Update
```
zoq: I would put the header file in the smorms3 folder, it's not necessary related with the SGD class.

src/mlpack/core/optimizers/sgd/update_policies/smorms3_update.hpp
```diff
@@ -0,0 +1,108 @@
...
+
+  /**
+   * Update step for SMORMS3.
+   *
+   * @param iterate Parameter that minimizes the function.
+   * @param stepSize Step size to be used for the given iteration.
+   * @param gradient The gradient matrix.
+   */
+  void Update(arma::mat& iterate,
+              double stepSize,
```
zoq: Please use const for ``stepSize`` and ``gradient``.

src/mlpack/core/optimizers/sgd/sgd.hpp
```diff
@@ -16,6 +16,7 @@
 #include <mlpack/prereqs.hpp>
 #include <mlpack/core/optimizers/sgd/update_policies/vanilla_update.hpp>
 #include <mlpack/core/optimizers/sgd/update_policies/momentum_update.hpp>
+#include <mlpack/core/optimizers/sgd/update_policies/smorms3_update.hpp>
```
zoq: I don't think we have to include the smorms3_update header file here, we include vanilla and momentum header files, to create an alias where this to classes are necessary.

src/mlpack/core/optimizers/sgd/update_policies/smorms3_update.hpp
```diff
@@ -0,0 +1,108 @@
...
+ * @endcode
+ */
+
+class SMORMS3Update
+{
+ public:
+  /**
+   * Construct the SMORMS3 update policy with given epsilon parameter.
+   *
+   * @param psilon Value used to initialise the mean squared gradient parameter.
```
zoq: Misspelling -> epsilon not psilon, also I think the description should match with the parameters so, we should use epsilon and not eps.

src/mlpack/core/optimizers/sgd/update_policies/smorms3_update.hpp
```diff
@@ -0,0 +1,108 @@
...
+
+  /**
+   * The Initialize method is called by SGD::Optimize method with UpdatePolicy
+   * SMORMS3Update before the start of the iteration update process.
+   *
+   * @param n_rows number of rows in the gradient matrix.
+   * @param n_cols number of columns in the gradient matrix.
+   * @param lRate 
+   */
+  void Initialize(const size_t n_rows,
```
zoq: I think we missed this in the momentum sgd PR, we use camel casing for all names. I guess we can just go with rows and cols.

src/mlpack/core/optimizers/sgd/update_policies/smorms3_update.hpp
```diff
@@ -0,0 +1,108 @@
...
+
+    g = (1 - r) % g;
+    g += r % gradient;
+
+    g2 = (1 - r) % g2;
+    g2 += r % (gradient % gradient);
+
+    arma::mat x = (g % g) / (g2 + eps);
+
+    stepSizeMat.fill(stepSize);
```
zoq: Instead of resetting the stepSizeMat in each iteration we could check if stepSize != previousStepSize and reset the matrix only if the condition is true. What do you think?

src/mlpack/tests/smorms3_test.cpp
```diff
@@ -0,0 +1,110 @@
...
+BOOST_AUTO_TEST_SUITE(SMORMS3Test);
+
+/**
+ * Tests the SMORMS3 optimizer using a simple test function.
+ */
+BOOST_AUTO_TEST_CASE(SimpleSMORMS3TestFunction)
+{
+  SGDTestFunction f;
+  SMORMS3Update smorms3Update(1e-16);
+  SGD<SGDTestFunction, SMORMS3Update> s(f, 1e-3, 5000000, 1e-9, true, smorms3Update);
```
zoq: I would use SMORMS3 instead of ``SMORMS3Update smorms3Update(1e-16);`` in combination with the ``SGD<SGDTestFunction, SMORMS3Update>`` that was the idea of introducing a new class.

ivmarkp: May be I didn't understand correctly; you mean to use SMORMS3 constructor like
```cpp
SMORMS3<SGDTestFunction> s(f, 1e-16, 1e-3, 5000000, 1e-9, true);
```
here in place of 
```cpp
SMORMS3Update smorms3Update(1e-16);
SGD<SGDTestFunction, SMORMS3Update> s(...);
```
or define a new template typedef like we did for `StandardSGD` and `MomentumSGD` in sgd.hpp?

zoq: To use ``SMORMS3<SGDTestFunction> s(f, 1e-16, 1e-3, 5000000, 1e-9, true);``

ivmarkp: Ok, thanks for the quick review. Working now to address your comments.

ivmarkp: @zoq I've made the changes but there's a problem with compilation now. Build fails with
```bash
error: invalid use of member function (did you forget the ‘()’ ?)
``` 
at `double Optimize(arma::mat& iterate) { return optimizer.Optimize(iterate); }`.

Detailed error [here](https://paste.debian.net/920268/). Couldn't find anything helpful related to this error elsewhere. What do you suggest?

ivmarkp:
@zoq What was the other issue you were referring to? Is it fixed now?

src/mlpack/core/optimizers/smorms3/smorms3_impl.hpp
```diff
@@ -0,0 +1,39 @@
...
+
+template<typename DecomposableFunctionType>
+SMORMS3<DecomposableFunctionType>::SMORMS3(DecomposableFunctionType& function,
+                                           const double stepSize,
+                                           const double epsilon,
+                                           const size_t maxIterations,
+                                           const double tolerance,
+                                           const bool shuffle) :
+    function(function),
+    stepSize(stepSize),
```
zoq: No need to save the ``stepSize``or ``maxIterations``etc here, since all those parameters are already initialized via the optimizer instantiation and we access them via optimizer object e.g. `òptimizer.StepSize()``.

src/mlpack/core/optimizers/smorms3/smorms3.hpp
```diff
@@ -0,0 +1,159 @@
...
+  //! Modify the instantiated function.
+  DecomposableFunctionType& Function() { return function; }
+
+  //! Get the step size.
+  double StepSize() const { return optimizer.StepSize(); }
+  //! Modify the step size.
+  double& StepSize() { return optimizer.StepSize(); }
+
+  //! Get the value used to initialise the mean squared gradient parameter.
+  double Epsilon() const { return optimizer.Epsilon(); }
```
zoq: Since optimizer is SGD<...> there is no ``optimizer.Epsilon()`` but the policy has the ``Epsilon()`` function. So it should be ``updatePolicy.Epsilon()`` under the assumption there is an updatePolicy object.

src/mlpack/core/optimizers/smorms3/smorms3_impl.hpp
```diff
@@ -0,0 +1,39 @@
...
+namespace optimization {
+
+template<typename DecomposableFunctionType>
+SMORMS3<DecomposableFunctionType>::SMORMS3(DecomposableFunctionType& function,
+                                           const double stepSize,
+                                           const double epsilon,
+                                           const size_t maxIterations,
+                                           const double tolerance,
+                                           const bool shuffle) :
+    function(function),
```
ivmarkp: Can you elaborate a bit more? I've tried all I know but this is turning out to be a confusing aspect of meta-programming for me. I have the initialisation as follows:
```cpp
SMORMS3<DecomposableFunctionType>::SMORMS3(DecomposableFunctionType& function,
                                           const double stepSize,
                                           const double epsilon,
                                           const size_t maxIterations,
                                           const double tolerance,
                                           const bool shuffle) :
    optimizer(SGD<DecomposableFunctionType, SMORMS3Update>(function,
                                                           stepSize,
                                                           maxIterations,
                                                           tolerance,
                                                           shuffle,
                                                           SMORMS3Update))
{ /* Nothing to do. */ }
```
but it results in a couple of [errors](https://paste.debian.net/922572/) and a warning for unused `epsilon`. Can you please take a look at those errors?

src/mlpack/core/optimizers/smorms3/smorms3.hpp
ivmarkp: Can we just do `return epsilon` here? Because `SMORMS3Update` (Update policy class) has no function named `Epsilon` and thus calling a function that doesn't exist won't work.

src/mlpack/core/optimizers/smorms3/smorms3_impl.hpp
ivmarkp: Yes, these initialisations will go after we have the updated initialisation of `optimizer` object in place.

src/mlpack/core/optimizers/smorms3/smorms3_impl.hpp
ivmarkp: @zoq Please ignore my last comments. I've fixed those errors and the instantiation looks pretty clean now. :) Pushing the changes shortly.

src/mlpack/core/optimizers/smorms3/smorms3.hpp
zoq: What we like to do here is to modify the epsilon value of the policy class right? So, I think we should add the Function to the Policy class and use ``double Epsilon() const { return smorms3Update.Epsilon(); }`` here. What do you think?

ivmarkp: I have a query regarding this (just so that I fully understand) -- we need `Epsilon()` function in the policy class to facilitate the direct modification of epsilon after the initial value was assigned to it through a SMORMS3 object like `SMORMS3<SGDTestFunction> s(f, 0.001, 1e-16, ...);`. Is that correct?

ivmarkp:
@zoq  Could you take another look? There are some changes that I made after your last review. Thanks.

src/mlpack/core/optimizers/smorms3/smorms3_update.hpp
```diff
@@ -0,0 +1,118 @@
...
+    iterate -= gradient % arma::min(x, stepSizeMat) / (arma::sqrt(g2) + epsilon);
+
+    mem %= (1 - x);
+    mem += 1;
+
+    previousStepSize = stepSize;
+  }
+ private:
+   //! The value used to initialise the mean squared gradient parameter.
+   double epsilon, previousStepSize = 0;
```
zoq: Can you use the constructor initialization list for the initialization of previousStepSize?

src/mlpack/core/optimizers/smorms3/smorms3.hpp
```diff
@@ -0,0 +1,162 @@
...
+  bool Shuffle() const { return optimizer.Shuffle(); }
+  //! Modify whether or not the individual functions are shuffled.
+  bool& Shuffle() { return optimizer.Shuffle(); }
+
+ private:
+  //! The instantiated function.
+  DecomposableFunctionType& function;
+
+  //! The step size for each example.
+  double stepSize;
```
zoq: The StepSize, maxIterations, tolerance, shuffle as modified via the optimizer object, so no need to save them here.

ivmarkp:
@zoq Hi, I addressed the issue you mentioned earlier. Do you think this PR is ready to be merged now?

src/mlpack/core/optimizers/smorms3/smorms3.hpp
```diff
@@ -0,0 +1,149 @@
...
+  //! Modify the instantiated function.
+  DecomposableFunctionType& Function() { return function; }
+
+  //! Get the step size.
+  double StepSize() const { return optimizer.StepSize(); }
+  //! Modify the step size.
+  double& StepSize() { return optimizer.StepSize(); }
+
+  //! Get the value used to initialise the mean squared gradient parameter.
+  double Epsilon() const { return smorms3Update.Epsilon(); }
```
zoq: Please use ``optimizer. UpdatePolicy. Epsilon()``here, that way it has an effect on the policy hold in the SGD instantiation.

src/mlpack/core/optimizers/smorms3/smorms3.hpp
```diff
@@ -0,0 +1,149 @@
...
+   * be modified to store the finishing point of the algorithm, and the final
+   * objective value is returned.
+   *
+   * @param iterate Starting point (will be modified).
+   * @return Objective value of the final point.
+   */
+  double Optimize(arma::mat& iterate) { return optimizer.Optimize(iterate); }
+
+  //! Get the instantiated function to be optimized.
+  const DecomposableFunctionType& Function() const { return function; }
```
zoq: Please use ``optimizer.Function()`` here.

src/mlpack/core/optimizers/smorms3/smorms3.hpp
```diff
@@ -0,0 +1,149 @@
...
+  bool Shuffle() const { return optimizer.Shuffle(); }
+  //! Modify whether or not the individual functions are shuffled.
+  bool& Shuffle() { return optimizer.Shuffle(); }
+
+ private:
+  //! The instantiated function.
+  DecomposableFunctionType& function;
+
+  //! The value used to initialise the mean squared gradient parameter.
+  double epsilon;
```
zoq: I think we can remove epsilon here, since it isn't used.

src/mlpack/core/optimizers/smorms3/smorms3.hpp
```diff
@@ -0,0 +1,149 @@
...
+  double& Tolerance() { return optimizer.Tolerance(); }
+
+  //! Get whether or not the individual functions are shuffled.
+  bool Shuffle() const { return optimizer.Shuffle(); }
+  //! Modify whether or not the individual functions are shuffled.
+  bool& Shuffle() { return optimizer.Shuffle(); }
+
+ private:
+  //! The instantiated function.
+  DecomposableFunctionType& function;
```
zoq: If we use ``optimizer.Function`` we can remove the parameter, here what do you think?

src/mlpack/core/optimizers/smorms3/smorms3.hpp
```diff
@@ -0,0 +1,149 @@
...
+
+  //! The value used to initialise the mean squared gradient parameter.
+  double epsilon;
+
+  //! The SMORMS3Update update policy object.
+  SMORMS3Update smorms3Update;
```
zoq: If we use ``Optimizer.UpdatePolicy()``in all places, we can remove the parameter.

src/mlpack/core/optimizers/smorms3/smorms3_update.hpp
```diff
@@ -0,0 +1,120 @@
...
+    {
+      stepSizeMat.fill(stepSize);
+    }
+
+    iterate -= gradient % arma::min(x, stepSizeMat) / (arma::sqrt(g2) + epsilon);
+
+    mem %= (1 - x);
+    mem += 1;
+
+    previousStepSize = stepSize;
```
zoq: Should we put this into the if case?

src/mlpack/core/optimizers/smorms3/smorms3_update.hpp
```diff
@@ -0,0 +1,120 @@
...
+  }
+ private:
+   //! The value used to initialise the mean squared gradient parameter.
+   double epsilon, previousStepSize;
```
zoq: Can you use one line for each parameter, that way we can add a comment that describes the previousStepSize.

src/mlpack/core/optimizers/smorms3/smorms3.hpp
```diff
@@ -0,0 +1,149 @@
...
+ *
+ * For more information, see the following.
+ *
+ * @code
+ * @misc{Funk2015,
+ *   author    = {Simon Funk},
+ *   title     = {RMSprop loses to SMORMS3 - Beware the Epsilon!},
+ *   year      = {2015}
+ * }
+ * @endcode
```
rcurtin: If you want to use the BiBTeX citation here we should probably add the URL: http://sifter.org/~simon/journal/20150420.html

Alternately you could just cite the URL and we can hope Simon never takes his website down. :)

src/mlpack/core/optimizers/smorms3/smorms3_update.hpp
```diff
@@ -0,0 +1,120 @@
...
+ * of RMSprop and Yann LeCun’s method in "No more pesky learning rates".
+ *
+ * For more information, see the following.
+ *
+ * @code
+ * @misc{Funk2015,
+ *   author    = {Simon Funk},
+ *   title     = {RMSprop loses to SMORMS3 - Beware the Epsilon!},
+ *   year      = {2015}
+ * }
```
rcurtin: Same BiBTeX comment applies here.

src/mlpack/core/optimizers/smorms3/smorms3_update.hpp
```diff
@@ -0,0 +1,120 @@
...
+{
+ public:
+  /**
+   * Construct the SMORMS3 update policy with given epsilon parameter.
+   *
+   * @param epsilon Value used to initialise the mean squared gradient parameter.
+   */
+  SMORMS3Update(const double epsilon = 1e-16) :
+              epsilon(epsilon),
+              previousStepSize(0)
```
rcurtin: The spacing here seems weird, the initialization lists should only be 2 indents in if they are on newlines.

src/mlpack/core/optimizers/smorms3/smorms3_update.hpp
```diff
@@ -0,0 +1,120 @@
...
+
+    g2 = (1 - r) % g2;
+    g2 += r % (gradient % gradient);
+
+    arma::mat x = (g % g) / (g2 + epsilon);
+
+    if (stepSize != previousStepSize)
+    {
+      stepSizeMat.fill(stepSize);
+    }
```
rcurtin: Allocating an entire matrix here to hold the same value, I think, is going to be inefficient.   We should find a way to avoid this, which probably means adapting the `arma::min(x, stepSizeMat)` call.  (Though I wouldn't be surprised if `arma::min(arma::mat&, double)` works fine; you should try and see.)

src/mlpack/core/optimizers/smorms3/smorms3_update.hpp
rcurtin: If we can avoid the need for `stepSizeMat` then we can remove the `previousStepSize` member also.

ivmarkp:
Thanks again @zoq and @rcurtin for your reviews. I'll update the PR shortly.

src/mlpack/core/optimizers/smorms3/smorms3_update.hpp
rcurtin: Right---I think maybe we can use `for_each()` and a lambda here to solve the problem.  What do you think?

ivmarkp: Sorry, I'm not entirely sure what for we need `for_each()` and a lambda here? Could you please elaborate a bit? Thanks.

rcurtin: The idea is that we can avoid the creation of `stepSizeMat` entirely, which is only necessary because `arma::min()` requires two matrix-like objects.  We can use a lambda:

```
x.for_each([](double& v) { v = std::min(v, stepSize); });
```

Then the `arma::min()` isn't necessary, and we don't need `stepSizeMat` at all.  Let me know if I can clarify anything.  I didn't test the code above so there could be small errors, but it at least gets across the general idea.

ivmarkp: Thanks, now I get it. That's a nice solution. I'll make the changes and push in a few minutes.

ivmarkp:
@rcurtin Travis build is failing with `error: ‘arma::mat’ has no member named ‘for_each’`. It's strange because there are no issues locally.

rcurtin:
Try using `transform()`, not `for_each()`; looks like only recent Armadillo versions have `for_each()`.

ivmarkp:
Ah, okay. If possible, could we also try installing the latest Armadillo version for Travis builds?

rcurtin:
Nope, because then we would not catch issues with old Armadillo versions. :)

ivmarkp:
Oh, I see. Hopefully, using `transform()` should do the trick. :)

src/mlpack/core/optimizers/smorms3/smorms3_update.hpp
```diff
@@ -0,0 +1,111 @@
...
+  //! Modify the value used to initialise the mean squared gradient parameter.
+  double& Epsilon() { return epsilon; }
+
+  /**
+   * The Initialize method is called by SGD::Optimize method with UpdatePolicy
+   * SMORMS3Update before the start of the iteration update process.
+   *
+   * @param rows number of rows in the gradient matrix.
+   * @param cols number of columns in the gradient matrix.
+   * @param lRate
```
zoq: Looks like we can delete the parameter comment here.

src/mlpack/core/optimizers/smorms3/smorms3_update.hpp
```diff
@@ -0,0 +1,111 @@
...
+ * @author Vivek Pal
+ *
+ * SMORMS3 update for Stochastic Gradient Descent.
+ *
+ * mlpack is free software; you may redistribute it and/or modify it under the
+ * terms of the 3-clause BSD license.  You should have received a copy of the
+ * 3-clause BSD license along with mlpack.  If not, see
+ * http://www.opensource.org/licenses/BSD-3-Clause for more information.
+ */
+#ifndef MLPACK_CORE_OPTIMIZERS_SMORMS3_UPDATE_HPP
```
zoq: Can you use ``#ifndef MLPACK_CORE_OPTIMIZERS_SMORMS3_SMORMS3_UPDATE_HPP`` here, to match the rest of the codebase?

src/mlpack/core/optimizers/smorms3/smorms3_update.hpp
ivmarkp: Yeah, I missed that. Will remove that comment.

src/mlpack/core/optimizers/smorms3/smorms3_update.hpp
ivmarkp: Sure, will make the change.

rcurtin:
Thanks for the hard work. :)

ivmarkp:
That's great, thanks for reviewing the PR!

PR #901, opened 2017-03-01
Suppressing Warning from unit test

govg:
The file load_save_test.cpp generate warnings that were not suppressed when using mlpack_test to run unit tests. A guard has been added for all test invocations, that sets Log::Warn.ignoreInput to true temporarily while tests are run, and resets to its earlier value. 

The earlier setting set Log::Warn.ignoreInput to false at the end of a test, this caused subsequent files to pick up this flag and output warnings.

### DIFF
--- a/src/mlpack/tests/load_save_test.cpp
+++ b/src/mlpack/tests/load_save_test.cpp
@@ -28,10 +28,7 @@ BOOST_AUTO_TEST_SUITE(LoadSaveTest);
 BOOST_AUTO_TEST_CASE(NoExtensionLoad)
 {
   arma::mat out;
-  bool lflag = Log::Warn.ignoreInput;
-  Log::Warn.ignoreInput = true;
   BOOST_REQUIRE(data::Load("noextension", out) == false);
-  Log::Warn.ignoreInput = lflag;
 }
 
 /**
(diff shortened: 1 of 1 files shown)


rcurtin:
Thanks for the contribution.  I see you wrote:

> A guard has been added for all test invocations, that sets Log::Warn.ignoreInput to true temporarily while tests are run, and resets to its earlier value.

But I don't see this part, maybe there is a piece that didn't get checked in?  Or maybe I have overlooked it in the code?

Thanks!

govg:
What I meant was this bit :   
```cpp

   bool lflag = Log::Warn.ignoreInput;
   Log::Warn.ignoreInput = true;
   BOOST_REQUIRE(data::Load("noextension", out) == false);
-  Log::Warn.ignoreInput = false;
+  Log::Warn.ignoreInput = lflag;

```
The earlier code(-) reset it to false, I've simply set it to the previous value of Log::Warn.ignoreInput

Is there a cleaner way to do this? I figured setting it to false actually causes issues because a test later on might get triggered and send output. Reverting to the earlier value seems to suppress all warnings.

rcurtin:
Hmm... the default value of `Log::Warn.ignoreInput` should be false.  So the code that reset it to false should simply be restoring it to its original value, just like your modification.  I was confused, then, why what you are doing is false... then I found `src/mlpack/tests/mlpack_test.cpp`, which I guess I had forgotten about.  This file defines a test fixture that sets `Log::Warn.ignoreInput = true`, meaning that all of the `Log::Warn.ignoreInput` modification is actually unnecessary, and simply the issue is that the code I wrote that resets it to `false` introduces a bug!

So I think actually you could modify the PR to simply remove all of the handling of `Log::Warn.ignoreInput` altogether (except what's in `mlpack_test.cpp`), and it will solve the problem.

govg:
Got it, will update PR by tomorrow.

govg:
Updated, I guess this should work?

rcurtin:
Perfect.  Thanks!  If you want, you can add some lines (instead of removing them like in this PR!) by adding your name to `src/mlpack/core.hpp` and `COPYRIGHT.txt`.

Issue #902, opened 2017-03-01
Upper Camel Case


zoq:
As I implemented the first activation function in ``src/mlpack/methods/ann/activation_functions`` I used Lower Camel Case for the function names instead of Upper Camel Case as used throughout the codebase and as stated in the design guidelines: https://github.com/mlpack/mlpack/wiki/DesignGuidelines

The goal of this ticket is to find and to replace them with the correct style:

```
fn -> Fn
deriv -> Deriv
inv -> Inv
```

sam0410:
Hi @zoq !
I'll pull a request soon !

sam0410:
Hi @zoq !
I created a pull request for this.

sam0410:
Hi @zoq !
The PR is now ready to merge. Please review the same
Thanks!

zoq:
Fixed in 5215c3df3ec.

PR #906, opened 2017-03-03
fix for #902

sinjan25:
fix for #902


src/mlpack/methods/policy_gradient/policy_gradient.cpp
```diff
@@ -0,0 +1 @@
+
```
zoq: Can you remove this file?

sinjan25: Okay. I was actually planning to add my implementation of policy gradient there.

sinjan25: @zoq I updated elu.hpp and removed the policy_gradient.cpp

sam0410:
Hi @zoq !
Actually I was thinking, what is wrong in the code in my pull request? Because, it has been 3 days, u haven't reviewed it yet.

sinjan25:
@zoq All checks have successfully passed.

sam0410:
@sinjan25 If another person has already pulled the request which passes all the test cases, it is pointless to work on the same issue. You could take up another issue. This way more effective work can be done.

zoq:
@sinjan25 I checked the issue and @sinjan25 was the first how said he likes to solve the issue and he was also the first who created an PR for that, so I think it's fair to merge his PR.

We are working on adding other issues to the issue tracker.

zoq:
Issue #902 is fixed in 5215c3d, however thanks for taking a look into the issue and for the contribution.

PR #907, opened 2017-03-03
Made ARFF Loader Case insensitive.

chvsp:
Converted the tags (@relation, @data, @attribute) into lowercase before comparing. Also created a test to check the successful load of an ARFF file which has these tags in upper case.


chvsp:
@rcurtin Kindly have a look at this PR. It is a minor change, hence shouldn't take much time to review. Thanks :)

src/mlpack/tests/load_arff_test.cpp
```diff
@@ -0,0 +1,39 @@
...
+  arma::mat dataset;
+
+  DatasetMapper<IncrementPolicy> info;
+
+  LoadARFF<double, IncrementPolicy>("casecheck.arff", dataset, info);
+
+  BOOST_CHECK_EQUAL(dataset.n_rows, 2);
+  BOOST_CHECK_EQUAL(dataset.n_cols, 3);
+
+}
```
rcurtin: I think we can just drop this in `load_save_test.cpp`---or split the other ARFF tests into this file.  It's your call which you'd like to do.

rcurtin:
Sweet, thank you for the contribution.  I made some trivial style fixes in d032540e.

PR #913, opened 2017-03-04
Issue#593

ShihaoJing:
For issue#593.
Added a command line parameter to output posterior of each class to a file.


src/mlpack/methods/naive_bayes/naive_bayes_classifier.hpp
```diff
@@ -123,7 +123,7 @@ class NaiveBayesClassifier
    * @param data List of data points.
    * @param results Vector that class predictions will be placed into.
    */
-  void Classify(const MatType& data, arma::Row<size_t>& results);
+  void Classify(const MatType& data, arma::Row<size_t>& results, MatType& results_proba);
```
rcurtin: This will have to be of type `arma::mat` not `MatType`---if the user sets `MatType` to, i.e. `arma::umat`, we can't return posteriors as a `umat`, we need double-precision.

src/mlpack/methods/naive_bayes/naive_bayes_classifier_impl.hpp
```diff
@@ -162,8 +162,7 @@ void NaiveBayesClassifier<MatType>::Train(const VecType& point,
 }
 
 template<typename MatType>
-void NaiveBayesClassifier<MatType>::Classify(const MatType& data,
-                                             arma::Row<size_t>& results)
+void NaiveBayesClassifier<MatType>::Classify(const MatType& data, arma::Row<size_t>& results, MatType& results_proba)
```
rcurtin: This line is longer than 80 characters, and you shouldn't use underscores in variable names.  (These comments apply throughout the PR.)  Take a look at the style guidelines and please make sure your code conforms: https://github.com/mlpack/mlpack/wiki/DesignGuidelines#style-guidelines

src/mlpack/tests/nbc_test.cpp
```diff
@@ -60,16 +61,27 @@ BOOST_AUTO_TEST_CASE(NaiveBayesClassifierTest)
 
   arma::mat testData;
   arma::Mat<size_t> testRes;
+  arma::mat testResProba;
   arma::Row<size_t> calcVec;
+  arma::mat calcProba;
   data::Load(testFilename, testData, true);
   data::Load(testResultFilename, testRes, true);
+  data::Load(testResultProbaFilename, testResProba, true);
```
rcurtin: How were the results in this file calculated?

src/mlpack/methods/naive_bayes/nbc_main.cpp
```diff
@@ -77,6 +77,8 @@ PARAM_FLAG("incremental_variance", "The variance of each class will be "
 PARAM_MATRIX_IN("test", "A matrix containing the test set.", "T");
 PARAM_UMATRIX_OUT("output", "The matrix in which the predicted labels for the"
     " test set will be written.", "o");
+PARAM_MATRIX_OUT("output_proba", "The matrix in which the predicted probability of labels for the"
```
rcurtin: Go with `output_probs` not `output_proba` please, for consistency across other programs.

src/mlpack/methods/naive_bayes/naive_bayes_classifier_impl.hpp
```diff
@@ -162,8 +162,7 @@ void NaiveBayesClassifier<MatType>::Train(const VecType& point,
 }
 
 template<typename MatType>
-void NaiveBayesClassifier<MatType>::Classify(const MatType& data,
-                                             arma::Row<size_t>& results)
```
rcurtin: The user may not always want the probabilities, so we should leave the other overload in place and add a new method, instead of simply changing the signature of the existing method (which also breaks reverse compatibility).

src/mlpack/methods/naive_bayes/naive_bayes_classifier_impl.hpp
```diff
@@ -195,18 +194,25 @@ void NaiveBayesClassifier<MatType>::Classify(const MatType& data,
...
-    testProbs.col(i) += (data.n_rows / -2.0 * log(2 * M_PI) - 0.5 *
+    testProbs.col(i) += (data.n_rows / - 2.0 * log(2 * M_PI) - 0.5 *
         log(arma::det(arma::diagmat(variances.col(i)))) + exponents);
   }
 
+  arma::vec log_prob_x(data.n_cols); // log(Prob(X))
+  for (size_t j = 0; j < data.n_cols; ++j)
+    log_prob_x(j) = log(arma::accu(exp(testProbs.row(j))));
+
+  testProbs -= arma::repmat(log_prob_x, 1, testProbs.n_cols);
```
rcurtin: I am not sure I understand what this calculation is.  Can you please clarify?

src/mlpack/methods/naive_bayes/naive_bayes_classifier_impl.hpp
ShihaoJing: Can I refractor this according to the design of decision_tree? Like these four methods:
`template<typename VecType>
  size_t Classify(const VecType& point) const;

  template<typename VecType>
  void Classify(const VecType& point,
                size_t& prediction,
                arma::vec& probabilities) const;

  template<typename MatType>
  void Classify(const MatType& data,
                arma::Row<size_t>& predictions) const;

  template<typename MatType>
  void Classify(const MatType& data,
                arma::Row<size_t>& predictions,
                arma::mat& probabilities) const;`

src/mlpack/methods/naive_bayes/naive_bayes_classifier_impl.hpp
ShihaoJing: This calculates total probability of features. Since we use log probability, we need to subtract it. I will add some comments to clarify this.

src/mlpack/tests/nbc_test.cpp
ShihaoJing: I used GaussianNB module in scikit-learn to calculate posteriors. I don't know if it's the appropriate way to do such testing.

ShihaoJing:
@rcurtin Thanks for code review. I will fix these problems as soon as possible.

src/mlpack/methods/naive_bayes/naive_bayes_classifier_impl.hpp
rcurtin: Sure, I think that would be a nice improvement.

ShihaoJing:
@rcurtin Travis build failed and it seems due to the failure of gmm test.

rcurtin:
Hi Shihao,

I'm sorry for the delay on my review here.  Thanks for the contribution!  I went ahead and changed a couple of things in 6751acf.  Mostly it's style changes but important things to note are that I changed the return type of `JointLogLikelihood()` to `void` in order to avoid temporary objects.  Instead it's better to pass in the output matrix or vector.  In addition, Armadillo is column-major, so it is best to organize data in columns, because rows are contiguous in memory; it's typically much faster to do operations that call `col()` instead of `row()`.

Also, now I think I understand what you were saying about the probability calculation many weeks ago.  It is definitely true that when you are not calculating the posterior probabilities, there's no need to normalize (subtract log(P(X))) since we only need the max value.  I agree that the calculations you added here do properly normalize.

Thanks!

PR #917, opened 2017-03-06
confidence scores 

samO410:
Hi @rcurtin !
As discussed, I reopened a pull request for issue #593 
Thanks!


rcurtin:
This approach doesn't give the user a clean way to get the confidence scores and increases the size of the `NaiveBayesClassifier` class unnecessarily.  Instead the right way to do this would be to create a new overload of `Classify()`.

I think that Shihao's approach to this PR is in much more of the right direction, so I'd like to go ahead and close this one. However, I have not been able to fully review the math in that one, so if you want, you can review the relevant math there (#913) and help the PR converge to the correct implementation.

sam0410:
Thanks @rcurtin ! I'll do that.

rcurtin:
Thanks, I will go ahead and close this then.

PR #923, opened 2017-03-06
NAG

kris-singh:
This adds Nesterov Accelerated Gradient PR #893.
All tests have passed locally.
I think this is needed as separate optimizer and
not as extension to SGD as has been discussed 
for SGD+Momentum.
Ping @zoq @rcurtin


kris-singh:
All test pass locally. I don't understand why they are failing travis CI?

kris-singh:
Only the gmm test is failling. This is ready for your comments. Also following this i have an idea to allow any method to incorporate momentum i would discuss that on the irc.

src/mlpack/core/optimizers/nestrov/nag_impl.hpp
```diff
@@ -0,0 +1,127 @@
...
+    if (shuffle)
+      function.Gradient(iterate + momentum * velocity, visitationOrder[currentFunction], gradient);
+    else
+      function.Gradient(iterate + momentum * velocity, currentFunction, gradient);
+
+    //iterate += momentum * velocity;
+    velocity = velocity*momentum - stepSize*gradient;
+
+    // And update the iterate.
+    iterate = iterate + velocity;
```
zoq: I think you missed some steps here, I think it should be:

```
iterate += momentum * momentum * velocity;
iterate -= (1 + momentum) * stepSize * gradient;
```

and

```
function.Gradient(iterate, visitationOrder[currentFunction], gradient);
```

kris-singh: No, I made the update in function.gradient function(iterate + momentum * velocity). because we need to take the gradient at updated parameters. with this gradient. Then, update the velocity vector.
You can look here http://sebastianruder.com/optimizing-gradient-descent/index.html#nesterovacceleratedgradient

kris-singh: Thanks for the review. I will update this with the policy design pattern.

zoq: I see, the reference I used was: https://arxiv.org/pdf/1212.0901.pdf which uses a different formulation, that avoids the peeking ahead computation of the first run, which is zero since when momentum is initialized with zero.

kris-singh: @zoq I am thinking we could have  2 base classes 1. sgd(arun) 2.nestrov. I want to implement nestrov as a base class where we all the accelerated variants of optimisers can be implemented with nestrov being the base the class and all other being policies. This is similar to how caffe does things(though) they use inheritance. Also why do we prefer policy based design pattern over inheritance(they are pretty similar).

zoq: I'm not sure I get what you mean here, if we implement it as an policy any other optimizer can use the method.

Take a look at https://github.com/mlpack/mlpack/wiki/DesignGuidelines#inheritance to get the answer why we prefer the policy design over inheritance here.

kris-singh: I was thinking about how we could implement this as Policy class for sgd. We need to temporarily update the iterate before the gradient step also. Now we could have a policy that implement TempUpdate function. But the problem is we can't sfINAE here because the type is deduced as soon as the object is instantiated. Basically look [here](https://gist.github.com/kris-singh/9ed1d88c781de14fd62264ce6b1dd1b8) . So what do you think how should i go about it.

kris-singh: A simple solution would be to add to add the nestrov_momentum=False parameter to every policies constructor. And check in sgd_impl file if nestrov_momentum=True then we could to the temp update. but the problem is we would have to do that at every iteration which defeats the whole purpose

zoq: What do you think about the reformulation in: https://arxiv.org/pdf/1212.0901.pdf I pointed out earlier, that way we don't have to modify the iterate parameter twice.

kris-singh:
Ready for review. Could you have a look @zoq  and @rcurtin. Thanks.

kris-singh:
Hi, @zoq  could review this. Thanks

kris-singh:
@zoq did you have time to have a look at this. Are there any changes needed to be done

src/mlpack/core/optimizers/nestrov/nestrov_update.hpp
```diff
@@ -0,0 +1,107 @@
...
+  void Update(arma::mat& iterate,
+              const double stepSize,
+              const arma::mat& gradient,
+              T... args)
+  {
+    velocity = momentum * velocity - stepSize * gradient;
+    iterate += momentumLookAhead(momentum)*momentum*velocity - (1+momentumLookAhead(momentum))*stepSize*gradient;
+  }
+
+  double momentumLookAhead(double momentum)
```
zoq: I'm not sure this function necessary. The momentum shouldn't change, what could change is the stepSize, which is passed in each iteration for the case, someone uses a decay policy.

src/mlpack/core/optimizers/nestrov/nestrov_update.hpp
```diff
@@ -0,0 +1,107 @@
...
+  NestrovUpdate(const double momentum = 0.5) : momentum(momentum)
+  { /* Do nothing. */ };
+
+  /**
+   * The Initialize method is called by SGD Optimizer method before the start of
+   * the iteration update process.  In the momentum update policy the velocity
+   * matrix is initialized to the zeros matrix with the same size as the
+   * gradient matrix (see mlpack::optimization::SGD::Optimizer )
+   *
+   * @param n_rows number of rows in the gradient matrix.
```
zoq: Please use ``rows`` and ``cols`` instead of ``n_rows`` and ``n_cols`` here to match the rest of the codebase.

src/mlpack/core/optimizers/nestrov/nestrov_update.hpp
```diff
@@ -0,0 +1,107 @@
...
+   * @param gradient The gradient matrix.
+   */
+  template<typename... T>
+  void Update(arma::mat& iterate,
+              const double stepSize,
+              const arma::mat& gradient,
+              T... args)
+  {
+    velocity = momentum * velocity - stepSize * gradient;
+    iterate += momentumLookAhead(momentum)*momentum*velocity - (1+momentumLookAhead(momentum))*stepSize*gradient;
```
zoq: Minor style issue can you put a space between the operations.

src/mlpack/core/optimizers/nestrov/nestrov_update.hpp
```diff
@@ -0,0 +1,107 @@
...
+   *
+   * @param iterate Parameters that minimize the function.
+   * @param stepSize Step size to be used for the given iteration.
+   * @param gradient The gradient matrix.
+   */
+  template<typename... T>
+  void Update(arma::mat& iterate,
+              const double stepSize,
+              const arma::mat& gradient,
+              T... args)
```
zoq: Not sure why we have to pass args here. To pass a decay policy?

kris-singh: Yes, actually you could pass decay policy or other parameters like function or something else. This would not break any existing methods and would also be useful if somebody requires some other parameters in future. What do you think ?

src/mlpack/core/optimizers/nestrov/nestrov_update.hpp
kris-singh: If you look at the link the momentum(t) and momentum(t-1). I have used this function for that. If somebody would lik to apply decay policy on momentum they could change this function. Thinking about it this function should rather take a function as a parameters that could implement the decay policy.

kris-singh:
@zoq ready for review i have made the changes you asked. Let me know if any thing else is required. For the point regarding variadic argument please see the comments.

src/mlpack/core/optimizers/nestrov/nestrov_update.hpp
zoq: I think, that it's not the right way to do it here and probably not necessary:

- Update is called from within the optimizer class e.g. SGD so if someone likes to pass something else, he has to modify the class anyway and not only that class, all other policy classes.
- You can use the constructor do pass something else that's required for the update process.

src/mlpack/core/optimizers/nestrov/nestrov_update.hpp
```diff
@@ -0,0 +1,108 @@
...
+  {
+    velocity = momentum * velocity - stepSize * gradient;
+    iterate += momentumLookAhead(momentum) * momentum * velocity 
+      - (1 + momentumLookAhead(momentum)) * stepSize * gradient;
+  }
+
+  double momentumLookAhead(double momentum)
```
zoq: I get your point but, if someone likes to use momentum decay, he has to modify the codebase. You can't just link against mlpack, implement your own policy and use it for the NestrovUpdate, which is possible for the rest of the codebase. E.g. you can implement your own update policy and use it with the existing SGD class, without any modification. What you could do is to provide another policy for the NestrovUpdate policy that enables someone to use it with decay method for the momentum term, but in this case, we should implement a method that accelerates the NestrovUpdate besides the standard static momentum.

kris-singh:
Changes Made ready for merge. I have added the DecayPolicy as a templated function rather than as templated class 1)because it works without breaking anything we don't need to change much 2) It makes more of logical sense to me. Let me know what you think ?

src/mlpack/core/optimizers/nestrov/nag.hpp
```diff
@@ -0,0 +1,155 @@
...
+ * Stochastic Gradient Descent(SGD) with Nestrov Accelrated Gradient
+ * update(NAG) is an approach 
+ * that enjoys accelerated convergence rates compared to vanilla SGD.
+ *
+ * mlpack is free software; you may redistribute it and/or modify it under the
+ * terms of the 3-clause BSD license.  You should have received a copy of the
+ * 3-clause BSD license along with mlpack.  If not, see
+ * http://www.opensource.org/licenses/BSD-3-Clause for more information.
+ */
+#ifndef MLPACK_CORE_OPTIMIZERS_MOMENTUM_SGD_MOMENTUM_SGD_HPP
```
zoq: The header guard should be named like:  ``MLPACK_CORE_OPTIMIZERS_NESTROV_NAG_HPP``.

src/mlpack/core/optimizers/nestrov/nag.hpp
```diff
@@ -0,0 +1,155 @@
...
+   * @param function Function to be optimized (minimized).
+   * @param stepSize Step size for each iteration.
+   * @param maxIterations Maximum number of iterations allowed (0 means no
+   *     limit).
+   * @param tolerance Maximum absolute tolerance to terminate algorithm.
+   * @param momentum The momentum hyperparameter
+   * @param shuffle If true, the function order is shuffled; otherwise, each
+   *     function is visited in linear order.
+   */
+  NAG(DecomposableFunctionType& function,
```
zoq: Can you match the order of parameters with the one in the description?

src/mlpack/core/optimizers/nestrov/nag.hpp
```diff
@@ -0,0 +1,155 @@
...
+
+  /**
+   * Optimize the given function using NAG.  The given
+   * starting point will be modified to store the finishing point of the
+   * algorithm, and the final objective value is returned.
+   *
+   * @param iterate Starting point (will be modified).
+   * @return Objective value of the final point.
+   */
+  double Optimize(arma::mat& iterate){return optimizer.Optimize(iterate);};
```
zoq: Minor format issue, a space between ``{`` and ``return`` the same at the end.

src/mlpack/core/optimizers/nestrov/nag.hpp
```diff
@@ -0,0 +1,155 @@
...
+
+  //! Get the tolerance for termination.
+  double Tolerance() const { return optimizer.Tolerance(); }
+  //! Modify the tolerance for termination.
+  double& Tolerance() { return optimizer.Tolerance(); }
+
+  //! Get whether or not the individual functions are shuffled.
+  bool Shuffle() const { return optimizer.Shuffle(); }
+  //! Modify whether or not the individual functions are shuffled.
+  bool& Shuffle() { return optimizer.Shuffle(); }
```
zoq: Should we add another function for the momentum parameter, what do you think?

src/mlpack/core/optimizers/nestrov/nag_impl.hpp
```diff
@@ -0,0 +1,41 @@
...
+ * @author Kris Singh
+ *
+ * Implementation of stochastic gradient descent.
+ *
+ * mlpack is free software; you may redistribute it and/or modify it under the
+ * terms of the 3-clause BSD license.  You should have received a copy of the
+ * 3-clause BSD license along with mlpack.  If not, see
+ * http://www.opensource.org/licenses/BSD-3-Clause for more information.
+ */
+#ifndef MLPACK_CORE_OPTIMIZERS_NAG_IMPL_HPP
```
zoq: The header guard should be named like: MLPACK_CORE_OPTIMIZERS_NESTROV_NAG_IMPL_HPP.

src/mlpack/core/optimizers/nestrov/nag_impl.hpp
```diff
@@ -0,0 +1,41 @@
...
+#define MLPACK_CORE_OPTIMIZERS_NAG__IMPL_HPP
+
+#include <mlpack/methods/regularized_svd/regularized_svd_function.hpp>
```
zoq: Maybe I missed something, but I think we can remove the header here.

src/mlpack/core/optimizers/nestrov/nag_impl.hpp
```diff
@@ -0,0 +1,41 @@
...
+
+// In case it hasn't been included yet.
+#include "nag.hpp"
+#include "nestrov_update.hpp"
```
zoq: ``nag.hpp`` already includes ``nestrov_update.hpp``so no need to include it here again.

src/mlpack/core/optimizers/nestrov/nag_impl.hpp
```diff
@@ -0,0 +1,41 @@
...
+namespace optimization {
+
+template<typename DecomposableFunctionType>
+NAG<DecomposableFunctionType>::NAG(DecomposableFunctionType& function,
+                                   const double momentum,
+                                   const double stepSize,
+                                   const size_t maxIterations,
+                                   const double tolerance,
+                                   const bool shuffle):
+nestrovupdate(NestrovUpdate(momentum)),
```
zoq: I think there is no need to hold the update policy, since it is already hold inside of the SGD instance and we can access it with ``optimizer.PolicyUpdate()``.

src/mlpack/core/optimizers/nestrov/nestrov_update.hpp
```diff
@@ -0,0 +1,112 @@
...
+  {
+    double momentumLookAhead;
+    velocity = momentum * velocity - stepSize * gradient;
+    UpdateMomentum(momentum, momentumLookAhead);
+    iterate += momentumLookAhead * momentum * velocity 
+      - (1 + momentumLookAhead) * stepSize * gradient;
+  }
+
+  template<typename DecayPolicy = VanillaMomentum>
+  void UpdateMomentum(const double& momentum, double& momentumLookAhead)
```
zoq: I think the reason for this function is to use it with another policy, but how can I call that function with another policy? Also, right now we create a new decay instance for each run, which could result in an unnecessary slowdown.

src/mlpack/core/optimizers/nestrov/nestrov_update.hpp
zoq: I think this comment this applies.

src/mlpack/core/optimizers/nestrov/nestrov_update.hpp
kris-singh: At present if you look at the paper https://arxiv.org/pdf/1212.0901.pdf. This uses a momentum decay term which other policies such adam and other dont use hence i have only included the templated function for nestrov_update.hpp.

zoq: I get the point behind the function, but I'm not sure how I could use another policy. Maybe you can provide a toy example?

kris-singh: Okay so basically if somebody wants to have a new policy he would have to go ahead build another policy for the momentum update. Something decay_momentum.hpp. I see that this would not be very useful. Should I provide some more decay policies. Or rather have the UpdateMomentum accept and function that can be implemented by the user that takes momentum and returns the updated momentum. Let me know what you think?

zoq: Let's suppose I have an momentum policy called ``MyMomentum``, how could I use ``NAG`` using ``MyMomentum`` to optimize e.g. LogisticRegression, I'm not sure I get that part.

rcurtin:
I think we can close this one since we have Nesterov momentum SGD in ensmallen now:

http://ensmallen.org/docs.html#nesterov-momentum-sgd

(feel free to reopen if I am wrong)

PR #925, opened 2017-03-07
Sgd momentum updates with policy class.

arunreddy:
Implemented SGD with the momentum update using policy class towards  #893.

### DIFF
--- a/src/mlpack/core/optimizers/sgd/sgd.hpp
+++ b/src/mlpack/core/optimizers/sgd/sgd.hpp
@@ -13,6 +13,7 @@
 #define MLPACK_CORE_OPTIMIZERS_SGD_SGD_HPP
 
 #include <mlpack/prereqs.hpp>
+#include <mlpack/core/optimizers/sgd/update_policies/empty_update.hpp>
 
 namespace mlpack {
 namespace optimization {
@@ -72,7 +73,8 @@ namespace optimization {
  * @tparam DecomposableFunctionType Decomposable objective function type to be
  *     minimized.
  */
-template<typename DecomposableFunctionType>
+template<typename DecomposableFunctionType,
+         typename UpdatePolicyType = EmptyUpdate>
 class SGD
 {
  public:
@@ -93,6 +95,7 @@ class SGD
    *     function is visited in linear order.
    */
   SGD(DecomposableFunctionType& function,
+      UpdatePolicyType updatePolicyType = UpdatePolicyType(),
       const double stepSize = 0.01,
       const size_t maxIterations = 100000,
       const double tolerance = 1e-5,
@@ -112,6 +115,14 @@ class SGD
   const DecomposableFunctionType& Function() const { return function; }
   //! Modify the instantiated function.
   DecomposableFunctionType& Function() { return function; }
+
+//  //! Get the instantiated update policy type.
+//  const UpdatePolicyType UpdatePolicyType() const
+//  { return updatePolicyType; }
+//
+//  //! Modify the instantiated update policy type.
+//  UpdatePolicyType& UpdatePolicyType()
+//  { return updatePolicyType; }
 
   //! Get the step size.
   double StepSize() const { return stepSize; }
@@ -136,6 +147,9 @@ class SGD
  private:
   //! The instantiated function.
   DecomposableFunctionType& function;
+
+  //! The update policy used to update the parameters in each iteration.
+  UpdatePolicyType updatePolicyType;
 
   //! The step size for each example.
   double stepSize;
--- a/src/mlpack/core/optimizers/sgd/sgd.hpp
+++ b/src/mlpack/core/optimizers/sgd/sgd.hpp
@@ -95,7 +95,13 @@ class SGD
    *     function is visited in linear order.
    */
   SGD(DecomposableFunctionType& function,
-      UpdatePolicyType updatePolicyType = UpdatePolicyType(),
+      const double stepSize = 0.01,
+      const size_t maxIterations = 100000,
+      const double tolerance = 1e-5,
+      const bool shuffle = true);
+
+  SGD(DecomposableFunctionType& function,
+      UpdatePolicyType updatePolicyType,
       const double stepSize = 0.01,
       const size_t maxIterations = 100000,
       const double tolerance = 1e-5,
@@ -164,6 +170,9 @@ class SGD
   //! iterating.
   bool shuffle;
 };
+
+template<typename DecomposableFunctionType>
+using StandardSGD = SGD<DecomposableFunctionType, EmptyUpdate>;
 
 } // namespace optimization
 } // namespace mlpack
--- a/src/mlpack/core/optimizers/sgd/sgd.hpp
+++ b/src/mlpack/core/optimizers/sgd/sgd.hpp
@@ -98,7 +98,7 @@ class SGD
       const size_t maxIterations = 100000,
       const double tolerance = 1e-5,
       const bool shuffle = true,
-      const UpdatePolicy updatePolicy = VanillaUpdate());
+      const UpdatePolicy updatePolicy = UpdatePolicy());
 
   /**
    * Optimize the given function using stochastic gradient descent.  The given
--- a/src/mlpack/core/optimizers/sgd/sgd.hpp
+++ b/src/mlpack/core/optimizers/sgd/sgd.hpp
@@ -29,19 +29,14 @@ namespace optimization {
  * \f]
  *
  * and our task is to minimize \f$ A \f$.  Stochastic gradient descent iterates
- * over each function \f$ f_i(A) \f$, producing the following update scheme:
- *
- * \f[
- * A_{j + 1} = A_j + \alpha \nabla f_i(A)
- * \f]
- *
- * where \f$ \alpha \f$ is a parameter which specifies the step size.  \f$ i \f$
- * is chosen according to \f$ j \f$ (the iteration number).  The SGD class
- * supports either scanning through each of the \f$ n \f$ functions \f$ f_i(A)
- * \f$ linearly, or in a random sequence.  The algorithm continues until \f$ j
- * \f$ reaches the maximum number of iterations---or when a full sequence of
- * updates through each of the \f$ n \f$ functions \f$ f_i(A) \f$ produces an
- * improvement within a certain tolerance \f$ \epsilon \f$.  That is,
+ * over each function \f$ f_i(A) \f$, based on the specified update policy. By
+ * default vanilla update policy (see mlpack::optimization::VanillaUpdate) is
+ * used. The SGD class supports either scanning through each of the \f$ n \f$
+ * functions \f$ f_i(A)\f$ linearly, or in a random sequence.  The algorithm
+ * continues until \f$ j\f$ reaches the maximum number of iterations---or when a
+ * full sequence of updates through each of the \f$ n \f$ functions \f$ f_i(A)
+ * \f$ produces an improvement within a certain tolerance \f$ \epsilon \f$.
+ * That is,
  *
  * \f[
  * | f(A_{j + n}) - f(A_j) | < \epsilon.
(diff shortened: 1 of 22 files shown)


rcurtin:
Should we close #896 then in favor of this one?

arunreddy:
Hello @rcurtin, Yes please.

src/mlpack/core/optimizers/sgd/sgd.hpp
```diff
@@ -99,6 +102,26 @@ class SGD
...
+   *     during the iterations.
+   * @param stepSize Step size for each iteration.
+   * @param maxIterations Maximum number of iterations allowed (0 means no
+   *     limit).
+   * @param tolerance Maximum absolute tolerance to terminate algorithm.
+   * @param shuffle If true, the function order is shuffled; otherwise, each
+   *     function is visited in linear order.
+   */
+  SGD(DecomposableFunctionType& function,
+      UpdatePolicyType updatePolicyType,
```
rcurtin: I would actually suggest making this the last parameter and giving it a default: `UpdatePolicyType updatePolicyType = UpdatePolicyType()`.  That would save you from needing to have two different constructor overloads.

src/mlpack/methods/logistic_regression/logistic_regression.hpp
```diff
@@ -98,9 +99,9 @@ class LogisticRegression
    *
    * @param optimizer Instantiated optimizer with instantiated error function.
    */
-  template<template<typename> class OptimizerType>
+  template<template<typename, typename ... PolicyTypeArg> class OptimizerType, typename ... PolicyTypeArg>
```
rcurtin: I think you can make the line shorter by giving no name to the first `PolicyTypeArg`:

```
template<template<typename, typename...> class OptimizerType, typename... OptimizerTypeArgs>
```

You can also remove the other overload of `Train(OptimizerType&)`, this will work in its place.

src/mlpack/tests/momentum_sgd_test.cpp
```diff
@@ -0,0 +1,44 @@
...
+  MomentumSGD<SGDTestFunction> s(f, momentumUpdate, 0.0003, 5000000, 1e-9, true);
+
+  arma::mat coordinates = f.GetInitialPoint();
+  double result = s.Optimize(coordinates);
+
+  BOOST_REQUIRE_CLOSE(result, -1.0, 0.05);
+  BOOST_REQUIRE_SMALL(coordinates[0], 1e-3);
+  BOOST_REQUIRE_SMALL(coordinates[1], 1e-7);
+  BOOST_REQUIRE_SMALL(coordinates[2], 1e-7);
+}
```
rcurtin: This is a very short test file, can you either add some more tests for the optimizer (this would be my preference) or merge it with the file that is testing the `StandardSGD` class please?

src/mlpack/core/optimizers/sgd/update_policies/momentum_update.hpp
```diff
@@ -0,0 +1,72 @@
...
+#ifndef MLPACK_CORE_OPTIMIZERS_SGD_MOMENTUM_UPDATE_HPP
+#define MLPACK_CORE_OPTIMIZERS_SGD_MOMENTUM_UPDATE_HPP
+
+namespace mlpack {
+namespace optimization {
+
+/**
+ * Momentum update policy for SGD.
+ *
+ */
```
rcurtin: Could I ask you to please provide some more documentation on the momentum update policy (what it is, what it does, and how it works) in the comments for this class please?

src/mlpack/core/optimizers/sgd/update_policies/vanilla_update.hpp
```diff
@@ -0,0 +1,46 @@
...
+   * @param n_cols
+   */
+  void Initialize(const size_t n_rows,
+                  const size_t n_cols)
+  {/* Do Nothing */}
+
+  arma::mat Update(double stepSize, arma::mat gradient)
+  {
+    // perform the vanilla SGD update.
+    return - stepSize * gradient;
```
rcurtin: Why the space between `-` and `stepSize`?  In my opinion that's a little confusing, can you make it `-stepSize` please?

src/mlpack/core/optimizers/sgd/update_policies/momentum_update.hpp
```diff
@@ -0,0 +1,72 @@
...
+   * Momentum update constructor.
+   *
+   * @param momentum The momentum hyperparameter
+   */
+  MomentumUpdate(const double momentum = 0.5):
+      momentum(momentum)
+  {/* do nothing */};
+
+  /**
+   * Intialize the velocity matrix to the zeros matrix of the given size
```
rcurtin: It's probably worth pointing out in the comments here that this method is called by the optimizer at the start of optimization.

src/mlpack/core/optimizers/sgd/update_policies/vanilla_update.hpp
zoq: I'm not really a fan of ``arma::mat Update(double stepSize, arma::mat gradient)`` (creating a copy twice return and gradient), what do you think about ``void Update(arma::mat& iterate, const double stepSize, const arma::mat& gradient)``?

src/mlpack/core/optimizers/sgd/sgd.hpp
```diff
@@ -72,7 +75,7 @@ namespace optimization {
  * @tparam DecomposableFunctionType Decomposable objective function type to be
  *     minimized.
  */
-template<typename DecomposableFunctionType>
+template<typename DecomposableFunctionType, typename UpdatePolicyType>
```
zoq: In favor of backward compatibility, I would go with a default for UpdatePolicyType.

src/mlpack/core/optimizers/sgd/sgd.hpp
```diff
@@ -137,6 +160,9 @@ class SGD
   //! The instantiated function.
   DecomposableFunctionType& function;
 
+  //! The update policy used to update the parameters in each iteration.
+  UpdatePolicyType updatePolicyType;
```
zoq: I think we could use ``updatePolicy``instead of ``updatePolicyType``, it's not a datatype but it's a policy type, so I think it also works. What do you think?

src/mlpack/core/optimizers/sgd/update_policies/vanilla_update.hpp
```diff
@@ -0,0 +1,46 @@
...
+ * Vanilla update policy for SGD.
+ *
+ */
+class VanillaUpdate {
+ public:
+
+  /**
+   * Function for runtime initialization.
+   *
+   * @param n_rows
```
zoq: We should add a description for each parameter. That also applies for the Update function.

src/mlpack/core/optimizers/sgd/update_policies/vanilla_update.hpp
rcurtin: Oh, yeah, I missed the pass-by-value of the gradient, that should definitely be `const arma::mat&`.  I think for `MomentumPolicy` the compiler would be smart enough to avoid the unnecessary copy, but I don't think that will apply here, so I agree, `void Update(arma::mat& iterate, const double stepSize, const arma::mat& gradient)` would be the right way to go.

arunreddy: I thought at compile time the compiler is going to avoid copying when templates are used. I can update it to pass-by-reference.

src/mlpack/core/optimizers/sgd/update_policies/momentum_update.hpp
arunreddy: @rcurtin: Can I move the comments(as shown below) related to vanilla update in sgd.hpp to vanilla_update.hpp ? 

``` c++
Stochastic gradient descent iterates
 * over each function \f$ f_i(A) \f$, producing the following update scheme:
 *
 * \f[
 * A_{j + 1} = A_j + \alpha \nabla f_i(A)
 * \f]
 *
 * where \f$ \alpha \f$ is a parameter which specifies the step size.  \f$ i \f$
 * is chosen according to \f$ j \f$ (the iteration number).

```

src/mlpack/core/optimizers/sgd/sgd.hpp
```diff
@@ -96,7 +97,8 @@ class SGD
       const double stepSize = 0.01,
       const size_t maxIterations = 100000,
       const double tolerance = 1e-5,
-      const bool shuffle = true);
+      const bool shuffle = true,
+      const UpdatePolicy updatePolicy = VanillaUpdate());
```
zoq: I think if you like to use MomentumSGD without passing the Momentum policy this fails, because  in that case UpdatePolicy is MomentumUpdate and can't be assigned with VanillaUpdate(), instead if you use ``const UpdatePolicy updatePolicy = UpdatePolicy()`` it should be fine.

src/mlpack/core/optimizers/sgd/update_policies/momentum_update.hpp
```diff
@@ -0,0 +1,110 @@
...
+ *
+ * For more information, please refer to the Section 8.3.2 of the following book
+ *
+ * @code
+ * @book{Goodfellow-et-al-2016,
+ *  title={Deep Learning},
+ *  author={Ian Goodfellow and Yoshua Bengio and Aaron Courville},
+ *  publisher={MIT Press},
+ *  note={\url{http://www.deeplearningbook.org}},
+ *  year={2016}
```
zoq: I'm a fan of giving credit to the first paper, that proposed the method, which I guess is "Learning representations by back-propagating errors" by Rumelhart et al. if you like we can cite both papers/books. What do you think?

src/mlpack/core/optimizers/sgd/update_policies/momentum_update.hpp
```diff
@@ -0,0 +1,110 @@
...
+   * way as momentum term increases for dimensions pointing in the same and
+   * reduces updates for dimensions whose gradients change directions.
+   *
+   * @param iterate Parameters that minimize the function.
+   * @param stepSize Step size to be used for the given iteration.
+   * @param gradient The gradient matrix.
+   */
+  void Update(arma::mat& iterate,
+              const double stepSize,
+              arma::mat& gradient)
```
zoq: Do you think we could use ``const arma::mat& gradient`` instead of ``arma::mat& gradient`` here?

src/mlpack/tests/momentum_sgd_test.cpp
```diff
@@ -0,0 +1,79 @@
...
+
+using namespace std;
+using namespace arma;
+using namespace mlpack;
+using namespace mlpack::optimization;
+using namespace mlpack::optimization::test;
+
+BOOST_AUTO_TEST_SUITE(MomentumSGDTest);
+
+BOOST_AUTO_TEST_CASE(MomentumSGDSpeedUpTestFunction)
```
zoq: Can you tell us something about the runtime of this test? ``time bin/mlpack_test -t MomentumSGDTest/MomentumSGDSpeedUpTestFunction``

zoq:
There are a couple minor style issues that we can fix once it's merged, besides that, I think it's ready to be merged, @rcurtin any more comments?

arunreddy:
@zoq: The `MomentumSGDSpeedUpTestFunction` test runs in 4.538s on an intel 6850k processor at 3.5Ghz.

zoq:
@arunreddy Thanks for the timings. Just wanted to make sure the test doesn't take to long to run.

 I guess once we have the improved build setup I can just take a look at: http://masterblaster.mlpack.org/job/mlpack%20-%20git%20commit%20test/2642/testReport/mlpackTest/

rcurtin:
Hi @arunreddy, thanks for this PR, it is a nice improvement.  I made some minor style fixes in b70700c9 and 3b9917fa.

arunreddy:
hello @rcurtin, good to see the pr merged. I will ensure the styling is handled well next time. Thanks.

Issue #921, opened 2017-03-06
Use new Load() overloads when only a vector is required


rcurtin:
There are many places in command-line programs (e.g. `logistic_regression_main.cpp` and others) where we have code like

```
arma::Mat<size_t> labelsIn;
data::Load(labelsFile, labelsIn, true);
// convert this to an arma::Row<size_t> or whatever
```

It would be much better if we dispensed with this step and simply loaded into the correct Armadillo type directly, and this would allow us to reduce the size of the codebase a little bit.  A PR to fix this issue should handle all occurrences of this pattern in the mlpack codebase.  But a PR can't be merged until #920 is fixed---we need to ensure that a user, when they pass an invalid labels file or something, gets an exception so they know something is wrong.

ShihaoJing:
@rcurtin I'd like to take this.

ShihaoJing:
I noticed that in `logistic_regression_main.cpp` and others, lables can loaded this way:
```
responsesMat = std::move(CLI::GetParam<arma::Mat<size_t>>("labels"));
if (responsesMat.n_cols == 1)
      responses = responsesMat.col(0).t();
    else
      responses = responsesMat.row(0);
```
When I tried to change parma loading to 
`responsesMat = std::move(CLI::GetParam<arma::Col<size_t>>("labels"));`
It gives me some errors:
```
/Users/shihaojing/Projects/mlpack/src/mlpack/../mlpack/core/util/param_data_impl.hpp:70:5: error: no matching function for call to 'Load'
    data::Load(value, matrix, true, !d.noTranspose);
    ^~~~~~~~~~
/Users/shihaojing/Projects/mlpack/src/mlpack/../mlpack/core/util/cli_impl.hpp:194:16: note: in instantiation of function template specialization
      'mlpack::util::HandleParameter<arma::Col<double> >' requested here
  return util::HandleParameter<T>(v, d);
               ^
/Users/shihaojing/Projects/mlpack/src/mlpack/methods/linear_regression/linear_regression_main.cpp:143:34: note: in instantiation of function template
      specialization 'mlpack::CLI::GetParam<arma::Col<double> >' requested here
      responses = std::move(CLI::GetParam<vec>("training_responses"));
                                 ^
/Users/shihaojing/Projects/mlpack/src/mlpack/../mlpack/core/data/load_impl.hpp:112:6: note: candidate function [with eT = double] not viable: no known
[...]

Issue #920, opened 2017-03-06
Make Load() fail correctly when not in debugging mode for column vectors


rcurtin:
@Lakshya1605 just contributed some nice support that allows us to call `data::Load()` with column vectors and row vectors.  That could clean up a lot of the code that loads labels (which need to be one dimensional).

But the problem is, when the file that is loaded has more than one dimension, no exception is thrown when mlpack is compiled without debugging symbols (`-DDEBUG=OFF`, the default).  However, we would like some kind of failure to happen, of course, otherwise users will have their data incorrectly loaded silently and will encounter weird errors later in their code.

This should be a simple fix at its core; we should not have to do anything crazy like partially load the CSV or anything, we should be able to do this with existing Armadillo functionality.

sam0410:
Hi @rcurtin !
I would work on this

sam0410:
Hi @rcurtin!
I was thinking of this-
In row vector, we can check the number of lines(rows) in the file and that should be equal to one, otherwise, an error is thrown.
In column vector, we can find number of spaces(excluding spaces which are there in the beginning and the end) in the first row of the file input, if it is not equal to zero, we can throw an error.
Can I go on and implement this?
Thanks!

rcurtin:
Like I wrote:

> This should be a simple fix at its core; we should not have to do anything crazy like partially load the CSV or anything, we should be able to do this with existing Armadillo functionality.

I am fairly convinced this can be done without having to open the file manually.

SND96:
Hello @rcurtin!
I was thinking we could use data::load for both loading the same data into a column/row vector and a matrix. We could then use the size() function to check whether the column/row vector has the same dimensions as the matrix. If not, an error will be shown.
If this is feasible, can I go on and implement this?

rcurtin:
That doubles the loading time.  That won't be a good idea.

sam0410:
We can just load it into a matrix and check the size of matrix if number of columns =1 or number of rows=1

SND96:
@sam0410  Since `data::load()` can now be called with column/row vectors, wouldn't using a matrix for loading and then converting that to a vector not use that new functionality at all?

sam0410:
Hi @SND96 ! I was thinking,it would be loaded to a vector later and new matrix which we created will be destroyed after the function scope is over,

SND96:
@sam0410 I had suggested something similar but that would double the load time as rcurtin had said before.

sam0410:
[...]

PR #927, opened 2017-03-10
Issue#921

ShihaoJing:
For #921.
Add PARAM_COL and PARAM_ROW for loading vectors from files directly.


ShihaoJing:
@rcurtin code will be ready for merge when #920 is merged.

src/mlpack/core/util/output_param_impl.hpp
```diff
@@ -48,9 +48,16 @@ void OutputParamImpl(
 {
   const T& output = *boost::any_cast<T>(&data.mappedValue);
   const std::string& filename = *boost::any_cast<std::string>(&data.value);
-
+  
   if (output.n_elem > 0 && filename != "")
-    data::Save(filename, output, false, !data.noTranspose);
+  {
+    //call the correct data::Save()
+    if (data.tname.find("Col") != std::string::npos || data.tname.find("Row") != std::string::npos)
```
rcurtin: This is a clever way to do it, but unfortunately there's no guarantee that the typename (as given by RTTI) has the class name in it.  On gcc and clang that will probably always work, but it won't be portable to every possible compiler. :)

I think, maybe a better way to do this, might be to use the (undocumented) `arma::is_Row<T>` and `arma::is_Col<T>` structs:

```
if (arma::is_Row<T>::value || arma::is_Col<T>::value)
```

Technically these structs are not documented by Armadillo and we run the risk that future versions may remove or change the names, but in my opinion it doesn't make much sense to reinvent the wheel by making our own identical structs, and if Armadillo changes we can backport or otherwise handle it when that happens.

src/mlpack/methods/decision_stump/decision_stump_main.cpp
```diff
@@ -128,15 +128,12 @@ int main(int argc, char *argv[])
...
     {
-      labelsIn = std::move(CLI::GetParam<Mat<size_t>>("labels"));
+      labelsIn = std::move(CLI::GetParam<Row<size_t>>("labels"));
 
-      // Do the labels need to be transposed?
-      if (labelsIn.n_cols == 1)
-        labelsIn = labelsIn.t();
-      if (labelsIn.n_cols == 1)
+      if (labelsIn.n_rows > 1)
         Log::Fatal << "Labels must be one-dimensional!" << endl;
```
rcurtin: This check isn't needed anymore, because `labelsIn` can't possibly have more than one row now.  So we can simplify this code.  Later on, when `labelsIn` is used (I believe it's usually used as `labelsIn.row(0)`), that can be simplified to just `labelsIn`.  We can even change the name of the variable to simply `labels`.

rcurtin: This type of pattern appears in a lot of other files too.

ShihaoJing:
Hi Ryan,

Thanks for your review and I'll fix these problems soon. 
I am very willing to do improve `cli_test.cpp`. Please give me some instructions.

PR #928, opened 2017-03-11
Avoid use of internal Armadillo structures in prefixedoustream_impl.hpp

shikharbhardwaj:
Changes `prefixedoustream_impl.hpp`


rcurtin:
Hi Shikhar, thanks for this!  I made some really trivial style changes in 5b97d377.

PR #930, opened 2017-03-11
Add Docker build support

NaxAlpha:
GSoC: Build testing with Docker and VMs
Hi, I have added basic docker build support

https://hub.docker.com/r/naxalpha/mlpack/
http://pastebin.com/K1dEDpMp

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,16 @@
+# Select Image
+FROM gcc:5
+# Developer Info
+MAINTAINER Ryan Curtin <ryan@ratml.org>
+# Update Packages
+RUN apt-get update && apt-get -y upgrade
+# Install Dependencies
+RUN apt-get -y install git cmake subversion libboost-math-dev libboost-program-options-dev libboost-test-dev libboost-serialization-dev libarmadillo-dev binutils-dev
+# Copy Source
+COPY . /mlpack
+# Current Directory
+WORKDIR /mlpack
+# Build mlpack
+RUN mkdir build && cd build && cmake ../ && make && make mlpack_test
+# Start Shell
+CMD ["/bin/bash"]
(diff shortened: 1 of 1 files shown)


NaxAlpha:
To use mlpack from Docker, follow these steps

- Install Docker
- Run Docker (Quickstart Terminal in Windows)
- `docker pull naxalpha/mlpack` This command will download docker image (Max Size: 1.5 GB)
- ` docker run --rm -it -v $PWD:/mlpack-run -w /mlpack-run naxalpha/mlpack` To run mlpack apps from current directory
- `mlpack_knn --k=5 --reference_file=input.csv --distances_file=distances.csv --neighbors_file=neighbors.csv` Make sure `input.csv` is in current directory, after running command, `distance.csv` and `neighbors.csv` will be in current directory of your **host OS**

rcurtin:
Hi there,

I'm not sure what the goal of this Docker container is.  Is this for the build testing project?  That project is actually focused on using Docker containers to get different versions of the library dependencies, so that the build system can use these Docker containers to build mlpack.  So a Docker container that just contains mlpack would not be what's needed there.  If you can clarify the purpose here, we can figure out where to go with it.

Thanks!

NaxAlpha:
Hi @rcurtin, 

At the moment, basic goal of Docker container is to provide build for testing project and testing mlpack latest version without install it on system. Later on, we can have build for multiple version of mlpack.

Thanks.

rcurtin:
Right, I see.  If that is the case, the Dockerfiles to build the containers should be included in the `jenkins-conf` repository since that is where the containers will actually be built.  But I don't think there is reason to include this Dockerfile there until the build system is ready, and I don't think it's appropriate to include a Dockerfile in the main mlpack repository, so I'll go ahead and close this.  But, it is still a good start towards the testing project!

NaxAlpha:
Ok got it!

PR #931, opened 2017-03-11
Add PARAM_COL and PARAM_ROW for loading vectors from files directly

ShihaoJing:
For #920.
@rcurtin Sorry, I messed up my original PR so I create a new one. Problems you mentioned in that PR have been addressed. 
By the way, can you explain more about how to improve `cli_test.cpp`?

### DIFF
--- a/src/mlpack/core/data/save_impl.hpp
+++ b/src/mlpack/core/data/save_impl.hpp
@@ -28,6 +28,22 @@ namespace data {
 
 template<typename eT>
 bool Save(const std::string& filename,
+          const arma::Col<eT>& vec,
+          const bool fatal)
+{
+  return Save(filename, vec, fatal, false);
+}
+
+template<typename eT>
+bool Save(const std::string& filename,
+          const arma::Row<eT>& rowvec,
+          const bool fatal)
+{
+  return Save(filename, rowvec, fatal, false);
+}
+
+template<typename eT>
+bool Save(const std::string& filename,
           const arma::Mat<eT>& matrix,
           const bool fatal,
           bool transpose)
--- a/src/mlpack/core/util/output_param_impl.hpp
+++ b/src/mlpack/core/util/output_param_impl.hpp
@@ -50,7 +50,12 @@ void OutputParamImpl(
   const std::string& filename = *boost::any_cast<std::string>(&data.value);
 
   if (output.n_elem > 0 && filename != "")
-    data::Save(filename, output, false, !data.noTranspose);
+  {
+      if (arma::is_Row<T>::value || arma::is_Col<T>::value)
+        data::Save(filename, output, false);
+      else
+        data::Save(filename, output, false, !data.noTranspose);
+  }
 }
 
 //! Output a model option (this saves it to file).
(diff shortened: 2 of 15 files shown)


rcurtin:
Hi Shihao,

Thanks again for this PR.  The idea with the tests would just be to test that we can properly use Row and Col parameters.  Take a look at `InputMatrixParamTest` for an idea:

```
BOOST_AUTO_TEST_CASE(InputMatrixParamTest)
{
  AddRequiredCLIOptions();

  // --matrix is an input parameter; it won't be transposed.
  CLI::Add<arma::mat>(arma::mat(), "matrix", "Test matrix", 'm', false, true,
      false);

  // Set some fake arguments.
  const char* argv[3];
  argv[0] = "./test";
  argv[1] = "-m";
  argv[2] = "test_data_3_1000.csv";

  int argc = 3;

  // The const-cast is a little hacky but should be fine...
  CLI::ParseCommandLine(argc, const_cast<char**>(argv));

  // The --matrix parameter should exist.
  BOOST_REQUIRE(CLI::HasParam("matrix"));
  // The --matrix_file parameter should not exist (it should be transparent from
  // inside the program).
  BOOST_REQUIRE_THROW(CLI::HasParam("matrix_file"), runtime_error);

  arma::mat dataset = CLI::GetParam<arma::mat>("matrix");
  arma::mat dataset2 = CLI::GetParam<arma::mat>("matrix");

  BOOST_REQUIRE_EQUAL(dataset.n_rows, 3);
  BOOST_REQUIRE_EQUAL(dataset.n_cols, 1000);
  BOOST_REQUIRE_EQUAL(dataset2.n_rows, 3);
  BOOST_REQUIRE_EQUAL(dataset2.n_cols, 1000);

  for (size_t i = 0; i < dataset.n_elem; ++i)
    BOOST_REQUIRE_CLOSE(dataset[i], dataset2[i], 1e-10);
}
```

There is also `OutputMatrixParamTest`.  We should add a test like this for each of the parameter types you've implemented.  The tests are pretty simple, so it shouldn't be hard.  It will probably be a little repetitive, but that's ok.

ShihaoJing:
@rcurtin PR is ready for merge.

rcurtin:
Tests look good, but could you add some for `PARAM_UCOL` and `PARAM_UROW` also please? It's important to test these also since they will be so widely used.

I think there is a typo in some of the tests too, it seems like you compare the right size of `vec1` twice after `GetParam()`, I think the second comparison should be for `vec2`.

ShihaoJing:
@rcurtin I've added tests for PARAM_UCOL and PARAM_UROW and fixed some typos. Please let me know if there is any problem. Thanks :)

rcurtin:
Hi Shihao,

Thanks for the nice contribution.  If you like you can add your name to the list of contributors in `src/mlpack/core.hpp` and `COPYRIGHT.txt`.  It also looks like your Github account is not linked to the email you used to commit, so your Github user page won't show these commits.  That should be easy to fix if that is a thing that is important to you. :)

I had to make some modifications to related code to get all of this to work.  `data::Load()` for vectors had to be totally rewritten so that we could load transposed vectors also.  i.e. you could load a file like "1\n2\n3" into a row vector (which is a common scenario).  That was done in 5dab608 and 8831683.  I also made some slight style changes to your changes in 7f9edcc and 9419bc3.  It looks like your editor uses tab characters, please use 2-space tabs in the future instead.

I also tested all of the modified executables and this revealed that the output that predictions were being saved as was changed, so I made commit 120c178 which makes sure that every vector that's output (whether row or column vector) is one element per line (i.e. a column vector).

rcurtin:
I actually did merge this but it looks like I messed up Github's detection of the merge.  In either case, all the commits are part of master now so I'll close the PR.

PR #933, opened 2017-03-12
Add Dependencies' install command for Ubuntu/Debian

outofnames:
It would be better to provide dependencies install command to assist someone who is building by following this readme. (Otherwise we get an error in cmake which may confuse people)

### DIFF
--- a/README.md
+++ b/README.md
@@ -77,6 +77,11 @@ All of those should be available in your distribution's package manager.  If
 not, you will have to compile each of them by hand.  See the documentation for
 each of those packages for more information.
 
+In Ubuntu and Debian, you can get all of these dependencies through apt:
+
+    $ apt-get install cmake libboost-math-dev libboost-program-options-dev
+    libboost-test-dev libboost-serialization-dev libarmadillo-dev binutils-dev
+
 If you are compiling Armadillo by hand, ensure that LAPACK and BLAS are enabled.
 
 4. Building mlpack from source


zoq:
Hello, thanks for the contribution, I'm not sure that adding OS related install steps to the main README is a good idea. The amount of information might be overwhelming and is probably only interesting for some of the users. I guess, we should even thing about removing the section about using 'sudo apt-get install libmlpack-dev' to install all ubuntu dependencies, which is only true if a newer version is used that already installs boost serialization and not libxml. On the other side providing build informations is crucial and maybe a better idea is to add a link to: http://www.mlpack.org/docs/mlpack-2.1.1/doxygen.php?doc=build.html#build

Eager to hear some more input on this, so let's discuss this.

outofnames:
@zoq , your view on too much information makes sense. A link explicitly redirecting Ubuntu/linux users would do the job.
We can either strip down the building section and add a link to the website's tutorial page(just like windows) or we should provide the complete commands for Ubuntu (I was following this readme to build and didn't installed dependencies, and got cmake errors. A novice user with insufficient knowledge about cmake may get stuck in it)

I think it would be better to strip down the whole section and redirect users to the website if you want to make things precise.

rcurtin:
I agree with @zoq that OS-related install steps should be generally avoided where possible.  If the names of certain packages change or the dependencies of mlpack change, it is very easy for our documentation to go out of date.

I took a look through the readme and honestly I don't see a need to make changes:

 * The build section is preceded by a section on what the dependencies are, which mentions that `All of those should be available in your distribution's package manager.`
 * The build section itself has a link to the more detailed build tutorial generated by Doxygen (as well as the Windows build).
 * The build section already suggests that installing `libmlpack-dev` (or whatever other package on another distribution) might be easier.
 * The build instructions are distribution-agnostic (with the exception of the `libmlpack-dev` command above, which was only added because lots of Ubuntu users complained; but technically that is not actually part of the build).

I'm not against improving the readme, but I'm not convinced that anything here needs to be improved.  In your case it sounds like the answer---that you needed dependencies---was laid out in the section before the build instructions.  Have I overlooked something?  I suppose, maybe we could add a sentence to the build instructions pointing out that dependencies must be installed for CMake to configure correctly, and that could help clarify the confusion that you had.

For what it's worth, typically novice users who only want to use the library should stick to installing `libmlpack-dev` instead of compiling from source.

rcurtin:
Closing for inactivity and for the reasons above.  Feel free to reopen if needed.

PR #934, opened 2017-03-13
Implement a framework of DQN and asynchronous learning methods

ShangtongZhang:
DQN with experience replay and target network.
Tested in Mountain Car task.


ShangtongZhang:
I use the latest version of armadillo. But it seems some new functions don't exist in old version (based on the error log of Jenkins). Which version of armadillo is recommended?
Thanks,

rcurtin:
Hi Shangtong,

The minimum required version of Armadillo for mlpack is currently Armadillo 4.200.0.  I think that is the version that Travis builds with.  You can take a look at this page to get an idea of what has changed over different versions: http://arma.sourceforge.net/docs.html#api_additions

ShangtongZhang:
Hi Ryan,
Thanks for your response. I think I should then install armadillo 4.2 and try to make it compiled in Travis CI.

zoq:
Hello @ShangtongZhang just noticed that implemented the CartPole task, do you think we could use the code Bang has written last year? https://github.com/BangLiu/mlpack/blob/ne/src/mlpack/methods/ne/tasks.hpp If you think that's a good idea, I can merge it in, so that we can test it. Let me know what you think.

ShangtongZhang:
Hi @zoq , maybe not. Although the dynamics are same, but the interfaces are quite different. And his implementation seems to be coupled with NE. Also I think the code style and readability of that implementation is poor...

ShangtongZhang:
Hi @zoq, would you like to apply [this](https://github.com/mlpack/mlpack/pull/934/commits/effa86d84fe95f437cbd29e2913b901a8a072a98) commit first? Currently if I train FFN with SGD it will lead to compile error as SGD has two template arguments.

zoq:
Yeah, I was going to fix this, can you open a separate PR for the change or should I cherry pick the commit?

ShangtongZhang:
I think cherry pick is ok

ShangtongZhang:
Oh.. My fix works for gcc and clang, but fails with msvc.. So I just pushed a new fix.
https://godbolt.org/g/cqIuhD
http://webcompiler.cloudapp.net/ (Need to copy the code manually)

ShangtongZhang:
(Double) DQN works well in my local environment with MountainCar and CartPole. I don't know why VanillaNetworkTest always fails in travis-ci -- it also succeeds in my local environment.

zoq:
Maybe Travis found an efficient method to intercept the test process, just to run each test with a bad seed selected from a project depended pool of bad seeds. I guess in this case they have to update the pool, since I changed the code: ae93d749a2fed8c5418df14dbfbff95

.gitignore
```diff
@@ -3,3 +3,7 @@ xcode*
 .DS_Store
 src/mlpack/core/util/gitversion.hpp
 src/mlpack/core/util/arma_config.hpp
+.idea
+cmake-build-*
+
+
```
rcurtin: I think the last two empty lines are unnecessary here.

src/mlpack/methods/ann/ffn.hpp
```diff
@@ -58,6 +61,12 @@ class FFN
   FFN(OutputLayerType&& outputLayer = OutputLayerType(),
       InitializationRuleType initializeRule = InitializationRuleType());
 
+  //! Copy CTOR
+  FFN(const NetworkType&);
+
+  //! Assignment CTOR
+  NetworkType& operator = (const NetworkType&);
```
rcurtin: It might be clearer to use `FFN` here instead of `NetworkType`.

src/mlpack/methods/ann/init_rules/gaussian_init.hpp
```diff
@@ -35,7 +35,7 @@ class GaussianInitialization
    * @param variance Variance of the gaussian
    */
   GaussianInitialization(const double mean = 0, const double variance = 1) :
-      mean(mean), variance(variance)
+          mean(mean), variance(variance)
```
rcurtin: I think the indentation was correct before.

src/mlpack/methods/ann/visitor/assign_visitor.hpp
```diff
@@ -0,0 +1,37 @@
...
+#ifndef MLPACK_METHODS_ANN_VISITOR_ASSIGN_VISITOR_HPP
+#define MLPACK_METHODS_ANN_VISITOR_ASSIGN_VISITOR_HPP
+
+#include <mlpack/methods/ann/layer/layer_types.hpp>
+#include <boost/variant.hpp>
+
+namespace mlpack {
+namespace ann {
+
+class AssignVisitor : public boost::static_visitor<void>
```
rcurtin: Can you add some comments on what this class does and why it's needed?

src/mlpack/methods/reinforcement_learning/async_learning_impl.hpp
```diff
@@ -0,0 +1,108 @@
+#ifndef MLPACK_METHODS_RL_ASYNC_LEARNING_IMPL_HPP
```
rcurtin: Don't forget the file header with name, comment, and license too. :)

src/mlpack/tests/async_learning_test.cpp
```diff
@@ -0,0 +1,58 @@
...
+
+using namespace mlpack;
+using namespace mlpack::ann;
+using namespace mlpack::rl;
+
+BOOST_AUTO_TEST_SUITE(AsyncLearningTest);
+
+BOOST_AUTO_TEST_CASE(CartPoleTest)
+{
+  arma::arma_rng::set_seed_random();
```
rcurtin: We shouldn't set random seeds in the test, so that a failing test is reproducible when it's compiled.

ShangtongZhang: Thanks for the response. The implementation of A3C (`bootstrap`, `async_learning.hpp`) in this code is incomplete, it can only compile, just to show the design of API in my proposal. And recently I'm busy with my final so I haven't had time to comment these files and debug them further. The other files (implementation of DQN) is well documented and I have tested the DQN implementation to make sure it works well. I will add more comments this weekend for all the files. But I'm afraid I'm not able to debug A3C further until after next week when all my courses end.

rcurtin:
Sure, no hurry! Good luck on your finals. :)

ShangtongZhang:
I just added more detailed comments for all the code, looking forward to any feedback.

zoq:
Hello @ShangtongZhang do you mind to split this PR into multiple parts? One PR for the environments and another PR for the DQN? I think it's easier to catch potential issues and it could be beneficial for others if e.g. they could work with the environments. let me know what you think and then we can go from there.

ShangtongZhang:
Hi @zoq , I think it's a good idea. I can split this PR into Environment, DQN and A3C. As for environment part, do you think I should still put it in reinforcement learning directory, or just create a new directory? I prefer creating a new directory directly under `methods`

zoq:
I don't see the environments as an independent algorithms/method more as a tool, so I would create the directory under the reinforcement learning director, at least for now.

ShangtongZhang:
Yeah it looks wired to treat it as a method. But in the future neuron evolutionary methods also need such kind of environment. So maybe it's better to have an independent directory.

zoq:
You are absolutely right, but I think there is no problem to include the files from the rl directory, in fact we already do this e.g. we use different svd method inside the pca class or different kernels inside the kernel pca, etc.

ShangtongZhang:
This PR will only serve for a track of my progress. I will break it down to small incremental PRs to make it easy to review. BTW, does anyone know how to make `Log::Info` work in test cases? My current cmake config is `-D DEBUG=ON -D PROFILE=OFF -D FORCE_CXX11=ON -D TEST_VERBOSE=ON` and environment is `CXX=/usr/local/bin/clang++-omp;CC=/usr/local/bin/clang-omp;CXXFLAGS=-std=c++11`. But it doesn't output anything to console. Did I miss something?

zoq:
Thanks for breaking it down into smaller parts. Regarding ``Log::Info``, you can set ``mlpack::Log::Info.ignoreInput = false;`` in the test case or ``Log::Debug`` should work if you build with ``DEBUG=ON``.

ShangtongZhang:
Hi @zoq , would you like to look into [this](https://github.com/mlpack/mlpack/pull/934/commits/ab3c5a1854a7768791bbd9d64ed33af6dcea7d65). Currently if you add `std::cout << gaussianModel.Parameters();` at `init_rules_test.cpp` before line 284. You will find the `gaussianModel` isn't initialized properly -- only the last 12 parameters are initialized, all others are zero. And it happens only for gaussian init (i.e. random init and ovis init work well).  I did a tentative fix in the above commit. It works. But I don't know why it works. I don't understand how an assignment operator could affect the memory -- I assume assignment operator should only copy values.

zoq:
Nice catch, that's a really strange issue, not only that it looks like it's allocating new memory; on top of that, it works for the last layer. I have to take a deeper look into the issue. But for now, do you like to open a PR for the issue or should I fix it?

ShangtongZhang:
I'm back ^>^
This week I'll work on implementing async one-step q-learning.
I have two questions now:

- Can this PR https://github.com/mlpack/mlpack/pull/1034 be merged? My new work will depend on that PR, if that's not merged, I cannot make new PR

- Can I use inheritance in my implementation? (without virtual function)

zoq:
Welcome back, how was it?

- It was an easy merge, it doesn't modify any existing interface.
- In general, we avoid inheritance: https://github.com/mlpack/mlpack/wiki/DesignGuidelines#inheritance, maybe we can find a solution around it? Can you elaborate on this?

ShangtongZhang:
@zoq Thanks for merging that PR and the summer school was pretty good! 
I have several classes, they share same member variables and same constructor. The only difference is a member function called `Episode`. Maybe I can use tag dispatch
```
class WorkerTypeA {};

class WorkerTypeB {};

template <typename WorkerType, typename T1, typename T2>
class Worker {
public:
	int data;
	Worker(int i) { data = i; }

	void foo() { foo(WorkerType()); }

	void foo(WorkerTypeA tag) {}
	void foo(WorkerTypeB tag) {}
};

int main() {
	Worker<WorkerTypeA, int, int> worker(1);
	worker.foo();
	return 0;
}
```

zoq:
I guess depending on the use case, yes we could use the tag dispatch design pattern or maybe we could use the policy design pattern: http://www.drdobbs.com/policy-based-design-in-the-real-world/184401861

That way, we  could add more classes, without modifying the main code, what do you think?

ShangtongZhang:
Not sure whether policy based design is applicable. Because in my case it's those policy classes that share a common constructor. Let me have a try to see which is the best practice.

zoq:
I thought something like:

```c++
class OneStepQLearningWorker
{
  static double Episode(...)
  {
  }
}

template<typename WorkerType>
class AsyncLearning
{
  void Episode(...)
  {
    WorkerType::Episode(...);
  }

  
}
````

in this case you would have to implement the logic inside the Worker, or at least part of the logic.

**Comment by ShangtongZhang** [comment]
The problem is `WorkerType::Episode` needs to have access to many member variables of `AsyncLearning`, if I pass them as function parameters, the parameter list will be very long (about 10 or more). If I pass those parameters to CTOR of `worker` directly rather than CTOR of `AsyncLearning`, then those `workers` will share a common CTOR, which will also be very long.

**Comment by zoq** [comment]
I see I guess another solution would be to write a utility class e.g. ``AsyncLearningInfo`` that holds the parameter and then you could pass an instantiation of ``AsyncLearningInfo``. I was also thinking about template specialization but since the class also takes a bunch of other template parameters that's not really an option. What do you think?

**Comment by ShangtongZhang** [comment]
A config class looks good, I think I can do that.

**Comment by ShangtongZhang** [comment]
I think the compiler error (lack of `arma::clamp`) can be solved by https://github.com/mlpack/mlpack/pull/1050. BTW, how do you make your irc always online? Is there a free solution?

**Comment by zoq** [comment]
ah, right, ``arma::clamp`` was introduced in 4.400.

I'm not aware of a free solution, I use irssi (https://irssi.org/) + tmux on a remote machine, which is always online.

**Comment by ShangtongZhang** [comment]
I run the test with `RandomSeed(std::time(NULL));` in my mac 20 times, it all passed. The problem is the algorithm is multi-thread (18 threads in total), and isn't synchronized. So the performance of the physical machine will heavily influence the test result. I guess that's why it gets timeout in Travis CI. Does anyone have any idea about how to deal with this?

**Comment by zoq** [comment]
We should add the test to the list of long running parallel tests in https://github.com/mlpack/mlpack/blob/master/src/mlpack/tests/CMakeLists.txt

That should speed things up a little bit, but not sure it's enough, can we decrease the amount of worker?

**Comment by ShangtongZhang** [comment]
I can have a try to see what's the minimum amount of the workers.

**Comment by zoq** [comment]

**Comment by ShangtongZhang** [comment]
I think once https://github.com/mlpack/mlpack/pull/1056 is merged, this PR will be ready for review -- it's difficult to break this PR down further.  The failed AppVeyor is due to the failure of downloading `boost 1.60`

**Comment by zoq** [comment]
About the AppVeyor build: https://appveyor.statuspage.io/incidents/m2vdvw39kdk8, I'll restart the build once the issue is solved.

**Comment by zoq** [comment]
Also, thanks for tunning the AsyncLearningTest test. I'm not sure it's feasible to run the test on a regular basis, it looks like it takes about 10 minutes, which is a bummer, since I can't see a good way to test the code without it. But if it takes about 10 minutes in release mode it probably takes a lot more time in debug mode.

**Comment by ShangtongZhang** [comment]
oops, I didn't realize it takes so much time in the server. In my mac, it only takes 2 seconds in debug mode. So it may be hard to tune it further.

**Comment by ShangtongZhang** [comment]
I meet a problem about sharing layers between two `FFN` objects. Say we have two `FFN` instances, i.e. `actor` and `critic`, I want to share some layers, so I have following code:
```
    for (auto& layer : sharedLayers)
    {
      actor.Add(layer);
      critic.Add(layer);
    }
    actor.Add<Linear<>>(featureSize, actionSize);
    actor.Add<LogSoftMax<>>();
    critic.Add<Linear<>>(featureSize, 1);
```
The problem happens when `actor` and `critic` deconstructs
```
FFN<OutputLayerType, InitializationRuleType>::~FFN()
{
  std::for_each(network.begin(), network.end(),
      boost::apply_visitor(deleteVisitor));
}
```
The shared layers will be deleted twice.
The solution I can come up with is to use `shared_ptr` everywhere, i.e. 
```
using LayerTypes = boost::variant<std::shared_ptr<Add<arma::mat, arma::mat>>, ... >
```
with this approach we can also remove the DTOR of `FFN`.
But I assume it will be a huge change. Should I do this? Or Did I just miss some better solution?
Looking forward to any feedback @rcurtin @zoq

**Comment by zoq** [comment]
I see, the only solution I can see right now is to use ``shared_ptr`` as you suggested.

**Comment by ShangtongZhang** [comment]
Ok. I think I can make a separate PR to apply shared_ptr.

**Comment by zoq** [comment]
Great, thanks!

File: src/mlpack/methods/ann/layer/policy_impl.hpp [hunk 7]
```diff
@@ -0,0 +1,63 @@
...
+}
+
+template<typename InputDataType, typename OutputDataType>
+template<typename eT>
+void Policy<InputDataType, OutputDataType>::Backward(
+    const arma::Mat<eT> &&prob, arma::Mat<eT> &&advantage, arma::Mat<eT> &&g)
+{
+  arma::mat adv = arma::repmat(arma::sum(advantage), advantage.n_rows, 1);
+  arma::mat gradientPolicy = advantage - adv % prob;
+  arma::mat logProb = arma::log(prob);
```
zoq: We should use ``trunc_log`` or add a small probability factor to prevent NaNs. Also did you wrote a test for the ``Policy`` layer, without using the Reinforce layer?

ShangtongZhang: Thanks for pointing that out, I will use `trunc_log` instead and add a unit test. BTW, did you find any mistake in math in computing `d(plogp)/dx`?

zoq: It looks good for me, maybe the test can help us to figure out what went wrong.

ShangtongZhang: It works well in the unit test. I computed the ground truth gradient by pytorch
```
    import torch
    from torch.autograd import Variable
    import torch.nn.functional as F
    x = Variable(torch.DoubleTensor([[1, 3, 9]]), requires_grad=True)
    prob = F.softmax(x)
    log_prob = F.log_softmax(x)
    policy_loss = -log_prob.gather(1, Variable(torch.LongTensor([[1]]))) * 2
    kl_loss = (prob * log_prob).sum()
    (policy_loss + 0.1 * kl_loss).backward()
    print x.grad.data.numpy()
```
maybe I need to implement same architecture in pytorch to see whether it works or not.

ShangtongZhang: same architecture works well in pytorch :< there must be something wrong

zoq: Same architecture, means you used the Reinforce layer right? Maybe someone strange happens there.

ShangtongZhang: By same architecture I mean I used non-shared actor critic and same hyper parameters. pytorch doesn't have something like reinforce layer. The `Reinforce` layer I implemented is just meant to be a placeholder in place of layer like `MeanSquredError`, simply passing the `advantage` to `Policy` layer. Did I miss something?

zoq: Ah, I see I thought you used the existing ``ReinforceNormal`` layer. Can you post the pytorch architecture you compared against?

ShangtongZhang: Sure. https://github.com/ShangtongZhang/DeepRL/blob/mlpack/async_worker/actor_critic.py
And the definition of actor network and critic network is here https://github.com/ShangtongZhang/DeepRL/blob/mlpack/main.py, the entrance function is `a3c_cart_pole`

zoq: Okay, thanks, let me take a look.

zoq: Since you used the same hyperparameter, I think we don't have to test different parameter sets.

zoq:
> This PR will only serve for a track of my progress. I will break it down to small incremental PRs to make it easy to review. 

@ShangtongZhang should we close this for now to avoid any confusions?

ShangtongZhang:
Sure I think so.

PR #937, opened 2017-03-14
Fix GMMTest comparisions

### LINKED ISSUE #932: GMMTest/GMMTrainEMOneGaussian fails when component co-variance is close to 0
Parent issue : #922 

Test in `src/mlpack/tests/gmm_test.cpp`.

The test uses `BOOST_REQUIRE_CLOSE` to check the equality of floating point numbers, which is based on Knuth's tolerance predicate. It compares the relative errors of the numbers to a required tolerance %. The problem arises when the numbers are close to 0 where the relative errors can grow arbitrarily large. This leads to the test failing randomly, as is shown by the multiple instance outputs [here](https://gist.github.com/shikharbhardwaj/d757d9ef9dec5424fe3d65800409c9a9), whenever the co-variance of the component is close to 0. 

The same can be spotted on many failed Travis builds, such as [this one](https://travis-ci.org/mlpack/mlpack/builds/210031049).

Relevant discussion on StackOverflow : http://stackoverflow.com/questions/29196130/boost-check-close-fails-even-if-the-two-values-are-below-the-threshold

Possible solutions : 

 * Use [`BOOST_CHECK_SMALL`](http://www.boost.org/doc/libs/1_34_0/libs/test/doc/components/test_tools/reference/BOOST_CHECK_SMALL.html) on the difference between the values of the computed value and correct value. The tolerance here has to be a multiple of `std::numeric_limits<double>::epsilon()`, but I am unaware of the error propagation in the EM algorithm implemented here. Since this method ignores the relative errors, it may cause problems when the values are large.

 * Conditionally check if the value to be tested is close to 0, and use different methods to establish an equality in that case.

shikharbhardwaj:
Fixes #932.

I have changed the interface of `CheckMatrices` in test_tools.hpp to take tolerance values, with a default parameter to avoid changing existing code.

### DIFF
--- a/src/mlpack/tests/gmm_test.cpp
+++ b/src/mlpack/tests/gmm_test.cpp
@@ -111,17 +111,8 @@ BOOST_AUTO_TEST_CASE(GMMTrainEMOneGaussian)
     arma::mat actualCovar = ccov(data, 1 /* biased estimator */);
 
     // Check the model to see that it is correct.
-    BOOST_REQUIRE_CLOSE(gmm.Component(0).Mean()[0], actualMean(0), 1e-5);
-    BOOST_REQUIRE_CLOSE(gmm.Component(0).Mean()[1], actualMean(1), 1e-5);
-
-    BOOST_REQUIRE_CLOSE(gmm.Component(0).Covariance()(0, 0),
-        actualCovar(0, 0), 1e-5);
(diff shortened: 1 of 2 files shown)


rcurtin:
Thanks, let's see what this fixes in the matrix build. :)

PR #939, opened 2017-03-15
Xavier init

kris-singh:
Adds Xavier Uniform and Normal Init method. Also adds He Uniform and Normal init methods. 
Test added for only uniform type.


kris-singh:
@zoq Ready for review

kris-singh:
@zoq  can you have a look at this. Thanks

kris-singh:
Ready for review. Thanks

src/mlpack/methods/ann/init_rules/he_normal.hpp
```diff
@@ -0,0 +1,54 @@
...
+#include <mlpack/prereqs.hpp>
+ 
+namespace mlpack {
+namespace ann /** Artificial Neural Network. */ {
+
+/**
+ * This class is used to initialize weights according to 
+ * the formula N(0, sqrt(2/fanin))
+ */
+class HeNormal
```
zoq: Couldn't we use the GaussianInit method here, to do the same?

src/mlpack/methods/ann/init_rules/he_init.hpp
```diff
@@ -0,0 +1,73 @@
...
+
+/**
+ * Base class for He Intilisation Policy
+ * Refrence 
+ * Kaiming He et al. (2015):
+ * Delving deep into rectifiers: Surpassing human-level performance on
+ * imagenet classification. arXiv preprint arXiv:1502.01852.
+ */
+template<typename InitializerType>
+class HeInit: public InitializerType
```
zoq: Not sure why we have to inherent from the InitializerType here. Couldn't we just do 

```
InitializerType init;
init. Initialize(...);
````

Maybe I missed something?

File: src/mlpack/methods/ann/init_rules/he_uniform.hpp [hunk 3]
```diff
@@ -0,0 +1,54 @@
...
+
+  /**
+   * Initialize the elements of the specified weight matrix.
+   *
+   * @param W Weight matrix to initialize.
+   * @param rows Number of rows.
+   * @param cols Number of columns.
+   */ 
+  template<typename eT>
+  void Initialize(arma::Mat<eT>& W, const size_t rows, const size_t cols, const size_t seed = 21)
```
zoq: Couldn't we use the GaussianInit method here, to do the same?

kris-singh:
@zoq I have made the changes. I do not prefer on using Xavier init here.

zoq:
> I do not prefer on using Gaussian init here, hence i have used RandNormal.

Can you elaborate on that, I think someone could just use the Gaussian Init class to do the same, wouldn't that reduce code duplication?

kris-singh:
I think it's an overkill to create a new object just for doing to random normal initialisation. Also, it would reduce the code duplication. But it would lead to dependence of Xavier Init on Gaussian Init which is logically not justifiable. it should be independent.

zoq:
I see your point, I was thinking that we could reduce code duplication by using some template metaprogramming:

```
class HeUniform {}
class HeNormal {}


template<typename InitializerType>
class XavierInit
{

  template<typename eT, typename InitType = InitializerType>
  void Initialize(arma::Mat<eT>& W,
                  const size_t rows,
                  const size_t cols,
                  const typename std::enable_if_t<std::is_same<
                      InitType, HeUniform>::value>* junk = 0)
  {
    // Use GaussianInit
  }

  template<typename eT, typename InitType = InitializerType>
  void Initialize(arma::Mat<eT>& W,
                  const size_t rows,
                  const size_t cols,
                  const typename std::enable_if_t<std::is_same<
                      InitType, HeNormal>::value>* junk = 0)
  {
    // Use GaussianInit
  }
}


template<typename InitializerType = HeNormal>
using XavierInitNormal = XavierInit<HeNormal>;

template<typename InitializerType = HeUniform>
using XavierInitUniform = XavierInit<HeUniform>;
```

kris-singh:
Okay i get your point. i will make the necessary changes by tonight and resubmit. Thanks for the review.

kris-singh:
I have used templated functions instead of templated class. Please go through solution and let me know. What do you think

src/mlpack/tests/init_rules_test.cpp
```diff
@@ -126,6 +129,35 @@ BOOST_AUTO_TEST_CASE(OivsInitTest)
   BOOST_REQUIRE_EQUAL(1, 1);
 }
 
+BOOST_AUTO_TEST_CASE(XavierInitTest)
+{
+  arma::mat weights;
+  //RandomSeed(21);
+  XavierInit xavierUniform;
+  XavierInit xavierNormal;
+  xavierUniform.Initialize<XavierUniform>(weights, 100, 100);
```
zoq: Since the ``Initialize`` function is used inside e.g. the FFN class there is no easy way to call  `` xavierUniform.Initialize<XavierUniform>`` since, there is no option for a user to provide XavierUniform. An easy solution would be to templatize the ``XavierInit`` class. What do you think?

kris-singh: I did try the solution you suggested previously. The problems were if you use something like this 
template<typename InitializerType>
class XavierInit
{
.....
......
 template<typename eT, typename InitType = InitializerType>
  void Initialize(
....
}
}
This would work fine in GCC, but I think it does not work in [vc++](https://stackoverflow.com/questions/6972368/stdenable-if-to-conditionally-compile-a-member-function).  Second ans comments. Hence that's why i choose this approach. This is equivalent to partial specialisation of functions which i think is not permitted in c++11. Correct me if i am wrong.

rcurtin: We require Visual Studio 2015 minimum, since earlier versions don't support a lot of C++11 features that we need.  The comment that you referenced is about VS2012, so I wouldn't worry about it.

src/mlpack/core/math/shapiro_wilk_test.hpp
```diff
@@ -0,0 +1,108 @@
...
+        boost::math::normal_distribution<>myNormal (mean, sd);
+        return quantile(myNormal, prob);
+};
+template<typename eT>
+Pval Shapiro(arma::Mat<eT>& dist, double alpha=0.05)
+{
+  //All the variable are consitent with refrence provided above
+  std::mt19937 engine;  // Mersenne twister random number engine
+  std::normal_distribution<> distr(0, 1.0);
+  dist.reshape(dist.n_rows*dist.n_cols, 1);
```
rcurtin: Please don't use your own RNG, you can use `math::Random()` and other similar functionality.  The reason for this is that `math::RandomSeed()` sets the seed for all RNGs that we use, which makes it way easier to debug tests.

kris-singh:
okay then i will make the necessary changes and push them.

kris-singh:
Changes Made. Ready for merge.

src/mlpack/tests/math_test.cpp
```diff
@@ -583,4 +584,21 @@ BOOST_AUTO_TEST_CASE(RangeContainsRange)
   BOOST_REQUIRE_EQUAL(b.Contains(a), true);
 }
 
+/**
+ * Test the shapiro wilk normality test.
+ * The values have been taken from https://tinyurl.com/q9gx9m2
```
rcurtin: Could you avoid a tinyurl here please? Better to link to the original source.

src/mlpack/tests/init_rules_test.cpp
```diff
@@ -126,6 +129,35 @@ BOOST_AUTO_TEST_CASE(OivsInitTest)
...
+  arma::mat weights;
+  //RandomSeed(21);
+  XavierInit<XavierUniform> xavierUniform;
+  XavierInit<XavierNormal> xavierNormal;
+  xavierUniform.Initialize(weights, 100, 100);
+  bool b = arma::all(arma::vectorise(weights) >= -(1e-2) || arma::vectorise(weights) <= (1e-2));
+  BOOST_REQUIRE_EQUAL(b, 1);
+  xavierNormal.Initialize(weights, 100, 100);
+  //std::cout << weights << std::endl;
+  auto ret = Shapiro(weights, 0.05);
```
rcurtin: Honestly I don't see the need for the Shapiro Wilk test. I think that if you really wanted to test that the distribution is correct you can use the `GaussianDistribution` class and use `Train()` and ensure that the mean and variance are close to what is expected.

I realize that the Shapiro Wilk test might be a 'better' test to use here, but my concern is that the test code is itself less thoroughly tested than the `GaussianDistribution` test.

src/mlpack/tests/init_rules_test.cpp
```diff
@@ -126,6 +129,35 @@ BOOST_AUTO_TEST_CASE(OivsInitTest)
...
+  arma::mat weights;
+  //RandomSeed(21);
+  HeInit<HeUniform> Uniform;
+  HeInit<HeNormal> Normal;
+  Uniform.Initialize(weights, 100, 100);
+  bool b = arma::all(arma::vectorise(weights) >= -(1e-2) || arma::vectorise(weights) <= (1e-2));
+  BOOST_REQUIRE_EQUAL(b, 1);
+  Normal.Initialize(weights, 100, 100);
+  auto ret = Shapiro(weights, 0.05);
+  BOOST_REQUIRE_EQUAL(ret.accept, true);
```
rcurtin: Could you split these into two tests please? Better to test each thing separately.

src/mlpack/methods/ann/init_rules/xavier_init.hpp
```diff
@@ -0,0 +1,94 @@
...
+using namespace mlpack::math;
+
+ 
+namespace mlpack {
+namespace ann /** Artificial Neural Network. */ {
+
+class XavierNormal
+{};
+class XavierUniform
+{};
```
rcurtin: There's so little overlap in the code here that I think maybe it is better to simply have a XavierUniformInit and XavierNormalInit class. (The same applies for HeInit.)

src/mlpack/methods/ann/init_rules/xavier_init.hpp
```diff
@@ -0,0 +1,94 @@
...
+  scalingFactor(scalingFactor)
+  {}
+
+  template<typename eT, typename Init = InitializerType>
+  void Initialize(arma::Mat<eT>& W, 
+    const size_t rows, 
+    const size_t cols, 
+    std::enable_if_t<std::is_same<Init, XavierNormal>::value>* = 0)
+  {
+    double var = sqrt (2 / ((double)(rows) + (double)(cols)));
```
rcurtin: Why not also include `scalingFactor` in this calculation?

src/mlpack/methods/ann/init_rules/xavier_init.hpp
```diff
@@ -0,0 +1,94 @@
...
+
+  template<typename eT, typename Init = InitializerType>
+  void Initialize(arma::Mat<eT>& W, 
+    const size_t rows, 
+    const size_t cols,
+    std::enable_if_t<std::is_same<Init, XavierUniform>::value>* = 0)
+  {
+    double var = sqrt(12 / static_cast<double>(rows + cols));
+    W = arma::zeros(rows, cols);
+    W.imbue( [&]() {return Random(-var, var); });
```
rcurtin: No need to call `zeros()` before calling `imbue()`.

src/mlpack/tests/init_rules_test.cpp
kris-singh: Maybe we could discuss what other tests we would need for the Shapiro-Wilk test. Because I have used the train method from GaussianDistribution before but it does converge where I mean even to come close to original mean and variance util a high amount of iterations

rcurtin: Can you tell me how you are using `GaussianDistribution::Train()`?  That should have no problem converging in this instance.

kris-singh: Checkout the test of Gaussian Init.

rcurtin:
It was pointed out in #1510 that the Xavier initialization is merged as of #1272 so we can go ahead and close this now. :+1:

PR #942, opened 2017-03-15
AdaGrad Update Policy

abhinavmoudgil95:
__Implementation of AdaGrad optimizer__ 
* Added AdaGrad update policy class to SGD. 
* All tests passed. 
* Detailed discussion for reducing code duplication in #897. 
* Reference: [Adaptive Subgradient Methods for Online Learning and Stochastic Optimization](http://jmlr.org/papers/v12/duchi11a.html).


abhinavmoudgil95:
@rcurtin @zoq PR is ready to be merged. Please suggest edits, if any.

abhinavmoudgil95:
@rcurtin @zoq PR is ready to be merged. Please review and suggest edits, if any.

src/mlpack/core/optimizers/ada_grad/ada_grad.hpp
```diff
@@ -0,0 +1,172 @@
...
+  //! The tolerance for termination.
+  double tolerance;
+
+  //! Controls whether or not the individual functions are shuffled when
+  //! iterating.
+  bool shuffle;
+
+  //! Stochastic Gradient Descent object with AdaGrad policy.
+  SGD<DecomposableFunctionType, AdaGradUpdate>
+  optimizer(DecomposableFunctionType&,
```
zoq: I'm wondering if this works for you? Shouldn't that be ``SGD<DecomposableFunctionType, AdaGradUpdate> optimizer;``?

src/mlpack/tests/ada_grad_test.cpp
```diff
@@ -0,0 +1,107 @@
...
+ *
+ * Test file for AdaGrad (stochastic gradient descent with AdaGrad updates).
+ *
+ * mlpack is free software; you may redistribute it and/or modify it under the
+ * terms of the 3-clause BSD license.  You should have received a copy of the
+ * 3-clause BSD license along with mlpack.  If not, see
+ * http://www.opensource.org/licenses/BSD-3-Clause for more information.
+ */
+#include <mlpack/core.hpp>
+#include <mlpack/core/optimizers/sgd/sgd.hpp>
```
zoq: I think we can remove the header file here. We include ``SGDTestFunction`` with ``#include <mlpack/core/optimizers/sgd/test_function.hpp>``.

src/mlpack/core/optimizers/sgd/update_policies/CMakeLists.txt
```diff
@@ -1,6 +1,7 @@
 set(SOURCES
   vanilla_update.hpp
   momentum_update.hpp
+  ada_grad_update.hpp
```
zoq: I looks like we moved the file from ``/core/optimizers/sgd/update_policies/``to the adagrad folder, so we should remove the entry from the sgd policies CMakeLists.txt file.

src/mlpack/core/optimizers/ada_grad/CMakeLists.txt
```diff
@@ -0,0 +1,10 @@
+set(SOURCES
+  ada_grad.hpp
```
zoq: We should include the update policy here.

abhinavmoudgil95:
@zoq Thanks a lot for the review. Sorry, I got busy in my mid term examinations, so couldn't update the PR. I am making changes now.

src/mlpack/core/optimizers/ada_grad/ada_grad.hpp
abhinavmoudgil95: Yes, I have made the change. I was just testing a few things to resolve build errors. 

Actually, I am not able to build locally as CMake gets stuck after executing few steps when I run command `cmake ..` (details [here](https://paste.debian.net/hidden/e457f447/)). Could you give an idea why it might be happening?

abhinavmoudgil95: Yes, I have made the change. I was just testing a few things to resolve build errors. 

Actually, I am not able to build locally now as CMake gets stuck after executing few steps when I run command `cmake ..` (details [here](https://paste.debian.net/hidden/e457f447/)). Earlier when I opened up AdaGrad PR, I was able to build and test. Could you give an idea why it might be happening?

src/mlpack/core/optimizers/ada_grad/CMakeLists.txt
abhinavmoudgil95: Fixed.

src/mlpack/core/optimizers/sgd/update_policies/CMakeLists.txt
abhinavmoudgil95: Fixed.

abhinavmoudgil95: Done.

src/mlpack/tests/ada_grad_test.cpp
abhinavmoudgil95: Done.

src/mlpack/core/optimizers/ada_grad/ada_grad.hpp
rcurtin: The link doesn't work, so I can't provide any help.  Are you still having the issue?

src/mlpack/core/optimizers/ada_grad/ada_grad.hpp
```diff
@@ -0,0 +1,164 @@
...
+
+  //! The maximum number of allowed iterations.
+  size_t maxIterations;
+
+  //! The tolerance for termination.
+  double tolerance;
+
+  //! Controls whether or not the individual functions are shuffled when
+  //! iterating.
+  bool shuffle;
```
rcurtin: Why do all these parameters need to be held in the class if they are not used and are also held in `optimizer`?  I think we could remove them.

src/mlpack/core/optimizers/ada_grad/ada_grad.hpp
```diff
@@ -0,0 +1,164 @@
...
+
+  //! The AdaGradUpdate policy object.
+  AdaGradUpdate adagradUpdate;
```
rcurtin: The update policy is also held in `optimizer`, so I think we could remove this member too?

zoq: I think it's need for ``double Epsilon() const { return adagradUpdate.Epsilon(); }`` but that shouldn't have an effect on the optimizer class since it uses: 

```
const UpdatePolicy updatePolicy = UpdatePolicy());
```

src/mlpack/core/optimizers/ada_grad/ada_grad.hpp
abhinavmoudgil95: Yes, I think we can remove all the parameters as they are held by optimizer. Just not sure about `adagradUpdate`.

src/mlpack/core/optimizers/ada_grad/ada_grad.hpp
abhinavmoudgil95: Can we change the `Epsilon()` function to something like below? I am not sure about this but I feel that `UpdatePolicy` is also part of SGD, so it should work. 
```c++
 double Epsilon() const { return optimizer.AdaGradUpdate.Epsilon(); }
```

Accordingly, I would change the constructor: 

```c++
AdaGrad<DecomposableFunctionType>::AdaGrad(DecomposableFunctionType& function,
                                           const double stepSize,
                                           const double epsilon,
                                           const size_t maxIterations,
                                           const double tolerance,
                                           const bool shuffle)
{
  AdaGradUpdate adagradUpdate(epsilon);
  optimizer = SGD<DecomposableFunctionType, AdaGradUpdate>(function, stepSize, 
         maxIterations, tolerance, shuffle, adagradUpdate);
}
```

rcurtin: Yes, I think it would be `optimizer.UpdatePolicy().Epsilon()` but that should work and would be a good way to do it without duplicating any options anywhere.

src/mlpack/core/optimizers/ada_grad/ada_grad.hpp
abhinavmoudgil95: Yes, issue is still there. Sorry, the link got expired. This is the updated [link](http://pastebin.com/tGFjGdmk). I work on a GPU server where I don't have root access. Hence, I updated the `cmake` command accordingly. It remains stuck after the last line on console:
` Found Doxygen: /home/abhinav.moudgil/doxygen_installed/bin/doxygen (found version "1.8.14") `.

src/mlpack/core/optimizers/ada_grad/ada_grad.hpp
abhinavmoudgil95: Ah thanks, I will make the changes asap.

zoq: ``optimizer.UpdatePolicy().Epsilon()`` That only works if SGD would implement a function that returns the UpdatePolicy. So I guess we either have to implement that function or we could change the constructor to use a reference for the policy.

abhinavmoudgil95: Keeping consistency in mind, I think implementing a function `UpdatePolicy()` just like other variables would be better?

rcurtin: I agree, I am sure that `UpdatePolicy()` will come in useful in other situations too.

abhinavmoudgil95:
@rcurtin @zoq I am facing some issues in removing `adagradUpdate` parameter from AdaGrad class. 

We are declaring the SGD optimizer in AdaGrad class without any arguments like `SGD<DecomposableFunctionType, AdaGradUpdate> optimizer;` whereas the default constructor of SGD has 6 arguments. To counter this, we must initialize the `optimizer` in AdaGrad constructor with initializer list like below otherwise it doesn't work as default constructor expects 6 arguments and we are passing 0 in the declaration.
```c++
AdaGrad<DecomposableFunctionType>::AdaGrad(DecomposableFunctionType& function,
                                           const double stepSize,
                                           const double epsilon,
                                           const size_t maxIterations,
                                           const double tolerance,
                                           const bool shuffle):
adagradUpdate(epsilon),
optimizer(function, stepSize ...) // requires AdaGradUpdate object
{/* Nothing to do */}
```
But initializing `optimizer` in intializer list like above requires the `AdaGradUpdate` object which we don't have.

We can resolve this by adding a constructor without any arguments `SGD();` to SGD class. Then I can reinitialize my `optimizer` object in AdaGrad constructor as I suggested before:
```c++
AdaGrad<DecomposableFunctionType>::AdaGrad(DecomposableFunctionType& function,
                                           const double stepSize,
                                           const double epsilon,
                                           const size_t maxIterations,
                                           const double tolerance,
                                           const bool shuffle)
{
  AdaGradUpdate adagradUpdate(epsilon);
  optimizer = SGD<DecomposableFunctionType, AdaGradUpdate>(function, stepSize, 
         maxIterations, tolerance, shuffle, adagradUpdate);
}
``` 
Please suggest if there is any other better way to solve this problem.

abhinavmoudgil95:
* Removed all the parameters from AdaGrad class except `SGD` and `AdaGradUpdate` object. 
* Added `GetUpdatePolicy()` method to SGD to access `UpdatePolicy` object. Then we can access `epsilon` directly from optimizer rather than accessing it from `AdaGradUpdate` object. 
* Had to name `GetUpdatePolicy()` rather than `UpdatePolicy()` otherwise template parameter gets shadowed and it doesn't work. Also, it can't be overloaded `UpdatePolicy& GetUpdatePolicy()` like other parameter methods.

src/mlpack/core/optimizers/sgd/sgd.hpp
```diff
@@ -135,6 +136,9 @@ class SGD
   //! Modify whether or not the individual functions are shuffled.
   bool& Shuffle() { return shuffle; }
 
+  //! Get the update policy. 
+  UpdatePolicy GetUpdatePolicy() const { return updatePolicy; }
```
zoq: Instead of using ``GetUpdatePolicy`` because of the template parameter shadowed issue, you can just rename the template parameter to ``updatePolicyType`` that way we are still consistent with the rest of the codebase.

zoq: Also please add ``UpdatePolicy& UpdatePolicy() { return updatePolicy; }``, so that we can modify the policy.

abhinavmoudgil95: That's a good idea. So, I should make the following updates in SGD class, right? 
* `updatePolicy` to `updatePolicyType`
* `typename UpdatePolicy = VanillaUpdate` to `typename UpdatePolicyType = VanillaUpdate`
* `GetUpdatePolicy()` to `UpdatePolicy()`

zoq: yes, but please only rename the type not the parameter so:

```
const UpdatePolicy updatePolicy = UpdatePolicy());
```

would become 

```
const UpdatePolicyType updatePolicy = UpdatePolicyType());
```

and not

```
const UpdatePolicyType updatePolicyType = UpdatePolicyType());
```

and also don't forget to add:

```
UpdatePolicyType& UpdatePolicy() { return updatePolicy; }
```

to the sgd class.

abhinavmoudgil95: Sure, thanks a lot.

src/mlpack/core/optimizers/ada_grad/ada_grad.hpp
```diff
@@ -0,0 +1,145 @@
...
+  double& Tolerance() { return optimizer.Tolerance(); }
+
+  //! Get whether or not the individual functions are shuffled.
+  bool Shuffle() const { return optimizer.Shuffle(); }
+  //! Modify whether or not the individual functions are shuffled.
+  bool& Shuffle() { return optimizer.Shuffle(); }
+
+ private:
+  //! The AdaGradUpdate policy object.
+  AdaGradUpdate adagradUpdate;
```
rcurtin: This member isn't used and I think we should remove it---there is already one of these held inside `optimizer`.

src/mlpack/core/optimizers/ada_grad/ada_grad_impl.hpp
```diff
@@ -0,0 +1,39 @@
...
+                                           const size_t maxIterations,
+                                           const double tolerance,
+                                           const bool shuffle) :
+    adagradUpdate(epsilon),
+    optimizer(function,
+              stepSize,
+              maxIterations,
+              tolerance,
+              shuffle,
+              adagradUpdate)
```
rcurtin: You can replace `adagradUpdate` with `AdagradUpdate(epsilon)`.

rcurtin:
Looks good to me, although since Travis and AppVeyor refuse to build since there are merge conflicts, it should be tested during the merge process.  @zoq: I can do that if you want and merge this in, just let me know.

rcurtin:
Also, Abhinav, thanks for the hard work and patience with this. :)  I know there were lots of changes and it ended up looking way different than the original version.

zoq:
Let's check the code one more time :)

abhinavmoudgil95:
Thanks a lot, Ryan and Marcus for guiding me so well. Honestly, it has been a great learning experience for me. :)

zoq:

abhinavmoudgil95:
Glad to see the PR merged. Really excited to contribute more!

PR #944, opened 2017-03-15
Implementation of HDBSCAN.

s1998:


src/mlpack/methods/hdbscan/hdbscan_impl.hpp
```diff
@@ -0,0 +1,755 @@
...
+        lambda = std::numeric_limits<double>::infinity();
+
+      if(leftCount >= minClusterSize && rightCount >= minClusterSize)
+      {
+        relabel[leftChild] = nextLabel;
+        nextLabel++;
+        result(resultCounter,0) = relabel[currNode];
+        result(resultCounter,1) = relabel[leftChild];
+        result(resultCounter,2) = lambda;
+        result(resultCounter,3) = leftCount;
```
s1998: @rcurtin  Is there a shorter way to update entire row at once ? (I looked through the documentation, I am missing it somehow) I think it will reduce the code length a lot !!

src/mlpack/methods/hdbscan/hdbscan_impl.hpp
```diff
@@ -0,0 +1,755 @@
...
+  arma::Mat<size_t> neighbors;
+  arma::mat distances;
+  neighborSearch.Train(data);
+  neighborSearch.Search(minPoints, neighbors, distances);
+  arma::mat dataWithDcore = arma::conv_to<arma::mat>::from(data);
+  dataWithDcore.resize(dataWithDcore.n_rows+1, dataWithDcore.n_cols); 
+  dataWithDcore.row(dataWithDcore.n_rows-1) = distances.row(distances.n_rows-1);
+ 
+  //computing the mst of the given data and sort the obtained result
+  emst::DualTreeBoruvka<metric::HdbscanDistance, arma::mat, TreeType> dtb(dataWithDcore, true, metric);
```
s1998: I tried this with value false, but it fails (http://pastebin.com/1Ke1gRLe). So for the time being, I am using the naive implementation.

src/mlpack/core/metrics/lmetric.hpp
```diff
@@ -116,6 +116,10 @@ typedef LMetric<2, true> EuclideanDistance;
  */
 typedef LMetric<INT_MAX, false> ChebyshevDistance;
 
+/**
+ * The HDBSCAN metric.
+ */
+typedef LMetric<INT_MAX, true> HdbscanDistance;
```
s1998: Instead of doing this, if I define the metric in namespace HDBSCAN{} , the user will not be able to select the treeType because of "static_assert(meta::IsLMetric<MetricType>::Value == true" in hrectbound.hpp .

src/mlpack/methods/hdbscan/hdbscan_impl.hpp
rcurtin: Is `metric::HdbscanDistance` an actual metric?  If it doesn't satisfy the triangle inequality then it will not work with trees.

src/mlpack/methods/hdbscan/hdbscan_impl.hpp
rcurtin: Armadillo is column major, so you should really be updating a column at once, not a row at once.  In either case since you don't have all four values available in an `arma::vec` you probably have to update this manually.

src/mlpack/tests/hdbscan_test.cpp
```diff
@@ -0,0 +1,74 @@
...
+  {
+    BOOST_REQUIRE((assignments(i)==0) || (assignments(i)==SIZE_MAX));
+  }
+}
+
+BOOST_AUTO_TEST_CASE(noiseTest)
+{
+  arma::mat points;
+  if (!data::Load("circle.csv", points))
+    BOOST_FAIL("Cannot load test dataset circle.csv!");
```
rcurtin: Is the circle dataset just points on the 2-d circle?  If so can you generate this data instead?

src/mlpack/core/metrics/lmetric.hpp
rcurtin: That's there for a reason... the BinarySpaceTree type *requires* the distance to be `LMetric` or some variant thereof because the `MinDistance()` and `MaxDistance()` functions are specialized.  So overloading the distance here will cause the bounds to fail and therefore the results you get are wrong (this is why you have to use naive mode---but you should still check whether it follows the triangle inequality).  You will need to use a different type of tree (like `BallTree` or something else) by default for this to work.

src/mlpack/methods/hdbscan/hdbscan_impl.hpp
s1998: I think it satisfies the triangle inequality because : 
`hdbscanDistance(A, B) = max(dcore(A), dcore(B), euclideanDistance(A, B))`
`hdbscanDistance(A, B) >= euclideanDistance(A, B)`

Lets consider triangle ABC and side AB : 
if `hdbscanDistance(A, B) = euclideanDistance(A, B)`
then `hdbscanDistance(A, C) + hdbscanDistance(B, C) >=`
`euclideanDistance(A, C) + euclideanDistance(B,C) >= `
`euclideanDistance(A, B) = hdbscanDistance(A, B)`
other wise `hdbscanDistance(A, B) = dcore(A) or dcore(B).`
Lets say `hdbscanDistance(A, B) = dcore(A) `
then `hdbscanDistance(A, C) + hdbscanDistance(B, C) >=`
`hdbscanDistance(B, C) + dcore(A) >= dcore(A) = hdbscanDistance(A, B)`
We can argue similarly if `hdbscanDistance(A, B) = dcore(B) `
Hope, I am not mistaken.

s1998:
@rcurtin. I have tried to fix the style changes and test generates the points on the 2d circle.
Should i set `TreeType` as `BallTree` for default and `naive` as `false` ? In the documentation I will add that to use any other `TreeType` use `naive` as `true` ?

rcurtin:
Yes, we should be using `BallTree` or `CoverTree` as the default, and `naive` should be `false`.  Using other TreeTypes doesn't mean `naive` should be set to `false`; if an invalid tree type is used, this will cause a compilation error.

rcurtin:
In addition the `HdbscanDistance` should not be an invalid specialization of `LMetric`, it should be its own type.

s1998:
@rcurtin  Defined `HdbscanMetric` inside the metric folder and set naive as `false`.

rcurtin:
Thanks for making the change.  I have some basic high level comments before we get too deep into details:

 * Can you go through the style guide and make sure that your code matches it? You can find that as a wiki page.  Max line length 80 characters, capitalized method names, horizontal whitespace, etc.

 * There are some inefficiencies with how Armadillo matrices are being copied but as we get deeper into the code I can help with that.

 * I'd really like to add more tests. Right now there are only two tests which make sure that faraway outliers are classified as noise and that HDBSCAN can recover one cluster.  We should add more difficult tests like those done for DBSCAN and we should also add some tests that ensure HDBSCAN works better than DBSCAN where it should (some paper review will probably be needed to determine that).

Thanks for your hard work on this. I think it will be nice when it is ready. :)

s1998:
I fixed the style issues. As you had said earlier, that there should be tests that prove HDBSCAN outperform DBSCAN, I actually tried different datasets, but for each one of them I was able to tune DBSCAN's parameters to get results similar to HDBSCAN, any advice on this or how to come up with tests ? I actually looked the paper, but it doesn't give much clue about it :(
The paper contains tests on some datasets and has compared the algorithm's output with other algorithms (OPTICS and AUTO-HDS) and has comparisons based on other metrics(%coverage, ARI and FScore).

PR #945, opened 2017-03-17
Named FFN

kris-singh:
This PR adds functionality of adding Layer Names to the Layer. This is really helpful in any layer by name in large networks.
typeid(object).name() this would not work as all linear layer types. would return the same thing.

### DIFF
--- a/src/mlpack/methods/ann/ffn.hpp
+++ b/src/mlpack/methods/ann/ffn.hpp
@@ -160,12 +160,38 @@ class FFN
   template <class LayerType, class... Args>
   void Add(Args... args) { network.push_back(new LayerType(args...)); }
 
+  template <class LayerType, class... Args>
+  void AddNamed(std::string name, Args... args)
+  { 
+    network.push_back(new LayerType(args...));
+    // should raise error if duplicate
+    dictionary.insert(std::make_pair(name, new LayerType(args...)));
+  }
+
   /*
    * Add a new module to the model.
    *
    * @param layer The Layer to be added to the model.
    */
-  void Add(LayerTypes layer) { network.push_back(layer); }
+  void Add(LayerTypes layer, std::string name="")
+  { 
+    network.push_back(layer);
+    if(std::is_empty<std::string>(name))
+      dictionary["Layer" + network.size()] = layer;
+    else
+    {
+      // should raise error if duplicate
+      dictionary.insert(std::make_pair(name, layer));
+    }
+
+  }
+
+  std::vector<LayerTypes> Model(){return network;}
+
+  LayerTypes& getByName(std::string name)
+  {
+      return dictionary[name];
+  }
 
   //! Return the number of separable functions (the number of predictor points).
   size_t NumFunctions() const { return numFunctions; }
@@ -292,6 +318,9 @@ private:
 
   //! Locally-stored gradient parameter.
   arma::mat gradient;
+
+  //! Locally-stored dictionary.
+  std::map<std::string, LayerTypes> dictionary;
 }; // class FFN
 
 } // namespace ann
--- a/src/mlpack/tests/feedforward_network_test.cpp
+++ b/src/mlpack/tests/feedforward_network_test.cpp
@@ -65,10 +65,8 @@ void BuildVanillaNetwork(MatType& trainData,
   model.Add<Linear<> >(trainData.n_rows, hiddenLayerSize);
   model.Add<SigmoidLayer<> >();
   model.Add<Linear<> >(hiddenLayerSize, outputSize);
-  model.AddNamed<LogSoftMax<> >("output");
-
-  std::cout << model.getByName("output") << std::endl;
-
+  model.Add<LogSoftMax<> >();
+  
   RMSprop<decltype(model)> opt(model, 0.01, 0.88, 1e-8,
       maxEpochs * trainData.n_cols, -1);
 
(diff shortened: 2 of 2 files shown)


src/mlpack/methods/ann/ffn.hpp
```diff
@@ -292,6 +318,9 @@ class FFN
 
   //! Locally-stored gradient parameter.
   arma::mat gradient;
+
+  //! Locally-stored dictionary.
+  std::map<std::string, LayerTypes> dictionary;
```
rcurtin: This ends up storing two copies of the network.  If anything you should be using a reference to the layers here, not the direct layers themselves.  I suspect also that the code will not work if you build the network, then train it, then try to access a layer by name, since the extraneous copy does not get trained.

src/mlpack/methods/ann/ffn.hpp
```diff
@@ -160,12 +160,38 @@ class FFN
...
+    {
+      // should raise error if duplicate
+      dictionary.insert(std::make_pair(name, layer));
+    }
+
+  }
+
+  std::vector<LayerTypes> Model(){return network;}
+
+  LayerTypes& getByName(std::string name)
```
rcurtin: This doesn't follow the mlpack naming conventions; please take a look at the style guide.

src/mlpack/tests/feedforward_network_test.cpp
```diff
@@ -410,4 +410,14 @@ BOOST_AUTO_TEST_CASE(DropConnectNetworkTest)
...
 }
 
+BOOST_AUTO_TEST_CASE(NamedNetworkTest)
+{
+  FFN<NegativeLogLikelihood<> > model;
+  model.AddNamed<LinearNoBias<> >("input", 4, 10);
+  model.Add<SigmoidLayer<> >();
+  model.Add<LinearNoBias<> >(10, 2);
+  model.Add<LogSoftMax<> >();
+  std::cout << boost::apply_visitor(WeightSizeVisitor(), model.getByName("input") ) << std::endl;
```
rcurtin: This test doesn't actually test anything.  What are you trying to test?  You should be using the `BOOST_REQUIRE_EQUAL()` or similar tests to ensure that the functionality is correct here.

kris-singh:
Thanks @rcurtin I will make necessary the changes.

zoq:
Thanks @kris-singh for taking the time to look into this issue.

I like the idea, but I'm not sure this is the right way to do this. The main problem I see is that right now this only works for first-level models, e.g.  no container is used (split, concat layers). I think, if we provide such a feature we should make sure it shows the full model hearchiy.

Example:

```
                |------------>layer 1---->layer 3 -------------->|
model ---> concat layer                                          |---merge layer ---> output layer
                |------------>layer 2----> layer 4 ---> layer 5->|
```

would only print the concat layer, because all the other layers aren't held inside the FFN class. This is just a simple example but there are a couple more take a look at the LSTM layer which internally uses a bunch of layers for the layer construction. The output I would expect is something like:

```
model:
  concat layer
  |
  |-----layer1
  |       |
  |     layer 3
  |
  |-----layer 2
  |        |
  |     layer 4
  |        |
  |      layer 5
  |
  merger layer
  |
  output layer
```

Another problem I see is we have to implement the feature for each class, since a user might ask why can I print the model structure for class X but not for class Y. I'm not sure there is a simple solution, maybe my expectations aren't reasonable, eager to get some feedback on this.

kris-singh:
@zoq I do agree the that we should implement this also in the concat and split layer(container layers). I firstly wanted to add them to the ffn network because most of the large models are in general CNN. Which would benefit from the use of this functionality. Hence i went with the idea of doing it first for ffn and then possibly extend this to rnn. I did not understand the last point though could you elaborate a bit more. "Implementing the functionality for each class ......."

zoq:
> I did not understand the last point though could you elaborate a bit more. "Implementing the functionality for each class ......."

That we have to implement the print feature not only for the FNN class but also for the Concat, RNN, Merge classes. And every time someone implements a new layer/container he has to reimplement the feature. I'd like to keep the functions someone has to implement low, I'm not sure that in this particular case there is an easy solution, I guess there is one if you reintroduce the toString function, but that has some drawback too.

kris-singh:
Yes, I think if somebody does implement the container layers. The only thing we would need to add is the dictionary (key -> name, value -> layer) this is only 3-4 lines of code. Like we have for the ffn model.
This doesn't have to be done for every layer though only layers that contain more than one layer(concat layer). I think i would go ahead with idea and implment it for concat, sequential and merge layers. That would be good enough for time being and later we could extend this to rnn and other layers. @zoq  what do you think

kris-singh:
Should i go ahead @zoq.

zoq:
My main concern is that using something like a map and user-defined names isn't the optimal way. I think we can come up with something that automatically figures out some meaningful way to print the models and also prints the exact model hierarchy.

Maybe we can exploit the serialization feature somehow, it already saves model/layer information and is able to represent it in a meaningful way.

kris-singh:
Maybe....I actually was thinking about other libraries that give this features for the users to name a layer for identification.

rcurtin:
I would caution against any kind of automatic printing method or anything.  Having each layer have its own name for access is one thing, but once you say "okay we can provide a method to automatically print the network", it's not going to work for all networks: for very very large networks, what will be printed is a massive amount of output, and for very small networks, the user may actually want more output.  This is why I removed the `ToString()` methods that used to exist: it was a lot to implement for every mlpack class, and the output wasn't actually often useful to a user.

rcurtin:
I'm going to go ahead and close this one for inactivity.  We can revisit it again if needed.

Issue #722, opened 2016-07-12
Reduce compile time by removing unnecessary headers


rcurtin:
The compile time for mlpack is very long compared to other libraries of the same size.  There are many reasons for this; Armadillo (which is basically header-only) is one; boost is another; the template metaprogramming inside of mlpack is yet another.  But there are some ways that we can reduce compile time.

 * Some of the headers in `src/mlpack/core/prereqs.hpp` are not used in all parts of the project and can be removed (or in some cases, included only where they are necessary).
 * Where possible, we can move header files into .cpp files (so they are not included by everything).  Some examples here might be the boost math headers.
 * We may be able to use `extern template` definitions for some of the various parts mlpack provides, specifying to common types, e.g., `extern template AllkNN<arma::mat, EuclideanDistance, KDTree>` and then having that specialization declared in some .cpp file... or something like this.

It'll be easy to check if you've reduced compile times when Travis builds a PR; right now Travis takes about 31 minutes for a complete build, and AppVeyor takes about 45 minutes.  I doubt we can get order-of-magnitude compile speedups or anything, but we can probably help out incrementally.

anshuman23:
Hi Ryan, I'd like to try my hand at this. I just have a trivial question, and I hope it isn't too kiddish:
- For the last point, do you mean that I should look for similar/repetitive definitions in mlpack and hence implement ```extern template``` definitions for them where I find them repeating? 

As for the first 2 points, I've been looking at the source code and I'll try my best to redistribute the headers so as to minimize compile time.

rcurtin:
Great, I look forward to incorporating some changes to speed things up. :)

You're right about the last point, that is indeed what I meant.  Another nice thing is that if we are providing instantiated templates for common mlpack objects, then users who compile things against mlpack downstream should have faster compile times.

anshuman23:
Alright, thank you! :D
Also, if I have to check for a reduction in compile time, would I have to compile from source every time to check?

rcurtin:
Yeah, if you wanted to check, you would need to `make clean` then `make` (or `make -jX` if you want to build with X cores).  Alternately, you could open a PR and push changes to it, and Travis will build it, but Travis will take about 30-45 minutes per build and there will be a delay before the build starts, so that may not be the most efficient way to do it.

sbrodehl:
[...]

PR #946, opened 2017-03-18
Removed Unncessary Headers - PR for issue #722

pvskand:
Referencing issue #722 , I have removed the headers in the following files:

- det_main.cpp

- dt_utils.hpp

- edge_pair.hpp

- emst_main.cpp

- kernel_pca.hpp

- kernel_pca_main.cpp

- kernel_pca_main.cpp

- kernel_rules/naive_method.hpp

- kernel_rules/nystroem_method.hpp

I have ran the tests on their respective modules and found that they have passed the tests. Please do tell me if I have done anything wrong!

### DIFF
--- a/src/mlpack/methods/det/det_main.cpp
+++ b/src/mlpack/methods/det/det_main.cpp
@@ -10,7 +10,7 @@
  * http://www.opensource.org/licenses/BSD-3-Clause for more information.
  */
 #include <mlpack/prereqs.hpp>
-#include <mlpack/core/util/cli.hpp>
+
 #include "dt_utils.hpp"
 
 using namespace mlpack;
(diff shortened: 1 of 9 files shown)


rcurtin:
Unfortunately we really can't remove these headers.  Although the library still compiles, the files that you've modified now can't be included by a user.  So with these changes, if I write the code:

```
#include <mlpack/methods/kernel_pca/kernel_pca.hpp>
```

this will fail.  I would have to first include `mlpack/prereqs.hpp` or `mlpack/core.hpp`.  This isn't desirable and could break user code, so we shouldn't merge these changes.

For what it's worth, the extra inclusion of a header that has already been included shouldn't be a problem---since the header guard macro is already defined, the preprocessor will not include the file multiple times.  I think that the difference in parsing time from including a file multiple times will be negligible.

PR #947, opened 2017-03-18
make decision tree support weighted traning

MikeLing:


src/mlpack/methods/decision_tree/decision_tree.hpp
```diff
@@ -129,13 +170,15 @@ class DecisionTree :
...
+   * @param weights Weights of all the labels
    * @param minimumLeafSize Minimum number of points in each leaf node.
    */
-  template<typename MatType>
+  template<bool UseWeights, typename MatType, typename WeightVecType>
   void Train(const MatType& data,
              const data::DatasetInfo& datasetInfo,
              const arma::Row<size_t>& labels,
              const size_t numClasses,
+             const WeightVecType& weights,
```
rcurtin: Should we restrict to `arma::rowvec` here also?

src/mlpack/methods/decision_tree/gini_gain.hpp
```diff
@@ -33,26 +33,63 @@ class GiniGain
...
+      double accWeights = 0.0;
+
+      // the weight must have the same number of labels
+      if (weights.n_elem != labels.n_elem)
+      {
+        std::ostringstream oss;
+        oss << "the weights' length is' " << weights.n_elem << " but we have  " 
+            << labels.n_elem << " labels, but they must be equal";
+        throw std::invalid_argument(oss.str());
+      }
```
rcurtin: Given that the `Evaluate()` function will be called so many times, I am fine with removing this check for speed (though it will be a minor boost if any) and simply writing in the documentation that the weight length must be the same as the label length (or the number of points).  The user is not likely to use this function directly anyway.  We can perform this check at a higher level in the code.

src/mlpack/tests/decision_tree_test.cpp
```diff
@@ -355,6 +592,26 @@ BOOST_AUTO_TEST_CASE(BasicConstructionTest)
...
+  weights.ones();
+
+  for (size_t i = 0; i < 1000; ++i)
+    labels[i] = i % 3; // 3 classes.
+
+  // Use default parameters.
+  DecisionTree<> d(dataset, labels, 3, weights, 50);
+
+  // Now require that we have some children.
+  BOOST_REQUIRE_GT(d.NumChildren(), 0);
```
rcurtin: We can go further than this check, we can check that the tree is identical to one without any weights.

src/mlpack/tests/decision_tree_test.cpp
```diff
@@ -428,8 +722,12 @@ BOOST_AUTO_TEST_CASE(SimpleGeneralizationTest)
   if (!data::Load("vc2_labels.txt", labels))
     BOOST_FAIL("Cannot load labels for vc2_labels.txt");
 
+  // init a weight martix
+  arma::Mat<double> weights = arma::ones<arma::Mat<double>>(labels.n_rows, labels.n_cols);
```
rcurtin: You can just call this `arma::mat` for short.

src/mlpack/tests/decision_tree_test.cpp
```diff
@@ -454,112 +752,77 @@ BOOST_AUTO_TEST_CASE(SimpleGeneralizationTest)
...
 
+/**
+ * Test that we can build a decision tree with weighted on a simple categorical dataset.
+ */
+BOOST_AUTO_TEST_CASE(CategoricalBuildTestWithWeight)
+{
+  mockCategoricalData mockData = mockCategoricalData();
+  arma::mat d = mockData.get_data();
+  arma::Row<size_t> l = mockData.get_label();
+  data::DatasetInfo di = mockData.get_data_info();
```
rcurtin: It might be nicer to just have a single function to create the mock categorical data:

```
void MockCategoricalData(arma::mat& data, arma::Row<size_t>& labels, data::DatasetInfo datasetInfo);
```

src/mlpack/tests/decision_tree_test.cpp
MikeLing: Hi @rcurtin,sorry for the late reply. I haven't got the pointer here actually. Is that mean something like:
`BOOST_REQUIRE_EQUAL(d.NumChildren(), wd.NumChildren())`?

rcurtin: Yes, you could check `NumChildren()` but you should check it on the node and all its children also, as well as their properties.  Take a look at `hoeffding_tree_test.cpp`, in `ParameterChangeTest`, for an idea of how you can check an entire tree to see if it's identical.

src/mlpack/tests/decision_tree_test.cpp
```diff
@@ -152,12 +306,30 @@ BOOST_AUTO_TEST_CASE(InformationGainEvenSplitManyClassTest)
...
+{
+  arma::Row<size_t> labels(10);
+  arma::rowvec weights("1 1 1 1 1 0 0 0 0 0");
+  for (size_t i = 0; i < 5; ++i)
+    labels[i] = 0;
+  for (size_t i = 5; i < 10; ++i)
+    labels[i] = 1;
+
+  // Zero is not a good result as gain, but we just need to prove cacluation works.
+  BOOST_REQUIRE_CLOSE(InformationGain::Evaluate<true>(labels, 2, weights), 0, 1e-5);
```
rcurtin: Zero is the best possible result for information gain, so I don't think this is 'not a good result'. :)

src/mlpack/tests/decision_tree_test.cpp
```diff
@@ -355,6 +557,28 @@ BOOST_AUTO_TEST_CASE(BasicConstructionTest)
...
+  for (size_t i = 0; i < 1000; ++i)
+    labels[i] = i % 3; // 3 classes.
+
+  // Use default parameters.
+  DecisionTree<> wd(dataset, labels, 3, weights, 50);
+  DecisionTree<> d(dataset, labels, 3, 50);
+
+  // Now require that we have some children.
+  BOOST_REQUIRE_GT(wd.NumChildren(), 0);
+  BOOST_REQUIRE_EQUAL(wd.NumChildren(), d.NumChildren());
```
rcurtin: You should make the leaf size smaller than 50 so that a full tree is built.  For this case I would say use a leaf size of 1.  Then you will have to adapt the test code here; you can take a look at the Hoeffding tree test to see some code that iterates over all nodes in a tree to ensure that they are the same.  We should definitely end up with the exact same tree as `d` since all the weights are 1.

src/mlpack/tests/decision_tree_test.cpp
```diff
@@ -454,112 +719,77 @@ BOOST_AUTO_TEST_CASE(SimpleGeneralizationTest)
...
   arma::mat trainingData = d.cols(0, 4999);
   arma::mat testData = d.cols(5000, 9999);
   arma::Row<size_t> trainingLabels = l.subvec(0, 4999);
   arma::Row<size_t> testLabels = l.subvec(5000, 9999);
 
+  arma::Row<double> weights = arma::ones<arma::Row<double>>(trainingLabels.n_elem);
+
   // Build the tree.
-  DecisionTree<> tree(trainingData, di, trainingLabels, 5, 10);
+  DecisionTree<> tree(trainingData, di, trainingLabels, 5, weights, 10);
```
rcurtin: It would be fine in this test just to build another tree on unweighted data and ensure that the results are identical.

rcurtin:
If you don't mind, let's add a few more tests.  I'll write one here.  The basic idea is that we will take a dataset that we know we can get good classification on, then we will add lots of noisy points to it, but with very low weight.  Then we will train a decision tree on that dataset (with the noise included) and ensure that we still get good performance.  We could do this like the following:

```
arma::mat dataset;
arma::Row<size_t> labels;
data::Load("vc2.csv", dataset);
data::Load("vc2_labels.txt", labels);

const size_t trueObservations = dataset.n_cols;

// Add some noise.
arma::mat noise(dataset.n_rows, 1000, arma::fill::randu);
arma::Row<size_t> noiseLabels(1000);
for (size_t i = 0; i < noiseLabels.n_elem; ++i)
  noiseLabels[i] = math::RandInt(3); // Random label.

// Concatenate data matrices.
arma::mat data = arma::join_cols(data, noise);
arma::Row<size_t> fullLabels = arma::join_cols(labels, noiseLabels);

// Now set weights.
arma::rowvec weights(dataset.n_cols + 1000);
for (size_t i = 0; i < dataset.n_cols; ++i)
  weights[i] = math::Random(0.9, 1.0);
for (size_t i = dataset.n_cols; i < dataset.n_cols + 1000; ++i)
  weights[i] = math::Random(0.0, 0.01); // Low weights for false points.

// Now build the decision tree.  I think the syntax is right here.
DecisionTree<> d(data, fullLabels, weights, 10);

// Now we can check that we get good performance on the VC2 test set.
arma::mat testData;
arma::Row<size_t> testLabels;
data::Load("vc2_test.csv", testData);
data::Load("vc2_test_labels.txt", testLabels);

// You can use other classification performance test code from other tests now to check that accuracy is good...
```

Let me know if there's anything I can clarify about that idea.  I think that everything is right syntax-wise, but I have not explicitly tested it so there could be small errors.

rcurtin:
@MikeLing: thanks for the contribution!  I had to make several more changes; you can look at them if you like from 2fd5333..b701267.

I am glad this support is here now.  :)

rcurtin:
Ah, also, I see your name is not in `src/mlpack/core.hpp` or `COPYRIGHT.txt`, would you like to add it?

MikeLing:
Yes, please. Thank you very much

rcurtin:
Sure, what name/email should I use? :)

MikeLing:
Same name or email with GitHub account. I  believe MikeLing is just fine;)

PR #950, opened 2017-03-19
Follow standard procedure for 'TestBooleanOption'

NaxAlpha:
Applied this fix in cli_test.cpp>>TestBooleanOption
Fix


rcurtin:
Hi there,

I'm not sure that there was a problem that needed to be solved here, but I don't mind merging this code anyway---it removes a few lines from the existing code and doesn't change the functionality.

However, I see that the commit history here includes several unrelated commits.  Could you please close this PR and open a new one with only the relevant commits?  (a1e9f87 and b31fa3a)

Thanks!

NaxAlpha:
Ok got it. And Thanks for help!

PR #955, opened 2017-03-21
Implemented the BatchNorm Layer

chvsp:
Implemented the Batch Normalization Layer and have written tests to check the working of the same.


chvsp:
@zoq Ready for review.

src/mlpack/tests/ann_layer_test.cpp
```diff
@@ -231,6 +232,104 @@ BOOST_AUTO_TEST_CASE(JacobianConstantLayerTest)
...
+ * the values from another implementation.
+ */
+BOOST_AUTO_TEST_CASE(BatchNormTest)
+{
+  arma::mat dataset, output;
+  data::Load("iris.csv", dataset, true);
+
+  arma::mat input = dataset.submat(0, 0, dataset.n_rows-1, 2);
+
+  BatchNorm<> model(input.n_rows);
```
zoq: It looks like this line causes the windows build error

```
3949: [00:07:04] C:\projects\mlpack\src\mlpack\tests\ann_layer_test.cpp(245): error C2664: 'mlpack::ann::BatchNorm<arma::mat,arma::mat>::BatchNorm(mlpack::ann::BatchNorm<arma::mat,arma::mat> &&)': cannot convert argument 1 from 'const arma::uword' to 'const mlpack::ann::BatchNorm<arma::mat,arma::mat> &' [C:\projects\mlpack\build\src\mlpack\tests\mlpack_test.vcxproj]
```

maybe if we cast ``input.n_rows`` to size_t it work's?

zoq:
Okay, here is another idea, remove ``#include <mlpack/methods/ann/layer/batchnorm.hpp>``from ``layer_types.hpp`` and put the header into ``layer.hpp``. This is once again a really strange windows related error.

chvsp:
@zoq, I have removed the headers as suggested. I have also tried after removing the line `using namespace std;`, it still gives the same error

zoq:
Here is another idea: make the constructor explicit ``explicit BatchNorm(const size_t size, const double eps);``. If that doesn't work, I'll take a closer look.

chvsp:
@zoq The same error persists

src/mlpack/methods/ann/layer/batchnorm_impl.hpp
```diff
@@ -0,0 +1,136 @@
...
+
+// In case it is not included.
+#include "batchnorm.hpp"
+
+namespace mlpack {
+namespace ann { /** Artificial Neural Network. */
+
+template <typename InputDataType, typename OutputDataType>
+BatchNorm<InputDataType, OutputDataType>::BatchNorm(
+    const size_t size, const double eps = 0.001) : size(size), eps(eps)
```
zoq: Can you put the default parameter (eps = 0.001) into the header file.

zoq: If someone includes just the hpp file, all it sees is the header file, since there is no default value for the second parameter, the code should fail if you pass just the first parameter. I wonder why g++ is fine with it.

chvsp:
@zoq, The AppVeyor build has succeeded. I had overlooked the default declaration part. Kindly review the code. Thank you.

src/mlpack/tests/ann_layer_test.cpp
```diff
@@ -231,6 +232,106 @@ BOOST_AUTO_TEST_CASE(JacobianConstantLayerTest)
...
+
+  BatchNorm<> model(numUnits);
+  model.Reset();
+
+  // Non-Deteministic Forward Pass Test.
+  model.Deterministic() = false;
+  model.Forward(std::move(input), std::move(output));
+
+  arma::mat result;
+  result << 1.20240722e+00 << 5.33976074e-15 << -1.20240722e+00 << arma::endr
```
zoq: Can you add a comment about where this values come from?

src/mlpack/methods/ann/layer/batchnorm_impl.hpp
```diff
@@ -0,0 +1,136 @@
...
+        arma::as_scalar(arma::sqrt(arma::var(input.row(i), 1, 1) + eps));
+  }
+
+  gradient.submat(0, 0, gamma.n_elem - 1, 0) = arma::sum(normalized % error, 1);
+  gradient.submat(gamma.n_elem, 0, gradient.n_elem - 1, 0) = arma::sum(error, 1);
+}
+
+template<typename InputDataType, typename OutputDataType>
+template<typename Archive>
+void BatchNorm<InputDataType, OutputDataType>::Serialize(
```
zoq: Can you add a test for the serialization, take a look at: https://github.com/mlpack/mlpack/blob/master/src/mlpack/tests/serialization_test.cpp for more information, I'm not certain that it works, so let's test it out.

src/mlpack/tests/ann_layer_test.cpp
chvsp: I have added the link to the implementation.

src/mlpack/methods/ann/layer/batchnorm_impl.hpp
```diff
@@ -0,0 +1,136 @@
...
+    {
+      stats(input.col(i));
+    }
+  }
+
+    for (size_t i = 0; i < output.n_rows; i++)
+    {
+      output.row(i) = input.row(i) - arma::as_scalar(mean.row(i));
+      output.row(i) /= (arma::as_scalar(arma::sqrt(variance.row(i) + eps)));
+      output.row(i) *= arma::as_scalar(gamma.row(i));
```
sumedhghaisas: Should we use '.each_row' function here? I am not sure how its implemented in Armadillo, but Armadillo is known to implement optimisations everywhere. Maybe it is using some clever row iterator form. Nonetheless, I think it will be at least as good as the implementation we are doing. What do you think?

rcurtin: Nice catch; definitely at the very least this can be condensed significantly into one expression and that will give some speedup.

Issue #821, opened 2016-11-30
mlpack_adaboost error


cypro666:
./mlpack_adaboost -t train.csv -l labels.csv -M model.txt
./mlpack_adaboost -m model.txt -T train.csv -o result.csv

[FATAL] input stream error
terminate called after throwing an instance of 'std::runtime_error'
  what():  fatal error; see Log::Fatal output
Aborted (core dumped)

train.csv : 
2,3
2,3
2,4
1,0
2,3
6,6
7,8
2,3
2,3
2,4
1,0
2,3
6,6
7,8
2,3
2,3
2,4
1,0
2,3
6,6
7,8

lables.csv:
0
0
0
0
0
1
1
0
0
0
0
0
1
1
0
0
0
0
0
1
1

version:
mlpack 2.1.0
boost 1.58.0
blas openblas 2.18

rcurtin:
I can reproduce this with both 2.1.0 and git master.  But I found that if you use binary models (i.e. `-M model.bin` and `-m model.bin`) there is no problem and everything seems to work.  Text and XML models fail.  I'll try to look into this in the coming days.

cypro666:
Aha， bin format works well ，Thx！ 
Infact bin format makes more efficiency， but sometimes texts get better readability ：）

rcurtin:
So the issue here is twofold:

 * The training set you are using can be perfectly modeled by the weak learner in a single iteration of AdaBoost.  But the AdaBoost code does not handle this as well as it could.
 * There is some problem serializing `nan` values in text and binary format.

I have some ideas for how to fix it but at the moment I don't have the time.  If you train on a more difficult dataset, I think that you are not likely to see this problem.  Still, I will fix it when I have a chance! :)

nitinkgp23:
Hey Ryan, 

I am a newbie here, and I am in a position of just jumping in and contribute. I wanted to take up this issue, since most of the 'easy'  labelled isues have already been taken up or have been fixed. I will require help from you on how to go ahead with this issue.

shikharbhardwaj:
I'll try and isolate this issue if no one else is actively working on it right now.

rcurtin:
The Adaboost problem is fixed with #956 being merged (thanks @nish21!).  The nan serialization issue still remains.

shikharbhardwaj:
Thanks for the update @rcurtin.
I checked the reported error and cannot replicate it after #956 .

[...]

PR #956, opened 2017-03-21
Adaboost error fix

nish21:
A fix for issue in #821. The problem was that for the given training set, ```rt``` value was a very small value greater than ```1``` which caused the calculation of log of a negative value. Hence the additional check of the value of ```rt```. If the value of ```rt = 1``` then the denominator in the calculation of ```alphat``` is zero. ```DBL_MAX``` is introduced to avoid calculating log of ```inf```.

After this change
- The reported crash is fixed. 
- The adaboost algorithm now converges in one iteration on the given dataset. 
- The problem with serializing and deserializing ```nan``` is behind the curtains. I'll look into it next.

### DIFF
--- a/src/mlpack/methods/adaboost/adaboost_impl.hpp
+++ b/src/mlpack/methods/adaboost/adaboost_impl.hpp
@@ -135,14 +135,20 @@ void AdaBoost<WeakLearnerType, MatType>::Train(
     if ((i > 0) && (std::abs(rt - crt) < tolerance))
       break;
 
+    // Check if model has converged.
+    if ((rt >= 1))
+    {
+      // Save the weak learner and terminate.
+      alpha.push_back(1.0);
+      wl.push_back(w);
+      break;
+    }
+
     crt = rt;
 
     // Our goal is to find alphat which mizimizes or approximately minimizes the
     // value of Z as a function of alpha.
-    if (rt >= 1)
-      alphat = 0.5 * log(DBL_MAX);
-    else
-      alphat = 0.5 * log((1 + rt) / (1 - rt));
+    alphat = 0.5 * log((1 + rt) / (1 - rt));
 
     alpha.push_back(alphat);
     wl.push_back(w);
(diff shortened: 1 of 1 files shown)


rcurtin:
Thanks for looking into this.  Would it not also work to simply terminate when rt >= 1, on line 135?

nish21:
@rcurtin Thanks for the review. Checking rt >= 1 on line 135 will not work because in iteration zero, we will still try to calculate log of a negative value and that will reintroduce the bug.

rcurtin:
Have you tested that strategy?  If you check on line 135 then you will break out of the loop and thus there will never be a log calculation of an invalid value.

nish21:
yes, the check on line 135 will break out only if the iteration count is greater than zero due to ```(i > 0)```. Since the iteration in question is iteration zero, it'll calculate the log value. I tested it out.

It makes sense to check rt value in line 135. But in this training example, an invalid value of rt is reached in the first iteration itself, which won't be caught by the check.

rcurtin:
Or you can just modify the condition:

```
if (((i > 0) && (std::abs(rt - crt) < tolerance)) || (rt >= 1.0))
```

Have I overlooked something?

nish21:
I just tested that condition as well. The model didn't learn. This is probably because we will never execute the statement 
```wl.push_back(w);```

rcurtin:
Then move when `wl.push_back(w)` is called.  There may be some amount of debugging to do to make it work but the general strategy I just outlined would be the better way to go.

nish21:
Yes, I realize the condition checking on line 135 would be better. Will look into this further and update the PR soon.

rcurtin:
Right, looking forward to the update.  My feeling is just that it's a little bit of a hack to approximate `alphat` as some very large value when `rt = 1` even though this is not actually what should happen---in reality the algorithm should not proceed past line 135 since it is already converged.

rcurtin:
Thanks, I appreciate you taking the time to look into this.  I think this fix is good.  It seems like it is better to have this condition as an `if` and not an `else if` like you've done it, because a weak learner perfectly fitting the data can happen at any iteration---and in that case the algorithm must terminate otherwise `alpha` explodes.  If you'd like to add your name to the list of contributors in `src/mlpack/core.hpp` and `COPYRIGHT.txt`, please go ahead and do that and open another PR, and I'll merge it in.

Issue #948, opened 2017-03-19
Training constructors for LARS


micyril:
Right now LARS has two constructors
```
  LARS(const bool useCholesky = false,
       const double lambda1 = 0.0,
       const double lambda2 = 0.0,
       const double tolerance = 1e-16);

  LARS(const bool useCholesky,
       const arma::mat& gramMatrix,
       const double lambda1 = 0.0,
       const double lambda2 = 0.0,
       const double tolerance = 1e-16);
```
and one training method
```
  void Train(const arma::mat& data,
             const arma::vec& responses,
             arma::vec& beta,
             const bool transposeData = true);
```
Every classification and regression algorithm I have encountered in mlpack has a training constructor (a constructor with a body containing a call of a Train method), what is considered to be a good practice according to the design guidelines. But there is a lack of such ones for LARS. Should we consider adding some training constructors to LARS? I can imagine such signatures:
```
  LARS(const arma::mat& data,
       const arma::vec& responses,
       const bool transposeData = true,
       const bool useCholesky = false,
       const double lambda1 = 0.0,
       const double lambda2 = 0.0,
       const double tolerance = 1e-16);

  LARS(const arma::mat& data,
       const arma::vec& responses,
       const bool transposeData,
       const bool useCholesky,
       const arma::mat& gramMatrix,
       const double lambda1 = 0.0,
       const double lambda2 = 0.0,
       const double tolerance = 1e-16);
```

rcurtin:
I agree, this would be a welcome improvement.

micyril:
Ok, I will handle it

rcurtin:
Fixed with #957.

PR #957, opened 2017-03-22
Add training constructors to LARS

micyril:
Fix for [#948](https://github.com/mlpack/mlpack/issues/948)


rcurtin:
Looks good to me.  It may also be useful here to add an overload of `Train()` that doesn't need the `beta` parameter, because the user can access `beta` as `BetaPath()[BetaPath().size() - 1]` (it may also be useful to add a `Beta()` function that just returns that).

PR #958, opened 2017-03-23
Edited output parameter for Predict in Ann and Rnn

Trion129:
The current parameter name 'responses' raised a confusion as it makes it look same as Train function in declaration. I changed it 'results', so did all the other variables that had response in name in the Predict function.
**Changes in:**
1. Predict in ann
2. Predict in rnn
3. SinglePredict in rnn


zoq:
Looks good, If you'd like to add your name to the list of contributors in ``src/mlpack/core.hpp`` and ``COPYRIGHT.txt`` please go ahead and do that.

zoq:
Thanks for the contribution, I think it's easier now to distinguish both functions.

PR #962, opened 2017-03-25
Overloading Train and accessing Beta in LARS

micyril:
Implementing suggestions from the comment to [the PR #957](https://github.com/mlpack/mlpack/pull/957).


rcurtin:
Looks good, thanks!

PR #966, opened 2017-03-27
Add ELU in layer types

gvsi:


zoq:
Nice catch, thanks for the contribution.

PR #967, opened 2017-03-28
Fix bug of variadic template parameters of Optimizer

ShangtongZhang:
I just found I can't use cherry pick... So I create a separate PR for this fix.

### DIFF
--- a/src/mlpack/methods/ann/ffn.hpp
+++ b/src/mlpack/methods/ann/ffn.hpp
@@ -92,11 +92,12 @@ class FFN
    * @param optimizer Instantiated optimizer used to train the model.
    */
   template<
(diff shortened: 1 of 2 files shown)


zoq:
I forgot to push the changes; there where a couple of other places that I updated too. Also when using variadict templates  we use the convention:

```
template<typename... Arguments> class classname;
```

And if the type isn't used like in this case:

```
template<
    template<typename> class OptimizerType = mlpack::optimization::RMSprop
    template<typename...> class OptimizerType = mlpack::optimization::RMSprop
>
void Train(const arma::mat& predictors, const arma::mat& responses);
```

we don't have to provide the OptimizerTypeArgs args. For more information take a look at: 2344eba81ca927d

Issue #922, opened 2017-03-06
Identify commonly failing tests and address them


rcurtin:
This is a large ticket that many people can collaborate on.

mlpack has very many tests, and many of these tests are probabilistic.  This means that they can randomly fail, even when something isn't actually wrong with the code.  The purpose of this ticket is to identify these tests, open issues to address those failing tests, and fix the tests.  Typically this will involve understanding an algorithm, understanding the test, and adjusting the tolerance.  This is a good way to get your hands dirty with mlpack, so I think this is a good starter ticket for anyone looking to get involved.  I've labeled it all of 'easy', 'moderate', and 'difficult', because depending on the test, it might be easy to solve and it might not.  Identifying failing tests should be an easy task too.

mlpack has a build server that runs a nightly matrix build.  This can be found at

http://masterblaster.mlpack.org/job/mlpack%20-%20nightly%20matrix%20build/

If you click on any of the 'yellow' builds, you can get to the list of failing tests.  You can also look at the Travis build history for mlpack, or run tests locally.  We should identify tests that fail sometimes, and then open tickets to handle those individually.  I will start with a couple here.

- [ ] `RANNTest/RAModelTest`: #734 (difficult)
- [ ] `LARSTest/NoCholeskyTest`: #355 (difficult) -- only on i386
- [x] `GMMTest/GMMTrainEMOneGaussian`
- [x] `VanillaNetworkTest` in `convolutional_network_test.cpp`
- [x] `EmbeddedReberGrammarTest` in `recurrent_network_test.cpp`
- [x] `LocalCoordinateCodingTest`: #943 (difficult) -- only on sparc64

Debugging tests that fail randomly can be done (or finding how often they fail) can be done by setting the random seed at the top of the test in question:

```
math::RandomSeed(std::time(NULL));
```

Then, you can run that individual test in a loop to see how often it fails:

```
while(true); do bin/mlpack_test -t TestSuiteName/TestCaseName; sleep 1; done
```

and this will run the test over and over again, and you can see how often it fails.  That is usually a good first step to start understanding why the test is failing.

These failing tests are important to fix, because users (and new contributors) often get confused by failing Travis builds.  We can close this bug when we've got the list down to 0. :)

If you'd like to contribute to this effort:

 * Comment with the name of a test that seems to be commonly failing (and I'll update the list above).
 * Open a separate issue for a test that is commonly failing (and I'll update the list above).
 * Investigate a test that is commonly failing and submit a PR to fix it.

[...]

PR #971, opened 2017-03-29
Fix VanillaNetworkTest

amoudgl:
### Issue
CNN used to get stuck in local minima and couldn't converge in the given number of iterations. #922

### Hyperparameter Tuning
* Since the network is performing only binary classification task of recognizing 4 and 9, so number of output units in the CNN network has been updated to 2. This reduces the number of parameters in the network, hence avoiding wrong classification due to weights which could not be trained in the given number of iterations.
* Reduced number of units in full connected layers to reduce number of trainable parameters. 
* Reduced learning rate of optimizer to converge easily in given number of iterations. 

### Updates in PR
* Removed loop which performs the same test five times, hence reducing the testing time. 
* Renamed `error` variable as `correct` for more clarity because it counts the number of correct predictions by the network.
* Did Hyperparameter tuning as explained above.

### Results
Test was run for 75 times with `math::RandomSeed(std::time(NULL))` added to the top. Mean classification error over the runs was around `0.03` (rough estimate). Console log (with `classificationError` printed in each run) can be viewed [here](https://pastebin.com/1vBsHv3U). Please let me know if you are able to reproduce similar results.

### DIFF
--- a/src/mlpack/tests/convolutional_network_test.cpp
+++ b/src/mlpack/tests/convolutional_network_test.cpp
@@ -1,6 +1,7 @@
 /**
  * @file convolutional_network_test.cpp
  * @author Marcus Edel
+ * @author Abhinav Moudgil
  *
  * Tests the convolutional neural network.
  *
@@ -22,7 +23,6 @@
 using namespace mlpack;
 using namespace mlpack::ann;
 using namespace mlpack::optimization;
-
 
 BOOST_AUTO_TEST_SUITE(ConvolutionalNetworkTest);
 
@@ -47,11 +47,13 @@ BOOST_AUTO_TEST_CASE(VanillaNetworkTest)
   {
     if (i < nPoints / 2)
     {
-      Y(i) = 4;
+      // class 1 - digit = 4
+      Y(i) = 1;
     }
     else
     {
-      Y(i) = 9;
+      // class 2 - digit = 9
+      Y(i) = 2;
     }
   }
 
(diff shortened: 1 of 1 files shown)


rcurtin:
Hi Abhinav,

Thanks for the detailed investigation.  It looks like maybe you and @zoq were working on this in parallel; I think that https://github.com/mlpack/mlpack/commit/ae93d749a2fed8c5418df14dbfbff9536177bdb6 solves the problem also.

I went ahead and tested both fixes for 1000 iterations to see if there were any failures; I saw no failures for either test strategy.  I also saw that there was no significant runtime difference for either strategy; it takes about 5.5s on my system for either strategy.

@zoq, @amoudgl: What do you think we should do here?  At the very least renaming `error` to `correct` is still a strategy that could work; maybe there are a couple other improvements, but in either case the big issue (the failing test) seems solved.

amoudgl:
@rcurtin Yes, I saw that commit by @zoq. I made the changes with the thought in mind that why not do it in one shot like other tests rather than running a loop. I also hoped that hyperparameter tuning would reduce the testing time but thanks for your analysis, I didn't actually check the testing time difference. I see that it doesn't make any difference so we can go with either solution. I will try to perform multiple experiments with multiple iterations in each one and come up with some detailed analysis of time and number of failures, to converge to one solution.

zoq:
If @amoudgl solution runs for 1000 iterations without any error I think we should go with it instead of running the test again. As I tested my solution I noticed that in like 5 of 100 cases the tests runs a second time. @amoudgl what do you think?

amoudgl:
@zoq Thanks for the info. Yes, I agree with your view point. Please let me know if I should make any edits in my solution otherwise I think it is merge ready.

src/mlpack/tests/convolutional_network_test.cpp
```diff
@@ -47,11 +47,13 @@ BOOST_AUTO_TEST_CASE(VanillaNetworkTest)
   {
     if (i < nPoints / 2)
     {
-      Y(i) = 4;
+      // class 1 - digit = 4
```
zoq: Pedantic style issue  it's helpful to always have comments that are complete sentences with proper grammar and punctuation.

amoudgl: Oops, thanks for pointing this out. Fixed now. 👍

amoudgl:
@zoq I updated the PR to reflect code review changes. Do you think this PR is ready to be merged now?

PR #972, opened 2017-03-30
Adam and AdaMax update policy implementation

ivmarkp:
Implementation of the Adam optimizer using the AdamUpdate policy class. All tests pass locally. Please review, thanks.

### DIFF
--- a/src/mlpack/core/optimizers/adam/adam.hpp
+++ b/src/mlpack/core/optimizers/adam/adam.hpp
@@ -64,7 +64,7 @@ namespace optimization {
  * @tparam DecomposableFunctionType Decomposable objective function type to be
  *     minimized.
  */
-template<typename DecomposableFunctionType>
+template<typename DecomposableFunctionType, bool adaMax = false>
 class Adam
 {
  public:
@@ -87,8 +87,6 @@ class Adam
    * @param tolerance Maximum absolute tolerance to terminate algorithm.
    * @param shuffle If true, the function order is shuffled; otherwise, each
    *        function is visited in linear order.
-   * @param adaMax If true, then the AdaMax optimizer is used; otherwise, by
-   *        default the Adam optimizer is used.
    */
   Adam(DecomposableFunctionType& function,
       const double stepSize = 0.001,
@@ -97,8 +95,7 @@ class Adam
       const double eps = 1e-8,
       const size_t maxIterations = 100000,
       const double tolerance = 1e-5,
-      const bool shuffle = true,
-      const bool adaMax = false);
+      const bool shuffle = true);
 
   /**
    * Optimize the given function using Adam. The given starting point will be
@@ -153,14 +150,9 @@ class Adam
   //! Modify whether or not the individual functions are shuffled.
   bool& Shuffle() { return optimizer.Shuffle(); }
 
-  //! Get whether or not the AdaMax optimizer is specified.
-  bool AdaMax() const { return optimizer.UpdatePolicy().AdaMax(); }
-  //! Modify wehther or not the AdaMax optimizer is to be used.
-  bool& AdaMax() { return optimizer.UpdatePolicy().AdaMax(); }
-
  private:
   //! The Stochastic Gradient Descent object with Adam policy.
-  SGD<DecomposableFunctionType, AdamUpdate> optimizer;
+  SGD<DecomposableFunctionType, AdamUpdate<adaMax> > optimizer;
 };
 
 } // namespace optimization
--- a/src/mlpack/core/optimizers/adam/adam.hpp
+++ b/src/mlpack/core/optimizers/adam/adam.hpp
@@ -94,13 +94,13 @@ class AdamType
    *        function is visited in linear order.
    */
   AdamType(DecomposableFunctionType& function,
-      const double stepSize = 0.001,
-      const double beta1 = 0.9,
-      const double beta2 = 0.999,
-      const double eps = 1e-8,
-      const size_t maxIterations = 100000,
-      const double tolerance = 1e-5,
-      const bool shuffle = true);
+          const double stepSize = 0.001,
+          const double beta1 = 0.9,
+          const double beta2 = 0.999,
+          const double eps = 1e-8,
+          const size_t maxIterations = 100000,
+          const double tolerance = 1e-5,
+          const bool shuffle = true);
 
   /**
    * Optimize the given function using Adam. The given starting point will be
--- a/src/mlpack/core/optimizers/adam/adam.hpp
+++ b/src/mlpack/core/optimizers/adam/adam.hpp
@@ -20,6 +20,9 @@
 
 #include <mlpack/prereqs.hpp>
 
+#include <mlpack/core/optimizers/sgd/sgd.hpp>
+#include "adam_update.hpp"
+
 namespace mlpack {
 namespace optimization {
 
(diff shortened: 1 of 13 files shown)


rcurtin:
My opinion would be that instead of modifying the `Update()` API for all `UpdatePolicy` types, instead you could keep an internal `iteration` variable (probably of type `double` to avoid the cast later) and increment that at each iteration.  Also, I think you could turn the `adaMax` boolean parameter into a template parameter.  What do you think?

ivmarkp:
Alright, I'll try that. Thanks for taking a quick look.

zoq:
Looks like, if we do something like this and use variadic templates the compiler cannot deduce the type.

```c++
template<typename DecomposableFunctionType, bool adaMax = false>
class Adam
{
}
```

in contrast

```c++
class AdaMax {};

template<typename DecomposableFunctionType, class adaMax = AdaMax>
class Adam
{
}
```

works just fine. I guess a simple solution would be to provide one update rule for each method (they don't really share any code, so there's basically no code duplication). One for adamax and another one for standard adam, that way we could even provide another update rule for adamin as proposed in the paper. So something like this should work:

```c++
// Standard adam update.
class AdamUpdate
{

}

// AdaMax update.
class AdaMaxUpdate
{

}

// AdaMin update.
class AdaMinUpdate
{

}

// Simple Proxy.
template<typename DecomposableFunctionType, typename UpdateRule = AdamUpdate>
class AdamType
{

}

// Provide convenience typedefs.

using Adam = AdamType<DecomposableFunctionType, AdamUpdate>;

using AdaMax = AdamType<DecomposableFunctionType, AdaMaxUpdate>;

using AdaMin = AdamType<DecomposableFunctionType, AdaMinUpdate>;
```

What do you think?

ivmarkp:
This looks good. Thanks for a handy solution. I'm now making the changes and will the update the PR shortly.

>  we could even provide another update rule for adamin as proposed in the paper. 

Sorry, I might have missed something but I couldn't find AdaMin update rule in the paper. Could you please tell me where to look for it? Thanks.

zoq:
> Sorry, I might have missed something but I couldn't find AdaMin update rule in the paper. Could you please tell me where to look for it? Thanks.

I thought they talked about Adamin but it turns out they haven't.

src/mlpack/core/optimizers/adam/adam.hpp
```diff
@@ -84,18 +92,15 @@ class Adam
    * @param tolerance Maximum absolute tolerance to terminate algorithm.
    * @param shuffle If true, the function order is shuffled; otherwise, each
    *        function is visited in linear order.
-   * @param adaMax If true, then the AdaMax optimizer is used; otherwise, by
-   *        default the Adam optimizer is used.
    */
-  Adam(DecomposableFunctionType& function,
+  AdamType(DecomposableFunctionType& function,
       const double stepSize = 0.001,
```
zoq: Minor style issue, can you line up the parameters?

src/mlpack/core/optimizers/adam/adam_update.hpp
```diff
@@ -0,0 +1,148 @@
...
+ * @endcode
+ */
+class AdamUpdate
+{
+ public:
+  /**
+   * Construct the Adam update policy with the given epsilon parameter.
+   *
+   * @param epsilon The epsilon value used to initialise the squared gradient
+   *        parameter.
```
zoq: Can you add a parameter description for ``beta1`` and ``beta2``?

src/mlpack/core/optimizers/adam/adamax_update.hpp
```diff
@@ -0,0 +1,143 @@
...
+ * @endcode
+ */
+class AdaMaxUpdate
+{
+ public:
+  /**
+   * Construct the AdaMax update policy with the given epsilon parameter.
+   *
+   * @param epsilon The epsilon value used to initialise the squared gradient
+   *        parameter.
```
zoq: Can you add a parameter description for beta1 and beta2?

src/mlpack/core/optimizers/adam/adam.hpp
ivmarkp: Thanks for pointing that out. Will fix that.

src/mlpack/core/optimizers/adam/adamax_update.hpp
ivmarkp: Yeah, sorry I missed that. Will add the description.

ivmarkp:
Changed the PR title to reflect the separate implementation of AdaMax update policy.

src/mlpack/core/optimizers/adam/adam_update.hpp
```diff
@@ -0,0 +1,150 @@
...
+
+#include <mlpack/prereqs.hpp>
+
+namespace mlpack {
+namespace optimization {
+
+/**
+ * Adam is an optimizer that computes individual adaptive learning rates for
+ * different parameters from estimates of first and second moments of the
+ * gradients. AdaMax is a variant of Adam based on the infinity norm as given
```
zoq: Since this file implements the standard Adam update, we can probably remove the section about AdaMax, what do you think?

src/mlpack/core/optimizers/adam/adamax_update.hpp
```diff
@@ -0,0 +1,145 @@
...
+
+    // And update the iterate.
+    m *= beta1;
+    m += (1 - beta1) * gradient;
+
+    // Update the exponentially weighted infinity norm.
+    u *= beta2;
+    u = arma::max(u, arma::abs(gradient));
+
+    const double biasCorrection1 = 1.0 - std::pow(beta1, (double) iteration);
```
zoq: The cast is redundant since iteration is already from type double.

src/mlpack/core/optimizers/adam/adam_update.hpp
ivmarkp: Yes, true. I'll make the change.

src/mlpack/core/optimizers/adam/adam_update.hpp
```diff
@@ -0,0 +1,148 @@
...
+ * gradients as given in the section 7 of the following paper.
+ *
+ * For more information, see the following.
+ *
+ * @code
+ * @article{Kingma2014,
+ *   author    = {Diederik P. Kingma and Jimmy Ba},
+ *   title     = {Adam: {A} Method for Stochastic Optimization},
+ *   journal   = {CoRR},
+ *   year      = {2014}
```
rcurtin: I might suggest adding `url       = {http://arxiv.org/abs/1412.6980}` to the citation here.

ivmarkp: @rcurtin  I'll get this covered but I'm not sure if we really want to add `url` here because [citation guidelines](https://github.com/mlpack/mlpack/wiki/DesignGuidelines#citations) suggest otherwise. Although, I'm aware that you wrote all design guidelines for mlpack but just wanted to check if there was a decision in the recent past to also add urls in the citations :)

rcurtin: Ah, ok, I see.  In this case I would say that a URL is fine because ICLR papers aren't actually published in a journal, they're only put on arXiv, and we can probably assume that arXiv has URLs that won't change.  I think when I originally wrote to avoid URLs, I was trying to avoid linking to the author's homepage, which could go down if the author changes affiliation or moves or simply restructures their website.  What do you think, should we modify the guideline there?

ivmarkp: >  I think when I originally wrote to avoid URLs, I was trying to avoid linking to the author's homepage, which could go down if the author changes affiliation or moves or simply restructures their website. 

Ah, I see -- thanks for clarifying. I'll update the citation with url.

> What do you think, should we modify the guideline there?

Yes, definitely we should. :)

rcurtin:

ivmarkp:
@rcurtin Thanks, more to come! :)

PR #974, opened 2017-04-02
#include "add_merge.hpp" instead of #include "add_merge_impl.hpp" in add_merge_impl.hpp

sagarbhathwar:
Hey,

While going through the NN layers code, came across this small typo, which is harmless. But just for mlpack's sake!


zoq:
Nice catch, thanks for the contribution.

PR #975, opened 2017-04-02
Correct minor typo in ffn.hpp

sidak:
This is a small pull request that restores balance to the universe by changing '_infomration_' to '_information_'. 

Thanks Marcus for sharing that tweet! :)


rcurtin:
Important fix.  Thanks! :)

Issue #420, opened 2015-03-14
Considering using c++ linting tools


stephentu:
Not sure how much we care about this, but it might be nice to start running the mlpack code through a c++ linter, such as [cpplint](http://google-styleguide.googlecode.com/svn/trunk/cpplint/cpplint.py).

Here is an example of what we would get from cpplint on our codebase:  [https://gist.github.com/anonymous/7ba0015c0a527a381bb5](https://gist.github.com/anonymous/7ba0015c0a527a381bb5)

Some of the complaints are silly and can be removed by filtering them out, but I think some are reasonable. I know you guys are thinking about updating the style guidelines, so it might be nice to have some sort of automated enforcement.

rcurtin:
I agree with this. Not sure I have time to look into the configuration too much but I am happy to help adapt code. I think we all ought to build a design document and coding standards first, and then go from there. (I know I've been saying that for a while... it's close enough to the top of the stack that I think I'll write something up within a few weeks...)

stephentu:
Once you finalize the coding style, I can adapt cpplint to enforce as much of it as possible. Ping this bug when that's ready to go.

theaverageguy:
Are there any auto linters existing ? Like we have autopep8 for python? Maybe I can help.

theaverageguy:
Can I start with Uncrustify @stephentu ? It seems that it could fix a lot of it. 
I would like your help for this PR.

stepthom:
I've had great success using [ClangFormat](http://clang.llvm.org/docs/ClangFormat.html) to stylize code and ensure coding standards. When you make it part of the testing process, code is always ensured to be consistently styled before it's committed.

zoq:
I heard good things about Clang Format. It looks like nobody is working on the issue right now, so if you like to pick it up feel free but don't feel obligated.

sam0410:
stephentu: the link mentioned ('cpplint') is not available

outofnames:
@stephentu , the link you've posted isn't working, you can update it to [this link (github)](https://github.com/google/styleguide/tree/gh-pages/cpplint) 

And is someone working on it? I am kinda interested in it to start, just need a green flag

rcurtin:
Sure, nobody is working on this now; if you wanted to do it, it would be appreciated.

amoudgl:
[...]

PR #981, opened 2017-04-08
Add cpplint to Travis for style check

amoudgl:
This PR implements automated style checks in Travis. #420


amoudgl:
@rcurtin Google cpplint uses a different braces style convention for all cases like class, if/else, loops etc. with opening bracket on the same line. Rather than modifying the regex in [cpplint.py](https://github.com/google/styleguide/blob/gh-pages/cpplint/cpplint.py), I decided to shift to [clang-format](https://clang.llvm.org/docs/ClangFormat.html), which is much more customizable and easy to use. The only limitation of clang-format I can see is that it doesn't have nice reporting of formatting errors like Google's cpplint. We can only reformat a file using a command: 

```
clang-format -i -style=file /path/to/file
```
where `-i` option does the formatting of file inplace according to style guidelines in `.clang-format` file otherwise it prints the formatted output to stdout. To find style errors and report in Travis build, we may need to apply formatting changes by clang-format and then check difference between user commit and clang-format output. What do you think?

zoq:
I was thinking instead of using Travis we could use Jenkins to show more details. If I'm right if the style is not conform with the guidelines the build is marked as failed. But we don't provide any further information, unfortunately, there is no Jenkins plugin to show diffs, but we could use diff2html in combination with the HTML Publisher Plugin to give some pretty neat hints. Let me know what you think.

amoudgl:
Hi @zoq, in `script.sh` I am reporting two things to the user: 

- The file in which style error is there. User will get an error message like this: 
  ```
  The file $filename is not compliant with the coding style.
  ```
- The diffs. Since there may be lot of style errors in a single pull request, I have truncated the diffs to 50 i.e. if there are more than 50 files with faulty style, no more further diffs will be shown. User will get a message: 

  ```
  Too many style errors encountered previously, this diff is hidden.
  ```
I only check the modified files in the pull request. If there are any modified source code files (i.e. with extension `.hpp` or `.cpp`), I further check if they lie in `src/mlpack/core/arma_extend/` or `src/mlpack/core/boost_backport/` because files in these directory do not follow explicity the style guidelines and style check on them should be ignored as stated by Ryan in #420. If there is any file with faulty style, build is marked as failed otherwise pass.

.travis.yml
```diff
@@ -6,14 +6,20 @@ before_install:
   - sudo add-apt-repository -y ppa:ubuntu-toolchain-r/test
   - sudo apt-get update -qq
   - sudo apt-get install -qq g++-4.8 libarmadillo-dev libarmadillo4 libboost1.54-all-dev
+  - curl http://apt.llvm.org/llvm-snapshot.gpg.key | sudo apt-key add -
+  - sudo add-apt-repository "deb http://apt.llvm.org/trusty/ llvm-toolchain-trusty-4.0 main"
+  - sudo apt-get update -qq
+  - sudo apt-get install -qq clang-4.0 clang-tidy-4.0
```
rcurtin: Is clang-tidy-3.8 acceptable?  That's already available in the trusty repositories.

zoq:
@amoudgl thanks for the insights, I'm still not sure that using Travis for this is the best we can do. The plain diff in the Travis log isn't always that easy to read, especially if the line exceeds 80 columns and even if you cut the results to 50 diffs, which is a lot Travis will hide the results by default.

amoudgl:
@rcurtin @zoq clang-format has an issue that we can't omit some particular checks done by it which are not relevant for mlpack. If we don't assign any value to configuration option in `.clang-format` file, it automatically picks up the default one and reformats the file. In `clang-tidy`, we can filter checks but its error reporting format is not so neat. 

So, I tried again Google [cpplint](https://github.com/google/styleguide/tree/gh-pages/cpplint) because of two main reasons: 

- We can filter the checks we want to do and ignore the rest. 
- Error reporting is pretty neat and clear.

I ran each check of Google cpplint on mlpack codebase individually and analyzed the ones which should be excluded (or needs modification) for mlpack. Following is the final command I came up with, after removing unnecessary filters: 

```bash
python cpplint.py --extensions=hpp,cpp --filter=\
-whitespace/braces,\ # checks if opening braces are on the same line
-whitespace/newline,\ # reports error when else statement is not the same line with }
-build/header_guard,\ # recommends to append _ after header guard
-build/include_order,\ # should be excluded otherwise it breaks the code
-build/storage_class,\ # checks for static, extern, typedef, etc at the beginning of function declaration
-build/namespaces,\ # enforces to not use namespace using-directives, suggests to use using-declarations instead.
-build/include_what_you_use,\ # e.g. if std::string is found, enforces to include <string>
-legal/copyright,\ #  copyright year and name should be there in license header at the top of file
-readability/casting,\ # discourages to use C type casting like (double) a
-readability/alt_tokens,\ # enforces to use operator "!" instead of "not" even in comments
-readability/todo,\ # # Checks for username in TODO comments like TODO (username) comment
-readability/multiline_string,\  # checks for multiline string and suggests to improve it 
-runtime/explicit,\ # single-parameter constructors should be marked explicit
-runtime/int,\ # Use int16/int64/etc, rather than the C type short
-runtime/references \ # e.g. Is this a non-const reference? If so, make const or use a pointer
$(find ./src/mlpack/core -type f ! -path "./src/mlpack/core/arma_extend/*" ! -path "./src/mlpack/core/boost_backport/*")
```

View the console log generated by above command [here](https://pastebin.com/sKT4RQT9). The issues reported in the log seems valid to me. Only `whitespace/braces` filter needs modification, which currently checks for opening brace on the same line. If this filter is handled well, we can go ahead with Google cpplint. What do you think?

rcurtin:
Ah, yeah, this looks much better to me.  I see only one issue with a quick inspection; it's the one you already pointed out:

```
./src/mlpack/core/tree/binary_space_tree/ub_tree_split.hpp:33:  Missing spaces around <  [whitespace/operators] [3]
```

Looks like it's not correctly picking up that that's a template definition not a less than operator.  I wonder if that's a cpplint bug or something.

I think you can also modify the find command a little, and add `-iname '*.[hc]pp'` to catch only cpp/hpp files.

Does cpplint automatically fix these issues, or would we have to do that ourselves to get the current code in conformance with that?  (I don't think it would be a big problem to do so, that's not an absurd number of things to fix.)

amoudgl:
Hi Ryan, 

Sorry for the delayed response. The operator issue you pointed out is a cpplint bug. They have a method `CheckOperatorSpacing` which checks for `<` in `#include` statements. If it is not a include statement, it checks if `<` is surrounded by any spaces and raises an error in case there is no space around. Following is the code block from google cpplint script responsible for this: 
```python
match = Match(r'^(.*[^\s<])<[^\s=<,]', line)
    if match:
      (_, _, end_pos) = CloseExpression(
          clean_lines, linenum, len(match.group(1)))
      if end_pos <= -1:
        error(filename, linenum, 'whitespace/operators', 3,
              'Missing spaces around <')

```

Earlier I thought of fixing by clang-format which already have Google stylesheet but it points lot of other errors too (#982). So, I would go with manual fixing which is quite doable now. Also, I would not able to work on this PR till 27th April due to end semester examinations and project evaluations. I will try to get back as soon as possible.

rcurtin:
Yeah, maybe we can postprocess the output or something, depending on the format.

Good luck with your exams!  I am looking forward to getting this merged when you are ready. :)

amoudgl:
Hi Ryan! The only way I can see to resolve the operator issue in google cpplint is by changing regex in the main source code. I didn't get your previous statement completely, could you please elaborate a bit? 
 
> maybe we can postprocess the output or something, depending on the format.

Thanks.

rcurtin:
Ah, yeah, no reason to change the main source.  We could instead run cpplint something like:

```
python cpplint.py --extensions=hpp,cpp --filter=\
-whitespace/braces,\ # checks if opening braces are on the same line
-whitespace/newline,\ # reports error when else statement is not the same line with }
-build/header_guard,\ # recommends to append _ after header guard
-build/include_order,\ # should be excluded otherwise it breaks the code
-build/storage_class,\ # checks for static, extern, typedef, etc at the beginning of function declaration
-build/namespaces,\ # enforces to not use namespace using-directives, suggests to use using-declarations instead.
-build/include_what_you_use,\ # e.g. if std::string is found, enforces to include <string>
-legal/copyright,\ #  copyright year and name should be there in license header at the top of file
-readability/casting,\ # discourages to use C type casting like (double) a
-readability/alt_tokens,\ # enforces to use operator "!" instead of "not" even in comments
-readability/todo,\ # # Checks for username in TODO comments like TODO (username) comment
-readability/multiline_string,\  # checks for multiline string and suggests to improve it 
-runtime/explicit,\ # single-parameter constructors should be marked explicit
-runtime/int,\ # Use int16/int64/etc, rather than the C type short
-runtime/references \ # e.g. Is this a non-const reference? If so, make const or use a pointer
$(find ./src/mlpack/core -type f ! -path "./src/mlpack/core/arma_extend/*" ! -path "./src/mlpack/core/boost_backport/*") | \
  grep -v 'Missing spaces around <'
```

That's not a great solution, since that will filter out even valid 'Missing spaces around <' messages.  But it's at least a start, and maybe we could come up with something better that actually looks in the affected file to see if there is a 'template' or 'typedef' on that line, and filters accordingly.  In either case I would be fine with just the basic filtering above; what do you think?

amoudgl:
@rcurtin I agree with the idea of basic filtering you suggested. I took a step forward in this direction and updated the PR with google cpplint style check. There are a few minor changes I did in google cpplint script: 

- cpplint prints to console done processing messages like `Done processing /Users/abhinavmoudgil/mlpack/src/mlpack/core/arma_extend/glue_ccov_meat.hpp`. I removed these as they really don't matter to user and make things crowded in log. 
- If there is any comment on the same line which contains code, cpplint suggests that there should be atleast two spaces between code and comment. For example, if we are adding namespace comments like `} // namespace mlpack`, google cpplint expects two spaces after closing bracket, which is not followed in mlpack. So, I modified the google cpplint script to check just for one space between code and comment in commits 95cfe325d4b595ca2d68a2f90c8735e89e115a1d and 9e8b1f19504591773e7655236716e1a1217d6eca.

I also added a bash script to extract number of style errors from cpplint output. As per the script, Travis build exits with status 1, if number of style errors are greater than 0 otherwise it passes successfully. Kindly review the updates. Thanks!

rcurtin:
Great, looks good to me.  Thanks for the update!  I think I will try to integrate this as Jenkins as a separate check instead of piggybacking on top of Travis.  So probably I will end up cherry-picking the commits from here to go into the `mlpack/jenkins-conf` repository instead, and get it set up as a job that gets run when PRs are submitted.

amoudgl:
It would be great to integrate this with Jenkins. Let me know if any further updates are required from my side. One minor question: should I open a separate PR for fixing style issues pointed by cpplint?

rcurtin:
Sure, if you want to handle the individual style issues, I would be happy to merge those changes, but I agree, a separate PR would be better for that.

zoq:
@amoudgl do you mind to open a PR on the mlpack/jenkins-conf repository that contains all the great work you did around the cpplint style check?

amoudgl:
@zoq @rcurtin I would be happy to open a PR in `mlpack/jenkins-conf` repository. Should I do that and close this one?

zoq:
If you don't mind I'd like to use this PR as a test, so I'd like to keep this open just for now.

amoudgl:
Sure, I will keep this one open and submit a new PR in `mlpack/jenkins-conf` repository too.

zoq:
@amoudgl great, thanks!

zoq:
@amoudgl have you had a chance to open a PR for the style check? if you haven't, no worries, I could do it for you if you like.

amoudgl:
@zoq I worked on fixing style issues pointed by cpplint and I am halfway through it. I would open a PR for that in a day or two but yes, please go ahead with the PR for implementing style check in jenkins as I am really caught up in some other work commitments and it is really needed to quickly implement this style check for upcoming PRs in mlpack. Thanks.

zoq:
@amoudgl happy to help, thanks for all the great work!

rcurtin:
I think since everything is set up with Jenkins and Github now, this PR isn't necessary, so I'll go ahead and close it.  @amoudgl: thanks so much, this is really nice and has needed to happen for... 7 years now :)

PR #982, opened 2017-04-10
clang-format style fixes

amoudgl:
I have fixed all the style issues in this PR using `clang-format` tool. `.clang-format` file from PR #981 has been used as style guide, which configures all the style options. Please review and let me know how this looks. I'll update the `.clang-format` file accordingly. Thanks!


src/mlpack/core/data/binarize.hpp
```diff
@@ -45,10 +45,10 @@ void Binarize(const arma::Mat<T>& input,
   output.copy_size(input);
 
   const int totalElems = static_cast<int>(input.n_elem);
-  const T *inPtr = input.memptr();
-  T *outPtr = output.memptr();
+  const T* inPtr = input.memptr();
+  T* outPtr = output.memptr();
 
-  #pragma omp parallel for
+#pragma omp parallel for
```
rcurtin: I think it's fine to indent the preprocessor macros, I don't think they should be always at 0 indentation.

src/mlpack/core/data/dataset_mapper.hpp
```diff
@@ -121,8 +119,8 @@ class DatasetMapper
   template<typename Archive>
   void Serialize(Archive& ar, const unsigned int /* version */)
   {
-    ar & data::CreateNVP(types, "types");
-    ar & data::CreateNVP(maps, "maps");
+    ar& data::CreateNVP(types, "types");
+    ar& data::CreateNVP(maps, "maps");
```
rcurtin: This looks like an error, it should be `ar & data::CreateNVP(...)`.

src/mlpack/core/data/dataset_mapper_impl.hpp
```diff
@@ -21,17 +21,16 @@ namespace data {
 
 // Default constructor.
 template<typename PolicyType>
-inline DatasetMapper<PolicyType>::DatasetMapper(const size_t dimensionality) :
-    types(dimensionality, Datatype::numeric)
+inline DatasetMapper<PolicyType>::DatasetMapper(const size_t dimensionality)
+    : types(dimensionality, Datatype::numeric)
```
rcurtin: Everything in mlpack has the `:` on the preceding line so that the initialization lists can line up at four-space indentation, so we should move that back to how it was.

src/mlpack/core/data/dataset_mapper_impl.hpp
```diff
@@ -133,15 +130,14 @@ inline Datatype DatasetMapper<PolicyType>::Type(const size_t dimension) const
 template<typename PolicyType>
 inline Datatype& DatasetMapper<PolicyType>::Type(const size_t dimension)
 {
-  if (dimension >= types.size())
-    types.resize(dimension + 1, Datatype::numeric);
+  if (dimension >= types.size()) types.resize(dimension + 1, Datatype::numeric);
```
rcurtin: I don't like one-line ifs, I think they are harder to read because they don't currently appear anywhere.  Can we change this back to the way it was?

src/mlpack/core/data/load_arff_impl.hpp
```diff
@@ -94,13 +93,12 @@ void LoadARFF(const std::string& filename,
       else
       {
         throw std::runtime_error("unknown ARFF annotation '" + (*tok.begin()) +
-            "'");
+                                 "'");
```
rcurtin: I have seen this a few times now... if the following line is double-indented, it's fine; it doesn't need to line up with the previous opening brace.

src/mlpack/core/data/load_csv.hpp
```diff
@@ -114,16 +114,15 @@ class LoadCSV
...
       }
 
       // I guess this is technically a second pass, but that's ok... still the
       // same idea...
       if (MapPolicy::NeedsFirstPass)
       {
         // In this case we must pass everything we parse to the MapPolicy.
-        auto firstPassMap = [&](const iter_type& iter)
-        {
+        auto firstPassMap = [&](const iter_type& iter) {
```
rcurtin: The opening brace should be on the following line here, not moved to be on the same line.

src/mlpack/core/data/load_impl.hpp
```diff
@@ -127,7 +124,8 @@ bool Load(const std::string& filename,
     {
       if (extension == "tsv")
         Log::Warn << "'" << filename << "' is comma-separated, not "
-            "tab-separated!" << std::endl;
+                                        "tab-separated!"
+                  << std::endl;
```
rcurtin: This is a strange change, I think it was fine before.

src/mlpack/core/data/load_impl.hpp
```diff
@@ -231,32 +229,34 @@ bool Load(const std::string& filename,
...
     stringType = "HDF5 data";
-  #if ARMA_VERSION_MAJOR == 4 && \
-      (ARMA_VERSION_MINOR >= 300 && ARMA_VERSION_MINOR <= 400)
+#if ARMA_VERSION_MAJOR == 4 && \
+    (ARMA_VERSION_MINOR >= 300 && ARMA_VERSION_MINOR <= 400)
     Timer::Stop("loading_data");
     if (fatal)
-      Log::Fatal << "Attempted to load '" << filename << "' as HDF5 data, but "
+      Log::Fatal
+          << "Attempted to load '" << filename << "' as HDF5 data, but "
```
rcurtin: I think this also was fine before, and looks a bit weird now.

src/mlpack/core/data/load_vec_impl.hpp
```diff
@@ -70,24 +68,22 @@ bool Load(const std::string& filename,
...
+      *((arma::Mat<eT>*)&vec) = std::move(tmp);
       return true;
     }
   }
   else
   {
     // It's loaded as a column vector.  We can call the move constructor
     // directly.
-    *((arma::Mat<eT>*) &vec) = std::move(tmp);
+    *((arma::Mat<eT>*)&vec) = std::move(tmp);
```
rcurtin: I think the space should still be there between the cast and the argument.

src/mlpack/core/data/map_policies/missing_policy.hpp
```diff
@@ -85,7 +85,8 @@ class MissingPolicy
               MapType& maps,
               std::vector<Datatype>& /* types */)
   {
-    static_assert(std::numeric_limits<T>::has_quiet_NaN == true,
+    static_assert(
+        std::numeric_limits<T>::has_quiet_NaN == true,
```
rcurtin: Here I also don't see a reason to put the condition on a second line, I think it would have been fine on the first.

src/mlpack/core/data/serialization_shim.hpp
```diff
@@ -123,13 +135,13 @@ inline FirstShim<T> CreateNVP(
...
-boost::serialization::nvp<T> CreateNVP(
-    T& t,
-    const std::string& name,
-    typename std::enable_if_t<!HasSerialize<T>::value>* = 0,
-    typename std::enable_if_t<!std::is_pointer<T>::value>* = 0)
+    boost::serialization::nvp<T>
+    CreateNVP(T& t,
+              const std::string& name,
+              typename std::enable_if_t<!HasSerialize<T>::value>* = 0,
+              typename std::enable_if_t<!std::is_pointer<T>::value>* = 0)
```
rcurtin: I know this is a weird section of the code, but I think the changes to indentation here aren't necessary; it should still have zero indentation in the function definition.

src/mlpack/core/data/serialization_shim.hpp
```diff
@@ -35,37 +35,49 @@ struct HasSerializeFunction
...
+template<typename T>
+struct FirstNormalArrayShim;
+template<typename T>
+struct SecondShim;
+template<typename T>
+struct SecondArrayShim;
+template<typename T>
+struct SecondNormalArrayShim;
+template<typename T>
+struct PointerShim;
```
rcurtin: No need to split the lines here in my opinion.

src/mlpack/core/data/serialization_shim.hpp
```diff
@@ -278,8 +292,10 @@ template<typename T>
 struct FirstNormalArrayShim
 {
   //! Construct the first shim with the given objects, length, and name.
-  FirstNormalArrayShim(T* t, const size_t len, const std::string& name) :
-      t(t), len(len), name(name) { }
+  FirstNormalArrayShim(T* t, const size_t len, const std::string& name)
+      : t(t), len(len), name(name)
+  {
+  }
```
rcurtin: I think it's also fine to have an empty function as `{ }` on the line of the definition.

src/mlpack/core/data/binarize.hpp
amoudgl: There is no option for intending preprocessor separately in `.clang-format` file. It just finds statements with `#` and intends accordingly. One thing we can do is add comments `clang-format off` and `clang-format on` in this particular code section?

src/mlpack/core/data/dataset_mapper_impl.hpp
amoudgl: Yes, this is an issue in `clang-format` configuration. We have an option for configuring initializer list in clang-format [style guide](https://clang.llvm.org/docs/ClangFormatStyleOptions.html) i.e. `BreakConstructorInitializersBeforeComma`. I have already set it to`false` which matches with mlpack style guide but here I guess length of constructor is exceeding more than 80 chars, so it intends `:` to next line, not so sure why.

src/mlpack/core/data/dataset_mapper.hpp
amoudgl: I don't know about this in detail but is it not pass by reference? Because clang-format simply treats it as pass by reference and aligns `&` with the datatype returned.

src/mlpack/core/data/load_arff_impl.hpp
amoudgl: There is an option for this `AlignAfterOpenBracket` in the style guide. Three possible values are: 

1. `Align`: Align parameters on the open bracket, e.g.:
   ```c++
   someLongFunction(argument1,
                    argument2);
   ```
2. `DontAlign`:  Don’t align, instead use 
   ContinuationIndentWidth, e.g.:
   ```c++
   someLongFunction(argument1,
       argument2);
   ```
3. `AlwaysBreak`: Always break after an open bracket, if the parameters don’t fit on a single line, e.g.:
   ```c++
   someLongFunction(
       argument1, argument2);
   ```

I can handle this by setting this configuration option `AlignAfterOpenBracket` to `DontAlign` but it will change the alignment in constructor initializer lists and function declarations too with 4 indent width. Should I go ahead with this?

src/mlpack/core/data/load_csv.hpp
amoudgl: I couldn't find any configuration option till now to handle such cases. All the braces alignment have been correctly configured as far as I know with opening braces on a new line. Figuring out any way to handle this.

src/mlpack/core/data/serialization_shim.hpp
amoudgl: There is an option `AlwaysBreakTemplateDeclarations` to handle this. It has two options `true` or `false`. 

```c++
true:                                  false:
template <typename T>          vs.     template <typename T> class C {};
class C {};
```
In most of the mlpack codebase, I have observed that there is always break after template declaration statements. So I set it to `true`, hence this change is there. If I change it to `false`, all the template declaration statements will change. What should we do?

src/mlpack/core/data/dataset_mapper_impl.hpp
amoudgl: Yes, fixed this by setting `AllowShortIfStatementsOnASingleLine` in the configuration file to false.

src/mlpack/core/data/load_vec_impl.hpp
amoudgl: `SpaceAfterCStyleCast` handles this in clang-format. It has two values `true` or `false`. 

If `true`, a space is inserted after C style casts.
```c++
true:                                  false:
(int)i;                        vs.     (int) i;
```
I have set to `false` but it looks a bug in clang-format for dealing with such complex cases (not sure).

amoudgl:
Hi Ryan,

Thanks a lot for taking out time and reviewing this in detail. I have addressed most of your comments, I am still figuring out the rest. Could you please take a look at them?

src/mlpack/core/data/dataset_mapper.hpp
rcurtin: This is a use of `operator&()`, not a reference.

src/mlpack/core/data/dataset_mapper_impl.hpp
rcurtin: Great, thanks!

src/mlpack/core/data/serialization_shim.hpp
rcurtin: I think adding `clang-format` comments all over the code just for the sake of formatting is really confusing and we should definitely try to avoid that.  But I know that makes the clang-format configuration process harder... sorry about that... :(

src/mlpack/core/data/load_vec_impl.hpp
rcurtin: Hmm, maybe if we can show a simple reproducible example we can file a bug upstream.

rcurtin:
I went looking through the clang-format documentation, and I see that really there is not a way to have clang-format "ignore" certain parts of the code.  So I have to wonder, is clang-format the right tool for the job here?  Or should we instead be looking at something more flexible (if something else exists)?

amoudgl:
We can ignore certain parts of the code by adding comments `clang-format off` and `clang-format on` before and after. But we can't ignore any configuration options e.g. there is a configuration option `AlwaysBreakTemplateDeclarations` which can be set as `true` or `false` but if we want that clang-format should ignore all the template declarations leaving them untouched, we can't do it.

amoudgl:
I did some basic search on how to ignore certain checks in clang-format but I couldn't find any way to do that. However, I found that in `clang-tidy` we can pass the checks we want to do and ignore the unnecessary ones. Should I try it?

src/mlpack/core/data/serialization_shim.hpp
amoudgl: Yes I agree. :) Need to some more hard work on this but I will try my best to come up with some decent solution.

rcurtin:
If we are going to use cpplint instead of clang-format I will go ahead and close this PR.

PR #985, opened 2017-04-12
new constructors of decision_tree to avoid unnecessary copies

### LINKED ISSUE #951: mlpack_decision_tree memory exploded on dataset "reuters"
I have just finished benchmark script of mlpack_decision_tree, and I noticed that on reuters dataset the memory increases very fast and I have to drop this dataset to run benchmark on my machine. Maybe there is something wrong? (this dataset is extreme because it has so many features)

ThyrixYang:
fix #951


src/mlpack/methods/decision_tree/decision_tree_impl.hpp
```diff
@@ -56,6 +56,98 @@ DecisionTree<FitnessFunction,
...
+             ElemType,
+             NoRecursion>::DecisionTree(MatType&& data,
+                                        const data::DatasetInfo& datasetInfo,
+                                        arma::Row<size_t>&& labels,
+                                        const size_t numClasses,
+                                        const size_t minimumLeafSize)
+{
+  // move the data to avoid copy
+  MatType tmpData = std::move(data);
+  arma::Row<size_t> tmpLabels = std::move(labels);
```
rcurtin: I realized that this parameter `MatType&&` is a universal reference and can be either an lvalue (`MatType&`) or an rvalue (`MatType&&`). This means that we have the opportunity to do perfect forwarding here. If you changed the labels to also be a universal reference (i.e. `LabelsType&& labels`) then you could do:

```
MatType tmpData(std::forward<MatType>(data));
arma::Row<size_t> tmpLabels(std::forward<LabelsType>(labels));
```

This will copy the data if a reference is given and move the data if an rvalue reference is given, which is exactly the behavior we want.  Then we can remove the other constructor overload that takes a `const MatType&`.  You could do the same with `Train()` too.

I've done some changes just now in another part of the codebase that do a similar thing in 3f7d138, maybe you can look at those as an example.  I regret to say that I had not fully learned about perfect forwarding and `std::forward` in C++ until just today. Learning about it earlier probably could have saved me some pain in the past. Either way I just ordered Scott Meyers' 'Effective Modern C++' so I don't get caught out by stuff like this again. :)

ThyrixYang: I think const MatType& can't be replaced with MatType&& like following:
```cpp
  template<typename MatType>
  DecisionTree(MatType&& data,
               const size_t numClasses,
               const size_t minimumLeafSize = 10){
MatType tmpData(std::forward<MatType>(data));
}
```
In this case, if I pass a lvalue mat<double>.
Then the MatType will be deduced as mat<double>&, not mat<double>.
When I tried this:
```cpp
template<typename T>
vector<int> F(T&& v) {
	T a = std::forward<T>(v);
	a.push_back(2);
	return a;
}
int main() {
	vector<int> x;
	x.push_back(33);
	vector<int> b = F(x);
	cout << x.size() << endl;
	cout << b.size() << endl;
	return 0;
}
```
It shows both x.size() and b.size() are 2.
Maybe I've not understood correctly? :)

rcurtin: The `std::vector<>` example you showed looks like it is working fine to me.

I see the problem with `MatType` being deduced as `mat&`.  We could use `std::remove_reference` here:

```
typedef std::remove_reference_t<MatType> TrueMatType;
TrueMatType tmpData(std::forward(data));
```

What do you think?

ThyrixYang: I'm reading *Effective Modern C++* today, I think it's a nice solution.

And I think it's a common consideration in mlpack. Whether it's suggested to write a rvalue constructor like this whenever we can't use a single const& without copy?  Maybe this will be added to DesignGuidelines?

rcurtin: Yeah, I think it would be nice to adapt some of the mlpack code to use universal references like this, but there are a few problems that can make it tricky in some places:

 * universal references will match anything, so often some SFINAE may need to be done in order to restrict the matching to, e.g., only Armadillo types or something like this
 * since the universal reference requires the type to be deduced for the method itself, then you have to actually have a separate template parameter for the input matrix; i.e. the `NeighborSearch` class has a template parameter `MatType`, but you can't do this:

```cpp
template<..., typename MatType, ...>
class NeighborSearch
{
  NeighborSearch(MatType&& data, ...); // not a universal reference
};
```

but instead you have to do something like

```cpp
template<..., class MatType, ...>
class NeighborSearch
{
  template<typename MatInType>
  NeighborSearch(MatInType&& m, ...); // universal reference, but also requires SFINAE to restrict to Armadillo types
};
```

There may be some other minor concerns I haven't thought of too.  Anyway, I do agree, it would be nice to look through the mlpack codebase again and see what we can adapt to use universal references.  I think if I have time, I might start by looking at everything in `src/mlpack/core/tree/` to start with; the tree constructors could definitely be condensed using universal references.

ThyrixYang:
I've changed to universal reference.
Seems mlpack is not using c++14 now, so I can't use remove_reference_t

src/mlpack/methods/decision_tree/decision_tree_impl.hpp
```diff
@@ -286,48 +413,50 @@ void DecisionTree<FitnessFunction,
       numChildren = NumericSplit::NumChildren(classProbabilities, *this);
 
     // Calculate all child assignments.
-    arma::Col<size_t> childAssignments(data.n_cols);
+    arma::Row<size_t> childAssignments(data.n_cols);
```
rcurtin: We are only using elements `begin` to `begin + count` here, I think we could save memory by reducing this to have size `count` instead of `data.n_cols`.

src/mlpack/methods/decision_tree/decision_tree.hpp
```diff
@@ -264,6 +264,86 @@ class DecisionTree :
...
+   * @param labels Labels for each training point.
+   * @param numClasses Number of classes in the dataset.
+   * @param minimumLeafSize Minimum number of points in each leaf node.
+   */
+  template<typename MatType>
+  void Train(MatType& data,
+             const size_t begin,
+             const size_t count,
+             arma::Row<size_t>& labels,
+             const size_t numClasses,
```
rcurtin: Since these are only used by children, do you think we should mark them `private`?

src/mlpack/methods/decision_tree/decision_tree_impl.hpp
```diff
@@ -247,18 +366,26 @@ void DecisionTree<FitnessFunction,
   // numericAux and categoricalAux (and clear them later if we make not split),
   // and use classProbabilities as auxiliary information.  Later we'll overwrite
   // classProbabilities to the empirical class probabilities if we do not split.
-  double bestGain = FitnessFunction::Evaluate(labels, numClasses);
+
+  double bestGain = FitnessFunction::Evaluate(
+    labels.subvec(begin, begin + count - 1), numClasses);
```
rcurtin: Minor comment---this should be indented twice since it is a continuation of a previous line.

src/mlpack/methods/decision_tree/decision_tree_impl.hpp
```diff
@@ -247,18 +366,26 @@ void DecisionTree<FitnessFunction,
...
+        numClasses, minimumLeafSize,
+        classProbabilities, *this);
     else if (datasetInfo.Type(i) == data::Datatype::numeric)
-      dimGain = NumericSplit::SplitIfBetter(bestGain, data.row(i), labels,
-          numClasses, minimumLeafSize, classProbabilities, *this);
+      dimGain = NumericSplit::SplitIfBetter(bestGain, 
+        data.cols(begin, begin + count - 1).row(i), 
+        labels.subvec(begin, begin + count - 1), 
+        numClasses, minimumLeafSize,
+        classProbabilities, *this);
```
rcurtin: Same comment about indentation through here. :)

src/mlpack/methods/decision_tree/decision_tree.hpp
ThyrixYang: I placed them under the "private:" at line 236, so I think there is no need to mark them again?

rcurtin: Oh!  Right.  I see that now.  Sorry...

rcurtin:
Thanks so much for taking the time to do this, this is a very nice improvement!

PR #986, opened 2017-04-12
CodeAi fixed 4 null pointer dereferences in hoeffding_tree_model.cpp

C0deAi:
Potential null dereferences of other.giniHoeffdingTree, other.giniBinaryTree, other.infoHoeffdingTree, other.infoBinaryTree on lines 72, 74, 76, and 78 respectively.  CodeAi fixes these issues by tightening the if condition to verify that each pointer variable is not null before use in the constructor.  

See the bug reports below for more information:

[CodeAiBugReport_mlpack_src_mlpack_methods_hoeffding_trees_hoeffding_tree_model_cpp_78.pdf](https://github.com/mlpack/mlpack/files/917104/CodeAiBugReport_mlpack_src_mlpack_methods_hoeffding_trees_hoeffding_tree_model_cpp_78.pdf)
[CodeAiBugReport_mlpack_src_mlpack_methods_hoeffding_trees_hoeffding_tree_model_cpp_76.pdf](https://github.com/mlpack/mlpack/files/917101/CodeAiBugReport_mlpack_src_mlpack_methods_hoeffding_trees_hoeffding_tree_model_cpp_76.pdf)
[CodeAiBugReport_mlpack_src_mlpack_methods_hoeffding_trees_hoeffding_tree_model_cpp_74.pdf](https://github.com/mlpack/mlpack/files/917102/CodeAiBugReport_mlpack_src_mlpack_methods_hoeffding_trees_hoeffding_tree_model_cpp_74.pdf)
[CodeAiBugReport_mlpack_src_mlpack_methods_hoeffding_trees_hoeffding_tree_model_cpp_72.pdf](https://github.com/mlpack/mlpack/files/917103/CodeAiBugReport_mlpack_src_mlpack_methods_hoeffding_trees_hoeffding_tree_model_cpp_72.pdf)

### DIFF
--- a/src/mlpack/methods/hoeffding_trees/hoeffding_tree_model.cpp
+++ b/src/mlpack/methods/hoeffding_trees/hoeffding_tree_model.cpp
@@ -68,13 +68,13 @@ HoeffdingTreeModel& HoeffdingTreeModel::operator=(
 
   // Create the right tree.
   type = other.type;
-  if (type == GINI_HOEFFDING)
+  if (other.giniHoeffdingTree && (type == GINI_HOEFFDING))
     giniHoeffdingTree = new GiniHoeffdingTreeType(*other.giniHoeffdingTree);
-  else if (type == GINI_BINARY)
+  else if (other.giniBinaryTree && (type == GINI_BINARY))
     giniBinaryTree = new GiniBinaryTreeType(*other.giniBinaryTree);
-  else if (type == INFO_HOEFFDING)
+  else if (other.infoHoeffdingTree && (type == INFO_HOEFFDING))
     infoHoeffdingTree = new InfoHoeffdingTreeType(*other.infoHoeffdingTree);
-  else if (type == INFO_BINARY)
+  else if (other.infoBinaryTree && (type == INFO_BINARY))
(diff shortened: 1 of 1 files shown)


rcurtin:
Yes, this is a legitimate issue.  The `HoeffdingTreeModel` class is designed in such a way that it's not supposed to be used without calling `BuildModel()` first, but it is indeed susceptible to the null pointer reference found here.

A deeper issue is that the `HoeffdingTreeModel` assignment operator isn't safe for self-assignment, but given that this class is only meant to be used carefully by the command-line programs and bindings, I'm not going to worry too much about that for now.  This fix doesn't actually handle the self-assignment problem, but it does handle the situation where someone calls, e.g.,

```
HoeffdingTreeModel m, n;
m = n; // would previously cause NULL dereference; now fixed
```

Before I merge this... one question... how can I add CodeAi to the list of contributors?  I don't imagine it has an email address or a name other than CodeAi. :)  Let me know what you'd like me to do.

C0deAi:
For now, its a GitHub account like any other.  GitHub ID: C0deAi email: benjamin.bales@assrc.us  Eventually, we plan to offer it as a GitHub Integration.

keon:
This is so cool :)

PR #989, opened 2017-04-25
Implement two classical control problems for testing reinforcement learning method

ShangtongZhang:


ShangtongZhang:
```
*****************************************unknown location(0): fatal error in "LogChebychevApproxSdp": std::runtime_error: chol(): failed to converge
/home/travis/build/mlpack/mlpack/src/mlpack/tests/sdp_primal_dual_test.cpp(217): last checkpoint
**********
```

src/mlpack/methods/reinforcement_learning/environment/cart_pole.hpp
```diff
@@ -0,0 +1,182 @@
...
+
+#ifndef MLPACK_METHODS_RL_ENVIRONMENT_CART_POLE_HPP
+#define MLPACK_METHODS_RL_ENVIRONMENT_CART_POLE_HPP
+
+#include <mlpack/prereqs.hpp>
+
+namespace mlpack {
+namespace rl {
+
+namespace cart_pole_details {
```
zoq: I think it would be a good idea to make the parameters part of the CartPole  constructor, that way, someone could use the environment with e.g. a different mass or gravity. I know it's somewhat unlikely that someone is going for it, but at least there is the possibility to do that without modifying the codebase if you just link against mlpack.

src/mlpack/methods/reinforcement_learning/environment/cart_pole.hpp
```diff
@@ -0,0 +1,182 @@
...
+  {
+   public:
+    //! Construct a state instance
+    State() : data(4) { }
+
+    //! Construct a state instance from given data
+    State(arma::colvec data) : data(data) { }
+
+    //! Get position
+    double X() const
```
zoq: Maybe we can use some more descriptive names for the methods, what do you think? Also, minor style idea, we can use one line if it fit's, we do this all the time for the parameter functions.

src/mlpack/methods/reinforcement_learning/environment/cart_pole.hpp
```diff
@@ -0,0 +1,182 @@
...
+   * Each state is a tuple of (position, velocity, angle, angular velocity)
+   */
+  class State
+  {
+   public:
+    //! Construct a state instance
+    State() : data(4) { }
+
+    //! Construct a state instance from given data
+    State(arma::colvec data) : data(data) { }
```
zoq: Would pass the state by reference, to avoid a copy and we can also make it const.

src/mlpack/methods/reinforcement_learning/environment/cart_pole.hpp
zoq: Another idea is to return the complete state instead of providing one function for each parameter, because it's basically used internally. That way, we could provide the same function set (State) for each environment.

src/mlpack/methods/reinforcement_learning/environment/cart_pole.hpp
```diff
@@ -0,0 +1,182 @@
...
+
+  /**
+   * Dynamics of Cart Pole
+   * Get next state and next action based on current state and current action
+   * @param state Current state
+   * @param action Current action
+   * @param nextState Next state
+   * @param reward Reward is always 1
+   */
+  void Sample(const State& state, const Action::Actions& action,
```
zoq: Do you think it's a good idea to provide a default state and nextState parameter and to return the reward instead of passing it by reference?

zoq:
Don't worry about the test, many mlpack tests are probabilistic, that fail in some cases e.g. if they don't converge in the specified number of iterations.

src/mlpack/methods/reinforcement_learning/environment/cart_pole.hpp
ShangtongZhang: It looks like a good idea

src/mlpack/methods/reinforcement_learning/environment/cart_pole.hpp
ShangtongZhang: Previously I don't want to store a reference because sometimes the `state` needs to be scaled (i.e. scale to [-1, 1]) and I want to cache the scaled state directly in the `data` variable. But after completing DQN I realize this is not quite necessary, and even we want to do this, it's better to do this by other class rather than `State` class. I will change it to a reference.

src/mlpack/methods/reinforcement_learning/environment/cart_pole.hpp
ShangtongZhang: Yeah I can change it to some more descriptive methods and in the meantime add the function to get/set complete state

src/mlpack/methods/reinforcement_learning/environment/cart_pole.hpp
ShangtongZhang: I think it's good to provide a default dummy state for `nextState` in case user only cares about reward and set reward to return value.

ShangtongZhang:
I refactored the implementation, looking forward to any feedback

src/mlpack/tests/rl_environment_test.cpp
```diff
@@ -0,0 +1,46 @@
...
+#include "test_tools.hpp"
+
+using namespace mlpack;
+using namespace mlpack::rl;
+
+BOOST_AUTO_TEST_SUITE(RLEnvironmentTest)
+
+BOOST_AUTO_TEST_CASE(MountainCarTest)
+{
+  const auto task = MountainCar();
```
zoq: I'd like to avoid the use of auto here because I think in this case it doesn't enhance code readability; in fact, a user could take a look at the test case to get an impression on how to use the code and might wondering what action, reward, etc. is. Of course, he could look into the code or get it from the context but since the type isn't complex we could just use the type instead of auto here. Let me know what you think.

src/mlpack/tests/rl_environment_test.cpp
```diff
@@ -0,0 +1,46 @@
...
+
+#include <boost/test/unit_test.hpp>
+#include "test_tools.hpp"
+
+using namespace mlpack;
+using namespace mlpack::rl;
+
+BOOST_AUTO_TEST_SUITE(RLEnvironmentTest)
+
+BOOST_AUTO_TEST_CASE(MountainCarTest)
```
zoq: I would call the test ``SimpleMountainCarTest`` since we might add some more tests here for the different environments.

src/mlpack/methods/reinforcement_learning/environment/mountain_car.hpp
```diff
@@ -0,0 +1,172 @@
...
+    state.Position() = arma::as_scalar(arma::randu(1)) * 0.2 - 0.6;
+    return state;
+  }
+
+  /**
+   * Whether given state is terminal state
+   * @param state desired state
+   * @return true if @state is terminal state, otherwise false
+   */
+  bool IsTerminal(const State& state) const { return std::abs(state.Position() - positionMax) <= 1e-5; }
```
zoq: Another picky format issue: "Lines should be no more than 80 characters wide." for more informations take a look at the DesignGuidelines (https://github.com/mlpack/mlpack/wiki/DesignGuidelines).

src/mlpack/methods/reinforcement_learning/environment/mountain_car.hpp
```diff
@@ -0,0 +1,172 @@
...
+
+/**
+ * Implementation of Mountain Car task
+ */
+class MountainCar
+{
+ public:
+
+  /**
+   * Implementation of state of Mountain Car
```
zoq: Picky format issue, please use proper punctuation for all comments and parameter descriptions.

src/mlpack/methods/reinforcement_learning/environment/mountain_car.hpp
```diff
@@ -0,0 +1,172 @@
+/**
+ * @file mountain_car.hpp
+ * @author Shangtong Zhang
+ *
+ * This file is an implementation of Mountain Car task
+ * https://gym.openai.com/envs/MountainCar-v0
+ *
+ * TODO: refactor to OpenAI interface
```
zoq: I'm not sure that we should fully compliant with the OpenAI Interface. What do you think?

src/mlpack/methods/reinforcement_learning/environment/mountain_car.hpp
```diff
@@ -0,0 +1,172 @@
...
+    State(const arma::colvec& data): data(data) { }
+
+    //! Encode the state to a column vector
+    const arma::colvec& Encode() const { return data; }
+
+    /**
+     * Set the internal data to given value
+     * @param data desired internal data
+     */
+    void Set(const arma::colvec& data) { this->data = data; }
```
zoq: We don't use setter/getter functions, I guess a simple solution would be to use ``arma::colvec& Data() { return data; }`` here, and do: ``task.Data() = data;``.

src/mlpack/methods/reinforcement_learning/environment/mountain_car.hpp
ShangtongZhang: Oh maybe this comment is kind of misleading, I didn't mean to have same API as OpenAI gym. I just want to provide an alternative implementation, i.e. the logic of `Sample` can also be done via OpenAI gym if user specifies this.

src/mlpack/tests/rl_environment_test.cpp
ShangtongZhang: yeah I agree with you, `auto` makes code hard to understand

ShangtongZhang:
Hopefully it's now ready to merge

src/mlpack/methods/reinforcement_learning/environment/mountain_car.hpp
```diff
@@ -0,0 +1,172 @@
...
+    State nextState;
+    return Sample(state, action, nextState);
+  }
+
+  /**
+   * Initial position is randomly generated within [-0.6, -0.4].
+   * Initial velocity is 0.
+   * @return Initial state for each episode.
+   */
+  State InitialSample() const
```
zoq: We could avoid a copy if we pass the state by reference if the State is complex this could speed things up, but I'm not sure that is necessary here. What do you think?

ShangtongZhang: I think it will be optimized by `Return Value Optimization` even with `-o0` unless user specifies `-fno-elide-constructors`. In the other hand, by using `const State & s = InitialSample()` user can explicitly extend the lifetime of the local variable, so there is also no overhead. If we refactor it to `const State& InitialSample() const`, it seems to be an undefined behavior per [this](http://stackoverflow.com/questions/1465851/returning-const-reference-to-local-variable-from-a-function) and clang does gives an warning for it.

ShangtongZhang: I think it will be optimized by `Return Value Optimization` even with `-o0` unless user specifies `-fno-elide-constructors`. In the other hand, by using `const State & s = InitialSample()` user can explicitly extend the lifetime of the local variable, so there is also no overhead. If we refactor it to `const State& InitialSample() const`, it seems to be an undefined behavior per [this](http://stackoverflow.com/questions/1465851/returning-const-reference-to-local-variable-from-a-function) and clang does gives an warning for it.

zoq: Thanks for your thoughts and clarification; besides some minor style issues, I think this is ready to be merged. I'll go and fix the issues once I merge this in.

ShangtongZhang: That's cool! Thanks for your review!

zoq:
Merged, I made some minor style fixes in: 859d8139a1d433d284a17e8a70b5a3c5ce4df99f let me know if I missed something.

PR #990, opened 2017-04-25
Stochastic Gradient Descent with Restarts and Snapshot ensembles

zoq:
Implementation of the Stochastic Gradient Descent with Restarts (SGDR) as described in: "SGDR: Stochastic Gradient Descent with Warm Restarts" by I. Loshchilov et al and the Snapshot ensembles technique described in: "Snapshot ensembles: Train 1, get m for free" by G. Huang et al. The implementation reuses the mini batch SGD optimizer class and the main idea is implemented as update and decay policy.

### DIFF
--- a/src/mlpack/core/optimizers/sgdr/cyclical_decay.hpp
+++ b/src/mlpack/core/optimizers/sgdr/cyclical_decay.hpp
@@ -63,21 +63,20 @@ class CyclicalDecay
       constStepSize(stepSize),
       nextRestart(epochRestart),
       batchRestart(0),
-      epochBatches(numFunctions / (double) batchSize)
+      epochBatches(numFunctions / (double) batchSize),
+      epoch(0)
   { /* Nothing to do here */ }
 
   /**
    * This function is called in each iteration after the policy update.
    *
-   * @param stepSize The stepSize to be adjusted.
-   * @param epoch The current epoch.
-   * @param batch The current batch.
-   * @param iterate Function parameters.
+   * @param iterate Parameters that minimize the function.
+   * @param stepSize Step size to be used for the given iteration.
+   * @param gradient The gradient matrix.
    */
-  void Update(double& stepSize,
-              const size_t epoch,
-              const size_t /* batch */,
-              const arma::mat& /* iterate */)
+  void Update(arma::mat& /* iterate */,
+              double& stepSize,
+              const arma::mat& /* gradient */)
   {
     // Time to adjust the step size.
     if (epoch >= epochRestart)
@@ -101,6 +100,8 @@ class CyclicalDecay
       // Update the time for the next restart.
       nextRestart += epochRestart;
     }
+
+    epoch++;
   }
 
   //! Get the step size.
@@ -130,6 +131,9 @@ class CyclicalDecay
 
   //! Locally-stored restart fraction.
   double epochBatches;
+
+  //! Locally-stored epoch.
+  size_t epoch;
 };
 
 } // namespace optimization
--- a/src/mlpack/core/optimizers/sgdr/cyclical_decay.hpp
+++ b/src/mlpack/core/optimizers/sgdr/cyclical_decay.hpp
@@ -1,0 +1,138 @@
+/**
+ * @file cyclical_decay.hpp
+ * @author Marcus Edel
+ *
+ * Definition of the warm restart technique (SGDR) described in:
+ * "SGDR: Stochastic Gradient Descent with Warm Restarts" by
+ * I. Loshchilov et al.
+ *
+ * You should define your own decay update that looks like EmptyDecay.
+ *
+ * mlpack is free software; you may redistribute it and/or modify it under the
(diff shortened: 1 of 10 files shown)


src/mlpack/core/optimizers/sgdr/cyclical_decay.hpp
```diff
@@ -0,0 +1,140 @@
...
+ *
+ * For more information, please refer to:
+ *
+ * @code
+ * @article{Loshchilov2016,
+ *   title   = {Learning representations by back-propagating errors},
+ *   author  = {Ilya Loshchilov and Frank Hutter},
+ *   title   = {{SGDR:} Stochastic Gradient Descent with Restarts},
+ *   journal = {CoRR},
+ *   year    = {2016}
```
rcurtin: Do you think we should add a URL here because this is on arXiv?

src/mlpack/core/optimizers/sgdr/sgdr.hpp
```diff
@@ -0,0 +1,256 @@
...
+ * }
+ * @endcode
+ *
+ * @code
+ * @inproceedings{Huang2017,
+ *   title     = {Snapshot ensembles: Train 1, get m for free},
+ *   author    = {Gao Huang, Yixuan Li, Geoff Pleiss, Zhuang Liu,
+ *                John E. Hopcroft, and Kilian Q. Weinberger},
+ *   booktitle = {Proceedings of the International Conference on Learning
+ *                Representations (ICLR)},
```
rcurtin: ICLR proceedings are just on arXiv, I think?  In that case it may also be useful to add a URL here.

src/mlpack/core/optimizers/sgdr/sgdr.hpp
```diff
@@ -0,0 +1,256 @@
...
+ * @tparam UpdatePolicyType Update policy used during the iterative update
+ *         process. By default the vanilla update policy
+ *         (see mlpack::optimization::VanillaUpdate) is used.
+ * @tparam DecayPolicyType Decay policy used during the iterative update
+ *         process to adjust the step size (CyclicalDecay or SnapshotEnsembles).
+ */
+template<
+    typename DecomposableFunctionType,
+    typename UpdatePolicyType = MomentumUpdate,
+    typename DecayPolicyType = CyclicalDecay
```
rcurtin: It looks like `DecayPolicyType` can only be `SnapshotEnsembles` or `CyclicalDecay`, and the implementation is different in either case.  So do you think it would be better to simply have two different classes?  I guess that might obviate the need for `DecayPolicyType` in general too; not sure.

src/mlpack/core/optimizers/sgdr/sgdr_impl.hpp
```diff
@@ -0,0 +1,193 @@
...
+  }
+
+  // If a user changed the batch size we have to update the restart fraction
+  // of the cyclical decay instantiation.
+  if (optimizer.BatchSize() != batchSize)
+  {
+    batchSize = optimizer.BatchSize();
+    optimizer.DecayPolicy().EpochBatches() = function.NumFunctions() /
+        double(batchSize);
+  }
```
rcurtin: One possibility here might be to add a function to `DecayPolicy()` called something like `UpdateParameters(OptimizerType&)`, which for `CyclicalDecay` or `SnapshotEnsembles` could do the same thing as this code and for `NoDecay` could do nothing.  This could help reduce the need for two separate implementations here.

src/mlpack/core/optimizers/sgdr/sgdr_impl.hpp
```diff
@@ -0,0 +1,193 @@
...
+    {
+      iterate += optimizer.DecayPolicy().Snapshots()[i];
+    }
+    iterate /= (optimizer.DecayPolicy().Snapshots().size() + 1);
+
+    // Calculate final objective.
+    overallObjective = 0;
+    for (size_t i = 0; i < function.NumFunctions(); ++i)
+      overallObjective += function.Evaluate(iterate, i);
+  }
```
rcurtin: But this part I am less sure of how to move into `SnapshotEnsembles`.  The only thing I could think of is to make a template specialization for `SGD<..., SnapshotEnsembles>` and `MiniBatchSGD<..., SnapshotEnsembles>` but that would be ugly and weird because we can't do partial template member specialization.

src/mlpack/core/optimizers/sgdr/cyclical_decay.hpp
zoq: Good idea, I thought I already did this but looks like I didn't.

src/mlpack/core/optimizers/sgdr/sgdr.hpp
zoq: I agree, two separate classes one for ``SGDR`` and another one for ``SnapshotSGDR`` is the cleanest way.

PR #991, opened 2017-04-28
Build against Visual Studio 2015 and Visual Studio 2017

zoq:
Appveyor matrix build for Visual Studio 2015 and Visual Studio 2017.


rcurtin:
Will this fail if the total job time is greater than an hour? Or does the limit get increased to two hours?  I can't seem to determine from the AppVeyor documentation.

Either way looks like a nice change to me!

zoq:
I'm not sure but since they don't run in parallel, I think the time limit is 60 min for each so 2 hours.

PR #994, opened 2017-05-05
Reduce compilation time - PR for #722

Farqd:
Change some includes to reduce compilation time. 
PR for issue #722

* Remove #include \<queue\> and \<iostream\> from prereqs.hpp. Add includes in files that need them. 
* Use AdaBoost forward declaration in adaboost_model
* Move some includes from header to cpp file in util/cli.cpp
* Remove unnecessary includes in some other files

### DIFF
--- a/src/mlpack/methods/adaboost/adaboost_model.hpp
+++ b/src/mlpack/methods/adaboost/adaboost_model.hpp
@@ -9,6 +9,7 @@
 
 #include <mlpack/core.hpp>
 
+// Use forward declaration instead of include to accelerate compilation.
 class AdaBoost;
 
 namespace mlpack {
--- a/src/mlpack/methods/adaboost/adaboost_model.hpp
+++ b/src/mlpack/methods/adaboost/adaboost_model.hpp
@@ -8,7 +8,8 @@
 #define MLPACK_METHODS_ADABOOST_ADABOOST_MODEL_HPP
 
 #include <mlpack/core.hpp>
-#include "adaboost.hpp"
+
+class AdaBoost;
 
 namespace mlpack {
 namespace adaboost {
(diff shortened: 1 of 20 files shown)


src/mlpack/methods/adaboost/adaboost_model.hpp
```diff
@@ -8,7 +8,8 @@
 #define MLPACK_METHODS_ADABOOST_ADABOOST_MODEL_HPP
 
 #include <mlpack/core.hpp>
-#include "adaboost.hpp"
+
+class AdaBoost;
```
rcurtin: Could you please add a comment here indicating that `adaboost.hpp` isn't included to accelerate compilation?

src/mlpack/tests/serialization.hpp
```diff
@@ -22,7 +22,6 @@
 #include <mlpack/core.hpp>
 
 #include <boost/test/unit_test.hpp>
-#include "test_tools.hpp"
```
rcurtin: It's probably a good idea to include this anyway, in case `CheckMatrices()` is ever used in this test.  The file is very short anyway.  (For what it's worth, it looks like some of the reverse compatibility code in `test_tools.hpp` is very out of date and can be removed---the minimum Boost version is greater than 1.36, so that bit can go if you would like to remove it.)

rcurtin:
Thanks for the contribution; it looks like there is some minor speedup in Travis with this.  If you'd like to add your name to the list of contributors in `src/mlpack/core.hpp` and `COPYRIGHT.txt` and address the couple comments I made, I can go ahead and merge this.  Thanks!

rcurtin:
Thanks, much appreciated!

PR #995, opened 2017-05-09
Refactor FFN

ShangtongZhang:
- Support copy/assignment CTOR

- Add a new function `SetTrainingData` for async rl methods
This is mainly for compute gradients
```
    learningNetwork.SetTrainingData(state, target);
    arma::mat gradients;
    learningNetwork.Gradient(learningNetwork.Parameters(), 0, gradients);
```

- Expose a private function `ResetParameters` to public


src/mlpack/methods/ann/init_rules/gaussian_init.hpp
```diff
@@ -76,10 +76,10 @@ class GaussianInitialization
 
  private:
   //! Mean of the gaussian.
-  const double mean;
+  double mean;
```
ShangtongZhang: class with const member variable doesn't support assignment operator, so I just removed the const qualifier.

src/mlpack/methods/ann/ffn_impl.hpp
```diff
@@ -379,6 +384,48 @@ void FFN<OutputLayerType, InitializationRuleType>::Serialize(
...
+  {
+    this->network.push_back(boost::apply_visitor(copyVisitor, network.network[i]));
+  }
+};
+
+template<typename OutputLayerType, typename InitializationRuleType>
+FFN<OutputLayerType, InitializationRuleType>&
+FFN<OutputLayerType, InitializationRuleType>::operator = (const FFN& network)
+{
+  // Replicate layers from source network
```
zoq: We should deconstruct the network first, to avoid memory leaks. Also maybe it's a good idea to use the  copy-and-swap idiom here, could also be useful if we provide a move constructor, what do you think?

src/mlpack/methods/ann/ffn.hpp
```diff
@@ -177,6 +186,20 @@ class FFN
...
   arma::mat& Parameters() { return parameter; }
 
+  /**
+   * Make the network ready for the given training data.
+   * This function won't actually trigger training process.
+   *
+   * @param predictors Input training variables.
+   * @param responses Outputs results from input training variables.
+   */
+  void SetTrainingData(const arma::mat& predictors, const arma::mat& responses);
```
zoq: Do you think instead of using ``SetTrainingData`` followed by ``Gradient``we could provide another ``Gradient`` function, that takes the ``predictors`` and ``responses``?

src/mlpack/methods/ann/ffn_impl.hpp
ShangtongZhang: It's the first time I hear about copy-and-swap idiom. It's really an interesting and useful design! Also I will look into move CTOR.

src/mlpack/methods/ann/ffn.hpp
ShangtongZhang: That sounds good, more easy to use ^.^

ShangtongZhang:
Hope it's ready to merge now

src/mlpack/methods/ann/ffn.hpp
```diff
@@ -215,6 +217,15 @@ class FFN
   void Forward(arma::mat&& input);
 
   /**
+   * Prepare the network for the given training data.
+   * This function won't actually trigger training process.
+   *
+   * @param predictors Input training variables.
+   * @param responses Outputs results from input training variables.
+   */
+  void SetTrainingData(const arma::mat& predictors, const arma::mat& responses);
```
zoq: Do we have to provide this function? I guess it's only used internally, I'd like to keep the function set minimal, so maybe we can delete the function.

src/mlpack/tests/feedforward_network_test.cpp
```diff
@@ -410,4 +410,18 @@ BOOST_AUTO_TEST_CASE(DropConnectNetworkTest)
       (dataset, labels, dataset, labels, 2, 10, 50, 0.2);
 }
 
+/**
+ * Test miscellaneous things of FFN.
+ */
```
rcurtin: Could you be a little more specific in this comment please? :)

src/mlpack/methods/ann/ffn.hpp
ShangtongZhang: It's a common part of `Gradient()` and `Train()`, I marked it as private. Without it we have to do copy-paste for some code snippet. I'm not sure which is preferred. If we decide to keep it, maybe we can change the name of this function. I don't think current name is so elegant but I can't come up with a better one..

src/mlpack/tests/feedforward_network_test.cpp
ShangtongZhang: Sure

src/mlpack/methods/ann/ffn.hpp
zoq: You are right, avoiding code duplication. I think the function name is misleading, it's not necessarily the training set, what you do you think about ``ResetData(...)``?

ShangtongZhang: That looks good. I have changed the name to `ResetData`.

src/mlpack/methods/ann/ffn.hpp
```diff
@@ -217,13 +217,13 @@ class FFN
...
 
   /**
-   * Prepare the network for the given training data.
+   * Prepare the network for the given data.
    * This function won't actually trigger training process.
    *
-   * @param predictors Input training variables.
-   * @param responses Outputs results from input training variables.
+   * @param predictors Input data variables.
+   * @param responses Outputs results from input datagit variables.
```
zoq: Looks like a combination of ``data`` and ``git``.

src/mlpack/methods/ann/ffn.hpp
```diff
@@ -219,6 +248,12 @@ class FFN
    */
   void ResetGradients(arma::mat& gradient);
 
+  /**
+   * Swap the content of this network with given network.
+   * @param network Desired source network.
```
zoq: I know it is a pedantic request, but could you add newline between the description and the function parameters?

ShangtongZhang: Oh, I didn't realize this before :)

zoq:
Thanks for the great work!

PR #996, opened 2017-05-10
Readme header

zoq:
I'm not sure which version looks better the one with a header or the one without.

With header: https://github.com/zoq/mlpack/tree/readme_header
Without header: https://github.com/mlpack/mlpack


stites:
As a new watcher I'm a fan of the header -- I like that there's a reminder of community, help, and irc directly in the repo.

rcurtin:
@stites: I agree, I also like the header.  It helps emphasize that there is a complete mlpack website instead of just the Github page, which I think the previous README did not.  I vote to merge it. :)

mentekid:
I like the pointers to the website and IRC too :)

zoq:
Thanks for the feedback, since you all like the idea, I'll go and merge it.

PR #1000, opened 2017-05-17
Fix a bug in gaussian init

ShangtongZhang:
Please find more discussions about this [here](https://github.com/mlpack/mlpack/pull/934).
Briefly speaking, it seems the init function goes wrong when the desired mat has pre-allocated memory. But I don't know why.


zoq:
Nice catch and thanks for the patch.

rcurtin:
Here's a trophy for getting issue #1000!



:)

PR #1001, opened 2017-05-18
Add random replay for DQN

ShangtongZhang:
Random replay with a simple unit test

### DIFF
--- a/src/mlpack/methods/reinforcement_learning/replay/random_replay.hpp
+++ b/src/mlpack/methods/reinforcement_learning/replay/random_replay.hpp
@@ -29,11 +29,11 @@ namespace rl {
  * For more information, see the following.
  *
  * @code
- * @phdthesis {lin1993reinforcement,
- *  title  =  {Reinforcement learning for robots using neural networks},
- *  author =  {Lin, Long-Ji},
- *  year   =  {1993},
- *  school =  {Fujitsu Laboratories Ltd}
+ * @phdthesis{lin1993reinforcement,
+ *  title  = {Reinforcement learning for robots using neural networks},
+ *  author = {Lin, Long-Ji},
+ *  year   = {1993},
+ *  school = {Fujitsu Laboratories Ltd}
  * }
  * @endcode
  *
--- a/src/mlpack/methods/reinforcement_learning/replay/random_replay.hpp
+++ b/src/mlpack/methods/reinforcement_learning/replay/random_replay.hpp
@@ -49,8 +49,8 @@ class RandomReplay
   /**
    * Construct an instance of random experience replay class.
    *
-   * @param batchSize # of examples returned at each sample.
-   * @param capacity Total memory size in terms of # of examples.
+   * @param batchSize Number of examples returned at each sample.
+   * @param capacity Total memory size in terms of number of examples.
    * @param dimension The dimension of an encoded state.
    */
   RandomReplay(const size_t batchSize,
--- a/src/mlpack/methods/reinforcement_learning/replay/random_replay.hpp
+++ b/src/mlpack/methods/reinforcement_learning/replay/random_replay.hpp
@@ -88,7 +88,8 @@ class RandomReplay
     nextStates.col(position) = nextState.Encode();
     isTerminal(position) = isEnd;
     position++;
-    if (position == capacity) {
+    if (position == capacity)
+    {
       full = true;
       position = 0;
     }
--- a/src/mlpack/methods/reinforcement_learning/replay/random_replay.hpp
+++ b/src/mlpack/methods/reinforcement_learning/replay/random_replay.hpp
@@ -60,10 +60,10 @@ class RandomReplay
       capacity(capacity),
       position(0),
       states(dimension, capacity),
-      actions(dimension, capacity),
-      rewards(dimension, capacity),
+      actions(capacity),
+      rewards(capacity),
       nextStates(dimension, capacity),
-      isTerminal(dimension, capacity),
+      isTerminal(capacity),
       full(false)
   { /* Nothing to do here. */ }
 
--- a/src/mlpack/methods/reinforcement_learning/replay/random_replay.hpp
+++ b/src/mlpack/methods/reinforcement_learning/replay/random_replay.hpp
@@ -20,6 +20,23 @@ namespace rl {
 /**
  * Implementation of random experience replay.
  *
+ * At each time step, interactions between the agent and the
+ * environment will be saved to a memory buffer. When necessary,
+ * we can simply sample previous experiences from the buffer to
+ * train the agent. Typically this would be a random sample and
+ * the memory will be a First-In-First-Out buffer.
+ *
+ * For more information, see the following.
+ *
+ * @code
+ * @phdthesis {lin1993reinforcement,
+ *  title  =  {Reinforcement learning for robots using neural networks},
+ *  author =  {Lin, Long-Ji},
+ *  year   =  {1993},
+ *  school =  {Fujitsu Laboratories Ltd}
+ * }
+ * @endcode
+ *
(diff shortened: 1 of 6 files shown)


src/mlpack/methods/reinforcement_learning/replay/random_replay.hpp
```diff
@@ -0,0 +1,142 @@
...
+        arma::randi<arma::uvec>(batchSize, arma::distr_param(0, upperBound - 1));
+    sampledStates = states.cols(sampledIndices);
+    sampledActions = actions.elem(sampledIndices);
+    sampledRewards = rewards.elem(sampledIndices);
+    sampledNextStates = nextStates.cols(sampledIndices);
+    isTerminal = this->isTerminal.elem(sampledIndices);
+  }
+
+  /**
+   * Get the # of transitions in the memory.
```
zoq: I would use Number instead of ``#`` here.

src/mlpack/methods/reinforcement_learning/replay/random_replay.hpp
```diff
@@ -0,0 +1,142 @@
...
+   *
+   * @return Actual memory size
+   */
+  size_t Size()
```
zoq: Just to be consistent with the rest of the codebase, I would do:  ``const size_t& Size() const { return size; }``.

src/mlpack/tests/rl_components_test.cpp
```diff
@@ -57,4 +58,42 @@ BOOST_AUTO_TEST_CASE(SimpleCartPoleTest)
   BOOST_REQUIRE_EQUAL(2, CartPole::Action::size);
 }
 
+/**
+ * Compare two matrix
+ */
+bool Equal(const arma::mat& m1, const arma::mat& m2)
```
zoq: I guess ``CheckMatrices`` form ``tests/test_tools.hpp`` would come in handy.

src/mlpack/methods/reinforcement_learning/replay/random_replay.hpp
```diff
@@ -0,0 +1,142 @@
+/**
+ * @file random_replay.hpp
```
ShangtongZhang: Yeah there is also prioritized replay but I'm not going to implement it in the near future. But I sill think a separate folder is better because once we add a new replay method, a separate folder can provide backward compatibility automatically. Otherwise user may have to modify their `#include` clause.

src/mlpack/methods/reinforcement_learning/replay/random_replay.hpp
```diff
@@ -0,0 +1,142 @@
...
+   * @param state Given state.
+   * @param action Given action.
+   * @param reward Given reward.
+   * @param nextState Given next state.
+   * @param isEnd Whether next state is terminal state.
+   */
+  void Store(const StateType& state, ActionType action,
+             double reward, const StateType& nextState, bool isEnd)
+  {
+    if (isTerminal.n_elem < capacity)
```
ShangtongZhang: I'm not sure. Two days ago it's preallocated, but sometimes user may declare a pretty large capacity and actually only small part of the memory is used.  So I modify the code to dynamic allocation. Maybe I can add a parameter in the constructor to determine whether to preallocate or not.

src/mlpack/methods/reinforcement_learning/replay/random_replay.hpp
zoq: I guess if we would use the root folder, we would also put other replay methods in the root folder. But if you prefer the extra folder, that's fine.

ShangtongZhang:
I ended up with preallocated memory. Today when I play Atari game with DQN in PyTorch I realized even it's a very large memory buffer, it will still be filled up easily. So there is no need for dynamic allocation.

zoq:
Great that you tested the approach. I think even if someone overestimates the capacity parameter the performance is still comparable with the dynamic approach.

src/mlpack/tests/rl_components_test.cpp
```diff
@@ -57,4 +58,34 @@ BOOST_AUTO_TEST_CASE(SimpleCartPoleTest)
   BOOST_REQUIRE_EQUAL(2, CartPole::Action::size);
 }
 
+/**
+ * Construct a random replay instance and check if it works as
+ * it should be.
+ */
+BOOST_AUTO_TEST_CASE(RandomReplayTest)
+{
+  RandomReplay<MountainCar> replay(1, 1);
```
zoq: I was thinking it's probably a good idea to test RandomReplay with capacity > 1. We could write a new test or integrate that into the existing.

src/mlpack/methods/reinforcement_learning/replay/random_replay.hpp
```diff
@@ -0,0 +1,165 @@
...
+ * At each time step, interactions between the agent and the
+ * environment will be saved to a memory buffer. When necessary,
+ * we can simply sample previous experiences from the buffer to
+ * train the agent. Typically this would be a random sample and
+ * the memory will be a First-In-First-Out buffer.
+ *
+ * For more information, see the following.
+ *
+ * @code
+ * @phdthesis {lin1993reinforcement,
```
zoq: Really minor issue: the build tool doesn't tolerate a whitespace between the command and the bracket: ``@phdthesis{lin1993reinforcement``

src/mlpack/methods/reinforcement_learning/replay/random_replay.hpp
```diff
@@ -0,0 +1,165 @@
...
+             const StateType& nextState,
+             bool isEnd)
+  {
+    states.col(position) = state.Encode();
+    actions(position) = action;
+    rewards(position) = reward;
+    nextStates.col(position) = nextState.Encode();
+    isTerminal(position) = isEnd;
+    position++;
+    if (position == capacity)
```
zoq: Shouldn't we use ``if ((position % capacity) == 0)``here, we only reset position once, so we could run into a index out of bounds error for the second run, maybe I missed something?

src/mlpack/methods/reinforcement_learning/replay/random_replay.hpp
```diff
@@ -0,0 +1,165 @@
...
+    {
+      full = true;
+      position = 0;
```
ShangtongZhang: Every time `position` reaches `capacity`, it will be rest to `0`

zoq: You are absolutely right, somehow I thought you check against full. Thanks for the clarification.

zoq:
Thanks for another great contribution that bring us one step closer.

src/mlpack/methods/reinforcement_learning/replay/random_replay.hpp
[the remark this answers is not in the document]
robotcator: > Yeah there is also prioritized replay but I'm not going to implement it in the near future. But I sill think a separate folder is better because once we add a new replay method, a separate folder can provide backward compatibility automatically. Otherwise user may have to modify their `#include` clause.

Hi, @ShangtongZhang, do you have any plan to implement prioritized replay buffer ? I would like to contribute as a process to be familiar with the mlpack codebase and get involved in this community. I also  implement it in Python before. [PER Implementation](https://github.com/robotcator/Assignment)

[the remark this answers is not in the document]
ShangtongZhang: @robotcator Yeah, it would be great if you can contribute PER.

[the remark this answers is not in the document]
robotcator: Sorry for late reply, I miss the notification. I written the schedule for PER and I think I will begin to code soon.

Issue #929, opened 2017-03-11
Normalizing signatures of Train methods and constructors


micyril:
As have been mentioned on the GSoC ideas page under the title "Cross-validation and hyper-parameter tuning infrastructure", many Train methods of classifiers have similar signatures of the form 
```
template<typename MatType>
void Train(const MatType& data, const arma::Row<size_t>& labels, ... other parameters ...)
```
On the other hand, the DecisionTree class has the method
```
  template<typename MatType>
  void Train(const MatType& data,
             const data::DatasetInfo& datasetInfo,
             const arma::Row<size_t>& labels,
             const size_t numClasses,
             const size_t minimumLeafSize = 10);
```
and the corresponding constructor
```
  template<typename MatType>
  DecisionTree(const MatType& data,
               const data::DatasetInfo& datasetInfo,
               const arma::Row<size_t>& labels,
               const size_t numClasses,
               const size_t minimumLeafSize = 10);
```
I guess the following signatures would be more conventional:
```
  template<typename MatType>
  DecisionTree(const MatType& data,
               const arma::Row<size_t>& labels,
               const data::DatasetInfo& datasetInfo,
               const size_t numClasses,
               const size_t minimumLeafSize = 10);

  template<typename MatType>
  void Train(const MatType& data,
             const arma::Row<size_t>& labels,
             const data::DatasetInfo& datasetInfo,
             const size_t numClasses,
             const size_t minimumLeafSize = 10);
```

micyril:
The class HoeffdingTree also has a less conventional order of arguments in one of its training constructors:
```
  template<typename MatType>
  HoeffdingTree(const MatType& data,
                const data::DatasetInfo& datasetInfo,
                const arma::Row<size_t>& labels,
                const size_t numClasses,
                const bool batchTraining = true,
                const double successProbability = 0.95,
                const size_t maxSamples = 0,
                const size_t checkInterval = 100,
                const size_t minSamples = 100,
                const CategoricalSplitType<FitnessFunction>& categoricalSplitIn
                    = CategoricalSplitType<FitnessFunction>(0, 0),
                const NumericSplitType<FitnessFunction>& numericSplitIn =
                    NumericSplitType<FitnessFunction>(0));
```
Analogously to the DecisionTree class, I think it would be more conventional to provide the following interface:
```
  template<typename MatType>
  HoeffdingTree(const MatType& data,
                const arma::Row<size_t>& labels,
                const data::DatasetInfo& datasetInfo,
[...]

PR #1002, opened 2017-05-22
Column-major responses in Linear Regression and LARS

### LINKED ISSUE #963: Regarding Linear Regression and LARS
As of now linear regression has constructor : 
`LinearRegression::LinearRegression(`
`const arma::mat& predictors,`
   `                                                       const arma::vec& responses,`
`                                                       const double lambda,`
`                                                       const bool intercept,`
`                                                       const arma::vec& weights)`
And training method as : 
` LinearRegression::Train(`
`const arma::mat& predictors,`
`                                     const arma::vec& responses,`
`                                     const bool intercept,`
`                                     const arma::vec& weights)`

And LARS has constructor : 
`LARS(`
`const arma::mat& data,`
`         const arma::vec& responses,`
`         const bool transposeData = true,`
`          const bool useCholesky = false,`
`          const double lambda1 = 0.0,`
       `const double lambda2 = 0.0,`
       `const double tolerance = 1e-16);`
And training method as : 
`void Train(`
            `const arma::mat& data,`
             `const arma::vec& responses,`
             `arma::vec& beta,`
             `const bool transposeData = true);

Since all the methods are column major (i.e. each input is a column vector) in mlpack, is it a good idea to convert these to column major form too ? 

Also since decision stump and perceptron (though it is less frequently used) have weights as 3rd parameter in the train method, does it sound reasonable to change the above mentioned methods (Train and constructor) to have weights as the third parameter ? (Similar to issue [929](https://github.com/mlpack/mlpack/issues/929) )
(It will be good to have uniform interface across the library and also help in simplifying the cross validation project)

Modified APIs may look like : 
`LinearRegression(`
`const arma::mat& predictors,`
   `                                                       const arma::rowvec& responses,`
` const arma::vec& weights,`
`                                                       const double lambda,`
`                                                       const bool intercept)`

` LinearRegression::Train(`
`const arma::mat& predictors,`
`                                     const arma::rowvec& responses,`
`                                     const arma::rowvec& weights`
`                                     const bool intercept)`

`LARS(`
`const arma::mat& data,`
`         const arma::rowvec& responses,`
`         const bool transposeData = true,`
`          const bool useCholesky = false,`
`          const double lambda1 = 0.0,`
       `const double lambda2 = 0.0,`
       `const double tolerance = 1e-16);`

`void Train(`
            `const arma::mat& data,`
             `const arma::rowvec& responses,`
             `arma::vec& beta,`
             `const bool transposeData = true);`

This will also probably eliminate the `transposeData` variable.

Also `LARS::Predict` will be changed. 
`void Predict(const arma::mat& points,`
           `   arma::vec& predictions,`
`const bool rowMajor = false) const;`

It will look like
`void Predict(const arma::mat& points,`
           `   arma::rowvec& predictions)`
and will have one less variable.

micyril:
Fix #963. Also separate signatures of constructors and Train methods in Linear Regression for learning with weights and without them as discussed in #929.

### DIFF
--- a/src/mlpack/core/dists/regression_distribution.hpp
+++ b/src/mlpack/core/dists/regression_distribution.hpp
@@ -48,14 +48,25 @@ class RegressionDistribution
    * @param predictors Matrix of predictors (X).
    * @param responses Vector of responses (y).
    */
+  mlpack_deprecated RegressionDistribution(const arma::mat& predictors,
+                                           const arma::vec& responses) :
+    RegressionDistribution(predictors, arma::rowvec(responses.t()))
+  {}
+
+  /**
+   * Create a Conditional Gaussian distribution with conditional mean function
+   * obtained by running RegressionFunction on predictors, responses.
+   *
+   * @param predictors Matrix of predictors (X).
+   * @param responses Vector of responses (y).
+   */
   RegressionDistribution(const arma::mat& predictors,
-                         const arma::vec& responses)
+                         const arma::rowvec& responses)
   {
-    arma::rowvec rowResponses = responses.t();
-    rf.Train(predictors, rowResponses);
+    rf.Train(predictors, responses);
     err = GaussianDistribution(1);
     arma::mat cov(1, 1);
-    cov(0, 0) = rf.ComputeError(predictors, rowResponses);
+    cov(0, 0) = rf.ComputeError(predictors, responses);
     err.Covariance(std::move(cov));
   }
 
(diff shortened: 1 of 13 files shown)


src/mlpack/methods/linear_regression/linear_regression_main.cpp
```diff
@@ -140,10 +140,10 @@ int main(int argc, char* argv[])
     {
       // The initial predictors for y, Nx1.
       Timer::Start("load_responses");
-      responses = std::move(CLI::GetParam<vec>("training_responses"));
+      responses = CLI::GetParam<vec>("training_responses").t();
```
rcurtin: We can change the definition of `training_responses` from `PARAM_COL_IN` to `PARAM_ROW_IN`, then the `.t()` isn't necessary.

src/mlpack/core/dists/regression_distribution.hpp
```diff
@@ -49,12 +49,13 @@ class RegressionDistribution
    * @param responses Vector of responses (y).
    */
   RegressionDistribution(const arma::mat& predictors,
-                         const arma::vec& responses) :
-      rf(regression::LinearRegression(predictors, responses))
+                         const arma::vec& responses)
```
rcurtin: Should we change this to `rowvec` also?

PR #1008, opened 2017-05-25
cpplint style fixes

amoudgl:
Objective is to bring down the count of style errors (in the whole mlpack codebase) to 0.


amoudgl:
@zoq All the style issues are not fixed yet, still working on them. The style check interface looks pretty awesome though!

zoq:
Thanks for going through the issue and thanks again for the really easy to use cpplint script. Btw. here: https://github.com/mlpack/jenkins-conf/commit/ab3b059c7dd3d96d7e6ee3af8242f25bb64f5900 is the commit, let me know if I missed something.

I agree the Jenkins integration is pretty neat. It took some time to convert and integrate the cpplint output into the cppchecks format, but it looks like it works. We have to define some thresholds to indicate that the build fails, but I guess once we did that the configuration is ready.

amoudgl:
Once we fix all the style issues, may be we can run cpplint just on the changed files rather than the whole codebase (correct me if I am wrong). If cpplint raises errors on the changed files in the pull request, we can mark as build failed. What do you think? 

Also, I have a suggestion. Currently we omit the style checks on the files which lies in `*/src/mlpack/core/arma_extend/*` and `*/src/mlpack/core/boost_backport/*` directories. But there are some more files where style check should not be applied I think like `./src/mlpack/core/data/serialization_template_version.hpp`. 
In future if new such files gets added to mlpack, rather than changing the command everytime, may be we can maintain a whitelist text file which would contain paths to all the files where style check should not be applied. We can just add new files to that whitelist text file and style check command in jenkins linter would remain the same. Let me know if this sounds good.

zoq:
> Once we fix all the style issues, may be we can run cpplint just on the changed files rather than the whole codebase (correct me if I am wrong). If cpplint raises errors on the changed files in the pull request, we can mark as build failed. What do you think?

That's a good idea, we can introduce another command line parameter and pass the files.

> Also, I have a suggestion. Currently we omit the style checks on the files which lies in */src/mlpack/core/arma_extend/* and */src/mlpack/core/boost_backport/* directories. But there are some more files where style check should not be applied I think like ./src/mlpack/core/data/serialization_template_version.hpp.
In future if new such files get added to mlpack, rather than changing the command everytime, we can maintain a whitelist text file which would contain paths to all the files where style check should not be applied. We can just add new files to that whitelist text file and style check command in jenkins linter would remain the same. Let me know if this sounds good.

@rcurtin  can probably say more about ``serialization_template_version.hpp`` but I like the idea to use a whitlist to exclude folder and files.

rcurtin:
I only see three issues that I just fixed with `serialization_template_version.hpp` in 80b03dbc, so I don't think it's necessary to exclude it. :)

I do agree though that a whitelist to exclude other folders and files is a good idea though.

amoudgl:
I have fixed majority of the style issues pointed by cpplint, please review. Remaining ones are kinda tricky to deal, in which length of macro statements (`#define`, `#ifndef`) exceeds 80 characters. Any ideas to handle them?

rcurtin:
I think for the header guards that are too long, we'll just have to abbreviate the names.  i.e. `MLPACK_CORE_TREE_RECTANGLE_TREE_R_PLUS_PLUS_TREE_AUXILIARY_INFORMATION_HPP` to `MLPACK_CORE_TREE_RECTANGLE_TREE_RPP_TREE_AUXILIARY_INFORMATION_HPP`.

The only other tricky one is the escape sequences... I think that `\%` can be changed to `%` and then on the following line the parser may be getting confused by the odd `"\"{"`, maybe it could be changed to `"{\""`.

I agree, the style check interface is really nice

amoudgl:
All the style issues have been resolved. Kindly review.

rcurtin:
Everything looks good to me and ready to merge. Thank you so much for taking the time to do this, this is such a nice improvement. @zoq: I don't know if you have any more comments; whenever you are ready go ahead and merge.

src/mlpack/core/dists/discrete_distribution.hpp
```diff
@@ -143,9 +144,10 @@ class DiscreteDistribution
       // Ensure that the observation is within the bounds.
       if (obs >= probabilities[dimension].n_elem)
       {
-        Log::Debug << "DiscreteDistribution::Probability(): received observation "
-             << obs << "; observation must be in [0, " << probabilities[dimension].n_elem
-             << "] for this distribution." << std::endl;
+        Log::Debug << "DiscreteDistribution::Probability(): "
+             << "received observation " << obs << "; observation must be"
```
zoq: Can you add a space after "must be" or before " received"?

src/mlpack/core/kernels/gaussian_kernel.hpp
```diff
@@ -126,8 +126,9 @@ class GaussianKernel
   template<typename VecTypeA, typename VecTypeB>
   double ConvolutionIntegral(const VecTypeA& a, const VecTypeB& b)
   {
-    return Evaluate(sqrt(metric::SquaredEuclideanDistance::Evaluate(a, b) / 2.0)) /
-        (Normalizer(a.n_rows) * pow(2.0, (double) a.n_rows / 2.0));
+    return Evaluate(sqrt(metric::
```
zoq: This looks strange to me, I think we shouldn't split the namespace ``metric::``from the function ``Evaluate``, also is the indentation correct, it should be 4 spaces right? Let me know what you think.

src/mlpack/core/math/lin_alg.cpp
```diff
@@ -257,7 +257,9 @@ void mlpack::math::Svec(const arma::sp_mat& input, arma::sp_vec& output)
 
 void mlpack::math::Smat(const arma::vec& input, arma::mat& output)
 {
-  const size_t n = static_cast<size_t>(ceil((-1. + sqrt(1. + 8. * input.n_elem))/2.));
+  const size_t n = static_cast<size_t>
```
zoq: Looks like the indentation isn't correct, also I think we can remove the ``static_cast`` here and write everything in a single line. @rcurtin what do you think?

src/mlpack/core/metrics/mahalanobis_distance.hpp
```diff
@@ -111,7 +111,7 @@ class MahalanobisDistance
   arma::mat covariance;
 };
 
-} // namespace distance
+} // namespace metric
```
zoq: Nice catch.

src/mlpack/core/optimizers/sgdr/sgdr.hpp
```diff
@@ -76,7 +76,7 @@ class SGDR
    * @param updatePolicy Instantiated update policy used to adjust the given
    *        parameters.
    */
- SGDR(DecomposableFunctionType& function,
+  SGDR(DecomposableFunctionType& function,
       const size_t epochRestart = 50,
```
zoq: The parameter should align with the first parameter: ``DecomposableFunctionType& function,``.

src/mlpack/core/optimizers/sgdr/snapshot_sgdr.hpp
```diff
@@ -91,7 +91,7 @@ class SnapshotSGDR
    * @param updatePolicy Instantiated update policy used to adjust the given
    *        parameters.
    */
- SnapshotSGDR(DecomposableFunctionType& function,
+  SnapshotSGDR(DecomposableFunctionType& function,
               const size_t epochRestart = 50,
```
zoq: Can you fix the parameter alignment?

src/mlpack/core/tree/address.hpp
```diff
@@ -262,6 +262,6 @@ bool Contains(const AddressType1& address, const AddressType2& loBound,
 
 } // namespace addr
 } // namespace bound
-} // namespave mlpack
+} // namespace mlpack
```
zoq: Another nice catch!

src/mlpack/core/tree/rectangle_tree/x_tree_split_impl.hpp
```diff
@@ -492,7 +492,8 @@ bool XTreeSplit::SplitNonLeafNode(TreeType *tree,std::vector<bool>& relevels)
             sorted2[i].second = sorted[i].second;
           }
         }
-        std::sort(sorted2.begin(), sorted2.end(), PairComp<ElemType, TreeType*>);
+        std::sort(sorted2.begin(), sorted2.end(),
+                  PairComp<ElemType, TreeType*>);
```
zoq: We should use the correct indentation here (4 spaces).

src/mlpack/core/tree/rectangle_tree/x_tree_split_impl.hpp
```diff
@@ -652,7 +653,8 @@ bool XTreeSplit::SplitNonLeafNode(TreeType *tree,std::vector<bool>& relevels)
...
```
zoq: We should use the correct indentation here (4 spaces).

amoudgl:
@rcurtin Thanks for your kind words. I could empathize with your burden of checking style errors in each PR, so thought of working on this issue. :) 

@zoq Thanks for reviewing this in detail, really appreciate it. I have tried to fix the issues you pointed out in my recent commit. Let me know if there are any more.

src/mlpack/core/math/lin_alg.cpp
rcurtin: I think that without the `static_cast` a compiler warning may be generated, but if it doesn't I agree that we can put this all on one line.  If it does, I think the opening `(` of the cast can be on the first line.

zoq:
Thanks for looking into the issues and thanks for the great work.

Issue #1009, opened 2017-05-26
Compile failure with clang


ShangtongZhang:
When I try latest version of mlpack, it failed to compile.
`git clone https://github.com/mlpack/mlpack.git mlpack --depth=1`
`mkdir build`
`cd build`
`cmake ../`
`-- The C compiler identification is AppleClang 8.1.0.8020042`
`-- The CXX compiler identification is AppleClang 8.1.0.8020042`
`Shangtong@Shangtong:~/Downloads/mlpack$ git lg`
`4144f66da2d91978e60282746d274c771c6719b2 Use correct types in program.`
Here is the error info http://paste.debian.net/942279/
It seems to have something to do with latest decision tree commit.

zoq:
The following fixed the problem for me, but maybe @rcurtin or @MikeLing have another/better solution.

```
diff --git a/src/mlpack/methods/decision_tree/decision_tree_impl.hpp b/src/mlpack/methods/decision_tree/decision_tree_impl.hpp
index 7c9ef7ecc..05dd95b20 100644
--- a/src/mlpack/methods/decision_tree/decision_tree_impl.hpp
+++ b/src/mlpack/methods/decision_tree/decision_tree_impl.hpp
@@ -592,15 +592,20 @@ void DecisionTree<FitnessFunction,
       }
 
       // Now build the child recursively.
-      children.push_back(new DecisionTree());
+      DecisionTree* child = new DecisionTree();
       if (NoRecursion)
-        children.back()->Train<UseWeights>(data, currentChildBegin,
+      {
+        child->Train<UseWeights>(data, currentChildBegin,
             currentCol - currentChildBegin, datasetInfo, labels, numClasses,
             weights, currentCol - currentChildBegin);
+      }
       else
-        children.back()->Train<UseWeights>(data, currentChildBegin,
+      {
+        child->Train<UseWeights>(data, currentChildBegin,
             currentCol - currentChildBegin, datasetInfo, labels, numClasses,
             weights, minimumLeafSize);
+      }
+      children.push_back(child);
     }
   }
   else
@@ -718,15 +723,20 @@ void DecisionTree<FitnessFunction,
       }
 
       // Now build the child recursively.
-      children.push_back(new DecisionTree());
+      DecisionTree* child = new DecisionTree();
       if (NoRecursion)
-        children.back()->Train<UseWeights>(data, currentChildBegin,
+      {
+        child->Train<UseWeights>(data, currentChildBegin,
[...]

PR #1010, opened 2017-05-26
Fix non-static member function issue

zoq:
For more Information take a look at #1009.


rcurtin:
I'm not sure I understand what the original error was or why this fixes it, but it fixes it :)

PR #1012, opened 2017-05-28
Add epsilon greedy policy for DQN

ShangtongZhang:
Add epsilon greedy policy and a simple test case.

### DIFF
--- a/src/mlpack/methods/reinforcement_learning/policy/greedy_policy.hpp
+++ b/src/mlpack/methods/reinforcement_learning/policy/greedy_policy.hpp
@@ -1,0 +1,103 @@
+/**
+ * @file greedy_policy.hpp
+ * @author Shangtong Zhang
+ *
+ * This file is an implementation of epsilon greedy policy.
+ *
+ * mlpack is free software; you may redistribute it and/or modify it under the
+ * terms of the 3-clause BSD license.  You should have received a copy of the
+ * 3-clause BSD license along with mlpack.  If not, see
+ * http://www.opensource.org/licenses/BSD-3-Clause for more information.
+ */
+#ifndef MLPACK_METHODS_RL_POLICY_GREEDY_POLICY_HPP
+#define MLPACK_METHODS_RL_POLICY_GREEDY_POLICY_HPP
+
+#include <mlpack/prereqs.hpp>
+
+namespace mlpack {
+namespace rl {
+
+/**
+ * Implementation for epsilon greedy policy.
+ *
+ * In general we will select an action greedily based on the action value,
+ * however under sometimes we will also randomly select an action to
+ * encourage exploration.
+ *
+ * @tparam EnvironmentType The reinforcement learning task.
+ */
+template <typename EnvironmentType>
+class GreedyPolicy {
+ public:
+  using ActionType = typename EnvironmentType::Action;
+
+  /**
+   * Constructor for epsilon greedy policy class.
+   * @param initialEpsilon The initial probability to explore (select a random action).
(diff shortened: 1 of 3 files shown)


src/mlpack/methods/reinforcement_learning/policy/greedy_policy.hpp
```diff
@@ -0,0 +1,103 @@
...
+    double exploration = math::Random();
+
+    // Select the action randomly.
+    if (exploration < epsilon)
+      return static_cast<ActionType>(math::RandInt(ActionType::size));
+
+    // Select the action greedily.
+    size_t bestAction = 0;
+    double maxActionValue = actionValue[0];
+    for (size_t action = 1; action < ActionType::size; ++action)
```
zoq: hm, do you think we should backport the ``index_max`` function, I guess we could also do ``arma::as_scalar(arma::find(actionValue.max(), 1))``. Let me know what you think.

src/mlpack/methods/reinforcement_learning/policy/greedy_policy.hpp
```diff
@@ -0,0 +1,103 @@
...
+ public:
+  using ActionType = typename EnvironmentType::Action;
+
+  /**
+   * Constructor for epsilon greedy policy class.
+   * @param initialEpsilon The initial probability to explore (select a random action).
+   * @param annealInterval The steps during which the probability to explore will anneal.
+   * @param minEpsilon Epsilon will never be less than this value.
+   */
+  GreedyPolicy(double initialEpsilon,
```
zoq: Can we use ``const``for the parameter?

src/mlpack/methods/reinforcement_learning/policy/greedy_policy.hpp
```diff
@@ -0,0 +1,103 @@
...
+#include <mlpack/prereqs.hpp>
+
+namespace mlpack {
+namespace rl {
+
+/**
+ * Implementation for epsilon greedy policy.
+ *
+ * In general we will select an action greedily based on the action value,
+ * however under sometimes we will also randomly select an action to
```
zoq: I think we can remove ``under`` here.

src/mlpack/tests/rl_components_test.cpp
```diff
@@ -87,18 +88,34 @@ BOOST_AUTO_TEST_CASE(RandomReplayTest)
...
 }
 
+/**
+ * Construct a greedy policy instance and check if it works as
+ * it should be.
+ */
+BOOST_AUTO_TEST_CASE(GreedyPolicyTest)
+{
+  GreedyPolicy<CartPole> policy(1.0, 10, 0.0);
+  for (int i = 0; i < 15; ++i)
```
zoq: Do you mind to use ``size_t`` instead of ``int`` here, I know it is kind of a pedantic request.

src/mlpack/methods/reinforcement_learning/policy/greedy_policy.hpp
ShangtongZhang: I was considering this too. It's more concise but costs twice time. But the performance here may not matter. I will do it.

zoq: I agree, performance wise it's probably not the best decision, I don't mind to keep the existing code, so we can go with the solution you like the most.

zoq:
Thanks, looking forward to review the next PR.

rcurtin:
I made some really minor style changes in 48676e8.  In future PRs, the style checker should give useful results now that we have merged #1008. :)

PR #1014, opened 2017-05-31
Basic DQN

ShangtongZhang:
This (double) DQN implementation is ready for classical control problems (e.g. CartPole). But it's not ready for Atari games with pixel input.


src/mlpack/methods/ann/ffn.hpp
```diff
@@ -142,6 +142,16 @@ class FFN
...
   /**
+   * Predict the responses to a given set of predictors. The responses will
+   * reflect the output of the given output layer as returned by the
+   * output layer function.
+   *
+   * @param predictors Input predictors.
+   * @return Output predictions of responses into.
+   */
+  arma::mat Predict(arma::mat predictors);
+
```
ShangtongZhang: Original `Predict` function is fairly inconvenient to use especially when I want to preserve the `predictors`. So I just add a wrapper.

src/mlpack/methods/reinforcement_learning/q_learning.hpp
```diff
@@ -0,0 +1,162 @@
...
+
+#include "replay/random_replay.hpp"
+
+namespace mlpack {
+namespace rl {
+
+/**
+ * Implementation of various Q-Learning algorithms, such as DQN, double DQN.
+ *
+ * For more details, see the following:
```
ShangtongZhang: Not sure whether to refer the DQN Nature paper. (The bibtex of the nature paper is really really long).

src/mlpack/methods/reinforcement_learning/q_learning_impl.hpp
```diff
@@ -0,0 +1,204 @@
...
+  arma::mat nextActionValues = targetNetwork.Predict(sampledNextStates);
+
+  arma::icolvec bestActions;
+  if (doubleQLearning)
+    // If use double Q-Learning, use learning network to select the best action.
+    bestActions = BestAction(learningNetwork.Predict(sampledNextStates));
+  else
+    bestActions = BestAction(nextActionValues);
+
+  // Compute the update target.
```
ShangtongZhang: Here we can improve performance by exposing `Forward` and `Backward`.

src/mlpack/methods/reinforcement_learning/q_learning.hpp
zoq: What do you think about:

```bibtex
@article{Mnih2013,
  author    = {Volodymyr Mnih and
               Koray Kavukcuoglu and
               David Silver and
               Alex Graves and
               Ioannis Antonoglou and
               Daan Wierstra and
               Martin A. Riedmiller},
  title     = {Playing Atari with Deep Reinforcement Learning},
  journal   = {CoRR},
  year      = {2013},
  url       = {http://arxiv.org/abs/1312.5602}
}
```

it's not that short, but I think this will work just fine.

src/mlpack/methods/reinforcement_learning/q_learning_impl.hpp
zoq: Sounds fine for me, do you like to do it in the context of this PR or later? I think if we expose the the ``Forward`` and ``Backward`` function for the ``FFN`` class we should do the same for the ``RNN`` class, just to be consistent.

src/mlpack/tests/q_learning_test.cpp
```diff
@@ -0,0 +1,133 @@
...
+          << std::endl;
+      break;
+    }
+  }
+}
+
+//! Test Double DQN in Cart Pole task.
+BOOST_AUTO_TEST_CASE(CartPoleWithDoubleDQN)
+{
+  arma::arma_rng::set_seed(0);
```
zoq: We should avoid setting the random seed in the tests, this can make specific test errors really hard to reproduce. What we like to do is to set the random seed like you did here and run like 1000 tests a local machine to make sure it works, then remove the line that sets the seed.

src/mlpack/methods/ann/ffn.hpp
zoq: Sounds good for me, I think we should use a reference here, there might be compilers which are not going to optimize this way.

src/mlpack/methods/reinforcement_learning/q_learning_impl.hpp
```diff
@@ -0,0 +1,204 @@
...
+  arma::mat sampledNextStates;
+  arma::icolvec isTerminal;
+  replayMethod.Sample(sampledStates, sampledActions, sampledRewards,
+                      sampledNextStates, isTerminal);
+
+  // Compute action value for next state with target network.
+  arma::mat nextActionValues = targetNetwork.Predict(sampledNextStates);
+
+  arma::icolvec bestActions;
+  if (doubleQLearning)
```
zoq: Can you use brackets for a multiline if condition?

src/mlpack/methods/reinforcement_learning/q_learning_impl.hpp
```diff
@@ -0,0 +1,204 @@
...
+    // If use double Q-Learning, use learning network to select the best action.
+    bestActions = BestAction(learningNetwork.Predict(sampledNextStates));
+  else
+    bestActions = BestAction(nextActionValues);
+
+  // Compute the update target.
+  arma::mat target = learningNetwork.Predict(sampledStates);
+  for (size_t i = 0; i < sampledNextStates.n_cols; ++i)
```
zoq: Can you use brackets for a multiline loop?

src/mlpack/methods/reinforcement_learning/q_learning_impl.hpp
```diff
@@ -0,0 +1,204 @@
...
+  // Get the initial state from environment.
+  state = environment.InitialSample();
+
+  // Track the steps in this episode.
+  size_t steps = 0;
+
+  // Track the return of this episode.
+  double totalReturn = 0.0;
+
+  // Running until get ot the terminal state.
```
zoq: Minor spelling issue here.

src/mlpack/tests/q_learning_test.cpp
```diff
@@ -0,0 +1,133 @@
...
+ *
+ * mlpack is free software; you may redistribute it and/or modify it under the
+ * terms of the 3-clause BSD license.  You should have received a copy of the
+ * 3-clause BSD license along with mlpack.  If not, see
+ * http://www.opensource.org/licenses/BSD-3-Clause for more information.
+ */
+
+#include <mlpack/core.hpp>
+
+#include <mlpack/core/optimizers/adam/adam.hpp>
```
zoq: Looks like we can remove the ``adam`` and ``orthogonal_init`` here.

src/mlpack/methods/ann/ffn.hpp
ShangtongZhang: But returning a non-const reference to a local variable is an undefined behavior. And I think `NRVO` is fairly reliable unless [conditional return](https://stackoverflow.com/questions/10522585/when-cant-a-compiler-use-rvo-or-nrvo).

src/mlpack/methods/reinforcement_learning/q_learning.hpp
ShangtongZhang: It looks good.

src/mlpack/methods/reinforcement_learning/q_learning_impl.hpp
ShangtongZhang: Yeah I'll do it for RNN too. I'm planning to do it after A3C.

src/mlpack/methods/ann/ffn.hpp
zoq: I was talking about ``predictors``.

src/mlpack/tests/q_learning_test.cpp
ShangtongZhang: I'm still confused. Here I'm setting a constant seed rather that a time dependent seed, so this test case should then become totally deterministic.

src/mlpack/methods/ann/ffn.hpp
ShangtongZhang: Oh. The reason why I didn't use `arma::mat Predict(arma::mat& predictors);` is that I want `predictors` to stay unchanged after calling `Predict`. So I want a copy here. If the signature of the original function is `void Predict(const arma::mat& predictors, arma::mat& results);`, I won't write this wrapper. But I find it's not so easy to change the original function to const reference.

zoq: Ah okay, thanks for the clarification!

src/mlpack/tests/q_learning_test.cpp
zoq: Right, so, what we would like to see is does the test pass just for this seed or 'all' seeds, maybe it will fail for another seed, and lead us to some issue in the code. Of course, it might fail for another seed but I think that's okay. We already tracked down some issues by avoiding static seeds.

rcurtin: Just for further clarification---if you set a static random seed, you may be lucky and the code may work for that specific seed only.

If you set a seed like `math::RandomSeed(std::time(NULL))` (which will set both random seed for Armadillo and mlpack RNGs, so that is probably the better way to do it), then it will be different on every single run, making debugging difficult.

If instead you leave the random seed completely unset across all tests, then effectively the random seed is set at compile time, meaning that across systems (or across the matrix build) you'll get a good picture of whether or not the test is robust across random seeds, and if it does fail, you can put a debugger on the program without recompiling with a specific random seed in order to see what's wrong.

Typically, before I check in a test that has randomness that I'm not confident of, I'll use `math::RandomSeed(std::time(NULL))` at the start of the test, then run many times:

```
while(true); do bin/mlpack_test -t TestSuiteName/TestName; sleep 1; done
```

(or sometimes we get it wrong and we only find out through the matrix build that some of the tests are failing intermittently, that's ok too as long as we fix it eventually...)

ShangtongZhang: That makes sense. Thanks for the clarification!

src/mlpack/tests/q_learning_test.cpp
```diff
@@ -0,0 +1,126 @@
...
+  QLearning<CartPole, decltype(model), decltype(opt), decltype(policy)>
+      agent(model, opt, 0.9, policy, replayMethod, 100, 100, true, 200);
+
+  arma::running_stat<double> averageReturn;
+  while (true)
+  {
+    double episodeReturn = agent.Episode();
+    averageReturn(episodeReturn);
+    /**
+     * Reaching running average return 40 is enough to show it works.
```
zoq: Do you think we can decrease the test time for the CartPoleWithDoubleDQN and CartPoleWithDQN test a little bit more? My timings: -DRELEASE=ON ~40 seconds and -DDEBUG=ON ~75 seconds.

src/mlpack/methods/ann/ffn.hpp
rcurtin: Why not add two overloads:

```
void Predict(const mat& predictors, mat& output);
void Predict(mat&& predictors, mat& output);
```

This would allow you to avoid having two dissimilar calls that do the same thing, and it's easy to specify when you want the original matrix unchanged and when you don't.

zoq: Actually we don't need ``void Predict(mat&& predictors, mat& output);`` fixed in: 
400abaa89c89. So I think we can remove ``arma::mat Predict(arma::mat predictors);``.

ShangtongZhang: If `void Predict(const mat& predictors, mat& output);` and `void Predict(mat& predictors, mat& output);` exist together, I think it will be somehow confusing. In that case, assume we have a non-const `predictors`, then the function call will be dispatched to `void Predict(mat& predictors, mat& output);` but actually we may still want the `predictors` to stayed unchanged even we didn't mark it as const. So I think if we use these two overloads, we should delete the original one or move it to private. Furthure more, I don't like the idea to putut `output` as a parameter. It's very user-unfriendly -- almost every time we call the function we need to write an extra line. So what I propose is deleting the original one and adding `mat Predict(const mat&)` and `mat Predict(mat&&)`. If both of you agree, I can do this for ffn and rnn. Can I assume there would not be many regression issues due to this change?

rcurtin: The non-const version would be an rvalue reference (`mat&&`) not a non-const reference (`mat&`).  What this means is that if I want to avoid a copy and I don't need those predictors back, I can do

```
network.Predict(std::move(predictors));
```

and now I save time because the predictors didn't need to be copied.  Otherwise I can just use it as normal

```
network.Predict(predictors);
```

and a copy will be incurred.  This is the strategy I've been employing throughout other parts of the code, like the trees and the constructors for other models like LogisticRegression, etc.: if you pass with `std::move()`, no copy happens but your data matrix may be lost; otherwise, the data will be copied if needed.

ShangtongZhang: Hi @zoq , I suspect this fix will lead to a forever copy. 
`this->predictors = std::move(predictors);` 
After this fix the RHS will be `const mat&&`, then the `=` operator will call `=(const mat&)` rather than `=(mat&&)`, which I think will lead to a copy.
See cpp.sh/2itol and [this](https://stackoverflow.com/questions/38917130/does-stdmove-works-with-lvalue-references-how-stdmove-works-on-standard-con)

ShangtongZhang: Hi @rcurtin , that's what exactly I mean. My point is we shouldn't allow `Predict(mat&)` and `Predict(const mat&)` exist together (So we need to delete current `Predict`). Instead we use `Predict(const mat&)` and `Predict(mat&&)` as you said.

rcurtin: Ah!  Right.  Sorry for writing the same thing twice then. :)

src/mlpack/tests/q_learning_test.cpp
ShangtongZhang: Sure. I will observe the statistics carefully and select a minimal threshold.

src/mlpack/methods/ann/ffn.hpp
zoq: @ShangtongZhang right, we should also provide another Constructor ``FFN(arma::mat&& predictors, arma::mat&& responses, ...)``, just as @rcurtin did for the trees.

zoq: I still don't like the idea to return the output, basically because it's the only place we would do this if I switch to another method I might wonder why I can't call ``output = model.Predict(input)``.

zoq: Or do we plan to provide 4 instances for the ``Predict`` function?

ShangtongZhang: It looks good for me to provide 4 `Predict`. The simplest way to do this I think is implementing `Predict(mat&&, mat&)`, then all other three methods can be easily implemented as a wrapper over this one.

ShangtongZhang: And it's also fine for me to only provide 2 instances and delete `mat Predict(...)`.

rcurtin: I'm a little confused, I only see the need for two (one with const reference, one with rvalue reference); what are the other two?

ShangtongZhang:
```
void Predict(mat&&, mat&)
void Predict(const mat&, mat&)
mat Predict(mat&&)
mat Predict(const mat&)
```

rcurtin: Ah, ok.  I missed your comment at 12:38pm, I must have been typing mine out then.

I agree with @zoq here---there is no other `Predict()` function in the library that operates like `mat Predict(const mat&)`, so I think it is better to be consistent and only provide the top two overloads.  If we eventually want to switch all `Predict()` and similar calls to `mat Predict()`, we can discuss that, but the place to discuss that would be some other issue I think (and it would also probably be a huge amount of work).

ShangtongZhang: ok that makes sense. Then I think what we need to do here is to implement `void Predict(mat&&, mat&)` directly and implement `void Predict(const mat&, mat&)` as a wrapper. (Perhaps need to revert the fix).

zoq: Sounds like a good plan for me.

ShangtongZhang: So will you do this or I make it in this PR?

zoq: If you like you can do here, don't feel obligated, just let me know.

ShangtongZhang: Yeah I can do it here

src/mlpack/tests/q_learning_test.cpp
```diff
@@ -59,16 +59,16 @@ BOOST_AUTO_TEST_CASE(CartPoleWithDQN)
     double episodeReturn = agent.Episode();
     averageReturn(episodeReturn);
     /**
-     * Reaching running average return 60 is enough to show it works.
+     * Reaching running average return 35 is enough to show it works.
```
zoq: Looks like a nice runtime/testtime improvement, thanks for looking into the issue.

zoq:
Looks ready for me, do you have anything else to add?

src/mlpack/methods/ann/ffn.hpp
```diff
@@ -88,6 +88,23 @@ class FFN
       OutputLayerType&& outputLayer = OutputLayerType(),
       InitializationRuleType initializeRule = InitializationRuleType());
 
+  /**
+   * Create the FFN object with the given predictors and responses set (this is
+   * the set that is used to train the network) and the given optimizer.
+   * Optionally, specify which initialize rule and performance function should
+   * be used.
```
rcurtin: It seems like there is no optimization happening with this constructor, so I think the comment is inaccurate.  (I guess this applies to the previous constructor too).

src/mlpack/methods/reinforcement_learning/q_learning_impl.hpp
```diff
@@ -0,0 +1,215 @@
...
+
+
+template <
+  typename EnvironmentType,
+  typename NetworkType,
+  typename OptimizerType,
+  typename PolicyType,
+  typename ReplayType
+>
+arma::icolvec QLearning<
```
rcurtin: Do you think this type should be `arma::Col<size_t>` to match the rest of the codebase better?

src/mlpack/methods/ann/ffn.hpp
```diff
@@ -88,6 +88,23 @@ class FFN
...
+   * @param predictors Input training variables.
+   * @param responses Outputs results from input training variables.
+   * @param outputLayer Output layer used to evaluate the network.
+   * @param initializeRule Optional instantiated InitializationRule object
+   *        for initializing the network parameter.
+   */
+  FFN(arma::mat&& predictors,
+      arma::mat&& responses,
+      OutputLayerType&& outputLayer = OutputLayerType(),
+      InitializationRuleType initializeRule = InitializationRuleType());
```
rcurtin: This may be a question more for @zoq: what is the intended use of this constructor and the version that takes `const arma::mat&`s?  I don't see it used in this code, and I'm not sure why a user would want to initialize the network with some data but not train it.  Maybe there is some reason I haven't thought of. :)

src/mlpack/tests/q_learning_test.cpp
```diff
@@ -0,0 +1,126 @@
...
+  GreedyPolicy<CartPole> policy(1.0, 1000, 0.1);
+  RandomReplay<CartPole> replayMethod(10, 10000);
+
+  // Set up the DQN agent.
+  QLearning<CartPole, decltype(model), decltype(opt), decltype(policy)>
+      agent(model, opt, 0.9, policy, replayMethod, 100, 100, true, 200);
+
+  arma::running_stat<double> averageReturn;
+  while (true)
+  {
```
rcurtin: Should we set a maximum number of iterations here?  I think that if there is a bug, this test might not terminate.

src/mlpack/methods/reinforcement_learning/q_learning.hpp
```diff
@@ -0,0 +1,170 @@
...
+ private:
+  /**
+   * Select the best action based on given action value.
+   * @param actionValues Action values.
+   * @return Selected actions.
+   */
+  arma::icolvec BestAction(const arma::mat& actionValues);
+
+  //! Reference of the learning network.
+  NetworkType& learningNetwork;
```
rcurtin: Reference members can make the class really hard to serialize; I might suggest that you provide a constructor instead that takes a const reference and copies, and then a constructor that takes an rvalue reference.  The same applies to the optimizer.  This only applies if you intend that this class will ever be serialized.

src/mlpack/methods/ann/ffn.hpp
ShangtongZhang: I'll look into it.. I just copied the old comment :)

src/mlpack/methods/reinforcement_learning/q_learning_impl.hpp
ShangtongZhang: That makes sense. I didn't realize it before.

src/mlpack/tests/q_learning_test.cpp
ShangtongZhang: Yeah I expect the build to fail with time limit exceed if there is bug. Bug it looks better to set a maximum iterations. I will make it.

rcurtin: Yeah---users may run the tests without any time limits, so it's good to prevent hanging when possible.

src/mlpack/methods/reinforcement_learning/q_learning.hpp
ShangtongZhang: It sounds good. I didn't think about serialization before. I'll make it.

ShangtongZhang:
I suddenly realized we don't need to provide `Predict(mat&&, mat&)` and `Predict(const mat&, mat&)`. We can only provide `Predict(mat m, mat&)`. Then in the function implementation, we make sure we use `std::move(m)` to do things. So if user wants to save a copy, he can invoke the function with `Predict(std::move(..), ..)`, there will be no copy. If user wants to preserve the matrix, he can simply call `Predict(..., ...)`, there is still only one copy as before. Did I miss something?

zoq:
I think, providing ``Predict(const mat&, mat&)`` is still a good idea, even if we do it just to explicitly tell the user the function doesn't change the input.

ShangtongZhang:
But I think `Predict(mat, mat&)` is also explicitly telling user this function won't change the input.

rcurtin:
I read http://cpptruths.blogspot.com/2012/03/rvalue-references-in-constructor-when.html , I need to think over it.  Maybe we can save a lot of code with this approach, but if we do this:

 * we should apply it to all mlpack code (so we should just open an issue about it, not do it all here of course).  Probably it is best applied to user-facing interfaces like constructors, `Train()`, and `Predict()` only.
 * we must make sure that the documentation is clear that whenever you can you should pass your data with `std::move()`---this would be my big sticking point, since now it is way easier for the user to accidentally copy their data and get slow code.  How would we best make this clear to user?  A comment in every constructor?  Something else?
 * we should make sure that gcc and clang are able to optimize that the same (or better) than what we do now.

But the benefit would be big, because it allows users to pass in _some_ values to be copied and some to be moved, and it allows us to reduce the amount of boilerplate code we have by a pretty large amount.

ShangtongZhang:
* I agree we should open a new issue for this to refactor existing codebase. But in the meantime can we still do it incrementally in newly added code like this PR? Because I think the old function is compatible with the new calling convention. If user calls it with `std::move`, both old and new functions won't have copy. If user calls it directly, new function will lead to one extra copy, while old function will have at most one extra copy. But according to our new calling convention, if user doesn't use `std::move`, we can assume user can afford an extra copy.
* Yeah this could be a headache. What I can think besides clear documentation is that we should also refactor our test cases carefully to show user a correct usage example. When I was a beginner with mlpack, I just turned to test case to find usage of some class.
* I don't think it would be a big issue. As far as I know, this new style doesn't depend on any special optimizations like `NRVO`. It's only a move constructor and works somehow like the `copy-and-swap` idiom.

zoq:
> I agree we should open a new issue for this to refactor existing codebase. But in the meantime can we still do it incrementally in newly added code like this PR? 

I'm fine to do it for the ann code, it's one part that's not released yet, so I think we don't run into issues any issues.

rcurtin:
I thought about it more, I can't see any reason not to switch to pass-by-value instead of having an lvalue-reference overload and an rvalue-reference overload.  So I think we should do it throughout the mlpack codebase.  I'd say go ahead and do it here too.

What's really nice about this is that it doesn't break the API, since you still interact with the methods in the exact same way.  I opened #1021 to handle the rest of the code.

src/mlpack/methods/ann/ffn.hpp
zoq: Good point, I can't think of any situation where that might be necessary. Maybe I just looked at the LinearRegression/LogisticRegression class and thought it's a good idea.

src/mlpack/methods/reinforcement_learning/q_learning.hpp
ShangtongZhang: I just realized I can't do this easily. My constructor should accept `model` and `opt`, however the problem comes from that `opt` itself locally stores a reference to `model`. If I `std::move` the `model` object, the reference in `opt` will be invalid.

zoq: Actually, do we have to pass the model? Instead we could get the model from the optimizer itself.

ShangtongZhang: Getting the model from the optimizer looks very anti-intuitive for me and doesn't solve the problem from root. What I think is the most elegant way is to design a mechanism to allow rebinding for optimizer or delayed binding. It's even more necessary for async methods -- each thread will have its own model and may have their own optimizer, so the model must be bound to its corresponding optimizer dynamically. Currently it seems we cannot do this.
I have some solutions, one is delayed binding by lambda function. In stead of passing in a optimizer, we pass in a optimizer generator (that's what I did in my PyTorch implementation) like this:
`auto opt = [](auto& model) { return StandardSGD<decltype(model)>(model, 0.0001, 2); }`
The drawback is user has to write this lambda function and the code will become fairly overstaffed so I don't want to do this in C++.
Another delayed binding solution is to add a overload for `Optimize` of optimizer like this:
`double Optimize(arma::mat& iterate, FunctionType& function);`
It looks like the easiest way. But the problem is if we pass in the `function` every time we call the function, why do we still store it in constructor? If we do this, do we need to change the original constructor? Although this, totally separating optimizer and network, is the most efficient design I think, I don't know what's the influence of this change on existing codebase.
Rebinding is also a solution, we can refactor the optimizer with `reference_wrapper` like this:
```
template<typename FunctionType>
class GradientDescent
{
 public:
  GradientDescent(FunctionType& function,
                  const double stepSize = 0.01,
                  const double tolerance = 1e-5);

  //! Get the instantiated function to be optimized.
  const FunctionType& Function() const { return function.get(); }
  //! Modify the instantiated function.
  FunctionType& Function() { return function.get(); }

  void Rebind(std::reference<FunctionType> f) { function = f; };

 private:
  //! The instantiated function.
  std::reference_wrapper<FunctionType> function;
};
```
This would be totally compatible with existing codebase.
If I do all the things from scratch, I'd prefer the second solution. But if refactoring the constructor of optimizer class will lead to too many regression issues, I'm fine to implement the third one.

zoq: For solution 2 we still have to provide the existing constructor and also the existing ``Optimize(arma::mat& iterate, FunctionType& function);`` to be backward compatible at least for mlpack < 3.0, so that's this. Also despite the refactoring that we have to do, I can't think of any problem that comes with the solution. I would also prefer solution 2 over 3, not just because you don't have to think about rebinding the function in certain cases. But the big point is backward compatibility. Eager to here @rcurtin on this point.

rcurtin: I suspect that any development here will be part of 3.0.0, not 2.2.x---I think that only backported fixes will go into 2.2.x at this point really.

I think that the second solution is better here; in my view I don't see a reason to bind the `FunctionType` to the `OptimizerType` instance.  Instead I agree with Shangtong and think the design would be better as:

```
class GradientDescent
{
 public:
  GradientDescent(...);

  template<typename FunctionType>
  void Optimize(FunctionType& function, arma::mat& iterate);

  ...
};
```

I think that this refactoring would also make @micyril's hyperparameter tuner easier to design; a little more background on that in the second half of this email:

http://knife.lugatgt.org/pipermail/mlpack/2017-May/003327.html

When I originally made the `OptimizerType` idea, I thought that it was necessary to bind the function so that the optimizer could use `GetInitialPoint()` for the start of the optimization.  But as time has gone on, I think that `GetInitialPoint()` is not used internally in any optimizer, so now there is no reason to bind the optimizer type to the function type that it is optimizing.

I guess, to summarize, I think option 2 is the right way to do this.  If we all agree I can open a ticket for it, and help with the refactoring if needed.  (It should be a fairly simple refactoring, might take a little bit of time though.)

ShangtongZhang: I agree. I can make another PR to do this.

zoq: That would be greatly appreciated.

ShangtongZhang: When I refactor the optimizer, I found some optimizer like `lbfgs` has many member functions and many of them use the locally stored `function` reference, and even the constructor also uses the `function`:
`const size_t rows = function.GetInitialPoint().n_rows;`,
So for those classes, will storing a reference be a better choice? After seeing `lbfgs`, I would vote for the the `std::reference_wrapper` idea, by this we only need to modify the code within optimizer folder, all other code won't be affected. (And `aug_lagrangian` seems also very complicated)

zoq: If I didn't missed something, as for ``L_BFGS`` ``function`` is only used in private member functions, as for ``aug_lagrangian`` it looks a little bit more complicated since we also have to provide an interface for the utility AugLagrangianFunction. I guess, if the proposed change would make @rcurtin and  @micyril's hyperparameter tuner project easier, it's worth the work. Maybe they can provide more information on that point. If they say it would be helpful, we could split the work?

ShangtongZhang: As far as I can see, `function` is used in constructor in `L_BFGS` and `SA`. I can do this refactor for some simple optimizers like `ada_deltas`, `ada_grad`, `adam`, `gradient_descent`, `minibatch_sgd`, `rmsprop`, `sgdr`, `smorms3`. But for other complicated optimizers I am afraid I'm not suitable to do that -- I don't have any idea how them work, even this is the first time I hear those names :<

zoq: Sounds more than fair for me.

ShangtongZhang:
This PR is still not fully backward compatible -- Using optimizers not shown in this https://github.com/mlpack/mlpack/pull/1026 with FFN and RNN will lead to compile error. But I doubt whether there exists someone that did use such unusual optimizers with FFN and RNN. I'm fine to delay merging this PR until all the optimizers are ready.

src/mlpack/methods/reinforcement_learning/q_learning.hpp
```diff
@@ -0,0 +1,173 @@
...
+   * @param discount Discount for future return.
+   * @param policy Behavior policy of the agent.
+   * @param replayMethod Experience replay method.
+   * @param targetNetworkSyncInterval Interval (steps) to sync the target network.
+   * @param explorationSteps Steps before starting to learn.
+   * @param doubleQLearning Whether to use double Q-Learning.
+   * @param stepLimit Maximum steps in each episode, 0 means no limit.
+   * @param environment Reinforcement learning task.
+   */
+  QLearning(NetworkType network,
```
zoq: Do you mind to use ``const`` for the ``discount``, ``targetNetworkSyncInterval``, etc. parameter?  To be consistent with the style of the codebase?

src/mlpack/methods/reinforcement_learning/q_learning_impl.hpp
```diff
@@ -0,0 +1,216 @@
...
+  EnvironmentType,
+  NetworkType,
+  OptimizerType,
+  PolicyType,
+  ReplayType
+>::BestAction(const arma::mat& actionValues)
+{
+  arma::Col<size_t> bestActions(actionValues.n_cols);
+  arma::rowvec maxActionValues = arma::max(actionValues, 0);
+  for (size_t i = 0; i < actionValues.n_cols; ++i)
```
zoq: To improve readability maybe we should use braces here?

zoq:
There is one more thing, can you remove ``move`` from ``model.Train(X, Y, opt);`` in ``convolutional_network_test.cpp``,  should resolve the issue with the failing test.

ShangtongZhang:
Hope it's ready to merge now.

zoq:
Thanks for looking into the issues.

src/mlpack/tests/q_learning_test.cpp
```diff
@@ -0,0 +1,143 @@
...
+  size_t episodes = 0;
+  while (true)
+  {
+    double episodeReturn = agent.Episode();
+    averageReturn(episodeReturn);
+    episodes += 1;
+
+    if (episodes > 1000)
+    {
+      Log::Fatal << "Cart Pole with DQN failed." << std::endl;
```
zoq: I was wondering if we could run into the same problem as @rcurtin encountered here: https://github.com/mlpack/mlpack/commit/531aae9eaa3ff2168717be4a3c117b81df942dff

Would be easy to fix by replacing ``Log::Fatal`` with a boolean that we can check at the end.

zoq:
I run the ``CartPoleWithDoubleDQN`` test in a loop to see how often it fails, turns out it fails relative often, so I tried to stabalize the test and also tested another network architecture to reduce the test time. Let me know if I messed something up. Here is the commit: https://github.com/mlpack/mlpack/commit/2c51181068d9184530264eab25e56a7e47f4761c

ShangtongZhang:
Thanks for looking into this issue! It's amazing that such a small network will work.

PR #1016, opened 2017-06-02
Metrics

micyril:
Add the metrics accuracy and mean squared error for cross-validation.


src/mlpack/core/cv/metrics/mse_impl.hpp
```diff
@@ -0,0 +1,60 @@
...
+{
+  model.Predict(data, responses);
+}
+
+template<typename MLAlgorithm, typename DataType>
+void MSE::Predict(MLAlgorithm& model,
+                  const DataType& data,
+                  arma::mat& responses)
+{
+  // In the case of neural networks data should be passed without const
```
zoq: Fixed in 400abaa89c89eb20190fc3a3e1d0febc3d415033, so I think we can remove the extra function here.

src/mlpack/core/cv/CMakeLists.txt
```diff
@@ -0,0 +1,13 @@
+# Define the files we need to compile
+# Anything not in this list will not be compiled into mlpack.
+set(SOURCES
+)
```
rcurtin: Either we should add the `metrics/*` files here, or an `add_subdirectory(metrics)` call somewhere in this file so that the new files are properly added to `MLPACK_SOURCES`.  Let me know if I can clarify anything about that.

micyril: I don't have too much experience with cmake, and I was looking at src/mlpack/core/data/CMakeLists.txt as an example while writing src/mlpack/core/cv/CMakeLists.txt, since src/mlpack/core/data also has some directories. Probably we should fix src/mlpack/core/data/CMakeLists.txt too. I will fix src/mlpack/core/cv/CMakeLists.txt shortly.

rcurtin: Ah, nice catch, I will commit changes for `data/` shortly.

PR #1005, opened 2017-05-23
[GSoC] Augmented RNN models - benchmarking framework

partobs-mdp:
This PR is part of my GSoC project "Augmented RNNs".
Imeplemented:

- class `CopyTask` for evaluating models on the sequence copy problem, showcasing benchmarking framework;
- unit test for it (a simple non-ML model that is hardcoded to copy the sequence required number of times is expected to ace the `CopyTask`).

PR #1018, opened 2017-06-05
GRU layer for ANN module

sumedhghaisas:
Implementation of Grated Recurrent Unit for recurrent networks in ANN
Tests are implemented similar to the tests of LSTM


src/mlpack/methods/ann/layer/gru_impl.hpp
```diff
@@ -0,0 +1,279 @@
...
+    const size_t rho) :
+    inSize(inSize),
+    outSize(outSize),
+    rho(rho),
+    forwardStep(0),
+    backwardStep(0),
+    gradientStep(0),
+    deterministic(false)
+{
+  // input specific linear layers(for zt, rt, ot)
```
zoq: Can you take a look at: https://github.com/mlpack/mlpack/wiki/DesignGuidelines#comments and use the correct format (punctuation) for each comment?

src/mlpack/methods/ann/layer/gru_impl.hpp
```diff
@@ -0,0 +1,279 @@
...
+  // where cmul1 is input gate * prevOutput and
+  // cmul2 is (1 - input gate) * hidden gate.
+  output = (boost::apply_visitor(outputParameterVisitor, inputGateModule) 
+      % prevOutput) + 
+      ((arma::ones<arma::vec>(outSize) - 
+      boost::apply_visitor(outputParameterVisitor, inputGateModule)) % 
+      boost::apply_visitor(outputParameterVisitor,
+      hiddenStateModule));
+      
+  prevOutput = output;
```
zoq: I wonder can we avoid the extra copy if we do

```
if (!deterministic)
{
  outParameter.push_back(prevOutput);
}
```

here. The problem I see is, that the first output is zero, but we could preallocated the vector and use forwardStep as index and push the first output in the preallocated step.

src/mlpack/tests/recurrent_network_test.cpp
```diff
@@ -635,9 +652,18 @@ void DistractedSequenceRecallTestNetwork()
...
+}
+
+/**
+ * Train the specified networks on the Derek D. Monner's distracted sequence
+ * recall task.
+ */
+BOOST_AUTO_TEST_CASE(GRUDistractedSequenceRecallTest)
 {
-  DistractedSequenceRecallTestNetwork();
+  DistractedSequenceRecallTestNetwork<GRU<>>();
```
zoq: I like the idea to reuse the existing test code, but can we tune the GRU parameter to accelerate the test, or do we require the same settings to get stable results. The problem I see is are the current runtimes on the sparc machine: http://masterblaster.mlpack.org/job/mlpack%20-%20nightly%20matrix%20build/arch=sparc64,armadillo=armadillo-5.500.2,buildmode=debug/lastCompletedBuild/testReport/mlpackTest/RecurrentNetworkTest/

I thought the latest change improved the timings, but it looks like that's not true at least for the sparc machines.

src/mlpack/methods/ann/layer/gru.hpp
```diff
@@ -0,0 +1,209 @@
...
+
+#include <boost/ptr_container/ptr_vector.hpp>
+
+#include "../visitor/delta_visitor.hpp"
+#include "../visitor/output_parameter_visitor.hpp"
+
+#include "layer_types.hpp"
+#include "add_merge.hpp"
+#include "sequential.hpp"
+
```
sumedhghaisas: Ahh yes. I will add that first thing.

src/mlpack/methods/ann/layer/gru_impl.hpp
sumedhghaisas: Surely. I will add better documentation than that.

src/mlpack/methods/ann/layer/gru_impl.hpp
sumedhghaisas: Hmm... I agree. We could use the last entry in outParameters rather than prevOutput, right? or am I getting it wrong? I think it will save 2 copies. If we do push the temporary computed output directly to outParameters with c++11 move semantics and then make a copy of that for returning. I think we can do the same thing for LSTM as well. What do you think?

src/mlpack/tests/recurrent_network_test.cpp
sumedhghaisas: Sure. I will try to tweak the parameters for GRU specific tests. The current tests are just for validating the implementation. I have usually observed that GRU converges faster than LSTM, do you think we should add that as a test as well?

zoq: I like the idea, but I'm not sure the test would provide much insight about the correctness of the GRU or LSTM implementation. Like in some situations LSTM might converge faster as the GRU layer.

src/mlpack/methods/ann/layer/gru_impl.hpp
zoq: Yes, I was thinking the same, the impact might be minimal, but I think it's worth a test.

src/mlpack/methods/ann/layer/gru_impl.hpp
zoq: I'm fine with the documentation: Just like to follow the style guidelines: so instead of ``input specific linear layers(for zt, rt, ot)`` we do: ``Input specific linear layers(for zt, rt, ot).``. I think using equations/symbols from the paper is a great idea.

src/mlpack/methods/ann/layer/gru.hpp
```diff
@@ -0,0 +1,209 @@
...
+  LayerTypes outputHidden2GateModule;
+
+  //! Locally-stored input gate module.
+  LayerTypes inputGateModule;
+
+  //! Locally-stored hidden state module.
+  LayerTypes hiddenStateModule;
+
+  //! Locally-stored forget gate module.
+  LayerTypes forgetGateModule;
```
rcurtin: A couple comments here---

 * It might be better to use the actual type of the layer here just for clarity in the definition.

 * It's possible that it may be better to manually implement each of the layers here.  This could lead to a more optimized implementation of `Forward()` and `Backward()`.  In order to determine whether it's necessary to take the time to do that, I'd suggest doing a quick benchmark of mlpack vs. Keras on some simple GRU network.

src/mlpack/methods/ann/layer/gru_impl.hpp
rcurtin: You could also replace the `std::vector<arma::mat>` with an `arma::cube`, that may provide some very minor additional gain.

src/mlpack/methods/ann/layer/gru_impl.hpp
sumedhghaisas: Sure. Will do that.

src/mlpack/methods/ann/layer/gru_impl.hpp
sumedhghaisas: hmm... I think it will increase a locality of the data. I will try to do a benchmarking check.

src/mlpack/methods/ann/layer/gru.hpp
sumedhghaisas: I agree to changing the definition to actual types. It will provide more clarity while reading the headers.

sumedhghaisas: About implementing Forward and Backward manually. I am not so sure. I did a quick and dirty check on the paper (writing down the equations involved), and did not find any operation that could be avoided with manual implementation. All the matrices that currently saved inside layers will be saved in GRU layer and they all help in Backward and Gradient. We will be saving on locality and function calls. Nonetheless, I will do the calculations again tomorrow, in case I missed out on something.

rcurtin: If you don't see an easy route toward any optimization with hand implementation and the existing implementation is comparably fast to, e.g., Keras or whatever other tools, then I think it's fine as-is. :)

src/mlpack/methods/ann/layer/gru_impl.hpp
zoq: If we test the cube approach we should also test resize, since we do support a dynamic sequence length, e.g. rho might change over time.

zoq: You said, ``cube`` is slightly faster as ``std::vector`` to you think it's worth to introduce a parameter that allows us to select the 'right' container, e.g. if the sequence length is const we could use cube?

src/mlpack/tests/recurrent_network_test.cpp
```diff
@@ -183,7 +183,7 @@ void GenerateReber(const arma::Mat<char>& transitions, std::string& reber)
         grammerIdx + 2)) - '0';
   } while (idx != 0);
 
-  reber =  "BPTVVE";
+  //reber =  "BPTVVE";
```
zoq: Nice catch!

src/mlpack/tests/recurrent_network_test.cpp
```diff
@@ -448,20 +452,59 @@ void ReberGrammarTestNetwork(bool embedded = false)
...
+BOOST_AUTO_TEST_CASE(LSTMEmbeddedReberGrammarTest)
+{
+  ReberGrammarTestNetwork<LSTM<>>(true);
+}
+
 /**
  * Train the specified networks on a Reber grammar dataset.
  */
-BOOST_AUTO_TEST_CASE(ReberGrammarTest)
+BOOST_AUTO_TEST_CASE(GRUReberGrammarTest)
```
zoq: I end up with ``error: Mat::rows(): indices out of bounds or incorrectly used`` running ``RecurrentNetworkTest/GRUReberGrammarTest`` wonder if you missed to check in something crucial?

zoq: Same error for ``LSTMReberGrammarTest``.

src/mlpack/methods/ann/layer/gru_impl.hpp
```diff
@@ -0,0 +1,307 @@
+/**
+ * @file lstm_impl.hpp
```
zoq: Minor issue, use the correct file name.

src/mlpack/methods/ann/layer/gru_impl.hpp
```diff
@@ -0,0 +1,307 @@
...
+ * @author Sumedh Ghaisas
+ *
+ * Implementation of the LSTM class, which implements a lstm network
```
zoq: Use the correct layer description.

src/mlpack/methods/ann/layer/gru_impl.hpp
```diff
@@ -0,0 +1,307 @@
...
+ * mlpack is free software; you may redistribute it and/or modify it under the
+ * terms of the 3-clause BSD license.  You should have received a copy of the
+ * 3-clause BSD license along with mlpack.  If not, see
+ * http://www.opensource.org/licenses/BSD-3-Clause for more information.
+ */
+#ifndef MLPACK_METHODS_ANN_LAYER_GRU_IMPL_HPP
+#define MLPACK_METHODS_ANN_LAYER_GRU_IMPL_HPP
+
+// In case it hasn't yet been included.
+#include "linear.hpp"
```
zoq: We should include ``gru.hpp`` here right, it looks like I did the same mistake for ``lstm_impl.hpp``.

src/mlpack/methods/ann/layer/gru_impl.hpp
```diff
@@ -0,0 +1,307 @@
...
+
+  // Pass the second through forgetGate
+  boost::apply_visitor(ForwardVisitor(std::move(output.submat(
+      1 * outSize, 0, 2 * outSize - 1, 0)), std::move(boost::apply_visitor(
+      outputParameterVisitor, forgetGateModule))), forgetGateModule);
+
+  arma::mat modInput = (boost::apply_visitor(outputParameterVisitor, 
+      forgetGateModule) % *prevOutput);
+
+  // Pass that through the outputHidden2GateModule
```
zoq: Pedantic style comment: missing ``.`` at the end.

src/mlpack/methods/ann/layer/gru_impl.hpp
```diff
@@ -0,0 +1,307 @@
...
+  output = (boost::apply_visitor(outputParameterVisitor, 
+      input2GateModule).submat(0, 0, 2 * outSize - 1, 0) +
+      boost::apply_visitor(outputParameterVisitor, output2GateModule));
+
+  // Pass the first outSize through inputGate(it).
+  boost::apply_visitor(ForwardVisitor(std::move(output.submat(
+      0, 0, 1 * outSize - 1, 0)), std::move(boost::apply_visitor(
+      outputParameterVisitor, inputGateModule))), inputGateModule);
+
+  // Pass the second through forgetGate
```
zoq: Pedantic style comment: missing ``.`` at the end.

src/mlpack/methods/ann/layer/gru_impl.hpp
```diff
@@ -0,0 +1,307 @@
...
+      input2GateModule).submat(2 * outSize, 0, 3 * outSize - 1, 0) + 
+      boost::apply_visitor(outputParameterVisitor, outputHidden2GateModule);
+
+  // Pass it through hiddenGate.
+  boost::apply_visitor(ForwardVisitor(std::move(outputH), std::move(
+      boost::apply_visitor(outputParameterVisitor, hiddenStateModule))), 
+      hiddenStateModule);
+
+  // Update the output (nextOutput): cmul1 + cmul2
+  // Wwhere cmul1 is input gate * prevOutput and
```
zoq: Looks like you mean ``Where ``.

src/mlpack/methods/ann/layer/gru_impl.hpp
```diff
@@ -0,0 +1,307 @@
...
+      outputHidden2GateModule);
+
+  gradIterator--;
+}
+
+template<typename InputDataType, typename OutputDataType>
+void GRU<InputDataType, OutputDataType>::ResetCell()
+{
+  outParameter.clear();
+  outParameter.push_back(arma::zeros<arma::mat>(outSize, 1));
```
zoq: I was wondering if we could preserve the existing data, and only allocate if the ``list.size < rho``. This way we don't have to push ``arma::zeros<arma::mat>(outSize, 1)`` again?

sumedhghaisas: I am sorry I did not get that. Yo mean we clear the list till the first element?

zoq: Yes, not sure that is feasible or even necessary, just an idea.

rcurtin:
Hey Sumedh, I wanted to ask what the status of this PR was.  Based on the weekly report it seems like you are done with GRU, but maybe there is something I misunderstood?  Let me know, I am looking forward to when we can merge this. :)

sumedhghaisas:
Hey Ryan. If the current checks all pass the branch is ready to be merged :)

src/mlpack/methods/ann/layer/gru.hpp
```diff
@@ -0,0 +1,246 @@
+/**
+ * @file gru.hpp
+ * @authorSumedh Ghaisas
```
zoq: Minor style issue, missing space between tag and author.

src/mlpack/methods/ann/layer/gru.hpp
```diff
@@ -0,0 +1,246 @@
...
+#include "add_merge.hpp"
+#include "sequential.hpp"
+
+namespace mlpack {
+namespace ann /** Artificial Neural Network. */ {
+
+/**
+ * An implementation of a gru network layer.
+ *
+ * This class allows specification of the type of the activation functions used
```
zoq: The comment doesn't match with the layer.

src/mlpack/tests/recurrent_network_test.cpp
```diff
@@ -572,8 +583,8 @@ void DistractedSequenceRecallTestNetwork()
   {
     RNN<MeanSquaredError<> > model(rho);
     model.Add<IdentityLayer<> >();
-    model.Add<Linear<> >(inputSize, 14);
-    model.Add<LSTM<> >(14, 7, rho);
+    model.Add<Linear<> >(inputSize, 16);
```
zoq: I wonder if a hiddensize < 16 would pass the test too?

rcurtin:
@sumedhghaisas: do you think you could add some unit tests for the GRU, like for the `Gradient()`, `Forward()`, and `Baclward()` functions? It might be a little tricky to do this outside of a network but I think it should be possible, and I think that will be a good supplement to the grammar tests.

It looks like there are some other comments to be addressed too; if you can take care of those, we can get closer to merging this.

src/mlpack/tests/recurrent_network_test.cpp
sumedhghaisas: I tried some tests... to my surprise Reber grammar tests pass with just 1 unit of each. Distracted sequence tests pass with 3 units.

sumedhghaisas:
Hey Ryan. I am not sure how I would write the Forward, Backward and Gradient tests cause it would involve setting weights for every parameter used in the cell. Cause random initialization would give different results. Exactly what do you mean by Forward test?

zoq:
What about checking the Gradient, as done in ``ann_layer_test.cpp``? We can also test the forward function using some generated input and compare against manually calculated values?

sumedhghaisas:
So according to the current implementation, the parameters have to be accessed through the model stored inside the layer. If we set all the parameters to value '1' we can check first couple of outputs by sending in '1's as inputs. Is that a good idea?

zoq:
Sounds reasonable to me.

src/mlpack/methods/ann/init_rules/unit_init.hpp
```diff
@@ -0,0 +1,66 @@
...
+
+#include <mlpack/prereqs.hpp>
+
+namespace mlpack {
+namespace ann /** Artificial Neural Network. */ {
+
+/**
+ * This class is used to initialize randomly the weight matrix.
+ */
+class UnitInitialization
```
zoq: This class is only for testing right? So I think we can use the ``RandomInitialization`` class and set ``lowerBound`` and ``upperBound`` to 1. What do you think?

src/mlpack/methods/ann/rnn.hpp
```diff
@@ -202,6 +202,8 @@ class RNN
   //! Modify the initial point for the optimization.
   arma::mat& Parameters() { return parameter; }
 
+  size_t& Rho() { return rho; }
```
zoq: Can you add a comment here?

src/mlpack/tests/ann_layer_test.cpp
```diff
@@ -779,6 +780,92 @@ BOOST_AUTO_TEST_CASE(GradientLSTMLayerTest)
...
+    arma::mat input, target;
+  } function;
+
+  BOOST_REQUIRE_LE(CheckGradient(function), 1e-4);
+}
+
+/**
+ * GRU layer manual forward test.
+ */
+BOOST_AUTO_TEST_CASE(ForwardGRULayerTest)
```
zoq: Can we add a similar test for the backward function?

sumedhghaisas: umm... I can. But I thought if the gradients are right at each layer then backward has to be correct right?

src/mlpack/methods/ann/init_rules/unit_init.hpp
sumedhghaisas: Yeah we can. I thought a separate initialization policy is maybe simpler to read. And also can be used for different purposes later. The better design would be to have a constant initialization function and make zero and 1 special cases. Is that better?

zoq: ``ConstIntialization`` sounds like a good to me.

src/mlpack/tests/ann_layer_test.cpp
zoq: Yeah, you are right, it's just an additional test.

src/mlpack/methods/ann/init_rules/const_init.hpp
```diff
@@ -1,33 +1,37 @@
...
 namespace ann /** Artificial Neural Network. */ {
 
 /**
- * This class is used to initialize randomly the weight matrix.
+ * This class is used to initialize weight matrix with constant values.
+ *
+ * @tparam initConst Const value to initialize.
  */
-class ZeroInitialization
+template<size_t initConst>
```
zoq: I think we should use the constructor to pass the const value, that way we could also use ``double``as type and pass e.g.: ``0.5``.

rcurtin: Agreed, I think the time savings from using a compile-time constant aren't going to be noticeable in this case, and it's pretty restrictive to disallow non-size_t values.

src/mlpack/tests/ann_layer_test.cpp
```diff
@@ -779,6 +780,92 @@ BOOST_AUTO_TEST_CASE(GradientLSTMLayerTest)
 }
 
 /**
+ * GRU layer numerically gradient test.
```
rcurtin: This is a nice test, thanks for writing it.  It might be good to expand the comment a little bit here; it took me a few minutes to fully understand it.  To me it looks like the test ensures that an approximation of the gradient is close enough to the actual gradient returned by the `GRU<>` layer.

src/mlpack/tests/ann_layer_test.cpp
```diff
@@ -779,6 +780,92 @@ BOOST_AUTO_TEST_CASE(GradientLSTMLayerTest)
...
+
+  arma::mat input = arma::ones(3, 1);
+  arma::mat output;
+  gru.Forward(std::move(input), std::move(output));
+
+  arma::mat expectedOutput = arma::ones(3, 1);
+  expectedOutput *= -4;
+  expectedOutput = arma::exp(expectedOutput);
+  expectedOutput = arma::ones(3, 1) / (arma::ones(3, 1) + expectedOutput);
+  expectedOutput = (arma::ones(3, 1)  - expectedOutput) % expectedOutput;
```
rcurtin: If you can add a comment here on why this is the correct output, it will probably be really helpful to someone in the future who is trying to debug a failure. :)

src/mlpack/tests/ann_layer_test.cpp
rcurtin: I took a look at `Backward()`, I think the code there is non-trivial, not just a call to `Gradient()` or anything.  So even if the gradient test is passing, I think it would still be a good idea to write a test for `Backward()`.

src/mlpack/methods/ann/layer/gru_impl.hpp
rcurtin: Correct me if I am misunderstanding---I think the first element of `outParameter` will always be a matrix of zeros and it is never modified.  This leads me to three thoughts:

1) If `outParameter.front()` is always `arma::zeros<arma::mat>(outSize, 1)` and `outSize` does not change between calls to `ResetCell()` we can just do:

```
arma::mat tmp = std::move(outParameter[0]);
outParameter.clear();
outParameter.push_back(arma::mat()); // not 100% sure on the syntax here, I think this is right
outParameter.front() = std::move(tmp); // maybe this could be compressed with the previous line but you get the idea
```

2) In another place you also push back a zeros matrix onto `outParameter`.  If `outParameter` held pointers to matrices instead, you could do `outParameter.push_back(outParameter.front())`.  But you could also create an alias matrix with one of the special Armadillo constructor, of `outParameter.front()`, and push that onto the list.

3) Would it be possible to avoid allocating the zeros matrix entirely?  And then detect when you are at the front of `outParameter`?  You use `outParameter` in code like...

```
arma::mat d_zt = gy % (*backIterator - stuff);
```

but if we could instead do something like

```
if (we are at the first element)
  arma::mat d_zt = gy % stuff;
else
  arma::mat d_zt = gy % (*backIterator - stuff);
```

then we can save a possibly significant amount of computation.  Since we are pushing the zeros matrix into something of type `arma::mat`, it is no longer known at compile time that the matrix only has zeros.  Thus Armadillo is not able to see something like `*backIterator - stuff` and know that `*backIterator` is actually a matrix of zeros, and an unnecessary loop over the values happens.

The third comment is the most complex of the three options, but it will result in the most speedup so I think it would be worth it to implement.  If you don't want to do that now, we should open an issue for it.  But at the very least we should do #1 (and #2) if not, plus possibly the improvement suggested by Marcus above.

src/mlpack/methods/ann/init_rules/const_init.hpp
sumedhghaisas: I agree. Will make the appropriate change.

src/mlpack/tests/ann_layer_test.cpp
sumedhghaisas: For Backward() test, the code is definitely non-trivial but the way we test gradients, if the Backward function is not correct for all the layers, the gradients of the input won't be correct. The gradients computed at each layer depends on the signal coming from the layer above, which is computed by the backward call. Unless the backward function of the layer above is correct, the gradient of the layer below won't be correct. Writing the test is not a very complex task, but I feel that it is unnecessary.

rcurtin: Fair enough---having a separate test for `Backward()` can help rule out whether a failure in the gradient test is caused by an incorrect `Backward()` implementation or an incorrect implementation in some other part of the code that is being tested.  But if it is tedious to add, I am ok with leaving it out, as long as the `Backward()` implementation is indeed being tested at least somewhere.

src/mlpack/methods/ann/layer/gru_impl.hpp
```diff
@@ -0,0 +1,324 @@
...
+    gy += boost::apply_visitor(deltaVisitor, output2GateModule);
+  }
+
+  if (backIterator == outParameter.end())
+  {
+    backIterator = --(--outParameter.end());
+  }
+
+  // Delta zt.
+  arma::mat d_zt = gy % (*backIterator -
```
zoq: Pedantic style issue; can you use camel casing for all names?

src/mlpack/methods/ann/layer/gru_impl.hpp
```diff
@@ -0,0 +1,324 @@
...
+  backIterator = outParameter.end();
+  gradIterator = outParameter.end();
+}
+
+template<typename InputDataType, typename OutputDataType>
+template<typename eT>
+void GRU<InputDataType, OutputDataType>::Forward(
+    arma::Mat<eT>&& input, arma::Mat<eT>&& output)
+{
+  arma::mat temp = arma::zeros<arma::mat>(outSize, 1);
```
zoq: It looks like this parameter is unused.

src/mlpack/methods/ann/layer/gru.hpp
```diff
@@ -0,0 +1,248 @@
...
+ * http://www.opensource.org/licenses/BSD-3-Clause for more information.
+ */
+#ifndef MLPACK_METHODS_ANN_LAYER_GRU_HPP
+#define MLPACK_METHODS_ANN_LAYER_GRU_HPP
+
+#include <list>
+
+#include <mlpack/prereqs.hpp>
+
+#include <boost/ptr_container/ptr_vector.hpp>
```
zoq: I think, we can remove the header.

src/mlpack/methods/ann/layer/gru_impl.hpp
```diff
@@ -0,0 +1,322 @@
...
+      *backIterator;
+
+  // Delta of forget gate.
+  boost::apply_visitor(BackwardVisitor(std::move(boost::apply_visitor(
+      outputParameterVisitor, forgetGateModule)), std::move(dRt),
+      std::move(boost::apply_visitor(deltaVisitor, forgetGateModule))),
+      forgetGateModule);
+
+  // Put delta zt.
+  prevError.submat(0, 0, 1 * outSize - 1, 0) = boost::apply_visitor(
```
zoq: That works if we use a batchSize = 1, do you mind to refactor the parameter so that it works with batchSize > 1? You could get the batchSize by looking at ``input.n_cols``, so instead of using ``prevError = arma::zeros<arma::mat>(3 * outSize, 1);`` we have to use ``prevError.set_size(4 * outSize, batchSize);`` and ``prevError.submat(outSize, 0,  2 * outSize - 1, batchSize - 1)``. I can refactor the LSTM layer, but it would be great if you could update the GRU layer.

zoq:
I think once, we fixed the issue with the batch size, this is ready to be merged.

src/mlpack/methods/ann/layer/gru_impl.hpp
sumedhghaisas: Ohh. I did not know that batchSize is also a factor. I will refactor GRU and LSTM.

sumedhghaisas: Is the batchSize constant, or may change during training?

rcurtin: I am not 100% sure on this, so @zoq please correct me if I am wrong, but I think that for now `batchSize` will be constant during training.  If we wanted the user to be able to set `batchSize` differently during training, we would probably need to add a `BatchSizeVisitor` that could operate on many different layer types, so a change like that I think would be outside the scope of this PR.

zoq: The layer should be able to handle different sizes, but I agree adding another visitor shouldn't be part of this PR.

It's correct that in most cases the batchSize is constant, during training, I guess we could say that we only train on full batches, however it probably makes sense to evaluate more or less samples during the prediction process.

sumedhghaisas: So should I resize prevError in every call or set it once and keep using it?

zoq: We could do something like:

```
if (input.n_cols != batchSize)
{
  batchSize = input.n_cols;
  // reset prevError;
}
```

What do you think?

rcurtin:
@sumedhghaisas: just a quick question, have you managed to fix the batch size issue yet?  I'd like to merge this one, it would make the NTM PR a lot easier to review. :)

It looks like there are a few other comments that haven't been addressed, can you go back through and make sure you take care of them all?  Here is a list of what I see:

> Do we also need to serialize network here?

> Can you add a comment here?

> I think, we can remove the header.

Great work with the code here, I am excited to get this merged and start playing with it (if I can ever find some free time to do that!). :)

sumedhghaisas:
@zoq @rcurtin I am not sure how the serialization exactly works. I looked at the code in FFN and RNN, so I think it stores the weight matrix and loads it with weight set visitor. Weight set visitor iterates through Model as well, for layers which have an internal model. Where is the individual Serialize function call for all the layers? Are those necessary? Cause we are storing all the parameters in global weight matrix.

zoq:
As you already pointed out serialization of layer weights isn't necessary, but serialization of layer specific parameters; like the dropout rate for the Dropout layer is something that we need to restore the layer or the number of steps to backpropagate through time for the GRU.

sumedhghaisas:
In that case... this branch is finally ready to be merged :)

zoq:
Unfortunately we have to update the PR, since https://github.com/mlpack/mlpack/pull/1005 was merged before, mainly adjust the rho part. Let me know if you need help with the modifications.

rcurtin:
@sumedhghaisas: can you handle the updates for the #1005 merge?  I was hoping I would be able to merge this today, but this PR is waiting on those changes. :)  Thanks!

sumedhghaisas:
@zoq Hey Marcus. I looked at the related code changes in PR 1005. I think there we are assuming the rho of the network and the rho of the layer will always be same? cause the RhoSetVisitor sets the rho for all the layers. Maybe I missed something? I remember discussing this with you. The whole point of adding zeros in the middle of BPTT was to break the chain, which both LSTM and GRU now support with this PR. For this the rho of layers may differ from the rho of RNN.

zoq:
Yes, I remember the discussion; we needed something to test the tasks, your solution is way more elegant, so I propose to replace the current solution.

sumedhghaisas:
So I rebased the branch on master and made necessary changes to RNN module regarding 'rho'.

sumedhghaisas:
I also replaced the EmbeddedReberTest with RecursiveReberTest. RecursiveReberTest is a generalization of EmbeddedReberTest. In embedded tests, a reber is embedded inside another reber, where as in recursive reber recursively embeds a reber inside given depth. 

But then, I thought of making it better so rather than constant depth in recursive reber, I added options of average depth and max depth. This is where things get interesting. In a normal reber, a start symbol is 'B' and then a 'P' or 'T'. So I assigned 'P' to recursion and 'T' to termination. So after selecting 'B', the reber generator selects either 'P' or 'T' given probability, which you can guess is 1 / (average depth). If 'P' comes, it selects again, but if 'T' comes then the selection stops and a normal reber string is attached to the current construction. After this, the entire string is closed with proper closing sequence.
In context free grammar - 
s -> B a E
a -> P a P
a -> T r T
r -> {reber string}

For precaution, I have added the max depth option which terminates 'P'/'T' selection if the depth reaches maximum, and forces the selection of 'T'.

Experiments with such recursive reber are more interesting that constant depth one. Here, the LSTM has  to remember how many 'P' to reproduce while ending the sequence, which is kind of stack storage. Both GRU and LSTM, pass recursive reber test for recursive depth of 3 (max 5) with cell state size of 15. No cell size less than 15 passes the test. 

It would be an interesting experiment to see the relation between average depth and the cell state size required. I will set that up in models when I get time.

zoq:
Really nice idea and description, also I like the idea running some experiments on the task, I guess if we do something like that a simple graphic might help to understand the **"Recursive" Embedded Reber Grammar task**, something like: https://www.willamette.edu/~gorr/classes/cs449/reber.html

rcurtin:
I think some of the refactoring here has introduced a bug in the LSTM class, based on the Travis output.  Is it better to revert the changes for `Rho()` and merge GRU, then open another PR for the `Rho()` changes?

Overall I agree though, I think `ResetCells()` is a much nicer solution.

One additional thought I have is that allocating memory for the `allZeros` matrix is not the fastest thing to do.  Armadillo will still go through all the effort of using that `allZeros` matrix, but really it might be better to do something like

```
if (firstStep)
{
  // Perform computation with arma::zeros()
}
else
{
  // Perform computation as usual.
}
```

Although the code gets a little longer, it is surely faster to check the condition than to perform the operation with zeros.

sumedhghaisas:
hmm... I am not sure. The gradient tests are passing. And the Rho() changes majorly in RNN module. Do you want me to remove the ResetCell code?

sumedhghaisas:
So I checked the diff between the current RNN, GRU and LSTM vs my version of them before I rebased. There is not any algorithmic difference. I wonder why is this happening.

sumedhghaisas:
And regarding the AllZeroes thing. Most of the current usage of LSTM and GRU or any other recurrent unit uses a linear layer initialization. We may need to come up for with a design for that. Thats why I thought there is no point in specializing for zeroes. But we can maybe use some specialization tricks?

sumedhghaisas:
Okay. So I reverted back to the old network sizes for the tests. That solved the problem.

rcurtin:
> And regarding the AllZeroes thing. Most of the current usage of LSTM and GRU or any other recurrent unit uses a linear layer initialization. We may need to come up for with a design for that. Thats why I thought there is no point in specializing for zeroes. But we can maybe use some specialization tricks?

Hm, I am not sure, I think maybe we are talking about different things.  But I think, in the end, my idea will not work, because the visitors require an `arma::mat` to be passed (so we can't pass some expression like `arma::zeros<arma::mat>(...)` which simply represents a zeros matrix without allocating it).  Therefore I think what you've done seems like the best option for now, and nevermind about my comments. :)

Thanks for solving the LSTM test issue!

rcurtin:
I merged and also fixed the static code analysis issues.  Nice work with the GRU!  I am happy that we have merged it now.  :)

PR #1019, opened 2017-06-05
Style fixes - Part 2

zoq:
I've noticed that the style check only checked the source files in src/mlpack/core, this was fixed in https://github.com/mlpack/jenkins-conf/commit/b63e8ae30ce46c8f16e03d49e5fe61142427c079.


zoq:
Okay, great so +409 issues, I think I fixed most of the issues, but the question is, is a single commit fine, or should we split the fixes e.g. one commit for each file/method, etc.?

rcurtin:
One commit is fine with me, mostly I'm just going to scroll through the combined PR diff in Github and see if anything seems out of place.

zoq:
Okay, I'm not sure what the issues are with ``core.hpp``, I guess unless anybody else has an idea, we can whitelist the file. Which I think is fine, since we check the style anyway and if somebody changes something in ``core.hpp`` which is rare we are probably able to point out the issues.

zoq:
Not sure what we should do with the ``rand`` issue, we can filter it out.

src/mlpack/methods/ann/visitor/copy_visitor.hpp
```diff
@@ -26,7 +26,7 @@ class CopyVisitor : public boost::static_visitor<LayerTypes>
 {
  public:
   template <typename LayerType>
-  LayerTypes operator () (LayerType* ) const;
+  LayerTypes operator () (LayerType*) const;
```
rcurtin: I think in other places for operators we've done, e.g., `operator=(args& args...)`, so I would expect here it might be `operator()(LayerType*)`; is this a gray area in the style check rules?

src/mlpack/methods/rann/krann_main.cpp
```diff
@@ -31,10 +31,10 @@ PROGRAM_INFO("K-Rank-Approximate-Nearest-Neighbors (kRANN)",
     "This program will calculate the k rank-approximate-nearest-neighbors of a "
     "set of points. You may specify a separate set of reference points and "
     "query points, or just a reference set which will be used as both the "
-    "reference and query set. You must specify the rank approximation (in \%) "
+    "reference and query set. You must specify the rank approximation (in %%) "
```
rcurtin: Since this isn't being given to printf, I think that just `%` is fine here.

src/mlpack/tests/load_save_test.cpp
```diff
@@ -1738,15 +1738,15 @@ BOOST_AUTO_TEST_CASE(HarderARFFTest)
   f << endl;
   f << "@attribute @@@@flfl numeric" << endl;
   f << endl;
-  f << "\% comment" << endl;
+  f << "%% comment" << endl;
```
rcurtin: Same here, I think one `%` is fine.

src/mlpack/methods/ann/visitor/copy_visitor.hpp
zoq: hm, good point, I'll take a look into the config.

src/mlpack/methods/rann/krann_main.cpp
zoq: Right, will change that.

PR #1020, opened 2017-06-06
Add a method form detection tool

micyril:
Add a tool that allows to check at compile time whether a given class
has a method of the requested form with the specified number of
additional arguments. See #929 for additional comments.


micyril:
It looks like some tests randomly fail since my last commit just adds a description of the test.

src/mlpack/core/util/sfinae_utility.hpp
```diff
@@ -15,6 +15,110 @@
...
+ * @tparam MethodForm A template class member pointer type to a method of the
+ *   form to look for.
+ * @tparam Class A class in which a method of the requested form should be
+ *   looked for.
+ * @tparam AdditionalArgsCount A number of additional arguments.
+ */
+template<typename Class,
+         template<typename...> class MethodForm,
+         size_t AdditionalArgsCount>
+struct MethodFormDetecter;
```
rcurtin: Can you change the name to `MethodFormDetector` please?  Just a slight misspelling :)

rcurtin:
I should add, the Radical test failure has nothing to do with your code, so I am not worried about it.  See #922 for more information.

src/mlpack/core/util/sfinae_utility.hpp
micyril: Oh, sure

micyril:
I want to add some more related code here, so to review it all at once.

rcurtin:
Ok, sure, I will not merge it until you let me know that it is ready then, and I will review more as you add it.

micyril:
I think it's ready for reviewing now.

src/mlpack/tests/sfinae_test.cpp
```diff
@@ -0,0 +1,124 @@
...
+
+  static_assert(!HasM<B, MForm1>::value, "value should be false");
+
+  static_assert(!HasM<A, MForm2>::value, "value should be false");
+
+  static_assert(HasM<B, MForm2>::value, "value should be true");
+
+  static_assert(HasTemplatedM<B, MForm2>::value, "value should be true");
+
+  static_assert(HasVeryTemplatedM<B, MForm3>::value, "value should be true");
```
rcurtin: I think you should use `BOOST_REQUIRE_EQUAL()` or `BOOST_REQUIRE()` here, so that upon failure the Boost UTF will fail in the expected way (and also so it correctly counts the successes in the test output).

micyril: My motivation was to check that values can be evaluated at compile time, but if `BOOST_REQUIRE()` is more convenient, let's use it.

micyril:
Yes, it's ready if there are no more questions.

src/mlpack/tests/sfinae_test.cpp
rcurtin: I see, I agree with you then, evaluating at compile time would probably be better.  I'll merge and then revert the change to `BOOST_REQUIRE()`.  Not a big deal either way I don't think :)

Issue #1011, opened 2017-05-26
PVS-Studio static code analyzer results


zoq:
The following is the output of the PVS-Studio static code analyzer, I think we can probably ignore most of the warnings like ``Not all members of a class are initialized inside the constructor.``, but there are some warnings that might be interesting. I've also tested clang's static code analyzer, no remarks at all.

- **mlpack/core/util/cli.hpp**  173 warn  V690 
``Copy constructor is declared as private in the 'CLI' class, but the default '=' operator will still be generated by compiler. It is dangerous to use such a class.``

- **mlpack/core/util/cli_impl.hpp** 68  warn  V688 
``The 'parameters' local variable possesses the same name as one of the class members, which can result in a confusion.``

- **mlpack/core/util/cli_impl.hpp** 70  warn  V688
``The 'aliases' local variable possesses the same name as one of the class members, which can result in a confusion.``

- **mlpack/core/util/cli_impl.hpp** 131 warn  V688
``The 'desc' local variable possesses the same name as one of the class members, which can result in a confusion.``

- **mlpack/core/dists/gaussian_distribution.hpp** 45  warn  V730
``Not all members of a class are initialized inside the constructor. Consider inspecting: logDetCov.``

- **mlpack/core/dists/gaussian_distribution.cpp** 22  warn  V730
``It is possible that not all members of a class are initialized inside the constructor. Consider inspecting: logDetCov.``

- **mlpack/core/optimizers/lbfgs/test_functions.cpp** 94  warn  V636
``The '1 / 10' expression was implicitly cast from 'int' type to 'double' type. Consider utilizing an explicit type cast to avoid the loss of a fractional part. An example: double A = (double)(X) / Y;.``

- **mlpack/methods/adaboost/adaboost_impl.hpp** 56  warn  V730
``Not all members of a class are initialized inside the constructor. Consider inspecting: classes, ztProduct.``

- **mlpack/core/tree/ballbound_impl.hpp** 78  err V591
``Non-void function should return a value.``

- **mlpack/core/tree/binary_space_tree/binary_space_tree.hpp**  54  warn  V690
``The 'BinarySpaceTree' class implements a copy constructor, but lacks the '=' operator. It is dangerous to use such a class.``

- **mlpack/core/tree/rectangle_tree/rectangle_tree.hpp**  53  warn  V690
``The 'RectangleTree' class implements a copy constructor, but lacks the '=' operator. It is dangerous to use such a class.``

[...]

Issue #1021, opened 2017-06-07
Use pass-by-value instead of two overloads for copy and move


rcurtin:
Instead of doing the following:

```
void Method(const arma::mat& data); // makes copy of data
void Method(arma::mat&& data); // doesn't copy data, just moves it
```

we should instead do this:

```
void Method(arma::mat data);
```

In this way, a user can choose whether they want to copy data by calling `Method(data)` or avoid a copy by calling `Method(std::move(data))`, and it allows us to reduce the size of the codebase (sometimes significantly).

More information here:
http://cpptruths.blogspot.com/2012/03/rvalue-references-in-constructor-when.html
and this issue comes out of the discussion for #1014 so there is some more discussion there too.

We can close this issue when we have found and replaced all instances of the pattern, and when we have provided good documentation for how to use these methods and avoid copies.  I would suggest that we add some phrase like `be sure to pass the parameters with std::move() to avoid copies when possible` to the documentation of every function like this, since I don't necessarily expect the C++ users to be familiar with rvalue references in C++11, but that isn't the only option for providing good documentation.

ShangtongZhang:
I suddenly have a new concern. Assume user has a `arma::mat data` and two models `m1`, `m2`, then user wants to call `m1.Predict(data)` then `m2.Predict(data)`. (Assume the logic within Predict can actually work well with `const mat&` and won't lead to any extra copy). In this case user can't use `m1.Predict(std::move(data))` then `m2.Predict(std::move(data))`. If user wants to call `Predict` of several different models `n` times, there would be at least `n-1` copies. But actually there can be no copy. Should we worry about this case? Or should the pass-by-value only be applied to constructor? As we have decided not to store reference in class so at least one copy in constructor is necessary.

rcurtin:
There is only a need to pass-by-value in instances where the method actually modifies the data.  So if we have a method `Predict()` which doesn't need to modify the input data, then its signature should be

```
Predict(const MatType&);
```

but if you have a method that's actually going to modify the input data, then you need to copy it.  i.e. when we build kd-trees, we modify the ordering of the dataset and thus we can do pass-by-value there.  But when we build cover trees, the dataset on which the tree is built is not modified, so there we can do pass-by-const-reference.

ShangtongZhang:
[...]

PR #1022, opened 2017-06-08
PVS-Studio fixes

rcurtin:
This is a work-in-progress PR for #1011.  I've solved some of the issues, but not yet all of them.

### DIFF
--- a/src/mlpack/core/dists/gaussian_distribution.cpp
+++ b/src/mlpack/core/dists/gaussian_distribution.cpp
@@ -19,7 +19,7 @@ using namespace mlpack::distribution;
 
 GaussianDistribution::GaussianDistribution(const arma::vec& mean,
                                            const arma::mat& covariance)
-  : mean(mean)
+  : mean(mean), logDetCov(0) /* will be reset in Covariance() */
 {
   Covariance(covariance);
 }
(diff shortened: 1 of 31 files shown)


ShangtongZhang:
I suddenly have a new concern. Assume user has  a `arma::mat data` and two models `m1`, `m2`,  then user wants to call `m1.Predict(data)` then `m2.Predict(data)`. (Assume the logic within `Predict` can actually work well with `const mat&` and won't lead to any extra copy). In this case user can't use `m1.Predict(std::move(data))` then `m2.Predict(std::move(data))`.  If user wants to call `Predict` of several different models `n` times, there would be at least `n-1` copies. But actually there can be no copy. Should we worry about this case? Or should the pass-by-value only be applied to constructor? As we have decided not to store reference in class so at least one copy in constructor is necessary.

zoq:
@ShangtongZhang do you mind to ask the question over here: https://github.com/mlpack/mlpack/issues/1021? I'm not sure you can remove the comment, if you like I should be able to do this for you; not sure that's necessary.

ShangtongZhang:
@zoq Sure, I moved the comment to that issue. But here comes a new one.. please feel free to delete this one.

Issue #514, opened 2016-02-02
mlpack KMeans class much slower than armadillo kmeans()


jciminogh:
I seem to be getting much faster clustering from armadillo's [kmeans()](http://arma.sourceforge.net/docs.html#kmeans) function in comparison to mlpack's kmeans::KMeans<> class, which is about 2x to 6x slower. Using the latest mlpack code from the git repo. Am I doing something wrong?

Using the code below I get the following timings on my machine (Intel i5, 64 bit, g++ version 5.3).

Compiled without openmp:
`g++ kmeans_test.cpp -o kmeans_test -O3 -std=c++11 -larmadillo -lmlpack`
mlpack_kmeans time: 17.3024
arma::kmeans time: 9.16399

Compiled with openmp:
`g++ kmeans_test.cpp -o kmeans_test -O3 -std=c++11 -larmadillo -lmlpack -fopenmp`
mlpack_kmeans time: 17.7575
arma::kmeans time: 2.87675

```
#include <fstream>
#include <mlpack/methods/kmeans/kmeans.hpp>
#include <armadillo>

int main() {
    arma::uword dims = 20; // number of dimensions
    arma::uword samples = 5000000;

    arma::uword max_iterations = 10;
    arma::uword k = 10;  // number of clusters

    arma::arma_rng::set_seed_random(); // random start

    std::cout << "Generating some synthetic data ..." << std::endl;

    arma::mat data(dims, samples, arma::fill::zeros);

    // generate data with unique centroids, added with a small amount of noise
    for (arma::uword i=0; i<samples; i++) {
        arma::uword c = as_scalar( arma::randi<arma::uvec>(1, arma::distr_param(0,k-1)) );
        data.col(i) = arma::linspace<arma::vec>(c, c+dims-1, dims) + 0.25*arma::randn<arma::vec>(dims);
      }

    arma::wall_clock timer;

    std::cout << "mlpack_kmeans start ..." << std::endl;

    arma::Row<size_t> mlpack_assignments;
    arma::mat mlpack_centroids;

    mlpack::kmeans::KMeans<> mlpack_kmeans(max_iterations);

    timer.tic();
    mlpack_kmeans.Cluster(data, k, mlpack_assignments, mlpack_centroids);

    std::cout << "mlpack_kmeans time: " << timer.toc() << std::endl;

    std::cout << "---" << std::endl;
    std::cout << "arma::kmeans start ..." << std::endl;

    arma::mat arma_centroids;

    timer.tic();
[...]

PR #1024, opened 2017-06-12
Parallel implementation of NaiveKMeans

shikharbhardwaj:
Add a parallel implementation of NaiveKMeans, with the same interface.

### Testing
Right now, I just replicated the first test in the existing tests on KMeans, changing the `LloydStepType` from the default (`NaiveKMeans`) to `ParallelNaiveKMeans`.

### Speedup
Around 2.33x speedup on 4 threads with a simple benchmark using the Adult dataset with 32560 instances. [The measurements.](https://gist.github.com/shikharbhardwaj/89bf1f1e58b3730b4bfc00bb6373e222)


mentekid:
Thanks for opening the PR. These are some nice speedups.

I think, since the code remains simple enough after the parallelization, we could get away with directly modifying [naive_kmeans_impl.hpp](https://github.com/mlpack/mlpack/blob/master/src/mlpack/methods/kmeans/naive_kmeans_impl.hpp) instead of adding another file to the folder.

If we do it like that, people will benefit from the parallelism without explicitly specifying the parallel module. It might even speed up the other implementations such as dual trees, which you mentioned use the naive method.
What do you think?

@rcurtin, @zoq would that make the codebase too complicated in your opinions?

rcurtin:
I believe that there is no difference between `NaiveKMeans` and `ParallelNaiveKMeans` other than the OpenMP bits, so in my opinion it's better to just modify `NaiveKMeans` and add OpenMP support there.  If a user doesn't want OpenMP to parallelize it for some weird reason, they can either specify `OMP_NUM_THREADS=1` or they can simply disable OpenMP in the CMake configuration (hmm, I'm not sure if that's actually an option that's currently implemented.  If it's not, I'm not too worried about it though, because I doubt anyone will want *no* OpenMP at all, unless it's not installed, in which case CMake already ignores it.  Sorry for the tangent!).

shikharbhardwaj:
Great. I'll delete the extra file and test case.

shikharbhardwaj:
I was going through the issues with kmeans. I will test whether this will be fast enough to solve #514 .

rcurtin:
I think Armadillo's `kmeans()` implementation is only parallel via OpenBLAS, meaning only some of the operations will be parallelized, so it's possible that the OpenMP work you've done here will outperform that, which would be great (I love when old bugs can be closed!).

shikharbhardwaj:
I tried the benchmark code given in the issue. This implementation is still ~2x slower than Armadillo with OpenMP. This is faster than the previous implementation, which was 4x slower with OpenMP enabled.

shikharbhardwaj:
I tried a few optimizations and have some more speedups.

1. Using `unsafe_col` in `metric.Evaluate` speeds things up quite a bit. But this function is not available in the Sparse matrix type. A template specialisation for sparse matrices is needed.

2. Avoid the copy by constructing `arma::vec` at line 80. Since the class accepts an arbitrary matrix type, this is needed.

Overall, these two make the mlpack implementation around 1.2x slower than the Armadillo one.

src/mlpack/methods/kmeans/naive_kmeans_impl.hpp
```diff
@@ -38,28 +40,52 @@ double NaiveKMeans<MetricType, MatType>::Iterate(const arma::mat& centroids,
...
 
-    // We now have the minimum distance centroid index.  Update that centroid.
-    newCentroids.col(closestCluster) += arma::vec(dataset.col(i));
-    counts(closestCluster)++;
+      // We now have the minimum distance centroid index.  Update that centroid.
+      localCentroids.unsafe_col(closestCluster) += dataset.col(i);
+      localCounts(closestCluster)++;
+    }
+    // Combine calculated state from each thread
+    #pragma omp critical
```
mentekid: I haven't tested this and you probably have, so what I'm saying is dumb - but: Have you tried expressing these as an OpenMP reduction? Does it work for the overloaded `+` operator of armadillo matrices?
If it did, we could remove the critical section, right?

shikharbhardwaj: I had checked for this. :) Yes, that would be enough to remove the critical section.
But OpenMP as of now does not support reductions on non-POD types (or custom user-specified reductions). https://stackoverflow.com/questions/16176155/openmp-reduction-with-overloaded-operator
I think this is quite minimal synchronisation overhead. Armadillo's `kmeans` does this the hard way by splitting up the data among the number of threads, keeping track of that data and reducing it outside the parallel region. That makes for quite a lot of code.

shikharbhardwaj: One thing I had in my mind, we need not do both the operations in the critical section at once. I'll try and give separate critical sections to both the operations to see if there is any effect on the runtime.

rcurtin: I am not sure if this is helpful but you can get the memory pointer of a matrix with `.memptr()`, which might allow you to write a reduction.  Though even if you did I am not sure how much benefit would be seen.

shikharbhardwaj: I am not too sure about how that would make a reduction possible. Any variable inside the reduction clause is restricted to be a scalar (OpenMP does not define what scalar means, but it most probably  means numeric POD types).
Each thread calculates its share of centroids and counts, so the reduction must be able to sum up the centroids and counts which are vectors themselves. I can't think of a way to achieve this in the  absence of user-defined reductions.
I'll measure how much time is spent on waiting for the critical section to unlock.

shikharbhardwaj: I made a few measurements and found that the time spent in the critical section by a thread is quite varied, but small. With a sample size of 8000 measurements, the average time was around 365ns. This measurements is quite shaky, as the resolution of the clock on my system is around 20ns. The maximum time spent was around 82us. But this is too low to affect the runtime significantly.

mentekid: Nice, thanks for doing the correct thing (profiling) and not what I did (guessing).

I'll take a closer look on the algorithm (this time, I solemnly swear to profile), and see if I can come up with any other potential optimizations.

However, 1.2x the Armadillo times is still a significant bump :)

shikharbhardwaj:
This final commit sets the mlpack implementation at max around 10% slower than the Armadillo implementation of KMeans(when both are run with 4 threads), according to the benchmark given by #514 although the benchmark is not very stable (both the implementation may start at random centroids).

The mlpack implementation simply does more work than `arma::kmeans` by computing assignments for each point in the dataset. Without the assignment computation, the mlpack implementation finishes within 1% of `arma::kmeans`.

src/mlpack/methods/kmeans/naive_kmeans_impl.hpp
```diff
@@ -38,28 +40,52 @@ double NaiveKMeans<MetricType, MatType>::Iterate(const arma::mat& centroids,
...
+                             arma::fill::zeros);
+    arma::Col<size_t> localCounts(centroids.n_cols, arma::fill::zeros);
 
-    for (size_t j = 0; j < centroids.n_cols; j++)
+    #ifdef _WIN32
+    // Tiny workaround: Visual Studio only implements OpenMP 2.0, which doesn't
+    // support unsigned loop variables. If we're building for Visual Studio, use
+    // the intmax_t type instead.
+    #pragma omp for
+    for (intmax_t i = 0; i < (intmax_t) dataset.n_cols; ++i)
```
rcurtin: Hmm, I think this pattern appears a bunch of places now.  We might want to add to `prereqs.hpp`:

```
// This can be removed with Visual Studio supports an OpenMP version with unsigned loop variables.
#ifdef _WIN32
  #define omp_size_t intmax_t
#else
  #define omp_size_t size_t
#endif
```

Then for the loops,

```
#pragma omp for
for (omp_size_t i = 0; i < (omp_size_t) dataset.n_cols; ++i)
```

But maybe that is out of scope for this PR.  I just wanted to point it out as an idea.

rcurtin:
Nice!  I think that would be sufficient to close #514; Armadillo doesn't calculate the assignments and that definitely is going to affect the resulting runtime.  The benchmarking system (https://github.com/mlpack/benchmarks) is setup to start from the same initial centroids when comparing against other libraries, so it's possible that when this code is run on that system we will get some more stable benchmark comparisons.  Everything looks great to me, thank you for looking into this so effectively.  Looking forward to the merge. :)

src/mlpack/methods/kmeans/naive_kmeans_impl.hpp
shikharbhardwaj: I found this in the LSH search code. I'll find other places where this comes up and just do it here in this PR. Then I think this will be ready for merge. :)

mentekid:
If there are no more comments I will go ahead and merge this one.

src/mlpack/methods/kmeans/naive_kmeans_impl.hpp
```diff
@@ -38,28 +40,44 @@ double NaiveKMeans<MetricType, MatType>::Iterate(const arma::mat& centroids,
...
+      // Find the closest centroid to this point.
+      double minDistance = std::numeric_limits<double>::infinity();
+      size_t closestCluster = centroids.n_cols; // Invalid value.
 
-      if (distance < minDistance)
+      for (size_t j = 0; j < centroids.n_cols; j++)
       {
-        minDistance = distance;
-        closestCluster = j;
+        const double distance = metric.Evaluate(dataset.col(i),
```
zoq: Pedantic style issue: when wrapping a line, the next line should be tabbed twice: for more information take a look at: https://github.com/mlpack/mlpack/wiki/DesignGuidelines#line-length-and-wrapping. This also applies to  ``localCentroids``.

shikharbhardwaj:
I have made the style fixes. Ready to merge. :)
Just a reminder, this merge closes #514.

mentekid:
Thanks for noticing that Marcus.
I think this is ready to merge but I'll leave it open for the weekend, that way anybody that has comments can pitch in.

PR #1027, opened 2017-06-14
[WIP]Restricted Boltzmann Machine

kris-singh:
I have impmented the 
1. Base layer(visible and hidden layer)
2. RBM(wrapper)
3. Cd-k algorithm(In progress)

I will add test after the cd-k algorithm is added


kris-singh:
I am also confused about the way to handle the biases of the hidden and visible layers.
Visible layer uses the biases of the hidden layer. Right now we have 2 biases for every layer 
ownBias and otherBias the problem with that is when we initialize the parameters of the layer we have to swap the how the parameters is initialized using the initializer.

src/mlpack/methods/ann/layer/binary_rbm.hpp
```diff
@@ -0,0 +1,130 @@
...
+  /**
+   * Sample the output given the input parameters
+   * The sample are obtained from a distribution 
+   * specified by the sampler. 
+   *
+   * @param input the input parameters for 
+   * @param output samples from the parameters
+   */
+  template<typename eT>
+  void Sample(arma::Mat<eT>&& input, arma::Mat<eT>&& output);
```
lozhnikov: What is the difference between `Sample()` and `Forward()`? Why do you don't want to merge these functions into one?

src/mlpack/methods/ann/layer/binary_rbm.hpp
```diff
@@ -0,0 +1,130 @@
...
+  OutputDataType weights;
+
+  //! Locally-stored weight paramters.
+  OutputDataType weight;
+
+  //! Locally-stored weight paramters .
+  OutputDataType ownBias;
+
+  //! Locally-stored bias of the other layer.
+  OutputDataType otherBias;
```
lozhnikov: Hidden variables are sampled from `P(h=1|v)=Sigm(W^t * v + c)`. The weight matrix `W` is shared between the visible and the hidden layers. The matrix `W` is denoted by `weight` (in your notation).  The bias `c` is denoted by `ownBias` (I suggest just `bias`) So, you need to store only two matrices: `weight` and `bias`.

src/mlpack/methods/ann/layer/binary_rbm_impl.hpp
```diff
@@ -0,0 +1,85 @@
...
+namespace ann /** Artificial Neural Network. */ {
+
+template<typename InputDataType, typename OutputDataType>
+VisibleLayer<InputDataType, OutputDataType>::VisibleLayer(
+  const size_t inSize,
+	const size_t outSize):
+  inSize(inSize), 
+  outSize(outSize)
+{
+  weights.set_size(outSize * inSize + inSize, 1);
```
lozhnikov: I guess it is better to allocate memory inside the main RBM class.

src/mlpack/methods/ann/layer/binary_rbm_impl.hpp
```diff
@@ -0,0 +1,85 @@
...
+template<typename InputDataType, typename OutputDataType>
+template<typename eT>
+void VisibleLayer<InputDataType, OutputDataType>::Forward(
+  arma::Mat<eT>&& input, 
+  arma::Mat<eT>&& output)
+{ 
+  if(Parameters().empty())
+    Reset();
+  else
+    LogisticFunction::Fn(weight * input + otherBias, output);
```
lozhnikov: This is just a forward pass through `Linear` and then through `SigmoidLayer`. 
We discussed that it is better to use the existing API.
So this line should look like
```
linear.Forward(std::move(input), std::move(linear.OutputParameters()));
sigmoid.Forward(std::move(linear.OutputParameters()), std::move(output));
```

src/mlpack/methods/ann/layer/binary_rbm_impl.hpp
```diff
@@ -0,0 +1,85 @@
...
+}
+template<typename InputDataType, typename OutputDataType>
+template<typename eT>
+void VisibleLayer<InputDataType, OutputDataType>::Sample(arma::Mat<eT>&& input, arma::Mat<eT>&& output)
+{
+    Forward(std::move(input), std::move(output));
+    for(size_t i = 0; i < output.n_elem; i++)
```
lozhnikov: We also discussed that this is just a forward pass through some layer. I sent an example.
Moreover binomial distribution should look like `math::Random()` functions (see `core/math/random.hpp`).

src/mlpack/methods/ann/vanilla_rbm.hpp
```diff
@@ -0,0 +1,189 @@
...
+#include "activation_functions/softplus_function.hpp"
+#include "init_rules/gaussian_init.hpp"
+
+namespace mlpack {
+namespace ann /** Artificial Neural Network. */ {
+
+template<typename InitializationRuleType,
+         typename VisibleLayerType,
+         typename HiddenLayerType>
+class VanillaRBM
```
lozhnikov: I guess if you pass layer types as template parameters these parameters completely define the type of the RBM. So, for example `RBM<... , BinaryRBMLayer, BinaryRBMLayer>` represents the binary RBM and `RBM<..., SSRBMVisibleLayer, SSRBMHiddenLayer>` represents the spike and slab RBM. In such a way it is better to call the class just `RBM`.

src/mlpack/methods/ann/vanilla_rbm.hpp
```diff
@@ -0,0 +1,189 @@
...
+  /*
+   * Calculates the gradients for the rbm network
+   *
+   * @param input the visible layer/data point
+   * @param neg_input stores the negative samples computed usign the gibbs 
+   * @param k number of steps for gibbs sampling
+   * @param persistence pcdk / not 
+   * @param output store the gradients
+   */
+  void Gradient(arma::mat input,
```
lozhnikov: You can pass `k` and `persistence` at the constructor. If you do so, the method will look like other `Gradient()` methods.

src/mlpack/methods/ann/vanilla_rbm.hpp
```diff
@@ -0,0 +1,189 @@
...
+  // Sigmoid Layer
+  LayerTypes sigmoid;
+  // ResetVisitor
+  ResetVisitor resetVisitor;
+  // DeleteVistor
+  DeleteVisitor deleteVisitor;
+  //! The matrix of data points (predictors).
+  arma::mat predictors;
+  // Network
+  std::vector<LayerTypes> network;
```
lozhnikov: The network consists of only two layers: visible and hidden. So, this variable is not needed.

src/mlpack/tests/ann_layer_test.cpp
```diff
@@ -923,4 +923,38 @@ BOOST_AUTO_TEST_CASE(SimpleLogSoftmaxLayerTest)
   BOOST_REQUIRE_EQUAL(arma::accu(delta), 0);
 }
 
+/**
+ * Simple add module test.
+ */
+BOOST_AUTO_TEST_CASE(SimpleBinaryRbmLayerTest)
```
lozhnikov: It seems you forgot to remove this test.

src/mlpack/methods/ann/layer/binary_rbm.hpp
kris-singh: If we look at the deeplearning.net example. Forward function actually computes the activation form the sigmoid layer and sample actually uses the activations to sample visible/ hidden layer. I think we need both

src/mlpack/methods/ann/layer/binary_rbm.hpp
kris-singh: Okay... so you mean to say that bias(ownBias + otheBIas). because c actually denotes the other bias  always.

src/mlpack/methods/ann/layer/binary_rbm_impl.hpp
kris-singh: The reason for not doing so is that we would have to declare the linear and sigmoid layer which mean we would have to store there parameters and reset them also(overhead). This seems a overhead to me. Also all layer in mlpack are independent that's why i decided to go with this

src/mlpack/methods/ann/vanilla_rbm.hpp
kris-singh: Yes i think we can do that

src/mlpack/methods/ann/layer/binary_rbm.hpp
lozhnikov: I guess `Forward()` can compute both the activation and samples.

src/mlpack/methods/ann/layer/binary_rbm.hpp
lozhnikov: No, `c` denotes `ownBias`. `otherBias` is not needed at all. So, for visible layer `bias` means visible biases and  for hidden layer `bias` means hidden biases.

lozhnikov:
Regarding your question about shared parameters:
I suggest the following:
1. The main RBM class allocates memory:
```
template<.........>
void RBM<VisibleLayerType, HiddenLayerType, .........>::ResetParameters()
{
  size_t weights = visible.Parameters().n_elem;

  parameter.set_size(weights, 1);
  initializeRule.Initialize(parameter, parameter.n_elem, 1);

  visible.Parameters() = arma::mat(parameters.memptr(),
      visible.Parameters().n_rows, visible.Parameters().n_cols, false, false);

  // And the same for the hidden layer
  hidden.Parameters() = arma::mat(parameters.memptr(),
      visible.Parameters().n_rows, visible.Parameters().n_cols, false, false);

  // The Reset() function have to set up links properly
  visible.Reset();
  hidden.Reset();
}
```

2. Implementation of the Reset() method:
The implementation assumes that you have to define the `visible` variable (at the constructor for example).
```
void VisibleLayer<InputDataType, OutputDataType>::Reset()
{
  if (visible)
  {
    // The number of input (visible) variables is equal to inSize
    // The number of output (hidden) variables is equal to outSize
    weight = arma::mat(weights.memptr(), outSize, inSize, false, false);

    // The offset is equal to the number of elements of the weight matrix
    bias = arma::mat(weights.memptr() + weight.n_elem, outSize, 1, false, false);
  }
  else
  {
    // The number of input (hidden) variables is equal to inSize
    // The number of output (visible) variables is equal to outSize
    weight = arma::mat(weights.memptr(), inSize, outSize, false, false);

    // The offset is equal to the number of elements of the weight matrix and the size of visible biases
    bias = arma::mat(weights.memptr() + weight.n_elem + inSize, outSize, 1, false, false);
  }
}
```

So, the `weights` matrix contains all parameters. The weight matrix `W` is denoted by `weight`. And each layer has only its own biases.

src/mlpack/methods/ann/layer/binary_rbm_impl.hpp
lozhnikov: I mean you should add a function to `core/math/random.hpp`:) for example `RandBinomial`

src/mlpack/methods/ann/layer/binary_rbm.hpp
lozhnikov: Hmm.. I looked again through the article, the forward pass of the visible layer needn't hidden biases.  Hidden variables are sampled from `P(h=1|v)=Sigm(W^t * v + c)` where `W` denotes the shared weight matrix and `c` denotes visible biases.

kris-singh: I think you can look at the equation 7 c: reprsents the hidden biases http://deeplearning.net/tutorial/rbm.html#equation-rbm_propup

lozhnikov: oh, you are right, nevertheless you can use the technique that I suggested below.

kris-singh:
Just a few minor things I think
`visible.Parameters() = arma::mat(parameters.memptr(), 
  visible.Parameters().n_elem, 1, false, false);`
Since parameters is a 1d vector
Also here
`// The offset is equal to the number of elements of the weight matrix
    bias = arma::mat(weights.memptr() + weight.n_elem, inSize, 1, false, false);`

src/mlpack/methods/ann/layer/binary_rbm.hpp
kris-singh: In your solution we have only one bias for the binaryRBMLayer. So the problem is that since the forward function uses the bias of the other layer(hidden layer uses the bias of the visible layer) how can we code that up. Are you suggesting we move the forward function to the wrapper class.

src/mlpack/methods/ann/layer/binary_rbm.hpp
kris-singh: Well in many cases you need the preactivation outputs. I think we need both the funcitons

src/mlpack/methods/ann/layer/binary_rbm.hpp
lozhnikov: yeah, of course

src/mlpack/core/math/random.hpp
```diff
@@ -66,6 +66,31 @@ inline double Random(const double lo, const double hi)
 }
 
 /**
+ * Generates a binomial random number in the specified range.
+ */
+inline double BinomialRandom(const int input)
```
lozhnikov: I guess the comment is invalid since `input` is equal to the success probability (and the mean). Moreover `input` is a real variable and belongs to `[0, 1]`.
Actually this is the Bernoulli distribution since the binomial distribution depends on the number of trials. And the name of the function should be similar to others (`RandNormal`, `RandInt`, etc).
So, I suggest `RandBernoulli`.

src/mlpack/core/math/random.hpp
```diff
@@ -66,6 +66,31 @@ inline double Random(const double lo, const double hi)
...
+  if(Random() > input)
+    return 1;
+  else
+    return 0;
+}
+
+/**
+ * Generates a binomial random number in the specified range.
+ */
+template<typename InputDatatype, typename OutputDatatype>
```
lozhnikov: I think this function is not needed (e.g. since `arma::mat::size()` and others return the dimensions).

src/mlpack/core/optimizers/cdk/cdk_impl.hpp
```diff
@@ -0,0 +1,94 @@
...
+  arma::mat gradient(iterate.n_rows, iterate.n_cols);
+  for (size_t i = 1; i != maxIterations; ++i, ++currentFunction)
+  {
+    // Is this iteration the start of a sequence?
+    if ((currentFunction % numFunctions) == 0)
+    {
+      if (shuffle) // Determine order of visitation.
+        visitationOrder = arma::shuffle(visitationOrder);
+    }
+
```
lozhnikov: Well, right now the implementation of the CDk optimizer is similar to the SGD optimizer. I guess there is no need to duplicate code.

src/mlpack/methods/ann/layer/binary_rbm_impl.hpp
```diff
@@ -19,61 +19,63 @@ namespace mlpack {
...
 template<typename InputDataType, typename OutputDataType>
-void VisibleLayer<InputDataType, OutputDataType>::Reset()
+void BinaryLayer<InputDataType, OutputDataType>::Reset()
 {
-  
-  weight = arma::mat(weights.memptr(), outSize, inSize, false, false);
-  ownBias = arma::mat(weights.memptr() + weight.n_elem, inSize, false, false);
-  otherBias = arma::mat(weight.memptr() + weight.n_elem + inSize, outSize, false, false);
-
+  if(typeVisible)
```
lozhnikov: I think there are some errors since `visible.inSize = hidden.outSize` and `visible.outSize = hidden.inSize`.

src/mlpack/methods/ann/vanilla_rbm_impl.hpp
```diff
@@ -76,45 +74,65 @@ void VanillaRBM<InitializationRuleType,
...
+}
+
+template<typename InitializationRuleType,
+  typename VisibleLayerType,
+  typename HiddenLayerType>
+double RBM<InitializationRuleType, 
+  VisibleLayerType, 
+  HiddenLayerType>::Evaluate(const arma::mat& /* parameters*/, const size_t i)
+{
+  arma::mat freeEnergy, freeEnergyCorrupted, output;
```
lozhnikov: Could you send me the link to the source?
P.S. `FreeEnergy()` returns `double`.

src/mlpack/core/optimizers/cdk/cdk_impl.hpp
kris-singh: Hmmm, i agree for the most part. But the problem remains that sgd function utilities a evaluate function as stopping criteria. Maybe we can set the tolerance to very low so that the difference evaluate function of the previous and current function is never less than the tolerance level but we cannot attach any guarantees with that. Also i will look at the ssRBM and let you know if we nedd the cdk optimizer or not.

src/mlpack/methods/ann/layer/binary_rbm_impl.hpp
kris-singh: Do you mean to say here that when we are initializing the 1.hidden layer(outsize, insize, 0).
According to that i think you might be correct but i wrote it with the idea that 2.hidden layer(insize, outsize, 0). Should i change this to keeping the 1 in mind.

src/mlpack/methods/ann/vanilla_rbm_impl.hpp
kris-singh: Ohhh sorry. I would change this. So there are 2 sources i am following
1. https://github.com/scikit-learn/scikit-learn/blob/14031f6/sklearn/neural_network/rbm.py#L28
2. http://deeplearning.net/tutorial/code/rbm.py you can have look at the pseudo likelihood cost function

src/mlpack/core/optimizers/cdk/cdk_impl.hpp
lozhnikov: If I remember right the ssRBM paper don't mention the stopping criteria. You can use the SGD optimizer with `tolerance < 0`.

src/mlpack/methods/ann/layer/binary_rbm_impl.hpp
lozhnikov: I think `inSize` should denote the number of input units. I guess if the meaning of `inSize` depends on the type of the level it is confusing.

src/mlpack/methods/ann/layer/binary_rbm_impl.hpp
```diff
@@ -0,0 +1,81 @@
...
+}
+
+template<typename InputDataType, typename OutputDataType>
+void BinaryLayer<InputDataType, OutputDataType>::Forward(
+  const InputDataType&& input, 
+  OutputDataType&& output)
+{ 
+  if(Parameters().empty())
+    Reset();
+  else
```
lozhnikov: I guess you forgot braces.

src/mlpack/methods/ann/vanilla_rbm.hpp
```diff
@@ -0,0 +1,208 @@
...
+  };
+
+  /*
+   * Helper function for Gradient
+   * calculates the gradients for both
+   * positive and negative samples.
+   */
+  void CalcGradient(arma::mat&& input, arma::mat&& output)
+  {
+    arma::mat temp1, temp2;
```
lozhnikov: Could you reduce memory allocations here? As for me it is better to define these temporary variables as class members.

src/mlpack/methods/ann/vanilla_rbm_impl.hpp
```diff
@@ -0,0 +1,201 @@
...
+  arma::mat temp, temp1;
+  temp = input;
+
+  if(persistence && !state.is_empty())
+    input = state;
+  
+  for(size_t j = 0; j < num_steps - 1; j++)
+  {
+    // Use probabilties for updation till the last step(section 3 hinton)
+    ForwardVisible(std::move(temp), std::move(temp1));
```
lozhnikov: I think you forgot to sample variables after each Forward() call.

src/mlpack/methods/ann/layer/binary_rbm_impl.hpp
kris-singh: I did not understand your comment your saying that when we declare a new layer we do not allocate memory to it. Other layer allocate memory when they are created. Also we cannot this in the vanilla_rbm as we do not have acess to the insize and outsize there.

src/mlpack/tests/ann_layer_test.cpp
kris-singh: I think we need to test the rbm_binary layer. So there should be this test.

src/mlpack/methods/ann/vanilla_rbm_impl.hpp
kris-singh: I think you can read about it here. https://www.cs.toronto.edu/~hinton/absps/guideTR.pdf.
But earlier i was doing just sampleVisible and sampleHidden for numSteps then also the result was not converging

kris-singh:
How can i fix the appveyor check. I want to add a new test to ann_layer_test.cpp. Also all the style checks on my files pass i don't understand why the test is failing.

zoq:
If you rebase the PR against the latest master branch the style checks should pass and fix the appveyor and travis build issue. For more information: https://github.com/edx/edx-platform/wiki/How-to-Rebase-a-Pull-Request

src/mlpack/core/math/random.hpp
zoq: I agree, you can just use ``for_each``or ``imbue`` to do the same without introducing another function.

src/mlpack/core/optimizers/cdk/cdk_impl.hpp
```diff
@@ -0,0 +1,98 @@
...
+    if (shuffle)
+      rbm.Gradient(visitationOrder[currentFunction], k, persistent, gradient);
+    else
+      rbm.Gradient(currentFunction, k, persistent,  gradient);
+
+    if (i % batchSize == 0)
+    {
+      std::cout << rbm.Evaluate(iterate, orderFunction) << std::endl;
+      // Update the step
+      iterate += stepSize * cumgradient;
```
zoq: I haven't checked the Gradient function yet, but should we subtract or add here? Also, should we divide ``cumgradient`` by ``batchSize``?

src/mlpack/tests/rbm_network_test.cpp
```diff
@@ -0,0 +1,77 @@
...
+   */
+  arma::mat output;
+  BinaryLayer<> visible(trainData.n_rows, hiddenLayerSize, 1);
+  BinaryLayer<> hidden(hiddenLayerSize, trainData.n_rows, 0);
+  GaussianInitialization gaussian(0, 0.1);
+  RBM<GaussianInitialization, BinaryLayer<>, BinaryLayer<> > model(trainData,
+      gaussian, visible, hidden);
+  CDK<RBM<GaussianInitialization, BinaryLayer<>, BinaryLayer<> >> cdk(model,
+      100, 0.1, 15 * trainData.n_cols, 20, true, true);
+  model.Train(trainData, cdk);
```
zoq: Since you used ``std::move()`` inside the RBM constructor you can't reuse ``trainData`` here. You could remove ``std::move`` from the RBM constructor or pass a new copy to the Train function.

src/mlpack/methods/ann/vanilla_rbm_impl.hpp
```diff
@@ -0,0 +1,201 @@
...
+  arma::Col<double> output(orderFunction.n_elem);
+  size_t corruptIdx = math::RandInt(0, visible.Bias().n_elem);
+  for (auto i : orderFunction)
+  {
+    // Do not use unsafe col predictor as we change the input
+    arma::vec currentInput = predictors.col(i);
+    freeEnergy = FreeEnergy(std::move(currentInput));
+    currentInput(corruptIdx) =  1 - predictors(corruptIdx);
+    freeEnergyCorrupted = FreeEnergy(std::move(currentInput));
+    output(i) = SoftplusFunction::Fn(freeEnergyCorrupted - freeEnergy + 1);
```
zoq: ``orderFunction`` holds the visitation order of the current batch and is of size batchSize right? So let's say ``orderFunction`` is ``[120, 121, 122, ...]`` I guess I run into an issue if I do ``output(i)``with ``i = 120`` since ``output`` has the same size as ``orderFunction`` which is ``batchSize``. Maybe I missed something?

zoq: So I guess, you meant to use the index of the current iteration instead of using the value?

zoq: Also, we should avoid ``auto`` here, I don't think it makes it easier for someone trying to understand the code if you have to lookup the type.

src/mlpack/core/math/random.hpp
kris-singh: Okay i get it. I will remove this now.

src/mlpack/core/optimizers/cdk/cdk_impl.hpp
kris-singh: I did subtract and divide the cumgradient by batchsize but it didn't help with the training. Thought i agree with the scaling by batch size if you look at the update for the RBM they the gradients are added to the parameters.

src/mlpack/methods/ann/vanilla_rbm_impl.hpp
kris-singh: Yes i meant to use the iteration index here for the output. But funnily the code actually does not give error on this which is strange.

src/mlpack/tests/rbm_network_test.cpp
kris-singh: I have remove the reference in the model constructor now.

src/mlpack/methods/ann/vanilla_rbm_impl.hpp
zoq: What test case do you use? It will not fail if ``numFunctions = batchSize`` which I think is the case for the ``SGDFunction``.

kris-singh:
The training error / evaluate function now only increases rather than going up and down.
The samples are also always one.

src/mlpack/core/optimizers/cdk/cdk_impl.hpp
```diff
@@ -0,0 +1,97 @@
...
+  if (shuffle)
+  {
+    visitationOrder = arma::shuffle(arma::linspace<arma::Col<size_t>>(0,
+        (numFunctions - 1), numFunctions));
+  }
+
+  // Now iterate!
+  arma::mat gradient(iterate.n_rows, iterate.n_cols);
+  arma::mat cumgradient(iterate.n_rows, iterate.n_cols);
+
```
lozhnikov: You forgot to add here
```
cumgradient.zeros();
```
Or just add to the constructor `arma::fill::zeros`.

src/mlpack/core/optimizers/cdk/cdk_impl.hpp
```diff
@@ -0,0 +1,97 @@
...
+    // Evaluate the gradient for this iteration.
+    if (shuffle)
+      rbm.Gradient(visitationOrder[currentFunction], gradient);
+    else
+      rbm.Gradient(currentFunction, gradient);
+
+    if (i % batchSize == 0)
+    {
+      std::cout << overallCost << std::endl;
```
lozhnikov: There is no need to print `overallCost` each time. I think it is reasonable to print that after each training epoch (in order to debug of course).
I suggest the following:
```
    if (shuffle)
      rbm.Gradient(visitationOrder[currentFunction], gradient);
    else
      rbm.Gradient(currentFunction, gradient);

    cumgradient += gradient;

    if (shuffle)
      overallCost +=rbm.Evaluate(iterate, visitationOrder[currentFunction]);
    else
      overallCost +=rbm.Evaluate(iterate, currentFunction);

    if (i % batchSize == 0)
    {
      // Update the step
      iterate -= stepSize * (cumgradient / batchSize);
      cumgradient.zeros();
    }

    if (i % numFunctions == 0)
    {
      overallCost /= numFunctions;
      std::cout << overallCost << std::endl;
      overallCost = 0;
    }

```

src/mlpack/methods/ann/vanilla_rbm_impl.hpp
```diff
@@ -0,0 +1,201 @@
...
+  // Collect the positive gradients
+  CalcGradient(std::move(predictors.col(input)), std::move(positiveGradient));
+
+  // Collect the negative samples
+  Gibbs(std::move(predictors.col(input)), std::move(negativeSamples));
+
+  // Collect the negative gradients
+  CalcGradient(std::move(negativeSamples), std::move(negativeGradient));
+
+  output = positiveGradient - negativeGradient;
```
lozhnikov: This is the main issue: the sign is invalid.
So write the following
```
output = -(positiveGradient - negativeGradient);
```
Or change the signs of `positiveGradient` and `negativeGradient`.
If you calculate the derivatives of the energy function you will see that they has got the minus sign in the beginning.

src/mlpack/tests/rbm_network_test.cpp
```diff
@@ -0,0 +1,79 @@
...
+  std::cout << output << std::endl;
+}
+
+/**
+ * Train the vanilla network on a larger dataset.
+ */
+BOOST_AUTO_TEST_CASE(VanillaNetworkTest)
+{
+  arma::mat dataset;
+  dataset.load("/Users/kris/Desktop/GsoC/mlpack/src/mlpack/tests/data/mnist_first250_training_4s_and_9s.arm");
```
lozhnikov: Of course here should be just `dataset.load("mnist_first250_training_4s_and_9s.arm");`

lozhnikov:
If I forget to comment something, here is the patch that seems works. Don't commit this patch (there are 
a lot of debug prints and so on).
[patch.zip](https://github.com/mlpack/mlpack/files/1091436/patch.zip)

src/mlpack/methods/ann/vanilla_rbm_impl.hpp
```diff
@@ -0,0 +1,201 @@
...
+  * the free energy of the model
+  */
+template<typename InitializationRuleType, typename VisibleLayerType,
+    typename HiddenLayerType>
+double RBM<InitializationRuleType, VisibleLayerType, HiddenLayerType>::
+    FreeEnergy(arma::mat&& input)
+{
+  arma::mat output;
+  visible.Forward(std::move(input), std::move(output));
+  SoftplusFunction::Fn(output, output);
```
zoq: I'm not sure I completely get this point, we do ``softplus(sigmoid(x))`` but shouldn't it be ``log(sigmoid(x))`` or, I guess you can directly use ``SoftplusFunction`` inside the ``BinaryLayer``?

src/mlpack/methods/ann/vanilla_rbm.hpp
```diff
@@ -0,0 +1,207 @@
...
+   * Helper function for Gradient
+   * calculates the gradients for both
+   * positive and negative samples.
+   */
+  void CalcGradient(arma::mat&& input, arma::mat&& output)
+  {
+    arma::mat temp1;
+    ForwardVisible(std::move(input), std::move(temp1));
+    output = temp1 * input.t();
+    output.reshape(output.n_rows * output.n_cols, 1);
```
lozhnikov: As for me, this code is unsafe since it assumes that armadilo stores matrices in the column major order.
Could you rewrite that using `arma::mat(something.memptr(), ....)`?

src/mlpack/methods/ann/vanilla_rbm.hpp
```diff
@@ -0,0 +1,207 @@
...
+  * 
+  */
+  double Evaluate(const arma::mat& parameters, const size_t i);
+
+ /** 
+  * This function calculates
+  * the free energy of the model
+  * @param: input data point 
+  */
+  double FreeEnergy(arma::mat&& input);
```
lozhnikov: Why do you use rvalue references here? As for me the input should be`const arma::mat&` or even a template since you calculate free energy of `arma::vec`.

src/mlpack/methods/ann/vanilla_rbm_impl.hpp
```diff
@@ -0,0 +1,201 @@
...
+  return  -arma::accu(output) -arma::dot(input, visible.Bias());
+}
+
+template<typename InitializationRuleType, typename VisibleLayerType,
+    typename HiddenLayerType>
+double RBM<InitializationRuleType, VisibleLayerType, HiddenLayerType>::
+    Evaluate(const arma::mat& /* parameters*/, const size_t i)
+{
+  
+    double freeEnergy, freeEnergyChain;
```
lozhnikov: Could you fix spaces here?

src/mlpack/methods/ann/vanilla_rbm.hpp
```diff
@@ -0,0 +1,224 @@
...
+   * Helper function for Gradient
+   * calculates the gradients for both
+   * positive and negative samples.
+   */
+  void CalcGradient(arma::mat&& input, arma::mat&& output)
+  {
+    ForwardVisible(std::move(input), std::move(inputForward));
+    output = inputForward * input.t();
+    // Weights, hidden bias, visible bias
+    output = arma::join_cols(arma::join_cols(arma::vectorise(output),
```
lozhnikov: Are you sure that `arma::join_cols` doesn't allocate new memory?

src/mlpack/methods/ann/vanilla_rbm.hpp
```diff
@@ -0,0 +1,224 @@
...
+  // Visible layer
+  VisibleLayerType visible;
+  // Hidden Layer
+  HiddenLayerType hidden;
+  // Sigmoid Layer
+  LayerTypes sigmoid;
+  // ResetVisitor
+  ResetVisitor resetVisitor;
+  // DeleteVistor
+  DeleteVisitor deleteVisitor;
```
lozhnikov: I am not sure that the variables above are used.

src/mlpack/methods/ann/vanilla_rbm_impl.hpp
```diff
@@ -0,0 +1,232 @@
...
+template<typename InitializationRuleType, typename VisibleLayerType,
+    typename HiddenLayerType>
+void RBM<InitializationRuleType, VisibleLayerType, HiddenLayerType>::Reset()
+{
+  // Since we are dealing with a static stucture no need for visitor
+  // Reset the parameters of all the layers
+  size_t weight = 0;
+  weight+= visible.Parameters().n_elem;
+  parameter.set_size(weight, 1);
+  parameter.zeros();
```
lozhnikov: I think this line is not needed since you invoke `InitializationRule`.

src/mlpack/methods/ann/vanilla_rbm_impl.hpp
```diff
@@ -0,0 +1,232 @@
...
+}
+
+ /* 
+  * This function calculates
+  * the free energy of the model
+  */
+template<typename InitializationRuleType, typename VisibleLayerType,
+    typename HiddenLayerType>
+double RBM<InitializationRuleType, VisibleLayerType, HiddenLayerType>::
+    FreeEnergy(arma::mat input)
```
lozhnikov: Why did you remove the reference? You needn't to copy the matrix since the function doesn't change that.

src/mlpack/methods/ann/vanilla_rbm_impl.hpp
```diff
@@ -0,0 +1,232 @@
...
+template<typename InitializationRuleType, typename VisibleLayerType,
+    typename HiddenLayerType>
+double RBM<InitializationRuleType, VisibleLayerType, HiddenLayerType>::
+    Evaluate(const arma::mat& /* parameters*/, const size_t i)
+{
+  double freeEnergy, freeEnergyChain;
+  arma::mat negativeSample;
+  arma::vec currentInput = predictors.unsafe_col(i);
+  // Do not use unsafe col predictor as we change the input
+  freeEnergy = FreeEnergy(currentInput);
```
lozhnikov: Moreover if you apply `FreeEnergy` to `arma::vec` you should add a template parameter in order to prevent copying.

For example:
```
template<typename VectorType>
FreeEnergy(VectorType& input)
```

src/mlpack/methods/ann/vanilla_rbm_impl.hpp
```diff
@@ -0,0 +1,232 @@
...
+    input = state;
+
+  for (size_t j = 0; j < steps; j++)
+  {
+    // Use probabilties for updation till the last step(section 3 hinton)
+    ForwardVisible(std::move(input), std::move(output));
+    ForwardHidden(std::move(output), std::move(input));
+  }
+  // Set state to binarised output
+  SampleVisible(std::move(output), std::move(state));
```
lozhnikov: Don't forget that `SampleVisible` does an extra forward pass that is not needed here.

src/mlpack/tests/rbm_network_test.cpp
```diff
@@ -46,34 +49,67 @@ void BuildVanillaNetwork(MatType& trainData,
...
+   L_BFGS<SoftmaxRegressionFunction> optimizer(srf, numBasis, numIterations);
+   SoftmaxRegression<L_BFGS> regressor2(optimizer);
+ 
+   arma::Row<size_t> predictions1, predictions2; // Vectors to store predictions in.
+
+   double classificationAccuray = regressor2.ComputeAccuracy(Y, Ylabel);
+   std::cout << "classificationAccuray" << classificationAccuray << std::endl;
+
+
+   SoftmaxRegressionFunction srf1(XRbm, Xlabel, numClasses, 0.001);
```
lozhnikov: Why do you use hidden samples for training?

src/mlpack/methods/ann/vanilla_rbm.hpp
```diff
@@ -0,0 +1,224 @@
...
+  * 
+  */
+  double MonitoringCost(const size_t i);
+
+ /** 
+  * This function calculates
+  * the free energy of the model
+  * @param: input data point 
+  */
+  double FreeEnergy(arma::mat input);
```
lozhnikov: You should avoid copy-constructors.

src/mlpack/methods/ann/vanilla_rbm_impl.hpp
```diff
@@ -0,0 +1,232 @@
...
+{
+  double freeEnergy, freeEnergyChain;
+  arma::mat negativeSample;
+  arma::vec currentInput = predictors.unsafe_col(i);
+  // Do not use unsafe col predictor as we change the input
+  freeEnergy = FreeEnergy(currentInput);
+  Gibbs(currentInput, std::move(negativeSample));
+  freeEnergyChain = FreeEnergy(negativeSample);
+
+  return arma::mean(freeEnergy) - arma::mean(freeEnergyChain);
```
lozhnikov: `arma::mean` is not needed here.

src/mlpack/methods/ann/vanilla_rbm_impl.hpp
```diff
@@ -0,0 +1,232 @@
...
+    corruptInput.row(idx) = 1 - corruptInput.row(idx);
+    // Use mean when showing the if more than one input
+    return std::log(LogisticFunction::Fn(FreeEnergy(corruptInput) - 
+        FreeEnergy(predictors.col(i)))) * predictors.n_rows;
+  }
+  else
+  {
+    visible.Forward(std::move(predictors.col(i)), std::move(output));
+    // Use mean when showing the if more than one input
+    return arma::mean(arma::accu( predictors.col(i) * arma::log(output).t() +
```
lozhnikov: `arma::mean` is not needed here.

src/mlpack/methods/ann/vanilla_rbm.hpp
```diff
@@ -0,0 +1,224 @@
...
+  void SampleVisible(arma::mat&& input, arma::mat&& output);
+
+ /*
+  * This function does the k-step
+  * gibbs sampling.
+  *
+  * @param input: input to the gibbs function
+  * @param output: stores the negative sample
+  */
+  void Gibbs(arma::mat input, arma::mat&& output, size_t steps = 1);
```
kris-singh: Here we need this since input is changed if you lookt at the implementation of gibbs.

src/mlpack/methods/ann/vanilla_rbm_impl.hpp
kris-singh: Are you getting better results with this strategy i am not getting better results with using probabilities?

src/mlpack/core/optimizers/cdk/cdk_impl.hpp
```diff
@@ -0,0 +1,112 @@
...
+      rbm.Gradient(visitationOrder[currentFunction], gradient);
+    else
+      rbm.Gradient(currentFunction, gradient);
+
+    cumgradient += gradient;
+
+    if (shuffle)
+    {
+      overallCost += rbm.Evaluate(iterate, visitationOrder[currentFunction]);
+      overallmCost += rbm.MonitoringCost(visitationOrder[currentFunction]);
```
kris-singh: I don't understand what you mean by the template parameter here. Can you give the code snippet. That would mean that i create 2 classes one for MonitioringClass and other for Evaluate. and then template the rbm class in the evaluate function inside the rbm class call the templated evaluate function. Is that correct ?

src/mlpack/tests/rbm_network_test.cpp
kris-singh: I implemented the strategy we were using that is using the transformed input matrix for training and testing of the softmax regression in sklearn. And yes you are correct the get the get the classification  accuracy around 51% in that case but our implementation is still under perfoming we are getting around 10 %

src/mlpack/core/optimizers/cdk/cdk_impl.hpp
lozhnikov: For example
```
template<....., typename EvaluationMethod>
double RBM<......., EvaluationMethod>::Evaluate(.......)
{
  return EvaluationMethod::Evaluate(......);
}
```
Or without template:
```
RBM<......>::Evaluate(......)
{
   if (evaluateMonitoringCost)
   {
// evaluate monitoring cost
   }
   else
   {
// evaluate loss
   }
}

File: src/mlpack/methods/ann/vanilla_rbm.hpp (hunk 39 above)
Comment (lozhnikov) [inline src/mlpack/methods/ann/vanilla_rbm.hpp]: I mean that the memory should be preallocated.

**Comment by lozhnikov** [comment]
I wrote above that there are a lot of memory allocations inside your code. I think you should avoid copy-constructors, temporary variables (I mean complex structures like `arma::mat`, `arma::vec` etc).
I tried to rewrite the `CalcGradient()` function and I got considerable speedup (1.7 seconds for an epoch against 4.5 seconds). Here is the patch.
[speedup-patch.zip](https://github.com/mlpack/mlpack/files/1099828/speedup-patch.zip)
(Don't merge as is, I didn't take care of the coding style).

File: src/mlpack/methods/ann/vanilla_rbm_impl.hpp (hunk 34 above)
Comment (lozhnikov) [inline src/mlpack/methods/ann/vanilla_rbm_impl.hpp]: No, I didn't. I tried to speedup the code since it's too slow (I wrote a comment about that below).

File: src/mlpack/methods/ann/vanilla_rbm_impl.hpp [hunk 41]
```diff
@@ -0,0 +1,232 @@
...
+    Gibbs(arma::mat input, arma::mat&& output, size_t steps)
+{
+  steps = (steps == 1)?this-> numSteps: steps;
+  if (persistence && !state.is_empty())
+    input = state;
+
+  for (size_t j = 0; j < steps; j++)
+  {
+    // Use probabilties for updation till the last step(section 3 hinton)
+    ForwardVisible(std::move(input), std::move(output));
```
lozhnikov: It seems I overlooked an issue yesterday. I looked again through Hinton's paper. He wrote: 

> It is very important to make these hidden states binary, rather than using the probabilities themselves. ...
> ...For the last update of the hidden units, it is silly to use stochastic binary states because nothing depends on which state is chosen. 

So, according to the paper you should replace `ForwardVisible()` by `SampleHidden()` for all steps except the last one.

kris-singh:
[save.patch.zip](https://github.com/mlpack/mlpack/files/1100137/save.patch.zip)
I created another version which follows the deeplearning.net implementation more closely let me know what you think

src/mlpack/methods/ann/rbm_impl.hpp
```diff
@@ -0,0 +1,266 @@
...
+    // Do not use unsafe col predictor as we change the input
+    Gibbs(i, std::move(tempInput));
+    return FreeEnergy(currentInput) - FreeEnergy(tempInput);
+  }
+  else
+  {
+    if(persistence)
+    {
+      size_t idx = RandInt(0, currentInput.n_rows);
+      tempInput = currentInput;
```
lozhnikov: I just realized that you forgot to round the `currentInput` vector.

src/mlpack/methods/ann/rbm_impl.hpp
```diff
@@ -0,0 +1,266 @@
...
+}
+
+template<typename InitializationRuleType, typename VisibleLayerType,
+    typename HiddenLayerType>
+void RBM<InitializationRuleType, VisibleLayerType, HiddenLayerType>::
+    Gibbs(const size_t i, arma::mat&& output, size_t steps)
+{
+  // Do not use unsafe col
+  gibbsInput = predictors.col(i);
+  steps = (steps == 1)?this-> numSteps: steps;
```
lozhnikov: I guess it is much better to use `step = SIZE_MAX` instead of 1.

src/mlpack/methods/ann/rbm_impl.hpp
```diff
@@ -0,0 +1,266 @@
...
+  else
+  {
+    for (size_t j = 0; j < steps - 1 ; j++)
+    {
+      // Use probabilties for updation till the last step(section 3 hinton)
+      SampleHidden(std::move(gibbsInput), std::move(output));
+      SampleVisible(std::move(output), std::move(gibbsInput));
+    }
+    // Set state to binarised output
+    this->VisibleLayer().Forward(std::move(gibbsInput), std::move(state));
```
lozhnikov: I guess `state` is a binary variable.

src/mlpack/methods/ann/rbm_impl.hpp
```diff
@@ -0,0 +1,266 @@
...
+    gibbsInput = state;
+
+  if (steps == 1)
+  {
+    this->VisibleLayer().Sample(std::move(gibbsInput), std::move(output));
+    this->HiddenLayer().Sample(std::move(output), std::move(output));
+  }
+  else
+  {
+    for (size_t j = 0; j < steps - 1 ; j++)
```
lozhnikov: Why do you perform `state-1` steps instead of `steps`?

src/mlpack/methods/ann/rbm_impl.hpp
lozhnikov: Try the following implementation:
```
template<typename InitializationRuleType, typename VisibleLayerType,
    typename HiddenLayerType>
void RBM<InitializationRuleType, VisibleLayerType, HiddenLayerType>::
    Gibbs(arma::mat&& input, arma::mat&& output, size_t steps)
{
  steps = (steps == SIZE_MAX) ? this-> numSteps : steps;

  if (persistence && !state.is_empty())
  {
    SampleHidden(std::move(state), std::move(gibbsTemporary));
    SampleVisible(std::move(gibbsTemporary), std::move(output));

//  According to Hinton's paper we can use probabilities instead of samples i.e.
//  hidden.Forward(std::move(gibbsTemporary), std::move(output));
  }
  else
  {
    SampleHidden(std::move(input), std::move(gibbsTemporary));
    SampleVisible(std::move(gibbsTemporary), std::move(output));

//  According to Hinton's paper we can use probabilities instead of samples i.e.
//  hidden.Forward(std::move(gibbsTemporary), std::move(output));
  }

  for (size_t j = 1; j < steps; j++)
  {
    SampleHidden(std::move(output), std::move(gibbsTemporary));
    SampleVisible(std::move(gibbsTemporary), std::move(output));

//  According to Hinton's paper we can use probabilities instead of samples i.e.
//  hidden.Forward(std::move(gibbsTemporary), std::move(output));
  }

  if (persistence)
    state = output;
}
```

lozhnikov:
Regarding your version which follows the deeplearning.net implementation:
I don't think that we need two almost equal variants of the `Gibbs()` function.

src/mlpack/methods/ann/rbm_impl.hpp
kris-singh: Okay if you look closely at the deep learning.net example the gibbshvh to do the negative pass where you do this SampleVisible ---> SampleHidden(Negative Hidden) ---> SampleVisible(Negative Visible)x'
Also state = NegativeHidden and output = Negative Visible.
The same strategy cannot be applied for Sampling for sampling they use 
gibbsvhv where they do something like 
SampleHidden--> VisibleForwardPass --> SampleVisible(
here output = Visible.ForwardPass(input) and state = SampleVisible.
That was the mistake we were making.

kris-singh: Also state is not a binary variable it holds the persistence state which is basically vector which we use to start the gibbs chain. That is why is was doing step-1

lozhnikov:
I think I fixed the test.The classification accuracy is equal to ~84% on the dataset from the scikit example.

Here is the code (Don't merge as is, I didn't take care of the coding style and so on).
The results are obtained with the version I sent you via IRC.
[test-fix.zip](https://github.com/mlpack/mlpack/files/1100411/test-fix.zip)

Just in case here is that old version of the `RBM` class that I use:
[old_rbm_files.zip](https://github.com/mlpack/mlpack/files/1100414/old_rbm_files.zip)

kris-singh:
I am not able replicate your results. I added all your changes to my files i am just using currentinput instead of predictor.col(i) i don't think that would have caused the problem are we using the same files for training and testing?

kris-singh:
I am not able replicate your results. I added all your changes to my files i am just using currentinput instead of predictor.col(i) i don't think that would have caused the problem are we using the same files for training and testing?  I am getting around 21% accuracy

lozhnikov:
No, I didn't use the MNIST dataset in order to calculate the computation accuracy. I used the dataset from the scikit example.

kris-singh:
Strange i am also using the same scikit dataset. I have pushed the data to the commit.

kris-singh:
Added the test for RBM network. I wanted to add test for gibbs and the evaluation function. I was looking at the sklearn implementation but it heavily depends on the random state i tried to repeat calculations to get the rid of the randomness but they are no conclusive results that i got. Let me know if we need to change something or add something.

lozhnikov:
Regarding the datasets:
I think you should try existing datasets first. If the test doesn't show good results, try some small parts of the dataset from the scikit example. If the minimum part required for test weights too much, upload that in the binary format.

src/mlpack/tests/rbm_network_test.cpp
```diff
@@ -0,0 +1,210 @@
...
+      mainOutput.submat(s * 28, c * 28, s * 28 + 27, c * 28 + 27) = copy;
+    }
+  }
+  BOOST_REQUIRE_LE(result / (numChains*numSamples), 700);
+}
+
+BOOST_AUTO_TEST_CASE(ClassificationTest)
+{
+  // Normalised dataset
+  arma::mat X, Y;
```
lozhnikov: Could you point out which variable corresponds to the train dataset and which variable corresponds to the test dataset? I suggest `trainData`, `testData`, `trainLabels`, `testLabels` for example.

src/mlpack/tests/rbm_network_test.cpp
```diff
@@ -0,0 +1,210 @@
...
+  arma::Row<size_t> Xlabel, Ylabel;
+  data::Load("mnist_regression.txt", X, true);
+  data::Load("mnist_regression_label.txt", Xlabel, true);
+  data::Load("mnist_regression_test.txt", Y, true);
+  data::Load("mnist_regression_test_label.txt", Ylabel, true);
```
lozhnikov: Hmm.. I guess these filenames are incorrect. Did you try other datasets? For example some datasets in `src/tests/data`.

src/mlpack/tests/rbm_network_test.cpp
```diff
@@ -0,0 +1,210 @@
...
+  BinaryLayer<> visible(trainData.n_rows, hiddenLayerSize, 1);
+  BinaryLayer<> hidden(hiddenLayerSize, trainData.n_rows, 0);
+  GaussianInitialization gaussian(0, 0.1);
+  RBM<GaussianInitialization, BinaryLayer<>, BinaryLayer<> > model(trainData,
+      gaussian, visible, hidden, 1,  true);
+  CDK<RBM<GaussianInitialization, BinaryLayer<>, BinaryLayer<> >> cdk(model,
+      0.06, trainData.n_cols * 20, 10, true);
+  model.Reset();
+  // Set the parmaeters from a learned rbm
+  model.Parameters() = {-0.23224054, -0.23000632,
```
kris-singh: I got these values after training the sklearn. And i actually used there parameters to intialise out model and test the freeEnergy function.

src/mlpack/tests/rbm_network_test.cpp
```diff
@@ -0,0 +1,210 @@
...
+  BOOST_REQUIRE_LE(arma::accu(calcultedFreeEnergy - freeEnergy), 1);
+  
+}
+BOOST_AUTO_TEST_CASE(MiscTest)
+{
+  /**
+   * Train and evaluate a vanilla network with the specified structure.
+   */
+
+  arma::mat X;
```
kris-singh: I will change the seed. The test actually does not depend upon the seed value if we intialize the parameter value.

src/mlpack/tests/rbm_network_test.cpp
```diff
@@ -0,0 +1,210 @@
...
+      }
+      model.HiddenLayer().Forward(std::move(tmp), std::move(output));
+      arma::mat copy = output;
+      result += arma::norm(copy - input, 2);
+      copy.reshape(28, 28);
+      copy = copy.t();
+      mainOutput.submat(s * 28, c * 28, s * 28 + 27, c * 28 + 27) = copy;
+    }
+  }
+  BOOST_REQUIRE_LE(result / (numChains*numSamples), 700);
```
kris-singh: Hmmm.... okay i will change that

kris-singh:
Sorry for the mistakes. I will take care of these in the future. I will the fix errors asap

src/mlpack/core/optimizers/cdk/cdk.hpp
```diff
@@ -0,0 +1,100 @@
...
+  //! Modify the maximum number of iterations (0 indicates no limit).
+  size_t& MaxIterations() { return maxIterations; }
+
+  //! Get whether or not the individual functions are shuffled.
+  bool Shuffle() const { return shuffle; }
+  //! Modify whether or not the individual functions are shuffled.
+  bool& Shuffle() { return shuffle; }
+
+ private:
+  //! The instantiated function.
```
kris-singh: I did not understand this comment.

src/mlpack/methods/ann/rbm.hpp
```diff
@@ -0,0 +1,243 @@
...
+
+  /* 
+   * Intalise all the parameters of the network
+   * using the intialise rule. 
+   *
+   * @tparam: IntialiserType rule to intialise the parameters of the network
+   * @tparam: VisibleLayerType visible layer 
+   * @tparam: HiddenLyaerType hidden layer
+   */
+  RBM(arma::mat predictors, InitializationRuleType initializeRule,
```
kris-singh: I think we discussed this previously i think the data is stored locally here. It is consistent with FFN class as welll

src/mlpack/methods/ann/rbm_impl.hpp
```diff
@@ -0,0 +1,271 @@
...
+      return std::log(LogisticFunction::Fn(FreeEnergy(std::move(corruptInput)) -
+          FreeEnergy(std::move(arma::round(predictors.col(i)))))) *
+          predictors.n_rows;
+    }
+    else
+    {
+      // Cross entropy loss
+      visible.Forward(std::move(predictors.col(i)),
+        std::move(activation));
+      return arma::accu( predictors.col(i) * arma::log(activation).t() +
```
kris-singh: Well the activation can't be greater than 1 as it is output of the sigmoid unit whose range is 0 to 1. I will dig deeper to find the problem.

src/mlpack/core/optimizers/cdk/cdk.hpp
lozhnikov: > //! The instantiated FUNCTION.

I mean the comment is incorrect.

src/mlpack/methods/ann/rbm.hpp
lozhnikov: No, take a look at the FFN class.

kris-singh: Looking at line 92 constructor of the FFN class
FFN(arma::mat predictors,
      arma::mat responses,
      OutputLayerType outputLayer = OutputLayerType(),

src/mlpack/methods/ann/rbm_impl.hpp
lozhnikov: Ah, I see. I agree.

src/mlpack/tests/rbm_network_test.cpp
lozhnikov: I don't think that there is a simple solution for estimating the conformity of outputs by means of matrix norms.

kris-singh: I just wanted to check what the reconstruction error was here. Maybe what i could do is just take one sample from the data and test the reconstruction error on that say that value should be always less than some threshold.

kris-singh:
Don't know Travis is failing on X = {{1,0,1}.....}}. The intializer list is valid with c++11.

lozhnikov:
Try the following:
```
arma::mat("1.0 0.0 1.0;"
                  "1.0 0.0 0.0;"
                  /*etc*/);
```

src/mlpack/tests/rbm_network_test.cpp
```diff
@@ -0,0 +1,173 @@
...
+  BinaryLayer<> visible(trainData.n_rows, hiddenLayerSize, 1);
+  BinaryLayer<> hidden(hiddenLayerSize, trainData.n_rows, 0);
+  GaussianInitialization gaussian(0, 0.1);
+  RBM<GaussianInitialization, BinaryLayer<>, BinaryLayer<>> model(trainData,
+      gaussian, visible, hidden, 1,  true);
+  MiniBatchSGD msgd(10, 0.06, trainData.n_cols * 20, 0, true);
+  model.Train(trainData, msgd);
+  model.Reset();
+  // Set the parmaeters from a learned rbm
+  model.Parameters() = {-0.23224054, -0.23000632,
```
lozhnikov: Could you also fix the initializer list here?

src/mlpack/tests/rbm_network_test.cpp
```diff
@@ -0,0 +1,173 @@
...
+                        -0.25701271, -0.25122418,
+                        -0.20716651,  -0.20962217,
+                        -0.59922456, -0.60003836,
+                        -0.6, -0.625, -0.475};
+  // Check Weight Shared
+  BOOST_REQUIRE_CLOSE(arma::accu(model.VisibleLayer().Weight() -
+      model.HiddenLayer().Weight()), 0, 1e-2);
+
+  // Check free energy
+  arma::vec freeEnergy = {-0.87523715, 0.50615066, 0.46923476, 1.21509084};
```
lozhnikov: And here.

lozhnikov:
Regarding the serialization test:
You couldn't compile the test since the `BinaryLayer` class contains constants such as `inSize`, `outSize` and `typeVisible`. So, you should remove `const` modifiers.

kris-singh:
Only krann test is failing so you can review this.

src/mlpack/tests/rbm_network_test.cpp
```diff
@@ -0,0 +1,172 @@
...
+      testLabelsTemp.n_rows);
+
+  for (size_t i = 0; i < trainLabelsTemp.n_rows; ++i)
+  {
+    trainLabels(i) = arma::as_scalar(trainLabelsTemp.row(i));
+  }
+
+  for (size_t i = 0; i < testLabels.n_rows; ++i)
+  {
+    testLabels(i) = arma::as_scalar(testLabels.row(i));
```
lozhnikov: I found a critical error. That's equal to `testLables = testLabels`.

src/mlpack/tests/rbm_network_test.cpp
```diff
@@ -0,0 +1,172 @@
...
+}
+
+BOOST_AUTO_TEST_CASE(ClassificationTest)
+{
+  // Normalised dataset
+  int hiddenLayerSize = 100;
+  arma::mat trainData, testData, dataset;
+  arma::vec trainLabelsTemp, testLabelsTemp;
+  data::Load("mnist_data_small.csv", dataset, true);
+  trainData = dataset.submat(0, 0,
```
lozhnikov: As for me it is much better to divide the dataset by four parts e.g. `test.csv`, `train.csv`, `test_labels.csv`, `train_labels.csv` since your way doesn't allow anyone else to use the dataset as the the splitting algorithm is not documented anywhere.

src/mlpack/methods/ann/rbm_impl.hpp
```diff
@@ -0,0 +1,283 @@
...
+  // Collect the negative samples
+  Gibbs(std::move(predictors.col(input)), std::move(negativeSamples));
+
+  // Collect the negative gradients
+  visible.Forward(std::move(negativeSamples),
+      std::move(hiddenBiasNegativeGrad));
+  weightNegativeGrad = hiddenBiasNegativeGrad * negativeSamples.t();
+  visibleBiasNegativeGrad = negativeSamples;
+
+  output = (positiveGradient - negativeGradient);
```
lozhnikov: I guess the sign is invalid since you switched to the mini-batch SGD optimizer.

src/mlpack/methods/ann/rbm_impl.hpp
```diff
@@ -0,0 +1,283 @@
...
+  this->predictors = std::move(predictors);
+
+  if (!reset)
+  {
+    Reset();
+  }
+
+  // Train the model.
+  Timer::Start("rbm_optimization");
+  const double out = optimizer.Optimize(*this, parameter);
```
lozhnikov: Tthe `out` variable is not used.

src/mlpack/tests/rbm_network_test.cpp
```diff
@@ -0,0 +1,172 @@
...
+      gaussian, visible, hidden, 1,  true);
+  MiniBatchSGD msgd(10, 0.06, trainData.n_cols * 20, 0, true);
+  model.Train(trainData, msgd);
+  model.Reset();
+  // Set the parmaeters from a learned rbm
+  model.Parameters() = arma::mat(
+      "-0.23224054, -0.23000632, -0.25701271, -0.25122418, -0.20716651"
+      "-0.20962217, -0.59922456, -0.60003836, -0.6, -0.625, -0.475;");
+  // Check Weight Shared
+  BOOST_REQUIRE_CLOSE(arma::accu(model.VisibleLayer().Weight() -
```
lozhnikov: I think the weights should be equal. Why did you choose 1e-2 accuracy? The precision of `double` around 1 is equal to `1e-16` so you can set something like `1e-14`.

src/mlpack/tests/rbm_network_test.cpp
```diff
@@ -0,0 +1,172 @@
...
+   *        
+   */
+  arma::mat output;
+  BinaryLayer<> visible(trainData.n_rows, hiddenLayerSize, 1);
+  BinaryLayer<> hidden(hiddenLayerSize, trainData.n_rows, 0);
+  GaussianInitialization gaussian(0, 0.1);
+  RBM<GaussianInitialization, BinaryLayer<>, BinaryLayer<>> model(trainData,
+      gaussian, visible, hidden, 1,  true);
+  MiniBatchSGD msgd(10, 0.06, trainData.n_cols * 20, 0, true);
+  model.Train(trainData, msgd);
```
lozhnikov: I don't understand why you train the model if you don't use the parameters.

src/mlpack/tests/rbm_network_test.cpp
```diff
@@ -0,0 +1,172 @@
...
+  arma::vec calcultedFreeEnergy(4);
+  calcultedFreeEnergy.zeros();
+  for (size_t i = 0; i < trainData.n_cols; i++)
+  {
+    model.VisibleLayer().ForwardPreActivation(std::move(trainData.col(i)),
+        std::move(output));
+    calcultedFreeEnergy(i) = model.FreeEnergy(std::move(trainData.col(i)));
+  }
+
+  BOOST_REQUIRE_LE(arma::accu(calcultedFreeEnergy - freeEnergy), 1);
```
lozhnikov: I think here should be `BOOST_REQUIRE_CLOSE()`

src/mlpack/tests/rbm_network_test.cpp
kris-singh: I did not understand your point here can you elaborate a bit.
I have used arma::as_scalar for converting the 1x1 matrix to a scalar quantity

kris-singh:
Strange the accuray i was getting was around 90% i will check again and let you know..

src/mlpack/tests/rbm_network_test.cpp
lozhnikov: You don't initialize `testLabels` properly. You initialize `testLabelsTemp` but you don't copy that to `testLabels`, so `testLabels` is equal to zero. Moreover `arma::as_scalar(testLabels.row(i))` is equal to `testLabels(i)` since `testLabels` is a vector.

kris-singh: okay got your point so 
testLabels(i) = arma::as_scalar(testLabelsTemp.row(i));
even with this i am getting the classification accuarcy for softmax as 10%
and for rbm as 84%.
Are you using the same dataset that i sen't

lozhnikov: Yes, I use the same dataset. And I build the project from scratch.
If I change 
`testLabels(i) = arma::as_scalar(testLabels.row(i));` 
to 
`testLabels(i) = testLabelsTemp(i);`
the accuracy becomes 27% for the RBM.
But I think you didn't notice another error:
`for (size_t i = 0; i < testLabels.n_rows; ++i)`
`testLabels` is a row, so `testLabels.n_rows` is equal to 1.
In such a way, you should write `for (size_t i = 0; i < testLabelsTemp.n_rows; ++i)`.

When I fixed that I got 67% and 27% for regression and RBM respectively.

src/mlpack/tests/rbm_network_test.cpp
```diff
@@ -0,0 +1,172 @@
...
+    testLabels(i) = arma::as_scalar(testLabelsTemp.row(i));
+
+  std::cout << arma::size(trainLabels) << std::endl;
+  std::cout << arma::size(testLabels) << std::endl;
+
+  trainData = trainData.t();
+  testData = testData.t();
+
+  arma::mat output, XRbm(hiddenLayerSize, trainData.n_cols),
+      YRbm(hiddenLayerSize, trainLabels.n_cols);
```
lozhnikov: I think I found the fix. You don't define the dimensions of `YRbm` properly.
Try the following:
`YRbm(hiddenLayerSize, testData.n_cols)`

kris-singh:
Ready to merge.

src/mlpack/tests/rbm_network_test.cpp
```diff
@@ -0,0 +1,158 @@
...
+  BuildVanillaNetwork<>(X, 2);
+}
+
+BOOST_AUTO_TEST_CASE(ClassificationTest)
+{
+  // Normalised dataset
+  int hiddenLayerSize = 100;
+  arma::mat trainData, testData, dataset;
+  arma::vec trainLabelsTemp, testLabelsTemp;
+  data::Load("mnisttrain.txt", trainData, true);
```
lozhnikov: 1. If I am not mistaken, you didn't upload the MNIST dataset, you uploaded the Digit dataset that I sent you. In such a way the file names should look like `digit_train.txt`, `digit_train_labels.txt` etc.
2. Moreover you uploaded the same binary files two times: in 646c6c67363441a09490dcc4316304e9da4567a9 and in c2bd07c59cb3c0aa016c03012b236df142306350. So, I guess you should do git-rebase.
3. The official webpage of the dataset (http://archive.ics.uci.edu/ml/datasets/Pen-Based+Recognition+of+Handwritten+Digits) points out the citation policy (http://archive.ics.uci.edu/ml/citation_policy.html). So, I think we have to refer to the repo in the source code.

zoq: I'm curious, have you tested to store the datasets as ``arma_binary``? I think it would be great if we could keep the size of the repo and datasets low.

src/mlpack/methods/ann/layer/binary_layer_impl.hpp
```diff
@@ -0,0 +1,95 @@
...
+      output = weight.t() * input + otherBias;
+}
+
+template<typename InputDataType, typename OutputDataType>
+void BinaryLayer<InputDataType, OutputDataType>::Sample(InputDataType&& input,
+    OutputDataType&& output)
+{
+  Forward(std::move(input), std::move(output));
+  for (size_t i = 0; i < output.n_elem; i++)
+    output(i) = math::RandBernoulli(output(i));
```
zoq: I'm wondering if using ``arma::randi<arma::mat>(10, 10, arma::distr_param(0, 1));`` is faster.

src/mlpack/tests/rbm_network_test.cpp
kris-singh: The digits dataset(mnist-small) is form the sklearn example which in turn is taken form the MNIST dataset. So there is not much difference the only difference being since the sklearn example uses only a small subset they have augmented the data set by nudge in the the data set which is given in their code. I do agree i can cite the mnist dataset as they have done.

src/mlpack/methods/ann/layer/binary_layer_impl.hpp
kris-singh: I tested this and rand Bernoulli is faster

zoq: Okay, thanks for checking.

kris-singh:
If they are no further comments i think you can merge this.

lozhnikov:
Could you add binary data in a separate commit?

src/mlpack/tests/rbm_network_test.cpp
```diff
@@ -0,0 +1,182 @@
...
+
+BOOST_AUTO_TEST_CASE(ClassificationTest)
+{
+  // Normalised dataset
+  int hiddenLayerSize = 100;
+  arma::mat trainData, testData, dataset;
+  arma::mat trainLabelsTemp, testLabelsTemp;
+  trainData.load("digits_train.arm");
+  testData.load("digits_test.arm");
+  trainLabelsTemp.load("digits_train_label.arm");
```
lozhnikov: It seems I overlooked that. There is no sense in binary labels since each `double` utilize 8 bytes and each label is a digit. So, plain labels should take `2*numLabels` bytes (digit+separator). In my opinion it is better to save `arma::Row<size_t>` in plain format. That should simplify code since that allow to avoid temporary matrices (such as `trainLabelsTemp` and `testLabelsTemp`).

src/mlpack/tests/rbm_network_test.cpp
```diff
@@ -0,0 +1,182 @@
...
+  testLabelsTemp.load("digits_test_label.arm");
+
+  std::cout << arma::size(testLabelsTemp) << std::endl;
```
lozhnikov: Could you remove all unnecessary output?

lozhnikov:
I'll close this PR since the code was merged into PR #1046 (SpikeSlabRBM) and the code was much refactored in #1046 . So the binary RBM will be merged together with the spike and slab RBM.

PR #1029, opened 2017-06-15
Optimizer api

rcurtin:
@ShangtongZhang: this is a refactoring of the L-BFGS, SA, AugLagrangian, and SDP optimizers.  Now all that should need to happen is the SGD refactoring, then the various uses of the optimizers will have to be updated, and we can call it finished.  I didn't refactor any of the methods since SGD wasn't done, but I tested my code by commenting out all of the other tests.

I guess, you can either open a PR to my repository with updates and I'll merge it there, or you can cherry-pick my commits here and then I can close this PR.  Let me know what you think.


ShangtongZhang:
Thanks! I think I can cherry pick the commit.

src/mlpack/core/optimizers/lbfgs/lbfgs_impl.hpp
```diff
@@ -452,7 +282,12 @@ double L_BFGS<FunctionType>::Optimize(arma::mat& iterate)
     }
 
     // Overwrite an old basis set.
+<<<<<<< Updated upstream
```
zoq: Stashed changes.

src/mlpack/core/optimizers/sa/sa_impl.hpp
```diff
@@ -33,55 +29,51 @@ SA<FunctionType, CoolingScheduleType>::SA(
...
 //! Optimize the function (minimize).
-template<
-    typename FunctionType,
-    typename CoolingScheduleType
->
-double SA<FunctionType, CoolingScheduleType>::Optimize(arma::mat &iterate)
+template<typename CoolingScheduleType>
+template<typename FunctionType>
+double SA<CoolingScheduleType>::Optimize(FunctionType& function,
+                                         arma::mat &iterate)
```
zoq: Pedantic style issue.

src/mlpack/core/optimizers/aug_lagrangian/aug_lagrangian.hpp
```diff
@@ -70,20 +63,23 @@ class AugLagrangian
...
+   * @tparam LagrangianFunctionType Function which can be optimized by this
+   *     class.
    * @param coordinates Output matrix to store the optimized coordinates in.
    * @param maxIterations Maximum number of iterations of the Augmented
    *     Lagrangian algorithm.  0 indicates no maximum.
    * @param sigma Initial penalty parameter.
    */
-  bool Optimize(arma::mat& coordinates,
+  template<typename LagrangianFunctionType>
+  bool Optimize(LagrangianFunctionType& function,
```
zoq: I guess, we can update the documentation afterwards.

rcurtin: Ah, let me do that now.  I did this as a proof-of-concept mostly, and did some of the documentation, but it's easier to see what all needs to be fixed from the Github diffs.

src/mlpack/core/optimizers/lbfgs/lbfgs_impl.hpp
rcurtin: Oops, the merge wasn't clean I guess.

rcurtin:
Ok, if you can cherry pick, then I'll go ahead and close this PR and you can just pull from the branch.  If the merge was bad and some of the L-BFGS, AugLagrangian, SA, or SDP tests are failing (or failing to compile), let me know and I can work out the issue.

ShangtongZhang:
There seems something wrong within LBFGS, `UpdateBasisSet` is declared and never implemented, which leads to a link error.

rcurtin:
Oops, sorry about that, forgot to add lbfgs.cpp.  Just pushed the change now.

ShangtongZhang:
Thanks for the fix. Could you also look at `logistic_regression_impl.hpp`, there are function calls like `optimizer.Function()`, `GetInitialPoint()`.

rcurtin:
Hmm, I see, for this one, I think that we will have to construct `LogisticRegressionFunction` in `Train()` right before we do the training.  Inside `LogisticRegressionFunction` we can remove `GetInitialPoint()` and `initialPoint`, then make `lambda` a reference to be passed to the constructor.  The overload `LogisticRegression(OptimizerType<LogisticRegressionFunction<MatType>, OptimizerTypeArgs...>& optimizer)` can be replaced with

```
template<typename OptimizerType>
LogisticRegression(const MatType& predictors, const arma::Row<size_t>& responses, OptimizerType& optimizer);
```

and we can do the same modification to the `Train()` call.

I think, since now we are working in different branches, it would be easier for you to apply these changes than me---what do you think?  It is a bit hard to coordinate at this level, both working on one PR, I think :)

ShangtongZhang:
That makes sense. I can do that.

PR #1030, opened 2017-06-16
Adding dummy Evaluate function to FNN and RNN modules

sumedhghaisas:
The dummy functions are added to avoid an extra Forward pass. The dummy functions will be called from the optimizer. The dummy functions will return the cost computed by the forward pass done in Gradient call.

### DIFF
--- a/src/mlpack/methods/ann/ffn_impl.hpp
+++ b/src/mlpack/methods/ann/ffn_impl.hpp
@@ -169,6 +169,7 @@ template<typename OutputLayerType, typename InitializationRuleType>
 double FFN<OutputLayerType, InitializationRuleType>::ForwardCall(
     const arma::mat& /* parameters */, const size_t i, const bool deterministic)
 {
+  currentFunctionIndex = i;
   if (parameter.is_empty())
   {
     ResetParameters();
@@ -194,7 +195,7 @@ template<typename OutputLayerType, typename InitializationRuleType>
 double FFN<OutputLayerType, InitializationRuleType>::Evaluate(
     const arma::mat& parameters, const size_t i, const bool deterministic)
 {
-  if (currentCost == -1)
+  if (numFunctions == 1 || currentCost == -1 || currentFunctionIndex != i)
   {
     return ForwardCall(parameters, i, deterministic);
   }
--- a/src/mlpack/methods/ann/ffn_impl.hpp
+++ b/src/mlpack/methods/ann/ffn_impl.hpp
@@ -166,7 +166,7 @@ void FFN<OutputLayerType, InitializationRuleType>::Predict(
 }
 
 template<typename OutputLayerType, typename InitializationRuleType>
-double FFN<OutputLayerType, InitializationRuleType>::Evaluate(
+double FFN<OutputLayerType, InitializationRuleType>::ForwardCall(
     const arma::mat& /* parameters */, const size_t i, const bool deterministic)
 {
   if (parameter.is_empty())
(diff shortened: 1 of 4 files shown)


src/mlpack/methods/ann/ffn_impl.hpp
```diff
@@ -184,10 +184,22 @@ double FFN<OutputLayerType, InitializationRuleType>::Evaluate(
...
-  double res = outputLayer.Forward(std::move(boost::apply_visitor(
+  currentCost = outputLayer.Forward(std::move(boost::apply_visitor(
       outputParameterVisitor, network.back())), std::move(currentTarget));
 
-  return res;
+  return currentCost;
+}
+
+template<typename OutputLayerType, typename InitializationRuleType>
+double FFN<OutputLayerType, InitializationRuleType>::Evaluate(
```
zoq: Currently, if the optimizer calls the Gradient function we do Evaluate -> Backward -> Gradient inside the FFN class. Afterward, the Optimizer calls the Evaluate function again to keep track of the the latest changes (did the performance improve).

But with this change the performance analysis, the Evaluation call inside the Optimizer after the Gradient returns the same value as before, which is problematic if the input is just a single element since SGD will return with:
```
if (std::abs(lastObjective - overallObjective) < tolerance)
{
  Log::Info << "SGD: minimized within tolerance " << tolerance << "; "
      << "terminating optimization." << std::endl;
  return overallObjective;
}
```
The question is, do we handle that situation or do we expect samples > 1?

sumedhghaisas: I am not sure I am getting the problem. Maybe I am missing something. These return conditions are above both Gradient call and Evaluate call in the loop. So I think the overallObjective is correct at that point for returning even if the number of functions is 1.

and I have realized another mistake. If we run out of iterations for ANN. The last loop which computes overall Objective again has only Evaluate without gradient call. Thus all the Evaluate will return the same objective. I should check for objective function index there.

zoq: It depends on what you expect if I optimize something using a single iteration what I get is the performance before I updated the parameters right? That's why the gradient checks are failing.

rcurtin:
If I understand right, this PR aims to solve the following problem: the mlpack `SGD` and other optimizers often call both `Evaluate()` and `Gradient()`, because they have to calculate the objective function.  However there is often some overlap here and the assumptions we have on our optimizers don't allow for getting the objective function and the gradient at the same time, and therefore some computation is redundant.

But I want to take a step back: typically when a neural network is trained, one doesn't wait 'for convergence' to terminate the optimization.  Instead one just chooses a number of epochs that they want to do.  So calculating the objective function in order to find when convergence has been reached is not actually necessary for that use case.

So I might suggest instead, that we modify the `SGD` optimizer (since that's the most common neural network optimizer) to have either a template parameter or runtime parameter `useTolerance` (or similar), and when that is set to `false`, then the `Evaluate()` calls to compute the objective are simply not performed---in that mode, the only termination for the optimizer is if it hits the maximum number of iterations.

What do you think?  Is this a valid way to solve the problem?  Let me know if I have misunderstood anything.

rcurtin:
Any further thoughts on this one?  I am still not sure merging it is a good idea, because these changes assume that `Evaluate()` and `Gradient()` will always be called in a certain way, which may not be true in general.  So either we need to update the abstractions we are using for functions that we can optimize, or alternately we could go with the solution I suggested above.  (Or, we could decide that this is not a huge problem right now and just close the PR and move on too.)

rcurtin:
I'm going to go ahead and close this PR---I think the better way to do this optimization is just to implement a simple `EvaluateWithGradient()` for `FNN` and `RNN` and that should avoid any extra forward passes.

Issue #290, opened 2014-12-29
support arbitrary-precision types


rcurtin:
**Reported by rcurtin on 6 Nov 43643786 13:21 UTC**
Right now everything works on double and arma::mat.  Based on #300, it would be useful to support arbitrary-precision types (within the confines of what Armadillo offers).  This ticket right now just proposes the ideas.  There will be numerous difficult implementation details that may make this very difficult in the end; those can be detailed in the comments as they come up.

Migrated-From: http://trac.research.cc.gatech.edu/fastlab/ticket/302

shrit:
Probably we will close this one before mlpack 5.0

rcurtin:
Agreed, I think all the documentation-related refactoring and other work going on right now will end up with all mlpack algorithms accepting a matrix type soon. :+1:

shrit:
I think we can close this one, since we have this now on our list and we have already started working on it

DKGH:
In which release is this expected to be available?

rcurtin:
@DKGH it's supported by many but not all of the machine learning techniques that mlpack provides.  For instance, the documentation for the naive Bayes classifier has a [section showing how to use different element types](https://www.mlpack.org/doc/user/methods/naive_bayes_classifier.html#advanced-functionality-different-element-types).  Most of the other documented techniques have similar sections in the 'advanced functionality' parts near the bottom.

DKGH:
> [@DKGH](https://github.com/DKGH) it's supported by many but not all of the machine learning techniques that mlpack provides. For instance, the documentation for the naive Bayes classifier has a [section showing how to use different element types](https://www.mlpack.org/doc/user/methods/naive_bayes_classifier.html#advanced-functionality-different-element-types). Most of the other documented techniques have similar sections in the 'advanced functionality' parts near the bottom.

I was looking to do a standard normalization. Didn't work with arma::fmat. Worked with arma::mat.

PR #1031, opened 2017-06-16
Extracting meta information

micyril:
Add a tool for extracting meta information about machine learning algorithms.

### DIFF
--- a/src/mlpack/core/cv/meta_info_extractor.hpp
+++ b/src/mlpack/core/cv/meta_info_extractor.hpp
@@ -38,7 +38,7 @@ template<typename MatType,
          bool NumClasses>
 struct TrainForm;
 
-template<typename PT, typename WT, typename...SignatureParams>
+template<typename PT, typename WT, typename... SignatureParams>
 struct TrainFormBase
 {
   using PredictionsType = PT;
@@ -47,7 +47,7 @@ struct TrainFormBase
   /* A minimum number of parameters that should be inferred */
   static const size_t MinNumberOfAdditionalArgs = 1;
 
-  template<typename Class, typename RT, typename...Ts>
+  template<typename Class, typename RT, typename... Ts>
   using Type = RT(Class::*)(SignatureParams..., Ts...);
 };
 
@@ -101,20 +101,20 @@ struct NotFoundMethodForm
  * combination succeeds, or when there are no more combinations.
  */
 template<typename MLAlgorithm,
-         template<class, template<class...> class, size_t> class...HMFs>
+         template<class, template<class...> class, size_t> class... HMFs>
 struct SelectMethodForm;
 
 template<typename MLAlgorithm,
          template<class, template<class...> class, size_t> class HasMethodForm,
-         template<class, template<class...> class, size_t> class...HMFs>
+         template<class, template<class...> class, size_t> class... HMFs>
 struct SelectMethodForm<MLAlgorithm, HasMethodForm, HMFs...>
 {
-  template<typename...Forms>
+  template<typename... Forms>
   class From
   {
     /* Declaration and definition of Implementation for the case when
      * RemainingForms are empty */
-    template<typename...RemainingForms>
+    template<typename... RemainingForms>
     struct Implementation
     {
       using NextSMF = SelectMethodForm<MLAlgorithm, HMFs...>;
@@ -122,7 +122,7 @@ struct SelectMethodForm<MLAlgorithm, HasMethodForm, HMFs...>
     };
 
     /* The case when there is at least one remaining form */
-    template<typename Form, typename...RemainingForms>
+    template<typename Form, typename... RemainingForms>
     struct Implementation<Form, RemainingForms...>
     {
       using Type = typename std::conditional<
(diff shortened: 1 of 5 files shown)


src/mlpack/core/cv/meta_info_extractor.hpp
```diff
@@ -0,0 +1,249 @@
...
+ * @tparam DatasetInfo An indicator whether a data::DatasetInfo parameter should
+ *   be present.
+ * @tparam NumClasses An indicator whether the numClasses parameter should be
+ *   present.
+ */
+template<typename MatType,
+         typename PredictionsType,
+         typename WeightsType,
+         bool DatasetInfo,
+         bool NumClasses>
```
rcurtin: If I am understanding correctly, you will use this to determine regression or classification types, right?  `NumClasses` should be true for all classification types (some changes still necessary there, would you like me to do them?) and false for all regression types.

src/mlpack/core/cv/meta_info_extractor.hpp
```diff
@@ -0,0 +1,249 @@
...
+struct TrainForm;
+
+template<typename PT, typename WT, typename...SignatureParams>
```
rcurtin: Minor style comment---should there be a space between `typename...` and `SignatureParams`?

src/mlpack/core/cv/meta_info_extractor.hpp
```diff
@@ -0,0 +1,249 @@
...
+  using TF5 = TrainForm<MT, PT, void, true, true>;
+
+  /* Forms of Train with weights for regression algorithms */
+  using WTF1 = TrainForm<MT, arma::rowvec, WT, false, false>;
+  using WTF2 = TrainForm<MT, arma::mat, WT, false, false>;
+  using WTF3 = TrainForm<MT, PT, WT, false, false>;
+
+  /* Forms of Train with weights for classification algorithms */
+  using WTF4 = TrainForm<MT, PT, WT, false, true>;
+  using WTF5 = TrainForm<MT, PT, WT, true, true>;
```
rcurtin: I like the names of these variables. :)

src/mlpack/tests/cv_test.cpp
```diff
@@ -90,4 +98,78 @@ BOOST_AUTO_TEST_CASE(MSEMatResponsesTest)
...
+  CheckPredictionsType<FFN<>, arma::mat>();
+
+  CheckPredictionsType<LogisticRegression<>, arma::Row<size_t>>();
+  CheckPredictionsType<SoftmaxRegression<>, arma::Row<size_t>>();
+  CheckPredictionsType<HoeffdingTree<>, arma::Row<size_t>, arma::mat>();
+  CheckPredictionsType<HoeffdingTree<>, arma::Row<size_t>, arma::imat>();
+  CheckPredictionsType<DecisionTree<>, arma::Row<size_t>, arma::mat,
+      arma::Row<size_t>>();
+  CheckPredictionsType<DecisionTree<>, arma::Row<char>, arma::mat,
+      arma::Row<char>>();
```
rcurtin: Just out of curiosity, did you try alternate types for other classes than `DecisionTree<>` and have a failure?  I think since #290 is not done yet, there may be problems with that for some classes.

src/mlpack/core/cv/meta_info_extractor.hpp
```diff
@@ -0,0 +1,249 @@
...
+#include <mlpack/core/util/sfinae_utility.hpp>
+
+namespace mlpack {
+namespace cv {
+
+/*
+ * A wrapper struct for holding a Train form.
+ *
+ * @tparam MatType The type of data.
+ * @tparam PredictionsType The type of predictions.
```
rcurtin: The assumption here is that `LabelsType == PredictionsType`, right?  I think that's completely reasonable, I just want to make sure I am comprehending fully.

micyril: Yes, you're right.

src/mlpack/core/cv/meta_info_extractor.hpp
micyril: Yes, you're right again. It will be nice if you normalize API across classification algorithms.

src/mlpack/core/cv/meta_info_extractor.hpp
micyril: I don't mind adding a space. Maybe it's worth to explicitly state about this convention in the style guidelines?

src/mlpack/tests/cv_test.cpp
micyril: I was trying it also for HoeffdingTree<> in the line 121:
`CheckPredictionsType<HoeffdingTree<>, arma::Row<size_t>, arma::imat>();`
Here arma::imat is used is as the data type in the Train method.

src/mlpack/core/cv/meta_info_extractor.hpp
micyril: :)

src/mlpack/tests/cv_test.cpp
```diff
@@ -113,7 +113,7 @@ void CheckPredictionsType()
 BOOST_AUTO_TEST_CASE(PredictionsTypeTest)
 {
   CheckPredictionsType<LinearRegression, arma::rowvec>();
-  CheckPredictionsType<FFN<>, arma::mat>();
```
micyril: I have disabled the check since API for FFN have changed (now the Train method takes data and predictions as values, rather than as const references).

micyril:
By the way, if you don't see any problems, feel free to merge.

PR #1032, opened 2017-06-18
Update Optimizer API

ShangtongZhang:
Finally made it successfully compile.. Let's see if some test fails unexpectedly.


ShangtongZhang:
Should we delete `SoftmaxRegressionTrainTest` and `SoftmaxRegressionOptimizerTrainTest`? The initialization includes `randn`, so we cannot expect they have same result as before. (It works before because we could specify the function to optimize at that time)

src/mlpack/core/optimizers/sgdr/snapshot_sgdr_impl.hpp
```diff
@@ -64,16 +61,17 @@ double SnapshotSGDR<DecomposableFunctionType, UpdatePolicyType>::Optimize(
...
+  optimizer.DecayPolicy().EpochBatches() = function.NumFunctions() /
+      double(batchSize);
+
   // If a user changed the batch size we have to update the restart fraction
   // of the cyclical decay instantiation.
   if (optimizer.BatchSize() != batchSize)
   {
     batchSize = optimizer.BatchSize();
-    optimizer.DecayPolicy().EpochBatches() = function.NumFunctions() /
-        double(batchSize);
```
rcurtin: I'm not sure I follow the change here, can you explain why this was needed?  I think it is probably a simple reason I am overlooking.

src/mlpack/core/optimizers/smorms3/smorms3.hpp
```diff
@@ -97,29 +91,12 @@ class SMORMS3
    * @param iterate Starting point (will be modified).
    * @return Objective value of the final point.
    */
+  template<typename DecomposableFunctionType>
```
rcurtin: I know this is a bit tedious, but can you add the `@tparam DecomposableFunctionType ...` documentation to the `Optimize()` functions please since it's now removed from the class documentation?

src/mlpack/methods/ann/ffn.hpp
```diff
@@ -113,14 +113,10 @@ class FFN
...
-  template<
-      template<typename, typename...> class OptimizerType =
-          mlpack::optimization::RMSProp,
-      typename... OptimizerTypeArgs
-  >
+  template<typename OptimizerType>
   void Train(arma::mat predictors,
              arma::mat responses,
-             OptimizerType<NetworkType, OptimizerTypeArgs...>& optimizer);
+             OptimizerType& optimizer);
```
rcurtin: I think this is a nice improvement, this code is clearer now. :)

src/mlpack/methods/regularized_svd/regularized_svd_function.cpp
```diff
@@ -129,8 +129,8 @@ void RegularizedSVDFunction::Gradient(const arma::mat& parameters,
 namespace mlpack {
 namespace optimization {
 
-template<>
-double StandardSGD<mlpack::svd::RegularizedSVDFunction>::Optimize(
+template <> template <>
```
rcurtin: I think we should put these on two separate lines.  Also, I notice a lot of `template <typename ...>`, but I think we should use `template<typename ...>` (i.e. no space).  I'm not sure why the style check doesn't pick that up, I haven't looked into it.

src/mlpack/methods/softmax_regression/softmax_regression.hpp
```diff
@@ -98,15 +96,25 @@ class SoftmaxRegression
...
+   * @param numClasses Number of classes for classification.
+   * @param optimizer Desired optimizer.
+   * @param lambda L2-regularization constant.
+   * @param fitIntercept add intercept term or not.
    */
-  SoftmaxRegression(OptimizerType<SoftmaxRegressionFunction>& optimizer);
+  SoftmaxRegression(const arma::mat& data,
+                    const arma::Row<size_t>& labels,
+                    const size_t numClasses,
+                    OptimizerType& optimizer,
```
rcurtin: Related to #1021, I think we could just use `OptimizerType optimizer = OptimizerType()` here and use that as the last argument to the other constructor.  But I don't see the need to refactor everything here, we can save that for later.

src/mlpack/methods/softmax_regression/softmax_regression.hpp
```diff
@@ -59,9 +59,7 @@ namespace regression {
  * regressor2.Classify(test_data, predictions2);
  * @endcode
  */
-template<
-  template<typename> class OptimizerType = mlpack::optimization::L_BFGS
->
+template<typename OptimizerType = mlpack::optimization::L_BFGS>
```
rcurtin: I think we can templatize only `Train()` and the constructors that train the model, no need to store an optimizer internally.

src/mlpack/methods/sparse_autoencoder/sparse_autoencoder.hpp
```diff
@@ -62,9 +62,7 @@ namespace nn {
  * @tparam OptimizerType The optimizer to use; by default this is L-BFGS.  Any
  *     mlpack optimizer can be used here.
  */
-template<
-  template<typename> class OptimizerType = mlpack::optimization::L_BFGS
->
+template<typename OptimizerType = mlpack::optimization::L_BFGS>
```
rcurtin: Same here, I think only `Train()` and the constructors that train the model need to have `OptimizerType` as a template parameter.

src/mlpack/tests/sgdr_test.cpp
```diff
@@ -43,7 +43,8 @@ BOOST_AUTO_TEST_CASE(CyclicalResetTest)
     {
       double epochStepSize = stepSize;
 
-      CyclicalDecay cyclicalDecay(restart, double(mult), stepSize, 10, 1000);
+      CyclicalDecay cyclicalDecay(restart, double(mult), stepSize);
+      cyclicalDecay.EpochBatches() = (double)1000 / 10;
```
rcurtin: Can you add a space between the cast and the argument?  For this one you could also just write `100.0`. :)

src/mlpack/tests/softmax_regression_test.cpp
```diff
@@ -362,12 +362,8 @@ BOOST_AUTO_TEST_CASE(SoftmaxRegressionTrainTest)
 
   // This should be the same as the default parameters given by
   // SoftmaxRegression.
-  SoftmaxRegressionFunction srf(dataset, labels, 2, 0.0001, false);
-  L_BFGS<SoftmaxRegressionFunction> lbfgs(srf);
-  SoftmaxRegression<> sr(lbfgs);
-
+  SoftmaxRegression<> sr(dataset, labels, 2, 0.0001, false);
   SoftmaxRegression<> sr2(dataset.n_rows, 2);
-  sr2.Parameters() = srf.GetInitialPoint(); // Start from the same place.
```
rcurtin: The idea here was just to make sure that the optimizations start from the same initial point, but we have to be a little more tricky now about how to do that since we can't get the initial point as easily---we have to use the constructor that initializes the parameters but does not train:

```
SoftmaxRegression<> sr(dataset.n_rows, 2);
SoftmaxRegression<> sr2(dataset.n_rows, 2);

// Force the same initial optimization point.
sr.Parameters() = sr2.Parameters();

// Now train both models.
sr.Train(dataset, labels, 2);
sr2.Train(dataset, labels, 2);
```

src/mlpack/tests/softmax_regression_test.cpp
```diff
@@ -392,14 +388,13 @@ BOOST_AUTO_TEST_CASE(SoftmaxRegressionOptimizerTrainTest)
...
+  L_BFGS lbfgs;
+  SoftmaxRegression<> sr(dataset, labels, 2, lbfgs, 0.01, true);
 
+  L_BFGS lbfgs2;
   SoftmaxRegression<> sr2(dataset.n_rows, 2, true);
-  L_BFGS<SoftmaxRegressionFunction> lbfgs2(srf);
-  sr2.Parameters() = srf.GetInitialPoint();
-  sr2.Train(lbfgs2);
+  sr2.Lambda() = 0.01;
+  sr2.Train(dataset, labels, 2, lbfgs2);
```
rcurtin: For this one, we can initialize the parameters in the same way, but then train as

```
sr.Train(dataset, labels, 2, lbfgs);
sr2.Train(dataset, labels, 2, lbfgs2);
```

src/mlpack/core/optimizers/sgdr/snapshot_sgdr_impl.hpp
ShangtongZhang: See the change of `src/mlpack/core/optimizers/sgdr/cyclical_decay.hpp`. In the past we compute `epochBatches` at CTOR, but now we cannot do this as we won't pass in `function` to CTOR anymore. So I just compute this value every time before calling the `Optimize`

src/mlpack/tests/softmax_regression_test.cpp
ShangtongZhang: But does this fix really make sense? The original test is to make sure the training result from CTOR and from `Train` function call are same. By this fix it looks trivial -- they are both calling `Train`.

ShangtongZhang:
Hope it is ready to merge now.

src/mlpack/core/optimizers/sa/sa.hpp
```diff
@@ -193,8 +172,11 @@ class SA
    * @param sweepCounter Current counter representing how many sweeps have been
    *      completed.
    */
-  void GenerateMove(arma::mat& iterate,
+  template<typename FunctionType>
+  void GenerateMove(FunctionType& function,
+                    arma::mat& iterate,
                     arma::mat& accept,
+                    arma::mat& moveSize,
```
zoq: Do you mind to add a comment about ``moveSize``, don't feel obligated we can do this later.

src/mlpack/tests/softmax_regression_test.cpp
rcurtin: Ah yes, sorry, we should train the second object with an L_BFGS object:

```
L_BFGS lbfgs; // Use default options.
sr2.Train(dataset, labels, 2, lbfgs);
```

This way we test that passing an optimizer still gives us the same result.

rcurtin:
Thanks!  I am happy with this change, I think it is a good improvement. :)

PR #1034, opened 2017-06-20
Support batched forward and backward for FFN

ShangtongZhang:
- This PR exposes `Forward` and `Backward`  of `FFN` and it works in real batch mode ( By fake batch mode I refer to current `minibatch-sgd`, which does update column by column rather than by matrix multiplication).

- Until now I only refactored `Linear` layer. I don't know exactly which layers are not matrix compatible. Do you have any idea about this? @zoq 

- I no longer need `Gradient()`, which I added several weeks ago. So I just removed it. Hope nobody is using it.

- If this PR works, this https://github.com/mlpack/mlpack/pull/1030 may become unnecessary.


zoq:
> This PR exposes Forward and Backward of FFN and it works in real batch mode ( By fake batch mode I refer to current minibatch-sgd, which does update column by column rather than by matrix multiplication).

Nice idea, maybe in the future we can add real batch mode to the optimizer classes as well. So that we can update the parameter in a single call instead of doing Forward/Backward/Update. Would avoid some boilerplate code as well.

> Until now I only refactored Linear layer. I don't know exactly which layers are not matrix compatible. Do you have any idea about this? @zoq

Every layer should be matrix compatible, the conv layer might be tricky, but I can help you with that if you like.

> I no longer need Gradient(), which I added several weeks ago. So I just removed it. Hope nobody is using it.

It's save to delete the ``Gradient`` function.

> If this PR works, this #1030 may become unnecessary.

#1030 follows another idea which allows us to reuse the existing optimizer classes. Which is helpful if someone likes to use the e.g. cross validation in the future.

zoq:
Actually adding real batch support for the existing optimizer isn't a problem, we could reuse the existing interface and internally do batch training, by using the given index (``Gradient``function) as starting point.

ShangtongZhang:
So I think the remaining part of this PR would be to make sure all the layers are compatible with matrix. It would be great if you can help me with conv layers.

zoq:
Sounds good, I'll take a look at the conv layer.

ShangtongZhang:
I read all the implementations of layers except conv related layer. To my best knowledge, I only find `LogSoftmax` needs refactoring. And I find many layers already compatible with matrix (i.e. they have a for loop over all columns). I'm not sure whether I missed something. It will be appreciated if you could double check them.

zoq:
I think we should also test ``Add`` and ``Constant``, but besides that I think this is ready.

rcurtin:
Since we haven't yet released 3.0.0, this is a good time to semi-painlessly (i.e. without thinking too much about reverse compatibility) revisit our optimizer API and see if it meets the needs that we have.  Right now an optimizer will call the functions

```
double Evaluate(const arma::mat& coordinates);
void Gradient(const arma::mat& coordinates, arma::mat& gradient);
```

and if the function type is a "decomposable" function, the functions

```
double Evaluate(const arma::mat& coordinates, const size_t i, arma::mat& gradient);
void Gradient(const arma::mat& coordinates, const size_t i, arma::mat& gradient);
```

where each `i` represents a single part of the decomposable objective function.  In the case of neural networks, each decomposed function in the `DecomposableFunctionType` class corresponds to a single point.  For many machine learning algorithms, this abstraction requires the `DecomposableFunctionType` to hold the data internally, and then use the `i` parameter to find the corresponding correct part of the data.  This is a little bit awkward, and I think the code can be a little hard to understand because of this.  As a result of this awkward abstraction, it's very difficult to fit batch computation into the existing framework that we have.

Here is an alternate idea: instead of the `DecomposableFunctionType` abstraction, we instead replace this with a `DataDependentFunctionType` that has the following functions:

```
// Evaluate the objective function for the given data at the given coordinates
// (for neural networks the coordinates are the parameters of the network).
template<typename DataType>
double Evaluate(const arma::mat& coordinates, const DataType& data);

// Evaluate the gradient for the given data at the given coordinates
// (for neural networks the coordinates are the parameters of the network).
template<typename DataType>
void Gradient(const arma::mat& coordinates, const DataType& data, arma::mat& gradient);
```

In this way, we could call, e.g., `Evaluate(coordinates, data.cols(1, 10))` or similar, and this would be equivalent to what could have been done before as:

```
double obj = 0;
for (size_t i = 1; i < 10; ++i)
  obj += Evaluate(coordinates, i);
```

and for neural networks (and probably many other problems) this can allow us to get significant speedup via batch computations, and hopefully keep the abstractions kind of clean.

What does everyone think?  I know there are a lot of abstraction changes that I've proposed and we've made recently, but I think it's important to try and get these right now instead of suffering from an awkward API that makes life hard later in the 3.x lifecycle.  I can help implement (or completely implement) this one, it will take a little time until I am able to though.

ShangtongZhang:
It looks cool and will make the code more clear. And for this I think we mainly need to work on implementation, test cases will be barely influenced, which is excellent -- refactoring so many tests are really painful. But I'm afraid I don't have time to do this in the near future, at least before I finish async RL methods. BTW I didn't realize there is a difference between `DecomposableFunctionType` and `FunctionType` before this post, I thought it was just a random event... some guy uses `DecomposableFunctionType` while other guys prefer `FunctionType` :)

zoq:
What I like about the current abstraction is that its data independent if we make the data part of the Evaluation function, we have to make sure every data type implements ``data.cols(...)``. Which might be difficult for arma::cube and arma::field, arma::vec. We could pass the indices, instead of a single index but as @rcurtin pointed out this is sometimes a little bit awkward, at least for someone trying to understand/follow the code.

Also, if we go that way, each method has to support real batches or we have to loop over the specified batch internally. I guess right now I could do the same, but the other way around, if the batch size and index is known I could do data.cols(index + batchSize).

ShangtongZhang:
Hi @zoq , could you help me `constant`? I don't know how to refactor it -- I notice in `Forward` and `Backward` of `constant`, the parameter type is `DataType` rather than `Mat`, does that mean it would be `cube` some time?

zoq:
I can refactor the ``Constant`` layer, about the ``DataType``, not ``cube`` but maybe ``arma::mat<int>`` or something in that direction in the future.

rcurtin:
> Also, if we go that way, each method has to support real batches or we have to loop over the specified batch internally. I guess right now I could do the same, but the other way around, if the batch size and index is known I could do data.cols(index + batchSize).

So, this could work, if we could call `Evaluate(coordinates, i, batchSize)` and it would evaluate `batchSize` points, but this doesn't work well for optimizers that want to shuffle the data.  If `SGD` shuffles data now, it simply reorders all the indices, then passes those indices.  But if we want to do batches of shuffled data, either we need a way to accept non-contiguous indices (which I think will be slow), or we need some way to tell the function being optimized to shuffle its data.

I am not sure, I guess I could see a function interface like this:

```
class DecomposableFunctionType
{
  void Shuffle(); // shuffle any functions
  double Evaluate(const arma::mat& coordinates, const size_t begin, const size_t batchSize);
  double Evaluate(const arma::mat& coordinates);
  void Gradient(const arma::mat& coordinates, const size_t begin, const size_t batchSize, arma::mat& gradient);
  void Gradient(const arma::mat& coordinates, arma::mat& gradient);
  size_t NumFunctions();
};
```

We could also remove the other `Evaluate()` and `Gradient()` overload, instructing optimizers instead to call `Evaluate(coordinates, 0, NumFunctions())` and `Gradient(coordinates, 0, NumFunctions(), gradient)`.

This preserves the data independence of the optimizers, but requires the functions to now have a way to shuffle their data*.  What do you think, is that a better way to go?

* I suppose it could also be possible to use SFINAE to only call `Shuffle()` if it exists, and do something else otherwise.  But that's getting into details.

zoq:
> But if we want to do batches of shuffled data, either we need a way to accept non-contiguous indices (which I think will be slow), or we need some way to tell the function being optimized to shuffle its data.

I think that depends on the premise, whether we like to support batches over shuffles data, if I remember correctly, the Mini Batch SGD shuffles batches and preserves the data order, which I think is okay. So do we plan to support batches over shuffled data?

> This preserves the data independence of the optimizers, but requires the functions to now have a way to shuffle their data*. What do you think, is that a better way to go?

I like the interface, it's similar to the current interface. Also using ``Evaluate(coordinates, 0, NumFunctions())`` to avoid an additional function sounds like a reasonable idea to me.

rcurtin:
>> But if we want to do batches of shuffled data, either we need a way to accept non-contiguous indices (which I think will be slow), or we need some way to tell the function being optimized to shuffle its data.
>
> I think that depends on the premise, whether we like to support batches over shuffles data, if I remember correctly, the Mini Batch SGD shuffles batches and preserves the data order, which I think is okay. So do we plan to support batches over shuffled data?

Hmm, the problem with non-shuffled contiguous minibatches is that our gradient can be very biased: i.e. suppose I have my dataset with half class 0 and half class 1, but it's ordered such that all the class 0 points come first, then all the class 1 points.  Even if I split into many minibatches, all minibatches (except perhaps one) will have only one class, meaning that every gradient step will be biased towards the class of the minibatch.  I think (I am not sure) that this will affect how quickly minibatch SGD will converge to a good solution, so I think it is better to shuffle the entire dataset, then do contiguous minibatches, as opposed to just doing contiguous minibatches.  Let me know what you think.  It sounds like we are close to a solution, we just need to work out some small details. :)

zoq:
> Hmm, the problem with non-shuffled contiguous minibatches is that our gradient can be very biased: i.e. suppose I have my dataset with half class 0 and half class 1, but it's ordered such that all the class 0 points come first, then all the class 1 points. Even if I split into many minibatches, all minibatches (except perhaps one) will have only one class, meaning that every gradient step will be biased towards the class of the minibatch. 

I thought about the issue, and you are right if someone passes a sorted dataset this could be a problem, I guess, we can't assume that that the data is shuffled before training. I'm wondering what is the tradeoff between non-contiguous indices and contiguous indices. For the contiguous indices, we have to create a copy of the dataset and depending on the optimizer, shuffle the complete dataset, for SGD we would shuffle every NumFunctions? For the non-contiguous case, we don't have to copy the dataset and can shuffle over the indices which should be less expensive. Not sure one of the approaches is superior in every case. Maybe we should write a simple microbenchmark to test this out unless you have some insights that could be helpful, or maybe I missed something?

rcurtin:
Sorry for the slow response on this one.  If we do a non-contiguous access of data we are almost guaranteed that `Evaluate()` and `Gradient()` will be slower by possibly an order of magnitude (depending) since the prefetcher can't work as well and we won't be accessing contiguous data anymore.  So I think it is important to access data in a contiguous manner.

However I should also point out that `SGD`, which currently has a boolean `shuffle` option, cannot access data contiguously when `shuffle == true` because it has no way to know anything about the data at all.

But like you pointed out, shuffling in minibatch SGD has some extra complexity because you can shuffle two ways: shuffle the batches themselves (i.e. do batch 5 then 2 then 3... etc...) or shuffle the data and proceed through the batches as normal.

It seems to me like we could provide both of these possibilities for functionality if we add a `Shuffle()` function to the requirements for `DecomposableFunctionType`.  The first strategy (shuffle batches only) can operate the same way `SGD` does now.  The second strategy can call `Shuffle()` at the beginning of each iteration.

What do you think?  I think this is a decent way to solve the problem and also allow the user flexibility in their desired strategy.  I would say that by default with minibatch SGD we should use the shuffle batches strategy (not the shuffle data strategy).

zoq:
> It seems to me like we could provide both of these possibilities for functionality if we add a Shuffle() function to the requirements for DecomposableFunctionType. The first strategy (shuffle batches only) can operate the same way SGD does now. The second strategy can call Shuffle() at the beginning of each iteration.

Sounds reasonable to me, a good combination of both possibilities. I was thinking if we add ``Shuffle()`` to the requirements for DecomposableFunctionType, we have to implement the function for each method that support's SGD/ SGD derivatives, can we implement a general ``Shuffle``, and each method can specialize the ``Shuffle`` method if necessary. But I guess, in this case, we have to provide a function to access the data, which is probably more work as providing the ``Shuffle`` function.

rcurtin:
Hmm, maybe we can use sone SFINAE and if `Shuffle()` doesn't exist in the function, we throw an exception if the user asked for a data shuffle.  Do you think that sounds reasonable?

zoq:
That will do the trick, yes. I think we should open a new issue, and wait if someone likes to pick it up, if not I can do the refactoring, probably by the end of the week. Does this sounds reasonable?

rcurtin:
Sure, that works for me.  Unfortunately I don't think I will be able to make time to do it this week, but if you can, I think it would be great to see the change. :)

zoq:
Note this PR is independent of #1047 and introduces batch support for the majority of the ann layer.

PR #1035, opened 2017-06-21
Random forest implementation

rcurtin:
Here is an implementation of Breiman's random forests based on the existing decision tree implementation.  I've also included a nice speedup of the Gini gain calculation by rewriting the loop so that gcc will autovectorize and use SIMD instructions.  The speedup seems to be 2.5x+ in some cases.

This implementation uses OpenMP to parallelize the training of trees, but I do wonder if the decision trees themselves can be parallelized to get additional speedup.  I have played with it a little bit but not enough to have success yet.


src/mlpack/methods/decision_tree/gini_gain.hpp
```diff
@@ -44,37 +44,128 @@ class GiniGain
     if (labels.n_elem == 0)
       return 0.0;
 
-    // Count the number of elements in each class.
-    arma::vec counts(numClasses, arma::fill::zeros);
+    // Count the number of elements in each class.  Use four auxiliary vectors
```
zoq: Really nice, maybe there are other places we can do the same, and get about the same speedup.

src/mlpack/methods/decision_tree/multiple_random_dimension_select.hpp
```diff
@@ -0,0 +1,88 @@
...
+            unique = false;
+            break;
+          }
+        }
+      }
+
+      values[i] = value;
+    }
+
+    values[NumDimensions] = size_t(-1);
```
zoq: huh, isn't ``size_t(-1)`` = 0?

src/mlpack/methods/random_forest/bootstrap.hpp
```diff
@@ -0,0 +1,47 @@
...
+{
+  bootstrapDataset.set_size(dataset.n_rows, dataset.n_cols);
+  bootstrapLabels.set_size(labels.n_elem);
+  if (UseWeights)
+    bootstrapWeights.set_size(weights.n_elem);
+
+  // Random sampling with replacement.
+  for (size_t i = 0; i < dataset.n_cols; ++i)
+  {
+    const size_t index = math::RandInt(dataset.n_cols);
```
zoq: I'm wondering if using ``randi`` is faster, if we generate the random values once, but in this case we have to store them.

src/mlpack/methods/random_forest/random_forest.hpp
```diff
@@ -0,0 +1,297 @@
...
+             const size_t numTrees = 50,
+             const size_t minimumLeafSize = 20);
+
+  /**
+   * Train the random forest on the given labeled training data with the given
+   * dataset info and the given number of trees.  The minimumLeafSize parameter
+   * is given to each individual decision tree during tree building.  This
+   * overload can be used to train on categorical data.
+   *
+   * @param dataset Dataset to train on.
```
zoq: I think the parameter should match with the function.

src/mlpack/methods/random_forest/bootstrap.hpp
rcurtin: A simple test:

```
#include <mlpack/core.hpp>
#include <mlpack/methods/random_forest/bootstrap.hpp>

using namespace mlpack;
using namespace arma;

int main()
{
  mat dataset(10, 100000, arma::fill::randu);
  Col<size_t> labels(100000, fill::zeros);
  vec weights;

  mat newset;
  Col<size_t> newlabels;

  for (size_t i = 0; i < 1000; ++i)
  {
    tree::Bootstrap<false>(dataset, labels, weights, newset, newlabels, weights);
  }
}
```

With randi:

```
$ time ./test-randi

real    0m20.356s
user    0m20.376s
sys     0m0.348s
```

Without (i.e. the old code):

```
$ time ./test

real    0m33.492s
user    0m33.512s
sys     0m0.348s
```

I had not thought this could be slow!  Thanks for pointing it out.

src/mlpack/methods/decision_tree/gini_gain.hpp
rcurtin: Yeah, this will be a primary place to look for optimizations: if SIMD instructions aren't being used where they could be, this is like a 4x slowdown in some cases.  Unfortunately digging deep to figure out what is being used is a little tedious.

I updated the code so that the information gain uses this approach also.

src/mlpack/methods/decision_tree/multiple_random_dimension_select.hpp
rcurtin: I thought `size_t(-1)` was `std::numeric_limits<size_t>::max()`, but I'll just go ahead and change it.

rcurtin:
The code here is done, but the travis build fails because of `RAModelTest`.  So I think this one is ready to merge.

src/mlpack/methods/random_forest/random_forest_main.cpp
```diff
@@ -0,0 +1,190 @@
...
+        << "); must be greater than 0!" << endl;
+  if (CLI::HasParam("predictions") && !CLI::HasParam("test"))
+    Log::Warn << "--predictions_file ignored because --test_file not specified."
+        << endl;
+  if (CLI::HasParam("probabilities") && !CLI::HasParam("test"))
+    Log::Warn << "--probabilities_file ignored because --test_file not "
+        << "specified." << endl;
+  if (CLI::HasParam("minimum_leaf_size") &&
+      CLI::GetParam<int>("minimum_leaf_size") <= 0)
+    Log::Fatal << "Invalid minimum leaf size ("
```
zoq: I think it would improve the readability if we would use clamps here.

rcurtin:
Ok, I will leave this open for 5 days in case anyone has any comments, then merge it.

PR #1036, opened 2017-06-22
Trust Region Newton method.

zoq:
Here is an implementation of the Trust Region Newton method as described in "Trust region Newton methods for large-scale logistic regression". Tested on the isolet dataset I get TrustRegionNewton: 326.386, L_BFGS: 464.662 ms. About to run benchmarks on all files mentioned in the paper, but this might take some time. It might also be interesting to look into: "A Study on Trust Region Update Rules in Newton Methods for Large-scale Linear Classification".

### DIFF
--- a/src/mlpack/core/optimizers/trust_region_newton/trust_region_newton_impl.hpp
+++ b/src/mlpack/core/optimizers/trust_region_newton/trust_region_newton_impl.hpp
@@ -1,0 +1,214 @@
+/**
+ * @file trust_region_newton_impl.hpp
+ * @author Marcus Edel
+ *
+ * Implementation of the Trust Region Newton Method.
+ *
+ * mlpack is free software; you may redistribute it and/or modify it under the
+ * terms of the 3-clause BSD license.  You should have received a copy of the
+ * 3-clause BSD license along with mlpack.  If not, see
+ * http://www.opensource.org/licenses/BSD-3-Clause for more information.
+ */
+#ifndef MLPACK_CORE_OPTIMIZERS_TRUST_REGION_NEWTON_TRUST_REGION_NEWTON_IMPL_HPP
+#define MLPACK_CORE_OPTIMIZERS_TRUST_REGION_NEWTON_TRUST_REGION_NEWTON_IMPL_HPP
+
+// In case it hasn't been included yet.
+#include "trust_region_newton.hpp"
+
+namespace mlpack {
+namespace optimization {
+
+template<typename FunctionType>
+TrustRegionNewton<FunctionType>::TrustRegionNewton(
+    FunctionType& function,
+    const double minGradientNorm,
+    const size_t maxIterations,
+    const size_t maxConjugateIterations,
+    const double eta0,
+    const double eta1,
+    const double eta2,
+    const double sigma1,
+    const double sigma2,
+    const double sigma3) :
+    function(function),
+    minGradientNorm(minGradientNorm),
+    maxIterations(maxIterations),
+    maxConjugateIterations(maxConjugateIterations),
+    eta0(eta0),
+    eta1(eta1),
+    eta2(eta2),
+    sigma1(sigma1),
+    sigma2(sigma2),
+    sigma3(sigma3)
+{ /* Nothing to do. */ }
+
+template<typename FunctionType>
+void TrustRegionNewton<FunctionType>::ConjugateGradient(
+    FunctionType& function,
+    const double delta,
+    const arma::mat& iterate)
+{
+  // The initial gradient values.
+  r = -gradient;
+  arma::mat d = r;
+  arma::mat hessian;
+
+  s.zeros();
+  double rTr = arma::dot(r, r);
+
+  // Whether to optimize until convergence.
+  const bool optimizeUntilConvergence = (maxConjugateIterations == 0);
+
+  const double minGradientNormConjugate = 0.5 * arma::norm(gradient, 2);
+  for (size_t itNum = 0; optimizeUntilConvergence ||
+      (itNum != maxConjugateIterations); ++itNum)
+  {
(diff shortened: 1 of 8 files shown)


zoq:
First results, each experiment used the default optimizer parameter and the same stop criterium:

```
a9a
----------------------------
# Number of classes: 2
# Number of data: 32,561 / 16,281 (testing)
# Number of features: 123 / 123 (testing)

TrustRegionNewton: 278.213 ms
Accuracy: 83.6277

L_BFGS: 18357 ms
Accuracy: 84.9206
```

```
HIGGS
----------------------------
# Number of classes: 2
# Number of data: 11,000,000
# Number of features: 28

TrustRegionNewton: 26276.5 ms
Accuracy: 64.4766

L_BFGS: 191399 ms
Accuracy: 64.8545
```

```
gisette_scale
----------------------------
# Number of classes: 2
# Number of data: 6,000 / 1,000 (testing)
# Number of features: 5,000

TrustRegionNewton: 1578.3 ms
Accuracy: 84.85

L_BFGS: 15872.1 ms
Accuracy: 93.6167
```

src/mlpack/methods/logistic_regression/logistic_regression_function_impl.hpp
```diff
@@ -170,6 +171,47 @@ void LogisticRegressionFunction<MatType>::Gradient(
...
+{
+  hessian.set_size(gradient.n_elem);
+  hessian(0) = gradient(0) * parameters(0);
+
+  arma::mat hessianTemp = arma::mat(
+      hessian.colptr(0) + 1, 1, gradient.n_elem - 1, false, true);
+  hessianTemp = (lambda *
+      (gradient.col(0).subvec(1, parameters.n_elem - 1).t() *
+      predictors % derivative)) * predictors.t() +
+      gradient.col(0).subvec(1, parameters.n_elem - 1).t();
```
rcurtin: I think this is computing (according to the notation in the extended paper) `((d^i)^T \nabla^2 f(w^k) d^i)`, is that correct?  If so I think we could simplify the required interface for functions here.  Since `d` is from the TRON algorithm only and not specific to linear regression, I think it is possible to do the calculation above in `ConjugateGradient()` and restrict this to only compute `\nabla^2 f(w^k)`, which could let us simplify the interface in one of two ways:

```
// set both gradient and hessian
void Hessian(const mat& iterate, mat& gradient, mat& hessian);
```

or

```
void Gradient(const mat& iterate, mat& gradient);
void Hessian(const mat& iterate, mat& hessian);
```

Of course this comment only applies if my initial statement is true.

src/mlpack/core/optimizers/trust_region_newton/trust_region_newton_impl.hpp
```diff
@@ -0,0 +1,214 @@
...
+
+    if (itNum > 0)
+    {
+      const double rTrUpdate = arma::dot(r, r);
+      d *= (rTrUpdate / rTr);
+      d += r;
+      rTr = rTrUpdate;
+    }
+
+    function.Hessian(iterate, d, derivative, hessian);
```
rcurtin: If what I suggested in the comment about `logistic_regression_function_impl.hpp` is right, then I think that we can calculate the Hessian only once at the beginning of the whole conjugate gradient inner optimization.

src/mlpack/core/optimizers/trust_region_newton/trust_region_newton_impl.hpp
```diff
@@ -0,0 +1,214 @@
...
+  {
+    // If ||r^i|| <= e_k||detla f(w^k)|| then output sk =s^(i) and stop.
+    if (arma::norm(r, 2) <= minGradientNormConjugate)
+      break;
+
+    if (itNum > 0)
+    {
+      const double rTrUpdate = arma::dot(r, r);
+      d *= (rTrUpdate / rTr);
+      d += r;
```
rcurtin: Might be better to have these as one line, but I'm not sure how much of a difference it would make for Armadillo---I guess that all on one line means that this can be done in only one loop over `d` instead of 2.

src/mlpack/methods/logistic_regression/logistic_regression_function.hpp
```diff
@@ -28,14 +29,33 @@ template<typename MatType = arma::mat>
...
+   * for L2-regularization.  If not specified, it is set to 1,
+   *
+   * @param predictors Input training variables.
+   * @param responses Outputs results from input training variables.
+   * @param lambda L2-regularization parameter.
+   */
   LogisticRegressionFunction(const MatType& predictors,
                              const arma::Row<size_t>& responses,
-                             const double lambda = 0);
+                             const double lambda = 1);
```
rcurtin: If we change the default here we should probably document it in `HISTORY.txt`. Any particular reason for the change?

src/mlpack/core/optimizers/trust_region_newton/trust_region_newton_impl.hpp
```diff
@@ -0,0 +1,214 @@
...
+  double rTr = arma::dot(r, r);
+
+  // Whether to optimize until convergence.
+  const bool optimizeUntilConvergence = (maxConjugateIterations == 0);
+
+  const double minGradientNormConjugate = 0.5 * arma::norm(gradient, 2);
+  for (size_t itNum = 0; optimizeUntilConvergence ||
+      (itNum != maxConjugateIterations); ++itNum)
+  {
+    // If ||r^i|| <= e_k||detla f(w^k)|| then output sk =s^(i) and stop.
```
rcurtin: Small typo :)

rcurtin:
We can move this one to ensmallen too.  I think here, maybe we would need to introduce a new `FunctionType` to be optimized (`TwiceDifferentiableFunctionType`?).

PR #1037, opened 2017-06-24
Implementation of parallel SGD

shikharbhardwaj:
Initial ideas for the implementation of parallel SGD based on HOGWILD! in mlpack.

The code is untested and may have a lot of style issues. I'm trying to find reasonable a way to test parallel SGD.


shikharbhardwaj:
I think the last commit fixes the issues with Visual Studio not being able to compile in-class member initializer with braces. I moved the init to the default constructor.

I have added a simple test for parallel SGD right now. Any other ways we could test this?

zoq:
I think it would be neat to test it on the matrix completion method, I don't think it would be feasible to replicate the results from the paper, at least not as a unit test. Perhaps we can create a small synthetic dataset?

shikharbhardwaj:
I was able to run this on the RCV1 dataset for training a sparse SVM classifier. The problem I had was of defining exactly what metric would be a good test.
If we create a synthetic dataset for Matrix Completion, what property would signify the correctness?
 (reduction in the objective function?)
Or do we create a synthetic dataset such that it is guaranteed to optimize to a particular solution, or a particular threshold value of the objective function within a fixed number of iterations?

zoq:
Testing the optimizer on a synthetic dataset with a fixed number of iterations where the resulting objective lies in a reasonable range is definitely something we could do. Perhaps testing it on the maxcut problem is another option.

src/mlpack/core/optimizers/parallel_sgd/parallel_sgd_impl.hpp
```diff
@@ -0,0 +1,141 @@
...
+    overallObjective = 0;
+
+    // Get the stepsize for this iteration
+    double stepSize = decayPolicy.StepSize(i);
+
+    arma::Col<size_t> visitationOrder;
+    GenerateVisitationOrder(visitationOrder);
+
+    #pragma omp parallel
+    {
```
rcurtin: I'm not sure on this one, but is it better to do `#pragma omp parallel` at a higher scope (at the beginning of the function)?  The overhead to start threads can be quite high so I am not sure if this will cause that overhead to happen every iteration.

src/mlpack/core/optimizers/parallel_sgd/parallel_sgd_impl.hpp
```diff
@@ -0,0 +1,141 @@
...
+}
+
+template <typename SparseFunctionType, typename DecayPolicyType>
+void ParallelSGD<
+    SparseFunctionType,
+    DecayPolicyType>::GenerateVisitationOrder(
+        arma::Col<size_t>& visitationOrder)
+{
+  visitationOrder = arma::shuffle(arma::linspace<arma::Col<size_t>>(0,
+        (function.NumFunctions() - 1), function.NumFunctions()));
```
rcurtin: Another idea that might be better for data locality is to split the points that are worked on before shuffling: so thread 0 gets [0, 10), thread 1 gets [10, 20), and so forth.  Then shuffle those individual sets for each thread.  I think this will be faster and won't have a significant effect on convergence, but I'm not 100% certain.

src/mlpack/core/optimizers/parallel_sgd/sparse_svm_function.hpp
```diff
@@ -0,0 +1,55 @@
...
+  SparseSVMLossFunction() = default;
+
+  //! Member initialization constructor.
+  SparseSVMLossFunction(arma::sp_mat& dataset, arma::vec& labels);
+
+  //! Evaluate a function.
+  double Evaluate(const arma::mat& weights, size_t id);
+
+  //! Evaluate the gradient of a function.
+  void Gradient(const arma::mat& weights, size_t id, arma::sp_mat& gradient);
```
rcurtin: If this is all that is necessary for the function that interfaces with Hogwild, do you think we could adapt our existing `LogisticRegression` and `SoftmaxRegression` implementations to handle this?  I think all that would need to change would be that the type of the gradient should be templatized.

src/mlpack/core/optimizers/parallel_sgd/parallel_sgd_impl.hpp
shikharbhardwaj: I placed it inside the loop because the epoch time is expected to be large if the `batchSize` is large(much larger than the time taken to take threads and set up), although I did see that if the `batchSize` is small, each thread is not fully utilised. Also, at the start of each iteration, the threads need to agree upon the stepsize to use in the next iteration. This is discussed in section 5.2 of the paper.

src/mlpack/core/optimizers/parallel_sgd/parallel_sgd_impl.hpp
shikharbhardwaj: I think that would somewhat improve the locality of reference of the memory accesses. But I don't think that our sampling would remain uniform after restricting shuffling in `batchSize` batches.

shikharbhardwaj:
Thanks for the initial thoughts, Ryan.

Regarding the tests, I am currently finding how to generate the synthetic datasets for testing both sparse SVM and Matrix Completion methods.

I'll look into the changes required in `LogisticRegression` and `SoftmaxRegression`.

src/mlpack/core/optimizers/parallel_sgd/parallel_sgd_impl.hpp
rcurtin: Agreed, it would definitely break the sampling. I don't know how much that would matter, but I simply haven't thought about it very much.  Probably better to leave as is.

rcurtin:
Sounds good, looking forward to future improvements. :) Thanks!

src/mlpack/core/optimizers/parallel_sgd/sparse_test_function.hpp
```diff
@@ -0,0 +1,63 @@
...
+// a reasonable number of iterations, each value in the decision variable should
+// be at the vertex of the parabola in that dimension.
+class SparseTestFunction
+{
+ public:
+  //! Set members in the default constructor.
+  SparseTestFunction();
+
+  //! Return 6 (the number of functions).
+  size_t NumFunctions() const { return 4; }
```
shikharbhardwaj: Yes. I was planning to add 6 functions but then decided to do just 4. :)

shikharbhardwaj:
I am trying to find a way to test both the matrix completion and sparse SVM methods. Here is what I have been able to find so far : 

For testing the sparse matrix completion method, I ran it with the same data used for testing the existing `matrix_completion` method in mlpack. The results are here : https://gist.github.com/shikharbhardwaj/91fe7e5984f32ccd95b78c99c9fcbf20, the error is around the order of 1e-3, while the implemented tests have it to be around 1e-5.

I generated some more "synthetic" data to see how the algorithm works. Here are the runs : https://gist.github.com/shikharbhardwaj/e487fff3f8d6d67a8a02c724991797ac.

I am afraid that I have not been able to find a way to test the sparse SVM method. Any inputs are appreciated :)

shikharbhardwaj:
The final commit sets the implementation ready for merge from my side. The one thing remaining is a test for the sparse SVM function. I had tried generating some random sparse data, but the run terminated with 0 RMSE, I don't think it makes for a good test.

The run on RCV1 can be added as a test, but the dataset is ~27MB in Armadillo's binary format, a bit large to include within the repo.

I decided to remove the implementation of the sparse matrix completion loss function as the implementation was quite similar to Regularized SVD.

mentekid:
Thank you Shikhar.

Have you run any benchmarks of HOGWILD! vs "vanilla" SGD? How much boost does plugging in HOGWILD! give us?

I'll play with this more extensively in the weekend and leave you any comments I have.

shikharbhardwaj:
The difference is pretty large I think.

I ran HOGWILD!  with 4 threads and StandardSGD on the RCV1 dataset (the RMSE here is computed over the training set only).

```
# HOGWILD!
λ mlpack_spike/hogwild ∴ time OMP_NUM_THREADS=4 ./a.out
RMSE : 0.797936
Initial loss : 70176.4
Final loss : 114.276
RMSE : 0.0496217
OMP_NUM_THREADS=4 ./a.out  16.12s user 0.18s system 392% cpu 4.155 total

# StandardSGD
λ mlpack_spike/hogwild ∴ time ./a.out
RMSE : 0.797936
Initial loss : 70176.4
Final loss : 5969.06
RMSE : 0.227869
./a.out  7.53s user 1.44s system 102% cpu 8.763 total
```

HOGWILD! finishes 2x faster, with a much better RMSE.

src/mlpack/core/optimizers/parallel_sgd/decay_policies/constant_step.hpp
```diff
@@ -0,0 +1,47 @@
...
+namespace mlpack {
+namespace optimization {
+
+/**
+ * Implementation of the ConstantStep stepsize decay policy for parallel SGD.
+ */
+class ConstantStep
+{
+ public:
+  ConstantStep(double initalStep) : step(initalStep) { /* Nothing to do */ }
```
zoq: Can you add a comment for the constructor and the parameter? Also do you think you could use ``const double initalStep`` here, to be consistent with the rest of the codebase?

Not sure, can we find a reasonable default step value?

src/mlpack/core/optimizers/parallel_sgd/decay_policies/constant_step.hpp
```diff
@@ -0,0 +1,47 @@
...
+
+  /**
+   * This function is called in each iteration before the gradient update.
+   *
+   * @param n_epoch The iteration number for which the stepsize is to be
```
zoq: Can you use camel casing for all names, I know the parameter isn't used, but doxygen should pick it up anyway.

src/mlpack/core/optimizers/parallel_sgd/decay_policies/exponential_backoff.hpp
```diff
@@ -0,0 +1,87 @@
...
+class ExponentialBackoff
+{
+ public:
+  /**
+   * Construct the exponential backoff policy with the required parameters.
+   *
+   * @param firstBackoffEpoch The number of updates to run before the first
+   * stepsize backoff.
+   * @param step The initial stepsize(gamma).
+   * @param beta The reduction factor.
```
zoq: I think it's a good idea to mention that beta has to be in range (0, 1).

src/mlpack/core/optimizers/parallel_sgd/parallel_sgd_impl.hpp
```diff
@@ -0,0 +1,148 @@
...
+      Log::Info << "SGD: minimized within tolerance " << tolerance << "; "
+        << "terminating optimization." << std::endl;
+      return overallObjective;
+    }
+
+    // Get the stepsize for this iteration
+    double stepSize = decayPolicy.StepSize(i);
+
+    arma::Col<size_t> visitationOrder;
+    GenerateVisitationOrder(visitationOrder, function.NumFunctions());
```
zoq: Can we reuse the existing order vector and just shuffel the order in each iterations?

src/mlpack/core/optimizers/parallel_sgd/parallel_sgd_impl.hpp
```diff
@@ -0,0 +1,148 @@
...
+template <typename DecayPolicyType>
+template <typename SparseFunctionType>
+double ParallelSGD<DecayPolicyType>::Optimize(
+    SparseFunctionType& function,
+    arma::mat& iterate)
+{
+  double overallObjective = DBL_MAX;
+  double lastObjective;
+
+  for (size_t i = 1; i != maxIterations; ++i)
```
mentekid: Maybe add a comment mentioning (again) that maxIterations = 0 means repeat until convergence?

src/mlpack/core/optimizers/parallel_sgd/parallel_sgd_impl.hpp
```diff
@@ -0,0 +1,148 @@
...
+    lastObjective = overallObjective;
+    overallObjective = 0;
+
+    for (size_t j = 0; j < function.NumFunctions(); ++j)
+    {
+      overallObjective += function.Evaluate(iterate, j);
+    }
+
+    // Output current objective function.
+    Log::Info << "Parallel SGD: iteration " << i << ", objective "
```
mentekid: I would print this to `Log::Debug` I think, to avoid sending too much information to the screen. However, in SGD, we also print this in `Log::Info`.

src/mlpack/core/optimizers/parallel_sgd/parallel_sgd_impl.hpp
```diff
@@ -0,0 +1,148 @@
...
+      << overallObjective << "." << std::endl;
+
+    if (std::isnan(overallObjective) || std::isinf(overallObjective))
+    {
+      Log::Warn << "Parallel SGD: converged to " << overallObjective
+        << "; terminating" << " with failure.  Try a smaller step size?"
```
mentekid: you could join "terminating" and "with failure [...]" into one string (makes no difference anyway)

src/mlpack/core/optimizers/parallel_sgd/decay_policies/constant_step.hpp
shikharbhardwaj: Ah, right. I missed that.

Regarding the default step size, we can keep something like 0.01 (like SGD).

src/mlpack/core/optimizers/parallel_sgd/sparse_test_function.hpp
```diff
@@ -0,0 +1,63 @@
...
+#define MLPACK_CORE_OPTIMIZERS_PARALLEL_SGD_SPARSE_TEST_FUNCTION_HPP
+
+#include <mlpack/prereqs.hpp>
+
+namespace mlpack {
+namespace optimization {
+namespace test {
+
+// A simple test function. Each dimension has a parabola with a
+// distinct minima. Each update is guaranteed to be sparse(only a single
```
mentekid: pedantic comment - "minimum" I think is more appropriate, since it is singular.

src/mlpack/core/optimizers/parallel_sgd/decay_policies/constant_step.hpp
zoq: Sounds good, we should mention something like: "The defaults here are not necessarily good for the given problem, so it is suggested that the values used be tailored to the task at hand." as done for the SGD class, what do you think?

src/mlpack/core/optimizers/parallel_sgd/parallel_sgd_impl.hpp
```diff
@@ -0,0 +1,148 @@
...
+
+        // Evaluate the sparse gradient.
+        function.Gradient(iterate, instances[j], gradient);
+
+        // Update the decision variable with non-zero components of the
+        // gradient.
+        for (size_t i = 0; i < gradient.n_cols; ++i)
+        {
+          // Iterate over the non-zero elements.
+          for (auto cur = gradient.begin_col(i); cur != gradient.end_col(i);
```
zoq: In general we should avoid ``auto``when using armadillo objects; from the armadillo page:

> Can I use the C++11 auto keyword with Armadillo objects and/or expressions? 
> Use of C++11 auto is not recommended with Armadillo objects and expressions. Armadillo has a template meta-programming framework which creates lots of short lived temporaries that are not handled by auto.
> 

I think in this case it's okay, but it would be easy to use the type of the iterator instead of ``auto``. Let me know what you think.

src/mlpack/core/optimizers/parallel_sgd/decay_policies/constant_step.hpp
shikharbhardwaj: Sure, will do that.

src/mlpack/core/optimizers/parallel_sgd/parallel_sgd_impl.hpp
shikharbhardwaj: I think the reason for using `Log::Info` here is to allow the user to choose to display the output (with the `-v` verbose flag to the executable) without recompiling the library in debug mode.

shikharbhardwaj:
Thanks for the reviews @zoq and @mentekid. I pushed some changes to according to the comments.

The Windows build is failing, I don't think its related to the code.

Is there some way to run the style checks for mlpack locally?

zoq:
About the AppVeyor build: https://appveyor.statuspage.io/incidents/m2vdvw39kdk8, we can restart the build once the issue is resolved.

About the style checks, you can clone the jenkins repository (https://github.com/mlpack/jenkins-conf.git) and run 
```
mkdir reports
./lint.sh --root . --reports reports/cpplint.xml
```
Let us know if we should clarify anything.

src/mlpack/core/optimizers/parallel_sgd/decay_policies/exponential_backoff.hpp
```diff
@@ -0,0 +1,88 @@
...
+   *
+   * @param firstBackoffEpoch The number of updates to run before the first
+   * stepsize backoff.
+   * @param step The initial stepsize(gamma).
+   * @param beta The reduction factor. This should be a value in range (0, 1).
+   */
+  ExponentialBackoff(size_t firstBackoffEpoch, double step, double beta) :
+    firstBackoffEpoch(firstBackoffEpoch), step(step), beta(beta)
+  {
+      cutoffEpoch = firstBackoffEpoch;
```
zoq: Constructor initialization-list evaluation order depends on the order of member variable declaration in the class. In this case ``cutoffEpoch`` is initialized after ``firstBackoffEpoch`` which means we could use the initialization-list  for ``cutoffEpoch`` too. Also, do you mind to use ``const size_t firstBackoffEpoch`` etc. here, to be consistent wit the rest of the codebase?

src/mlpack/core/optimizers/parallel_sgd/decay_policies/exponential_backoff.hpp
```diff
@@ -0,0 +1,88 @@
...
+  }
+  /**
+   * Get the step size for the current gradient update.
+   *
+   * @param numEpoch The iteration number of the current update.
+   * @return The stepsize for the current iteration.
+   */
+  double StepSize(size_t numEpoch)
```
zoq: Pedantic style issue; I would go with ``const size_t numEpoch`` here.

src/mlpack/core/optimizers/parallel_sgd/parallel_sgd_impl.hpp
zoq: I think since, ``Log::Info``is only printed if the user ask for it, it's fine here to use it.

src/mlpack/methods/regularized_svd/regularized_svd_function.hpp
```diff
@@ -19,6 +19,13 @@
 namespace mlpack {
 namespace svd {
 
+/**
+ * The data is stored in a matrix of type MatType, so that this class can be
+ * used with both dense and sparse matrix types.
+ *
+ * @tparam MatType The matrix type of the dataset.
+ */
+template <typename MatType>
```
zoq: I think it's a good idea to set a default parameter here (``arma::mat``).

src/mlpack/methods/sparse_svm/sparse_svm_function.hpp
```diff
@@ -0,0 +1,78 @@
+/**
+ * @file sparse_svm_function.hpp
+ * @author Shikhar Bhardwaj
+ *
+ * Implementation of the hinge loss function for training a sparse SVM with the
+ * parallel SGD algorithm
```
zoq: Pedantic style issue; missing ``.`` at the end.

shikharbhardwaj:
Thanks for the suggestions. I will update the code ASAP.

A few things I noticed while running this algorithm on the Netflix dataset : 

1.  Shuffling `visitationOrder` takes a large amount of time. The number of indices is of the order 1e7 for that problem. `arma::shuffle` creates a copy of the vector, which can be avoided (with `std::shuffle` I think). Can we use any other approach to partition data among the threads? I remember the authors mentioned running the algorithm on a 12 core machine, out of which 2 cores were just used for handing out data to the worker threads.

2. Performing the update as is done currently (iterating over the columns of the gradient and updating non-zero components) is wasteful when the gradient has a very large size(as just running an empty loop is expensive when n ≈ 1e6). The way the existing regularized SVD function solves this is to provide a specialisation for `Optimize` which updates only the needed components. I can't think of a less intrusive way to allow for the function to relay this to the Optimizer, without introducing some extra bookkeeping.

src/mlpack/core/optimizers/parallel_sgd/parallel_sgd_impl.hpp
```diff
@@ -0,0 +1,146 @@
...
+      }
+    }
+  }
+  Log::Info << "\n Parallel SGD terminated with objective : "
+    << overallObjective << std::endl;
+  return overallObjective;
+}
+
+template <typename DecayPolicyType>
+arma::Col<size_t> ParallelSGD<DecayPolicyType>::ThreadShare(
```
zoq: The compiler might optimize this away, but in case if not, we could pass the thread share as a reference. Also do you think we should make shuffel optinal, that way we could entirely avoid the vector copy.

zoq:
About 1. What if we would use ``#pragma omp task`` and run the shuffel process along with the gradient update process.

src/mlpack/core/optimizers/parallel_sgd/parallel_sgd_impl.hpp
shikharbhardwaj: I made a few measurements, and it didn't make a difference using index calculations within the loop, without calling `ThreadShare`. I think this gets optimized away.

A few other measurements I made : https://gist.github.com/shikharbhardwaj/3b43de28e90e1ce9f3a1b2e60d61c7f0.

zoq: Did you use the same seed for each run, e.g. by setting: ``math::RandomSeed(3);``?

shikharbhardwaj: I didn't, but I don't think a different seed would affect runtime, as the number of gradient updates is same in all the runs. I repeated with fixed seed and it was the same.

Also, the randomness in the initial value of the decision variable would be seen as a difference in the initial objective, which is same in all the runs.

src/mlpack/methods/regularized_svd/regularized_svd_function_impl.hpp
```diff
@@ -186,5 +218,109 @@ double StandardSGD::Optimize(
...
+      Log::Info << "SGD: minimized within tolerance " << tolerance << "; "
+        << "terminating optimization." << std::endl;
+      return overallObjective;
+    }
+
+    // Get the stepsize for this iteration
+    double stepSize = decayPolicy.StepSize(i);
+
+    // Shuffle for uniform sampling of functions by each thread.
+    std::random_shuffle(visitationOrder.begin(), visitationOrder.end());
```
zoq: I think if this is faster as ``arma::shuffel``, we should use this anywhere else, do you mind to write some sort of microbenchmark for this, or maybe you already did that?

shikharbhardwaj: I didn't write one separately. But I just wrote one up and it shows ~2.5 x difference : https://gist.github.com/shikharbhardwaj/33daba392167f8752ac4bb9018c02e9d

src/mlpack/core/optimizers/parallel_sgd/parallel_sgd_impl.hpp
```diff
@@ -0,0 +1,125 @@
...
+      Log::Info << "SGD: minimized within tolerance " << tolerance << "; "
+        << "terminating optimization." << std::endl;
+      return overallObjective;
+    }
+
+    // Get the stepsize for this iteration
+    double stepSize = decayPolicy.StepSize(i);
+
+    // Shuffle for uniform sampling of functions by each thread.
+    visitationOrder = arma::shuffle(visitationOrder);
```
shikharbhardwaj: I didn't add one because I initially thought running without shuffling would make no sense. But maybe the user has pre-shuffled data(?) and does not want to incur the cost of a shuffle, while maintaining good locality of reference.

src/mlpack/methods/regularized_svd/regularized_svd_function_impl.hpp
zoq: hm, the difference is still minimal, we are talking about nanoseconds. Just did some experiments using:

```
void ShuffelArma(arma::Col<size_t>& data)
{
  auto start = chrono::steady_clock::now();

  data = std::move(arma::shuffle(data));

  auto end = chrono::steady_clock::now();
  auto diff = end - start;

  typedef std::chrono::duration<float> floatSeconds;
  auto secs = std::chrono::duration_cast<floatSeconds>(diff);

  cout << "arma shuffel: " << chrono::duration <double, milli> (diff).count() << " ms - " << secs.count() << " sec" << endl;
}

void ShuffelStd(arma::Col<size_t>& data)
{
  auto start = chrono::steady_clock::now();

  std::random_shuffle(data.begin(), data.end());

  auto end = chrono::steady_clock::now();
  auto diff = end - start;

  typedef std::chrono::duration<float> floatSeconds;
  auto secs = std::chrono::duration_cast<floatSeconds>(diff);

  cout << "std shuffel: " << chrono::duration <double, milli> (diff).count() << " ms - " << secs.count() << " sec" << endl;
}
```

my results:

```
size: 10
std shuffel: 0.012911 ms - 1.2911e-05 sec
arma shuffel: 0.018432 ms - 1.8432e-05 sec
---------------------------------
size: 100
std shuffel: 0.006969 ms - 6.969e-06 sec
arma shuffel: 0.008546 ms - 8.546e-06 sec
---------------------------------
size: 1000
std shuffel: 0.063325 ms - 6.3325e-05 sec
arma shuffel: 0.071951 ms - 7.1951e-05 sec
---------------------------------
size: 10000
std shuffel: 0.664872 ms - 0.000664872 sec
arma shuffel: 0.751739 ms - 0.000751739 sec
---------------------------------
size: 100000
std shuffel: 6.71557 ms - 0.00671557 sec
arma shuffel: 8.78779 ms - 0.00878779 sec
---------------------------------
size: 1000000
std shuffel: 66.6233 ms - 0.0666233 sec
arma shuffel: 92.9679 ms - 0.0929679 sec
---------------------------------
size: 10000000
std shuffel: 891.552 ms - 0.891552 sec
arma shuffel: 1077.56 ms - 1.07756 sec
---------------------------------
size: 100000000
std shuffel: 10738.9 ms - 10.7389 sec
arma shuffel: 12356.7 ms - 12.3567 sec
---------------------------------
```

So for 100000000 elements the difference is about 1.6 seconds, which I think is still minimal, but I don't mind to use ``std::random_shuffle`` instead of ``arma::shuffel``.

shikharbhardwaj: I agree that the absolute difference begins to make a significant affect only when n ~ 1e8. One more thing `arma::shuffle` does is to create another copy of the matrix, which is not needed.

```
λ mlpack_spike/random ∴ ./a.out
Run on (4 X 3700 MHz CPU s)
2017-07-24 01:42:06
Benchmark             Time           CPU Iterations
---------------------------------------------------
baseline     3614062113 ns 3610419547 ns          1
arma_shuffle 16072122317 ns 16071446490 ns          1
std_shuffle  8582545673 ns 8582178549 ns          1
```

I ran the benchmark I wrote for size = 1e6, and `arma::shuffle` took 3x more time as compared to `std::random_shuffle`, while consuming ~2x more memory (due to the extra copy).

Should I switch back to `arma::shuffle`?

zoq: Right, there is also the memory overhead, I would go with ``std::random_shuffel`` you could also change the shuffel function in the optimzer itself not only in the svd function.

src/mlpack/core/optimizers/parallel_sgd/parallel_sgd_impl.hpp
```diff
@@ -45,6 +45,10 @@ double ParallelSGD<DecayPolicyType>::Optimize(
   arma::Col<size_t> visitationOrder = arma::linspace<arma::Col<size_t>>(0,
       (function.NumFunctions() - 1), function.NumFunctions());
 
+  // A random number generator instance to be used for shuffling the order of
+  // visitation.
+  std::mt19937 gen{ std::random_device()() };
```
zoq: Instead of creating a new generator, we could use the existing: https://github.com/mlpack/mlpack/blob/master/src/mlpack/core/math/random.hpp#L27 That way, we can also use e.g. ``RandomSeed`` to set a static seed for debugging.

mentekid:
I agree, I'll merge it on Friday if there are no more comments.

mentekid:
Merged. Thank you for the great contribution.

@zoq, thanks for the detailed code reviews, I was a bit out of my depth on this one.

Issue #891, opened 2017-02-25
Standardize error handling


rcurtin:
A long time ago, the general rule of thumb (at least that I used) was to use `Log::Fatal` to handle errors, but sometimes e.g. `Log::Debug` was used to just print the error instead and continue going.  But in other places, sometimes exceptions are being thrown (and `Log::Fatal` itself will throw an exception when a newline is encountered).

One of the most frustrating things about using a new library or tool is when it doesn't give good or consistent error handling output, so we should work to ensure that all of mlpack's output is consistent and easy to handle.  We have some nice pieces:

 * A logging system
 * A backtrace system (only available on Linux at the moment though and only in debugging mode)
 * The C++ exception infrastructure

So how should we make something that makes mlpack easy to debug?  Ideally, if a user gets an error, they should be able to quickly understand what went wrong (usually the error will probably be with their data) and maybe get some guidance on how it might be fixed, such as in the situation where the data has invalid entries, or some guidance on what to learn about to debug further, such as in the situation where some algorithm doesn't converge.

For this issue, before we close it, we should converge on some idea of what we should do, then ensure that all of the error messages in mlpack are consistent.  My current idea (which may not be the best idea!) is that we could return an exception that inherits from `std::exception` with a string field containing the backtrace and the explanation, or something like this.  I think we can also modify the backtrace code so that it works even without debugging symbols---it just doesn't have line numbers.  Using that idea instead of `Log::Fatal` offers the advantage that we can be more clear about the error using the type of the exception---`Log::Fatal` only throws `std::runtime_error`.

This is an open-ended ticket, so it is definitely open for discussion.  I've labeled the difficulty "easy", but by this I only mean that it's going to be implementationally easy, not that coming up with the right idea will be easy. :)

zoq:
I like the idea of unifying the error handling and also your idea of using/extending the C++ exception infrastructure. The way I see it, it's basically what Log::Fatal does but extended with some useful information.

IlievskiV:
First of all, the idea of proper exception handling and using the class Exception is a good idea. Usually, it is a well known software engineering practice, to include and plat the exception handling since the very beginnings of the development of the system, because doing that afterward might be difficult. Having said that, if I understood correctly, there is not such a feature in mlpack.

The following image neatly summarizes the exception handling:
[http://tutorials.jenkov.com/images/exception-handling-strategies/overview.png](url)

[...]

PR #1038, opened 2017-06-25
Refactor classifiers to take numClasses parameters

rcurtin:
This is needed for the cross-validation module.  For each of the mlpack classifiers, I have made sure that the API can match the following signature:

```
void Train(const arma::mat& data, const arma::Row<size_t>& labels, const size_t numClasses);
```

and constructors have been updated appropriately also.  The following classes needed changes:

 - `AdaBoost`
 - `DecisionStump`
 - `LogisticRegression` -- this one is special because logistic regression is two-class only, so I just added an extra overload and corresponding documentation
 - `NaiveBayesClassifier`
 - `Perceptron`

@micyril: take a look at this, let me know if this is what you need or if there are any further changes necessary.


src/mlpack/methods/logistic_regression/logistic_regression.hpp
```diff
@@ -53,6 +53,30 @@ class LogisticRegression
...
+   * it requires the extra numClasses parameter (which must always be 2).
+   *
+   * @param predictors Input training variables.
+   * @param responses Outputs resulting from input training variables.
+   * @param numClasses Number of classes in the dataset (must be 2).
+   * @param lambda L2-regularization parameter.
+   */
+  LogisticRegression(const MatType& predictors,
+                     const arma::Row<size_t>& responses,
+                     const size_t numClasses,
```
micyril: I guess my comment about lacking the `numClasses` parameter in `LogisticRegression` was a bit missleading. When we disccussed the possibility to add the `numClasses` parameter in CV constructors, I perceived it as we should always require this parameter if it is present in one of `Train` methods. According to this strategy we will require users to pass `numClasses` when they want to apply cross-validation to LogisticRegression. Do we want it? If no, we can leave `LogisticRegression` without changes.

micyril: The same considerations are applied to `AdaBoost`.

src/mlpack/methods/naive_bayes/naive_bayes_classifier.hpp
```diff
@@ -92,11 +92,14 @@ class NaiveBayesClassifier
...
    *
    * @param data The dataset to train on.
+   * @param labels The labels for the dataset.
+   * @param numClasses The numbe of classes in the dataset.
    * @param incremental Whether or not to use the incremental algorithm for
    *      training.
    */
   void Train(const MatType& data,
              const arma::Row<size_t>& labels,
+             const size_t numClasses,
```
micyril: Is it ok just to add a new parameter to the `Train` method? Shouldn't we deprecate the old method and add a new one?

rcurtin: That should only be necessary if we plan to release another 2.x.x version with this new signature, but since the next release should be 3.0.0, it is ok to break reverse compatibility at this level.

rcurtin:
Hm, ok, I had thought that we should synchronize the `Train()` methods across all classifiers and then the presence of the `numClasses` parameter could be used to determine whether the method was classification or regression.  Did you have something else in mind?  I thought we can enforce that each classifier must implement

```
ClassifierType(const MatType& data, const arma::Row<size_t>& labels, const size_t numClasses);
void ClassifierType::Train(const MatType& data, const arma::Row<size_t>& labels, const size_t numClasses);
```

but then a question also is what to do with the `DatasetInfo` parameter, which is used by some classifiers that can accept categorical data.  We could have other overloads:

```
ClassifierType(const MatType& data, const DatasetInfo& data, const arma::Row<size_t>& labels, const size_t numClasses);
void ClassifierType::Train(const MatType& data, const DatasetInfo& data, const arma::Row<size_t>& labels, const size_t numClasses);
```

Or, we could force the `DatasetInfo` to be accepted as another later optional parameter to the constructor, i.e., `ClassifierType(const MatType& data, const arma::Row<size_t>& labels, const size_t numClasses, const DatasetInfo& datasetInfo = DatasetInfo())`.

Whichever of those we choose, I think it is easiest if you specify the best interface for your work, and then we can apply it through the codebase.  Since the next release will be 3.0.0 and this allows us to break reverse compatibility, this is the easiest time to make these changes. :)

micyril:
We can support cross-validation for `LogisticRegression` and `AdaBoost` without requirement to pass the unnecessary `numClasses` parameter. The changes in `DecisionStump`, `NaiveBayesClassifier` and `Perceptron` are more appropriate since users always need to pass the `numClasses` parameter into constructors of these classes.
Regarding detection of whether a method is classification or regression - the cross-validation module don't need to know this information as long as it knows which parameters it should require to pass  and which ones it can take as optional.

rcurtin:
I see, thanks for the clarification.  Do you have an opinion of how the APIs for classification and regression methods should be structured?  If we can standardize this, we make it easier to implement new classifiers and regressors.

micyril:
I guess we can use LinearRegression as a de facto standard for regression algorithms, and DecisionTree as a de facto standard for classification algorithms.

rcurtin:
I see, the only issue is that these don't account for categorical variables.  I suggest that we do this:

regression:

```
// I guess usually MatType == arma::Mat<eT> and ResponsesType == arma::Col<eT>
RegressionType(const MatType& data, const ResponsesType& responses, /* all optional arguments past here... */);
void RegressionType::Train(const MatType& data, const ResponsesType& responses);

// Sometimes the input could be categorical for some methods too, so these can be optional:
RegressionType(const MatType& data, const DatasetInfo& info, const ResponsesType& responses, /* all optional arguments past here... */);
void RegressionType::Train(const MatType& data, const DatasetInfo& info, const ResponsesType& responses);
```

classification:

```
ClassificationType(const MatType& data, const arma::Row<size_t>& labels, const size_t numClasses, /* all optional arguments past here... */);
void ClassificationType::Train(const MatType& data, const arma::Row<size_t>& labels, const size_t numClasses);

// optional for classes that support categorical data also
ClassificationType(const MatType& data, const DatasetInfo& info, const arma::Row<size_t>& labels, const size_t numClasses, /* all optional arguments past here... */);
void ClassificationType::Train(const MatType& data, const DatasetInfo& info, const arma::Row<size_t>& labels, const size_t numClasses);
```

In these cases it is not required that the overload is exactly implemented as specified, but the compiler should at least be able to instantiate it with `MatType = arma::mat` and `ResponsesType = arma::vec`.  So it is ok if the signature is implemented a little differently.  It would be nice to eventually require that `MatType` can be an Armadillo object of any type, but there is other support that would be needed first.

@zoq: I wouldn't mind your input here too, this is a fairly fundamental design decision so it would be best to get eyes on this before things start changing.

Whenever we decide on something, I will have to create some documentation detailing exactly what is required of a classification and a regression type, a little like http://mlpack.org/docs/mlpack-2.2.3/doxygen.php?doc=trees.html .

micyril:
For completeness we can also specify an optional parameter for weights.

rcurtin:
Ah, right, I forgot about weights.  I've updated my comment.

src/mlpack/methods/logistic_regression/logistic_regression.hpp
rcurtin: Hm, so I spent some time thinking about this point.  I think that in order to preserve compatibility of two-class classifiers with other classifiers, we can simply provide two overloads, one that takes `numClasses` (so that the classifier satisfies the requirements of other classifiers) and one that doesn't, for user convenience.

AdaBoost is a multiclass classifier (or it can be, depending on its weak learner) so I think leaving `numClasses` there is just fine.

Do you think that is ok, or does that cause any problems?

src/mlpack/methods/adaboost/adaboost_main.cpp
```diff
@@ -202,8 +202,10 @@ int main(int argc, char *argv[])
     else if (weakLearner == "perceptron")
       m.WeakLearnerType() = AdaBoostModel::WeakLearnerTypes::PERCEPTRON;
 
+    const size_t numClasses = arma::max(labels) + 1;
```
zoq: Do we require continuous labels starting with 0? If not I think this doesn't return the correct number of labels.

zoq:
As for me, the proposed interface looks good, as long as we provide the same interface for all methods. I'm a little bit worried about the LogisticRegression method since someone could get the impression that the method works with ``numClasses > 2``, but if can make that clear in the documentation this is just a minor issue. I guess we could also through an exception or something like that in the future, once we decided on the error handling: https://github.com/mlpack/mlpack/issues/891.

src/mlpack/methods/adaboost/adaboost_main.cpp
rcurtin: Yes, that is required, but I am not sure how clear the documentation is.  I'll update it.

rcurtin: Ack, sorry, I thought this was in a different place than it is.  In this code the call to `NormalizeLabels()` forces the labels to be between 0 and numClasses.  But I can still make the call a little more efficient, because `numClasses == m.Mappings().n_elem`.

rcurtin:
I updated the documentation somewhat to make it clearer that `LogisticRegression` only supports two classes even though the `numClasses` parameter is sometimes available.  If you think that is clear, should we go ahead and merge this?

zoq:
Completely missed the last message, it's clear for me and I think we waited long enough for any more comments, so let's merge this.

rcurtin:
Agreed, so in this case I will go ahead and merge.  If anyone complains we can revisit the issue then :)

src/mlpack/methods/logistic_regression/logistic_regression.hpp
micyril: Sorry, I missed this reply. The main disadvantage of keeping `numClasses` in `LogisticRegression` is that users will be required to pass it when they use `LogisticRegression` with the cross-validation and hyper-parameter tuning modules. If we want to keep it for the reason described in the documentation "as necessary for the cross-validation code and hyper-parameter tuner to work", it is not actually the case - cross-validation and hyper-parameter tuning will work without it. So, we need to keep it only if there is some other reason(s). Is there any other code that depends on the assumption of the presence of `numClasses` in classifiers?

rcurtin: I see, so are you saying that we can remove the `numClasses` from `LogisticRegression` and other two-class classifiers entirely, and the functionality of CV or the hyper-parameter tuner will not be affected?  If that is true, I guess we can revert the changes to `LogisticRegression`; would you agree with that?

micyril: Cross-validation and hyper-parameter tuning will work in both cases. The only difference is that in one case it will be required to pass `numClasses`, in another - no. I agree that we can revert the changes for `LogisticRegression`.

PR #1040, opened 2017-06-25
Passing const data to DecisionTree constructors

### LINKED ISSUE #1039: Passing const data to DecisionTree constructor
The `DecisionTree` class has several constructors that accept data through universal references.
```
  template<typename MatType, typename LabelsType>
  DecisionTree(MatType&& data,
               LabelsType&& labels,
               const size_t numClasses,
               const size_t minimumLeafSize = 10);
```
Universal references can take the form of const lvalue references, which is good. The problem is that after accepting data the constructors pass it to one of the private `Train` methods, and these `Train `methods require non-const lvalue reference for the `labels` parameter. 
```
  template<bool UseWeights, typename MatType>
  void Train(MatType& data,
             const size_t begin,
             const size_t count,
             arma::Row<size_t>& labels,
             const size_t numClasses,
             arma::rowvec& weights,
             const size_t minimumLeafSize = 10);
```
To fix ability to use DecisionTree with const data we need to remove the const qualifier inside the constructors.

micyril:
Fix #1039

### DIFF
--- a/src/mlpack/methods/decision_tree/decision_tree_impl.hpp
+++ b/src/mlpack/methods/decision_tree/decision_tree_impl.hpp
@@ -72,6 +72,7 @@ DecisionTree<FitnessFunction,
       typename std::remove_reference<MatType>::type>::type;
   using TrueLabelsType = typename std::remove_const<
       typename std::remove_reference<LabelsType>::type>::type;
+
   TrueMatType tmpData(std::forward<MatType>(data));
   TrueLabelsType tmpLabels(std::forward<LabelsType>(labels));
 
--- a/src/mlpack/methods/decision_tree/decision_tree_impl.hpp
+++ b/src/mlpack/methods/decision_tree/decision_tree_impl.hpp
@@ -34,12 +34,12 @@ DecisionTree<FitnessFunction,
                                         const size_t numClasses,
                                         const size_t minimumLeafSize)
 {
-  // Copy or move data.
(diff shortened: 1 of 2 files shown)


micyril:
Ok. My next PR will depend on this patch (I need it to test the cross-validation module with decision trees since DecisionTree has some unique interfaces of Train methods), so I will wait for getting it merged.

src/mlpack/methods/decision_tree/decision_tree_impl.hpp
```diff
@@ -34,10 +34,12 @@ DecisionTree<FitnessFunction,
                                         const size_t numClasses,
                                         const size_t minimumLeafSize)
 {
-  // Copy or move data.
-  typedef typename std::remove_reference<MatType>::type TrueMatType;
-  typedef typename std::remove_reference<LabelsType>::type TrueLabelsType;
+  using TrueMatType = typename std::remove_const<
```
zoq: Could we use ``std::decay`` or maybe it makes sense to define a type alias since the expression is used multiple times? e.g.:

```
template<typename T>
using BaseType = typename std::remove_const<typename std::remove_reference<T>::type>::type;
```

micyril: Sure, a nice suggestion

rcurtin:
Thanks for the nice fix!  I agree that `std::decay<>` is a better way to go; somehow I had not heard of it. :)

PR #1041, opened 2017-06-27
Classic FrankWolfe and OMP

czdiao:
The classic FrankWolfe and OMP is working with this version. They passed simple tests. Please review it.


src/mlpack/core/optimizers/fw/constr_lpball.hpp
```diff
@@ -0,0 +1,104 @@
...
+      {
+      // lp ball with 1<p<inf
+      s = -sign(v) % pow(abs(v), p-1);
+      return;
+      }
+      else if (p == 1.0)
+      {
+      // l1 ball, used in OMP
+      arma::mat tmp = abs(v);
+      arma::uword k = tmp.index_max();  // linear index of matrix
```
zoq: ``index_max()`` was introduced in Version 7.300, mlpack supports Armadillo >= 4.200.0, so either we use another approach e.g. ``arma::as_scalar(arma::find(actionValue.max(), 1))`` or we backport the function. Let me know what you think. If we go with ``arma::find`` we should make sure it's comparable fast.

rcurtin: If it's time to bump up the required Armadillo version somewhat we can do that, although I'd like to make sure that we don't go past 6.500.5 since that's all that's available in Ubuntu 16.04.  The RHEL6 and RHEL7 versions are really old, but... I think that I am the maintainer for those, so technically that is my responsibility that I am failing at...

czdiao: I guess it doesn't matter to me~ Or I could even write a function that go through all the elements in the matrix, and put it somewhere you like ^_^

ShangtongZhang: @rcurtin I think it would be really helpful if we can bump up the required Armadillo version.

rcurtin: @ShangtongZhang: sure---can you propose a version we should jump to, and the functionality that it would get us?  Like I said before, unfortunately I don't think we should go past 6.500.5.

zoq: Since, ``index_min()`` and ``index_max()`` was introduced in 7.300 and ``.index_min()`` and ``.index_max() `` in 7.200, and as @rcurtin pointed out we can't  go past 6.500.5 I'm not sure we can get away without backporting ``index_max``. I don't like the idea to write a separate function, if we could use the existing armadillo feature. Let me know what you all think.

rcurtin: Sure, how about this---we can backport `index_min()` and `index_max()`, and then find the minimum version to use from there.  Here are the API additions that I think are useful inside of mlpack, that could help us set the right minimum version:

 - 5.000: 64-bit indices automatically enabled
 - 6.400: `ind2sub()` and `sub2ind()` (currently backported)
 - 6.500: `conv2()` and other `conv()` improvements (maybe useful for CNNs, not sure)

My thought is, we can go with 6.500.5 in this case, if the `conv()` functions are useful, and 6.400 otherwise, and in either case we can remove the `ind2sub()` backport.  In either case we'll still have to backport `index_min()` and `index_max()` and `.index_min()` and `.index_max()`, but that is not impossible.  If everyone agrees that this is reasonable, I'll open an issue and either I can try and find some time to do it or someone else can if they have time.  (The backporting is typically not hard at all.)

zoq: I think ``conv2`` could be interesting, at least in some cases, but I'm also fine with 6.400.

rcurtin: Ok, let's go with 6.500 then.  I'll open an issue about it.

src/mlpack/core/optimizers/fw/frank_wolfe.hpp
```diff
@@ -0,0 +1,177 @@
...
+   * optimized is stored in update_rule.
+   *
+   * @param function Function to be optimized
+   * @param linear_constr_solver Solver for linear constrained problem.
+   * @param update_rule Rule for updating solution in each iteration.
+   * @param maxIterations Maximum number of iterations allowed (0 means no
+   *     limit).
+   * @param tolerance Maximum absolute tolerance to terminate algorithm.
+   */
+  FrankWolfe(FunctionType& function,
```
rcurtin: #1032, which was recently merged, changes the optimizer API so that the `FunctionType` template parameter is only given to the `Optimize()` function.  Do you think you could make the same change here?  I can clarify what is needed more if you like.

I guess in this situation the constraint domain should also be moved to be input to `Optimize()`.

src/mlpack/core/optimizers/fw/frank_wolfe.hpp
```diff
@@ -0,0 +1,177 @@
...
+
+  //! The maximum number of allowed iterations.
+  size_t maxIterations;
+
+  //! The tolerance for termination.
+  double tolerance;
+};
+
+//! Orthogonal Matching Pursuit
+using OMP = FrankWolfe<FuncSq, ConstrLpBallSolver, UpdateSpan<FuncSq>>;
```
rcurtin: If this is only used in the tests, I think we should avoid giving a typedef here, unless we plan to use it somewhere else in the code.

src/mlpack/core/optimizers/fw/constr_lpball.hpp
```diff
@@ -0,0 +1,104 @@
...
+      tmp = 0 * tmp;
+      tmp(k) = v(k);
+      s = -sign(tmp);
+      return;
```
rcurtin: Could you just do this:

```
arma::mat tmp = arma::abs(v);
arma::uword k = tmp.max_index();
s.zeros(v.n_rows, v.n_cols);
s(k) = sign(tmp(k));
```

I think this could save some computation.

src/mlpack/core/optimizers/fw/constr_lpball.hpp
```diff
@@ -0,0 +1,104 @@
...
+  * Optimizer of Linear Constrained Problem for FrankWolfe.
+  *
+  * @param v Input local gradient.
+  * @param s Output optimal solution in the constrained domain (lp ball).
+  */
+  void Optimize(const arma::mat& v,
+      arma::mat& s)
+  {
+      if (p == -1.0)
+      {
```
rcurtin: It might be more straightforward to use `std::numeric_limits<double>::infty()` to represent that the user wants the l-infinity norm.

src/mlpack/core/optimizers/fw/frank_wolfe_impl.hpp
```diff
@@ -0,0 +1,98 @@
...
+        Log::Info << "FrankWolfe: minimized within tolerance "
+        << tolerance << "; " << "terminating optimization." << std::endl;
+        return CurrentObjective;
+    }
+
+
+    // Update solution, save in iterate_new
+    update_rule.Update(iterate, s, iterate_new, i);
+
+    iterate = iterate_new;
```
rcurtin: I think you could use std::move() here to avoid a copy.

src/mlpack/core/optimizers/fw/update_classic.hpp
```diff
@@ -0,0 +1,88 @@
...
+  }
+
+  //! Get the instantiated function to be optimized.
+  const FunctionType& Function() const { return function; }
+  //! Modify the instantiated function.
+  FunctionType& Function() { return function; }
+
+ private:
+  //! The instantiated function.
+  FunctionType& function;
```
rcurtin: If the function is not used, is there any need for this parameter?

src/mlpack/core/optimizers/fw/update_span.hpp
```diff
@@ -0,0 +1,144 @@
...
+      arma::mat& new_coords,
+      const size_t num_iter)
+  {
+      // add atom here
+      arma::uvec ind = find(s, 1);
+      arma::uword d = ind(0);
+      AddAtom(d);
+
+      arma::vec b = function.Vectorb();
+      arma::mat x = solve(atoms_current, b);
```
rcurtin: This uses `Vectorb()`, which isn't part of the required `FunctionType` interface.  Can this be adapted to remove that requirement?

czdiao: This might be a design problem.

The original FrankWolfe solves "constrained" optimization problem, while this class is supposed to update the solution from the space spanned by all previous solutions, which is unconstrained.  So theoretically it won't use this `UpdateSpan` class. This class is used to solve the unconstrained problem with sparse solution, such as OMP.

So the type of problems that would use this method I currently have in mind are just the atom domain optimization problems, that is, find sparse solution to ||y-Ax||^2. So in vector case, it is just the `FuncSq` class, while I am also implementing a matrix case for matrix completion problem.

In summary, I think we don't need the `FunctionType` interface in `FrankWolfe` class to have `MatrixA()` and `Vectorb()` function, but we can require the `FunctionType` interface in `UpdateSpan` class to have them. I think that will make more sense. I can add them in the documentation.

src/mlpack/core/optimizers/fw/frank_wolfe.hpp
czdiao: OMP seems to be a standard traditional algorithm. The way that I implement it in this FrankWolfe wrapper looks pretty implicit to those people who are not familiar with FW.  I make a typedef here just to remind people how to use it. I don't know if any other people would like to use OMP directly somewhere else. If not, I could just delete it.

src/mlpack/core/optimizers/fw/frank_wolfe.hpp
czdiao: This is a very good idea~ It would be much more reasonable to do it in this way. I will change both classes.

src/mlpack/core/optimizers/fw/constr_lpball.hpp
```diff
@@ -0,0 +1,107 @@
...
+ /**
+  * Optimizer of Linear Constrained Problem for FrankWolfe.
+  *
+  * @param v Input local gradient.
+  * @param s Output optimal solution in the constrained domain (lp ball).
+  */
+  void Optimize(const arma::mat& v,
+      arma::mat& s)
+  {
+      if (p == std::numeric_limits<double>::infinity())
```
zoq: Tabs should be two spaces wide. for more information take a look at: https://github.com/mlpack/mlpack/wiki/DesignGuidelines#tabbing.

src/mlpack/core/optimizers/fw/constr_lpball.hpp
```diff
@@ -0,0 +1,107 @@
...
+      {
+      // l-inf ball
```
zoq: When wrapping a line, the next line should be tabbed twice from where the previous line began, for more information take a look at: https://github.com/mlpack/mlpack/wiki/DesignGuidelines#line-length-and-wrapping.

src/mlpack/core/optimizers/fw/constr_lpball.hpp
```diff
@@ -0,0 +1,107 @@
...
+      s = -sign(v);
+      return;
+      }
+      else if (p > 1.0)
+      {
+      // lp ball with 1<p<inf
```
zoq: Please use complete sentences with proper grammar and punctuation: https://github.com/mlpack/mlpack/wiki/DesignGuidelines#comments in most cases you can just add ``.`` at the end you are fine.

src/mlpack/core/optimizers/fw/constr_lpball.hpp
```diff
@@ -0,0 +1,107 @@
...
+      }
+  }
+
+ private:
+  //! lp norm, take 1<p<inf,
+  // use std::numeric_limits<double>::infinity() for inf norm.
+  double p;
+
+  //! Signum function for double.
+  double sign_double(const double x) const {return (x > 0) - (x < 0);}
```
zoq: Please use camel casing for all names. Also, maybe it makes sense to put the function in ``core/math/lin_alg.hpp`` what do you think? I guess if we go that way, we should use a template for the data type:

```
template <typename T>
T Sign(const T x)
{
    return (T(0) < x) - (x < T(0));
}
```

src/mlpack/core/optimizers/fw/frank_wolfe.hpp
```diff
@@ -0,0 +1,168 @@
...
+   * optimized is stored in update_rule.
+   *
+   * @param linear_constr_solver Solver for linear constrained problem.
+   * @param update_rule Rule for updating solution in each iteration.
+   * @param maxIterations Maximum number of iterations allowed (0 means no
+   *     limit).
+   * @param tolerance Maximum absolute tolerance to terminate algorithm.
+   */
+  FrankWolfe(const LinearConstrSolverType linear_constr_solver,
+        const UpdateRuleType update_rule,
```
zoq: Can you align the upcoming parameter with the first parameter?

src/mlpack/core/optimizers/fw/frank_wolfe.hpp
```diff
@@ -0,0 +1,168 @@
...
+   * at the initialization of linear_constr_solver, the function to be  
+   * optimized is stored in update_rule.
+   *
+   * @param linear_constr_solver Solver for linear constrained problem.
+   * @param update_rule Rule for updating solution in each iteration.
+   * @param maxIterations Maximum number of iterations allowed (0 means no
+   *     limit).
+   * @param tolerance Maximum absolute tolerance to terminate algorithm.
+   */
+  FrankWolfe(const LinearConstrSolverType linear_constr_solver,
```
zoq: Please use camel casing for all names.

src/mlpack/core/optimizers/fw/update_classic.hpp
```diff
@@ -0,0 +1,63 @@
...
+  */
+  template<typename FunctionType>
+  void Update(FunctionType& function,
+      const arma::mat& old_coords,
+      const arma::mat& s,
+      arma::mat& new_coords,
+      const size_t num_iter)
+  {
+      double gamma = 2.0/(num_iter + 2.0);
+      new_coords = (1.0-gamma)*old_coords + gamma*s;
```
zoq: Please use a space between all operations.

src/mlpack/core/optimizers/fw/constr_lpball.hpp
czdiao: It looks like std library doesn't have sign() function. I could put it inside `lin_alg.hpp` if everybody agrees for this option.

Also, thanks for pointing out the coding style problems. I will read the DesignGuideline page gain.

zoq: Putting the ``Sign(...)`` function in ``in_alg.hpp`` sounds resonable for me.

stephentu:
@rcurtin: is this mergable now? what are the outstanding issues that need to be resolved? thanks!

src/mlpack/core/math/lin_alg.hpp
```diff
@@ -124,6 +124,19 @@ inline size_t SvecIndex(size_t i, size_t j, size_t n);
  */
 void SymKronId(const arma::mat& A, arma::mat& op);
 
+/**
+ * Signum function.
+ * Return 1 if x>0; return 0 if x=0; return -1 if x<0.
+ * Return type are the same as input type.
+ *
+ * @param x
```
zoq: Can you write a complete sentence that describes the parameter, that way doxygen can pick that up to create the documentation.

src/mlpack/core/optimizers/fw/frank_wolfe.hpp
```diff
@@ -0,0 +1,166 @@
...
+  double Optimize(FunctionType& function, arma::mat& iterate);
+
+  //! Get the linear constrained solver.
+  LinearConstrSolverType LinearConstrSolver()
+      const { return linearConstrSolver; }
+  //! Modify the linear constrained solver.
+  LinearConstrSolverType& LinearConstrSolver() { return linearConstrSolver; }
+
+  //! Get the update rule.
+  UpdateRuleType UpdateRule() const { return updateRule; }
```
rcurtin: The type here should be `const UpdateRuleType&`, otherwise this will cause a copy.  This applies to each of the other functions of this type, could you fix this please?

src/mlpack/core/optimizers/fw/frank_wolfe.hpp
```diff
@@ -0,0 +1,166 @@
...
+
+  //! The maximum number of allowed iterations.
+  size_t maxIterations;
+
+  //! The tolerance for termination.
+  double tolerance;
+};
+
+//! Orthogonal Matching Pursuit.
+using OMP = FrankWolfe<ConstrLpBallSolver, UpdateSpan>;
```
rcurtin: Could you add a little extra documentation here please on what OMP is, what it's for, etc.?  Nothing too in-depth, just something simple that can point users in the right direction.  Users may come across this in the Doxygen documentation, kind of like this:

http://mlpack.org/docs/mlpack-git-old/doxygen/doxygen.php?doc=namespacemlpack_1_1neighbor.html#ad4e1ff28336994a46a680b750edaf814

so if there's not much information, it will only say `Orthogonal Matching Pursuit` without much further explanation for users to learn what it is about.

src/mlpack/core/optimizers/fw/func_sq.hpp
```diff
@@ -0,0 +1,85 @@
...
+class FuncSq
+{
+ public:
+  /**
+   * Construct the square loss function.
+   *
+   * @param A matrix A.
+   * @param b vector b.
+   */
+  FuncSq(const arma::mat A, const arma::mat b) : A(A), b(b)
```
rcurtin: Do you mean `const arma::mat& A, const arma::mat& b)` here?  Also if b should be a vector, maybe better to use `const arma::vec&`?

czdiao:
@rcurtin Thanks for pointing out the problem. I hope this works now.

src/mlpack/core/optimizers/fw/frank_wolfe.hpp
czdiao: I added a brief document in the new commit. However, if users want to know how to use it, I guess they need to consult the test file. I can make it a method in `src/mlpack/methods` if you think necessary.

src/mlpack/core/optimizers/fw/frank_wolfe.hpp
czdiao: I changed it in the new commit. I hope my C++ knowledge is correct in this change. 😄 I think I did this originally because I learned from the `UpdatePolicyType UpdatePolicy() const { return updatePolicy; }` in `src/mlpack/core/optimizers/sgd/sgd.hpp`. BTW, why is it different from my case?

src/mlpack/core/optimizers/fw/constr_lpball.hpp
```diff
@@ -0,0 +1,102 @@
...
+    else if (p > 1.0)
+    {
+      // lp ball with 1<p<inf.
+      s = -sign(v) % pow(abs(v), p-1);
+    }
+    else if (p == 1.0)
+    {
+      // l1 ball, also used in OMP.
+      arma::mat tmp = arma::abs(v);
+      arma::uword k = tmp.index_max();  // linear index of matrix.
```
zoq: ``.index_max()`` was introduced in armadillo 7.200, but you could use:

```
arma::uword k;
tmp.max(k);
```

czdiao: Thanks for pointing this out. It seems that the documentation page of Armadillo doesn't mention that `.max()` function could be used in this way. Now we don't need `.index_max()` function any more. 👍

stephentu:
@rcurtin these changes look good to me. I will merge this in 3 days!

src/mlpack/core/optimizers/fw/frank_wolfe_impl.hpp
```diff
@@ -0,0 +1,95 @@
...
+  // To keep track of the function value.
+  double currentObjective = function.Evaluate(iterate);
+  double previousObjective = DBL_MAX;
+
+  arma::mat gradient(iterate.n_rows, iterate.n_cols);
+  arma::mat s(iterate.n_rows, iterate.n_cols);
+  arma::mat iterateNew(iterate.n_rows, iterate.n_cols);
+  double gap = 0;
+
+  for (size_t i=1; i != maxIterations; ++i)
```
zoq: Pedantic style issue, please use a space between the operation: ``i = 1;``.

src/mlpack/core/optimizers/fw/frank_wolfe_impl.hpp
```diff
@@ -0,0 +1,95 @@
...
+  {
+    // Output current objective function.
+    Log::Info << "Iteration " << i << ", objective "
+        << currentObjective << "." << std::endl;
+
+    // Reset counter variables.
+    previousObjective = currentObjective;
```
zoq: Looks like this parameter is unused, can we remove it?

src/mlpack/core/optimizers/fw/frank_wolfe_impl.hpp
```diff
@@ -0,0 +1,95 @@
...
+
+    // Calculate the gradient.
+    function.Gradient(iterate, gradient);
+
+    // Solve linear constrained problem, solution saved in s.
+    linearConstrSolver.Optimize(gradient, s);
+
+    // Check duality gap for return condition.
+    gap = std::fabs(dot(iterate-s, gradient));
```
zoq: another pedantic style issue, ``iterate -  s``.

src/mlpack/core/optimizers/fw/test_func_fw.hpp
```diff
@@ -0,0 +1,63 @@
...
+  {/* Nothing to do. */}
+
+  /**
+   * Evaluation of the function.
+   *
+   * @param coords input vector x.
+   */
+  double Evaluate(const arma::mat& coords)
+  {
+    double f = std::pow(coords[0]-0.1, 2);
```
zoq: Can you use a space between the minus operation?

src/mlpack/core/optimizers/fw/update_span.hpp
```diff
@@ -0,0 +1,132 @@
...
+  void Update(FuncSq& function,
+              const arma::mat& oldCoords,
+              const arma::mat& s,
+              arma::mat& newCoords,
+              const size_t numIter)
+  {
+    // Add new atom into soluton space.
+    arma::uvec ind = find(s, 1);
+    arma::uword d = ind(0);
+    AddAtom(function, d);
```
zoq: We could directly pass ind(0) here, just to avoid an extra copy, unless it's used anywhere else.

src/mlpack/core/optimizers/fw/update_span.hpp
```diff
@@ -0,0 +1,132 @@
...
+  //! Current atoms in the solution space, ordered as currentIndices.
+  arma::mat currentAtoms;
+
+  //! Flag current indices is empty.
+  bool isEmpty = true;
+
+  //! Add atom into the solution space.
+  void AddAtom(FuncSq& function, const arma::uword k)
+  {
+    if (isEmpty)
```
zoq: Can we use the armadillo function ``is_empty`` to do the same with the extra ``isEmpty`` parameter?

src/mlpack/core/optimizers/fw/update_span.hpp
```diff
@@ -0,0 +1,132 @@
...
+    }
+  }
+
+  /**
+   * Recover the solution coordinate from the coefficients of current atoms.
+   *
+   * @param n dimension of original solution space.
+   * @param x coefficients of current atoms.
+   */
+  arma::vec RecoverVector(const size_t n, const arma::vec& x)
```
zoq: The compiler might be smart to optimize the copy away, but I was wondering if we should pass ``y`` as a function parameter?

src/mlpack/tests/frankwolfe_test.cpp
```diff
@@ -0,0 +1,77 @@
...
+#include "test_tools.hpp"
+
+using namespace std;
+using namespace arma;
+using namespace mlpack;
+using namespace mlpack::optimization;
+
+BOOST_AUTO_TEST_SUITE(FrankWolfeTest);
+
+BOOST_AUTO_TEST_CASE(OMPTest)
```
zoq: Can you add a comment for each test, a single sentence is often enough, no details needed.

src/mlpack/core/optimizers/fw/frank_wolfe.hpp
rcurtin: I think the `sgd.hpp` implementation was wrong; I fixed it in 726946d21.  If we have `T Method() const` this will return a copy of the `T`, but we don't want that---the goal here is to provide access to the `T` but not a copy, so we need `const T& Method() const`.  If the user wants to change the value of the `T`, that's what the other overload is for.

PR #1046, opened 2017-07-03
SpikeSlabRBM

kris-singh:
1)Right now this implements ssRBM layer types.(ssRBM.hpp is not ready i will finish that by tonight or so)


src/mlpack/methods/rbm/ssRBM.hpp
```diff
@@ -0,0 +1,31 @@
...
+#include <mlpack/core.hpp>
+#include <mlpack/prereqs.hpp>
+
+namespace mlpack{
+namespace rbm{
+  template<typename VisibleLayer, typename HiddenLayer>
+  class ssRBM
+  {
+  public:
+    ssRBM(VisibleLayer visible, HiddenLayer hidden, const size_t numSteps, bool persisitence,)
```
lozhnikov: There is no need to implement a separate class for the ssRBM.  I think you should implement the ssRBM on the basis of the `RBM` class i.e. `RBM<BinaryLayer, BinaryLayer>` represents the binary RBM and `RBM<SpikeSlabLayer, SpikeSlabLayer>` corresponds to the ssRBM.

kris-singh: Yes, i was thinking of implementing this as policy class to the existing RBM class. There would have to a lot refactoring though. We would have to convert the existing RBM impl to the a policy class also. Would that be a good idea.

lozhnikov: Yeah, I think it is a good idea to add a policy class that calculates the gradient and free energy. I think that don't require a lot of refactoring.

kris-singh:
I removed the serialise test because i am not sure how to write the serialisation function with policy based implmentation.

src/mlpack/methods/rbm/CMakeLists.txt
```diff
@@ -0,0 +1,21 @@
+# Define the files we need to compile
+# Anything not in this list will not be compiled into mlpack.
+set(SOURCES
+  binary_layer.hpp
+  binary_layer_impl.hpp
+  binary_rbm.hpp 
+  rbm.hpp
```
lozhnikov: In my opinion it is better to move these files back to `mlpack/methods/ann` since other neural networks are located there.

src/mlpack/methods/rbm/binary_layer.hpp
```diff
@@ -0,0 +1,123 @@
...
+using namespace mlpack::ann;
+
+namespace mlpack{
+namespace rbm{
+
+template <
+    typename InputDataType = arma::mat,
+    typename OutputDataType = arma::mat
+>
+class BinaryLayer
```
lozhnikov: Hmm.. I thought it is better to merge this class and the policy.

src/mlpack/methods/rbm/rbm_impl.hpp
```diff
@@ -0,0 +1,206 @@
...
+    Gradient(arma::mat& /*parameters*/, const size_t input, arma::mat& output)
+{
+  // Collect the negative samples
+  rbmPolicy.PositivePhase(std::move(predictors.col(input)));
+  for (size_t i = 0; i < mSteps; i++)
+  {
+    Gibbs(std::move(predictors.col(input)), std::move(tempNegativeSmaples));
+    negativeSamples.col(i) = tempNegativeSmaples;
+  }
+  rbmPolicy.NegativePhase(std::move(negativeSamples));
```
lozhnikov: As for me that is quite difficult. Why don't call `NegativePhase()` `mSteps` times?

src/mlpack/methods/rbm/rbm_impl.hpp
```diff
@@ -0,0 +1,206 @@
...
+        FreeEnergy(std::move(negativeSamples)));
+  }
+  else
+  {
+    if (persistence)
+    {
+      size_t idx = RandInt(0, predictors.n_rows);
+      corruptInput.row(idx) = 1 - corruptInput.row(idx);
+      return std::log(LogisticFunction::Fn(FreeEnergy(std::move(corruptInput)) -
+          FreeEnergy(std::move(arma::round(predictors.col(i)))))) *
```
lozhnikov: Are you sure that this code is valid for any RBM type e.g. for the ssRBM?

src/mlpack/methods/rbm/rbm_impl.hpp
```diff
@@ -0,0 +1,206 @@
...
+{
+  // Collect the negative samples
+  rbmPolicy.PositivePhase(std::move(predictors.col(input)));
+  for (size_t i = 0; i < mSteps; i++)
+  {
+    Gibbs(std::move(predictors.col(input)), std::move(tempNegativeSmaples));
+    negativeSamples.col(i) = tempNegativeSmaples;
+  }
+  rbmPolicy.NegativePhase(std::move(negativeSamples));
+  output = (negativeGradient - positiveGradient);
```
lozhnikov: Hmm.. I guess it is hard to understand that `NegativePhase()` affects `negativeGradient`. I suggest to add a function to tho policy API e.g. `NegativeGradient()`.

src/mlpack/methods/rbm/spike_slab_layer.hpp
```diff
@@ -0,0 +1,167 @@
...
+  * variance = lambda^-1
+  * We only samples that are in ball of size of R > max_t || v_t ||_2
+  * t indexes over all training samples
+  *
+  * @param slab the slab variable(KxN)
+  * @param hidden the spike variables(1xN)
+  * @param output the sampled visible variable
+  * @param returnMean return the mean of the distribution 
+  */
+  void Pvgivensh(arma::mat&& slab, arma::mat&& hidden, arma::mat&& outMean,
```
lozhnikov: Could you replace the name of the function by something more meaningful?

src/mlpack/methods/rbm/spike_slab_layer_impl.hpp
```diff
@@ -0,0 +1,145 @@
...
+    arma::mat&& hidden, arma::mat&& outMean, arma::mat&& output)
+{
+  variance = arma::diagmat(lambdaBias.col(0)).i();
+  mean = arma::zeros(variance.n_rows, 1);
+  for (size_t i = 0; i < weight.n_slices; i++)
+  {
+    mean += weight.slice(i).t() * slab.col(i) * arma::as_scalar(hidden(i));
+  }
+  mean = variance * mean;
+  GaussianDistribution dist(mean, variance);
```
lozhnikov: I suggest to replace that by `math::RandNormal()` since `GaussianDistribution` is too heavy.

src/mlpack/methods/rbm/spike_slab_layer_impl.hpp
```diff
@@ -0,0 +1,145 @@
...
+  sample = dist.Random();
+  // Rejection sampling
+  while (arma::norm(sample) > radius)
```
lozhnikov: I guess we should restrict the number of iterations.

src/mlpack/methods/rbm/ssRBM.hpp
```diff
@@ -0,0 +1,282 @@
...
+    visible.Sample(std::move(input), std::move(tempHidden));
+
+    visible.Psgivenvh(std::move(input), std::move(tempHidden),
+        std::move(tempMean), std::move(tempSlab));
+
+    // positive weight gradient
+    for (size_t i = 0 ; i < tempHidden.n_rows; i++)
+    {
+      weightPositiveGrad.slice(i) = tempMean * input.t() *
+          arma::as_scalar(tempHidden.row(i));
```
lozhnikov: If I am not mistaken `arma::as_scalar(tempHidden.row(i))` is equal to `tempHidden[i]` since actually `tempHidden` is a vector.

lozhnikov:
Regarding the serialization test:
Add `ar & data::CreateNVP(rbmPolicy, "rbmPolicy");` to `RBM::Serialize()`.

src/mlpack/methods/rbm/rbm_impl.hpp
kris-singh: Well i think the Pseudo likelihood loss should be valid since it is independent of the RBM(depends on cd-k algo). I have to see how to make the cross entropy loss work since ssRBM doesn't have a ForwardFunction.

src/mlpack/methods/rbm/binary_layer.hpp
kris-singh: I don't see a good reason for this. Also Layer is independent of the policy for eg a layer doesn't have access to the hidden layer in anyway. A new policy can have one layer as Binary and other spikeslab layer.

src/mlpack/methods/rbm/rbm_impl.hpp
lozhnikov: Oh, I mean the code uses the assumption that visible variables are binary.
For example `corruptInput.row(idx) = 1 - corruptInput.row(idx)` and `arma::round(predictors.col(i))`.

src/mlpack/methods/rbm/rbm_impl.hpp
kris-singh: Do you mean here to rename the NegativePhase() to NegativeGradient()?

src/mlpack/methods/rbm/spike_slab_layer_impl.hpp
kris-singh: This can't be done since math::RandomNormal only accepts double as parameters. Here both the mean and variance are of arma::mat type.

src/mlpack/methods/rbm/rbm_impl.hpp
kris-singh: True my solution to this is that i will std::is_same<RBMPolicy, BinaryRBM> criteria along with the persistence criteria for testing

src/mlpack/methods/rbm/rbm_impl.hpp
kris-singh: if i call NegativePhase() mSteps number of times. That would mean that i would be storing the negativeGradient in a tempvariable(since negativeGradient is overwritten) and then divide it by mSteps. How is this better than the previous approach.

src/mlpack/methods/rbm/spike_slab_layer_impl.hpp
lozhnikov: I mean it is better to write a loop.

src/mlpack/methods/rbm/rbm_impl.hpp
lozhnikov: No, I  mean a function that returns a locally stored gradient.

src/mlpack/methods/rbm/rbm_impl.hpp
lozhnikov: My approach requires less temporary memory.

src/mlpack/methods/rbm/rbm_impl.hpp
lozhnikov: That's too complex. I prefer to add a rule to the policy class.

src/mlpack/methods/rbm/spike_slab_layer_impl.hpp
```diff
@@ -0,0 +1,149 @@
...
+    hiddenSize = outSize;
+  }
+  else
+  {
+    visibleSize = outSize;
+    hiddenSize = inSize;
+  }
+  // Weight shape = k * d * n
+  weight = arma::cube(weights.memptr(), poolSize,
+      visibleSize, hiddenSize, false,
```
lozhnikov: It seems the shape is incorrect, `weight.slice(i)` should be a `visibleSize * poolSize` matrix. The constructor looks like `cube(const ptr_aux_mem, n_rows, n_cols, n_slices)`.

src/mlpack/methods/rbm/spike_slab_layer_impl.hpp
```diff
@@ -0,0 +1,149 @@
...
+    Pvgivensh(std::move(output), std::move(input), std::move(mean),
+        std::move(output));
+  }
+  else
+  {
+    for (size_t i = 0; i < hiddenSize; i++)
+    {
+      if (output.is_empty())
+        output = arma::zeros(hiddenSize, 1);
+      output(i) = LogisticFunction::Fn(arma::as_scalar(input.t() *
```
lozhnikov: This formula is incorrect. You forgot `1/2` and mixed up `weight.slice(i)` and `weight.slice(i).t()`.

P.S. I forgot that you transposed the weight matrix. It seems the formula is correct except `1/2`.

src/mlpack/methods/rbm/spike_slab_layer_impl.hpp
```diff
@@ -0,0 +1,149 @@
...
+  sample.set_size(variance.n_rows, 1);
+  for (size_t i = 0; i < weight.n_slices; i++)
+  {
+    mean += weight.slice(i).t() * slab.col(i) * arma::as_scalar(hidden(i));
+  }
+  mean = variance * mean;
+  for (size_t i = 0; i < mean.n_rows; i++)
+      sample(i) = math::RandNormal(mean(i), variance(i));
+  // Rejection sampling
+  while (arma::norm(sample) > radius)
```
lozhnikov: I think we have to restrict the number of iterations here.

src/mlpack/methods/rbm/ssRBM.hpp
```diff
@@ -0,0 +1,284 @@
...
+    }
+
+    // positive hidden bias gradient
+    for (size_t i = 0; i < visible.Weight().n_slices; i++)
+      hiddenBiasPositiveGrad.col(i) = -((
+          visible.Weight().slice(i) *input).t() *
+          tempSlab.col(i)) +
+          arma::as_scalar(visible.SpikeBias().col(i));
+    // positive lambda bias
+    visibleBiasPositiveGrad = (0.5 * input * input.t());
```
lozhnikov: `input` is a column. Hence `input * input.t()` is a `n*n` matrix. I think the correct formula is `0.5 * input.t() * input`.

src/mlpack/methods/rbm/ssRBM.hpp
```diff
@@ -0,0 +1,284 @@
...
+    visibleBiasPositiveGrad = visibleBiasPositiveGrad.diag();
```
lozhnikov: `visibleBiasPositiveGrad.diag()` is a `n*n` matrix, but `visibleBiasPositiveGrad` should be a vector.

src/mlpack/methods/rbm/spike_slab_layer_impl.hpp
kris-singh: Well i agree with that. But i think the weight. shape in ann are generally outsize * insize. That's why i followed the convention here.

lozhnikov:
It seems I fixed the spike slab RBM classification test. I decided to merge RBM layers and policies as I mentioned above. And I moved the implementation back to `mlpack/methods/ann` as I mentioned above.

It seems you committed some unnecessary code here 55ab816acc5305ebe7e777d122808ad004269df9 (see methods/ann/ffn.hpp) and here e714528547ed387cf9b9cd4bc4968b65594c4c27 (see tests/CMakeLists.txt) and I was unable to compile the full source code. So, I decided to squash the five latest commits in such a way that the patch below is designed for 256b9827a8b3db00fac94c3d7830c35701db8377 ("Fix RBM Test").

There are still some issues:
1. The ssRBM is too slow, I think it is reasonable to add support for the `float` datatype i.e. add some templates to the policy classes.
2. The `Evaluate()` function needs some refactoring.
3. The binary RBM classification test shows better results than the spike slab RBM classification test.
4. I tried to test the ssRBM on the MNIST dataset yesterday. First, the ssRBM is too slow. Secondly the penalty for the visible variables may become less than zero if the initial penalty is too small.
5. The `Evaluate()` function may return INF or NAN. I think that happens due to computational issues with the ssRBM free energy. I think it is worth to rearrange products and logarithms.

Finally, here [ssrbm-fix.zip](https://github.com/mlpack/mlpack/files/1169715/ssrbm-fix.zip) is the patch (for 256b9827a8b3db00fac94c3d7830c35701db8377).

Here https://github.com/lozhnikov/mlpack/tree/ssRBM is the full source code.

lozhnikov:
1. There are two possibilities (add a template for the type of elements or add a template for the whole input):
For example
`RBMPolicy<ElemType>::SamleVisible(arma::Mat<ElemType>&& input,......)`
or
`RBMPolicy<InputDatatype, AndSoOn>::SamleVisible(InputDatatype&& input,......)`
2. Yes, that happens if `useMonitoringCost=false` and when you are calculating mean squared error.
3. Well, in general `slabPenalty` represents `N` diagonal matrices, where `N` represents the number of hidden units. We can make all diagonal elements equal, in such a way we can represent each matrix as `alpha_i * E`, where `alpha_i` is a scalar and `E` is the identity matrix. Moreover we require that all `alpha_i` are equal. If so, `slabPenalty` is a scalar.

kris-singh:
- [ ] The ssRBM is too slow. The code requires some profiling. I think it is reasonable to add support for the float datatype i.e. add some templates to the policy classes. (In progress)
-[x] The Evaluate() function does an extra Gibbs sampling. We could cache the latest negative samples in order to avoid that. This can only be done for `useMonitoringCost = false` since we need the `gibbsTemporary` for mean squared test. Even for `useMonitoringCost = false` see my comments in the code.
-[] The Evaluate() function needs some refactoring.
 Can you elaborate here ? 
-[] The binary RBM classification test shows better results than the spike slab RBM classification test. 
I think it not safe to compare the two since one works on binary data and other is known to work on the continuous data.
-[] I tried to test the ssRBM on the MNIST dataset yesterday. First, the ssRBM is too slow. Secondly the penalty for the visible variables may become less than zero if the initial penalty is too small.
-[x]The Evaluate() function may return INF or NAN. I think that happens due to computational issues with the ssRBM free energy. I think it is worth to rearrange products and logarithms.
-[x]The serialization test needs some review.
-[x]There is a lack of comments in the code.
-[x]The paper states the expectation of the gradient with respect to Lambda is equal to 1/2 * v^T * v. But this is correct only if Lambda has equal diagonal elements. Is it better to use constant-diagonal (i.e. scalar) Lambdas?
-[x]Is there any sense in slab penalty matrices (slabPenalty) with different diagonal elements? I think this feature will never be used. Is it better to replace slabPenalty by single scalar?
-[x]I think it is worth to add an ability to use an arbitrary norm in the visible sampling algorithm (SpikeSlabRBMPolicy::SampleVisible()) since it is not reasonable to use the euclidean norm for the MNIST dataset for example.

Regarding some extra commits. I can cherry pick the commits later when we decide to merge this.
P.S : Sorry for delay.

src/mlpack/methods/ann/rbm/spike_slab_rbm_policy.hpp
```diff
@@ -0,0 +1,224 @@
...
+   * Sample Visible function sample 
+   * the visible outputs from the normal distribution with
+   * mean \Lambda^{-1} \sum_{i=1}^N W_i * s_i * h_i and 
+   * variance \Lambda^{-1}
+   * 
+   * @param input consists of spike and slab variables
+   * @param output consits of visible layer.
+   * @param norm norm used for rejection sampling
+   */
+  void SampleVisible(arma::mat&& input, arma::mat&& output, size_t norm = 2);
```
lozhnikov: First, armadillo supports `"inf"` and `"-inf"` norms, you can't use them since the datatype is `size_t`. Second, it is better to use the same API for both `BinaryRBMPolicy` and `SpikeSlabRBMPolicy`. Third, did you try other norms? I suggested that but later I realized that some norms cause failures in `SampleVisible()` and the model may fall into instability. So, right now I am not sure that this is really needed. Sorry, that I didn't say that earlier.

src/mlpack/methods/ann/rbm/spike_slab_rbm_policy_impl.hpp
```diff
@@ -0,0 +1,341 @@
...
+  // spike bias shape N * 1
+  spikeBias = arma::mat(parameter.memptr() + weight.n_elem, hiddenSize, 1,
+      false, false);
+
+  // visible penalty 1 * 1 => D * D(when used)
+  visiblePenalty = arma::mat(parameter.memptr() + weight.n_elem +
+      spikeBias.n_elem, 1, 1, false, false);
+
+  diagInvSlabPenalty = 1.0 / slabPenalty * arma::eye(poolSize, poolSize);
+  diagSlabPenalty = slabPenalty * arma::eye(poolSize, poolSize);
```
lozhnikov: Probably this is not the best idea. These variables are of type `arma::mat`, hence the multiplication by a vector requires `poolSize * poolSize` operations. I think we could avoid matrices at all.

src/mlpack/methods/ann/rbm/spike_slab_rbm_policy_impl.hpp
```diff
@@ -0,0 +1,341 @@
...
+{
+  assert(input.n_rows == visibleSize);
+  assert(input.n_cols == 1);
+
+  scalarVisiblePenalty = visiblePenalty(0, 0);
+
+  double freeEnergy = 0.5 * arma::as_scalar(input.t() *
+      scalarVisiblePenalty * input);
+
+  assert(std::isfinite(std::log(2.0 * M_PI / slabPenalty)));
```
lozhnikov: I think this is not needed since `slabPenalty` is a constant. Right now (after I fixed `SampleVisible()` and `MeanVisible()`) I can't reproduce errors with NANs and INFs. So, it seems the error has already been fixed.

src/mlpack/methods/ann/rbm/spike_slab_rbm_policy_impl.hpp
```diff
@@ -0,0 +1,341 @@
...
+
+  // positive weight gradient
+  for (size_t i = 0 ; i < hiddenSize; i++)
+    weightGrad.slice(i) = input * slabMean.col(i).t() * spikeMean(i);
+
+  // positive hidden bias gradient
+  for (size_t i = 0; i < hiddenSize; i++)
+    spikeBiasGrad(i) = spikeMean(i);
+  // positive lambda bias
+  visiblePenaltyGrad = -0.5 * input.t() * input;
```
lozhnikov: The same comment. Did you compare these approaches? Are you sure that the matrix with equal diagonal elements is better? I think this question requires some research.

src/mlpack/methods/ann/rbm_impl.hpp
```diff
@@ -0,0 +1,224 @@
...
+
+template<typename InitializationRuleType, typename RBMPolicy>
+double RBM<InitializationRuleType, RBMPolicy>::Evaluate(
+    const arma::mat& /* parameters*/, const size_t i)
+{
+  if (!useMonitoringCost)
+  {
+    // negative samples used here are from previous iteration
+    if (negativeSamples.is_empty())
+      Gibbs(std::move(predictors.col(i)), std::move(negativeSamples));
```
lozhnikov: The problem is tricky since `MiniBatchSGD` updates parameters before it calls `Evaluate()`. Hence, the last negative samples are outdated. I suggest to leave as is right now, this is not a high priority problem.

src/mlpack/methods/ann/rbm/spike_slab_rbm_policy_impl.hpp
kris-singh: Well, most of the calculations would not feasible if we change the `slabPenalty` to a scalar.
like P(s | v, h). if `slabPenalty` is scalar it would require the mean to scalar as well.
But the mean is given by `spike(i) * slabPenalty * weight.slice(i).t() * visible` here if slabPenalty is a scalar then the mean would not be a scalar or `1*1` matrix

src/mlpack/methods/ann/rbm/spike_slab_rbm_policy_impl.hpp
kris-singh: when you same compare? Means with respect to what time or accuarcy or both ?

lozhnikov:
I tried to investigate why the ssRBM is so slow. `gprof` tells that `arma::eop_core::apply_inplace_plus` wastes 69% of time. I think if we reject `poolSize` at all, we can increase performance significantly. If I remember right the ssRBM paper states that different values of `poolSize` don't give advantages. So, I guess we have to test different values of `poolSize`, maybe it is better to remove the parameter at all.

kris-singh:
Ahhhh seems right. I used `poolSize = 1` when i was testing it for various parameters values and ssRBM was pretty quick.

src/mlpack/methods/ann/rbm/spike_slab_rbm_policy_impl.hpp
lozhnikov: I think you can simplify this expression. Try `arma::cumsum()`.

src/mlpack/methods/ann/rbm/spike_slab_rbm_policy_impl.hpp
lozhnikov: I mean accuracy.

kris-singh:
I tested the above-mentioned points on the cifar dataset with 500 images only. Here is the [gist](https://gist.github.com/kris-singh/8f0336b67df29687831cd697780a5f32)
1. I changed the values of the visibleBias to different diagonal values but did not see much change in accuracy. I think it's better to have only a scalar value for visibleBias since it is easier to provide a intial value to it. The paper says the same.
2. I was able to test slabBias  only  on scalar values.
3. Tested for different pool sizes pool size = 3 accuracy = 79.5 and poolSize = 1 accuracy ~ 75.5%.
Also for 500 images it takes around 3 mintues(Wall Clock Time) on poolSize = 3.

src/mlpack/methods/ann/rbm/spike_slab_rbm_policy_impl.hpp
```diff
@@ -0,0 +1,356 @@
...
+ * @param input the visible layer
+ */ 
+template<typename InputDataType, typename OutputDataType>
+inline double SpikeSlabRBMPolicy<InputDataType, OutputDataType>
+    ::FreeEnergy(InputDataType&& input)
+{
+  assert(input.n_rows == visibleSize);
+  assert(input.n_cols == 1);
+
+  scalarVisiblePenalty = visiblePenalty(0, 0);
```
lozhnikov: It seems you initialize this variable each time when you are going to use that. Thus I guess it is better to define the variable as a local variable i.e. inside each function rather the class.

src/mlpack/methods/ann/rbm/binary_rbm_policy.hpp
```diff
@@ -0,0 +1,162 @@
...
+   */
+  void HiddenPreActivation(InputDataType&& input, OutputDataType&& output);
+
+ private:
+  //! Locally stored number of visible neurons
+  size_t visibleSize;
+  //! Locally stored number of hidden neurons
+  size_t hiddenSize;
+  //! Locally stored  Parameters of the network
+  InputDataType parameter;
```
kris-singh: That is what i also thought until i saw convolution.hpp and convlution_impl.hpp
The weights is defined as OutputDataType and then used weight = arma::cube(weights.memtpr(), .....) where weight is of type arma::cube.

src/mlpack/methods/ann/rbm/binary_rbm_policy.hpp
```diff
@@ -0,0 +1,162 @@
...
+  //! Locally stored weight of the network
+  InputDataType weight;
+  //! Locally stored biases of the visible layer
+  OutputDataType visibleBias;
+  //! Locally stored biases of hidden layer
+  OutputDataType hiddenBias;
+  //! Locally-stored output of the preActivation function used in FreeEnergy
+  InputDataType preActivation;
+  //! Locally-stored corrupInput used for Pseudo-Likelihood
+  InputDataType corruptInput;
```
kris-singh: Well this is a good question. There is actually dosen't make sense to have diffrent input and out datatypes. But the same logic applies for FFN layers as well.

src/mlpack/methods/ann/rbm/binary_rbm_policy_impl.hpp
```diff
@@ -0,0 +1,177 @@
...
+}
+
+// Reset function
+template <typename InputDataType, typename OutputDataType>
+inline void BinaryRBMPolicy<InputDataType, OutputDataType>::Reset()
+{
+  weight = arma::mat(parameter.memptr(), hiddenSize, visibleSize, false, false);
+  hiddenBias = arma::mat(parameter.memptr() + weight.n_elem,
+      hiddenSize, 1, false, false);
+  visibleBias = arma::mat(parameter.memptr() + weight.n_elem +
```
kris-singh: See the above comment

src/mlpack/methods/ann/rbm/spike_slab_rbm_policy_impl.hpp
kris-singh: Yes you have to initialise it again and again since visible Penalty can change. Or i could just reference to visiblePenalty(0,0) variable with scalarVisiblePenalty as double pointer type.

src/mlpack/methods/ann/rbm/binary_rbm_policy.hpp
lozhnikov: hmm... I see. In such a way I think there is an error in `convolution_impl.hpp`. I think the code below (`convolution_impl.hpp:220-230`) shouldn't work properly if `eT` is not equal to `double`.
```
arma::Mat<eT>&& error;
arma::cube mappedError;
mappedError = arma::cube(error.memptr(), outputWidth / padW,
        outputHeight / padH, outSize);
```
@zoq How do you think?

src/mlpack/methods/ann/rbm/binary_rbm_policy.hpp
lozhnikov: Well, `FFN` hasn't got loops, but `RBM` has.

src/mlpack/methods/ann/rbm/spike_slab_rbm_policy_impl.hpp
lozhnikov: As for me it is better to remove `scalarVisiblePenalty` at all.

src/mlpack/methods/ann/rbm/binary_rbm_policy.hpp
zoq: You are absolutely right, the refactoring of the convolution layer is on my todo list, I think there is a lot of room for optimizations.

src/mlpack/methods/ann/rbm/spike_slab_rbm_policy.hpp
```diff
@@ -0,0 +1,223 @@
...
+  
+  /**
+   * Sample Visible function sample 
+   * the visible outputs from the normal distribution with
+   * mean \Lambda^{-1} \sum_{i=1}^N W_i * s_i * h_i and 
+   * variance \Lambda^{-1}
+   * 
+   * @param input consists of spike and slab variables
+   * @param output consits of visible layer.
+   * @param norm norm used for rejection sampling
```
lozhnikov: We rejected `norm`. So, the comment is out of date.

src/mlpack/methods/ann/rbm/spike_slab_rbm_policy.hpp
```diff
@@ -0,0 +1,223 @@
...
+   */
+  void SampleHidden(DataType&& input, DataType&& output);
+
+  // Serialize function
+  template<typename Archive>
+  void Serialize(Archive& ar, const unsigned int /* version */);
+
+  //! Return the initial point for the optimization.
+  const DataType& Parameters() const { return parameter; }
+  //! Modify the initial point for the optimization.
```
lozhnikov: The same comment here. To be precise, this is not an initial point.

src/mlpack/methods/ann/rbm/spike_slab_rbm_policy.hpp
```diff
@@ -0,0 +1,223 @@
...
+  //! Get the regulaliser associated with spike variables
+  DataType const& SpikeBias() const { return spikeBias; }
+  DataType& SpikeBias() { return spikeBias; }
+
+  //! Get the regulaliser associated with slab variables
+  ElemType const& SlabPenalty() const { return invSlabPenalty; }
+
+  //! Get the regulaliser associated with visible variables
+  DataType const& VisiblePenalty() const { return visiblePenalty; }
+  DataType& VisiblePenalty() { return visiblePenalty; }
```
lozhnikov: Some functions above haven't got a description.

src/mlpack/methods/ann/rbm/spike_slab_rbm_policy.hpp
```diff
@@ -0,0 +1,223 @@
...
+   * 
+   * @param visible the visible layer units
+   * @param spike the spike variables from hidden layer
+   * @param slabMean the mean of the normal distribution
+   */
+  void SlabMean(DataType&& visible, DataType&& spike,
+      DataType&& slabMean);
+  /**
+   * SampleSlab function calculates the
+   * normal distribution P(s|v,h).
```
lozhnikov: The function doesn't calculate the distribution. The function samples slab variables.

src/mlpack/methods/ann/rbm/spike_slab_rbm_policy.hpp
```diff
@@ -0,0 +1,223 @@
...
+   * @slabMean mean of the normal distribution
+   * @slab sample slab variable from the normal distribution
+   */
+  void SampleSlab(DataType&& slabMean, DataType&& slab);
+
+  //! Locally stored parameters number of visible neurons
+  size_t visibleSize;
+  //! Locally stored parameters number of hidden neurons
+  size_t hiddenSize;
+  //! Locally stored parameters poolSize
```
lozhnikov: Did you forget to remove `parameters` in the comments above?

src/mlpack/methods/ann/rbm/spike_slab_rbm_policy.hpp
```diff
@@ -0,0 +1,223 @@
...
+  //! Locally stored slabPenalty
+  ElemType slabPenalty;
+  //! Locally stored iverse of slabPenalty
+  ElemType invSlabPenalty;
+  //! Locally stored radius used for rejection sampling
+  ElemType radius;
+  //! Locally stored visible Penalty(1 * 1)
+  DataType visiblePenalty;
+  //! Locally stored saclar visible Penalty
+  ElemType scalarVisiblePenalty;
```
lozhnikov: This variable is not used anymore.

src/mlpack/methods/ann/rbm/spike_slab_rbm_policy_impl.hpp
```diff
@@ -0,0 +1,343 @@
...
+  slabMean.set_size(poolSize, hiddenSize);
+  invSlabPenalty = 1.0 / slabPenalty;
+};
+
+// Reset function
+template<typename DataType>
+inline void SpikeSlabRBMPolicy<DataType>::Reset()
+{
+  // Weight shape D * K * N
+  weight = arma::cube(parameter.memptr(), visibleSize, poolSize, hiddenSize,
```
lozhnikov: The constructor is incorrect. We have already discussed that.

src/mlpack/methods/ann/rbm/spike_slab_rbm_policy_impl.hpp
```diff
@@ -0,0 +1,343 @@
...
+ * @param input the visible input
+ * @param output the computed gradient
+ */
+template<typename DataType>
+inline void SpikeSlabRBMPolicy<DataType>
+    ::PositivePhase(DataType&& input, DataType&& gradient)
+{
+  assert(input.n_rows == visibleSize);
+  assert(input.n_cols == 1);
+  arma::cube weightGrad = arma::cube(gradient.memptr(), visibleSize, poolSize,
```
lozhnikov: The same comment here.

src/mlpack/methods/ann/rbm_impl.hpp
```diff
@@ -0,0 +1,222 @@
...
+  ar & data::CreateNVP(numSteps, "numSteps");
+  ar & data::CreateNVP(mSteps, "mSteps");
+  ar & data::CreateNVP(useMonitoringCost, "useMonitoringCost");
+  ar & data::CreateNVP(persistence, "persistence");
+  ar & data::CreateNVP(rbmPolicy, "rbmPolicy");
+
+  // If we are loading, we need to initialize the weights.
+  if (Archive::is_loading::value)
+  {
+  size_t weight = rbmPolicy.Parameters().n_elem;
```
lozhnikov: Could you fix spaces here?

src/mlpack/methods/ann/rbm/spike_slab_rbm_policy.hpp
```diff
@@ -0,0 +1,223 @@
...
+  //! Locally stored parameters
+  DataType parameter;
+  //! Locally stored weight of the network (visibleSize * poolSize * hiddenSize)
+  arma::cube weight;
+  //! Locally stored spikeBias (hiddenSize * 1)
+  DataType spikeBias;
+  //! Locally stored slabPenalty
+  ElemType slabPenalty;
+  //! Locally stored iverse of slabPenalty
+  ElemType invSlabPenalty;
```
lozhnikov: It seems you did an optimization here. But I think the compiler is clever enough, so we can remove `invSlabPenalty` at all in order to improve readability.

src/mlpack/methods/ann/rbm/spike_slab_rbm_policy_impl.hpp
```diff
@@ -0,0 +1,343 @@
...
+
+    freeEnergy -= SoftplusFunction::Fn(spikeBias(i) - sum);
+  }
+
+  return freeEnergy;
+}
+
+template<typename DataType>
+inline typename SpikeSlabRBMPolicy<DataType>::ElemType
+SpikeSlabRBMPolicy<DataType>::SpikeSlabRBMPolicy<DataType>
```
lozhnikov: That's why the Travis build errored.
`SpikeSlabRBMPolicy<DataType>::SpikeSlabRBMPolicy<DataType>::Evaluate`

lozhnikov:
I think it is reasonable to add some tests for the `float` datatype.

kris-singh:
Should i add the test for cifar images or should we should only test this for mnist. 
Is there any thing remaining for this PR.

kris-singh:
Here is patch against the latest build 
[error.patch.zip](https://github.com/mlpack/mlpack/files/1204625/error.patch.zip)

kris-singh:
Should I add a test for float data type that checks the classification accuracy with float type.

lozhnikov:
> Should I add a test for float data type that checks the classification accuracy with float type.

Looks reasonable to me.

> Should i add the test for cifar images or should we should only test this for mnist.

I don't like this idea since it requires to add new dataset to the repo.

src/mlpack/methods/ann/rbm.hpp
```diff
@@ -0,0 +1,183 @@
...
+  using NetworkType = RBM<InitializationRuleType, RBMPolicy>;
+  typedef typename RBMPolicy::ElemType eT;
+
+  /* 
+   * Intalise all the parameters of the network
+   * using the intialise rule. 
+   *
+   * @tparam IntialiserType rule to intialise the parameters of the network
+   * @param predictors training data
+   * @param numSteps Number of gibbs steps sampling
```
lozhnikov: You didn't describe `mSteps`.

src/mlpack/methods/ann/rbm.hpp
```diff
@@ -0,0 +1,183 @@
...
+      const bool persistence = false);
+
+  // Reset the network
+  void Reset();
+
+  /* 
+   * Train the network using the Opitimzer with given set of args.
+   * the optimiser sets the parameters of the network for providing
+   * most likely parameters given the inputs
+   * @param: predictors data points
```
lozhnikov: Maybe that's not an issue but I am not sure that doxygen processes the colon.
```@param:```

File: src/mlpack/methods/ann/rbm.hpp [hunk 38]
```diff
@@ -0,0 +1,183 @@
...
+  *
+  * @param parameters Matrix model parameters.
+  * @param i Index of point to use for objective function evaluation.
+  */
+  double Evaluate(const arma::Mat<eT>& parameters, const size_t i);
+
+ /** 
+  * This function calculates
+  * the free energy of the model
+  * @param: input data point
```
lozhnikov: The colon is not needed.

src/mlpack/methods/ann/rbm.hpp
```diff
@@ -0,0 +1,183 @@
...
+  */
+  void SampleVisible(arma::Mat<eT>&& input, arma::Mat<eT>&& output);
+
+ /*
+  * This function does the k-step
+  * gibbs sampling.
+  *
+  * @param input: input to the gibbs function
+  * @param output: stores the negative sample
+  * @param steps: number of gibbs sampling steps
```
lozhnikov: Again, the colons are not needed. Maybe doxygen processes that correctly but I am not sure. Next time, I'll not comment that. So, make sure that doxygen works correctly or remove all unnecessary colons.

src/mlpack/methods/ann/rbm/binary_rbm_policy_impl.hpp
```diff
@@ -0,0 +1,159 @@
...
+#include <mlpack/core.hpp>
+#include <mlpack/prereqs.hpp>
+
+#include "binary_rbm_policy.hpp"
+
+namespace mlpack {
+namespace ann {
+
+template<typename DataType>
+BinaryRBMPolicy<DataType>
```
lozhnikov: Correct if I am mistaken, but it seems according to the style guide we should write
`BinaryRBMPolicy<DataType>::BinaryRBMPolicy(`
without line-breaks.

src/mlpack/methods/ann/rbm/binary_rbm_policy_impl.hpp
```diff
@@ -0,0 +1,159 @@
...
+template<typename DataType>
+void BinaryRBMPolicy<DataType>
+    ::VisiblePreActivation(DataType&& input, DataType&& output)
+{
+  output = weight.t() * input + visibleBias;
+}
+
+template<typename DataType>
+void BinaryRBMPolicy<DataType>
+    ::HiddenPreActivation(DataType&& input, DataType&& output)
```
lozhnikov: Maybe I am mistaken, but it seems the style guide suggested to write code without line-breaks between
class names and `::`.

src/mlpack/methods/ann/rbm/spike_slab_rbm_policy.hpp
```diff
@@ -0,0 +1,222 @@
...
+#include <mlpack/methods/ann/activation_functions/logistic_function.hpp>
+
+namespace mlpack {
+namespace ann {
+template<typename DataType = arma::mat>
+class SpikeSlabRBMPolicy
+{
+ public:
+  typedef typename DataType::elem_type ElemType;
+  // Intialise the visible and hidden layer of the network
```
lozhnikov: You forgot to describe the parameters.

src/mlpack/methods/ann/rbm/spike_slab_rbm_policy_impl.hpp
```diff
@@ -0,0 +1,332 @@
...
+  {
+    for (size_t i = 0; i < visibleSize; i++)
+    {
+      assert(visiblePenalty(0) > 0);
+      output(i) = math::RandNormal(visibleMean(i), 1.0 / visiblePenalty(0));
+    }
+    if (arma::norm(output, 2) < radius)
+      break;
+  }
+  if (arma::norm(output, 2) > radius)
```
lozhnikov: I am not sure that compiler optimizes this. As for me it is better to declare `k` as an outer variable (outside tho loop) and then compare `k` and `numMaxTrials`.

src/mlpack/tests/rbm_network_test.cpp
```diff
@@ -0,0 +1,265 @@
...
+   testLabels);
+  std::cout << "Softmax Accuracy = " << classificationAccuray << std::endl;
+
+  L_BFGS optimizer1(numBasis, numIterations);
+  SoftmaxRegression regressor1(XRbm, trainLabels, numClasses,
+        0.001, false, optimizer1);
+  double classificationAccuray1 = regressor1.ComputeAccuracy(YRbm, testLabels);
+  std::cout << "ssRBM Accuracy = " <<classificationAccuray1 << std::endl;
+
+  BOOST_REQUIRE_GE(classificationAccuray1, 70);
```
lozhnikov: Why did you remove `BOOST_REQUIRE_GE(classificationAccuray1, classificationAccuray);`?

src/mlpack/tests/rbm_network_test.cpp
```diff
@@ -0,0 +1,265 @@
...
+      slabPenalty, radius);
+  RBM<GaussianInitialization, SpikeSlabRBMPolicy<>> modelssRBM(trainData,
+      gaussian, ss_rbm, 1, 1, true, false);
+
+  size_t numRBMIterations = trainData.n_cols * numEpoches;
+  numRBMIterations /= batchSize;
+
+  MiniBatchSGD msgd(batchSize, 0.06, numRBMIterations, 0, true);
+  modelssRBM.Reset();
+  modelssRBM.Policy().VisiblePenalty().fill(0);
```
lozhnikov: hmm.. The penalty should be greater than zero.

src/mlpack/tests/rbm_network_test.cpp
kris-singh: I have changed the poolSize = 1. Since we want it to fast rather than accurate since test for mnist are some what meaningless. poolSize = 3 actually takes a lot of time.

src/mlpack/methods/ann/rbm/binary_rbm_policy_impl.hpp
kris-singh: Actually no. The design guide lines https://github.com/mlpack/mlpack/wiki/DesignGuidelines under method declaration on point 3 states that this is allowed. Still, i will correct all the possible errors.

src/mlpack/tests/rbm_network_test.cpp
lozhnikov: hmm.. I disagree since the version that I sent you (https://github.com/lozhnikov/mlpack/tree/ssRBM) shows the accuracy around 80% with `poolSize=1`. It seems we didn't change the algorithm. Moreover, if the ssRBM accuracy is less than the regression accuracy then the ssRBM doesn't work.

src/mlpack/methods/ann/rbm_impl.hpp
```diff
@@ -0,0 +1,229 @@
...
+
+template<typename InitializationRuleType, typename RBMPolicy>
+double RBM<InitializationRuleType, RBMPolicy>::Evaluate(
+    const arma::Mat<eT>& /* parameters*/,
+    const size_t i)
+{
+  if (!useMonitoringCost)
+  {
+    Gibbs(std::move(predictors.col(i)), std::move(negativeSamples));
+    return (FreeEnergy(std::move(predictors.col(i))) -
```
lozhnikov: I suggest to add `std::fabs` here i.e.
```std::fabs(FreeEnergy(std::move(predictors.col(i))) -FreeEnergy(std::move(negativeSamples)));```

File: src/mlpack/methods/ann/rbm.hpp [hunk 47]
```diff
@@ -0,0 +1,183 @@
...
+ * mlpack is free software; you may redistribute it and/or modify it under the
+ * terms of the 3-clause BSD license.  You should have received a copy of the
+ * 3-clause BSD license along with mlpack.  If not, see
+ * http://www.opensource.org/licenses/BSD-3-Clause for more information.
+ */
+#ifndef MLPACK_METHODS_ANN_RBM_HPP
+#define MLPACK_METHODS_ANN_RBM_HPP
+
+#include <mlpack/core.hpp>
+#include <mlpack/prereqs.hpp>
```
zoq: ``core.hpp`` already includes ``prereqs.hpp``, so we can remove the header here.

src/mlpack/methods/ann/rbm.hpp
```diff
@@ -0,0 +1,183 @@
...
+#include <mlpack/core/math/random.hpp>
```
zoq: Same as for ``prereqs.hpp``, ``core.hpp`` already includes ``random.hpp``.

src/mlpack/methods/ann/rbm.hpp
```diff
@@ -0,0 +1,183 @@
...
+
+#include <mlpack/methods/ann/activation_functions/softplus_function.hpp>
+#include <mlpack/methods/ann/init_rules/gaussian_init.hpp>
+#include <mlpack/methods/ann/init_rules/random_init.hpp>
```
zoq: Looks like this can be removed, since it's unused, maybe I missed something?

src/mlpack/methods/ann/rbm.hpp
```diff
@@ -0,0 +1,183 @@
...
+
+namespace mlpack {
+namespace ann /** Artificial neural networks.  */ {
+
+template<typename InitializationRuleType, typename RBMPolicy>
```
zoq: Can you comment on both template parameter, using ``@tparam ...``?

src/mlpack/methods/ann/rbm.hpp
```diff
@@ -0,0 +1,183 @@
...
+class RBM
+{
+ public:
+  using NetworkType = RBM<InitializationRuleType, RBMPolicy>;
+  typedef typename RBMPolicy::ElemType eT;
```
zoq: I would go with ``typedef typename VecType::elem_type ElemType;``to be consistent with the rest of the codebase, for more information take a look at: https://github.com/mlpack/mlpack/blob/master/doc/policies/elemtype.hpp

src/mlpack/methods/ann/rbm.hpp
```diff
@@ -0,0 +1,183 @@
...
+   * Intalise all the parameters of the network
+   * using the intialise rule. 
+   *
+   * @tparam IntialiserType rule to intialise the parameters of the network
+   * @param predictors training data
+   * @param numSteps Number of gibbs steps sampling
+   * @param useMonitoringCost evaluation function to use
+   * @param persistence indicates to use persistent CD
+   */
+  RBM(arma::Mat<eT> predictors, InitializationRuleType initializeRule,
```
zoq: Can you put the ``InitializationRuleType`` on a new line?

src/mlpack/methods/ann/rbm.hpp
zoq: Also can you use  proper grammar and punctuation, for more information take a look at: https://github.com/mlpack/mlpack/wiki/DesignGuidelines#comments in most cases this means adding ``.`` at the end of the line and to start with an upper-case letter.

src/mlpack/methods/ann/rbm.hpp
```diff
@@ -0,0 +1,183 @@
...
+
+ /*
+  * This function does the k-step
+  * gibbs sampling.
+  *
+  * @param input: input to the gibbs function
+  * @param output: stores the negative sample
+  * @param steps: number of gibbs sampling steps
+  */
+  void Gibbs(arma::Mat<eT>&& input, arma::Mat<eT>&& output,
```
zoq: Can you align the parameter with the rest and use a new line for the output parameter, if the line ends up being longer than 80 characters.

src/mlpack/methods/ann/rbm/binary_rbm_policy.hpp
```diff
@@ -0,0 +1,177 @@
...
+#define MLPACK_METHODS_ANN_RBM_BINARY_RBM_POLICY_HPP
+
+#include <mlpack/core.hpp>
+#include <mlpack/prereqs.hpp>
+#include <mlpack/methods/ann/activation_functions/softplus_function.hpp>
+#include <mlpack/methods/ann/activation_functions/logistic_function.hpp>
+
+namespace mlpack{
+namespace ann{
+template <typename DataType = arma::mat>
```
zoq: I think we should add a line between the namespace and the class template to improve the readability, what do you think?

src/mlpack/methods/ann/rbm/binary_rbm_policy.hpp
```diff
@@ -0,0 +1,160 @@
...
+class BinaryRBMPolicy
+{
+ public:
+  // Intialise the visible and hiddenl layer of the network
+  BinaryRBMPolicy(size_t visibleSize, size_t hiddenSize);
```
zoq: Maybe we can use ``const`` here to be consistent with the rest of the codebase?

src/mlpack/methods/ann/rbm/binary_rbm_policy_impl.hpp
```diff
@@ -0,0 +1,159 @@
...
+  visibleBias = DataType(parameter.memptr() + weight.n_elem +
+      hiddenBias.n_elem , visibleSize, 1, false, false);
+}
+
+template<typename DataType>
+typename BinaryRBMPolicy<DataType>::ElemType BinaryRBMPolicy<DataType>
+    ::FreeEnergy(DataType&& input)
+{
+  HiddenPreActivation(std::move(input), std::move(preActivation));
+  SoftplusFunction::Fn(preActivation, preActivation);
```
zoq: Maybe it makes sense to revisit the Softplus function and to use arma::trunc_exp instead of testing for ``< DBL_MAX``, what do you think?

src/mlpack/methods/ann/rbm/spike_slab_rbm_policy_impl.hpp
```diff
@@ -0,0 +1,332 @@
...
+      visibleSize, poolSize, hiddenSize,
+      false, false);
+
+  // spike bias shape N * 1
+  spikeBias = DataType(parameter.memptr() + weight.n_elem, hiddenSize, 1,
+      false, false);
+
+  // visible penalty 1 * 1 => D * D(when used)
+  visiblePenalty = DataType(parameter.memptr() + weight.n_elem +
+      spikeBias.n_elem, 1, 1, false, false);
```
zoq: Should we use ``visibleSize`` size here, since that's what is used here:

```
parameter.set_size(visibleSize * hiddenSize * poolSize + visibleSize + hiddenSize);
```

src/mlpack/methods/ann/rbm/spike_slab_rbm_policy_impl.hpp
```diff
@@ -0,0 +1,332 @@
...
+  freeEnergy -= 0.5 * hiddenSize * poolSize *
+      std::log((2.0 * M_PI) / slabPenalty);
+
+  for (size_t i = 0; i < hiddenSize; i++)
+  {
+    ElemType sum = 0;
+
+    for (size_t k = 0; k < poolSize; k++)
+    {
+      sum += arma::as_scalar(input.t() * weight.slice(i).col(k)) *
```
zoq: I was think we could use ``arma::pow`` here to compress the line, what do you think?

src/mlpack/methods/ann/rbm/spike_slab_rbm_policy_impl.hpp
```diff
@@ -0,0 +1,332 @@
...
+  SampleSpike(std::move(spikeMean), std::move(spikeSamples));
+  SlabMean(std::move(input), std::move(spikeSamples), std::move(slabMean));
+
+  // positive weight gradient
+  for (size_t i = 0 ; i < hiddenSize; i++)
+    weightGrad.slice(i) = input * slabMean.col(i).t() * spikeMean(i);
+
+  // positive hidden bias gradient
+  for (size_t i = 0; i < hiddenSize; i++)
+    spikeBiasGrad(i) = spikeMean(i);
```
zoq: Do we need the for loop here, looks like we could just do ``spikeBiasGrad = spikeMean`` since both is already of size ``hiddenSize``; maybe I missed something?

src/mlpack/methods/ann/rbm/spike_slab_rbm_policy_impl.hpp
kris-singh: `visibleSize` indicated the number of neurons in the visible layer.
Though the `visiblePenalty` we are using is 1 * 1 scalar. This has to be actually D*D diagonal matrix but we reduced it to 1*1 scalar since for better performance we end using scalar * Identity matrix for `visiblePenalty`

src/mlpack/methods/ann/rbm/binary_rbm_policy_impl.hpp
kris-singh: Can you explain this in more detail ?

zoq: it might be faster to use ``arma::trunc_exp`` instead of checking using

```c++
if (x < DBL_MAX)
      return x > -DBL_MAX ? std::log(1 + std::exp(x)) : 0;
    return 1.0;
```

which performs a check for each value, the performance boost might be neglectable but I think it's worth a check.

kris-singh:
Fixed most of the changes. Let me know if some other changes are needed.

kris-singh:
Don't understand why Travis second test case is failing.

zoq:
About the travis error, it looks like the time to run the RBM test is pretty high, do you think we can reduce the time shomehow?

src/mlpack/methods/ann/rbm/spike_slab_rbm_policy_impl.hpp
```diff
@@ -0,0 +1,344 @@
...
+    {
+      output(i) = math::RandNormal(visibleMean(i), 1.0 / visiblePenalty(0));
+    }
+    if (arma::norm(output, 2) < radius)
+      break;
+  }
+
+  if (k == numMaxTrials)
+  {
+    Log::Warn << "Outputs are still not in visible unit"
```
zoq: Missing space after ``unit`` and before ``terminating`` also no need to align the streams here (tab twice is fine).

src/mlpack/methods/ann/rbm_impl.hpp
```diff
@@ -0,0 +1,228 @@
...
+void RBM<InitializationRuleType, RBMPolicy>::Reset()
+{
+  size_t weight = rbmPolicy.Parameters().n_elem;
+
+  positiveGradient.set_size(weight, 1);
+  negativeGradient.set_size(weight, 1);
+  negativeSamples.set_size(rbmPolicy.VisibleSize(), 1);
+  tempNegativeGradient.set_size(weight, 1);
+
+  parameter.set_size(weight, 1);
```
zoq: You could just use ``.zeros(...)`` here no need to set the size first.

src/mlpack/tests/rbm_network_test.cpp
```diff
@@ -0,0 +1,265 @@
...
+  XRbm.zeros();
+  YRbm.zeros();
+  double slabPenalty = 5;
+
+  SpikeSlabRBMPolicy<> ss_rbm(trainData.n_rows, hiddenLayerSize, poolSize,
+      slabPenalty, radius);
+  RBM<GaussianInitialization, SpikeSlabRBMPolicy<>> modelssRBM(trainData,
+      gaussian, ss_rbm, 1, 1, true, false);
+
+  size_t numRBMIterations = trainData.n_cols * numEpoches;
```
zoq: Maybe we decrease the number of epochs to decrease the test time?

src/mlpack/tests/rbm_network_test.cpp
```diff
@@ -0,0 +1,265 @@
...
+  double classificationAccuray1 = regressor1.ComputeAccuracy(YRbm, testLabels);
+  std::cout << "RBM Accuracy = " <<classificationAccuray1 << std::endl;
+  BOOST_REQUIRE_GE(classificationAccuray1, classificationAccuray);
+}
+
+BOOST_AUTO_TEST_CASE(ssRBMClassificationTest)
+{
+  size_t batchSize = 10;
+  size_t numEpoches = 30;
+  int hiddenLayerSize = 100;
```
zoq: Can we decrease the hidden layer size, to reduce the test time?

src/mlpack/tests/serialization_test.cpp
```diff
@@ -1690,4 +1695,95 @@ BOOST_AUTO_TEST_CASE(HoeffdingTreeTest)
   }
 }
 
+BOOST_AUTO_TEST_CASE(BinaryRBMTest)
```
zoq: Can you add a comment and describe what you test here?

lozhnikov:
@zoq That's a known issue. Right now the ssRBM is ten times slower than the binary RBM. I tried to investigate why the ssRBM is so slow. `gprof` tells that `arma::eop_core::apply_inplace_plus` wastes 69% of time. My guess is that `armadillo` allocates some temporary memory. I didn't dig into that yet.

src/mlpack/methods/ann/rbm/spike_slab_rbm_policy_impl.hpp
```diff
@@ -0,0 +1,344 @@
...
+      input);
+
+  freeEnergy -= 0.5 * hiddenSize * poolSize *
+      std::log((2.0 * M_PI) / slabPenalty);
+
+  for (size_t i = 0; i < hiddenSize; i++)
+  {
+    ElemType sum = 0;
+
+    for (size_t k = 0; k < poolSize; k++)
```
kris-singh: Can you elaborate on this i don't seem to get it fully.

lozhnikov: `sum = arma::accu(arma::pow(input.t() * weight.slice(i), 2)) / (2.0 * slabPenalty);`

src/mlpack/methods/ann/rbm.hpp
```diff
@@ -0,0 +1,186 @@
...
+ /*
+  * This function does the k-step gibbs sampling.
+  *
+  * @param input Input to the gibbs function.
+  * @param output Used for storing the negative sample.
+  * @param steps Number of gibbs sampling steps taken.
+  */
+  void Gibbs(arma::Mat<ElemType>&& input,
+             arma::Mat<ElemType>&& output,
+             size_t steps = SIZE_MAX);
```
kris-singh: the steps actually change so they can't be declared const here.

src/mlpack/tests/rbm_network_test.cpp
zoq: Have you tested the code with a lower number of iterations? Currently you use 17970 iterations, but maybe we can get reasonable result with less iterations?

src/mlpack/methods/ann/rbm/spike_slab_rbm_policy_impl.hpp
```diff
@@ -0,0 +1,336 @@
...
+  spikeSamples.set_size(hiddenSize, 1);
+  slabMean.set_size(poolSize, hiddenSize);
+};
+
+// Reset function
+template<typename DataType>
+void SpikeSlabRBMPolicy<DataType>::Reset()
+{
+  // Weight shape D * K * N
+  weight = arma::Cube<ElemType>(parameter.memptr(),
```
zoq: Do you think it might be possible to use ``arma::mat`` instead of ``arma::cube`` for the weights, that way we could vectorize some expressions (avoid a for loop), since you can't e.g. multiply a cube with a matrix and store the result as matrix. It looks like there are a lot of places that would come in handy.

src/mlpack/tests/rbm_network_test.cpp
kris-singh: So i was able to reduce the num epochs to 22 with the hidden layer size 100. And this was still beating the softmax classification function. But yes the time decrease was not that much on my local machine. Though i would push that code and see the results of travis.

src/mlpack/methods/ann/rbm/spike_slab_rbm_policy_impl.hpp
kris-singh: So, if understand this right i think you are proposing the we `weight = arma::Mat<ElemType>(parameter.memeptr(), visibleSize, poolSize * hiddenSize);`
and replace all the expr that were usign the weight matrix by `weight.subMat(....,....)`. I still think we would require the for loop though.

zoq: Using ``.cols`` and ``.each_col`` but you are right we can't avoid for loops entirely.

src/mlpack/tests/rbm_network_test.cpp
zoq: I think we don't have to beat the Softmax function, if the results are reasonable that's probably fine in this case.

src/mlpack/methods/ann/rbm/spike_slab_rbm_policy_impl.hpp
kris-singh: Well, I just saw the code again. Most of the function's actually use `weight.slice(i)`
so I think we would not be able to vectorize most of the expressions. We could use submat something like this `weight.submat(i * visibleSize * poolSize, 0, (i + 1) * visibleSize * poolSize , poolSize)` that would not get rid of the loop but i don't if armadillo matrix is preferred over cubes for performance considerations.

zoq: I'll take a second look, and let you know if I find an expression that could benefit from ``arma::mat`` over ``arma::cube``, in the meantime, can you test different numbers of iterations?

kris-singh:
Finally, I was able to make it work with time constraints. The accuracy is also good around ~77-78%.

lozhnikov:
@kris-singh I played with parameters a little. Try the following patch
[ssrbm_test.zip](https://github.com/mlpack/mlpack/files/1253591/ssrbm_test.zip)

kris-singh:
Should we merge this if everything is okay.

lozhnikov:
@kris-singh How do you think is it better to remove `useMonitoringCost` at all? As for me it is better to rewrite the `Evaluate()` function in the following way:
```
template<typename InitializationRuleType, typename RBMPolicy>
 double RBM<InitializationRuleType, RBMPolicy>::Evaluate(
     const arma::Mat<ElemType>& /* parameters*/,
     const size_t i)
{
  Gibbs(std::move(predictors.col(i)), std::move(negativeSamples));
  return std::fabs(FreeEnergy(std::move(predictors.col(i))) -
      FreeEnergy(std::move(negativeSamples)));
}
```
and remove the rest of code. Tell me if you'll fix that or I can refactor that later as soon as I merge the PR.

Regarding the merge: let me try the implementation on the CIFAR dataset, I didn't try yet, I was focused on the GAN PR.

kris-singh:
Okay i will remove the use monitoring cost. Regarding the cifar data set here is code for both creating patches and test. https://gist.github.com/kris-singh/55a4934c463bbddb3c8d321dd889d194.
I remembered that i was getting good accuracy only for a subset of data set > 75% but as I increased the size the accuracy kept falling.

ShikharJ:
@kris-singh Thanks for your work! These commits have been re-used in #1208.
@zoq This can be closed now.

zoq:
Agreed, thanks @kris-singh and @lozhnikov for the all the time you put into the code.

PR #1048, opened 2017-07-04
[GSoC] Implementing Hierarchical Memory Unit

17minutes:

Implementing HAM unit API and tree memory structure for it, as described in [arxiv paper](https://arxiv.org/abs/1602.03218).

As far as I understand, the entire HAM-related discussion is transfered here, in contrast to task-specific discussion.


src/mlpack/methods/ann/CMakeLists.txt
```diff
@@ -21,3 +21,4 @@ add_subdirectory(activation_functions)
 add_subdirectory(init_rules)
 add_subdirectory(layer)
 add_subdirectory(convolution_rules)
+add_subdirectory(augmented)
```
zoq: Can you remove non related code from this PR?

partobs-mdp:
I've written up some provisional code stub + comments with my understanding of the paper. Could you glance over it and point out some logical mistakes?

src/mlpack/methods/ann/augmented/ham_unit_impl.hpp
```diff
@@ -0,0 +1,115 @@
...
+void HAMUnit::Gradient(arma::mat&& input,
+                       arma::mat&& error,
+                       arma::mat&& gradient)
+{
+  // Compute gradients for attention values for all nodes, startng from root.
+  // For root it's (of course) zero, for nodes it's computed as
+  // grad(parent value * step probability).
+
+  // Input gradient is sum(attention gradient * leaf value)
+  // - how to get node value *before* write phase?
```
zoq: Since we store a sequence of all decisions, we know which node we have to access right? Maybe I missed something?

src/mlpack/methods/ann/augmented/ham_unit_impl.hpp
```diff
@@ -0,0 +1,115 @@
...
+  typename J,
+  typename S,
+  typename W
+>
+arma::vec HAMUnit::Attention() const
+{
+  size_t nodesCnt = 2 * memory.ActualMemorySize() - 1;
+  arma::vec probabilities(nodesCnt);
+  probabilities(0) = 1;
+  // TODO Query the controller state.
```
zoq: Do you think we should start with the Raw HAM as described in 4.3 and later extend the model with the LSTM controller? Let me know what you think.

partobs-mdp: I also think it's a good idea, since we'll get one layer of complexity out (of course, the task is still a little bit scary)

src/mlpack/methods/ann/augmented/ham_unit_impl.hpp
partobs-mdp: Yes, we do, but after every time step we do the write phase (rewriting all leaf values, calling `JOIN` for the rest)

partobs-mdp:
As a next step, how can I create a *hard-coded* FFN model? More precisely, I want to hard-code some parts of HAM to see if `Forward` pass works correctly.

zoq:
What do you mean with hard-coded FNN model:

```c++
odel.Add<Linear<> >(dataset.n_rows, 50);
model.Add<SigmoidLayer<> >();
model.Add<Linear<> >(50, 10);
model.Add<LogSoftMax<> >();
```

is probably not what you like? If you set a static seed you work with fixed parameter settings.

src/mlpack/methods/ann/augmented/tree_memory_impl.hpp
```diff
@@ -0,0 +1,119 @@
...
+template<typename T, typename J, typename W>
+void TreeMemory<T, J, W>::Initialize(vector<T>& leafValues) {
+  assert(leafValues.size() <= memorySize);
+  // First, write in the leaf nodes.
+  for (size_t i = 0; i < leafValues.size(); ++i) {
+    memory[actualMemorySize - 1 + i] = leafValues[i];
+  }
+  if (actualMemorySize < 2) return;
+  size_t lastWrittenIdx = actualMemorySize - 1 + leafValues.size() - 1;
+  // After that, write into inner nodes as prescribed by writeFunction.
```
rcurtin: I think the comment is wrong here, the JOIN function should be used (and that is what the code does), but the comment says the WRITE function will be used.

src/mlpack/methods/ann/augmented/tree_memory_impl.hpp
```diff
@@ -0,0 +1,119 @@
...
+  for (size_t i = actualMemorySize - 2; ; --i) {
+    size_t l = Left(i), r = Right(i);
+    assert(l <= r);
+    if (r > lastWrittenIdx) continue;
+    memory[i] = joinFunction(memory[l], memory[r]);
+
+    if (i == Root()) break;
```
rcurtin: You could use this as a conditional to break the loop instead of a separate line in the loop, I think that would be more intuitive.

src/mlpack/methods/ann/augmented/ham_unit_impl.hpp
```diff
@@ -0,0 +1,99 @@
...
+    size_t parent = memory.Parent(node);
+    bool dir = node == memory.Left(parent);
+    arma::vec h(hController.n_elem + memory.GetCell(parent).n_elem);
+    h.rows(0, hController.n_elem - 1) = hController;
+    h.rows(hController.n_elem,
+           hController.n_elem + memory.GetCell(parent).n_elem - 1)
+        = memory.GetCell(parent);
+    double prob = search.Forward(h);
+    if (!dir) prob = 1. - prob;
+    probabilities(node) = prob * probabilities(parent);
```
rcurtin: Does this implement Figure 3 from the HAM paper correctly?  I thought that at a given node in the tree, we calculated p = SEARCH(h_node, h_lstm), and then went left with probability (1 - p) and went right with probability p.  So we would have to same a random number to do that.  Then when we reach a leaf, the value held in the TreeMemory there would be what we use as the input for the LSTM for the output stage.  Maybe I have read this wrong, but it seems like this does not implement that.  Or is this implementing 'soft attention' from the appendix?

src/mlpack/methods/ann/augmented/ham_unit_impl.hpp
```diff
@@ -0,0 +1,99 @@
...
+  memory.Initialize(sequence);
+  t = 0;
+
+  arma::vec attention = Attention();
+  arma::vec input = arma::zeros(memory.Get(0).n_elem);
+  for (size_t i = 0; i < memorySize; ++i)
+  {
+    input += leafAttention.at(i) * memory.Get(i);
+  }
+  // TODO Feed input to the controller and get output from it.
```
rcurtin: Figure 4 is a little hard to understand for me, but I'm not sure why `leafAttention` is being multiplied by the memory, can you clarify that please?  Maybe there is part of the figure I missed.  Also they introduce this hyperparameter eta in that figure, do you have a plan for how that will be handled?

src/mlpack/methods/ann/augmented/ham_unit_impl.hpp
```diff
@@ -0,0 +1,99 @@
...
+
+  // Write phase.
+  for (size_t i = 0; i < memory.MemorySize(); ++i)
+  {
+    memory.Get(i) *= attention.at(i);
+    memory.Get(i) += (1. - attention.at(i)) * write.Forward(sequence.col(t++));
```
rcurtin: I think WRITE takes `attention` and `memory.Get(i)` as input parameters, is that correct?  Maybe I have overlooked something and this code is equivalent to that part of the paper, I am not sure.

src/mlpack/methods/ann/augmented/ham_unit_impl.hpp
rcurtin: Maybe one easy idea would be to copy the initial values of the nodes during the write phase, if the model is training (I think this can be accessed via the `Deterministic()` parameter?  I am not sure of that.).

partobs-mdp:
@zoq: No, I mean feeding my own weights to all MLPs in HAM unit. For example, if SEARCH is sigmoid(W * x + b), then I want to create SEARCH function with `W = [0 0 ... 0]` and `b = -ln(2)` to get a function that turns left 'blindly' with probability 1/3.

@rcurtin: Thanks, it's not overwhelming - I was starting to drown in the complexity anyway ^_^ Implementing.

partobs-mdp:
> If you have a chance to remove the unrelated code from this PR, it would make it nicer to scroll around and review. I can help you with that if you like, just let me know.

I know, I just wait for benchmarking PR to be merged - I'll rebase as soon as it will be merged, because there is a lot of useful code there :)

src/mlpack/methods/ann/augmented/ham_unit_impl.hpp
partobs-mdp: Yes, it's a soft attention

src/mlpack/methods/ann/augmented/tree_memory_impl.hpp
```diff
@@ -0,0 +1,119 @@
...
+inline T& TreeMemory<T, J, W>::Get(size_t index) {
+  assert(0 <= index && index < memory.size());
+  return memory[index];
+}
+
+template<typename T, typename J, typename W>
+inline T TreeMemory<T, J, W>::GetCell(size_t index) {
+  assert(0 <= index && index < memory.size());
+  return memory[index];
+}
```
partobs-mdp: Well, they're not quite the same - the first one extracts *leafs* and the seconds one extracts any node via "pointer". I do agree that current name for the first function is a misnomer - changed that to `GetLeaf`.

src/mlpack/methods/ann/augmented/tree_memory_impl.hpp
```diff
@@ -0,0 +1,119 @@
...
+  if (child == 0) return actualMemorySize;
+  return ((child + 1) >> 1) - 1;
+}
+
+template<typename T, typename J, typename W>
+void TreeMemory<T, J, W>::Initialize(vector<T>& leafValues) {
+  assert(leafValues.size() <= memorySize);
+  // First, write in the leaf nodes.
+  for (size_t i = 0; i < leafValues.size(); ++i) {
+    memory[actualMemorySize - 1 + i] = leafValues[i];
```
partobs-mdp: Well, my initial thought was to do the embedding *before* creating `TreeMemory` instance - not sure if this is a good idea, what do you think?

src/mlpack/methods/ann/augmented/tree_memory_impl.hpp
```diff
@@ -0,0 +1,119 @@
...
+void TreeMemory<T, J, W>::Update(size_t pos, T el) {
+  assert(pos >= 0 && pos < memorySize);
+  size_t start = actualMemorySize - 1 + pos;
+  memory[start] = writeFunction(memory[start], el);
+  if (start > 0) {
+    while (true) {
+      start = Parent(start);
+      size_t l = Left(start), r = Right(start);
+      memory[start] = joinFunction(memory[l], memory[r]);
+      if (start == Root()) break;
```
partobs-mdp: Well, yes, but `0` here feels so much like a magic constant :)

src/mlpack/methods/ann/augmented/ham_unit_impl.hpp
partobs-mdp: `leafAttention` is probability distribution over tree leaves. In soft attention `input` is actually the expected value of the input fed to the controller after traversing the tree as `SEARCH` prescribes.

zoq:
> @zoq: No, I mean feeding my own weights to all MLPs in HAM unit. For example, if SEARCH is sigmoid(W * x + b), then I want to create SEARCH function with W = [0 0 ... 0] and b = -ln(2) to get a function that turns left 'blindly' with probability 1/3.

What you could do is to set the parameters manually e.g.

```c++
model.Parameters().subvec(0, 10).zeros();
model.Parameters()(10) = 3;
```
does this help? I could also implement some function that enables you to use the layer do to the same, e.g.

```c++
layer.Parameters().zeros();
```

partobs-mdp:
I have problems with implementing JOIN (as `0.5 * x + 0.5 * y`) - the network outputs some garbage, even though the parameter values fed into FFN look OK. Could you take a look at the issue?

partobs-mdp:
Ok, so the primitive parts are ready :) Now switching to implementing forward pass of HAM - with these primitives we can test it quite simply, because the soft attention will always assign a constant probability distribution to leaves (for `L = 4` it will be `[1/9 2/9 2/9 4/9]`)

src/mlpack/methods/ann/augmented/tree_memory_impl.hpp
```diff
@@ -12,18 +12,22 @@
 #ifndef MLPACK_METHODS_AUGMENTED_TREE_MEMORY_IMPL_HPP
 #define MLPACK_METHODS_AUGMENTED_TREE_MEMORY_IMPL_HPP
 
+#include <mlpack/methods/ann/layer/layer.hpp>
```
zoq: We should include the header file in ``tree_memory.hpp`` so that ``LayerTypes`` is known. Also, currently the Constructor definition doesn't match with the implementation; looks like there are two options using ``FFN`` or ``LayerTypes``, if you go for ``FFN`` you don't have to use the visitor pattern, but on the other side you are restricted to a forward model, so passing a recurrent model isn't something we could do. think. I think either FFN or LayerTypes is fine here, let me know what you think.

src/mlpack/methods/ann/augmented/ham_unit_impl.hpp
```diff
@@ -0,0 +1,90 @@
...
+HAMUnit<E, J, S, W>::HAMUnit(size_t memorySize,
+                 size_t memoryDim,
+                 E& embed,
+                 J& join,
+                 S& search,
+                 W& write)
+  : memorySize(memorySize), memoryDim(memoryDim),
+    search(search), embed(embed), t(0)
+{
+  memory = TreeMemory<double, J, S>(memorySize, join, write);
```
zoq: Since there is no default constructor for  ``TreeMemory`` we should use the initialization list:

```c++
...
embed(embed),
t(0), 
memory(TreeMemory<double, J, S>(memorySize, memoryDim, join, write))
```

Also you missed the ``memoryDim`` parameter.

src/mlpack/tests/ham_test.cpp
```diff
@@ -0,0 +1,266 @@
...
+  writeModel.Parameters().rows(2 * nDim * nDim, 2 * nDim * nDim + nDim - 1) = arma::zeros(nDim);
+  // Search model is a constant model that ignores its input and returns 1 / 3.
+  FFN<MeanSquaredError<> > searchModel;
+  searchModel.Add<Linear<> >(2 * nDim, 1);
+  searchModel.ResetParameters();
+  searchModel.Parameters().rows(0, 2 * nDim - 1) = arma::zeros(2 * nDim);
+  searchModel.Parameters().at(2 * nDim) = -log(2);
+  searchModel.Add<SigmoidLayer<> >();
+
+  HAMUnit<> hamUnit(seqLen, embedModel, joinModel, searchModel, writeModel);
```
zoq: Looks like you missed the ``memoryDim`` parameter here.

src/mlpack/methods/ann/augmented/ham_unit_impl.hpp
```diff
@@ -0,0 +1,96 @@
...
+                 C& controller)
+  : memorySize(memorySize), memoryDim(memoryDim),
+    search(search), embed(embed), controller(controller),
+    memory(TreeMemory<double, J, S>(memorySize, memoryDim, join, write))
+{
+  this->t = 0;
+}
+
+template<typename E, typename J, typename S, typename W, typename C>
+arma::vec HAMUnit<E, J, S, W, C>::Attention()
```
zoq: To you mind to pas the output by reference here, to be consistent with the rest of the codebase?

src/mlpack/methods/ann/augmented/ham_unit_impl.hpp
```diff
@@ -0,0 +1,96 @@
...
+{
+  size_t nodesCnt = 2 * memory.ActualMemorySize() - 1;
+  arma::vec probabilities(nodesCnt);
+  probabilities(0) = 1;
+  arma::vec hController = sequence.col(t);
+  for (size_t node = 1; node < nodesCnt; ++node) {
+    size_t parent = memory.Parent(node);
+    bool dir = node == memory.Left(parent);
+    arma::vec h(hController.n_elem + memory.Cell(parent).n_elem);
```
zoq: We could instantiate ``h``once and reuse it for each node.

zoq:
You could also use ``LayerTypes``, Sumedh also uses the FFN class as controller so he made the ``FNN`` class part of ``LayerTypes``: https://github.com/mlpack/mlpack/pull/1072/files#diff-1ad44b5de9eab9004eb0a4ccc45c88dbR112

partobs-mdp:
@zoq: About #1072 - what is the status of that PR? I hope it will be merged soon, since it (as far as I can see) already contains tested integration of `FFN` with `LayerTypes`.

zoq:
#1072 isn't done yet, I would recommend to copy the changes from the PR you think are helpful.

src/mlpack/tests/ham_test.cpp
```diff
@@ -0,0 +1,280 @@
...
+  /*AddJoiner<double> J;
+  ReplaceWriter<double> W;*/
+  size_t memSize = 4;
+  FFN<MeanSquaredError<> > J = AddJoiner(memSize), W = ReplaceWriter(memSize);
+  TreeMemory<double> mem(
+      8, memSize, J, W);
+  std::vector<std::vector<double>> initMemSTL =
+      {{0, 0, 0, 0}, {0, 0, 0, 0}, {0, 0, 0, 0}, {0, 0, 0, 0},
+       {0, 0, 0, 0}, {0, 0, 0, 0}, {0, 0, 0, 0}, {0, 0, 0, 0}};
+  arma::mat initMem = convertToArma(initMemSTL);
```
rcurtin: Why not just initialize the Armadillo matrix directly? I think that would allow you to simplify a lot of code.

src/mlpack/tests/ham_test.cpp
```diff
@@ -0,0 +1,280 @@
...
+  }
+  BOOST_REQUIRE_EQUAL(mem.Cell(
+    mem.Parent(mem.LeafIndex(0))).n_elem, memSize);
+  BOOST_REQUIRE_EQUAL(mem.Cell(
+    mem.Parent(mem.Parent(mem.LeafIndex(0)))).n_elem, memSize);  
+  BOOST_REQUIRE_EQUAL(mem.Cell(
+    mem.Parent(mem.Parent(mem.Parent(mem.LeafIndex(0))))).n_elem, memSize); 
+}
+
+BOOST_AUTO_TEST_CASE(TreeMemoryTestArbitrary) {
```
rcurtin: Can you be sure to add comments to each test saying what it does, and also comments to what you are testing? It takes a lot more time to try and read the code to understand what the test is doing.  (This comment applies to all the tests, not just this one.)

src/mlpack/methods/ann/augmented/tree_memory_impl.hpp
```diff
@@ -0,0 +1,158 @@
...
+}
+
+template<typename T, typename J, typename W>
+void TreeMemory<T, J, W>::Update(size_t pos, arma::Col<T> el)
+{
+  assert(pos >= 0 && pos < memorySize);
+  size_t start = actualMemorySize - 1 + pos;
+  arma::Mat<T> newCellValue;
+  writeFunction.Predict(Stack(Cell(start), el), newCellValue);
+  Cell(start) = newCellValue;
```
partobs-mdp: I tried it, but it doesn't really work since `Cell(node)` doesn't really return reference to the corresponding column - it returns `arma::subview_col<T>`, and there goes the C++ type system :)

src/mlpack/methods/ann/augmented/tree_memory_impl.hpp
```diff
@@ -0,0 +1,158 @@
...
+  {
+    start = Parent(start);
+    size_t l = Left(start), r = Right(start);
+    joinFunction.Predict(Stack(Cell(l), Cell(r)), newCellValue);
+    Cell(start) = newCellValue;
+  }
+}
+
+template<typename T, typename J, typename W>
+arma::Mat<T> TreeMemory<T, J, W>::Stack(arma::Mat<T> left, arma::Mat<T> right) const
```
partobs-mdp: Well, `Stack` is used for *arbitrary* columns, not necessarily the ones from the memory:

```C++
writeFunction.Predict(Stack(Cell(start), el), newCellValue);
```

Rewrote the function with const-reference parameters, since we don't really need to modify them or copy them inside a function.

src/mlpack/tests/ham_test.cpp
partobs-mdp: Well, I did it for clarity (just like in every other test). In the "internal" `arma::mat` representation memory vectors are aligned column-wise, which imho is not human-intuitive. Besides, there is a convenient "standard" way to initialize `std::vector` (in `arma` it turns out to be platform-dependent). Taking these two together, I decided to initialize STL vector (intuitive) and then convert it to `arma` (hides complexity)

rcurtin: Couldn't you just do, e.g.,

```
arma::mat matrix = {{1, 2, 3, 4}, {5, 6, 7, 8}};
matrix = matrix.t();
```

I think that gives you the same result, and uses significantly less code (and `convertToArma()` can be removed).

Also, in the particular case above this comment here, I think you can just do

```
arma::mat initMem(4, 8, arma::fill::zeros);
```

or if you really want to have it be row-major when you initialize it,

```
arma::mat initMem(8, 4, arma::fill::zeros);
initMem = initMem.t();
```

I'd advise treating Armadillo matrices in C++ as column-major, even if you find it a little counterintuitive, because the rest of the library does.  One workaround would be to add a comment clarifying what rows represent and what columns represent.  i.e. with `arma::mat m(8,
 4, arma::fill::zeros)` you could write a comment like `// 4 points in 8 dimensions.` or similar.  (For what it's worth: every textbook in the world implies row-major, but FORTRAN and the low-level scientific computing world imply column-major.  I think even numpy stores matrices in column-major format like Armadillo by default.)

src/mlpack/methods/ann/augmented/tree_memory_impl.hpp
rcurtin: I'd suggest avoiding the functions `Cell()` and `Leaf()` entirely, and instead access the correct column of your memory directly.  After all, `Cell()` and `Leaf()` are just calling `.col()` and then forcing the type of the result.  I think that should be avoided.

Instead try the following:

```
writeFunction.Predict(Stack(memory.unsafe_col(start), el), memory.unsafe_col(start));
```

The `unsafe_col()` function returns a new column vector that is an alias of the same memory, so no copy happens.

src/mlpack/methods/ann/augmented/tree_memory_impl.hpp
rcurtin: I think there might still be optimization opportunities here, but at least the const-reference should help.  If you are only passing single columns at a time, I'd still suggest using `const arma::Col<T>&` just to be a little more specific.  That would also obviate the need for the `assert()` check in `Stack()`.

rcurtin:
It looks like the `RebuildParameters()` method is necessary because the various parts of the `HAMUnit` aren't stored in contiguous memory.  I think it should be possible to, when you create the `HAMUnit` object, ensure that all the memory used is contiguous.  Consider this code:

```
// Assume we know the size of everything...
extern size_t embedCount, searchCount, controllerCount, joinCount, writeCount;

// Now create the parameters matrix.
parameters = arma::mat(embedCount + searchCount + controllerCount + joinCount + writeCount, 1);

// Next, set up the parameters matrices of other things to be aliases
// of the right part of the parameters matrix.
embed.Parameters() = arma::mat(parameters.memptr(), embedCount, 1, false /* no copy */, true /* no resize */);
search.Parameters() = arma::mat(parameters.memptr() + embedCount, searchCount, 1, false, true);
controller.Parameters() = arma::mat(parameters.memptr() + embedCount + searchCount, controllerCount, 1, false, true);
memory.JoinObject().Parameters() = arma::mat(parameters.memptr() + embedCount + searchCount + controllerCount, joinCount, 1, false, true);
memory.WriteObject.Parameters() = arma::mat(parameters.memptr() + embedCount + searchCount + controllerCount + joinCount, writeCount, 1, false, true);
```

That uses one of the Armadillo advanced constructors that re-uses other memory.  So after that code, the memory of each component of `HAMUnit` is set up as an alias of `HAMUnit::parameters`, and no `RebuildParameters()` function would be necessary.

Also as a side note, be careful with statements like this:

```
arma::mat embedParams = embed.Parameters();
```

That incurs a full copy of `embed.Parameters()`.  To avoid the copy, either don't make the temporary variable `embedParams`, or use a `const arma::mat&` for its type.

Let me know if I can clarify anything; I think the idea I just proposed would help make the design simpler (and help make debugging the rest of it simpler).

src/mlpack/methods/ann/augmented/ham_unit_impl.hpp
```diff
@@ -123,7 +124,34 @@ void HAMUnit<E, J, S, W, C>::ResetParameters() {
...
+
+  size_t embedCount = embedParams.n_elem,
+         searchCount = searchParams.n_elem,
+         controllerCount = controllerParams.n_elem,
+         joinCount = joinParams.n_elem,
+         writeCount = writeParams.n_elem;
+
+  parameters = arma::mat(embedCount + searchCount + controllerCount + joinCount + writeCount, 1);
+
+  embed.Parameters() = arma::mat(parameters.memptr(), embedCount, 1, false, true);
```
zoq: > Implemented Parameters() method as rcurtin described (make individual function parameters as memory pointers to some contiguous memory block), but there are still only zeros in the HAMUnit paramters. Can you take a look at the issue?

It should work if you use ``embed.Parameters() = arma::mat(parameters.memptr(), embedCount, 1, false, false);`` instead of ``embed.Parameters() = arma::mat(parameters.memptr(), embedCount, 1, false, true);``. Let us know if this works for you.

src/mlpack/methods/ann/augmented/ham_unit_impl.hpp
```diff
@@ -139,19 +114,30 @@ void HAMUnit<E, J, S, W, C>::ResetParameters() {
 
   parameters = arma::mat(embedCount + searchCount + controllerCount + joinCount + writeCount, 1);
 
-  embed.Parameters() = arma::mat(parameters.memptr(), embedCount, 1, false, true);
-  search.Parameters() = arma::mat(parameters.memptr() + embedCount, searchCount, 1, false, true);
-  controller.Parameters() = arma::mat(parameters.memptr() + embedCount + searchCount, controllerCount, 1, false, true);
-  memory.JoinObject().Parameters() = arma::mat(parameters.memptr() + embedCount + searchCount + controllerCount, joinCount, 1, false, true);
-  memory.WriteObject().Parameters() = arma::mat(parameters.memptr() + embedCount + searchCount + controllerCount + joinCount, writeCount, 1, false, true);
+  embed.Parameters() = arma::mat(parameters.memptr(), embedCount, 1, false, false);
```
zoq: The problem with the current solution is, that we also have to reset the memory pointer for each layer, so I propose to use the ``NetworkInitialization`` class, which does all this for us.

Here is the updated ``ResetParameters`` method:


```c++
template<typename E, typename J, typename S, typename W, typename C>
void HAMUnit<E, J, S, W, C>::ResetParameters()
{
  embed.ResetParameters();
  search.ResetParameters();
  controller.ResetParameters();
  memory.ResetParameters();

  const size_t embedCount = embed.Parameters().n_elem;
  const size_t searchCount = search.Parameters().n_elem;
  const size_t controllerCount = controller.Parameters().n_elem;
  const size_t joinCount = memory.JoinObject().Parameters().n_elem;
  const size_t writeCount = memory.WriteObject().Parameters().n_elem;

  parameters = arma::mat(embedCount + searchCount + controllerCount +
      joinCount + writeCount, 1);

  NetworkInitialization<> networkInit;
  networkInit.Initialize(embed.Model(), parameters);
  networkInit.Initialize(search.Model(), parameters, embedCount);
  networkInit.Initialize(controller.Model(), parameters, embedCount +
      searchCount);
  networkInit.Initialize(memory.JoinObject().Model(), parameters, embedCount +
      searchCount + controllerCount);
  networkInit.Initialize(memory.WriteObject().Model(), parameters, embedCount +
      searchCount + controllerCount + joinCount);

  reset = true;
}
```

and here is the updated NetworkInitialization class: https://gist.github.com/zoq/a895915d2da1dc36770e2aea583b1147 it's bacially the same as used in #1066, just set a default ``InitializationRuleType``.

I've also passed the ``joiner`` and ``writer`` as reference in the ``TreeMemory`` class.

zoq:
Thanks for the comments, make it easier to understand the code, can you take a look at the build issues? It would be great if the code could be build automatically, it looks like some file are just missing.

rcurtin:
Looks like this compiles and passes the tests correctly, which is great. :)  Have you had a chance to benchmark the task against the LSTM model?

PR #1050, opened 2017-07-04
Updated Armadillo version to 6.500.0

Rajiv2605:
I have updated the Armadillo version from 4.200.0 to 6.500.0 in README.md, HISTORY.md and /doc/guide/build.hpp files. Please review it and suggest changes if any.


zoq:
Thanks for opening the PR, can you also take a look at: ``CMakeLists.txt`` and ``.travis.yml``?

Rajiv2605:
@zoq After pushing the code, I am getting error in travis-ci checks. I am not able to understand why. It would be great if you could help me.

zoq:
We can test if ``dist: xenial`` instead of ``dist: trusty`` works. The problem I see is that xenial is only available as docker container, which doesn't run on GCE, so it's probably slower, but let's see.

zoq:
Turns out Travis don't support xenial for now: https://github.com/travis-ci/travis-ci/issues/5821, So I was wondering, if it's a good idea to bump the version, I guess we could install armadillo from source but using the package manager is way easier. Let me all know what you think.

Rajiv2605:
@zoq So what should I do now?

zoq:
We should wait for further input, I thought Ubuntu Trusty would support Armadillo 6 out of the box, but it turns out that's not the case, so the question is, do we still require Armadillo 6.500.0.

rcurtin:
I think it's ok to build Armadillo 6.500.0 from source in the Trusty container; it should only take a few seconds at most.

In the longer term we should request a backport of Armadillo to trusty:

https://wiki.ubuntu.com/UbuntuBackports

I will put it on my todo list for this week to open a bug there.  I'm not sure how long it will take to get that issue resolved though. It would be bad to require all our users who want to use modern mlpack to install Armadillo from source.

rcurtin:
Just to add to that if you would like me to clarify the Armadillo build steps I am happy to, just let me know. :)

Rajiv2605:
@zoq @rcurtin So what is the plan for now?

zoq:
We install armadillo 6.500.0 from source the basic steps are:

```
before_install:
  ...
  - sudo apt-get install libopenblas-dev
  - download and unpack armadillo
  - cd to download dir && cmake . && make && sudo make install
  - sudo cp .travis/config.hpp /usr/include/armadillo_bits/config.hpp
```

.travis.yml
```diff
@@ -1,14 +1,16 @@
...
 
 before_install:
   - sudo add-apt-repository -y ppa:ubuntu-toolchain-r/test
   - sudo apt-get update -qq
-  - sudo apt-get install -qq g++-4.8 libarmadillo-dev libarmadillo4 libboost1.54-all-dev
   - export CC=gcc-4.8
   - export CXX=g++-4.8
   - printenv
+  - sudo apt-get install libopenblas-dev
+  - wget http://sourceforge.net/projects/arma/files/armadillo-7.950.1.tar.xz && tar xf armadillo-7.950.1.tar.xz
```
zoq: You can simplify the command:

```
curl https://distfiles.macports.org/armadillo/armadillo-6.500.5.tar.gz | tar xvz && cd armadillo*
```

That way, we are already in the correct folder and can call ``make`` ...

Rajiv2605:
@zoq Still not getting accepted by travis... Should I try changing dist to trusty and then try?

zoq:
Have you seen my comment? https://github.com/mlpack/mlpack/pull/1050#discussion_r125899222

Rajiv2605:
@zoq I added the Armadillo's building from source commands to the .travis.yml yet it isn't getting accepted.

zoq:
The problem is 
```
wget https://launchpad.net/ubuntu/+archive/primary/+files/armadillo_6.500.5+dfsg.orig.tar.gz && tar xf armadillo_6.500.5+dfsg.orig.tar.gz
```
doesn't download the expected file, you can test it on your machine. So I proposed:

```
curl https://distfiles.macports.org/armadillo/armadillo-6.500.5.tar.gz | tar xvz && cd armadillo*
```

which has the additional benefit that you don't have to know the armadillo version, so:

```
cd armadillo_6.500.5
```

is obsolet.

zoq:
Looks like you accidentally removed: ``libboost1.54-all-dev``

Rajiv2605:
@zoq OhOh... I'll check

Rajiv2605:
@zoq TRAVIS SAYS: The command "mkdir build && cd build && cmake -DDEBUG=OFF -DPROFILE=OFF .. && make -j2" failed and exited with 2 during .

rcurtin:
@Rajiv2605: in this case, it seems like Armadillo is not built in a way that will link against LAPACK.  I think you need to install liblapack-dev.

Rajiv2605:
@zoq @rcurtin It passed!!:dancer::dancer::dancer:

rcurtin:
Looks good.  Are you willing to do the next thing, which is look through `arma_extend/` and remove anything that is no longer necessary for backporting?  Basically there is a lot of code that will look like

```
#if ARMA_MAJOR_VERSION < 4
... do something...
#endif
```

and since the minimum version of Armadillo is now 6.500.0, then we don't need that code anymore.  I think the majority of that code will be in `src/mlpack/core/arma_extend/` but there may also be a little in `src/mlpack/tests/`.

Once that piece is done, then I think this will be basically ready for merge; I will need to handle the other things (upgrade EPEL package and request Ubuntu trusty package upgrade).

Rajiv2605:
@rcurtin Fine I'll do that:thumbsup:

Rajiv2605:
@rcurtin I have removed the unnecessary codes for backporting in `src/mlpack/core/arma_extend/`. I am not sure about what all to remove in `src/mlpack/tests` as it has variations like `ARMA_MAJOR_VERSION > 4` and else cases to `ARMA_MAJOR_VERSION < 4`. It would be great if you could help me in this.

src/mlpack/core/arma_extend/Mat_extra_bones.hpp
```diff
@@ -29,136 +29,4 @@ typedef Row<elem_type>   row_type;
  * row_col_iterator for Mat<eT>. This iterator can return row and column index
  * of the entry its pointing too. The functionality of this iterator is similar
  * to sparse matrix iterators.
```
rcurtin: We can remove the comments too since we no longer need to add row_col_iterator to Mat. :)

src/mlpack/core/arma_extend/SpMat_extra_bones.hpp
```diff
@@ -27,17 +27,4 @@ typedef SpRow<elem_type>   row_type;
 /*
  * Extra functions for SpMat<eT>
  * Adding definition of row_col_iterator to generalize with Mat<eT>::row_col_iterator
- */
```
rcurtin: We can also remove this comment.

src/mlpack/core/arma_extend/fn_ind2sub.hpp
```diff
@@ -5,74 +5,4 @@
...
-    arma_extra_debug_sigprint();
-
-    arma_debug_check( ((row >= s.n_rows) || (col >= s.n_cols) || (slice >= s.n_slices)), "sub2ind(): subscript out of range" );
-
-    return uword( (slice * s.n_rows * s.n_cols) + (col * s.n_rows) + row );
-    }
-#endif
-
-
+// Written by Conrad Sanderson - http://conradsanderson.id.au
```
rcurtin: There's nothing left in this file, we can remove it.

src/mlpack/core/arma_extend/operator_minus.hpp
```diff
@@ -9,27 +9,4 @@
...
-  {
-  arma_extra_debug_sigprint();
-
-  typedef typename T1::elem_type eT;
-
-  return SpOp<T1,spop_scalar_times>(X, eT(-1));
-  }
-
-#endif
+// older.
```
rcurtin: This file can be removed too.

rcurtin:
For the tests, you can take a look at the conditions, which might look like this:

```
#if ARMA_VERSION_MAJOR > 4 ...

do one thing

#else

do other thing

#endif
```

Now we know that `ARMA_VERSION_MAJOR` will always be 6 or greater (since the minimum version of Armadillo will be 6.500.0), so in situations like that we can simply remove the condition, and keep only the `do one thing` code from the conditional.  Does that make sense?  I can try and explain better if you like.

Rajiv2605:
@rcurtin The build doesn't take place when I deleted the files `fn_ind2sub.hpp` and `operator_minus.hpp`. I have attached the travis logs which were generated due to the absence of these files.
[travis_log_1.txt](https://github.com/mlpack/mlpack/files/1130015/travis_log_1.txt)
[travis_log_2.txt](https://github.com/mlpack/mlpack/files/1130014/travis_log_2.txt)

rcurtin:
I would strongly recommend trying to build your changes on your local system before pushing to have Travis build it.  It will save you a lot of time and grief.  In this case the error indicates that you need to modify the CMake configuration so it does not try to build files that do not exist.

zoq:
You can also remove ``ind2sub_test.cpp`` in ``tests/`` and update the ``CMakeLists.txt`` accordingly.

Rajiv2605:
@zoq @rcurtin I think I have done all the changes.

Rajiv2605:
@zoq @rcurtin Is it fine? Or is there anything more I am supposed to do?

zoq:
@Rajiv2605 Thanks for looking into the issue, can you also take a look at:

```
primal_dual_impl.hpp
load_impl.hpp
gaussian_distribution.cpp
```

.travis.yml
zoq: @Saurabh3012 @rcurtin If I remember right you work with some older armadillo versions, maybe we can use masterblaster to provide the package?

Rajiv2605:
@zoq @rcurtin I have updated code for backporting in all places I guess. Is it fine?

rcurtin:
@Saurabh3012: I added some Armadillo versions I found on the MacPorts site to masterblaster.  But it's not fully comprehensive.

@zoq: @Rajiv2605: we can use http://masterblaster.mlpack.org:5005/armadillo-6.500.5.tar.gz as the source to build from.

.travis.yml
```diff
@@ -1,14 +1,17 @@
 sudo: required
-dist: trusty
+dist: xenial
```
rcurtin: Ah, sorry, one more thing I overlooked, should we stick with the `trusty` container because of https://github.com/travis-ci/travis-ci/issues/5821 ?  The build logs show that it is still building on trusty anyway.

zoq: It will use the xenial dist once Travis provides a container, which will probably happen in the next month, so at the moment it shouldn't matter.

rcurtin: Got it, in that case, no need to change anything.

Rajiv2605:
@zoq @rcurtin All changes done!

rcurtin:

Rajiv2605:
Thanks!

PR #1051, opened 2017-07-06
Turn VecType into an alias in NaiveBayesClassifier

### LINKED ISSUE #1049: Do we need the VecType template parameter in NaiveBayesClassifier?
`NaiveBayesClassifier` has the following `Train` interfaces:
```
             const bool incremental = true);

  template<typename VecType>
  void Train(const VecType& point, const size_t label);
```
Do we really need for the second signature to have the `VecType` template parameter? Can't we just use the alias `using VecType = arma::Col<typename MatType::value_type>`?

micyril:
Fix #1049

### DIFF
--- a/src/mlpack/methods/naive_bayes/naive_bayes_classifier.hpp
+++ b/src/mlpack/methods/naive_bayes/naive_bayes_classifier.hpp
@@ -48,8 +48,8 @@ template<typename MatType = arma::mat>
 class NaiveBayesClassifier
 {
  public:
-   //! A short alias for the data point type.
-   using VecType = arma::Col<typename MatType::value_type>;
+  //! A short alias for the data point type.
+  using VecType = arma::Col<typename MatType::value_type>;
 
   /**
    * Initializes the classifier as per the input and then trains it by
--- a/src/mlpack/methods/naive_bayes/naive_bayes_classifier.hpp
+++ b/src/mlpack/methods/naive_bayes/naive_bayes_classifier.hpp
@@ -48,6 +48,9 @@ template<typename MatType = arma::mat>
 class NaiveBayesClassifier
 {
  public:
+   //! A short alias for the data point type.
+   using VecType = arma::Col<typename MatType::value_type>;
+
   /**
    * Initializes the classifier as per the input and then trains it by
    * calculating the sample mean and variances.
@@ -107,7 +110,6 @@ class NaiveBayesClassifier
    * @param point Data point to train on.
    * @param label Label of data point.
    */
-  template<typename VecType>
   void Train(const VecType& point, const size_t label);
 
   /**
@@ -116,7 +118,6 @@ class NaiveBayesClassifier
    *
    * @param point Point to classify.
    */
-  template<typename VecType>
   size_t Classify(const VecType& point) const;
 
   /**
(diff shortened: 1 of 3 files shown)


rcurtin:
Hmm, there is one other thing that I overlooked.  Previously I could do this:

```
model.Train(data.col(i));
```

but now with this change we would not be able to do that.  So, maybe it would be a better idea to simply enforce that the element type is the same, and leave `VecType` as a template parameter.  Do you agree?

micyril:
The bad thing is we can't apply MetaInfoExtractor to a class that contains both templated and non-templated Train methods (I have mentioned this restriction in the MetaInfoExtractor documentation). If you think it's better to avoid turning VecType to the alias, maybe we should make other Train methods templated by adding `template<typename = void>`?

rcurtin:
I spent a while reading through the `MetaInfoExtractor` code.  To make sure that I understand the problem, I think that for `NaiveBayesClassifier`, the train form `TF3` is matched by `template<typename VecType> void Train(VecType&, size_t)` before form `TF4` matches `void Train(MatType&, arma::Row<size_t>)`, which causes `PredictionsType` to be incorrectly deduced as `size_t`.  Is that correct?

If this is true, I think maybe we could modify `TrainFormBase` so that it cannot deduce `PT = size_t`.  I think that would be a better solution than adding `template<typename = void>` to the other `Train()` calls, because that can make the documentation more confusing for users.

Let me know if I have understood the problem correctly.  It takes me a long time to think through what happens with complex template metaprogramming like this...

micyril:
It looks like something strange happened - I thought I already answered you a long time ago, but probably it was in the preview mode, and I forgot to press the "Comment" button.

What is going on can be illustrated by the following code snippets. Suppose we have the following Train form
```
  template<typename Class, typename...T>
  using TrainForm =
      void(Class::*)(const arma::mat&, const arma::Row<size_t>&, T...);
```
and a function that takes a pointer to a method of the Train form with one additional argument
```
  template<typename Class, typename T1>
  void f(TrainForm<Class, T1>) {};
```

Then, the following code works just fine (it compiles).
```
  class A
  {
  public:
    void Train(const arma::mat&, const arma::Row<size_t>&, double) {}
  };

  void main()
  {
    f(&A::Train);
  }
```

But if we try to compile the following code
```
  class B
  {
  public:
    void Train(const arma::mat&, const arma::Row<size_t>&, double) {}

    template<typename VecType>
    void Train(const VecType&, size_t, double) {}
  };

  void main()
  {
    f(&B::Train);
  }
```

we get the error
```
error: no matching function for call to 'f'
    f(&B::Train);
note: candidate template ignored: couldn't infer template argument 'Class'
  void f(TrainForm1<Class, T1>) {};
```
The following code works again.
  ```
class C
  {
  public:
    template <typename = void>
    void Train(const arma::mat&, const arma::Row<size_t>&, double) {}

    template<typename VecType>
    void Train(const VecType&, size_t, double) {}
  }

  void main()
  {
    f(&C::Train<>);
  }
```

rcurtin:
It seems to me that it should be possible to get the class `B` to have the right template parameters inferred by the call to `f()`, but I am not sure it's worth the time to debug all that.  I thought about it for a while, and I realized that we can also resolve the situation by adding a little extra functionality to the `NaiveBayesClassifier` class.  Take a look at #1068---what do you think?  I believe it solves the problem, and I think that provides a little more in the way of functionality than adding the `template<typename = void>` to the other overload of `Train()`.

rcurtin:
Let's use #1068 instead.

Issue #1052, opened 2017-07-07
Parallel serialization testing can fail


rcurtin:
This would be a great issue for someone looking to get involved in the project.

Here is a fascinating unintended consequence of parallel tests: the `SerializeObjectAll()` function, used in `fastmks_test.cpp` as well as other tests, uses boost::serialization to save the model to a specific file.  But since we might be testing in parallel, then files may get overwritten arbitrarily, thus causing tests to fail in strange ways like here:

https://travis-ci.org/mlpack/mlpack/builds/250830590?utm_source=github_status&utm_medium=notification

In order to fix this problem, we'll need to modify `SerializeObjectAll()` in `src/mlpack/tests/serialization.hpp` in a way that it uses unique filenames instead of `"test"`.  There are very many ways to do this, so there is a lot of freedom in how to fix the issue.

@shikharbhardwaj: CC'ing you here just to let you know (feel free to fix it if you like, but no obligation!).  Even if we think about everything beforehand and are careful about what we merge, there is always something that gets overlooked. :)

shikharbhardwaj:
That slipped under my radar. Thanks for the heads up. :)

Should be an easy fix indeed. I'll wait for some time, if no one takes this up, I'll fix it. (As you mentioned on IRC, this is a good beginner bug).

shikharbhardwaj:
I started to look into some ways this could be done, in a cross platform manner.

 * There's `std::tempnam()`, which returns a filename, but is generally deprecated (not thread-safe and can be an attack vector, although it should not matter much here, but still, fewer warnings are always better.)

 * `std::tmpfile()` which returns a C `FILE*` handle, I'm not sure if that would be useful here.

Other ways require `Boost.Filesystem` or platform specific functions like `mkstemp()` on Linux or `GetTempFileName()` on Windows.

rcurtin:
Another idea could be just assembling a name based on the test, or the types given to the test.  RTTI could probably be used here to get a string from a type that should be unique enough.

shikharbhardwaj:
I had thought about that. Although the string returned by `typeid(t).name()` is not guaranteed to be unique for each type, most implementations keep it that way. We could just truncate and filter the string to be a valid filename.

rcurtin:
[...]

PR #1053, opened 2017-07-10
Use unique filenames in Serialization tests

shikharbhardwaj:
Fix for #1052 

`FilterFileName()` does not handle the corner case of `inputString` being empty, but I think it should be fine here.

### DIFF
--- a/src/mlpack/tests/test_tools.hpp
+++ b/src/mlpack/tests/test_tools.hpp
@@ -48,4 +48,18 @@ inline void CheckMatrices(const arma::Mat<size_t>& a,
     BOOST_REQUIRE_EQUAL(a[i], b[i]);
 }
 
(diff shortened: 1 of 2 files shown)


src/mlpack/tests/test_tools.hpp
```diff
@@ -48,4 +48,18 @@ inline void CheckMatrices(const arma::Mat<size_t>& a,
     BOOST_REQUIRE_EQUAL(a[i], b[i]);
 }
 
+inline std::string FilterFileName(const std::string& inputString)
```
rcurtin: Can you add a comment to mention what this function is for please? Even in the tests we should be sure we have enough documentation. :)

mentekid:
I like it. Thanks for being so fast with this :)

PR #1055, opened 2017-07-12
Use arma::gmm_diag for diagonal GMM training

### LINKED ISSUE #215: GMM::Classify() is slow
**Reported by rcurtin on 16 Sep 42238823 05:55 UTC**
Currently, in gmm.cpp, `GMM::Classify()` is defined like so:

```

void GMM::Classify(const arma::mat& observations,
                   arma::Col<size_t>& labels) const
{
  // This is not the best way to do this!

  // We should not have to fill this with values, because each one should be
  // overwritten.
  labels.set_size(observations.n_cols);
  for (size_t i = 0; i < observations.n_cols; ++i)
  {
    // Find maximum probability component.
    double probability = 0;
    for (size_t j = 0; j < gaussians; ++j)
    {
      double newProb = Probability(observations.unsafe_col(i), j);
      if (newProb >= probability)
      {
        probability = newProb;
        labels[i] = j;
      }
    }
  }
}
```

This is an O(n*m) operation, where m is the number of Gaussians and n is the number of points.  It could be sped up greatly with the use of trees -- using the means of the Gaussians as "reference points" and the phi() function as the distance metric.

In addition, at the very least, running `GMM::Probability()` over and over again for each point is slow, as it will invert each covariance matrix each time the method is called.  It could be replaced with a better call, perhaps to one of the overloads of `phi()` which takes many observations at once.

Migrated-From: http://trac.research.cc.gatech.edu/fastlab/ticket/221

rcurtin:
This finally fixes #215 by using `arma::gmm_diag` when the GMM fitting constraint is `DiagonalConstraint`.  Summary of changes:

 * use `std::is_same<>` to detect when `DiagonalConstraint` is used, then call `EMFit::ArmadilloGMMWrapper()`
 * add `-d` option to `mlpack_gmm_train` so that a user can train diagonal GMMs
 * add test case for diagonal GMM training

Unfortunately `gmm_diag` doesn't support weighted training, so we still have to use our implementation for that, this PR does not improve that.  Some basic timing on training diagonal GMMs:

```
old:
covertype-5k: 7.027s
isolet: 37.580s
cloud: 1.636s

new:
covertype-5k: 0.245s
isolet: 19.341s
cloud: 0.163s
```

I'd say that is a pretty significant improvement. :)


src/mlpack/methods/gmm/diagonal_constraint.hpp
```diff
@@ -27,8 +27,7 @@ class DiagonalConstraint
   static void ApplyConstraint(arma::mat& covariance)
   {
     // Save the diagonal only.
-    arma::vec diagonal = covariance.diag();
-    covariance = arma::diagmat(diagonal);
+    covariance = arma::diagmat(arma::clamp(covariance.diag(), 1e-10, DBL_MAX));
```
zoq: Good that we bump the armadillo version, that way we can use ``arma::clamp``.

conradsnicta:
@rcurtin @zoq - I've recently made the training in arma::gmm_diag even faster under OpenMP, by removing a bottleneck in the calculation of initial covariance matrices.

The speedup will be released as part of Armadillo 7.960, which I'm aiming to finalise before the end of the month.

src/mlpack/methods/gmm/em_fit_impl.hpp
```diff
@@ -316,6 +325,63 @@ void EMFit<InitialClusteringType, CovarianceConstraintPolicy>::Serialize(
...
+    arma::mat covs(observations.n_rows, dists.size());
+    for (size_t i = 0; i < dists.size(); ++i)
+    {
+      means.col(i) = dists[i].Mean();
+      covs.col(i) = dists[i].Covariance().diag();
+    }
+
+    g.set_means(std::move(means));
+    g.set_dcovs(std::move(covs));
+    g.set_hefts(std::move(weights));
```
conradsnicta: You may want to use `g.set_params(means, dcovs, hefts)` here instead of setting them separately.  The `.set_means()` etc functions first check if the given parameters are consistent with the existing model (number of dimensions, gaussians, etc), so they may fail if the model isn't setup correctly in the first place.

rcurtin:
Finally got around to working out the issues with this one.

PR #1056, opened 2017-07-13
Add aggregated policy for async rl methods

ShangtongZhang:


src/mlpack/methods/reinforcement_learning/policy/aggregated_policy.hpp
```diff
@@ -0,0 +1,78 @@
...
+class AggregatedPolicy
+{
+ public:
+  //! Convenient typedef for action.
+  using ActionType = typename PolicyType::ActionType;
+
+  /**
+   * @param policies Child policies.
+   * @param distribution Probability distribution for each child policy.
+   *   User should make sure its size is same as the number of policies
```
zoq: Just a minor style issue: "When wrapping a line, the next line should be tabbed twice from where the previous line began." https://github.com/mlpack/mlpack/wiki/DesignGuidelines#line-length-and-wrapping

src/mlpack/methods/reinforcement_learning/policy/aggregated_policy.hpp
```diff
@@ -0,0 +1,78 @@
...
+   *   and the sum of its element is equal to 1.
+   */
+  AggregatedPolicy(std::vector<PolicyType> policies,
+                   arma::colvec distribution) :
+          policies(std::move(policies)),
```
zoq: Can you tab twice here?

src/mlpack/methods/reinforcement_learning/policy/aggregated_policy.hpp
```diff
@@ -0,0 +1,78 @@
...
+      return policies.front().Sample(actionValue, true);
+    return policies[RandomIndex(distribution)].Sample(actionValue, false);
+  }
+
+  /**
+   * Exploration probability will anneal at each step.
+   */
+  void Anneal()
+  {
+    for (auto& policy : policies)
```
zoq: Do you mind to use ``PolicyType``instead of ``auto`` here?

ShangtongZhang:
@zoq Can you trigger the test process for https://github.com/mlpack/mlpack/pull/934? As arma is now updated, I want to see whether it works or not.

zoq:
@ShangtongZhang I think the latest commit already triggered the build: https://travis-ci.org/mlpack/mlpack/builds/253458551?utm_source=github_status&utm_medium=notification but it looks like it timed out.

zoq:
@ShangtongZhang Anyway, I can restart the build if you like.

ShangtongZhang:
@zoq Sure. Thanks!

src/mlpack/core/math/random.hpp
```diff
@@ -58,6 +58,27 @@ inline double Random()
...
+  double oops = Random();
+  double density = distribution(0);
+  size_t index = 0;
+  while (oops >= density)
+  {
+    index++;
+    density += distribution(index);
+  }
+  return index;
+}
```
rcurtin: Do you think that `DiscreteDistribution` could be used here instead?  You can create a `DiscreteDistribution` with the `distribution` colvec, then sample with `DiscreteDistribution::Random()`.

ShangtongZhang: It's amazing, I even don't know that class. And I find a better solution, we can use `std:: discrete_distribution`

rcurtin: It might be better to use `DiscreteDistribution` just to be consistent; I'm not sure of how it compares (speedwise) to `std::discrete_distribution`, although it does offer some more functionality.  I'm not picky on what you do in the end, I just wanted to suggest it as a possibility.

rcurtin: Ah, it might also be good to just use `DiscreteDistribution` directly where you are currently using `RandomIndex()`.  Your call.

ShangtongZhang: That makes sense. I have updated the code.

ShangtongZhang:
Could anyone merge this PR?

rcurtin:
Sure, I can hit the button. :)

PR #1057, opened 2017-07-14
Update optimizer batch support

Rajiv2605:
I have refractored SGDTestFunction as of now.

### DIFF
--- a/src/mlpack/methods/logistic_regression/logistic_regression_function_impl.hpp
+++ b/src/mlpack/methods/logistic_regression/logistic_regression_function_impl.hpp
@@ -99,6 +99,40 @@ double LogisticRegressionFunction<MatType>::Evaluate(
 }
 
 /**
+ * Evaluate the logistic regression objective function given the estimated
+ * parameters for a given batch from a given point.
+ */
+template<typename MatType>
+  double Evaluate(const arma::mat& parameters,
+                  const size_t begin,
+                  const size_t batchSize) const
+{
+  // Calculating the regularization term.
+  const double regularization = 0.5 * lambda *
+    arma::dot(parameters.col(0).subvec(begin, batchSize - 1),
+              parameters.col(0).subvec(begin, batchSize - 1));
+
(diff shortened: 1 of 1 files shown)


Rajiv2605:
@zoq Which all codes need to be updated?

zoq:
@Rajiv2605 If you like you can continue with ``LogisticRegressionFunction``.

Rajiv2605:
@zoq I am trying to understand the code base. From where should I start?

rcurtin:
Hi Rajiv,

Thanks for taking a look into this one.  I just wanted to say---please keep in mind that there is a lot of activity in this repository so maintainers typically won't have the time to provide in-depth guidance for each bug.  For learning the codebase, I suggest looking at http://www.mlpack.org/involved.html and http://www.mlpack.org/gsoc.html for guides on how to get started, and if elements of the codebase still don't make sense, it is probably a good idea to read up on some C++ template metaprogramming resources or machine learning resources.

I hope this is helpful! :)

Rajiv2605:
Thanks! I'll do that:+1:

Rajiv2605:
@zoq I don't get why we should refractor to support batch optimization when minibatch-sgd is already existant.

zoq:
Minibatch SGD doesn't work on "real" batches, we evaluate each example and accumulate the the result later, but there are some existing function e.g. neural network which could benefit from "real" batch computation e.g. we only have to iterate over a specific matrix once. I hope this is helpful, let me know if I should clarify anything further.

Rajiv2605:
@zoq Thanks! Currently, I am trying to understand the code-base completely. So, I am taking time to go through each method in the library... :smile:

Rajiv2605:
@zoq My local repo got deleted and the branch in which I was pushing got deleted along with it. After that, I was unable to push changes to the branch even after re-cloning it. Hence, I am closing this pull request and creating a new one #1073.

zoq:
Okay, no problem.

PR #1061, opened 2017-07-18
[GSoC] Highway network

17minutes:

Implementing a stub for Highway network class. When the time comes: can you give me directions on what to do next to make it work?


zoq:
Could you take a look into the travis build issue and the style checks?

src/mlpack/core/optimizers/sgd/sgd_impl.hpp
```diff
@@ -107,6 +107,9 @@ double SGD<UpdatePolicyType>::Optimize(
     else
       function.Gradient(iterate, currentFunction, gradient);
 
+    // Time to do something really dirty.
+    // gradient.transform( [](double val) { return std::min(std::max(val, -3.), 3.); } );
```
zoq: I think we can remove this one here, once gradient clipping is merged.

src/mlpack/methods/ann/CMakeLists.txt
```diff
@@ -21,3 +21,4 @@ add_subdirectory(activation_functions)
 add_subdirectory(init_rules)
 add_subdirectory(layer)
 add_subdirectory(convolution_rules)
+add_subdirectory(augmented)
```
zoq: Can you clean this PR, and remove all unrelated files?

zoq:
Do you like to finish this first before continuing with the HAM implementation? Both options are fine for me.

partobs-mdp:
> Do you like to finish this first before continuing with the HAM implementation? Both options are fine for me.

Not sure if this is OK, but I would do it in this order:

- Resolve benchmarking PR (since it's almost finished);
- Implement and merge HAM (since this one is the project focus, not the highway network);
- *Resolve this PR*.

Is it OK or should I do it in some other order?

zoq:
@partobs-mdp  Sounds good for me, let's do this.

sourabhvarshney111:
Looks like this is pending from GSoC, 2017. @zoq Can you tell me the issue with this PR? I might be able to do the remaining task.

zoq:
@sourabhvarshney111 sorry for the slow response, this PR contains the augmented related code as well, so the first step would be to separate the Highway network from the rest and see that the test are passing.

sourabhvarshney111:
@zoq Does the augmented rnn code is merged already?

PR #1064, opened 2017-07-19
Implementation of async one step q-learning

ShangtongZhang:
Although the test time is still a bummer, I think in general it's ready for review.


src/mlpack/methods/reinforcement_learning/training_config.hpp
```diff
@@ -0,0 +1,124 @@
...
+ * 3-clause BSD license along with mlpack.  If not, see
+ * http://www.opensource.org/licenses/BSD-3-Clause for more information.
+ */
+#ifndef MLPACK_METHODS_RL_TRAINING_CONFIG_HPP
+#define MLPACK_METHODS_RL_TRAINING_CONFIG_HPP
+
+namespace mlpack {
+namespace rl {
+
+class TrainingConfig
```
zoq: I like the approach, but it's not conform with what's proposed here: https://github.com/mlpack/mlpack/pull/1044#issuecomment-314816704. I think for now this is fine. Not sure adding each parameter to the constructor is helpful here. What be great to get @rcurtin's view on this.

src/mlpack/tests/async_learning_test.cpp
```diff
@@ -0,0 +1,88 @@
...
+  model.Add<Linear<>>(20, 20);
+  model.Add<ReLULayer<>>();
+  model.Add<Linear<>>(20, 2);
+
+  // Set up the policy.
+  using Policy = GreedyPolicy<CartPole>;
+  AggregatedPolicy<Policy> policy({Policy(0.7, 5000, 0.1),
+                                   Policy(0.7, 5000, 0.01),
+                                   Policy(0.7, 5000, 0.5)},
+                                  arma::colvec("0.4 0.3 0.3"));
```
zoq: Pedantic style issue: missing space.

src/mlpack/tests/async_learning_test.cpp
```diff
@@ -0,0 +1,88 @@
...
+  auto measure = [&rewards, &testEpisodes](double reward) {
+    size_t maxEpisode = 20000;
+    if (testEpisodes > maxEpisode)
+      BOOST_REQUIRE(false);
+    testEpisodes++;
+    rewards.push_back(reward);
+    size_t windowSize = 20;
+    if (rewards.size() > windowSize)
+      rewards.erase(rewards.begin());
+    double avgReward = std::accumulate(
```
zoq: Looks like windows has a problem with this, maybe we have to include ``numeric``to solve the problem.

zoq:
I'm not sure there is anything we can do about the test right now, but at the same time I'm not sure it's feasible to include the test, what do you think? I don't like to remove it, since it could be helpful for everyone looking for a simple example, but running it on a daily basis is probably difficult, especially if the python bindings are merged.

src/mlpack/methods/reinforcement_learning/training_config.hpp
ShangtongZhang: `CV` shouldn't be a problem, as I don't think there exists some `cross validation` in RL. Maybe in the future I need to refactor the API to make it consistent with other parts of mlpack, but it won't block the `CV` project anyway for now. As for hyper-parameter tuner, I'm not very clear how it works and it seems in that discussion there hasn't be a clear standard. 

> In the hyperparameter tuner it only makes sense for the user to pass all of the hyperparameters to the Optimize() function

RL does need a hyper parameter tuner, but if we need to pass all the hyper parameters to `Train()`, it may be a disaster for user.

src/mlpack/tests/async_learning_test.cpp
ShangtongZhang: I can do it with `arma::vec` to make this code more mlpackish.

ShangtongZhang:
Yeah it's important to keep this test because without this simple example, user may never know the correct way to use it. In my understanding, the most important role for this test case is to teach user how to use it rather than to make sure it converges (after all we can test convergency off-line when we modify rl-related code). Based on this following is my suggestion:

- Load a pre-trained network
- Simply comment out the `agent.Train(...)` and leave some comments there

Which one do you prefer?

src/mlpack/methods/reinforcement_learning/training_config.hpp
zoq: I should have mentioned that, I was referring to the standard mlpack interface: especially "Hyperparameters are passed to the constructor, but not to Train().", if we do that we should probably do this for all methods to avoid any confusion. But I I think we can do this later and I agree we don't have to wait for the change.

zoq:
I like the idea of using a pre-trained network, but not sure what the speedup in this case would be.

ShangtongZhang:
I can try to load a fully converged network to see what happens.

zoq:
I guess, the interesting part is the travis runtime here, so yeah let's see.

ShangtongZhang:
```
    std::ifstream ifs(fileName, std::ios::binary);
    boost::archive::binary_iarchive i(ifs);
    i >> data::CreateNVP(model, "network");
    ifs.close();

    std::ofstream ofs(fileName, std::ios::binary);
    boost::archive::binary_oarchive o(ofs);
    o << data::CreateNVP(agent.Network(), "network");
    ofs.close();
```
Is my serialization wrong? It works well in my local environment but gets 
```
unknown location(0): fatal error in "OneStepQLearningTest": std::exception: unsupported version
```
in the server.

zoq:
hm, I'll have to take a deeper look into the issue; have you tested:

```c++
data::Save("output_model_file.bin", "model", model, false);
data::Load("input_model_file.bin", "model", model);
```

I think it's easier to use.

zoq:
Looks like it works with ``data::save``, ``data::load``.

ShangtongZhang:
It's weird. It still costs more than 500 seconds. In my mac it only needs less than 20 episodes (before this is about 5000 episodes). Let's see what will happen if I set # of workers to 0.

ShangtongZhang:
It looks good now. It took about 3s.

src/mlpack/methods/reinforcement_learning/async_learning_impl.hpp
```diff
@@ -0,0 +1,108 @@
...
+  // Make sure OpenMP will start desired number of threads.
+  omp_set_dynamic(0);
+  // An extra thread for test.
+  omp_set_num_threads(config.NumOfWorkers() + 1);
+
+  // All the variables that doesn't show up in this list will be copied.
+  #pragma omp parallel for shared(learningNetwork, targetNetwork, \
+      stop, totalSteps, policy)
+  // OpenMP only support signed index variable
+  for (int i = 0; i <= (int)config.NumOfWorkers(); ++i)
```
zoq: You could use ``omp_size_t`` as defined in https://github.com/mlpack/mlpack/blob/980ba7e49a12379b7c1837ba933ff45f9dbf3bdb/src/mlpack/prereqs.hpp

src/mlpack/methods/reinforcement_learning/async_learning_impl.hpp
```diff
@@ -0,0 +1,108 @@
...
+    learningNetwork.ResetParameters();
+  NetworkType targetNetwork = learningNetwork;
+  bool stop = false;
+  size_t totalSteps = 0;
+  PolicyType policy = this->policy;
+
+  // Make sure OpenMP will start desired number of threads.
+  omp_set_dynamic(0);
+  // An extra thread for test.
+  omp_set_num_threads(config.NumOfWorkers() + 1);
```
zoq: Should we use ``OMP_NUM_THREADS`` to set the number of threads/worker, as we do this for the other methods, if it's not set we use the available number of cores. What do you think?

ShangtongZhang: I'm not sure whether my understanding is right. Do you mean remove `NumOfWorkers` property from `TrainingConfig`, and get this value from `OMP_NUM_THREADS` and then set # of threads by `omp_set_num_threads`? My concern is I want to make sure this value is set. I assume user may easily forget to set this environment variable if the CTOR doesn't require it, so we may end up with # of cores, which is fine for most algorithms but not suitable for async rl. Because async rl doesn't care too much about the speed, it relies on uncorrelated experiences. So in general we need more agents than # of cores. Another concern is to make the test completed in short time in the server, we need to use pre-trained network and set # of workers to 1. Is there a way to set `OMP_NUM_THREADS` to 1 only for this test?

zoq: I agree with you that there should be an option to specify the number of worker, but I'm not sure, it's a good idea to create the same amount of threads. It could be faster to create a pool of threads = # of cores, to keep the overhead low. We could test the performance by setting #worker >> #cores and compare the results with both strategies, what do you think?

About the test, we could set omp_set_num_threads(1) before the test starts and reset the value afterwards.

src/mlpack/methods/reinforcement_learning/worker/one_step_q_learning_worker.hpp
```diff
@@ -0,0 +1,172 @@
...
+    // Get initial state.
+    StateType state = environment.InitialSample();
+
+    size_t steps = 0;
+    double totalReturn = 0.0;
+
+    // Store pending transitions for update later on.
+    std::vector<TransitionType> pending;
+
+    while (!stop && (!config.StepLimit() || steps < config.StepLimit())) {
```
zoq: Pedantic style issue; can you put the opening brace on a new line?

src/mlpack/methods/reinforcement_learning/worker/one_step_q_learning_worker.hpp
```diff
@@ -0,0 +1,172 @@
...
+
+        // Sync the local network with the global network.
+        network = learningNetwork;
+
+        pending.clear();
+      }
+
+      if (terminal)
+        break;
+      else
```
zoq: We could remove the else branch here, since we will break if the condition is true, just wanted to note this here, since it works both ways.

src/mlpack/methods/reinforcement_learning/worker/one_step_q_learning_worker.hpp
```diff
@@ -0,0 +1,172 @@
...
+    NetworkType network = learningNetwork;
+
+    // Get initial state.
+    StateType state = environment.InitialSample();
+
+    size_t steps = 0;
+    double totalReturn = 0.0;
+
+    // Store pending transitions for update later on.
+    std::vector<TransitionType> pending;
```
zoq: Do you think, we should reserve the size here, to avoid ``pending.clear();`` in line 148? Depending on how often we call ``clear`` we could speed things up?

src/mlpack/methods/reinforcement_learning/worker/one_step_q_learning_worker.hpp
```diff
@@ -0,0 +1,172 @@
...
+          // Compute gradient.
+          arma::mat gradients;
+          network.Backward(actionValue, gradients);
+
+          // Accumulate gradients.
+          totalGradients += gradients;
+        }
+
+        // Clamp the accumulated gradients.
+        totalGradients = arma::clamp(totalGradients,
```
zoq: In #1070 Konstantin used ``transform``: https://github.com/mlpack/mlpack/pull/1070/files#diff-42e728c2a9d60b90643d92b5a352168cR76 not sure it's faster, but it avoid's an extra copy.

src/mlpack/methods/reinforcement_learning/async_learning_impl.hpp
ShangtongZhang: As far as I'm concerned, I think setting # of threads equal to # of workers is important. Say we have 16 workers and 4 cores(4 threads), so we have a for loop from 1 to 16. In my understanding, OpenMP will let thread one run 1 to 4 sequentially, thread two run 5 to 8 sequentially and so on. So in any given time, there are at most 4 different workers running, so we can gain only 4 different uncorrelated experiences. So our program will be equivalent to having 4 workers and more training episodes, which won't be as good as true 16 workers. (Although in cart pole the difference may be not so clear) So the overhead may be necessary.

src/mlpack/methods/reinforcement_learning/worker/one_step_q_learning_worker.hpp
ShangtongZhang: Maybe not. If `terminal` is `false`, we need to go to next step directly without touching pending buffer and doing any update, since we are now in `deterministic` mode (i.e. test process).

zoq: You mean we can't write:

```
if (terminal)
  break;

state = nextState;
```

instead of:

```
if (terminal)
  break;
else
  state = nextState;
```

ShangtongZhang: oh I got it!

src/mlpack/methods/reinforcement_learning/async_learning_impl.hpp
zoq: Ah, I missed that you share parameter, in this case, we can't achieve the same results with # of worker != # of threads:

I was thinking about:

#pragma omp parallel for
for(int i = 0; i < numerOfWorker; ++i)
{
  // do something

  std::cout << workerId << std::endl;
}

if we use:

numerOfWorker = 8;
omp_set_num_threads(numerOfWorker);

we would see something like:

```
1
2
3
4
5
6
7
8
```

but with:

```
omp_set_num_threads(4); // number of virtual cores
```

we would see something like:

```
1
2
3
4
1
2
3
4
```

depending on the task the second strategy could be faster since we can use on thread per core and reduce schedule overhead etc.

Thanks for the clarification.

src/mlpack/methods/reinforcement_learning/async_learning.hpp
```diff
@@ -0,0 +1,144 @@
...
+  PolicyType& Policy() { return policy; }
+  const PolicyType& Policy() const { return policy; }
+
+  //! Get and modify optimizer.
+  UpdaterType& Updater() { return updater; }
+  const UpdaterType& Updater() const { return updater; }
+
+  //! Get and modify the environment.
+  EnvironmentType& Environment() { return environment; }
+  const EnvironmentType& Environment() const { return environment; }
```
rcurtin: Very minor doxygen issue---can you make one comment for each of these methods instead of a comment that covers both?  Doxygen won't pick up any documentation for `const EnvironmentType& Environment() const` like this is now.

src/mlpack/methods/reinforcement_learning/training_config.hpp
```diff
@@ -0,0 +1,124 @@
...
+ public:
+  TrainingConfig() : stepLimit(0), gradientLimit(40), doubleQLearning(false)
+  { /* Nothing to do here. */ }
+
+  /**
+   * Get and modify the number workers.
+   * This is valid only for async RL agent.
+   **/
+  size_t NumOfWorkers() const { return numOfWorkers; }
+  size_t& NumOfWorkers() { return numOfWorkers; }
```
rcurtin: I'd call this `NumWorkers()` just to be consistent with other mlpack code that looks like `NumX()` and not `NumOfX()`.

src/mlpack/methods/reinforcement_learning/async_learning_impl.hpp
rcurtin: I'm sorry to throw this comment in so late.  My understanding is that ideally `NumOfWorkers()` needs to be large for the algorithm to work.  But the code like this doesn't allow a user to limit the number of cores that the program uses, which is I think an important use case in resource-limited environments or other environments where there may be some contention.

But I see that it is important that there be enough workers, otherwise we will not get updates for each worker at each time step.  I think it might be better to adjust the design so that `Episode` doesn't run all iterations at once, but instead runs a single iteration at a time.  i.e., the loop here would look like:

```
for (size_t i = 0; i < numIterations; ++i)
{
  #pragma omp for ...
  for (omp_size_t j = 0; j < config.NumOfWorkers(); ++i)
  {
    ...
  }
}
```

and in this way you can get an update from each iteration of each episode even when having `OMP_NUM_THREADS` set to fewer than the number of workers.  If a problem with that approach is that each worker could take a different amount of time per iteration, you could adapt the idea above to use OpenMP tasks too.

src/mlpack/methods/reinforcement_learning/training_config.hpp
rcurtin: Sorry for the slow response here---I was unavailable over the weekend so I could only get to this now.  I think for the hyperparameter tuner, we would only need to add a constructor to `TrainingConfig` that also takes all of the parameter values.  I believe the HP tuner will be able to pass hyperparameters to "nested" constructors, so even if not all hyperparameters are directly set in the top-level RL class constructors, it's ok if they are set in constructors of objects that are passed to RL classes.

src/mlpack/methods/reinforcement_learning/worker/one_step_q_learning_worker.hpp
```diff
@@ -0,0 +1,174 @@
...
+    StateType state = environment.InitialSample();
+
+    size_t steps = 0;
+    double totalReturn = 0.0;
+
+    // Store pending transitions for update later on.
+    std::vector<TransitionType> pending(config.UpdateInterval());
+    size_t pendingIndex = 0;
+
+    while (!stop && (!config.StepLimit() || steps < config.StepLimit())) {
```
rcurtin: Picky style comment--- the opening brace should be on the following line. :)

src/mlpack/methods/reinforcement_learning/worker/one_step_q_learning_worker.hpp
```diff
@@ -0,0 +1,174 @@
...
+            learningNetwork.Parameters().n_cols);
+        for (size_t i = 0; i < pending.size(); ++i) {
+          TransitionType &transition = pending[i];
+
+          // Compute the target state-action value.
+          arma::colvec actionValue;
+          #pragma omp critical
+          {
+            targetNetwork.Predict(
+                std::get<3>(transition).Encode(), actionValue);
```
rcurtin: I think this would fit onto just one line.

ShangtongZhang: It's 81 characters :<

src/mlpack/methods/reinforcement_learning/async_learning_impl.hpp
ShangtongZhang: Do you think `GOMP_CPU_AFFINITY` can be used by user to limit the amount of cores used?
https://gcc.gnu.org/onlinedocs/libgomp/GOMP_005fCPU_005fAFFINITY.html

ShangtongZhang: But I assume few people know this environment variable and it's less consistent with other part of mlpack :( So I will break down `Episode` into `Step` and use a worker pool (queue) with OpenMP task.

ShangtongZhang:
Finally made it. Now the # of threads is total independent with # of workers.

src/mlpack/methods/reinforcement_learning/async_learning_impl.hpp
```diff
@@ -0,0 +1,139 @@
...
+    tasks.push(i);
+
+  // Synchronization for get/put task.
+  omp_lock_t tasksLock;
+  omp_init_lock(&tasksLock);
+
+  #pragma omp parallel shared(stop, workers, tasks, learningNetwork, \
+      targetNetwork, totalSteps, policy)
+  {
+    #pragma omp task
```
zoq: Looks like Visual Studio only supports OpenMP 2.0 is way outdated. We could try to build with clang-cl, if that doesn't work I don't mind to do a ifdef here. Let me know what you think. Also I can look into the clang-cl setup, if you like.

ShangtongZhang: I'm not sure it's good to compile it with clang-omp in windows, because most windows user may only have MSVC. Although my envision is to use clang everywhere instead of gcc and msvc, it looks impractical. I have a workaround. I can write it in a for loop instead of `task`, where the length of the for loop is equal to the # of threads. `omp_get_num_threads` barely works (at least it doesn't work in my mac and always return 1) So I need to use some trick https://stackoverflow.com/questions/11071116/i-got-omp-get-num-threads-always-return-1-in-gcc-works-in-icc . What do you think about this?

zoq: I think clang-cl comes with the recent MSVS, but if you could think of a workaround that would be even better. By workaround I guess you mean:

```
int n = 0;
#pragma omp parallel reduction(+:n)
n += 1;
```

From the OpenMP documentation:

> If this call is made from a serial portion of the program, or a nested parallel region that is serialized, it will return 1.

That could be the issue, but I don't mind to use a workaround here to make everyone happy.

ShangtongZhang: oops.. I misunderstood this function by its name. If it's only valid in parallelized region, I still cannot use it. I will apply that trick, hope it's reliable.

zoq: Okay, let me know what you find out.

ShangtongZhang:
Here is some stats with random seed:
With 8 threads, it fails twice out of 20 trials, with 1 thread, it fails once out of 20 trials. In general it's stable.

src/mlpack/methods/reinforcement_learning/async_learning_impl.hpp
```diff
@@ -89,42 +89,50 @@ void AsyncLearning<
   omp_lock_t tasksLock;
   omp_init_lock(&tasksLock);
 
-  #pragma omp parallel shared(stop, workers, tasks, learningNetwork, \
+  /**
+   * Compute the number of threads for the for-loop. In general, we should use
+   * OpenMP task rather than for-loop, here we do so to be compatible with some
+   * compiler.
```
zoq: I guess, we could be more specific here, e.g. we can switch to OpenMP task once MSVC supports OpenMP 3.0; might help us later to find that code block again.

zoq:
> Here is some stats with random seed:
> With 8 threads, it fails twice out of 20 trials, with 1 thread, it fails once out of 20 trials. In general it's stable.

I think that's fine, I guess a quick solution would be to run multiple trials like we did here: https://github.com/mlpack/mlpack/blob/master/src/mlpack/tests/recurrent_network_test.cpp#L571. But that might be unnecessary here, what do you think?

ShangtongZhang:
Maybe we don't need that for now. The probability for failure is small enough I think.

zoq:
Agreed, we can adjust the test case if needed later.

zoq:
Thanks for looking into the issue, @rcurtin any more comments on this one, if not I think we can merge it.

zoq:
Let us wait one more day, before we merge the code.

src/mlpack/methods/reinforcement_learning/async_learning_impl.hpp
```diff
@@ -0,0 +1,147 @@
...
+    workers.push_back(WorkerType(updater, environment, config, !i));
+    workers.back().Initialize(learningNetwork);
+  }
+  // Set up task queue corresponding to worker pool.
+  std::queue<size_t> tasks;
+  for (size_t i = 0; i <= config.NumWorkers(); ++i)
+    tasks.push(i);
+
+  // Synchronization for get/put task.
+  omp_lock_t tasksLock;
```
zoq: There is one last thing we should talk about that I missed, if OpenMP isn't supported the build would fail, so do you think we should surround the code with ``#ifdef HAS_OPENMP`` or just specific parts as I did in #1077. Let me know what you think.

ShangtongZhang: yeah it makes sense and most users may simply don't have OpenMP installed. I'll make it.

src/mlpack/methods/reinforcement_learning/async_learning_impl.hpp
```diff
@@ -0,0 +1,164 @@
...
+
+  #ifdef HAS_OPENMP
+    #pragma omp parallel for shared(stop, workers, tasks, learningNetwork, \
+        targetNetwork, totalSteps, policy)
+  #endif
+  for (omp_size_t i = 0; i < numThreads; ++i)
+  {
+    #ifdef HAS_OPENMP
+      #pragma omp critical
+    #endif
```
rcurtin: I think you only need the `#ifdef HAS_OPENMP` around the `omp_*()` function calls; any unknown `#pragma`s will typically be ignored by the compiler.

src/mlpack/methods/reinforcement_learning/async_learning_impl.hpp
```diff
@@ -0,0 +1,164 @@
...
+          policy, episodeReturn) && !task)
+      {
+        stop = measure(episodeReturn);
+      }
+
+      // Put task back to queue.
+      #ifdef HAS_OPENMP
+        omp_set_lock(&tasksLock);
+      #endif
+      tasks.push(task);
```
rcurtin: Instead of using OpenMP locks here, could you instead use a critical section?

rcurtin:
Sorry for the slow response.  I think the new parallel implementation is a lot better, thank you for taking the time to do that!  I left a couple comments on it, but they are minor comments.  If you can handle those please I think this is ready for merge.

src/mlpack/methods/reinforcement_learning/async_learning_impl.hpp
ShangtongZhang: maybe not. At any time there should be only at most one thread in `push` and `pop` section, so the two parts need to share one lock. If i use `critical` for each part, it looks they have two independent lock.

src/mlpack/methods/reinforcement_learning/async_learning_impl.hpp
ShangtongZhang: Thanks for pointing this out. I will make it.

src/mlpack/methods/reinforcement_learning/async_learning_impl.hpp
rcurtin: Ah, I see what you mean.  You could possibly restructure something like...

```
bool hasTask = false;

#pragma omp parallel ...
while (!stop) // worker loop
{
  // Push old task if needed, get new task.
  #pragma omp critical
  {
    if (hasTask)
      tasks.push(task);
    task = tasks.front();
    tasks.pop();
    hasTask = true;
  }

  // Do task...
}
```

If this is possible, I think it removes all need for `omp_set_lock()` and `omp_unset_lock()`.  What do you think?  It's possible I overlooked something.

ShangtongZhang: Oh I got it. The lock is at the beginning and end of the loop, so we could cat them together in another order!

ShangtongZhang:
Would anyone have more comments?

zoq:
Thanks for the awesome work!

PR #1065, opened 2017-07-19
First stab at adding matrix types 🚀

etano:
Opening this now not to be immediately merged, but just to get a flavor for what this change might look like.

More files will be touched in the end (e.g. AdaMax), and it's already quite a large merge...

Comments welcome!

### DIFF
--- a/src/mlpack/core/optimizers/adam/adam.hpp
+++ b/src/mlpack/core/optimizers/adam/adam.hpp
@@ -50,10 +50,10 @@ namespace optimization {
  * is required. This class must implement the following function:
  *
  *   size_t NumFunctions();
- *   double Evaluate(const arma::mat& coordinates, const size_t i);
- *   void Gradient(const arma::mat& coordinates,
+ *   double Evaluate(const MatrixType& coordinates, const size_t i);
+ *   void Gradient(const MatrixType& coordinates,
  *                 const size_t i,
- *                 arma::mat& gradient);
+ *                 MatrixType& gradient);
  *
  * NumFunctions() should return the number of functions (\f$n\f$), and in the
  * other two functions, the parameter i refers to which individual function (or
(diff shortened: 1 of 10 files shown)


src/mlpack/core/optimizers/adam/adam.hpp
```diff
@@ -107,7 +110,7 @@ class AdamType
    * @return Objective value of the final point.
    */
   template<typename DecomposableFunctionType>
-  double Optimize(DecomposableFunctionType& function, arma::mat& iterate)
+  double Optimize(DecomposableFunctionType& function, MatrixType& iterate)
```
zoq: I think we could write:

```c++
template<typename DecomposableFunctionType, typename MatType>
double Optimize(DecomposableFunctionType& function, MatType& iterate)
{
...
}
```

That way, we can avoid the extra class template parameter. What do you think?

Note, I just used ``MatType`` instead of ``MatrixType`` for consistency reasons.

src/mlpack/methods/ann/rnn.hpp
```diff
@@ -281,22 +282,22 @@ class RNN
...
-  arma::mat predictors;
+  MatrixType predictors;
 
   //! The matrix of responses to the input data points.
-  arma::mat responses;
+  MatrixType responses;
 
   //! Matrix of (trained) parameters.
-  arma::mat parameter;
+  MatrixType parameter;
```
rcurtin: Hi @etano, thanks again for taking the time to do this.  I think that if we templatize the internal parameters like this, we can end up with a bad situation: the user may want to pass `arma::sp_mat` (a sparse matrix) as their input, but we don't want to store a sparse matrix internally for the parameters (because they will generally be non-sparse).

One option is that we can do something like this:

```
typedef MatrixType::elem_type ElemType;

arma::Mat<ElemType> parameter;
```

and this could sidestep the problem.  What do you think?

(I'd also suggest changing to `MatType` like @zoq suggested.)

etano:
Hey, sorry it's been awhile. I'll go head and close this for now while I work on it. :)

rcurtin:
No worries, whenever you want to continue with it we are happy to help out. :)

PR #1066, opened 2017-07-20
Generative Adversarial Network

kris-singh:
I still have to make the test work and add the evaluate function.
Also can somebody help me regarding what loss function to use for 0/1 loss.


src/mlpack/methods/ann/gan_impl.hpp
```diff
@@ -0,0 +1,268 @@
...
+  fakeLables.set_size(1, batchSize);
+  tempTrainData.set_size(predictors.n_rows, 2 * batchSize);
+  tempLabels.set_size(1, 2 * batchSize);
+  fakeData.zeros();
+  fakeLables.zeros();
+  parameter.set_size(generator.Parameters().n_rows + 
+      discriminator.Parameters().n_rows,
+      generator.Parameters().n_cols + generator.Parameters().n_cols);
+  initializeRule.Initialize(parameter, parameter.n_elem, 1);
+  generator.Parameters() = arma::mat(parameter.memptr(),
```
lozhnikov: I think this line is incorrect since you can not change this pointer. I mean that the layers of the generator network still store data in the previous pointer. That leads to segfault. So, you can not change the pointer after `ResetParameters()`.

src/mlpack/methods/ann/gan_impl.hpp
```diff
@@ -0,0 +1,293 @@
...
+    gradient.zeros();
+  }
+
+  // Gradient for generator network
+  gradientGenerator = arma::mat(gradient.memptr(),
+      generator.Parameters().n_rows,
+      generator.Parameters().n_cols, false, false);
+
+  // Gradient for discriminator network
+  gradientDisriminator = arma::mat(gradient.memptr(),
```
zoq: We don't have to share the gradient's between the generator and discriminator right? But that's what you do here, should this be: ``gradientDisriminator = arma::mat(gradient.memptr() + generator.Parameters().n_elem, ...);``? Maybe I missed something?

src/mlpack/methods/ann/gan_impl.hpp
```diff
@@ -0,0 +1,293 @@
...
+      discriminator.Parameters().n_rows,
+      discriminator.Parameters().n_cols, false, false);
+
+  // get the gradients of the discriminator
+  discriminator.Gradient(parameters, i, gradientDisriminator);
+  // apply error to the first layer
+  boost::apply_visitor(BackwardVisitor(std::move(boost::apply_visitor(
```
zoq: I think you could avoid this step by using an Identity layer as first layer of the discriminator network. You should test that, but I think it's cleaner and probably less error prone.

src/mlpack/methods/ann/gan_impl.hpp
```diff
@@ -0,0 +1,293 @@
...
+        discriminator.network[0]);
+    generator.Backward();
+    generator.ResetGradients(gradient);
+    generator.currentInput = noiseData.col(i);
+    generator.Gradient();
+  }
+
+  // Freeze weights of discrminator if generator is training
+  if (!trainGenerator)
+    gradientGenerator.zeros();
```
zoq: The gradient is zero for each network, since you shared the gradient, take a look at the comment above.

src/mlpack/methods/ann/gan_impl.hpp
```diff
@@ -0,0 +1,293 @@
...
+      // Training discriminator
+      this->predictors = std::move(tempTrainData);
+      this->responses = std::move(tempLabels);
+      numFunctions = predictors.n_cols;
+      // set here discrminator evaluate function
+      discriminator.predictors = predictors;
+      discriminator.responses = responses;
+      // do one pass over the dataset
+      Optimizer.MaxIterations() = numFunctions;
+      Optimizer.Optimize(*this, parameter);
```
zoq: The problem with this strategy is that you reset the update policy for each call of ``Optimize``, which is fine for the vanilla update policy, but maybe it's a better idea to generate the complete training set and call ``Optimize`` once?

src/mlpack/methods/ann/gan_impl.hpp
kris-singh: Yes that's what i wanted to do. Sorry i changed this on my local branch did not push it.

zoq: In this case, can you push your latest changes, so that we look at the same state?

lozhnikov:
@kris-singh  Remember, I asked you in #mlpack if it is possible to refactor the `Gradient()` function in such a way that the `Train()` function contains only `optimizer.Optimize()`? I mean is it possible to call the optimizer once? I think it is possible. If so, you are doing unnecessary work right now since the code requires full refactoring in order to do that.

src/mlpack/methods/ann/gan_impl.hpp
```diff
@@ -0,0 +1,328 @@
...
+    generator(generator),
+    discriminator(discriminator),
+    trainGenerator(false),
+    batchSize(batchSize),
+    iterations(iterations),
+    disIteration(disIteration),
+    generatorInSize(generatorInSize),
+    reset(false)
+{
+  discriminator.network.insert(discriminator.network.begin(), new Join<>());
```
lozhnikov: If I am not mistaken, you add this line in order get the error, right? In such a way I suggest `IdentityLayer` (see 'ann/layer/base_layer.hpp').

src/mlpack/methods/ann/gan_impl.hpp
```diff
@@ -0,0 +1,328 @@
...
+::Train(OptimizerType& Optimizer)
+{
+  if (!reset)
+    Reset();
+  std::chrono::time_point<std::chrono::system_clock> start, end;
+  for (size_t i = 0; i < iterations; i++)
+  {
+    std::cout << "iteration #" << i << std::endl;
+    start = std::chrono::system_clock::now();
+    size_t offset = 0;
```
lozhnikov: I think the variable should be defined and initialized outside the loop.

kris-singh:
When you say have the`Optimizer.Optimize()` function just once. I do understand that i am calling function `2* iteration` number of times. but i think with your fix also I would still have to call it `iteration` number of times right.

kris-singh:
@lozhnikov can you review this.  The samples generated are still random. Can you look at any obvious mistakes that i might be overlooking.

kris-singh:
The main error here is that the gradient values are too small and when multiplied with a smaller learning rate they parameters end up not changing much any ideas on how to counter this. I was sgd as optimizer with learning rate 0.001 there is very less learning with 0.1 aslo.

src/mlpack/methods/ann/gan_impl.hpp
```diff
@@ -0,0 +1,332 @@
...
+
+  assert(generatorWeights > 0);
+
+  for (size_t i = 0; i < discriminator.network.size(); ++i)
+    discriminatorWeights += boost::apply_visitor(weightSizeVisitor,
+        discriminator.network[i]);
+
+  assert(discriminatorWeights > 0);
+
+  parameter.set_size(generatorWeights + discriminatorWeights, 1);
```
zoq: In order to get stable results, we have to be careful with the choice of architecture and parameter initialization. I think instead of using random weights we should test the ``GaussianInitialization`` or ``NguyenWidrowInitialization`. So instead of reimplementing the weight initialization, we should use the ``NetworkInitialization`` class, that way we are going to initialize each layer using the correct parameter space.

The only problem I see is that we would have to adjust the NetworkInitialization class so that we could call it with a predefined parameter matrix. So that once we call Initialize we don't resize the matrix.

```c++
void Initialize(const std::vector<LayerTypes>& network, arma::mat& parameter, const size_t parameterOffset = 0)
{
    if (parameter.is_empty())
    {
      // Determine the number of parameter/weights of the given network.
      size_t weights = 0;
      for (size_t i = 0; i < network.size(); ++i)
        weights += boost::apply_visitor(weightSizeVisitor, network[i]);
      parameter.set_size(weights, 1);
    }

    // Initialize the network layer by layer or the complete network.
    if (ann::InitTraits<InitializationRuleType>::UseLayer)
    {
      for (size_t i = 0, offset = parameterOffset; i < network.size(); ++i)
      {
        // Initialize the layer with the specified parameter/weight
        // initialization rule.
        const size_t weight = boost::apply_visitor(weightSizeVisitor,
            network[i]);
        arma::mat tmp = arma::mat(parameter.memptr() + offset,
            weight, 1, false, false);
        initializeRule.Initialize(tmp, tmp.n_elem, 1);

        // Increase the parameter/weight offset for the next layer.
        offset += weight;
      }
    }
    else
    {
      initializeRule.Initialize(parameter, parameter.n_elem, 1);
    }

    // Note: We can't merge the for loop into the for loop above because
    // WeightSetVisitor also sets the parameter/weights of the inner modules.
    // Inner Modules are held by the parent module e.g. the concat module can
    // hold various other modules.
    for (size_t i = 0, offset = parameterOffset; i < network.size(); ++i)
    {
      offset += boost::apply_visitor(WeightSetVisitor(std::move(parameter),
          offset), network[i]);

      boost::apply_visitor(resetVisitor, network[i]);
    }
}
```

Afterwards we can call the function like:

```c++
// Reset the network parameter with the given initialization rule.

parameter.set_size(generatorWeights + discriminatorWeights, 1);

NetworkInitialization<InitializationRuleType> networkInit(initializeRule);
networkInit.Initialize(generator, parameter);

networkInit.Initialize(discriminator, parameter, generatorWeights);
```

src/mlpack/tests/gan_test.cpp
```diff
@@ -0,0 +1,76 @@
...
+{
+  size_t hiddenLayerSize1 = 500;
+  size_t hiddenLayerSize2 = 500;
+  size_t gOutputSize = 784;
+  size_t gInputSize = 100;
+  size_t dOutputSize = 1;
+  size_t batchSize = 100;
+  // Load the dataset
+  arma::mat trainData, dataset;
+  trainData.load("train4.txt");
```
zoq: Can you make that file available somehow?

src/mlpack/methods/ann/gan_impl.hpp
```diff
@@ -0,0 +1,332 @@
...
+    {
+      trainGenerator = false;
+      offset = (offset + batchSize) % trainData.n_cols;
+      GenerateData();
+      iterationDiscriminator--;
+    }
+  }
+  iterations++;
+  std::cout << "gradient generator" << std::endl;
+  gradient.rows(0, 5).print();
```
zoq: Have you checked the full gradient? We should make sure that each parameter is updated e.g. by setting the gradient to e.g. ``DBL_MAX`` and afterwards we could check if some of the gradient are still ``DBL_MAX``.

src/mlpack/tests/gan_test.cpp
```diff
@@ -0,0 +1,76 @@
...
+using namespace std::placeholders;
+
+BOOST_AUTO_TEST_SUITE(GANNetworkTest);
+
+
+BOOST_AUTO_TEST_CASE(GanTest)
+{
+  size_t hiddenLayerSize1 = 500;
+  size_t hiddenLayerSize2 = 500;
+  size_t gOutputSize = 784;
```
zoq: Maybe we should test it on something smaller e.g. mnist?  What do you think?

src/mlpack/methods/ann/gan_impl.hpp
kris-singh: I am not entirely sure about this comment. In the present initialization strategy, we use Gaussian initialisation. We initialize the whole parameters of the GAN(generator + discriminator) and then initialise the generator parameters by making them use the parameters of the GAN network using the arma::mat constructor. I don't get point about changing of size here.

kris-singh:
The file is just the mnist dataset where i have extracted all 4 digits. Here is the file 
[train4.txt.zip](https://github.com/mlpack/mlpack/files/1206111/train4.txt.zip)

src/mlpack/methods/ann/gan_impl.hpp
zoq: Right, currently you use the gaussian initialization strategy for all network parameter, but usually you like to use a distribution with with zero mean and unit variance for each layer not for the the complete network. If we switch to ``NetworkInitialization`` we use the initialization rule for each layer. Does this make sense?

kris-singh: But we could it do it another way around as well right we can initialise the generator and discriminator parameters and then make ` parameters.cols(0, gan.Parameters().n_rows) = arma::mat(generator.Parameters(), 0, ....). Hmm, i am not entirely sure that is safe. we do not change the memptr() of generator.Parmeters() as such anywhere in the code.

zoq: The problem with the other way around is, there is no easy way to return the joint parameter of the generator and discriminator using ``memptr()``.

kris-singh: I tried the strategy that you stated above. The problem seems to be that both Network parameters ie Generator network and Discriminator network are initialized to the same value.

zoq: Maybe I missed something, can you post the modification somewhere?

src/mlpack/methods/ann/gan_impl.hpp
```diff
@@ -0,0 +1,332 @@
...
+  discriminator.Gradient(parameters, i, gradientDiscriminator);
+  if (trainGenerator)
+  {
+    // pass the error from discriminator to generator
+    generator.error = boost::apply_visitor(deltaVisitor,
+        discriminator.network[1]);
+    generator.Backward();
+    generator.ResetGradients(gradientGenerator);
+    // set the current input requrired for gradient computation
+    generator.currentInput = predictors.col(i);
```
lozhnikov: I think this is incorrect since the generator accepts noise data as input rather than dataset vectors.

src/mlpack/methods/ann/gan_impl.hpp
```diff
@@ -0,0 +1,332 @@
...
+    generator.Gradient();
```
lozhnikov: It seems the formula doesn't correspond to the paper.
If I am not mistaken `discriminator.responses[i] = 1`. Since you use `CrossEntropyLayer` as the error function the gradient is equal to the gradient of `-log(D(G(noise)))` (since the cross entropy error is equal to 
`-[y log x + (1 - y) log (1 - x)]`,
where `y` is the response and `x` is equal to the output of the discriminator) while as the paper states the gradient should be equal to `log(1 - D(G(noise)))`.

src/mlpack/methods/ann/gan_impl.hpp
kris-singh: Actually maybe this could because of confusing name but when trainGenerator = true then the GenerateData function actually makes    noiseData = arma::mat(predictors.memptr(), noiseInSize,
+        batchSize, false, false); and noiseData function is filled by GenerateNoise function.

src/mlpack/methods/ann/gan_impl.hpp
kris-singh: I am just trying to make sense of the other places that i have read where Generator network lables are = 1. Which makes intuitive sense to me as the generator is trying to fool the discriminator why would the generator itself say that the data is fake.
The best i was able to get was this https://github.com/soumith/ganhacks Point 2 which basically says that you maximise the log D(G(z)) which is equivalent to min - log(D(G(z)) so i think the current formulation is better ie is more intutive and it gets rid of the vanishing gradient part as stated by chintala et al

kris-singh:
@zoq  I have added the method of Initialization as you had suggested but the problem is that now both parameters of the `Generator` and `Discriminator` Network are initialized to the same values.

src/mlpack/methods/ann/init_rules/network_init.hpp
```diff
@@ -52,7 +52,8 @@ class NetworkInitialization
    * @param network Network that should be initialized.
    * @param parameter The network parameter.
    */
-  void Initialize(const std::vector<LayerTypes>& network, arma::mat& parameter)
+  void Initialize(const std::vector<LayerTypes>& network, arma::mat& parameter,
+      size_t parameterOffset = 0)
   {
     // Determine the number of parameter/weights of the given network.
     size_t weights = 0;
```
zoq: We should do the following only if ``parameter.is_empty()`` is true, that way we use the specified ``parameter`` matrix and don't allocated a new one every time we call ``Initialize``.

src/mlpack/methods/ann/gan_impl.hpp
lozhnikov: Sorry for the slow response. I don't assert that you did an error since
the sign of `grad[-log(D(G(noise)))]` is equal to the sign of `grad[log(1 - D(G(noise)))]`. I just pointed out that the formula doesn't match. I think we should test both approaches and compare them.
P.S. If you implement something according to another source (i.e. not the GAN papers) you should point that out in the code (i.e. provide the link), otherwise it is hard to verify the code. Moreover, the authors didn't provide any test that verify their approach. Hence, as for me the statement is doubtful and needs verification.

src/mlpack/methods/ann/gan_impl.hpp
```diff
@@ -0,0 +1,295 @@
...
+    discWeights += boost::apply_visitor(weightSizeVisitor, 
+        discriminator.network[i]);
+
+  parameter.set_size(genWeights + discWeights, 1);
+
+  // Intialise the parmaeters generator
+  networkInit.Initialize(generator.network, parameter);
+  // Intialise the parameters discriminator
+  networkInit.Initialize(discriminator.network, parameter, genWeights);
+  generator.Parameters() = arma::mat(
```
zoq: Have you checked that the weights from the generator/discriminator are shared with the parameter matrix in the GAN class? The generator/discriminator should call the ResetParameters function on the first Evaluation call, since reset is false, so you probably end up with different weights.

What you could do here is:

```c++
generator.ResetParameters();
discriminator.ResetParameters();

size_t genWeights = generator.Parameters().n_elem;
size_t discWeights = discriminator.Parameters().n_elem;

parameter.set_size(generator.Parameters().n_elem +
    discriminator.Parameters().n_elem, 1);

NetworkInitialization<InitializationRuleType> networkInit(initializeRule);
networkInit.Initialize(generator.network, parameter);
networkInit.Initialize(discriminator.network, parameter, generator.Parameters().n_elem);
```

lozhnikov:
I refactored GAN. The code is available here https://github.com/lozhnikov/mlpack/tree/gan . And I wrote a simple test that trains GAN and generates samples. The test is available here [gan-test.zip](https://github.com/mlpack/mlpack/files/1219938/gan-test.zip) . I tried to test the implementation on the Digit dataset and got these samples 
digits -v -s -e 500 -r 0 1 -g 2 -n 32 -b 10 -g 128 -d 32 -t 0
the first row is obtained from the generator and the second row contains some samples from the Digit dataset. In order to obtain these results run the test with `-i path/to/the/digit/dataset -o path/to/output -v -s -e 500 -r 0 1 -g 2 -n 32 -b 10 -g 128 -d 32 -t 0` on the Digit dataset.

Unfortunately, I am unable to get suitable results on the MNIST dataset. I noticed that the network heavily depends on parameters (Layer types, the number of units and so on). So, maybe I did an error in the implementation, maybe some parameters are incorrect. I think it is worth to find a working example, try to run it on the MNIST dataset (some small part of the MNIST dataset) and if the example shows good results then run our implementation with the same layers and the same arguments.

P.S. Since you changed `minibatch-SGD` in the latest commit and I couldn't compile your code, I did the branch from the previous commit (1cd793844f708eca53edaa13def594e3b8b78e6d). So, do
`git reset --hard 1cd793844f708eca53edaa13def594e3b8b78e6d`
before merge.

kris-singh:
Sorry for the late reply. Actually just went through the changes I think i would have to manually merge them as I in the last commit I had made some changes to the CreateBatch function and the CreateNoise function I will look through in more detail and run this on the mnist dataset right now for mnist with non cnn based architecture i can only think of this [example](https://www.analyticsvidhya.com/blog/2017/06/introductory-generative-adversarial-networks-gans). I was thinking that we should extend this to work with CNN layers also let me know what you think on that.

kris-singh:
Just a few questions on the implementation. Now we fix the predictors matrix to have `n - 1` cols of the traindata and one column of the fake/noise data. We also now train the generator network only one sample of noise. Is that correct, if so this does not comply with the original Goodfellow's paper?

lozhnikov:
> I was thinking that we should extend this to work with CNN layers also let me know what you think on that.

Looks reasonable to me. I ran this example (https://www.oreilly.com/learning/generative-adversarial-networks-for-beginners , the code is available here https://github.com/jonbruner/generative-adversarial-networks) ten hours ago. The example should finish in the evening or tonight. I'll tell you the results.

> Now we fix the predictors matrix to have n - 1

`discriminator.predictors` contains the whole train data and a column of generated data which I update each `Gradient()` call.

kris-singh:
But shouldn't the number of samples for training the generator and discriminator should be m samples.  For discriminator it would mean that there are m/2 samples of real data and m/2 samples of fake data. So i did not understand your logic behind just using 1 sample of fake data for both the discriminator and generator. 
PS. If it is giving better results then i think i can't argue with that but just wanted to know if there is anything more to it.

lozhnikov:
> So i did not understand your logic behind just using 1 sample

> `discriminator.predictors` contains the whole train data and a column of generated data **which I update each Gradient() call**

I resample noise variables each time.

kris-singh:
Can you explain the evaluate function. Now the evaluate function evaluates in every iteration on the both current input and the noise input why?
Also, same question for the discriminator if this is sgd we should only evaluate the gradient at the 
One point why evaluate it on the noise as well.

Was -log(D(G(noise)) not giving good results ?? That you switched the code.
I am confused why did you change the sign of the gradient int the calculation of the generator gradient if we are trying to minimize log(1 - D(z)) Can you explain that.
I rather propose that we do something like this :
```
arma::mat noiseDataDiscriminator;
discriminator.predictors.set_size(predictors.n_rows, predictors.n_cols + 	noiseDataDiscriminator.n_cols);
discriminator.predictors.cols(0, predictors.n_cols - 1) = predictors;
discriminator.predictors.cols(predictors.n_cols, noiseDataDiscriminator.n_cols - 1) = noiseData;
discriminator.responses.set_size(1, predictors.n_cols + noiseDataDiscriminator.n_cols);
discriminator.responses.ones();
  discriminator.responses.cols(predictors.n_cols, noiseDataDiscriminator.n_cols - 1).zeros();
```
With this evaluate function changes to
```
if (!reset)
    Reset();
  currentInput = this->predictors.unsafe_col(i);
  currentTarget = this->responses.unsafe_col(i);
  discriminator.Forward(std::move(currentInput));
  double res = discriminator.outputLayer.Forward(
      std::move(boost::apply_visitor(
          outputParameterVisitor, 
          discriminator.network.back())), std::move(currentTarget));
return res
```
The gradient function would change accordingly 
```
  static bool print = true;
  if (!reset)
    Reset();

  if (gradient.is_empty())
  {
    if (parameter.is_empty())
    {
      Reset();
    }
    gradient = arma::zeros<arma::mat>(parameter.n_elem, 1);
  }
  else
  {
    gradient.zeros();
  }

  if (noiseGradientDiscriminator.is_empty())
  {
    noiseGradientDiscriminator = arma::zeros<arma::mat>(gradientDiscriminator.n_elem, noiseData.n_cols);
  }
  else
  {
    noiseGradientDiscriminator.zeros();
  }

  gradientGenerator = arma::mat(gradient.memptr(),
      generator.Parameters().n_elem, 1, false, false);

  gradientDiscriminator = arma::mat(gradient.memptr() + 
      gradientGenerator.n_elem,
      discriminator.Parameters().n_elem, 1, false, false);

  // get the gradients of the discriminator
  discriminator.Gradient(discriminator.parameter, i, gradientDiscriminator);

  if (currentBatch % generatorUpdateStep == 0 &&/* counter % batchSize == 0 &&*/ currentBatch != 0)
  {
    // generate noise here
    predictors.cols(predicts.n_cols, noiseData.n_cols - 1).imbue(…….)
    // Minimize log(1 - D(G(noise)))
    // pass the error from discriminator to generator
    generator.error = boost::apply_visitor(deltaVisitor,
        discriminator.network[1]);
    generator.currentInput = predictors.cols(i);
    generator.Backward();
    generator.ResetGradients(gradientGenerator);
    // set the current input requrired for gradient computation
    generator.Gradient();

    double multiplier = 1.0;
    // not sure here
    gradientGenerator = -gradientGenerator;
    gradientGenerator *= multiplier;

    if (counter % batchSize == 0 && print)
    {
      Log::Info << "gradientDiscriminator = " << std::max(std::fabs(gradientDiscriminator.min()), std::fabs(gradientDiscriminator.max())) << std::endl;
      Log::Info << "gradientGenerator = " << std::max(std::fabs(gradientGenerator.min()), std::fabs(gradientGenerator.max())) << std::endl;
      print = false;
    }
  }

  counter++;

// TODO: I know the code below looks like crutches.
// I think it is better to remove batch handling from GAN.
// For example we could add some functions to the Optimizer API
// in order to obtain the number of current batch.
  if (counter >= numFunctions)
  {
    counter = 0;
    currentBatch++;
    print = true;
  }
  else if (counter % batchSize == 0)
    currentBatch++;
}
```
I will have to look into the gradient function a bit more particularly the gradientGenerator part. But we now allow the user to give the noiseData so he can indeed use any noise distribution he pleases.

kris-singh:
I think your above comment does explain that by just using 1 column of noise your able to actually "simulate n noise data points'' which is really cool. But i did not understand the Evaluate function and the gradient function calculations being done for both the noise and predictors variables hence my above comments still hold. Let me know what you think.

kris-singh:
Okay i think i got what you are trying to do with. You are evaluating both the noise data and predictor at the same time for `log(D(z)) + 1 - log(D(G(z))` the same goes for the gradient function. Just one thing can you explain this 
Was -log(D(G(noise)) not giving good results ?? That you switched the code.
I am confused why did you change the sign of the gradient int the calculation of the generator gradient if we are trying to minimize log(1 - D(z)). Particularly changing the gradient direction.

lozhnikov:
1. > Can you explain the evaluate function. Now the evaluate function evaluates in every iteration on the both current input and the noise input why?

The `Evaluate()` function calculates
`-[log D(x) + log ( 1 - D(G(z)))]` i.e. the functions that we are going to minimize.

2. > Also, same question for the discriminator if this is sgd we should only evaluate the gradient at the
One point why evaluate it on the noise as well.

The paper states that the gradient of discriminator is equal to the gradient of
`-[log D(x) + log (1 - D(G(z)))]`.

3. > Was -log(D(G(noise)) not giving good results ?? That you switched the code.

I got the results above using `log(1 - D(G(z)))`. I couldn't select valid parameters for `-log(D(G(noise)))`.

4. > I am confused why did you change the sign of the gradient int the calculation of the generator gradient if we are trying to minimize log(1 - D(z)) Can you explain that.

The cross entropy function changes the sign.

5. > But we now allow the user to give the noiseData so he can indeed use any noise distribution he pleases.

I think we should change `noiseData` each time.

src/mlpack/core/optimizers/minibatch_sgd/minibatch_sgd.hpp
```diff
@@ -165,6 +166,9 @@ class MiniBatchSGDType
   //! are reset before Optimize call.
   bool& ResetPolicy() { return resetPolicy; }
 
+  //! Get the current batch being processed.
+  size_t& CurrentBatch() { return currentBatch; }
```
lozhnikov: Since the Optimizer API influence on the whole library, I think that any changes of the API require a separate discussion and consequently a separate PR. If I am not mistaken only `mini-batch SGD` supports mini-batches right now. So, I guess there is no sense to change the API until other optimizers allow to use mini-batches.

src/mlpack/core/optimizers/minibatch_sgd/decay_policies/no_decay.hpp
```diff
@@ -32,13 +32,12 @@ class NoDecay
...
-  * @param gradient The gradient matrix.
-  */
-  void Update(arma::mat& /* iterate */,
-              double& /* stepSize */,
-              const arma::mat& /* gradient */)
+   * @param stepSize Step size to be used for the given iteration.
+   * @param epoch Size of one epoch
+   * @param decaySteps decay learning after these many steps
+   */
+  void Update(double& /*stepSize*/,
```
lozhnikov: The same comment here. I think these changes need a separate discussion. That's why I removed that commit and did a branch from the previous one.

src/mlpack/methods/ann/gan_impl.hpp
```diff
@@ -0,0 +1,255 @@
...
+
+
+template<typename Model, typename InitializationRuleType, typename NoiseType>
+template<typename OptimizerType>
+void GAN<Model, InitializationRuleType, NoiseType>::Train(
+    OptimizerType& Optimizer)
+{
+  if (!reset)
+    Reset();
+  currentBatch = &Optimizer.CurrentBatch();
```
lozhnikov: Probably, it is not a good idea to use the pointer here.

src/mlpack/methods/ann/gan_impl.hpp
```diff
@@ -0,0 +1,255 @@
...
+    generator.ResetGradients(gradientGenerator);
+    // set the current input requrired for gradient computation
+    generator.Gradient();
+
+    double multiplier = 1.0;
+
+    gradientGenerator = -gradientGenerator;
+    gradientGenerator *= multiplier;
+
+    if (counter % batchSize == 0)
```
lozhnikov: I suggest to restore batch handling here (see the comment above) until the rest of optimizers get refactored.

lozhnikov:
The example that I pointed out above (https://www.oreilly.com/learning/generative-adversarial-networks-for-beginners) shows the following results.
samples
So, I think we could use the same arguments and the same layers in order to test our implementation.
Next, I'll test the example that you mentioned (without convolutional layers) (https://www.analyticsvidhya.com/blog/2017/06/introductory-generative-adversarial-networks-gans).

kris-singh:
I already implmented that example in the gan_test.cpp file. You can have a look there.

lozhnikov:
I want to test the original example. Maybe it doesn't work at all, in that case there is no sense to implement that.

src/mlpack/methods/ann/image_functions/bilinear_function.hpp
```diff
@@ -0,0 +1,194 @@
...
+   * DownSample the given input.
+   *
+   * @param input The input matrix
+   * @param output The resulting down-sampled output image.
+   */
+  template<typename eT>
+  void DownSample(const arma::Mat<eT>& input, arma::Mat<eT>& output)
+  {
+    if (output.is_empty())
+      output.set_size(inRowSize * inColSize, 1);
```
lozhnikov: That's incorrect. You should write
`output.set_size(inRowSize * inColSize * depth, 1);`

src/mlpack/methods/ann/image_functions/bilinear_function.hpp
```diff
@@ -0,0 +1,194 @@
...
+   * UpSample the given input.
+   *
+   * @param input The input matrix
+   * @param output The resulting interpolated output image.
+   */
+  template<typename eT>
+  void UpSample(const arma::Mat<eT>& input, arma::Mat<eT>& output)
+  {
+    if (output.is_empty())
+      output.set_size(outRowSize * outColSize, 1);
```
lozhnikov: That's incorrect. You should write
`output.set_size(outRowSize * outColSize * depth, 1);`

src/mlpack/methods/ann/image_functions/bilinear_function.hpp
kris-singh: Yup i changed it in the local branch. Will push it soon.

src/mlpack/core/optimizers/minibatch_sgd/minibatch_sgd_impl.hpp
```diff
@@ -92,6 +92,10 @@ double MiniBatchSGDType<
       Log::Info << "Mini-batch SGD: iteration " << i << ", objective "
           << overallObjective << "." << std::endl;
 
+      Log::Info << "Last objective" << lastObjective << std::endl;
+
+      Log::Info << "Gradient = " << arma::max(gradient) << std::endl;
```
lozhnikov: I understand you need that in order to debug the implementation, but we should remove that in the future.

src/mlpack/methods/ann/gan.hpp
```diff
@@ -0,0 +1,187 @@
...
+ * http://www.opensource.org/licenses/BSD-3-Clause for more information.
+ */
+#ifndef MLPACK_METHODS_ANN_GAN_HPP
+#define MLPACK_METHODS_ANN_GAN_HPP
+
+#include <mlpack/core.hpp>
+
+#include <mlpack/methods/ann/layer/layer.hpp>
+#include <mlpack/methods/ann/layer/base_layer.hpp>
+#include <mlpack/methods/ann/ffn.hpp>
```
lozhnikov: Correct if I am mistaken, but it seems you needn't other includes except this one.

src/mlpack/methods/ann/gan.hpp
```diff
@@ -0,0 +1,187 @@
...
+#include <mlpack/methods/ann/visitor/reset_visitor.hpp>
+#include <mlpack/methods/ann/visitor/weight_size_visitor.hpp>
+#include <mlpack/methods/ann/visitor/weight_set_visitor.hpp>
+#include <mlpack/methods/ann/init_rules/random_init.hpp>
+
+using namespace mlpack;
+using namespace mlpack::ann;
+using namespace mlpack::optimization;
+using namespace mlpack::math;
+using namespace mlpack::distribution;
```
lozhnikov: Could you remove `using namespace`s? Usually we don't use them outside of tests.

src/mlpack/methods/ann/gan.hpp
```diff
@@ -0,0 +1,187 @@
...
+
+namespace mlpack {
+namespace ann /** artifical neural network **/ {
+template<
+typename Model,
```
lozhnikov: Pedantic style issue: incorrect spacing.

src/mlpack/methods/ann/gan.hpp
```diff
@@ -0,0 +1,187 @@
...
+   * @tparam Model The class type of generator and discriminator.
+   * @tparam InitializationRuleType Type of Intializer.
+   * @param generator Generator network.
+   * @param trainData The real data.
+   * @param noiseData The data generated from randomly.
+   * @param discriminator Discriminator network.
+   * @param batchSize BatchSize to be used for training.
+   * @param noiseInSize Input size of the generator network.
+   * @param disIteration Ratio of number of training step for Disc to Gen
+   */
```
lozhnikov: The comment is outdated.

src/mlpack/methods/ann/gan.hpp
```diff
@@ -0,0 +1,187 @@
...
+
+  //! Return the number of separable functions (the number of predictor points).
+  size_t NumFunctions() const { return numFunctions; }
+
+  //! Serialize the model.
+  template<typename Archive>
+  void Serialize(Archive& ar, const unsigned int /* version */);
+
+ private:
+  //! Locally stored parameter for training data.
```
lozhnikov: The comment is incorrect.

src/mlpack/methods/ann/gan.hpp
```diff
@@ -0,0 +1,187 @@
...
+  Noise noiseFunction;
+  size_t noiseDim;
+  //! Locally stored number of data points.
+  size_t numFunctions;
+
+  //! Locally stored batch size parameter.
+  size_t batchSize;
+
+  //! Locally stored offset for predictors and noise data.
+  size_t offset;
```
lozhnikov: I thought we don't use this variable anymore.

src/mlpack/methods/ann/gan.hpp
```diff
@@ -0,0 +1,187 @@
...
+  //! Locally-stored reset visitor.
+  ResetVisitor resetVisitor;
+  //! Locally stored gradient parameters.
+  arma::mat gradient;
+  //! Locally stored gradient for discriminator.
+  arma::mat gradientDiscriminator;
+
+  arma::mat noiseGradientDiscriminator;
+
+  arma::mat noise;
```
lozhnikov: It seems the variable is not used.

src/mlpack/methods/ann/gan_impl.hpp
```diff
@@ -0,0 +1,269 @@
...
+#include <mlpack/methods/ann/layer/layer.hpp>
+#include <mlpack/methods/ann/layer/base_layer.hpp>
+#include <mlpack/methods/ann/ffn.hpp>
+#include <mlpack/methods/ann/init_rules/network_init.hpp>
+#include <mlpack/methods/ann/visitor/output_parameter_visitor.hpp>
+
+#include <mlpack/methods/ann/activation_functions/softplus_function.hpp>
+#include <mlpack/methods/ann/init_rules/gaussian_init.hpp>
+#include <mlpack/core/dists/gaussian_distribution.hpp>
+#include <mlpack/methods/ann/init_rules/random_init.hpp>
```
lozhnikov: I think some includes aren't needed.

src/mlpack/methods/ann/gan_impl.hpp
```diff
@@ -0,0 +1,269 @@
...
+
+using namespace mlpack;
+using namespace mlpack::ann;
+using namespace mlpack::optimization;
+using namespace mlpack::math;
+using namespace mlpack::distribution;
```
lozhnikov: Could you remove that?

src/mlpack/methods/ann/gan_impl.hpp
```diff
@@ -0,0 +1,269 @@
...
+    discriminator(discriminator),
+    initializeRule(initializeRule),
+    noiseFunction(noiseFunction),
+    noiseDim(noiseDim),
+    batchSize(batchSize),
+    generatorUpdateStep(generatorUpdateStep),
+    preTrainSize(preTrainSize),
+    reset(false)
+{
+  // Insert IdentityLayer for Joing the generator and discriminator
```
lozhnikov: Looks like a typo (`Joing`).

src/mlpack/methods/ann/gan_impl.hpp
```diff
@@ -0,0 +1,269 @@
...
+    discWeights += boost::apply_visitor(weightSizeVisitor,
+        discriminator.network[i]);
+
+  parameter.set_size(genWeights + discWeights, 1);
+  generator.Parameters() = arma::mat(parameter.memptr(), genWeights, 1, false,
+      false);
+  discriminator.Parameters() = arma::mat(parameter.memptr() + genWeights,
+      discWeights, 1 , false, false);
+
+  // Intialise the parmaeters generator
```
lozhnikov: Looks like a typo (`parmaeters`).

src/mlpack/methods/ann/gan_impl.hpp
```diff
@@ -0,0 +1,269 @@
...
+  discriminator.Forward(std::move(currentInput));
+  double res = discriminator.outputLayer.Forward(
+      std::move(boost::apply_visitor(
+          outputParameterVisitor,
+          discriminator.network.back())), std::move(currentTarget));
+  noise.imbue( [&]() { return noiseFunction();} );
+  generator.Forward(std::move(noise));
+  arma::mat temp = boost::apply_visitor(
+      outputParameterVisitor, generator.network.back());
+  temp.reshape(discriminator.predictors.n_rows, 1);
```
lozhnikov: I thought the shape is correct without this code, isn't it?.

src/mlpack/methods/ann/gan_impl.hpp
```diff
@@ -0,0 +1,269 @@
...
+    discriminator.responses(numFunctions) = 1;
+    discriminator.Gradient(discriminator.parameter, numFunctions,
+      noiseGradientDiscriminator);
+    generator.error = boost::apply_visitor(deltaVisitor,
+        discriminator.network[1]);
+    generator.currentInput = noise;
+    generator.Backward();
+    generator.ResetGradients(gradientGenerator);
+    generator.Gradient();
+    double multiplier = 1.5;
```
lozhnikov: I think this variable could be useful if we want to set different step sizes. In that case it is better to add an argument to the constructor. How do you think?

src/mlpack/methods/ann/gan_impl.hpp
```diff
@@ -0,0 +1,269 @@
...
+  }
+  else if (counter % batchSize == 0)
+  {
+    currentBatch++;
+    if (preTrainSize > 0)
+    {
+      preTrainSize--;
+    }
+    if (preTrainSize)
+      Log::Info << "Pre Training batch " << preTrainSize << std::endl;
```
lozhnikov: I leave a comment as a reminder. I guess we should remove that in the future.

src/mlpack/methods/ann/gan_impl.hpp
```diff
@@ -0,0 +1,269 @@
...
+  }
+}
+
+template<typename Model, typename InitializationRuleType, typename Noise>
+void GAN<Model, InitializationRuleType, Noise>::Forward(arma::mat&& input)
```
lozhnikov: I am not sure of the functions below. Do we need them?

src/mlpack/methods/ann/image_functions/bilinear_function.hpp
```diff
@@ -0,0 +1,199 @@
...
+            cOrigin = inColSize - 2;
+
+          double deltaR = i * scaleRow - rOrigin;
+          double deltaC = j * scaleCol - cOrigin;
+          coeff1 = (1 - deltaR) * (1 - deltaC);
+          coeff2 = deltaR * (1 - deltaC);
+          coeff3 = (1 - deltaR) * deltaC;
+          coeff4 = deltaR * deltaC;
+
+          double depth = k * inRowSize * inColSize;
```
lozhnikov: That's not an error but I suggest to rename the variable. I think you could save some space if you define `ptr = k * inRowSize * inColSize + cOrigin * inColSize + rOrigin`. In that case
```
output(k * outRowSize * outColSize + j * outColSize + i) =
              input(ptr) * coeff1 +
              input(ptr + 1) * coeff2 +
              input(ptr + inColSize) * coeff3 +
              input(ptr + inColSize + 1) * coeff4;
```
Could you make these changes in a separate commit in order to cherry pick them to the ResizeLayer branch?

src/mlpack/methods/ann/gan.hpp
kris-singh: It is still used. Look at the evaluate function and the gradient function.

src/mlpack/methods/ann/gan_impl.hpp
kris-singh: Could be required for some cases when extending Gan's . Not required now though.

kris-singh:
Fixed your comments. I have cherry picked the commit to the Resize Layer and would add the make the changes there.

ShikharJ:
Thanks @kris-singh and @lozhnikov for your contribution, this has been incorporated in #1204! @zoq This can be closed now.

zoq:
I agree with @ShikharJ thanks for the awesome work.

Issue #1045, opened 2017-07-02
Require Armadillo 6.500 as a minimum


rcurtin:
Several things need to happen for this:

- [x] Update README.md, HISTORY.md, and any documentation in doc/
- [x] Remove any unnecessary backports from `src/mlpack/core/arma_extend` (I think this will be just about everything)
- [x] Update Jenkins build matrix to remove builds against older Armadillo versions
- [x] Update Armadillo EPEL packages (only I can do this)

Rajiv2605:
I would like to contribute to this. It would be great if you could help me as I am new to this library and to open source. It would be really helpful if you could guide me through this.

zoq:
Hello @Rajiv2605, you can start with the first point, search through the codebase (``README.md``, ``HISTORY.md``, https://github.com/mlpack/mlpack/tree/master/doc) and search for ``Armadillo  4.200.0`` or something along the lines and update the line according to the new version.

Rajiv2605:
I have updated README.md, HISTORY.md and /doc folder files with the latest version of Armadillo. I have submitted a pull request **Updated Armadillo version to 6.500.0 #1050**. Please review it and suggest changes if any.

rcurtin:
I filed a bug on Launchpad to update the Armadillo version on trusty: https://bugs.launchpad.net/bugs/1703454 .

rcurtin:
Trying again with a different bug, https://bugs.launchpad.net/trusty-backports/+bug/1703626

Rajiv2605:
@rcurtin @zoq Should I close this issue?

Rajiv2605:
@rcurtin @zoq The wiki still shows minimum version for Armadillo as `4.200.0`
This is the screenshot of the wiki:
<img width="516" alt="version" src="https://user-images.githubusercontent.com/16632603/30323677-bb24d44e-97db-11e7-995d-64d1e3c1d252.PNG">

zoq:
Nice catch! I think, instead of updating the page: https://github.com/mlpack/mlpack/wiki we can remove the part you pointed out. The main readme or the build tutorial provides that kind of information so it's redundant; for me, it's just another page we have to update if something changed.

Rajiv2605:
@zoq But, shouldn't the page wiki have all the info about the repository?

zoq:
I think it does, but instead of providing the information on the same page, we reference the existing page that provides all the information about building mlpack, dependencies, etc. Let me know what you think.

[...]

PR #1067, opened 2017-07-20
Memory head of NTM module

sumedhghaisas:
Memory Head will be layer used in NTM module to get the weights with which the memory will be read or written.


sumedhghaisas:
Some minor improvisations are yet to be done. I will keep updating the PR with those. The main problem I see right now is the Armadillo version. I am using latest arma functions such as arma::shift which is essential for the circular convolution. Can we bump up the version of Armadillo?

zoq:
unfortunately that's something we can't do right now; here are some pointers:

- https://github.com/mlpack/mlpack/issues/1045
- https://github.com/mlpack/mlpack/pull/1041#pullrequestreview-46955870

So, we could backport the ``shift`` function, let us know what you think.

rcurtin:
I think `shift()` would be straightforward to backport; let me know if you need any help doing that.

PR #1068, opened 2017-07-20
Refactor NaiveBayesClassifier to have templated Train() functions.

rcurtin:
This is an alternate solution to #1051.  It actually allows more flexibility---now I can pass in subviews, etc. to `NaiveBayesClassifier::Train()`, and I can still use `NaiveBayesClassifier` with non-double types (i.e. `NaiveBayesClassifier<float>`.  Since all overloads of `Train()` are templatized, this should satisfy the conditions needed for the hyper-parameter tuner.

@micyril: what do you think?  Is this an acceptable alternate approach?

### DIFF
--- a/src/mlpack/methods/naive_bayes/naive_bayes_classifier.hpp
+++ b/src/mlpack/methods/naive_bayes/naive_bayes_classifier.hpp
@@ -43,11 +43,20 @@ namespace naive_bayes /** The Naive Bayes Classifier. */ {
  *
  * nbc.Classify(testing_data, results);
  * @endcode
+ *
+ * The ModelMatType template parameter specifies the internal matrix type that
+ * NaiveBayesClassifier will use to hold the model.  This can be arma::mat,
+ * arma::fmat, or any other Armadillo (or Armadillo-compatible) object.
+ *
+ * @tparam ModelMatType Internal matrix type to use to store the model.
  */
-template<typename MatType = arma::mat>
+template<typename ModelMatType = arma::mat>
 class NaiveBayesClassifier
 {
  public:
+  // Convenience typedef.
+  typedef typename ModelMatType::elem_type ElemType;
+
   /**
    * Initializes the classifier as per the input and then trains it by
    * calculating the sample mean and variances.
@@ -66,6 +75,7 @@ class NaiveBayesClassifier
    *     calculate the variance; this can prevent loss of precision in some
    *     cases, but will be somewhat slower to calculate.
    */
+  template<typename MatType>
   NaiveBayesClassifier(const MatType& data,
                        const arma::Row<size_t>& labels,
                        const size_t classes,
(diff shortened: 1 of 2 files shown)


micyril:
Yes, it should work.

src/mlpack/methods/naive_bayes/naive_bayes_classifier.hpp
```diff
@@ -43,11 +43,20 @@ namespace naive_bayes /** The Naive Bayes Classifier. */ {
  *
  * nbc.Classify(testing_data, results);
  * @endcode
+ *
+ * The ModelMatType template parameter specifies the internal matrix type that
+ * NaiveBayesClassifier will use to hold the model.  This can be arma::mat,
+ * arma::fmat, or any other Armadillo (or Armadillo-compatible) object.
```
zoq: hm, I think we should make it clear why we do this here, the PR summary sounds for me more informative.

src/mlpack/methods/naive_bayes/naive_bayes_classifier.hpp
```diff
@@ -169,59 +182,51 @@ class NaiveBayesClassifier
    * @param probabilities This will be filled with class probabilities for each
    *      point. Each row represents a point.
    */
+  template<typename MatType>
   void Classify(const MatType& data,
                 arma::Row<size_t>& predictions,
-                arma::mat& probabilities) const;
+                ModelMatType& probabilities) const;
```
zoq: Should we introduce a template parameter for the probabilities, I guess a user might ask why we introduced ``ProbabilitiesVecType`` for:

```
template<typename VecType, typename ProbabilitiesVecType>
void Classify(const VecType& point,
              size_t& prediction,
              arma::vec& probabilities) const;
              ProbabilitiesVecType& probabilities) const;
```

but not for the matrix version?

rcurtin: I think that `ProbabilitiesVecType` should always have the same type as the underlying `ModelMatType`, but I think it is fine to templatize it.  I'll push changes soon.

src/mlpack/methods/naive_bayes/naive_bayes_classifier.hpp
rcurtin: I added some more documentation, what do you think?

rcurtin:
Yeah, an unfortunate side effect is that the code gets more complex.  Do you think this PR is reasonable, or should more documentation be added? I can't think of anything else to help other than documentation; maybe there is some other cleaner design...?

zoq:
I think it's the cleanest we can come up with, that supports everything we like to see; no more comments from my side, looks good.

rcurtin:
Ok, in this case I will merge this in 3 days.

PR #1070, opened 2017-07-21
[GSoC] Adding optimization features not related with my GSoC project

partobs-mdp:

Implemented:

- `CrossEntropyLayer` for evaluating the performance of the model on binary vector targets.
- Initial version of the gradient clipping (albeit very dirty, as @zoq and @rcurtin have already mentioned in #1005)

As far as I understand, the conversation related to these two points (including but not limited to the reusable update API for gradient clipping) is transferred here.


src/mlpack/tests/sgd_test.cpp
```diff
@@ -39,6 +41,23 @@ BOOST_AUTO_TEST_CASE(SimpleSGDTestFunction)
   BOOST_REQUIRE_SMALL(coordinates[2], 1e-7);
 }
```
zoq: Might be a good idea to test the policy independently from the optimizer, we could simply test if the gradient is clipped correctly.

src/mlpack/methods/ann/layer/cross_entropy_error_impl.hpp
```diff
@@ -0,0 +1,58 @@
...
+}
+
+template<typename InputDataType, typename OutputDataType>
+template<typename eT>
+void CrossEntropyError<InputDataType, OutputDataType>::Backward(
+    const arma::Mat<eT>&& input,
+    const arma::Mat<eT>&& target,
+    arma::Mat<eT>&& output)
+{
+  output = (1. - target) / (1. - input + 1e-2) - target / (input + 1e-2);
```
zoq: ``1e-2`` is somewhat hight, I would use something like ``1e-10``.

src/mlpack/methods/ann/layer/cross_entropy_error.hpp
```diff
@@ -0,0 +1,100 @@
...
+ * @tparam InputDataType Type of the input data (arma::colvec, arma::mat,
+ *         arma::sp_mat or arma::cube).
+ * @tparam OutputDataType Type of the output data (arma::colvec, arma::mat,
+ *         arma::sp_mat or arma::cube).
+ */
+template <
+    typename InputDataType = arma::mat,
+    typename OutputDataType = arma::mat
+>
+class CrossEntropyError
```
zoq: Can we add a simple test for the cross entropy error function?

src/mlpack/core/optimizers/sgd/update_policies/gradient_clipping.hpp
```diff
@@ -0,0 +1,96 @@
...
+   * Constructor for creating a GradientClipping instance.
+   * 
+   * @param minGradient Minimum possible value of gradient element.
+   * @param maxGradient Maximum possible value of gradient element.
+   * @param updatePolicy An instance of the UpdatePolicy
+   *                     used for actual optimization.
+   */
+  GradientClipping(const double minGradient,
+                   const double maxGradient,
+                   UpdatePolicy updatePolicy) :
```
partobs-mdp: Likewise, it looks reasonable to add `MinGradient` and `MaxGradient` methods (done in the last commit)

src/mlpack/core/optimizers/sgd/update_policies/gradient_clipping.hpp
```diff
@@ -0,0 +1,96 @@
...
+   * @param iterate Parameters that minimize the function.
+   * @param stepSize Step size to be used for the given iteration.
+   * @param gradient The gradient matrix.
+   */
+  void Update(arma::mat& iterate,
+              const double stepSize,
+              const arma::mat& gradient)
+  {
+    // First, clip the gradient.
+    gradient.transform(
```
partobs-mdp: Didn't know about `clamp`, implementing. By the way, it also resolves the issue with const reference.

partobs-mdp: Speaking about performance, they should have the comparable performance, as they're both element-wise operations (if `transform` is parallelized, which I think is true)

src/mlpack/tests/sgd_test.cpp
partobs-mdp: I also thought about this, but how could we extract the gradient history from SGD optimizer object?

src/mlpack/methods/ann/layer/cross_entropy_error_impl.hpp
partobs-mdp: Introduced the `eps` parameter for handling this kind of trade-off without breaking the code.

src/mlpack/core/optimizers/sgd/update_policies/gradient_clipping.hpp
partobs-mdp: Benchmarked that one. Here's what I've got: https://gist.github.com/partobs-mdp/02b3bb93be0496ad6866528359d5ba3d

zoq: Thanks for taking a look into the issue; looks like transform is slightly faster, but I think it's 
negligible.

src/mlpack/tests/gradient_clipping_test.cpp
```diff
@@ -0,0 +1,73 @@
...
+BOOST_AUTO_TEST_CASE(ClippedVanillaUpdateTest)
+{
+  VanillaUpdate vanillaUpdate;
+  GradientClipping<VanillaUpdate> update(-3.0, +3.0, vanillaUpdate);
+  update.Initialize(3, 3);
+
+  arma::mat coordinates = arma::zeros(3, 3);
+  // Setting step = 1 to make math easy.
+  double stepSize = 1.0;
+  arma::mat dummyGradient = { {-6, +6, 0}, {1, 2, 3}, {-3, 0, +4} };
```
zoq: To make the windows build happy you could use:

```
mat input;
input << 1 << 2 << 3 << endr
      << 4 << 5 << 6 << endr
      << 7 << 8 << 9
      << ...;
```

or

```
mat input("1 2 3; \
           4 5 6 \
          ...");
```

src/mlpack/methods/ann/layer/cross_entropy_error_impl.hpp
```diff
@@ -0,0 +1,59 @@
...
+  // Nothing to do here.
+}
+
+template<typename InputDataType, typename OutputDataType>
+template<typename eT>
+double CrossEntropyError<InputDataType, OutputDataType>::Forward(
+    const arma::Mat<eT>&& input, const arma::Mat<eT>&& target)
+{
+  return -arma::accu(target % arma::log(input + eps) +
+                     (1. - target) % arma::log(1. - input + eps));
```
rcurtin: It's a little late so I'm not sure I'm thinking 100% clearly, but this appears that it will work in the multiclass setting as long as the `input` and `target` matrices are one-hot encoded.  So labels like `[0 2 1]` will not work but labels like `[[1 0 0] [0 0 1] [0 1 0]]` will.  Correct me if I am wrong.  (i.e., I think this is right and works the way I would expect it to.)

partobs-mdp: Well, it won't work as it stands, but the computations in the case you've mentioned would be easier: `-arma::accu(target % arma::log(input + eps))` (the formula get easier due to data representation redundancy)

src/mlpack/core/optimizers/sgd/sgd.hpp
```diff
@@ -132,8 +138,15 @@ class SGD
...
+  //! Get whether or not the update policy parameters
+  //! are reset before Optimize call.
+  bool ResetPolicy() const { return resetPolicy; }
+  //! Modify whether or not the update policy parameters
+  //! are reset before Optimize call.
+  bool& ResetPolicy() { return resetPolicy; }
+
   //! Get the update policy.
-  const UpdatePolicyType& UpdatePolicy() const { return updatePolicy; }
+  UpdatePolicyType UpdatePolicy() const { return updatePolicy; }
```
zoq: We should return a reference instead of a copy here.

src/mlpack/core/optimizers/sgd/update_policies/gradient_clipping.hpp
```diff
@@ -0,0 +1,96 @@
...
+   * Update step. First, the gradient is clipped, and then the actual update
+   * policy does whatever update it needs to do.
+   * 
+   * @param iterate Parameters that minimize the function.
+   * @param stepSize Step size to be used for the given iteration.
+   * @param gradient The gradient matrix.
+   */
+  void Update(arma::mat& iterate,
+              const double stepSize,
+              const arma::mat& gradient)
```
zoq: Agreed, that is another good option, @partobs-mdp what do you think?

src/mlpack/methods/ann/layer/cross_entropy_error_impl.hpp
```diff
@@ -0,0 +1,59 @@
...
+  output = (1. - target) / (1. - input + eps) - target / (input + eps);
+}
+
+template<typename InputDataType, typename OutputDataType>
+template<typename Archive>
+void CrossEntropyError<InputDataType, OutputDataType>::Serialize(
+    Archive& /* ar */,
+    const unsigned int /* version */)
+{
+  // Nothing to do here.
```
zoq: We should serialize ``eps``  here, since it's now a parameter.

src/mlpack/methods/ann/layer/cross_entropy_error_impl.hpp
```diff
@@ -0,0 +1,59 @@
...
+}
+
+template<typename InputDataType, typename OutputDataType>
+template<typename eT>
+void CrossEntropyError<InputDataType, OutputDataType>::Backward(
+    const arma::Mat<eT>&& input,
+    const arma::Mat<eT>&& target,
+    arma::Mat<eT>&& output)
+{
+  output = (1. - target) / (1. - input + eps) - target / (input + eps);
```
zoq: Not sure, if the compile would optimize that away, but we could rewrite the expression as:

``output = (target - input) / ((x - 1) % x); `` and save one extra division and the eps addition.

partobs-mdp: Checked it on a piece of paper and got the precise expression we should get after simplifying the original expression so that it would still be the gradient of the loss function with `log(x + eps)`:

```output = (input - target + eps * (1. - 2 * target)) / ((1. - input + eps) % (input + eps))```

This one, however, doesn't really optimize much (or, at least, I think so), because even though it runs only one (element-wise) division, it runs *three* multiplications, which are also slow (as compared with additions - that's why I didn't count them).

File: src/mlpack/core/optimizers/sgd/update_policies/gradient_clipping.hpp (hunk 9 above)
Comment (partobs-mdp) [inline src/mlpack/core/optimizers/sgd/update_policies/gradient_clipping.hpp]: Sure, but *should* we do this? As we have seen, the `clamp` performance is not much of an issue, and (imho) we shouldn't break the natural assumption that the update policy doesn't break the variable which stores the gradient.

File: src/mlpack/methods/ann/layer/cross_entropy_error_impl.hpp (hunk 11 above)
Comment (zoq) [inline src/mlpack/methods/ann/layer/cross_entropy_error_impl.hpp]: Right, I guess the only benefit I see is that we could avoid adding eps since: ``((x - 1) % x)`` should be stable. Anyway, I don't mind to leave it as it is.

File: src/mlpack/core/optimizers/sgd/update_policies/gradient_clipping.hpp (hunk 9 above)
Comment (zoq) [inline src/mlpack/core/optimizers/sgd/update_policies/gradient_clipping.hpp]: Right, the performance difference is insignificant, we could use ``arma::clamp`` here, no need to change it if you don't think that's a good idea. We just wanted to point out there is another solution.

**Comment by partobs-mdp** [comment]
What's wrong with the Internet connection on Travis CI? It didn't even manage to install boost from apt :'(

**Comment by zoq** [comment]
Let me restart the build.

**Comment by partobs-mdp** [comment]
Is there anything else that should be done on this PR on my side?

File: src/mlpack/tests/gradient_clipping_test.cpp [hunk 12]
```diff
@@ -0,0 +1,73 @@
...
+  arma::mat targetCoordinates("3 -3 0; -1 -2 -3; 3 0 -3;");
+  // On the first Update() call the parameters
+  // should just be equal to (-gradient).
+  BOOST_REQUIRE_SMALL(arma::abs(coordinates - targetCoordinates).max(), 1e-7);
+  update.Update(coordinates, stepSize, dummyGradient);
+  // On the second Update() call the Momentum update will subtract
+  // the gradient from the momentum, which gives 2 * gradient value
+  // for the momentum on that step. Adding that to the gradient which
+  // was subtracted earlier yiels the 3 * gradient in the following check.
+  BOOST_REQUIRE_SMALL(arma::abs(coordinates - 3 * targetCoordinates).max(),
```
zoq: Really pedantic style issue; I would probably write:

```
BOOST_REQUIRE_SMALL(
    arma::abs(coordinates - targetCoordinates).max(), 1e-7);
```

src/mlpack/methods/ann/layer/cross_entropy_error.hpp
```diff
@@ -0,0 +1,111 @@
...
+
+#include <mlpack/prereqs.hpp>
+
+namespace mlpack {
+namespace ann /** Artificial Neural Network. */ {
+
+/**
+ * The cross-entropy performance function measures the network's
+ * performance according to the cross-entropy
+ * between the input and target distributions.
```
rcurtin: I think you can reduce the number of lines here, it seems like they are nowhere near as long as 80 characters. :)  (I think this also applies elsewhere)

src/mlpack/methods/ann/layer/cross_entropy_error_impl.hpp
rcurtin: Fair enough; in this case, can you clarify the limitation on how the labels should be in the documentation for the class?  We should definitely support multiclass cross-entropy at some point, so if you don't want to do that here that's ok, but in that case could I ask you to open a new issue for it, detailing basically what needs to be done and where someone could look to get started with it?

src/mlpack/tests/sgd_test.cpp
```diff
@@ -11,6 +11,8 @@
  */
 #include <mlpack/core.hpp>
 #include <mlpack/core/optimizers/sgd/sgd.hpp>
+#include <mlpack/core/optimizers/sgd/update_policies/gradient_clipping.hpp>
+#include <mlpack/core/optimizers/sgd/update_policies/vanilla_update.hpp>
```
rcurtin: Is there a need to add these here if they aren't being used by the tests?

zoq:
Thanks for the contributions!

PR #1072, opened 2017-07-29
Implementation of Neural Turing Machine.

sumedhghaisas:
After million tries to fix the gradients here is the implementation which passed gradient checks. :)
Tests are also performed with Reber grammar task.

This implementation introduces the concept of memory first time in our framework thus some new 'Forward' and 'Backward' signatures are introduced (backward compatible) to deal with this extra parameter of memory. 

I found out that layers that deal with memory are harder to debug cause the gradient come back in 2 directions, input and memory. To deal with this, I have added a new layer called 'MemoryTest' which fakes a memory content with linear layer and checks the gradient w.r.t. both input and memory. Thus all the layers that accept memory can be checked individually before adding them into bigger framework such as NTM which simulates memory with the combination of these layers.

NTM implementation is broken down into 4 main modules
1) Memory Head
2) Read Memory
3) Write Memory
4) Controller

1) Memory head module takes controller output and generates weights for memory locations.
2) Read memory module incorporates a Memory Head to read memory content given controller output and memory
3) Write memory module incorporates a Memory Head to create new memory content using older memory content and controller output
4) Currently controller is hardcoded as 2 layer feedforward network but a better design has to be implemented to abstract the controller network from NTM

This modularization will allow us to implement more complicated networks than NTM in later future. For example, NTM with multiple read and write heads. Best example of this is Differential Neural Computer.

1 major thing that is missing is design for different initialization of memory. Current the memory is initialized as constant ones, but according to the paper should be initialised with a bias layer so that the network also learnes the correct initialization. The framework for that is actually similar to the framework used in MemoryTest layer but some design decisions are needed.

I still need to go through the implementation once more to try to optimize wherever possible cause till now I have been paying more attention towards the correct implementation rather than fastest implementation. This PR would help in that as well.


sumedhghaisas:
Found another bug in the gradients of Memory Head while testing for Copy Task. Maybe that was the reason for failing tests on Travis. Although the gradients tests are passing on my system. Don;t know the reason for that. Will update as soon as it is resolved...

sumedhghaisas:
@zoq @rcurtin The PR is ready for review. I have fixed the gradient errors and optimized the implementation of Memory Head a bit. I am currently trying to run the copy task for NTM. Will send the PR to models repo.

sumedhghaisas:
The gradient test for NTM is failing online, but on my machine, they are passing. I checked the gradients of NTM again but could not find anything. The tests I am running with copy tasks are performing correctly. I am getting above 70 accuracy with sequence length of 10. But the network could be learning to compensate for the wrong gradients. @zoq Can you take a quick look at my gradients in NeuralTuringMachine class?

zoq:
Couldn't see anything right away, I'll have to take a closer look into the issue; I'll do that later today.

zoq:
Okay, I can reproduce the issue, you said on your system it passes the Gradient check?

zoq:
Just printed the gradients (currentGradient in the RNN class), and it looks like a big chunk is just zero. Can you reproduce the issue? I think that might be a good starting point to look into the issue. Let's see if we can identify the corresponding layer.

sumedhghaisas:
I am going to do clean build gain. Lets see if I can reproduce the issue. But re the gradients passing for MemoryHead, ReadMemory and WriteMemory? Then we can concentrate on NTM module alone. Its just the combination all 3 modules.

zoq:
GradientMemoryHeadTest, GradientWriteMemoryTest, GradientReadMemoryTest is fine at least on my machine.

src/mlpack/tests/ann_layer_test.cpp
```diff
@@ -779,6 +780,379 @@ BOOST_AUTO_TEST_CASE(GradientLSTMLayerTest)
...
+  mh.ForwardWithMemory(std::move(input), std::move(memory), std::move(weights));
+
+  // approximate k_t gate output
+  arma::mat k_t = arma::ones<arma::mat>(5, 1);
+
+  double b_t = 6.0;
+  double g_t = 1.0;
+  double gamma_t = 6.0;
+
+  arma::mat cosSimilarity = arma::normalise(memory, 1) * k_t /
```
zoq: Wondering if we could use https://github.com/mlpack/mlpack/blob/master/src/mlpack/core/kernels/cosine_distance.hpp here.

src/mlpack/tests/ann_layer_test.cpp
```diff
@@ -779,6 +780,379 @@ BOOST_AUTO_TEST_CASE(GradientLSTMLayerTest)
...
+  arma::mat w_g = g_t * w_c;
+
+  arma::mat shiftMatrix = arma::ones<arma::mat>(5, 5);
+  arma::mat w_tilde = arma::trans(arma::trans(w_g) * shiftMatrix);
+
+  arma::mat w_dash = arma::pow(w_tilde, gamma_t + 1);
+  w_dash = w_dash / arma::as_scalar(arma::sum(arma::sum(w_dash)));
+
+  BOOST_REQUIRE_CLOSE(arma::as_scalar(arma::trans(
+      arma::normalise(cosSimilarity)) * arma::normalise(weights)), 1, 1e-2);
```
zoq: Can we be more rigorous here, e.g. use 1e-4?

src/mlpack/tests/ann_layer_test.cpp
sumedhghaisas: ahh didn't think of that. Although that class only accepts vectors. Multipying the memory matrix directly with a vector is I think faster than multiplying the vectors individually. What do you think? Should we add a overload there? or just go ahead with this implementation?

src/mlpack/tests/ann_layer_test.cpp
sumedhghaisas: Sure. Will do that.

sumedhghaisas:
Any clue how I could fix the errors on windows build? I am not sure where the problem lies. is it the lambda function support in Armadillo?

zoq:
Appveyor builds against armadillo-7.800.2, which should support lambda function, looks like we have to switch to a for loop to solve the problem.

src/mlpack/tests/ann_layer_test.cpp
zoq: hm, the function uses a template so it should also accept an matrix instance.

sumedhghaisas:
@rcurtin @zoq I am not sure if for-loops are equivalent to for_each. Is it? I am scared to make any assumption about Armadillo to be honest :)

src/mlpack/methods/ann/layer/write_memory_impl.hpp
```diff
@@ -0,0 +1,205 @@
...
+      writeHead);
+  const arma::mat& writeWeights = boost::apply_visitor(outputParameterVisitor,
+      writeHead);
+
+  output = memory;
+
+  auto addVecIt = addVec.begin();
+  auto eraseVecIt = eraseVec.begin();
+
+  output.each_col([&](arma::vec& v)
```
zoq: I'm wondering if you remove the first ``&`` if that would solve the windows problem.

partobs-mdp:
Looks like this one already has a lot of neat stuff implemented :+1: About `LayerTypes` and `FFN` integration: is there any way to add `FFN<T>` for *all* `T`? As far as I understand, the running code version uses only `NegativeLogLikelihood<>` as `T`:

```C++
FFN<NegativeLogLikelihood<>, RandomInitialization>*,
```

It would be really neat, however, if one could also use MSE error (e.g., `EMBED` function in HAM), cross-entropy (for `OUTPUT` function of HAM) or, in fact, any predefined error.

zoq:
We are working on that, right now you have to add the types you need manually.

src/mlpack/methods/ann/layer/read_memory.hpp
```diff
@@ -0,0 +1,276 @@
...
+   * This function is used in testing the layer with MemoryTest layer.
+   *
+   * @param input Input data used for evaluating the specified function.
+   * @param output Resulting output activation.
+   * @param memory Current memory content.
+   */
+  template<typename eT>
+  void ForwardWithMemoryTest(arma::Mat<eT>&& input,
+                             const arma::Mat<eT>&& memory,
+                             arma::Mat<eT>&& output)
```
rcurtin: I see lots of code for `MemoryTest`, like this function, that class, and associated visitors, but I can't find anywhere where it was used.  Did I miss where to look, or has it not been checked in?

In general I don't like the idea of adding specific testing functions.  If you can add the tests that you were using, I can take a look at how it could be refactored so that this function wouldn't be necessary.

src/mlpack/methods/ann/layer/neural_turing_machine_impl.hpp
```diff
@@ -0,0 +1,249 @@
...
+
+  // Acess to current memory.
+  arma::mat& cMemory = memoryHistory.back();
+
+  // Pass the controller output through read memory layer.
+  boost::apply_visitor(ForwardWithMemoryVisitor(std::move(controllerOutput),
+      std::move(cMemory),
+      std::move(boost::apply_visitor(outputParameterVisitor, readMem))),
+      readMem);
+  lReads.push_back(boost::apply_visitor(outputParameterVisitor, readMem));
```
rcurtin: So it looks like you have built a lot of code like `ForwardWithMemory()` and `BackwardWithMemory()` and the associated visitors so that you can pass an `arma::mat memory` to the read memory and write memory.  But I think you could remove the need for all that extra code if `ReadMemory` and `WriteMemory` also held a reference to `memoryHistory`.  You could modify the constructor of `ReadMemory`:

```
ReadMemory(const size_t, const size_t, const size_t, const size_t, std::list<arma::mat>& memoryHistory)
```

and then add a member `std::list<arma::mat>& memoryHistory` to `ReadMemory`.  (Or maybe that would need to be a pointer so that serialization and assignment wouldn't have a problem.)  And you could do the same thing for `WriteMemory`.

Alternately, and I think this may be the better idea, simply add `arma::mat* currentMemory` members to both `ReadMemory` and `WriteMemory`, as well as a `CurrentMemory()` function to access and modify the pointer.  Then, instead of calling `ForwardWithMemoryVisitor`... do this:

```
readMem->CurrentMemory() = &memoryHistory.back();
boost::apply_visitor(ForwardVisitor(std::move(controllerOutput),
    std::move(boost::apply_visitor(outputParameterVisitor, readMem))),
    readMem);
```

and then use the same idea for the other instances where you would otherwise use `ForwardWithMemoryVisitor` or `BackwardWithMemoryVisitor`.

The nice thing about this situation is that you always know that the only classes that need to hold memory during `Forward()`, `Backward()`, or `Gradient()` will only be a `ReadMemory` or `WriteMemory`, so you can simply write the functions you need to set the memory correctly and access those directly, instead of through the visitor pattern.  The visitor pattern I think is more necessary for the `FFN` class, which is different because the actual layers that make up the FFN are in general not known, so we must use boost::variant + the visitor paradigm.

Let me know what you think of that idea.

src/mlpack/methods/ann/layer/layer_types.hpp
```diff
@@ -92,6 +109,7 @@ using LayerTypes = boost::variant<
     DropConnect<arma::mat, arma::mat>*,
     Dropout<arma::mat, arma::mat>*,
     ELU<arma::mat, arma::mat>*,
+    FFN<NegativeLogLikelihood<>, RandomInitialization>*,
```
rcurtin: It seems like this has to be added here so that you can pass a single layer or an entire `FFN` to the constructor of `NeuralTuringMachine`.  But unfortunately this means there is a restriction, that `NegativeLogLikelihood<>` and `RandomInitialization` must be used.  I have a different idea which maybe could be more helpful and robust; let me know what you think.

Instead of adding `FFN<>` to `LayerTypes`, add an additional constructor to `NeuralTuringMachine` that takes some `FFN` type (I guess a template type) and extracts the layers from it and adds them to `NeuralTuringMachine::network`.  I think we might have to add some functions to `FFN` so that individual layers could be accessed (not sure about that).

rcurtin: @partobs-mdp: I think this comment is relevant to the comment you made on this PR too---if you need to add networks to internal layers that you're building (I think you have to for HAM), this could be another approach that could work for that and not have the drawback you pointed out.

src/mlpack/methods/ann/ffn.hpp
```diff
@@ -172,9 +172,10 @@ class FFN
    * @param i Index of points to use for objective function gradient evaluation.
    * @param gradient Matrix to output gradient into.
    */
+  template<typename eT>
   void Gradient(const arma::mat& parameters,
                 const size_t i,
-                arma::mat& gradient);
+                arma::Mat<eT>& gradient);
```
sumedhghaisas: Ahh.. ooofff. Took me a while to figure out why I actually did this :) The change successfully enables the use of FNN as a controller in NTM. So with this w don't have to create a new controller layer type. A user can create an FFN, like he always does, and pass it to NTM. NTM will use this as the controller. Although I do have a question here. Here the user can add recurrent layers to this FFN cause the underlying call is made by RNN of NTM. Thats why I did not make the changes to RNN. Do you think thats confusing? Should I also make RNN LayerType supported?

src/mlpack/tests/cv_test.cpp
```diff
@@ -81,7 +81,8 @@ BOOST_AUTO_TEST_CASE(MSEMatResponsesTest)
   arma::mat data("1 2");
   arma::mat trainingResponses("1 2; 3 4");
 
-  FFN<MeanSquaredError<>, ZeroInitialization> ffn;
+  FFN<MeanSquaredError<>, ConstInitialization> ffn(MeanSquaredError<>(),
+    ConstInitialization(0));
```
sumedhghaisas: ahh... on it :)

src/mlpack/tests/init_rules_test.cpp
```diff
@@ -236,7 +236,8 @@ BOOST_AUTO_TEST_CASE(NetworkInitTest)
 
   // Create a simple network and use the ZeroInitialization rule to
   // initialize the network parameters.
-  FFN<NegativeLogLikelihood<>, ZeroInitialization> zeroModel;
+  FFN<NegativeLogLikelihood<>, ConstInitialization>
+    zeroModel(NegativeLogLikelihood<>(), ConstInitialization(0));
```
sumedhghaisas: I need to be bit careful with this :)

src/mlpack/methods/ann/layer/base_layer.hpp
```diff
@@ -168,6 +169,17 @@ template <
...
+/**
+ * Standard SoftPlus layer.
+ */
+template <
+    class ActivationFunction = SoftplusFunction,
+    typename InputDataType = arma::mat,
+    typename OutputDataType = arma::mat
+>
+using SoftplusLayer = BaseLayer<
+    ActivationFunction, InputDataType, OutputDataType>;
```
sumedhghaisas: You are right. Thats a better way. SoftPlus functions is described in its activation function file. But I agree. I will add some comments about softplus and redirect to the file for more details.

src/mlpack/methods/ann/layer/neural_turing_machine.hpp
```diff
@@ -0,0 +1,256 @@
+/**
+ * @file neural_turing_machine.hpp
+ * @author Sumedh Ghaisas
+ *
+ * Definition of the Neural Turing Machine class.
+ *
+ * //TODO add reference
```
sumedhghaisas: Ohh yes. :)

src/mlpack/methods/ann/layer/neural_turing_machine.hpp
```diff
@@ -0,0 +1,256 @@
...
+
+#include "memory_head.hpp"
+#include "read_memory.hpp"
+#include "write_memory.hpp"
+
+namespace mlpack {
+namespace ann /** Artificial Neural Network. */ {
+
+/**
+ * An implementation of Neural Turing Machine Layer.
```
sumedhghaisas: Yup. I planned to do that with the TODO of reference. But got very excited with the working version and started testing it right away.

src/mlpack/methods/ann/layer/neural_turing_machine.hpp
```diff
@@ -0,0 +1,256 @@
...
+  OutputDataType weights;
+
+  //! Locally-stored output parameter visitor.
+  OutputParameterVisitor outputParameterVisitor;
+
+  //! Locally-stored delta visitor.
+  DeltaVisitor deltaVisitor;
+
+  //! Locally-stored list of network modules.
+  std::vector<LayerTypes> network;
```
sumedhghaisas: I agree. But it makes the code way more readable. Using it like network[0] is annoying. A more clever way is to add references. But anyways they are pointers. Will it make a noticeable difference?

src/mlpack/methods/ann/layer/memory_head.hpp
```diff
@@ -0,0 +1,377 @@
...
+
+#include "layer_types.hpp"
+#include "add_merge.hpp"
+#include "sequential.hpp"
+
+namespace mlpack {
+namespace ann /** Artificial Neural Network. */ {
+
+/**
+ * An implementation of a memory head used in NTM.
```
sumedhghaisas: I agree. I will add comments.

src/mlpack/methods/ann/layer/read_memory.hpp
sumedhghaisas: ForwardWithMemoryTest is used by MemoryTest layer to call layer while testing the gradients. Although I think I can make some design changes  to counter this problem. Although that would involve deleting ReadMemory and WriteMemory layers and incorporating them into NTM itself. I thought of implementing this design but then decided to go with a more modularized approach. 

Okay so the main reason this function exists is because of WriteMemory layer. This layer accepts a memory, and the controller output and outputs a new memory content. Although for gradient testing this layer is used alone with MemoryTest layer (classic Unit testing... haha :) ). So this function reshapes the output to a vec so that this layer is just like a normal layer rather than being a layer which outputs a memory. This function is thus only called by MemoryTest layer while testing.

To bypass this, we need to make the memory changes in NTM so that WriteMemory just outputs the memory location to make the change. But then its just a MemoryHead ;P So no need of that layer. Let me know what you think.

Ahh and MemoryTest for the layers are done in ann_layer_test.cpp

src/mlpack/methods/ann/layer/neural_turing_machine_impl.hpp
sumedhghaisas: ahh man. I totally agree. This is a better design. I feel like an idiot now :) It would add some more iterators in ReadMemory and WriteMemory but it would eliminate the use of ForwardWithMemory, which I always thought as a bad design but didn't think of any solution at that point. I totally missed this point that ReadMemory and WriteMemory are not made for FFN and RNN so we can create non-classical constructors for them.

src/mlpack/methods/ann/layer/layer_types.hpp
sumedhghaisas: Ohh thats not the case. So the template parameters passed on to the FFN are irrelevant. They won;t be used at all. The initialization is done by the base class, which in this case is the RNN in which NTM layer will be added. There the user can specify the appropriate initialization and error function. Both those template arguments, for this use of FFN as a controller are dummies.

sumedhghaisas:
@partobs-mdp For the case of NTM the parameter passed on to the FFN object, which is used as a controller are dummy parameter. The parameter passed to the base RNN class will be used instead to initialize the controller network. So my design is, NTM is just a layer, which is added in RNN. So the user can add multiple layers on top of that and use appropriate Error in RNN class. 

I am not sure about the design that you are using. Can you tell me the details of your design, maybe I can help you

sumedhghaisas:
@partobs-mdp Also let me rebase this branch on the merged GRU. This is on the older GRU. I can imagine the horrible errors that have come up for you :) I apologize for that. I will get it done as soon as possible.

sumedhghaisas:
@rcurtin The gradient and forward tests are implemented in ann_layer_test.cpp. But I am yet to add recurrent tests in recurrent_layer_test.cpp. I just have to copy the LSTM and GRU tests for NTM.

sumedhghaisas:
@zoq I like the idea of creating a visual representation of the recursive grammar. It would make an amazing blog post. I also need to write those scripts for models to get the minimum cell state size required versus the average recursive depth. I am so looking forward to that. Although this M.Sc. dissertation submission is keeping me away from these pleasures. 4 more days and I will be back with some fun experiments :)

src/mlpack/methods/ann/layer/read_memory.hpp
sumedhghaisas: Ohh no. Another thing I realized, Its not a good design to keep the WriteMemory module. Its better to make it internal to NTM. Cause there are 2 operations in memory write, erase and add. And I ignored the fact that in the case of extending NTM with multiple heads, all erases have to be done first then all the adds but the current design won;t let us do that.

sumedhghaisas: So maybe its better to go to NTM internal implementation of ReadMemory and WriteMemory.

partobs-mdp:
@sumedhghaisas: HAM mostly follows the same design (use FFNs for running some sub-steps), but is more complicated, because it uses five different FFNs: one for embedding sequence vectors into memory, one - for traversing the memory (which is tree-based - the noteworthy part of HAM), one - for replacing memory contents, one - for evaluating inner memory nodes by it children, and one - for emitting output sequence.

@rcurtin Before (unsuccessfully) trying to migrate to `LayerTypes`, I was trying to make a templated class. Personally, the only drawback I see about that is that the class definition will be really messy (`HAM<FFN, FFN, FFN, FFN, RNN>`, where `FFN` and `RNN` are themselves templated) What do you think?

As a follow-up: can I use `boost::visitor` for FFN (or RNN) in the templated setting? (Maybe a dumb question, but I can't clearly answer it)

src/mlpack/methods/ann/ffn.hpp
zoq: It's a good idea to support the ``RNN``class as well, that way we can easily use an ``LSTM`` as controller.

zoq: Also, I guess the question here was what is the reason to use ``arma::Mat<eT>`` instead of ``arma::mat``.

src/mlpack/methods/ann/layer/read_memory.hpp
zoq: I would also go with the internal implementation, I don't think someone is going to reuse the code for anything else.

zoq:
I think both strategies have their advantages, as pointed out by @partobs-mdp  the class definition is much cleaner but on the other side, we are more restricted since we have to each type to layer types. On the other side, we could provide reasonable default values for ``HAM<FFN, FFN, FFN, FFN, RNN>`` so that we could use ``HAM<>`` later on.

About the visitor, if we know the type we don't have to use any visitor we can directly use the function.

src/mlpack/methods/ann/layer/base_layer.hpp
rcurtin: Sounds good, thanks!

src/mlpack/methods/ann/layer/neural_turing_machine.hpp
```diff
@@ -0,0 +1,256 @@
...
+  arma::mat dMem;
+
+  //! Locally stored previous memory content error.
+  arma::mat dMemPrev;
+
+  //! All zeros vector for initializing read.
+  arma::mat allZeros;
+
+  //! All ones matrix for initializing memory.
+  arma::mat allOnes;
```
rcurtin: You can ignore this comment, we discussed it in the GRU PR.  I agree that this is the best way we can handle this for now.  I forgot to delete it from the review before I hit 'post' :)

src/mlpack/methods/ann/layer/read_memory.hpp
rcurtin: I agree also---probably the internal implementation is better.  I think it will simplify the code too; probably visitors won't be necessary to use internally.  You could just do the matrix operations directly.  I think this will be a pretty significant change, so I think if you'll redo the implementation internally, I'll save further comments for once that's done.

src/mlpack/methods/ann/layer/neural_turing_machine_impl.hpp
rcurtin: No worries, sometimes just a second set of eyes is all that's needed. :)

src/mlpack/methods/ann/ffn.hpp
sumedhghaisas: Sure. I will make the RNN class supported as well. Ahh... I think the reason was that Gradient function already exists in that class apart from this function. Thus I had to templatize it to make it work.

zoq:
Have you tested the code on the recursive embedded reber grammar task?

sumedhghaisas:
@zoq Some documentation changes and recursive reber grammar checks are remaining. Will add them today.

sumedhghaisas:
@zoq @rcurtin For windows, maybe armadillo does not have lambda function overload for each_row and each_col. Do you know how to solve this?

zoq:
Have you tested: 

> I'm wondering if you remove the first & if that would solve the windows problem.

sumedhghaisas:
@zoq You mean the the '&' inside the closure? that would mean we will incur copy overhead.

zoq:
Just wanted to test if that works, since MSVC supports lambda functions.

zoq:
At the end we could replace ``.each_col``with a for loop, what do you think?

sumedhghaisas:
That should work for now. But I am not sure about the performance comparison.

src/mlpack/methods/ann/layer/neural_turing_machine.hpp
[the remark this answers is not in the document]
rcurtin: Ah, sorry for the slow response.  You are right, since it is a boost::variant to a pointer, I think it will not be a problem to also store the other objects.

src/mlpack/methods/ann/layer/neural_turing_machine_impl.hpp
```diff
@@ -0,0 +1,367 @@
...
+    const size_t numMem,
+    const size_t memSize,
+    const size_t shiftSize,
+    LayerTypes controller) :
+    inSize(inSize),
+    outSize(outSize),
+    numMem(numMem),
+    memSize(memSize),
+    shiftSize(shiftSize),
+    controller(controller),
```
rcurtin: You can use `controller(std::move(controller))` here.  See https://github.com/mlpack/mlpack/issues/1021 for a little more detail on why.

src/mlpack/methods/ann/layer/neural_turing_machine_impl.hpp
```diff
@@ -0,0 +1,367 @@
...
+  backwardStep = 0;
+  gradientStep = 0;
+  forwardStep = 0;
+}
+
+template<typename InputDataType, typename OutputDataType>
+template<typename Archive>
+void NeuralTuringMachine<InputDataType, OutputDataType>::Serialize(
+    Archive& ar, const unsigned int /* version */)
+{
```
rcurtin: Hm, I think that we actually do need to do serialization here, can you handle that?  You can take a look at the PR I just opened: #1103.  (Actually it would be good if you looked at that anyway, I made some changes to `GRU`.)

src/mlpack/methods/ann/layer/neural_turing_machine.hpp
```diff
@@ -0,0 +1,273 @@
...
+   * @param memSize Size of each memory location.
+   * @param shiftSize Circular shift rotation size.
+   * @param controller The controller network to use.
+   */
+  NeuralTuringMachine(const size_t inSize,
+                      const size_t outSize,
+                      const size_t numMem,
+                      const size_t memSize,
+                      const size_t shiftSize,
+                      LayerTypes controller);
```
rcurtin: I think a destructor also needs to be added to loop over `network` and use a `DeleteVisitor()`.

src/mlpack/methods/ann/layer/neural_turing_machine_impl.hpp
sumedhghaisas: You are right, but I am not sure how much performance difference it would make since LayerTypes is only holding the pointers.

src/mlpack/methods/ann/layer/neural_turing_machine_impl.hpp
sumedhghaisas: Sure. I will add the changes according to your changes to GRU.

src/mlpack/methods/ann/layer/neural_turing_machine.hpp
sumedhghaisas: Sure thing...

sumedhghaisas:
@zoq @rcurtin Ready to me merged?

zoq:
Did you test the NTM code on some tasks from the paper e.g. Copy? Would be an easy test to see if  we haven't missed something, and the NTM is able to learn each problem to some extend.

sumedhghaisas:
I have tested it with copy task. Should I add some tasks as a boost test? maybe of smaller length sequence?

zoq:
We could open a new PR on the models repo, to provide an executable to run the code against the tasks, just like Konstantin did for the HAM code. What do you think?

zoq:
Thanks for the updates, I'll review the code over the next days.

src/mlpack/methods/ann/layer/memory_head_impl.hpp
```diff
@@ -0,0 +1,482 @@
...
+
+  // Build kT with non linearity.
+  boost::apply_visitor(ForwardVisitor(std::move(lOutput.submat(0, 0,
+      memSize - 1, 0)), std::move(boost::apply_visitor(outputParameterVisitor,
+      kTNonLinear))), kTNonLinear);
+  const arma::mat& kT = boost::apply_visitor(outputParameterVisitor,
+      kTNonLinear);
+
+  // Build bT with non linearity
+  lBt.push_back(bTNonLinear.Fn(lOutput(memSize, 0)));
```
zoq: Do you think instead of using a vector to hold the states, we could use ``arma::cube``? I think this would improve the overall readability.

```c++
lBt.push_back(bTNonLinear.Fn(lOutput(memSize, 0)));
const double& bT = lBt.back();
```

would become

```c++
gT.slice(step) = gTNonLinear.Fn(lOutput(memSize + 1, 0);
```

src/mlpack/methods/ann/layer/memory_head_impl.hpp
```diff
@@ -0,0 +1,482 @@
...
+  const double& gammaT = lGammaT.back();
+
+  // Perform cosine similarity with memory content
+  lConsineT.push_back(arma::normalise(memory, 2, 1) * arma::normalise(kT));
+  const arma::vec& cosSimilarity = lConsineT.back();
+
+  // Build wC with bT and softmax
+  lWe.push_back(arma::exp(bT * cosSimilarity));
+  const arma::vec& wE = lWe.back();
+
```
zoq: We should remove the line here, since ``lwe`` doesn't contain the softmax, ``lWc`` does.

src/mlpack/tests/recurrent_network_test.cpp
```diff
@@ -529,36 +580,102 @@ void ReberGrammarTestNetwork(size_t hiddenSize = 4,
...
+BOOST_AUTO_TEST_CASE(NTMReberGrammarTest)
+{
+  size_t hiddenSize = 4;
+  size_t numMem = 3;
+  size_t memSize = 5;
+  size_t shiftSize = 1;
+
+  auto fun = [&](RNN<MeanSquaredError<> >& model)
+  {
+    FFN<>* controller = new FFN<>();
```
zoq: Should we test the NTM code with an LSTM controller as well?

src/mlpack/methods/ann/layer/neural_turing_machine_impl.hpp
```diff
@@ -0,0 +1,406 @@
...
+  // Generate write weights
+  boost::apply_visitor(ForwardVisitor(std::move(controllerOutput),
+      std::move(boost::apply_visitor(outputParameterVisitor, writeHead))),
+      writeHead);
+  const arma::mat& writeWeights = boost::apply_visitor(outputParameterVisitor,
+      writeHead);
+
+  arma::mat nMemory = cMemory;
+
+  auto addVecIt = addVec.begin();
```
zoq: Do you mind to avoid ``auto`` here, I think it helps to understand the code if the type is immediately visible.

zoq:
I think it would be a good idea, to test the code on more tasks, like the copy task mentioned on the paper.

himanshupathak21061998:
@zoq I think this pr contains some extra commits which are not necessary I have to start from new one just by copying changes. As you mentioned their is pr on Neural Turing Machine I found this out. After completing https://github.com/mlpack/mlpack/pull/2099 I will try to resurect this pr.

PR #1074, opened 2017-08-01
Precision, recall and F1 for the cross-validation and hyper-parameter tuning modules

micyril:


micyril:
For F1 it is unclear for me what we should return if precision and recall turn out to be zeros. I see the following options.
1. To return NaN (the result of 2 * 0.0  * 0.0 / (0.0 + 0.0))
2. To return 0.0 (since F1 should serve as a harmonic mean of precision and recall)
3. To throw an exception.

src/mlpack/core/cv/metrics/precision.hpp
```diff
@@ -0,0 +1,71 @@
...
+/**
+ * Precision is a metric of performance for classification algorithms that for
+ * binary classification is equal to tp / (tp + fp), where tp and fp are the
+ * numbers of true positives and false positives respectively. For multiclass
+ * classification the precision metric can be used with the following strategies
+ * for averaging.
+ * 1. Micro. If there are N + 1 classes in total, the result is equal to
+ * (tp0 + tp1 + ... + tpN) / (tp0 + tp1 + ... + tpN + fp0 + fp1 + ... fpN),
+ * where tpI and fpI are the numbers of true positives and false positives
+ * respectively for the class (label) I.
```
rcurtin: You can use inline LaTeX here with Doxygen, it would probably look nicer in the resulting generated documentation:

```
an inline formula is @f$ x + y @f$, and a standalone one is

@f[
(tp_0 + tp_1 + \ldots + tp_N) / (\ldots)
@f]
```

I think this comment would apply a few other places too.

src/mlpack/core/cv/metrics/precision_impl.hpp
```diff
@@ -0,0 +1,84 @@
...
+  AssertSizes(data, labels, "Precision<Micro>::Evaluate()");
+
+  arma::Row<size_t> predictedLabels;
+  model.Classify(data, predictedLabels);
+
+  size_t numClasses = arma::max(labels) + 1;
+
+  size_t tpTotal = 0;
+  for (size_t c = 0; c < numClasses; ++c)
+    tpTotal += arma::sum((labels == c) % (predictedLabels == c));
```
rcurtin: Could you do `arma::sum(labels == predictedLabels)` here and drop the for loop?

micyril: A nice catch. I guess we can just reuse Accuracy.

rcurtin:
Ah sorry I realize I didn't answer the question about F1. I believe returning 0 would be the best option in this case; this is what a user is most likely to expect (something in the range of 0 to 1).

Another couple questions---

1. Now the Micro averaging strategy is just Accuracy in all cases that you are adding here. Is there a reason to keep the Micro enum then, or should users simply use Accuracy directly?

2. Should the binary strategy allow specifying a different target class than class 0? This could provide some extra flexibility for users.

micyril:
> Now the Micro averaging strategy is just Accuracy in all cases that you are adding here. Is there a reason to keep the Micro enum then, or should users simply use Accuracy directly?

Maybe some users don't know that microaveraged precision, recall and F1 are the same as accuracy (It was the case for me before I started to write the code). You can also see that some other libraries, like  [scikit-learn](http://scikit-learn.org/stable/modules/classes.html#module-sklearn.metrics), keep the microaveraged metrics along with accuracy.

> Should the binary strategy allow specifying a different target class than class 0? This could provide some extra flexibility for users.

I can imagine the following interface:

```
static const size_t targetClass = 5;
double precision = Precision<Binary, targetClass>::Evaluate(model, data, labels);
```

What do you think?

rcurtin:
> Maybe some users don't know that microaveraged precision, recall and F1 are the same as accuracy (It was the case for me before I started to write the code). You can also see that some other libraries, like scikit-learn, keep the microaveraged metrics along with accuracy.

That seems fine to me, not a bad idea to keep an alias around like this.

I think the interface you proposed is good, it's the idea I had in mind too.

src/mlpack/core/cv/metrics/f1.hpp
```diff
@@ -29,19 +31,17 @@ namespace cv {
...
  *
  * The returned value for F1 will be zero if both precision and recall turn out
  * to be zeros.
  *
  * @tparam AS An average strategy.
+ * @tparam PositiveClass In the case of binary classification (AS = Binary)
+ *     positives are assumed to have labels equal to this value.
  */
-template<AverageStrategy AS>
+template<AverageStrategy AS, size_t PositiveClass = 1>
```
rcurtin: I'd suggest a default positive label of `0` since that's the smallest possible label.

src/mlpack/core/cv/metrics/f1.hpp
```diff
@@ -62,6 +62,43 @@ class F1
...
+   */
+  template<AverageStrategy _AS,
+           typename MLAlgorithm,
+           typename DataType,
+           typename = std::enable_if_t<_AS == Macro>,
+           typename = void,
+           typename = void>
+  static double Evaluate(MLAlgorithm& model,
+                        const DataType& data,
+                        const arma::Row<size_t>& labels);
```
rcurtin: Wow, the refactoring necessary to allow the user to specify a positive class is a lot harder than I thought!  I think that you might be able to use SFINAE in the arguments instead of adding the extra `typename = void`s:

```
template<AverageStrategy _AS, typename MLAlgorithm, typename DataType>
static double Evaluate(
    MLAlgorithm& model,
    const DataType& data,
    const std::enable_if_t<_AS == Micro>* = 0);
```

and likewise for `_AS == Macro` and `_AS == Binary`.

rcurtin:
@micyril: do you want to work out the merge conflict?  If so I will merge this.  Also, do you want to change the default positive label to `0`?

micyril:
Ok

micyril:
Are you sure you want to see 0 as the default positive label? For me it is more natural to have 1 for that. In [scikit-learn](http://scikit-learn.org/stable/modules/generated/sklearn.metrics.precision_score.html#sklearn.metrics.precision_score) also 1 is used as default.

rcurtin:
Fair, compatibility here is probably a good idea.  So I agree we should leave it as 1.

PR #1075, opened 2017-08-01
Implementation of Stochastic Coordinate Descent

shikharbhardwaj:
Here is an implementation of another optimizer, SCD. I have started with a serial implementation first, and made some changes to `LogisticRegressionFunction` to test this optimizer. The optimizer requires another addition to the requirements for the function type to be optimized.

Currently, I am working on getting two things done : 

1. Implement the greedy descent policy, using ideas from : http://www.maths.ed.ac.uk/~prichtar/Optimization_and_Big_Data_2015/slides/Schmidt.pdf.

2. Optimize the `FeatureGradient` method in `LogisticRegressionFunction` to not use `Gradient`.

I'll update soon with more changes.


src/mlpack/core/optimizers/scd/descent_policies/greedy_descent.hpp
```diff
@@ -0,0 +1,55 @@
...
+   *
+   * @tparam ResolvableFunctionType The type of the function to be optimized.
+   * @param numEpoch The iteration number for which the feature is to be
+   *    obtained.
+   * @param iterate The current value of the decision variable.
+   * @param function The function to be optimized.
+   * @return The index of the coordinate to be descended.
+   */
+
+  // TODO: Find a way to implement this.
```
zoq: I'm not sure I get the point of the TODO, can you elaborate on this?

shikharbhardwaj: I am planning to implement this descent policy which gives the feature with the maximum guaranteed descent, based on : http://www.maths.ed.ac.uk/~prichtar/Optimization_and_Big_Data_2015/slides/Schmidt.pdf and ideas from this implementation of SCD : http://ttic.uchicago.edu/~tewari/code/scd/

The above mentioned implementation uses a parameter "rho", which is a part of the objective function, which is 0.25 for logistic and hinge loss and 1 for quadratic loss in the part for calculating the predicted descent and finding the feature with the steepest descent.

src/mlpack/core/optimizers/scd/scd.hpp
```diff
@@ -0,0 +1,139 @@
...
+ *  size_t NumFeatures();
+ *  double Evaluate(const arma::mat& coordinates);
+ *  void FeatureGradient(const arma::mat& coordinates,
+ *                       const size_t j,
+ *                       arma::sp_mat& gradient);
+ *
+ *  NumFeatures() should return the number of features in the decision variable.
+ *  Evaluate gives the value of the loss function at the current decision
+ *  variable and FeatureGradient is used to evaluate the partial gradient with
+ *  respect to the jth feature.
```
rcurtin: Thanks for documenting this.  What happens if a user passes a `FunctionType` that isn't a `ResolvableFunctionType`?  How difficult is the error message to understand?

shikharbhardwaj: I guess if someone passes in a `FunctionType` without the `ResolvableFunctionType` interface, the compiler will complain about the absence of member functions `NumFeatures` and `FeatureGradient`. Are there ways we can make this more understandeable to the user?

shikharbhardwaj:
Thanks for the suggestions Ryan.
I am currently working on some more improvements. I'll add more functions with the Resolvable requirements soon.

Documenting the policies added is a nice improvement. I'll do that.

src/mlpack/core/optimizers/parallel_sgd/sparse_test_function.hpp
```diff
@@ -27,21 +27,59 @@ class SparseTestFunction
...
   arma::mat GetInitialPoint() const { return arma::mat("0; 0; 0; 0;"); }
 
+  //! Evaluate all the functions.
+  double Evaluate(const arma::mat& coordinates, const size_t i) const
+  {
+    return coordinates[i] * coordinates[i] + bi[i] * coordinates[i] +
+      intercepts[i];
+  }
+
   //! Evaluate a function.
```
mentekid: Minor - I think this comment and the one on the overload above should be swapped?

src/mlpack/core/optimizers/scd/descent_policies/random_descent.hpp
```diff
@@ -0,0 +1,49 @@
...
+ * descent scheme picks a the co-ordinate for the descent uniformly randomly.
+ */
+class RandomDescent
+{
+ public:
+  /**
+   * The DescentFeature method is used to get the descent coordinate for the
+   * current iteration.
+   *
+   * @tparam ResolvableFunctionType The type of the function to be optimized.
```
mentekid: Should we have a pointer to the more detailed description of what the signature of this function is in `scd.hpp`? Something like "For more information see the SCD implementation"

src/mlpack/core/optimizers/scd/scd.hpp
```diff
@@ -0,0 +1,139 @@
...
+ *
+ *  @tparam ResolvableFunctionType A function whose partial gradients with
+ *      respect to the jth feature can be obtained.
+ *  @tparam DescentPolicy Descent policy to decide the order in which the
+ *      coordinate for descent is selected.
+ */
+template <typename DescentPolicyType = RandomDescent>
```
mentekid: I like that you have a default policy here.

src/mlpack/core/optimizers/scd/scd.hpp
```diff
@@ -0,0 +1,139 @@
...
+  size_t maxIterations;
+
+  //! The tolerance for termination.
+  double tolerance;
+
+  //! The update interval for reporting objective and testing for convergence.
+  size_t updateInterval;
+
+  //! The descent policy used to pick the coordinates for the update.
+  DescentPolicyType descentPolicy;
```
mentekid: I would argue that the user would be unlikely to change the descent policy during the object's lifetime - would it make sense to make it const and remove the set function? I realise it's a very minor point though.

shikharbhardwaj:
I finished up the changes in `SoftmaxRegressionFunction` with the last commit. A minor inconsistency I noticed while working with the code in softmax and logistic regression is the way the decision variable is represented. Softmax regression has the features arranged column-wise, where as logistic regression has it row-wise (so it gives out a single column vector as the output from `Gradient`).

I guess it would be nice to make this uniform across all the functions (SCD could then do the update on the relevant column instead of working on the entire decision variable) easing up the parallelisation?

zoq:
I think we should see if we can work out the inconsistency, @rcurtin already looked into the logistic regression method so he can probably provide some more insights, regarding what needs to be done.

shikharbhardwaj:
I am working on the changes in Logistic regression (using a `rowvec` for the decision variable and obtaining the submatrix views with `tail_cols`). I am done with the changes in the function, but they break some other tests (which assumed the gradient to be in the other shape).

zoq:
Can you refactor the test cases too, don't feel obligated.

shikharbhardwaj:
Sure, I am already on it. :)

src/mlpack/methods/logistic_regression/logistic_regression_function_impl.hpp
```diff
@@ -110,13 +110,14 @@ double LogisticRegressionFunction<MatType>::Evaluate(
 {
   // Calculate the regularization term.  We must divide by the number of points,
   // so that sum(Evaluate(parameters, [1:points])) == Evaluate(parameters).
+  double norm = arma::norm(parameters.tail_cols(parameters.n_elem - 1));
+
   const double regularization = lambda * (1.0 / (2.0 * predictors.n_cols)) *
-      arma::dot(parameters.col(0).subvec(1, parameters.n_elem - 1),
-                parameters.col(0).subvec(1, parameters.n_elem - 1));
+    norm * norm;
```
mentekid: indent 2 more spaces here

src/mlpack/methods/softmax_regression/softmax_regression_function.cpp
```diff
@@ -205,3 +205,16 @@ void SoftmaxRegressionFunction::Gradient(const arma::mat& parameters,
                lambda * parameters;
   }
 }
+
+void SoftmaxRegressionFunction::FeatureGradient(const arma::mat& parameters,
+    const size_t j,
```
mentekid: You could also tab this to be under `c` in the above line, but that's really minor

src/mlpack/core/optimizers/parallel_sgd/parallel_sgd_impl.hpp
```diff
@@ -54,10 +54,7 @@ double ParallelSGD<DecayPolicyType>::Optimize(
     lastObjective = overallObjective;
     overallObjective = 0;
```
mentekid: Also, you can remove this line since you're now evaluating it all in one go.

mentekid:
This looks ready to merge on my end - AppVeyor failure seems to be a random timeout.

I am leaving it open to any comments or reviews, and I'll merge this on Friday if nothing comes up.

@shikharbhardwaj nice work - thank you and well done :)

zoq:
Let's restart the windows build, looks like the build could not fetch the repo.

src/mlpack/methods/logistic_regression/logistic_regression_function_impl.hpp
```diff
@@ -158,19 +159,47 @@ void LogisticRegressionFunction<MatType>::Gradient(
...
+template <typename MatType>
+void LogisticRegressionFunction<MatType>::FeatureGradient(
+    const arma::mat& parameters,
+    const size_t j,
+    arma::sp_mat& gradient) const
+{
+  const arma::rowvec sigmoids = (1 / (1 + arma::exp(-parameters(0, 0)
+      - parameters.tail_cols(parameters.n_elem - 1) * predictors)));
+
+  arma::mat diffs = responses - sigmoids;
```
zoq: I think we could inline the operation here, since it only needed once.

src/mlpack/tests/scd_test.cpp
```diff
@@ -0,0 +1,165 @@
...
+  // the feature with index 2.
+  arma::mat point("1; 2; 3; 4;");
+
+  SparseTestFunction f;
+
+  GreedyDescent descentPolicy;
+
+  BOOST_REQUIRE_EQUAL(descentPolicy.DescentFeature(0, point, f), 2);
+
+  point[1] = 10;
```
zoq: Can you eloborate on this (add a comment)?

doc/policies/functiontype.hpp
```diff
@@ -0,0 +1,117 @@
...
+@endcode
+
+To evaluate the loss function at the given coordinates.
+
+@code
+void Gradient(const arma::mat& coordinates, arma::mat& gradient);
+@endcode
+
+To evaluate the gradient at the given coordinates, where \c gradient is an
+out-param for the required gradient.
```
rcurtin: It might be easier to read if you include this comment as part of the `@code` snippet:

```
// Evaluate the gradient at the given coordinates, where 'gradient' is an output
// parameter for the required gradient.
void Gradient(const arma::mat& coordinates, arma::mat& gradient);
```

There are lots of little `@code` blocks you could modify like this.  What do you think?

src/mlpack/core/optimizers/scd/descent_policies/cyclic_descent.hpp
```diff
@@ -0,0 +1,50 @@
...
+
+#include <mlpack/core.hpp>
+
+namespace mlpack {
+namespace optimization {
+
+/**
+ * Cyclic descent policy for Stochastic Co-ordinate Descent(SCD). This
+ * descent scheme picks a the co-ordinate for the descent in a cyclic manner
+ * serially.
```
rcurtin: Is there an easy paper to cite here for SCD?  It's always nice to provide references to the original paper.

doc/policies/functiontype.hpp
```diff
@@ -0,0 +1,117 @@
...
+
+@code
+void FeatureGradient(const arma::mat& coordinates, const size_t j, arma::sp_mat& gradient);
+@endcode
+
+To evaluate the gradient at the given coordinates, where \c gradient is an
+out-param for the required gradient. The out-param is a sparse matrix(with
+dimensions equal to the decision variable), for storing the gradient of the
+jth feature.  The \c gradient matrix is supposed to be non-zero in the jth
+column, which contains the relavant partial gradient.
```
rcurtin: Minor misspelling: 'relevant' :)

doc/policies/functiontype.hpp
```diff
@@ -0,0 +1,117 @@
...
+
+@code
+double Evaluate(const arma::mat& coordinates);
+@endcode
+
+To evaluate the loss function at the given coordinates, same as the
+\c FunctionType interface.
+
+@code
+void FeatureGradient(const arma::mat& coordinates, const size_t j, arma::sp_mat& gradient);
```
rcurtin: Do you think it would be better to name this function `PartialGradient()` since it is the partial gradient with respect to one of the `coordinates` features?

src/mlpack/core/optimizers/scd/descent_policies/greedy_descent.hpp
```diff
@@ -0,0 +1,77 @@
...
+class GreedyDescent
+{
+ public:
+  /**
+   * The DescentFeature method is used to get the descent coordinate for the
+   * current iteration.
+   *
+   * @tparam ResolvableFunctionType The type of the function to be optimized.
+   * @param numEpoch The iteration number for which the feature is to be
+   *    obtained.
```
rcurtin: To me an `epoch` is a term used for an entire pass over the whole dataset, but it seems like the wrong word to use here.  I think it might be clearer to just name this parameter `iteration`.

src/mlpack/core/optimizers/scd/descent_policies/random_descent.hpp
```diff
@@ -0,0 +1,51 @@
...
+   * @param numEpoch The iteration number for which the feature is to be
+   *    obtained.
+   * @param iterate The current value of the decision variable.
+   * @param function The function to be optimized.
+   * @return The index of the coordinate to be descended.
+   */
+  template <typename ResolvableFunctionType>
+  size_t DescentFeature(const size_t /* numEpoch */,
+                        const arma::mat& /* iterate */,
+                        const ResolvableFunctionType& function)
```
rcurtin: You could mark this `static`.  This is true of all three `DescentPolicy` classes you wrote.

src/mlpack/core/optimizers/scd/scd.hpp
rcurtin: Sorry for the slow response here.  We could use a `static_assert()` check to ensure that it has the proper methods.  But for the sake of this PR I think that is not necessary.  If you are interested afterwards in making the error reporting more robust for template parameters, I have some ideas.  But if you are not interested that is ok too, you should not feel obligated. :)

doc/policies/functiontype.hpp
```diff
@@ -0,0 +1,117 @@
...
+for disjoint updates. The features should be arranged columnwise in the decision
+variable. For example, in \c SoftmaxRegressionFunction the decision variable has
+size \c numClasses x \c featureSize (+ 1 if an intercept also needs to be fit).
+Similarly, for \c LogisticRegression, the decision variable is a row vector,
+with the number of columns determined by the dimensionality of the dataset.
+
+The interface expects the following member functions from the function class
+
+@code
+size_t NumFeatures();
```
rcurtin: I am not sure on this---is it always true that `NumFeatures() == coordinates.n_elem`?  If so, I am not sure this function is needed.

src/mlpack/methods/logistic_regression/logistic_regression_function_impl.hpp
```diff
@@ -70,16 +70,16 @@ double LogisticRegressionFunction<MatType>::Evaluate(
   // multiplied by the squared l2-norm of the parameters then divided by two.
 
   // For the regularization, we ignore the first term, which is the intercept
-  // term.
-  const double regularization = 0.5 * lambda *
-      arma::dot(parameters.col(0).subvec(1, parameters.n_elem - 1),
-                parameters.col(0).subvec(1, parameters.n_elem - 1));
+  // term and take every term except the last one in the decision variable.
+  double norm = arma::norm(parameters.tail_cols(parameters.n_elem - 1));
```
rcurtin: I think the use of `tail_cols()` is a nice improvement, but is the switch of `parameters` from `arma::vec` to `arma::rowvec` necessary?  That's an API change and might break some user's code.  I think you could also just do `tail_rows()` and leave it as an `arma::vec`.  Let me know what you think.  Also I think it might be faster (very incrementally) to use `arma::dot()` instead of `arma::norm()` and then multiplying.

src/mlpack/methods/softmax_regression/softmax_regression_function.cpp
```diff
@@ -205,3 +205,16 @@ void SoftmaxRegressionFunction::Gradient(const arma::mat& parameters,
...
+                                                const size_t j,
+                                                arma::sp_mat& gradient) const
+{
+  arma::mat denseGrad;
+
+  Gradient(parameters, denseGrad);
+
+  gradient.set_size(arma::size(denseGrad));
+
+  gradient.col(j) = denseGrad.col(j);
```
rcurtin: Would it be faster to explicitly calculate only the part of the gradient that is needed here?

src/mlpack/tests/scd_test.cpp
```diff
@@ -0,0 +1,165 @@
...
+  BOOST_REQUIRE_CLOSE(iterate[0], 2, 0.02);
+  BOOST_REQUIRE_CLOSE(iterate[1], 1, 0.02);
+  BOOST_REQUIRE_CLOSE(iterate[2], 1.5, 0.02);
+  BOOST_REQUIRE_CLOSE(iterate[3], 4, 0.02);
+}
+
+/**
+ * Test the greedy descent policy.
+ */
+BOOST_AUTO_TEST_CASE(GreedyDescentTest)
```
rcurtin: It might be also useful to add a simple test for `RandomDescent` and `CyclicDescent`; very simple tests can just make sure they give the expected result.  These can be helpful later, to make sure that nothing is broken by later maintenance/refactorings of the code.

src/mlpack/tests/scd_test.cpp
```diff
@@ -0,0 +1,165 @@
...
+  {
+    arma::sp_mat fGrad;
+    f.FeatureGradient(testPoint, i, fGrad);
+
+    CheckMatrices(testGradient.col(i), arma::mat(fGrad.col(i)));
+  }
+}
+
+/**
+ * Test changes to Softmax regression function.
```
rcurtin: I think this might be an inaccurate comment (or at least it won't make sense after merge), maybe something more like `Test that SoftmaxRegressionFunction::FeatureGradient()` works as expected.

doc/policies/functiontype.hpp
shikharbhardwaj: Sure, I'll do that. I was thinking about improving the formatting but somehow it slipped my mind.

doc/policies/functiontype.hpp
shikharbhardwaj: Yes, in the current scheme, `NumFeatures() == coordinates.n_cols`. I guess we can remove this function then (I kept it for consistency with SGD).

doc/policies/functiontype.hpp
shikharbhardwaj: I agree, `PartialGradient`  describes the function better.

src/mlpack/core/optimizers/scd/scd.hpp
```diff
@@ -0,0 +1,139 @@
...
+   * @param tolerance Maximum absolute tolerance to terminate the algorithm.
+   * @param updateInterval The interval at which the objective is to be
+   *    reported and checked for convergence.
+   * @param descentPolicy The policy to use for picking up the coordinate to
+   *    descend on.
+   */
+  SCD(const double stepSize = 0.01,
+      const size_t maxIterations = 100000,
+      const double tolerance = 1e-5,
+      const size_t updateInterval = 1e3,
```
shikharbhardwaj: `updateInterval` is different from `batchSize`. It is the number of iterations after which we print the diagnostic information to the logs. As printing requires a call to `Evaluate` (which may take time), we need to make sure that the diagnostic info does not slow down the iteration (each iteration in SCD is expected to be much faster than `Evaluate`).

zoq: I see, thanks for the clarification.

rcurtin:
@shikharbhardwaj: thanks for the responses, everything looks good to me.  I think there are still two comments worth addressing---`SoftmaxRegressionFunction::PartialGradient()` could be accelerated still, and tests could be added for `RandomDescent` and `CyclicDescent`, but those are up to you.  I think it's fine to merge as-is regardless.  Thanks for the hard work, I think this is a nice addition. :+1:

shikharbhardwaj:
Sure, I had added tests for the descent policies (`CyclicDescentTest` and `RandomDescenTest` in the SCD test file). 
I had started with optimizing `PartialGradient` in  `SoftmaxRegressionFunction` but got confused. I'll start again with a clear head today.
I also looked into the idea of checking the function API for consistency at compile time. I guess this could be applied to almost all optimizers, so I'll do this as a separate PR.

rcurtin:
I think this is ready to merge; @mentekid: was there anything else you were waiting on for this one?

mentekid:
Sorry, I was planning to review this but forgot. I haven't looked at the latest push, but if you think it is ready to go just go ahead and merge.

Sorry for the delay, @shikharbhardwaj thanks for the contribution!

rcurtin:
Ok, sure, I'll go ahead and merge it then.  I think the latest changes we good.  Thanks! :)

PR #1077, opened 2017-08-03
Use omp_get_thread_num and omp_set_num_threads only if OpenMP supported.

zoq:
I was thinking how we could easily prevent such an issue, and I think we could configure travis to build with and without OpenMP support. I think using jenkins to test regular PR's might also be an option.


zoq:
@shikharbhardwaj Maybe you could look over the PR and let me know if there is anything that I missed?

shikharbhardwaj:
Thanks for the update. I'll go through other relevant places ASAP.

shikharbhardwaj:
I looked through and didn't find any other places with these issues (my first guess was the naive KMeans implementation, but it does not use these functions). I think its good to go.

I'll keep these things in mind in future OpenMP additions to the codebase.

PR #1085, opened 2017-08-09
Optimize tests

rcurtin:
I spent some time yesterday and today identifying tests that ran really slowly.  However, I am not sure how much of a difference this will make in general---on my desktop, running all tests takes only about 3 minutes.  On the other hand, running these tests on our sparc64 build servers takes 15 hours (according to Jenkins).  So I basically went through Jenkins and found tests that took more than about ~10 minutes on Jenkins, then attempted to speed them up.

Before, on my system (which has 8 cores);

```
$ time bin/mlpack_test
Running 978 test cases...

*** No errors detected

real    4m19.281s
user    6m48.364s
sys     9m45.672s
```

With these changes:

```
$ time bin/mlpack_test
Running 978 test cases...

*** No errors detected

real    2m45.430s
user    3m35.868s
sys 4m0.240s
```

That's an ok improvement, about 25-30% speedup I guess.  I think it will have a pretty big effect on the sparc64 systems.

This also fixes a few other test issues that I found; I added `MaxIterations()` to the `LRSDP` class, and fixed an apparent issue where `inputSize` and `targetSize` may not be set correctly for RNNs (that fix may not be the right way to do it, it should be double-checked).


src/mlpack/tests/krann_search_test.cpp
```diff
@@ -274,6 +274,7 @@ BOOST_AUTO_TEST_CASE(SingleDatasetSearch)
 // Test single-tree rank-approximate search with cover trees.
 BOOST_AUTO_TEST_CASE(SingleCoverTreeTest)
 {
+  math::RandomSeed(std::time(NULL));
```
zoq: Looks like you forgot to remove the line here, the same applies to ``DualCoverTreeTest`.

src/mlpack/methods/ann/rnn_impl.hpp
```diff
@@ -180,6 +180,10 @@ void RNN<OutputLayerType, InitializationRuleType>::SinglePredict(
     inputSize = predictors.n_elem / rho;
     prevRho = rho;
   }
+  else if (inputSize == 0)
```
zoq: I think we should initialize ``prevRho`` with 0, to remove the extra check here.

src/mlpack/methods/ann/rnn_impl.hpp
```diff
@@ -221,6 +225,11 @@ double RNN<OutputLayerType, InitializationRuleType>::Evaluate(
     targetSize = target.n_elem / rho;
     prevRho = rho;
   }
+  else if (inputSize == 0)
```
zoq: If we initalize ``prevRho`` we have to check the ``targetSize`` here, since if someone calls ``Predict`` on an untrained model first and trains it afterwards ``inputSize`` is set but ``targetSize`` isn't.

src/mlpack/tests/cf_test.cpp
```diff
@@ -32,15 +32,15 @@ BOOST_AUTO_TEST_CASE(CFGetRecommendationsAllUsersTest)
 {
   // Dummy number of recommendations.
   size_t numRecs = 3;
-  // GroupLens100k.csv dataset has 943 users.
-  size_t numUsers = 943;
+  // GroupLensSmall.csv dataset has 943 users.
```
zoq: But we only use 200 for testing?

rcurtin: Oops, this was just a typo---I made the dataset smaller, but I forgot to update the comment.

src/mlpack/methods/ann/rnn_impl.hpp
rcurtin: Ok, I think that I fixed this issue by initializing `prevRho`---or, at least, it passes the tests on my system.

PR #1088, opened 2017-08-09
Conventional Neural Evolution - Optimizer

kartik-nighania:
Works as an optimizer for vanilla network and converges using probability and evolution strategy


kartik-nighania:
Hello @zoq 
Can u look at the PR and suggest improvements and style fixes.. 

The test caseis under evaluation 
After this a tutorial is tobe made

src/mlpack/core/optimizers/cne/cne.hpp
```diff
@@ -0,0 +1,220 @@
...
+ * starting point will be modified to store the finishing point of the
+ * algorithm, and the final objective value is returned.
+ *
+ * @tparam DecomposableFunctionType Type of the function to be optimized.
+ * @param function Function to optimize.
+ * @param iterate Starting point (will be modified).
+ * @return Objective value of the final point.
+ */
+  template<typename DecomposableFunctionType>
+  double Optimize(DecomposableFunctionType& function, arma::mat& answer);
```
zoq: Should we use ``iterate`` instead of ``answer`` as parameter name, to be consistent with the rest of the codebase?

src/mlpack/core/optimizers/cne/cne_impl.hpp
```diff
@@ -0,0 +1,205 @@
...
+  << std::endl;
+
+  // Iterate till max number of generations
+  for (size_t gen = 1; gen <= maxGenerations; gen++)
+  {
+    // calculate fitness values of all candidates
+    for (size_t i = 0; i < populationSize; i++)
+     {
+       // select a candidate
+       parameters = population.row(i).t();
```
zoq: It looks like we could just use ``answer`` here instead to go with ``parameters``, that would avoid an extra copy. What do you think?

src/mlpack/core/optimizers/cne/cne_impl.hpp
```diff
@@ -0,0 +1,205 @@
...
+       // Insert the parameters in the function
+       answer = parameters;
+
+       // find the fitness
+       for (int j = 0; j < numFun; j++)
```
zoq: Can you use ``size_t`` instead of ``int`` her?

src/mlpack/core/optimizers/cne/cne_impl.hpp
```diff
@@ -0,0 +1,205 @@
...
+       fitness += function.Evaluate(parameters, j);
```
zoq: Pedantic style issue; missing space.

src/mlpack/core/optimizers/cne/cne_impl.hpp
```diff
@@ -0,0 +1,205 @@
...
+
+       // Save fitness values
+       fitnessValues[i] = fitness;
+       fitness = 0;
+     }
+
+      Log::Info << "Generation number: " << gen << " best fitness = "
+      << (double)fitnessValues.min() << std::endl;
```
zoq: The type cast is unnecessary, ``fitnessValues.min() `` should be ``double``.

src/mlpack/core/optimizers/cne/cne_impl.hpp
```diff
@@ -0,0 +1,205 @@
...
+
+      // create the next generation of species
+      Reproduce();
+
+      // check for termination
+      if (finalValue >= (double)fitnessValues.min())
+        {
```
zoq: The opening and closing ``{}``should align the the if condition.

src/mlpack/core/optimizers/cne/cne_impl.hpp
```diff
@@ -0,0 +1,205 @@
...
+
+      // Store the best fitness of this generation before update
+      lastBestFitness = (double)fitnessValues.min();
+  }
+
+  // Set the best candidate into the network
+  answer = population.submat(index[0], 0, index[0], populationSize-1);
+
+  // find the best fitness
+  for (int j = 0; j < numFun; j++)
```
zoq: Do you mind to use ``size_t`` as index type here?

src/mlpack/core/optimizers/cne/cne_impl.hpp
```diff
@@ -0,0 +1,205 @@
...
+          << " < " << fitnessHist << std::endl;
+          break;
+        }
+
+      // Store the best fitness of this generation before update
+      lastBestFitness = (double)fitnessValues.min();
+  }
+
+  // Set the best candidate into the network
+  answer = population.submat(index[0], 0, index[0], populationSize-1);
```
zoq: Pedantic style issue; missing space between ``populationSize``, ``-`` and ``1``.

src/mlpack/core/optimizers/cne/cne_impl.hpp
```diff
@@ -0,0 +1,205 @@
...
+  size_t numElite = floor(selectPercent * populationSize);
+
+  // Making sure we have even number of candidates to remove and create
+  if ((populationSize - numElite) % 2 != 0) numElite++;
+
+  for (size_t i = numElite; i < populationSize-1; i++)
+  {
+    // Select 2 parents from the elite group randomly [0, numElite)
+    size_t mom = mlpack::math::RandInt(0, numElite);
+    size_t dad = mlpack::math::RandInt(0, numElite);
```
zoq: We should make sure we are not going to select the same parent.

src/mlpack/core/optimizers/cne/cne_impl.hpp
```diff
@@ -0,0 +1,205 @@
...
+  // Remove the cadidates and instead place the parents
+  population.row(child1) = population.row(mom);
+  population.row(child2) = population.row(dad);
+
+  double rand;
+  // Randomly alter mom and dad genome data to get two childs
+  for (size_t i = 0; i < population.n_cols; i++)
+  {
+    // Select a random value from the normal distribution
+    rand = mlpack::math::RandNormal();
```
zoq: Let us use a uniform random number between 0 and 1 (``mlpack::math::Random``) here and ``0.5`` as condition.

src/mlpack/core/optimizers/cne/cne_impl.hpp
```diff
@@ -0,0 +1,205 @@
...
+{
+  // Helper variables
+  double noise;
+  double delta;
+
+  // Mutate the whole matrix with the given rate and probability
+  // Note: The best candidate is not altered
+  for (size_t i = 1; i < populationSize; i++)
+  {
+    for (size_t j = 0; j < population.n_cols; j++)
```
zoq: I think, depending on the value of  ``mutationProp`` something like:

```c++
population.row(index(i)) = (arma::randu<vec>(population) < mutationProb) % arma::randn<vec>(population);
```

could be faster, since we could avoid any branch prediction.

src/mlpack/core/optimizers/cne/cne.hpp
kartik-nighania: done

src/mlpack/core/optimizers/cne/cne_impl.hpp
kartik-nighania: ah yes.. i missed it

src/mlpack/core/optimizers/cne/cne_impl.hpp
kartik-nighania: done

src/mlpack/core/optimizers/cne/cne_impl.hpp
kartik-nighania: @zoq space around parameters ? (which is now iterate)

src/mlpack/core/optimizers/cne/cne_impl.hpp
kartik-nighania: done

src/mlpack/core/optimizers/cne/cne_impl.hpp
kartik-nighania: done

src/mlpack/core/optimizers/cne/cne_impl.hpp
kartik-nighania: done

src/mlpack/core/optimizers/cne/cne_impl.hpp
kartik-nighania: done .. i missed it actually..

src/mlpack/core/optimizers/cne/cne_impl.hpp
zoq: two space before ``titness``:

```c++
for (int j = 0; j < numFun; j++)
  fitness += function.Evaluate(parameters, j);
```

kartik-nighania:
@zoq 
im thinking of using XOR task as a test case ?

also for the xor using FNN
and training with 
0,0
1,1
0,1
1,0
and labels as
1
1
2
2

for 2 input 2 hidded and 2 output layer
in the FNN model 
and it gives me an error ..
its kindoff a silly doubt.. 
but i am not able to figure this out ..
the output is error : saying matrix incompatible

```
  model.Add<Linear<> >(t2, 2);
  model.Add<SigmoidLayer<> >();
  model.Add<Linear<> >(2, 2);
  model.Add<LogSoftMax<> >();

```

zoq:
Testing it on XOR sounds good to me; about the issue, you use two nodes as output, but your label is just a single node. I would use 0, 1 as the label and only a single output node:

```c++
model.Add<Linear<> >(t2, 2);
model.Add<SigmoidLayer<> >();
model.Add<Linear<> >(2, 1);
model.Add<LogSoftMax<> >();
```

kartik-nighania:
@zoq 
in the test case added
```
error: Mat::operator(): index out of bounds
  what():  Mat::operator(): index out of bounds
Aborted (core dumped)
```
why am i getting this error. Need help

kartik-nighania:
i get the same error using StandardSGD also

kartik-nighania:
@zoq 
please suggest me improvmenets for the tutorial doc added.
thanks

src/mlpack/tests/cne_test.cpp
```diff
@@ -0,0 +1,59 @@
...
+
+#include <mlpack/core/optimizers/cne/cne.hpp>
+
+using namespace mlpack;
+using namespace mlpack::ann;
+using namespace mlpack::optimization;
+
+BOOST_AUTO_TEST_SUITE(CNE_Test);
+
+BOOST_AUTO_TEST_CASE(CNE_XOR_TEST)
```
zoq: Would be interesting to see if we are able to optimize the logistic regression example.

src/mlpack/core/optimizers/cne/cne_impl.hpp
```diff
@@ -0,0 +1,202 @@
...
+namespace optimization {
+
+CNE::CNE(const size_t populationSize,
+         const size_t maxGenerations,
+         const double mutationProb,
+         const double mutationSize,
+         const double selectPercent,
+         const double finalValue,
+         const double fitnessHist) :
+         populationSize(populationSize),
```
zoq: Pedantic style issue; we should tap here once (two spaces), that only applies to the constructor initialization list.

src/mlpack/core/optimizers/cne/cne_impl.hpp
```diff
@@ -0,0 +1,202 @@
...
+template<typename DecomposableFunctionType>
+double CNE::Optimize(
+    DecomposableFunctionType& function,
+    arma::mat& iterate)
+{
+  // Get the number of functions to iterate
+  const size_t numFun = function.NumFunctions();
+
+  // Set the population size and fill random values [0,1]
+  population.set_size(populationSize, iterate.n_rows);
```
zoq: We could just use ``population = arma::randu(populationSize, iterate.n_rows));`` no need to set the size first.

src/mlpack/core/optimizers/cne/cne_impl.hpp
```diff
@@ -0,0 +1,202 @@
...
+  population.randu();
+
+  // initialize helper variables
+  fitnessValues.set_size(populationSize);
+  double fitness = 0;
+  double lastBestFitness;
+
+  Log::Info << "CNE initialized successfully. Optimization started"
+  << std::endl;
```
zoq: Pedantic style issue number 2; tape twice (4 spaces).

doc/tutorials/cne/cne.txt
```diff
@@ -0,0 +1,233 @@
...
+void Predict(arma::mat predictors, arma::mat& results);
+@endcode
+
+Given the data to predict in armadillo matrix format. Matrix result is modified
+and the output of prediction is stored in it.
+
+@subsection cne_ex1_cnetut The constructor parameters.
+
+@code
+CNE(const size_t populationSize = 15,
```
zoq: We should use the same default values here as for the definition, e.g. 500 for the ``populationSize`` instead of 15.

src/mlpack/core/optimizers/cne/cne_impl.hpp
kartik-nighania: done

src/mlpack/core/optimizers/cne/cne_impl.hpp
kartik-nighania: done

src/mlpack/core/optimizers/cne/cne_impl.hpp
kartik-nighania: done

src/mlpack/core/optimizers/cne/cne_impl.hpp
kartik-nighania: done

src/mlpack/core/optimizers/cne/cne_impl.hpp
kartik-nighania: thanks ..
will remember this 
done

doc/tutorials/cne/cne.txt
kartik-nighania: @zoq 
sorry for that ..
thanks

src/mlpack/tests/cne_test.cpp
kartik-nighania: done

src/mlpack/core/optimizers/cne/cne_impl.hpp
kartik-nighania: should i use hasing then ?

src/mlpack/core/optimizers/cne/cne_impl.hpp
kartik-nighania: @zoq 
delta = mlpack::math::RandNormal(0, mutationSize);

after this to call randnormal iteratively and add the delta values to non zero locations?

src/mlpack/core/optimizers/cne/cne_impl.hpp
kartik-nighania: @zoq 
u mean the same parent last time? as if select population will be just 10% this will cause a problem .. and same parent is bound to happen. Did you mean same pair of parents ?
sorry im not clear with your words..

zoq: I mean that we should make sure, that ``mon != dad`` since we choose both randomly, we could end up with the same random number.

src/mlpack/core/optimizers/cne/cne_impl.hpp
zoq: I think we can avoid any for loop here, in my case: ``arma::randn<vec>(population);`` is the delta and ``(arma::randu<vec>(population) < mutationProb)``  the if condition combined with the delta. We will multiply the delta vector with zero if the condition isn't true, which is the non existing else case in your code.

src/mlpack/tests/cne_test.cpp
```diff
@@ -0,0 +1,137 @@
...
+using namespace mlpack;
+using namespace mlpack::ann;
+using namespace mlpack::optimization;
+
+using namespace mlpack::distribution;
+using namespace mlpack::regression;
+
+BOOST_AUTO_TEST_SUITE(CNE_Test);
+
+BOOST_AUTO_TEST_CASE(CNE_XOR_TEST)
```
zoq: Can you add a simple comment for each test that sums up what the test does?

kartik-nighania: done
Also @zoq
 the thyroid dataset test for CNE takes 235.753 seconds in my local machine.
Is it okay?

src/mlpack/core/optimizers/cne/cne.hpp
```diff
@@ -0,0 +1,220 @@
...
+/**
+ * Constructor for the CNE optimizer.
+ * 
+ * The default values provided over here are not necessarily suitable for a
+ * given function. Therefore it is highly recommended to adjust the
+ * parameters according to the problem.
+ *
+ * @param populationSize The number of candidates in the population
+ * @param maxGenerations The maximum number of generations allowed for CNE
+ * @param mutationProb   Probability that a weight will get mutated
```
zoq: Minor style issue, the description should come right after the parameter, not need to align the descriptions.

src/mlpack/core/optimizers/cne/cne.hpp
```diff
@@ -0,0 +1,220 @@
...
+ * parameters according to the problem.
+ *
+ * @param populationSize The number of candidates in the population
+ * @param maxGenerations The maximum number of generations allowed for CNE
+ * @param mutationProb   Probability that a weight will get mutated 
+ * @param mutationSize   The range of mutation noise to be added. This range
+ *                       is between 0 and mutationSize
+ * @param selectPercent  The percentage of candidates to select to become the
+ *                       the next generation
+ * @param finalValue     The final value of the objective function for
```
zoq: Do you think we could rename the parameter to tolerance? That way we could use the same name as we did for SGD.

src/mlpack/core/optimizers/cne/cne.hpp
```diff
@@ -0,0 +1,220 @@
...
+ * @param populationSize The number of candidates in the population
+ * @param maxGenerations The maximum number of generations allowed for CNE
+ * @param mutationProb   Probability that a weight will get mutated 
+ * @param mutationSize   The range of mutation noise to be added. This range
+ *                       is between 0 and mutationSize
+ * @param selectPercent  The percentage of candidates to select to become the
+ *                       the next generation
+ * @param finalValue     The final value of the objective function for  
+ *                       termination. Not considered if not provided.
+ * @param fitnessHist    Minimum change in best fitness values between two
```
zoq: Not sure, but maybe ``objectiveChange`` is a better name?

src/mlpack/core/optimizers/cne/cne.hpp
```diff
@@ -0,0 +1,220 @@
...
+  void Reproduce();
+
+  //! Function to modify weights for the evolution of next generation
+  void Mutate();
+
+  /**
+   * Crossover parents and create new childs. 
+   * Two parents create two new childs.
+   *
+   * @param mom      First parent from the elite population
```
zoq: Same style issue as above, no need to align the description.

src/mlpack/core/optimizers/cne/cne.hpp
```diff
@@ -0,0 +1,220 @@
...
+   * @param dad      Second parent from the elite population
+   * @param dropout1 The place to delete the candidate of the present
+   *                 generation and place a child over there for the
+   *                 next generation
+   * @param dropout2 The place to delete the candidate of the present
+   *                 generation and place a child over there for the
+   *                 next generation
+   */
+  void Crossover(size_t mom, size_t dad, size_t dropout1, size_t dropout2);
```
zoq: Do you think we could use ``const`` here?

src/mlpack/core/optimizers/cne/cne.hpp
```diff
@@ -0,0 +1,220 @@
...
+  //! Index of the sorted fitness values
+  arma::uvec index;
+
+  //! The number of candidates in the population
+  size_t populationSize;
+
+  //! Maximum number of generations before termination criteria is met
+  size_t maxGenerations;
+
+  //!  The range of mutation noise to be added
```
zoq: Pedantic style issue, remove the additional space.

src/mlpack/core/optimizers/cne/cne_impl.hpp
```diff
@@ -0,0 +1,202 @@
...
+#ifndef MLPACK_CORE_OPTIMIZERS_CNE_CNE_IMPL_HPP
+#define MLPACK_CORE_OPTIMIZERS_CNE_CNE_IMPL_HPP
+
+#include "cne.hpp"
+
+namespace mlpack {
+namespace optimization {
+
+CNE::CNE(
+    const size_t populationSize,
```
zoq: I think you can start right after ``CNE`` so something like:

```c++
CNE::CNE(const size_t populationSize,
         const size_t maxGenerations,
         const double mutationProb,
         const double mutationSize,
         const double selectPercent,
         const double finalValue,
         const double fitnessHist) :
    populationSize(populationSize),
    maxGenerations(maxGenerations),
    mutationProb(mutationProb),
    mutationSize(mutationSize),
    selectPercent(selectPercent),
    finalValue(finalValue),
    fitnessHist(fitnessHist)
```

src/mlpack/core/optimizers/cne/cne_impl.hpp
```diff
@@ -0,0 +1,202 @@
...
+    mutationSize(mutationSize),
+    selectPercent(selectPercent),
+    finalValue(finalValue),
+    fitnessHist(fitnessHist)
+{ /* Nothing to do here. */ }
+
+//! Optimize the function
+template<typename DecomposableFunctionType>
+double CNE::Optimize(
+    DecomposableFunctionType& function,
```
zoq: I would start right after ``CNE::Optimize(``.

src/mlpack/core/optimizers/cne/cne_impl.hpp
```diff
@@ -0,0 +1,202 @@
...
+    arma::mat& iterate)
+{
+  // Get the number of functions to iterate
+  const size_t numFun = function.NumFunctions();
+
+  // Set the population size and fill random values [0,1]
```
zoq: Missing ``.`` at the end.

src/mlpack/core/optimizers/cne/cne_impl.hpp
```diff
@@ -0,0 +1,202 @@
...
+  Log::Info << "CNE initialized successfully. Optimization started"
+      << std::endl;
+
+  // Iterate till max number of generations
+  for (size_t gen = 1; gen <= maxGenerations; gen++)
+  {
+    // calculate fitness values of all candidates
+    for (size_t i = 0; i < populationSize; i++)
+    {
+       // select a candidate and insert the parameters in the function
```
zoq: Please use complete sentences with proper grammar and punctuation. In this case Starting with ``Select`` and add the missing ``.`` at the end will do the trick.

src/mlpack/core/optimizers/cne/cne_impl.hpp
```diff
@@ -0,0 +1,202 @@
...
+       iterate = population.row(i).t();
```
zoq: I think we can avoid the transpose here if we would use col instead of row? That would mean that we modify the reproduce code as well. Let me know what you think.

src/mlpack/core/optimizers/cne/cne_impl.hpp
```diff
@@ -0,0 +1,202 @@
...
+
+       // find the fitness
```
zoq: Minor style issue; please use complete sentences.

src/mlpack/core/optimizers/cne/cne_impl.hpp
```diff
@@ -0,0 +1,202 @@
...
+       for (size_t j = 0; j < numFun; j++)
+         fitness += function.Evaluate(iterate, j);
+
+       // Save fitness values
```
zoq: See my comment above.

src/mlpack/core/optimizers/cne/cne_impl.hpp
```diff
@@ -0,0 +1,202 @@
...
+    }
+
+      Log::Info << "Generation number: " << gen << " best fitness = "
+          << fitnessValues.min() << std::endl;
+
+      // create the next generation of species
+      Reproduce();
+
+      // check for termination
+      if (finalValue != DBL_MIN && finalValue >= fitnessValues.min())
```
zoq: Do we have to check for ``DBL_MIN`` here?

src/mlpack/core/optimizers/cne/cne_impl.hpp
```diff
@@ -0,0 +1,202 @@
...
+      {
+          Log::Info << "Terminating. Given fitness criteria " << finalValue
+              << " > " << fitnessValues.min() << std::endl;
+          break;
+      }
+
+      // check for termination
+      if (fitnessHist != DBL_MIN && gen != 1 &&
```
zoq: Instead of checking against ``gen == 1`` we could lastBestFitness fitness at the beginning by iterating through the values with the initial parameter.

src/mlpack/core/optimizers/cne/cne_impl.hpp
```diff
@@ -0,0 +1,202 @@
...
+          << " < " << fitnessHist << std::endl;
+          break;
+      }
+
+      // Store the best fitness of this generation before update
+      lastBestFitness = fitnessValues.min();
+  }
+
+  // Set the best candidate into the network
+  iterate = population.submat(index[0], 0, index[0], iterate.n_rows - 1).t();
```
zoq: Maybe I missed something, but I think we could use ``population.row(index(0))`` here.

src/mlpack/core/optimizers/cne/cne_impl.hpp
```diff
@@ -0,0 +1,202 @@
...
+  // Sort fitness value. The smaller the better
+  index = arma::sort_index(fitnessValues);
+
+  // Find the number of elite percentage
+  size_t numElite = floor(selectPercent * populationSize);
+
+  // Making sure we have even number of candidates to remove and create
+  if ((populationSize - numElite) % 2 != 0) numElite++;
+
+  for (size_t i = numElite; i < populationSize-1; i++)
```
zoq: Pedantic style issue; missing space between ``populationSize`` and the operation.

src/mlpack/tests/cne_test.cpp
```diff
@@ -0,0 +1,215 @@
...
+
+using namespace mlpack::distribution;
+using namespace mlpack::regression;
+
+BOOST_AUTO_TEST_SUITE(CNE_Test);
+
+/**
+ * Training a vanilla network for 2 input XOR function
+ */
+BOOST_AUTO_TEST_CASE(CNE_XOR_TEST)
```
zoq: We should remove the underscore here, to be consistent with the rest of the codebase.

src/mlpack/core/optimizers/cne/cne.hpp
kartik-nighania: done

src/mlpack/core/optimizers/cne/cne.hpp
kartik-nighania: done

src/mlpack/core/optimizers/cne/cne.hpp
kartik-nighania: done

src/mlpack/core/optimizers/cne/cne.hpp
kartik-nighania: done

src/mlpack/core/optimizers/cne/cne.hpp
kartik-nighania: done

src/mlpack/core/optimizers/cne/cne_impl.hpp
kartik-nighania: done

src/mlpack/core/optimizers/cne/cne_impl.hpp
kartik-nighania: done

src/mlpack/core/optimizers/cne/cne_impl.hpp
kartik-nighania: done

src/mlpack/core/optimizers/cne/cne_impl.hpp
kartik-nighania: @zoq 
ive done that to all comments. Please tell me if some more correction is needed.

Thanks a lot

src/mlpack/core/optimizers/cne/cne_impl.hpp
kartik-nighania: done for all

src/mlpack/core/optimizers/cne/cne_impl.hpp
kartik-nighania: done for all

src/mlpack/core/optimizers/cne/cne_impl.hpp
kartik-nighania: This is the default value which says this is not given as parameter by the user and so is ignored.
Ive mentioned this in the tutorial too for the user to understand.

src/mlpack/core/optimizers/cne/cne_impl.hpp
kartik-nighania: right. thanks

src/mlpack/core/optimizers/cne/cne_impl.hpp
kartik-nighania: done

src/mlpack/tests/cne_test.cpp
kartik-nighania: done

src/mlpack/core/optimizers/cne/cne_impl.hpp
zoq: I think it would be more intuitive to set ``finalValue`` to something small e.g. ``1e-5`` and if someone likes to run the maximum number of iterations he can set the value to ``-1`` or something like that. What do you think?

src/mlpack/core/optimizers/cne/cne_impl.hpp
kartik-nighania: done 
thanks for this one

src/mlpack/core/optimizers/cne/cne_impl.hpp
kartik-nighania: done

src/mlpack/core/optimizers/cne/cne_impl.hpp
kartik-nighania: yes changed the matrix. Now column has candidate

src/mlpack/core/optimizers/cne/cne.hpp
kartik-nighania: @zoq 
made all the parameters const in crossover.
and instead using 
```
    // Parents generates 2 childs replacing the droped out candidates.
    // The index of the candidates in the population is found first. 
    Crossover(index[mom], index[dad], index[i], index[i+1]);
```
and then 
```
//! Crossover parents to create new childs.
void CNE::Crossover(const size_t mom,
                    const size_t dad,
                    const size_t child1,
                    const size_t child2)
```

Just mentioning. I think the previous way of finding the index inside crossover was looking more beautiful and friendly to read. 

```
//! Crossover parents to create new childs.
void CNE::Crossover(size_t mom,
                    size_t dad,
                    size_t child1,
                    size_t child2)
{
  // Find the index of these candidates in the population matrix.
  mom = index[mom];
  dad = index[dad];
  child1 = index[child1];
  child2 = index[child2];
```

kartik-nighania:
@zoq 
- Test Iris dataset now works fine and the time is now down from timeout -> 247 sec -> 4.91 / 1.62 in both the machines.
- some variables are taken out of reproduction to avoid multiple time same calculations
- two false input checks have been added 

- 2 minor code change remains, as suggested by u today

Thanks a lot for spending so much time for this in depth code review. It was very helpful

src/mlpack/core/optimizers/cne/cne_impl.hpp
```diff
@@ -0,0 +1,227 @@
...
+  for (size_t i = numElite; i < populationSize - 1; i++)
+  {
+    // Select 2 different parents from elite group randomly [0, numElite).
+    do {
+         mom = mlpack::math::RandInt(0, numElite);
+         dad = mlpack::math::RandInt(0, numElite);
+       } while (mom == dad);
+
+    // Parents generates 2 childs replacing the droped out candidates.
+    Crossover(mom, dad, i, i+1);
```
zoq: Style issue; missing space between ``i`` and ``+``.

src/mlpack/core/optimizers/cne/cne_impl.hpp
```diff
@@ -0,0 +1,227 @@
...
+  }
+
+  // Mutating the weights with small noise values.
+  // This is done to bring change in the next generation.
+  Mutate();
+}
+
+//! Crossover parents to create new childs.
+void CNE::Crossover(size_t mom,
```
zoq: The function doesn't match with the definition (const).

src/mlpack/core/optimizers/cne/cne_impl.hpp
```diff
@@ -0,0 +1,227 @@
...
+    numElite(0)
+{ /* Nothing to do here. */ }
+
+//! Optimize the function.
+template<typename DecomposableFunctionType>
+double CNE::Optimize(DecomposableFunctionType& function, arma::mat& iterate)
+{
+  // Make sure for evolution to work at least four candidates are present.
+  if (populationSize < 4)
+  Log::Fatal << "Population size should be atleast 4." << std::endl;
```
kartik-nighania: @zoq 
Done.
Thanks
Just wanted to know how is better ?

src/mlpack/core/optimizers/cne/cne_impl.hpp
```diff
@@ -0,0 +1,227 @@
...
+          break;
+      }
+
+      // Check for termination criteria.
+      if (objectiveChange != DBL_MIN &&
+        (lastBestFitness - fitnessValues.min()) < objectiveChange)
+      {
+          Log::Info << "Terminating. Fitness History change "
+              << (lastBestFitness - fitnessValues.min())
+          << " < " << objectiveChange << std::endl;
```
kartik-nighania: done

src/mlpack/core/optimizers/cne/cne_impl.hpp
```diff
@@ -0,0 +1,227 @@
...
+      // Store the best fitness of present generation.
+      lastBestFitness = fitnessValues.min();
+  }
+
+  // Set the best candidate into the network parameters.
+  iterate = population.col(index(0));
+
+  // Find the objective function value from the best candidate found.
+  for (size_t j = 0; j < numFun; j++)
+      fitness += function.Evaluate(iterate, j);
```
kartik-nighania: done

src/mlpack/core/optimizers/cne/cne.hpp
```diff
@@ -0,0 +1,226 @@
...
+{
+ public:
+/**
+ * Constructor for the CNE optimizer.
+ * 
+ * The default values provided over here are not necessarily suitable for a
+ * given function. Therefore it is highly recommended to adjust the
+ * parameters according to the problem.
+ *
+ * @param populationSize The number of candidates in the population
```
kartik-nighania: done

src/mlpack/core/optimizers/cne/cne_impl.hpp
kartik-nighania: done

src/mlpack/core/optimizers/cne/cne_impl.hpp
zoq: If you throw an exception there is a chance to recover from that, ``Log::Fatal`` calls ``exit`` afterwards.

src/mlpack/core/optimizers/cne/cne_impl.hpp
```diff
@@ -0,0 +1,227 @@
...
+  size_t mom;
+
+  // Second parent.
+  size_t dad;
+
+  for (size_t i = numElite; i < populationSize - 1; i++)
+  {
+    // Select 2 different parents from elite group randomly [0, numElite).
+    do {
+         mom = mlpack::math::RandInt(0, numElite);
```
kartik-nighania: @zoq 
should i go for using an armadillo vector of shuffled number using linspace. That way i guess all the parents can get used for sure. But i think that will decrease the amount of randomness.

kartik-nighania: to use next or previous index will then call for check if index is out of bounds [0, numElite)

src/mlpack/core/optimizers/cne/cne_impl.hpp
kartik-nighania: oh.. thanks

src/mlpack/core/optimizers/cne/cne_impl.hpp
kartik-nighania:
```
   arma::mat delta = randu<arma::mat>( size(population) );
   delta.elem( arma::find(delta < mutationProb) ).fill( mlpack::math::RandNormal(0, mutationSize) );

   population += delta;
```

@zoq  I came up with this. 
What do you think ? the random values are the same. 
should i set the seed value to 0 or something to get differnt random values.

src/mlpack/core/optimizers/cne/cne_impl.hpp
zoq: I don't think we have to use shuffel here, we just have to handle the case where mom = dad efficiently, so, checking the bounds is probably the most efficient solution.

kartik-nighania: thanks 
done

src/mlpack/core/optimizers/cne/cne_impl.hpp
zoq: I think you mean ``RandNormal`` returns the same values if you rerun the method? But it does not return the same values if you call ``RandNormal`` multiple times in the same execution. Since you use a static seed this is expected behaviour. If you like to see different values each time you run the test you can write ``math::RandomSeed(std::time(NULL));`` before you run the test.

kartik-nighania:
@zoq 
```
//! Modify weights with some noise for the evolution of next generation.
void CNE::Mutate()
{
  arma::vec bestCandidate = population.col(index[0]);

  // Mutate the whole matrix with the given rate and probability.
  arma::mat delta = arma::randu<arma::mat>(size(population));
  delta.elem(arma::find(delta < mutationProb)).fill(mlpack::math::RandNormal(0, mutationSize));

  population += delta;
  population.col(index[0]) = bestCandidate;
}
```
.fill takes a finite value, that is the problem.
I cant think of anything that takes both mutationProb and mutationSize into consideration otherwise the method becomes more time consuming than the iterative process itself.
what should i do?

src/mlpack/core/optimizers/cne/cne_impl.hpp
kartik-nighania: done

kartik-nighania:
@zoq 
after the above suggested change, I think this PR is complete. Please let me know for further improvements.

zoq:
There is one thing that I missed, it's not granted that iterate is a single column matrix. So I propose to use arma::cube instead of arma::mat to store the population, so:

```c++
population = arma::randu(iterate.n_rows, populationSize);
```

becomes

```c++
population = arma::randu(iterate.n_rows, iterate.n_cols, populationSize);
```

here is the refactored Corossover function:

```c++
void CNE::Crossover(const size_t mom,
                    const size_t dad,
                    const size_t child1,
                    const size_t child2)
{
  // Replace the cadidates with parents at their place.
  population.slice(child1) = population.slice(mom);
  population.slice(child2) = population.slice(dad);

  double rand;

  // Preallocate random selection vector (values between 0 and 1).
  arma::vec selection = arma::randu(population.n_rows * population.n_cols);

  // Randomly alter mom and dad genome weights to get two different childs.
  for (size_t i = 0; i < population.n_elem; i++)
  {
    // Using it to alter the weights of the childs.
    if (selection(i) > 0.5)
    {
      population.slice(child1)(i) = population.slice(mom)(i);
      population.slice(child2)(i) = population.slice(dad)(i);
    }
    else
    {
      population.slice(child1)(i) = population.slice(dad)(i);
      population.slice(child2)(i) = population.slice(mom)(i);
    }
  }
}
```

About the ``Mutate`` I think we can write something like:

```c++
  double mutationProb = 0.5;
  double mutationSize = 0.02;

  // noise
  std::cout << arma::randu(5, 5) << std::endl;

  // noise < mutationProp
  std::cout << (arma::randu(5, 5) < mutationProb) << std::endl;

  // delta
  std::cout << mutationSize * arma::randn(5, 5) << std::endl;

  // complete update matrix
  std::cout << (arma::randu(5, 5) < mutationProb) % (mutationSize * arma::randn(5, 5)) << std::endl;
```

here is the output:

```
  // noise
   0.0002   0.0361   0.8205   0.1831   0.6022
   0.0254   0.0820   0.9466   0.4364   0.7958
   0.3789   0.6139   0.1122   0.1771   0.5465
   0.5490   0.6238   0.7952   0.8186   0.4000
   0.2536   0.4979   0.2296   0.5374   0.4618

  // noise < mutationProp
        1        0        0        0        0
        1        0        0        0        0
        1        1        1        0        1
        0        1        0        1        1
        1        0        1        0        1
  // delta
  -0.0070   0.0105   0.0165   0.0127  -0.0195
  -0.0398   0.0156  -0.0072   0.0012   0.0112
  -0.0107  -0.0040   0.0280   0.0252   0.0081
   0.0148   0.0249  -0.0201  -0.0005  -0.0198
  -0.0222   0.0085   0.0194  -0.0020   0.0025

  // complete update matrix
        0   0.0038   0.0457   0.0048   0.0024
   0.0212        0  -0.0203        0        0
        0        0  -0.0067        0  -0.0054
  -0.0004  -0.0026        0   0.0045        0
   0.0090        0        0  -0.0170        0
```

so ``Mutate`` becomes:

```c++
void CNE::Mutate()
{
  // Helper variables.
  double noise;
  double delta;

  // Mutate the whole matrix with the given rate and probability.
  // The best candidate is not altered.
  for (size_t i = 1; i < populationSize; i++)
  {
    population.slice(index(i)) += (arma::randu(
        population.n_rows, population.n_cols) < mutationProb) %
        (mutationSize * arma::randn(population.n_rows, population.n_cols));
  }
}
```

Let me know if I should clarify anything.

kartik-nighania:
@zoq
Ah.. i see. I was thinking of how to make sure less than mutationSize will be taken care of. But multiplying it with a quantity less than one ofcourse takes care of that. 
Thanks

kartik-nighania:
@zoq 
I think the PR is complete now.  :v:
Thanks

src/mlpack/core/optimizers/cne/cne.hpp
```diff
@@ -0,0 +1,231 @@
+/**
+ * @file cne.hpp
+ * @author Marcus Edel
+ * @author Kartik Nighania
+ *
+ * Conventional Neural Evolution
```
zoq: Can you elaborate on this, a simple sentence what this method does is just fine.

src/mlpack/core/optimizers/cne/cne.hpp
```diff
@@ -0,0 +1,231 @@
...
+ * double Evaluate(const arma::mat& coordinates, const size_t i);
+ *
+ * The parameter i refers to which individual function is being evaluated.
+ * The function will evaluate the objective function on the data point i in
+ * the dataset.
+ */
+class CNE
+{
+ public:
+/**
```
zoq: Tab once here (two spaces). This comment applies for the other definitions as well.

src/mlpack/core/optimizers/cne/cne_impl.hpp
```diff
@@ -0,0 +1,220 @@
+/**
+ * @file cne_impl.hpp
+ * @author Marcus Edel
+ * @author Kartik Nighania
+ *
+ * Conventional Neural Evolution
```
zoq: Can you elaborate on this, a simple sentence what this method does is just fine.

src/mlpack/core/optimizers/cne/cne_impl.hpp
```diff
@@ -0,0 +1,220 @@
...
+  for (size_t i = numElite; i < populationSize - 1; i++)
+  {
+    // Select 2 different parents from elite group randomly [0, numElite).
+    mom = mlpack::math::RandInt(0, numElite);
+    dad = mlpack::math::RandInt(0, numElite);
+
+    // Making sure both parents are not the same.
+    if (mom == dad)
+    {
+      if (dad != numElite - 1) dad++;
```
zoq: Pedantic style issue; you can write:

```c++
if (dad != numElite - 1)
{
  dad++;
}
else
{
  dad--;
}
```

you could also use a ternary operator:

```c++
dad = dad != numElite - 1 ? dad + 1 : dad - 1;
```

src/mlpack/tests/cne_test.cpp
```diff
@@ -0,0 +1,200 @@
...
+  /*
+   * Create the four cases for XOR with two variable
+   *
+   *  Input    Output
+   * 0 XOR 0  =  0
+   * 1 XOR 1  =  0 
+   * 0 XOR 1  =  1
+   * 1 XOR 0  =  1
+   */
+  arma::mat train("1,0,0,1;1,0,1,0");
```
zoq: Pedantic style issue; can you use a space between each ``,``?

kartik-nighania: done

src/mlpack/core/optimizers/cne/cne_impl.hpp
kartik-nighania: done

src/mlpack/core/optimizers/cne/cne_impl.hpp
kartik-nighania: done

src/mlpack/core/optimizers/cne/cne.hpp
kartik-nighania: done. Thanks

src/mlpack/core/optimizers/cne/cne.hpp
kartik-nighania: done

src/mlpack/core/optimizers/cne/cne.hpp
```diff
@@ -0,0 +1,234 @@
...
+ * The algorithm works by creating a fixed number of candidates. The candidates
+ * are simply the network parameters that are untrained. Given the train and 
+ * test set. Each candidate is tested upon the training set or 1 epoch and then
+ * a fitness score is assigned to it. Given the selection percentage of best
+ * candidates by the user, for a single generation that many percentage of 
+ * candidates were selected for the next generation and the rest were removed
+ * The selected candidates for a particular generation then become the parent 
+ * for the next generation and evolution takes place.
+ *
+ * The evolution process basically takes place in two types-
```
zoq: Should we end with ``types:`` and start a listing by using ``- Crossover`` and ``- Mutation``?

src/mlpack/core/optimizers/cne/cne.hpp
```diff
@@ -0,0 +1,234 @@
...
+ * Crossover
+ * Mutation 
+ *
+ * Crossover takes two parents and generates two childs from them. Both the
+ * child have properties inherited by their parents. The parameters of the
+ * parents gets mixed using equal probability selection, creating two childs.
+ * This is just a mix of link weights or the parameters.
+ *
+ * In mutation parameters are updated by pertubating small noise.
+ * If $$ \Lambda $$ is the number of weights in the network.
```
zoq: Have you tested the formulas renderd correctly? I guess you should use: ``\f[`` and ``\f$`` for more informatik take a look at: https://www.stack.nl/~dimitri/doxygen/manual/formulas.html

src/mlpack/core/optimizers/cne/cne.hpp
```diff
@@ -0,0 +1,234 @@
...
+   * Constructor for the CNE optimizer.
+   * 
+   * The default values provided over here are not necessarily suitable for a
+   * given function. Therefore it is highly recommended to adjust the
+   * parameters according to the problem.
+   *
+   * @param populationSize The number of candidates in the population.
+   *        This should be atleast 4 in size.
+   * @param maxGenerations The maximum number of generations allowed for CNE.
+   * @param mutationProb Probability that a weight will get mutated
```
zoq: I know I'm really picky about the style here, but can you add a ``.`` at the end of each parameter description? Please check the other descriptions as well.

src/mlpack/core/optimizers/cne/cne_impl.hpp
```diff
@@ -0,0 +1,228 @@
...
+    }
+
+      Log::Info << "Generation number: " << gen << " best fitness = "
+          << fitnessValues.min() << std::endl;
+
+      // Create next generation of species.
+      Reproduce();
+
+      // Check for termination criteria.
+      if (tolerance != -1 && tolerance >= fitnessValues.min())
```
zoq: The first check is redundant, if someone likes to run the method for a specific number of iterations, he can set the ``tolerance`` accordingly.

src/mlpack/core/optimizers/cne/cne_impl.hpp
```diff
@@ -0,0 +1,228 @@
...
+      {
+          Log::Info << "Terminating. Given fitness criteria " << tolerance
+              << " > " << fitnessValues.min() << std::endl;
+          break;
+      }
+
+      // Check for termination criteria.
+      if (objectiveChange != -1 &&
```
zoq: I think we can also remove the first check here, see my comment on the ``tolerance`` parameter.

src/mlpack/core/optimizers/cne/cne_impl.hpp
kartik-nighania: oh i get the catch here, a person putting -1 will automatically won't have this evaluated. 
This means that i dont have to change the document too that mentions -1 to turn this off.

src/mlpack/core/optimizers/cne/cne_impl.hpp
kartik-nighania: done

src/mlpack/core/optimizers/cne/cne_impl.hpp
kartik-nighania: but i can edit it and write that -ve value means no evaluation.
thanks a lot

kartik-nighania:
thanks Marcus :)

doc/tutorials/cne/cne.txt
```diff
@@ -0,0 +1,356 @@
...
+@file cne.txt
+@author Kartik Nighania
+@brief Tutorial on how to use the CNE optimizer class.
+
+@page cnetutorial CNE Optimizer tutorial (mlpack_cne)
+
+@section intro_cnetut Introduction
+
+Conventional Neural Evolution (CNE) is a class of evolutionary algorithms
+focused on dealing with fixed topology networks. This class implements this
```
rcurtin: Which class is `this class`?  I'm not sure here, maybe you can reference the class the tutorial is about?

doc/tutorials/cne/cne.txt
```diff
@@ -0,0 +1,356 @@
+/*!
+
+@file cne.txt
+@author Kartik Nighania
+@brief Tutorial on how to use the CNE optimizer class.
+
+@page cnetutorial CNE Optimizer tutorial (mlpack_cne)
```
rcurtin: I don't think there is any `mlpack_cne` program, so maybe you should reference the CNE class here instead.

doc/tutorials/cne/cne.txt
```diff
@@ -0,0 +1,356 @@
...
+        const double finalValue,
+        const double fitnessHist);
+@endcode
+
+2) Making a neural network model and giving CNE as an optimizer to train the model.
+   For our test, we will be using a feed forward network or vanilla network from the
+   artificial neural network class.
+
+
+3) The trained model can then be used by calling the predict method.
```
rcurtin: You can use Doxygen code blocks here: `...can then be used by calling the @c Predict() method`.

doc/tutorials/cne/cne.txt
```diff
@@ -0,0 +1,356 @@
...
+
+@code
+void Predict(arma::mat predictors, arma::mat& results);
```
rcurtin: Should this be `const arma::mat& predictors`?

doc/tutorials/cne/cne.txt
```diff
@@ -0,0 +1,356 @@
...
+
+  for (size_t i = 0; i < predictionTemp.n_cols; ++i)
+  {
+    prediction(i) = arma::as_scalar(arma::find(
+        arma::max(predictionTemp.col(i)) == predictionTemp.col(i), 1)) + 1;
+  }
+
+  // Print the results 
+  for(size_t i = 0; i < 4; i++)
+    std::cout << prediction[i] << std::endl;
```
rcurtin: You can also do `std::cout << prediction` here and there is no need for a loop. :)

src/mlpack/core/optimizers/cne/cne.hpp
```diff
@@ -0,0 +1,234 @@
...
+ *
+ * Crossover takes two parents and generates two childs from them. Both the
+ * child have properties inherited by their parents. The parameters of the
+ * parents gets mixed using equal probability selection, creating two childs.
+ * This is just a mix of link weights or the parameters.
+ *
+ * In mutation parameters are updated by pertubating small noise.
+ * If \f$ \Lambda \f$ is the number of weights in the network.
+ *
+ * $$ \sum_{n=1}^{\Lambda}\omega_n = \omega_n + \rho $$
```
rcurtin: Does Doxygen pick up `$$`?  I think you have to do

```
\f[
\sum_{n=1}^{\Lambda}\omega_n = \omega_n + \rho,
\f]
```

src/mlpack/core/optimizers/cne/cne.hpp
```diff
@@ -0,0 +1,234 @@
...
+ * size_t NumFunctions();
+ * 
+ * NumFunctions() should return the number of functions or data points in
+ * the dataset.
+ *
+ * double Evaluate(const arma::mat& coordinates, const size_t i);
+ *
+ * The parameter i refers to which individual function is being evaluated.
+ * The function will evaluate the objective function on the data point i in
+ * the dataset.
```
rcurtin: This documentation is great.  Thanks so much for the clear details, I feel like I understand the class fully before I have even read the code. :)

src/mlpack/core/optimizers/cne/cne_impl.hpp
```diff
@@ -0,0 +1,227 @@
...
+
+  // Find the number of elite canditates from population.
+  numElite = floor(selectPercent * populationSize);
+
+  // Making sure we have even number of candidates to remove and create.
+  if ((populationSize - numElite) % 2 != 0) numElite--;
+
+  // Terminate if two parents can not be created.
+  if (numElite < 2)
+    throw std::logic_error("Increase selection percentage.");
```
rcurtin: I think this could be a better error message---the comment above might help it turn into something more useful.  With this error the user just gets `Increase selection percentage.` and they aren't sure why.  I'd suggest something more like:

```
throw std::logic_error("Selection percentage is too low---cannot select two parents.  Increase selection percentage.");
```

This idea could probably be applied to a few more error messages here.

src/mlpack/core/optimizers/cne/cne_impl.hpp
```diff
@@ -0,0 +1,227 @@
...
+      if (lastBestFitness - fitnessValues.min() < objectiveChange)
+      {
+          Log::Info << "Terminating. Fitness History change "
+              << (lastBestFitness - fitnessValues.min())
+              << " < " << objectiveChange << std::endl;
+          break;
+      }
+
+      // Store the best fitness of present generation.
+      lastBestFitness = fitnessValues.min();
```
rcurtin: This block (starting with the `Generation number:` output) seems to be indented two spaces too many.

doc/tutorials/cne/cne.txt
kartik-nighania: done

doc/tutorials/cne/cne.txt
kartik-nighania: done

doc/tutorials/cne/cne.txt
kartik-nighania: yes thanks
done

doc/tutorials/cne/cne.txt
kartik-nighania: done

src/mlpack/core/optimizers/cne/cne.hpp
kartik-nighania: yes it does @rcurtin 
but the one u mentioned is standard one used in MLPack, and i think i should also stick to that one for now and future.
thanks
changed. done

src/mlpack/core/optimizers/cne/cne.hpp
kartik-nighania: Cant be anymore happy to hear that. Thanks a lot.
 Also the reason I exchanged so many messages is that I am a bad coder 😆 . Constantly learning something from my mentor. 
Thanks again 😄

src/mlpack/core/optimizers/cne/cne_impl.hpp
kartik-nighania: Indeed.
thanks 
done

src/mlpack/core/optimizers/cne/cne_impl.hpp
kartik-nighania: yes.. i missed it .. thanks for the detailed review

doc/tutorials/cne/cne.txt
kartik-nighania: done

zoq:
This looks ready for me, if you don't mind I'd like to squash this into three commits. This would go as follows:


```shell
# Backup the code just in case.

git remote add upstream https://github.com/mlpack/mlpack.git
git checkout cne
git fetch upstream master
git reset --soft upstream/master

# Unstage test changes.
git reset HEAD src/mlpack/tests/cne_test.cpp
git reset HEAD src/mlpack/tests/CMakeLists.txt

# Unstage tutorial changes.
git reset HEAD doc/tutorials/README.md
git reset HEAD doc/tutorials/cne/cne.txt
git reset HEAD doc/tutorials/tutorials.txt

# Commit CNE optimizer code.
git commit -m "Add Conventional Neural Evolution (CNE) optimizer."

# Stage test changes.
git add src/mlpack/tests/cne_test.cpp
git add src/mlpack/tests/CMakeLists.txt

# Commit CNE tests.
git commit -m "Add Conventional Neural Evolution (CNE) optimizer test cases."

# Stage tutorial changes.
git add doc/tutorials/README.md
git add doc/tutorials/cne/cne.txt
git add doc/tutorials/tutorials.txt

# Commit CNE tutorials.
git commit -m "Add Conventional Neural Evolution (CNE) optimizer tutorial."

git push --force
```

kartik-nighania:
Done
Thanks @zoq for this guide.
I searched for this on the internet but its a bit messy over there. 
Can u tell me a good place to learn about stashing commits like u did?

kartik-nighania:
@zoq
i didnt understood history.md
also the following test failed -
nmf_test.cpp

zoq:
Looks good, thanks for taking care of the issues; I'll merge the code later today. About git: I really like https://try.github.io/ it's not going into all the details, I guess if you run into some problem you still have to search the internet for solutions.

zoq:
Thanks for the great contribution!

Merged:  https://github.com/mlpack/mlpack/commit/7af5fd18639740e2cf375333d17393dae39f045a https://github.com/mlpack/mlpack/commit/99ce3b99b1b0adbec1d6d98f75b981e31f6c2c4e https://github.com/mlpack/mlpack/commit/49ff33b042e638de67d0c028b15562fb55cf5ab1

kartik-nighania:
wow! 
my first open source merge..

thanks a lot for your help.
@zoq

zoq:
That reminds me of something, feel free to open a new PR to add yourself to the contributors list: ``COPYRIGHT.txt`` and ``src/mlpack/core.hpp``

kartik-nighania:
@zoq
I think Marcus its too early that i deserve to be added in contributors now..
Ill keep working for now and would add some time later.. 
I hope u wont mind.
:)

zoq:
> I think Marcus its too early that i deserve to be added in contributors now..

I don't think that's true, your contributions are much appreciated.

rcurtin:
Typically we have added people after they have contributed a single line of code to mlpack, so please feel free to add yourself. :)

PR #1093, opened 2017-08-14
Remove boost::bimap dependency; maintain two std::maps in DatasetMapper

rcurtin:
I've been using `-ftime-report` with gcc to try and figure out what is causing slowdowns.  I noticed that when we include `boost::bimap` and associated headers, the RAM usage and parse time of the compiler increases.

So I made this change, which replaces the `boost::bimap` in `DatasetMapper` with two `std::map` objects.

In addition, this solves the problem of non-unique mappings: previously, it was possible to have two strings map to the same mapping but there was no way to recover both of those strings.  Now, the reverse map holds a std::vector of strings, and `UnmapString()`'s signature is modified to take an index of the string to unmap to.  `NumUnmappings()` is also added to determine how many unmappings there are.

Before the change, time to compile with `make -j6`:

```
real    14m44.051s
user    80m22.688s
sys     2m44.908s
```

After the change:

```
real    13m58.226s
user    77m37.932s
sys     2m36.992s
```

There should be no noticeable runtime improvement, just compile-time.

### DIFF
--- a/src/mlpack/core/data/dataset_mapper_impl.hpp
+++ b/src/mlpack/core/data/dataset_mapper_impl.hpp
@@ -81,12 +81,20 @@ inline T DatasetMapper<PolicyType>::MapString(const std::string& string,
 
 // Return the string corresponding to a value in a given dimension.
 template<typename PolicyType>
+template<typename T>
 inline const std::string& DatasetMapper<PolicyType>::UnmapString(
-    const size_t value,
-    const size_t dimension)
+    const T value,
+    const size_t dimension,
+    const size_t unmappingIndex)
 {
+  // If the value is std::numeric_limits<T>::quiet_NaN(), we can't use it as a
+  // key---so we will use something else...
+  const T usedValue = (value != value) ?
+      std::nexttoward(std::numeric_limits<T>::max(), T(0)) :
+      value;
+
   // Throw an exception if the value doesn't exist.
-  if (maps[dimension].second.count(value) == 0)
+  if (maps[dimension].second.count(usedValue) == 0)
   {
     std::ostringstream oss;
     oss << "DatasetMapper<PolicyType>::UnmapString(): value '" << value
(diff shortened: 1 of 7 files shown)


src/mlpack/core/data/dataset_mapper_impl.hpp
```diff
@@ -81,20 +81,54 @@ inline T DatasetMapper<PolicyType>::MapString(const std::string& string,
...
 inline const std::string& DatasetMapper<PolicyType>::UnmapString(
-    const size_t value,
-    const size_t dimension)
+    const T value,
+    const size_t dimension,
+    const size_t unmappingIndex)
 {
+  // If the value is std::numeric_limits<T>::quiet_NaN(), we can't use it as a
+  // key---so we will use something else...
+  const T usedValue = (value != value) ?
```
zoq: Currently we don't use ``-ffast-math`` so we don't run into any problems here: 
https://stackoverflow.com/questions/22931147/stdisinf-does-not-work-with-ffast-math-how-to-check-for-infinity

but maybe it's worth to take a closer look into `-ffast-math`` and to think about a workaround? Also this issue was raised by the static code analysis check.

rcurtin: Hmmm, so this is an interesting situation.  I replaced `(value != value)` with `std::isnan(value)` just for clarity.  But it turns out the code is actually safe for `-ffast-math`, for complicated reasons:

When we use `-ffast-math`, then the compiler assumes that `std::isnan()` will always be false---so the conditional will always evaluate as false.  However, that conditional (`std::isnan(value)`) exists because a NaN can't be used as an index for a `std::unordered_map` or other container:

https://stackoverflow.com/questions/8096817/is-nan-a-valid-key-value-for-associative-containers

Therefore, when we encounter NaN as a key (which we do for `MissingPolicy`), we use this strange "super unlikely" value as a workaround.  But if we compile with `-ffast-math`, then whatever the byte representation of NaN is is assumed to just be a valid `double` number, so then that value can actually be used as a key for an `unordered_map` without a problem.  So the code actually works with `-ffast-math` (and I *think* that should work on all platforms).  Here's some example code:

```
#include <unordered_map>
#include <limits>
#include <vector>
#include <string>
#include <iostream>

using namespace std;

int main()
{
  unordered_map<double, std::vector<std::string>> m;

  double nan = std::numeric_limits<double>::quiet_NaN();
  m[nan] = std::vector<std::string>();

  if (m.count(nan) > 0)
    m[nan].push_back("hello");
  if (m.count(nan) > 0)
    m[nan].push_back("goodbye");

  for (size_t i = 0; i < m[nan].size(); ++i)
    std::cout << m[nan][i] << "\n";
}
```

If you compile with `-ffast-math`, the output is

```
hello
goodbye
```

because `quiet_NaN()` can be used as a valid `unordered_map` key in that case, but if you compile without `-ffast-math`, then there is no output because `quiet_NaN()` can't be used as a key.

This is probably a longer response than you expected, but I certainly learned a lot in the process!  It might be interesting to find out if the rest of mlpack can be compiled and work successfully with `-ffast-math`.

rcurtin:
I also made a few more changes to make `UnmapString()` const, but otherwise I think this is ready.

rcurtin:
This one has sat for a while and seems to be working, so I'll let it sit until tomorrow before a merge in case anyone has any other comments.

PR #1095, opened 2017-08-15
Add k-fold cross-validation

micyril:
See the previous discussion [here](https://github.com/micyril/mlpack/pull/1).


micyril:
Thank you for your reviews. I would like to mention one more thing that came to my mind. For the case k=2 we can actually avoid data copying if the user have passed moved data (r-value references). I don't think that k-fold cross-validation is really intended to be used with k=2, and this implementation will make constructors seem more "heavier" since they will be templated, but anyway we can do that. Let me know if you would like me to add the change.

rcurtin:
Personally I think that the set of people using `k = 2` is so small it's not worth optimizing for that use case.  So my opinion would be to leave the code as it is.  Of course, if you want to change the code to handle that case, I don't have any problem with that, but it is up to you.

micyril:
Let's then leave it as it is.

rcurtin:
Great work! Glad to merge it. :)

PR #1096, opened 2017-08-16
Update README to discuss Python bindings

rcurtin:
I realized that the README doesn't mention anything about the Python bindings, so this helps fix that.  New readme example:

https://github.com/rcurtin/mlpack/tree/python-readme

If anyone has any feedback or ideas for how it could be improved (or if this is sufficient), I'd appreciate it.

### DIFF
--- a/HISTORY.md
+++ b/HISTORY.md
@@ -2,7 +2,11 @@
 ###### ????-??-??
   * Speed and memory improvements for DBSCAN.  --single_mode can now be used for
     situations where previously RAM usage was too high.
+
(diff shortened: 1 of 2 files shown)


zoq:
Have you thought about using syntax highlighting?

```python
>>> from mlpack import knn
```

I guess in this case it's not that helpful and it's just a matter of taste.

rcurtin:
Ah, I didn't know about that.  Fixed now.

zoq:
Looks and sounds good for me.

PR #1098, opened 2017-08-19
ResizeLayer implementation

kris-singh:


src/mlpack/methods/ann/image_functions/CMakeLists.txt
```diff
@@ -0,0 +1,14 @@
+# Define the files we need to compile
+# Anything not in this list will not be compiled into mlpack.
+set(SOURCES
+  bilinear_impl.hpp
```
lozhnikov: The cmake config is incorrect. The file name is `bilinear_function.hpp`.

lozhnikov: Moreover, you didn't add `image_functions` dir into `src/mlpack/methods/ann/CMakeLists.txt`.

src/mlpack/methods/ann/image_functions/bilinear_function.hpp
```diff
@@ -0,0 +1,143 @@
...
+ * The bilinear interpolation function
+ *
+ */
+class BiLinearFunction
+{
+ public:
+  BiLinearFunction(
+      const size_t inRowSize,
+      const size_t inColSize,
+      const size_t scale):
```
lozhnikov: I think it isn't difficult to support real (`double`) scales. Moreover, it is a good idea to support changing of the aspect ratio.
So, I suggest 4 input parameters: `inRowSize`, `inColSize`, `outRowSize`, `outColSize`.

src/mlpack/methods/ann/image_functions/bilinear_function.hpp
```diff
@@ -0,0 +1,143 @@
...
+      inRowSize(inRowSize),
+      inColSize(inColSize),
+      outRowSize(2 * inRowSize),
```
lozhnikov: I think that's a typo. Do you mean `scale * inRowSize`?

src/mlpack/methods/ann/image_functions/bilinear_function.hpp
```diff
@@ -0,0 +1,143 @@
...
+   * UpSample the given Image
+   *
+   * @param input Input image.
+   * @param output The resulting interpolated output image.
+   */
+  template<typename eT>
+  void UpSample(const arma::Mat<eT>& input, arma::Mat<eT>& output)
+  {
+    // Get dimensions.
+    arma::Mat<eT> tempInput = input;
```
lozhnikov: I guess it isn't difficult to avoid copy-constructors. As for me it is better to translate the pair `(row, col)` to the order number each time.

src/mlpack/methods/ann/image_functions/bilinear_function.hpp
```diff
@@ -0,0 +1,143 @@
...
+    if (input.n_rows != inRowSize || input.n_cols != inColSize)
+      tempInput.reshape(inRowSize, inColSize);
+
+    output.set_size(outRowSize, outColSize);
+
+    if (output.n_rows > tempInput.n_rows)
+      scaleRow = inRowSize / outRowSize;
```
lozhnikov: I think that's incorrect since you use integer-valued math here.

src/mlpack/methods/ann/image_functions/bilinear_function.hpp
```diff
@@ -0,0 +1,143 @@
...
+      scaleCol = inColSize / outColSize;
+    else
+      scaleCol = (inColSize - 1) / outColSize;
+
+    for (size_t i = 0; i < outRowSize; i++)
+      for (size_t j = 0; j < outColSize; j++)
+      {
+        r_origin = std::floor(i * scaleRow);
+        c_origin = std::floor(j * scaleCol);
+        GetCoff(i - r_origin, j - c_origin);
```
lozhnikov: Looks like a typo. Maybe `Coeff`? I don't like the idea to change shared variables inside this function. In my opinion it is better to place the code of `GetCoff` here.

src/mlpack/methods/ann/image_functions/bilinear_function.hpp
```diff
@@ -0,0 +1,143 @@
...
+        output = tempInput(r_origin, c_origin) * coff1 +
```
lozhnikov: The left side is a matrix. The right side is a scalar.

src/mlpack/methods/ann/image_functions/bilinear_function.hpp
```diff
@@ -0,0 +1,143 @@
...
+                 tempInput(r_origin + 1, c_origin) * coff2 +
+                 tempInput(r_origin, c_origin + 1) * coff3+
+                 tempInput(r_origin + 1, c_origin + 1) * coff4;
+      }
+  }
+
+  /**
+   * Computes the first derivative of the logistic function.
```
lozhnikov: The comment is incorrect.

src/mlpack/methods/ann/image_functions/bilinear_function.hpp
lozhnikov: The formula looks like incorrect. I think you should multiply `i` by `scaleRow` and `j` by `scaleCol`.

src/mlpack/methods/ann/image_functions/bilinear_function.hpp
```diff
@@ -0,0 +1,143 @@
...
+   *
+   * @param x Input data.
+   * @return f'(x)
+   */
+  void GetCoff(const size_t row, const size_t col)
```
lozhnikov: `row` and `col` aren't integers here.

src/mlpack/methods/ann/image_functions/bilinear_function.hpp
```diff
@@ -0,0 +1,143 @@
...
+  const size_t inColSize;
+  const size_t outRowSize;
+  const size_t outColSize;
+  const size_t scale;
+  double scaleRow;
+  double scaleCol;
+  size_t r_origin;
+  size_t c_origin;
+
+  size_t coff1;
```
lozhnikov: The coefficients are real.

src/mlpack/methods/ann/image_functions/bilinear_function.hpp
kris-singh: I wanted to use the case of 1d array as input to the bilinear function. If I use 1d array I would have to resize to the `inRowSize, inColSize` which will not be possible since I am using const for input. That why I used a temp variable here.

lozhnikov: No, you needn't resize the input. Since armadillo stores matrices in column-major order you could write `input(j * input.n_rows + i)` instead of `input(i, j)`.

kris-singh:
Write now i am getting this as input. For 4x
screen shot 2017-08-20 at 6 02 43 pm

lozhnikov:
hmm... Why the second image has grid-like artifacts?

kris-singh:
Since i am putting it on the same scale it seems there are grid-like artifacts. I will test this with a rgb image and check.

kris-singh:
This is the full sized image with the correct formula. The unzoomed image does not have the artifact though. 
screen shot 2017-08-20 at 8 31 51 pm

src/mlpack/methods/ann/image_functions/bilinear_function.hpp
```diff
@@ -44,43 +44,41 @@ class BiLinearFunction
...
         coff1 = (1 - deltaC) * (1 - deltaC);
         coff2 = deltaR * (1 - deltaC);
         coff3 = (1 - deltaR) * deltaC;
         coff4 = deltaR * deltaC;
-        output(i, j) = tempInput(r_origin, c_origin) * coff1 +
-                 tempInput(r_origin + 1, c_origin) * coff2 +
-                 tempInput(r_origin, c_origin + 1) * coff3+
-                 tempInput(r_origin + 1, c_origin + 1) * coff4;
+        eT temp = input(c_origin * input.n_rows + r_origin );
+        output(i, j) =  temp * coff1 + temp * coff2 + temp * coff3 +
```
lozhnikov: Why do you multiply each coefficient by the same value?

zoq:
For me it looks like you don't interpolate between the values, an easy test would be to pass an artificial image e.g. a 20x20 matrix filled with a constant value. My guess is that end up with gaps.

kris-singh:
It works now....
screen shot 2017-08-20 at 10 41 48 pm

src/mlpack/methods/ann/image_functions/bilinear_function.hpp
```diff
@@ -0,0 +1,141 @@
...
+namespace ann /** Artificial Neural Network. */ {
+
+/**
+ * The bilinear interpolation function
+ *
+ */
+class BiLinearFunction
+{
+ public:
+  BiLinearFunction(
```
lozhnikov: Could you describe the arguments?

src/mlpack/methods/ann/image_functions/bilinear_function.hpp
```diff
@@ -0,0 +1,141 @@
...
+    for (size_t i = 0; i < output.n_rows; i++)
+    {
+      for (size_t j = 0; j < output.n_cols; j++)
+      {
+        r_origin = std::floor(i * scaleRow);
+        c_origin = std::floor(j * scaleCol);
+
+        if (r_origin < 0)
+          r_origin = 0;
+        if (c_origin < 0)
```
lozhnikov: These checks aren't needed.

src/mlpack/methods/ann/image_functions/bilinear_function.hpp
```diff
@@ -0,0 +1,141 @@
...
+          c_origin = 0;
+
+        if (r_origin > input.n_rows - 2)
+          r_origin = input.n_rows - 2;
+        if (c_origin > input.n_cols - 2)
+          c_origin = input.n_cols - 2;
+
+        double deltaR = i * scaleRow - r_origin;
+        double deltaC = j * scaleCol - c_origin;
+        coff1 = (1 - deltaR) * (1 - deltaC);
```
lozhnikov: Pedantic issue. I suggest to write `coeff`.

src/mlpack/methods/ann/image_functions/bilinear_function.hpp
```diff
@@ -0,0 +1,141 @@
...
+      output = arma::resize(output, outRowSize, outColSize);
+
+    scaleRow = (double)input.n_rows / (double) output.n_rows;
+    scaleCol = (double)input.n_cols / (double) output.n_cols;
+
+    for (size_t i = 0; i < output.n_rows; i++)
+    {
+      for (size_t j = 0; j < output.n_cols; j++)
+      {
+        r_origin = std::floor(i * scaleRow);
```
lozhnikov: Could you fix the variable name according to the style guide?

src/mlpack/methods/ann/image_functions/bilinear_function.hpp
```diff
@@ -0,0 +1,141 @@
...
+        output(i, j) =  input(c_origin * input.n_rows + r_origin) * coff1 +
+                        input(c_origin * input.n_rows + r_origin + 1) * coff2 +
+                        input((c_origin + 1) * input.n_rows + r_origin) * coff3 +
+                        input((c_origin + 1) * input.n_rows + r_origin+1) * coff4;
+      }
+    }
+  }
+
+  template<typename eT>
+  void DownSample(const arma::Mat<eT>& input, arma::Mat<eT>& output)
```
lozhnikov: Could you describe the arguments?

kris-singh: While upsampling the input represents the image and output represent the zoomed image.
Here it is opposite input represents the image and output un-zoomed image.(smaller image).
This function is required for doing backprop through the layer if we get the image gradient from the layer l+1 we have to downsample it to apply to l-1.

kris-singh:
@lozhnikov  I think you can have a look now. The random forest test is failing.

src/mlpack/tests/ann_layer_test.cpp
```diff
@@ -1004,5 +1004,26 @@ BOOST_AUTO_TEST_CASE(SimpleMeanSquaredErrorLayerTest)
   BOOST_REQUIRE_EQUAL(output.n_elem, 1);
 }
 
+/*
+ * Simple test for the Resize layer
+ */
+
+BOOST_AUTO_TEST_CASE(SimpleResizeLayerTest)
```
lozhnikov: I suggest a more meaningful test:
1. Declare a n*m matrix.
2. Fill the matrix using a linear function (e.g. `x+y` or `ax+by`)
3. Scale the matrix.
4. Check if the scaled matrix represents the scaled linear function.
5. Transform the scaled matrix backward.
6. Check if this matrix is close to the original one.

src/mlpack/methods/ann/layer/resize_impl.hpp
```diff
@@ -0,0 +1,70 @@
...
+
+template<class InterpolationType,
+         typename InputDataType,
+         typename OutputDataType
+         >
+template<typename Archive>
+void Resize<InterpolationType, InputDataType, OutputDataType>::Serialize(
+    Archive& ar, const unsigned int /* version */)
+{
+  ar & data::CreateNVP(policy, "policy");
```
lozhnikov: In that case you should implement `BiLinearFunction::Serialize()`.

src/mlpack/methods/ann/image_functions/bilinear_function.hpp
```diff
@@ -0,0 +1,170 @@
...
+  //! Locally stored row size of the input.
+  const size_t inRowSize;
+  //! Locally stored column size of the input.
+  const size_t inColSize;
+  //! Locally stored row size of the output.
+  const size_t outRowSize;
+  //! Locally stored column size of the input.
+  const size_t outColSize;
+
+  //! Locally stored scaling factor along row.
```
lozhnikov: I guess all variable below should be defined locally inside the functions. How do you think?

src/mlpack/methods/ann/image_functions/bilinear_function.hpp
```diff
@@ -0,0 +1,170 @@
...
+   *
+   * @param input The input matrix
+   * @param output The resulting down-sampled output image.
+   */
+  template<typename eT>
+  void DownSample(const arma::Mat<eT>& input, arma::Mat<eT>& output)
+  {
+
+    if (output.is_empty() && (output.n_cols != inColSize || output.n_rows != inRowSize))
+      output = arma::resize(output, inRowSize, inColSize);
```
lozhnikov: I am not sure that the layer should reshape the output. Maybe it is better to represent the output as a column `(inRowSize * inColSize, 1)`. In that case the layer should work with the linear layer.

src/mlpack/methods/ann/image_functions/bilinear_function.hpp
kris-singh: Well, the variables are used by both the `Upsample` and the `DownSample` function. So rather than creating stack variables, it is better to just create heap variables. 
Could you tell the benefits of using stack variable here?

lozhnikov: As for me, it improves readability and points out that you needn't these variables outside of the function.

kris-singh:
This un zoomed image....
screen shot 2017-08-24 at 8 53 52 pm

src/mlpack/methods/ann/image_functions/bilinear_function.hpp
```diff
@@ -0,0 +1,187 @@
...
+        
+
+        double deltaR = i * scaleRow - rOrigin;
+        double deltaC = j * scaleCol - cOrigin;
+        coeff1 = (1 - deltaR) * (1 - deltaC);
+        coeff2 = deltaR * (1 - deltaC);
+        coeff3 = (1 - deltaR) * deltaC;
+        coeff4 = deltaR * deltaC;
+
+        output(i * outRowSize + j) =
```
lozhnikov: Probably, it is better to store matrices in column-major order
since Armadillo does the same i.e.
`output(i, j) = output(j * outColSize + i)`
and
`input(a, b) = input(b * inColSize + a)`

src/mlpack/methods/ann/image_functions/bilinear_function.hpp
```diff
@@ -0,0 +1,187 @@
...
+    assert(output.n_rows == inRowSize * inColSize);
+    assert(output.n_cols == 1);
+
+    if (input.n_elem == output.n_elem)
+    {
+      output = input;
+    }
+    else
+    {
+      scaleRow = (double)(outRowSize - 1) / inRowSize;
```
lozhnikov: I think the correct formula is
`scaleRow = ((double) outRowSize) / inRowSize;`
(without `-1`)

kris-singh: Look here https://ia802707.us.archive.org/23/items/Lectures_on_Image_Processing/EECE_4353_15_Resampling.pdf slide 60.
Here the Output is the zoomed image ie image of size `outRowSize, outColSize` hence these become R and C according to the slide. and `inRowSize` becomes R'

src/mlpack/methods/ann/image_functions/bilinear_function.hpp
kris-singh: So you are suggesting i change this `output.set_size(outRowSize * outColSize, 1);` to 
`output.set_size(1, outRowSize * outColSize)`;. Right.

src/mlpack/methods/ann/image_functions/bilinear_function.hpp
lozhnikov: Looks like the slide is incorrect.

src/mlpack/methods/ann/layer/resize_impl.hpp
```diff
@@ -0,0 +1,69 @@
...
+
+template<class InterpolationType,
+         typename InputDataType,
+         typename OutputDataType
+         >
+template<typename eT>
+void Resize<InterpolationType, InputDataType, OutputDataType>::Backward(
+    const arma::Mat<eT>&& /* input */, arma::Mat<eT>&& gy, arma::Mat<eT>&& g)
+{
+  policy.DownSample(gy, g);
```
lozhnikov: Hmmm... I think it is correct only in case of bilinear interpolation. For example quadratic interpolation would require the `input` matrix in order to perform the backward pass (If I am not mistaken). So, I guess we should add that parameter to the policy API.
Moreover, I think it isn't too evident that the backward pass is just the inverse interpolation of the backpropogated error in case of bilinear interpolation. I guess we should add a comment in order to describe that.

src/mlpack/tests/ann_layer_test.cpp
```diff
@@ -1004,5 +1004,37 @@ BOOST_AUTO_TEST_CASE(SimpleMeanSquaredErrorLayerTest)
...
+ * Simple test for the Resize layer
+ */
+BOOST_AUTO_TEST_CASE(SimpleResizeLayerTest)
+{
+  arma::mat input, output, unzoomedOutput, expectedOutput;
+  size_t inRowSize = 10;
+  size_t inColSize = 10;
+  size_t outRowSize = 20;
+  size_t outColSize = 20;
+  input.set_size(inRowSize * inColSize, 1);
```
lozhnikov: I think we should add the `depth` parameter to the test. We didn't test the third dimension yet.

src/mlpack/methods/ann/layer/resize_impl.hpp
kris-singh: Fixed the api. But not sure about the comment.

kris-singh:
I fixed all the comments. Also, i added the test for the 3d input but removed the 2d input test seems redundant to me.

lozhnikov:
> Fixed the api. But not sure about the comment.

I mean it is reasonable to add a comment that describes why the backward pass is just the inverse transform applied to the backpropogated error (in case of bilinear interpolation).

kris-singh:
Could we restart travis on this one.

lozhnikov:
> Moreover, I think it isn't too evident that the backward pass is just the inverse interpolation of the backpropogated error in case of bilinear interpolation. I guess we should add a comment in order to describe that.
> I mean it is reasonable to add a comment that describes why the backward pass is just the inverse transform applied to the backpropogated error (in case of bilinear interpolation).

```
 /**
    * DownSample the given input.
    * This is equivalent to the Backward the pass for the layer.
    * Since, the layer dose not have any learnable parameters
    * we just have downsample the gradient to make it's size
    * compatible with the input size.
    *		
```

The backward pass should calculate the product of the derivative and the backpropogated error. I suggested to leave a comment that explains why the inverse transform applied to the backpropogated error is the same (it seems that's true in case of bilinear interpolation, but in general, the statement is incorrect for any layer without `learnable parameters`, e.g. that's incorrect for cubic splines).

It is possible to implement the backward transform in terms of the ActivationFunction API i.e.
`g = derivative * error;`
But the derivative isn't a diagonal matrix therefore the approach is cost-inefficient. However, we could use sparse matrices instead. @zoq How do you think, is it reasonable?

kris-singh:
Well isn' this what we are exactly doing derivate in our case is 1 error is the gy(Which we have to downsample). g = 1 * gy.
I don't seem to get the problem.

zoq:
I'm not sure maybe I missed something , but instead of implementing the Interpolation as an Activation function we could just implement it as a Layer itself and remove the extra ``Resize`` layer. It looks like the ``Resize`` layer is just an extra wrapper around the interpolation function. I guess you could add an alias if you like to use Resize over BilinearInterpolation. Let me know what you think.

kris-singh:
@zoq  Our idea here was that there could be a lot of interpolations function in the future hence we implemented in the manner of the base_layer.hpp or activation function file. This way it is more extendable.

zoq:
Not sure it's more extendable since all the Resize layer does is to wrap two functions. I think implementing the current function as an independent layer has another benefit since you could directly pass additional parameter instead of passing a policy instantiation. Like currently you would have to write

```
InterpolationFunction inter(/* function parameter */);
model.Add<Resize<> >(inter);
```

if we implement it as an indepent layer we could just write:

```
model.Add<InterpolationFunction<> >(/* function parameter */);
```

Anyway, this is just hypothetical, since the BilinearFunction does not provide an interface for additional parameter.

kris-singh:
@zoq I think what i am trying to say if you look at the linear interpolation, bilinear interpolation, bi cubic interpolation all there are represented as policies of the resize layer otherwise you would have to implement all these as a different layer. It makes sense to define them as policies since they are doing the same task.
Are you following my point?

zoq:
I get your point, we did something similar with the optimizer classes, where we use SGD as a base class. But instead of providing a base class that takes an update policy we opted for a specilized class for each new Optimizer like Adam to improve the overall usability. So that you can use ``Adam(/* optimizer setting */)`` instead of 

```
Adam policy(/* optimizer settings */)
SGD<Adam>(policy);
```
I think using Adam directly, instead of creating an instance first is easier. Anyway, as you pointed out we could implement the interpolation function as a policy, but maybe providing an independent layer is just easier to use? Unless, each interpolation function takes the same parameter, in this case, we should pass the parameter instead of an instance. Let me know what you think.

kris-singh:
I think right now we can implement it as policy as you are suggesting by passing the same set of parameters. I would go ahead with that implementation i think if in future we come across an interpolation function we could always add the parameters that are being passed. I would go ahead and implement this. @lozhnikov do you have any comments here?

lozhnikov:
I agree with Marcus. Since `ResizeLayer` is just a wrapper I think it is reasonable to merge `BiLinearInterpolation` and `ResizeLayer`. I guess the only question we should answer is the following:
Are we going to use `BiLinearInterpolation` without `ResizeLayer`?
I think not since they do the same. Suppose we have a few interpolation functions. Anyway we have to add each instance to `LayerTypes`.

zoq:
Forwarding the interpolation parameter sounds fine for me, as @lozhnikov pointed if we only use the interpolation function for the resize layer the difference is negligible.

ShikharJ:
Thanks @kris-singh for your work here. This has been completed in #1238.
 @zoq This can be closed now.

zoq:
Thanks @kris-singh and @ShikharJ for the great work.

PR #1101, opened 2017-08-25
Add a hyper-parameter tuning module

micyril:
See the previous discussion [here](https://github.com/micyril/mlpack/pull/2).

### DIFF
--- a/src/mlpack/core/hpt/hpt_impl.hpp
+++ b/src/mlpack/core/hpt/hpt_impl.hpp
@@ -148,6 +148,7 @@ void HyperParameterTuner<MLAlgorithm,
   bestObjective = Metric::NeedsMinimization?
       optimizer.Optimize(cvFunction, bestParams) :
       -optimizer.Optimize(cvFunction, bestParams);
+  bestModel = std::move(cvFunction.BestModel());
 }
 
 template<typename MLAlgorithm,
--- a/src/mlpack/core/hpt/hpt_impl.hpp
+++ b/src/mlpack/core/hpt/hpt_impl.hpp
@@ -177,6 +177,14 @@ void HyperParameterTuner<MLAlgorithm,
   for (auto value : std::get<I>(args))
     datasetInfo.MapString<size_t>(value, dimension);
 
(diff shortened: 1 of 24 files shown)


micyril:
@rcurtin: when I was rebasing the branch to the master branch, there were a lot of conflicts with the commit ["Refactor DatasetMapper to map any input type"](https://github.com/mlpack/mlpack/pull/1101/commits/9391667d4f1f9ec45b88f92dbf5c98541a720606). So, please review this particular commit.

Also, I have noticed that the style checks haven't been run for my last submission. Is it expected behavior?

zoq:
@mlpack-jenkins style-check this

zoq:
Not sure, why the last commit did not trigger the style check job. But we can always ask @mlpack-jenkins to run the check.

src/mlpack/core/hpt/hpt_impl.hpp
```diff
@@ -0,0 +1,234 @@
...
+         typename WeightsType>
+template<typename... CVArgs>
+HyperParameterTuner<MLAlgorithm,
+                    Metric,
+                    CV,
+                    Optimizer,
+                    MatType,
+                    PredictionsType,
+                    WeightsType>::HyperParameterTuner(const CVArgs&... args) :
+    cv(args...), relativeDelta(0.01), minDelta(1e-10) {}
```
zoq: Should we provide a default constructor and set the default vales there? That way doxygen can pick up the default values and we can also add a note that the values are not necessarily good for the given problem, and suggest to adjust each to the task at hand. Let me know what you think.

micyril: It is problematic since we need to accept variadic template parameters in the constructor. I thought about mentioning the default values in the documentation for the getters and setters, but I hadn't seen anything like that in the mlpack codebase. Let me know if you think it's ok to do that.

zoq: Right I missed that, I think you could add a note to the class description itself, but adding it to the getter/setter documentation is also fine; it's up to you.

rcurtin:
Thanks for the great contribution. :)

PR #1103, opened 2017-08-25
Serialize the network for ANNs

rcurtin:
This PR refactors the `FFN` and `RNN` classes so that the networks themselves are serialized.  To make this work, I had to give up on the `SerializationShim` and `data::CreateNVP()`.  Honestly, I think it is a good thing to get rid of them... I had fun writing them, but I am not sure the cost in terms of code is worth it just so that we can have a `Serialize()` function instead of a `serialize()` function and follow the style guide.  So we can modify the style guide and make an exception...

(A comment from a friend when I was discussing this... "I'm glad you've finally come to your senses and realized that the shim is stupid" :))

I also added some quick tests for serialization, but they are not comprehensive.  I think more ideally, we should make a simple network that tests each possible layer type and makes sure that it serializes correctly.  There is some amount of trickiness for serializing layers, so I think it would be a worthwhile test.

So I think this PR is fine to merge as-is, but we should open issues if we merge this:

 * Remove the serialization shim: convert `Serialize()` to `serialize()`, and `CreateNVP()` to `BOOST_SERIALIZATION_NVP()` where possible (and `make_nvp()` in other places).
 * Write serialization tests that cover each layer type.

What's nice too is that this serialization strategy allows us to serialize any arbitrary network structure and then load it in another program that knows nothing about the network structure.  I am really happy about that.

### DIFF
--- a/src/mlpack/tests/recurrent_network_test.cpp
+++ b/src/mlpack/tests/recurrent_network_test.cpp
@@ -551,7 +551,6 @@ BOOST_AUTO_TEST_CASE(LSTMRecursiveReberGrammarTest)
  */
 BOOST_AUTO_TEST_CASE(GRUReberGrammarTest)
 {
-  math::RandomSeed(std::time(NULL));
   ReberGrammarTestNetwork<GRU<>>(4, false);
 }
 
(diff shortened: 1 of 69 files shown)


src/mlpack/tests/recurrent_network_test.cpp
```diff
@@ -550,6 +551,7 @@ BOOST_AUTO_TEST_CASE(LSTMRecursiveReberGrammarTest)
  */
 BOOST_AUTO_TEST_CASE(GRUReberGrammarTest)
 {
+  math::RandomSeed(std::time(NULL));
```
zoq: Looks like we can remove this line.

PR #1104, opened 2017-08-26
Cross entropy with logit

kris-singh:


kris-singh:
Not sure about the error from travis why this happened.

kris-singh:
@lozhnikov  This works fine should we merge this. Or just discard this.

src/mlpack/methods/ann/layer/cross_entropy_error_with_logits_impl.hpp
```diff
@@ -0,0 +1,79 @@
...
+  // Nothing to do here.
+}
+
+template<typename InputDataType, typename OutputDataType>
+template<typename eT>
+double CrossEntropyErrorLogits<InputDataType, OutputDataType>::Forward(
+    const arma::Mat<eT>&& input, const arma::Mat<eT>&& target)
+{
+  eT loss = 0;
+  for (size_t i = 0; i < input.n_elem; i++)
```
lozhnikov: I suggest to add braces in order to improve readability.

src/mlpack/methods/ann/layer/cross_entropy_error_with_logits.hpp
```diff
@@ -0,0 +1,104 @@
...
+ * @tparam InputDataType Type of the input data (arma::colvec, arma::mat,
+ *         arma::sp_mat or arma::cube).
+ * @tparam OutputDataType Type of the output data (arma::colvec, arma::mat,
+ *         arma::sp_mat or arma::cube).
+ */
+template <
+    typename InputDataType = arma::mat,
+    typename OutputDataType = arma::mat
+>
+class CrossEntropyErrorLogits
```
lozhnikov: I think the name `SigmoidCrossEntropyError` suits better than `CrossEntropyErrorLogits`.

src/mlpack/methods/ann/layer/cross_entropy_error_with_logits.hpp
```diff
@@ -0,0 +1,104 @@
...
+{
+ public:
+  /**
+   * Create the CrossEntropyErrorLogits object.
+   * 
+   * @param eps The minimum value used for computing logarithms
```
lozhnikov: There are no parameters.

src/mlpack/methods/ann/layer/cross_entropy_error_impl.hpp
```diff
@@ -9,8 +9,8 @@
  * 3-clause BSD license along with mlpack.  If not, see
  * http://www.opensource.org/licenses/BSD-3-Clause for more information.
  */
-#ifndef MLPACK_METHODS_ANN_LAYER_CROSS_ENTROPY_ERROR_IMPL_HPP
-#define MLPACK_METHODS_ANN_LAYER_CROSS_ENTROPY_ERROR_IMPL_HPP
+#ifndef MLPACK_METHODS_ANN_LAYER_CROSS_ENTROPY_ERROR_LOGIT_IMPL_HPP
```
lozhnikov: hmm.. Why did you change `cross_entropy_error_impl.hpp`?

src/mlpack/methods/ann/layer/cross_entropy_error_with_logits_impl.hpp
```diff
@@ -0,0 +1,79 @@
...
+      loss += input(i) - input(i) * target(i) +
+              SoftplusFunction::Fn(-std::abs(input(i)));
+    }
+    else
+    {
+      loss += input(i) * target(i) +
+              SoftplusFunction::Fn(-std::abs(input(i)));;
+    }
+
+  return loss / input.n_elem;
```
lozhnikov: `CrossEntropyError` doesn't divide the result by the number of dimensions. I suggest to do the same.

src/mlpack/methods/ann/layer/cross_entropy_error_with_logits_impl.hpp
```diff
@@ -0,0 +1,79 @@
...
+  eT loss = 0;
+  for (size_t i = 0; i < input.n_elem; i++)
+    if (input(i) > 0)
+    {
+      loss += input(i) - input(i) * target(i) +
+              SoftplusFunction::Fn(-std::abs(input(i)));
+    }
+    else
+    {
+      loss += input(i) * target(i) +
```
lozhnikov: The sign is incorrect.

src/mlpack/methods/ann/layer/cross_entropy_error_with_logits_impl.hpp
```diff
@@ -0,0 +1,79 @@
...
+              SoftplusFunction::Fn(-std::abs(input(i)));;
+    }
+
```
lozhnikov: Actually, you could simplify that. For example
```
arma::uvec positive = arma::find(input > 0);

loss = arma::accu(input(positive)) - arma::accu(input % target) + 
  arma::accu(SoftplusFunction::Fn(-arma::abs(input)));
```
However, it requires an extra memory allocation.

src/mlpack/methods/ann/layer/cross_entropy_error_with_logits_impl.hpp
```diff
@@ -0,0 +1,79 @@
...
+}
+
+template<typename InputDataType, typename OutputDataType>
+template<typename eT>
+void CrossEntropyErrorLogits<InputDataType, OutputDataType>::Backward(
+    const arma::Mat<eT>&& input,
+    const arma::Mat<eT>&& target,
+    arma::Mat<eT>&& output)
+{
+  output = input;
```
lozhnikov: I guess this line isn't needed.

src/mlpack/methods/ann/layer/cross_entropy_error_with_logits_impl.hpp
```diff
@@ -0,0 +1,79 @@
...
+  for (size_t i = 0; i < input.n_elem; i++)
```
lozhnikov: I think it is possible to vectorise this loop too.

kris-singh:
The tests pass locally. Checked thrice.

lozhnikov:
```
  8/114 Test   #9: ANNLayerTest .....................***Failed    2.91 sec
Running 31 test cases...
unknown location(0): fatal error in "SimpleSigmoidCrossEntropyLayerTest": memory access violation at address: 0x00000000: no mapping at fault address
/home/travis/build/mlpack/mlpack/src/mlpack/tests/ann_layer_test.cpp(1044): last checkpoint
```
> The tests pass locally. Checked thrice.

I think valgrind could help.

src/mlpack/methods/ann/layer/sigmoid_cross_entropy_error_impl.hpp
```diff
@@ -0,0 +1,75 @@
...
+    arma::Mat<eT>&& output)
+{
+  arma::mat temp;
+
+  arma::uvec positive = arma::find(input > 0);
+  arma::uvec negative = arma::find(input < 0 || input == 0);
+
+  SoftplusFunction::Deriv(static_cast<arma::mat>(-arma::abs(input(positive))),
+      temp);
+  output(positive) = 1 - target(positive) - temp;
```
zoq: Looks like we should set the size of output first, if we like to index it here.

kris-singh:
Finally, all tests pass. Let me know if anything else needs to be done.

kris-singh:
Should we merge this if there are no more comments.

src/mlpack/methods/ann/layer/sigmoid_cross_entropy_error_impl.hpp
```diff
@@ -0,0 +1,76 @@
...
+{
+  // Nothing to do here.
+}
+
+template<typename InputDataType, typename OutputDataType>
+template<typename eT>
+double SigmoidCrossEntropyError<InputDataType, OutputDataType>::Forward(
+    const arma::Mat<eT>&& input, const arma::Mat<eT>&& target)
+{
+  arma::uvec positive = arma::find(input > 0);
```
zoq: Since this is used only once should we inline the expression?

src/mlpack/methods/ann/layer/sigmoid_cross_entropy_error_impl.hpp
```diff
@@ -0,0 +1,76 @@
...
+
+template<typename InputDataType, typename OutputDataType>
+template<typename eT>
+void SigmoidCrossEntropyError<InputDataType, OutputDataType>::Backward(
+    const arma::Mat<eT>&& input,
+    const arma::Mat<eT>&& target,
+    arma::Mat<eT>&& output)
+{
+  arma::mat temp;
+  output = input;
```
zoq: Do you think we should use ``set_size`` here instead of a copy?

src/mlpack/tests/ann_layer_test.cpp
```diff
@@ -926,6 +926,42 @@ BOOST_AUTO_TEST_CASE(SimpleLogSoftmaxLayerTest)
       arma::mat("1.6487; 0.6487") - delta)), 1e-3);
 }
 
+BOOST_AUTO_TEST_CASE(SimpleSigmoidCrossEntropyLayerTest)
```
zoq: Would be great if you could add a comment that describes the test case.

src/mlpack/tests/ann_layer_test.cpp
```diff
@@ -926,6 +926,42 @@ BOOST_AUTO_TEST_CASE(SimpleLogSoftmaxLayerTest)
...
+{
+  arma::mat input1, input2, output, target1, target2, expectedOutput;
+  SigmoidCrossEntropyError<> module;
+
+  // Test the Forward function on a user generator input and compare it against
+  // the manually calculated result.
+  input1 = arma::mat("0.5 0.5 0.5 0.5 0.5 0.5 0.5 0.5");
+  target1 = arma::zeros(1, 8);
+  double error1 = module.Forward(std::move(input1), std::move(target1));
+  // value computed using tf
```
zoq: Pedantic style issue; start with capital letter and end with ``.``.

src/mlpack/methods/ann/layer/sigmoid_cross_entropy_error.hpp
```diff
@@ -0,0 +1,101 @@
...
+#include <mlpack/prereqs.hpp>
+
+namespace mlpack {
+namespace ann /** Artificial Neural Network. */ {
+
+/**
+ * The SigmoidCrossEntropyError performance function measures the network's
+ * performance according to the cross-entropy function.
+ * between the input and target distributions.
+ * For more detail look here goo.gl/tRjS6j
```
zoq: Instead of linking to tf I think it would be a good idea to provide some more information here, e.g. provide the formula and mention that the function is often more numerically stable.

kris-singh:
@zoq  All your comments were fixed.

src/mlpack/methods/ann/layer/sigmoid_cross_entropy_error.hpp
```diff
@@ -0,0 +1,107 @@
...
+
+/**
+ * The SigmoidCrossEntropyError performance function measures the network's
+ * performance according to the cross-entropy function.
+ * between the input and target distributions.
+ * This function calculates the cross entropy
+ * given the real values instead of providing the sigmoid activations.
+ * The functions is much more numerically stable as can be found from the 
+ * formula below.
+ * max(x, 0) - x * z + log(1 + exp(-abs(x)))
```
zoq: Can you use the doxygen forumla syntax here, so that we can render the expression? For more information take a look at: https://www.stack.nl/~dimitri/doxygen/manual/formulas.html

src/mlpack/methods/ann/layer/sigmoid_cross_entropy_error.hpp
```diff
@@ -0,0 +1,107 @@
...
+  /**
+   * Ordinary feed backward pass of a neural network.
+   *
+   * @param input The propagated input activation.
+   * @param target The target vector.
+   * @param output The calculated error.
+   */
+  template<typename eT>
+  inline void Backward(const arma::Mat<eT>&& input,
+                const arma::Mat<eT>&& target,
```
zoq: Pedantic style issue, can you align the ``target`` and ``output`` parameter with the first parameter?

src/mlpack/tests/ann_layer_test.cpp
```diff
@@ -927,6 +927,46 @@ BOOST_AUTO_TEST_CASE(SimpleLogSoftmaxLayerTest)
...
+  expectedOutput = arma::mat(
+      "0.7310586 0.88079709 -0.04742587 0.98201376 -0.00669285");
+  module.Backward(std::move(input2), std::move(target2), std::move(output));
+  for (size_t i = 0; i < output.n_elem; i++)
+    BOOST_REQUIRE_SMALL(output(i) - expectedOutput(i), 1e-5);
+
+  BOOST_REQUIRE_EQUAL(output.n_rows, input2.n_rows);
+  BOOST_REQUIRE_EQUAL(output.n_cols, input2.n_cols);
+}
+
```
zoq: Another pedantic style issue, can you remove the extra line here?

kris-singh:
Okay i fixed all the comments.

zoq:
Thanks, that was fast, looks good for me. I think this is ready to be merged.

src/mlpack/methods/ann/layer/sigmoid_cross_entropy_error.hpp
```diff
@@ -0,0 +1,107 @@
...
+
+#include <mlpack/prereqs.hpp>
+
+namespace mlpack {
+namespace ann /** Artificial Neural Network. */ {
+
+/**
+ * The SigmoidCrossEntropyError performance function measures the network's
+ * performance according to the cross-entropy function.
+ * between the input and target distributions.
```
rcurtin: Maybe this is an extra sentence, or the period from the previous line is in the wrong place?

src/mlpack/methods/ann/layer/sigmoid_cross_entropy_error.hpp
```diff
@@ -0,0 +1,107 @@
...
+ * This function calculates the cross entropy
+ * given the real values instead of providing the sigmoid activations.
+ * The functions is much more numerically stable as can be found from the 
+ * formula below.
+ * \f$max(x, 0) - x * z + \log(1 +  e^{-|x|})\f$
+ * For more detail look here goo.gl/tRjS6j
```
rcurtin: Ah, I don't know how long that link will be good, so we should avoid using it, put the details here, or maybe find a paper that we can cite.

src/mlpack/methods/ann/layer/sigmoid_cross_entropy_error_impl.hpp
```diff
@@ -0,0 +1,76 @@
...
+template<typename InputDataType, typename OutputDataType>
+template<typename eT>
+inline double SigmoidCrossEntropyError<InputDataType, OutputDataType>::Forward(
+    const arma::Mat<eT>&& input, const arma::Mat<eT>&& target)
+{
+  arma::uvec positive = arma::find(input > 0);
+  arma::mat output;
+  SoftplusFunction::Fn(static_cast<arma::mat>(-arma::abs(input)), output);
+  double loss = arma::accu(input(positive)) - arma::accu(input % target) +
+        arma::accu(output);
```
rcurtin: I think you could optimize this to avoid the `positive` vector, maybe like this:

```
double loss = arma::accu(arma::max(input, 0)) - arma::accu(input % target) + arma::accu(output);
```

I am not 100% sure that `arma::accu(arma::max(input, 0))` will avoid creating a temporary, but if it doesn't, you could also just loop over the `input` matrix.

src/mlpack/methods/ann/layer/sigmoid_cross_entropy_error_impl.hpp
```diff
@@ -0,0 +1,76 @@
...
+
+  arma::uvec positive = arma::find(input > 0);
+  arma::uvec negative = arma::find(input < 0 || input == 0);
+
+  SoftplusFunction::Deriv(static_cast<arma::mat>(-arma::abs(input(positive))),
+      temp);
+  output(positive) = 1 - target(positive) - temp;
+  SoftplusFunction::Deriv(static_cast<arma::mat>(-arma::abs(input(negative))),
+      temp);
+  output(negative) = -(target(negative)) - negative;
```
rcurtin: Do you think you could implement this all as a single loop over the `target` matrix?  i.e.

```
for (size_t i = 0; i < output.n_elem; ++i)
{
  const double val = target[i];
  if (val > 0)
  {
    output[i] = 1 - val - Deriv(-std::abs(input[i]));
  }
  else
  {
    output[i] = -val - Deriv(-std::abs(input[i]));
  }
}
```

Also, is `-(target(negative)) - negative` correct, or should that be `-(target(negative)) - temp`?

src/mlpack/tests/ann_layer_test.cpp
```diff
@@ -927,6 +927,45 @@ BOOST_AUTO_TEST_CASE(SimpleLogSoftmaxLayerTest)
...
+ */
+
+BOOST_AUTO_TEST_CASE(SimpleSigmoidCrossEntropyLayerTest)
+{
+  arma::mat input1, input2, output, target1, target2, expectedOutput;
+  SigmoidCrossEntropyError<> module;
+
+  // Test the Forward function on a user generator input and compare it against
+  // the manually calculated result.
+  input1 = arma::mat("0.5 0.5 0.5 0.5 0.5 0.5 0.5 0.5");
```
rcurtin: Since the behavior is different depending on whether the input is positive or negative, should you add some tests with negative inputs?

src/mlpack/methods/ann/layer/sigmoid_cross_entropy_error_impl.hpp
lozhnikov: I think `-(target(negative)) + temp` is correct i.e. the sign of the second term was incorrect.
If `x < 0` then the derivative of `log(1+exp(-abs(x)))` is equal to the derivative of `log(1 + exp(x))`.

rcurtin:
@kris-singh: did you want to handle the issues I pointed out?  Once those are done I think that everyone agrees this is ready for merge (@zoq or @lozhnikov correct me if I am wrong there please).

kris-singh:
I am just a bit busy right now. I will look at it on the weekend if that is
okay.

rcurtin:
Of course, look when you have a chance. You did a lot of hard work over the summer so I want to make sure we can merge as much as we can. :)

lozhnikov:
@rcurtin Let me look closely into the test. I don't understand why the test didn't point out this issue (https://github.com/mlpack/mlpack/pull/1104#discussion_r137060612  https://github.com/mlpack/mlpack/pull/1104#discussion_r137109612).

rcurtin:
@lozhnikov: I think that the issue wasn't pointed out by the tests because the Forward() function is only being tested with positive inputs:

```
+  // Test the Forward function on a user generator input and compare it against
+  // the manually calculated result.
+  input1 = arma::mat("0.5 0.5 0.5 0.5 0.5 0.5 0.5 0.5");
```

src/mlpack/methods/ann/layer/sigmoid_cross_entropy_error_impl.hpp
kris-singh: Could you explain arma::accu(arma::max(input, 0)). This  would return a max number along every column whereas arma::find(input > 0) find the indicies of every non-negative input. So, how could one replace another.

rcurtin: Ah yeah, sorry, I used the wrong function here.  Instead of `arma::accu(arma::max(input, 0))` I meant `arma::accu(arma::clamp(input, 0, std::numeric_limits<double>::max()))`.

ShikharJ: @rcurtin `arma::clamp` does create a copy of the original matrix. How do you wish to proceed here?

rcurtin: @ShikharJ: are you sure that the combination of `accu(clamp(...))` will create a temporary matrix?  It looks like `clamp()` returns an `arma::mtOp<...>` which signifies that the clamp operation should be done, but I think then `accu()` will use an `arma::Proxy<>` object, which I think in turn will create a temporary for an `mtOp<>`.  The best way to check for sure would be to create a simple program, compile with `-DARMA_EXTRA_DEBUG`, and parse through the output to see if the constructor for `arma::Mat<eT>` is called during the `accu(clamp())` operation.

In any case, if you do confirm that this makes a copy, you can just write a loop over the input matrix and manually clamp the elements to be greater than zero as you sum them.

ShikharJ: The `Armadillo` documentation mentions this about  `clamp()`:
```
Create a copy of X with each element clamped to be between min_val and max_val
```
`accu()` should just return the sum of the intended elements. So all in all, `accu(clamp(...))` will internally use a temporary, but just return the sum in the end. Am I misunderstanding the context of your concern?

rcurtin: Yes, so the documentation is not misleading but it doesn't mention that Armadillo is sometimes able to take advantage of expressions and avoid copies.  Based on what I saw in the Armadillo code, Armadillo is not currently able to avoid a copy in this case, so if it is ok with you, I think it would be better to manually write a loop to calculate the sum here.

rcurtin: Ah, I see in #1200 that you already wrote the loop.  Thanks! :)

ShikharJ:
@zoq I think this PR can also be closed now. Thanks @kris-singh for your work. This has been incorporated in #1200.

zoq:
Correct, thanks @kris-singh and @ShikharJ for the great work.

PR #1105, opened 2017-08-27
NEAT: Neural Evolution through Augmented Topology

kartik-nighania:
this class was created by @BangLiu which implements a neural evolution technique.

The goal of this new PR is to bring a better platform to discuss and document the conversation that i have with my mentor about the changes and fixes that went through while debugging.


zoq:
Nice commit message :)

sourabhvarshney111:
@kartik-nighania I am not sure whether this is complete or not. @zoq probably this can be merged.

zoq:
We encountered an issue were for some tasks, the population doesn't change/evolve after a number of iterations, I think ``IsSameSpecies`` does something strange since we end up with a single genome.

sourabhvarshney111:
@zoq Is that the only problem. If that is only problem, I would try to go through the code and try to figure out the issue.

zoq:
If I remember right yes, I think the first step would be to reproduce the issue, let me know if you need any help.

sourabhvarshney111:
Any idea when this issue used to occur @zoq because I don't have any idea regarding that.

zoq:
We used the mario task, but I think you could also use the XOR task to reproduce the issue.

PR #1106, opened 2017-08-29
Contributors list - added Kartik Nighania

kartik-nighania:


rcurtin:
Looks good, thanks. :)

PR #1107, opened 2017-08-30
Fast LSTM + Peephole connections

zoq:
A faster version of the LSTM. The input, forget and output gates, as well as the hidden state,  are computed at one step. Note, the Fast LSTM implementation does not support peephole connections and uses an approximation of the sigmoid function. This PR also refactors the current LSTM to support peephole connections as well as using a matrix instead of a list to accelerate the Forward and Backward pass.


rcurtin:
It seems like the `RecurrentNetworkTest` takes a very long time for the `FastLSTM`:

```
$ time bin/mlpack_test --log_level=all -t RecurrentNetworkTest
Running 7 test cases...
Entering test suite "mlpackTest"
Entering test suite "RecurrentNetworkTest"
Entering test case "SequenceClassificationTest"
/home/ryan/src/mlpack-zoq/src/mlpack/tests/recurrent_network_test.cpp(152): info: check successes >= 1 passed
Leaving test case "SequenceClassificationTest"; testing time: 820809mks
Entering test case "LSTMRecursiveReberGrammarTest"
/home/ryan/src/mlpack-zoq/src/mlpack/tests/recurrent_network_test.cpp(529): info: check successes >= 1 passed
Leaving test case "LSTMRecursiveReberGrammarTest"; testing time: 1025443mks
Entering test case "FastLSTMRecursiveReberGrammarTest"
/home/ryan/src/mlpack-zoq/src/mlpack/tests/recurrent_network_test.cpp(529): info: check successes >= 1 passed
Leaving test case "FastLSTMRecursiveReberGrammarTest"; testing time: 3854754009mks
Entering test case "GRURecursiveReberGrammarTest"
/home/ryan/src/mlpack-zoq/src/mlpack/tests/recurrent_network_test.cpp(529): info: check successes >= 1 passed
Leaving test case "GRURecursiveReberGrammarTest"; testing time: 1575843mks
Entering test case "LSTMDistractedSequenceRecallTest"
/home/ryan/src/mlpack-zoq/src/mlpack/tests/recurrent_network_test.cpp(716): info: check successes >= 1 passed
Leaving test case "LSTMDistractedSequenceRecallTest"; testing time: 6928566mks
Entering test case "FastLSTMDistractedSequenceRecallTest"
/home/ryan/src/mlpack-zoq/src/mlpack/tests/recurrent_network_test.cpp(716): info: check successes >= 1 passed
Leaving test case "FastLSTMDistractedSequenceRecallTest"; testing time: 993165627mks
Entering test case "GRUDistractedSequenceRecallTest"
/home/ryan/src/mlpack-zoq/src/mlpack/tests/recurrent_network_test.cpp(716): info: check successes >= 1 passed
Leaving test case "GRUDistractedSequenceRecallTest"; testing time: 694084mks
Leaving test suite "RecurrentNetworkTest"
Leaving test suite "mlpackTest"
```

So it looks like the FastLSTM tests are taking 2 orders of magnitude or more longer than the equivalent LSTM tests.  I think that the issue has to do with OpenMP:

```
$ time OMP_NUM_THREADS=1 bin/mlpack_test --log_level=all -t RecurrentNetworkTest
Running 7 test cases...
Entering test suite "mlpackTest"
Entering test suite "RecurrentNetworkTest"
Entering test case "SequenceClassificationTest"
/home/ryan/src/mlpack-zoq/src/mlpack/tests/recurrent_network_test.cpp(152): info: check successes >= 1 passed
Leaving test case "SequenceClassificationTest"; testing time: 431716mks
Entering test case "LSTMRecursiveReberGrammarTest"
/home/ryan/src/mlpack-zoq/src/mlpack/tests/recurrent_network_test.cpp(529): info: check successes >= 1 passed
Leaving test case "LSTMRecursiveReberGrammarTest"; testing time: 1040453mks
Entering test case "FastLSTMRecursiveReberGrammarTest"
/home/ryan/src/mlpack-zoq/src/mlpack/tests/recurrent_network_test.cpp(529): info: check successes >= 1 passed
Leaving test case "FastLSTMRecursiveReberGrammarTest"; testing time: 3229699mks
Entering test case "GRURecursiveReberGrammarTest"
/home/ryan/src/mlpack-zoq/src/mlpack/tests/recurrent_network_test.cpp(529): info: check successes >= 1 passed
Leaving test case "GRURecursiveReberGrammarTest"; testing time: 1562118mks
Entering test case "LSTMDistractedSequenceRecallTest"
/home/ryan/src/mlpack-zoq/src/mlpack/tests/recurrent_network_test.cpp(716): info: check successes >= 1 passed
Leaving test case "LSTMDistractedSequenceRecallTest"; testing time: 897240mks
Entering test case "FastLSTMDistractedSequenceRecallTest"
/home/ryan/src/mlpack-zoq/src/mlpack/tests/recurrent_network_test.cpp(716): info: check successes >= 1 passed
Leaving test case "FastLSTMDistractedSequenceRecallTest"; testing time: 1715269mks
Entering test case "GRUDistractedSequenceRecallTest"
/home/ryan/src/mlpack-zoq/src/mlpack/tests/recurrent_network_test.cpp(716): info: check successes >= 1 passed
Leaving test case "GRUDistractedSequenceRecallTest"; testing time: 620698mks
Leaving test suite "RecurrentNetworkTest"
Leaving test suite "mlpackTest"

*** No errors detected

real    0m9.508s
user    0m9.404s
sys     0m0.100s
```

When I make only one OpenMP thread, the FastLSTM code is much quicker (but still 2-3x slower than the equivalent LSTM test).

My guess is that the OpenMP-ized implementation of `FastSigmoid()` is actually causing slowdowns because of the overhead of OpenMP.  I think either using one of the Armadillo functions like `for_each()` (maybe that won't work here) or hand-writing a loop that exploits SIMD might be the better solution here.  When I wrote the SIMD code for the Gini index for the decision tree, I created a simple program that only called the SIMD-ized function with `-fopt-info-vec-missed` and other related options to make sure that the compiler's autovectorizer was working right.  Maybe there is a better way, I am not sure.

zoq:
Thanks for all the information, I'll take a look into the issue.

zoq:
I've tested different ideas to accelerate the sigmoid function:

- hard sigmoid: ``max(0, min(1, (x + 1) * 0.5))``, the function uses just two cut points (interpolation), the error is larger than for the ultra-fast version of the sigmoid function. In my tests, the error effected the numerical stability so I discard the idea.


- I've tested a preallocated lookup table but the speed was comparable with the current approximation or even slower, and if I include the time to build the table (build at compile time) I think the fast sigmoid function is the better choice.

- I've tried to use SIMD code to accelerate the function but couldn't see any speedup in that direction (branch elimination included).

I ended up with removing the OpenMP part from the FastSigmoid function, optimized the test to get the test time down. Using the same architecture the FastLSTM layer is about 0.3-0.4x faster on my machine.

I'll update the post once, I get the travis test times.

zoq:
**with OpenMP**
DEBUG=OFF : RecurrentNetworkTest  ``164.02 sec``
DEBUG=ON  : RecurrentNetworkTest  ``404.48 sec``

**Optimized test + without OpenMP**
Debug=OFF : RecurrentNetworkTest ``75.68 sec``
DEBUG=ON  : RecurrentNetworkTest ``165.17 sec``

rcurtin:
Agreed, now I see significant speedup with LSTM models:

```
Leaving test case "LSTMReberGrammarTest"; testing time: 941734mks
Leaving test case "FastLSTMReberGrammarTest"; testing time: 639095mks
```

```
Leaving test case "LSTMDistractedSequenceRecallTest"; testing time: 2103651mks
Leaving test case "FastLSTMDistractedSequenceRecallTest"; testing time: 691919mks
```

Really nice improvement! :)

src/mlpack/methods/ann/layer/fast_lstm.hpp
```diff
@@ -0,0 +1,304 @@
...
+ *
+ * @code
+ * @article{Hochreiter1997,
+ *   author  = {Hochreiter, Sepp and Schmidhuber, J\"{u}rgen},
+ *   title   = {Long Short-term Memory},
+ *   journal = {Neural Comput.},
+ *   year    = {1997}
+ * }
+ * @endcode
+ *
```
rcurtin: I think it might be worthwhile in this documentation to add a link to the regular `LSTM` class (and in the `LSTM` class, add a link to this class as something quicker that could be used).

src/mlpack/methods/ann/rnn_impl.hpp
```diff
@@ -121,7 +119,7 @@ void RNN<OutputLayerType, InitializationRuleType>::ResetCells()
 {
   for (size_t i = 1; i < network.size(); ++i)
```
rcurtin: Should this be from `i = 0` to `network.size()`?  This seems to fail if you have an LSTM as the first layer of a network.

zoq: You are right, we have to perform the backward pass (layer = 0) if the layer wraps another layer or parameter (in case of the LSTM class). However, it's unnecessary for some layers e.g. linear layer, where the error from the previous layer is used; in this case, we can avoid the extra backward function since there is no layer which might use the calculated error. I used an Identity layer as the first layer to solve the problem, but it's definitely not a good solution.

src/mlpack/methods/ann/layer/lstm_impl.hpp
```diff
@@ -31,351 +26,425 @@ LSTM<InputDataType, OutputDataType>::LSTM()
...
+  outParameter.cols(forwardStep + batchSize,
+      forwardStep + batchSize + batchStep) =
+      cellActivation.cols(forwardStep, forwardStep + batchStep) %
+      outputGateActivation.cols(forwardStep, forwardStep + batchStep);
+
+  output = OutputType(outParameter.memptr() +
+      (forwardStep + batchSize) * outSize, outSize, batchSize, false, false);
+
+  forwardStep += batchSize;
+  if ((forwardStep / batchSize) == rho)
```
rcurtin: Should this be `rhoSize` instead of `rho`?  When I was doing some timing simulations with this implementation, I had problems because `rho` was set to `SIZE_MAX` but `rhoSize` was 20, so `forwardStep` would never be reset.  It's possible there is something else I misunderstood.

zoq: Nice catch, but since ``rho`` controls the maximum number of steps to backpropagate through time, it could be < ``rhoSize``, so we should use ``std::min(rho, rhoSize)``.

PR #1114, opened 2017-09-08
Python bindings bugfixes

rcurtin:
This PR fixes a handful of small bugs that a colleague at Symantec reported and that I found:

 * Prevent segfaults on problems by modifying Cython definitions to point out that exceptions can be thrown (i.e. add `except +` to `cdef` lines).

 * Don't print any backtraces, because this seems to cause segfaults somewhere in libbfd.

 * Reset timers every time bindings are run, in case there was a failure last time that left the timers running.

 * When a parameter name is a Python keyword (i.e. `lambda`), append an underscore to the parameter in all documentation and code.


rcurtin:
Ok, I'll go ahead and merge on Monday then in case there are other comments.

PR #1115, opened 2017-09-08
Update Boost_ADDITIONAL_VERSIONS

zoq:
Update the ``Boost_ADDITIONAL_VERSIONS`` variable, so that ``findBoost`` works if cmake is older as the installed version of boost.


rcurtin:
Or I guess I should wait for the checks to complete first :)

PR #1116, opened 2017-09-11
Matrix Completion with Frank Wolfe Type algorithm

czdiao:
The matrix completion code is added with Frank Wolfe type algorithm.
mlpack already has matrix completion code with SDP solver implemented by @stephentu , I just added another solver here.

It is known that SDP solvers for matrix completion problems do not scale well, i.e., it would be quite expensive for large scale problems. I added the Frank-Wolfe type solver here. However, this solver converges quite slowly (with rate only 1/k). In the unit test file, I used the two solvers to solve the same matrix completion problem, the FW solver would use 40-50 seconds on my computer, with much larger error tolerance.

So we can see both solvers have their obvious drawbacks. Other popular methods for matrix completion problems are using SGD solver and Singular Value Thresholding algorithm. I am also interested in implementing those two in the future if you like them.

In order to have different algorithms for matrix completion problem exist together in mlpack, I implemented a template class `MatrixCompletion`. However, different algorithms require different input parameters, so they need to call different constructors. (You could see in `mlpack/methods/matrix_completion/mc_sdp_solver.hpp` and `mlpack/methods/matrix_completion/mc_fw_solver.hpp `) I don't know if you like the current implementation in this way. Or you could give me some other suggestions.


src/mlpack/methods/matrix_completion/mc_fw_update_matrix.hpp
```diff
@@ -0,0 +1,238 @@
...
+ * mlpack is free software; you may redistribute it and/or modify it under the
+ * terms of the 3-clause BSD license.  You should have received a copy of the
+ * 3-clause BSD license along with mlpack.  If not, see
+ * http://www.opensource.org/licenses/BSD-3-Clause for more information.
+ */
+#ifndef MLPACK_METHODS_MATRIX_COMPLETION_MC_FW_UPDATE_MATRIX_HPP
+#define MLPACK_METHODS_MATRIX_COMPLETION_MC_FW_UPDATE_MATRIX_HPP
+
+#include "mc_fw_function.hpp"
+#include <mlpack/core/optimizers/proximal/proximal.hpp>
```
zoq: Do you mind to move the ``Proximal`` class into the ``fw`` folder? I think it would improve the overall code structure since the ``Proximal`` class isn't directly an optimizer. Let me know what you think.

czdiao: Thanks for the review. Sorry for the late reply since I was busy last weekend. This is actually the idea of @stephentu  to make a `Proximal` class folder like this. Since Proximal Algorithms are a large group of optimization algorithms based on proximal operators, and the projections I implemented here are 2 cases of them, I don't know if any one would add more algorithms of this type in the future. Since it is not directly an optimizer here, I think it is ok for me to move them into `fw` folder for now. I will leave a message to Stephen to see what he thinks about it. :D

stephentu: Hi @zoq. The idea was, if anybody ever comes along and need prox operators or projection operators (ie if they implement projected gradient descent), then they can reuse these operators. Perhaps the naming is not the best, but I do think there is value in keeping it in a separate location to be shared.

src/mlpack/methods/matrix_completion/mc_fw_solver.hpp
```diff
@@ -0,0 +1,130 @@
...
+
+  //! This constructor type is not supported.
+  MCFWSolver(const size_t m,
+             const size_t n,
+             const arma::umat& indices,
+             const arma::vec& values,
+             const size_t r) :
+      function(indices, values, m, n),
+      fwSolver(ConstraintSolver(1), UpdateMatrix(0))
+  { Log::Fatal << "No such constructor!" << std::endl; }
```
stephentu: Why do we have these constructors if we don't support them?

src/mlpack/methods/matrix_completion/mc_fw_solver.hpp
```diff
@@ -0,0 +1,130 @@
...
+   *    length p).
+   * @param tau Nuclear norm in constraint.
+   */
+  MCFWSolver(const size_t m,
+             const size_t n,
+             const arma::umat& indices,
+             const arma::vec& values,
+             const double tau) :
+      function(indices, values, m, n),
+      fwSolver(ConstraintSolver(1), UpdateMatrix(tau), 140000)
```
stephentu: Can you document what this 140000 magic constant does?

src/mlpack/methods/matrix_completion/mc_fw_update_matrix.hpp
```diff
@@ -0,0 +1,238 @@
...
+    function.GetKnownEntries(oldCoords, oldEntries);
+    function.GetKnownEntries(v, vEntries);
+
+    double gamma = arma::dot(b - oldEntries, vEntries);
+    gamma = gamma / std::pow(arma::norm(vEntries, "fro"), 2);
+    gamma = std::min(gamma, 1.0);
+    currentCoeffs = (1.0 - gamma) * currentCoeffs;
+    AddAtom(s, gamma * tau);
+
+    ProjectedGradientEnhancement(function, 0.6);
```
stephentu: What does this 0.6 do here?

src/mlpack/methods/matrix_completion/mc_fw_update_matrix.hpp
```diff
@@ -0,0 +1,238 @@
...
+  {
+    // Line search, with explicit solution here.
+    arma::mat v = tau * s - oldCoords;
+    arma::vec b = function.Values();
+
+    arma::vec oldEntries, vEntries;
+    function.GetKnownEntries(oldCoords, oldEntries);
+    function.GetKnownEntries(v, vEntries);
+
+    double gamma = arma::dot(b - oldEntries, vEntries);
```
stephentu: Where does this sequence of updates for gamma come from? If it suggested by a paper, can you leave a short comment with a pointer.

src/mlpack/methods/matrix_completion/mc_fw_update_matrix.hpp
zoq: Thanks for the clarification, I guess if someone is going to implement e.g. projected gradient descent as you said, it would make sense to put the proximal operations inside the projected gradient descent optimizer folder. My point is, right now, each folder is basically an optimizer that follows some kind of unified interface; we can directly use each optimizer in combination with some existing method. The proximal folder is an exception. Not sure that's something we should worry about.

src/mlpack/core/optimizers/fw/constr_matrix_lp.hpp
```diff
@@ -0,0 +1,91 @@
...
+    S = U * diagmat(SingularValDual) * V.t();
+  }
+
+ private:
+  //! lp norm, 1<=p<=inf;
+  //! use std::numeric_limits<double>::infinity() for inf norm.
+  double p;
+
+  //! Solver for vector lp ball optimal dual.
+  ConstrLpBallSolver vector_lp_solver;
```
zoq: Can you use camel casing for all parameter names; for more information take a look at: https://github.com/mlpack/mlpack/wiki/DesignGuidelines#naming-conventions

src/mlpack/methods/matrix_completion/matrix_completion.hpp
```diff
@@ -13,50 +14,65 @@
...
+ * \f]
  *
  * where ||X||_* denotes the nuclear norm (sum of singular values of X).
  *
- * For a theoretical treatment of the conditions necessary for exact recovery,
- * see the following paper:
+ * For a theoretical treatment of the conditions necessary for exact recovery
+ * using the above model, see the following paper:
  *
  *   A Simpler Appoarch to Matrix Completion.
```
zoq: Do you mind to use bibtex here? (example: https://github.com/mlpack/mlpack/blob/master/src/mlpack/core/optimizers/parallel_sgd/parallel_sgd.hpp#L27)

src/mlpack/methods/matrix_completion/matrix_completion_impl.hpp
```diff
@@ -0,0 +1,106 @@
...
+  {
+    Log::Fatal << "MatrixCompletion::CheckValues(): the number of constraint "
+        << "indices (columns of constraint indices matrix) does not match the "
+        << "number of constraint values (length of constraint value vector)!"
+        << std::endl;
+  }
+
+  for (size_t i = 0; i < values.n_elem; i++)
+  {
+    if (indices(0, i) >= m || indices(1, i) >= n)
```
zoq: Can you use ``{`` for a multiline expression to improve the readability?

src/mlpack/methods/matrix_completion/mc_fw_function.hpp
```diff
@@ -0,0 +1,122 @@
...
+  /**
+   * Get the values of a given matrix at positions with known "indices".
+   *
+   * @param X Input given matrix.
+   * @param xValues Output entries of the matrix X.
+   */
+  void GetKnownEntries(const arma::mat& X, arma::vec& xValues)
+  {
+    xValues.set_size(arma::size(values));
+    for (size_t i = 0; i < indices.n_cols; i++)
```
zoq: I think instead of a for loop you could also write: ``xValues = X(indices.row(0), indices.row(1))``.

src/mlpack/methods/matrix_completion/mc_fw_update_matrix.hpp
```diff
@@ -0,0 +1,238 @@
...
+    double value = function.Evaluate(x);
+
+    for (size_t iter = 1; iter < maxIterations; iter++)
+    {
+      // Update currentCoeffs with gradient descent method.
+      // We caculate the gradient directly here, so we don't need to generate
+      // a large sparse matrix.
+      arma::vec entries;
+      function.GetKnownEntries(x, entries);
+      entries = entries - function.Values();
```
zoq: Just a minor note; you could just write: ``entries -= function.Values();``.

src/mlpack/methods/matrix_completion/mc_fw_update_matrix.hpp
```diff
@@ -0,0 +1,238 @@
...
+
+      arma::vec gradient(currentCoeffs.n_elem);
+      for (size_t i = 0; i < gradient.n_elem; i++)
+      {
+        arma::vec entriesAtom;
+        function.GetKnownEntries(currentAtoms.slice(i), entriesAtom);
+
+        gradient(i) = arma::dot(entriesAtom, entries);
+      }
+      currentCoeffs = currentCoeffs - stepSize * gradient;
```
zoq: Just a minor note; you could just write: ``currentCoeffs -= function.Values();``.

src/mlpack/methods/matrix_completion/mc_sdp_solver.hpp
```diff
@@ -0,0 +1,169 @@
...
+ * the remaining unknown values of X
+ *
+ *   min ||X||_* subj to X_ij = M_ij
+ *
+ * where ||X||_* denotes the nuclear norm (sum of singular values of X).
+ *
+ * For a theoretical treatment of the conditions necessary for exact recovery,
+ * see the following paper:
+ *
+ *   A Simpler Appoarch to Matrix Completion.
```
zoq: See my comment about bibtex.

src/mlpack/methods/matrix_completion/mc_sdp_solver.hpp
```diff
@@ -0,0 +1,169 @@
...
+      sdp(indices.n_cols, 0,
+          arma::randu<arma::mat>(m + n, DefaultRank(m, n, indices.n_cols)))
+    {
+      Log::Fatal << "No such constructor for SDP solver!" << std::endl;
+    }
+
+  /**
+   * Solve the underlying sdp to fill in the remaining values.
+   *
+   * @param recovered Will contain the completed matrix.
```
zoq: Can you add a comment about the other parameter as well?

src/mlpack/methods/matrix_completion/mc_sdp_solver.hpp
```diff
@@ -0,0 +1,169 @@
...
+  MCSDPSolver(const size_t m,
+              const size_t n,
+              const arma::umat& indices,
+              const arma::vec& values) :
+      sdp(indices.n_cols, 0,
+          arma::randu<arma::mat>(m + n, DefaultRank(m, n, indices.n_cols)))
+  { InitSDP(m, n, indices, values); }
+
+  //! This type of constructor is not feasible for SDP solver.
+  MCSDPSolver(const size_t m,
```
zoq: I guess, we can remove the constructor in this case?

src/mlpack/methods/matrix_completion/mc_sdp_solver.hpp
```diff
@@ -0,0 +1,169 @@
...
+    sdp.Optimize(recovered);
+    recovered = recovered * trans(recovered);
+    recovered = recovered(arma::span(0, m - 1), arma::span(m, m + n - 1));
+  }
+
+ private:
+    //! The underlying SDP to be solved.
+    optimization::LRSDP<optimization::SDP<arma::sp_mat>> sdp;
+
+    size_t DefaultRank(const size_t m, const size_t n, const size_t p)
```
zoq: Do you mind to add a simple description here?

src/mlpack/methods/matrix_completion/mc_sdp_solver.hpp
```diff
@@ -0,0 +1,169 @@
...
+        // For more details, see
+        //
+        //   On the rank of extreme matrices in semidefinite programs and the
+        //   multiplicity of optimal eigenvalues.
+        //   Pablo Moscato, Michael Norman, and Gabor Pataki.
+        //   Math Oper. Res., 23(2). 1998.
+        const size_t mpn = m + n;
+        float r = 0.5 + sqrt(0.25 + 2 * p);
+        if (ceil(r) > mpn)
+            r = mpn; // An upper bound on the dimension.
```
zoq: We could avoid the extra copy here, by using ``return ceil(mpn)``.

src/mlpack/methods/matrix_completion/matrix_completion.hpp
```diff
@@ -13,50 +14,65 @@
...
  * arma::mat recovered; // will contain the completed matrix
  *
- * MatrixCompletion mc(m, n, indices, values);
+ * MatrixCompletion<MCSolverType> mc(m, n, indices, values);
  * mc.Recover(recovered);
  * @endcode
  *
- * @see LRSDP
  */
+template<typename MCSolverType>
```
rcurtin: I like templatizing the solver, but I think there are a couple of ways we can improve it here.  From what I can tell, the two solvers you are considering---`MCSDPSolver` and `MCFWSolver`---have different APIs, and this means that half of the functions of `MatrixCompletion` can only be used with one solver and half can only be used with the other.  We should probably find a way to instead change the APIs of each solver to match, so that the templatization can work with either and the API is unified.

From what I can tell, the primary problem seems to be that `MCFWSolver()` accepts a parameter `tau` whereas `MCSDPSolver()` accepts either `initialPoint` or `r` (the maximum rank).  There is no clear relation (I think) between `tau` and `r`, so we can't combine the parameters.  Instead, I might suggest the following pattern which is used in other places in mlpack:

```
// Construct the MatrixCompletion object with the given optimizer.
MatrixCompletion(
    const size_t m,
    const size_t n,
    const arma::umat& indices,
    const arma::vec& values,
    MCSolverType solver = MCSolverType()); // use std::move(solver) to avoid copying the solver in the implementation

// Access the solver to get its parameters.
const MCSolverType& Solver() const { return solver; }
// Modify the solver.
MCSolverType& Solver() { return solver; }
```

Internally, I would suggest having `MatrixCompletion` hold the `MatrixCompletionFWFunction` or `LRSDP<SDP<arma::sp_mat>>` objects.  You could expose a typedef in `MCSDPSolver` and `MCFWSolver` and call it something like `FunctionType`; then inside of `MatrixCompletion` you can hold `FunctionType function`.  Separating the function from the optimizer becomes important for the second usage example below.

Once all this is done, a user can set their given tau in one of two ways:

```
MatrixCompletion<MCFWSolver> m(...);
m.Solver().Tau() = myFavoriteTauValue;
// or alternately
MCFWSolver solver(myFavoriteTauValue); // all function-related parameters should move out of the solver
MatrixCompletion<MCFWSolver> m(..., std::move(solver));
```

I hope what I've written here makes sense.  Basically, the key is that I think we can restructure the design of the solver and function classes so that they all fit nicely into `MatrixCompletion`.  Let me know if you agree, or if I can clarify what I mean.

src/mlpack/methods/matrix_completion/matrix_completion_impl.hpp
```diff
@@ -0,0 +1,106 @@
...
+                 const double tau) :
+    m(m), n(n), mcSolver(m, n, indices, values, tau)
+{
+  CheckValues(indices, values);
+}
+
+
+template<typename MCSolverType>
+void MatrixCompletion<MCSolverType>::
+CheckValues(const arma::umat& indices, const arma::vec& values)
```
rcurtin: It might be possible to simply rely on Armadillo throwing an exception in the `sp_mat` constructor instead of having this `CheckValues()` function; I know that function will fail if the given matrix is of the wrong size.

src/mlpack/tests/matrix_completion_test.cpp
```diff
@@ -64,4 +66,54 @@ BOOST_AUTO_TEST_CASE(UniformMatrixCompletionSDP)
...
+ * The test data is the same as the previous test.
+ *
+ * The matrix is recovered by solving the model:
+ *
+ * min \sum_{i,j} (X_ij - M_ij)^2
+ * s.t. ||X||_* <= tau
+ *
+ * where ||X||_* is the nuclear norm of X.
+ *
+ * The convergence is really slow.
```
rcurtin: I see that in the test times, like your comment says, the test takes significantly longer than the SDP solver.  Do we have a way to reduce the runtime of this test?  Maybe the tolerance can be reduced?  The second question is, do we know situations where the FW solver will outperform the existing SDP solver?

src/mlpack/methods/matrix_completion/matrix_completion.hpp
czdiao: Thanks a lot for this idea! I was thinking that the way I did is really not elegant. I will change the structure to your suggestion.
I think I basically understand what you are suggesting. I just need a constructor of each `MCSolverType` with no input, and different parameters could be set later. Right?

As to the `FunctionType`, I just have a C++ question:
If `FunctionType` is a typedef defined in each solver, and the solver is a member in `MatrixCompletion` class. Can `MatrixCompletion` class know how `FunctionType` is defined?

src/mlpack/tests/matrix_completion_test.cpp
czdiao: It is known that the convergence rate of FW is only O(1/k). So we will need much more iterations to get a small error. I guess there is no way to fix it. The benefit of FW is that it is really fast in each iteration. This could be seen for large scale problems (when the matrix size is huge). In that case, each iteration of SDP is very slow. So the comparison between FW and SDP is:
(fast in each step) x (more iterations)     versus    (slow in each step) x (less iterations)

In the unit test, the problem size is really small, so SDP is much faster to beat FW. I guess if you want to see the situation that FW outperforms SDP, we need to test with large scale problems, where SDP becomes infeasible.

There is one thing that is a little bit embarrassing for large problems tests so far. FW is fast in each iteration because it requires solving Singular Value Decomposition with only the largest singular value/vector. However, Armadillo doesn't have it now for dense matrices. Stephen used to say he would add it in Armadillo, but I guess he is too busy. So the current FW code performs a full SVD. (This kind of prohibit a large scale problem tests comparison for now.)

In the current unit test file, I tried the example Stephen saved for SDP solver. I manually tuned the iteration numbers for FW and the step size in the projected gradient step (the weird numbers Stephen asked in his comments). Stephen also suggested to try some other examples. I will try it later to see if we can find a better example that is in favor of FW.

Sorry that I am really busy recently, I will definitely fix the problems all you guys mentioned soon. (Hopefully this weekend.) Thanks a lot for the suggestions. 😄

src/mlpack/methods/matrix_completion/matrix_completion.hpp
rcurtin: Great, I think the new structure will be an improvement.  Thanks!

`MatrixCompletion` can know what `FunctionType` is by using `SolverType::FunctionType`.  So, e.g., if I have a simple set of code like...

```
class MySolver
{
 public:
  typedef MyFunction FunctionType;
};

template<typename SolverType>
class MatrixCompletion
{
  typename SolverType::FunctionType function;
};
```

then if I create a `MatrixCompletion<MySolver>` type, then its `function` value will have type `MyFunction`.  I hope I understood the question correctly; let me know if not.

src/mlpack/tests/matrix_completion_test.cpp
rcurtin: No worries about being busy, sometimes life gets in the way. :)  I think maybe we should open a bug noting that `eigs()` can only be used with sparse matrices.  It should be pretty easy to fix actually, then we can just send a patch upstream to Conrad.

If the convergence is slow for the test, I think that is okay---the only goal of the test is to see if it works.  So I think it is okay to make the problem even smaller, or limit the number of iterations to get the test time down, as long as the result conditions we are checking are still meaningful.

For large-scale simulations, if you have a belief that FW will be faster than SDP, I think that is ok with me, and we can confirm that later once dense `eigs()` is available.  (If you are interested in doing that too, I can show you how, just let me know.)  I just want to make sure we are not going to commit code that will be slower than SDP in all cases, otherwise there is no reason for anyone to use it. :)

sourabhvarshney111:
@zoq @rcurtin Can you add the label inactive please? Looks like nobody is working on it from September, 2017?

rcurtin:
@sourabhvarshney111: I don't want to close this one inactive quite yet, because I think the functionality could be really useful.  So by leaving it open I am hoping that someone may come along and finish the work.

sourabhvarshney111:
@rcurtin Can you give me the references for this? I may try to complete this.

czdiao:
So sorry for the delay for this PR for such a long time. I am having my Ph.D. defense in the late of May. I think I won't have time before that. I would fix this PR after my defense if you guys can wait~

sourabhvarshney111:
@czdiao No issues. I was just going through all open PRs and just found this inactive. You can complete whenever you feel free.

rcurtin:
Hi Chenzhe, congratulations on your upcoming defense!  I hope that it will go well. :)  I'm happy to wait until after your defense unless someone else comes along and wants to work on it.

@sourabhvarshney111: I see no problem if you want to take up the task, but we can also wait for Chenzhe, your call---there is no shortage of open PRs if you are interested in others. :)  One of the papers to look into would be "A Simpler Approach to Matrix Completion" by Ben Recht, but there are also some other papers referenced in the code that might be worth looking at.

sourabhvarshney111:
@rcurtin I think @czdiao can complete this. As this will take time for me to first understand the code and then complete it. Meanwhile, GSoC is coming up. I might not get time time to complete this. Till then, I would complete any other open PR or would open new for adding new functionalities.

PR #1117, opened 2017-09-11
Downpour SGD

zoq:
One of my coworkers wanted to train a model over multiple machines, so I implemented Downpour SGD in C++ using mlpack and Boost MPI (OpenMPI). We haven't talked about running methods across multiple machines; do we like/plan to support such scenarios? If we don't like to go that path, I don't mind to close the PR, if we do I'll write some test cases.


rcurtin:
Hmm, I spent a while thinking about it.  I don't have a problem with MPI support but it raises a bunch of issues we should address:

1. Users shouldn't be required to have MPI installed---your code already handles that though it looks like.  But since DownpourSGD is a template class, if a user compiles from the command line, `HAS_MPI` won't be defined.  So I guess we can either tell the user to define that, or to use something like suggested here instead: https://stackoverflow.com/questions/28353969/mpi-preprocessor-macro .

2. Typically machine learning practitioners won't have an MPI cluster set up (and may be scared of it); instead they are more likely to have Spark or Hadoop setups.  So I think if we provide MPI support then we should also have some tutorials on how to set up and run the MPI program if you have, e.g., a Spark cluster.  I guess that is just setting up the hosts file and using mpirun, maybe there are a few other details.

3. As far as I can tell there's not really an MPI machine learning library so there is some opportunity here, especially since I've seen some studies that show that MPI can give order-of-magnitude-or-more speed improvements over Spark or Hadoop.  If we do do that though, we should probably be careful to make more than just Downpour SGD MPI-capable.

4. It looks like here the entire data matrix gets sent to all nodes, whereas in some other cases it would be better to just have parts of the data available on some nodes.  I wonder if Jack Poulson's Elemental distributed matrix library could be useful here.  I always wondered how easy it would be to wrap an Elemental matrix inside of Armadillo; I'm not sure if it would be easy or hard or even possible though.

Overall I guess I feel that boost::mpi is reasonable and doesn't have too much boilerplate, but I want to be careful to make sure that if we do start incorporating MPI-enabled classes, that they get the attention they deserve, so we can say 'mlpack works well with an MPI cluster' and not 'mlpack has an MPI class or two...'.  If the latter is all you're interested in for now, we could make a supplementary library 'mlpack-mpi' and store our MPI-enabled classes there. Then someday if we have enough MPI support we can merge that back into mlpack itself.  What do you think, what is your preferred option?

I definitely think that this support would be nice to have, and maybe it would allow us to compare with larger-scale projects like Spark-ML or Mahout or Parameter Server or whatever else.

zoq:
Yeah, I agree about the things that you are worried about.

(1) Thanks for pointing that out, I'll take a look into the issue.

(2) Agreed, Setting up an MPI cluster is really straightforward as you pointed out create a hostfile, setup up ssh keys and you are done. It looks like you could also run MPI on a Hadoop cluster (https://github.com/alibaba/mpich2-yarn). Not sure something similar exists for Spark, but I'll take a look at this point as well.

(3) If we go that direction, I'll go and add DownpourAdam and DownpourAdaDelta, which is straightforward since we can just use DownpourSGD as the basis, just like we did for Adam and AdaDelta, etc. Also, I'd like to point out that since we have an SGD MPI implementation a bunch of existing methods like ``LogisticRegression``, ``SparseAutoencoder``, ``SoftmaxRegression`` or the Neural Network code are already MPI complaint. So, it's not just ``Downpour SGD``. Anyway, I'll definitely take a look at the ``Sandblaster L-BFGS`` idea from the Large Scale Distributed Deep Networks paper.

(4) I agree, transferring the entier parameter matrix isn't optimal, I think if you adapt the batch size the effect is probably negligible as compared to the computation itself. Anyway, the idea to use Poulson's Elemental distributed matrix library is interesting, I'll take a look at it.

Managing two versions is something I'd like to avoid, so I would prefer to work on the MPI idea and to merge it, once we all think it's ready.

rcurtin:
Ah, great, that sounds good to me.  You are right that if we have MPI optimizers then we have MPI support for a lot of mlpack already.  The distributed matrix ideas can come sometime later, I think the MPI support itself is a great addition as long as we have the documentation and benchmarks to appeal to people who already have a Spark or Hadoop cluster and want to do large-scale learning (or, to appeal to HPC people who already have an MPI cluster setup and want to do machine learning).  I can get access to some nice MPI clusters and Spark clusters for comparison.  If you want me to run some simulations, just let me know.

rcurtin:
I think we should reopen this in the ensmallen repository, and I think that perhaps we should allow the user to manually include it and note in the documentation that it depends on Boost.MPI.  Since it wouldn't be a part of mlpack that actually makes it a bunch easier to include, in my opinion.

In any case it would be nice to add---actually I had someone ask about distributed optimization algorithms next week and they knew about downpour SGD and found it interesting. :+1:

PR #1124, opened 2017-09-20
Test ld gold on travis.

zoq:
Just a simple test to see the effect on the build time.


zoq:
Okay, the results are almost the same.

Issue #1123, opened 2017-09-19
Error compiling cython file


newling:
Hello

(((probably unrelated to issue #1122, this is a clean build on a different (ubuntu) machine)))

I'm encoutering an error with `make install`, the output below. I'm posting my cython version (I've encountered issues specific to cython versions in the past, so just in case):

```
cython --version
Cython version 0.23.4
```

.
.
.

[ 95%] Building Cython binding hmm_generate.so...
/usr/lib/python2.7/dist-packages/setuptools/dist.py:294: UserWarning: The version specified ('3.0.x') is an invalid version, this may not work as expected with newer versions of setuptools, pip, and PyPI. Please see PEP 440 for more details.
  "details." % self.metadata.version
running build_ext
cythoning /home/james/clustering/mlpack/mlpack/build/src/mlpack/bindings/python/mlpack/gmm_train.pyx to /home/james/clustering/mlpack/mlpack/build/src/mlpack/bindings/python/mlpack/gmm_train.cpp

Error compiling Cython file:
  \ - -----------------------------------------------------------
...
from cython.operator import dereference

cdef extern from "</home/james/clustering/mlpack/mlpack/src/mlpack/methods/det/det_main.cpp>" nogil:
  cdef void mlpackMain() nogil except +RuntimeError
  
  cdef cppclass DTree[T=*]:
                      ^
 \  - -----------------------------------------------------------

mlpack/det.pyx:21:23: Expected ']', found '='
building 'mlpack.det' extension
[...]

Issue #1119, opened 2017-09-13
Python models cannot be pickled


rcurtin:
If you try to pickle a Python model from an mlpack binding, it fails:

```
>>> o = logistic_regression(training=x, labels=y, lambda_=0.5)
[WARN ] --output_file ignored because --test_file is not specified.
[WARN ] --output_probabilities_file ignored because --test_file is not specified.
>>> import pickle
>>> pickle.dump(o['output_model'], open('model.pkl', 'wb'))
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/usr/lib/python2.7/pickle.py", line 1376, in dump
    Pickler(file, protocol).dump(obj)
  File "/usr/lib/python2.7/pickle.py", line 224, in dump
    self.save(obj)
  File "/usr/lib/python2.7/pickle.py", line 306, in save
    rv = reduce(self.proto)
  File "/usr/lib/python2.7/copy_reg.py", line 70, in _reduce_ex
    raise TypeError, "can't pickle %s objects" % base.__name__
TypeError: can't pickle LogisticRegressionType objects
```

I think just the `__getstate__()` and `__setstate__()` functions need to be added to the mlpack model types.

rcurtin:
Fixed in #1125.

PR #1125, opened 2017-09-25
Update binding output, fix pickling, fix installation

### LINKED ISSUE #1118: Python binding warnings are inaccurate or misleading
The warnings in bindings to other languages are generally written for command-line programs; example:

```
>>> o = logistic_regression(training=x, labels=y, lambda_=0.5)
[WARN ] --output_file ignored because --test_file is not specified.
[WARN ] --output_probabilities_file ignored because --test_file is not specified.
```

but neither of those warnings make very much sense.  We need a way to write generic error or warning messages that can be automatically reformatted for the language at hand.

### LINKED ISSUE #1122: ImportError: cannot import name test_python_binding
Both with a clean install from source, and using the package from the repo: 

```
import mlpack
ImportError: cannot import name test_python_binding
```

```
..........:~$ uname -a
Linux idbean 3.16.0-4-amd64 #1 SMP Debian 3.16.43-2 (2017-04-30) x86_64 GNU/Linux
```

rcurtin:
This fixes a number of Python binding problems that have been observed recently.

 * Work with Python to make mlpack objects pickle-able using boost::serialization.
 * Use generic functions to print warning messages, instead of custom messages that may be language-specific.  These are the majority of the changes here, and I will double-check them over the next days to make sure they are all correct.
 * Require Cython minimum version of 0.24.
 * Properly set PYTHONPATH during installation so that a custom installation directory will not fail to have Python bindings installed to it.
 * Avoid duplicate building of bindings when installing.

This fixes #1118, #1119, #1122 and #1123 among other things.


src/mlpack/bindings/python/print_doc_functions_impl.hpp
```diff
@@ -218,6 +218,37 @@ inline std::string ParamString(const std::string& paramName, const T& value)
...
+    return true;
+  return false;
+}
+
+inline bool IgnoreCheck(const std::vector<std::string>& constraints)
+{
+  bool anyOutput = false;
+  for (size_t i = 0; i < constraints.size(); ++i)
+    if (!CLI::Parameters()[constraints[i]].input)
+      anyOutput = true;
```
zoq: Instead of using ``anyOutput`` we can just return ``true``?

src/mlpack/bindings/python/print_doc_functions_impl.hpp
```diff
@@ -218,6 +218,37 @@ inline std::string ParamString(const std::string& paramName, const T& value)
...
+inline bool IgnoreCheck(
+    const std::vector<std::pair<std::string, bool>>& constraints,
+    const std::string& paramName)
+{
+  bool anyOutput = false;
+  for (size_t i = 0; i < constraints.size(); ++i)
+    if (!CLI::Parameters()[constraints[i].first].input)
+      anyOutput = true;
+
+  anyOutput &= !CLI::Parameters()[paramName].input;
```
zoq: We could return one anyOutput is true.

src/mlpack/methods/sparse_coding/sparse_coding_main.cpp
```diff
@@ -103,55 +104,25 @@ void mlpackMain()
...
-      Log::Warn << "--normalize (-N) ignored because --training_file (-t) is "
-          << "not specified." << endl;
-    if (CLI::HasParam("objective_tolerance"))
-      Log::Warn << "--objective_tolerance (-o) ignored because --training_file "
-          << "(-t) is not specified." << endl;
-    if (CLI::HasParam("newton_tolerance"))
-      Log::Warn << "--newton_tolerance (-w) ignored because --training_file "
-          << "(-t) is not specified." << endl;
-  }
+  if (CLI::HasParam("training"))
```
zoq: Do you mind to use ``{`` here to improve the readability?

src/mlpack/methods/softmax_regression/softmax_regression_main.cpp
```diff
@@ -119,23 +120,18 @@ void mlpackMain()
...
 
-  if ((CLI::HasParam("training") || CLI::HasParam("labels")) &&
-      !(CLI::HasParam("training") && CLI::HasParam("labels")))
-    Log::Fatal << "--labels_file must be specified with --training_file!"
-        << endl;
-
-  if (maxIterations < 0)
-    Log::Fatal << "Invalid value for maximum iterations (" << maxIterations
-        << ")! Must be greater than or equal to 0." << endl;
+  RequireParamValue<int>("max_iterations", [](int x) { return x >= 0; }, true,
```
zoq: That is a really nice idea.

src/mlpack/methods/random_forest/random_forest_main.cpp
```diff
@@ -70,81 +71,31 @@ PARAM_MODEL_OUT(RandomForestModel, "output_model", "Model to save trained "
...
-  }
-
-  if (CLI::HasParam("training") &&
-      CLI::HasParam("num_trees") &&
-      CLI::GetParam<int>("num_trees") <= 0)
-  {
-    Log::Fatal << "Invalid number of trees (" << CLI::GetParam<int>("num_trees")
-        << "); must be greater than 0!" << endl;
-  }
+  if (CLI::HasParam("training"))
```
zoq: Do you mind to use ``{`` here? I guess this also applies to a bunch of other lines.

rcurtin:
I'm happy to report that I've finally debugged the issues here and this is ready to merge.  It took a really long time to dig to the bottom of the issues and double-check all the output...

I'll go ahead and merge this in a week in case anyone has any comments.

src/mlpack/bindings/python/mlpack/arma_util.hpp
```diff
@@ -24,4 +24,124 @@ void SetMemState(T& t, int state)
   const_cast<arma::uhword&>(t.mem_state) = state;
 }
 
+/**
+ * Return the matrix's allocated memory pointer, unless the matrix is using its
+ * internal preallocated memory, in which case we copy that and return a
+ * pointer to the memory we just made.
+ */
+inline double* GetMemory(arma::mat& m)
```
zoq: Couldn't you write something like:

```c++
template<typename MatType>
inline typename MatType::elem_type* GetMemory(MatType& m)
{
}
```

I was thinking if we are going for: https://github.com/mlpack/mlpack/blob/master/doc/policies/elemtype.hpp I would have to add a bunch more, perhaps I missed something.

rcurtin: I was afraid Cython would have problems with the `MatType::elem_type` bit but I never tried it---let me do that now.

src/mlpack/bindings/python/print_doc_functions_impl.hpp
```diff
@@ -218,6 +218,31 @@ inline std::string ParamString(const std::string& paramName, const T& value)
...
 }
 
+inline bool IgnoreCheck(const std::string& paramName)
+{
+  return !CLI::Parameters()[paramName].input;
+}
+
+inline bool IgnoreCheck(const std::vector<std::string>& constraints)
+{
+  for (size_t i = 0; i < constraints.size(); ++i)
```
zoq: Pedantic style comment; I think adding ``{}`` to the outer for loop would improve the readability.

src/mlpack/core/util/param_checks_impl.hpp
```diff
@@ -0,0 +1,253 @@
...
+inline void RequireOnlyOnePassed(
+    const std::vector<std::string>& constraints,
+    const bool fatal,
+    const std::string& errorMessage)
+{
+  if (BINDING_IGNORE_CHECK(constraints))
+    return;
+
+  size_t set = 0;
+  for (size_t i = 0; i < constraints.size(); ++i)
```
zoq: Same pedantic style comment as for the ``IgnoreCheck`` function.

PR #938, opened 2017-03-14
Neuro CMAES-Algorithm

kartik-nighania:
I have used and modified Bang's genome and have implemeted cmaes into it .. 
Tests are remaining for it ..

PR #1128, opened 2017-09-29
CMA-ES (Covariance Matrix Adaptation Evolution Strategy)

zoq:
Implementation of the Covariance Matrix Adaptation Evolution Strategy as proposed by N. Hansen et al. in "Completely Derandomized Self-Adaptation in Evolution Strategies". I used #938 as a basis and simplified the code here and there. @kartik-nighania maybe you have time to take a look at the code.


kartik-nighania:
@zoq 
I have my test going. Can i took at it in 5 days ?

kartik-nighania:
@zoq 
I have my test going. Can I look at it in 5 days?

kartik-nighania:
@zoq 
I'm not getting much. But I want to discuss more.

zoq:
Okay, if you like, go through the code add a comments, but don't feel obligated. Also instead of using a custom function to get the eigenvalues as you did, I used armadillo, it might be slower since we don't take the time for each iteration into account. Also, once there is an armadillo function to calculate only the leading eigenvector it might be interesting to test if that could accelerate the method.

src/mlpack/core/optimizers/cmaes/cmaes_impl.hpp
```diff
@@ -0,0 +1,260 @@
...
+  const size_t numFunctions = function.NumFunctions();
+
+  // Population size.
+  if (lambda == 0)
+    lambda = (4 + std::round(3 * std::log(iterate.n_elem))) * 10;
+
+  // Parent weights.
+  const size_t mu = std::round(lambda / 2);
+  arma::vec w = std::log(mu + 0.5) - arma::log(arma::conv_to<arma::vec>::from(
+      arma::linspace<arma::uvec>(1, mu, mu + 1)));
```
rcurtin: I think you can get away with `arma::linspace<arma::vec>(1, mu, mu + 1)` here, no need for the `conv_to`.

src/mlpack/core/optimizers/cmaes/cmaes_impl.hpp
```diff
@@ -0,0 +1,260 @@
...
+
+  // The current visitation order (sorted by population objectives).
+  arma::uvec idx = arma::linspace<arma::uvec>(0, lambda - 1, lambda);
+
+  // Now iterate!
+  for (size_t i = 1; i < maxIterations; ++i)
+  {
+    // To keep track of where we are.
+    const size_t idx0 = (i - 1) % 3;
+    const size_t idx1 = i % 3;
```
rcurtin: I think that according to the algorithm you only ever need to look at time step `g` and `g + 1`; so maybe it is possible to reduce the number of slices in `C` to 2?  I am not certain, I have not checked completely.  I also don't know if it would make much difference to the running time or memory usage of the code anyway---just an idea I had.

src/mlpack/core/optimizers/cmaes/cmaes_impl.hpp
```diff
@@ -0,0 +1,260 @@
...
+        pStep.slice(idx(j)) = arma::randn(iterate.n_rows, iterate.n_cols) *
+            covLower;
+      }
+
+      pPosition.slice(idx(j)) = mPosition.slice(idx0) + sigma(idx0) *
+          pStep.slice(idx(j));
+
+      pObjective(idx(j)) = function.Evaluate(pPosition.slice(idx(j)), 0);
+      for (size_t f = 1; f < numFunctions; ++f)
+        pObjective(idx(j)) += function.Evaluate(pPosition.slice(idx(j)), f);
```
rcurtin: I guess when we have batch support we can change this to a single call to `Evaluate()` with batch size `numFunctions`.

src/mlpack/core/optimizers/cmaes/cmaes_impl.hpp
```diff
@@ -0,0 +1,260 @@
...
+        C.slice(idx1) = C.slice(idx1) + cmu * w(j) *
+            pStep.slice(idx(j)).t() * pStep.slice(idx(j));
+      }
+    }
+
+    arma::eig_sym(eigval, eigvec, C.slice(idx1));
+    if (arma::any(eigval < 0))
+    {
+      eigval = arma::max(eigval, eigvalZero);
+      C.slice(idx1) = eigvec * diagmat(eigval) / eigvec;
```
rcurtin: Instead of `/ eigvec` should this be `* eigvec.t()`? My impression is that you are reassembling this matrix after removing all negative eigenvalues.

Another option, since `eig_sym()` gives back eigenvalues in order, is to find the first negative eigenvalue and do something like

```
C.slice(idx1) = eigvec.cols(0, firstNegativeEigenvalue - 1) * arma::diagmat(eigval.subvec(0, firstNegativeEigenvalue - 1) * eigvec.cols(0, firstNegativeEigenvalue - 1).t();
```

But if the calculation you are doing is different than what I thought, maybe my comments do not apply. :)

src/mlpack/core/optimizers/cmaes/cmaes_impl.hpp
zoq: Right, I used an incorrect starting point, so that I needed to convert doubles to uword. With the correct starting point I can directly use ``arma::linspace<arma::vec>``.

src/mlpack/core/optimizers/cmaes/cmaes_impl.hpp
zoq: I'll change that, no need to allocate additional memory.

src/mlpack/core/optimizers/cmaes/cmaes_impl.hpp
zoq: Agreed, not only here but in the CNE code as well.

src/mlpack/core/optimizers/cmaes/cmaes_impl.hpp
zoq: Nice, catch and I like the truncated operation, I'll push the changes soon.

zoq:
> I imagine one of the big problems with CMAES is that it wants to look over all data before taking a step. I wonder if it would find good solutions more quickly if some kind of batch size was added, so that instead of calling Evaluate() on all functions, it was only done on a small batch (to give a fast but possibly inaccurate gradient estimate)

Really like the idea, this could acclerate the whole method, depending on the choosen batch size. What I'd like to do is to open an issue; to take a closer look into the idea once #1137 and this PR is merged.

zoq:
Closing in favour of #1148.

PR #1129, opened 2017-09-29
IQN: An Incremental Quasi-Newton Method with Local Superlinear Convergence Rate

zoq:
Implementation of the incremental Quasi-Newton with local superlinear convergence rate as proposed by A. Mokhtari et al. in "IQN: An Incremental Quasi-Newton Method with Local Superlinear Convergence Rate".


zoq:
I'd like to merge this after #1137 is merged. I like the idea to reduce the memory footprint and opening an issue for that sounds like a good idea to me.

zoq:
Closed in favour of #1154.

PR #1130, opened 2017-09-29
SPALeRA Stochastic Gradient Descent

zoq:
Implementation of SPALeRA Stochastic Gradient Descent as proposed by A. Schoenauer-Sebag in "Stochastic Gradient Descent: Going As Fast As Possible But Not Faster".

### DIFF
--- a/src/mlpack/core/optimizers/spalera_sgd/spalera_sgd_impl.hpp
+++ b/src/mlpack/core/optimizers/spalera_sgd/spalera_sgd_impl.hpp
@@ -1,0 +1,208 @@
+/**
+ * @file spalera_sgd_impl.hpp
+ * @author Marcus Edel
+ *
+ * Implementation of SPALeRA SGD.
+ *
+ * mlpack is free software; you may redistribute it and/or modify it under the
+ * terms of the 3-clause BSD license.  You should have received a copy of the
+ * 3-clause BSD license along with mlpack.  If not, see
+ * http://www.opensource.org/licenses/BSD-3-Clause for more information.
+ */
+#ifndef MLPACK_CORE_OPTIMIZERS_SPALERA_SGD_SPALERA_SGD_IMPL_HPP
+#define MLPACK_CORE_OPTIMIZERS_SPALERA_SGD_SPALERA_SGD_IMPL_HPP
+
+// In case it hasn't been included yet.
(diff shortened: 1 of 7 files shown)


src/mlpack/core/optimizers/spalera_sgd/spalera_stepsize.hpp
```diff
@@ -0,0 +1,227 @@
...
+   */
+  bool Update(const double stepSize,
+              const double objective,
+              const size_t batchSize,
+              const size_t numFunctions,
+              arma::mat& iterate,
+              const arma::mat& gradient)
+  {
+    // The ratio of mini-batch size to training set size; needed for the
+    // Page-Hinkley relaxed objetive computations.
```
rcurtin: Small typo: `objective` :)

src/mlpack/core/optimizers/spalera_sgd/spalera_sgd_impl.hpp
```diff
@@ -0,0 +1,208 @@
...
+      }
+
+      // Ensure the last batch size isn't zero, to avoid division by zero before
+      // updating.
+      bool learningRateCheck = true;
+      if (lastBatchSize > 0)
+      {
+        // Now update the iterate.
+        learningRateCheck = updatePolicy.Update(stepSize, currentObjective,
+            lastBatchSize, numFunctions, iterate, gradient);
```
rcurtin: If I am not mistaken, everything "special" about SPALeRA is contained in the `SPALeRAStepSize` class.  So could we refactor `SGD` to add a `DecayPolicy` type and then use `SGD<VanillaUpdate, SPALeRAStepSize>` as the SPALeRA optimizer?

zoq:
> I guess the only question is how we handle this and the batch support PR (but I think that question exists for each of the four optimizer PRs you opened recently).

I'd like to merge this afterwards, I'll adjust the PR's once #1137 is merged, which I guess, should be soon.

src/mlpack/core/optimizers/spalera_sgd/spalera_sgd_impl.hpp
zoq: I thought about that, and discarded that idea, because the current policies just take ``iterate``, ``stepSize`` and ``gradient`` as parameter.

PR #1131, opened 2017-09-29
Big-batch Stochastic Gradient Descent (BBSGD)

zoq:
Implementation of Big-batch Stochastic Gradient Descent (BBSGD) as described in: "Big Batch SGD: Automated Inference using Adaptive Batch Sizes" by Soham De et al.

### DIFF
--- a/src/mlpack/core/optimizers/bigbatch_sgd/bigbatch_sgd_impl.hpp
+++ b/src/mlpack/core/optimizers/bigbatch_sgd/bigbatch_sgd_impl.hpp
@@ -41,27 +41,15 @@ template<typename DecomposableFunctionType>
 double BigBatchSGD<UpdatePolicyType>::Optimize(
     DecomposableFunctionType& function, arma::mat& iterate)
 {
-  // Find the number of functions.
+  // Find the number of functions to use.
   const size_t numFunctions = function.NumFunctions();
-  size_t numBatches = numFunctions / batchSize;
-  if (numFunctions % batchSize != 0)
-    ++numBatches; // Capture last few.
-
-  // Batch visitation order.
-  arma::Col<size_t> visitationOrder = arma::linspace<arma::Col<size_t>>(0,
-      (numBatches - 1), numBatches);
-
-  if (shuffle)
-    visitationOrder = arma::shuffle(visitationOrder);
 
   // To keep track of where we are and how things are going.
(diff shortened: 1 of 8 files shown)


Rajiv2605:
@zoq This looks interesting... Can I work on this?

src/mlpack/core/optimizers/bigbatch_sgd/bigbatch_sgd_impl.hpp
```diff
@@ -0,0 +1,250 @@
...
+      gB = std::pow(arma::norm(gradient / batchSize, 2), 2.0);
+
+      // Compute sample variance.
+      vB = std::pow(arma::norm(funcGradients[0] - (gradient / batchSize),
+          2.0), 2.0);
+      for (size_t j = 1; j < batchSize; ++j)
+      {
+        vB += std::pow(arma::norm(funcGradients[j] - (gradient / batchSize),
+            2.0), 2.0);
+      }
```
rcurtin: I think it might be possible to avoid holding the full vector `funcGradients` by using an incremental variance calculation, something like this:

https://en.wikipedia.org/wiki/Algorithms_for_calculating_variance#Online_algorithm

However I am not totally sure.  But for large batch sizes, I think the whole algorithm could be much faster if all the memory for `funcGradients` was not needed.  I think for large batch sizes and large networks being optimized, that might actually cause the system to run out of RAM in some cases.

src/mlpack/core/optimizers/bigbatch_sgd/bigbatch_sgd_impl.hpp
```diff
@@ -0,0 +1,250 @@
...
+      {
+        // Increase batch size at least by one.
+        size_t batchOffset = batchDelta * batchSize;
+        if (batchOffset <= 0)
+          batchOffset = 1;
+
+        funcGradients.resize(batchSize + batchOffset);
+
+        // Generate new batch indices.
+        arma::Col<size_t> batchVisitationOrder;
```
rcurtin: I guess once #1137 is merged, this part can be simplified greatly because we can just call `Shuffle()`.

zoq:
Kinda, looks like the only things that is left is to use the latest ``EvaluateWithGradient`` function. I'll set this on my list.

sourabhvarshney111:
Thanks @zoq. Looking forward to see this merged

PR #1134, opened 2017-10-05
CodeAi fixes: remove dead code, prevent possible null deref

C0deAi:
CodeAi suggest the removal of 3 lines of dead code in timers.cpp and ra_util.cpp.

CodeAi suggests a condition tightening to prevent possible null dereference in cosine_tree.cpp.

### DIFF
--- a/src/mlpack/core/tree/cosine_tree/cosine_tree.cpp
+++ b/src/mlpack/core/tree/cosine_tree/cosine_tree.cpp
@@ -94,7 +94,7 @@ CosineTree::CosineTree(const arma::mat& dataset,
   // Initialize Monte Carlo error estimate for comparison.
   double monteCarloError = root.FrobNormSquared();
 
-  while (monteCarloError > epsilon * root.FrobNormSquared())
+  while (treeQueue.top() && (monteCarloError > epsilon * root.FrobNormSquared()))
   {
     // Pop node from queue with highest projection error.
     CosineTree* currentNode;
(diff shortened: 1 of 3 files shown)


src/mlpack/core/tree/cosine_tree/cosine_tree.cpp
```diff
@@ -94,7 +94,7 @@ CosineTree::CosineTree(const arma::mat& dataset,
   // Initialize Monte Carlo error estimate for comparison.
   double monteCarloError = root.FrobNormSquared();
 
-  while (monteCarloError > epsilon * root.FrobNormSquared())
+  while (treeQueue.top() && (monteCarloError > epsilon * root.FrobNormSquared()))
```
rcurtin: In this case, I believe it shouldn't be possible to dereference a null pointer because when the tree queue is empty, then `monteCarloError` should be 0 (thus terminating the loop).  However, it's not bad to be paranoid so I think it's fine to include the fix.

rcurtin:
I'll handle the failing style and static analysis checks after merge, I don't imagine CodeAi is very good at following the style guide exactly. :)

rcurtin:
I merged this, but I must have mistyped one of the commands during the merge or something because Github isn't recognizing it as merged.  The commits are now 07056406 7a2dae01 bd4eb30d 1fd5f323 plus my fixes for the style and static analysis checks in 8b5765bb.

PR #1135, opened 2017-10-05
Make timers threadsafe by holding one timer for each thread.

### LINKED ISSUE #1133: Calling Train from multiple threads
Hi, 
I have the following code:

```c++ 
std::thread worker_threads[5];
for(int i = 0; i < 5; ++i)
{
  worker_threads[i]=std::thread([](){
          try{
                   mlpack::regression::SoftmaxRegression model(3,3,false);
                   arma::Row<size_t> label({0,1,2});
                   arma::mat data = arma::randu<arma::mat>(3,3);
                   model.Train(data,label,3);
                 }catch(const std::exception& e){
                      std::cerr<< "Multithreaded SoftmaxRegression::Train failed: "<<e.what() ;
                 }
   });
}

  for(int i = 0; i < 5; ++i)
  {
        worker_threads[i].join();
  }
```

This code fails with 

> Multithreaded SoftmaxRegression::Train failed: Timer::Start(): timer 'softmax_regression_optimization' has already been started

As far as I can tell this is due to the fact that Train calls Timers::Start and Timers::Stop, which in turn calls CLI::GetSingleton().timer.StartTimer(name);

Therefore, no matter what, even if my code is supposed to be thread safe (using one object per thread), because of the CLI object this particular usage is not possible. 

Is there another way of using the library by not triggering the timers? Or, alternatively, is it possible to disable those timers? Something like this would make me happy enough:

```c++
void Timer::Start(const std::string& name)
{
#ifndef DISABLE_TIMERS
  CLI::GetSingleton().timer.StartTimer(name);
#endif
}

/**
 * Stop the given timer.
 */
void Timer::Stop(const std::string& name)
{
#ifndef DISABLE_TIMERS
  CLI::GetSingleton().timer.StopTimer(name);
#endif
}
```

Thank you.

rcurtin:
This should fix #1133.  @sgiurgiu: you can try this branch and see if it fixes your problem.  I wrote a simple test to imitate your bug report.

Since `std::this_thread::get_id()` is guaranteed to produce a different result for every thread, I simply made the `Timers` class hold a set of timers for each possible thread, then when the result is printed, each timer across all threads is summed up.  So in the code in #1133, if you printed the value of the timers at the end of the program, it would be the sum of all softmax regression trainings.

### DIFF
--- a/src/mlpack/core/util/timers.hpp
+++ b/src/mlpack/core/util/timers.hpp
@@ -71,13 +71,31 @@ class Timer
    * @param name Name of timer to return value of.
    */
   static std::chrono::microseconds Get(const std::string& name);
+
+  /**
+   * Enable timing of mlpack programs.  Do not run this while timers are
+   * running!
+   */
+  static void EnableTiming();
+
+  /**
+   * Disable timing of mlpack programs.  Do not run this while timers are
+   * running!
+   */
+  static void DisableTiming();
+
+  /**
+   * Stop and reset all running timers.  This removes all knowledge of any
+   * existing timers.
+   */
+  static void ResetAll();
 };
 
 class Timers
 {
  public:
-  //! Nothing to do for the constructor.
-  Timers() { }
+  //! Default to disabled.
+  Timers() : enabled(false) { }
 
   /**
    * Returns a copy of all the timers used via this interface.
@@ -138,6 +156,11 @@ class Timers
   bool GetState(const std::string& timerName,
                 const std::thread::id& threadId = std::thread::id());
 
+  //! Modify whether or not timing is enabled.
+  bool& Enabled() { return enabled; }
+  //! Get whether or not timing is enabled.
+  bool Enabled() const { return enabled; }
+
  private:
   //! A map of all the timers that are being tracked.
   std::map<std::thread::id, std::map<std::string, std::chrono::microseconds>>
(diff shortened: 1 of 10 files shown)


sgiurgiu:
Thank you very much for your fast response.

sgiurgiu:
I've been looking at the patch and I have a few questions. Please correct me if I'm wrong:

The map of timers doesn't seem to be cleared ever. Which means that with a long running application (I'm talking about a web service, where the entire threading issue started, and which could be running for months/years), this map will only grow. Small growth, for sure, but still, more and more memory will be used with no way to clear it. 

Now, adding a "Clear" method would be easy, the problem would be about the concurrent access. std::map is not (as far as I know) thread safe. Therefore, deleting a value from it, while other threads access or modify other values can cause unexpected (even undefined) behaviour. 

I wonder, wouldn't it be easier that instead of having an "std::map<std::thread::id, std::map<std::string, std::chrono::microseconds>>" to just mark "Timers timer" as thread_local in the CLI object? As far as I can tell, according to http://en.cppreference.com/w/cpp/language/storage_duration a thread_local object would be created when a thread begins and destroyed when the thread finishes. Which would free us from any memory growth headaches.

It would be true, yes, that beyond the thread that called Start/Stop those timers would not be available, but I believe that would be an acceptable trade-off. If one cares about the timers, he/she should print them in the thread that generated them, right?

Again, please correct me if my assessment is wrong here.

rcurtin:
@mlpack-jenkins test this

src/mlpack/core/util/timers.hpp
```diff
@@ -15,7 +15,11 @@
 
 #include <map>
```
zoq: Can you add your name to the authors list and also adjust ``MLPACK``?

rcurtin: Good catch, done.

PR #1136, opened 2017-10-06
Making the CLI singleton thread local

sgiurgiu:
My take on the threading issue. This patch would make the CLI object thread local using the boost::thread_specific_ptr class. Downsides:
- pulls Boost Thread as a dependency
- Timers are not available across threads (in the unit test it can be clearly seen that outside the thread that created the them the timers do not exist)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -289,6 +289,7 @@ find_package(Boost 1.49
       program_options
       unit_test_framework
       serialization
+      thread
     REQUIRED
 )
 
--- a/src/mlpack/core/util/cli.cpp
+++ b/src/mlpack/core/util/cli.cpp
@@ -63,11 +63,7 @@ void CLI::StopTimers()
  */
 void CLI::Destroy()
 {
-  if (singleton != NULL)
-  {
-    delete singleton;
-    singleton = NULL; // Reset pointer.
-  }
+  singleton.reset();
 }
 
 void CLI::Add(ParamData&& data)
@@ -140,10 +136,10 @@ bool CLI::HasParam(const std::string& key)
 // Returns the sole instance of this class.
 CLI& CLI::GetSingleton()
 {
-  if (singleton == NULL)
-    singleton = new CLI();
+  if (!singleton.get())
+    singleton.reset(new CLI());
 
-  return *singleton;
+  return *(singleton.get());
 }
 
 /**
--- a/src/mlpack/core/util/cli.hpp
+++ b/src/mlpack/core/util/cli.hpp
@@ -20,6 +20,7 @@
 #include <string>
 
 #include <boost/any.hpp>
+#include <boost/thread/tss.hpp>
 
 #include <mlpack/prereqs.hpp>
 
@@ -309,7 +310,7 @@ class CLI
 
  private:
   //! The singleton itself.
-  static CLI* singleton;
+  static boost::thread_specific_ptr<CLI> singleton;
 
   //! True, if CLI was used to parse command line options.
  public:
--- a/src/mlpack/core/util/singletons.cpp
+++ b/src/mlpack/core/util/singletons.cpp
@@ -32,7 +32,7 @@ using namespace mlpack::util;
   #define BASH_CLEAR ""
 #endif
 
-CLI* CLI::singleton = NULL;
+boost::thread_specific_ptr<CLI> CLI::singleton;
 
 // Only output debugging output if in debug mode.
 #ifdef DEBUG
--- a/src/mlpack/tests/timer_test.cpp
+++ b/src/mlpack/tests/timer_test.cpp
@@ -18,6 +18,8 @@
   #include <windows.h>
 #endif
 
+#include <thread>
+
 #include <mlpack/core.hpp>
 
 #include <boost/test/unit_test.hpp>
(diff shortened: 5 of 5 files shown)


rcurtin:
Thanks for taking the time to put this together.  My only concern is that aggregating the time across timers in different threads is convenient functionality that I don't really want to lose.

I thought also of modifying the other PR to remove thread-specific timers after they are stopped, but there is no guarantee that the timer will stop and so if threads are crashing or exiting without stopping timers, then we have the same memory leak problem.  So I think it is more robust, perhaps, to have the destructor of a thread-local timers object simply add its time to a global timers object.

I think we can remove the boost::thread dependency by just holding a `thread_local` pointer directly instead of using boost's wrappers.

Let me know what you think, and we can proceed from there.  I want to make sure that whatever solution we go with actually solves the problems that you were having to begin with, instead of a halfway solution. :)

sgiurgiu:
The problem with global variables that are mutated in a multithreaded environment is that access to them needs to be protected. The simplest way to protect access to a variable is to use a mutex. The problem with using a mutex is that it becomes a choke point for the multithreaded program. It doesn't transform it into a single threaded one, but overall performance will decrease. As a guideline, mutexes should only be used as a last resort, when all other options are not available.
Using a lock-free structure (map-like) is another possibility, but that will add yet another dependency to the library. And those structures tend to be quite complex and complicated (for good reason). 

So, this seemingly simple problem can balloon into a very complex monster if we are not careful. As a general idea, sharing and mutating data between threads is an extremely dangerous adventure, which should be avoided whenever possible.

Secondly,  marking the CLI pointer as thread_local is not possible, as nobody will ever delete it. One would have to at least use an `std::unique_ptr` for it, to ensure no memory leaks occur. But, `std::unique_ptr` is only available in C++ 11 so that then will become a hard requirement for the library (I am not familiar enough with mlpack to know if that's acceptable). On the other hand, `thread_local` is a C++ 11 only feature also, so this would be a moot point. Therefore, the `boost::thread_specific_ptr` looked like the simplest solution to this problem.
Or, not make the singleton object  a pointer, just a simple value. Then yes, boost::thread is not needed and simply marking it as thread_local would suffice (with the entire C++ 11 requirement caveat).

This, however, doesn't solve the global timers problem.  And global timers would need to have a way to be cleared, to avoid the timers map to grow too large. 

Personally, since I do not need the timers feature, I would be perfectly happy with a simple `#ifndef` flag to disable them. This would solve all problems in the simplest way. And, to be honest, me discovering them was quite a surprise. All other approaches have downsides and increase complexity by quite a bit.

rcurtin:
I agree with what you've written.  This problem is more difficult than I anticipated.

It seems clear to me that the first and most important issue to solve is to disable timers by default.  The timers were only ever made so that detailed timing information could be returned for a user's convenience by the command-line programs (or bindings to other languages).  Unfortunately to give good timing information, this means calling the `Timer::Start()` and `Timer::Stop()` functions from down inside the actual machine learning algorithms themselves.  As you have pointed out, I think you are right that not only your situation but most situations don't need timing by default.  So I have made (on the other PR, #1135) timers disabled by default and if you _do_ want timing in your program, you can call `Timer::EnableTiming()` (and then `Timer::DisableTiming()` if you want to turn it back off).

But to provide the convenient cumulative timers to users even in a multithreaded environment, you are right that the task has a significant amount of extra complexity---and some drawbacks.  For one, it won't (and can't) scale: any reasonable way to do this will need a mutex when any timer is started or stopped.  So if you are running some program with hundreds or more threads, the timers can be a slowdown.

I'd like to avoid the extra boost dependency unless we absolutely need it, because including a Boost header can significantly slow down compilation (depending on the boost library in question) due to the large number of additional headers that will be included recursively.

So I think that, overall, disabling timers by default and using a stupid-but-relatively-simple mutex-based solution for multithreaded situations solves this problem "well enough" for its most important use case: the command-line programs and bindings to other languages where timing information is collected and displayed for a user's convenience.  With timers disabled, nobody will have a moment of surprise when they find some completely unrelated support causing a problem for them like you did---sorry about that!

Let me know what you think.  If you agree that's a reasonable solution that solves your problem satisfactorily, I'd like to go with #1135 instead of this one.  Nonetheless, thank you so much for getting involved at this level and submitting some code!

sgiurgiu:
Thank you very much for not abandoning the problem. I agree with your solution, it is better for everyone if timers are enabled only when needed.
Since you decided to go down the multi-threading rabbit-hole I have a couple of comments about the commit  #1135 :
- Instead of using mutex.lock() and unlock(), use std::lock_guard<std::mutex) _(mutex); For more information see http://en.cppreference.com/w/cpp/thread/lock_guard . 
- Using lock/unlock creates the potential to forget to unlock a mutes, like in StartTimer/StopTimer where when an exception is being thrown the mutex is never unlocked.
- All accesses to those three maps (timers,timerState and timerStartTime) have to be protected. You can't even look at them without the security of a mutex. This includes GetAllTimers,GetTimer,GetState and Reset among others.
- In StopAllTimers the timersState map is being iterated and deleted from at the same time. This can cause problems, 2 of which I can see right now:
  - The entire StopAllTimers method should be protected by the mutex. Now, this poses a problem because we want the StopTimer  method too to be protected by a mutex and you cannot call lock twice. So there are two choices here: Use a `recursive_mutex` that you can call lock twice on or call a different Stoptimer method, a private one, that doesn't do the lock. That method is then called from both StopTimer and StopAllTimers, with those methods taking care of locking the mutex before calling the internal one.
  - The second problem is that the maps are modified (elements erased from it) while being iterated upon. One solution to that is presented here: https://stackoverflow.com/questions/8234779/how-to-remove-from-a-map-while-iterating-it
- Is that enabled flag used anywhere? I see it set, I see it read, but shouldn't there be something like `if(!enabled) return` somewhere in both Start and Stop. If it's there and I just don't see it, please forgive me.
- Shouldn't GetAllTimers() return a copy not a reference? Because you want to protect it with the mutex, but then if it returns a reference ... isn't it all in vain?
- Maybe make the enabled member atomic? For example std::atomic_bool enabled;? In practice on an x86 platform it shouldn't make a difference (no idea about other platforms) but ... it would communicate the intent, right?

These are all small little nitpicks, sorry for being so intrusive. Shouldn't be much of a problem to fix.

rcurtin:
I appreciate the feedback even if it is nitpicky. :)  This is my first time working with the C++ builtin multithreading support, so your comments are definitely something I can learn from.

The enabling not making a difference is kind of facepalmy.  I fixed that and added a test, which is what I should have done in the first place.  I also handled all the other issues.  Definitely your comments pointed out a lot of issues that I had not fully thought through.  Thanks!

PR #1141, opened 2017-10-18
Add FunctionType API checks at compile time

shikharbhardwaj:
This PR adds static checking of FunctionType API at compile time. Currently, I have added a short testcase and made changes to the `AdaDelta` optimizer.

A few minor things : 
1. POD types passed by value are treated equally with or without the `const` qualifier. So `void Foo(size_t)` and `void Foo(const size_t)` are treated as same function prototypes.
2. `const` qualified member functions are enforced. So I think we should modify the `FunctionType` policy documentation to indicate that the functions must be `const` qualified.

I'll change other optimizers and add more method forms if these changes are fine.

### DIFF
--- a/src/mlpack/core/optimizers/fw/frank_wolfe_impl.hpp
+++ b/src/mlpack/core/optimizers/fw/frank_wolfe_impl.hpp
@@ -42,6 +42,11 @@ template<typename FunctionType>
 double FrankWolfe<LinearConstrSolverType, UpdateRuleType>::
 Optimize(FunctionType& function, arma::mat& iterate)
 {
+  static_assert(mlpack::static_checks::CheckEvaluate<FunctionType>::value,
+      "The FunctionType does not have a correct definition of Evaluate.");
+  static_assert(mlpack::static_checks::CheckGradient<FunctionType>::value,
+      "The FunctionType does not have a correct definition of Gradient.");
+
   // To keep track of the function value.
   double currentObjective = function.Evaluate(iterate);
 
--- a/src/mlpack/core/util/functiontype_method_forms.hpp
+++ b/src/mlpack/core/util/functiontype_method_forms.hpp
@@ -159,9 +159,66 @@ template <typename FunctionType>
 struct CheckGradientConstraint
 {
   const static bool value =
-    HasGradientConstraint<FunctionType, GradientConstraintFormConst>::value
-    || HasGradientConstraint<FunctionType,
-    GradientConstraintFormNonConst>::value;
+    HasGradientConstraint<FunctionType, GradientConstraintFormConst>::value ||
+    HasGradientConstraint<FunctionType, GradientConstraintFormNonConst>::value;
+};
+
+/*
+ * Definitions of method forms and checks for the SparseFunctionType API.
+ */
+HAS_METHOD_FORM(Gradient, HasSparseGradient);
+
+template <typename Class, typename...Ts>
+using SparseGradientFormConst = void(Class::*)(const arma::mat&, size_t,
+    arma::sp_mat&) const;
+
+template <typename Class, typename...Ts>
+using SparseGradientFormNonConst = void(Class::*)(const arma::mat&, size_t,
+    arma::sp_mat&);
+
+template <typename FunctionType>
+struct CheckSparseGradient
+{
+  const static bool value =
+    HasSparseGradient<FunctionType, SparseGradientFormConst>::value ||
+    HasSparseGradient<FunctionType, SparseGradientFormNonConst>::value;
+};
+
+/**
+ * Definitions of method forms and checks for the ResolvableFunctionType API.
+ */
+HAS_METHOD_FORM(NumFeatures, HasNumFeatures);
+
+template <typename Class, typename...Ts>
+using NumFeaturesFormConst = size_t(Class::*)() const;
+
+template <typename Class, typename...Ts>
+using NumFeaturesFormNonConst = size_t(Class::*)();
+
+template<typename FunctionType>
+struct CheckNumFeatures
+{
+  const static bool value =
+    HasNumFeatures<FunctionType, NumFeaturesFormConst>::value ||
+    HasNumFeatures<FunctionType, NumFeaturesFormNonConst>::value;
+};
+
+HAS_METHOD_FORM(PartialGradient, HasPartialGradient);
+
+template <typename Class, typename...Ts>
+using PartialGradientFormConst = void(Class::*)(const arma::mat&, size_t,
+    arma::sp_mat&) const;
+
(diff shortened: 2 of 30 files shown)


rcurtin:
Nice, thanks for taking the time to do this. :)

It should be possible to catch both const and non-const overloads with something like this:

```
HAS_METHOD_FORM(NumFunctions, HasNumFunctions);

template<typename Class, typename ...Ts>
using NumFunctionsFormConst = size_t(Class::*)(Ts...) const;
template<typename Class, typename ...Ts>
using NumFunctionsFormNonConst = size_t(Class::*)(Ts...);

template<typename Class, typename ...Ts>
struct NumFunctionsForm
{
  const static bool value = NumFunctionsFormConst<Class, Ts...>::value || NumFunctionsFormNonConst<Class, Ts...>::value;
};
```

Probably that can be simplified a little bit and the using declarations put inside the struct (or not used at all).  I think that is probably a better solution than enforcing a const requirement on some functions.

Otherwise everything looks good to me.  I think you can put the checks only in sgd_impl.hpp and save a lot of time, since all of the SGD-related optimizers call SGD::Optimize().

src/mlpack/core/util/sfinae_utility.hpp
```diff
@@ -16,6 +16,7 @@
 #define MLPACK_CORE_SFINAE_UTILITY
 
 #include <type_traits>
+#include <cstring>
```
rcurtin: I think this is included in some other places, is it needed here?

shikharbhardwaj: This file uses `size_t` but does not define it anywhere. I guess it would be fine (as `size_t` will probably be defined somewhere above it during a compile), I just changed it because I saw a bunch of errors while editing this file.

shikharbhardwaj:
Thanks for looking at this Ryan.

I made the changes you suggested. I also discovered that these MethodForms allow the method form to match any overload with the same starting parameters(so a method form for the normal Evaluate function also matched the decomposable Evaluate) : 

```
template<typename Class, typename ...Ts>
using FlexibleForm = size_t(Class::*)(Type1, Type2, Ts...);
```

So I changed the MethodForms to

```
template<typename Class, typename ...Ts>
using StrictForm = size_t(Class::*)(Type1, Type2);
```

But FFN and RNN include a decomposable `Evaluate` with a default parameter (`deterministic`), which is not matched by the strict method form.

I think this can be fixed by :

1. Renaming the current `Evaluate` function with a deterministic parameter to something else, and provide the right overload (the Optimizer only needs to be called while training, with deterministic = true).
2. Use the Flexible method form.

rcurtin:
Hm, I am fine either way here---I think flexible is just fine since RNN and FFN have a special `Evaluate()` overload.  I would also be fine with refactoring RNN and FFN to add an overload without the deterministic parameter, then using the `StrictForm`.  @zoq, what do you think?

zoq:
I'm fine with the flexible version and with adding an overload for ``Evaluate``.

src/mlpack/core/util/functiontype_method_forms.hpp
```diff
@@ -0,0 +1,227 @@
+/**
+ * @file functiontype_form.hpp
```
zoq: Looks like the file name changed.

src/mlpack/core/util/functiontype_method_forms.hpp
```diff
@@ -0,0 +1,227 @@
...
+using NumFunctionsFormConst = size_t(Class::*)() const;
+
+template <typename Class, typename...Ts>
+using NumFunctionsFormNonConst = size_t(Class::*)();
+
+template<typename FunctionType>
+struct CheckNumFunctions
+{
+  const static bool value =
+    HasNumFunctions<FunctionType, NumFunctionsFormConst>::value ||
```
zoq: The line should be tabbed twice; for more information take a look at: https://github.com/mlpack/mlpack/wiki/DesignGuidelines#line-length-and-wrapping.

src/mlpack/core/util/functiontype_method_forms.hpp
```diff
@@ -0,0 +1,227 @@
...
+
+template <typename Class, typename...Ts>
+using DecomposableEvaluateFormNonConst = double(Class::*)(const arma::mat&,
+    const size_t);
+
+template<typename FunctionType>
+struct CheckDecomposableEvaluate
+{
+  const static bool value =
+    HasDecomposableEvaluate<FunctionType, DecomposableEvaluateFormConst>::value
```
zoq: See my comment for ``CheckNumFunctions``. Perhaps, that way we can prettify the expression here to improve the readability.

src/mlpack/tests/functiontype_method_forms_test.cpp
```diff
@@ -0,0 +1,184 @@
+/**
+ * @file functiontype_method_forms.cpp
```
zoq: Looks like the file name changed.

src/mlpack/core/optimizers/fw/frank_wolfe_impl.hpp
```diff
@@ -42,6 +42,11 @@ template<typename FunctionType>
 double FrankWolfe<LinearConstrSolverType, UpdateRuleType>::
 Optimize(FunctionType& function, arma::mat& iterate)
 {
+  static_assert(mlpack::static_checks::CheckEvaluate<FunctionType>::value,
+      "The FunctionType does not have a correct definition of Evaluate.");
```
zoq: I'm wondering if it makes sense to improve the error message, and give some additional hints e.g. about the correct function definition. This is somehow related to: https://github.com/mlpack/mlpack/issues/891, let me know what you think, also we don't have to do it now.

shikharbhardwaj: Sure, I'll add the correct definition and maybe a reference to the `FunctionType` API documentation(?)
The documentation needs some updating, I guess I'll do that in a separate PR. :)

rcurtin: This is a minor comment, but I do think maybe we should use `Evaluate()` instead of `Evaluate` throughout to make it clear that it is a method that's required.  If we want to make the error messages more complex, maybe it is worthwhile to add some method to `CheckEvaluate<>` so that the error message is consistent across checks.

src/mlpack/core/util/sfinae_utility.hpp
rcurtin: Ah, fair, should we use `<cstdlib>` instead?  `size_t` should be directly defined in that header.

rcurtin:
Closing in favor of #1252.

PR #1142, opened 2017-10-21
Swish Activation Function

ivmarkp:
This PR adds Swish activation function as proposed by Prajit Ramachandran, Barret Zoph and Quoc V. Le in [Swish: A self-gated activation function](https://arxiv.org/pdf/1710.05941.pdf).

### DIFF
--- a/src/mlpack/methods/ann/activation_functions/swish_function.hpp
+++ b/src/mlpack/methods/ann/activation_functions/swish_function.hpp
@@ -64,7 +64,8 @@ class SwishFunction
    */
   static double Deriv(const double y)
   {
-    return y / (1 + std::exp(-y)) + (1 - y / (1 + std::exp(-y))) / (1 + std::exp(-y));
+    return y / (1 + std::exp(-y)) + (1 - y / (1 + std::exp(-y))) /
+                                             (1 + std::exp(-y));
   }
 
   /**
@@ -76,7 +77,8 @@ class SwishFunction
   template<typename InputVecType, typename OutputVecType>
   static void Deriv(const InputVecType& y, OutputVecType& x)
   {
-    x = y / (1 + arma::exp(-y)) + (1 - y / (1 + arma::exp(-y))) / (1 + arma::exp(-y));
+    x = y / (1 + arma::exp(-y)) + (1 - y / (1 + arma::exp(-y))) /
+                                           (1 + arma::exp(-y));
   }
 }; // class SwishFunction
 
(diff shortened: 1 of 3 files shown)


src/mlpack/methods/ann/activation_functions/swish_function.hpp
```diff
@@ -0,0 +1,88 @@
...
+   * @param x Input data.
+   * @param y The resulting output activation.
+   */
+  template<typename InputVecType, typename OutputVecType>
+  static void Fn(const InputVecType& x, OutputVecType& y)
+  {
+    y = x;
+
+    for (size_t i = 0; i < x.n_elem; i++)
+      y(i) = Fn(x(i));
```
zoq: Do you think you could use armadillo expressions to write the function for a matrix type (e.g. by using ``arma::exp``)? This way it could be parallelised.

src/mlpack/tests/activation_functions_test.cpp
```diff
@@ -484,4 +485,22 @@ BOOST_AUTO_TEST_CASE(PReLUFunctionTest)
   CheckPReLUGradientCorrect(activationData, desiredGradient);
 }
 
+/**
+ * Basic test of the swish function.
+ */
+BOOST_AUTO_TEST_CASE(SwishFunctionTest)
+{
+  const arma::colvec desiredActivations("-0.238405 3.07466 4.45055 \
```
zoq: Can you comment on where this values come from?

src/mlpack/methods/ann/activation_functions/swish_function.hpp
ivmarkp: Yes, sure thing. Will add the function for computing swish function using a matrix as input.

src/mlpack/tests/activation_functions_test.cpp
ivmarkp: I calculated these swish function values using the python interpreter using the column vector `activationData` values as input data.

zoq: Okay, sounds good, would be great if you could add a comment something like this: https://github.com/mlpack/mlpack/blob/master/src/mlpack/tests/logistic_regression_test.cpp#L41

ivmarkp: Added the comment with a new fixup commit.

zoq:
Thanks for the great contribution.

PR #1143, opened 2017-10-21
Reinclude 'total_time' timer.

zoq:
Add a ``total_time`` timer that is added by default for each mlpack program.


zoq:
I think starting a ``total_time`` timer for each command-line program is just fine, I'll adjust the wording and make clear that we start the timer for each cli program.

PR #1153, opened 2017-11-24
Implementation of SPSA optimizer

Rajiv2605:
As of now, I have just created the basic files necessary to implement the optimizer for the sake of creating the PR... I'll push the code in the subsequent commits :v:


zoq:
Great, looking forward to take a look over the implementation.

Rajiv2605:
Hey @zoq, sorry for the delay... I have started to work on the optimizer. It looks a bit difficult to implement as only material available is its research paper.

Rajiv2605:
@zoq Compadagrad looks a bit challenging... How about I start working on the following algos: Momentum, Adamax? I'll continue with Compadagrad after that. What do you think?

zoq:
Momentum and Adamax is already implemented but if you like you could look into: Simultaneous perturbation stochastic approximation (SPSA), that might be a better method to jump in. Let me know what you think.

Rajiv2605:
@zoq I'll look into it! Thanks!

Rajiv2605:
@zoq Hey, I have a small doubt... Why are all implementation and update algorithms written in `.hpp` and not in `.cpp`? For example: `ada_grad_update.hpp`. Shouldn't header files contain declarations and definitions only leaving the main logic implementations to the `.cpp` files?

zoq:
There is an exception; template classes need to have the method definitions inside the header file, see https://isocpp.org/wiki/faq/templates#templates-defn-vs-decl for an explanation.

Rajiv2605:
@zoq Thanks!

Rajiv2605:
@zoq As of now, I am working on SPSA algorithm. The only part left is passing the function to the `Update()` function for calculating the gradient. I found a template called `DecomposableFunctiontype` which is used to pass functions. But, I am not able to understand the parameters used in it. Like, its description talks about the no. of functions and the no. of points in the dataset at the same time. How can there be more than one function? Does it have anything to do with the points in the dataset?  It would be great if you could tell me how to use it. I will be able to complete the SPSA optimiser if I find a way to pass the function.

zoq:
Some functions (separable functions) can be expressed as a combination of functions (number of function) e.g. ``x * sin(y) + x^2`` is a product of ``f(x) = x``, ``g(y) = sin(y)`` and ``z(x) = x^2`` , in this case NumFunctions returns 2. The function parameter are x and y, in our case we often use the name ``iterate``. For a simple example take a look at: https://arxiv.org/pdf/1711.06581.pdf page 5.

In your case you don't need to worry about ``NumFunctions`` or ``Gradient``, you can just use ``Evaluate(iterate)``; keep in mind one step is just a gradient approximation, no need to use ``function.Gradient()``.

So what you end up with is something like this:

```c++
void Optimize(function, iterate)
{

  for(size_t i = 0; i < iterations; ++i)
  {
    // iterate are the function parameter, let's see if we can find some parameter that minimize function.

    // Let's modifiy the iterate parameter and see if that is a step in the right direction.
    iterate += 0.1;

    double loss = function.Evalaute(iterate);
  }
}
```

Rajiv2605:
@zoq Thanks! I'll try it out.

Rajiv2605:
@zoq Thanks! I'll try it out.

zoq:
Nice, before I'll take a closer look at the code, can you write a simple test for the optimizer: one option would be to replicate the sgd tests: https://github.com/mlpack/mlpack/blob/master/src/mlpack/tests/sgd_test.cpp

Rajiv2605:
Sure :+1:  I'll look into it.

Rajiv2605:
@zoq I have complemented the SPSA optimizer along with the tests(this is the first time I've written a test so please tell me if it's decent :p). I had to completely revamp the SPSA codes to build it perfectly with the tests(It will be visible when you would compare it with the older commits). For suiting the optimizer, I have created a new `Evaluate()` function and I'm skeptical about its correctness. Please suggest any improvements and corrections which have to be done 😄 . I'll implement the changes ASAP 👍

Rajiv2605:
@zoq The spsa test is failing 😞 but I'm not able to figure out why... Is there any change which I'm supposed to do to pass the test?

src/mlpack/core/optimizers/spsa/spsa_update.hpp
```diff
@@ -0,0 +1,153 @@
...
+    c(c),
+    max_iter(maxIterations)
+  {
+    Initialize();
+  }
+
+  void Initialize()
+  {
+    A = 0.1*a;
+    ak = a/pow((max_iter + 1 + A), alpha);
```
zoq: hm, we should update deltas in every iteration, see the pseudo code on page 490.

src/mlpack/core/optimizers/spsa/spsa_update.hpp
```diff
@@ -0,0 +1,153 @@
...
+
+  template<typename DecomposableFunctionType>
+  void Update(arma::mat& iterate,
+          DecomposableFunctionType& function)
+  {
+    const int s = iterate.n_elem;
+
+    for (size_t i = 0; i < max_iter; i++)
+    {
+      sp_vector = arma::conv_to<arma::vec>::from(
```
zoq: We should parameterized the the number of elements we change per iterations. Introduce ``p``, and maybe set it to ``iterate.n_elem / 2`` per default, we should test if we can find some reasonable value that works across different functions.

src/mlpack/core/optimizers/spsa/spsa_update.hpp
```diff
@@ -0,0 +1,153 @@
...
+  float Gamma() const { return gamma; }
+
+  float& Gamma() { return gamma; }
+
+  //! Get the maximum number of iterations (0 indicates no limit).
+  size_t MaxIterations() const { return max_iter; }
+  //! Modify the maximum number of iterations (0 indicates no limit).
+  size_t& MaxIterations() { return max_iter; }
+
+  float Gradient_scaling_parameter(const int& choice) const
```
zoq: Not sure, is this used somewhere? Maybe I missed something.

Rajiv2605: None of these functions are used anywhere. I used defined them in case anyone wanted to monitor the variations of those parameters in their code.

src/mlpack/core/optimizers/spsa/spsa.hpp
```diff
@@ -0,0 +1,102 @@
...
+class SPSA
+{
+ public:
+  SPSA(const float& alpha = 0.602,
+       const float& gamma = 0.101,
+       const float& a = 1e-6,
+       const float& c = 0.01,
+       const size_t& maxIterations = 100000)
+  {
+    spsaUpdate = new SPSAUpdate(alpha, gamma, a, c,
```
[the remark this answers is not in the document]
Rajiv2605: Yeah, I thought of it... But then I followed the structure of other optimizers.

src/mlpack/core/optimizers/problems/spsa_test_function.cpp
```diff
@@ -0,0 +1,40 @@
+/**
+ * @file sgd_test_function.cpp
```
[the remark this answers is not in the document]
Rajiv2605: I forgot to change the comments part... 😅 I created a new test function because I wanted `Evaluate()` to return a vector whereas `Evaluate()` from `SGDTestFunction` returns only a single value.

src/mlpack/core/optimizers/spsa/spsa_update.hpp
Rajiv2605: Oh yeah 😅 Just read the paper again... I'll change it 👍

src/mlpack/core/optimizers/spsa/spsa_update.hpp
zoq: I see, do you mind to remove those functions, to be consistent with the rest of the codebase. I think we can provide something similar that works across all optimization algortihmis. A cheap and hacky solution would be to redefine, ``private`` as ``public`` and access the internal parameter that way.

src/mlpack/core/optimizers/spsa/spsa.hpp
```diff
@@ -0,0 +1,102 @@
...
+ *   pages   = {482--492},
+ *   year    = {1998}
+ * }
+ * @endcode
+ *
+ */
+class SPSA
+{
+ public:
+  SPSA(const float& alpha = 0.602,
```
zoq: Can you add a comment for each parameter?

src/mlpack/core/optimizers/spsa/spsa_update.hpp
Rajiv2605: ohk... I'll do it 👍

src/mlpack/core/optimizers/problems/spsa_test_function.cpp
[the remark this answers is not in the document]
Rajiv2605: @zoq I used `SPSATestFunction` to implement a `Evaluate()` function which can process and return a vector which was not available in `SGDTestFunction`. What do you think?

src/mlpack/core/optimizers/spsa/spsa_update.hpp
Rajiv2605: @zoq `iterate` is actually a feature vector/ point right? Am I right?

Rajiv2605:
@zoq I have pushed with the changes you asked to make(except SPSATestFunction change and changing the number of elements; I have stated the reasons as a reply to your comments). I don't know why the Static Code Analysis Checks are failing.

zoq:
There was an issue in the build step, let me restart the Static Code Analysis Checks job.

zoq:

src/mlpack/core/optimizers/spsa/spsa.hpp
```diff
@@ -0,0 +1,98 @@
...
+
+    for (size_t i = 0; i < max_iter; i++)
+    {
+      sp_vector = arma::conv_to<arma::vec>::from(
+                  randi(s, arma::distr_param(-1, 1)));
+      ak = a/pow((max_iter + 1 + A), alpha);
+      ck = c/pow((max_iter + 1), gamma);
+      arma::vec f_plus = function.Evaluate(iterate + ck*sp_vector, s);
+      arma::vec f_minus = function.Evaluate(iterate - ck*sp_vector, s);
+      arma::mat gradient = ((f_plus - f_minus)/2)*ck*sp_vector.i();
```
zoq: I'm not sure I get the inverse step, can you elaborate on that? I think it should be:

```c++
arma::mat gradient = (f_plus - f_minus) % (1 / (2 * ck * sp_vector));
```

Maybe I missed something?

src/mlpack/core/optimizers/problems/spsa_test_function.hpp
```diff
@@ -0,0 +1,39 @@
...
+//! larger step size may be required to optimize it in a reasonable number of
+//! iterations.
+class SPSATestFunction
+{
+ public:
+  //! Get the starting point.
+  arma::mat GetInitialPoint() const { return arma::mat("6; -45.6; 6.2"); }
+
+  //! Evaluate a function.
+  arma::vec Evaluate(const arma::mat& coordinates, const int& size) const;
```
rcurtin: Why does this return a vector?  We should just be returning the objective value as a `double`, I think.

src/mlpack/core/optimizers/spsa/spsa.hpp
```diff
@@ -0,0 +1,98 @@
...
+    const int s = iterate.n_elem;
+
+    for (size_t i = 0; i < max_iter; i++)
+    {
+      sp_vector = arma::conv_to<arma::vec>::from(
+                  randi(s, arma::distr_param(-1, 1)));
+      ak = a/pow((max_iter + 1 + A), alpha);
+      ck = c/pow((max_iter + 1), gamma);
+      arma::vec f_plus = function.Evaluate(iterate + ck*sp_vector, s);
+      arma::vec f_minus = function.Evaluate(iterate - ck*sp_vector, s);
```
rcurtin: I think we can save some memory by using `iterate` in-place, instead of making two copies when we call `Evaluate()`:

```
iterate += ck * sp_vector;
f_plus = function.Evaluate(iterate);
iterate -= 2 * ck * sp_vector;
f_minus = function.Evaluate(iterate);
// We'll have to add ck * sp_vector back to iterate later...
```

src/mlpack/core/optimizers/spsa/spsa.hpp
```diff
@@ -0,0 +1,98 @@
...
+  void Optimize(DecomposableFunctionType& function, arma::mat& iterate)
+  {
+    const int s = iterate.n_elem;
+
+    for (size_t i = 0; i < max_iter; i++)
+    {
+      sp_vector = arma::conv_to<arma::vec>::from(
+                  randi(s, arma::distr_param(-1, 1)));
+      ak = a/pow((max_iter + 1 + A), alpha);
+      ck = c/pow((max_iter + 1), gamma);
```
rcurtin: A minor comment: we should use `std::pow()` here or it may not work with all compilers.

src/mlpack/core/optimizers/spsa/spsa.hpp
Rajiv2605: I wasn't sure how to use vector in the denominator... Hence I took it to the numerator and raised to power -1, which I think I considered as inverse. Looks like a very rookie math mistake... 😞 Is it fine if I have a vector in the denominator as I thought vector division doesn't exist?

Rajiv2605:
@zoq @rcurtin I think I have made all the changes you suggested 👍 Please review and let me know if anything else has to be changed :)

rcurtin:
It looks like the optimizer is failing.  To debug it, I'd suggest adding some output each iteration for debugging that will print what is going on.  It looks like the method is converging to `nan`, so definitely something is not right in the implementation.

If you can trace down the `nan` (probably a division by 0?) then I think maybe this will work right and we can proceed, adding some more tests and documentation.

Rajiv2605:
@zoq As you said in the irc, I ran the command `g++ foo.cpp -o foo -std=c++11 /path/to/lib/libmlpack.so -I/path/to/mlpack/include/ -larmadillo -lboost_serialization -lboost_program_options` with the right files and paths. 

I got the following error:
```
/usr/lib/gcc/x86_64-linux-gnu/5/../../../x86_64-linux-gnu/crt1.o: In function `_start':
(.text+0x20): undefined reference to `main'
/tmp/ccUlXyyX.o: In function `SPSATest::SimpleSPSATestFunction::test_method()':
spsa_test.cpp:(.text+0x4cd): undefined reference to `boost::unit_test::unit_test_log_t::set_checkpoint(boost::unit_test::basic_cstring<char const>, unsigned long, boost::unit_test::basic_cstring<char const>)'
spsa_test.cpp:(.text+0x62a): undefined reference to `boost::unit_test::unit_test_log_t::set_checkpoint(boost::unit_test::basic_cstring<char const>, unsigned long, boost::unit_test::basic_cstring<char const>)'
spsa_test.cpp:(.text+0x757): undefined reference to `boost::unit_test::unit_test_log_t::set_checkpoint(boost::unit_test::basic_cstring<char const>, unsigned long, boost::unit_test::basic_cstring<char const>)'
spsa_test.cpp:(.text+0x884): undefined reference to `boost::unit_test::unit_test_log_t::set_checkpoint(boost::unit_test::basic_cstring<char const>, unsigned long, boost::unit_test::basic_cstring<char const>)'
/tmp/ccUlXyyX.o: In function `SPSATest::GeneralizedRosenbrockTest::test_method()':
spsa_test.cpp:(.text+0xbf8): undefined reference to `boost::unit_test::unit_test_log_t::set_checkpoint(boost::unit_test::basic_cstring<char const>, unsigned long, boost::unit_test::basic_cstring<char const>)'
/tmp/ccUlXyyX.o:spsa_test.cpp:(.text+0xd2e): more undefined references to `boost::unit_test::unit_test_log_t::set_checkpoint(boost::unit_test::basic_cstring<char const>, unsigned long, boost::unit_test::basic_cstring<char const>)' follow
/tmp/ccUlXyyX.o: In function `__static_initialization_and_destruction_0(int, int)':
spsa_test.cpp:(.text+0x178c): undefined reference to `boost::unit_test::ut_detail::auto_test_unit_registrar::auto_test_unit_registrar(boost::unit_test::basic_cstring<char const>)'
spsa_test.cpp:(.text+0x17e7): undefined reference to `boost::unit_test::ut_detail::auto_test_unit_registrar::auto_test_unit_registrar(boost::unit_test::test_case*, unsigned long)'
spsa_test.cpp:(.text+0x184e): undefined reference to `boost::unit_test::ut_detail::auto_test_unit_registrar::auto_test_unit_registrar(boost::unit_test::test_case*, unsigned long)'
spsa_test.cpp:(.text+0x1869): undefined reference to `boost::unit_test::ut_detail::auto_test_unit_registrar::auto_test_unit_registrar(int)'
/tmp/ccUlXyyX.o: In function `boost::unit_test::unit_test_log_t::unit_test_log_t()':
spsa_test.cpp:(.text._ZN5boost9unit_test15unit_test_log_tC2Ev[_ZN5boost9unit_test15unit_test_log_tC5Ev]+0x25): undefined reference to `vtable for boost::unit_test::unit_test_log_t'
/tmp/ccUlXyyX.o: In function `boost::unit_test::make_test_case(boost::unit_test::callback0<boost::unit_test::ut_detail::unused> const&, boost::unit_test::basic_cstring<char const>)':
spsa_test.cpp:(.text._ZN5boost9unit_test14make_test_caseERKNS0_9callback0INS0_9ut_detail6unusedEEENS0_13basic_cstringIKcEE[_ZN5boost9unit_test14make_test_caseERKNS0_9callback0INS0_9ut_detail6unusedEEENS0_13basic_cstringIKcEE]+0x42): undefined reference to `boost::unit_test::ut_detail::normalize_test_case_name[abi:cxx11](boost::unit_test::basic_cstring<char const>)'
spsa_test.cpp:(.text._ZN5boost9unit_test14make_test_caseERKNS0_9callback0INS0_9ut_detail6unusedEEENS0_13basic_cstringIKcEE[_ZN5boost9unit_test14make_test_caseERKNS0_9callback0INS0_9ut_detail6unusedEEENS0_13basic_cstringIKcEE]+0x7c): undefined reference to `boost::unit_test::test_case::test_case(boost::unit_test::basic_cstring<char const>, boost::unit_test::callback0<boost::unit_test::ut_detail::unused> const&)'
/tmp/ccUlXyyX.o: In function `boost::unit_test::unit_test_log_t::~unit_test_log_t()':
spsa_test.cpp:(.text._ZN5boost9unit_test15unit_test_log_tD2Ev[_ZN5boost9unit_test15unit_test_log_tD5Ev]+0xd): undefined reference to `vtable for boost::unit_test::unit_test_log_t'
/tmp/ccUlXyyX.o: In function `bool boost::test_tools::tt_detail::check_frwd<boost::test_tools::check_is_small_t, double, double>(boost::test_tools::check_is_small_t, boost::unit_test::lazy_ostream const&, boost::unit_test::basic_cstring<char const>, unsigned long, boost::test_tools::tt_detail::tool_level, boost::test_tools::tt_detail::check_type, double const&, char const*, double const&, char const*)':
spsa_test.cpp:(.text._ZN5boost10test_tools9tt_detail10check_frwdINS0_16check_is_small_tEddEEbT_RKNS_9unit_test12lazy_ostreamENS5_13basic_cstringIKcEEmNS1_10tool_levelENS1_10check_typeERKT0_PSA_RKT1_SH_[_ZN5boost10test_tools9tt_detail10check_frwdINS0_16check_is_small_tEddEEbT_RKNS_9unit_test12lazy_ostreamENS5_13basic_cstringIKcEEmNS1_10tool_levelENS1_10check_typeERKT0_PSA_RKT1_SH_]+0x176): undefined reference to `boost::test_tools::tt_detail::check_impl(boost::test_tools::predicate_result const&, boost::unit_test::lazy_ostream const&, boost::unit_test::basic_cstring<char const>, unsigned long, boost::test_tools::tt_detail::tool_level, boost::test_tools::tt_detail::check_type, unsigned long, ...)'
/tmp/ccUlXyyX.o: In function `bool boost::test_tools::tt_detail::check_frwd<boost::test_tools::check_is_close_t, double, double, boost::test_tools::percent_tolerance_t<double> >(boost::test_tools::check_is_close_t, boost::unit_test::lazy_ostream const&, boost::unit_test::basic_cstring<char const>, unsigned long, boost::test_tools::tt_detail::tool_level, boost::test_tools::tt_detail::check_type, double const&, char const*, double const&, char const*, boost::test_tools::percent_tolerance_t<double> const&, char const*)':
spsa_test.cpp:(.text._ZN5boost10test_tools9tt_detail10check_frwdINS0_16check_is_close_tEddNS0_19percent_tolerance_tIdEEEEbT_RKNS_9unit_test12lazy_ostreamENS7_13basic_cstringIKcEEmNS1_10tool_levelENS1_10check_typeERKT0_PSC_RKT1_SJ_RKT2_SJ_[_ZN5boost10test_tools9tt_detail10check_frwdINS0_16check_is_close_tEddNS0_19percent_tolerance_tIdEEEEbT_RKNS_9unit_test12lazy_ostreamENS7_13basic_cstringIKcEEmNS1_10tool_levelENS1_10check_typeERKT0_PSC_RKT1_SJ_RKT2_SJ_]+0x1d7): undefined reference to `boost::test_tools::tt_detail::check_impl(boost::test_tools::predicate_result const&, boost::unit_test::lazy_ostream const&, boost::unit_test::basic_cstring<char const>, unsigned long, boost::test_tools::tt_detail::tool_level, boost::test_tools::tt_detail::check_type, unsigned long, ...)'
collect2: error: ld returned 1 exit status
```

zoq:
How does ``spsa_test.cpp`` look like, does it have a proper main?

src/mlpack/core/optimizers/spsa/spsa.hpp
```diff
@@ -0,0 +1,101 @@
...
+    const int s = iterate.n_elem;
+
+    for (size_t i = 0; i < max_iter; i++)
+    {
+      sp_vector = arma::conv_to<arma::vec>::from(
+                  randi(s, arma::distr_param(-1, 1)));
+      ak = a/std::pow((max_iter + 1 + A), alpha);
+      ck = c/std::pow((max_iter + 1), gamma);
+      iterate += ck * sp_vector;
+      double f_plus = function.Evaluate(iterate + ck*sp_vector, 0, s);
```
zoq: You should remove ``+ ck*sp_vector`` inside the function call, since you already updated the parameter.

src/mlpack/core/optimizers/spsa/spsa.hpp
```diff
@@ -0,0 +1,101 @@
...
+      iterate -= ck * sp_vector;
```
zoq: This should be ``iterate -= 2 * ck * sp_vector`` if you remove ``- ck*sp_vector`` from the ``Evaluate`` function.

Rajiv2605:
@zoq Oh yeah, it didn't have main() 😅 But even after adding main(), I'm not able to call the boost functions.
```

spsa_test.cpp:63:6: error: ISO C++ forbids declaration of ‘main’ with no type [-fpermissive]
 main()
      ^
In file included from /usr/include/boost/test/unit_test.hpp:20:0,
                 from spsa_test.cpp:17:
spsa_test.cpp: In function ‘int SPSATest::main()’:
spsa_test.cpp:65:3: error: a function-definition is not allowed here before ‘{’ token
   BOOST_AUTO_TEST_CASE(SimpleSPSATestFunction);
   ^
spsa_test.cpp:65:3: error: qualified-id in declaration before ‘(’ token
   BOOST_AUTO_TEST_CASE(SimpleSPSATestFunction);
   ^
spsa_test.cpp:66:3: error: a function-definition is not allowed here before ‘{’ token
   BOOST_AUTO_TEST_CASE(GeneralizedRosenbrockTest);
   ^
spsa_test.cpp:66:3: error: qualified-id in declaration before ‘(’ token
   BOOST_AUTO_TEST_CASE(GeneralizedRosenbrockTest);
```

zoq:
You should remove the boost specific test code, e.g. it should look like:

```c++
#include <mlpack/core.hpp>
#include <mlpack/core/optimizers/spsa/spsa.hpp>
#include <mlpack/core/optimizers/problems/sgd_test_function.hpp>
#include <mlpack/core/optimizers/problems/generalized_rosenbrock_function.hpp>

using namespace std;
using namespace arma;
using namespace mlpack;
using namespace mlpack::optimization;
using namespace mlpack::optimization::test;

int main()
{

  SGDTestFunction f;

  SPSA optimiser(0.602, 0.101, 1e-6, 0.01, 100000);

  arma::mat coordinates = f.GetInitialPoint();
  double result = optimiser.Optimize(f, coordinates);
}
```

src/mlpack/core/optimizers/spsa/spsa.hpp
Rajiv2605: Oh sorry! That was a silly mistake from my side... I'll make sure that such mistakes don't happen again.

Rajiv2605:
@zoq I tried the above code and compiled it using the command:
```
 g++ spsa_test.cpp -o spsa -std=c++11 ../../../build/lib/libmlpack.so -I ../../../build/include/ -l armadillo -l boost_serialization -l boost_program_options
```
I also tried changing `libmlpack.so` to `libmlpack.so.2`. Even then, when I print `ldd spsa`, it shows the following:
```
linux-vdso.so.1 =>  (0x00007fff7b363000)
libmlpack.so.2 => not found
libarmadillo.so.7 => /usr/lib/x86_64-linux-gnu/libarmadillo.so.7 (0x00007f565cdf3000)
libstdc++.so.6 => /usr/lib/x86_64-linux-gnu/libstdc++.so.6 (0x00007f565ca71000)
libm.so.6 => /lib/x86_64-linux-gnu/libm.so.6 (0x00007f565c768000)
libgcc_s.so.1 => /lib/x86_64-linux-gnu/libgcc_s.so.1 (0x00007f565c552000)
libc.so.6 => /lib/x86_64-linux-gnu/libc.so.6 (0x00007f565c188000)
/lib64/ld-linux-x86-64.so.2 (0x00007f565cff5000)
```

zoq:
Okay, just to make sure what is the output of ``ls ../../../build/lib/``?

Rajiv2605:
I get the following output:
```
libmlpack.so  libmlpack.so.2  libmlpack.so.2.0  pkgconfig
```

zoq:
Okay, can you put the example code into the lib folder and try again, just like to make sure the lib is in your search path.

Rajiv2605:
@zoq I copied it to the lib folder and used the following command:
```
g++ spsa_test.cpp -o spsa -std=c++11 libmlpack.so -I ../include/ -l armadillo -l boost_serialization -l boost_program_options
```

Rajiv2605:
@zoq I also used `sudo` to link `libmlpack.so`. It still says that it's not found.

rcurtin:
I think you need to set `LD_LIBRARY_PATH`... try something like this:

```
$ export LD_LIBRARY_PATH=/path/to/mlpack/build/lib/
$ ldd spsa_test
```

(make sure you put the right path in place when you set `LD_LIBRARY_PATH`; it should just be the directory that contains `libmlpack.so`)

I think that should fix your issue; let me know if we can clarify anything.

Rajiv2605:
@rcurtin Worked! I'll now try to print and find out the reason for the test to fail 👍

Rajiv2605:
@zoq @rcurtin I have fixed the `nan` issue! It was caused by the zeros from the randomly generated vector `sp_vector` in the denominator. But, for 100000 iterations, the function value reduced from 3672.31 to 3577.93. Also, the value of the function seems to drop by 0.1 after 11 iterations exactly every time. That 11 changed to a larger constant when I changed the number of  iterations. Isn't the convergence too slow? Is the pattern of 11 iterations suggesting something? 🤔

rcurtin:
I think that this should definitely take less than 100k iterations to converge for the `SGDTestFunction`, so maybe there is a bug in the code.

Keep in mind also, SPSA is a random technique, so you will get the same results every run unless you set the random seed.  So if you want to try a number of different runs with different random vectors each time, you can add `math::RandomSeed(std::time(NULL));` to the top of your program for testing, then each time you run it the random seed will be different and you'll get different results.

Rajiv2605:
@rcurtin For the vector, I'm using the following method:
```
      sp_vector = arma::conv_to<arma::vec>::from(
                  randi(s, arma::distr_param(-1, 1)));
```
For removing the zero, I do the following:
```
bool one = false;
      for (size_t j = 0; j < 3; j++)
        if (sp_vector[j]==0)
        {
          if(one)
          {
            sp_vector[j] = -1;
            one = false;
          }
          else
          {
            sp_vector[j] = 1;
            one = true;
          }
        }
```
I use the `bool` to alternatively replace the zeros with -1 and 1.

Can I do anything better instead of the above methods?

zoq:
You can add a really small value: ``const double epsilon = 1e-8; sp_vector  = ... + epsilon``.

Rajiv2605:
@zoq After adding `espilon`, the final function value blew up to `e+103`!

rcurtin:
I do think that you have to remove the zeros from `sp_vector` and only use 1s or -1s.  When you do this, does the optimization perform any better?

Rajiv2605:
@rcurtin The cause of `nan` was those zeros. So, I had to remove them.

zoq:
Ah I missed that you are using: ``randi(s, arma::distr_param(-1, 1)``, what about:

```c++
randi(s, arma::distr_param(0, 1) * 2 - 1;
```

Rajiv2605:
@zoq Nice! I have changed that line... Even then when I print, I get the following results:
```
Initial coordinates: (6, -45.6, 6.2)
Initial Function value: 3672.31
Final Result after SPSA: 3579.9
```
Should I push the latest code? 
Meanwhile, I'll try debugging step by step all over again and see if I can find anything. I think the main problem is in the generation of `sp_vector`. I got better results when I generated it with random values between `-100 to 100` instead of `-1 to -1` which I am using currently(I saw a Python implementation which used -1 to 1, hence I decided to go with it)... I'll check if there is a better choice of distribution.

zoq:
If you push the latest updates, I'll take a closer look at the code once I get a chance.

src/mlpack/core/optimizers/spsa/spsa.hpp
```diff
@@ -0,0 +1,109 @@
...
+  {
+    const int s = iterate.n_elem;
+    const double epsilon = 1e-8;
+
+    for (long long int i = 0; i < max_iter; i++)
+    {      
+      sp_vector = arma::conv_to<arma::vec>::from(
+                  randi(s, arma::distr_param(0, 1)))*2 - 1;
+      
+      ak = a/std::pow((max_iter + 1 + A), alpha);
```
zoq: This should be ``ak = a/std::pow(i + 1 + A, alpha);`` same for ``ck``.

zoq:
The defaults you used are not necessarily good for the given problem, so it is suggested that the values used be tailored to the task at hand. Take a look at secion IIIB of the paper, which offers some suggestions for picking the coefficients to achive some good results.

E.g with: ``a = 0.16, A = 0.001 * max_iter and c = 0.3`` I was able to solve: ``StyblinskiTangFunction f(8);`` and ``BoothFunction f;``.


Also, I introduced another loop for the gradient approximation, to make it more stable:

```c++
arma::mat gradient = arma::zeros(iterate.n_rows, iterate.n_cols);

for (size_t i = 0; i < max_iter; i++)
{

  ak = a/std::pow((i + 1 + A), alpha);
  ck = c/std::pow((i + 1), gamma);


  gradient.zeros();
  for (size_t b = 0; b < 10; ++b)
  {
    sp_vector = arma::conv_to<arma::vec>::from(
              randi(s, arma::distr_param(0, 1)))*2 - 1;

    iterate += ck * sp_vector;
    double f_plus = function.Evaluate(iterate, 0, s);

    iterate -= 2 * ck * sp_vector;
    double f_minus = function.Evaluate(iterate, 0, s);
    iterate += ck * sp_vector;

    gradient += (f_plus - f_minus) / (2 * ck * sp_vector);
  }
}

gradient /= 10;
```

Rajiv2605:
@zoq I had read that paper before, but the parameters given there were for a particular example. I had found a python implementation, hence I went with its default parameters as I thought it was a general case.

I made the changes you suggested but still it doesn't seem to pass the `Rosenbrock function`. I get the following results:
```
Initial coordinates: (6, -45.6, 6.2)
Final function value: 3456.14
```
Maybe, as you have suggested, I'll have to try different parameter values until I get the converging value.

rcurtin:
@Rajiv2605: have you tried with a number of different random seeds?  That can give a better idea of how robust the implementation is.  It's possible that the result you are seeing is an unlucky artifact of one specific random seed.

Rajiv2605:
@rcurtin I'll try out an exhaustive grid search to find out the set of optimal parameters for the data.

Rajiv2605:
@rcurtin @zoq I made a set of 10 observations for the following parameters:
`alpha:0.1, gamma:0.102, a:0.16, c:0.3`
For the same above parameters, I get the following values when I run the code 10 times:
```
-1.49889e-05
1.79812e+75
1.79812e+75
-1.49948e-05
1.79812e+75
-1.49258e-05
-1.49218e-05
1.79812e+75
-1.51336e-05
-1.48727e-05
```
The convergence is affected by the random values generated in the `sp_vector`.

rcurtin:
The objective function value of something as large as `1e+75` is unexpected.  You might consider taking a look at the objective value each iteration to see if the step it is taking is bad or something like this; it certainly looks to me like something may be wrong.

zoq:
Instead of testing against the generalized rosenbrock function, do you think you could test against logistic regression and the sphere function?

Rajiv2605:
@zoq The sphere functiioin is converging! I get the final result as: `4.87053e-34`

Rajiv2605:
@rcurtin @zoq  The code is converging for the sphere function perfectly for all the combinations of the number of functions and random initial points. But it is blowing up to the range of `1e105` for Rosenbrock function. I'll try a few other test functions and let you know.

rcurtin:
To me the real key here is '''how often''' is it converging and how often is it not?  For this I would suggest trying with different random seeds, and getting some statistics like the percentage of the time it converges for the sphere function (including for each of the different configurations) and for the Rosenbrock function.

I really think the failure rate here should be pretty low---it's hard for me to see why SPSA would explode.  The gradient estimate will be noisy, but over many iterations it should still be able to find some optima.

Rajiv2605:
@rcurtin @zoq  The sphere function converges each and every time. I used the following code:
```
    int conv = 0;
    for(size_t i = 10; i<=50; i++)
    {
      GeneralizedRosenbrockFunction g(i);
      SphereFunction h(i);
      SPSA optimiser(0.1, 0.102, 0.16, 0.3, 100000);
      arma::mat coordinates = f.GetInitialPoint();
      double result = optimiser.Optimize(h, coordinates);
      cout << result << endl;

      conv += (result<0.00001)?1:0;
    }
     cout << "Convergence rate for sphere function: " << conv/40 << endl;
```

I get the following results:
```
1.48176e-33
3.85186e-34
3.64596e-34
4.58543e-34
5.7566e-34
4.30885e-34
1.92593e-34
3.34465e-34
5.4668e-34
9.83013e-34
5.77779e-34
3.85186e-34
5.16441e-34
3.44294e-34
5.4668e-34
5.4668e-34
0
5.79102e-34
3.85186e-34
3.85186e-34
5.7566e-34
0
9.36242e-34
5.88852e-34
5.84263e-34
5.4668e-34
9.77135e-34
8.62487e-34
9.56258e-34
5.16441e-34
5.4668e-34
3.85186e-34
3.44437e-34
5.16441e-34
4.87062e-34
5.76808e-34
5.7566e-34
3.34465e-34
0
5.16441e-34
3.34465e-34
Convergence rate for sphere function: 1
```

zoq:
Okay, looks like the ``SphereFunction`` is fine, do you think we can find some stable parameter for the SimpleSPSA function?

Rajiv2605:
@zoq I think we can use the defaults used in the paper. For the sphere function, when I used the default parameters, I got the following values:
```
3.5832e-29
5.95375e-29
5.15806e-29
4.93109e-29
5.23269e-29
4.99956e-29
5.05558e-29
5.06215e-29
5.53465e-29
4.853e-29
4.95614e-29
5.00693e-29
5.70236e-29
4.92177e-29
4.94044e-29
4.87087e-29
5.01422e-29
5.0043e-29
4.90963e-29
5.30989e-29
5.40629e-29
5.71824e-29
5.00021e-29
4.83245e-29
4.861e-29
5.18315e-29
4.90137e-29
4.95337e-29
5.35572e-29
5.08953e-29
5.04981e-29
5.01872e-29
5.3888e-29
5.05105e-29
5.77949e-29
5.14828e-29
5.78113e-29
5.31659e-29
5.049e-29
6.34776e-29
5.99371e-29
Convergence rate for sphere function: 1
```
For better convergence, we can tune the parameters anytime :+1:

zoq:
I think the parameter for the Sphere function are okay, the remaining issue is the SimpleSPSA function, see the travis build.

Rajiv2605:
@zoq @rcurtin I had to remove `BOOST_REQUIRE_CLOSE` as the function converged at `0` and the boost function has `0` as its singularity point as it uses Knuth function. I replaced it with `BOOST_CHECK_SMALL` but it still is failing. What test should I exactly run?(This suggestion was provided in: https://stackoverflow.com/questions/29196130/boost-check-close-fails-even-if-the-two-values-are-below-the-threshold)

I get the following error:
```
/home/travis/build/mlpack/mlpack/src/mlpack/tests/spsa_test.cpp(44): error in "SimpleSPSATestFunction": absolute value of result{1.1555579666323415e-33} exceeds 0
```
The co-ordinates also converged very close to `0` but when I use `BOOST_REQUIRE_SMALL(coordinate[0], 0)`, I get the following error:
```
/home/travis/build/mlpack/mlpack/src/mlpack/tests/spsa_test.cpp(45): fatal error in "SimpleSPSATestFunction": absolute value of coordinates[0]{0} exceeds 0
```
Is there any other BOOST function which I can use for testing in this case?

Rajiv2605:
@zoq @rcurtin It has passed the build test! I changed the tolerance value and the convergence to a non-zero point and it worked.

zoq:
Looks good, can you look into the Static Code Analysis issues?

Rajiv2605:
@zoq I have fixed the Static Code Analysis issues!

src/mlpack/tests/spsa_test.cpp
```diff
@@ -0,0 +1,50 @@
...
+{
+  for (size_t i = 10; i <= 50; i++)
+  {
+      SphereFunction h(i);
+      GeneralizedRosenbrockFunction f(i);
+
+      SPSA optimiser(0.1, 0.102, 0.16,
+                     0.3, 100000);
+
+      arma::mat coordinates = f.GetInitialPoint();
```
zoq: Right now, you are using the initial coordinates of the Rosenbrock function for the Sphere function. Can we use the Sphere function for the same?

src/mlpack/tests/spsa_test.cpp
```diff
@@ -0,0 +1,50 @@
...
+
+BOOST_AUTO_TEST_CASE(SimpleSPSATestFunction)
+{
+  for (size_t i = 10; i <= 50; i++)
+  {
+      SphereFunction h(i);
+      GeneralizedRosenbrockFunction f(i);
+
+      SPSA optimiser(0.1, 0.102, 0.16,
+                     0.3, 100000);
```
zoq: When wrapping a line, the next line should be tabbed twice from where the previous line began.  See https://github.com/mlpack/mlpack/wiki/DesignGuidelines#line-length-and-wrapping for more information.

src/mlpack/tests/spsa_test.cpp
```diff
@@ -0,0 +1,50 @@
...
+using namespace std;
+using namespace arma;
+using namespace mlpack;
+using namespace mlpack::optimization;
+using namespace mlpack::optimization::test;
+using namespace mlpack::regression;
+
+BOOST_AUTO_TEST_SUITE(SPSATest);
+
+BOOST_AUTO_TEST_CASE(SimpleSPSATestFunction)
```
zoq: Can you add a simple comment about what this test does?

src/mlpack/tests/spsa_test.cpp
```diff
@@ -0,0 +1,50 @@
...
+ * terms of the 3-clause BSD license.  You should have received a copy of the
+ * 3-clause BSD license along with mlpack.  If not, see
+ * http://www.opensource.org/licenses/BSD-3-Clause for more information.
+ */
+#include <mlpack/core.hpp>
+#include <mlpack/core/optimizers/spsa/spsa.hpp>
+#include <mlpack/core/optimizers/problems/sgd_test_function.hpp>
+#include <mlpack/core/optimizers/problems/generalized_rosenbrock_function.hpp>
+#include <mlpack/core/optimizers/problems/sphere_function.hpp>
+#include <mlpack/methods/logistic_regression/logistic_regression_function.hpp>
```
zoq: Since you include the logistic regression function here, do you think we could add a test for this function as well?

src/mlpack/core/optimizers/spsa/spsa.hpp
```diff
@@ -0,0 +1,117 @@
...
+  }
+
+  template<typename DecomposableFunctionType>
+  double Optimize(DecomposableFunctionType& function, arma::mat& iterate)
+  {
+    const int s = iterate.n_elem;
+    const double epsilon = 1e-8;
+    arma::mat gradient = arma::zeros(iterate.n_rows, iterate.n_cols);
+
+    for (long long int i = 0; i < max_iter; i++)
```
zoq: We use ``size_t`` instead of ``long long int``, so it would be great if we could be consistent with the rest of the codebase.

src/mlpack/core/optimizers/spsa/spsa.hpp
```diff
@@ -0,0 +1,117 @@
...
+    ck(0)
+  {
+    // Nothing to do.
+  }
+
+  template<typename DecomposableFunctionType>
+  double Optimize(DecomposableFunctionType& function, arma::mat& iterate)
+  {
+    const int s = iterate.n_elem;
+    const double epsilon = 1e-8;
```
zoq: Looks like ``epsilon`` isn't used, maybe I missed something?

src/mlpack/core/optimizers/spsa/spsa.hpp
```diff
@@ -0,0 +1,117 @@
...
+    ak(0),
+    ck(0)
+  {
+    // Nothing to do.
+  }
+
+  template<typename DecomposableFunctionType>
+  double Optimize(DecomposableFunctionType& function, arma::mat& iterate)
+  {
+    const int s = iterate.n_elem;
```
zoq: I guess, we could directly use ``iterate.n_elem``, no need to create an alias, what do you think?

src/mlpack/core/optimizers/spsa/spsa.hpp
```diff
@@ -0,0 +1,117 @@
...
+      ak = a/std::pow((i + 1 + A), alpha);
+      ck = c/std::pow((i + 1), gamma);
+
+      gradient.zeros();
+      for (size_t b = 0; b < 10; b++)
+      {
+        sp_vector = arma::conv_to<arma::vec>::from(
+                    randi(s, arma::distr_param(0, 1)))*2 - 1;
+
+        iterate += ck * sp_vector;
```
zoq: Can you use camel casing for all names, see https://github.com/mlpack/mlpack/wiki/DesignGuidelines#naming-conventions for more information.

src/mlpack/core/optimizers/spsa/spsa.hpp
```diff
@@ -0,0 +1,117 @@
...
+    for (long long int i = 0; i < max_iter; i++)
+    {
+      ak = a/std::pow((i + 1 + A), alpha);
+      ck = c/std::pow((i + 1), gamma);
+
+      gradient.zeros();
+      for (size_t b = 0; b < 10; b++)
+      {
+        sp_vector = arma::conv_to<arma::vec>::from(
+                    randi(s, arma::distr_param(0, 1)))*2 - 1;
```
zoq: Can you use a space between operations?

src/mlpack/core/optimizers/spsa/spsa.hpp
```diff
@@ -0,0 +1,117 @@
...
+    const double epsilon = 1e-8;
+    arma::mat gradient = arma::zeros(iterate.n_rows, iterate.n_cols);
+
+    for (long long int i = 0; i < max_iter; i++)
+    {
+      ak = a/std::pow((i + 1 + A), alpha);
+      ck = c/std::pow((i + 1), gamma);
+
+      gradient.zeros();
+      for (size_t b = 0; b < 10; b++)
```
zoq: I think ``b`` should be configurable by the user, what do you think?

src/mlpack/core/optimizers/spsa/spsa.hpp
```diff
@@ -0,0 +1,117 @@
...
+ */
+class SPSA
+{
+ public:
+  // Specifies the maximum number of iterations.
+  long long int max_iter;
+  arma::vec sp_vector;
+
+  // Control the amount of gradient update.
+  float a;
```
zoq: Can you use ``double`` to be consistent with the rest of the codebase?

src/mlpack/core/optimizers/spsa/spsa.hpp
```diff
@@ -0,0 +1,117 @@
...
+  float A;
+
+  // Non-negative co-efficients controlling the optimizer.
+  float alpha;
```
zoq: Usually we provide separated functions to modify the method parameter, see https://github.com/mlpack/mlpack/blob/master/src/mlpack/core/optimizers/sgd/sgd.hpp for an example.

src/mlpack/core/optimizers/spsa/spsa.hpp
Rajiv2605: Users are getting to control `maxIter`, the loop outside `b`. So, I thought that it would be enought to control the optimizer. What do you say?

zoq: I guess, since this is somewhat similar to the ``batchSize``parameter, we could avoid any confusions if we use the same construct here, also ``10`` might work for the sphere function, but it's not necessary good for other cases.

src/mlpack/tests/spsa_test.cpp
Rajiv2605: Yeah sure!

Rajiv2605:
@zoq Sorry I didn't reply to you in the IRC on 10th April regarding the [mlpack_test issue](https://imagebin.ca/v/3xucl1iet744). I use Ubuntu 16.04 LTS with kernel version: 4.13.0-38-generic

Rajiv2605:
@zoq Travis-CI build is producing weird results... :thinking:

zoq:
Let me restart the travis build.

Rajiv2605:
This is my CMake output:
```
-- cotire 1.7.10 loaded.
-- Armadillo libraries: /usr/lib/x86_64-linux-gnu/libarmadillo.so
-- Boost version: 1.58.0
-- Found the following Boost libraries:
--   program_options
--   unit_test_framework
--   serialization
-- Updated gitversion.hpp.
-- Regenerating arma_config.hpp.
-- Found Python: /usr/bin/python
CMake Warning at CMake/FindPythonModule.cmake:44 (message):
  Unsuitable version of Python module Cython (0.24 or greater required).
Call Stack (most recent call first): 
 src/mlpack/bindings/python/CMakeLists.txt:31 (find_python_module)


-- Could NOT find PY_Cython (missing: PY_CYTHON) 
-- Cython not found; not building Python bindings.
-- Configuring done
-- Generating done
-- Build files have been written to: /home/rajiv/Programming/OSF/mlpack/build
```
Armadillo version being used:
```
ARMA version: 7.950.1 (Northern Banana Republic)
```

zoq:
I guess, you installed lapack via the package manger, can you post ``arma_config.hpp``?

Rajiv2605:
Is this the file are you talking about?
Path: `/src/mlpack/core/util/arma_config.hpp'
Content:
```
/**
 * @file arma_config.hpp
 *
 * This is an autogenerated file which contains the configuration of Armadillo
 * at the time mlpack was built.  If you modify anything in here by hand, your
 * warranty is void, your house may catch fire, and we're not going to call the
 * police when your program segfaults so hard that robbers come to your house
 * and take everything you own.  If you do decide, against better judgment, to
 * modify anything at all in this file, and you are reporting a bug, be
 * absolutely certain to mention that you've done something stupid in this file
 * first.
 *
 * In short: don't touch this file.
 */
#ifndef MLPACK_CORE_UTIL_ARMA_CONFIG_HPP
#define MLPACK_CORE_UTIL_ARMA_CONFIG_HPP

#define MLPACK_ARMA_64BIT_WORD

#endif
```

Rajiv2605:
Yes, I installed LAPACK using the package manager

Rajiv2605:
@zoq The 2/3 builds have passed. The reason for the 3rd build to fail looks a bit odd as it says the cause of build to fail is incompatible matrix dimensions. But, if that is the case, the other two builds should have also failed... :thinking: Another error in it is that a particular `test.csv` file is not found.

zoq:
The third build does build with `´DEBUG=ON``, means the bounds check is enabled, so I would say there is something wrong with the matrix multiplication. Also, to temporary fix your build issue, does it work if you comment the failing tests (build step) in ``https://github.com/mlpack/mlpack/blob/master/src/mlpack/tests/CMakeLists.txt``?

Rajiv2605:
@zoq The reason the 3rd build is failing because of the Logistic Regression test in SPSA. But it is the exact same as that of the Adagrad test.

rcurtin:
It may be useful to use a debugger like gdb to find the issue.  You could have gdb catch the exception (`catch throw`) and then you can check the backtrace and inspect the variables.

Rajiv2605:
@zoq @rcurtin I'll try out your suggestions :+1:

zoq:
No worries, happy to get back to this one.

zoq:
Moved to: https://github.com/mlpack/ensmallen/pull/44

@Rajiv2605 let me know what you think.

PR #1159, opened 2017-12-07
SVRG, SARAH, Katyusha

zoq:
Implementation of:

- SVRG as proposed as proposed by R. Johnson et al. in "Accelerating Stochastic Gradient Descent using
Predictive Variance Reduction".

- Barzilai-Borwein SVRG as proposed by C. Tan et al. in "Barzilai-Borwein Step Size for Stochastic Gradient Descent".

- SARAH as proposed by L. Nguyen et al. in "SARAH: A Novel Method for Machine Learning Problems Using Stochastic Recursive Gradient".

- Katyusha as proposed by Z. Allen-Zhu in "Katyusha: The First Direct Acceleration of Stochastic Gradient Methods".


src/mlpack/core/optimizers/katyusha/katyusha_impl.hpp
```diff
@@ -0,0 +1,188 @@
...
+    {
+      // Find the effective batch size (the last batch may be smaller).
+      effectiveBatchSize = std::min(batchSize, numFunctions - f);
+
+      function.Gradient(iterate, f, gradient, effectiveBatchSize);
+      fullGradient += gradient;
+
+      f += effectiveBatchSize;
+    }
+    fullGradient /= (double) numFunctions;
```
rcurtin: It looks like in the algorithm, how often the full gradient is computed is different across the paper and I think the only requirement is that m is O(n).  Should we add a way for the user to specify how many epochs should pass before they want to calculate the gradient?  I remember sitting through a tutorial on this and I believe the intuition was that you can choose to trade off between calculating the gradient fully more often, or possibly taking more SGD steps and calculating the gradient less often.

src/mlpack/core/optimizers/katyusha/katyusha_impl.hpp
```diff
@@ -0,0 +1,188 @@
...
+    // To keep track of where we are and how things are going.
+    double bb = 0;
+    size_t iteration = 0;
+
+    for (size_t f = 0; f < numFunctions; /* incrementing done manually */)
+    {
+      // Find the effective batch size (the last batch may be smaller).
+      effectiveBatchSize = std::min(batchSize, numFunctions - f);
+
+      iterate = tau * z + 0.5 * iterate0 + y * tau1;
```
rcurtin: I think maybe I have misunderstood some algebra you have done here to simplify; shouldn't the last term be `(1.0 - tau1 - tau2) * y?  I'm looking at line 15 of `Katyusha1` in the paper.

src/mlpack/core/optimizers/katyusha/katyusha_impl.hpp
```diff
@@ -0,0 +1,188 @@
...
+
+      // Calculate variance reduced gradient.
+      function.Gradient(iterate, f, gradient, effectiveBatchSize);
+      function.Gradient(iterate0, f, gradient0, effectiveBatchSize);
+
+      const arma::mat z0 = z;
+      z -= stepSize * (fullGradient + (gradient - gradient0) /
+        (double) batchSize);
```
rcurtin: Hmmm, I am not sure on this one but it might be possible to avoid an extra allocation and copy here (depending on how Armadillo handles the `z -= ...` line:

```
arma::mat zNew = z - stepSize * (fullGradient + (gradient - gradient0) / (double) batchSize);
y = iterate + (zNew - z) * tau;
z = std::move(zNew);
```

But I am not sure if it is actually faster (or if this is a bottleneck in the algorithm), so I dunno if it is worth trying.

src/mlpack/core/optimizers/katyusha/katyusha_impl.hpp
```diff
@@ -0,0 +1,188 @@
...
+      w += y * std::pow(tau2, (double) iteration);
+
+      bb += pow(tau2, iteration);
+
+      f += effectiveBatchSize;
+
+      iteration++;
+    }
+
+    iterate = w / (double) bb;
```
rcurtin: Hmm, I didn't spend much time thinking about it but I don't know how line 22 of `Katyusha1` turns into this.  I think maybe you did something clever here to reduce the computation?

src/mlpack/core/optimizers/svrg/barzilai_borwein.hpp
```diff
@@ -0,0 +1,109 @@
...
+ *   booktitle = {Advances in Neural Information Processing Systems 29},
+ *   editor    = {D. D. Lee and M. Sugiyama and U. V. Luxburg and I. Guyon
+ *                and R. Garnett},
+ *   pages     = {685--693},
+ *   year      = {2016},
+ *   publisher = {Curran Associates, Inc.}
+ * }
+ * @endcode
+ */
+class BarzilaiBorwein
```
rcurtin: Do you think we should call it `BarzilaiBorweinDecay` for clarity?  I'm fine with either way, just a suggestion.

src/mlpack/core/optimizers/CMakeLists.txt
```diff
@@ -17,6 +17,7 @@ set(DIRS
   sdp
   sgd
   smorms3
+  svrg
```
rcurtin: I think the `katyusha` and `sarah` directories need to be added too. :)

src/mlpack/core/optimizers/svrg/svrg_impl.hpp
```diff
@@ -0,0 +1,179 @@
...
+    {
+      // Find the effective batch size (the last batch may be smaller).
+      effectiveBatchSize = std::min(batchSize, numFunctions - f);
+
+      function.Gradient(iterate, f, gradient, effectiveBatchSize);
+      fullGradient += gradient;
+
+      f += effectiveBatchSize;
+    }
+    fullGradient /= (double) numFunctions;
```
rcurtin: It seems to me like the calculation of the full gradient here could be in the conditional `if ((currentFunction % numFunctions) == 0)`, since it seems with SVRG the full gradient is calculated at the start of an iteration.

src/mlpack/core/optimizers/svrg/svrg_impl.hpp
```diff
@@ -0,0 +1,179 @@
...
+      // Calculate variance reduced gradient.
+      function.Gradient(iterate, f, gradient, effectiveBatchSize);
+      function.Gradient(iterate0, f, gradient0, effectiveBatchSize);
+
+      // Use the update policy to take a step.
+      updatePolicy.Update(iterate, fullGradient, gradient, gradient0,
+          effectiveBatchSize, stepSize);
+
+      f += effectiveBatchSize;
+    }
```
rcurtin: This seems like it is looping over all functions, but I think you could take the inner part of this loop out (lines 142-148) and put them at line 155 and everything would work fine.

src/mlpack/core/optimizers/svrg/barzilai_borwein.hpp
```diff
@@ -0,0 +1,109 @@
...
+  void Initialize(const size_t /* rows */, const size_t /* cols */)
+  { /* Do nothing. */ }
+
+  /**
+   * Barzilai-Borwein update step for SVRG.
+   *
+   * @param iterate Parameters that minimize the function.
+   * @param fullGradient The computed full gradient.
+   * @param gradient The current gradient matrix at time t.
+   * @param gradient The old gradient matrix at time t - 1.
```
rcurtin: I think the documentation is off here a little bit.

src/mlpack/core/optimizers/svrg/barzilai_borwein.hpp
```diff
@@ -0,0 +1,109 @@
...
+    {
+      // Step size selection based on Barzilai-Borwein (BB).
+      stepSize = std::pow(arma::norm(iterate - iterate0), 2.0) /
+          (arma::dot(iterate - iterate0, fullGradient - fullGradient0) + eps) /
+          (double) numBatches;
+
+      stepSize = std::min(stepSize, maxStepSize);
+    }
+
+    fullGradient0 = fullGradient;
```
rcurtin: Hmm, I am not sure, but maybe we could use `std::move()` here to avoid a copy---but only if we know that `fullGradient` won't be used by the caller of `Update()` after this call.  But I think the code in SVRG might need to be just a little restructured before that happens.

src/mlpack/core/optimizers/katyusha/katyusha_impl.hpp
```diff
@@ -0,0 +1,188 @@
...
+  size_t numBatches = numFunctions / batchSize;
+  if (numFunctions % batchSize != 0)
+    ++numBatches; // Capture last few.
+
+  const size_t actualMaxIterations = (maxIterations == 0) ?
+      std::numeric_limits<size_t>::max() : maxIterations;
+  for (size_t i = 0; i < actualMaxIterations; /* incrementing done manually */)
+  {
+    // Is this iteration the start of a sequence?
+    if ((currentFunction % numFunctions) == 0)
```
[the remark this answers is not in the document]
zoq: My intention was to use:
```c++
effectiveBatchSize = std::min(batchSize, numFunctions - currentFunction);
overallObjective += function.Evaluate(iterate, currentFunction, effectiveBatchSize);
```
the evaluation after the inner loop as a proxy for the overall performance, we could remove the if condition and evaluate the complete function, but I don't think that's necessary here. Let me know what you think.

> And should we set actualMaxIterations to the user-supplied maxIterations, or to maxIterations / numFunctions?

Yes, I will do that.

src/mlpack/core/optimizers/katyusha/katyusha.hpp
```diff
@@ -0,0 +1,164 @@
...
+   * @param lambda The regularization parameter.
+   * @param tau The momentum parameter.
+   * @param batchSize Batch size to use for each step.
+   * @param maxIterations Maximum number of iterations allowed (0 means no
+   *     limit).
+   * @param tolerance Maximum absolute tolerance to terminate algorithm.
+   * @param shuffle If true, the function order is shuffled; otherwise, each
+   *     function is visited in linear order.
+   */
+  Katyusha(const double stepSize = 1000,
```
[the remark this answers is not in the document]
zoq: Agreed, not sure what I was thinking.

src/mlpack/core/optimizers/katyusha/katyusha_impl.hpp
```diff
@@ -0,0 +1,188 @@
...
+    }
+
+    iterate0 = iterate;
+    y = iterate;
+    z = iterate;
+    w.zeros();
+
+    // Compute the full gradient.
+    size_t effectiveBatchSize = std::min(batchSize, numFunctions);
+    function.Gradient(iterate, 0, fullGradient, effectiveBatchSize);
```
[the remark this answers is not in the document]
zoq: The idea is to avoid an extra copy for the accumulation inside the loop:

```c++
function.Gradient(iterate, f, gradient, effectiveBatchSize);
fullGradient += gradient;
```

Maybe I missed something?

src/mlpack/core/optimizers/katyusha/katyusha_impl.hpp
zoq: Good idea will, do that.

src/mlpack/core/optimizers/katyusha/katyusha_impl.hpp
zoq: hm, I put the calculation inside the loop, nothing special, will add some more comments.

src/mlpack/core/optimizers/svrg/svrg_impl.hpp
zoq: What about introducing an update frequency for the inner loop.

src/mlpack/core/optimizers/svrg/svrg_impl.hpp
zoq: If we introduce the update frequency for the inner loop a user can control the full gradient calculation, what do you think? We can also remove the inner loop and put the full gradient inside the if condition.

src/mlpack/core/optimizers/katyusha/katyusha_impl.hpp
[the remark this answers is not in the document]
rcurtin: Oh!  Now I understand.  Thanks for clarifying. :)

src/mlpack/core/optimizers/katyusha/katyusha_impl.hpp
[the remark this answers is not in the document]
rcurtin: Okay, I've spent some more time with the code and I think I understand the transformation from `Katyusha1()` a bit better now.

Based on my understanding of `Katyusha1()` (just to be clear I mean Algorithm 3 of https://arxiv.org/pdf/1603.05953.pdf, in case there are different versions of the paper out there), there is an outer loop from line 11 to 23 and an inner loop from line 13 to 21.  The outer loop is performed S times, and updates the full gradient, then updates the parameter with the results of the inner loop (the parameter update can be moved into the inner loop and it is in your implementation).  So I would have expected code kind of like the form...

```
size_t maxOuterIterations = actualMaxIterations / numFunctions; // or something like this
for (size_t s = 0; s < maxOuterIterations; ++s)
{
  // calculate full gradient, print objective or an estimate of it

  // do inner loop stuff
}
```

so what was confusing me was that the outer iteration counter (`i`) is being incremented by `effectiveBatchSize` each time where I was expecting that either `i` would be incremented by `numFunctions` (with a maximum of `actualMaxIterations`), or `i` would be incremented by 1 (with a maximum of `numFunctions / actualMaxIterations` or something like that).  I hope that helps explain my confusion a little bit.

So it seems like the reason you are incrementing by `effectiveBatchSize` and not `numFunctions` is to avoid a potentially costly objective function evaluation every outer iteration, and then the objective will be printed to the user less often (once every `numFunctions / batchSize` outer iterations).  I wonder if we should instead print an objective function estimate based on the single-batch calculation you are doing, or something else.  It is really useful for users to see some kind of objective function so they know the optimization is converging, but it is unfortunate that sometimes it costs a lot of time to calculate it...

(Since I am not sure how well I explained my thinking, if my comment does not make any sense: I think I understand this code now; sorry for my confusion earlier, and let's not worry about my comment anymore. :))

src/mlpack/core/optimizers/svrg/svrg_impl.hpp
rcurtin: I see, I think that would be good to add.  For the default I guess the update frequency would be once per full pass over the data.

zoq:
I think this is ready.

src/mlpack/core/optimizers/katyusha/katyusha_impl.hpp
rcurtin: I think this is the parameter `innerIterations`, did I understand that right?  It might be worth saying in the documentation somewhere that the full gradient is only calculated in outer iterations; not sure, I think it's fine either way.

src/mlpack/core/optimizers/sarah/CMakeLists.txt
```diff
@@ -0,0 +1,13 @@
+set(SOURCES
+	sarah.hpp
```
rcurtin: Seems like strange indentation for this file (I know, minor comment...).

src/mlpack/core/optimizers/svrg/svrg_impl.hpp
```diff
@@ -0,0 +1,176 @@
...
+    overallObjective = 0;
+    for (size_t f = 0; f < numFunctions; f += batchSize)
+    {
+      const size_t effectiveBatchSize = std::min(batchSize, numFunctions - f);
+      overallObjective += function.Evaluate(iterate, f, effectiveBatchSize);
+    }
+
+    if (std::isnan(overallObjective) || std::isinf(overallObjective))
+    {
+      Log::Warn << "Katyusha: converged to " << overallObjective
```
rcurtin: Oops, I think this should say `SVRG` not `Katyusha` (there are a couple messages like this, in `SARAH` too).

src/mlpack/tests/sarah_test.cpp
```diff
@@ -0,0 +1,137 @@
...
+
+/**
+ * Create the data for the logistic regression test case.
+ */
+void CreateLogisticRegressionTestData(arma::mat& data,
+                                      arma::mat& testData,
+                                      arma::mat& shuffledData,
+                                      arma::Row<size_t>& responses,
+                                      arma::Row<size_t>& testResponses,
+                                      arma::Row<size_t>& shuffledResponses)
```
rcurtin: We could add this to an auxiliary header, because I think it's the same across all three tests (and also used in some other places); what do you think?

src/mlpack/tests/sarah_test.cpp
```diff
@@ -0,0 +1,137 @@
...
+
+using namespace std;
+using namespace arma;
+using namespace mlpack;
+using namespace mlpack::optimization;
+
+using namespace mlpack::distribution;
+using namespace mlpack::regression;
+
+BOOST_AUTO_TEST_SUITE(SAHRATest);
```
rcurtin: Oops, minor misspelling here.

src/mlpack/core/optimizers/katyusha/katyusha_impl.hpp
```diff
@@ -0,0 +1,202 @@
...
+      iterate = tau1 * z + tau2 * iterate0 + (1 - tau1 - tau2) * y;
+
+      // Calculate variance reduced gradient.
+      function.Gradient(iterate, currentFunction, gradient,
+          effectiveBatchSize);
+      function.Gradient(iterate0, currentFunction, gradient0,
+          effectiveBatchSize);
+
+      arma::mat zNew = z - alpha * (fullGradient + (gradient - gradient0) /
+          (double) batchSize);
```
rcurtin: I believe this comes from the top of page 11:

> By the minimality definition of z_{k + 1}, we have that
>   z_{k+1} − z_k + α∇_{k+1} + αg = 0

Do you think it would be useful to add a comment to point that out?  I spent a little while looking at the original line 10 in the Katyusha() algorithm before I realized the equivalence here.

src/mlpack/core/optimizers/katyusha/katyusha_impl.hpp
```diff
@@ -0,0 +1,202 @@
...
+
+      // Proximal update, choose between Option I and Option II. Shift relative
+      // to the Lipschitz constant or take a constant step using the given step
+      // size.
+      if (proximal)
+      {
+        y = iterate + 1.0 / (3.0 * lipschitz) * (zNew - z);
```
rcurtin: I guess there is a conversion in the paper too for line 11 to this, but I don't seem to be able to find it (I have also not read the paper in full detail, mostly just read the algorithm and skipped the proofs); do you want to add a note about it?

Issue #1152, opened 2017-11-17
Add tests for command-line and Python bindings


rcurtin:
This is a great issue for anyone looking to get involved with mlpack development.  No huge amount of C++ or machine learning knowledge is necessary; in order to write a test, essentially all you need to know is how the particular binding you are writing a test for works, and you can use the existing test code as boilerplate for the new tests.  So I think this is a good place to get started.

If you don't know about the automatic bindings system, you can read about it at http://www.mlpack.org/docs/mlpack-git/doxygen/bindings.html .

In essence, the problem is that we have all of these command-line programs and Python bindings like `mlpack_knn`, `mlpack_sparse_coding`, and so forth, but these are not rigorously tested.  Thanks to the automatic binding system, we now have an easy way to test these, which allows us to avoid all kinds of little bugs that have happened in the past where a command-line program might access the name of a variable incorrectly or something like this.  However, the tests have not been written yet.  So the purpose of this issue is to ensure that these tests actually do get written.

So, for example, let's take a look at a test we already have for the PCA binding.  This is from `src/mlpack/tests/pca_test.cpp`.

```
/**
 * Check that we can't specify an invalid new dimensionality.
 */
BOOST_AUTO_TEST_CASE(PCATooHighNewDimensionalityTest)
{ 
  arma::mat x = arma::randu<arma::mat>(5, 5);
  
  SetInputParam("input", std::move(x));
  SetInputParam("new_dimensionality", (int) 7); // Invalid.

  Log::Fatal.ignoreInput = true;
  BOOST_REQUIRE_THROW(mlpackMain(), std::runtime_error);
  Log::Fatal.ignoreInput = false;
}
```

This is a pretty simple test, which just checks that a user can't ask for a new dimensionality that is too high, and this is the kind of test we're looking for here.  We don't need to check if the exact output of the algorithm is correct---we are already doing that in the other tests we have for the algorithms.  Instead we want to test the input/output of these bindings.

So, if you'd like to write a test, the basic procedure is this:

 * Pick an uncompleted binding from the list below (I am trying to keep this up to date):
   * [x] adaboost (`src/mlpack/methods/adaboost_main.cpp`): AdaBoost
   * [x] approx_kfn (`src/mlpack/methods/approx_kfn_main.cpp`): approximate k-nearest-neighbor search
   * [x] cf (`src/mlpack/methods/cf/cf_main.cpp`): collaborative filtering
[...]

PR #1165, opened 2017-12-12
Bindings tests for emst

manish7294:
Bindings Tests of mlpackMain() for emst_main.cpp.

### DIFF
--- a/src/mlpack/tests/main_tests/emst_test.cpp
+++ b/src/mlpack/tests/main_tests/emst_test.cpp
@@ -9,7 +9,11 @@
  * 3-clause BSD license along with mlpack.  If not, see
  * http://www.opensource.org/licenses/BSD-3-Clause for more information.
  */
+#include <string>
+
 #define BINDING_TYPE BINDING_TYPE_TEST
+static const std::string testName = "EMST";
+
 #include <mlpack/core.hpp>
 #include <mlpack/core/util/mlpack_main.hpp>
 #include <mlpack/methods/emst/emst_main.cpp>
@@ -21,31 +25,13 @@
 
 using namespace mlpack;
 
-namespace mlpack {
-namespace bindings {
-namespace tests {
-
-extern std::string programName;
-
-}
-}
-}
-
-// Utility function to set a parameter and mark it as passed, using copy
-// semantics.
-template<typename T>
-void SetInputParam(const std::string& name, const T& value)
-{
-  CLI::GetParam<T>(name) = value;
-  CLI::SetPassed(name);
-}
-
-// Utility function to set a parameter and mark it as passed, using move
-// semantics.
+// Utility function to set a parameter and mark it as passed,
+// using copy semantics for lvalues and move semantics for rvalues.
 template<typename T>
 void SetInputParam(const std::string& name, T&& value)
 {
-  CLI::GetParam<T>(name) = std::move(value);
+   CLI::GetParam<typename std::remove_reference<T>::type>(name)
+        = std::forward<T>(value);
   CLI::SetPassed(name);
 }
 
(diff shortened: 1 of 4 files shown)


rcurtin:
Can you add this to the CMake configuration so that the new test gets built?  However be aware that it will not build successfully because of a problem pointed out in #1152.  We will work on fixing that so this will build right...

In the mean time you can ensure this works by commenting out `main_tests/pca_test.cpp` in `src/mlpack/tests/CMakeLists.txt`.

src/mlpack/tests/main_tests/emst_test.cpp
```diff
@@ -0,0 +1,108 @@
...
+  arma::mat x = arma::randi<arma::mat>(6, 2, distr_param(0, 10));
+
+  // Input random data points.
+  SetInputParam("input", std::move(x));
+  SetInputParam("leaf_size", (int) 2);
+
+  mlpackMain();
+
+  // Now check that the output has 3 dimensions.
+  BOOST_REQUIRE_EQUAL(CLI::GetParam<arma::mat>("output").n_cols, 3);
```
rcurtin: Can we also check the number of points in the output here?  I think we can also ensure that all values in the first two dimensions are very close to integers (since they are indices).

manish7294: yes we can compare the number of output points to total number of input points -1. As MST will will have exactly one edge between each points. Please correct me if I am wrong.

manish7294:
Sure, I will be making the change to CMake configuration.

manish7294:
@rcurtin please help me out with this.
screenshot from 2017-12-15 20-43-19
This is the check which I employed for number of output points.

manish7294:
@rcurtin please look into this.
instance #3517.2  of travis-ci pass the build and test while as instance #3517.1 failed just after completion of build stating below output:

> 
```
$ CTEST_OUTPUT_ON_FAILURE=1 travis_wait 30 ctest -j2


Still running (1 of 30): ctest -j2
Still running (2 of 30): ctest -j2
Still running (3 of 30): ctest -j2
Still running (4 of 30): ctest -j2

The job exceeded the maximum time limit for jobs, and has been terminated.


```

rcurtin:
> @rcurtin please look into this.

Unfortunately I do not have time to deal with all issues.  You should try running these tests on your local system to try and reproduce the problem, and if you cannot do that you should attempt to imitate the Travis build environment, possibly using a Docker container, and reproduce the issue and solve it that way.

manish7294:
I already made that check on my system. There was no error while running tests.

rcurtin:
This looks good, thank you for taking the time to put this together (and sorry I couldn't help more in depth).  I think that the reason the test is failing is because the build was running out of time (there is a 70 minute limit).  This seems strange because you only added one extra file to compile, and typically the build takes only 1 hour.  So I'm surprised, and perhaps this was a Travis load issue.  I'll have rebuild the PR a few times just to make sure that it is not something else.

It looks like all the basic flows through the program are covered (you are lucky or chose well---EMST is one of the simplest bindings :)), so I think there is no need to add more tests, and the tests look reasonable to me.

However let's wait until #1172 is resolved to merge this.  I'll send another note when that is handled.

rcurtin:
Now that #1172 is resolved, can you resolve the merge conflict here so that Travis can test the build please?

manish7294:
Done!

src/mlpack/tests/main_tests/emst_test.cpp
```diff
@@ -0,0 +1,142 @@
...
+
+using namespace mlpack;
+
+// Utility function to set a parameter and mark it as passed,
+// using copy semantics for lvalues and move semantics for rvalues.
+template<typename T>
+void SetInputParam(const std::string& name, T&& value)
+{
+  CLI::GetParam<typename std::remove_reference<T>::type>(name)
+        = std::forward<T>(value);
```
rcurtin: A minor style comment---can you put the `=` on the preceding line please?

rcurtin:
I'll go ahead and merge this in 4 days in case anyone else has any comments.

manish7294:
Thanks @rcurtin, I will more than happy to receive mlpack stickers. I have done the style change.

rcurtin:
Thanks again for the contribution! :)

PR #1166, opened 2017-12-12
Fix Haiku: link against librt

rcurtin:
This is a patch from @extrowerk.  (Thanks!)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,10 +9,10 @@ env:
 before_install:
   - sudo systemctl stop apt-daily.service apt-daily.timer
   - sudo systemctl disable apt-daily.service apt-daily.timer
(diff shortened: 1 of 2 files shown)


rcurtin:
@extrowerk: thanks again for the patch.  Even though it is small you are now a contributor to mlpack, so if you like I can add your name and email in `COPYRIGHT.txt` and `src/mlpack/core.hpp`.  Also, we can mail you some mlpack stickers if you want; just send me an email with your mailing address and we'll get them on the way... :)

extrowerk:
Oh, how cool!
I definitely need a sticker for my laptop!

All of my contribution is released under BSD-3, if you so hard insists, then you can add me to the contributors list :)

rcurtin:
@extrowerk: your call, I'm happy to add or not add.  If you can just send me an email with your address we can send them out to you. :)

PR #1169, opened 2017-12-13
Hdbscan implementation

s1998:


src/mlpack/core/metrics/hdbscan_metric.hpp
```diff
@@ -0,0 +1,43 @@
...
+ *  
+ */
+
+#ifndef MLPACK_CORE_HDBSCAN_METRIC_HPP
+#define MLPACK_CORE_HDBSCAN_METRIC_HPP
+
+namespace mlpack {
+namespace metric {
+
+class HdbscanMetric
```
rcurtin: Could you add a little documentation to this class please?  To the best of my understanding you are doing a really clever trick here---the vectors that are given to this by HDBSCAN have d_core appended to them.  Whatever the "trick" ends up being, we should document it here because it will make it clear how the class is meant to be used.

src/mlpack/core/metrics/hdbscan_metric_impl.hpp
```diff
@@ -0,0 +1,36 @@
...
+// in (a-b) vector but this contains the last
+// entry too. Remove it using lastEle.
+  template<typename VecTypeA, typename VecTypeB>
+  typename VecTypeA::elem_type HdbscanMetric::Evaluate(
+      const VecTypeA& a,
+      const VecTypeB& b)
+  {
+    typename VecTypeA::elem_type dcoreA = a(a.n_rows-1);
+    typename VecTypeB::elem_type dcoreB = b(b.n_rows-1);
+    typename VecTypeB::elem_type distSq = arma::accu(arma::square(a - b));
```
rcurtin: Consider calling out to SquaredEuclideanDistance::Evaluate() here.

src/mlpack/core/metrics/hdbscan_metric_impl.hpp
```diff
@@ -0,0 +1,36 @@
...
+    typename VecTypeB::elem_type lastEle = (dcoreA - dcoreB)*(dcoreA - dcoreB);
```
rcurtin: It might be easier to read to use std::pow(dcoreA - dcoreB, 2.0) here.

src/mlpack/core/metrics/hdbscan_metric_impl.hpp
```diff
@@ -0,0 +1,36 @@
...
+    typename VecTypeB::elem_type dist = sqrt(distSq - lastEle);
```
rcurtin: I think that we have to specify `std::sqrt()` here otherwise some other compilers may have a problem with it.

src/mlpack/methods/hdbscan/hdbscan.hpp
```diff
@@ -0,0 +1,218 @@
...
+#include <mlpack/core.hpp>
+#include <boost/dynamic_bitset.hpp>
+#include <mlpack/methods/neighbor_search/neighbor_search.hpp>
+#include <mlpack/methods/emst/dtb.hpp>
+#include <mlpack/core/metrics/hdbscan_metric.hpp>
+
+namespace mlpack {
+namespace hdbscan {
+
+template<typename NeighborSearch = neighbor::NeighborSearch
```
rcurtin: Could you please add some documentation discussing what the class does?  This can be similar to the comment at the top of the file, but it should also detail each of the template parameters and what they do.

src/mlpack/methods/hdbscan/hdbscan.hpp
```diff
@@ -0,0 +1,218 @@
...
+class HDBSCAN
+{
+ public:
+  /**
+   * Construct the HDBSCAN object with the given parameters.
+   *
+   * @param minPoints The number of points present in a cluster.
+   */
+  HDBSCAN(const size_t minPoints = 10,
+          bool allowSingleCluster = false);
```
rcurtin: Can you document 'allowSingleCluster' please?

src/mlpack/methods/hdbscan/hdbscan_impl.hpp
```diff
@@ -0,0 +1,763 @@
...
+
+/**
+ * Construct the HDBSCAN object with the given parameters.
+ */
+template<typename NeighborSearch,
+         typename MetricType,
+         template<typename TreeMetricType,
+                  typename TreeStatType,
+                  typename TreeMatType> class TreeType >
+HDBSCAN<NeighborSearch,
```
rcurtin: This could fit on one line:

```
HDBSCAN<NeighborSearch, MetricType, TreeType>::HDBSCAN(
```

src/mlpack/methods/hdbscan/hdbscan_impl.hpp
```diff
@@ -0,0 +1,763 @@
...
+        arma::Row<size_t>& assignments)
+{
+  // genearting a matrix with the values of dcore
+  arma::Mat<size_t> neighbors;
+  arma::mat distances;
+  neighborSearch.Train(data);
+  neighborSearch.Search(minPoints, neighbors, distances);
+  arma::mat dataWithDcore = arma::conv_to<arma::mat>::from(data);
+  dataWithDcore.resize(dataWithDcore.n_rows+1, dataWithDcore.n_cols);
+  dataWithDcore.row(dataWithDcore.n_rows-1) = distances.row(distances.n_rows-1);
```
rcurtin: I see that what you are doing here is a clever trick to make sure that the HDBSCANMetric properly has access to d_core.  But each step of the conversion here causes a runtime penalty, and it seems a little strange to me to force the data the user gives (which could have non-double type) into a double. Instead what I might suggest here is a little bit different... let me know what you think:

Currently the HdbscanMetric class does not hold anything internally.  I suggest adding some instantiated members to the class:

```
MatType& data;
arma::mat coreDistances;
```

Then, for the DualTreeBoruvka class, the data that you call it with is created like this:

```
// Fill a vector with increasing indices.
arma::Row<size_t> mstData = arma::linspace<arma::Row<size_t>>(0,
    data.n_cols - 1, data.n_cols);
```

Then, you could modify HdbscanMetric::Evaluate() to look more like the following:

```
double Evaluate(const VecTypeA& a, const VecTypeB& b)
{
  const size_t aIndex = a[0];
  const size_t bIndex = b[0];

  const double dCoreA = coreDistances(0, aIndex);
  const double dCoreB = coreDistances(0, bIndex);

  const double dist =
      SquaredEuclideanDistance::Evaluate(data.col(aIndex), data.col(bIndex));
  ...
}
```

In this way we can preserve the correct type of the data and avoid some expensive copies.  Let me know what you think of this idea, or if you'd like me to clarify any of my thoughts.

src/mlpack/methods/hdbscan/hdbscan_impl.hpp
```diff
@@ -0,0 +1,763 @@
...
+  arma::Mat<size_t> result;
+  GetClusters(condensedTree, result);
+
+  // save results obtained
+  assignments = result;
+}
+
+//  A class to help merge two clusters to form a
+// single linkage tree.
+class UnionForSingleLinkageClustering
```
rcurtin: Is this the same as UnionFind from src/mlpack/methods/emst/union_find.hpp?  If so we can replace this with that.

src/mlpack/methods/hdbscan/hdbscan_impl.hpp
```diff
@@ -0,0 +1,763 @@
...
+  size_t GetSize(size_t i)
+  {
+    return sizeVector(i);
+  }
+};
+
+/**
+ * Converts the input (sorted MST) to a single linkage tree.
+ * Each row of the output contaisn 3 values
+ * (cluster1 cluster2 d size)
```
rcurtin: Don't be afraid to use multiple types here.  I would suggest using one object of type arma::Mat<size_t> to hold cluster1 and cluster2 and size, and then a third MatType to hold the distances between the clusters.

src/mlpack/methods/hdbscan/hdbscan_impl.hpp
```diff
@@ -0,0 +1,763 @@
...
+             MetricType,
+             TreeType>::
+SingleLinkageTreeClustering(const MatType& inputMST,
+                          MatType& singleLinkageTree)
+{
+  singleLinkageTree.set_size(4, inputMST.n_cols);
+  size_t no = inputMST.n_cols + 1;
+  UnionForSingleLinkageClustering u(no);
+  size_t parentOfv1, parentOfv2;
+  double delta;
```
rcurtin: Just a note, it can be a little hard for me (or another maintainer) to jump in and immediately understand what's going on here.  I figured it out after a little while, but it would be really helpful if you could go through and add some comments to some of the code.

src/mlpack/methods/hdbscan/hdbscan_impl.hpp
```diff
@@ -0,0 +1,763 @@
...
+  for (size_t i = 0; i < inputMST.n_cols; i++)
+  {
+    delta = inputMST(2, i);
+    parentOfv1 = u.Find(inputMST(0, i));
+    parentOfv2 = u.Find(inputMST(1, i));
+    singleLinkageTree.col(i) = arma::vec({(double)parentOfv1,
+      (double)parentOfv2, delta,
+      (double)(u.GetSize(parentOfv1) + u.GetSize(parentOfv2))});
+    u.UnionOfTwoClusters(parentOfv1, parentOfv2);
```
rcurtin: This calculation does look pretty similar to the DualTreeBoruvka::EmitResults() calculation.  If you had access to the edge list produced by DualTreeBoruvka, would it make this code unnecessary?  I think (but I am not sure) that basically this code is the reverse of the code in EmitResults().  If that's true we can refactor DualTreeBoruvka so you get what you need out of it.  But I may have misunderstood, so please correct me if I am wrong.

src/mlpack/methods/hdbscan/hdbscan_impl.hpp
```diff
@@ -0,0 +1,763 @@
...
+      result[currentPoint] = SIZE_MAX;
+    else
+      // assign the correct cluster
+      result[currentPoint] = labelOfClusters[parentOfCurrentPoint];
+  }
+} // function getLabels ends
+
+/**
+ * This functin provides label to each and every point
+ * Noise points are labelled as ( total number of points + 1)
```
rcurtin: The DBSCAN implementation labels these as SIZE_MAX; you could do the same here, or if you have a good reason for your choice we can change DBSCAN too.

src/mlpack/methods/hdbscan/hdbscan_main.cpp
```diff
@@ -0,0 +1,84 @@
...
+#include <mlpack/core/util/cli.hpp>
+#include <mlpack/core/util/mlpack_main.hpp>
+
+#include "hdbscan.hpp"
+
+using namespace mlpack;
+using namespace mlpack::hdbscan;
+using namespace std;
+
+PROGRAM_INFO("HDBSCAN algorithm", "This program helps in clustering "
```
rcurtin: I think we should make this documentation a lot longer and add some example usage.  I can help with this part, and we can do it when we get closer to merge.

src/mlpack/methods/hdbscan/hdbscan_main.cpp
```diff
@@ -0,0 +1,84 @@
...
+    "\n\n"
+    "The output is saved in a column matrix");
+
+PARAM_STRING_OUT("output_file",
+                 "Output file. Stores cluster label of each point.",
+                 "o");
+PARAM_STRING_IN("input_file",
+                "Input dataset to cluster",
+                "i",
+                "noFileSelected");
```
rcurtin: I think that this was written a long time ago, and since then the infrastructure has changed a little bit.  I can help do the refactoring---just let me know.  In any case the two parameters above can change to:

```
PARAM_MATRIX_IN_REQ("input", "Input dataset to cluster.", "i");
PARAM_UROW_OUT("output", "Output cluster assignments for each point.",
    "o");
```

src/mlpack/methods/hdbscan/hdbscan_main.cpp
```diff
@@ -0,0 +1,84 @@
...
+PARAM_FLAG("single_cluster", "Allow single cluster.", "s");
```
rcurtin: Does this mean allow all the points to lie in a single cluster, or allow clusters with a single point?  I think it is the former but we should be clear in the documentation.

src/mlpack/methods/hdbscan/hdbscan_main.cpp
```diff
@@ -0,0 +1,84 @@
...
+      allowSingleCluster = false;
+  }
+
+  // Default value of minimum cluster size is 10
+  if (!CLI::HasParam("minimumClusterSize"))
+  {
+  Log::Warn << "--minimumClusterSize is not specified. Default value is 10!"
+            << endl;
+      minimumClusterSize = 10;
+  }
```
rcurtin: Here's another place where the code has changed somewhat.  Now we can use functions like `CheckParamValue()` and `ReportIgnoredParam()` and `RequireAtLeastOnePassed()`, etc., to do the input checks.  I can help refactor this; just let me know if you'd like me to.

src/mlpack/tests/hdbscan_test.cpp
```diff
@@ -0,0 +1,101 @@
...
+using namespace mlpack::hdbscan;
+using namespace mlpack::distribution;
+
+BOOST_AUTO_TEST_SUITE(HDBSCANTest);
+
+BOOST_AUTO_TEST_CASE(singleClusterTest)
+{
+  // Generate 5000 points on 0.1 radius circle.
+  arma::mat points(2, 5000);
+  double pi = 3.14156;
```
rcurtin: You can use M_PI here.

src/mlpack/tests/hdbscan_test.cpp
```diff
@@ -0,0 +1,101 @@
...
+    theeta = (double)i/5000.0;
+    points(0, i) = 0.1*sin(2*pi*theeta);
+    points(1, i) = 0.1*cos(2*pi*theeta);
+  }
+
+  HDBSCAN<> h1(5, true);
+  arma::Row<size_t> assignments;
+  h1.Cluster(points, assignments);
+
+  // Some points in the circle will get classified as noise
```
rcurtin: This doesn't seem to mesh with the test condition; is the comment correct?

src/mlpack/tests/hdbscan_test.cpp
```diff
@@ -0,0 +1,101 @@
...
+  for (size_t i = 0; i < assignments.n_cols; i++)
+  {
+    BOOST_REQUIRE((assignments(i) == 0));
```
rcurtin: I'd suggest `BOOST_REQUIRE_EQUAL(assignments[i], 0);`.

src/mlpack/methods/hdbscan/hdbscan.hpp
```diff
@@ -0,0 +1,218 @@
...
+   * The assigned clusters have values from 0 onwards. The noise points  
+   * are labelled as (total nuber of points + 1)
+   *
+   * @param MatType Type of matrix (arma::mat or arma::sp_mat).
+   * @param data Dataset to cluster.
+   * @param assignments Vector to store cluster assignments.
+   */
+  template<typename MatType>
+  void Cluster(const MatType& data,
+               arma::Row<size_t>& assignments);
```
rcurtin: I am not sure, but I wanted to see what you thought: do you think that a user might benefit from being able to return a full hierarchy of the clusters?  Let me know what you think.  Maybe it would be some nice functionality.

s1998: Yes, this is the idea behind Single Linkage Tree, merge the closest clusters to obtain the next cluster and in this way it goes on forming the hierarchy. Should I create a separate method
```
getSLT() 
```

src/mlpack/tests/hdbscan_test.cpp
s1998: Yes, unfortunately it is.

Initially I thought there was an error and added a rounding term eps_error, but then I realized eps_error dpends on distance between points in dataset and is not the correct way to go.

I tried to dig into it. Here is what I found (I don't know if it helps but this is the reason for some points getting classified as noise): When we are creating a MST, though all the points are equivalent, some points will be considered first and will be present with shorter edges in the MST. 

Example : For minClusterSize = 5
Say point p6 has neighbors p1, p2, p3, p4, p5. Once all these 5 points are inserted, when p6 will be considered all edges that are considered with p6 will have distance ( ``` > dCore(p6)```) and p6 will be not equivalent to p1 in the MST, though we know it is.

src/mlpack/methods/hdbscan/hdbscan_impl.hpp
s1998: No, they are slightly different.
In emst/uinon_find.hpp when we do union, a point becomes parent of another point.
Instead what needs to be done is when we are doing union of two clusters, give the merged cluster a new index.
```
  void UnionOfTwoClusters(size_t m, size_t n)
  {
    // Each union will form a new cluster
    // with size as sum of its children
    sizeVector(nextLabel) = sizeVector(m) + sizeVector(n);
    parentVector(m) = nextLabel;
    parentVector(n) = nextLabel;
    nextLabel++;
  }
```
The nextLabel is making the difference.

src/mlpack/methods/hdbscan/hdbscan_impl.hpp
s1998: To construct SLT from MST, we go on merging the two closest clusters. 
For this to happen, the edges need to be sorted (w.r.t length), so I think EmitResults() is necessary . 
Though I think this part is not necessary 
```
// Make sure the edge list stores the smaller index first to
// make checking correctness easier
if (ind1 < ind2)
      {
        edges[i].Lesser() = ind1;
        edges[i].Greater() = ind2;
      }
      else
      {
        edges[i].Lesser() = ind2;
        edges[i].Greater() = ind1;
      }
```
If I have edge list and oldFromNew, this can be skipped (though I doubt if this will give any speed up).

s1998:
@rcurtin
I need some help with the metric implementation (I was stuck here last time too). 
I modified hdbscan_metric to have
```
  const MatType& data;
  const arma::mat& coreDistances;
```
and it's constructor to
```
HdbscanMetric(const MatType& dataInp,
    const arma::mat& dcore):
    data(dataInp),
    coreDistances(dcore)
    { }
```
I am getting this error [](https://pastebin.com/fSiVRdwF)

s1998:
https://pastebin.com/fSiVRdwF

src/mlpack/methods/hdbscan/hdbscan_impl.hpp
rcurtin: So, I am really a big fan of code reuse, so I think it is a good idea to try and use the existing `UnionFind` for this.  The difference that you are saying, basically, is that you are using `nextLabel` to label each cluster from `0` to the number of clusters.  But you already have a unique index---the index of the point.  I believe you can use this, and I would prefer that you do this: having two union find structures in the code is more maintenance for later and more tests for now that need to be written.

In any case... since the comments and the code are not really clear through `SingleLinkageTreeClustering()` and `CondenseTree()`, it is difficult for me to quickly understand the other changes that would be necessary.  Please consider what would need to be done to make this work though and let me know what you think.

src/mlpack/tests/hdbscan_test.cpp
rcurtin: I think we are talking about something different... what I mean is this: the comment says that some points will be classified as noise, but the actual test condition (on line 40) tests that no points are classified as noise.

Unfortunately I don't understand the situation that the rest of your explanation is referring to; can you clarify please?

src/mlpack/methods/hdbscan/hdbscan.hpp
rcurtin: I think it would be a good idea, but how about we come back to this point when the rest of the code is ready?  Then the design we should use should be a little clearer.

rcurtin:
Yeah, I think you are right that some of the other tree types are going to try to call and instantiate the metric with a default constructor.  I would consider adding a default constructor, but then when you are holding `MatType& data` and `arma::mat& coreDistances` internally, don't hold references, and consider using the advanced constructor to make an alias:

http://arma.sourceforge.net/docs.html#Mat

Basically I can make an alias like this:

```
arma::mat matrix(10, 10, arma::fill::randu);
arma::mat alias(matrix.memptr(), 10, 10, false, true);
```

You could do that if the metric is constructed with matrix arguments, or just set it to an empty matrix otherwise (and warn in the documentation that the metric will not be useful unless those are set).  That should help get you closer to working.

rcurtin:
I'll close this for inactivity for now, and when it is updated we can reopen it.  (I am just trying to prune the list of open PRs.)

PR #1170, opened 2017-12-13
Fix recurrent layers' compatibility with batch sizes > 1.

rcurtin:
I tried to build an LSTM network with batch size > 1, but there were some Armadillo shape errors.  I believe this code fixes that; however, the tests do not work right now.  I need to redesign the tests...

### DIFF
--- a/src/mlpack/core/optimizers/sgd/sgd_impl.hpp
+++ b/src/mlpack/core/optimizers/sgd/sgd_impl.hpp
@@ -103,7 +103,8 @@ double SGD<UpdatePolicyType, DecayPolicyType>::Optimize(
     }
 
     // Find the effective batch size (the last batch may be smaller).
-    const size_t effectiveBatchSize = std::min(batchSize,
+    const size_t effectiveBatchSize = std::min(
+        std::min(batchSize, actualMaxIterations - i),
         numFunctions - currentFunction);
 
     function.Gradient(iterate, currentFunction, gradient, effectiveBatchSize);
--- a/src/mlpack/core/optimizers/sgd/sgd_impl.hpp
+++ b/src/mlpack/core/optimizers/sgd/sgd_impl.hpp
@@ -102,7 +102,12 @@ double SGD<UpdatePolicyType, DecayPolicyType>::Optimize(
         function.Shuffle();
     }
 
-    // Find the effective batch size (the last batch may be smaller).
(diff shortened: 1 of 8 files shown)


rcurtin:
I won't be able to fix the static analysis checks on this one (the warning is invalid in this case) but otherwise I have no other changes to make to this one and I think it is ready.

src/mlpack/core/optimizers/sgd/sgd_impl.hpp
```diff
@@ -103,7 +103,8 @@ double SGD<UpdatePolicyType, DecayPolicyType>::Optimize(
     }
 
     // Find the effective batch size (the last batch may be smaller).
-    const size_t effectiveBatchSize = std::min(batchSize,
+    const size_t effectiveBatchSize = std::min(
```
zoq: I think the idea here is to make sure that the ``batchSize`` does not change? But it looks like this only works if ``actualMaxIterations >= numFunctions >= batchSize`` right?

rcurtin: I made the change here to handle the situation where, e.g., `batchSize` is 50 but `actualMaxIterations` is, say, 5.  The current code will still execute 50 iterations of SGD because it will choose `effectiveBatchSize = 50`.  So here to choose the effective batch size we want to take the minimum of three quantities:

 - the user-specified batch size (`batchSize`)
 - the number of points remaining until we have looked at all functions (`numFunctions - currentFunction`)
 - the number of iterations remaining until the user wants optimization to terminate (`actualMaxIterations - i`)

Let me know if I've overlooked something or if that's actually not a good idea.

zoq: I see, this is fine for me, it might be good to write that down at some point.

zoq: Also, it might be good to do the same for the other optimizer as well, I can do that.

rcurtin: No problem, let me add comments and I'll go ahead and add it to the other optimizers too (it's an easy fix).

rcurtin:
I'll merge once the builds pass.  I think SPALeRA is the only other optimizer where the fix with `actualMaxIterations` fits; it seemed like the CMAES methods and IQN did not have the same issue.

PR #1172, opened 2017-12-15
Fixed issue with redifenition at binding tests

evgfreyman:
If you follow proposed way of writing tests (similar to `src/mlpack/tests/main_tests/pca_test.cpp`) then you get `mlpackMain` and `mlpack::bindings::tests::programName` redefined. This happens because we include all main files of bindings into one executable (mlpack_test)

Solution:

- Unique naming for each `mlpackMain` in case of binding tests. (through macros MAIN)
- All implementations of class `Option` consume an extra parameter - `programName`. It is defined unique in case of each binding test and dummy in other cases. Only second level macros are changed, so there is no need to change bindings themself.
- Only `TestOption` uses `programName` extra parameter, `PyOption` still uses global `programName`. I left it in order not to breake anything. This probably should be refactored

The existing code of bindings and tests are modified only a little:

- Using `MAIN()` - instead of `mlpackMain()` 
- At each binding tests two unique lines are added:

```
#define PROGRAM_NAME pcaProgramName
static const std::string pcaProgramName = "PrincipalComponentAnalysis";
```

Here is also a test suit for LinearRegression included. It has only one test case in it, just to show that this approach works.

### DIFF
--- a/src/mlpack/bindings/tests/test_option.hpp
+++ b/src/mlpack/bindings/tests/test_option.hpp
@@ -61,7 +61,8 @@ class TestOption
              const std::string& cppName,
              const bool required = false,
              const bool input = true,
-             const bool noTranspose = false)
+             const bool noTranspose = false,
+	     const std::string& programName = "")
   {
     // Create the ParamData object to give to CLI.
     util::ParamData data;
--- a/src/mlpack/bindings/tests/test_option.hpp
+++ b/src/mlpack/bindings/tests/test_option.hpp
@@ -54,7 +54,7 @@ class TestOption
    * @param noTranspose If the parameter is a matrix and this is true, then the
    *      matrix will not be transposed on loading.
    * @param testName Name of the test (used for identifiying which binding test 
-   * this option belongs to)
+   *      this option belongs to)
    */
   TestOption(const N defaultValue,
              const std::string& identifier,
--- a/src/mlpack/bindings/tests/test_option.hpp
+++ b/src/mlpack/bindings/tests/test_option.hpp
@@ -53,6 +53,8 @@ class TestOption
    * @param input Whether or not the option is an input option.
    * @param noTranspose If the parameter is a matrix and this is true, then the
    *      matrix will not be transposed on loading.
+   * @param testName Name of the test (used for identifiying which binding test 
+   * this option belongs to)
    */
   TestOption(const N defaultValue,
              const std::string& identifier,
@@ -62,7 +64,7 @@ class TestOption
              const bool required = false,
              const bool input = true,
              const bool noTranspose = false,
-             const std::string& programName = "")
+             const std::string& testName = "")
   {
     // Create the ParamData object to give to CLI.
     util::ParamData data;
@@ -82,7 +84,7 @@ class TestOption
 
     const std::string tname = data.tname;
 
-    CLI::RestoreSettings(programName, false);
+    CLI::RestoreSettings(testName, false);
 
     // Set some function pointers that we need.
     CLI::GetSingleton().functionMap[tname]["GetPrintableParam"] =
@@ -95,7 +97,7 @@ class TestOption
     if (!input)
       CLI::SetPassed(identifier);
 
-    CLI::StoreSettings(programName);
+    CLI::StoreSettings(testName);
     CLI::ClearSettings();
   }
 };
--- a/src/mlpack/bindings/tests/test_option.hpp
+++ b/src/mlpack/bindings/tests/test_option.hpp
@@ -62,7 +62,7 @@ class TestOption
              const bool required = false,
              const bool input = true,
              const bool noTranspose = false,
-	     const std::string& programName = "")
+             const std::string& programName = "")
   {
     // Create the ParamData object to give to CLI.
     util::ParamData data;
--- a/src/mlpack/core/util/param.hpp
+++ b/src/mlpack/core/util/param.hpp
@@ -15,24 +15,6 @@
 #ifndef MLPACK_CORE_UTIL_PARAM_HPP
 #define MLPACK_CORE_UTIL_PARAM_HPP
 
-#include <string>
-
-// PROGRAM_NAME is used for mlpackMain modification (mlpackMain##PROGRAM_NAME)
-// PROGRAM_NAME_SUBSTUTIDE is used for passing unique program name to TestOption
-#ifdef PROGRAM_NAME
-#define PROGRAM_NAME_SUBSTITUDE PROGRAM_NAME
-#else
-#define PROGRAM_NAME
-#define PROGRAM_NAME_SUBSTITUDE programNameSubstitude
-static const std::string programNameSubstitude = "programNameSubstitude";
-#endif
-// The MAIN macro is mlpackMain##PROGRAM_NAME. The goal is to have different
-// names of main procedure for main tests when all method's  main functions
-// are linked into one executable (mlpack_test)
-#define TOKENPASTE1(x, y) x ## y
-#define TOKENPASTE2(x, y) TOKENPASTE1(x, y)
-#define MAIN TOKENPASTE2(mlpackMain, PROGRAM_NAME)
-
 // Required forward declarations.
 namespace mlpack {
 namespace data {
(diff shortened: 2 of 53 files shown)


rcurtin:
Hey @evgfreyman,

Thank you so much for pointing this issue out and proposing a solution.  But I think I have found an easier and cleaner way to solve the first problem.  We can simply add a forward definition of `mlpackMain()` and call it inline.  So I added

```
inline void mlpackMain();
```

to the section of `mlpack_main.hpp` under `#elif (BINDING_TYPE == BINDING_TYPE_TEST)`; the whole section looks like this:

```
...

#elif(BINDING_TYPE == BINDING_TYPE_TEST) // This is a unit test.

#include <mlpack/bindings/tests/test_option.hpp>
#include <mlpack/bindings/tests/ignore_check.hpp>

// These functions will do nothing.
#define PRINT_PARAM_STRING(A) std::string(" ")
#define PRINT_PARAM_VALUE(A, B) std::string(" ")
#define PRINT_DATASET(A) std::string(" ")
#define PRINT_MODEL(A) std::string(" ")
#define PRINT_CALL(...) std::string(" ")
#define BINDING_IGNORE_CHECK mlpack::bindings::tests::IgnoreCheck

inline void mlpackMain();                                                                                                                                                                                                                        
                                                                                                                                                                                                                                                 
namespace mlpack {
namespace util {

template<typename T>
using Option = mlpack::bindings::tests::TestOption<T>;

}
}

...
```

For the second problem (`programName`) my idea did not work, but I think that yours is closer to the right solution.  Let me comment on that in the diffs...

src/mlpack/bindings/tests/test_option.hpp
```diff
@@ -61,7 +61,8 @@ class TestOption
              const std::string& cppName,
              const bool required = false,
              const bool input = true,
-             const bool noTranspose = false)
+             const bool noTranspose = false,
+             const std::string& programName = "")
```
rcurtin: Adding an extra parameter is definitely the right way to go here.

src/mlpack/core/util/param.hpp
```diff
@@ -15,6 +15,24 @@
...
+#define PROGRAM_NAME
+#define PROGRAM_NAME_SUBSTITUDE programNameSubstitude
+static const std::string programNameSubstitude = "programNameSubstitude";
+#endif
+// The MAIN macro is mlpackMain##PROGRAM_NAME. The goal is to have different
+// names of main procedure for main tests when all method's  main functions
+// are linked into one executable (mlpack_test)
+#define TOKENPASTE1(x, y) x ## y
+#define TOKENPASTE2(x, y) TOKENPASTE1(x, y)
+#define MAIN TOKENPASTE2(mlpackMain, PROGRAM_NAME)
```
rcurtin: If we can declare `mlpackMain()` as `inline` I think we can simplify this greatly... let me know what you think if we adapted this to the following (or something like it---I haven't tested what is below):

```
// PROGRAM_NAME might be defined for some binding types and may need to be passed
// to the individual parameters.  If it isn't defined, we'll pass a dummy empty string to the
// parameter constructor.
#ifndef PROGRAM_NAME
#define PROGRAM_NAME ""
#endif
```

src/mlpack/core/util/param.hpp
```diff
@@ -1012,53 +1030,55 @@ using DatasetInfo = DatasetMapper<IncrementPolicy, std::string>;
   #define PARAM_IN(T, ID, DESC, ALIAS, DEF, REQ) \
       static mlpack::util::Option<T> \
       JOIN(cli_option_dummy_object_in_, __COUNTER__) \
-      (DEF, ID, DESC, ALIAS, #T, REQ, true, false);
+      (DEF, ID, DESC, ALIAS, #T, REQ, true, false, PROGRAM_NAME_SUBSTITUDE);
```
rcurtin: Then here we can just use `PROGRAM_NAME`.

src/mlpack/tests/main_tests/linear_regression_test.cpp
```diff
@@ -0,0 +1,71 @@
...
+ * @file linear_regression_test.cpp
+ * @author Eugene Freyman
+ *
+ * Test mlpackMain() of linear_regression_main.cpp.
+ */
+#include <string>
+
+#define BINDING_TYPE BINDING_TYPE_TEST
+#define PROGRAM_NAME linearRegressionProgramName
+static const std::string linearRegressionProgramName = "LinearRegression";
```
rcurtin: We could even simplify this a little bit---

```
#define PROGRAM_NAME "LinearRegression"
```

src/mlpack/tests/main_tests/pca_test.cpp
```diff
@@ -48,7 +52,7 @@ struct PCATestFixture
   PCATestFixture()
   {
     // Cache in the options for this program.
-    CLI::RestoreSettings(mlpack::bindings::tests::programName);
+    CLI::RestoreSettings(pcaProgramName);
```
rcurtin: This could be replaced with `CLI::RestoreSettings(PROGRAM_NAME)` to avoid the static variable.

evgfreyman:
Thank you for detailed comments on my code. You are right. The inline way looks much more elegant. I'll fix this. 

As for programName there is one more possibility: we can define `static const std::string programName = "pcaProgramName"` at each binding test and use it directly in `PARAM_...` macros. As **static** global variable it will have internal linkage and so we will have a unique copy of this variable at each translation unit (binding test in our case). This way we can get rid of `PROGRAM_NAME` macro. However, I didn't try this approach yet

rcurtin:
Hmm, if the static string approach works I'd be fine with that, but I think either way can work.  Do you want to look into that further or do you want me to do that?

evgfreyman:
I'd like to implement this

evgfreyman:
I thought deeper about `inline void mlpackMain();` This [artice](http://en.cppreference.com/w/cpp/language/inline) sayas that

> There may be more than one definition of an inline function or variable (since C++17) in the program as long as each definition appears in a different translation unit and (for non-static inline functions and variables (since C++17)) **all definitions are identical**


and

> If an inline function or variable (since C++17) with external linkage is defined differently in different translation units, the behavior is undefined. 

So, I guess I should try to make `mlpackMain()` to be static and may be it'll be enough for the code to work even without `inline void mlpackMain();` I'll try these ideas today evening (I live in GMT+2 timezone) and update this PR

rcurtin:
You are right, and when I read more the idea I had is depending on undefined behavior---defining `inline void mlpackMain()` differently in multiple translation units is not supported.  I believe the same is true with static functions.

So I think the macro solution you have is best, but I'm a bit resistant to changing the function `void mlpackMain()` to something that looks like macro magic.  Since this issue only arises with the tests, I would be okay doing something a bit underhanded:

```
#define mlpackMain JOIN(TEST_NAME, mlpackMain)
```

for some `JOIN()` function that puts the two together.  We can assume that the token `mlpackMain` will only appear once in the actual code of a single binding.  What do you think?  Like I said it is a little bit underhanded, but it allows us to solve the problem in a way that doesn't require someone writing bindings to understand more of the system than they need---we're hiding the ugly details in the infrastructure that actually runs the tests.

evgfreyman:
I disagree about `static` functions. An example from [here](http://www.goldsborough.me/c/c++/linker/2016/03/30/19-34-25-internal_and_external_linkage_in_c++/#internal-linkage)(Usage paragraph) shows the case when two static functions with the same name has different definition in each cpp file. I think this is very close to what we want to achieve

> Imagine you have a helper function foo in your file1.hpp which you use in file1.cpp, but then you also have a helper function foo in your file2.hpp which you use in file2.cpp. The first foo does something completely different than the second foo, but you cannot think of a better name for them. So, you can declare them both static. Unless you include both file1.hpp and file2.hpp in some same translation unit, this will hide the respective foos from each other

rcurtin:
Ah, I see what you mean.  I misread the "One Definition Rule" section of http://en.cppreference.com/w/cpp/language/definition and thought it also applied to static functions or members.  However, it seems that it doesn't.

I haven't dug into the actual standard to try and figure out whether, in our case, we could get away with a static function with internal linkage.  But if you want to try it and see if it works (be sure to compile both with and without debugging symbols), and if it does work, then I'll dig through the standard to try and make sure we're not relying on undefined behavior.

evgfreyman:
Although I'm pretty sure that this approach should work (compiling it right now), the real question is whether it makes the project more readable and coherent. I believe you know the answer better than anyone else. So, please advise me if it worth trying. Otherwise I can implement your approach with a macro

evgfreyman:
The solution above is the one that uses only `static` keyword. 

About the code (difference with master branch, not with previous approach): 

- Introduced a new symbol `testName`, it should be defined at each binding test with internal linkage before inclusion `mlpack_main.h`. The name is this because I didn't want it to collide with old `programName` global variable that is still serves its purpose in python bindings. This is, I guess, the next candidate for refactoring.

- Each `void mlpackMain()` became `static void mlpackMain()`

Is specifying DEBUG=ON/OFF at cmake enough to be sure that debugging information included/excluded? If so, it compiles correctly (Travis compiled it in both ways).

Just realized that I didn't check if python binding itself work (I'm suspicious here because I changed all `mlpackMain `functions to be `static `and it seems that python uses them, not sure if it affects them). I'll check this in coming days. 

By the way, are you okay with this approach? I can implement your approach with macros if you find that more suitable or discuss other changes in the propesed solution.

rcurtin:
I double-checked the Python bindings, they work fine.  I think the approach with `static void mlpackMain()` is just fine.  Here's a patch where I've updated the documentation on the bindings:

https://gist.github.com/rcurtin/3b1afaa3bb31dfd21389418d4ab6cb90

I tend to prefer solutions where the user is not presented with a macro, unless we really have to.  With something like `MAIN()`, a user has to wonder what is going on, but with `static void mlpackMain()` I think it's a lot more obvious and the user doesn't have to go digging.  I actually don't really like the Boost Unit Test Framework for this reason, but it's hard for me to imagine another way to do it (and I also don't *really* mind ugly boilerplate in the tests...).  As always if you strongly disagree and have good justification my opinions can be changed, but if not I think `static void mlpackMain()` is fine and let's go with that. :)

I spent some time this afternoon playing with the `programName` variable that's used by the Python bindings, and I am not sure it's possible to refactor that to use the same approach as `testName` without having the user declare an extra macro variable in the program before `mlpack_main.hpp` is included (and I'd prefer to avoid that, I like minimizing required boilerplate where we can).  So I guess we should leave the name as `testName`, but then we should also change the name of the parameter to `CLIOption` and `PyOption` and `TestOption` accordingly, and add documentation saying that it is not used for `CLIOption` or `PyOption`.  What do you think?  Does that sound reasonable?

Making this whole system work right is tricky because we are trying to do things far outside the bounds of "typical" C++... thank you for being willing to dig into its complexities. :)

evgfreyman:
I completely agree with the idea of avoiding macros where possible as well as other C legacy. I like the last solution, I think it's pretty clean. As for naming of option's parameters and comments - I fixed that. Please mark what else should be fixed before merging if any

src/mlpack/methods/sparse_coding/sparse_coding.cpp
```diff
@@ -12,7 +12,6 @@
  */
 #include "sparse_coding.hpp"
 #include <mlpack/core/math/lin_alg.hpp>
-#include <mlpack/core/util/param.hpp>
```
rcurtin: Nice catch :)

evgfreyman:
This was my first experience with an open source contribution and I found it very positive, thank you very much for welcoming and supportive atmosphere. Hope to continue to work on MLPack further. I have some experience with implementing numerical methods in C++ and I'm studying ML at this moment, so I guess I can add value to this project as well as get valuable experience. 

My email is` evg.freyman@gmail.com` please send me all the stuff. I would be glad if you also add me as a contributer - want show this to my friends

As for tests for linear regression. I will implement them. At least it was my original intention. I think it's ok to add them in separate PR.

Thank you again for such a cool start in open source

rcurtin:
Sure, I am happy to help!  I really appreciate the contribution.  :)  I got in touch via email about shipping details for the stickers.

rcurtin:
Thanks again for the contribution! :)

PR #1173, opened 2017-12-18
AMF, remove C++11 alternative branch.

zoq:
Simple adjustment.

### DIFF
--- a/src/mlpack/methods/amf/amf.hpp
+++ b/src/mlpack/methods/amf/amf.hpp
@@ -142,8 +142,7 @@ typedef amf::AMF<amf::SimpleResidueTermination,
                  amf::RandomAcolInitialization<>,
                  amf::NMFALSUpdate> NMFALSFactorizer;
 
-//! Add simple typedefs
-#ifdef MLPACK_USE_CXX11
+//! Convenience typedefs.
 
 /**
  * SVDBatchFactorizer factorizes given matrix V into two matrices W and H by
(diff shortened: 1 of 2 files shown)


zoq:
Looks like it's not a simple adjustment, have to take a closer look.

rcurtin:
Looks like `MLPACK_USE_CXX11` was never actually used completely, so it seems like something strange has been in there.

If I am remembering correctly, the original idea was that you would use `FactorizerType<MatType>` to specify what you wanted.  But pre-C++11, template typedefs were not supported.  So Sumedh chose to define `SVDBatchFactorizer` and `SparseSVDBatchFactorizer`, and when the switch to C++11 was done, a user would use `SVDBatchFactorizer<arma::mat>` or `SVDBatchFactorizer<arma::sp_mat>` instead.

So I think to make this work, all the `SparseSVD...` types need to be replaced in the code with `SVD...<arma::sp_mat>` and others replaced with `SVD...<>` (and probably a default parameter should be added to those template typedefs for `arma::mat`.

Let me know if I can clarify further.  Sorry for the strangeness here...

zoq:
I see, so I guess the question is, can we remove the typedef? At least it's not backwards compatible. Perhaps there is no need to provide backwards compatibility, if we incorporate the change only in mlpack 3.0.

rcurtin:
I think it's fine to remove backward compatibility, since the next release will be mlpack 3.0.0.

PR #1174, opened 2017-12-20
Decision Tree and Preprocess Bindings Tests 

manish7294:
Includes Bindings Tests of mlpackMain() for  decision_tree_test.cpp, preprocess_imputer_main.cpp, preprocess_binarizer_main.cpp and preprocess_split_main.cpp.


src/mlpack/tests/CMakeLists.txt
```diff
@@ -122,7 +122,11 @@ add_executable(mlpack_test
   ub_tree_test.cpp
   union_find_test.cpp
   vantage_point_tree_test.cpp
-  main_tests/pca_test.cpp
+  main_tests/decision_tree_test.cpp
+# main_tests/pca_test.cpp
+# main_tests/preprocess_binarize_test.cpp
+# main_tests/preprocess_imputer_test.cpp
+# main_tests/preprocess_split_test.cpp
```
rcurtin: Now that #1172 is merged, can you update your branch and uncomment these tests?  Thanks!

src/mlpack/methods/preprocess/preprocess_imputer_main.cpp
```diff
@@ -150,8 +124,33 @@ static void mlpackMain()
...
+      {
+        CustomImputation<double> strat(customValue);
+        Imputer<double, MapperType, CustomImputation<double>> imputer(
+            info, strat);
+        imputer.Impute(input, missingValue, dimension);
+      }
+      else
+      {
+        Log::Fatal << "'" <<  strategy << "' imputation strategy does not exist"
+          << endl;
```
rcurtin: This is a really minor style comment, but new lines should be indented two tabs (four spaces) from the previous line.  If you prefer I can handle this (and comments like it) after the merge; just let me know if you'd like me to do that.

src/mlpack/methods/preprocess/preprocess_imputer_main.cpp
```diff
@@ -162,7 +161,33 @@ static void mlpackMain()
 
       for (size_t i : dirtyDimensions)
       {
-        imputer.Impute(input, missingValue, i);
+        if (strategy == "mean")
+        {
+          Imputer<double, MapperType, MeanImputation<double>> imputer(info);
+          imputer.Impute(input, missingValue, i);
```
rcurtin: This works, but I think it might simplify the code a little if you included the loop over `dirtyDimensions` inside the if statements, i.e.

```
          Imputer<double, MapperType, MeanImputation<double>> imputer(info);
          for (size_t i : dirtyDimensions)
            imputer.Impute(input, missingValue, i);
```

What do you think?

src/mlpack/tests/main_tests/decision_tree_test.cpp
```diff
@@ -0,0 +1,168 @@
...
+  SetInputParam("test", std::move(testData));
+
+  mlpackMain();
+
+  // Check that number of output points
+  // are equal to number of input points.
+  BOOST_REQUIRE_EQUAL(CLI::GetParam<arma::Row<size_t>>("predictions").n_cols,
+                      testSize);
+  BOOST_REQUIRE_EQUAL(CLI::GetParam<arma::mat>("probabilities").n_cols,
+                      testSize);
```
rcurtin: You could also check that `n_rows` is equal to the number of classes for the probabilities matrix; I think that would be useful.  `vc2_labels.txt` has 3 different classes in it, so there should be 3 rows.

src/mlpack/tests/main_tests/decision_tree_test.cpp
```diff
@@ -0,0 +1,168 @@
...
+                std::move(CLI::GetParam<DecisionTreeModel>("output_model")));
+
+  mlpackMain();
+
+  // Check that number of output points
+  // are equal to number of input points.
+  BOOST_REQUIRE_EQUAL(CLI::GetParam<arma::Row<size_t>>("predictions").n_cols,
+                      testSize);
+  BOOST_REQUIRE_EQUAL(CLI::GetParam<arma::mat>("probabilities").n_cols,
+                      testSize);
```
rcurtin: I think that you could improve this test a little bit by passing a test matrix in the original call to `mlpackMain()`, storing the output `predictions` and `probabilities` matrices, and then checking the exact values of those matrices against the predictions and probabilities matrices produced by the second call to `mlpackMain()` with the input model.  You could use the `CheckMatrices()` utility function for that.

src/mlpack/tests/main_tests/preprocess_binarize_test.cpp
```diff
@@ -0,0 +1,110 @@
...
+ *
+ * mlpack is free software; you may redistribute it and/or modify it under the
+ * terms of the 3-clause BSD license.  You should have received a copy of the
+ * 3-clause BSD license along with mlpack.  If not, see
+ * http://www.opensource.org/licenses/BSD-3-Clause for more information.
+ */
+#include <string>
+
+#define BINDING_TYPE BINDING_TYPE_TEST
+static const std::string testName = "PreprocessBinarize";
```
rcurtin: You can define `testName` after including `mlpack/core.hpp`, which will obviate the need for including `<string>` at the top.  (I know, it is a minor comment...)

src/mlpack/tests/main_tests/preprocess_binarize_test.cpp
```diff
@@ -0,0 +1,110 @@
...
+/**
+ * Check that input and output have same dimensions.
+ */
+BOOST_AUTO_TEST_CASE(PreprocessBinarizeDimensionTest)
+{
+  // synthetic dataset.
+  arma::mat inputData = arma::randu<arma::mat>(2, 5);
+
+  // Store size of input dataset.
+  size_t input_size  = inputData.n_cols;
```
rcurtin: Can you rename this `inputSize` so that it conforms with the mlpack style guide? (no underscores in variable names)

https://github.com/mlpack/mlpack/wiki/DesignGuidelines#StyleGuidelines

src/mlpack/tests/main_tests/preprocess_binarize_test.cpp
```diff
@@ -0,0 +1,110 @@
...
+
+  SetInputParam("input", std::move(inputData));
+  SetInputParam("threshold", (double) 0.5);
+  SetInputParam("dimension", (int) 6); // Invalid.
+
+  Log::Fatal.ignoreInput = true;
+  BOOST_REQUIRE_THROW(mlpackMain(), std::runtime_error);
+  Log::Fatal.ignoreInput = false;
+}
+
```
rcurtin: There are a couple additional tests I think we could add---let me know what you think.

 - A test to ensure that binarization actually did turn the values into 0 and 1 according to the threshold.
 - A test that ensures all dimensions are binarized when `--dimension` is not specified.

If you don't want to implement these, that's perfectly okay; we can just make a note of it in #1152. :)

src/mlpack/tests/main_tests/preprocess_imputer_test.cpp
```diff
@@ -0,0 +1,138 @@
...
+  data::Load("preprocess_imputer_test.csv", inputData);
+
+  // Store size of input dataset.
+  size_t input_size  = inputData.n_cols;
+
+  // Input custom data points and labels.
+  SetInputParam("input_file", (string) "preprocess_imputer_test.csv");
+  SetInputParam("missing_value", (string) "nan");
+  SetInputParam("strategy", (string) "mean");
+  SetInputParam("output_file", (string) "preprocess_imputer_output_test.csv");
```
rcurtin: Yeah, this binding is a little bit weird, because it's not yet known how we can make this binding work both in Python and from the command line.  So for now it just takes filenames.  The test is just fine, I just wanted to point out why this binding is a little strange.

src/mlpack/tests/main_tests/preprocess_imputer_test.cpp
```diff
@@ -0,0 +1,138 @@
...
+
+  mlpackMain();
+
+  // Now check that the output has desired dimensions.
+  arma::mat outputData;
+  data::Load(CLI::GetParam<string>("output_file"), outputData);
+  BOOST_REQUIRE_EQUAL(outputData.n_cols, input_size);
```
rcurtin: Just to be paranoid we should probably check `n_rows` too.  It would probably be worthwhile to make the imputer test file 2-dimensional (or more) to ensure that multiple dimensions are handled correctly.

src/mlpack/tests/main_tests/preprocess_imputer_test.cpp
```diff
@@ -0,0 +1,138 @@
...
+  data::Load("preprocess_imputer_test.csv", inputData);
+
+  // Store size of input dataset.
+  size_t input_size  = inputData.n_cols;
+  size_t countNaN = 0;
+
+  // Count number of unavailable entries.
+  for (size_t i = 0; i < input_size; i++)
+  {
+    if (std::to_string(inputData(1, i)) == "nan")
```
rcurtin: I think that the issue you are having with the failing Travis build is that this should be `inputData(0, i)`.

src/mlpack/tests/main_tests/preprocess_imputer_test.cpp
```diff
@@ -0,0 +1,138 @@
...
+};
+
+BOOST_FIXTURE_TEST_SUITE(PreprocessImputerMainTest,
+                         PreprocessImputerTestFixture);
+
+/**
+ * Check that input and output have same dimensions
+ * except for listwise_deletion strategy.
+ */
+BOOST_AUTO_TEST_CASE(PreprocessImputerDimensionTest)
```
rcurtin: It looks like this only tests the `mean` strategy, could we also add tests for `median` and `custom`?  Like the binarization tests, we can also make a note in #1152 and do those later---your choice.

src/mlpack/tests/main_tests/preprocess_split_test.cpp
```diff
@@ -0,0 +1,190 @@
...
+                      0);
+  BOOST_REQUIRE_EQUAL(CLI::GetParam<arma::mat>("test").n_cols,
+                      input_size);
+
+  BOOST_REQUIRE_EQUAL(CLI::GetParam<arma::Mat<size_t>>
+                      ("training_labels").n_cols, 0);
+  BOOST_REQUIRE_EQUAL(CLI::GetParam<arma::Mat<size_t>>
+                      ("test_labels").n_cols, labels_size);
+}
+
```
rcurtin: These tests look good; my only suggestion for another test is one where labels are not given.  It's possible to use `mlpack_preprocess_split` while only specifying a matrix with data, and this should work fine also.

src/mlpack/methods/preprocess/preprocess_imputer_main.cpp
manish7294: Yes, that's definitely a better way. It will reduce unnecessary repeative checks.

src/mlpack/tests/main_tests/preprocess_imputer_test.cpp
manish7294: I have added check for all three strategies in one test.

src/mlpack/tests/main_tests/preprocess_binarize_test.cpp
manish7294: If I am not wrong, we are already taking care of above two tests through `binarize_test.cpp`. If you want me to implement them here please just let me know.

src/mlpack/tests/main_tests/preprocess_imputer_test.cpp
manish7294: Thanks for pointing this out! I totally made a childish mistake (°◇°)

src/mlpack/tests/main_tests/decision_tree_test.cpp
manish7294: Done!

src/mlpack/tests/main_tests/decision_tree_test.cpp
manish7294: Done! Thanks for suggesting this.

manish7294:
@rcurtin Please if you have some spare time try to look into this.
I have seen many times that travis build with 

> CMAKE_OPTIONS="-DDEBUG=OFF-DPROFILE=OFF-DPYTHON=/usr/bin/python3"

usually fails at start with following log:

> `The command "sudo rm -f /var/lib/dpkg/lock && sudo add-apt-repository -y ppa:ubuntu-toolchain-r/test" failed and exited with 1 during .`

zoq:
> usually fails at start with following log:
> 
> The command "sudo rm -f /var/lib/dpkg/lock && sudo add-apt-repository -y ppa:ubuntu-toolchain-r/test" failed and exited with 1 during .

One option might be to add ``|| mkdir -p workaround`` to each command to prevent the error and to move on, looks like we add more and more commands to somehow deactivate the automatic update.

src/mlpack/tests/main_tests/preprocess_binarize_test.cpp
rcurtin: It seems to me like `PreprocessBinarizeNegativeDimensionTest` and `PreprocessBinarizeLargerDimensionTest` both test that the `dimension` parameter is valid, and `PreprocessBinarizeDimensionTest` checks if the output shape of the matrix is correct when only one dimension is specified via `dimension`.  So I think there is no test to ensure that binarization did actually happen, and there is no test to check if all dimensions are binarized.  Let me know if I missed something...

Like I said it's ok if you want to leave it out; these can be added or handled some other time---your call.

rcurtin:
Thanks again for the contribution!  It is really nice to have these tests as part of the testing suite.  I made a few tiny style changes in b28857102; let me know if I introduced any problems.

manish7294:
Thanks for your help @rcurtin :)

PR #1178, opened 2017-12-22
Fix randomly failing tests

rcurtin:
I took a look at the nightly matrix build on masterblaster and took a note of all the tests that failed.  Then for each of them I went through and diagnosed whether the failure was random or if it was some other issue.  In every case except one (`regularizedOMP`) the tolerances of the test just needed to be adjusted to account for the random failures.

This is related to #922.

### DIFF
--- a/src/mlpack/tests/cne_test.cpp
+++ b/src/mlpack/tests/cne_test.cpp
@@ -171,41 +171,52 @@ BOOST_AUTO_TEST_CASE(VanillaNetworkWithCNETest)
   data::Load("iris_test_labels.csv", testLabels, true);
   testLabels += 1;
 
-  // Create vanilla network with 4 input, 4 hidden and 3 output nodes.
-  FFN<NegativeLogLikelihood<> > model;
-  model.Add<Linear<> >(trainData.n_rows, 4);
-  model.Add<SigmoidLayer<> >();
-  model.Add<Linear<> >(4, 3);
-  model.Add<LogSoftMax<> >();
-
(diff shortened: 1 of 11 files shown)


src/mlpack/tests/cne_test.cpp
```diff
@@ -127,7 +139,7 @@ BOOST_AUTO_TEST_CASE(CNELogisticRegressionTest)
     testResponses[i] = 1;
   }
 
-  CNE opt(30, 500, 0.2, 0.2, 0.3, 65, -1);
+  CNE opt(200, 10000, 0.2, 0.2, 0.3, 65, -1);
```
zoq: hm, maybe we instead of increase the iterations we could train a few times and break once a single run is successful, in favor of the travis test time; not sure that is necessary here.

rcurtin: I did some simple test comparison; on my system the old test takes ~0.1s whereas the new one takes more like ~0.3s.  That's short enough that I'm not sure it makes sense to change the strategy here.  What do you think?  I can easily change it if you like.

zoq: If the test time doesn't blow up, I don't think we have to change anything, let's wait for the travis timings.

zoq: Okay, about 5 seconds for the CNE test suite, so I agree no need to change the strategy here.

rcurtin:
Great, now let's see if the builds turn more green :)

PR #1179, opened 2017-12-23
Avoid line wrap which made invalid .pc files.

rcurtin:
CMake was including the line wrap.  Thanks to @extrowerk for pointing this one out. :)


extrowerk:
Just let me know if you need any help or assistance.

PR #1180, opened 2017-12-26
LinearRegression binding fixes and tests

evgfreyman:
This PR contains tests for LinearRegression bindings as well as some fixes of binding itself. Aside from `main_tests/linear_regression_test.cpp` the following changes were made 

In `linear_regression_main.cpp`:
- output_predictions became `PARAM_ROW_OUT` (it was `PARAM_COL_OUT` and **never worked** because `LinearRegression` class outputs `rowvec predictions`). **Please confirm if this is correct and there is no need to use a transformation from row to column instead**.
- `std::move` trained model to `output_model` was moved to the end, after prediction proceudre. Earlier we tried to predict responses with using moved away model (and got an error)
- Lambda parameter was never used. Fixed that.
- Added error message if we trying to extract responses from a training data that consist only from 1 row

In `pca_test.cpp`:
- Modified `SetInputParam` in a way that it can accept both rvalue and lvalue. Although I used this ability in LinearRegression tests only, I guess it could help those who start implementing tests by looking at `pca_test.cpp`

I also added myself as a contributer to `COPYRIGHT.txt` and to `core.hpp`

### DIFF
--- a/src/mlpack/methods/linear_regression/linear_regression_main.cpp
+++ b/src/mlpack/methods/linear_regression/linear_regression_main.cpp
@@ -79,7 +79,7 @@ PARAM_MODEL_OUT(LinearRegression, "output_model", "Output LinearRegression "
 PARAM_MATRIX_IN("test", "Matrix containing X' (test regressors).", "T");
 
 // This is the future name of the parameter.
-PARAM_COL_OUT("output_predictions", "If --test_file is specified, this "
+PARAM_ROW_OUT("output_predictions", "If --test_file is specified, this "
     "matrix is where the predicted responses will be saved.", "o");
 
 PARAM_DOUBLE_IN("lambda", "Tikhonov regularization for ridge regression.  If 0,"
@@ -97,11 +97,9 @@ static void mlpackMain()
   rowvec responses;
 
   LinearRegression lr;
-  lr.Lambda() = lambda;
 
-  bool computeModel = false;
-  if (!CLI::HasParam("input_model"))
-    computeModel = true;
+  const bool computeModel = !CLI::HasParam("input_model");
+  const bool computePrediction = CLI::HasParam("test");
 
   // If they specified a model file, we also need a test file or we
   // have nothing to do.
@@ -127,6 +125,11 @@ static void mlpackMain()
     if (!CLI::HasParam("training_responses"))
     {
       // The initial predictors for y, Nx1.
+      if (regressors.n_rows < 2)
+      {
+        Log::Fatal << "Can't get responses from training data "
+            "since it has less than 2 rows" << endl;
+      }
       responses = regressors.row(regressors.n_rows - 1);
       regressors.shed_row(regressors.n_rows - 1);
     }
@@ -139,31 +142,26 @@ static void mlpackMain()
 
       if (responses.n_cols != regressors.n_cols)
       {
-        Log::Fatal << "The responses must have the same number of rows as the "
-            "training set." << endl;
+        Log::Fatal << "The responses must have the same number of columns "
+            "as the training set." << endl;
       }
     }
 
     Timer::Start("regression");
-    lr = LinearRegression(regressors, responses);
+    lr = LinearRegression(regressors, responses, lambda);
     Timer::Stop("regression");
-
-    // Save the parameters.
-    if (CLI::HasParam("output_model"))
-      CLI::GetParam<LinearRegression>("output_model") = std::move(lr);
+  }
+  else
+  {
+    // A model file was passed in, so load it.
+    Timer::Start("load_model");
+    lr = std::move(CLI::GetParam<LinearRegression>("input_model"));
+    Timer::Stop("load_model");
   }
 
   // Did we want to predict, too?
-  if (CLI::HasParam("test"))
+  if (computePrediction)
   {
-    // A model file was passed in, so load it.
-    if (!computeModel)
-    {
-      Timer::Start("load_model");
-      lr = std::move(CLI::GetParam<LinearRegression>("input_model"));
-      Timer::Stop("load_model");
-    }
-
     // Load the test file data.
     Timer::Start("load_test_points");
     mat points = std::move(CLI::GetParam<mat>("test"));
@@ -186,6 +184,10 @@ static void mlpackMain()
 
     // Save predictions.
     if (CLI::HasParam("output_predictions"))
-      CLI::GetParam<vec>("output_predictions") = std::move(predictions);
+      CLI::GetParam<rowvec>("output_predictions") = std::move(predictions);
   }
+
+  // Save the model if needed.
+  if (CLI::HasParam("output_model"))
+    CLI::GetParam<LinearRegression>("output_model") = std::move(lr);
 }
--- a/src/mlpack/tests/main_tests/linear_regression_test.cpp
+++ b/src/mlpack/tests/main_tests/linear_regression_test.cpp
@@ -4,6 +4,7 @@
  *
  * Test mlpackMain() of linear_regression_main.cpp.
  */
+
 #include <string>
 
 #define BINDING_TYPE BINDING_TYPE_TEST
@@ -11,22 +12,13 @@ static const std::string testName = "LinearRegression";
 
 #include <mlpack/core.hpp>
 #include <mlpack/core/util/mlpack_main.hpp>
+#include <mlpack/core/util/test_helper.hpp>
 #include <mlpack/methods/linear_regression/linear_regression_main.cpp>
 
 #include <boost/test/unit_test.hpp>
 #include "../test_tools.hpp"
 
 using namespace mlpack;
-
-// Utility function to set a parameter and mark it as passed,
-// using copy semantics for lvalues and move semantics for rvalues
-template<typename T>
-void SetInputParam(const std::string& name, T&& value)
-{
-  CLI::GetParam<typename std::remove_reference<T>::type>(name)
-    = std::forward<T>(value);
-  CLI::SetPassed(name);
-}
 
 struct LRTestFixture
 {
@@ -54,22 +46,23 @@ BOOST_FIXTURE_TEST_SUITE(LinearRegressionMainTest, LRTestFixture);
 
 /**
  * Training a model with different regularization parameter
- * and ensuring that predictions are different
+ * and ensuring that predictions are different.
  */
 BOOST_AUTO_TEST_CASE(LRDifferentLambdas)
 {
-  // required minimal difference between solutions
+  // A required minimal difference between solutions.
   const double delta = 0.1;
 
-  arma::mat trainX = { 1.0, 2.0, 3.0 };
-  arma::mat testX = { 4.0 };
-  arma::rowvec trainY = { 1.0, 4.0, 9.0 };
+  arma::mat trainX("1.0, 2.0, 3.0");
+  arma::mat testX("4.0");
+  arma::rowvec trainY("1.0, 4.0, 9.0");
+
   SetInputParam("training", trainX);
   SetInputParam("training_responses", trainY);
   SetInputParam("test", testX);
   SetInputParam("lambda", 0.1);
 
-  // first solution
+  // The first solution.
   mlpackMain();
   const double testY1 = CLI::GetParam<arma::rowvec>("output_predictions")(0);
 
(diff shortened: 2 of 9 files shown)


src/mlpack/tests/main_tests/linear_regression_test.cpp
```diff
@@ -18,49 +18,268 @@ static const std::string testName = "LinearRegression";
...
-// Utility function to set a parameter and mark it as passed, using move
-// semantics.
+// Utility function to set a parameter and mark it as passed,
+// using copy semantics for lvalues and move semantics for rvalues
 template<typename T>
 void SetInputParam(const std::string& name, T&& value)
 {
-  CLI::GetParam<T>(name) = std::move(value);
+  CLI::GetParam<typename std::remove_reference<T>::type>(name)
+    = std::forward<T>(value);
```
rcurtin: A minor style comment---I think that this can all fit on one line, I think it is less than 80 characters.  But maybe I am wrong (probably...); still, when you wrap, can you put the `=` on the previous line and indent the new line two tabs (four spaces) from the previous line? :)

src/mlpack/tests/main_tests/linear_regression_test.cpp
```diff
@@ -18,49 +18,268 @@ static const std::string testName = "LinearRegression";
...
+
+BOOST_FIXTURE_TEST_SUITE(LinearRegressionMainTest, LRTestFixture);
+
+/**
+ * Training a model with different regularization parameter
+ * and ensuring that predictions are different
+ */
+BOOST_AUTO_TEST_CASE(LRDifferentLambdas)
+{
+  // required minimal difference between solutions
```
rcurtin: This is a really pedantic comment and I can handle it after merge if you don't want to---comments in mlpack should be complete sentences preferably with punctuation and capitalization: https://github.com/mlpack/mlpack/wiki/DesignGuidelines#comments

src/mlpack/tests/main_tests/linear_regression_test.cpp
```diff
@@ -18,49 +18,268 @@ static const std::string testName = "LinearRegression";
...
+  SetInputParam("test", std::move(testX));
+  SetInputParam("lambda", 1.0);
+
+  // second solution
+  mlpackMain();
+  const double testY2 = CLI::GetParam<arma::rowvec>("output_predictions")(0);
+
+  // second solution has stronger regularization,
+  // so the predicted value should be smaller
+  BOOST_REQUIRE(testY1 - delta > testY2);
```
rcurtin: You can use `BOOST_REQUIRE_GT` here, which will provide slightly nicer output in the event of a failing test.

src/mlpack/tests/main_tests/pca_test.cpp
```diff
@@ -18,21 +18,13 @@ static const std::string testName = "PrincipalComponentAnalysis";
...
-void SetInputParam(const std::string& name, const T& value)
-{
-  CLI::GetParam<T>(name) = value;
-  CLI::SetPassed(name);
-}
-
-// Utility function to set a parameter and mark it as passed, using move
-// semantics.
+// Utility function to set a parameter and mark it as passed,
+// using copy semantics for lvalues and move semantics for rvalues
```
rcurtin: Nice, this is a good improvement. :+1:

Since this will almost certainly be useful in many places, maybe it would be useful to split this into a separate header file?

src/mlpack/methods/linear_regression/linear_regression_main.cpp
```diff
@@ -97,11 +97,9 @@ static void mlpackMain()
   rowvec responses;
 
   LinearRegression lr;
-  lr.Lambda() = lambda;
 
-  bool computeModel = false;
-  if (!CLI::HasParam("input_model"))
-    computeModel = true;
+  const bool computeModel = !CLI::HasParam("input_model");
```
rcurtin: Heh, this is a nice simplification :)

src/mlpack/methods/linear_regression/linear_regression_main.cpp
```diff
@@ -139,31 +142,26 @@ static void mlpackMain()
...
-        Log::Fatal << "The responses must have the same number of rows as the "
-            "training set." << endl;
+        Log::Fatal << "The responses must have the same number of columns "
+            "as the training set." << endl;
       }
     }
 
     Timer::Start("regression");
-    lr = LinearRegression(regressors, responses);
+    lr = LinearRegression(regressors, responses, lambda);
```
rcurtin: Yes, this is another nice catch, it looks like `lambda` would have done nothing before.

src/mlpack/methods/linear_regression/linear_regression_main.cpp
```diff
@@ -186,6 +184,10 @@ static void mlpackMain()
 
     // Save predictions.
     if (CLI::HasParam("output_predictions"))
-      CLI::GetParam<vec>("output_predictions") = std::move(predictions);
+      CLI::GetParam<rowvec>("output_predictions") = std::move(predictions);
   }
+
+  // Save the model if needed.
+  if (CLI::HasParam("output_model"))
+    CLI::GetParam<LinearRegression>("output_model") = std::move(lr);
```
rcurtin: Wow, the model was also not being saved?  The fact that you found so many issues with this single program that I have hand-checked a few times really makes me wonder how many other issues will be found as we add more tests...

src/mlpack/tests/main_tests/linear_regression_test.cpp
```diff
@@ -18,49 +18,268 @@ static const std::string testName = "LinearRegression";
...
+ * and ensuring that predictions are different
+ */
+BOOST_AUTO_TEST_CASE(LRDifferentLambdas)
+{
+  // required minimal difference between solutions
+  const double delta = 0.1;
+
+  arma::mat trainX = { 1.0, 2.0, 3.0 };
+  arma::mat testX = { 4.0 };
+  arma::rowvec trainY = { 1.0, 4.0, 9.0 };
```
rcurtin: Looks like the version of Armadillo on AppVeyor cannot use initialization lists.  If you change this to the following it should fix the AppVeyor issue:

```
arma::mat trainX("1.0 2.0 3.0"); // This might need to be transposed after construction, not sure.
arma::mat testX("4.0");
arma::rowvec trainY("1.0 4.0 9.0");
```

src/mlpack/methods/linear_regression/linear_regression_main.cpp
evgfreyman: The model was being saved in previous version, but it was moved to output parameters before calculating predictions so we couldn't make a prediction and creating a model at the same time

src/mlpack/tests/main_tests/linear_regression_test.cpp
evgfreyman: My code works if we use C++11 (I compiled it locally with Visual Studio), but it seems that Armadillo at AppVeyor build doesn't use

```
#define ARMA_USE_CXX11
```

Do have some agreement on this?

src/mlpack/tests/main_tests/linear_regression_test.cpp
evgfreyman: I really wan't to write a good code. Feel free to criticize me. Thank you for detailed review.

src/mlpack/tests/main_tests/linear_regression_test.cpp
rcurtin: I don't have a particular problem with forcing `ARMA_USE_CXX11` on Visual Studio, but I know that Conrad has had a lot of issues over the years with Visual Studio.  And VS C++11 support was pretty bad before the VS2015 release.  I would say, if it doesn't cause any problems, we could try it---after all, mlpack is using a lot of C++11 features even from the Visual Studio compiler.  mlpack already requires VS2015, so I *think* it will be ok.

I'm a bit concerned when you say that 7 tests are failing, and realized that we don't run the tests on AppVeyor to save time, since we have a limited time budget on AppVeyor.  I'll play with a PR to run the tests (#1188) and see what the results are.

src/mlpack/methods/linear_regression/linear_regression_main.cpp
rcurtin: Ah, right, sorry---I overlooked the part where it was saved previously in the diff. :)

src/mlpack/tests/main_tests/linear_regression_test.cpp
evgfreyman: As for failing tests on Windows, I created an issue #1187 about memory leaks. I'll put there info about which tests are failing as well as about memory leaks

rcurtin: Sounds good---we can handle the failing tests there.  In any case, if you want to force-enable `ARMA_USE_CXX11` on Windows I don't see a problem with it.  Probably it should just be added to `mlpack/prereqs.hpp` before Armadillo is included and that should solve the issue.

evgfreyman: I'm worried about the fact that I didn't find in Armadillo documentation any mention that compiled binary library doesn't depend on `ARMA_USE_CXX11`. 

However, when I tried, it worked fine and binary file seems the same (but I use an old version - 7.800.3). Nonetheless, I'm not sure that it's ok in general to compile Armadillo without this define and use with it because they could add something depending on `ARMA_USE_CXX11` in a cpp file in future releases. This could lead to some very hard memory issues.

My be the best solution is to merge this PR as it is (I removed using initializer_list constructors and now it compiles well on AppVeyor) and put a new task in todo list about enabling Armadillo's C++ features on AppVeyor. However, I think that it's important to enable it as described [here (search for ARMA_USE_CXX11)](http://arma.sourceforge.net/docs.html) and recompile

evgfreyman: Just realized that you are the developer of Armadillo. I guess you know this stuff very well :) I'll implement your suggestions about adding `ARMA_USE_CXX11` in mlpack/prereqs.hpp

rcurtin: Ah, yeah, it looks like 6f213a3d07 actually added `ARMA_USE_CXX11` incorrectly.  Moving that up above the include of `arma_extend.hpp` should fix it just fine.

I think you edited the comment, but it seemed like a concern was that enabling `ARMA_USE_CXX11` would cause ABI incompatibilities for different versions of Armadillo.  The Armadillo wrapper library is only an optional wrapper around LAPACK/BLAS/other support libraries; its reason for existence is so that you can link with `-larmadillo` and not `-lblas -llapack -lother_stuff -lmore_other_stuff ...`.  So it doesn't actually have any Armadillo types in it, and I'm confident enough in the future development of Armadillo that `ARMA_USE_CXX11` won't ever affect the ABI.  I am a contributor to Armadillo, by the way, but it's Conrad who calls the shots. :)

evgfreyman:
I fixed the issue with ARMA_DONT_USE_CXX11 define. Now initializer_list constructors works well in binding tests

rcurtin:
Looks good to me; thanks for taking the time to make the fixes.  I'll go ahead and merge this in 5 days in case anyone else has any comments.

src/mlpack/tests/main_tests/linear_regression_test.cpp
```diff
@@ -4,63 +4,275 @@
...
+  mlpackMain();
+
+  const arma::rowvec testY2 = CLI::GetParam<arma::rowvec>("output_predictions");
+
+  double norm = arma::norm(testY1 - testY2, 2);
+  BOOST_REQUIRE(norm < delta);
+}
+
+/**
+ * Ensuring that response size is is checked.
```
zoq: Oops, looks like there is an extra ``is`` here.

src/mlpack/methods/linear_regression/linear_regression_main.cpp
```diff
@@ -127,6 +125,11 @@ static void mlpackMain()
     if (!CLI::HasParam("training_responses"))
     {
       // The initial predictors for y, Nx1.
+      if (regressors.n_rows < 2)
+      {
+        Log::Fatal << "Can't get responses from training data "
+            "since it has less than 2 rows" << endl;
```
zoq: Minor punctuation issue: I think this should end with ``.``.

rcurtin:
Thanks again for the contribution! :)  I made some minor fixes based on @zoq's suggestions in 01060636.

rcurtin:
I also made some minor changes in a39c6817, including moving `test_helper.hpp` to the `main_tests/` directory (to me it makes more sense to put the testing-related infrastructure in the tests/ directory).  Let me know if there are any problems I caused.

PR #1181, opened 2017-12-29
binding-tests-for-preprocess_split_main-my-first-test-on-mlpack  #1152

wkk91193:
Bindings Tests of mlpackMain() for preprocess_split_main.cpp.

### DIFF
--- a/src/mlpack/tests/main_tests/preprocess_split_test.cpp
+++ b/src/mlpack/tests/main_tests/preprocess_split_test.cpp
@@ -1,0 +1,66 @@
+
+#include <string>
+
+#define BINDING_TYPE BINDING_TYPE_TEST
+static const std::string testName = "PreprocessSplit";
+
+#include <mlpack/core.hpp>
+#include <mlpack/core/util/mlpack_main.hpp>
+#include <mlpack/methods/preprocess/preprocess_split_main.cpp>
+
+#include <boost/test/unit_test.hpp>
+#include "../test_tools.hpp"
+
+using namespace mlpack;
+
+// Utility function to set a parameter and mark it as passed, using copy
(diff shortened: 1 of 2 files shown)


src/mlpack/tests/main_tests/preprocess_split_test.cpp
```diff
@@ -0,0 +1,66 @@
...
+
+#include <boost/test/unit_test.hpp>
+#include "../test_tools.hpp"
+
+using namespace mlpack;
+
+// Utility function to set a parameter and mark it as passed, using copy
+// semantics.
+template<typename T>
+void SetInputParam(const std::string& name, const T& value)
```
evgfreyman: It's not good to override [universal references](https://isocpp.org/blog/2012/11/universal-references-in-c11-scott-meyers) because you will almost always get them invoked instead of other functions. Here for example, this function won't be called if you pass non cost lvalue. I'd recommend using following code instead of both your `SetInputParam `functions:

// Utility function to set a parameter and mark it as passed,
```
// using copy semantics for lvalues and move semantics for rvalues
template<typename T>
void SetInputParam(const std::string& name, T&& value)
{
  CLI::GetParam<typename std::remove_reference<T>::type>(name)
    = std::forward<T>(value);
  CLI::SetPassed(name);
}
```

manish7294:
@wkk91193 I already opened a PR  #1174 for this test.

rcurtin:
Hi @wkk91193, we'll go ahead and use Manish's PR #1174 for this test, since he opened that PR first.  Sorry about that, but thank you regardless for taking the time to put this together.  If you'd like to write a test for a different binding please feel free. :)

PR #1182, opened 2017-12-30
Fixes for python3

rcurtin:
This PR contains a number of small fixes that solve problems that dan_dan pointed out in IRC, mostly having to do with making sure that the bindings use Python3.


zoq:
Do you think it might be useful to setup another travis build for python3?

rcurtin:
Definitely, that would be a great idea.  Let's see if the change I pushed works...

PR #1186, opened 2018-01-01
WIP: Build time reduction

sbrodehl:
This PR works on build time improvements, as discussed in #722.

- [ ] Split up `prereqs.hpp`
- [ ] Get rid of `core.hpp` includes
- [ ] Use `forward declarations` where applicable
- [x] Create precompiled headers if necessary

The TL;DL is: *include only what's necessary*

### prereqs.hpp
For now, `prereqs.hpp` is split up into seperate blocks, to have a finer control of the includes:
- arma include (`arma_inc.hpp`)
- boost include (`boost_inc.hpp`)
- std libs (`std_inc.hpp`)
- common mlpack libs (`common_inc.hpp`)

*For now*, those headers are again used in `prereqs.hpp`, so no functionality should have changed. Later, during the next commits, the idea is to only include headers which are actually needed, e.g. arma / boost only, common mlpack classes and common std classes.

### core.hpp
IMO this header exists for the user, to include basically all of the mlpack core. Internally it should not be used.

### forward declarations
Forward declarations  are kind of tricky in our case, because mlpack uses templates a lot, which are implemented in the header files. Currently code looks like this:
```c++
// foo.hpp
class Foo {
    ...
}
#include "foo_impl.hpp"
```
```c++
// foo_impl.hpp
#include "foo.hpp"
...
```
Correct me if I am wrong, but I think there are only a few cases were we can use forward declarations.

### precompiled headers
At some point we need to use precompiled headers, at least under windows, because a big chunk of the std libraries can be precompiled.
E.g. can use tools like https://github.com/sakra/cotire

----

Another thought: How about splitting the project up into smaller modules, build up libraries and link them together at the end? Maybe we can cache untouched modules and only rebuild those edited? Does that work with travis et al.?


rcurtin:
Hey there,

Glad to see some work on this---it's a difficult issue and the build time is definitely problematic.  A couple of thoughts:

> Correct me if I am wrong, but I think there are only a few cases were we can use forward declarations.

So the way I have found forward declarations most useful is for explicit template instantiations.  Take a look at `load_impl.hpp`, `load.hpp`, and `load.cpp`.  The whole reason those things are split up in that way is so that including `load.hpp` does not include all of boost::spirit (which in turn includes all kinds of other things and increases parse time significantly).  So instead we forward-declare explicit template instantiations of `Load<>()` for the most commonly-used types, and then those instantiations only exist inside of load.cpp and thus boost::spirit is only included once.  If a user needs a not-commonly-used type, then they have to include `load_impl.hpp` to have access to the actual templated definitions.

I've only found this useful to reduce dependencies on boost.  I think that checking through `util/` and using this strategy to reduce (or eliminate) boost includes could be helpful.  I noticed that `boost/program_options.hpp` is included in `gamma_distribution.hpp`, it looks like maybe that could be removed too.

gcc has some options for printing all of the included files (plus a breakdown of the time it took to compile) which I've found useful, via `-ftime-report` and `-M`.

> Another thought: How about splitting the project up into smaller modules, build up libraries and link them together at the end? Maybe we can cache untouched modules and only rebuild those edited? Does that work with travis et al.?

So CMake should already be recompiling only what is necessary, but the tricky part with templates is that very little actually ends up in `libmlpack.so`; most of the machine learning algorithms actually do not.  The best trick I can think of to help is explicit template instantiations, which help us compile an algorithm only once (or a few times) and stuff it into a single .so, instead of compiling it once for every program or binding that uses it.

> At some point we need to use precompiled headers, at least under windows, because a big chunk of the std libraries can be precompiled.

Yeah, I bet that this could provide a significant improvement.  Have you tried using cotire like you linked to?  It looks easy to drop into mlpack (or, well, as "easy" as working with CMake ever is :)).

sbrodehl:
I looked into this a little bit more and what takes the most time is building the cli executables and the bindings, and although we could optimize the includes in `mlpack`, it won't speed up the build/test prozess significantly.

See #1189 for the pch usage.

The mlpack target is currently build in ~2.5 min, and we can cut this in half by using `prereqs.hpp` as pch. It looks like the current *include usage* (include `prereqs.hpp` everywhere) is in favor of this. (I'm not happy with this usage, but for pch it's needed anyway.)

We can also use the same pch during the build of the bindings / cli executables, cotire supports something like this, as documented [(prefix-header-for-multiple-targets)](https://github.com/sakra/cotire/blob/master/MANUAL.md#using-a-generated-prefix-header-for-multiple-targets).
However, in practice, we need to restructure the cmake project a little.
The sources from `bindings/matlab` and `bindings/python` are not required to build the target `mlpack`, right?

For the `methods`, all sources are appended to the parent scope, but the cli executables could be defined at a later stage and use a pch. Here we could precompile `mlpack/prereqs.hpp`, `mlpack/core/util/cli.hpp` and `mlpack/core/util/mlpack_main.hpp`. We need to make sure that this pch is only build once.

For the `tests` we can use `core.hpp` as pch, which works out of the box.

---
How can one reproduce the *Static Code Analysis* done with Jenkins?
Found a [cpp_static.txt](https://github.com/mlpack/mlpack/files/1606522/cpp_static.txt) script in the build tree, is that the one used?

rcurtin:
> The sources from bindings/matlab and bindings/python are not required to build the target mlpack, right?

Yes, nothing in src/mlpack/bindings/ or src/mlpack/tests/ should be in the resulting `libmlpack.so`.

> For the methods, all sources are appended to the parent scope, but the cli executables could be defined at a later stage and use a pch. Here we could precompile mlpack/prereqs.hpp, mlpack/core/util/cli.hpp and mlpack/core/util/mlpack_main.hpp. We need to make sure that this pch is only build once.

I *think* that this would work, at least for the command-line programs.  I think it won't work for the Python bindings (or at least not easily) because those depend on different instances of a variable `programName` in different translation units.  We could adapt that code if needed but it might be tricky.

The Python bindings do take a very long time to build in general.

> How can one reproduce the Static Code Analysis done with Jenkins?

The static code analysis uses PVS-studio, and here's the exact script as run on Jenkins:

```
mkdir -p reports

# Run cppcheck on mlpack source.
curl https://gist.githubusercontent.com/zoq/c33368ced8ff6657b4e1fa2fab8ef13e/raw/af5fa36af96007b090a28f976963f0b79de1aae8/cpp_check.sh > cpp_static.sh
chmod u+x cpp_static.sh
./cpp_static.sh --src . --reports reports/static.xml

# Build mlpack with compile commands.
rm -rf build
mkdir build && cd build
cmake -DDEBUG=ON -DCMAKE_EXPORT_COMPILE_COMMANDS=ON ..
make -j4 >/dev/null
cd ..

# Build PVS utility.
set +x
rm -rf pvs-util
git clone https://github.com/viva64/how-to-use-pvs-studio-free.git pvs-util
cd pvs-util && mkdir build && cd build && cmake -DCMAKE_BUILD_TYPE=Release .. && make -j4 && cd ../../

# Add PVS header.
pvs-util/build/./how-to-use-pvs-studio-free -c 2 src/mlpack >/dev/null
rm -rf pvs-util

# Run static PVS.
cd build
set +x
pvs-studio-analyzer analyze -o project.log -e -j4
set +x
git stash save --keep-index >/dev/null
set +x
git stash clear >/dev/null
cd ..

set +x
curl https://gist.githubusercontent.com/zoq/86485d534d688120b35c04e9d9abf3eb/raw/9a56a650ce54f793729f1f8d37006e0a4183e470/pvs_check.py > pvs_static.py
set +x
plog-converter -a GA:1,2 -t tasklist build/project.log | python pvs_static.py / 2>> reports/static.xml
rm build/project.log
rm pvs_static.py

# Get information about the current PR:
curl -o files.txt https://api.github.com/repos/mlpack/mlpack/pulls/$ghprbPullId/files
grep -o 'filename.*' files.txt | sed -e "s/^filename\": \"//" -e "s/\",//" | uniq | awk '/.cpp/ || /.hpp/' > filenames.txt

set +x
curl https://gist.githubusercontent.com/zoq/9e45d4e13e61d4d4be4767783b3c667b/raw/063be5ba9ad045086e1fb708e1b83d8b957206da/clean_static.py > clean_static.py
python clean_static.py -r reports/static.xml -f filenames.txt
rm clean_static.py

rm -f reports/results.xml
mv reports/static.xml reports/results.xml
```

Probably at some point that should be moved into the jenkins-conf repo, but I don't think that's a huge priority at the moment.

Thanks again for your work on this---reducing the build time has been an outstanding issue since I have been working on mlpack.

rcurtin:
Just to check in, is this waiting on a review from one of us, or is it still a WIP?  If so I think we should close it for inactivity until you have a chance to look into it further (just to reduce the list of open PRs to something more manageable).

sbrodehl:
This PR is now a little bit redundant, because precompiled headers were introduced in #1189 which uses `prereqs.hpp`, so that can't be split up.
`core.hpp` includes are removed in this PR, but burried in other work I guess.
No work was done on forward declarations, so this is still open.

Since one needs to merge the recent changes anyway, I think it is best to close this PR and reopen another if work is done on the matter.

rcurtin:
Sounds good---I'll go ahead and close this one then.  Thank you for looking into it. :+1:

PR #1188, opened 2018-01-02
Have AppVeyor run the tests too

rcurtin:
Let's see what this does...


zoq:
Should be save to use ``set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} /bigobj")`` in release mode as well.

zoq:
Perhaps, you can't enable the test upload inside a PR, so we would have to merge this first. I think this is ready once, all the test commands are removed.

rcurtin:
Sure, we can merge this now if you like and see if the tests magically start working.  At the very least, running the tests at all is an improvement.  I can open an issue after this is merged about getting the test output to upload correctly.

zoq:
Sounds good.

rcurtin:

Issue #183, opened 2014-12-29
MVU does not work


rcurtin:
**Reported by rcurtin on 8 Jun 42063788 04:40 UTC**
I can't get it to successfully unroll even simple swiss roll problems.  I suspect this is related to #188, or, it may be that my implementation of the MVU objective function is incorrect.

PR #1190, opened 2018-01-06
cmake project

sbrodehl:
Add missing files to cmake project.

Rearrange `add_subdirectory` usage so that files are included in parent scope.


sbrodehl:
What's up with that include?
```c++
mlpack/core/math/range_serialization.hpp:11:46: fatal error: mlpack/core/data/serialization.hpp: No such file or directory
compilation terminated.
```

Can't quite figure out how to solve this.

rcurtin:
Huh, looks like `range_serialization.hpp` and `range_serialization.cpp` snuck into the repository somehow.  Those were experiments I was playing with to try and remove all boost::serialization code from being included by `mlpack/core.hpp`.  You can simply delete those files from the repository...

(I'll comment on the rest of the changes sometime soon---nice to see all the places where includes got missed or other minor little issues get fixed...)

sbrodehl:
So I added a comment to the cmake file regarding the MVU copied from the c++ files, but it should be #183, right?

src/mlpack/core/util/CMakeLists.txt
```diff
@@ -11,15 +12,18 @@ set(SOURCES
   cli_deleter.cpp
   cli_impl.hpp
   deprecated.hpp
+  gitversion.hpp
```
rcurtin: This is an auto-generated file; it's actually handled by a block in the top-level CMakeLists.txt:

```
if (GIT_FOUND)
  # Run 'git rev-parse HEAD' to find out if this is a working copy. If the
  # return code is not 0, then it isn't.
  execute_process(COMMAND ${GIT_EXECUTABLE} rev-parse HEAD
      WORKING_DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR}
      OUTPUT_VARIABLE MLPACK_TMP_REV_INFO
      ERROR_VARIABLE MLPACK_TMP_REV_INFO_ERROR
      RESULT_VARIABLE MLPACK_TMP_REV_INFO_RESULT
      OUTPUT_STRIP_TRAILING_WHITESPACE)
  if (${MLPACK_TMP_REV_INFO_RESULT} EQUAL 0)
    set (USING_GIT "YES")
    add_definitions(-DMLPACK_GIT_VERSION)
    include(CMake/CreateGitVersionHeader.cmake)

    add_custom_target(mlpack_gitversion ALL
        COMMAND ${CMAKE_COMMAND} -P CMake/CreateGitVersionHeader.cmake
        WORKING_DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR}
        COMMENT "Updating gitversion.hpp (if necessary)")
  # Add gitversion.hpp to the list of sources.
  set(MLPACK_SRCS ${MLPACK_SRCS}
      "${CMAKE_CURRENT_SOURCE_DIR}/src/mlpack/core/util/gitversion.hpp")
  endif ()
endif ()
```

so there's no need to add it here (since it may not exist if the code is not in a git repository).

src/mlpack/core/util/CMakeLists.txt
```diff
@@ -11,15 +12,18 @@ set(SOURCES
...
   is_std_vector.hpp
   log.hpp
   log.cpp
   mlpack_main.hpp
   nulloutstream.hpp
+  param.hpp
   param_checks.hpp
   param_checks_impl.hpp
   param_data.hpp
+  param_data_impl.hpp
```
rcurtin: This file is actually unused (it's from a previous setup of the parameter handling code) and we can just remove it.

src/mlpack/methods/amf/termination_policies/CMakeLists.txt
```diff
@@ -3,7 +3,7 @@
 set(SOURCES
   simple_residue_termination.hpp
   simple_tolerance_termination.hpp
-  validation_rmse_termination.hpp
+  validation_RMSE_termination.hpp
```
rcurtin: Hmm, none of the other source files in mlpack have capital letters---do you think we could change the file itself to `validation_rmse_termination.hpp` instead?

sbrodehl:
Looks like travis has been misbehaving.

zoq:
@sbrodehl Let me restart the travis job.

rcurtin:
Looks like #1191 will need to be sorted out before Travis will consistently work correctly, but I am ok with merging this since it works for 2/3 configurations on Travis.

PR #1191, opened 2018-01-08
Simplify travis configuration.

rcurtin:
Let's see if this works.  I'll try rebuilding this many times to see if it ever has any intermittent problems.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,10 +10,8 @@ env:
 before_install:
   - sudo systemctl stop apt-daily.service apt-daily.timer
   - sudo systemctl disable apt-daily.service apt-daily.timer
-  - sudo rm -f /var/lib/dpkg/lock && sudo add-apt-repository -y ppa:ubuntu-toolchain-r/test
-  - sudo rm -f /var/lib/dpkg/lock && sudo apt-get update -qq
-  - printenv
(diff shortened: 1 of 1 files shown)


sbrodehl:
Why is it such a pain to set up a travis VM?!

rcurtin:
Yeah, I have been trying to disable the automatic apt updates that run when the VM starts for some time now but not really having any sustained success.

sbrodehl:
How is this `while !(systemctl list-units --all apt-daily.service | grep -q dead); do sleep 1; done` doing?

rcurtin:
Right, exactly, this is my frustration with this issue. :)

The code you wrote ideally should be waiting until the apt-daily service is dead.  Right now I am thinking maybe there are multiple systemd units that are causing issues, so maybe the next run can help expose what is going on.

The response in the Travis issue is a little crazy to me, that 16.04 isn't supported... that was already nearly two years ago...

rcurtin:
I was hoping to restart this job 10 times, but Travis has limits on how many times an individual job should be restarted, so I can't restart it all 10 times.  Still, of the 8 successful restarts, I saw no issues with `/var/lib/dpkg/lock` being locked, so I think that we can go ahead and merge this.  If there are no other comments, I'll merge this in 5 days.

sbrodehl:
Unless travis supports 16.04 this is probably the best way to go.

rcurtin:
They have been saying they are working on 16.04 support for quite a while but 18.04 will be out soon so I think they are a little behind... :)

Issue #513, opened 2016-01-31
Serialization of ann


stereomatchingkiss:
I would like to make the ann support serialization(atleast for the ffn related parts).The optimizers of ann use reference to keep the address of the object, could I change them to pointer or shared_pointer?

zoq:
If you do this, we can't create a default optimizer in the way we do it right now, Because in C++ 'this' (keyword) is not an lvalue, so it isn't possible to create a pointer to this. Correct me if I'm on the wrong track.

stereomatchingkiss:
According to the [answer of stackoverflow](http://stackoverflow.com/questions/5058349/is-it-safe-to-use-the-this-pointer-in-an-initialization-list), it should be safe to that as long as we do not access the member haven't construct yet(including virtual member function, but mlpack do not use any virtual, so I think this would not be an issue). 

What I mean is, change the reference to pointer as following

```
RMSPROP(DecomposableFunctionType* function,
          const double lr = 0.01,
          const double alpha = 0.99,
          const double eps = 1e-8) :
      function(function),
      lr(lr),
      alpha(alpha),
      eps(eps)
```

So we could construct it as 

```
SparseBiasLayer(const size_t outSize,
                  const size_t batchSize,
                  WeightInitRule weightInitRule = WeightInitRule()) :
      outSize(outSize),
      batchSize(batchSize),
      optimizer(new OptimizerType<SparseBiasLayer<OptimizerType,
                                                  WeightInitRule,
                                                  InputDataType,
                                                  OutputDataType>,
                                                  InputDataType>(this)),
      ownsOptimizer(true)


```
Is this the problem you are worrying about?Or we are talking different problems?

ps : I forgot I open the same issue on #495, rcurtin has an [example](https://github.com/mlpack/mlpack/commit/60c81702c432014544351948c091d03deeef6985). If you think  it is ok, I would find a way to serialize it.

zoq:
I see that would solve the issue, you are right. If you make the changes, can you also modify the other ann optimizer classes?

I guess it would be a good idea, to close one of the pull requests and make a reference to the other pull request, what do you think?

rcurtin:
One of the keys of the refactoring I did for serialization is that I attempted to keep the interfaces using references, and not pointers, for the sake of simplicity.  i.e.

```
RMSPROP(DecomposableFunctionType* function,
          const double lr = 0.01,
          const double alpha = 0.99,
          const double eps = 1e-8) :
      function(&function),
[...]

PR #1192, opened 2018-01-11
Reorder data for RNN::Train() and other RNN functions

rcurtin:
This is related to a discussion from long ago, but I am not sure I am finding the right ticket to reference: #513, #831, #541?  Anyway, if I am remembering right, there was some discussion about whether to use `mat` or `cube` to represent input to neural networks.  It's completely possible my views have changed over time and I don't realize the change, but to me now it seems the best way is to have `mat` be the input to FFNs (i.e. one column per point, one row per dimension), but `cube` be the input to RNNs (i.e. one column per point, one row per dimension, one slice per timestep).

This PR implements that support and adds some documentation to the `RNN` class about how the data is expected to be formatted.  There are also a couple of other warning fixes that I noticed and handled.

One of the important reasons for doing this change is that internally, at each time step, the layer expects a matrix of shape (n_dims x n_points).  But RNN data is currently embedded in a matrix of shape ((n_dims * n_timesteps) * n_points), with a single column containing all of the time steps for a given point.  This means that when the batch size is greater than 1 (which it typically will be after #1137) we have to extract a non-contiguous submatrix, which Armadillo will do as a copy.  If we reorder our data into a cube where each timestep is a separate slice, this allows us to avoid that copy.  I also think the cube structure is a easier to understand for a user, but we can change that around if there are other opinions.

This doesn't solve the problem of how to pass sequences of different length to RNNs---you still must either pass them individually, or zero-pad them all into a cube and pass that.  That problem can be solved some other day...

### DIFF
--- a/src/mlpack/core/math/shuffle_data.hpp
+++ b/src/mlpack/core/math/shuffle_data.hpp
@@ -53,7 +53,6 @@ void ShuffleData(const MatType& inputPoints,
   // Generate ordering.
   arma::uvec ordering = arma::shuffle(arma::linspace<arma::uvec>(0,
       inputPoints.n_cols - 1, inputPoints.n_cols));
-//  std::cout << "ordering:\n" << ordering.t();
 
   // Extract coordinate list representation.
   arma::umat locations(2, inputPoints.n_nonzero);
@@ -75,8 +74,7 @@ void ShuffleData(const MatType& inputPoints,
     MatType newOutputPoints(locations, values, inputPoints.n_rows,
         inputPoints.n_cols, true);
     LabelsType newOutputLabels(inputLabels.n_elem);
-    for (size_t i = 0; i < inputLabels.n_elem; ++i)
-      newOutputLabels[ordering[i]] = inputLabels[i];
+    newOutputLabels.cols(ordering) = inputLabels;
 
     outputPoints = std::move(newOutputPoints);
     outputLabels = std::move(newOutputLabels);
@@ -86,8 +84,7 @@ void ShuffleData(const MatType& inputPoints,
     outputPoints = MatType(locations, values, inputPoints.n_rows,
         inputPoints.n_cols, true);
     outputLabels.set_size(inputLabels.n_elem);
-    for (size_t i = 0; i < inputLabels.n_elem; ++i)
-      outputLabels[ordering[i]] = inputLabels[i];
+    outputLabels.cols(ordering) = inputLabels;
   }
 }
 
--- a/src/mlpack/core/math/shuffle_data.hpp
+++ b/src/mlpack/core/math/shuffle_data.hpp
@@ -24,7 +24,8 @@ void ShuffleData(const MatType& inputPoints,
                  const LabelsType& inputLabels,
                  MatType& outputPoints,
                  LabelsType& outputLabels,
-                 const std::enable_if_t<!arma::is_SpMat<MatType>::value>* = 0)
+                 const std::enable_if_t<!arma::is_SpMat<MatType>::value>* = 0,
+                 const std::enable_if_t<!arma::is_Cube<MatType>::value>* = 0)
 {
   // Generate ordering.
   arma::uvec ordering = arma::shuffle(arma::linspace<arma::uvec>(0,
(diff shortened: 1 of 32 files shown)


src/mlpack/core/math/shuffle_data.hpp
```diff
@@ -46,11 +47,13 @@ void ShuffleData(const MatType& inputPoints,
...
                  MatType& outputPoints,
                  LabelsType& outputLabels,
-                 const std::enable_if_t<arma::is_SpMat<MatType>::value>* = 0)
+                 const std::enable_if_t<arma::is_SpMat<MatType>::value>* = 0,
+                 const std::enable_if_t<!arma::is_Cube<MatType>::value>* = 0)
 {
   // Generate ordering.
   arma::uvec ordering = arma::shuffle(arma::linspace<arma::uvec>(0,
       inputPoints.n_cols - 1, inputPoints.n_cols));
+//  std::cout << "ordering:\n" << ordering.t();
```
zoq: Looks like we can remove the debug message here.

src/mlpack/core/math/shuffle_data.hpp
```diff
@@ -67,11 +70,80 @@ void ShuffleData(const MatType& inputPoints,
...
-  outputLabels.set_size(inputLabels.n_elem);
-  for (size_t i = 0; i < inputLabels.n_elem; ++i)
-    outputLabels[ordering[i]] = inputLabels[i];
+  if (&inputPoints == &outputPoints || &inputLabels == &outputLabels)
+  {
+    MatType newOutputPoints(locations, values, inputPoints.n_rows,
+        inputPoints.n_cols, true);
+    LabelsType newOutputLabels(inputLabels.n_elem);
+    for (size_t i = 0; i < inputLabels.n_elem; ++i)
+      newOutputLabels[ordering[i]] = inputLabels[i];
```
zoq: Can you use a non-contiguous view here, to avoid the for loop? It's probably the same.

src/mlpack/tests/math_test.cpp
```diff
@@ -631,6 +631,8 @@ BOOST_AUTO_TEST_CASE(SparseShuffleTest)
     data(0, i) = i;
     labels[i] = i;
   }
+  // This appears to be a necessary workaround for an Armadillo 8 bug.
+  data *= 1.0;
```
zoq: Interesting, perhaps we should update the matrix build, maybe we already test against armadillo 8?

src/mlpack/tests/recurrent_network_test.cpp
```diff
@@ -690,13 +698,23 @@ void DistractedSequenceRecallTestNetwork(
...
       model.Predict(input, output);
-      data::Binarize(output, output, 0.5);
+      for (size_t j = 0; j < output.n_slices; ++j)
+      {
+        arma::mat outputSlice = output.slice(j);
+        data::Binarize(outputSlice, outputSlice, 0.5);
+        output.slice(j) = outputSlice;
+      }
+
+//      std::cout << "label:\n" << label << "\noutput:" << output
```
zoq: Looks like another debug output.

src/mlpack/tests/recurrent_network_test.cpp
```diff
@@ -690,13 +698,23 @@ void DistractedSequenceRecallTestNetwork(
...
+      arma::cube output;
+      arma::cube input(testInput.at(0, i).memptr(), inputSize, 1,
+          testInput.at(0, i).n_elem / inputSize, false, true);
 
       model.Predict(input, output);
-      data::Binarize(output, output, 0.5);
+      for (size_t j = 0; j < output.n_slices; ++j)
+      {
+        arma::mat outputSlice = output.slice(j);
+        data::Binarize(outputSlice, outputSlice, 0.5);
```
zoq: Do you think we should add ``cube`` support, might be helpful to postprocess the data. If you agree, this could be a good beginner task.

src/mlpack/tests/math_test.cpp
rcurtin: I'm in the process of updating the matrix build, but it's not ready yet.  I haven't isolated the Armadillo bug yet, but I'll either open an issue or send a fix to Conrad.

src/mlpack/tests/recurrent_network_test.cpp
rcurtin: Yeah, I think it would be a nice beginner task to add.  I'll go ahead and write up an issue for it.  I think no new code is necessary; just the existing `Binarize()` can be adapted for this case, then this code can be simplified.

src/mlpack/core/math/shuffle_data.hpp
rcurtin: Ah, I didn't think of that---fixed.

src/mlpack/core/math/shuffle_data.hpp
rcurtin: Oops, fixed.

PR #1193, opened 2018-01-12
Bindings Tests

manish7294:
Include bindings tests for random_forest_main.cpp, nbc_main.cpp, decision_stump_main.cpp, perceptron_main.cpp & softmax_regression_main.cpp.


manish7294:
@rcurtin please review this PR. And is there any problem with masterblaster as I am not able to see the failing style checks.

src/mlpack/tests/main_tests/nbc_test.cpp
```diff
@@ -0,0 +1,216 @@
...
+
+  // Input pre-trained model.
+  SetInputParam("input_model",
+                std::move(CLI::GetParam<NBCModel>("output_model")));
+
+  Log::Fatal.ignoreInput = true;
+  BOOST_REQUIRE_THROW(mlpackMain(), std::runtime_error);
+  Log::Fatal.ignoreInput = false;
+}
+
```
rcurtin: We should also add a test for the incremental variance option.  That one can be simple: you can just train with and without incremental variance and ensure the output predictions and probabilities are the same.

src/mlpack/tests/main_tests/perceptron_test.cpp
```diff
@@ -0,0 +1,200 @@
...
+  SetInputParam("test", std::move(testData));
+
+  mlpackMain();
+
+  // Check that number of output points are equal to number of input points.
+  BOOST_REQUIRE_EQUAL(CLI::GetParam<arma::Row<size_t>>("output").n_cols,
+                      testSize);
+
+  // Check output have only single row.
+  BOOST_REQUIRE_EQUAL(CLI::GetParam<arma::Row<size_t>>("output").n_rows, 1);
```
rcurtin: Same thought as the NBC test---we could build two models, then ensure the predictions and probabilities are the same across both.

src/mlpack/tests/main_tests/perceptron_test.cpp
```diff
@@ -0,0 +1,200 @@
...
+  // Delete the last row containing labels from test dataset.
+  testData.shed_row(testData.n_rows - 1);
+
+  size_t testSize = testData.n_cols;
+
+  // Input training data.
+  SetInputParam("training", std::move(inputData));
+
+  // Input test data.
+  SetInputParam("test", std::move(testData));
```
rcurtin: You could just make a copy here, and then you wouldn't need to reload the test set later.  I think this is also true for some other tests you've written in this PR.

src/mlpack/tests/main_tests/random_forest_test.cpp
```diff
@@ -0,0 +1,219 @@
...
+
+  // Input pre-trained model.
+  SetInputParam("input_model",
+                std::move(CLI::GetParam<RandomForestModel>("output_model")));
+
+  Log::Fatal.ignoreInput = true;
+  BOOST_REQUIRE_THROW(mlpackMain(), std::runtime_error);
+  Log::Fatal.ignoreInput = false;
+}
+
```
rcurtin: If you are willing to write some slightly more complicated tests, I think these could be added:

 * Ensure that training accuracy goes up when minimum_leaf_size is decreased: we could train on the vc2 dataset with minimum leaf size of 20, 10, and then 1, and ensure that the training error (which you'd have to calculate, but it's not too hard) is smallest for leaf size 1 and largest for leaf size 20.

 * Ensure that adding more trees decreases test set error (this one might need a little noise in the tolerance since it's not guaranteed).  You could build an RF on the vc2 set with 1 tree and with 5 trees, then ensure the test set accuracy of the 5-tree model is higher than the 1 tree model.

Let me know what you think.  Really here I am not aiming to test the random forest algorithm itself, but I do want to test that the leaf size and number of trees are properly getting passed to the random forest algorithm, and this is the best way I can think of to test that.  Maybe another idea is simply to inspect the RandomForestModel that is produced... that might be easier (I haven't fully thought it through yet).

src/mlpack/tests/main_tests/softmax_regression_test.cpp
```diff
@@ -0,0 +1,289 @@
...
+
+  // Input pre-trained model.
+  SetInputParam("input_model",
+                std::move(CLI::GetParam<SoftmaxRegression>("output_model")));
+
+  Log::Fatal.ignoreInput = true;
+  BOOST_REQUIRE_THROW(mlpackMain(), std::runtime_error);
+  Log::Fatal.ignoreInput = false;
+}
+
```
rcurtin: Looks good, but I have some simple ideas for a couple more tests that can help ensure that the values of lambda, no_intercept, and max_iterations are actually being used:

 * Build two models with different lambda values... ensure that the `SoftmaxRegression` objects returned have different parameters.

 * Build two models with different maximum numbers of iterations... ensure that the `SoftmaxRegression` objects returned have different parameters.

 * Ensure that a `SoftmaxRegression` model with no intercept term has one fewer parameter than one with an intercept term.

Like I said with the random forest tests, these are just to make sure that the parameters are actually properly being passed to the algorithm.  If you'd like to add these tests (or similar tests with the same goal), that would be great; and if not, that's no problem at all, we can make a note of it in the issue.

ShikharJ:
@manish7294 Could you please mention in the PR description as to what tests you'll be writing? Saying this because I think #1199 and this PR may have an overlap.

manish7294:
@ShikharJ your concern is right as this is the second time someone opened a PR for which there already exist a PR since I am working. I will be adding a comment regarding this PR in Issue #1152.

src/mlpack/tests/main_tests/nbc_test.cpp
```diff
@@ -0,0 +1,216 @@
...
+  // Check that number of output points are equal to number of input points.
+  BOOST_REQUIRE_EQUAL(CLI::GetParam<arma::Row<size_t>>("output").n_cols,
+                      testSize);
+  BOOST_REQUIRE_EQUAL(CLI::GetParam<arma::mat>("output_probs").n_cols,
+                      testSize);
+
+  // Check output have only single row.
+  BOOST_REQUIRE_EQUAL(CLI::GetParam<arma::Row<size_t>>("output").n_rows, 1);
+  BOOST_REQUIRE_EQUAL(CLI::GetParam<arma::mat>("output_probs").n_rows, 2);
```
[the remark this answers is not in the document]
manish7294: Done!

src/mlpack/tests/main_tests/nbc_test.cpp
manish7294: Done!

src/mlpack/tests/main_tests/perceptron_test.cpp
manish7294: Done!

src/mlpack/tests/main_tests/perceptron_test.cpp
manish7294: Yup! that's a better way around :)

src/mlpack/tests/main_tests/softmax_regression_test.cpp
manish7294: Added all the above suggested tests. 
Not able to use checkmatrices() here for checking different parameters as log from BOOST_REQUIRE_EQUAL() wasn't getting suppressed.
Earlier thought of something like this:
> Log::Fatal.ignoreInput = true;
>  BOOST_REQUIRE_THROW(CheckMatrices(modelParam,
>        CLI::GetParam<SoftmaxRegression>("output_model").Parameters()),
>        std::exception);
> Log::Fatal.ignoreInput = false;

src/mlpack/tests/main_tests/random_forest_test.cpp
manish7294: Done!

src/mlpack/tests/main_tests/perceptron_test.cpp
rcurtin: Ah, I meant something a little different---you could do 

```
SetInputParam("test", testData);
```

here, and then later

```
SetInputParam("test", std::move(testData));
```

The first call to `SetInputParam()` will make a copy implicitly.

manish7294: Sorry, I misunderstood. Will correct it shortly.

src/mlpack/tests/main_tests/softmax_regression_test.cpp
```diff
@@ -0,0 +1,513 @@
...
+  mlpackMain();
+
+  // Check that initial parameters and final parameters matrix
+  // using saved model are different.
+  bool flag = false;
+  bool* flagPtr = &flag;
+  for (size_t i = 0; i < modelParam.n_elem; ++i)
+  {
+    if ((int) (modelParam[i] * 1e+6) != (int) (CLI::GetParam<SoftmaxRegression>
+                                      ("output_model").Parameters()[i] * 1e+6))
```
rcurtin: This seems like a strange way to do the test; I think it would be sufficient to replace this with just

```
BOOST_REQUIRE_NE(modelParam[i], CLI::GetParam<SoftmaxRegression>("output_model").Parameters()[i]);
```

Let me know if I overlooked something.

src/mlpack/tests/main_tests/softmax_regression_test.cpp
```diff
@@ -0,0 +1,513 @@
...
+  SetInputParam("labels", std::move(labels2));
+  SetInputParam("test", std::move(testData2));
+
+  mlpackMain();
+
+  // Check that initial parameters has 1 more parameter than
+  // final parameters matrix.
+  BOOST_REQUIRE_EQUAL(
+    CLI::GetParam<SoftmaxRegression>("output_model").Parameters().n_cols,
+    modelParam.n_cols + 1);
```
rcurtin: A minor style issue---the wrapped lines here should be indented two tabs from the first line, not one. :)

src/mlpack/tests/main_tests/perceptron_test.cpp
rcurtin: This pattern appears lots of other places in this PR too, so if you can handle it everywhere, I'd really appreciate it.  Thanks! :)

src/mlpack/tests/main_tests/softmax_regression_test.cpp
manish7294: That's definitely  a lot better way.

rcurtin:
Thanks for updating the PR with the new tests.  I left a few comments for how we could improve it further.  Another thing I did was test that the `RandomForestDiffNumTreeTest` was robust, by adding a `math::RandomSeed(std::time(NULL))` as the first line and running the test in a loop:

```
i=0; while(true); do echo $i; bin/mlpack_test -t RandomForestMainTest/RandomForestDiffNumTreeTest 2>&1 | grep -v 'Running ' | grep -v '^[ ]*$' | grep -v 'No errors'; i=$(($i + 1)); sleep 1; done
```

I wanted to check that the test never failed for different random seeds (since it is not guaranteed that a 5-tree model will outperform a 1-tree model).  After 1000+ runs, there were no failures, so it seems the test is robust.

Once you have a chance to handle the other comments, I think that this is ready to merge.  Thank you for your contributions! :)

manish7294:
@rcurtin I have made the required changes. Thanks for continuous guidance :)

PR #1195, opened 2018-01-13
categorical dataset support

### LINKED ISSUE #885: Support for categorical datasets with mlpack_decision_tree program
Right now the `mlpack_decision_tree` program does not support categorical datasets---it only supports numeric datasets.  So I can do

```
$ mlpack_decision_tree -t data.csv -l labels.csv
```

and all the features will be treated as numeric.  But I can't pass, e.g., `data.arff` to the program, which contains categorical features.  We should refactor the program `decision_tree_main.cpp` in the same way as `src/mlpack/methods/hoeffding_tree/hoeffding_tree_main.cpp` so that it accepts categorical datasets also.

A PR to fix this issue should have a tested `decision_tree_main.cpp` program that accepts .arff files too, and properly serializes models with categorical features for later re-use.  So, e.g., the following should work:

```
$ mlpack_decision_tree -t data.arff -l labels.csv -T test_data.arff -P P.csv -p p.csv -M model.bin
$ mlpack_decision_tree -m model.bin -T test_data.arff -P P2.csv -p p2.csv
```

and after that, `P.csv` should be the same as `P2.csv`, and `p.csv` should be the same as `p2.csv`.

haritha1313:
Resolves #885


haritha1313:
@rcurtin I am working on the errors, but I would like some help with serialization. As per comments, support of categorical dataset needs serialization of DatasetInfo. 
Is defining `class DataSet
{
 public:
  DatasetInfo info;

  template<typename Archive>
    void serialize(Archive& ar, const unsigned int /* version */)
    {
      ar & BOOST_SERIALIZATION_NVP(info);
    }
};`
such a class and using its object for datasetinfo the right method?

src/mlpack/tests/data/t1_test.arff
```diff
@@ -0,0 +1,89 @@
...
+%                (.csv) form, and Microsoft Excel (.xls) form.
+% 
+% NOTICE: These data sets may be used freely for scientific, educational and/or
+%         noncommercial purposes, provided suitable acknowledgment is given (by citing
+%         the above-named reference).
+% 
+% Further details concerning the book, including information on statistical software
+% (including sample S-PLUS/R and SAS code), are available at the web site
+% 
+%             http://www.stern.nyu.edu/~jsimonof/AnalCatData
```
rcurtin: Wow, shortening "Analyzing Categorical Data" to "AnalCatData" is pretty awful!  Thank you for not naming this file `analcatdata.arff` or something... :)  However, should we rename it `braziltourism_test.arff` or something?  I'm not sure what the `t1` refers to; let me know if I'm missing something.

src/mlpack/tests/main_tests/decision_tree_test.cpp
```diff
@@ -88,6 +90,49 @@ BOOST_AUTO_TEST_CASE(DecisionTreeOutputDimensionTest)
...
+      testSize);
+  BOOST_REQUIRE_EQUAL(CLI::GetParam<arma::mat>("probabilities").n_cols,
+      testSize);
+
+  // Check number of output rows equals number of classes in case of
+  // probabilities and 1 for predictions.
+  BOOST_REQUIRE_EQUAL(
+      CLI::GetParam<arma::Row<size_t>>("predictions").n_rows, 1);
+  BOOST_REQUIRE_EQUAL(CLI::GetParam<arma::mat>("probabilities").n_rows, 6);
+}
```
rcurtin: Can I ask you to add one more test please?  I want to make sure that the DatasetInfo mappings are properly loaded for saved models.

So the test could build a model on some categorical data in one run and store predictions on test data, and then also store the `output_model`.  Then the stored model could be given as `input_model` with a test set, and we should ensure that the predictions and class probabilities are identical with both runs (you can use the `CheckMatrices()` function for that).

Let me know what you think or if I can clarify what I mean any further.

rcurtin:
Ah, sorry, I did not respond to your comment.  I would suggest serializing the `DatasetInfo` as a member of the `DecisionTreeModel` class.

src/mlpack/tests/data/t1_test.arff
haritha1313: Sure, I'll do the renaming. t1 was just a random name.

src/mlpack/tests/main_tests/decision_tree_test.cpp
haritha1313: I did try this manually as per your description in the issue itself and it is working as desired. I agree, a proper test will be good though. I'll work on it. :+1:

rcurtin:
Thanks for the great contribution---I'll go ahead and merge it now since it has two approvals. :+1:

PR #1197, opened 2018-01-16
Added a new macro to test if a function with a given name exists,

scrcro:
irrespective of the parameters or the return type.


zoq:
Thanks for looking into the issue, would be great if you could write a test; the existing SFINAE test: https://github.com/mlpack/mlpack/blob/master/src/mlpack/tests/sfinae_test.cpp should be helpful here.

src/mlpack/core/util/sfinae_utility.hpp
```diff
@@ -46,56 +46,56 @@ struct MethodFormDetector;
...
-  void operator()(MethodForm<Class, T1, T2, T3, T4, T5, T6>);
+  void operator() ( MethodForm<Class, T1, T2, T3, T4, T5, T6> );
 };
 
 template<typename Class, template<typename...> class MethodForm>
 struct MethodFormDetector<Class, MethodForm, 7>
 {
   template<class T1, class T2, class T3, class T4, class T5, class T6, class T7>
-  void operator()(MethodForm<Class, T1, T2, T3, T4, T5, T6, T7>);
+  void operator() ( MethodForm<Class, T1, T2, T3, T4, T5, T6, T7> );
```
rcurtin: Not sure I see the need to add the extra spaces here, could we revert this bit please?

src/mlpack/core/util/sfinae_utility.hpp
```diff
@@ -182,6 +182,45 @@ struct NAME                                                                    \
...
+template < typename T >                                                                       \
+struct has_member_function                                                                    \
+{                                                                                             \
+  template < typename Q = T >                                                                 \
+  static typename                                                                             \
+  std::enable_if< std::is_member_function_pointer< decltype( &Q::FUNC ) >::value, int >::type \
+  f( int t )                                                                                  \
+  {                                                                                           \
+    return 1;                                                                                 \
+  }                                                                                           \
```
rcurtin: I think you could one-line this to save a little space:

```
f(int t) { return 1; }
```

src/mlpack/core/util/sfinae_utility.hpp
```diff
@@ -182,6 +182,45 @@ struct NAME                                                                    \
...
+    return 0;                                                                                 \
+  }                                                                                           \
+                                                                                              \
+  static const bool value = sizeof( f< T >( 0 ) ) != sizeof( char );                          \
+};                                                                                            \
+                                                                                              \
+template < typename T >                                                                       \
+struct NAME                                                                                   \
+{                                                                                             \
+  static const bool value = has_member_function< T >::value;                                  \
```
rcurtin: Do you think you could simplify this a bit by removing the `has_member_function` struct entirely?  Maybe this works:

```
static const bool value = std::is_member_function_pointer<decltype(&T::FUNC)>::value;
```

scrcro:
Thanks for looking at it. I may be able to get rid of the structure. Let me
write up some tests.

zoq:
Can you take a look at the style issues pointed out by the "Style Checks" build? Looks like it's mainly removing the extra spaces between ``( WORD )``. For more information about the style you can take a look at: https://github.com/mlpack/mlpack/wiki/DesignGuidelines.

src/mlpack/core/util/sfinae_utility.hpp
```diff
@@ -182,6 +182,32 @@ struct NAME                                                                    \
...
+{                                                                                             \
+  template < typename Q = T >                                                                 \
+  static typename                                                                             \
+  std::enable_if< std::is_member_function_pointer< decltype( &Q::FUNC ) >::value, int >::type \
+  f( int t ) { return 1;}                                                                     \
+                                                                                              \
+  template < typename Q = T >                                                                 \
+  static char f( char t ) { return 0; }                                                       \
+                                                                                              \
+  static const bool value = sizeof( f< T >( 0 ) ) != sizeof( char );                          \
```
rcurtin: Hm, I think the full definition of this struct can be simplified further... let me know if it doesn't work:

```
template<typename T>
struct NAME
{
  static const bool value = std::is_member_function_pointer<decltype(&T::FUNC)>::value;
};
```

scrcro:
If T does not have FUNC defined at all this will  lead to compilation error
not just substitution error , I think...

rcurtin:
Ah, yeah, I think you are right, sorry about that.  In that case I think once the style checks are fixed, this is ready to merge from my side.

src/mlpack/core/util/sfinae_utility.hpp
```diff
@@ -182,6 +182,34 @@ struct NAME                                                                    \
       WithGreaterOrEqualNumberOfAdditionalArgs<MinN>::value;                   \
 };
 
+/**
+ * Constructs a template structure, which will define a boolean static variable
```
zoq: Super pedantic style issue; this looks a little bit odd, do you think we could go with:

```c++
/**
 * Constructs a template structure, which will define a boolean static
 * variable, to true if the passed template parameter, has a member function
 * with the specified name. The check does not care about the signature or the
 * function parameters
```

src/mlpack/core/util/sfinae_utility.hpp
```diff
@@ -182,6 +182,34 @@ struct NAME                                                                    \
...
+ * ,to true if the passed template parameter, has a member function with the
+ * specified name. The check does not care about the signature or the function
+ * parameters
+ *
+ * @param FUNC the name of the function, whose existence is to be detected
```
zoq: Another pedantic style issue; do you mind to use the correct punctuation here, e.g. end with a ``.``.

src/mlpack/core/util/sfinae_utility.hpp
```diff
@@ -182,6 +182,34 @@ struct NAME                                                                    \
...
+  template <typename Q = T>                                                  \
+  static typename                                                            \
+  std::enable_if<std::is_member_function_pointer<decltype(&Q::FUNC)>::value, \
+                 int>::type                                                  \
+  f(int t) { return 1;}                                                      \
+                                                                             \
+  template <typename Q = T>                                                  \
+  static char f(char t) { return 0; }                                        \
+                                                                             \
+  static const bool value = sizeof( f<T>(0) ) != sizeof(char);               \
```
zoq: Do you mind to remove the surrounding spaces here: ``f<T>(0)``.

src/mlpack/tests/sfinae_test.cpp
```diff
@@ -134,4 +159,12 @@ BOOST_AUTO_TEST_CASE(HasMethodFormTest)
   static_assert(HasTemplatedM<B, MForm4, 1>::value, "value should be true");
 }
 
+BOOST_AUTO_TEST_CASE(HasMethodNameTest)
```
zoq: Can you add a simple test description, as done for the other test cases?

src/mlpack/core/util/sfinae_utility.hpp
scrcro: Sure. I assume you meant that each subsequent line should be longer or equal.

src/mlpack/core/util/sfinae_utility.hpp
scrcro: Yes. Sorry that was a bad miss.

src/mlpack/core/util/sfinae_utility.hpp
scrcro: Done that too. Thanks for pointing out.

src/mlpack/tests/sfinae_test.cpp
scrcro: Added the description.

scrcro:
Thank you. I also checked the macro by changing HasModel and the respective visitors, where it was being used. Works fine. 
For the custom layer types too, I made the changes to see if that works, it compiles alright. Though I have questions regarding some classes whether they should be also support custom layers.

rcurtin:
Ok, good to know.  Do you want to continue by changing `HasModel` and related code in this PR, or in another PR?  Either way is fine with me.

I'm happy to try and help answer any questions you might have, feel free to ask. :)

scrcro:
Sure, I would be glad to do that. I can continue with the same PR. 
While looking at the code I found that these classes perhaps needs an additional parameter to support custom layers.
`AddMerge`,`AddVisitor`, `Concat`, `CopyVisitor`, `Recurrent`, `Sequential`. These will need an additional template parameter like `typename... CustomLayers`.
Not so sure about `VRClassReward`. I am guessing this does not need to support custom layers.
Please suggest.

zoq:
> AddMerge,AddVisitor, Concat, CopyVisitor, Recurrent, Sequential. These will need an additional template parameter like typename... CustomLayers.

Agreed.

> Not so sure about VRClassReward. I am guessing this does not need to support custom layers.

Once  ``AddVisitor`` is adjusted (same modification as for the Model check) we should be able to add various layer. Once that is done, ``VRClassReward`` should be just fine.

scrcro:
Sure. Happy to contribute.
I will proceed with the other changes too. Any suggestion about writing tests for the custom layers? I think both RNN and FFN needs to be tested. What about Recurrent? Are there existing tests that I can look into?

zoq:
You can take a look at: https://github.com/mlpack/mlpack/blob/master/src/mlpack/tests/feedforward_network_test.cpp, https://github.com/mlpack/mlpack/blob/master/src/mlpack/tests/feedforward_network_test.cpp and https://github.com/mlpack/mlpack/blob/master/src/mlpack/tests/ann_layer_test.cpp. As simple test could compare the network parameter of two network: one with the already implemented layer and another one that uses a custom layer that replicates e.g. the sigmoid function.

scrcro:
Thank you, let me take a look. 
On a different note, there are a lot of files that need to change. Do you want me to submit everything together, or in phases?

zoq:
I would split it up into smaller parts, makes the review easier and we can merge faster. And we can get this stickers shipped 😄 . Let me know what you think.

scrcro:
:) I figured as much. Added the files with the changes to HasModel checks. This should compile as is, though not yet support the custom layers.

rcurtin:
The changes look good to me; happy to see we are closer to supporting custom layers. :)

scrcro:
A couple of questions:
a. I wanted to test the custom layers by modifying the current `FeedForwardNetworkTest`. In the `BuildVanillaNetwork` method I wanted to change the third layer to be a custom layer, but it will mimic what `Linear` layer does today. Same could be done for `RecurrentNetworkTest` too. Does that sound good? 
b. If the above is fine, I am thinking of writing a new class in the test source, which will inherit from `Linear`. Is that within the boundaries of current design guidelines? This is for test only.

rcurtin:
I think that sounds fine for testing, but I would avoid using inheritance---ideally, we want users to be able to create an entirely new class and use it as a custom layer, so we should do that.  The test only really needs to check if the network compiles, so there's no need for the layer to work correctly (you could just have a "fake" `Forward()`, `Backward()`, and `Gradient()` implementation).  That should make the tests easier to implement if you do that.

zoq:
Agreed a "fake" layer sounds good, I think since we use the tests as examples we should make sure it works without inheritance.

scrcro:
Alright, then how about creating a custom layer ala SigmoidLayer using a different activation function?  Or just use one of the existing activation function and create a custom layer?

zoq:
Using a activation function as a layer sounds like a good idea. If you think replicating the ``SigmoidLayer`` is the easiest way to do that, go for it. I guess you could also just implement something that returns zero, so we could model something that produces a zero output if the layer is incorporated.

scrcro:
I submitted my changes, but I guess there are some conflicts. I do not understand the conflicts from line 204 to 229. Can you please take a look?

rcurtin:
@mlpack-jenkins test this

rcurtin:
It looks like the conflict is caused by the change of all RNN parameters from `arma::mat` to `arma::cube`.  The specific section you're referencing looks like you can entirely take the master branch---the `SinglePredict()` function has been removed.

scrcro:
I submitted the changes. Looks like one of the build failed while the others succeeded. Appears strange to me. What do you suggest?

rcurtin:
It looks like the travis job when compiling with debugging symbols is failing; have you taken a look at the log?

When you configure with CMake you can specify `-DDEBUG=ON`; that should allow you to reproduce the failing test locally.

scrcro:
Alright. Thank you. Let me try with debug ON.

scrcro:
I am done with the changes. There is one failure in `Sequential` class' constructor. This is asking for initializing the member variables in the constructor. Thoughts....

zoq:
Adding width(0) and height(0) to the constructor list should solve the problem.

src/mlpack/tests/customlayer.hpp
```diff
@@ -0,0 +1,24 @@
+#ifndef CUSTOMLAYER_HPP
+#define CUSTOMLAYER_HPP
```
rcurtin: Just to match the rest of the mlpack style, I'd consider naming this `custom_layer.hpp` and use the header guard `MLPACK_TESTS_CUSTOM_LAYER_HPP`.  Also, we should add a comment at the start of the file with the filename, what it contains, who wrote it, etc., and the license.  You can see any of the other files in mlpack for an idea. :)

src/mlpack/tests/customlayer.hpp
```diff
@@ -0,0 +1,24 @@
...
+
+#include <mlpack/prereqs.hpp>
+#include <mlpack/methods/ann/layer/layer.hpp>
+
+
+namespace mlpack {
+namespace ann{
```
rcurtin: I guess the style checker did not catch it, but if you can put a space between `ann` and `{` I'd appreciate it. :)

src/mlpack/methods/ann/layer/layer_types.hpp
```diff
@@ -118,7 +139,8 @@ using LayerTypes = boost::variant<
     ReinforceNormal<arma::mat, arma::mat>*,
     Select<arma::mat, arma::mat>*,
     Sequential<arma::mat, arma::mat>*,
-    VRClassReward<arma::mat, arma::mat>*
+    VRClassReward<arma::mat, arma::mat>*,
+    CustomLayers*...
```
rcurtin: Hm, I am not sure on this comment, but does the code `CustomLayers*...` properly expand into a pointer type for every type in `CustomLayers`?  If you are not sure, we could add another test to `feedforward_network_test.cpp` to test with two custom layers, but if you are sure that the code above is correct, then no need to do anything.

src/mlpack/tests/customlayer.hpp
scrcro: No problem, will make the changes. Thanks for looking into it. I removed the include to specifically show that the visitors are not dependent on the LayerTypes. Hopefully it will also help reduce triggering compilations when not required. I need to run some local tests for the layer types that you mentioned. Will send an update when ready.

src/mlpack/methods/ann/layer/layer_types.hpp
scrcro: I just tested with a very simple class like this:
```
class ForwardVisitor : public boost::static_visitor<>
{
  private:
  arma::mat input;
  arma::mat output;

  public:
  ForwardVisitor( arma::mat& input, arma::mat& output )
  : input( input ), output( output )
  {
  }
  template < typename T >
  void operator()( T* operand )
  {
    operand->Forward( input, output );
  }
};

template < typename... Args >
using LayerTypes = boost::variant< BaseLayer< double, double, double >*, Args*... >;

template < typename... Args >
class CustomFFN
{
  public:
  template < class LayerType, class... T >
  void Add( T... args )
  {
    network.push_back( new LayerType( args... ) );
  }

  void Add( LayerTypes<> layer ) { network.push_back( layer ); }

  void apply( ForwardVisitor& v )
  {
    std::for_each( network.begin(), network.end(), boost::apply_visitor( v ) );
  }

  private:
  std::vector< LayerTypes< Args... > > network;
};
```
And then use the following to test:
```
template < typename Activation, typename InputType = double, typename OutputType = double >
class EmbeddedLayer : public BaseLayer< double >
{
  public:
  template < typename eT >
  void Forward( eT& input, eT& output )
  {
    std::cout << "EmbeddedLayer" << std::endl;
  }

  size_t model;
};

template < typename Activation, typename InputType = double, typename OutputType = double >
class NewCustomLayer
{
  public:
  template < typename eT >
  void Forward( eT& input, eT& output )
  {
    std::cout << "NewCustomLayer" << std::endl;
  }

  size_t model;
};

int main()
{
  CustomFFN< EmbeddedLayer< double >, NewCustomLayer< double > > v;

  v.Add< EmbeddedLayer< double > >();
  v.Add< NewCustomLayer< double > >();
  arma::mat input;
  arma::mat output;
  ForwardVisitor visitor ( input, output );

  v.apply ( visitor );
}
```
The output is correctly printed.

rcurtin: Perfect, thanks for looking into that. :+1:

scrcro:
I am finally done with the changes. Please have a look. I copied the ReberGrammarTest function, for testing RNN with a custom layer. My earlier attempt was to use a template version. But that does not seem to compile with GCC 5 or 6.

src/mlpack/tests/recurrent_network_test.cpp
```diff
@@ -904,4 +907,183 @@ BOOST_AUTO_TEST_CASE(SerializationTest)
...
+ */
+void ReberGrammarTestCustomNetwork(const size_t hiddenSize = 4,
+                                   const bool recursive = false,
+                                   const size_t averageRecursion = 3,
+                                   const size_t maxRecursion = 5,
+                                   const size_t iterations = 10)
+{
+  // Reber state transition matrix. (The last two columns are the indices to the
+  // next path).
+  arma::Mat<char> transitions;
```
zoq: I think what we could do to shorten the test file is to write a ``GenerateReberGrammarData`` function and use that in both tests, what do you think?

src/mlpack/tests/custom_layer.hpp
```diff
@@ -0,0 +1,36 @@
...
+#define MLPACK_TESTS_CUSTOM_LAYER_HPP
+
+#include <mlpack/prereqs.hpp>
+#include <mlpack/methods/ann/layer/layer.hpp>
+
+
+namespace mlpack {
+namespace ann {
+  /**
+   * Standard hyperbolic tangent layer.
```
zoq: Looks like this is using the sigmoid function.

src/mlpack/tests/recurrent_network_test.cpp
scrcro: Yeah I thought of it and then thought against it because I did not want to shake things up too much. But I will give it a shot.

src/mlpack/tests/recurrent_network_test.cpp
```diff
@@ -456,11 +457,11 @@ void ReberGrammarTestNetwork(const size_t hiddenSize = 4,
...
-    RNN<MeanSquaredError<>, RandomInitialization, Args...> model(5);
-    model.template Add<Linear<> >(inputSize, hiddenSize);
-    model.template Add<RecurrentLayerType>(hiddenSize, hiddenSize);
-    model.template Add<Linear<> >(hiddenSize, outputSize);
-    model.template Add<OutputLayerType>();
+    RNN<MeanSquaredError<> > model(5);
+    model.Add<Linear<> >(inputSize, hiddenSize);
+    model.Add<RecurrentLayerType>(hiddenSize, hiddenSize);
+    model.Add<Linear<> >(hiddenSize, outputSize);
+    model.Add<SigmoidLayer<> >();
```
rcurtin: It should be possible to make the templatization work on gcc 5 and 6; this shouldn't be too much of an issue.  Maybe this is a C++ standard issue, so if you are interested in debugging further you might be able to reproduce the issue on your version of gcc by additionally making sure that the `-std=c++11` flag is given to the compiler.  I think this can be done by setting `CMAKE_CXX_FLAGS="-std=c++11"` when you configure.

scrcro: I tried those. Problem is the same code compiled with GCC-7 and Clang-5 but not with lower versions of GCC. 
I tried to add a variadic template parameter to the `BuildNetwork` function and add one more template parameter for the output layer type. The variadic parameter is necessary because we cannot add the same type in `LayerTypes` more than once.  
I will investigate some more the cause of the error. Something to do with template substitution. Perhaps a GCC bug, which cannot handle the variadic template properly. But I thought it was not necessary to hold up this work for that issue. I will create a new pull request if I find a more elegant solution.

rcurtin: Agreed, no need to hold up this PR.  We can return to the issue later if you figure something out.  I think what you had written was valid C++11 and older versions should support it too.  It's possible it could be a compiler bug; we have encountered those before.

scrcro:
The static code analysis and one of the CI build failed, but perhaps due to some issue in the env. Any idea what went wrong here?

rcurtin:
https://github.com/viva64/how-to-use-pvs-studio-free/issues/10

scrcro:
Thank you. Is there anything to be done for the CI build failure?

zoq:
https://github.com/mlpack/mlpack/pull/1217 should fix the Travis issue, let me restart the build.

zoq:

scrcro:
Thank you! I am done with the changes. Please have a look.

zoq:

scrcro:
Awesome! Glad to contribute. Look forward to add more.

src/mlpack/tests/feedforward_network_test.cpp
```diff
@@ -567,4 +568,55 @@ BOOST_AUTO_TEST_CASE(SerializationTest)
...
+        arma::max(testLabelsTemp.col(i)) == testLabelsTemp.col(i), 1)) + 1;
+  }
+
+  FFN<NegativeLogLikelihood<>, RandomInitialization, CustomLayer<> > model;
+  model.Add<Linear<> >(trainData.n_rows, 8);
+  model.Add<CustomLayer<> >();
+  model.Add<Linear<> >(8, 3);
+  model.Add<LogSoftMax<> >();
+
+  RMSProp opt(0.01, 32, 0.88, 1e-8, 10 * trainData.n_cols, -1);
```
zoq: I think since, we just like to see if this compiles without any errors, we can decrease the number of training iterations, e.g. 5 should be just fine: ``RMSProp opt(0.01, 32, 0.88, 1e-8, 15, -1);``, to keep the test time low.

scrcro:
Thanks I am done. The CI build failed because I think there is some problem in the env. I updated my name and the email id.

zoq:
Thanks, with two approvals I will go ahead and merge this now.

PR #1198, opened 2018-01-16
making label parameter optional

haritha1313:
hoeffding_tree didn't support using last column of dataset as labels. This PR makes label parameter optional and accepts last column of dataset as labels if it is not provided explicitly.

### DIFF
--- a/src/mlpack/methods/hoeffding_trees/hoeffding_tree_main.cpp
+++ b/src/mlpack/methods/hoeffding_trees/hoeffding_tree_main.cpp
@@ -38,7 +38,7 @@ PROGRAM_INFO("Hoeffding trees",
     " parameters, respectively. Optionally, if " +
     PRINT_PARAM_STRING("labels") + " is not specified, the labels are assumed "
     "to be the last dimension of the training dataset."
-    "\n\n" 
+    "\n\n"
     "The training may be performed in batch mode "
     "(like a typical decision tree algorithm) by specifying the " +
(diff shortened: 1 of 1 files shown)


src/mlpack/methods/hoeffding_trees/hoeffding_tree_main.cpp
```diff
@@ -182,7 +183,19 @@ static void mlpackMain()
...
           << i << "." << endl;
 
-    labels = CLI::GetParam<arma::Row<size_t>>("labels");
+    if (CLI::HasParam("labels"))
+    {
+      labels = std::move(CLI::GetParam<arma::Row<size_t>>("labels"));
+    }
+    else
+    {
+      // Extract the labels as the last
```
zoq: Pedantic spelling/style issue; can you check the punctuation. Also, should we use ``of the training set`` here? Let me know what you think.

haritha1313: Sure, I'll make the necessary edit.

haritha1313:
@rcurtin If there is a need of binding tests for hoeffding trees I can work on it once #1195 is completed.

manish7294:
@haritha1313 you could work on hoeffding_tree bindings tests if you want to but just make a note of that in  #1152.

rcurtin:
@haritha1313: sure, if you want to work on that I think that's just fine.  Thanks again for the contribution! :)

PR #1199, opened 2018-01-16
Binding Tests 

KARTHEEKCIC:
Adding Bindings tests for perceptron_main.cpp w.r.t. the issue #1152


KARTHEEKCIC:
@rcurtin. I saw that #1193 already contains binding tests for perceptron but there are some tests which are missing there and I have covered them. So should I keep this PR or add them there.

rcurtin:
You are right, there are some tests here that aren't in the other PR.  Let's merge #1193 first, then take the tests you've written that are different here and incorporate them in.  Thanks for taking the time to write these tests!  I am looking forward to merging them.

KARTHEEKCIC:
Thanks. I will wait for it.

KARTHEEKCIC:
@rcurtin. I have pushed binding tests for logistic regression here by mistake. So should I continue it here or should I make a new PR?

manish7294:
@KARTHEEKCIC It's probably okay to have them here, just update the info of PR.

rcurtin:
That's just fine, we can do both the new perceptron tests and the logistic regression tests in this PR.  Just let me know when they are ready for me to review.

KARTHEEKCIC:
Ok. I have made the changes and you can now review it. Let me know if any changes need to be made.

manish7294:
@KARTHEEKCIC please add perception test in CMake so that it can be build.

KARTHEEKCIC:
Done. Thanks for reminding.

rcurtin:
It looks like there are some test failures, could you please debug those before I review this?  You can run only the perceptron tests locally:

```
bin/mlpack_test -t PerceptronMainTest
```

KARTHEEKCIC:
@rcurtin. I have resolved errors. But still some test cases are failing because there are no checks for them in perceptron_main.cpp. Should I open I PR adding the checks in the perceptron_main.cpp. The following tests are failing - 
1) Checking the Absence of responses
2) Ensuring Response size is checked

rcurtin:
I think it would be reasonable to just add the checks to perceptron_main.cpp in this PR.

KARTHEEKCIC:
@rcurtin. I have made the changes and fixed all the errors. You can now review them.

rcurtin:
It looks like maybe you forgot to add the training data; the Travis build fails with the error:

```
/home/travis/build/mlpack/mlpack/src/mlpack/tests/main_tests/perceptron_test.cpp(85): fatal error in "PerceptronReTrainWithWrongClasses": Could not load the train data train_data_3_classes.csv
```

For what it's worth, if you need a 3-class dataset, the vc2 dataset does that just fine and it's already in the repository (`vc2.csv`).

src/mlpack/tests/main_tests/logistic_regression_test.cpp
```diff
@@ -0,0 +1,378 @@
...
+struct LogisticRegressionTestFixture
+{
+ public:
+  LogisticRegressionTestFixture()
+  {
+  	try { 
+    	// Cache in the options for this program.
+    	CLI::RestoreSettings(testName);
+    } catch (std::invalid_argument e) {
+    	Log::Fatal << "Invalid Test Name : " << e.what() << std::endl;
```
rcurtin: I'm not sure this is necessary; we haven't had a problem with it before.  Were you having issues that required this code?

manish7294:
@KARTHEEKCIC Somehow the changes you made in CMake are not available now may be due to some kind of merge conflict or other. Because of which logistic_regression_test is not building, please try to correct it.

src/mlpack/tests/main_tests/logistic_regression_test.cpp
KARTHEEKCIC: No. I didn't have any issues with the code. I had just added it for safety, if in any case it produces an exception.

KARTHEEKCIC: Should I remove it?

src/mlpack/methods/logistic_regression/logistic_regression_main.cpp
```diff
@@ -141,9 +141,15 @@ static void mlpackMain()
   const double tolerance = CLI::GetParam<double>("tolerance");
   const double stepSize = CLI::GetParam<double>("step_size");
   const size_t batchSize = (size_t) CLI::GetParam<int>("batch_size");
+
+  // Checking that max iterations is non negative
+  if(CLI::GetParam<int>("max_iterations") <= 0) {
+    Log::Fatal << "Max Iterations (" << CLI::GetParam<int>("max_iterations") << ") cannot be negative" << endl;
+  }
```
rcurtin: Can you use the `RequireParamValue()` function here please?  I suppose also we should check that the batch size is greater than 0 too.  Also, 0 is an allowable value for the maximum number of iterations.

src/mlpack/tests/main_tests/logistic_regression_test.cpp
rcurtin: Yes, I think removing it is fine.  The method itself should throw an exception if there is an issue.

src/mlpack/tests/main_tests/perceptron_test.cpp
```diff
@@ -1,21 +1,18 @@
...
- *
- * Test mlpackMain() of perceptron_main.cpp.
- *
- * mlpack is free software; you may redistribute it and/or modify it under the
- * terms of the 3-clause BSD license.  You should have received a copy of the
- * 3-clause BSD license along with mlpack.  If not, see
- * http://www.opensource.org/licenses/BSD-3-Clause for more information.
- */
+  * @file perceptron_test.cpp
+  * @author B Kartheek Reddy
```
rcurtin: Why did you remove Manish's name? He wrote many of these tests and we should leave them in.

src/mlpack/tests/main_tests/logistic_regression_test.cpp
```diff
@@ -0,0 +1,378 @@
...
+  trainY << 0 << 1 << 0 << 1 << 1 << 1 << 0 << 1 << 0 << 0 << arma::endr; // 10 responses
+  arma::mat testX = arma::randu<arma::mat>(D, M);
+
+  SetInputParam("training", std::move(trainX));
+  SetInputParam("labels", std::move(trainY));
+  SetInputParam("test", std::move(testX));
+
+  mlpackMain();
+
+  const arma::Row<size_t> testY = CLI::GetParam<arma::Row<size_t>>("output");
```
rcurtin: This is a minor issue that doesn't matter much for the tests, but I would suggest using a reference type here (`const arma::Row<size_t>&`) to avoid a copy.

src/mlpack/tests/main_tests/logistic_regression_test.cpp
```diff
@@ -0,0 +1,378 @@
...
+  arma::mat trainX1({{1.0, 2.0, 3.0}, {1.0, 4.0, 9.0}, {0,1,1}});
+  arma::mat testX({{4.0,5.0},{1.0,6.0}});
+
+  SetInputParam("training",std::move(trainX1));
+  SetInputParam("test", testX);
+
+  // The first solution.
+  mlpackMain();
+
+  const arma::Row<size_t> testY1 = CLI::GetParam<arma::Row<size_t>>("output");
```
rcurtin: Here I would suggest using `std::move()`, but again, since these objects are so small for the tests, it doesn't matter too much.  But if you were writing other mlpack code where the data was larger I would be more insistent. :)

src/mlpack/tests/main_tests/logistic_regression_test.cpp
```diff
@@ -0,0 +1,378 @@
...
+  SetInputParam("training", std::move(trainX2));
+  SetInputParam("labels", std::move(trainY2));
+  SetInputParam("test", std::move(testX));
+
+  // The second solution.
+  mlpackMain();
+
+  const arma::Row<size_t> testY2 = CLI::GetParam<arma::Row<size_t>>("output");
+
+  BOOST_REQUIRE_EQUAL_COLLECTIONS(testY1.begin(), testY1.end(), testY2.begin(), testY2.end());
```
rcurtin: Ah, nice, I did not know about `BOOST_REQUIRE_EQUAL_COLLECTIONS()`.

src/mlpack/tests/main_tests/logistic_regression_test.cpp
```diff
@@ -0,0 +1,378 @@
...
+  constexpr int D = 3;
+
+  arma::mat trainX = arma::randu<arma::mat>(D,N);
+  arma::Row<size_t> trainY;
+
+  trainY << 0 << 1 << 0 << 1 << 0 << 1 << 0 << 1 << 0 << 1 << arma::endr; // 10 responses
+
+  SetInputParam("training", std::move(trainX));
+  SetInputParam("labels", std::move(trainY));
+  SetInputParam("max_iterations", int(0.01));
```
rcurtin: I'm a bit confused about this one, shouldn't `int(0.01)` be 0?  0 is an allowable value for `max_iterations`, it means "no limit".

src/mlpack/tests/main_tests/perceptron_test.cpp
KARTHEEKCIC: Sorry for the inconvenience. I think I did a mistake in resolving merge conflicts. I misunderstood. I thought I have to push my file and then afterwards you will merge it with Manish's tests. Now I understand. Don't worry I will correct it now by adding my tests to Manish's Perceptron Test and push that file. I will remember this next time. Thanks for Reviewing.

rcurtin: No worries, when you get it fixed I'll take a look at the perceptron tests and we can finish the review process. :)

src/mlpack/tests/main_tests/logistic_regression_test.cpp
KARTHEEKCIC: Shouldn't int(0.01) value of max iterations give an error as it is an non integer value instead converting it into int type i.e, 0.

rcurtin: No, when you write `int(x)` that is an explicit conversion to type `int`, so in your case `int(0.01)` will create an `int` with value `0`.

KARTHEEKCIC: Yes. I did that mistake. I should use double instead of int.

rcurtin: I wouldn't suggest that, because that will cause `SetInputParam` to be called with a template type of `double`, which will cause all kinds of other disasters.  You can assume that whenever `mlpackMain()` is called, every parameter is of the right type.  Even if I do the following:

```
CLI::GetParam<int>("max_iterations") = double(0.01);
```

if it compiles (and it will at least issue a warning), the double will be implicitly cast to an int, resulting in the `0` value.

KARTHEEKCIC: Ok. So I don't think this test case is needed.

rcurtin: Right, I would agree with that.

KARTHEEKCIC:
@rcurtin. I have written the tests you have mentioned above and made changes according to your reviews. You can check them now. I have also pushed the perceptron test merged with Manish's Perceptron tests. You can now review it.

src/mlpack/tests/main_tests/logistic_regression_test.cpp
KARTHEEKCIC: Done. I made the changes accordingly. Thanks for your suggestion. I will remember that next time.

src/mlpack/tests/main_tests/logistic_regression_test.cpp
KARTHEEKCIC: Made the changes. Thanks for the tip. I will remember that.

src/mlpack/methods/logistic_regression/logistic_regression_main.cpp
KARTHEEKCIC: Done.

src/mlpack/tests/main_tests/perceptron_test.cpp
```diff
@@ -22,298 +25,380 @@ using namespace mlpack;
...
+  // Input training data.
+  SetInputParam("training", std::move(inputData));
+  SetInputParam("max_iterations", (int) -1);
 
   Log::Fatal.ignoreInput = true;
   BOOST_REQUIRE_THROW(mlpackMain(), std::runtime_error);
   Log::Fatal.ignoreInput = false;
 }
 
 /**
```
KARTHEEKCIC: You can start reviewing from here. These are the tests that I have written.

KARTHEEKCIC:
@rcurtin. I was trying to fix style checks in logistic_regression_main.cpp but I am not able to figure why is it failing. Can you help me?

manish7294:
Well you got a whole bunch of style issues here :) . You can easily solve them following this [style check log](http://masterblaster.mlpack.org/job/pull-requests%20mlpack%20style%20checks/1474/cppcheckResult/). Please try to follow [mlpack design guidelines](https://github.com/mlpack/mlpack/wiki/DesignGuidelines).
Let me know if you need any help.

rcurtin:

src/mlpack/tests/main_tests/perceptron_test.cpp
```diff
@@ -229,4 +235,170 @@ BOOST_AUTO_TEST_CASE(PerceptronMaxItrTest)
...
+  PerceptronModel model = std::move(CLI::GetParam<PerceptronModel>("output_model"));
+
+  resetSettings();
+
+  arma::mat trainX2;
+
+  // loading a train dataset with 5 classes
+  if(!data::Load("train_data_5_classes.csv",trainX2)) {
+    BOOST_FAIL("Could not load the train data train_data_5_classes.csv");
+  }
```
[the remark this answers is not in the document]
KARTHEEKCIC: Done

KARTHEEKCIC:
@rcurtin. I have made the changes. Travis CI build could not complete. Can you help me out?

KARTHEEKCIC:
I have tested all the tests. They are running fine. I guess there is some problem with Travis ci.

src/mlpack/tests/data/train_data_5_classes.csv
```diff
@@ -0,0 +1,63 @@
+4.7,3.2,1.6,0.2,0
```
rcurtin: If this file isn't used, can you remove it from the repository please?

src/mlpack/tests/main_tests/perceptron_test.cpp
```diff
@@ -138,9 +146,7 @@ BOOST_AUTO_TEST_CASE(PerceptronLabelsLessDimensionTest)
   arma::Row<size_t> output;
   output = std::move(CLI::GetParam<arma::Row<size_t>>("output"));
 
-  bindings::tests::CleanMemory();
-
-  // Now train perceptron with labels provided.
+  // Now train pereptron with labels provided.
```
rcurtin: I think we should not remove `CleanMemory()` here; is there a reason I have overlooked that we should remove it?

src/mlpack/tests/main_tests/perceptron_test.cpp
```diff
@@ -232,4 +238,186 @@ BOOST_AUTO_TEST_CASE(PerceptronMaxItrTest)
...
+
+  // wrong dimensionality of test data. Should give runtime error.
+  Log::Fatal.ignoreInput = true;
+  BOOST_REQUIRE_THROW(mlpackMain(), std::runtime_error);
+  Log::Fatal.ignoreInput = false;
+}
+
 BOOST_AUTO_TEST_SUITE_END();
+
+
```
rcurtin: Can you remove the extra lines at the end of the file please?

src/mlpack/tests/main_tests/perceptron_test.cpp
```diff
@@ -40,6 +40,14 @@ struct PerceptronTestFixture
   }
 };
 
+// reset the parameters
+void resetSettings()
+{
+  bindings::tests::CleanMemory();
+  CLI::ClearSettings();
+  CLI::RestoreSettings(testName);
+}
```
rcurtin: It looks like this function is never used; should we remove it?

src/mlpack/tests/data/train_data_5_classes.csv
KARTHEEKCIC: Done

src/mlpack/tests/main_tests/perceptron_test.cpp
KARTHEEKCIC: Yeah, you are right. It is never used. I overlooked that.

src/mlpack/tests/main_tests/perceptron_test.cpp
KARTHEEKCIC: Done

src/mlpack/tests/main_tests/perceptron_test.cpp
KARTHEEKCIC: Actually, I didn't remove it. The file contained the older tests that Manish had written and the changes that you made to it as per 1214 didn't apply. I will fix that.

KARTHEEKCIC:
I have checked comments and made changes. 

Sure, I would be happy to find my name in your list of contributors and receive mlpack stickers. Thanks. :)

rcurtin:

Issue #1062, opened 2017-07-19
Using floats instead of double


etano:
Hi, long time armadillo user, newer mlpack user (looks great!).

I'd like to be able to change the datatype in the ann implementation to float (instead of double).

I'm willing to go through and add a template parameter for the datatype as it seems pretty straightforward. But before I go through the effort, I'm just wondering if there's a possibility something like this would be accepted.

Cheers!

rcurtin:
I think actually this should not be too hard---all of the layers (or at least most of them) in `src/mlpack/methods/ann/layer/` already support templated matrix types, meaning you could use floats with them.  But still, I think that `FFN` and `RNN` should probably be refactored at some point to support generic matrix types.  I'm not sure what the exact best way to do that will be---@zoq, any ideas in particular?

This is related to the long-standing issue #290.

etano:
Yeah looking at it, it seems as you say. I think there is an option between templating on a matrix type or templating on an underlying datatype, or both.

Looking forward, having arbitrary types might be convenient when integrating with bandicoot: https://github.com/conradsnicta/bandicoot-code

rcurtin:
Absolutely, I want to be able to drop in bandicoot types when the time comes.  I'm surprised, word is spreading fast about that project I guess... :)

I think that templating on a matrix type would be better---like you pointed out, this would allow us to accept bandicoot types and also `arma::sp_mat`, etc.  I would think the way to do this would be to simply templatize the `Train()` and other related functions to accept arbitrary input types, so that you can train on whatever.

etano:
Well I opened a pull request (not meant to be merged yet), just to get an idea of what this change might look like.

I'm realizing now, though, that it will be a bit more involved because of this LayerTypes boost::variant thing.

zoq:
Right, we have to find a good solution for ``boost::variant``, have to think about it.

rcurtin:
Do you think we could do something like this?

```
template<typename ElemType>
using LayerTypes = boost::variant<
    Add<arma::Mat<ElemType>, arma::Mat<ElemType>>,
    ...
>;
```

dheeraj135:
@rcurtin Is anyone working on it? If not, I would like to look into this.

zoq:
Please feel free to look into the issue.

dheeraj135:
@zoq Okk..

Tanishq-Banyal:
Any updates ?

PR #1200, opened 2018-01-17
Cross Entropy With Logits

ShikharJ:
This PR is basically a rebase to master of the [original PR](https://github.com/mlpack/mlpack/pull/1104). Some minor fixes have been made as well.
TODOs:

- [x] Look for optimizations in `SigmoidCrossEntropyError::Forward` and `SigmoidCrossEntropyError::Backward`
- [x] Provide additional test cases for negative inputs


ShikharJ:
@lozhnikov I was unable to build this locally and test it prior to pushing. What is the minimal way of doing that? Each time I run `make`, my system starts to build the python bindings as well. Also, from where can I cross-check the targeted output for the test-cases?

lozhnikov:
> What is the minimal way of doing that? Each time I run make, my system starts to build the python bindings as well.
You can comment out some lines in e.g. `mlpack/src/test/CMakeLists.txt`. This could reduce the build time considerably (but don't push it).
> Also, from where can I cross-check the targeted output for the test-cases?
Actually, I think the only way is to calculate some test-cases by hand.

ShikharJ:
@lozhnikov I think this is ready for a review.

lozhnikov:
@ShikharJ Thanks for the contribution. I'll look through the code tomorrow.

src/mlpack/tests/ann_layer_test.cpp
```diff
@@ -1144,6 +1144,61 @@ BOOST_AUTO_TEST_CASE(SimpleLogSoftmaxLayerTest)
...
+  for (size_t i = 0; i < output.n_elem; i++)
+    BOOST_REQUIRE_SMALL(output(i) - expectedOutput(i), 1e-5);
+  BOOST_REQUIRE_EQUAL(output.n_rows, input2.n_rows);
+  BOOST_REQUIRE_EQUAL(output.n_cols, input2.n_cols);
+
+  module.Backward(std::move(input3), std::move(target3), std::move(output));
+  for (size_t i = 0; i < 8; ++i)
+  {
+    double el = output.at(0, i);
+    if (input3.at(i) == 0)
```
lozhnikov: I think the exact comparison isn't an error here. Nevertheless, could you replace this by an approximate comparison with some precision.

src/mlpack/methods/ann/layer/sigmoid_cross_entropy_error.hpp
```diff
@@ -0,0 +1,120 @@
...
+ *
+ * Hence, to ensure stability and avoid overflow, the implementation uses this
+ * equivalent formulation:
+ * \f$max(x, 0) - x * z + \log(1 +  e^{-|x|})\f$
+ *
+ * @tparam InputDataType Type of the input data (arma::colvec, arma::mat,
+ *         arma::sp_mat or arma::cube).
+ * @tparam OutputDataType Type of the output data (arma::colvec, arma::mat,
+ *         arma::sp_mat or arma::cube).
+ */
```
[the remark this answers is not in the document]
ShikharJ: The details actually follow from [here](https://github.com/mlpack/mlpack/pull/1104#discussion_r137058610). I was unable to find a paper to cite for the above.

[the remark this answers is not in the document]
lozhnikov: I know, looks like I missed that in summer.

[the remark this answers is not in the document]
zoq: I think it would be fine to drop derivation and just use: ``f(x,t) = x - x * t + log(1 + exp(-x))`` where ``x`` is the input and ``t`` are the target values.

At the end it's just Sigmoid + CrossEntropy nothing special, if you like you can cite "On Loss Functions for Deep Neural Networks in Classification" by K. Janocha et al.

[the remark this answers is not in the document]
ShikharJ: @rcurtin had previously advocated for an optimized implementation. I'd like to wait and see if he concurs with the same.

ShikharJ:
@rcurtin Please advise what should be done [here](https://github.com/mlpack/mlpack/pull/1200#discussion_r162702238).

src/mlpack/methods/ann/layer/sigmoid_cross_entropy_error.hpp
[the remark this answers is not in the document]
rcurtin: In the context of this comment I think it's just fine to take @zoq's suggestion.  No need for a full derivation in the comments, in my view.

src/mlpack/methods/ann/layer/sigmoid_cross_entropy_error_impl.hpp
```diff
@@ -0,0 +1,84 @@
...
+    }
+  }
+  double loss = max - arma::accu(input % target) + arma::accu(output);
+
+  return loss;
+}
+
+template<typename InputDataType, typename OutputDataType>
+template<typename eT>
+inline  void SigmoidCrossEntropyError<InputDataType, OutputDataType>::Backward(
```
rcurtin: Looks like an extra space here.

src/mlpack/methods/ann/layer/sigmoid_cross_entropy_error_impl.hpp
```diff
@@ -0,0 +1,84 @@
...
+}
+
+template<typename InputDataType, typename OutputDataType>
+template<typename eT>
+inline double SigmoidCrossEntropyError<InputDataType, OutputDataType>::Forward(
+    const arma::Mat<eT>&& input, const arma::Mat<eT>&& target)
+{
+  double max = 0;
+  arma::mat output;
+  SoftplusFunction::Fn(static_cast<arma::mat>(-arma::abs(input)), output);
```
rcurtin: I realize this was not your code originally, but I think this can be removed.  Keeping in mind that all we want in the end is `arma::accu(output)`, and `output` is made by looping over the elements of `input` individually, we can simply add the elementwise calculation to the loop below:

```
max += SoftplusFunction::Fn(-std::abs(input[i]));
```

and maybe it would be worthwhile to change the name of the `max` variable.

(Another note: the static_cast<> is a bad idea and could cause some really strange problems.  The code I'm suggesting above is limited to `eT = double`, but we can leave that for #1062.)

ShikharJ:
Thanks @rcurtin. This is updated now. A final review?

src/mlpack/tests/ann_layer_test.cpp
```diff
@@ -1165,21 +1166,37 @@ BOOST_AUTO_TEST_CASE(SimpleSigmoidCrossEntropyLayerTest)
...
   module.Backward(std::move(input1), std::move(target1), std::move(output));
   for (size_t i = 0; i < output.n_elem; i++)
-    BOOST_REQUIRE_SMALL(output(i) - 0.62245929, 1e-5);
+    BOOST_REQUIRE_SMALL(output(i) - 0.37754066, 1e-5);
   BOOST_REQUIRE_EQUAL(output.n_rows, input1.n_rows);
   BOOST_REQUIRE_EQUAL(output.n_cols, input1.n_cols);
 
   expectedOutput = arma::mat(
-      "0.7310586 0.88079709 -0.04742587 0.98201376 -0.00669285");
+      "0.2689414 0.1192029 -0.04742587 0.0179862 -0.00669285");
```
rcurtin: Could I ask why these numbers have changed from the original test?  I am not sure that there was anything wrong with that test.

src/mlpack/tests/ann_layer_test.cpp
```diff
@@ -1165,21 +1166,37 @@ BOOST_AUTO_TEST_CASE(SimpleSigmoidCrossEntropyLayerTest)
...
+
+  module.Backward(std::move(input3), std::move(target3), std::move(output));
+  for (size_t i = 0; i < 8; ++i)
+  {
+    double el = output.at(0, i);
+    if (std::abs(input3.at(i) - 0.0) < 1e-5)
+      BOOST_REQUIRE_SMALL(el - 0.5, 2e-6);
+    else
+      BOOST_REQUIRE_SMALL(el - 1.2689414, 2e-6);
+  }
```
rcurtin: I also don't follow where these numbers came from, could you add a comment describing why this is the correct output please?

src/mlpack/tests/ann_layer_test.cpp
ShikharJ: The previous implementation was slightly incorrect. See [here](https://github.com/mlpack/mlpack/pull/1104#discussion_r137109612)

src/mlpack/tests/ann_layer_test.cpp
ShikharJ: I checked them using `tensorflow.nn.sigmoid_cross_entropy_with_logits()`. The comment is added [here](https://github.com/mlpack/mlpack/pull/1200/files#diff-d686e5871ec51f01e597deaae104def3R1161)

ShikharJ:
@rcurtin Please let me know if you want anything else done here?

rcurtin:
There are lots of PRs to look at (and lots of other work I have to do), I will get to this when I have a chance.  I need to check the tests, then we need to see if anyone else has any comments.

src/mlpack/methods/ann/layer/sigmoid_cross_entropy_error_impl.hpp
```diff
@@ -31,39 +31,43 @@ template<typename eT>
...
-  output(positive) = 1 - target(positive) - temp;
-  SoftplusFunction::Deriv(static_cast<arma::mat>(-arma::abs(input(negative))),
-      temp);
-  output(negative) = -(target(negative)) - negative;
+  for (size_t i = 0; i < output.n_elem; ++i)
+  {
+    const double val = target[i];
+    if (val > 0)
+    {
+      output[i] = 1 - val - SoftplusFunction::Deriv(-std::abs(input[i]));
```
zoq: Not sure if this is faster, but what about using:

```c++
output = 1 - target - SoftplusFunction::Deriv(-std::abs(input));
output = (-1 * output > 0) + 1;
```

do you mind to give this is try? A simple test would use both snippets and time everything e.g. with the C++11 time class.

src/mlpack/tests/ann_layer_test.cpp
```diff
@@ -1165,21 +1166,37 @@ BOOST_AUTO_TEST_CASE(SimpleSigmoidCrossEntropyLayerTest)
   double error2 = module.Forward(std::move(input2), std::move(target2));
   BOOST_REQUIRE_SMALL(error2 / input2.n_elem - 1.5027283, 1e-6);
 
+  input3 = arma::mat("0 -1 -1 0 -1 0 0 -1");
+  target3 = arma::mat("0 -1 -1 0 -1 0 0 -1");
+  double error3 = module.Forward(std::move(input3), std::move(target3));
+  BOOST_REQUIRE_SMALL(error3 / input3.n_elem - 0.00320443, 1e-6);
```
zoq: This is just fine, just wanted to point out that you could also use: ``BOOST_REQUIRE_CLOSE``.

src/mlpack/tests/ann_layer_test.cpp
zoq: I think it might improve the overall readability if we define the expected results in above instead of using them inside the condition. What do you think?

src/mlpack/methods/ann/layer/sigmoid_cross_entropy_error_impl.hpp
ShikharJ: That will lead us back to Kris' original implementation which was avoided to prevent the usage of a temporary matrix in the first place.

ShikharJ: Doing this:
```
  auto start = std::chrono::system_clock::now();
  double error1 = module.Forward(std::move(input1), std::move(target1));
  auto end = std::chrono::system_clock::now();
  std::chrono::duration<double> elapsed_seconds = end - start;
```
and adding up the `elapsed_seconds` over all the calls, I'm getting an average of `3.5e-06` seconds.

zoq: Thanks for looking into the issue, just wanted to make sure we don't miss anything here.

Is this for the current implementation? What was the input size?

ShikharJ: Yes, this is for the current implementation. I timed it for only the test cases we have written, what input size would you like to know the benchmarks for?

src/mlpack/tests/ann_layer_test.cpp
ShikharJ: This has to be done for the other instances of tests in this sub-module I presume?

src/mlpack/methods/ann/layer/sigmoid_cross_entropy_error_impl.hpp
zoq: Thanks for the fast response, what do you think about 256, 512, 1024 and 2048? If you have time can you time the other approach as well, don't feel obligated I can also run the tests.

zoq: @ShikharJ @kris-singh Did you compare the results of the ``Backward`` function against Tensorflow, since I don't think the derivative is correct for the case ``x <= 0`` http://www.wolframalpha.com/input/?i=d%2Fdx++max(x,+0)+-+x+*+z+%2B+log(1+%2B+exp(-abs(x)))

If you checked the backward pass, we should take a second look, if not we can simplify the overall backward function.

lozhnikov: I guess the derivative is correct:
Let x > 0. Then
f  ( - abs ( x ) ) = f ( - x ) and
d/dx [ f ( - abs ( x ) ) ] = d/dx [ f ( - x ) ] = - d f ( y ) / d y, where y = - x.

Let x < 0. Then
f  ( - abs ( x ) ) = f ( x ) and
d/dx [ f ( - abs ( x ) ) ] = d/dx [ f ( x ) ] =  d f ( x ) / d x

zoq: Right, but did you check the sign of the second term, shouldn't this be:

```
z - 1/(e^x+1)
```
also the current method uses target inside the loop, where Kris original method used input.

rcurtin: In this discussion `x` is `input` and `z` is `target` (unless I've misunderstood), so it seems like at the very least the conditional should be `if (input[i] > 0)` and not `if (val > 0)`.  Let me know if I am wrong about that.

But to the best of my (quick) analysis, I think both the derivative that @lozhnikov and the Wolfram Alpha results are correct; but it seems to me that the Wolfram Alpha results (specifically the first "alternate form") offers us a really nice simplification:

```
// I'm avoiding SoftplusFunction::Deriv() here so the entire expression is evaluated at once by Armadillo.
output = 1.0 / (1.0 + arma::exp(-input)) - target;
```

Like I said, let me know if I am wrong.  I am not fully confident but this is what it seems like to me.

rcurtin: > the current method uses target inside the loop, where Kris original method used input.

This is my error from https://github.com/mlpack/mlpack/pull/1104/files#r137060612, sorry about that. :(

ShikharJ: @zoq The derivative is correct to the best of my knowledge as well. Also, we can just ignore, `SoftplusFunction::Deriv()` and go by the implementation given by @rcurtin (and yes, my bad that I didn't notice that input should be inside the loop). What made you arrive at this though?
```
z - 1/(e^x+1)
```

ShikharJ: I ran this snippet:
```
input1 = arma::ones(1, 256);
  target1 = arma::ones(1, 256);
  auto start = std::chrono::system_clock::now();
  error1 = module.Forward(std::move(input1), std::move(target1));
  module.Backward(std::move(input1), std::move(target1), std::move(output));
  auto end = std::chrono::system_clock::now();
  std::chrono::duration<double> elapsed_seconds = end - start;
  std::cout << elapsed_seconds.count() << std::endl;

  input1 = arma::ones(1, 512);
  target1 = arma::ones(1, 512);
  start = std::chrono::system_clock::now();
  error1 = module.Forward(std::move(input1), std::move(target1));
  module.Backward(std::move(input1), std::move(target1), std::move(output));
  end = std::chrono::system_clock::now();
  elapsed_seconds = end - start;
  std::cout << elapsed_seconds.count() << std::endl;

  input1 = arma::ones(1, 1024);
  target1 = arma::ones(1, 1024);
  start = std::chrono::system_clock::now();
  error1 = module.Forward(std::move(input1), std::move(target1));
  module.Backward(std::move(input1), std::move(target1), std::move(output));
  end = std::chrono::system_clock::now();
  elapsed_seconds = end - start;
  std::cout << elapsed_seconds.count() << std::endl;

  input1 = arma::ones(1, 2048);
  target1 = arma::ones(1, 2048);
  start = std::chrono::system_clock::now();
  error1 = module.Forward(std::move(input1), std::move(target1));
  module.Backward(std::move(input1), std::move(target1), std::move(output));
  end = std::chrono::system_clock::now();
  elapsed_seconds = end - start;
  std::cout << elapsed_seconds.count() << std::endl;
```
And got the following output for the current implementation (averaged over 5 instances):
```
2.0128e-05 (256)
5.8301e-05 (512)
9.0369e-05 (1024)
0.000166852 (2048)
```
and this on @rcurtin's approach (again, averaged over 5 instances):
```
2.4565e-05 (256)
0.0132882 (512) (weird?)
8.9329e-05 (1024)
0.000164627 (2048)
```

src/mlpack/tests/ann_layer_test.cpp
ShikharJ: @rcurtin The tests have reverted to original, after changing the incorrect conditional.

src/mlpack/methods/ann/layer/sigmoid_cross_entropy_error_impl.hpp
ShikharJ: @zoq It seems that for smaller instances, the single loop works better, but for larger sets, the `arma::exp()` approach might pan out just slightly better. How do you wish to proceed here?

zoq: If there are no objections let's go with:

```c++
// I'm avoiding SoftplusFunction::Deriv() here so the entire
// expression is evaluated at once by Armadillo.
output = 1.0 / (1.0 + arma::exp(-input)) - target;
```

I think there is no need for an additional check. Also thanks for all the timings.

ShikharJ: Would you have any idea why I got that wild timing for the case of `512`? Was that probably because of a problem with my machine?

zoq: hm, not sure, if you rerun the test do you get similar results?

ShikharJ: I'll have to build and check again. But that test was averaged over 5 instances. And the output for the case was pretty much the same each time.

zoq: Will also run the timings and post the results here.

zoq: Results:

```
256: 0.281 ms
512: 0.563 ms
1014: 0.979 ms
2048: 1.989 ms
```

ShikharJ: Ah, thanks, I rechecked again. No problems this time. Thanks, I updated the implementation here.

ShikharJ:
@zoq @lozhnikov Any further suggestions from your side?

src/mlpack/methods/ann/layer/sigmoid_cross_entropy_error.hpp
```diff
@@ -0,0 +1,117 @@
+/**
+ * @file sigmoid_cross_entropy_error.hpp
+ * @author Kris Singh
```
zoq: Feel free to add your name here.

src/mlpack/methods/ann/layer/sigmoid_cross_entropy_error.hpp
```diff
@@ -0,0 +1,117 @@
...
+
+namespace mlpack {
+namespace ann /** Artificial Neural Network. */ {
+
+/**
+ * The SigmoidCrossEntropyError performance function measures the network's
+ * performance according to the cross-entropy function between the input and
+ * target distributions. This function calculates the cross entropy
+ * given the real values instead of providing the sigmoid activations.
+ * The function is much more numerically stable and uses this
```
zoq: I think we should drop

> The function is much more numerically stable and uses this equivalent formulation:

this there is no further context about the previous "unstable" function.

src/mlpack/methods/ann/layer/sigmoid_cross_entropy_error.hpp
```diff
@@ -0,0 +1,117 @@
...
+ * where x = input and z = target.
+ *
+ * For more information, see the following paper.
+ *
+ * @code
+ * @article{1702.05659,
+ *   title={On Loss Functions for Deep Neural Networks in Classification},
+ *   author={Katarzyna Janocha, Wojciech Marian Czarnecki},
+ *   eprint={arXiv:1702.05659},
+ *   year={2017}
```
zoq: Can you add ``url       = {http://arxiv.org/abs/1702.05659}`` and ``journal   = {CoRR},``?

src/mlpack/methods/ann/layer/sigmoid_cross_entropy_error_impl.hpp
```diff
@@ -0,0 +1,68 @@
...
+ * @author Kris Singh
+ *
+ * Implementation of the sigmoid cross entropy error performance function.
+ *
+ * mlpack is free software; you may redistribute it and/or modify it under the
+ * terms of the 3-clause BSD license.  You should have received a copy of the
+ * 3-clause BSD license along with mlpack.  If not, see
+ * http://www.opensource.org/licenses/BSD-3-Clause for more information.
+ */
+#ifndef MLPACK_METHODS_ANN_LAYER_SIGMOID_CROSS_ENTROPY_ERROR
```
zoq: Can you add ``_IMPL``to the header guard?

src/mlpack/methods/ann/layer/sigmoid_cross_entropy_error_impl.hpp
```diff
@@ -0,0 +1,68 @@
...
+template<typename InputDataType, typename OutputDataType>
+template<typename eT>
+inline double SigmoidCrossEntropyError<InputDataType, OutputDataType>::Forward(
+    const arma::Mat<eT>&& input, const arma::Mat<eT>&& target)
+{
+  double maximum = 0;
+  for (size_t i = 0; i < input.n_elem; ++i)
+  {
+    maximum += std::max(input[i], 0.0) +
+        SoftplusFunction::Fn(-std::abs(input[i]));
```
zoq: Since we deal with logit as input, we should directly use: ``std::log(1 + std::exp(x))`` here to avoid an unnecessary check.

src/mlpack/methods/ann/layer/sigmoid_cross_entropy_error_impl.hpp
```diff
@@ -0,0 +1,68 @@
...
+  }
+  double loss = maximum - arma::accu(input % target);
```
zoq: Looks like we can remove the extra return at the end and return right here, what do you think?

src/mlpack/methods/ann/layer/sigmoid_cross_entropy_error_impl.hpp
```diff
@@ -0,0 +1,68 @@
...
+}
+
+template<typename InputDataType, typename OutputDataType>
+template<typename eT>
+inline void SigmoidCrossEntropyError<InputDataType, OutputDataType>::Backward(
+    const arma::Mat<eT>&& input,
+    const arma::Mat<eT>&& target,
+    arma::Mat<eT>&& output)
+{
+  output.set_size(input.n_rows, input.n_cols);
```
zoq: No need to set the output size anymore.

src/mlpack/methods/ann/layer/sigmoid_cross_entropy_error_impl.hpp
ShikharJ: I'm fine with, declaring a `double` seems redundant anyways.

ShikharJ:
@zoq Thanks for your suggestions, the PR has been updated. However, it looks like a memory leak has occurred in another part of the test suite (which hadn't been touched). I have no idea regarding how that can be overcome. A little help?

zoq:
Don't worry about the memory issues, they show up since you changed the test file, will fix them later.

ShikharJ:
Thanks! It's amazing to be a part of this community.

PR #1204, opened 2018-01-19
Generative Adversarial Network

ShikharJ:
A rebase to master of the [original PR](https://github.com/mlpack/mlpack/pull/1066) with some minor fixes.
Relevant: #1206
TODOs:

- [x] Apply Implementation Changes
- [x] Fix Build and Test Errors
- [x] Add Extensive Tests


lozhnikov:
@ShikharJ This summer we tried to reproduce this test
https://www.oreilly.com/learning/generative-adversarial-networks-for-beginners
Unfortunately, we hadn't succeeded. I think the test is the main criterion. You could find Kris's implementation of the test here https://gist.github.com/kris-singh/40dc1db9f257c1694245393392e4e9cf (gan_conv.cpp)

Our GAN implementation differs from the oreilly implementation due to mlpack Optimizer API. As for me, that shouldn't be a problem, however, I can't be sure since we didn't try the exact implementation.
Feel free to ask questions.

ShikharJ:
@lozhnikov I noticed that this shares some code with #1098, which has been worked upon a bit in addition to this. Should I first get that PR completed and then this?

lozhnikov:
@ShikharJ It's up to you. If I remember right the PR doesn't depend on `ResizeLayer`, it is needed for the oreilly test.

ShikharJ:
Actually, the issue coming up here is also present in that implementation. If I can get that merged in, this PR would be in a much better state than it is now.

ShikharJ:
@lozhnikov Sorry for taking too long for getting into this. I had exams, and had shifted my focus to the PR on Bilinear Interpolation for some time. 
I went through the O'Reilly blog, and it seems that they have used three `Adam` optimizer instantiations (one for training generator and one each for training discriminator on real and generated images). I don't really think having a single `Adam` optimizer should be a technical problem, as far as the O'Reilly test is concerned, but from [this source](https://github.com/soumith/ganhacks) it seems like a better idea to be able to set the optimizer types for generator and discriminator separately (like `Adam` for Generator and `SGD` for Discriminator). 
It also seems a number of implementation-related details have been borrowed from there, and so I think it makes sense to try the remaining ideas out. Though, do we want to replicate the O'Reilly test code by code and stick with the `Adam` optimizer implementation only? I know that it would mean we have to break the API, but that would give us better results probably. cc @zoq

lozhnikov:
@ShikharJ Sorry for the slow response. Actually, I don't know. If two different optimisers do make sense, we definitely should replace the current implementation. But in order to be sure we have to prove that by example. That is if we implement GAN with 2 optimizers and show that it really works better than the previous implementation then we should use the second implementation.

zoq:
Agreed, let's  test if different optimization methods are useful first.

ShikharJ:
@zoq After the recent change with the Optimizer API, this seems to fail, and I'm unable to figure out why. Can you help me with this?

zoq:
The travis build issue isn't related with the recent Optimizer API changes, increasing: https://github.com/mlpack/mlpack/blob/130918a5b3bf961eb1f88b251ad56035e9ad7e3c/src/mlpack/prereqs.hpp#L68

should at least solve the travis build issue, maybe I missed something?

ShikharJ:
I am getting the following error on increasing it to 41:
```
In file included from /usr/include/boost/math/policies/policy.hpp:9:0,
                 from /usr/include/boost/math/special_functions/math_fwd.hpp:28,
                 from /usr/include/boost/math/special_functions/sign.hpp:17,
                 from /usr/include/boost/lexical_cast/detail/inf_nan.hpp:34,
                 from /usr/include/boost/lexical_cast/detail/converter_lexical_streams.hpp:63,
                 from /usr/include/boost/lexical_cast/detail/converter_lexical.hpp:54,
                 from /usr/include/boost/lexical_cast/try_lexical_convert.hpp:42,
                 from /usr/include/boost/lexical_cast.hpp:32,
                 from /usr/include/boost/program_options/value_semantic.hpp:14,
                 from /usr/include/boost/program_options/options_description.hpp:13,
                 from /usr/include/boost/program_options.hpp:15,
                 from /home/shikhar/mlpack/src/mlpack/core/dists/gamma_distribution.hpp:23,
                 from /home/shikhar/mlpack/src/mlpack/core.hpp:265,
                 from /home/shikhar/mlpack/src/mlpack/bindings/cli/print_help.hpp:10,
                 from /home/shikhar/mlpack/src/mlpack/bindings/cli/print_help.cpp:8:
/usr/include/boost/mpl/list.hpp:36:1: fatal error: boost/mpl/list/list41.hpp: No such file or directory
 #   include BOOST_PP_STRINGIZE(boost/mpl/list/AUX778076_LIST_HEADER)
 ^~~~~~
```

src/mlpack/methods/ann/gan_impl.hpp
```diff
@@ -0,0 +1,392 @@
...
+    discriminator.Gradient(discriminator.parameter, numFunctions,
+      noiseGradientDiscriminator, batchSize);
+    //std::cout << "Updated Noise Discriminator Gradient:" << std::endl << noiseGradientDiscriminator << std::endl;
+    generator.error = boost::apply_visitor(deltaVisitor,
+        discriminator.network[1]);
+    //std::cout << "Generator Error:" << std::endl << generator.error << std::endl;
+    //std::cout << "Noise:" << std::endl << noise << std::endl;
+    generator.currentInput = noise;
+    std::cout << "Calling Generator Backward:" << std::endl;
+    generator.Backward();
```
zoq: ``generator.Gradient(..)`` already calls the backward function, so I think this is unnecessary.

src/mlpack/methods/ann/gan_impl.hpp
```diff
@@ -0,0 +1,394 @@
...
+    generator.error = boost::apply_visitor(deltaVisitor,
+        discriminator.network[1]);
+    //std::cout << "Generator Error:" << std::endl << generator.error << std::endl;
+    //std::cout << "Noise:" << std::endl << noise << std::endl;
+    //generator.currentInput = predictors;
+    //std::cout << "Calling Generator Backward:" << std::endl;
+    //generator.Backward();
+    //std::cout << "Generator Gradient:" << std::endl << gradientGenerator << std::endl;
+    std::cout << "Setting Predictors:" << std::endl;
+    generator.Predictors() = discriminator.predictors(numFunctions);
```
ShikharJ: @zoq I'm not sure I'm setting the predictors in the right way, since `generators.Predictors()` returns me a `1x1` zero matrix when this calls `Forward` inside `FFN::Gradient`. Can you help me out here?

ShikharJ:
@zoq @rcurtin @lozhnikov The O'Reilly test is way too big to provide any results within the Travis build time. Do you have any suggestions as to where can we shift the test or should we disable this test in a CI environment?

lozhnikov:
@ShikharJ Yeah, the O'Reilly test is too complicated for Travis. I think it is a good idea to implement a standalone test and describe the results and the code as a tutorial (e.g. http://mlpack.org/docs/mlpack-3.0.1/doxygen/tutorials.html).
@zoq How do you think?

ShikharJ:
@lozhnikov I would also appreciate a review of this code. Currently it works for `batchSize = 1`, and needs to be improved for `batchSize > 1`, and I also have a few questions, that I'd like to clear up.

lozhnikov:
@ShikharJ I'll take a look tomorrow.

src/mlpack/methods/ann/gan.hpp
```diff
@@ -0,0 +1,181 @@
...
+#include <mlpack/methods/ann/ffn.hpp>
+#include <mlpack/methods/ann/visitor/output_parameter_visitor.hpp>
+#include <mlpack/methods/ann/visitor/reset_visitor.hpp>
+#include <mlpack/methods/ann/visitor/weight_size_visitor.hpp>
+#include <mlpack/methods/ann/visitor/weight_set_visitor.hpp>
+
+
+namespace mlpack {
+namespace ann /** Artificial Neural Network. **/ {
+template<typename Model, typename InitializationRuleType, class Noise>
```
zoq: Can you add a comment for each template parameter (we can just move up the ones from the constructor), and a simple description what this class does would be great.

src/mlpack/methods/ann/gan.hpp
```diff
@@ -0,0 +1,181 @@
...
+  void Train(OptimizerType& Optimizer);
+
+  /**
+   * Evaluate function for the GAN gives the performance of the GAN on the
+   * current input.
+   *
+   * @param parameters The parameters of the network.
+   * @param i Index of the current input.
+   */
+  double Evaluate(const arma::mat& parameters, const size_t i,
```
zoq: Picky style comment, the parameter should align with the first one. Same for the ``Gradient``method.

src/mlpack/methods/ann/ffn.hpp
```diff
@@ -394,11 +394,13 @@ class FFN
 
   //! Locally-stored copy visitor
   CopyVisitor<CustomLayers...> copyVisitor;
+
+template<typename Model, typename InitializerType, class NoiseType>
```
zoq: Another picky style comment, can we tab this ones? And do you think we should add something like ``The GAN class should have access to internal members.``?

src/mlpack/methods/ann/gan_impl.hpp
```diff
@@ -0,0 +1,283 @@
...
+
+template<typename Model, typename InitializationRuleType, typename Noise>
+void GAN<Model, InitializationRuleType, Noise>::Reset()
+{
+  size_t genWeights = 0;
+  size_t discWeights = 0;
+
+  NetworkInitialization<InitializationRuleType> networkInit(initializeRule);
+
+  for (size_t i = 0; i < generator.network.size(); ++i){
```
zoq: Another picky style comment, can we put the opening ``{`` on an extra line?

src/mlpack/methods/ann/gan_impl.hpp
```diff
@@ -0,0 +1,283 @@
...
+    genWeights += boost::apply_visitor(weightSizeVisitor, generator.network[i]);
+  }
+
+  for (size_t i = 0; i < discriminator.network.size(); ++i){
```
zoq: See comment above.

src/mlpack/methods/ann/gan_impl.hpp
```diff
@@ -0,0 +1,283 @@
...
+
+  discriminator.Forward(std::move(currentInput));
+  double res = discriminator.outputLayer.Forward(
+      std::move(boost::apply_visitor(
+          outputParameterVisitor,
+          discriminator.network.back())), std::move(currentTarget));
+
+  noise.imbue( [&]() { return noiseFunction();} );
+  generator.Forward(std::move(noise));
+  arma::mat temp = boost::apply_visitor(
```
zoq: What about using ``discriminator.predictors.col(numFunctions) = ...`` here instead of an extra temp parameter.

ShikharJ: Sure, that would be an optimization.

zoq:
What we could do is to comment out the test for now, add a comment that it's currently not possible to run this every time due to time constraints. And write a simple example for the models repo, where we can use jenkins to run the code.

zoq:
Something like: https://github.com/mlpack/models/tree/master/Kaggle/DigitRecognizerCNN should work just fine.

zoq:
About the optimization process, do you think at this point, it should be possible to train the generator and discriminator with different optimizer?

I think if we provide such a support we could simplify the Gradient method since we could remove the extra step check. A simple solution might be to provide some form of a wrapper class, that takes two optimizers and runs everything once Train is called:

```c++
template<class OptimizerDiscriminator, class OptimizerGenerator>
class GANOptimizer
{
  GANOptimization(OptimizerDiscriminator& optDiscriminator, OptimizerGenerator& optGenerator)
  {
    ..
  }

  template<typename DecomposableFunctionTypeDiscriminator,
           typename DecomposableFunctionTypeGenerator>
  double Optimize(DecomposableFunctionTypeDiscriminator& discriminatorFunction,
                  DecomposableFunctionTypeGenerator generatorFunction,
                  arma::mat& iterate)
  {
    optDiscriminator.Optimize(discriminatorFunction, iterate.submat());
    optGenerator.Optimize(generatorFunction, iterate.submat());
  }
}
```

This is just a simple idea, let me know what you think.

ShikharJ:
@zoq This idea seems plausible to me, and I don't think we'll have to break the `mlpack` optimization API by doing this. However, this would be a reasonable big task, and I'd like to take this and batch-size support in another PR. Let's get this merged first? cc @lozhnikov

zoq:
Agreed, let's work on this in different PR's, this makes the review process also easier.

ShikharJ:
@zoq I'll post the results tomorrow, then we can merge this.

ShikharJ:
@zoq @lozhnikov Here's the output on `10,000` images, `300` pre-training and `10` epochs (under 10 hours, which is quite competitive with what other libraries have to offer). 
10_epoch

zoq:
Happy to merge this one, thanks @ShikharJ and @kris-singh for the great work.

Issue #1203, opened 2018-01-18
Use Clang-Format


ShikharJ:
Currently, one has to manually format the code after implementation. In my short time with `mlpack`, I had to push twice a couple of times just to fix the formatting issue, which leads to unnecessary CI backlogs. It'd be nice if a contributor could just run a script to automatically format the code. Thoughts?

zoq:
I think that's a good idea, as long as we can provide something that's compliant with our style guidelines (https://github.com/mlpack/mlpack/wiki/DesignGuidelines).

rcurtin:
Yeah, I agree, I think it would be reasonable to add a utility script to the repository to format code (or provide a configuration for a tool).  See also #420, #978.

sbrodehl:
Uhhh yes, I like.

Shashi456:
How i probably go about completing the travis integration @ShikharJ

ShikharJ:
Hi @Shashi456. As you can see from the discussion in #1205, we're still under discussion regarding what should be done regarding this issue. We're probably past the idea of `Travis` integration for `Clang-Format`, though you might want to dig into @rcurtin's idea [here](https://github.com/mlpack/mlpack/pull/1205#issuecomment-360147330)?

ShikharJ:
Closing this for now.

PR #1205, opened 2018-01-19
Use Clang-Format For Code Formatting

ShikharJ:
@rcurtin @zoq I've currently pushed some basic settings for using `clang-format`. The distribution being used is `clang-format-4.0`. We can build CI tests after we have finalized on the changes in the `.clang-format` file. Currently, one only needs to run the `test_format.sh` script. I'll also add the formatted files to portray the changes. Please advise.
Relevant: #1203 
TODOs:

- [ ] Complete Travis Integration


zoq:
The check should ignore any code in ``src/mlpack/core/arma_extend/`` and ``src/mlpack/core/boost_backport/`` both have there own coding conventions.

Also it would be great if you could look at: https://github.com/mlpack/mlpack/wiki/DesignGuidelines maybe we can reproduce the spacing so that ``template<typename ...>`` is not marked as an issue, the same applies to ``#ifdef``.

ShikharJ:
@zoq The spacing for `template<typename...>` has been fixed. However, I'm not sure if the spacing for the `#ifdef` directives would be possible with `clang-format` or not. Currently, it supports only this type of indentation:
```
#if FOO
#  if BAR
#    include <foo>
#  endif
#endif
```

zoq:
hm, not sure I missed something, but it looks like:

```
#if FOO
#  if BAR
#    include <foo>
#  endif
#endif
```

is what we like see, right now it does:

```
#ifdef HAS_OPENMP
 -  #include <omp.h>
+#include <omp.h>
#endif
```

Maybe we can set ``PPDirectiveIndentStyle`` to ``PPDIS_AfterHash`` and see if that works for us?

ShikharJ:
What I meant to say was, that `clang-format` cannot account for our style of preprocessor directive indentation, which is:
```
#ifdef HAS_OPENMP
  #include <omp.h>
#endif
```
We can use `IndentPPDirectives` command and set it to `AfterHash` to obtain the following:
```
#if FOO
#  if BAR
#    include <foo>
#  endif
#endif
```
But that command is only available for `clang-format-6.0` and above. Currently I am only able to find the `apt` packages for `clang-format-5.0` at maximum.

ShikharJ:
@zoq How do you wish to proceed here? Seems like `clang-format-6.0` is only available as a development branch at the [apt.llvm.org](https://apt.llvm.org/). Should I let it be as it is right now or should I make use of the `AfterHash` command? cc @rcurtin

zoq:
Okay, I guess since we don't really use a lot of ``#ifded`` it's fine to leave it as it is. I'll make some comments in the modified files, maybe we can fix this issues.

.travis/config.hpp
```diff
@@ -10,45 +10,54 @@
 // of the License or (at your option) any later version.
 // (see http://www.opensource.org/licenses for more info)
 
-
-
 #if !defined(ARMA_USE_LAPACK)
```
zoq: We should irgnore this file too; you can take a look at the existing linter script which file is excluded from the check: https://github.com/mlpack/jenkins-conf/blob/master/linter/lint.sh

src/mlpack/core/cv/cv_base.hpp
```diff
@@ -149,7 +146,7 @@ class CVBase
    * numClasses parameter and a data::DatasetInfo parameter.
    */
   template<typename... MLAlgorithmArgs,
-           bool Enabled = MIE::TakesNumClasses & MIE::TakesDatasetInfo,
+           bool Enabled = MIE::TakesNumClasses& MIE::TakesDatasetInfo,
```
zoq: Instead of the bitwise operator we should use ``&&`` here, however since the bool type is guaranteed to be only either a true or a false, it's just fine here to use the bitwise operator. Just might be out of the scope of the PR, so maybe we should fix this in another PR.

src/mlpack/core/cv/k_fold_cv.hpp
```diff
@@ -56,8 +56,8 @@ template<typename MLAlgorithm,
          typename PredictionsType =
              typename MetaInfoExtractor<MLAlgorithm, MatType>::PredictionsType,
          typename WeightsType =
-             typename MetaInfoExtractor<MLAlgorithm, MatType,
-                 PredictionsType>::WeightsType>
+             typename MetaInfoExtractor<MLAlgorithm, MatType, PredictionsType>::
```
zoq: I think the previous one is easier to read, not sure if we can avoid a break once a namespace or something like that is used.

src/mlpack/core/cv/simple_cv_impl.hpp
```diff
@@ -21,134 +21,118 @@ template<typename MLAlgorithm,
...
-                                MIT&& xs,
-                                PIT&& ys) :
-    SimpleCV(Base(), validationSize, std::forward<MIT>(xs),
-        std::forward<PIT>(ys))
-{ /* Nothing left to do. */ }
+SimpleCV<MLAlgorithm, Metric, MatType, PredictionsType, WeightsType>::SimpleCV(
+    const double validationSize, MIT&& xs, PIT&& ys)
+  : SimpleCV(
+        Base(), validationSize, std::forward<MIT>(xs), std::forward<PIT>(ys))
+{ /* Nothing left to do. */
```
zoq: Can we put the comment on a new line?

src/mlpack/tests/qdafn_test.cpp
```diff
@@ -155,18 +155,18 @@ BOOST_AUTO_TEST_CASE(SerializationTest)
...
   {
     BOOST_REQUIRE_EQUAL(qdafnXml.CandidateSet(i).n_rows,
-        qdafn.CandidateSet(i).n_rows);
+                        qdafn.CandidateSet(i).n_rows);
     BOOST_REQUIRE_EQUAL(qdafnText.CandidateSet(i).n_rows,
-        qdafn.CandidateSet(i).n_rows);
+                        qdafn.CandidateSet(i).n_rows);
     BOOST_REQUIRE_EQUAL(qdafnBinary.CandidateSet(i).n_rows,
-        qdafn.CandidateSet(i).n_rows);
+                        qdafn.CandidateSet(i).n_rows);
```
zoq: hm, this looks odd, not sure, why it aligns the parameter, maybe because it's a macro?

ShikharJ: This was mentioned in the `Design Guidelines` under `Method Declaration`. By default, it should align in the above manner. The previous format was only for those cases where even the above format leads to an exceed on the 80 character line length.

src/mlpack/core/cv/cv_base.hpp
ShikharJ: I think this can be fixed within this PR.

src/mlpack/core/cv/k_fold_cv.hpp
ShikharJ: I'll have to customize the sentence braking, I'll see what can be done.

rcurtin:
We've had a handful of style checking utilities in the past, which have led to the existing style checks Jenkins job that is very useful.  However one thing I have found over the years is that automatic formatters don't really leave any room for exceptions.  There are cases where people will choose to do

```
Function(argument1,
    argument2,
    argument3);
```

for readability and there are other cases where people will choose to do

```
Function(
    argument, argument2, argument3);
```

Both are okay, and an automatic formatter can't effectively choose between the two for a subjective question like this of "which is more readable".  In these cases I think it's just fine for the author of the code to choose what they would like to do.  There are a lot more examples like this too.

So I have to wonder, would it be better if instead of reformatting all the code (again) in this PR, should we instead basically add the utility we have in the jenkins-conf repository to the main mlpack repository so that people can self-lint their code before they submit it?  I think that might be a lot easier.

ShikharJ:
I agree. But the reason I wanted to have this utility, is to automate the changes for bigger PRs. For a few minor changes, probably it doesn't make much sense to use the script as the changes can be made at will in minimal time. However, when we head to larger PRs (like Kris' GAN PR for example), it becomes an additional burden to lint across the files, correcting the style, especially for newer contributors. So ultimately, it boils down to the argument of convenience vs control. I am comfortable either way, what are your views on this @zoq @sbrodehl?

sbrodehl:
I think the point here is not what linter tool we use, but how we use them.

Currently unlinted code causes a failed style check CI test, but this does not affect the build jobs, so they try to build the library every time. We can improve the CI test suite if we condition all builds and test on the style check, so that they are build only in case of no errors.

The output of the current linter ist not very good. First, as a new contributor it is hard to find where the output actually is and second finding the actual error is like finding a needle in a haystack. I am not sure if automating the whole process is a god idea, but I feel like if we can improve how the contributor perceives the linter-errors it is not so important to automate the correction.

So it is possible to only proceed with the CI jobs if the style checks succeeds?
Of course, we could use a client side hook to run the linter pre-push (?!) or on the server pre-receive (?!) - not sure what’s best here.
Anyway, it would be good to be able to run the linter on the client side regardless, so that a contributor can easily check their code (and correct it).

rcurtin:
Both of your perspectives are very valuable---thank you for sharing.  I have not been a new contributor since... a long time now, so I don't know how the experience is for someone new.  From what I can tell we have a number of issues:

**1. Local linting**

The linter that Jenkins uses is right here:

https://github.com/mlpack/jenkins-conf/tree/master/linter

If we simply pull this code into the master repository (and provide documentation on how to use it), then this can allow a contributor to do the style review locally.  Do you think that would be sufficient to help break the irritating push-to-Github-then-debug-the-Jenkins-job cycle?

**2. Automatic fixes vs. pointing out errors**

The PR here provides a script that can actually go through and fix all the code, but like I said earlier, this shouldn't necessary be applied to the whole codebase because of intentional ambiguities in the style guide that allow people to write the code in ways they have specifically chosen for readability.

At the same time, maybe there would be a nice way to run the automatic style fixes on any code that's modified from the master branch, and this could be useful to a contributor who might optionally choose to use it.

**3. Early termination of PR builds**

It might be nice to run the style check first, and conditionally on its passing we could do the other builds.  I wonder if there might be a case where we would need to override a style check failure that can't easily be fixed (sometimes this happens with the static code analysis checks), but as long as we have a way to override it I think this might be a nice improvement.  @sbrodehl: do you think that's possible?  If you have an idea of how to get this set up right, I'd personally be fine with merging it in.

Hopefully I got everything with those 3 things; if there are more concerns this brings up we can add to the list and tackle each issue in discussion.  :)

ShikharJ:
Local linting sounds good to me. There shouldn't be any issues as long as we mention about the Jenkins linter in the `README`.
If we're thinking about bringing the linter into the master repository, implementing an automation script becomes a low priority task in my opinion, as local linting in between the commits should be sufficient.
Early termination of PRs is a good option, in case someone fails to check for the style changes.

zoq:
> 3. Early termination of PR builds
> 
> It might be nice to run the style check first, and conditionally on its passing we could do the other builds.

I guess that depends on how you use the PR, sometimes you just like to get some input about build errors or some input about the modification (especially during GSoC). I think in this case it would be border some to fix potential style issues first before actually building the project.

> The output of the current linter ist not very good. First, as a new contributor
> it is hard to find where the output actually is and second finding the actual
> error is like finding a needle in a haystack. I am not sure if automating the
> whole process is a god idea, but I feel like if we can improve how the
> contributor perceives the linter-errors it is not so important to automate the
> correction.

@sbrodehl  Would be great if you could provide some more input about the difficulties you experienced. Right now you get an initial overview of the style issues with a short description and the line number. If you click the line number or e.g. the ``all`` option you see every issue with the corresponding code snippet.

**- overview:** http://masterblaster.mlpack.org/job/pull-requests%20mlpack%20style%20checks/1346/cppcheckResult/
**- all:** http://masterblaster.mlpack.org/job/pull-requests%20mlpack%20style%20checks/1346/cppcheckResult/source.all/?before=5&after=5

I think the first click could be improved if we forward the user directly to the overview page if you think that would help I'll look into that.

> 1. Local linting
> 
> The linter that Jenkins uses is right here:
> 
> https://github.com/mlpack/jenkins-conf/tree/master/linter
> 
> If we simply pull this code into the master repository (and provide documentation on how to use it), then this can allow a contributor to do the style review locally. 

The problem I see is that the output of the cpplint.py script isn't as convenient as e.g. clang-format, which formats the code for you. I think providing a clang-format script that locally runs on changed files is something that could simplify the overall process if the uses choose to use it. We could also provide a combination of both, e.g. use option one if you like to see potential style errors and option two uses clang-format.

rcurtin:
Maybe a good place to start is a git pre-push script that runs cpplint, what do you guys think?  That could at least let users run the style checks locally, and then after that if we like we can work on a clang-format script that can be used or something like that based on the code here.

zoq:
I haven't looked at pre-push scripts yet, but I guess if we provide a script it would automatically integrate into the users workflow if they clone the repo? Or do I have to active the script? My convern is what happens if the user has some weird python or no python at all.

sbrodehl:
It basically is a script which is run at the beginning of the push, if it succeeds, the 'normal' push continues, if not, the push will be rejected, that is, `git push` will return a non zero exit code and nothing will be pushed to the remote.
The script will be run with every `git push` command, and can be skipped adding the argument `--no-verify`.

So, if we check for a weird python / no python and exit with code 0 in case the linter can't be run, the push continues without linting.

zoq:
Thanks for the clarification, I'm not really sure I like the idea to reject a push for style issues, see my concerns here: 

> > Early termination of PR builds
> It might be nice to run the style check first, and conditionally on its passing we could do the other builds.
> 
> I guess that depends on how you use the PR, sometimes you just like to get some input about build errors or some input about the modification (especially during GSoC). I think in this case it would be border some to fix potential style issues first before actually building the project.

Perhaps this concerns are  unsubstantiated.

rcurtin:
Hm, is it possible that a pre-push (or post-commit) hook could simply print a warning like:

```
Style errors found.  Run scripts/check-style.py to see more details.
```

or something along these lines?  So it wouldn't interrupt a user who is trying to push, but does give them information about the failing style checks that they can then further investigate if they choose.  What do you think?

zoq:
That would work for me, not sure if @sbrodehl, @ShikharJ likes the idea, since at the moment I push it's already checked in. Also, this is just my opinion if you like the idea to always reject the push this is fine, I can configure my git to use ``--no-verify`` as @sbrodehl pointed out.

sbrodehl:
I like the idea @rcurtin, how about something like
```
Style errors found.  Run scripts/check-style.py to see more details. Continue push? (Y/n)
```
so one can abort the push if one doesn't want to push after all?

rcurtin:
I think that could work fine, my only question is can we add some custom git configuration bit to disable those warnings?  i.e. I could do `git config mlpack.stylechecks false`.  I don't know anything about custom variables in git or if that is possible, maybe so, maybe not...

rcurtin:
@ShikharJ: I saw your comment in #1203; I think that in this case if you can implement the idea that @sbrodehl implemented, that would be fine.  It would be great also if we could find out if a git custom configuration option could be used, but if not a Y/n dialog is just fine in my opinion.

ShikharJ:
Closing this for now. `clang-format` isn't probably the best solution.

Issue #1207, opened 2018-01-21
SpikeSlab RBM tests implementation


lozhnikov:
In summer 2017 @kris-singh wrote an implementation of Binary RBM and Spike and Slab RBM. The first one is implemented according to this [article](http://deeplearning.net/tutorial/rbm.html). And the second one is implemented according to this [paper](http://www.jmlr.org/proceedings/papers/v15/courville11a.html).
We reproduced the deeplearning.net MNIST generation test. Here are some samples generated by the binary RBM:

![mnist_out](https://user-images.githubusercontent.com/10090844/35193117-43cb15bc-feaf-11e7-80e7-abbf9b3aeaa0.png)

Moreover we tried these models for classification on [the Digit dataset](http://archive.ics.uci.edu/ml/datasets/Pen-Based+Recognition+of+Handwritten+Digits).

Unfortunately, we didn't implement extensive tests for the spike and slab RBM implementation. It would be nice to reproduce some tests from the original [paper](http://www.jmlr.org/proceedings/papers/v15/courville11a.html) e.g. see section 6.3. "Learning Features for classification".

Ongoing work: [Kris's RBM PR](https://github.com/mlpack/mlpack/pull/1046), [some thoughts on the ssRBM classification test](https://gist.github.com/kris-singh/55a4934c463bbddb3c8d321dd889d194).

ShikharJ:
@lozhnikov I had actually started working on some of Kris' PRs, after the discussion on the [IRC](http://www.mlpack.org/irc/mlpack.20180111.html). I have a branch [here](https://github.com/ShikharJ/mlpack/tree/RBM). Again it contains only minor conflict resolutions. I'll file a PR soon.

lozhnikov:
@ShikharJ ok, feel free to ask questions

Rajiv2605:
Can I work on this issue?

KunvarThaman:
@lozhnikov I'm looking to work on the _Essential Deep learning modules_ project. Is it okay if I work on this issue?

sundar-10:
Hello,
I have seen the 2 papers- Binary RBM and Spike and Slab RBM. And i would like to work on this issue..Is it okay?

PrashantSaroj:
@lozhnikov @ShikharJ @Rajiv2605  I am willing to work on the "Essential Deep learning modules" project. How do I get started?

zoq:
[...]

PR #1208, opened 2018-01-21
Restricted Boltzmann Machine and Spike Slab Restricted Boltzmann Machine

ShikharJ:
Another rebase to master of the [original PR](https://github.com/mlpack/mlpack/pull/1046) with some minor changes.
Relevant: #1207
TODOs:

- [x] Fix build errors
- [ ] Add tests
- [x] Add support for mini-batches


src/mlpack/methods/ann/rbm_impl.hpp
```diff
@@ -0,0 +1,236 @@
...
+    FreeEnergy(arma::Mat<ElemType>&& input)
+{
+  return rbmPolicy.FreeEnergy(std::move(input));
+}
+
+template<typename InitializationRuleType, typename RBMPolicy>
+double RBM<InitializationRuleType, RBMPolicy>::Evaluate(
+    const arma::Mat<ElemType>& /* parameters*/,
+    const size_t i,
+    const size_t batchSize)
```
ShikharJ: @lozhnikov This argument is currently unused, can anything be done here? Same with [here](https://github.com/mlpack/mlpack/pull/1208/files#diff-4ba957d6d102420d5311acd6af750ca8R174)

src/mlpack/methods/ann/rbm_impl.hpp
```diff
@@ -0,0 +1,236 @@
...
+
+    negativeGradient += tempNegativeGradient;
+  }
+  output = ((negativeGradient / negSteps) - positiveGradient);
+}
+
+template<typename InitializationRuleType, typename RBMPolicy>
+void RBM<InitializationRuleType, RBMPolicy>::Shuffle()
+{
+  // TODO: Complete This
```
ShikharJ: I am unsure what needs to be done here. Please advise.

src/mlpack/methods/ann/rbm_impl.hpp
ShikharJ: Also, should this be updated in `..ann/rbm/` subdirectory files as well?

lozhnikov: Actually, it depends on the optimizers. You can just comment out the name of the argument e.g.
```
const size_t i,
const size_t /*batchSize*/)
```
in order to avoid the warning.

src/mlpack/methods/ann/rbm_impl.hpp
lozhnikov: Take a look at ffn_impl.hpp:267. The function should shuffle the order of function visitation.

ShikharJ:
@lozhnikov The accuracy for `BinaryRBM` seems to lag behind `Softmax` accuracy. It'll take me some time to read through the paper, and look for optimizations. Do you have any suggestions for improvement?

lozhnikov:
@ShikharJ Could you elaborate a bit? What do you mean by "the accuracy for BinaryRBM"?

src/mlpack/tests/rbm_network_test.cpp
```diff
@@ -0,0 +1,250 @@
...
+   testLabels);
+  std::cout << "Softmax Accuracy = " << classificationAccuray << std::endl;
+
+  L_BFGS rbmOptimizer(numBasis, numIterations);
+  SoftmaxRegression rbmRegressor(XRbm, trainLabels, numClasses,
+        0.001, false, rbmOptimizer);
+  double rbmClassificationAccuracy = rbmRegressor.ComputeAccuracy(YRbm,
+      testLabels);
+  std::cout << "RBM Accuracy = " << rbmClassificationAccuracy << std::endl;
+  BOOST_REQUIRE_GE(rbmClassificationAccuracy, classificationAccuray);
```
ShikharJ: @lozhnikov I was referring to this test:
```
Softmax Accuracy = 76.2382
RBM Accuracy = 70.0612
/home/travis/build/mlpack/mlpack/src/mlpack/tests/rbm_network_test.cpp(118): fatal error in "ClassificationTest": critical check rbmClassificationAccuracy >= classificationAccuray failed [70.061213132999441 < 76.238174735670569]
ssRBM Accuracy = 80.801335559265439
```

ShikharJ: @lozhnikov This test seemed to be passing before the rebase, when `MiniBatchSGD` was being used. Any idea what can be done here?

src/mlpack/methods/ann/rbm_impl.hpp
```diff
@@ -0,0 +1,237 @@
...
+  output = ((negativeGradient / negSteps) - positiveGradient);
+}
+
+template<typename InitializationRuleType, typename RBMPolicy>
+void RBM<InitializationRuleType, RBMPolicy>::Shuffle()
+{
+  predictors = predictors.cols(arma::shuffle(arma::linspace<arma::uvec>(0,
+      predictors.n_cols - 1,
+      predictors.n_cols)));
+}
```
ShikharJ: @lozhnikov I'm not confident if this is the correct implementation. Please advise.

src/mlpack/methods/ann/rbm_impl.hpp
```diff
@@ -0,0 +1,237 @@
...
+
+template<typename InitializationRuleType, typename RBMPolicy>
+void RBM<InitializationRuleType, RBMPolicy>::Gradient(
+    arma::Mat<ElemType>& /*parameters*/,
+    const size_t input,
+    arma::Mat<ElemType>& output,
+    const size_t /*batchSize*/)
+{
+  positiveGradient.zeros();
```
lozhnikov: Sorry, I didn't see your comment. You should repeat the procedure below for each feature from `input` to `input + batchSize - 1`. The present implementation omits all features except the first one. That's why the accuracy is wrong.

src/mlpack/methods/ann/rbm_impl.hpp
```diff
@@ -0,0 +1,237 @@
...
+  return rbmPolicy.FreeEnergy(std::move(input));
+}
+
+template<typename InitializationRuleType, typename RBMPolicy>
+double RBM<InitializationRuleType, RBMPolicy>::Evaluate(
+    const arma::Mat<ElemType>& /* parameters*/,
+    const size_t i,
+    const size_t /*batchSize*/)
+{
+  if (!useMonitoringCost)
```
lozhnikov: The same issue as in `Gradient()`.

src/mlpack/methods/ann/rbm_impl.hpp
lozhnikov: Looks like the implementation is correct.

src/mlpack/methods/ann/rbm_impl.hpp
ShikharJ: @lozhnikov I've updated the changes here. The `BinaryRBM` Classification test is working well now, but seemingly the `ssRBM` test has taken a hit (went down from an `80.6` to `75.2`, a little below the intended baseline). Can you please check if the implementation below is correct?

ShikharJ:
@lozhnikov I checked out the CIFAR-10 dataset (which Kris' had mentioned would be about 178 MB in size). I'm presuming we don't want to commit the dataset, so should I rather create a script to fetch the dataset and have the test run on a separate `Travis` slot? Also, can you please tell me the reason behind the `ssRBM` classification benchmark of `76.18` [here](https://github.com/mlpack/mlpack/pull/1208/files#diff-75481ca8e30275b5f28119dddd36ffecR204)?

lozhnikov:
@ShikharJ I didn't dig into that yet. I looked through the Travis output:
```
105/133 Test #119: RbmNetworkTest ...................***Failed    0.20 sec
Running 3 test cases...
unknown location(0): fatal error in "ClassificationTest": std::logic_error: addition: incompatible matrix dimensions: 100x10 and 100x1
mlpack_test: /home/travis/build/mlpack/mlpack/src/mlpack/methods/ann/rbm/spike_slab_rbm_policy_impl.hpp:301: void mlpack::ann::SpikeSlabRBMPolicy<DataType>::SampleHidden(DataType&&, DataType&&) [with DataType = arma::Mat<double>]: Assertion `input.n_elem == visibleSize' failed.
unknown location(0): fatal error in "ssRBMClassificationTest": signal: SIGABRT (application abort requested)
```
Looks like, some matrix dimensions are incorrect. So, I think it is a good place to start.

Of course we don't want co commit the dataset. Moreover, the test should take too much CPU time. So, I guess the best way is to create a tutorial (e.g. http://mlpack.org/docs/mlpack-2.2.5/doxygen/tutorials.html).

src/mlpack/methods/ann/rbm_impl.hpp
```diff
@@ -0,0 +1,241 @@
...
+  // Collect the negative samples
+  rbmPolicy.PositivePhase(std::move(predictors.cols(input,
+                                                    input + batchSize - 1)),
+                          std::move(positiveGradient));
+
+  negativeGradient.zeros();
+
+  for (size_t i = 0; i < negSteps; i++)
+  {
+    Gibbs(std::move(predictors.cols(input, input + batchSize - 1)),
```
ShikharJ: `Looks like, some matrix dimensions are incorrect`.
Actually, this was because here, I tried to directly pass a batch of columns, which is wrong because internally, the present code expects input as a single column in one pass. For, now I can just introduce a `for` loop over the `batchSize`, and that should work. Do we want to leave that as it is, or should the implementation incorporate the use case of input being a bunch of columns rather than a single column? cc @zoq

zoq: hm, you should be able to pass a matrix where each column is another input, at least this is what we do for the FFN class and we do something similar for the RNN class, which uses arma::cube but just to handle data sequences.The format of the data should be as follows:

- each slice should correspond to a time step
- each column should correspond to a data point
- each row should correspond to a dimension

anyway, you are right that in case of a single input which is multidimensional it's represented as a single column, maybe it's worth to use ``arma::cube`` here as well. Let me know what you think.

zoq:
> I checked out the CIFAR-10 dataset (which Kris' had mentioned would be about 178 MB in size). I'm presuming we don't want to commit the dataset, so should I rather create a script to fetch the dataset and have the test run on a separate Travis slot?

We should test the code on smaller datasets, to keep the test time low, however we could write an example based on the CIFAR-10 dataset and push it to the models repository (https://github.com/mlpack/models).

src/mlpack/methods/ann/rbm_impl.hpp
ShikharJ: It sounds logical to me, I'll have to change the implementation a bit. @lozhnikov What do you prefer?

ShikharJ: Ping @lozhnikov.

lozhnikov: @ShikharJ Sorry for the slow response, I was preparing a report on gas dynamics. I agree with Marcus. The idea looks good to me.

src/mlpack/tests/rbm_network_test.cpp
```diff
@@ -0,0 +1,250 @@
...
+
+  size_t numRBMIterations = trainData.n_cols * numEpoches;
+  numRBMIterations /= batchSize;
+
+  StandardSGD msgd(0.06, batchSize, numRBMIterations, 0, true);
+  model.Reset();
+  model.Policy().VisibleBias().ones();
+  model.Policy().HiddenBias().ones();
+  // Test the reset function.
+  model.Train(trainData, msgd);
```
ShikharJ: @rcurtin Was the optimizer API changed recently? This seems to be failing to build here as follows:
```
/home/shikhar/mlpack/src/mlpack/core/optimizers/sgd/sgd_impl.hpp:118:22: error: no matching function for call to ‘mlpack::optimization::Function<mlpack::ann::RBM<mlpack::ann::GaussianInitialization, mlpack::ann::BinaryRBMPolicy<> > >::EvaluateWithGradient(arma::mat&, size_t&, arma::mat&, const size_t&)’
     overallObjective += f.EvaluateWithGradient(iterate, currentFunction,
```

zoq: Yes, for more information see: https://github.com/mlpack/mlpack/pull/1252/ realistically you just have to add

```c++
double Evaluate(const arma::mat& parameters,
                const size_t begin,
                const size_t batchSize)
{
   return Evaluate(parameters, begin, batchSize, true);
}
```

see for an example: https://github.com/mlpack/mlpack/pull/1252/files#diff-a778189f4c83e20d3e29c6c03039abfd

ShikharJ: @zoq I've updated the changes here, but it still doesn't fix the issue.

src/mlpack/methods/ann/rbm/rbm_impl.hpp
```diff
@@ -0,0 +1,229 @@
...
+    {
+      // Mean Squared Error
+      rbmPolicy.SampleHidden(std::move(predictors.col(i)),
+          std::move(hiddenReconstruction));
+      rbmPolicy.SampleVisible(std::move(hiddenReconstruction),
+          std::move(visibleReconstruction));
+      return arma::accu(arma::pow(visibleReconstruction -
+                                  predictors.col(i),
+          2)) / hiddenReconstruction.n_rows;
+    }
```
ShikharJ: @lozhnikov Could you comment why are we using mean squared error here? I don't see it being mentioned anywhere in the deepearning.net tutorial?

src/mlpack/methods/ann/rbm/binary_rbm_policy_impl.hpp
```diff
@@ -0,0 +1,167 @@
...
+  DataType hiddenBiasGrad = DataType(gradient.memptr() + weightGrad.n_elem,
+      hiddenSize, 1, false, false);
+
+  DataType visibleBiasGrad = DataType(gradient.memptr() + weightGrad.n_elem +
+      hiddenBiasGrad.n_elem, visibleSize, 1, false, false);
+
+  HiddenMean(std::move(negativeSamples), std::move(hiddenBiasGrad));
+  weightGrad = hiddenBiasGrad * negativeSamples.t();
+  visibleBiasGrad = negativeSamples;
+}
```
ShikharJ: @lozhnikov `PositivePhase()` and `NegativePhase()` have the same implementation. Is there something wrong here?

lozhnikov: @Wenhao-H I think it's correct. I just didn't notice that:) In this case you can get rid of either `PositivePhase()` or `NegativePhase()`.

src/mlpack/methods/ann/rbm/rbm_impl.hpp
lozhnikov: @ShikharJ Actually, I think you can get rid of
```
if (persistence)
{
  return rbmPolicy.Evaluate(predictors, i, batchSize);
}
```
either.

src/mlpack/methods/ann/rbm/rbm_impl.hpp
```diff
@@ -0,0 +1,343 @@
...
+  typename InitializationRuleType,
+  typename DataType,
+  typename PolicyType
+>
+template<typename Policy>
+typename std::enable_if<std::is_same<Policy, BinaryRBM>::value, double>::type
+RBM<InitializationRuleType, DataType, PolicyType>::FreeEnergy(
+    arma::Mat<ElemType>&& input)
+{
+  preActivation = arma::log(1 + arma::trunc_exp((weight * input) + hiddenBias));
```
zoq: This should fix the batch size issue:

```c++
preActivation = (weight * input);
preActivation.each_col() += hiddenBias;
preActivation = arma::log(1 + arma::trunc_exp(preActivation));
return -(arma::accu(preActivation) + arma::dot(input, arma::repmat(visibleBias, 1, input.n_cols)));
```

src/mlpack/methods/ann/rbm/rbm_impl.hpp
```diff
@@ -0,0 +1,343 @@
...
+  typename InitializationRuleType,
+  typename DataType,
+  typename PolicyType
+>
+template<typename Policy>
+typename std::enable_if<std::is_same<Policy, BinaryRBM>::value, void>::type
+RBM<InitializationRuleType, DataType, PolicyType>::VisibleMean(DataType&& input,
+    DataType&& output)
+{
+  output = weight.t() * input + visibleBias;
```
zoq:
```c++
output = weight.t() * input;
output.each_col() += visibleBias;
```

the same applies to the ``HiddenMean`` method.

ShikharJ:
@lozhnikov @zoq I would appreciate if we could review and merge, and implement the additional `SSRBM` test and other optimizations in another PR. It took a lot more time and effort than I was anticipating, to get the code corrected and cleaned, and I don't want this to get held up, as the changes are substantial.

src/mlpack/methods/ann/rbm/CMakeLists.txt
```diff
@@ -0,0 +1,17 @@
+# Define the files we need to compile.
+# Any file not in this list will not be compiled into mlpack.
+set(SOURCES
+  rbm.hpp
```
zoq: hm, not sure creating an extra directory is what we should do here, I think we could either move the directory in the method folder or move the rbm class into the ann folder. Let me know what you think.

Providing an executable for the rbm class is a lot easier since the model structure is somewhat static.

src/mlpack/methods/ann/rbm/rbm.hpp
```diff
@@ -0,0 +1,447 @@
...
+#define MLPACK_METHODS_ANN_RBM_RBM_HPP
+
+#include <mlpack/core.hpp>
+#include <mlpack/methods/ann/rbm/rbm_policies.hpp>
+
+namespace mlpack {
+namespace ann /** Artificial Neural Network. */ {
+
+/**
+ * The RBM class.
```
zoq: We could provide a more in detail description what this class does.

src/mlpack/methods/ann/rbm/rbm.hpp
```diff
@@ -0,0 +1,447 @@
...
+class RBM
+{
+ public:
+  using NetworkType = RBM<InitializationRuleType, DataType, PolicyType>;
+  typedef typename DataType::elem_type ElemType;
+
+  /**
+   * Initialize all the parameters of the network using initializeRule.
+   *
+   * @tparam initializeRule InitializationRule object for
```
zoq: This isn't a template parameter, also ``predictors`` should be swapped with ``initializeRule``.

src/mlpack/methods/ann/rbm/rbm.hpp
```diff
@@ -0,0 +1,447 @@
...
+   *         initializing the network parameter.
+   * @param predictors Training data to be used.
+   * @param visibleSize Number of visible neurons.
+   * @param hiddenSize Number of hidden neurons.
+   * @param batchSize Batch size to be used for training.
```
zoq: Should we increase the default batch size?

src/mlpack/methods/ann/rbm/rbm.hpp
```diff
@@ -0,0 +1,447 @@
...
+   * @param numSteps Number of Gibbs Sampling steps.
+   * @param negSteps Number of negative samples to average negative gradient.
```
zoq: Can you add a comment for the other parameter as well?

src/mlpack/methods/ann/rbm/rbm.hpp
```diff
@@ -0,0 +1,447 @@
...
+  typename std::enable_if<std::is_same<Policy, BinaryRBM>::value, void>::type
+  Reset();
+
+  // Reset the network.
+  template<typename Policy = PolicyType>
+  typename std::enable_if<std::is_same<Policy, SpikeSlabRBM>::value, void>::type
+  Reset();
+
+  /**
+   * Train the feed-forward network on the given input data.
```
zoq: I guess this should train the RBM on the given data.

src/mlpack/methods/ann/rbm/rbm.hpp
```diff
@@ -0,0 +1,447 @@
...
+  /**
+   * This function calculates the free energy of the BinaryRBM.
+   * The free energy is given by:
+   * $-b^Tv - \sum_{i=1}^M log(1 + e^{c_j+v^TW_j})$.
+   *
+   * @param input The visible neurons.
+   */
+  template<typename Policy = PolicyType>
+  typename std::enable_if<std::is_same<Policy, BinaryRBM>::value, double>::type
+  FreeEnergy(arma::Mat<ElemType>&& input);
```
zoq: hm, do you think it would make sense to put this into the policy itself, we would have to pass the weights and the bias, but perhaps this would make the class cleaner?

src/mlpack/methods/ann/rbm/rbm_impl.hpp
```diff
@@ -0,0 +1,351 @@
...
+>
+template<typename Policy>
+typename std::enable_if<std::is_same<Policy, BinaryRBM>::value, double>::type
+RBM<InitializationRuleType, DataType, PolicyType>::FreeEnergy(
+    arma::Mat<ElemType>&& input)
+{
+  preActivation = (weight * input);
+  preActivation.each_col() += hiddenBias;
+  preActivation = arma::log(1 + arma::trunc_exp(preActivation));
+  return -(arma::accu(preActivation) + arma::dot(input, arma::repmat(
```
zoq: Not sure the compiler will inline the expression (``arma::log(1 + arma::trunc_exp(preActivation));``) to avoid another copy.

src/mlpack/methods/ann/rbm/rbm_impl.hpp
```diff
@@ -0,0 +1,351 @@
...
+  typename DataType,
+  typename PolicyType
+>
+template<typename Policy>
+typename std::enable_if<std::is_same<Policy, BinaryRBM>::value, void>::type
+RBM<InitializationRuleType, DataType, PolicyType>::SampleVisible(
+    arma::Mat<ElemType>&& input,
+    arma::Mat<ElemType>&& output)
+{
+  VisibleMean(std::move(input), std::move(output));
```
zoq: Do you think, we could just use a reference for the ``input`` and ``output`` term?

src/mlpack/methods/ann/rbm/rbm.hpp
```diff
@@ -0,0 +1,447 @@
...
+  /**
+   * This function does the k-step Gibbs Sampling.
+   *
+   * @param input Input to the Gibbs function.
+   * @param output Used for storing the negative sample.
+   * @param steps Number of Gibbs Sampling steps taken.
+   */
+  void Gibbs(arma::Mat<ElemType>&& input,
+             arma::Mat<ElemType>&& output,
+             size_t steps = SIZE_MAX);
```
zoq: To be consistent with the rest of the codebase let's use ``const size_t steps`` here.

src/mlpack/methods/ann/rbm/spike_slab_rbm_impl.hpp
```diff
@@ -0,0 +1,311 @@
...
+  negativeSamples.set_size(visibleSize, batchSize);
+  visibleMean.set_size(visibleSize, 1);
+  spikeMean.set_size(hiddenSize, 1);
+  spikeSamples.set_size(hiddenSize, 1);
+  slabMean.set_size(poolSize, hiddenSize);
+
+  // Weight shape D * K * N
+  weightCube = arma::Cube<ElemType>(parameter.memptr(), visibleSize, poolSize,
+      hiddenSize, false, false);
+  // spike bias shape N * 1
```
zoq: Can you start with a capital ``S`` here?

src/mlpack/methods/ann/rbm/spike_slab_rbm_impl.hpp
```diff
@@ -0,0 +1,311 @@
...
+  spikeBias = DataType(parameter.memptr() + weight.n_elem, hiddenSize, 1,
+      false, false);
+  // visible penalty 1 * 1 => D * D(when used)
+  visiblePenalty = DataType(parameter.memptr() + weight.n_elem +
```
zoq: See comment above.

src/mlpack/methods/ann/rbm/spike_slab_rbm_impl.hpp
```diff
@@ -0,0 +1,311 @@
...
+  typename InitializationRuleType,
+  typename DataType,
+  typename PolicyType
+>
+template<typename Policy>
+typename std::enable_if<std::is_same<Policy, SpikeSlabRBM>::value, double>::type
+RBM<InitializationRuleType, DataType, PolicyType>::FreeEnergy(
+    arma::Mat<ElemType>&& input)
+{
+  ElemType freeEnergy = 0.5 * visiblePenalty(0) * arma::accu(input.t() * input);
```
zoq: We could also use ``arma::dot(input, input)`` instead of `` arma::accu(input.t() * input)`` here, it's not really shorter.

src/mlpack/methods/ann/rbm/spike_slab_rbm_impl.hpp
```diff
@@ -0,0 +1,311 @@
...
+
+  freeEnergy -= 0.5 * hiddenSize * poolSize *
+      std::log((2.0 * M_PI) / slabPenalty);
+
+  for (size_t i = 0; i < hiddenSize; i++)
+  {
+    ElemType sum = 0;
```
zoq: Perhaps I missed something but I guess we can directly use ``ElemType = arma::accu(arma::square(input.t() * weightCube.slice(i))) / (2.0 * slabPenalty);`` here.

src/mlpack/methods/ann/rbm/CMakeLists.txt
ShikharJ: This was done just to be in line with our gan implementation, but I think, since `GAN`s and `RBM`s are sub-variants of `ANN`s it makes sense to keep a folder here. I'm open to suggestions though.

src/mlpack/methods/ann/rbm/rbm.hpp
ShikharJ: By default, `RBM`s are expected to be stochastic, so I'd prefer it stay that way.

src/mlpack/methods/ann/rbm/rbm_impl.hpp
ShikharJ: I'll change this to directly return the expression, just to be sure.

src/mlpack/methods/ann/rbm/rbm_impl.hpp
ShikharJ: Hmm, to do that, we'll have to change the signatures of all the intermediate functions to use references. How should we proceed here?

src/mlpack/methods/ann/rbm/rbm.hpp
ShikharJ: One of my concerns is with the inconvenience of using the policy paradigm and passing parameters, and then the number of duplicate matrices and objects that Kris' original code had.

src/mlpack/methods/ann/rbm/CMakeLists.txt
zoq: I get your point, so I guess if we are going to write an executable it's in the rbm folder as well?

ShikharJ: Yes.

src/mlpack/tests/serialization_test.cpp
```diff
@@ -1690,4 +1693,97 @@ BOOST_AUTO_TEST_CASE(HoeffdingTreeTest)
...
+  CheckMatrices(Rbm.Weight(), RbmXml.Weight());
+  CheckMatrices(Rbm.Weight(), RbmText.Weight());
+  CheckMatrices(Rbm.Weight(), RbmBinary.Weight());
+}
+
+/**
+ * Build a ssRBM, then save it and make sure the parameters of the
+ * all the RBM are equal.
+ */
+
```
zoq: Looks like we can remove the extra line here.

src/mlpack/tests/serialization_test.cpp
```diff
@@ -1690,4 +1693,97 @@ BOOST_AUTO_TEST_CASE(HoeffdingTreeTest)
...
+{
+  arma::mat data;
+  size_t hiddenLayerSize = 5;
+  data.randu(3, 100);
+  double slabPenalty = 1;
+  double radius, tempRadius;
+  for (size_t i = 0; i < data.n_cols; i++)
+  {
+    tempRadius = arma::norm(data.col(i));
+    if (radius < tempRadius)
```
zoq: We should initialize radius with ``arma::norm(data.col(0));`` and start with ``i = 1`` in the for loop, to avoid unexpected behavior. Let me know what you think.

src/mlpack/methods/ann/rbm/rbm.hpp
```diff
@@ -0,0 +1,458 @@
...
+  const arma::Mat<ElemType>& Parameters() const { return parameter; }
+  //! Modify the parameters of the network.
+  arma::Mat<ElemType>& Parameters() { return parameter; }
+
+  //! Return the weights of the network.
+  const DataType& Weight() const { return weight; }
+  //! Modify the weights of the network.
+  DataType& Weight() { return weight; }
+  //! Get the weight of the network.
+  arma::cube const& WeightCube() const { return weightCube; }
```
zoq: Do you think we can remove the function? Not sure it's necessary, we could do the transformation based on the ``Weight`` function as well.

src/mlpack/tests/serialization_test.cpp
```diff
@@ -1690,4 +1693,97 @@ BOOST_AUTO_TEST_CASE(HoeffdingTreeTest)
   }
 }
 
+/**
+ * Build a Binary RBM, then save it and make sure the parameters of the
+ * all the RBM are equal.
+ */
+
```
zoq: See comment below.

src/mlpack/tests/rbm_network_test.cpp
```diff
@@ -0,0 +1,246 @@
...
+  for (size_t i = 0; i < freeEnergy.n_elem; i++)
+    BOOST_REQUIRE_CLOSE(calculatedFreeEnergy(i), freeEnergy(i), 1e-3);
+}
+
+BOOST_AUTO_TEST_CASE(MiscTest)
+{
+  /**
+   * Train and evaluate a Vanilla network with the specified structure.
+   */
+
```
zoq: Can we remove the extra empty line here.

src/mlpack/tests/rbm_network_test.cpp
```diff
@@ -0,0 +1,246 @@
...
+  calculatedFreeEnergy.zeros();
+  for (size_t i = 0; i < trainData.n_cols; i++)
+  {
+    calculatedFreeEnergy(i) = model.FreeEnergy(std::move(trainData.col(i)));
+  }
+
+  for (size_t i = 0; i < freeEnergy.n_elem; i++)
+    BOOST_REQUIRE_CLOSE(calculatedFreeEnergy(i), freeEnergy(i), 1e-3);
+}
+
```
zoq: Missing comment above the test case.

src/mlpack/tests/rbm_network_test.cpp
```diff
@@ -0,0 +1,246 @@
...
+void BuildVanillaNetwork(MatType& trainData,
+                         const size_t hiddenLayerSize)
+{
+  MatType output;
+  GaussianInitialization gaussian(0, 0.1);
+  RBM<GaussianInitialization, MatType, BinaryRBM> model(trainData, gaussian,
+      trainData.n_rows, hiddenLayerSize, 1, 1, 1, 2, 8, 1, true);
+
+  model.Reset();
+  // Set the parameters from a learned RBM Sklearn random state 23
```
zoq: Pedantic style issue; missing ``.`` at the end.

src/mlpack/tests/rbm_network_test.cpp
```diff
@@ -0,0 +1,246 @@
...
+  model.Parameters() = MatType(
+      "-0.23224054, -0.23000632, -0.25701271, -0.25122418, -0.20716651,"
+      "-0.20962217, -0.59922456, -0.60003836, -0.6, -0.625, -0.475;");
+
+  // Check free energy
```
zoq: See comment above.

src/mlpack/tests/rbm_network_test.cpp
```diff
@@ -0,0 +1,246 @@
...
+  arma::Mat<float> freeEnergy = MatType(
```
zoq: Do you see any reason do use ``float`` instead of ``double`` here?

src/mlpack/tests/rbm_network_test.cpp
```diff
@@ -0,0 +1,246 @@
...
+      "-0.87523715, 0.50615066, 0.46923476, 1.21509084;");
+  arma::vec calculatedFreeEnergy(4);
```
zoq: we could use ``calculatedFreeEnergy(4, arma::fill::zeros);`` to avoid the ``zeros();`` call in the next line.

src/mlpack/tests/rbm_network_test.cpp
```diff
@@ -0,0 +1,246 @@
...
+  const size_t numBasis = 5; // Parameter required for L-BFGS algorithm.
+  const size_t numIterations = 100; // Maximum number of iterations.
+
+  L_BFGS ssRbmOptimizer(numBasis, numIterations);
+  SoftmaxRegression ssRbmRegressor(XRbm, trainLabels, numClasses,
+        0.001, false, ssRbmOptimizer);
+  double ssRbmClassificationAccuracy = ssRbmRegressor.ComputeAccuracy(
+      YRbm, testLabels);
+
+  BOOST_REQUIRE_GE(ssRbmClassificationAccuracy, 76.18);
```
zoq: We should comment on ``76.18`` for future references.

src/mlpack/tests/rbm_network_test.cpp
```diff
@@ -0,0 +1,246 @@
+/**
+ * @file rbm_network_test.cpp
+ * @author Kris Singh
+ * @author Shikhar Jaiswal
+ *
+ * Tests the RBM Network
+ *
+ * digits dataset source:
```
zoq: Up to you, but I think we can remove the reference here. I guess it's more likely, that someone will read the class notes or the docs.

src/mlpack/tests/CMakeLists.txt
```diff
@@ -190,6 +192,14 @@ add_custom_command(TARGET mlpack_test
   COMMAND ${CMAKE_COMMAND} -E tar xjpf mnist_first250_training_4s_and_9s.tar.bz2
   WORKING_DIRECTORY ${PROJECT_BINARY_DIR}
 )
+add_custom_command(TARGET mlpack_test
+  POST_BUILD
+  COMMAND ${CMAKE_COMMAND} -E tar xjpf digits_train.tar.bz2
```
zoq: We could merge this with the mnist block, what do you think?

src/mlpack/methods/ann/rbm/spike_slab_rbm_impl.hpp
```diff
@@ -0,0 +1,310 @@
...
+    }
+  }
+
+  if (k == numMaxTrials)
+  {
+    Log::Warn << "Outputs are still not in visible unit "
+        << arma::norm(output, 2)
+        << " terminating optimization."
+        << std::endl;
+    return;
```
zoq: Looks like we could remove the ``return``statement here.

src/mlpack/methods/ann/rbm/rbm.hpp
ShikharJ: This is necessary for the `SSRBM` class, as it uses cubic parameters (unlike matrix parameters).

src/mlpack/tests/rbm_network_test.cpp
ShikharJ: Just to test our implementation of `arma::Mat<float>` as well. By default, `arma::mat` is a `arma::Mat<double>`.

src/mlpack/tests/rbm_network_test.cpp
ShikharJ: I'll rather keep it, in case someone wants to find the rationale behind the test.

src/mlpack/methods/ann/rbm/rbm.hpp
zoq: Sure, but it does use the internal ``weightCube`` parameter and not the ``WeightCube()``method right?

src/mlpack/tests/rbm_network_test.cpp
zoq: I see, thanks for the quick clarification, I think you put so much time into the code, that you become the expert.

src/mlpack/tests/rbm_network_test.cpp
zoq: Ah, since we replicate the results from the paper, good idea.

zoq:
Thanks for another great contribution!

PR #1209, opened 2018-01-21
Clean up CLI singleton implementation

rcurtin:
A discussion I had on Friday made it clear that the CLI singleton implementation is not optimal or even particularly good.  This minor refactoring obviates the need for the CLIDeleter class by putting the CLI singleton as a local static variable.


zoq:
Can you share some of the key points from the discussion?

rcurtin:
Basically, using a raw pointer for a singleton makes it difficult to clean up---that's why the `CLIDeleter` class existed from before.  If you can avoid explicitly allocating an object, then there's no need for `CLIDeleter`.  So using something like a `unique_ptr<>` might be an option, but an inline static object is cleaner in my opinion (and the opinion of the other guy in the discussion).  An inline static object is initialized at the time the method is first called, and destroyed during program termination (fortunately, because the `Log` objects and `CLI` objects are pretty decoupled, destruction order doesn't matter).  So this allows us to remove the `CLIDeleter` object and even the `NULL` check in `GetSingleton()`, which is nice.  It had been a long time since I had seen that, so I wish I had remembered earlier so I could have cleaned up the code years ago. :)

zoq:
Thanks for the clarification, no more comments from my side.

rcurtin:
Ok, if we both approve it I will merge it then.

PR #1210, opened 2018-01-23
adding BOOST_LIBRARYDIR info

ajkl:


rcurtin:
Hmm, I always avoided adding documentation about `BOOST_LIBRARYDIR` because I thought `BOOST_ROOT` was always good enough to find the Boost libraries.  Maybe this is not the case?  If there's a case where `BOOST_ROOT` doesn't work, maybe that needs to be removed or replaced entirely with `BOOST_LIBRARYDIR`, but I'm not sure.  Can you elaborate on why you needed this? :)

ajkl:
No good reason. I saw an error which mentioned it, so thought that was missed from the docs. If its not recommended, you can ignore this change.

rcurtin:
Hm, let's go ahead and close this then if that is okay with you.  If there's a reason to point users towards specifying `BOOST_LIBRARYDIR`, we should, but at the moment I don't know if there is one.  Should we delete the branch?

PR #1211, opened 2018-01-23
Binding test for hoeffding_tree

haritha1313:


src/mlpack/tests/main_tests/hoeffding_tree_test.cpp
```diff
@@ -0,0 +1,356 @@
...
+
+/**
+ * Check whether providing labels explicitly and extracting from last
+ * dimension give the same output.
+ */
+BOOST_AUTO_TEST_CASE(HoeffdingTreeLabelLessTest)
+{
+  arma::mat inputData;
+  DatasetInfo info;
+  if (!data::Load("tae.csv", inputData, info))
```
zoq: Do you think we could reuse an existing dataset here e.g. thyroid or reconstruct the iris dataset, the repo size is already quite high, so I think whenever we can we should avoid adding datasets. Let me know what you think.

haritha1313: Of course, a dataset with last column as labels was needed, but modifying iris would do the trick too. :+1:

haritha1313:
@rcurtin Thanks for the review. I can work on adding the tests, no problem. Let this PR remain open, I'll commit them in a few days time since I'm also trying to implement something in `src/mlpack/methods/ann` (and have exams coming too).

rcurtin:
Sure, no huge hurry.  Good luck with your exams! :)

rcurtin:

haritha1313:
@rcurtin Regarding the test for numeric_split_strategy, I don't think there exists any method to access the number of children of any node in the HoeffdingTreeModel class. Shall I write a utility function similar to HoeffdingTreeModel::NumNodes(), which can count NumChildren() or is there some other way around?

rcurtin:
Hmm, I think that `NumNodes()` should work fine for you here---it should print the full number of nodes in the tree, which is sufficient for the test.  If the tree only has one level (which you can control by passing in the right amount of data, or setting `max_samples` carefully), then I think it will be fine.  Let me know you'd like me to clarify what I mean; I hope what I wrote makes sense but I am not sure. :)

haritha1313:
I think the tests are working fine although I can't understand the travis build problem. Please let me know if any changes are to be made.

zoq:
We are working on a fix for the travis build: https://github.com/travis-ci/travis-ci/issues/9129 and https://github.com/mlpack/mlpack/pull/1221.

src/mlpack/tests/main_tests/hoeffding_tree_test.cpp
```diff
@@ -622,6 +622,115 @@ BOOST_AUTO_TEST_CASE(HoeffdingPassesTest)
...
+
+  SetInputParam("numeric_split_strategy", (string) "domingos");
+  SetInputParam("max_samples", 50);
+  SetInputParam("bins", 10);
+
+  mlpackMain();
+
+  // Check that both models have different number of nodes.
+  BOOST_CHECK_PREDICATE(std::not_equal_to<int>(),
+      ((CLI::GetParam<HoeffdingTreeModel*>("output_model"))->NumNodes())(nodes));
```
rcurtin: Why not just use `BOOST_REQUIRE_NE(...)` here?  Seems to me like that would be simpler.

rcurtin:
All of the tests look good to me, so if you want to handle the one comment I left I think we can get this merged.  Thanks for the nice tests! :)

haritha1313:
I have made the edit (some silly confusion caused it). It is great that the travis issue is being solved, failing travis build was becoming a persistent thing. :)

rcurtin:

PR #1214, opened 2018-01-24
Python memory fix

### LINKED ISSUE #1201: Memory leaks in the python wrappers
I ran knn several times iteratively within a loop and the memory used by python keeps increasing until it crashes.
Download the dataset
https://www.cs.toronto.edu/~kriz/cifar-10-python.tar.gz
And then run this script

```
from mlpack import knn
import copy
import cPickle
import numpy as np

class CIFAR10:
    def __init__(self, n_batches=1):
        def unpickle(file):
            with open(file, 'rb') as fo:
                dict = cPickle.load(fo)
                return dict
        directory = './cifar-10/cifar-10-batches-py/'
        self.data=[]
        for i in range(10):
            self.data.append([])
        for i in xrange(1,n_batches+1):
            d = unpickle(directory+"data_batch_"+str(i))
            index=[]
            data=[]
            for i in range(10):
                index.append([])
                data.append([])

            for j in range(len(d["labels"])):
                index[d["labels"][j]].append(j)

            for j in range(10):
                data[j]=d["data"][index[j], :]
            for j in range(10):
                self.data[j].append(data[j])
        for i in range(10):
            self.data[i]=np.vstack(self.data[i])
    def Data(self, class_id):
        length=int(0.9*len(self.data[class_id]))
        return self.data[class_id][0:length,: ]

    def TestData(self, class_id):
        length=int(0.9*len(self.data[class_id]))
        return self.data[class_id][length+1:,:]

data = CIFAR10(n_batches=5)
train=data.Data(0)
iteration = 0
while True:
    nbrs_plus = knn(reference=train, verbose=False)
    oo1=knn(input_model=copy.copy(nbrs_plus['output_model']), verbose=False, k=5, algorithm="greedy", qu
ery=train)
    (distances_intra_plus,indices_intra_plus)=(oo1["distances"], oo1["neighbors"])
    print iteration
    iteration+=1
```

rcurtin:
This is a set of bugfixes for the Python bindings that improve several issues including fixing #1201.  @nvasil: you can apply the patches from here or use this branch and it will fix your memory issue---I tested with the script from #1201.

A more clear summary of changes:

 * Avoid copies of matrices and other large types by using `std::move()` when calling `CLI::SetParam<>()` from the generated Cython bindings.

 * Clarify and correct ownership of matrix memory: now, in all cases, matrices passed down from Python will retain their ownership of the memory, and internally in mlpack we will use non-strict aliases so that we never delete the user's memory.

 * As a result of the code above, any binding that uses `std::move()` on the input matrix (most of them do) will now result in the Python user's matrix *not* being empty.

 * Also we print default values for optional string/double/int parameters now, which is nice.

There is more to do before I merge this, though:

- [x] Handle the model ownership issue: right now an input model gets wiped out and the user has to use `copy.copy()`.  Instead we should have `CLI` hold pointers to models instead, which allows us to make `output_model` point to the same model as `input_model`.

- [x] Force copies by default but provide a clear warning that extra copies are being done.  We can add an option `make_copies` and if `make_copies=False`, then we use aliases everywhere but the user's input matrices or models may be modified.  I'm not sure if I like defaulting `make_copies` to `True` or `False`; there are advantages and disadvantages to each.  I'm slightly inclined to set `make_copies=False` by default, then make clear in the documentation that if "weird things" are happening, try `make_copies=True`.

In any case, as it stands this fixes #1201.


ShikharJ:
@rcurtin Why is the #1200 referenced in the last line? I'm assuming you meant to refer to #1201?

rcurtin:
Because I remembered the correct number wrong when I originally wrote the description, then I went back and changed it but missed the third reference. :)

rcurtin:
Once I have this passing all the tests (we will see if it does) I think this is ready.  The key change to the Python bindings is that now I can do the following:

```
>>> from mlpack import knn
>>> import numpy as np
>>> x = np.random.rand(10, 10)
>>> out = knn(reference=x, k=3, verbose=True)
```

That will (or may) modify `x`, but we can force a copy:

```
>>> out = knn(reference=x, k=3, verbose=True, copy_all_inputs=True)
```

and now `x` will be copied before the binding is run.  This option is documented and should obviate the need for the clunky `copy.copy()` I was suggesting before.  By default inputs are not copied for the sake of speed (including models).

rcurtin:
@mlpack-jenkins build this

zoq:

src/mlpack/bindings/cli/get_allocated_memory.hpp
```diff
@@ -0,0 +1,59 @@
...
+  typedef std::tuple<T*, std::string> TupleType;
+  return std::get<0>(*boost::any_cast<TupleType>(&d.value));
+}
+
+template<typename T>
+void GetAllocatedMemory(const util::ParamData& d,
+                        const void* /* input */,
+                        void* output)
+{
+  *((void**) output) =
```
zoq: Such a pretty line :smiley:

src/mlpack/bindings/python/print_input_processing.hpp
```diff
@@ -31,6 +31,11 @@ void PrintInputProcessing(
     const typename boost::disable_if<std::is_same<T,
         std::tuple<data::DatasetInfo, arma::mat>>>::type* = 0)
 {
+  // The copy_all_inputs parameter must be handled first, and so is outside the
```
zoq: > and so is outside the

Is this correct?

src/mlpack/bindings/python/print_input_processing.hpp
```diff
@@ -161,12 +174,12 @@ void PrintInputProcessing(
...
-   *     MoveFromPtr[Model](CLI.GetParam[Model]('param_name'),
-   *         (<ModelType?> param_name).modelptr)
+   *     SetParamPtr[Model]('param_name', (<ModelType?> param_name).modelptr,
+   *         CLI.HasParam('copy_all_inputs'))
    *   except TypeError as e:
    *     if type(param_name).__name__ == "ModelType":
-   *       MoveFromPtr[Model](CLI.GetParam[Model]('param_name'),
-   *           (<ModelType> param_name).modelptr)
+   *       SetParamPtr[Model]('param_name', (<ModelType> param_name).modelptr,
+   *           CLI.HasParam('copy_all_inputs'))  TODO
```
zoq: Is this done?

src/mlpack/bindings/tests/delete_allocated_memory.hpp
```diff
@@ -0,0 +1,56 @@
+/**
+ * @file delete_allocated_memory.hpp
```
zoq: Is there any difference between this file and the file in ``src/mlpack/bindings/cli/delete_allocated_memory.hpp``? Maybe I missed something.

src/mlpack/bindings/python/print_input_processing.hpp
rcurtin: Yeah, I wrote that down when I was getting up to get dinner, and when I came back and wrote the code apparently I forgot to remove the TODO... :)

src/mlpack/bindings/tests/delete_allocated_memory.hpp
rcurtin: In this case yes, the CLI bindings have to delete the first element of a `tuple<T*, std::string>`, but the test bindings only hold the `T*` directly so the delete command is a little different.

src/mlpack/bindings/python/print_input_processing.hpp
rcurtin: Yeah, I guess it could also be written `and therefore is outside the`; I went ahead and changed it for clarity.

src/mlpack/bindings/cli/get_allocated_memory.hpp
rcurtin: I wish I knew a better way to do it, but the only way I could seem to was by casting around void pointers in all kinds of strange ways...

src/mlpack/bindings/tests/delete_allocated_memory.hpp
zoq: I see, thanks for the clarification.

rcurtin:
Ok, I will go ahead and merge this then.

PR #1215, opened 2018-01-26
Addressing failing tests (AdaBoost)

vmg27:
Changed the parameters such that there is no failure in tests. Parameters are set such that it is consistent with other tests i.e (perceptronIter = 800 and adaboost iterations = 50)

Issue #922

### DIFF
--- a/src/mlpack/tests/adaboost_test.cpp
+++ b/src/mlpack/tests/adaboost_test.cpp
@@ -44,12 +44,12 @@ BOOST_AUTO_TEST_CASE(HammingLossBoundIris)
 
   // Define your own weak learner, perceptron in this case.
   // Run the perceptron for perceptronIter iterations.
-  int perceptronIter = 400;
+  int perceptronIter = 800;
 
   Perceptron<> p(inputData, labels.row(0), numClasses, perceptronIter);
 
   // Define parameters for AdaBoost.
-  size_t iterations = 100;
+  size_t iterations = 50;
   double tolerance = 1e-10;
   AdaBoost<> a(inputData, labels.row(0), numClasses, p, iterations, tolerance);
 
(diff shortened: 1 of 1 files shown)


zoq:
I'm not sure, this test is one that often fails, so do you mind to share how often the test failed with the previous settings?

Ideally, we like to find a good balance between stability and test runtime.

vmg27:
For me every time I run AdaBoostTest I am getting error with previous setting
`/mlpack/src/mlpack/tests/adaboost_test.cpp(115): fatal error in "WeakLearnerErrorIris": critical check error <= weakLearnerErrorRate failed [0.040000000000000001 > 0.033333333333333333]`

rcurtin:
The tests are built in such a way that the random seed does not change each time you run it, so you will only ever receive one result.  Can I ask you to do the following please?  Add this line at line 74 of `src/mlpack/tests/adaboost_test.cpp`:

```
mlpack::math::RandomSeed(std::time(NULL));
```

Once that is added, then each time the test is run a different random seed will be used.  Next, you can run the test many times, printing only failing tests:

```
i=0; while(true); do echo $i; bin/mlpack_test -t AdaBoostTest/WeakLearnerErrorIris | grep -v '^[ ]*$' | grep -v 'Running ' | grep -v 'No errors '; i=$(($i + 1)); sleep 1; done
```

That will print just increasing numbers (that is the number of tests that have been run), and if a test fails it will print more output.  Let me know how that works on your system and we can see if the problem is that the tolerance needs to be adjusted or if it is another issue.

vmg27:
For some reason I had to change my system. In the new system there is no failure of tests. After running the script there is no error in any iteration.

rcurtin:
If you can't reproduce the issue, I will go ahead and close this then.  Thank you for looking into it and reporting what you found! :)

PR #1216, opened 2018-01-28
Optimized LRSDP Computations

manish7294:
Reduces number of computations by taking `R^T * C ` first in objective function and `R^T * A` in constraints converting objective function to `Tr((R^T * C) * R)` and similarly constraints to `Tr((R^T * A) * R)`.

### DIFF
--- a/src/mlpack/core/optimizers/sdp/lrsdp_function_impl.hpp
+++ b/src/mlpack/core/optimizers/sdp/lrsdp_function_impl.hpp
@@ -54,7 +54,8 @@ template <typename SDPType>
 double LRSDPFunction<SDPType>::Evaluate(const arma::mat& coordinates) const
 {
   // Note: We don't require to update the R*R^T matrix here as the current
-  // is used by AugLagrangian, which do not update the coordinate matrix.
+  // function is only used by AugLagrangian, which do not update the coordinates
+  // matrix.
   return accu(SDP().C() % rrt);
 }
 
@@ -72,7 +73,8 @@ double LRSDPFunction<SDPType>::EvaluateConstraint(
     const arma::mat& coordinates) const
 {
   // Note: We don't require to update the R*R^T matrix here as the current
-  // is used by AugLagrangian, which do not update the coordinate matrix.
+  // function is only used by AugLagrangian, which do not update the coordinates
+  // matrix.
 
   // Using cached R*R^T gives better optimization for sparse matrices.
   if (index < SDP().NumSparseConstraints())
@@ -95,8 +97,9 @@ void LRSDPFunction<SDPType>::GradientConstraint(
 }
 
 //! Utility function for updating R*R^T matrix.
-//! Note: Caching R*R^T provide computation optimization by reducing
-//! redundant R*R^T calculations.
+//! Note: Caching R*R^T provide significant computation optimization
+//! by reducing redundant R*R^T calculations in case of functions are not used
+//! updating coordinates matrix, hence leaving R*R^T unchanged.
 template <typename SDPType>
 void UpdateRRT(LRSDPFunction<SDPType>& function,
                const arma::mat& newrrt)
@@ -175,12 +178,19 @@ EvaluateImpl(LRSDPFunction<SDPType>& function,
   //
   // For computation optimization we will be taking R^T * C first.
   // Objective function = Tr((R^T * C) * R)
+
+  // Calculate R*R^T for updating cache.
   const arma::mat rrt = coordinates * trans(coordinates);
 
   // Update R*R^T matrix.
   // Note that we can only use this optimization in case of L-BFGS optimizer
-  // or any other optimizer which calls Evaluate() before Gradient()
-  // with same coordinates matrix.
+  // or any other similar optimizer which calls Evaluate() before Gradient()
+  // with same coordinates matrix and uses only Evaluate() to update
+  // coordinates matrix.
+
+  // Note: In case optimizer also uses Gradient() for updating coordinates
+  // matrix than the same line of code can be used to update R*R^T through
+  // Gradient().
   UpdateRRT(function, rrt);
 
   // Optimized objective function.
@@ -210,6 +220,8 @@ GradientImpl(const LRSDPFunction<SDPType>& function,
   //   with
   // S' = C - sum_{i = 1}^{m} y'_i A_i
   // y'_i = y_i - sigma * (Trace(A_i * (R R^T)) - b_i)
+
+  // Directly reterive R*R^T from cache.
   const arma::mat rrt = function.RRT();
   arma::mat s(function.SDP().C());
 
--- a/src/mlpack/core/optimizers/sdp/lrsdp_function_impl.hpp
+++ b/src/mlpack/core/optimizers/sdp/lrsdp_function_impl.hpp
@@ -94,10 +94,14 @@ void LRSDPFunction<SDPType>::GradientConstraint(
       << "for arbitrary optimizers!" << std::endl;
 }
 
+//! Utility function for updating R*R^T matrix.
+//! Note: Caching R*R^T provide computation optimization by reducing
+//! redundant R*R^T calculations.
 template <typename SDPType>
-void LRSDPFunction<SDPType>::UpdateRRT(const arma::mat& newrrt) const
+void UpdateRRT(LRSDPFunction<SDPType>& function,
+               const arma::mat& newrrt)
 {
-  *(arma::mat*)(&rrt) = newrrt;
+  function.RRT() = newrrt;
 }
 
 //! Utility function for calculating part of the objective when AugLagrangian is
@@ -149,7 +153,7 @@ UpdateGradient(arma::mat& s,
 
 template <typename SDPType>
 static inline double
-EvaluateImpl(const LRSDPFunction<SDPType>& function,
+EvaluateImpl(LRSDPFunction<SDPType>& function,
              const arma::mat& coordinates,
              const arma::vec& lambda,
              const double sigma)
@@ -177,7 +181,7 @@ EvaluateImpl(const LRSDPFunction<SDPType>& function,
   // Note that we can only use this optimization in case of L-BFGS optimizer
   // or any other optimizer which calls Evaluate() before Gradient()
   // with same coordinates matrix.
-  function.UpdateRRT(rrt);
+  UpdateRRT(function, rrt);
 
   // Optimized objective function.
   double objective = trace((trans(coordinates) * function.SDP().C())
--- a/src/mlpack/core/optimizers/sdp/lrsdp_function_impl.hpp
+++ b/src/mlpack/core/optimizers/sdp/lrsdp_function_impl.hpp
@@ -47,8 +47,8 @@ LRSDPFunction<SDPType>::LRSDPFunction(const size_t numSparseConstraints,
 template <typename SDPType>
 double LRSDPFunction<SDPType>::Evaluate(const arma::mat& coordinates) const
 {
-  const arma::mat rrt = coordinates * trans(coordinates);
-  return accu(SDP().C() % rrt);
+  // For computation optimization we will be taking CR first.
+  return trace((trans(coordinates) * SDP().C()) * coordinates);
 }
 
 template <typename SDPType>
@@ -64,11 +64,13 @@ double LRSDPFunction<SDPType>::EvaluateConstraint(
     const size_t index,
     const arma::mat& coordinates) const
 {
-  const arma::mat rrt = coordinates * trans(coordinates);
+  // For computation optimization we will be taking AR first.
   if (index < SDP().NumSparseConstraints())
-    return accu(SDP().SparseA()[index] % rrt) - SDP().SparseB()[index];
+    return trace((trans(coordinates) * SDP().SparseA()[index]) * coordinates)
+               - SDP().SparseB()[index];
   const size_t index1 = index - SDP().NumSparseConstraints();
-  return accu(SDP().DenseA()[index1] % rrt) - SDP().DenseB()[index1];
+  return trace((trans(coordinates) * SDP().DenseA()[index1]) * coordinates)
+                 - SDP().DenseB()[index1];
 }
 
 template <typename SDPType>
@@ -96,6 +98,9 @@ UpdateObjective(double& objective,
   for (size_t i = 0; i < ais.size(); ++i)
   {
     // Take the trace subtracted by the b_i.
+    // Here taking AR first is not recommended as we are already
+    // using pre-computed RR^T. Taking AR first will result in increase
+    // in number of computations.
     const double constraint = accu(ais[i] % rrt) - bis[i];
     objective -= (lambda[lambdaOffset + i] * constraint);
     objective += (sigma / 2.) * constraint * constraint;
@@ -116,6 +121,9 @@ UpdateGradient(arma::mat& s,
 {
   for (size_t i = 0; i < ais.size(); ++i)
   {
+    // Here taking AR first is not recommended as we are already
+    // using pre-computed RR^T. Taking AR first will result in increase
+    // in number of computations.
     const double constraint = accu(ais[i] % rrt) - bis[i];
     const double y = lambda[lambdaOffset + i] - sigma * constraint;
     s -= y * ais[i];
@@ -137,13 +145,18 @@ EvaluateImpl(const LRSDPFunction<SDPType>& function,
   // Let's start with the objective: Tr(C * (R R^T)).
   // Simple, possibly slow solution-- see below for optimization opportunity
   //
-  // TODO: Note that Tr(C^T * (R R^T)) = Tr( (CR)^T * R ), so
+  // Note that Tr(C^T * (R R^T)) = Tr( (CR)^T * R ), so
   // multiplying C*R first, and then taking the trace dot should be more memory
   // efficient
   //
   // Similarly for the constraints, taking A*R first should be more efficient
+  //
+  // For computation optimization we will be taking CR first.
   const arma::mat rrt = coordinates * trans(coordinates);
-  double objective = accu(function.SDP().C() % rrt);
+
+  // Optimized objective function.
+  double objective = trace((trans(coordinates) * function.SDP().C())
+                         * coordinates);
 
   // Now each constraint.
   UpdateObjective(objective, rrt, function.SDP().SparseA(),
--- a/src/mlpack/core/optimizers/sdp/lrsdp_function_impl.hpp
+++ b/src/mlpack/core/optimizers/sdp/lrsdp_function_impl.hpp
@@ -47,7 +47,7 @@ LRSDPFunction<SDPType>::LRSDPFunction(const size_t numSparseConstraints,
 template <typename SDPType>
 double LRSDPFunction<SDPType>::Evaluate(const arma::mat& coordinates) const
 {
-  // For computation optimization we will be taking CR first.
+  // For computation optimization we will be taking R^T * C first.
   return trace((trans(coordinates) * SDP().C()) * coordinates);
 }
 
@@ -64,7 +64,7 @@ double LRSDPFunction<SDPType>::EvaluateConstraint(
     const size_t index,
     const arma::mat& coordinates) const
 {
-  // For computation optimization we will be taking AR first.
+  // For computation optimization we will be taking R^T * A first.
   if (index < SDP().NumSparseConstraints())
     return trace((trans(coordinates) * SDP().SparseA()[index]) * coordinates)
                - SDP().SparseB()[index];
@@ -98,8 +98,8 @@ UpdateObjective(double& objective,
   for (size_t i = 0; i < ais.size(); ++i)
   {
     // Take the trace subtracted by the b_i.
-    // Here taking AR first is not recommended as we are already
-    // using pre-computed RR^T. Taking AR first will result in increase
+    // Here taking R^T * A first is not recommended as we are already
+    // using pre-computed R * R^T. Taking R^T * A first will result in increase
     // in number of computations.
     const double constraint = accu(ais[i] % rrt) - bis[i];
     objective -= (lambda[lambdaOffset + i] * constraint);
@@ -121,8 +121,8 @@ UpdateGradient(arma::mat& s,
 {
   for (size_t i = 0; i < ais.size(); ++i)
   {
-    // Here taking AR first is not recommended as we are already
-    // using pre-computed RR^T. Taking AR first will result in increase
+    // Here taking R^T * A first is not recommended as we are already
+    // using pre-computed R * R^T. Taking R^T * A first will result in increase
     // in number of computations.
     const double constraint = accu(ais[i] % rrt) - bis[i];
     const double y = lambda[lambdaOffset + i] - sigma * constraint;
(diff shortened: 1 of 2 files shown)


rcurtin:
If I am understanding right, this optimizations hinges around the fact that for (A*R^T*R), the matrix (A*R^T) is significantly smaller than (R^T*R) and therefore it's much quicker to compute (A*R^T)*R than A*(R^T*R).  This would happen when the number of constraints is fewer than the number of parameters, which I think is generally true.  Let me know if I have understood that right.  (I know I have written the algebra in the original form not the column-major Armadillo form where everything is transposed, I hope that is not confusing.)

Do you have any quick runtime tests to show that the evaluations and gradient calculations are now faster?  Simply running one of the test cases on LRSDP and comparing both versions over a handful of runs may be sufficient.

manish7294:
@rcurtin here are the running times for multiple run of Johnson844LovaszThetaSDP test.

> i=1; while(true); do echo -n "Run" $i":  "; bin/mlpack_test -t LRSDPTest/Johnson844LovaszThetaSDP 2>&1 | grep -v 'Running ' | grep -v '^[ ]*$' | grep -v 'No errors'; i=$(($i + 1)); sleep 1; done

Previous:
```
Run 1:  2.76081s
Run 2:  2.81437s
Run 3:  2.85808s
Run 4:  2.79318s
Run 5:  2.83648s
Run 6:  2.82382s
Run 7:  2.82247s
Run 8:  2.78188s
Run 9:  2.80789s
Run 10:  2.91141s
Run 11:  2.8192s
Run 12:  2.85181s
Run 13:  2.78982s
Run 14:  2.83226s
Run 15:  2.84832s
Run 16:  2.7895s
Run 17:  2.77255s
Run 18:  2.86243s
Run 19:  2.8154s
Run 20:  2.90054s
```

Proposed:                                 
```
Run 1:  1.32478s
Run 2:  1.31208s
Run 3:  1.32616s
Run 4:  1.31512s
Run 5:  1.31336s
Run 6:  1.30926s
Run 7:  1.31946s
Run 8:  1.32875s
Run 9:  1.31783s
Run 10:  1.32423s
Run 11:  1.31942s
Run 12:  1.33508s
Run 13:  1.31668s
Run 14:  1.3287s
Run 15:  1.35265s
Run 16:  1.31307s
Run 17:  1.3154s
Run 18:  1.32277s
Run 19:  1.31655s
Run 20:  1.29777s

```

src/mlpack/core/optimizers/sdp/lrsdp_function_impl.hpp
```diff
@@ -96,6 +98,9 @@ UpdateObjective(double& objective,
   for (size_t i = 0; i < ais.size(); ++i)
   {
     // Take the trace subtracted by the b_i.
+    // Here taking AR first is not recommended as we are already
+    // using pre-computed RR^T. Taking AR first will result in increase
+    // in number of computations.
```
rcurtin: I think it would be possible to refactor this in such a way that the `coordinates` matrix is passed directly to `UpdateObjective()`.  Do you think that this would provide additional speedup?

Also, a minor note, but I think that it would be more clear to write `A*R` instead of `AR` in the comments (and the same for the other matrix expression).  It's your call whether or not to make that change; I'm not picky.

manish7294: @rcurtin If I am understanding correctly then here refactoring means to directly pass coordinates matrix inplace of RR^T and then compute the constraint using A*R^T. If this is correct then as per my observation it will increase the number of computations drastically as in current state we are only doing computation in cumulative sum `accu()` and element-wise multiplication `ais[i] % rrt` (_here RR^T is being computed only once in the calling function_) where as using coordinates matrix we will be computing `trans((R^T * A_i) * R)` for each iteration 1 to `ais.size()` which will result in a large increase of computations w.r.t to previous one.
```
UpdateObjective(double& objective,
                const arma::mat& coordinates,
                const std::vector<MatrixType>& ais,
                const arma::vec& bis,
                const arma::vec& lambda,
                const size_t lambdaOffset,
                const double sigma)
{
  for (size_t i = 0; i < ais.size(); ++i)
  {
    // Take the trace subtracted by the b_i.
    const double constraint = trace((trans(coordinates) * ais[i]) * coordinates) - bis[i];
    objective -= (lambda[lambdaOffset + i] * constraint);
    objective += (sigma / 2.) * constraint * constraint;
  }
}
```
I have tried with the above code block and it increased the running time by about 6 times of the current time.
Please correct me if I am going wrong somewhere.

rcurtin: So, it's worth pointing out that in the tests you're running, the only functions actually being called are `EvaluateImpl()` and `GradientImpl()` (because the specializations for the `AugLagrangian` optimizer only call those two functions, not the `Evaluate()` and `Gradient()` functions).  So I would have expected that changing to `(A*R)*R^T` for each `A` would be faster, if `(C*R)*R^T` gives such a significant speedup.

The whole speedup depends on this expression being optimized such that the full matrix is not calculated:

```
trace((trans(coordinates) * ais[i]) * coordinates)
```

I'm pretty certain that is true with dense matrices, but I am not sure it is true with sparse matrices.  However, we do have one test case with all dense A matrices: `LRSDPTest/GaussianMatrixSensingSDP`.  Would you be willing to test the three versions of the code (original, optimized C calculation, optimized ais[i] calculations) on that problem?  I suspect we might find that the problem is that when `ais[i]` is sparse, Armadillo is computing the full matrix `trans(coordinates) * ais[i]) * coordinates` as a temporary, then taking the trace of it---which is probably the cause of the slowdown.

Now, I may also be wrong, but my intuition still suggests that if this is faster for the C matrix calculation, we should be able to get the same speedup for the A matrix calculations.

manish7294: Sure I will shortly provide the results for all three conditions.

manish7294:
_Running Time for different versions over multilple runs:_

**# GaussianMatrixSensingSDP**

Run Command Used:

> i=1; while(true); do echo -n "Run" $i": "; bin/mlpack_test -t LRSDPTest/GaussianMatrixSensingSDP 2>&1 | grep -v 'Running ' | grep -v '^[ ]*$' | grep -v 'No errors'; i=$(($i + 1)); sleep 1; done

Previous:
```
Run 1: 12.989s   (GaussianMatrixSensingSDP)
Run 2: 12.9791s   (GaussianMatrixSensingSDP)
Run 3: 12.9452s   (GaussianMatrixSensingSDP)
Run 4: 12.8985s   (GaussianMatrixSensingSDP)
Run 5: 13.0074s   (GaussianMatrixSensingSDP)
Run 6: 12.9212s   (GaussianMatrixSensingSDP)
Run 7: 13.0073s   (GaussianMatrixSensingSDP)
Run 8: 12.9725s   (GaussianMatrixSensingSDP)
Run 9: 12.899s   (GaussianMatrixSensingSDP)
Run 10: 12.9422s   (GaussianMatrixSensingSDP)

```

Using optimization only on dense matrices:

```
 const arma::mat rrt = coordinates * trans(coordinates);
  if (index < SDP().NumSparseConstraints())
    return accu(SDP().SparseA()[index] % rrt) - SDP().SparseB()[index];
  const size_t index1 = index - SDP().NumSparseConstraints();
  return trace((trans(coordinates) * SDP().DenseA()[index1]) * coordinates)
                 - SDP().DenseB()[index1];
```

```
Run 1: 12.4094s   (GaussianMatrixSensingSDP)
Run 2: 12.8643s   (GaussianMatrixSensingSDP)
Run 3: 12.1417s   (GaussianMatrixSensingSDP)
Run 4: 12.0798s   (GaussianMatrixSensingSDP)
Run 5: 12.6487s   (GaussianMatrixSensingSDP)
Run 6: 12.0841s   (GaussianMatrixSensingSDP)
Run 7: 11.9463s   (GaussianMatrixSensingSDP)
Run 8: 12.2213s   (GaussianMatrixSensingSDP)
Run 9: 12.3982s   (GaussianMatrixSensingSDP)
Run 10: 11.9218s   (GaussianMatrixSensingSDP)
```

Optimized ais[i]  (Here C is also optimized) _Tried only 5 times_ :

```
Run 1: 178.129s   (GaussianMatrixSensingSDP)
Run 2: 179.623s   (GaussianMatrixSensingSDP)
Run 3: 180.395s   (GaussianMatrixSensingSDP)
Run 4: 178.663s   (GaussianMatrixSensingSDP)
Run 5: 183.692s   (GaussianMatrixSensingSDP)
```

Current PR version:

```
Run 1: 11.6893s   (GaussianMatrixSensingSDP)
Run 2: 11.7374s   (GaussianMatrixSensingSDP)
Run 3: 11.7082s   (GaussianMatrixSensingSDP)
Run 4: 11.6893s   (GaussianMatrixSensingSDP)
Run 5: 11.6591s   (GaussianMatrixSensingSDP)
Run 6: 11.6397s   (GaussianMatrixSensingSDP)
Run 7: 11.6419s   (GaussianMatrixSensingSDP)
Run 8: 11.597s   (GaussianMatrixSensingSDP)
Run 9: 11.6155s   (GaussianMatrixSensingSDP)
Run 10: 11.6926s   (GaussianMatrixSensingSDP)
```




**# Johnson844LovaszThetaSDP**




Run Command Used:

> i=1; while(true); do echo -n "Run" $i": "; bin/mlpack_test -t LRSDPTest/Johnson844LovaszThetaSDP LRSDP/GaussianMatrixSensingSDP 2>&1 | grep -v 'Running ' | grep -v '^[ ]*$' | grep -v 'No errors'; i=$(($i + 1)); sleep 1; done

Previous:

```
Run 1: 2.90536s   (Johnson844LovaszThetaSDP)
Run 2: 2.93401s   (Johnson844LovaszThetaSDP)
Run 3: 3.15133s   (Johnson844LovaszThetaSDP)
Run 4: 2.87859s   (Johnson844LovaszThetaSDP)
Run 5: 2.9493s   (Johnson844LovaszThetaSDP)
Run 6: 2.98959s   (Johnson844LovaszThetaSDP)
Run 7: 2.96007s   (Johnson844LovaszThetaSDP)
Run 8: 2.8916s   (Johnson844LovaszThetaSDP)
Run 9: 2.90623s   (Johnson844LovaszThetaSDP)
Run 10: 2.88496s   (Johnson844LovaszThetaSDP)
```

Optimized ais[i]  (Here C is also optimized):

```
Run 1: 2.4925s   (Johnson844LovaszThetaSDP)
Run 2: 2.48819s   (Johnson844LovaszThetaSDP)
Run 3: 2.49341s   (Johnson844LovaszThetaSDP)
Run 4: 2.49682s   (Johnson844LovaszThetaSDP)
Run 5: 2.50309s   (Johnson844LovaszThetaSDP)
Run 6: 2.50949s   (Johnson844LovaszThetaSDP)
Run 7: 2.58773s   (Johnson844LovaszThetaSDP)
Run 8: 2.54184s   (Johnson844LovaszThetaSDP)
Run 9: 2.51936s   (Johnson844LovaszThetaSDP)
Run 10: 2.49226s   (Johnson844LovaszThetaSDP)
```

Using optimization only on dense matrices:

```
 const arma::mat rrt = coordinates * trans(coordinates);
  if (index < SDP().NumSparseConstraints())
    return accu(SDP().SparseA()[index] % rrt) - SDP().SparseB()[index];
  const size_t index1 = index - SDP().NumSparseConstraints();
  return trace((trans(coordinates) * SDP().DenseA()[index1]) * coordinates)
                 - SDP().DenseB()[index1];
```

```
Run 1: 2.03692s   (Johnson844LovaszThetaSDP)
Run 2: 2.05736s   (Johnson844LovaszThetaSDP)
Run 3: 2.08177s   (Johnson844LovaszThetaSDP)
Run 4: 2.03506s   (Johnson844LovaszThetaSDP)
Run 5: 2.09492s   (Johnson844LovaszThetaSDP)
Run 6: 2.08166s   (Johnson844LovaszThetaSDP)
Run 7: 2.05037s   (Johnson844LovaszThetaSDP)
Run 8: 2.04741s   (Johnson844LovaszThetaSDP)
Run 9: 2.0342s   (Johnson844LovaszThetaSDP)
Run 10: 2.0469s   (Johnson844LovaszThetaSDP)

```


Current PR version :

```
Run 1: 1.32836s   (Johnson844LovaszThetaSDP)
Run 2: 1.59912s   (Johnson844LovaszThetaSDP)
Run 3: 1.34011s   (Johnson844LovaszThetaSDP)
Run 4: 1.33653s   (Johnson844LovaszThetaSDP)
Run 5: 1.37482s   (Johnson844LovaszThetaSDP)
Run 6: 1.42944s   (Johnson844LovaszThetaSDP)
Run 7: 1.37366s   (Johnson844LovaszThetaSDP)
Run 8: 1.37739s   (Johnson844LovaszThetaSDP)
Run 9: 1.37261s   (Johnson844LovaszThetaSDP)
Run 10: 1.34749s   (Johnson844LovaszThetaSDP)
```

My Inference: The optimization is true for both sparse and dense matrices.

@rcurtin Here is what I found. Please suggest if anything else can be done or I if am wrong somewhere.
This comment might be a bit lengthy one :)

rcurtin:
Thanks for taking the time to run these simulations.  I think you are right that the optimization is occurring for both sparse and dense matrices (or, at least, that the same thing is happening for both).  However what is confusing me is that the optimizations you are performing should be the same for any A and C matrix.

For your optimization for the C matrix only (this is the optimization that is being used in the code currently in the PR), this provides significant speedup because it takes less time to calculate `C * R` followed by the trace of that times `R^T`.  If that is true, `C` has the same dimensions as any matrix `A_i` and therefore the same speedup should also exist for any `A_i`, instead of the 6x+ slowdown you saw.  In fact, when the calculation of `R*R^T` can be avoided altogether, it should be possible to provide even more speedup.  Hopefully this intuition makes sense, but to me it doesn't seem possible that an "in-between solution" like this where you have optimized C but not A_i calculations could be the fastest.  (It is still better than before, which is great, but I'm pretty sure we can get more here.)

I have to wonder, is it possible that the code you tested for the "full" optimization is causing extra copies somewhere?  I'm happy to help debug if you like, just let me know.

manish7294:
You are  right, optimizations should be the same for any A and C matrix and they are infact, as here I have also optimized A in `EvaluateConstraint` leaving `UpdateObjective` & `UpdateGradient` function.

> Here are the results when only C is optimized (Only `EvaluateImpl` & `Evaluate` functions are being optimized leaving `EvaluateConstraint` function):

```
Run 1: 2.00245s   (Johnson844LovaszThetaSDP)
Run 2: 1.97681s   (Johnson844LovaszThetaSDP)
Run 3: 1.99756s   (Johnson844LovaszThetaSDP)
Run 4: 2.00703s   (Johnson844LovaszThetaSDP)
Run 5: 2.0029s   (Johnson844LovaszThetaSDP)
Run 6: 1.97255s   (Johnson844LovaszThetaSDP)
Run 7: 1.97435s   (Johnson844LovaszThetaSDP)
Run 8: 1.98191s   (Johnson844LovaszThetaSDP)
Run 9: 2.01248s   (Johnson844LovaszThetaSDP)
Run 10: 1.98594s   (Johnson844LovaszThetaSDP)
```

```
Run 1: 13.5579s   (GaussianMatrixSensingSDP)
Run 2: 13.3436s   (GaussianMatrixSensingSDP)
Run 3: 13.2902s   (GaussianMatrixSensingSDP)
Run 4: 13.4847s   (GaussianMatrixSensingSDP)
Run 5: 13.1589s   (GaussianMatrixSensingSDP)
Run 6: 12.9112s   (GaussianMatrixSensingSDP)
Run 7: 12.975s   (GaussianMatrixSensingSDP)
Run 8: 12.9697s   (GaussianMatrixSensingSDP)
Run 9: 12.9949s   (GaussianMatrixSensingSDP)
Run 10: 12.9151s   (GaussianMatrixSensingSDP)

```

> Here is run time of  current PR code (Here `EvaluateImpl`(C) , `Evaluate`(C)  & `EvaluateConstraint` (A) functions are being optimized)

```
Run 1: 1.35244s   (Johnson844LovaszThetaSDP)
Run 2: 1.31905s   (Johnson844LovaszThetaSDP)
Run 3: 1.32695s   (Johnson844LovaszThetaSDP)
Run 4: 1.32961s   (Johnson844LovaszThetaSDP)
Run 5: 1.40477s   (Johnson844LovaszThetaSDP)
Run 6: 1.51283s   (Johnson844LovaszThetaSDP)
Run 7: 1.41424s   (Johnson844LovaszThetaSDP)
Run 8: 1.33381s   (Johnson844LovaszThetaSDP)
Run 9: 1.45415s   (Johnson844LovaszThetaSDP)
Run 10: 1.41317s   (Johnson844LovaszThetaSDP)
```


```
Run 1: 11.6515s   (GaussianMatrixSensingSDP)
Run 2: 11.6575s   (GaussianMatrixSensingSDP)
Run 3: 11.6564s   (GaussianMatrixSensingSDP)
Run 4: 11.636s   (GaussianMatrixSensingSDP)
Run 5: 11.6236s   (GaussianMatrixSensingSDP)
Run 6: 11.6292s   (GaussianMatrixSensingSDP)
Run 7: 11.6316s   (GaussianMatrixSensingSDP)
Run 8: 11.7187s   (GaussianMatrixSensingSDP)
Run 9: 11.6549s   (GaussianMatrixSensingSDP)
Run 10: 11.6014s   (GaussianMatrixSensingSDP)
```


To support that in current state we may not be able to apply optimization on `UpdateObjective` & `UpdateGradient` function, I would like to say that in these two function if we use optimization then we will be calculating `Tr((R^T * A) * R)` for `a_i.size()` times which is way more than just calculating `accu(A % B)` for `a_i.size()` times (Here `B` is just one time calculation of `R * R^T` & the fact that we are only using element-wise multiplication here also supports the inference).

Moreover it would be best if you could find out some time to debug it. Thanks @rcurtin.

manish7294:
@rcurtin any lead on this one :)

rcurtin:
@manish7294: I'm sorry, it has been hard to find the time to look into this.  I did some quick digging today and although the digging is somewhat incomplete I have a few conclusions (or at least directions) that maybe you could look into more if you were interested.

The augmented Lagrangian optimizer is a "two-stage" optimizer: it will, in outer iterations, evaluate the `LRSDPFunction` by calling `LRSDPFunction::Evaluate()` and `LRSDPFunction::EvaluateConstraint()` in order to update the penalty parameter `sigma` and other parameters.  Then, the inner iteration is that the augmented Lagrangian function (`AugLagrangianFunction`) is passed to the regular L-BFGS optimizer.  This L-BFGS optimizer is what ends up calling the overloads `EvaluateImpl()` and `GradientImpl()`, which in turn call `UpdateObjective()` and `UpdateGradient()`.

Now, like you pointed out, if you calculate `Tr((R^T * A) * R)` very many times, it is actually faster to have the matrix `R * R^T` precalculated, and I think that is what is happening here.  So you are right that optimizing `UpdateObjective()` and `UpdateGradient()` are unrealistic.  But maybe we can approach the problem from a different angle: properly calculating and caching `R * R^T`.  If you are interested in doing that, you can probably start by trying to figure out what the right way to handle the following code snippet from `aug_lagrangian_impl.hpp` is:

```
  // Track the last objective to compare for convergence.
  double lastObjective = function.Evaluate(coordinates);

  // Then, calculate the current penalty.
  double penalty = 0;
  for (size_t i = 0; i < function.NumConstraints(); i++)
    penalty += std::pow(function.EvaluateConstraint(i, coordinates), 2);
```

Maybe the `LRSDP` function API needs to be changed, or maybe it can be changed in a clever way to avoid recalculations of `R * R^T`; I am not sure.

Alternately, if you would prefer to leave this speedup as-is, we can merge this now and then open an issue that maybe someone else will be interested in sometime in the future (or that you can come back to some other day).

manish7294:
@rcurtin Thanks for looking into it and giving such a detail insight :) . Please let this PR open for now, I shall take a look into this to figure what can be done. I will notify you of any update from my side.

manish7294:
@rcurtin Here are the ideas I came up with according to your thought that caching `R*R^T` will be better:
1. Changing the function definition for Evaluate & Gradient to include `rrt`. But I myself don't like this one as we are using ` R*R^T` only in case of LRSDP, so it will only be violation of API design. And will result in changing of many different modules like lbfgs and others using or being used by Aug_Lagrangian.
```
Gradient(
    const arma::mat& coordinates,
    const arma::mat& rrt,
    arma::mat& gradient)
```
2. This one again is relative to 1st one, just a minor difference that instead of including coordinated separately in function definition, just make a tuple of the two.
```
std::make_tuple(coordinates, rrt)

Gradient(
    const tuple<arma::mat, arma::mat>& iterate,
    arma::mat& gradient)
```
3. This one I think may be the optimal solution. Here I am declaring `rrt` as private member of `LRSDPFunction class` and which is getting initialized as `rrt = initialPoint * trans(initialPoint)` through `LRSDPFunction constructor`. 
After going through whole process of `LRSDP` optimization I found that we are only updating the `coordinates` matrix through `lbfgs` optimizer. And which is using `EvaluateImpl` function while updating coordinates matrix. So it is pretty much clear that we only need to update value of `rrt` when `EvaluateImpl` is called (Here combining one of first two methods will give even slightly better result as `EvalateImpl` is also being called by `AugLagrangian`, which is not updating coordinates matrix. But doing this is a tradeoff between `API design` and slight `LRSDP improvement`).
Hence I am only updating the value of `rrt` as `rrt = coordinates * trans(coordinates)` when `EvaluateImpl` is called and directly using it elsewhere by calling` RRT()` function or directly `rrt` in member functions.

And the good news is that results got better using 3rd method:

> Johnson844LovaszThetaSDP

```
Run 1: 1.11352s 
Run 2: 1.11996s 
Run 3: 1.11193s 
Run 4: 1.11016s 
Run 5: 1.11213s 
Run 6: 1.1045s 
Run 7: 1.11028s 
Run 8: 1.11108s 
Run 9: 1.12134s 
Run 10: 1.11036s 
```

> GaussianMatrixSensingSDP

```
Run 1: 11.361s 
Run 2: 11.3493s 
Run 3: 11.4318s 
Run 4: 11.3945s 
Run 5: 11.3552s 
Run 6: 11.3976s 
Run 7: 11.4414s 
Run 8: 11.4287s 
Run 9: 11.3692s 
Run 10: 11.4163s 
```
And the last point (Specifically from your observation)
`accu(ais[i] % rrt)`  (with cached `rrt`) works slightly better than` trace((trans(coordinates) * C * coordinates) ` majorly in case of dense matrices.

If this is acceptable than I will make the commit. Please tell me if you have any other idea to improve it further.

rcurtin:
I think that option 3 is best, but we can be a little "safe" about it by including some comments in the calculation noting that this only works with L-BFGS or any optimizer that always calls Evaluate() before Gradient() with the same coordinates.

Something similar has been done for `src/mlpack/methods/nca/nca_softmax_error_function.hpp` with the `Precalculate()` function, but I think the situation there is a little bit different so the solution may not be fully adaptable.

src/mlpack/core/optimizers/sdp/lrsdp_function_impl.hpp
```diff
@@ -81,6 +94,12 @@ void LRSDPFunction<SDPType>::GradientConstraint(
       << "for arbitrary optimizers!" << std::endl;
 }
 
+template <typename SDPType>
+void LRSDPFunction<SDPType>::UpdateRRT(const arma::mat& newrrt) const
+{
+  *(arma::mat*)(&rrt) = newrrt;
+}
+
```
manish7294: This may sound lame but I was unable to update `rrt` through below code:
`rrt = newrrt`
If I am trying to use it then left operand is being retrieved as `const arma::mat` and hence can't change value directly. Please help me out here.

manish7294:
@rcurtin I have added the comment. Please let me know whether it will be sufficient. And thanks for helping through this. 
No doubt, will definitely open a PR if found anything which can be optimized :)

rcurtin:
:+1: looks good to me.  Thanks!

manish7294:
I think there is some problem going on with python bindings as it's test getting failed.

rcurtin:
The Python issue should be fixed since #1249 was merged.  I restarted the Travis build to see for sure.

manish7294:
@rcurtin  I think python bindings issue is still there as again the `python_bindings_test` failed. And earlier travis issue is also there as other 2 instances of travis errored with log "No output has been received in the last 10m0s".

rcurtin:
When Travis rebuilt it it did not use the newest configuration.  So I am not worried about it.  For the timeouts please see #1221 and the discussion there.  This is a Travis issue.

manish7294:
@rcurtin Any update for this :)

rcurtin:
I am a bit disappointed with the downtime for the build infrastructure.  I have a meeting with the team in charge of the move over lunch today and I will be letting them know what I think...

rcurtin:

rcurtin:
Thanks again for the contribution. :)

PR #1217, opened 2018-01-28
Sparse Coding  Bindings Tests

manish7294:
Include bindings tests for sparse_coding_main.cpp

### DIFF
--- a/src/mlpack/tests/main_tests/sparse_coding_test.cpp
+++ b/src/mlpack/tests/main_tests/sparse_coding_test.cpp
@@ -1,0 +1,465 @@
+/**
+ * @file sparse_coding_test.cpp
+ * @author Manish Kumar
+ *
+ * Test mlpackMain() of sparse_coding_main.cpp.
+ *
+ * mlpack is free software; you may redistribute it and/or modify it under the
+ * terms of the 3-clause BSD license.  You should have received a copy of the
+ * 3-clause BSD license along with mlpack.  If not, see
+ * http://www.opensource.org/licenses/BSD-3-Clause for more information.
+ */
+#include <string>
+
+#define BINDING_TYPE BINDING_TYPE_TEST
+static const std::string testName = "SparseCoding";
+
+#include <mlpack/core.hpp>
+#include <mlpack/core/util/mlpack_main.hpp>
+#include <mlpack/methods/sparse_coding/sparse_coding_main.cpp>
+#include "test_helper.hpp"
+
+#include <boost/test/unit_test.hpp>
+#include "../test_tools.hpp"
+
+using namespace mlpack;
+
+struct SparseCodingTestFixture
+{
+ public:
+  SparseCodingTestFixture()
+  {
+    // Cache in the options for this program.
+    CLI::RestoreSettings(testName);
+  }
+
+  ~SparseCodingTestFixture()
+  {
+    // Clear the settings.
+    CLI::ClearSettings();
+  }
+};
+
+BOOST_FIXTURE_TEST_SUITE(SparseCodingMainTest, SparseCodingTestFixture);
+
+/**
+ * Make sure that output points in dictionary equals number of
+ * atoms passed and codes have desired dimension.
+ */
+BOOST_AUTO_TEST_CASE(SparseCodingOutputDimensionTest)
+{
+  mat inputData;
+  inputData.load("mnist_first250_training_4s_and_9s.arm");
+
+  // Shuffle input dataset.
+  inputData = shuffle(inputData);
+
+  // Generate test dataset.
+  mat testData;
+  testData = inputData.cols(450, 499);
+
+  // Generate train dataset.
+  inputData.shed_cols(450, 499);
+
+  // Input data.
+  SetInputParam("training", std::move(inputData));
+  SetInputParam("atoms", (int) 30);
+  SetInputParam("max_iterations", (int) 500);
+  SetInputParam("normalize", (bool) true);
+  SetInputParam("test", std::move(testData));
+
+  mlpackMain();
+
+  // Check that number of output dictionary points are equals number of atoms.
+  BOOST_REQUIRE_EQUAL(CLI::GetParam<arma::mat>("dictionary").n_cols, 30);
+
+  // Check that number of output dictionary rows equal number of input rows
+  // which equal 784 for each data point.
+  BOOST_REQUIRE_EQUAL(CLI::GetParam<arma::mat>("dictionary").n_rows, 784);
+
+  // Check that number of output points are equal to number of test points.
+  BOOST_REQUIRE_EQUAL(CLI::GetParam<arma::mat>("codes").n_cols, 50);
+
+  // Check that number of output codes rows equal number of atoms.
+  BOOST_REQUIRE_EQUAL(CLI::GetParam<arma::mat>("codes").n_rows, 30);
+}
+
+/**
+ * Ensure that training data is normalized if normalize
+ * parameter is set to true.
+ */
+BOOST_AUTO_TEST_CASE(SparseCodingNormalizationTest)
+{
+  mat inputData;
+  inputData.load("mnist_first250_training_4s_and_9s.arm");
+
+  // Shuffle input dataset.
+  inputData = shuffle(inputData);
+
+  // Generate test dataset.
+  mat testData;
+  testData = inputData.cols(450, 499);
+
+  // Generate train dataset.
+  inputData.shed_cols(450, 499);
+
+  // Generate initial dictionary.
+  SetInputParam("training", inputData);
+  SetInputParam("atoms", (int) 30);
+  SetInputParam("max_iterations", (int) 10);
+  SetInputParam("normalize", (bool) true);
+
+  mlpackMain();
+
+  mat initialDictionary;
+  initialDictionary = std::move(CLI::GetParam<arma::mat>
+                                    ("dictionary"));
+
+  // Train for normalization set to true.
+
+  // Input data.
+  SetInputParam("training", inputData);
+  SetInputParam("atoms", (int) 30);
+  SetInputParam("initial_dictionary", initialDictionary);
+  SetInputParam("max_iterations", (int) 100);
+  SetInputParam("normalize", (bool) true);
+  SetInputParam("test", testData);
+
+  mlpackMain();
+
+  // Store outputs.
+  arma::mat dictionary;
+  arma::mat codes;
+  dictionary = std::move(CLI::GetParam<arma::mat>("dictionary"));
+  codes = std::move(CLI::GetParam<arma::mat>("codes"));
+
+  // Train for normalization set to false.
+
+  // Reset passed parameters.
+  CLI::GetSingleton().Parameters()["normalize"].wasPassed = false;
+
+  // Normalize train dataset.
+  for (size_t i = 0; i < inputData.n_cols; ++i)
+    inputData.col(i) /= norm(inputData.col(i), 2);
+
+  // Normalize test dataset.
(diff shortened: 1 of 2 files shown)


manish7294:
@rcurtin @zoq  Can you please re-run the travis as one of the build passed successfully while other two errored before even starting?

zoq:
Sure let me restart the build.

manish7294:
Thanks @zoq :)

src/mlpack/tests/main_tests/sparse_coding_test.cpp
```diff
@@ -0,0 +1,465 @@
...
+  SetInputParam("training", std::move(inputData));
+  SetInputParam("atoms", (int) 30);
+  SetInputParam("max_iterations", (int) 10);
+  SetInputParam("normalize", (bool) true);
+
+  mlpackMain();
+
+  mat initialDictionary;
+  initialDictionary = std::move(CLI::GetParam<arma::mat>
+                                    ("dictionary"));
```
rcurtin: This is a really minor comment, but you could initialize and assign on the same line, and this might make the overflow a little easier:

```
mat initialDictionary = 
    std::move(CLI::GetParam<arma::mat>("dictionary"));
```

manish7294:
@rcurtin I also wanted to add those tests. I had those tests in my mind while writing the tests and even wrote them but when I tried those, as for instance I trained two model's one with newton_tolerance 0.01 and other with 0.5 but I got the same result for output dictionaries. Don't know where I went wrong. Could you suggest me something to overcome this?

src/mlpack/tests/main_tests/sparse_coding_test.cpp
manish7294: Sure, will reformat it everywhere in PR.

src/mlpack/tests/main_tests/sparse_coding_test.cpp
```diff
@@ -462,4 +458,85 @@ BOOST_AUTO_TEST_CASE(SparseCodingModelReuseTest)
...
+  mlpackMain();
+
+  // Check that initial outputs and final outputs
+  // using two models model are different.
+  for (size_t i = 0; i < dictionary.n_elem; ++i)
+  {
+    if(dictionary[i]!=0 && CLI::GetParam<arma::mat>("dictionary")[i]!=0)
+    {
+      BOOST_REQUIRE_NE(dictionary[i],
+          CLI::GetParam<arma::mat>("dictionary")[i]);
```
manish7294: Currently I have used the above `if` condition to avoid failure when dictionary element is 0. Could you please suggest a better way so that it can be confirmed that atleast one element of the matrices is different. Using current code I am unable to do it properly as `BOOST_REQUIRE_NE` requires both element to be different.
One other way I tried to check that at least one element is different is through:
```
BOOST_REQUIRE_THROW(CheckMatrices(dictionary, CLI::GetParam<arma::mat>("dictionary")), std::runtime_error);
```
but it always reports the fatal error occured in `BOOST_REQUIRE_EQUAL`

rcurtin: I think you could do

```
BOOST_REQUIRE_LT(arma::accu(dictionary == CLI::GetParam<arma::mat>("dictionary")), dictionary.n_elem);
```

In that the left hand side will be the number of elements that are exactly equal, and we want this to be less than the total number of elements in the matrix.  I think this will work fine; let me know if there is an issue.

rcurtin:
Hm, you might try using larger values for the tests to force them to get different output dictionaries.  For instance you might try Newton tolerances set to 1e-5 and 10000.  If that doesn't work let me know and I will play with it a little bit.

src/mlpack/tests/main_tests/sparse_coding_test.cpp
manish7294: Wow! This is the finest way of doing that. Thanks @rcurtin.

rcurtin:
I ran the `mlpack_sparse_coding` program by hand with the `iris.csv` dataset as the training set and 2 atoms.  For a fixed initial dictionary which was just the first 2 points in `iris.csv`, I get different results when `objective_tolerance` is the default and when it is 10000, and also when `newton_tolerance` is the default and when it is 10000.  Maybe this can help you adapt the test.  I did have some trouble getting it to converge well for larger numbers of atoms on small datasets, so perhaps reducing that number can be helpful for this test.

manish7294:
@rcurtin Thanks for looking into it.
I tried with your parameters settings and they worked seamlessly. But I observed that these settings doesn't work with `mnist_first250_training_4s_and_9s.arm` dataset and also they doesn't work when normalization is set to true.

rcurtin:
That's fine, it doesn't need to work for every dataset, just one.  The idea of the test is just to make sure that the parameters are being passed to the `SparseCoding` class at all, so if even for one dataset it makes a difference, then the parameters are being passed, which is all I think we need to test here.  If this passes all the build tests then from my side I think it is ready for merge; did you have anything else you wanted to add?

manish7294:
I would like to do just one thing. As currently I have used `mnist_first250_training_4s_and_9s.arm`, I think it has increased the running time of tests as the file is pretty much big and might be a reason for slowdown. I would like to replace it with `iris.csv` dataset.

src/mlpack/tests/main_tests/sparse_coding_test.cpp
```diff
@@ -0,0 +1,746 @@
...
+BOOST_FIXTURE_TEST_SUITE(SparseCodingMainTest, SparseCodingTestFixture);
+
+/**
+ * Make sure that output points in dictionary equals number of
+ * atoms passed and codes have desired dimension.
+ */
+BOOST_AUTO_TEST_CASE(SparseCodingOutputDimensionTest)
+{
+  arma::mat inputData;
+  if (!data::Load("iris_train.csv", inputData))
```
zoq: Just an idea, maybe it makes sense to write an extra function for the dataset loading since it's basically used in every test. What do you think?

manish7294: Thanks @zoq for reviewing and suggestion. I like your idea to make a dataset loader function and tried to implement it.
I tried to design it in a way so that it can load training data, training labels, test data, test labels and training weights depending on data requirement.
Below is the working code tested over sparse coding bindings tests:
```
template<typename T  = arma::mat , typename G = T, typename W = G>
void LoadData(T& trainData = *(T*) NULL,
              const std::string& trainPath = *(std::string*) NULL,
              T& testData = *(T*) NULL,
              const std::string& testPath = *(std::string*) NULL,
              G& trainLabels = *(G*) NULL,
              const std::string& trainLabelsPath = *(std::string*) NULL,
              G& testLabels = *(G*) NULL,
              const std::string& testLabelsPath = *(std::string*) NULL,
              W& weights = *(W*) NULL,
              const std::string& weightsPath = *(std::string*) NULL)
{
  // Load training data.
  if(&trainPath != NULL)
    if(!data::Load(trainPath, trainData))
      Log::Fatal << "Cannot load train dataset" << trainPath << "!"
          << std::endl;

  // Load test data.
  if(&testPath != NULL)
    if(!data::Load(testPath, testData))
      Log::Fatal << "Cannot load test dataset" << testPath << "!"
          << std::endl;

  // Load training data labels.
  if(&trainLabelsPath != NULL)
    if(!data::Load(trainLabelsPath, trainLabels))
      Log::Fatal << "Cannot load train labels" << trainLabelsPath << "!"
          << std::endl;

    // Load test data labels.
  if(&testLabelsPath != NULL)
    if(!data::Load(testLabelsPath, testLabels))
      Log::Fatal << "Cannot load test labels" << testLabelsPath << "!"
          << std::endl;

    // Load test data labels.
  if(&weightsPath != NULL)
    if(!data::Load(weightsPath, weights))
      Log::Fatal << "Cannot load training weights" << weightsPath << "!"
          << std::endl;
}
```
Implementation may seem a bit weird as I couldn't find anything else to assign default values to the arguments passed by reference.
Please provide your views on it.

rcurtin: I think this could be useful, but I'd consider not using default arguments at all, but instead offering a few overloads:

```
template<typename MatType, typename LabelsType, typename WeightsType>
void LoadData(const std::string& trainFile,
              const std::string& trainLabelsFile,
              const std::string& testFile,
              const std::string& testLabelsFile,
              const std::string& trainWeightsFile,
              MatType& trainData,
              LabelsType& trainLabels,
              MatType& testData,
              LabelsType& testLabels,
              WeightsType& trainWeights);

template<typename MatType, typename LabelsType>
void LoadData(const std::string& trainFile,
              const std::string& trainLabelsFile,
              const std::string& testFile,
              const std::string& testLabelsFile,
              MatType& trainData,
              LabelsType& trainLabels,
              MatType& testData,
              LabelsType& testLabels);

template<typename MatType, typename LabelsType>
void LoadData(const std::string& trainFile,
              const std::string& trainLabelsFile,
              MatType& trainData,
              LabelsType& trainLabels);
```

(I ordered the parameters differently---the idea was all input parameters first, then all output parameters.)  That's just one possible idea.  I guess that it would be possible to generalize a little bit to pass in an arbitrary number of strings followed by an arbitrary number of objects to be loaded using variadic templates; that might result in the smallest implementation.

zoq: I was thinking more in the context of this particular test:

```c++
void LoadData(arma::mat& inputData, arma::mat& testData)
{
  // Load train dataset.
  if (!data::Load("iris_train.csv", inputData))
    BOOST_FAIL("Cannot load train dataset iris_train.csv!");

  // Load test dataset.
  if (!data::Load("iris_test.csv", testData))
    BOOST_FAIL("Cannot load test dataset iris_test.csv!");
}
```

rcurtin: Oh, fair, that works too. :)

manish7294: @rcurtin, @zoq : So, shall we generalize the data loading through overloading in test_helper.cpp as @rcurtin said or should we only employ the data loading for this particular test file as suggested by @zoq .

zoq: Should be fine, to implement a simple ``LoadData`` function inside the test, we can implement a generalized function later, if we think that might be helpful. Does this sound good?

manish7294:
I have a question, just out of curiosity :)
Why it is not recommended to use default values for function arguments?

manish7294:
@zoq, @rcurtin: The problem with travis build error is getting too frequent. Again 2 of 3 builds resulted in error. Would request you to please re-run those 2 instances. Thanks!

rcurtin:
Marcus opened #1221, which I think will fix the Travis build issues... I guess it was too optimistic to run on the experimental xenial images. :(

The only reason I recommended against default values for the `LoadData()` function you suggested is that I wanted to avoid the use of pointers in the function definition (typically mlpack functions don't use pointers if possible, preferring references instead for some amount of safety).  Or did you mean some other situation?

manish7294:
Thanks @rcurtin for answering that. I was just curious to know that, if the practice of using default values bad :-}

rcurtin:
Thanks again for the contribution.  I made some changes to reflect #1214 in c107ffc1; please let me know if there are any problems with the changes I made.  One thing to point out about those changes is that we can no longer `std::move()` the dictionary because it is an alias of the `SparseCoding` object's dictionary member, which gets deleted by `CleanMemory()`.  Everything else is pretty straightforward (I think), but let me know if there's anything I can clarify.

PR #1219, opened 2018-01-30
removed some warning messages

deepakks1995:
While building mlpack I encountered some warning messages corresponding to "uninitialised variables and unused variables". Hence I tried to remove them.


rcurtin:
Thanks again for the nice contribution.  If you like, I can add your name to the list of contributors in `src/mlpack/core.hpp` and `COPYRIGHT.txt` (or you can open a PR for that).  Also, we'd be happy to mail you some mlpack stickers.  If you would like some, send me your mailing address to ryan@ratml.org and I'll get them mailed out. :)

PR #1220, opened 2018-02-01
AdaBoost binding tests

nikhilgoel1997:
Contains binding test for adaboost algorithm as asked in issue #1152

### DIFF
--- a/src/mlpack/tests/main_tests/adaboost_test.cpp
+++ b/src/mlpack/tests/main_tests/adaboost_test.cpp
@@ -72,7 +72,7 @@ BOOST_AUTO_TEST_CASE(AdaBoostOutputDimensionTest)
   mlpackMain();
 
   // Check that number of predicted labels is equal to the input test points.
-  BOOST_REQUIRE_EQUAL(CLI::GetParam<arma::Row<size_t>>("output").n_cols, 
+  BOOST_REQUIRE_EQUAL(CLI::GetParam<arma::Row<size_t>>("output").n_cols,
                       testSize);
   BOOST_REQUIRE_EQUAL(CLI::GetParam<arma::Row<size_t>>("output").n_rows, 1);
 }
--- a/src/mlpack/tests/main_tests/adaboost_test.cpp
+++ b/src/mlpack/tests/main_tests/adaboost_test.cpp
@@ -31,6 +31,7 @@ struct AdaBoostTestFixture
   ~AdaBoostTestFixture()
   {
     // Clear the settings.
+    bindings::tests::CleanMemory();
     CLI::ClearSettings();
   }
 };
@@ -104,7 +105,7 @@ BOOST_AUTO_TEST_CASE(AdaBoostModelReuseTest)
 
   SetInputParam("test", std::move(testData));
   SetInputParam("input_model",
-                std::move(CLI::GetParam<AdaBoostModel>("output_model")));
+                CLI::GetParam<AdaBoostModel*>("output_model"));
 
   mlpackMain();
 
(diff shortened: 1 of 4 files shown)


nikhilgoel1997:
One of the 3 builds of travis terminated without starting and gave this in the log - 

No output has been received in the last 10m0s, this potentially indicates a stalled build or something wrong with the build itself.
Check the details on how to adjust your build configuration on: https://docs.travis-ci.com/user/common-build-problems/#Build-times-out-because-no-output-was-received
The build has been terminated

Is there something wrong from my end?

src/mlpack/tests/main_tests/adaboost_test.cpp
```diff
@@ -0,0 +1,267 @@
...
+
+  mlpackMain();
+
+  const string weakLearner = CLI::GetParam<string>("weak_learner");
+  if (weakLearner == "perceptron")
+  {
+  Log::Fatal.ignoreInput = true;
+  BOOST_REQUIRE_THROW(mlpackMain(), std::runtime_error);
+  Log::Fatal.ignoreInput = false;
+  }
```
rcurtin: I don't understand what you are trying to test here, could you clarify please?

nikhilgoel1997: If a model is given as input, and after that a different weak learner is specified (Maybe with the testing data), it should be ignored.
The same is written in adaboost_main.cpp line 132. I'm trying to test this. 
Is it wrong or have I misunderstood it in some way?

rcurtin: But if the weak learner is ignored, the program won't throw an exception (or at least it shouldn't), it should just print a warning message.  I don't think we have a good way to test if warning output is given, so I wouldn't worry about testing that.

rcurtin:
You are right, I overlooked that you had already written a test to ensure that `weak_learner` is valid.  Sorry about the confusion!

rcurtin:

nikhilgoel1997:
Thank you for your help!
I've removed the weakLearnerIgnoredTest and added the three tests that was asked.
Is there anything else I need to add?

src/mlpack/tests/main_tests/adaboost_test.cpp
```diff
@@ -0,0 +1,426 @@
...
+  int flag = 0;
+
+  for (size_t i = 0; i < output.n_elem; ++i)
+  {
+    if (output[i] != outputPerceptron[i])
+    {
+      flag = 1;
+      break;
+    }
+  }
```
rcurtin: You could also do

```
BOOST_REQUIRE_GT(arma::accu(output != outputPerceptron), 1);
```

nikhilgoel1997:
Thanks for the cleaner way to write that test!
I've added my name to COPYRIGHT and core.hpp! It has been a great learning experience and I'll try my best to apply this experience in my future contributions to scilab! Thank you for guiding me!
Yes, I would love the stickers.
I was unable to run these tests/valgrind on my machine. The console gave the error as "No test case matching the filter". Am I missing something?

nikhilgoel1997:
@rcurtin I don't understand why the Static Code Analysis check is failing. Is there something wrong with that build? Even the console output is not giving a good reason.

zoq:

zoq:
> If you like, you can check this test with valgrind:
> 
> valgrind --leak-check=full --track-origins=yes bin/mlpack_test -t AdaBoostMainTest

This check should now be covered by the Memory checks job.

rcurtin:
Sweet!  I will stop checking manually then. :)

src/mlpack/tests/main_tests/adaboost_test.cpp
```diff
@@ -0,0 +1,416 @@
...
+};
+
+BOOST_FIXTURE_TEST_SUITE(AdaBoostMainTest, AdaBoostTestFixture);
+
+/**
+ * Check that number of output labels and number of input
+ * points are equal.
+ */
+
```
zoq: Super pedantic style issue; remove the extra line.

rcurtin:
With two approvals I will go ahead and merge this now.  Thank you again Nikhil for the nice tests. :)

PR #1221, opened 2018-02-02
Use Trusty image for the travis build.

zoq:
I think we switch to Xenial to install a newer armadillo package, however, we never used the apt-get package. So maybe it's a good idea to switch back to Trusty, for now, to avoid some Travis related updated issues (the support of Xenial is still experimental).


rcurtin:
I think you are right that reverting to trusty is the right thing to do.  I see that the Python tests are failing, but I am not sure exactly why.  I'd need to run a Docker image to reproduce; it's probably an LD_LIBRARY_PATH issue or something similar.  I won't have time for that today, but if you can wait until next week I can look into it.

zoq:
Sounds good, let's fix the python issue over the next week. Sorry for all the time you put into the update issue.

rcurtin:
I determined the issue, but I am not sure exactly how I want to solve it.  It is an object initialization order issue (took a little while to track it down).  Basically, `mlpack::bindings::python::programName` is set as a static string in the same translation unit as where `PROGRAM_INFO()` is defined, but that is a different unit than `PARAM_FLAG("verbose")` and `PARAM_FLAG("copy_all_inputs")` is specified, and those are `PyOption` objects that depend on the value of `mlpack::bindings::python::programName`.  So I am not exactly sure how I want to handle it yet, maybe a quick hack, maybe something more complex.  I'll have an idea tomorrow.

rcurtin:
https://gist.github.com/rcurtin/42385b2d723697824ddb371bfd42eb93 is a fix for the issue.  If you apply it, it should work.  Basically `verbose` and `copy_all_inputs` are the only parameters that will ever exist in a different translation unit than `programName`, so this patch simply ensures that the `verbose` and `copy_all_inputs` parameters never actually try to access `programName`.

zoq:
Currently the pip3 install step takes about 15 minutes, which should be solved once the cache kicks in (probably after this is merged), but it looks like the python tests failed again.

rcurtin:
Turns out Ubuntu 14.04's setuptools is too old.  The following lines should fix it (I am not sure if the first is necessary):

```
pip3 install --upgrade --ignore-installed pip
python3 -m pip install --ignore-installed setuptools
```

If the first line isn't necessary it can just be simplified to

```
pip3 install --upgrade --ignore-installed setuptools
```

That's an awful idea on a system that is being maintained by someone but these are just ephemeral Travis VMs, so no problem. :)

zoq:
It looks like, we run into some limitations, since sometimes the build stucked after the worker information is printed, a direct restart also failed, a restart after like an hour works just fine.

zoq:
I limited the number of concurrent jobs, so let's see if that helps.

rcurtin:
I debugged the Python3 failure... the issue is that setuptools installed by the distribution is too old.  So we can either do

```
pip3 install --upgrade --ignore-installed cython setuptools numpy pandas
```

which gets us the newest versions of everything, or just

```
pip3 install cython setuptools numpy pandas
pip3 install --upgrade --ignore-installed setuptools
```

I'm not picky---I think either way is fine.  I don't know if the first way would cause any other issues.

For the stalled builds I can't find any useful issue... maybe something in here is helpful?  https://github.com/travis-ci/travis-ci/issues?utf8=%E2%9C%93&q=is%3Aissue+is%3Aopen+stalled

We could also try opening an issue there; the Travis team is usually pretty responsive.  Let me know what you think.

zoq:
I was going with: https://github.com/travis-ci/travis-ci/issues/9127.

rcurtin:
Ha, oops, I even looked at that issue but did not scroll all the way to the bottom.

rcurtin:
I guess, if you want to fix the Python3 issue in whichever way I mentioned above, now the issue is Travis's, and we can go ahead and merge this and hope they fix travis-ci/travis-ci#9127 or that we get some more information on what can be done.

zoq:
Let's open another PR for the python fix, this PR is currently just used for testing. I can open the PR, just let me know.

rcurtin:
Ah sorry I didn't realize this was just for testing.  I opened #1249 for the fixes.

rcurtin:
It looks like the code right now is using `xenial` and not `trusty`... do you want to try switching to `trusty` and see if we don't have any issues with the 10-minute build timeouts?  I guess I am not sure what your plan is here; I can also open another PR for simple testing, just let me know.

zoq:
The problem I've encountered is that the cache didn't work, and the pip install step takes > 10 minutes, which brings us to the time limit. Not sure what the cache issue is.

zoq:
Sounds like it's not going to be fixed anytime soon:  https://github.com/travis-ci/travis-ci/issues/9263

zoq:
I think there are a couple of things we can do:

- disable the python3 build
- get the cache working
- install the python3 packages via apt-get
- use jenkins

zoq:
Agreed, apt-get is probably the best option, let's see if we can find some recent packages. About using Jenkins, you are right running arbitrary code is an issue, I don't think docker would solve everything here.

rcurtin:
It looks like the imcode ppa is regularly updated (or at least recently updated): https://launchpad.net/~imcode/+archive/ubuntu/s3ql-trusty-backport, so I think maybe we can just go with that.  Realistically we only have to stay on the trusty image until a new one is stable, which hopefully won't be too long.

zoq:
With https://github.com/mlpack/mlpack/pull/1284 in place, let's close the playground.

PR #1224, opened 2018-02-03
Added tests for kmeans_main.cpp

Prabhat-IIT:
#1152 
Following Tests were added for Kmeans

1.Different Algorithms should yield similar Result
2.Number of Clusters should be Non Negative
3.Number of Clusters Should be Integer
4.Maximum Iteration should be Integer
5.Refined Percentage should be b/w 0 & 1
6.Size and Dimensionality of Output should be correct
7.Size and Dimensionality of Input File should be Correct when --inplace is specified
8. Absence of Input is checked
9. Absence of Number Of clusters is checked
10. Number of Clusters should be less than number of points

### DIFF
--- a/src/mlpack/methods/kmeans/kmeans_main.cpp
+++ b/src/mlpack/methods/kmeans/kmeans_main.cpp
@@ -176,7 +176,8 @@ void FindEmptyClusterPolicy(const InitialPartitionPolicy& ipp)
 {
   if (CLI::HasParam("allow_empty_clusters") ||
       CLI::HasParam("kill_empty_clusters"))
-    RequireOnlyOnePassed({ "allow_empty_clusters", "kill_empty_clusters" }, true);
+    RequireOnlyOnePassed({ "allow_empty_clusters", "kill_empty_clusters" },
+                         true);
 
   if (CLI::HasParam("allow_empty_clusters"))
     FindLloydStepType<InitialPartitionPolicy, AllowEmptyClusters>(ipp);
--- a/src/mlpack/methods/kmeans/kmeans_main.cpp
+++ b/src/mlpack/methods/kmeans/kmeans_main.cpp
@@ -245,10 +245,8 @@ void RunKMeans(const InitialPartitionPolicy& ipp)
   // Make sure we have an output file if we're not doing the work in-place.
   RequireAtLeastOnePassed({ "in_place", "output", "centroid" }, false,
       "no results will be saved");
-  if (!CLI::HasParam("input"))
-    Log::Fatal << "Must specify input"<< endl;
-  // Load our dataset.
-  arma::mat dataset = CLI::GetParam<arma::mat>("input");
+
+  arma::mat dataset = CLI::GetParam<arma::mat>("input");  // Load our dataset.
   arma::mat centroids;
 
   const bool initialCentroidGuess = CLI::HasParam("initial_centroids");
--- a/src/mlpack/methods/kmeans/kmeans_main.cpp
+++ b/src/mlpack/methods/kmeans/kmeans_main.cpp
@@ -174,7 +174,9 @@ static void mlpackMain()
 template<typename InitialPartitionPolicy>
 void FindEmptyClusterPolicy(const InitialPartitionPolicy& ipp)
 {
-  RequireOnlyOnePassed({ "allow_empty_clusters", "kill_empty_clusters" }, true);
+  if (CLI::HasParam("allow_empty_clusters") ||
+      CLI::HasParam("kill_empty_clusters"))
+    RequireOnlyOnePassed({ "allow_empty_clusters", "kill_empty_clusters" }, true);
 
   if (CLI::HasParam("allow_empty_clusters"))
     FindLloydStepType<InitialPartitionPolicy, AllowEmptyClusters>(ipp);
@@ -242,7 +244,8 @@ void RunKMeans(const InitialPartitionPolicy& ipp)
   // Make sure we have an output file if we're not doing the work in-place.
   RequireAtLeastOnePassed({ "in_place", "output", "centroid" }, false,
       "no results will be saved");
-
+  if (!CLI::HasParam("input"))
+    Log::Fatal << "Must specify input"<< endl;
   // Load our dataset.
   arma::mat dataset = CLI::GetParam<arma::mat>("input");
   arma::mat centroids;
--- a/src/mlpack/tests/CMakeLists.txt
+++ b/src/mlpack/tests/CMakeLists.txt
@@ -137,7 +137,8 @@ add_executable(mlpack_test
   main_tests/random_forest_test.cpp
   main_tests/softmax_regression_test.cpp
   main_tests/sparse_coding_test.cpp
-  main_tests/kmeans_test.cpp)
+  main_tests/kmeans_test.cpp
+  )
 
 # Link dependencies of test executable.
 target_link_libraries(mlpack_test
--- a/src/mlpack/tests/CMakeLists.txt
+++ b/src/mlpack/tests/CMakeLists.txt
@@ -137,7 +137,7 @@ add_executable(mlpack_test
   main_tests/random_forest_test.cpp
   main_tests/softmax_regression_test.cpp
   main_tests/sparse_coding_test.cpp
-)
+  main_tests/kmeans_test.cpp)
 
 # Link dependencies of test executable.
 target_link_libraries(mlpack_test
--- a/src/mlpack/tests/main_tests/kmeans_test.cpp
+++ b/src/mlpack/tests/main_tests/kmeans_test.cpp
@@ -49,7 +49,7 @@ BOOST_FIXTURE_TEST_SUITE(KmeansMainTest, KmTestFixture);
 BOOST_AUTO_TEST_CASE(NonNegativeClustersTest)
 {
   arma::mat inputData;
-  if (!data::Load("/home/prabhat/mlpack/src/mlpack/tests/data/vc2.csv", inputData))
+  if (!data::Load("vc2.csv", inputData))
     BOOST_FAIL("Unable to load train dataset vc2.csv!");
 
   SetInputParam("input", std::move(inputData));
@@ -88,7 +88,7 @@ BOOST_AUTO_TEST_CASE(RefinedStartPercentageTest)
   int C = 2;
   double P = 2.0;
   arma::mat inputData;
-  if (!data::Load("/home/prabhat/mlpack/src/mlpack/tests/data/vc2.csv", inputData))
+  if (!data::Load("vc2.csv", inputData))
     BOOST_FAIL("Unable to load train dataset vc2.csv!");
 
   SetInputParam("input", std::move(inputData));
@@ -110,7 +110,7 @@ BOOST_AUTO_TEST_CASE(NonNegativePercentageTest)
   int C = 2;
   double P = -1.0;
   arma::mat inputData;
-  if (!data::Load("/home/prabhat/mlpack/src/mlpack/tests/data/vc2.csv", inputData))
+  if (!data::Load("vc2.csv", inputData))
     BOOST_FAIL("Unable to load train dataset vc2.csv!");
 
   SetInputParam("input", std::move(inputData));
@@ -131,7 +131,7 @@ BOOST_AUTO_TEST_CASE(NonNegativePercentageTest)
 {
   int C = 2;
   arma::mat inputData;
-  if (!data::Load("/home/prabhat/mlpack/src/mlpack/tests/data/vc2.csv", inputData))
+  if (!data::Load("vc2.csv", inputData))
     BOOST_FAIL("Unable to load train dataset vc2.csv!");
 
   size_t col = inputData.n_cols;
@@ -147,9 +147,7 @@ BOOST_AUTO_TEST_CASE(NonNegativePercentageTest)
   BOOST_REQUIRE_EQUAL(CLI::GetParam<arma::mat>("output").n_cols, col);
   BOOST_REQUIRE_EQUAL(CLI::GetParam<arma::mat>("centroid").n_rows, row);
   BOOST_REQUIRE_EQUAL(CLI::GetParam<arma::mat>("centroid").n_cols, C);
-
 }
-
 
 /**
  * Checking that size and dimensionality of prediction is correct when --labels_only is specified
@@ -159,7 +157,7 @@ BOOST_AUTO_TEST_CASE(KmClusteringSizeCheckLabelOnly)
   int C = 2;
 
   arma::mat inputData;
-  if (!data::Load("/home/prabhat/mlpack/src/mlpack/tests/data/vc2.csv", inputData))
+  if (!data::Load("vc2.csv", inputData))
       BOOST_FAIL("Unable to load train dataset vc2.csv!");
   size_t col = inputData.n_cols;
   size_t row = inputData.n_rows;
@@ -188,7 +186,7 @@ BOOST_AUTO_TEST_CASE(KmClusteringEmptyClustersCheck)
   int iterations = 100;
 
   arma::mat inputData;
-  if (!data::Load("/home/prabhat/mlpack/src/mlpack/tests/data/vc2.csv", inputData))
+  if (!data::Load("vc2.csv", inputData))
       BOOST_FAIL("Unable to load train dataset vc2.csv!");
 
 
@@ -233,7 +231,6 @@ BOOST_AUTO_TEST_CASE(KmClusteringEmptyClustersCheck)
 
   BOOST_REQUIRE_GT(arma::accu(killEmptyOutput != allowEmptyOutput), 1);
   BOOST_REQUIRE_GT(arma::accu(normalOutput != allowEmptyOutput), 1);
-
 }
 
 /**
@@ -256,7 +253,7 @@ BOOST_AUTO_TEST_CASE(KmNoInputData)
 BOOST_AUTO_TEST_CASE(KmClustersNotDefined)
 {
   arma::mat inputData;
-  if (!data::Load("/home/prabhat/mlpack/src/mlpack/tests/data/vc2.csv", inputData))
+  if (!data::Load("vc2.csv", inputData))
       BOOST_FAIL("Unable to load train dataset vc2.csv!");
 
   SetInputParam("input", std::move(inputData));
@@ -274,7 +271,7 @@ BOOST_AUTO_TEST_CASE(AlgorithmsSimilarTest)
 {
   int C = 5;
   arma::mat inputData;
-  if (!data::Load("/home/prabhat/mlpack/src/mlpack/tests/data/vc2.csv", inputData))
+  if (!data::Load("vc2.csv", inputData))
       BOOST_FAIL("Unable to load train dataset vc2.csv!");
 
   std::string algo = "naive";
--- a/src/mlpack/tests/main_tests/kmeans_test.cpp
+++ b/src/mlpack/tests/main_tests/kmeans_test.cpp
@@ -300,9 +300,11 @@ BOOST_AUTO_TEST_CASE(KmClustersNotDefined)
 BOOST_AUTO_TEST_CASE(AlgorithmsSimilarTest)
 {
   int c = 5;
-  arma::mat inputData;
-  if (!data::Load("vc2.csv", inputData))
-    BOOST_FAIL("Unable to load train dataset vc2.csv!");
+  arma::mat inputData(10, 1000);
+  inputData.randu();
+
+  arma::mat initCentroids(10, 5);
+  initCentroids.randu();
 
   arma::mat initCentroid = arma::randu<arma::mat>(inputData.n_rows, c);
   std::string algo = "naive";
@@ -312,7 +314,6 @@ BOOST_AUTO_TEST_CASE(AlgorithmsSimilarTest)
   SetInputParam("algorithm", std::move(algo));
   SetInputParam("labels_only", true);
   SetInputParam("initial_centroids", initCentroid);
-  SetInputParam("allow_empty_clusters", false);
 
   mlpackMain();
 
@@ -330,7 +331,6 @@ BOOST_AUTO_TEST_CASE(AlgorithmsSimilarTest)
   SetInputParam("algorithm", std::move(algo));
   SetInputParam("labels_only", true);
   SetInputParam("initial_centroids", initCentroid);
-  SetInputParam("allow_empty_clusters", false);
 
   mlpackMain();
 
@@ -348,7 +348,6 @@ BOOST_AUTO_TEST_CASE(AlgorithmsSimilarTest)
   SetInputParam("algorithm", std::move(algo));
   SetInputParam("labels_only", true);
   SetInputParam("initial_centroids", initCentroid);
-  SetInputParam("allow_empty_clusters", false);
 
   mlpackMain();
 
@@ -366,7 +365,6 @@ BOOST_AUTO_TEST_CASE(AlgorithmsSimilarTest)
   SetInputParam("algorithm", std::move(algo));
   SetInputParam("labels_only", true);
   SetInputParam("initial_centroids", initCentroid);
-  SetInputParam("allow_empty_clusters", false);
 
   mlpackMain();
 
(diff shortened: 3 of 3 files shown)


Prabhat-IIT:
@rcurtin  could you please guide me to pass Travis CI. Should I remove the tests that are checking for Integer Inputs like Cluster and iterations?

KARTHEEKCIC:
Yes, you can remove them. They are not needed as It would be explicitly converted to an integer type.

Prabhat-IIT:
@KARTHEEKCIC  I've removed them :) but still Travis if failing :(

nikhilgoel1997:
I'd suggest to remove the std::move when giving data to train in the functions that are failing.

src/mlpack/tests/main_tests/kmeans_test.cpp
```diff
@@ -0,0 +1,262 @@
...
+        constexpr int N = 10;
+        constexpr int D = 4;
+        int C = 2;
+        double P = 2.0;
+        arma::mat InputData = arma::randu<arma::mat>(N, D);
+
+        SetInputParam("input", std::move(InputData));
+        SetInputParam("refined_start", true);
+        SetInputParam("clusters", std::move(C));
+        SetInputParam("percentage", std::move(P));     // Invalid
```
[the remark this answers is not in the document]
Prabhat-IIT: sure I'll do it :)

Prabhat-IIT:
Sure @rcurtin  I'll implement the remaining tests asap! A critical issue involving these parameters is that as we set the value of k to be large and if we specify any of the parameters `allow_empty_clusters` or `kill_empty_clusters` and set the `max_iterations` to be 0 (i.e. run until converge) then it will never converge. Let me know your thoughts about it and do you have anything specific in mind regarding tests related to these parameters. I'm very much willing to contribute :)

rcurtin:
You can set max_iterations to some intermediate value like 100 or something, although realistically I think that it should converge fine with max_iterations = 0 and `allow_empty_clusters` or `kill_empty_clusters`.  Realistically all we want to see is that one version returns empty clusters and the other version returns fewer than k clusters.

Prabhat-IIT:
@rcurtin whenever I set a high enough value of `k` and try to cluster them by keeping maximum iterations around `100` with `allow_empty_clusters` the centroid file doesn't contain any `Nans`.When we save our clusters, empty clusters should be filled with NaNs. This should be the expected behavior because if a cluster has no points, the concept of a centroid makes no sense.And I'm sure that there are empty clusters because the `Info` while running `kmeans` is showing many empty clusters. 
Also when we run the same case with `kill_empty_clusters` it is returning all the clusters i.e clusters=k and a thing to be noticed here is that residual remain close to `Inf` i.e. it's not converging and many clusters centroid have coordinates around `e+308` . How could I then frame a test if they are not performing as they should i.e `kill_empty_clusters` should have given less than `k` clusters and `allow_empty_clusters` should have given some `Nans` for high values of `k`

rcurtin:
You will need to read the code in detail to determine what it is doing, what it should be doing, and how you can successfully write a test for it.  So far I have given high level suggestions but I don't have time to dive into the details and help you figure everything out.  You may have to adapt what I suggest somewhat.

Prabhat-IIT:
Sorry to disturb you so much!! But since k-means was not giving the expected results that's why I was worried! Otherwise a simple test comparing the output of `kill_empty_clusters` and `allow_empty_clusters` would have been sufficient! I'll try to read the codebase more thoroughly. Thanks for tolerating my nuisances!

rcurtin:
Not a problem at all---but for writing these tests, it is important to know what the code is doing (and what it should do, which may not always be the same thing). :)

Note that `kill_empty_clusters` and `allow_empty_clusters` should both give different results than if neither is specified.  Realistically it is not necessary here to test that `kill_empty_clusters` and `allow_empty_clusters` are behaving correctly---that is for `src/mlpack/tests/kmeans_test.cpp` not the main test here---all we need to do here is check that when those flags are passed, they are getting to the `KMeans` class, which we can check just by seeing if the results are different in some way.  If you find that `allow_empty_clusters` or `kill_empty_clusters` is not behaving how you expect, then we should either open another PR to fix broken code or documentation (and make sure it is properly tested in the future).

Prabhat-IIT:
@rcurtin can you take a look at the new tests added :)

rcurtin:
Hi Prabhat,

I do not need a reminder.  I will get to it when I have time.

Prabhat-IIT:
Sorry to disturb you. Since I'm new to the org I didn't know that all the PRs are constantly reviewed. I'll keep that in mind for future :)

src/mlpack/tests/main_tests/kmeans_test.cpp
```diff
@@ -0,0 +1,376 @@
...
+        //Checking that the results are not identical
+
+        for (int j = 0; j <N ; ++j)
+        {
+            BOOST_CHECK_NE(NormalOutput(j), AllowEmptyOutput(j));
+        }
+        for (int j = 0; j <N ; ++j)
+        {
+            BOOST_CHECK_NE(KillEmptyOutput(j), AllowEmptyOutput(j));
+        }
```
rcurtin: These checks will ensure that every single element in the output is different, but that's not necessarily what we require---we only need that a single element in the output differs.  It's possible that the results could be different, but one single element of `NormalOutput` or `KillEmptyOutput` or `AllowEmptyOutput` could be the same. Can you adapt this please?

src/mlpack/tests/main_tests/kmeans_test.cpp
```diff
@@ -0,0 +1,376 @@
...
+        arma::mat InputData = arma::randu<arma::mat>(N, D);
+
+        SetInputParam("input", std::move(InputData));
+        SetInputParam("clusters", std::move(C));
+        SetInputParam("algorithm", std::move(algo));
+
+        mlpackMain();
+
+        arma::mat NaiveOutput;
+        NaiveOutput = std::move(CLI::GetParam<arma::mat>("output"));
```
rcurtin: Can you name variables with a lowercase first letter place?  `naiveOutput`, for instance.  This applies for all the code; see the style guide:

https://github.com/mlpack/mlpack/wiki/DesignGuidelines#StyleGuidelines

The automatic style checker is not working right now, but I hope to have it back up by the end of the day (it depends on a server being installed in a rack and coming back up cleanly; I am not sure of the status right now).  That should be helpful in working out the style issues.  Another big one is that the tabbing and spacing doesn't seem to be right; it should be two-space tabs (with space characters not tab characters).  If you can fix these things I'd really appreciate it. :)

src/mlpack/tests/CMakeLists.txt
```diff
@@ -137,7 +137,7 @@ add_executable(mlpack_test
   main_tests/random_forest_test.cpp
   main_tests/softmax_regression_test.cpp
   main_tests/sparse_coding_test.cpp
-)
+  main_tests/kmeans_test.cpp)
```
rcurtin: I think it's fine to leave the closing `)` on a separate line; it makes it just a little easier to add more files or move around their ordering.  I know that is a picky comment, sorry about that.

src/mlpack/tests/main_tests/kmeans_test.cpp
Prabhat-IIT: Sorry for the style issue! I'll keep the style guide in my mind. It's just I've a habit to write code as such. I hope style checker will be back soon and you'll not have to worry too much then :)

rcurtin: No worries, everyone writes code differently and that is ok. :)  But for the sake of one project we should try and keep it consistent.  Anyway, the infrastructure should hopefully be back online tomorrow but more likely Wednesday.

rcurtin:
Just an update, I'm sorry it is taking so long to get the build infrastructure to get back online.  We are now waiting on a firewall team to approve the request to open ports.  They say it should be done Monday, but I don't know if I trust them...

Prabhat-IIT:
No pb @rcurtin . Could you plz help me in passing travis-ci. Actually same Fatal error is coming in 5 of the tests and I couldn't figure it out even after debugging so many time. It's just that the error is at `SetInputParam("input", inputData)`

rcurtin:
What have you tried for debugging?

Prabhat-IIT:
I have tried gdb! The thing is gdb is terminating with an error code inside `mlpackMain()` when it is checking params like `refined_start` I've also tried printing at several places but all my efforts were futile as of now!

rcurtin:
Hi Prabhat, there is not time for me to teach you how to use gdb to isolate a stack trace given the large number of other requests we have to handle.  There are lots of tutorials online that you might consider looking at.  If you can't solve the failures then I would suggest that we remove any problematic tests and leave them for another contributor.  Let me know how you would like to proceed.

Prabhat-IIT:
Sorry, I'll solve the failures soon enough!!

rcurtin:

Prabhat-IIT:
@rcurtin Style check is failing due to other tests that were merged while masterblaster was down! What should I do? Moreover, all the tests are fine now. I've removed the test for `--inplace` flag as I was not getting the changed input through `GetParam()` I don't know hy but it was returning the original input repeatedly rather than the modified ones!

rcurtin:
We can't remove a test because it is failing.  The test should work so we need to debug either the test or the program.  I believe the problem is in the test, since I have hand-tested the in-place support and the last time I did that it worked fine.

Don't worry about the style issues not related to this code---Marcus submitted a PR to fix them that will be merged soon with #1283.

Prabhat-IIT:
Yup! It is indeed working fine!! But the thing is that the `GetParam()` is returning a copy of original input rather than a reference of changed input! I'll debug it soon. One thing I would like to ask is that all the algorithms are supposed to return the same result in kmeans but they are not exactly same at some points keeping in mind the floating point precision. That's why sometimes the test checking them to be same fails and sometimes pass. I've been thinking of increasing the difference between any two corresponding value. Presently, it is 1e-05% which is not big enough to pass the test. According to you what should I set this value so that it is acceptable?

rcurtin:
I'm okay if you want to loosen the tolerance a little bit for the centroid matrices, but we should check the cluster assignments and ensure those are definitely the same, since those aren't susceptible to floating point issues (or, at least, they shouldn't be).

Prabhat-IIT:
@rcurtin everythings fine now but cluster assignments results are not same for all the algorithms. I've tried these tests several times locally. They are passing locally in my system but they don't pass on Travis...some element always creates problem. In cluster assignments we will not be even able to increase the tolerance because assignments are themselves large numbers.

rcurtin:
It looks like you are not using the same initial centroids for each run.  Try setting those and I think this should fix your problem.  Note that because of #1264, setting `seed` doesn't actually do anything in the tests anymore.

Prabhat-IIT:
Thanks that really helped. I knew about the seed but since the patch was not in this fork so, I was trying to set a fixed seed for same initialization. Anyways, now it looks good :)

src/mlpack/tests/main_tests/kmeans_test.cpp
```diff
@@ -0,0 +1,408 @@
...
+#include <boost/test/unit_test.hpp>
+#include "../test_tools.hpp"
+
+using namespace mlpack;
+
+struct KmTestFixture
+{
+ public:
+    KmTestFixture()
+    {
```
rcurtin: This block is indented twice but it shpuld be indented only once.

src/mlpack/tests/main_tests/kmeans_test.cpp
```diff
@@ -0,0 +1,408 @@
...
+        // Clear the settings.
+        CLI::ClearSettings();
+    }
+};
+
+void ResetKmSettings()
+{
+    CLI::ClearSettings();
+    CLI::RestoreSettings(testName);
```
rcurtin: This should only be indented once.

src/mlpack/tests/main_tests/kmeans_test.cpp
```diff
@@ -0,0 +1,408 @@
...
+  Log::Fatal.ignoreInput = true;
+  BOOST_REQUIRE_THROW(mlpackMain(), std::runtime_error);
+  Log::Fatal.ignoreInput = false;
+}
+
+
+/**
+ * Checking that size and dimensionality of prediction is correct.
+ */
+    BOOST_AUTO_TEST_CASE(KmClusteringSizeCheck)
```
rcurtin: The spacing is off for this line.

src/mlpack/tests/main_tests/kmeans_test.cpp
```diff
@@ -0,0 +1,408 @@
...
+/**
+ * Checking that size and dimensionality of prediction is correct when --labels_only is specified
+ */
+BOOST_AUTO_TEST_CASE(KmClusteringSizeCheckLabelOnly)
+{
+  int C = 2;
+
+  arma::mat inputData;
+  if (!data::Load("vc2.csv", inputData))
+      BOOST_FAIL("Unable to load train dataset vc2.csv!");
```
rcurtin: This should only be indented once.

src/mlpack/tests/main_tests/kmeans_test.cpp
```diff
@@ -0,0 +1,408 @@
...
+  size_t col = inputData.n_cols;
+  size_t row = inputData.n_rows;
+
+  SetInputParam("input", std::move(inputData));
+  SetInputParam("clusters", std::move(C));
+  SetInputParam("labels_only", true);
+  SetInputParam("allow_empty_clusters", false);
+
+
```
rcurtin: No need for an empty extra line here.

src/mlpack/tests/main_tests/kmeans_test.cpp
```diff
@@ -0,0 +1,408 @@
...
+  SetInputParam("input", inputData);
+  SetInputParam("clusters", C);
+  SetInputParam("labels_only", true);
+  SetInputParam("allow_empty_clusters", true);
+  SetInputParam("max_iterations", iterations);
+
+  mlpackMain();
+
+  arma::mat allowEmptyOutput;
+  allowEmptyOutput = std::move(CLI::GetParam<arma::mat>("output"));
```
rcurtin: You can put both the declaration and the assignment in the same statement if you like.

src/mlpack/tests/main_tests/kmeans_test.cpp
```diff
@@ -0,0 +1,408 @@
...
+  if (!data::Load("vc2.csv", inputData))
+      BOOST_FAIL("Unable to load train dataset vc2.csv!");
+
+
+  SetInputParam("input", inputData);
+  SetInputParam("clusters", C);
+  SetInputParam("labels_only", true);
+  SetInputParam("max_iterations", iterations);
+  SetInputParam("allow_empty_clusters", false);
+
```
rcurtin: For these two runs with `allow_empty_clusters` and `kill_empty_clusters` you should set the initial centroids to be the same.

src/mlpack/tests/main_tests/kmeans_test.cpp
```diff
@@ -0,0 +1,408 @@
...
+ * Checking that that size and dimensionality of Final Input File is correct
+ * when flag --in_place is specified
+ */
+BOOST_AUTO_TEST_CASE(KmClusteringResultSizeCheck)
+{
+  int C = 2;
+
+  arma::mat inputData;
+  if (!data::Load("vc2.csv", inputData))
+      BOOST_FAIL("Unable to load train dataset vc2.csv!");
```
rcurtin: This should only be indented once (this issue occurs in several other places too).

src/mlpack/tests/main_tests/kmeans_test.cpp
```diff
@@ -0,0 +1,408 @@
...
+
+  size_t row = inputData.n_rows;
+  size_t col = inputData.n_cols;
+
+  SetInputParam("input", inputData);
+  SetInputParam("clusters", std::move(C));
```
rcurtin: No need to `std::move()` an int.

src/mlpack/tests/main_tests/kmeans_test.cpp
```diff
@@ -0,0 +1,408 @@
...
+  Log::Fatal.ignoreInput = false;
+}
+
+/**
+ * Checking that that size and dimensionality of Final Input File is correct
+ * when flag --in_place is specified
+ */
+BOOST_AUTO_TEST_CASE(KmClusteringResultSizeCheck)
+{
+  int C = 2;
```
rcurtin: Can you use a lowercase `c` to conform with the naming conventions please? This comment applies to many of the tests in this file.

src/mlpack/tests/main_tests/kmeans_test.cpp
Prabhat-IIT: I think since I've declared other matrices like this I would like to kept it like this otherwise if you want I'll change them all

Prabhat-IIT:
I've handled the style issues but one thing that is not going right is that all the empty cluster policies are returning same assignments therefore I resort to checking that the `centroid` should be different but even then `AllowEmptyCluster` centroids and normal centroids are sometimes same even when I have set number of clusters to be sufficiently high. Let me know what can be done in this regard?

src/mlpack/tests/main_tests/kmeans_test.cpp
```diff
@@ -0,0 +1,414 @@
...
+  arma::mat inputData;
+  if (!data::Load("vc2.csv", inputData))
+    BOOST_FAIL("Unable to load train dataset vc2.csv!");
+
+  size_t col = inputData.n_cols;
+  size_t row = inputData.n_rows;
+
+  SetInputParam("input", std::move(inputData));
+  SetInputParam("clusters", c);
+  SetInputParam("allow_empty_clusters", false);
```
rcurtin: It looks like this is a workaround: the program currently fails, saying that either `allow_empty_clusters` or `kill_empty_clusters` must be specified.  We should fix this in `kmeans_main.cpp` by changing the line

```
RequireOnlyOnePassed({ "allow_empty_clusters", "kill_empty_clusters" }, true);
```

to

```
if (CLI::HasParam("allow_empty_clusters") ||
    CLI::HasParam("kill_empty_clusters"))
  RequireOnlyOnePassed({ "allow_empty_clusters", "kill_empty_clusters" }, true);
```

Can you make that change in this PR?  Then we can remove all of the `SetInputParam("allow_empty_clusters", false);` calls.

src/mlpack/tests/main_tests/kmeans_test.cpp
```diff
@@ -0,0 +1,414 @@
...
+/**
+ * Checking that predictions are not same when --allow_empty_clusters or kill_empty_clusters are specified
+ */
+BOOST_AUTO_TEST_CASE(KmClusteringEmptyClustersCheck)
+{
+  int c = 400;
+  int iterations = 100;
+
+  arma::mat inputData;
+  if (!data::Load("vc2.csv", inputData))
```
rcurtin: Try this one with `test_data_3_1000.csv`.  I think maybe the VC2 dataset has too much structure to produce significantly different results.

src/mlpack/tests/main_tests/kmeans_test.cpp
Prabhat-IIT: I've made the changes but I've not removed the calls in last test because this workaround is assuring that all the algorithms yield same result. Otherwise some elements prove to be problematic and test won't pass. I think maybe this difference arises due to different empty cluster policies

src/mlpack/tests/main_tests/kmeans_test.cpp
Prabhat-IIT: Yup you are right VC2 dataset was not suitable for the test.

src/mlpack/tests/main_tests/kmeans_test.cpp
rcurtin: Let's not commit workarounds to the main codebase.  Please remove those, especially because `SetInputParam("allow_empty_clusters", false)` does not do what you think it does, as boolean-typed parameters should only ever be set to true or left unset.

src/mlpack/methods/kmeans/kmeans_main.cpp
```diff
@@ -242,7 +245,8 @@ void RunKMeans(const InitialPartitionPolicy& ipp)
   // Make sure we have an output file if we're not doing the work in-place.
   RequireAtLeastOnePassed({ "in_place", "output", "centroid" }, false,
       "no results will be saved");
-
+  if (!CLI::HasParam("input"))
+    Log::Fatal << "Must specify input"<< endl;
```
rcurtin: `input` is a required parameter, so this shouldn't be needed.

src/mlpack/tests/main_tests/kmeans_test.cpp
```diff
@@ -72,7 +72,7 @@ BOOST_AUTO_TEST_CASE(PointsLessThanClustersTest)
   arma::mat inputData = arma::randu<arma::mat>(N, D);
 
   SetInputParam("input", std::move(inputData));
-  SetInputParam("clusters", (int) 11); // Invalid
+  SetInputParam("clusters", (int) 0); // Invalid
```
rcurtin: Doesn't it break the idea of the test if you set the number of clusters to a value less than N?

src/mlpack/tests/main_tests/kmeans_test.cpp
```diff
@@ -398,17 +390,23 @@ BOOST_AUTO_TEST_CASE(AlgorithmsSimilarTest)
...
+    BOOST_REQUIRE_EQUAL(naiveOutput(0, i), dualCoverTreeOutput(0, i));
+
+  for (size_t i = 0; i < initCentroid.n_elem; ++i)
+    BOOST_REQUIRE_CLOSE(naiveCentroid(0, i), elkanCentroid(0, i), 1e-5);
+  for (size_t i = 0; i < initCentroid.n_elem; ++i)
+    BOOST_REQUIRE_CLOSE(naiveCentroid(0, i), hamerlyCentroid(0, i), 1e-5);
+  for (size_t i = 0; i < initCentroid.n_elem; ++i)
+    BOOST_REQUIRE_CLOSE(naiveCentroid(0, i), dualTreeCentroid(0, i), 1e-5);
+  for (size_t i = 0; i < initCentroid.n_elem; ++i)
+    BOOST_REQUIRE_CLOSE(naiveCentroid(0, i), dualCoverTreeCentroid(0, i), 1e-5);
```
rcurtin: I'm confused by this change; what was wrong with `CheckMatrices()`?

src/mlpack/methods/kmeans/kmeans_main.cpp
Prabhat-IIT: Actually when I'm not setting the parameter input then it doesn't throw a Fatal error. Although in CLI it does. I tried to look out for the cause but couldn't find it!

src/mlpack/tests/main_tests/kmeans_test.cpp
Prabhat-IIT: Actually, the idea here was changed because clusters demanded can be more than points according to existing implementation of `kmeans` So, here I'm testing that when 0 clusters is passed i.e. when clusters are to be auto detected then initial centroid must be passsed for autodetecting. I've changed the name of test now.

src/mlpack/tests/main_tests/kmeans_test.cpp
Prabhat-IIT: Don't get confused. Its just that while handling the issue I adopted this style for checking to have more control over the tolerance but now I 've reverted it back to `CheckMatrices()`

src/mlpack/methods/kmeans/kmeans_main.cpp
rcurtin: This is because `ParseCommandLine()` is what checks for required parameters, and that isn't called for the test bindings since we are calling `mlpackMain()`, so there is no need to test here that a required parameter not being passed causes a fatal error.

src/mlpack/tests/main_tests/kmeans_test.cpp
rcurtin: Sure, I guess when I review the actual code, it is possible to ask for more clusters but in the resulting clusters it is guaranteed that some will be empty.

Prabhat-IIT:
Thanks for your constant support @rcurtin From next time I'll try not to bother you much for petty issue. Thanx :smiley:

rcurtin:
Thanks again for the contribution; I'm happy to see it merged.

PR #1225, opened 2018-02-05
Introduce Particle Swarm Optimization for unconstrained problems

adl1995:
This is a work in progress pull request which adds Particle Swarm Optimization into mlpack's codebase. For testing out the implementation, I have added [Rosenbrock function](https://en.wikipedia.org/wiki/Rosenbrock_function) as the objective function. However, the test case currently fails as the PSO algorithm isn't fully implemented.

If someone has a link / document to some good resource which explains the PSO algorithm in detail, please share.

Also, I plan on referring this PR in my GSoC application to strengthen it, so I would really appreciate if the mlpack community could provide some feedback, this will help me get this merged quickly.


src/mlpack/core/optimizers/pso/test_function.hpp
```diff
@@ -0,0 +1,66 @@
+/**
+ * @file test_function.hpp
```
zoq: You can use the existing Rosenbrock function: ``https://github.com/mlpack/mlpack/tree/master/src/mlpack/core/optimizers/problems``

src/mlpack/core/optimizers/pso/pso.hpp
```diff
@@ -0,0 +1,188 @@
...
+ * (see mlpack::nca::NCA), NumFunctions() should return the number of points in
+ * the dataset, and Evaluate(coordinates, 0) will evaluate the objective function
+ * on the first point in the dataset (presumably, the dataset is held internally
+ * in the DecomposableFunctionType).
+ *
+ * @tparam SelectionPolicy The velocty / position update strategy used for the
+ * evaluation step.
+ */
+template<typename VelocityVectorType = InertiaWeight>
+class PSO
```
zoq: This is just an idea but we could name the class ``PSOType`` and define an alias called ``PSO``, that way we could use ``PSO`` instead of ``PSO<>`` later on, which makes more sense if we introduce another velocity type.

zoq:
The following links could be helpful:
- http://www.swarmintelligence.org/tutorials.php

- https://nathanrooy.github.io/posts/2016-08-17/simple-particle-swarm-optimization-with-python/

rcurtin:
Once you have it working for the Rosenbrock function, I would suggest also adding a few tests for other functions, and a test for a machine learning algorithm like logistic regression or similar.  This is done for a lot of other optimizers; for instance, see `spalera_sgd_test.cpp`.

src/mlpack/core/optimizers/pso/pso_impl.hpp
```diff
@@ -0,0 +1,106 @@
...
+    }
+
+    // Update velocity for each particle.
+    velocityType.Update(particlePosition,
+      particleVelocity, bestParticlePosition,
+      bestSwarmPosition, interiaWeight,
+      cognitiveAcceleration, socialAcceleration, dimension);
+
+    // Update position for each particle.
+    particlePosition = particleVelocity;
```
adl1995: This is incorrect. In the paper the position is updated like this:
```
particlePosition = particlePosition + particleVelocity;
```
However, when I tried this here `particleVelocity` gets extremely small (around `1e-200`) in a few iterations. So, adding it to `particlePosition` does not have any effect. This leads the objective function to return the same value in each iteration, hence it does not get minimized.

I have spent some time trying to fix this, but haven't been able to do so yet. However, I suspect this could be because of the following two reasons:

- In the other implementations I found online ([swarmops - Python](https://github.com/Hvass-Labs/swarmops/blob/master/swarmops/PSO.py#L346), [C](https://github.com/kkentzo/pso/blob/master/pso.c#L380)), there is a boundary enforcement after each update. I haven't done this as this would be a feature for constrained optimization (?).
- Currently, the particle position and velocity is being initialized with uniform random values. This is also how its done in the other implementations I could find. But, this may be causing the particle velocity to 'die out'. There is a [question regarding this on ResearchGate](https://www.researchgate.net/post/How_to_initialize_the_position_and_velocity_of_the_particle_swarm_optimization_technique_for_PV_model) which links to some papers, but I haven't gone through any of those.

I would really appreciate if someone could help me figure out how to resolve this issue.

adl1995:
@zoq - Thanks for the resources. I just updated my branch with the updated implementation. Please see my [inline comment](https://github.com/mlpack/mlpack/pull/1225#discussion_r167442017) regarding the particle position update.

I have added a simple test case which returns a squared sum of each element in the given vector. This function gets minimized in 10 iterations within the accepted tolerance (`1e-5`).

The other test case is the Rosenbrock function, but I now use the existing implementation present in `mlpack/src/mlpack/core/optimizers/problems`. The objective value starts from around `13.63` and is minimized to `0.80` in 200 iterations, which is the maximum limit, therefore it fails. So, the implementation still needs more work. I think the major issue is with how the particle position is being updated. Once that is resolved, this should get minimized within the accepted tolerance.

@rcurtin Thanks - I'll add those as soon as I'm finished with the Rosenbrock function.

src/mlpack/core/optimizers/pso/pso_impl.hpp
zoq: I don't think we could get around using a bound here, even if we find a good initialization. However, I like the idea to test different initalization methods, one idea we could test is, to use CNE (https://github.com/mlpack/mlpack/tree/master/src/mlpack/core/optimizers/cne) or CMAES (https://github.com/mlpack/mlpack/tree/master/src/mlpack/core/optimizers/cmaes) to initialize the velocities. Let me know what you think.

zoq:
Sounds like a good plan to me, I think let's test the bound option first and add new ideas afterwards, does this sound good?

adl1995:
@zoq - I just added [boundary checks](https://github.com/mlpack/mlpack/pull/1225/commits/1b0e614df8f9e2a7a005490b4d9dd4e30e51b4f0) for particle position and velocity, but this hasn't made any improvements.

Could you please explain on how to use CNE or CMA-ES for velocity initialization? Did you mean to evolve the velocities for a few generations and use that as an input to PSO?

zoq:
> @zoq - I just added boundary checks for particle position and velocity, but this hasn't made any improvements.

Okay, I'll take a closer look at the code tomorrow.

> Could you please explain on how to use CNE or CMA-ES for velocity initialization? Did you mean to evolve the velocities for a few generations and use that as an input to PSO?

Exactly, but I don't think it will make a huge difference.

zoq:
Can you push the latest version, it looks like travis can't build the code.

src/mlpack/core/optimizers/pso/pso_impl.hpp
```diff
@@ -0,0 +1,137 @@
...
+//! Optimize the function (minimize).
+template<typename VelocityVectorType>
+template<typename DecomposableFunctionType>
+double PSOType<VelocityVectorType>::Optimize(
+    DecomposableFunctionType& function, arma::mat& iterate)
+{
+  // Randomly initialize the particle position and velocity.
+  // Following a heuristic, the swarm size is set to 2 * dimension.
+  lambda = 2 * dimension;
+  particlePosition.randu(dimension);
```
zoq: We should use the iterate dimension here, which should contain the initial starting point. I think the idea here is that ``dimension`` defines the population size, correct me if I'm wrong. So I think it should look something like:

```c++
particlePosition.randu(iterate.n_rows, iterate.n_cols);
```

Which defines a single particle, I guess an easy option would be to use ``arma::cube``  to store populations. But depending on the population and function size this could be expensive. So another option would be to use ``arma::cube`` that stores only the current best solution and the solution we are testing:

```c++
arma::cube particlePosition = randu<arma::cube>(iterate.n_rows, iterate.n_cols, 2);
```

and in the evaluation step, which currently only uses a single particle we do something like:

```c++
for (size_t p = 0; p < populationSize; ++p)
{
  currentObjective = function.Evaluate(particlePosition.slice(0));
  if (currentObjective < lastObjectiveIndividual)
  {
   // slice(0) contains the current working particle and slice(1) the current best particle.
    particlePosition.slice(1) = particlePosition.slice(0);    
  }
  ... 
}
```

We should be able to do something clever to avoid the extra copy, but we can do that later. I hope something I said is helpful, let me know if I should clarify anything.

adl1995:
@zoq Thank you for the code review! I will make the required changes tomorrow. Quick question: If we store only the current and best particle in `arma::cube`, wouldn't we be updating the same particle in each iteration?

I pushed my latest version which fixes the previous build error. However, the build now fails at `pso_test.c`, as expected. Failing test case: https://travis-ci.org/mlpack/mlpack/jobs/342438786#L3555

adl1995:
@zoq Your suggestion on using the iterate dimension had a significant impact. The build is now passing: https://travis-ci.org/mlpack/mlpack/builds/342943314

I have initialized both the particle position and velocity with the iterate values. I'm not sure if this is the correct way, though. For now I have stored the entire population in `arma::cube`. I'm thinking of adding a few more tests to ensure the algorithm works for a variety of problems. After that, I will start working on optimizing the function.

The boundary constraints didn't have much impact, so I have removed them for now. Will add them later, if necessary.

zoq:
> I have initialized both the particle position and velocity with the iterate values. I'm not sure if this is the correct way, though. 

Yes, this way a user could easily modify the starting point.

> After that, I will start working on optimizing the function.

adl1995:
I have added a few more tests from the `mlpack/core/optimizers/problems` directory. They are passing except for the Booth function and Logistic regression: https://travis-ci.org/mlpack/mlpack/jobs/344165431#L4096

With the Logistic regression test, it always gives a flat 50% accuracy. So, I think there's some issue with the current implementation. Also, for the other tests, I have applied a check on the value returned by the optimizer i.e. the objective function value. However, in [adam_test.cpp](https://github.com/mlpack/mlpack/blob/119c410c56a1556749e2d98fc63edf21f40fd942/src/mlpack/tests/adam_test.cpp#L163) the check is applied on the coordinate values.

Currently, in PSO the iterate (or coordinate) values remain unaffected, and are only used to initialize the particle position and velocity. In SGD, the iterate is [updated in each iteration](https://github.com/mlpack/mlpack/blob/c1d43e602a17d416169f2891554db166b9c61532/src/mlpack/core/optimizers/sgd/sgd_impl.hpp#L118) using the update policy. Should I do something similar in PSO?

Also, I haven't used the `NumFunctions()` method in PSO class. In [cmaes_impl.hpp](https://github.com/mlpack/mlpack/blob/9f149767a08140dba7bf4f12f63f1462b3610c5f/src/mlpack/core/optimizers/cmaes/cmaes_impl.hpp#L95) the returned value from the function is accumulated in `currentObjective`. Is this done to optimize the cumulative sum of all the points in the dataset? 

I apologize if my questions are too broad. Please let me know if anything needs clarifying.

zoq:
Thanks for the update.

> With the Logistic regression test, it always gives a flat 50% accuracy. So, I think there's some issue with the current implementation. Also, for the other tests, I have applied a check on the value returned by the optimizer i.e. the objective function value. However, in adam_test.cpp the check is applied on the coordinate values.

If you run the Logistic regression test, make sure to use a new random see for each run, e.g. by adding ``math::RandomSeed(std::time(NULL));``. I think a simple idea would be to increase the number of iterations, and see if that has an effect.

> Currently, in PSO the iterate (or coordinate) values remain unaffected, and are only used to initialize the particle position and velocity. In SGD, the iterate is updated in each iteration using the update policy. Should I do something similar in PSO?

using: 

```
currentObjective = function.Evaluate(particlePosition.slice(k));
```

is just fine, but in the best case we use ``iterate`` to save memory.

> Also, I haven't used the NumFunctions() method in PSO class. In cmaes_impl.hpp the returned value from the function is accumulated in currentObjective. Is this done to optimize the cumulative sum of all the points in the dataset?

Right, but in this case it's only relevant if the function is seperable.

adl1995:
@zoq I increased the iteration count from 200 to 10,000 and also tried to seed the random number generator before each run. But, this still gives an accuracy of 50 - 51%: https://travis-ci.org/mlpack/mlpack/jobs/345837209#L4191

I have my exams starting from tomorrow. I will try and debug this issue once I get free.

zoq:
Okay, thanks for looking into the issue, I'll see if I can find something strange over the next days. Best of luck with your exams.

adl1995:
@zoq Thanks! The exams went well.

I just tried adding the constriction based variant, but as I had expected, it didn't make a difference for the logistic regression test. I can't seem to find a bug in the algorithm implementation, so, I'll test out different initialization techniques tomorrow. Please let me know if you have any suggestions on what could be a possible cause for this.

src/mlpack/core/optimizers/pso/pso_impl.hpp
```diff
@@ -0,0 +1,111 @@
...
+  // Initialize particle positions and velocities with the given points.
+  for (size_t i = 0; i < dimension; ++i)
+  {
+    particlePosition.slice(i) = iterate;
+    particleVelocity.slice(i) = iterate;
+  }
+
+  // Best particle and swarm position values are initialized
+  // from the first particle.
+  bestParticlePosition = particlePosition.slice(0);
```
zoq: Note ``iterate`` are the model parameter we like to optimize, so if we use PSO to train the logistic regression model, we have to update the model paramter; ``lr.ComputeAccuracy(..)`` internally uses the parameter that are passed once ``Optimize`` is called. So I think if you use ``iterate`` instead of ``bestParticlePosition`` it should work.

src/mlpack/tests/pso_test.cpp
```diff
@@ -0,0 +1,246 @@
...
+{
+  PSO optimizer;
+  EasomFunction f;
+
+  arma::mat iterate;
+  iterate << 3.14 << 3.14;
+
+  double result = optimizer.Optimize(f, iterate);
+
+  BOOST_REQUIRE_CLOSE(result, -1, 1e-3);
```
adl1995: I think this and some other functions are not being tested properly. In the `EasomFunction` class definition, it is stated: This should optimize to f(x) = -1, at x = [3.14, 3.14].

So, by providing the iterate `[3.14, 3.14]` as input, it would always provide -1 as the output. I tried passing in the iterate as `f.GetInitialPoint()`, but that returns the result as 0.
Maybe this should be dealt with differently. I tried to replicate the test from [adam_test.cpp](https://github.com/mlpack/mlpack/blob/119c410c56a1556749e2d98fc63edf21f40fd942/src/mlpack/tests/adam_test.cpp#L108), but that does not converge to the expected values.

adl1995:
@zoq That was a good catch. Thanks!

The build is passing again. I have some doubts regarding some other test functions, please see [this comment](https://github.com/mlpack/mlpack/pull/1225#discussion_r172620519).

src/mlpack/tests/pso_test.cpp
zoq: The defaults for the PSO are not necessarily good for every given problem, so some values used be tailored to the task at hand. In the context of this PR I think it's fine to select some functions (logistic regression, Bukin function, etc.) and see if we get stable results for each.

src/mlpack/core/optimizers/pso/constriction_factor.hpp
```diff
@@ -0,0 +1,103 @@
...
+   * @param particlePosition Position of the particles.
+   * @param particleVelocity Velocity of the particles.
+   * @param bestParticlePosition Best position of the particles.
+   * @param bestSwarmPosition Best position of the swarm.
+   * @param interiaWeight Inertia weight of the particles (omega).
+   * @param cognitiveAcceleration Cognitive acceleration of the particles.
+   * @param socialAcceleration Social acceleration of the particles.
+   */
+  void Update(const arma::cube& particlePosition,
+                    arma::cube& particleVelocity,
```
zoq: Can you align the parameter with the first one?

src/mlpack/core/optimizers/pso/constriction_factor.hpp
```diff
@@ -0,0 +1,103 @@
...
+                    const double& cognitiveAcceleration,
+                    const double& socialAcceleration,
+                    const double& dimension)
+  {
+    double phi = cognitiveAcceleration + socialAcceleration;
+    constrictionFactor = 2 / std::abs(2 - phi - std::sqrt(std::pow(phi, 2) - 4 * phi));
+
+    for (size_t i = 0; i < dimension; ++i)
+    {
+      for (size_t j = 0; j < particlePosition.slice(0).n_rows; ++j)
```
zoq: ``n_rows`` and ``n_cols`` should be implemented for ``arma::cube`` so, I would just go with that instead of using the first slice.

src/mlpack/core/optimizers/pso/constriction_factor.hpp
```diff
@@ -0,0 +1,103 @@
...
+      {
+        for (size_t k = 0; k < particlePosition.slice(0).n_cols; ++k)
+        {
+          particleVelocity.slice(i)[j, k] = constrictionFactor * (interiaWeight *
```
zoq: Do you think you could optimize the operation here, so that we don't have to go over each element? ``each_col`` might be helpful here.

zoq:
Looks like the lr test is failing again.

adl1995:
@zoq Yeah, I was able to update the velocity by updating each slice of `arma::cube` i.e. without iterating over each element, however, this leads the logistic regression test to failure. The `iterate` is in no way modified by the `Update` method, so I'm not sure why this happens.

I think there's more room for optimization here by using the [each_slice](http://arma.sourceforge.net/docs.html#each_slice) function. But, that requires conversion of other parameters to `arma::mat`, so we can take their product. For example, `interia_weight` has to be converted to a matrix:
```
arma::mat interiaWeight_matrix = arma::mat(particlePosition.n_rows, particlePosition.n_cols).fill(interiaWeight);
```
which can now be multiplied / added with `particleVelocity.each_slice()`. I haven't been able convert the velocity computation part to vectorized form completely, but I'll try to do so. This might also lead the logistic regression test to pass (although, I don't know the reason why its failing at the moment).

adl1995:
Ah, the problem was with the random numbers multiplication. Currently, I was only multiplying a single random number with each cube slice. This resulted in all the elements with the same value.

To solve this, I created a matrix of uniform random numbers `mat r1(particlePosition.n_rows, particlePosition.n_cols, fill::randu)` and performed element wise multiplication. The build should pass now.

adl1995:
The build is passing now. I had to increase the iteration count, so the accuracy could reach the acceptable tolerance.

Could you please let me know if I need to make further changes for this PR to be acceptable?

src/mlpack/core/optimizers/pso/pso_impl.hpp
```diff
@@ -0,0 +1,110 @@
...
+  {
+    for (size_t k = 0; k < dimension; ++k)
+    {
+      // Calculate the objective function.
+      currentObjective = function.Evaluate(particlePosition.slice(k));
+
+      // Check if the current position is an individual best.
+      if (currentObjective < lastObjectiveIndividual)
+      {
+        iterate = particlePosition.slice(k);
```
zoq: I think instead of updating ``iterate`` here, we should update an index and update iterate only once.

src/mlpack/core/optimizers/pso/constriction_factor.hpp
```diff
@@ -0,0 +1,105 @@
...
+              arma::cube& particleVelocity,
+              const arma::mat& bestParticlePosition,
+              const arma::mat& bestSwarmPosition,
+              const double& interiaWeight,
+              const double& cognitiveAcceleration,
+              const double& socialAcceleration,
+              const double& dimension)
+  {
+    // Calculate the constriction factor.
+    double phi = cognitiveAcceleration + socialAcceleration;
```
zoq: Do you think we should introduce something like a ``Reset`` or ``initialize`` method that takes the constant parameter, that way we calculate ``phi`` only once. Let me know what you think.

src/mlpack/core/optimizers/pso/pso.hpp
```diff
@@ -0,0 +1,177 @@
...
+   * @param cognitiveAcceleration Cognitive acceleration of the particles.
+   * @param socialAcceleration Social acceleration of the particles.
+   * @param maxIterations Maximum number of iterations allowed (0 means no
+   *     limit).
+   * @param tolerance Maximum absolute tolerance to terminate the algorithm.
+   * @param velocityType Velocity update policy used to calculate the
+   *     objective.
+   */
+  PSOType(const size_t dimension = 10,
+        const double interiaWeight = 0.9,
```
zoq: Can you align the parameter with the first one.

src/mlpack/core/optimizers/pso/constriction_factor.hpp
```diff
@@ -23,6 +23,23 @@ namespace optimization {
...
 {
  public:
+  /**
+   * Initialize the constant values required for updating velocity.
+   * In this case, the constriction factor is computed.
+   *
+   * @param cognitiveAcceleration Cognitive acceleration of the particles.
+   * @param socialAcceleration Social acceleration of the particles.
+   */
+  void Initialize(const double& cognitiveAcceleration,
```
adl1995: @zoq I have added an `Initialize` method which computes the constriction factor. However, in case of inertia weight variant of PSO, this method takes dummy parameters. Maybe this could be avoided using SFINAE, i.e. by calling `velocityType.Initialize` only if the `ConstrictionFactor` template is provided. But this might require code duplication.

This code executes different functions based on the template parameter: https://stackoverflow.com/a/30440624

zoq: I think the dummy parameter solution is just fine, SFINAE would be fine as well, but I think the current way is more readability and less complex, your call.

adl1995: Alright, I think the dummy parameter solution is good for now.

Regarding this PR, should I wait until the GSoC results are announced before this gets merged into master? Or, would it be fine to merge the basic PSO now? (assuming it doesn't require further changes)

zoq: No need to wait, once this is ready we can merge it. I'll have to recheck the update step.

sourabhvarshney111:
@zoq Can you review this? Although I am not sure, but I think after only few changes, we can merge this.

sourabhvarshney111:
@zoq I think this PR is waiting for review. Can you have a glance please?

rcurtin:
This would now be part of ensmallen, so if we'd like to get it merged, we should reopen the PR there.  Sorry that it went so many months without review---I do think it would be interesting to include, so @adl1995 if you'd like to adapt and reopen it there, I think we can review it and get it merged. :+1:

Issue #1223, opened 2018-02-03
use of internal armadillo functions (diskio)


conradsnicta:
Hi @rcurtin,

While implementing speedups in Armadillo's internal diskio class, I've discovered that mlpack is using an internal armadillo function:
https://github.com/mlpack/mlpack/blob/a16363a9c9cf3e9bf1145f5a593d384d74ab873f/src/mlpack/core/data/load_arff_impl.hpp#L200

I recommend changing the code so it doesn't use `arma::diskio::convert_naninf()`.  I'm planning to remove that function.

In general, it's not a good idea to use internal Armadillo functions, as they are subject to change and/or removal without notice.  Armadillo has strong API guarantees for public user-facing functions, but not for internal functions.

In a nutshell, functions and classes that are described in the Armadillo documentation ([docs.html](http://arma.sourceforge.net/docs.html)) are considered as public API. *Anything else (ie. not described in the documentation) is considered as internal implementation details.*  More info is given in the Armadillo [API policy](http://arma.sourceforge.net/docs.html#api_additions).

rcurtin:
Yep, you are fully right that we have left the reservation and are in uncharted territory.  But, rest assured that we are aware of the danger of our actions.  In this particular case, to give some context, we need to provide an ARFF reader, which Armadillo doesn't (and likely shouldn't since it has no notion of categorical variables; I think ARFF is out of scope for Armadillo).  We strongly desire that an ARFF will load identically to anything Armadillo loads, so we have to process non-numeric tokens in the same way Armadillo does.

This leaves us with two tough choices: (1) extract convert_naninf() and use it internally in mlpack, thus risking our ARFF reader getting out of sync when Armadillo's NaN/inf handling changes, or (2) use convert_naninf() in violation of the Armadillo API policy and then things may break with a new upstream release (which we try to monitor) and waste your time when you want to, e.g., make sure mlpack still builds and runs against a new experimental version of Armadillo.

I'd prefer for you to be able to avoid wasting your time screwing around with downstream software (although I don't at all mind the screwing around myself), so I am curious if you can think of a better idea for us to handle this issue.  I think there may be another couple instances where we have done things like this in the data loading code, and maybe we can come to a better solution that works for both projects.

conradsnicta:
I think option (2) is a non-starter, as it will continue to cause problems.  I think the best bet is a form of option (1), where mlpack is handling inf/nan using its own adapted form of convert_naninf().

[...]

PR #1226, opened 2018-02-06
Don't use Armadillo internal API.

rcurtin:
Pretty simple---I just pulled `convert_naninf()` out of Armadillo and switched to an internal implementation.  This is meant to solve #1223.

### DIFF
--- a/src/mlpack/core/data/is_naninf.hpp
+++ b/src/mlpack/core/data/is_naninf.hpp
@@ -1,0 +1,71 @@
+/**
+ * @file is_naninf.hpp
+ * @author Ryan Curtin
+ *
+ * This is an adapted version of Conrad Sanderson's implementation of
+ * arma::diskio::convert_naninf() from Armadillo.  It is here so as to avoid
+ * using Armadillo internal functionality.
+ */
+#ifndef MLPACK_CORE_DATA_HAS_NANINF_HPP
+#define MLPACK_CORE_DATA_HAS_NANINF_HPP
+
+#include <mlpack/prereqs.hpp>
+
+namespace mlpack {
+namespace data {
+
+/**
+ * See if the token is a NaN or an Inf, and if so, set the value accordingly and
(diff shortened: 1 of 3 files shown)


rcurtin:

rcurtin:

src/mlpack/core/data/is_naninf.hpp
```diff
@@ -0,0 +1,71 @@
...
+  if ((token.length() == 3) || (token.length() == 4))
+  {
+    const bool neg = (token[0] == '-');
+    const bool pos = (token[0] == '+');
+
+    const size_t offset = ((neg || pos) && (token.length() == 4)) ? 1 : 0;
+
+    const std::string token2 = token.substr(offset, 3);
+
+    if ((token2 == "inf") || (token2 == "Inf") || (token2 == "INF"))
```
zoq: I think this is faster as using: ``boost::algorithm::to_lower(str);`` and compare once; not sure.

src/mlpack/core/data/is_naninf.hpp
```diff
@@ -0,0 +1,71 @@
...
+    {
+      if (std::numeric_limits<T>::has_infinity)
+      {
+        if (!neg)
```
zoq: Should we use a ternary operator here, could improve the readability, if we say a ternary operator is easier to read 😄 .

rcurtin: Agreed, fixed in 69a85243e.

src/mlpack/core/data/is_naninf.hpp
rcurtin: I'll actually leave this one as-is so that it replicates Armadillo's functionality exactly.  When 8.400 comes out, I may update this function with whatever Armadillo is doing and then use some ifdefs to pick the same one as the Armadillo version.

PR #1229, opened 2018-02-08
Boost serialization 1.66 fix

rcurtin:
Right now the docker matrix build fails for boost 1.66:

http://masterblaster.mlpack.org/job/docker%20mlpack%20nightly%20build/armadillo_version=armadillo-7.100.3,boost_version=boost_1_66_0,buildmode=debug,compiler_version=gcc-7.2.0,label=linux-amd64/

I spent way too long digging into it, and this is the fix I came up with.  Basically we just need to ensure the ifstream/ofstream doesn't close before the archive is destructed.  Also I fixed a few warnings.

### DIFF
--- a/src/mlpack/tests/arma_extend_test.cpp
+++ b/src/mlpack/tests/arma_extend_test.cpp
@@ -70,7 +70,7 @@ BOOST_AUTO_TEST_CASE(ConstRowColIteratorTest)
   mat::const_row_col_iterator it;
   // Make sure ++ operator, operator* and comparison operators work fine.
   size_t count = 0;
-  for (it = X.begin_row_col(); it != X.end_row_col(); it++)
+  for (it = X.begin_row_col(); it != X.end_row_col(); ++it)
   {
     // Check iterator value.
     BOOST_REQUIRE_EQUAL(*it, (count % 5) * 3 + (count / 5));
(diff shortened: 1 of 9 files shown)


src/mlpack/tests/arma_extend_test.cpp
```diff
@@ -155,14 +155,14 @@ BOOST_AUTO_TEST_CASE(MatRowColIteratorDecrementOperatorTest)
   mat::row_col_iterator it1 = test.begin_row_col();
   mat::row_col_iterator it2 = it1;
 
-  // check that postfix-- does not decrement the position when position is
-  // pointing to the begining
-  it2--;
+  // Check that postfix-- does not decrement the position when position is
```
zoq: Somehow I was thinking about Postfix the mail server 😄 .

src/mlpack/tests/serialization.hpp
```diff
@@ -180,37 +194,43 @@ void SerializeObject(T& t, T& newT)
...
+    catch (boost::archive::archive_exception& e)
+    {
+      std::cout << e.what() << "\n";
+      success = false;
+    }
   }
   ifs.close();
 
-  remove(fileName.c_str());
+//  remove(fileName.c_str());
```
zoq: Looks like we can remove the line.

rcurtin: I actually commented that out for debugging purposes, and should re-add it.  I'll do that and then merge.

PR #1233, opened 2018-02-09
Modified CMakeLists.txt to not install as egg in python dist-packages

### LINKED ISSUE #1231: Python import takes too long.
I built mlpack from source according to the documentation.
Importing mlpack in python 2.7 takes too much time, I timed it:
It takes about 30 seconds to import mlpack or any of its specific parts like knn.
Is it due to it being kept as an egg file in python dist-packages?

akhandait:
Resolves #1231 
Added arguments to python setup.py install in CMakeLists.txt as mentioned here:[https://stackoverflow.com/questions/6301003/stopping-setup-py-from-installing-as-egg](url)
It now installs as a mlpack folder in python dist-packages and not an egg file, thus, drastically reducing the import time in python.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -23,7 +23,7 @@ before_install:
   # use.
   - sudo pip install cython setuptools numpy pandas
   - sudo pip3 install cython setuptools numpy pandas
-  - curl http://masterblaster.mlpack.org:5005/armadillo-6.500.5.tar.gz | tar xvz && cd armadillo*
+  - curl https://ftp.fau.de/macports/distfiles/armadillo/armadillo-6.500.5.tar.gz | tar xvz && cd armadillo*
   - cmake . && make && sudo make install && cd ..
   - sudo cp .travis/config.hpp /usr/include/armadillo_bits/config.hpp
 
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -280,6 +280,7 @@ set(MLPACK_LIBRARIES ${MLPACK_LIBRARIES} ${ARMADILLO_LIBRARIES})
 # Unfortunately this configuration variable is necessary and will need to be
 # updated as time goes on and new versions are released.
 set(Boost_ADDITIONAL_VERSIONS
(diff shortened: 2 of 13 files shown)


rcurtin:
Hi there, thanks for the contribution.  I spent a long time reading about it and there seems to be a lot of reports that using `single-version-externally-managed` can cause issues.  Worse, I couldn't find clear documentation on what that option actually does and whether it is the right thing to use in our case.

But I did find the following link:
https://stackoverflow.com/questions/27964960/pip-why-sometimes-installed-as-egg-sometimes-installed-as-files

This suggests that maybe `zip_safe` is the better way to go (and it seems better documented).  What do you think?  Would you be willing to try this and see if it works as expected?

akhandait:
Yeah sure, I will try this out an let you know.

akhandait:
Should we shift from python setup.py install to pip install?

rcurtin:
That would require pip to be available on the system, so I would personally prefer to avoid it.  However, I would be fine with detecting whether pip is available on the system, and if so, using it to install.  But that will be a little bit more code to write, so your call on if you want to pursue that.

akhandait:
Sorry, I think messed up with some commits.
I found the documentation of zip-safe to be quite concrete.[http://setuptools.readthedocs.io/en/latest/setuptools.html#setting-the-zip-safe-flag](url)
It now extracts itself and works.

akhandait:
I seem to have synced my working branch with all the other commits, should I create a new pull request?

rcurtin:
Hmm, I think you are right, looks like something went a little weird with the PR.  I would say go ahead and create a new branch with just the `zip_safe` change and open that as a PR (and close this one).  If it passes the Travis build then I think it is ready.  (I may have to restart the Travis build a few times, since we are using the xenial image which is unstable and has problems... we are in the process of switching back to trusty but it isn't ready yet: #1221.)

Thank you for looking into this!  And you are right, the zip_safe documentation is a lot clearer and more understandable.  I don't think there should be any problems with that.

Issue #1222, opened 2018-02-02
Greedy KNN doesn't always return enough results


rcurtin:
@nvasil pointed this one out.  If you try to use k-nearest-neighbor search with the greedy algorithm, it may not return results.  Here is a simple way to reproduce the issue:

```
$ bin/mlpack_knn -r test_data_3_1000.csv -k 3 -a greedy -n n.csv -d d.csv -l 1
```

That will calculate the nearest neighbors of every point in `test_data_3_1000.csv` (though any dataset will work) building a tree with a leaf size of 1 (meaning that every leaf in the tree holds only one point).  Since we are using the same query set as the reference set, a point cannot be its own nearest neighbor.  So for some point p, the greedy algorithm will descend the tree directly to the leaf containing only the point p, and no result will be saved.  In our example above, this means that no nearest neighbors are set; we can see that by looking at the results in n.csv and d.csv:

```
$ head d.csv
1.79769313486232e+308,1.79769313486232e+308,1.79769313486232e+308
1.79769313486232e+308,1.79769313486232e+308,1.79769313486232e+308
1.79769313486232e+308,1.79769313486232e+308,1.79769313486232e+308
1.79769313486232e+308,1.79769313486232e+308,1.79769313486232e+308
1.79769313486232e+308,1.79769313486232e+308,1.79769313486232e+308
1.79769313486232e+308,1.79769313486232e+308,1.79769313486232e+308
1.79769313486232e+308,1.79769313486232e+308,1.79769313486232e+308
1.79769313486232e+308,1.79769313486232e+308,1.79769313486232e+308
1.79769313486232e+308,1.79769313486232e+308,1.79769313486232e+308
1.79769313486232e+308,1.79769313486232e+308,1.79769313486232e+308
```

```
$ head n.csv
8017,8017,8017
8017,8017,8017
8017,8017,8017
8017,8017,8017
8017,8017,8017
8017,8017,8017
8017,8017,8017
8017,8017,8017
8017,8017,8017
8017,8017,8017
```

The responsible code is in `src/mlpack/core/tree/greedy_single_tree_traverser_impl.hpp`: the recursion there simply recurses until hitting a leaf node.  However, intuitively, the change that needs to happen is that for nearest neighbor search, we need to terminate the recursion and perform all the point-to-point base cases the level before `referenceNode.NumDescendants() < k`.

This would be a good issue for someone who is looking to learn about the tree recursion code in mlpack.

manish7294:
[...]

PR #1234, opened 2018-02-10
Greedy Traverser Fix for KNN

manish7294:
Fix for #1222 
Traverser traverse through greedy technique till more than `k` points are available and after that base case is taken over all the descendant points.
In order to access `k` value  through greedy, a private member `k` has been declared which is default set to 0 and hence do not interfere with the generality of traverser.For trees like kd-tree, ball tree which stores points in leaf nodes `k` can be set using `K()`.

### DIFF
--- a/src/mlpack/core/tree/greedy_single_tree_traverser.hpp
+++ b/src/mlpack/core/tree/greedy_single_tree_traverser.hpp
@@ -41,7 +41,10 @@ class GreedySingleTreeTraverser
   size_t NumPrunes() const { return numPrunes; }
 
   //! Set value of minBaseCases.
-  void MinBaseCases(size_t baseCases) { minBaseCases = baseCases; }
+  size_t& MinBaseCases() { return minBaseCases; }
+
+  //! Get value of minBaseCases.
+  size_t MinBaseCases() const { return minBaseCases; }
 
  private:
   //! Reference to the rules with which the tree will be traversed.
--- a/src/mlpack/core/tree/greedy_single_tree_traverser.hpp
+++ b/src/mlpack/core/tree/greedy_single_tree_traverser.hpp
@@ -40,8 +40,8 @@ class GreedySingleTreeTraverser
   //! Get the number of prunes.
   size_t NumPrunes() const { return numPrunes; }
 
-  //! Set value of k.
-  void K(size_t K) { k = K; }
+  //! Set value of minBaseCases.
+  void MinBaseCases(size_t baseCases) { minBaseCases = baseCases; }
 
  private:
   //! Reference to the rules with which the tree will be traversed.
@@ -50,9 +50,9 @@ class GreedySingleTreeTraverser
   //! The number of nodes which have been pruned during traversal.
   size_t numPrunes;
 
-  //! The number of results required. For example number of nearest
-  //! neighbours in case of knn.
-  size_t k;
+  //! The number of base cases required. For example the number of nearest
+  //! neighbours(k) in case of knn.
+  size_t minBaseCases;
 };
 
 } // namespace tree
--- a/src/mlpack/core/tree/greedy_single_tree_traverser.hpp
+++ b/src/mlpack/core/tree/greedy_single_tree_traverser.hpp
@@ -40,12 +40,19 @@ class GreedySingleTreeTraverser
   //! Get the number of prunes.
   size_t NumPrunes() const { return numPrunes; }
 
+  //! Set value of k.
+  void K(size_t K) { k = K; }
+
  private:
   //! Reference to the rules with which the tree will be traversed.
   RuleType& rule;
 
   //! The number of nodes which have been pruned during traversal.
   size_t numPrunes;
+
+  //! The number of results required. For example number of nearest
+  //! neighbours in case of knn.
+  size_t k;
 };
 
 } // namespace tree
--- a/src/mlpack/core/tree/greedy_single_tree_traverser_impl.hpp
+++ b/src/mlpack/core/tree/greedy_single_tree_traverser_impl.hpp
@@ -24,7 +24,8 @@ template<typename TreeType, typename RuleType>
 GreedySingleTreeTraverser<TreeType, RuleType>::GreedySingleTreeTraverser(
     RuleType& rule) :
     rule(rule),
-    numPrunes(0)
+    numPrunes(0),
+    k(0)
 { /* Nothing to do. */ }
 
 template<typename TreeType, typename RuleType>
@@ -32,17 +33,38 @@ void GreedySingleTreeTraverser<TreeType, RuleType>::Traverse(
     const size_t queryIndex,
     TreeType& referenceNode)
 {
-  // Run the base case as necessary for all the points in the reference node.
-  for (size_t i = 0; i < referenceNode.NumPoints(); ++i)
-    rule.BaseCase(queryIndex, referenceNode.Point(i));
+  if(referenceNode.IsLeaf())
+  {
+    // Run the base case as necessary for all the points in the reference node.
+    for (size_t i = 0; i < referenceNode.NumPoints(); ++i)
+      rule.BaseCase(queryIndex, referenceNode.Point(i));
+    return;
+  }
 
-  if (!referenceNode.IsLeaf())
+  size_t bestChild = rule.GetBestChild(queryIndex, referenceNode);
+  size_t numDescendants;
+
+  // Check that referencenode is not a leaf node while calculating number of
+  // descendants of it's best child.
+  if(!referenceNode.IsLeaf())
+    numDescendants = referenceNode.Child(bestChild).NumDescendants();
+  else
+    numDescendants = referenceNode.NumPoints();
+
+  // If number of descendants are more than k than we can go along with
+  // best child otherwise we need to traverse for each descendant to
+  // ensure that we get at least k nearest neighbors..
+  if (numDescendants > k)
   {
     // We are prunning all but one child.
     numPrunes += referenceNode.NumChildren() - 1;
     // Recurse the best child.
-    size_t bestChild = rule.GetBestChild(queryIndex, referenceNode);
     Traverse(queryIndex, referenceNode.Child(bestChild));
+  }
+  else
+  {
+    for (size_t i = 0; i < referenceNode.NumChildren(); ++i)
+      Traverse(queryIndex, referenceNode.Child(i));
   }
 }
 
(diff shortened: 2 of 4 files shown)


Prabhat-IIT:
It looks good @manish7294  .I was also doing the same thing but was confused on the part of `numChildren` and `numDescendants`. Congrats :)

src/mlpack/core/tree/greedy_single_tree_traverser_impl.hpp
```diff
@@ -24,26 +24,48 @@ template<typename TreeType, typename RuleType>
...
 template<typename TreeType, typename RuleType>
 void GreedySingleTreeTraverser<TreeType, RuleType>::Traverse(
     const size_t queryIndex,
     TreeType& referenceNode)
 {
-  // Run the base case as necessary for all the points in the reference node.
-  for (size_t i = 0; i < referenceNode.NumPoints(); ++i)
-    rule.BaseCase(queryIndex, referenceNode.Point(i));
+  if(referenceNode.IsLeaf())
+  {
```
rcurtin: So, this conditional actually is not necessary.  While the kd-tree only holds points in the leaves, other trees (like the cover tree) may hold some points in non-leaf nodes.  So we should calculate the base case of any points even as we go down.  We may have to track the total number of base cases computed in the traverser, but I think that is just fine to do.

src/mlpack/core/tree/greedy_single_tree_traverser_impl.hpp
```diff
@@ -24,26 +24,48 @@ template<typename TreeType, typename RuleType>
...
     numPrunes += referenceNode.NumChildren() - 1;
     // Recurse the best child.
-    size_t bestChild = rule.GetBestChild(queryIndex, referenceNode);
     Traverse(queryIndex, referenceNode.Child(bestChild));
   }
+  else
+  {
+    for (size_t i = 0; i < referenceNode.NumChildren(); ++i)
+      Traverse(queryIndex, referenceNode.Child(i));
+  }
```
rcurtin: It looks like this code will traverse even when the node is a leaf.  Instead if we do the pruning, we can just loop over all the points directly.  I think you could structure more like this:

```
if (!referenceNode.IsLeaf())
{
  const size_t bestChild = ... ;
  if (referenceNode.Child(bestChild).NumDescendants() > (k - baseCases))
  {
    Traverse(queryIndex, referenceNode.Child(bestChild);
  }
  else
  {
    for (size_t i = 0; i < referenceNode.NumDescendants(); ++i) // I guess we could terminate at k points also
    {
      rule.BaseCase(queryIndex, referenceNode.Descendant(i));
    }
  }
}
```

What do you think?  The idea there might not be quite right but I hope it is enough to go on.

src/mlpack/core/tree/greedy_single_tree_traverser.hpp
```diff
@@ -40,12 +40,19 @@ class GreedySingleTreeTraverser
   //! Get the number of prunes.
   size_t NumPrunes() const { return numPrunes; }
 
+  //! Set value of k.
+  void K(size_t K) { k = K; }
```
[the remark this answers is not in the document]
manish7294: Yeah, that way It will be more understandable.

src/mlpack/core/tree/greedy_single_tree_traverser_impl.hpp
manish7294: Right! Imposing this condition is just a reduction of results for trees storing points in other nodes also. This will just make them behave as the earlier greedy kd-tree.

src/mlpack/core/tree/greedy_single_tree_traverser_impl.hpp
manish7294: Curently this is not happening, but will take place once `referenceNode.IsLeaf()` condition is removed while calculating base_cases. So, this will be a good measure to impose `!referenceNode.IsLeaf()` condition here. But I don't quite understand the use of `(k - baseCases) `. I think it will cause a bit of uncertainity here. It think it will work fine to just use `referenceNode.Child(bestChild).NumDescendants() > k` as in case we do not determine k the code will just convert back to previous one.  And tracking BaseCases here may not be optimum, as we may have to declare baseCases globally or will have to change the class definition. Let, me know if I interpreted it correctly.
```

for (size_t i = 0; i < referenceNode.NumDescendants(); ++i) // I guess we could terminate at k points also
    {
      rule.BaseCase(queryIndex, referenceNode.Descendant(i));
    }
```
Nice, using NumDescendant is better way of doing the same and now we can just break the loop at ``i==k``. But breaking at` i==k` may give somewhat worse results for nearest neighbours. But as we know this is greedy technique, so everything is fair in greedy till we pick the best option before us  :)

src/mlpack/core/tree/greedy_single_tree_traverser_impl.hpp
```diff
@@ -24,26 +24,48 @@ template<typename TreeType, typename RuleType>
...
+
+  size_t bestChild = rule.GetBestChild(queryIndex, referenceNode);
+  size_t numDescendants;
 
-  if (!referenceNode.IsLeaf())
+  // Check that referencenode is not a leaf node while calculating number of
+  // descendants of it's best child.
+  if(!referenceNode.IsLeaf())
+    numDescendants = referenceNode.Child(bestChild).NumDescendants();
```
manish7294: We need to calculate `numDescendant` with `!referenceNode.IsLeaf()` condition imposed, otherwise on asking for child of leaf node it will give us segmentation fault.

src/mlpack/core/tree/greedy_single_tree_traverser_impl.hpp
```diff
@@ -36,13 +37,34 @@ void GreedySingleTreeTraverser<TreeType, RuleType>::Traverse(
...
 
+  size_t bestChild = rule.GetBestChild(queryIndex, referenceNode);
+  size_t numDescendants;
+
+  // Check that referencenode is not a leaf node while calculating number of
+  // descendants of it's best child.
+  if (!referenceNode.IsLeaf())
+    numDescendants = referenceNode.Child(bestChild).NumDescendants();
+  else
+    numDescendants = referenceNode.NumPoints();
```
rcurtin: You can move the whole definition of `numDescendants` into the `if (!referenceNode.IsLeaf())` block, and then you won't need the `if`/`else` here---you can just set

```
numDescendants = referenceNode.Child(bestChild).NumDescendants();
```

src/mlpack/core/tree/greedy_single_tree_traverser.hpp
```diff
@@ -40,12 +40,19 @@ class GreedySingleTreeTraverser
   //! Get the number of prunes.
   size_t NumPrunes() const { return numPrunes; }
 
+  //! Set value of minBaseCases.
+  void MinBaseCases(size_t baseCases) { minBaseCases = baseCases; }
```
rcurtin: I'm really sorry to be picky on this one, especially because I know there is ambiguity in the mlpack code.  I think it is better (or matches more of the mlpack codebase) to allow direct access:

```
//! Set value of minBaseCases.
size_t& MinBaseCases() { return minBaseCases; }
//! Get value of minBaseCases.
size_t MinBaseCases() const { return minBaseCases; }
```

src/mlpack/core/tree/greedy_single_tree_traverser_impl.hpp
manish7294: I wrote about this in a comment, may be you missed it as it went down with "show outdated" :)
Here it is :
> We need to calculate numDescendant with !referenceNode.IsLeaf() condition imposed, otherwise on asking for child of leaf node it will give us segmentation fault. This was the silly error that I mentioned about in #1222 comment. It cause me a lot of trouble.
Please let me know if you mean the same thing.

src/mlpack/core/tree/greedy_single_tree_traverser.hpp
manish7294: No problem, thanks for pointing it out. Will update it soon.

manish7294:
Thanks @rcurtin for helping through this. If it wasn't for you explaining it so clearly and patiently, I may not have been able to come up with this easily :) . Thanks again, you always help me understand things in a lot better way.

zoq:
Might be a good idea to write a simple test case, to make sure this doesn't happen again.

manish7294:
@zoq Is writing a test for greedy with all tree variants in knn bindings test a good idea? As this happens with few tree types.

rcurtin:
I agree, it would be a good idea to write a test case.  I would say just do it with one tree type, but if you want to add more types to test feel free.

manish7294:
Okay, then I will be doing this within knn bindings test as it's still need to implemented.

rcurtin:
The test should actually be a part of `knn_test.cpp` not `main_tests/knn_test.cpp` in my opinion.  I am ok with merging this without a test (I will run a check by hand when I merge it), but we should make a note in #1222 that we need to add a test before that issue can be closed (so we don't forget or overlook it).  Then you can add the test in another PR if you want.

manish7294:
Sure, that will be good. And if we are adding it in knn_test.cpp then we may just check for one tree type as you suggested. 
Should I add a comment regarding this in #1222 or will you be editing the issue?

src/mlpack/tests/knn_test.cpp
```diff
@@ -1434,4 +1434,36 @@ BOOST_AUTO_TEST_CASE(MoveOperatorNaiveTest)
...
+ * Check that no garbage value is returned when greedy tree traversal
+ * is performed over kd-tree.
+ */
+BOOST_AUTO_TEST_CASE(GreedyTreeSearch)
+{
+  // Initalize dataset.
+  arma::mat dataset = arma::randu<arma::mat>(3, 100);
+
+  KDTree<EuclideanDistance, NeighborSearchStat<NearestNeighborSort>,
+      arma::mat> tree(dataset);
```
rcurtin: To improve the robustness of this test, I'd suggest setting a leaf size of 1 when you build the tree.  It is still possible that you can get leaves in the tree with less than 5 points in them, but it's not guaranteed---so probably better to force a leaf size of 1 to make sure the problematic situation occurs for this test.

manish7294: Sure, I will make that change.

rcurtin:
Thanks for the contribution!  Glad to see this fixed. :)

Issue #1227, opened 2018-02-07
Avoid setting random seed in main tests


rcurtin:
This one is related to #1152 and would be a good issue for someone who is looking to contribute.

One of the keys about the test `mlpack_test` is that its results are reproducible for a single build.  So, the random seed should never be set during the test, instead preferring to leave it to whatever it is set to during compilation.  This means I can run `mlpack_test` multiple times, and if a single random test failed, it will fail again.  This is important for debugging.

However, the recent addition of the main tests in #1152 is actually setting the random seed in many cases to `std::time(NULL)`, which is different for every run.  That breaks the idea set out above... so we should fix it. :)

I would suggest the following steps as a good way to approach this issue:

 * Read through the code to understand how `math::RandomSeed(std::time(NULL))` is being called by some of the main tests in `src/mlpack/tests/*.cpp`.  The call will be in `mlpackMain()` of some of the algorithms being tested.

 * See that `math::RandomSeed()` is actually being called during the tests, perhaps by adding some printed debugging output to the call or setting a breakpoint or something.

 * Think about a way where, if we are in the tests (`mlpack_test` only), we can either avoid calling `math::RandomSeed()` altogether, or make sure that it doesn't actually set the random seed.

 * Implement it!

 * Test that setting the random seed still works for the command-line programs and Python bindings by manually running a program that has the `--seed` option with different seeds, and seeing that the output is different (k-means is one good option).

Regardless of whose PR is merged in the end, I think it is a great learning experience for the codebase for multiple people to work on this, so there is no need to "claim" the issue.  I'm likely to review and merge the first working PR that is submitted.

Prabhat-IIT:
@rcurtin here we have to avoid calling `math::RandomSeed(std::time(NULL))` or both `math::RandomSeed((size_t) CLI::GetParam<int>("seed"))` and `math::RandomSeed(std::time(NULL))`. I'm confused if we avoid calling both i.e. avoid calling `math::RandomSeed()` altogether then how will the randomseed will be set? What should be it's value when we don't pass the `seed` as input Param as well as when we are not allowing the `seed` to be it's default value i.e. 0 ?
If we've to just not allow the calling of `math::RandomSeed()` for `std::time(NULL)` then I think we should decide a fixed value of `seed` whenever `mlpackMain()` is called from any of the `mlpack_tests`. To do this I've two things in mind:

[...]

PR #1236, opened 2018-02-11
Resolved bug 

Prabhat-IIT:
Fix for #1227 
Whenever `mlpackMain()` is called from any of the test suite a random seed is set by calling `RandomSeed(time(NULL))` which is different for different runtimes. Thus, call to `RandomSeed()` is restricted from any of mlpack test suites
I've done it as of now for only `preprocess_split_test`. If it is acceptable then I can extend it for every test.
wip

### DIFF
--- a/src/mlpack/methods/preprocess/preprocess_split_main.cpp
+++ b/src/mlpack/methods/preprocess/preprocess_split_main.cpp
@@ -76,10 +76,13 @@ static void mlpackMain()
   // Parse command line options.
   const double testRatio = CLI::GetParam<double>("test_ratio");
 
-  if (CLI::GetParam<int>("seed") == 0)
-    mlpack::math::RandomSeed(std::time(NULL));
-  else
-    mlpack::math::RandomSeed((size_t) CLI::GetParam<int>("seed"));
+#if(BINDING_TYPE != BINDING_TYPE_TEST) // This is a unit test
+
+      if ((CLI::GetParam<int>("seed") == 0))
+          mlpack::math::RandomSeed(std::time(NULL));
+      else
+          mlpack::math::RandomSeed((size_t) CLI::GetParam<int>("seed"));
+#endif
 
   // Make sure the user specified output filenames.
   RequireAtLeastOnePassed({ "training" }, false, "no training set will be "


src/mlpack/methods/preprocess/preprocess_split_main.cpp
```diff
@@ -76,10 +76,13 @@ static void mlpackMain()
   // Parse command line options.
   const double testRatio = CLI::GetParam<double>("test_ratio");
 
-  if (CLI::GetParam<int>("seed") == 0)
-    mlpack::math::RandomSeed(std::time(NULL));
-  else
-    mlpack::math::RandomSeed((size_t) CLI::GetParam<int>("seed"));
+  if (BINDING_TYPE != BINDING_TYPE_TEST)
```
rcurtin: Two things:

1. By using `if` and not `#if`, this check gets compiled into the code (unless the code is properly optimized), so it is better to use the preprocessor here.  In general I would say if you are dealing with preprocessor macros, you should use the logical functionality provided by the preprocessor.

2. If we perform the check here, inside the bindings, we have to modify every single binding that has a `seed` option.  Wouldn't it be better to make changes to the `RandomSeed()` function directly?

src/mlpack/methods/preprocess/preprocess_split_main.cpp
```diff
@@ -76,10 +76,13 @@ static void mlpackMain()
   // Parse command line options.
   const double testRatio = CLI::GetParam<double>("test_ratio");
 
-  if (CLI::GetParam<int>("seed") == 0)
-    mlpack::math::RandomSeed(std::time(NULL));
-  else
-    mlpack::math::RandomSeed((size_t) CLI::GetParam<int>("seed"));
+#if(BINDING_TYPE != BINDING_TYPE_TEST) // This is a not unit test
```
rcurtin: I don't think you handled the second comment yet:

> If we perform the check here, inside the bindings, we have to modify every single binding that has a seed option. Wouldn't it be better to make changes to the RandomSeed() function directly?

Prabhat-IIT: I'm testing that. Please give me some time. I'll handle all the requirements after my exam. I'm extremely sorry for the delay :(

rcurtin: Sure, no worries.  There is no hurry at all and no reason to apologize. :)  I thought this was ready since you pushed some commits.  Good luck with your exam!

Prabhat-IIT:
@rcurtin there's a macro conflict in case I try to handle the issue inside `RandomSeed()` because the `random` function is included in both `CLI_BINDINGS` as well as `TEST_BINDINGS` due to 'BINDING_TYPE` macro has both the values in one program or another. Another thing to notice here that in `mlpack_tests` like `preprocess_split_test` `main_method` like `preprocess_split_main` is included after defining `BINDING_TYPE=BINDING_TYPE_TEST` which ultimately sets the `BINDING_TYPE` to be `BINDING_TYPE_CLI`. If you say I can fix this in each test by either first undefining the `BINDING_TYPE` and then defining it or by including main methods before defining the macro.

rcurtin:
Hi @Prabhat-IIT, this should be a very simple PR and I am not sure I understand your discussion above.  Let's have Roberto reopen his PR and see if that works; it was closed before I was able to review it.  I think that one worked fine.

Prabhat-IIT:
Hi @rcurtin that fix didn't worked for me. Even Roberto agreed on that! The present fix which Roberto proposed is very simple and I've tried it as my first attempt even before this pr but it didn't worked! Why it didn't worked for me? because whenever I'm testing to print something inside `RandomSeed()` it didn't print in case of `BINDING_TYPE_CLI` So, it means that `RandomSeed()` is not being called. The situation is same even now and I don't know how you tested that it works for `BINDING_TYPE_CLI` Moreover, since now it has been fixed according to you then I'll close this pr

PR #1237, opened 2018-02-11
mlpack_hmm_train binding tests

PlantsAndBuildings:
#1152. One of the tests is to check that the tolerance cannot be < 0. As of now, it is failing - maybe this needs a fix? Please let me know if I've overlooked any important tests, I'll be sure to add them as well.


src/mlpack/tests/main_tests/hmm_training_tests.cpp
```diff
@@ -0,0 +1,209 @@
...
+  double tol = - 100;  // Invalid
+
+  fileExists(inputFileName);
+  SetInputParam("input_file", std::move(inputFileName));
+  SetInputParam("states", states);
+  SetInputParam("type", std::move(hmmType));
+  SetInputParam("tolerance", tol);
+
+  Log::Fatal.ignoreInput = true;
+  BOOST_REQUIRE_THROW(mlpackMain(), std::runtime_error);
```
manish7294: @PlantsAndBuildings `hmm_train_main.cpp` currently doesn't have a check for the value of tolerance, that's may be the reason why there is an error as `BOOST_REQUIRE_THROW()` is expecting to receive a runtime error. You may solve this by adding a check in for tolerance in ` hmm_train_main`.  It's your choice to do so don't feel obligated.
May be something like this could be done:
```
RequireParamValue<int>("tolerance", [](int x) { return x >= 0; }, true,
        "tolerance must be non-negative");
```

PlantsAndBuildings: @manish7294 . Thanks, I've added the check like you suggested.

src/mlpack/methods/hmm/hmm.hpp
```diff
@@ -328,6 +328,9 @@ class HMM
   template<typename Archive>
   void serialize(Archive& ar, const unsigned int version);
 
+  bool ApproximatelyEqual(const HMM<Distribution>& other, 
+                          double tolerance) const;
+
```
rcurtin: Since this is used only for testing, could we instead add this in the test file only instead of the `HMM` code?

src/mlpack/methods/hmm/hmm_train_main.cpp
```diff
@@ -379,6 +379,12 @@ static void mlpackMain()
         "unknown HMM type");
   }
 
+  if (CLI::HasParam("tolerance"))
+  {
+    RequireParamValue<int>("tolerance", [](int x) { return x >= 0; }, true,
+        "tolerance must be non-negative");
+  }
```
rcurtin: Thanks for adding this.  I think the `if (CLI::HasParam("tolerance"))` is unnecessary though---if the parameter isn't passed, it will be the default, which is greater than 0.

src/mlpack/tests/CMakeLists.txt
```diff
@@ -138,6 +138,7 @@ add_executable(mlpack_test
   main_tests/random_forest_test.cpp
   main_tests/softmax_regression_test.cpp
   main_tests/sparse_coding_test.cpp
+  main_tests/hmm_training_tests.cpp
```
rcurtin: I know this is picky... do you think we could name the file `hmm_train_test.cpp` so it matches the names of the other files?  Thanks!

src/mlpack/tests/main_tests/hmm_training_tests.cpp
```diff
@@ -0,0 +1,209 @@
...
+  {
+    // Clear the settings.
+    bindings::tests::CleanMemory();
+    CLI::ClearSettings();
+  }
+};
+
+BOOST_FIXTURE_TEST_SUITE(HMMTrainMainTest, HMMTrainMainTestFixture);
+
+inline void fileExists(std::string fileName)
```
rcurtin: I know the style checker is not running right now, but I hope to have that back online in the next day or two.  It can tell you about these things automatically.  In any case, can we call this function `FileExists()` so it matches the style guide?

src/mlpack/tests/main_tests/hmm_training_tests.cpp
```diff
@@ -0,0 +1,209 @@
...
+  HMMModel h1 = *(CLI::GetParam<HMMModel*>("output_model"));
+
+  SetInputParam("input_model", std::move(ph1));
+
+  CLI::GetSingleton().Parameters()["type"].wasPassed = false;
+  CLI::GetSingleton().Parameters()["states"].wasPassed = false;
+
+  mlpackMain();
+
+  HMMModel h2 = *(CLI::GetParam<HMMModel*>("output_model"));
```
rcurtin: A minor note, but you could avoid the copy here by just working with the pointer directly.  Since this is a test on a small dataset it doesn't really matter, but something to keep in mind in any case.

src/mlpack/tests/main_tests/hmm_training_tests.cpp
```diff
@@ -0,0 +1,209 @@
...
+  SetInputParam("type", std::move(hmmType));
+  SetInputParam("gaussians", gaussians);
+
+  Log::Fatal.ignoreInput = true;
+  BOOST_REQUIRE_THROW(mlpackMain(), std::runtime_error);
+  Log::Fatal.ignoreInput = false;
+}
+
+// Make sure that model reuse is possible and work properly
+BOOST_AUTO_TEST_CASE(HMMTrainReuseModelTest)
```
rcurtin: I think it might be useful to also test this when training without labels (so the Baum-Welch algorithm is used for training instead of direct estimation).

PlantsAndBuildings:
@rcurtin . Thanks for the review; sorry about the style issues (I should have been careful about those). I'll make the changes you suggested today. I also do plan to write tests for the other hmm CLI utilities after this.

src/mlpack/methods/hmm/hmm_train_main.cpp
PlantsAndBuildings: Yes, you're right. Also I think that the template param here should be double instead of int -- will fix it.

src/mlpack/methods/hmm/hmm.hpp
PlantsAndBuildings: I also hesitated while adding this function -- it didnt seem very clean. However, the HMMModel class doesn't expose pointers to HMM<Distribution>. Should I provide accessors like
``` c++
HMM<distribution::DiscreteDistribution> * DiscreteHMM() { return discreteHMM; }
HMM<distribution::GaussianDistribution> * GaussianHMM() { return gaussianHMM; }
HMM<gmm::GMM> * GMMHMM() { return gmmHMM; }
```
in the HMMModel class and then an ApproximatelyEqual() function in the test file? Is there another (better) way?

src/mlpack/tests/main_tests/hmm_training_tests.cpp
PlantsAndBuildings: I think that atleast one copy is necessary. Consider the following
```c++
// Set some params
SetInputParam(...)

mlpackMain()

// Get the pointer to output model
HMMModel * ph1 = CLI::GetParam<HMMModel*>("output_model");

// Change some params
CLI::GetSingleton().Parameters()["..."].wasPassed = false;
SetInputParam(...)

mlpackMain();

// Again get the pointer to output model. But now the old output model is overwritten!
// Hence ph1 and ph2 both point to the same (latest) output model
HMMModel * ph2 = CLI::GetParam<HMMModel*>("output_model");
// Since we've lost out old output model, we cannot compare
```
Sorry if I'm missing something obvious

src/mlpack/methods/hmm/hmm.hpp
rcurtin: Yeah, I think that is a reasonable way to do it, since others may have use for those four functions you are proposing.  I would ask that you document those functions and point out that the pointers for the other types of HMMs will be NULL... i.e. if the HMMModel holds a discrete HMM, then GaussianHMM() will return NULL and so will GMMHMM(), and the user should be careful.

src/mlpack/tests/main_tests/hmm_training_tests.cpp
rcurtin: Yep, you're right, but what I was focusing on was the copy because `h2` is not a pointer:

```
HMMModel h2 = *(CLI::GetParam<HMMModel*>("output_model"));
```

Since we are not running `mlpackMain()` again, we could just use a pointer to the second model to avoid copying the `HMMModel`:

```
HMMModel* h2 = CLI::GetParam<HMMModel*>("output_model");
```

or even use std::move():

```
HMMModel h2 = std::move(*(CLI::GetParam<HMMModel*>("output_model")));
```

This is a pretty minor comment in any case, especially since `HMMModel` only holds a few pointers so the cost of a copy is very small.  But for implementing other tests or other mlpack code, it might be something worthwhile to keep in mind. :)

src/mlpack/tests/main_tests/hmm_train_test.cpp
```diff
@@ -51,6 +51,70 @@ inline void FileExists(std::string fileName)
...
+        tolerance
+        );
+    initialEqual = approx_equal(
+        h1.DiscreteHMM()->Transition(),
+        h2.DiscreteHMM()->Transition(),
+        "absdiff",
+        tolerance
+        );
+    // TODO
+    emissionEqual = true;
```
rcurtin: Don't forget the todo here. :)  (I am writing this comment in part so that I remember to double-check that this gets done before merge... probably you already know that you still need to do it. :))

src/mlpack/tests/main_tests/hmm_train_test.cpp
```diff
@@ -0,0 +1,404 @@
...
+{
+  if (h1.Type() != h2.Type())
+    return false;
+  HMMType  hmmType = h1.Type();
+  bool transitionEqual = false;
+  bool emissionEqual = false;
+  bool initialEqual = false;
+  if (hmmType ==  DiscreteHMM)
+  {
+    transitionEqual = approx_equal(
```
rcurtin: Hmm, this is unfortunate... the minimum version of Armadillo for mlpack is 6.500, but `approx_equal()` isn't available until 6.700.  I don't think it's realistic to bump the minimum version of Armadillo for this, so there are two things we could do:

 * Backport `fn_approx_equal.hpp`, by taking `fn_approx_equal.hpp` from the newest version of Armadillo, putting some `#if` guards around it so that it only gets included for sufficiently old versions of Armadillo (older than 6.700), and adding it to `arma_extend.hpp`.  Here's a commit where I did exactly that for a sparse matrix subtraction operator: https://github.com/mlpack/mlpack/commit/62def05b672f5a44dbdc2c721b522ac4bf10c7d0#diff-b3e69ecff96ebaae403c243f71c038db

 * Avoid `approx_equal` and write a loop by hand or use `CheckMatrices()` from `test_tools.hpp`.

I think `CheckMatrices` may be the easiest way to go, but it's up to you how you want to do it.

src/mlpack/tests/main_tests/hmm_train_test.cpp
```diff
@@ -206,4 +301,104 @@ BOOST_AUTO_TEST_CASE(HMMTrainBatchModeTest)
...
+  SetInputParam("input_file", std::move(inputObsFile2));
+  // Provide a labels file with more states than initially specified
+  SetInputParam("labels_file", std::move(inputLabFile2));
+  SetInputParam("type", std::move(type));
+  SetInputParam("input_model", CLI::GetParam<HMMModel*>("output_model"));
+
+  CLI::GetSingleton().Parameters()["type"].wasPassed = false;
+  CLI::GetSingleton().Parameters()["states"].wasPassed = false;
+
+  HMMModel h2 = *(CLI::GetParam<HMMModel*>("output_model"));
```
rcurtin: I think you are forgetting a call to `mlpackMain()`. :)  It may also be worthwhile to add a comment pointing out that when the emission type is changed between runs, a warning will be printed but it's not a fatal error.  (I think that is true...)

rcurtin:
Thanks for adding the extra tests; they look good to me.  I left a few more comments and clarified some others; let me know if you have any questions.  Once the build is fixed (see the `approx_equal()` comment) then I think this is pretty close to done. :+1:

PlantsAndBuildings:
@rcurtin: You were right, backporting fn_approx_equal.hpp from the latest armadillo is not straightforward. I tried it (according to the commit you had referred to) and quickly ran into errors. I gave up on that approach and used the CheckMatrices approach - which worked with relatively few modifications to my existing code. As a result, now the travis build succeeds. However, I have realized that there are still a few issues with my tests:

1. The TODOs in the ApproximatelyEqual() function. Comparing emission distributions in case of discrete HMMs is straightforward -- we just have to compare matrices. However, comparing Gaussian distributions (in case of gmmHMMs and GaussianHMMs is, IMHO, not as straightforward). I can compare the parameters (Ie. means and variances of two distributions) -- but then this is not consistent with how I am comparing DiscreteDistributions. For a consistent approach I would have to compare probabilities at an infinite number of points (which is obviously not possible). How do you suggest I proceed in this regard? I think that the best way to handle this is to sample a number of points from a uniform random distribution and check if the probabilities (as determined by calls to the Probability() functions provided in the distributions) are within a certain tolerance of each other.

2. In HMMTrainNoLabelsReuseModelTest -- the second model uses the first one as a "guess" of what the emission/transition probabilities should be. As a result, after training the second model, **sometimes** the probabilites (transition/emission) differ significantly from the first one. Hence ApproximatelyEqual fails. I think this is not necessarily a problem with the test or the HMM code -- since both work as they should. I think if we fix the seed for the random number generator, then we will get reproducible test results everytime. I would appreciate any inputs on this.

rcurtin:
Hey there,

Thanks for the improvements.  Some responses to your points---

> 1. ...For a consistent approach I would have to compare probabilities at an infinite number of points (which is obviously not possible). How do you suggest I proceed in this regard?

The `GaussianDistribution` and `GMM` are fundamentally different in how they are stored than the `DiscreteDistribution` so I think it is just fine to compare the parameters of the `GaussianDistribution` and `GMM` with `CheckMatrices()`, i.e.

```
CheckMatrices(d1[i].Mean(), d2[i].Mean(), tolerance);
CheckMatrices(d1[i].Covariance(), d2[i].Covariance(), tolerance);
```

> 2. As a result, after training the second model, sometimes the probabilites (transition/emission) differ significantly from the first one.

You're right, although there might be one way around this; could you try and and let me know if it works?  If you set the binding parameter `tolerance` to something very large, the Baum-Welch algorithm should terminate after the first iteration.  Then you *should* still be able to use `CheckMatrices()` on the transition matrix (honestly I think it would be fine to just check that one), possibly with a large tolerance.  Let me know if this doesn't work, but I think that it should.

PlantsAndBuildings:
Setting the tolerance very high causes the test to fail more frequently :)

rcurtin:
I can't reproduce the failure.  When I run

```
$ bin/mlpack_test -t HMMTrainMainTest/HMMTrainNoLabelsReuseModelTest
```

no errors are detected, even when the random seed is changed.  (In this particular case the random seed is changed every time, until #1264 is merged, so no extra code needs to be added for that.)  Can you tell me more about the output you are getting?

You could also get a little more output about the convergence of the algorithm by configuring with `-DDEBUG=ON -DTEST_VERBOSE=ON`:

```
$ bin/mlpack_test -t HMMTrainMainTest/HMMTrainNoLabelsReuseModelTest
Running 1 test case...
[DEBUG] Iteration 0: log-likelihood -3.27262.
[DEBUG] Iteration 1: log-likelihood -2.46304.
[DEBUG] Iteration 2: log-likelihood -1.70623.
[DEBUG] Iteration 3: log-likelihood -1.23373.
[DEBUG] Iteration 4: log-likelihood -0.878869.
[DEBUG] Iteration 5: log-likelihood -0.365109.
[DEBUG] Iteration 6: log-likelihood -0.042845.
[DEBUG] Iteration 7: log-likelihood -0.00049849.
[DEBUG] Iteration 8: log-likelihood -7.50658e-08.
[DEBUG] Iteration 9: log-likelihood -2.22045e-15.
[DEBUG] Converged after 9 iterations.
[DEBUG] Iteration 0: log-likelihood 0.
[DEBUG] Converged after 0 iterations.

*** No errors detected
```

PlantsAndBuildings:
Here is the output I get. As discussed on IRC, I'm still looking into why this is happening.
```
root@cd2a9dcef76e:~/mlpack/build# bin/mlpack_test -t HMMTrainMainTest/HMMTrainNoLabelsReuseModelTest
Running 1 test case...
Seed: 1519498001
[DEBUG] Iteration 0: log-likelihood -3.88649.
[DEBUG] Iteration 1: log-likelihood -2.73553.
[DEBUG] Iteration 2: log-likelihood -1.99.
[DEBUG] Iteration 3: log-likelihood -1.46121.
[DEBUG] Iteration 4: log-likelihood -1.20286.
[DEBUG] Iteration 5: log-likelihood -0.856085.
[DEBUG] Iteration 6: log-likelihood -0.459666.
[DEBUG] Iteration 7: log-likelihood -0.128357.
[DEBUG] Iteration 8: log-likelihood -0.0087673.
[DEBUG] Iteration 9: log-likelihood -3.86013e-05.
[DEBUG] Iteration 10: log-likelihood -7.45044e-10.
[DEBUG] Iteration 11: log-likelihood 0.
[DEBUG] Converged after 11 iterations.
First trained model:
Initial probs:
   1.0000
        0
        0

Transition probs:
            0   1.0603e-38   1.9258e-38
  9.0954e-195   1.0000e+00   1.0000e+00
   1.0000e+00            0            0

Emission probs:
   1.0000e+00
            0
   1.9258e-38

            0
  9.0954e-195
   1.0000e+00

        0
   1.0000
        0

[DEBUG] Iteration 0: log-likelihood 0.
[DEBUG] Converged after 0 iterations.
Second trained model:
Initial probs:
   1.0000
        0
        0

Transition probs:
            0   3.3333e-01   3.7087e-76
            0   3.3333e-01   1.0000e+00
   1.0000e+00   3.3333e-01            0

Emission probs:
   1.0000e+00
            0
   3.7087e-76

        0
        0
   1.0000

        0
   1.0000
        0

/root/mlpack/src/mlpack/tests/test_tools.hpp(34): fatal error in "HMMTrainNoLabelsReuseModelTest": absolute value of b[i]{0.33333333333333331} exceeds 0.05000
0000000000003

```

PlantsAndBuildings:
Here is what I've come up with:

1. In the `HMMTrainNoLabelsReuseModelTest`, the Baum-Welch training function follows these steps:
- Initialize `oldLoglik` to zero.
- Compute `loglik` over training data and current set of parameters (Expectation step)
- Update parameters to maximize `loglik` (Maximization step)
- Check if absolute difference between `loglik` and `oldLoglik` is less than `tolerance`. If it is, then break out of the E-M loop. Else, update the value of `oldLoglik` to the current value of `loglik` and loop back to the Expectation step.

Note that if, at the very beginning of the algorithm, the set of parameters already have a value such that the log-likelihood over the given training data is 0 -- then we do NOT need to update the parameters (Ie. maximization step is not required). However the algorithm (in its present implementation) updates the parameters not considering this fact. This is exactly why the output I posted above gives an error. I propose the following solution:
- Initialize `oldLoglik` to zero
- Compute `loglik` over training data and current set of parameters (Expectation Step).
- Check if the `abs(loglik - oldLoglik)` is less than `tolerance`. If it is, then DO NOT UPDATE PARAMETERS and break out of the E-M loop. Else, proceed to the next step 
- Update parameters to maximize `loglik` (Maximization step); set `oldLoglik` to current value of `loglik` and loop back to the Expectation step.

(I've changed this ^ in hmm_impl.hpp)

2. In `mlpack/src/tests/test_tools.hpp`, `CheckMatrices(arma::mat& a, arma::mat& b, double tolerance)` does something like this:

https://github.com/mlpack/mlpack/blob/555557f45a76aea1a89301297d2c425adc983cde/src/mlpack/tests/test_tools.hpp#L31-L37

The `BOOST_REQUIRE_SMALL` macro requires that `b[i]` be less than the number `tolerance / 2`. However, `BOOST_REQUIRE_CLOSE` requires `a[i]` and `b[i]` be within `tolerance`% of each other. Is this intentional? Shouldn't we be consistent by either comparing only the absolute differences or relative differences? I have forced all comparisons to use the relative difference (Ie. `BOOST_REQUIRE_CLOSE` by multiplying all Transition/Emission/Initial matrices by 100 before calling `CheckMatrices`) 

IMHO, none of the tests should fail with these changes (I got zero failures in 1000 runs of all tests). Let me know if you think something is wrong -- I think I'm too close to this to find any mistakes. Also, I'm extremely sorry about the messed up commit history, I had some issues with one of my docker images and using git from the other one resulted in lots of chaos.

rcurtin:
Thank you for the detailed explanation.  I really appreciate that you have dug so deep in this, and I hope it has been an enjoyable experience to get so much depth (or, at least, however enjoyable debugging like this can be!).

1.  It sounds like, no matter what, the termination of the loop should be after the E step and not after the M step.  In the previous code, it sounded like the M step would update the parameters even after convergence!  So I agree it is the right thing to do to perform the tolerance check where you have added it.  I think we can remove the original tolerance check at the bottom of the loop though, and if you think it is easier to read we could also move the assignment of `oldLoglik` but that's up to you.

1b.  I'm still a bit surprised that I couldn't reproduce this even with lots of runs and different random seeds, but if you have tried this fix and it works for you, I still think it's the right thing to do to merge it.  I'll still test it on my system for a lot of runs before merge just in case.

2. The reason it's split up like that is that when `a[i]` or `b[i]` get extremely small, tiny differences become larger percentage differences.  So the way you can see the `CheckMatrices()` function is, make sure that the relative difference between elements is within `tolerance`%, but if `a[i]` is small, the relative calculation may be more likely to fail, so just make sure `b[i]` is small too.

Don't worry about the commit history.  If you want to rewrite the history to make it cleaner feel free, otherwise I'm happy to leave it as-is.

src/mlpack/methods/hmm/hmm_model.hpp
```diff
@@ -164,6 +164,31 @@ class HMMModel
...
+   * type         --> DiscreteHMM
+   * gaussianHMM  --> NULL
+   * gmmHMM       --> NULL
+   * discreteHMM  --> HMM<DiscreteDistribution> object
+   * and hence, calls to GMMHMM() and GaussianHMM() will return NULL. Only the
+   * call to DiscreteHMM() will return a non NULL pointer.
+   *
+   * Hence, in practice, a user should be careful to first check the type of HMM
+   * (by calling the Type() accessor) and then perform subsequent actions, to
+   * avoid running into NullPointerExceptions.
```
rcurtin: Hehe, this is not Java, we don't have NullPointerExceptions here. :)  You could just write 'null pointer dereferences'.

rcurtin:

PlantsAndBuildings:
@rcurtin I've made the changes you suggested. The style checks that are failing have been rectified in PR #1283 . I dont know why one of the travis jobs is failing though.

PlantsAndBuildings:
Fixed all checks -- should be good for merge now 👍

rcurtin:
Thanks again for the nice contribution. :)

PR #1238, opened 2018-02-11
Bilinear Interpolation Layer

ShikharJ:
Another rebase to master of the [original PR](https://github.com/mlpack/mlpack/pull/1098) by @kris-singh with the improved API and build errors removed. Required in #1204.


ShikharJ:
@zoq Can you help me with the `serialization` errors? I'm frankly out of ideas why these are cropping up.

lozhnikov:
@ShikharJ Looks like you've fixed the errors. I can't reproduce.

zoq:
@ShikharJ did you resolve the issues, if not can you poi point me to the error message?

ShikharJ:
@zoq I was referring to the failing `Appveyor` tests. I am receiving the same errors on my machine as well. Something like this:
```
C:\projects\mlpack\boost.1.60.0.0\lib\native\include\boost/archive/detail/check.hpp(162): error C2338: typex::value [C:\projects\mlpack\build\src\mlpack\tests\mlpack_test.vcxproj]
```

zoq:
Remove the const qualifier from:

```c++
  //! Locally stored row size of the input.
  const size_t inRowSize;
  //! Locally stored column size of the input.
  const size_t inColSize;
  //! Locally stored row size of the output.
  const size_t outRowSize;
  //! Locally stored column size of the input.
  const size_t outColSize;
  //! Locally stored depth of the input.
  const size_t depth;
```

ShikharJ:
@zoq Thanks, this is ready for a review now.

src/mlpack/methods/ann/layer/bilinear_interpolation.hpp
```diff
@@ -0,0 +1,124 @@
...
+#ifndef MLPACK_METHODS_ANN_LAYER_BILINEAR_INTERPOLATION_HPP
+#define MLPACK_METHODS_ANN_LAYER_BILINEAR_INTERPOLATION_HPP
+
+#include <mlpack/prereqs.hpp>
+
+namespace mlpack {
+namespace ann /** Artificial Neural Network. */ {
+
+/**
+ * Definition and implementation of the Bilinear Interpolation layer.
```
rcurtin: Can you describe a little bit more about what the bilinear interpolation layer is and what it does in this comment please?  Keep in mind that users will read this when they are trying to figure out how and where to use this layer.  (I know that the original comment here may not have been yours, but it would be great if you could improve it regardless!)

src/mlpack/methods/ann/layer/bilinear_interpolation_impl.hpp
```diff
@@ -0,0 +1,171 @@
...
+                     rOrigin;
+
+        output(k * outRowSize * outColSize + j * outColSize + i) =
+            input(ptr) * coeff1 +
+            input(ptr + 1) * coeff2 +
+            input(ptr + inColSize) * coeff3 +
+            input(ptr + inColSize + 1) * coeff4;
+      }
+    }
+  }
```
rcurtin: Is there any way that we could express this as a clearer or simpler Armadillo expression?  Here it looks like we are looping over each output element individually (and even accessing the output as a vectorized block of memory), but I think it could be a lot easier to understand what is going on if we can simplify the expression, and that may also allow the compiler to optimize.

One place to start might be with reinterpreting `output` as a different shape (should it be a cube?) so instead of accessing as `output(k * outRowSize * outColSize + j * outColSize + i)` you can do something more like `output(i, j, k)`.  You might consider doing that using an Armadillo alias: http://arma.sourceforge.net/docs.html#Cube (see the second advanced constructor).

Once you can do that, maybe you can put `coeff1` through `coeff4` in some vector `v` and write something a little more like this...

```
output(i, j, k) = input.slice(k).submat(rOrigin, cOrigin, rOrigin + 1, cOrigin + 1) % v;
```

That is just an idea although it is not perfect and would need a little bit of thinking.  Once you were able to do that, you might find that the entire triple for loop can be simplified and written as a single (or a few) Armadillo expressions.

src/mlpack/tests/ann_layer_test.cpp
```diff
@@ -1282,4 +1282,47 @@ BOOST_AUTO_TEST_CASE(SimpleMeanSquaredErrorLayerTest)
...
+                                depth);
+
+  layer.Forward(std::move(input), std::move(output));
+  CheckMatrices(output - expectedOutput, arma::zeros(output.n_rows), 1e-12);
+
+
+  layer.Backward(std::move(output), std::move(output),
+        std::move(unzoomedOutput));
+  CheckMatrices(unzoomedOutput - input, arma::zeros(input.n_rows), 1e-12);
+}
```
rcurtin: Do you think you could add a numerical gradient test like, for instance, `GradientLSTMLayerTest`?  These use the `AddGradient()` function, and it should be simple to adapt the existing tests to this layer type.

src/mlpack/methods/ann/layer/bilinear_interpolation_impl.hpp
ShikharJ: @rcurtin I think I can fixate the output on being a cube, but I think that way we'll have to lose the `OutputDataType` parameter. It is not a hard requirement anywhere as of now. Its needed for the O'Reilly test in the GAN PR.

rcurtin: No need to change the output layer type, you can just make an alias like this, for instance:

```
arma::cube outputAsCube(output.memptr(), n_rows, n_cols, n_slices, false, true);
```

ShikharJ: @rcurtin Having given it some thought, I'm still a bit skeptical on how would this be achievable. I think, the triple `for` loop will have to be present, since the coefficients are calculated for all `i` and `j` and have to be applied on `k` slices as well. I came up with something like this (not sure of code correctness though):
```
if (output.is_empty())
    output.set_size(outRowSize * outColSize * depth, 1);

  assert(output.n_rows == outRowSize * outColSize * depth);
  assert(output.n_cols == 1);

  arma::cube InputAsCube(input.memptr(), inRowSize, inColSize, depth, false, true);
  arma::cube OutputAsCube(output.memptr(), outRowSize, outColSize, depth, false, true);

  double scaleRow = (double) inRowSize / (double) outRowSize;
  double scaleCol = (double) inColSize / (double) outColSize;

  arma::vec coeffs(4);
  for (size_t i = 0; i < outRowSize; i++)
  {
    for (size_t j = 0; j < outColSize; j++)
    {
      size_t rOrigin = (size_t) std::floor(i * scaleRow);
      size_t cOrigin = (size_t) std::floor(j * scaleCol);

      if (rOrigin > inRowSize - 2)
        rOrigin = inRowSize - 2;
      if (cOrigin > inColSize - 2)
        cOrigin = inColSize - 2;

      double deltaR = i * scaleRow - rOrigin;
      double deltaC = j * scaleCol - cOrigin;
      coeffs[0] = (1 - deltaR) * (1 - deltaC);
      coeffs[1] = deltaR * (1 - deltaC);
      coeffs[2] = (1 - deltaR) * deltaC;
      coeffs[3] = deltaR * deltaC;

      for (size_t k = 0; k < depth; k++)
      {
        OutputAsCube(i, j, k) = arma::accu(InputAsCube.slice(k).submat(
                rOrigin, cOrigin, rOrigin + 1,
                cOrigin + 1) % coeffs);
      }
    }
  }
```

rcurtin: I definitely think that's an improvement.  You might be able to remove that last for loop (the one indexed by `k`) by using `.tube()` to access across all `depth` slices at once.  You might need to use `repmat()` on `coeffs` to make that work.

In any case, assuming that the result of the computation is the same as before, I think it is better and we should go with the newer version that you've made.  Thanks!

ShikharJ: I don't think that `repmat()` would be of much use either, since it will only return a matrix, while doing something like this (again not sure of the implementation):
```
 OutputAsCube.tube(i, j) = arma::accu(InputAsCube.tube(rOrigin, cOrigin, rOrigin + 1, cOrigin + 1) % repmat(coeffs, 1, depth));
```
we will run into a `sub_cube` on the input, and a `mat` on the coeffs (correct me if I'm wrong). But element wise multiplication wouldn't work here. What we need ideally is a `cube` version of the coeffs (I think), then the above should work.

ShikharJ: @rcurtin Also, doing above, I'll have to drop the `arma::accu`, and then do a `column-wise` and then `row-wise` addition to the element wise product matrix obtained. All in all, I don't think there'd be much to gain from making those changes. Should I just go with this implementation?

src/mlpack/tests/ann_layer_test.cpp
zoq: Yes, the ``CheckGradient`` function: https://github.com/mlpack/mlpack/blob/346fd1f819122cdc6b5cf148291dcec05e1dcbce/src/mlpack/tests/ann_layer_test.cpp#L146-L178

rcurtin: Oops, you are right, sorry for my error.  I did indeed mean `CheckGradient()`.  It's still a useful test, even if the gradient is the same as the propagated error.

src/mlpack/tests/ann_layer_test.cpp
```diff
@@ -1282,4 +1282,89 @@ BOOST_AUTO_TEST_CASE(SimpleMeanSquaredErrorLayerTest)
...
+    }
+
+    arma::mat& Parameters() { return model->Parameters(); }
+
+    FFN<NegativeLogLikelihood<>, NguyenWidrowInitialization>* model;
+    arma::mat input, target;
+  } function;
+
+  BOOST_REQUIRE_LE(CheckGradient(function), 1e-4);
+}
```
ShikharJ: @zoq I'm not confident of this numerical test. Kris' original tests work fine, so probably the implementation is correct. A little help?

ShikharJ: Ping @zoq.

zoq: Did you check with the original code? Perhaps we missed something? I guess, we have to go through each line and make sure it works as expected. Might be helpful to check the output of the gradient and backward method.

ShikharJ: @zoq I checked the core algorithm code a number of times, and found no problems, the implementation is most likely correct. The issue is with this test case, where for some reason only the first column of the input image is being passed to the `BilinearInterpolation` Layer (possibly because of the model chosen?). Can you help me in correcting this?

src/mlpack/tests/ann_layer_test.cpp
```diff
@@ -170,6 +174,18 @@ double CheckGradient(FunctionType& function, const double eps = 1e-7)
 
     // Compute numerical gradients using the costs calculated above.
     estGradient(i) = (costPlus - costMinus) / (2 * eps);
+    if (k){
+      std::cout << i + 1 << "th Iteration:" << std::endl;
+      std::cout << "Weight: " << tmp << std::endl;
+      std::cout << "CostPlus: " << costPlus << std::endl;
+      std::cout << "CostMinus: " << costMinus << std::endl;
+      std::cout << "Established Gradient: " << estGradient(i) << std::endl << std::endl;
```
ShikharJ: @zoq I ran these changes over the iterations for the `BilinearInterpolation` Layer. It turns out that `costPlus` and `costMinus` values return the same `double` value, and  hence the `estGradient` becomes `0` for each output point.

zoq: I see, since the gradient is always one , I think there is no need to add the test here.

ShikharJ:
Thanks for your help @zoq. This is updated now.

ShikharJ:
@rcurtin @lozhnikov  Can I have a review of this please?

rcurtin:
Hi @ShikharJ, please be patient.  I will review this when I have a chance.  Just to give you an idea, I received 100 emails that require my attention this weekend---and that is over the weekend.  So I have quite a backlog, and I would expect similar situations to also be true for the other maintainers.  Everyone will review this when they are able to.

src/mlpack/methods/ann/layer/bilinear_interpolation.hpp
```diff
@@ -0,0 +1,159 @@
...
+ * Definition and Implementation of the Bilinear Interpolation Layer.
+ *
+ * Bilinear Interpolation is an mathematical technique, primarily used for
+ * image scaling purposes. It is an extension of linear interpolation, for
+ * interpolating functions of two variables on the image grid. The key idea
+ * is to perform linear interpolation first in one direction (e.g., along
+ * x-axis), and then again in the other direction (i.e., y-axis), on four
+ * different known pixel points in a rectangular grid. In this way, we
+ * represent any arbitrary pixel, present within the rectangular grid,
+ * as a function of those four pixels.
```
lozhnikov: I suggest to remove the paragraph above since anyone knows what Bilinear Interpolation is.

src/mlpack/methods/ann/layer/bilinear_interpolation.hpp
```diff
@@ -0,0 +1,159 @@
...
+  BilinearInterpolation();
+
+  /**
+   * The constructor for the Bilinear Interpolation.
+   *
+   * @param inRowSize: Number of input rows.
+   * @param inColSize: Number of input columns.
+   * @param outRowSize: Number of output rows.
+   * @param outColSize: Number of output columns.
+   * @param depth: Number of slices of input.
```
lozhnikov: The colons aren't needed here.

src/mlpack/methods/ann/layer/bilinear_interpolation_impl.hpp
```diff
@@ -0,0 +1,185 @@
...
+  weights.set_size(outRowSize * outColSize * depth, 1);
+}
+
+template<typename InputDataType, typename OutputDataType>
+template<typename eT>
+void BilinearInterpolation<InputDataType, OutputDataType>::Forward(
+    const arma::Mat<eT>&& input, arma::Mat<eT>&& output)
+{
+  if (output.is_empty())
+    output = weights;
```
lozhnikov: Hmm... I thought the layer hasn't got its own `weights`. What is the purpose of the variable?

src/mlpack/methods/ann/layer/bilinear_interpolation_impl.hpp
```diff
@@ -0,0 +1,185 @@
...
+
+  assert(input.n_rows == inRowSize * inColSize * depth);
+  assert(input.n_cols == 1);
+  assert(output.n_rows == outRowSize * outColSize * depth);
+  assert(output.n_cols == 1);
+  assert(inRowSize >= 2);
+  assert(inColSize >= 2);
+
+  arma::cube InputAsCube(input.memptr(), inRowSize, inColSize, depth);
+  arma::cube OutputAsCube(output.memptr(), outRowSize, outColSize, depth, false, true);
```
lozhnikov: According to the styleguide variables should begin with lowercase letters.

src/mlpack/methods/ann/layer/bilinear_interpolation.hpp
```diff
@@ -0,0 +1,159 @@
...
+  //! Locally-stored weight object.
+  OutputDataType weights;
+  //! Locally-stored gradient object.
+  OutputDataType gradient;
+  //! Locally-stored delta object.
+  OutputDataType delta;
+  //! Locally-stored input parameter object.
+  InputDataType inputParameter;
+  //! Locally-stored output parameter object.
+  OutputDataType outputParameter;
```
lozhnikov: Correct if I'm mistaken, but looks like the three variables above aren't used.

src/mlpack/methods/ann/layer/bilinear_interpolation.hpp
ShikharJ: Please refer [here](https://github.com/mlpack/mlpack/pull/1238#discussion_r167863833).

src/mlpack/methods/ann/layer/bilinear_interpolation.hpp
ShikharJ: You're correct, actually these came up while I was trying to add `CheckGradient()` function to the tests, however, that is not needed and I'll remove them.

src/mlpack/methods/ann/layer/bilinear_interpolation_impl.hpp
ShikharJ: Same as [here](https://github.com/mlpack/mlpack/pull/1238/files#r169643797).

src/mlpack/tests/ann_layer_test.cpp
rcurtin: My only remaining comment is that I'd still like to see a test for the gradient.  Even though the layer has no Gradient() method, that means the gradient should implicitly be all ones.  So we can adapt the function @zoq mentioned above and check against that.  In my view this is still an important check to add.

ShikharJ: @rcurtin I did add a test like that, but the issue with that test was that `costPlus` and `costMinus` values returned the same `double` value, and hence the `estGradient` became `0` for each output point. Hence @zoq's comment [here](https://github.com/mlpack/mlpack/pull/1238#discussion_r169102607).

rcurtin: Thanks for the link.  Github had hidden those comments.  In that case there are no more comments on this PR from me.  Thank you for working on it!

src/mlpack/methods/ann/layer/bilinear_interpolation_impl.hpp
```diff
@@ -0,0 +1,173 @@
...
+
+  double scaleRow = (double) inRowSize / (double) outRowSize;
+  double scaleCol = (double) inColSize / (double) outColSize;
+
+  arma::mat22 coeffs;
+  for (size_t i = 0; i < outRowSize; i++)
+  {
+    for (size_t j = 0; j < outColSize; j++)
+    {
+      size_t rOrigin = (size_t) std::floor(i * scaleRow);
```
zoq: Looks like we could move this one loop up.

src/mlpack/methods/ann/layer/bilinear_interpolation_impl.hpp
```diff
@@ -0,0 +1,173 @@
...
+      size_t cOrigin = (size_t) std::floor(j * scaleCol);
+
+      if (rOrigin > inRowSize - 2)
```
zoq: See my comment on the ``rOrigin`` parameter.

src/mlpack/methods/ann/layer/bilinear_interpolation_impl.hpp
```diff
@@ -0,0 +1,173 @@
...
+        rOrigin = inRowSize - 2;
+      if (cOrigin > inColSize - 2)
+        cOrigin = inColSize - 2;
+
+      double deltaR = i * scaleRow - rOrigin;
```
zoq: See my comment on the ``rOrigin`` parameter.

src/mlpack/methods/ann/layer/bilinear_interpolation_impl.hpp
```diff
@@ -0,0 +1,173 @@
...
+      double deltaC = j * scaleCol - cOrigin;
+      coeffs[0] = (1 - deltaR) * (1 - deltaC);
```
zoq: Not sure this is acutely faster if we also move this one loop up and just multiply by ``(1 - deltaC)`` here.

src/mlpack/methods/ann/layer/bilinear_interpolation_impl.hpp
```diff
@@ -0,0 +1,173 @@
...
+  arma::cube inputAsCube(input.memptr(), inRowSize, inColSize, depth);
+  arma::cube outputAsCube(output.memptr(), outRowSize, outColSize, depth, false, true);
+
+  double scaleRow = (double) inRowSize / (double) outRowSize;
+  double scaleCol = (double) inColSize / (double) outColSize;
+
+  arma::mat22 coeffs;
+  for (size_t i = 0; i < outRowSize; i++)
+  {
+    for (size_t j = 0; j < outColSize; j++)
```
zoq: I guess, this could be optimized if we would use an SIMD loop here, as done for the InformationGain caluclation here https://github.com/mlpack/mlpack/blob/88a7efe9fa87958190e6dc56dc86ec4d5ed3153a/src/mlpack/methods/decision_tree/information_gain.hpp. I don't think we have to do that in this context just a thought.

src/mlpack/methods/ann/layer/bilinear_interpolation_impl.hpp
```diff
@@ -0,0 +1,173 @@
...
+      {
+        size_t rOrigin = (size_t) std::floor(i * scaleRow);
+        size_t cOrigin = (size_t) std::floor(j * scaleCol);
+
+        if (rOrigin > outRowSize - 2)
+          rOrigin = outRowSize - 2;
+        if (cOrigin > outColSize - 2)
+          cOrigin = outColSize - 2;
+
+        double deltaR = i * scaleRow - rOrigin;
```
zoq: Looks like we could apply the same optimizations as done for the forward pass here.

src/mlpack/tests/ann_layer_test.cpp
```diff
@@ -1282,4 +1282,47 @@ BOOST_AUTO_TEST_CASE(SimpleMeanSquaredErrorLayerTest)
...
+        expectedOutput[ptr] = 1 * i / (double)outRowSize +
+            1 * j / (double)outColSize + 1 * d / (double)depth;
+      }
+
+  BilinearInterpolation<> layer(inRowSize, inColSize, outRowSize, outColSize,
+                                depth);
+
+  layer.Forward(std::move(input), std::move(output));
+  CheckMatrices(output - expectedOutput, arma::zeros(output.n_rows), 1e-12);
+
```
zoq: Pedantic style issue, remove the extra line here.

src/mlpack/tests/ann_layer_test.cpp
```diff
@@ -1282,4 +1282,47 @@ BOOST_AUTO_TEST_CASE(SimpleMeanSquaredErrorLayerTest)
   BOOST_REQUIRE_EQUAL(output.n_elem, 1);
 }
 
+/*
+ * Simple test for the BilinearInterpolation layer
+ */
+BOOST_AUTO_TEST_CASE(SimpleBilinearInterpolationLayerTest)
```
zoq: Another idea would be to compare against armadillo's interpolation method on an 1D vector.

src/mlpack/methods/ann/layer/bilinear_interpolation.hpp
lozhnikov: Ah, I see. I just don't like the comment since it is focused on images only. But Bilinear Interpolation is a much more general technique. Could you replace `images` by `functions` and `pixels` by `points`?

src/mlpack/methods/ann/layer/bilinear_interpolation_impl.hpp
ShikharJ: `(1 - deltaC)` would change at every iteration of `j`. Not sure if this would be a good idea, as I'd also have to divide by the same at the end of the each iteration of `j`. Only other way I can think of breaking this up is to store the `deltaR`'s and `deltaC`'s in different vectors and letting element wise multiplication do the job here, but that won't be faster than this.

zoq: Right my idea was, to set ``coeffs[0] = (1 - deltaR)`` inside the first loop and ``do coeffs[0] *= (1 - deltaC);`` here. Maybe I missed something?

ShikharJ: `deltaC` is defined as:
```
double deltaC = j * scaleCol - cOrigin;
```
which changes at every `j = 0, 1, 2...`. Now doing `coeffs[0] *= (1 - deltaC)` inside every `j`th iteration, would accumulate all the `(1 - deltaC)` values over the iterations, which can only be corrected if at the end of every `j`th iteration, I do `coeffs[0] /= (1 - deltaC)`, to get back `coeffs[0] = (1 - deltaR)`, as in the first loop.

zoq: Ahh, your are right, thanks for the clarification.

src/mlpack/tests/ann_layer_test.cpp
ShikharJ: I'll add a small example soon.

zoq: Great, thanks!

ShikharJ: @zoq I checked for `arma::interp1()`, and it seems that it doesn't support matrix types. So basically, what I'll have to do is create a number of row vectors and first interpolate them along one direction (along row), and then create another set of vectors from the column entries of these row vectors and interpolate them along the columns. Only then we'd be able to compare the methods. This would be tedious for a good test (2x2 to 3x3 matrix is doable, but any higher would just be time consuming). How big do we want the test to be?

src/mlpack/tests/ann_layer_test.cpp
```diff
@@ -1282,4 +1282,97 @@ BOOST_AUTO_TEST_CASE(SimpleMeanSquaredErrorLayerTest)
...
+  expectedOutput[2] = aa[2];
+  expectedOutput[3] = bb[0];
+  expectedOutput[4] = bb[1];
+  expectedOutput[5] = bb[2];
+  expectedOutput[6] = cc[0];
+  expectedOutput[7] = cc[1];
+  expectedOutput[8] = cc[2];
+  //std::cout << std::endl << "expectedOutput" << std::endl << expectedOutput << std::endl;
+  CheckMatrices(output - expectedOutput, arma::zeros(output.n_rows), 1e-12);
+  //std::cout << std::endl << "expectedOutput" << std::endl << expectedOutput << std::endl;
```
ShikharJ: @zoq I could only come up with this sort of hacky version, seemingly `arma::interp1` doesn't work the way bilinear-interpolation does.

src/mlpack/tests/ann_layer_test.cpp
zoq: I would say that 2x2 or 3x3 is just fine.

src/mlpack/tests/ann_layer_test.cpp
zoq: I see, do you think we should instead compare against a precomputed result?

ShikharJ: A precomputed result should be better. Probably from `Tensorflow` or `Skimage`? But we already have a test-case that tests that above.

ShikharJ: @zoq If you could suggest a source, it'd be great. I'd drop these changes then.

zoq: "But we already have a test-case that tests that above." Do you mean the three for loops?

zoq: I think comparing against skimage does make sense, this is a really simple test.

ShikharJ: Yes, those were a part of Kris' original test.

ShikharJ: I'll see what can be done.

zoq: Would be great if you could compare it a precomputed result, skimage or tensorflow is just fine for me.

src/mlpack/methods/ann/layer/bilinear_interpolation_impl.hpp
```diff
@@ -0,0 +1,173 @@
...
+
+template<typename InputDataType, typename OutputDataType>
+template<typename eT>
+void BilinearInterpolation<InputDataType, OutputDataType>::Forward(
+    const arma::Mat<eT>&& input, arma::Mat<eT>&& output)
+{
+  if (output.is_empty())
+    output.set_size(outRowSize * outColSize * depth, 1);
+
+  assert(input.n_rows == inRowSize * inColSize * depth);
```
zoq: Not sure, about the runtime parameter check here, usally we do not check the input values, however there is an open issue https://github.com/mlpack/mlpack/issues/891 that is related. I guess we could remove the check now and once we agreed on something put something in place. Let me know what you think.

src/mlpack/methods/ann/layer/bilinear_interpolation_impl.hpp
```diff
@@ -0,0 +1,173 @@
...
+      rOrigin = inRowSize - 2;
+    double deltaR = i * scaleRow - rOrigin;
+    for (size_t j = 0; j < outColSize; j++)
+    {
+      size_t cOrigin = (size_t) std::floor(j * scaleCol);
+
+      if (cOrigin > inColSize - 2)
+        cOrigin = inColSize - 2;
+
+      double deltaC = j * scaleCol - cOrigin;
```
zoq: Don't feel obligated, but I think adding some comments here would improve the comprehensibility.

src/mlpack/methods/ann/layer/bilinear_interpolation_impl.hpp
```diff
@@ -0,0 +1,173 @@
...
+        outputAsCube(i, j, k) = arma::accu(inputAsCube.slice(k).submat(
+                rOrigin, cOrigin, rOrigin + 1, cOrigin + 1) % coeffs);
+      }
+    }
+  }
+}
+
+template<typename InputDataType, typename OutputDataType>
+template<typename eT>
+void BilinearInterpolation<InputDataType, OutputDataType>::Backward(
```
zoq: Do you think it make sense, to write a function for the interpolation and just pass the correct parameter, since the forward and backward pass is almost the same?

ShikharJ: It is doable in a function, but we'd have to specify whether we want a forward or a backward pass, and introduce a number of `if-else` statements. Also, a number of variables have the same name, but are not computed the same way in both the function (see `scaleRow` and `scaleCol` in both the functions). I personally prefer the current implementation, but I'd say its your call.

zoq: Okay, let's go with the current implementation.

src/mlpack/tests/ann_layer_test.cpp
ShikharJ: This turned out to be a good idea @zoq. Found a bug in Kris' implementation logic that was going unnoticed.

ShikharJ:
This is ready for a final review @zoq.

src/mlpack/methods/ann/layer/bilinear_interpolation_impl.hpp
```diff
@@ -0,0 +1,185 @@
...
+  else
+  {
+    assert(output.n_rows == outRowSize * outColSize * depth);
+    assert(output.n_cols == 1);
+  }
+
+  assert(inRowSize >= 2);
+  assert(inColSize >= 2);
+
+  arma::cube inputAsCube(input.memptr(), inRowSize, inColSize, depth);
```
rcurtin: If `false` isn't specified as the fifth argument, this will by default copy all of the data in the cube.  Are you sure this is what you want to do here?

src/mlpack/methods/ann/layer/bilinear_interpolation_impl.hpp
```diff
@@ -0,0 +1,185 @@
...
+  else
+  {
+    assert(output.n_rows == inRowSize * inColSize * depth);
+    assert(output.n_cols == 1);
+  }
+
+  assert(outRowSize >= 2);
+  assert(outColSize >= 2);
+
+  arma::cube gradientAsCube(gradient.memptr(), outRowSize, outColSize, depth);
```
rcurtin: The same comment here applies about copying.  (You may have already realized it applies to both places but I want to explicitly point it out so we don't forget one.)

src/mlpack/methods/ann/layer/bilinear_interpolation_impl.hpp
ShikharJ: This is actually using `cube(const ptr_aux_mem, n_rows, n_cols, n_slices)` instead of `cube(ptr_aux_mem, n_rows, n_cols, n_slices, copy_aux_mem = true, strict = false)` as `input` is of `const` type.

rcurtin: Ah right, `input` is const, now it makes sense.  Thanks for the clarification.

ShikharJ:
@zoq @rcurtin Is there anything remaining to be done here? I'd like to get this merged for the O'Reilly GAN test.

src/mlpack/methods/ann/layer/bilinear_interpolation.hpp
```diff
@@ -0,0 +1,132 @@
...
+  /**
+   * The constructor for the Bilinear Interpolation.
+   *
+   * @param inRowSize Number of input rows.
+   * @param inColSize Number of input columns.
+   * @param outRowSize Number of output rows.
+   * @param outColSize Number of output columns.
+   * @param depth Number of slices of input.
+   */
+  BilinearInterpolation(
```
zoq: Do you mind to use:

```c++
BilinearInterpolation(const size_t inRowSize,
                      const size_t inColSize,
```
here to be consistent with the rest of the codebase?

src/mlpack/methods/ann/layer/bilinear_interpolation.hpp
```diff
@@ -0,0 +1,132 @@
...
+  BilinearInterpolation();
+
+  /**
+   * The constructor for the Bilinear Interpolation.
+   *
+   * @param inRowSize Number of input rows.
+   * @param inColSize Number of input columns.
+   * @param outRowSize Number of output rows.
+   * @param outColSize Number of output columns.
+   * @param depth Number of slices of input.
```
zoq: Should we go with number of input slices? Also, what do you think should we remove the default depth = 1, people might miss to set the parameter.

src/mlpack/methods/ann/layer/bilinear_interpolation.hpp
```diff
@@ -0,0 +1,132 @@
...
+   * @param output The resulting interpolated output matrix.
+   */
+  template<typename eT>
+  void Forward(const arma::Mat<eT>&& input, arma::Mat<eT>&& output);
+
+  /**
+   * Ordinary feed backward pass of a neural network, calculating the function
+   * f(x) by propagating x backwards through f. Using the results from the feed
+   * forward pass. Since the layer does not have any learn-able parameters,
+   * we just have down-sample the gradient to make its size compatible with
```
zoq: Should we use ``we just have to down-sample ...`` here?

src/mlpack/methods/ann/layer/bilinear_interpolation_impl.hpp
```diff
@@ -0,0 +1,185 @@
...
+        deltaC = 1.0;
+      coeffs[0] = (1 - deltaR) * (1 - deltaC);
+      coeffs[1] = deltaR * (1 - deltaC);
+      coeffs[2] = (1 - deltaR) * deltaC;
+      coeffs[3] = deltaR * deltaC;
+
+      for (size_t k = 0; k < depth; k++)
+      {
+        outputAsCube(i, j, k) = arma::accu(inputAsCube.slice(k).submat(
+                rOrigin, cOrigin, rOrigin + 1, cOrigin + 1) % coeffs);
```
zoq: This should be tabbed twice; for more information take a look at: https://github.com/mlpack/mlpack/wiki/DesignGuidelines#line-length-and-wrapping.

src/mlpack/tests/ann_layer_test.cpp
```diff
@@ -1282,4 +1282,39 @@ BOOST_AUTO_TEST_CASE(SimpleMeanSquaredErrorLayerTest)
...
+  expectedOutput.reshape(25, 1);
+  layer.Forward(std::move(input), std::move(output));
+  CheckMatrices(output - expectedOutput, arma::zeros(output.n_rows), 1e-12);
+
+  expectedOutput = arma::mat("1.0000 1.9000 1.9000 2.8000");
+  expectedOutput.reshape(4, 1);
+  layer.Backward(std::move(output), std::move(output),
+        std::move(unzoomedOutput));
+  CheckMatrices(unzoomedOutput - expectedOutput, arma::zeros(input.n_rows),
+                1e-12);
```
zoq: This should be tabbed twice, see comment above, perhaps

```
CheckMatrices(unzoomedOutput - expectedOutput,
    arma::zeros(input.n_rows), 1e-12);
```

is more readable?

zoq:
With two approvals I will go ahead and merge this now. Thanks again!

PR #1239, opened 2018-02-12
Removing empty clusters instead of killing them.

### LINKED ISSUE #1232: Documentation needs to be clarified
In the [allow_empty_clusters section of documentation](http://www.mlpack.org/docs/mlpack-2.2.5/doxygen/kmtutorial.html) it is written that 

> when you save your clusters, some of the clusters may be filled with NaNs. This is expected behavior – if a cluster has no points, the concept of a centroid makes no sense.

whereas in [EmptyCluster section of Documentation](http://www.mlpack.org/docs/mlpack-2.2.5/doxygen/classmlpack_1_1kmeans_1_1AllowEmptyClusters.html) it is stated that

> This function allows empty clusters to persist simply by leaving the empty cluster in its last position.

As `allow_empty_clusters` is the default constructor used for this policy of `EmptyCluster()` then I think we will not get any Nans when we save our clusters. I've experimented with several data sets and various values of clusters and I didn't get Nan even a single time. 

Moreover, it would be great if we could add a section of `kill_empty_clusters` to [command line kmeans](http://www.mlpack.org/docs/mlpack-2.2.5/doxygen/kmtutorial.html) describing how it will fill the results with `DBL_MAX` in case of empty clusters

Prabhat-IIT:
Fixes #1232 
While clustering instead of filling empty clusters with `DBL_MAX` empty clusters are entirely removed.

### DIFF
--- a/doc/tutorials/kmeans/kmeans.txt
+++ b/doc/tutorials/kmeans/kmeans.txt
@@ -132,9 +132,16 @@ $ mlpack_kmeans -c 5 -i dataset.csv -v -o assignments.csv -C centroids.csv
 
 If you would like to allow empty clusters to exist, instead of reinitializing
 them, simply specify the \c -e (\c --allow_empty_clusters) option.  Note that
-when you save your clusters, some of the clusters may be filled with NaNs.  This
-is expected behavior -- if a cluster has no points, the concept of a centroid
-makes no sense.
+when you save your clusters, even empty clusters will still have centroids.
+The centroids of the empty cluster will be the same as what they were on the
+last iteration when the cluster was not empty.
+
+@subsection cli_ex3_kmtut Killing empty clusters
+
+If you would like to kill empty clusters , instead of reinitializing
+them, simply specify the \c -E (\c --kill_empty_clusters) option.  Note that
+when you save your clusters, all the empty clusters will be removed and the
+final result may contain less than specified number of clusters.
 
 @code
 $ mlpack_kmeans -c 5 -i dataset.csv -v -o assignments.csv -C centroids.csv
--- a/doc/tutorials/kmeans/kmeans.txt
+++ b/doc/tutorials/kmeans/kmeans.txt
@@ -136,6 +136,10 @@ when you save your clusters, even empty clusters will still have centroids.
 The centroids of the empty cluster will be the same as what they were on the
 last iteration when the cluster was not empty.
 
+@code
+$ mlpack_kmeans -c 5 -i dataset.csv -v -e -o assignments.csv -C centroids.csv
+@endcode
+
 @subsection cli_ex3_kmtut Killing empty clusters
 
 If you would like to kill empty clusters , instead of reinitializing
@@ -144,7 +148,7 @@ when you save your clusters, all the empty clusters will be removed and the
 final result may contain less than specified number of clusters.
 
 @code
-$ mlpack_kmeans -c 5 -i dataset.csv -v -o assignments.csv -C centroids.csv
+$ mlpack_kmeans -c 5 -i dataset.csv -v -E -o assignments.csv -C centroids.csv
 @endcode
 
 @subsection cli_ex4_kmtut Limiting the maximum number of iterations
--- a/src/mlpack/methods/kmeans/kill_empty_clusters.hpp
+++ b/src/mlpack/methods/kmeans/kill_empty_clusters.hpp
@@ -51,18 +51,17 @@ class KillEmptyClusters
       const size_t emptyCluster,
       const arma::mat& /* oldCentroids */,
       arma::mat& newCentroids,
-      arma::Col<size_t>& counts /* clusterCounts */,
+      arma::Col<size_t>& clusterCounts,
       MetricType& /* metric */,
       const size_t /* iteration */)
   {
     // Remove the empty cluster.
-      if (emptyCluster < newCentroids.n_cols)
+    if (emptyCluster < newCentroids.n_cols)
       {
           newCentroids.shed_col(emptyCluster);
-          counts.shed_row(emptyCluster);
-          return 0; // one cluster removed
+          clusterCounts.shed_row(emptyCluster);
       }
-      else return 0; // no point changed
+    return 0; // cluster removed
   }
 
   //! Serialize the empty cluster policy (nothing to do).
--- a/src/mlpack/methods/kmeans/kill_empty_clusters.hpp
+++ b/src/mlpack/methods/kmeans/kill_empty_clusters.hpp
@@ -51,7 +51,7 @@ class KillEmptyClusters
       const size_t emptyCluster,
       const arma::mat& /* oldCentroids */,
       arma::mat& newCentroids,
-      arma::Col<size_t>& /* clusterCounts */,
+      arma::Col<size_t>& counts /* clusterCounts */,
       MetricType& /* metric */,
       const size_t /* iteration */)
   {
@@ -59,9 +59,10 @@ class KillEmptyClusters
       if (emptyCluster < newCentroids.n_cols)
       {
           newCentroids.shed_col(emptyCluster);
-          return 0; // No points were changed.
+          counts.shed_row(emptyCluster);
+          return 0; // one cluster removed
       }
-      else return 0;
+      else return 0; // no point changed
   }
 
   //! Serialize the empty cluster policy (nothing to do).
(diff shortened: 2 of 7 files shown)


Prabhat-IIT:
@rcurtin I would like to know that how can I reproduce the Doxygen Documentation locally? Because I would like to test the changes made to documentation locally.

doc/tutorials/kmeans/kmeans.txt
```diff
@@ -132,9 +132,16 @@ $ mlpack_kmeans -c 5 -i dataset.csv -v -o assignments.csv -C centroids.csv
...
+when you save your clusters, even empty clusters will still have centroids.
+The centroids of the empty cluster will be the same as what they were on the
+last iteration when the cluster was not empty.
+
+@subsection cli_ex3_kmtut Killing empty clusters
+
+If you would like to kill empty clusters , instead of reinitializing
+them, simply specify the \c -E (\c --kill_empty_clusters) option.  Note that
+when you save your clusters, all the empty clusters will be removed and the
+final result may contain less than specified number of clusters.
```
rcurtin: Looks good, but the code example below this doesn't have the `-E` option.  Also we should add an example to the previous subsection with the `-e` option (it can be the same command).

src/mlpack/methods/kmeans/kill_empty_clusters.hpp
```diff
@@ -55,9 +55,12 @@ class KillEmptyClusters
...
       const size_t /* iteration */)
   {
-    // Kill the empty cluster.
-    newCentroids.col(emptyCluster).fill(DBL_MAX);
-    return 0; // No points were changed.
+    // Remove the empty cluster.
+      if (emptyCluster < newCentroids.n_cols)
+      {
+          newCentroids.shed_col(emptyCluster);
+          return 0; // No points were changed.
```
rcurtin: Don't we have to update `counts` also?

Prabhat-IIT: I don't think we need to update the `count` as `count` represent the number of points in a particular cluster and dead cluster will not have any points in it. If we really want to modify the `counts` then we have to delete `count[i]` ( here `i` is the index of dead cluster) and then we also have to modify the `cluster` i.e. `cluster-=1` . I can do whichever way you want me to do :)

doc/tutorials/kmeans/kmeans.txt
Prabhat-IIT: Done :)

src/mlpack/methods/kmeans/kill_empty_clusters.hpp
rcurtin: I think we do have to update `counts`, but if you write a test we can know for sure if it is needed.

Prabhat-IIT: I've updated `counts`  as well as written the test :)

src/mlpack/methods/kmeans/kmeans_impl.hpp
```diff
@@ -228,11 +229,17 @@ Cluster(const MatType& data,
       {
         Log::Info << "Cluster " << i << " is empty.\n";
         if (iteration % 2 == 0)
-          emptyClusterAction.EmptyCluster(data, i, centroids, centroidsOther,
+        {
+          numberOfRemovedClusters = emptyClusterAction.EmptyCluster(data, i, centroids, centroidsOther,
               counts, metric, iteration);
+          clusters = clusters - numberOfRemovedClusters;
```
rcurtin: Hmm, I think we can actually simplify this code a bit.  If we change the loop

```
for (size_t i = 0; i < clusters; i++)
```

to

```
for (size_t i = 0; i < counts.n_elem; ++i)
```

then there should be no need for `numberOfRemovedClusters` and `EmptyCluster()` can still return `void`.  Let me know what you think.

Prabhat-IIT: Indeed its a good suggestion. I've simplified the code!

Prabhat-IIT:
Plz let me know what else refactoring would be required? Imo it should be ready now :)

src/mlpack/methods/kmeans/kill_empty_clusters.hpp
```diff
@@ -51,13 +51,18 @@ class KillEmptyClusters
       const size_t emptyCluster,
       const arma::mat& /* oldCentroids */,
       arma::mat& newCentroids,
-      arma::Col<size_t>& /* clusterCounts */,
+      arma::Col<size_t>& counts /* clusterCounts */,
```
zoq: Let's use ``clusterCounts`` instead of ``counts`` here to match with the parameter comment above, the reason for /* clusterCounts */ is basically to avoid a parameter unused warning.

src/mlpack/methods/kmeans/kill_empty_clusters.hpp
```diff
@@ -51,13 +51,18 @@ class KillEmptyClusters
...
       MetricType& /* metric */,
       const size_t /* iteration */)
   {
-    // Kill the empty cluster.
-    newCentroids.col(emptyCluster).fill(DBL_MAX);
-    return 0; // No points were changed.
+    // Remove the empty cluster.
+      if (emptyCluster < newCentroids.n_cols)
```
zoq: Looks like there are two extra spaces.

src/mlpack/methods/kmeans/kill_empty_clusters.hpp
```diff
@@ -51,13 +51,18 @@ class KillEmptyClusters
...
+      {
+          newCentroids.shed_col(emptyCluster);
+          counts.shed_row(emptyCluster);
+          return 0; // one cluster removed
```
zoq: Should this be 1? Or can we remove the return value here and just return at the end, since we always return 0.

src/mlpack/tests/kmeans_test.cpp
```diff
@@ -134,6 +135,41 @@ BOOST_AUTO_TEST_CASE(AllowEmptyClusterTest)
     BOOST_REQUIRE_EQUAL(counts[i], countsOld[i]);
 }
 
+/**
+ * Make sure kill empty cluster policy removes the empty cluster
```
zoq: Picky comment do you mind to use the correct punctuation here.

src/mlpack/tests/kmeans_test.cpp
```diff
@@ -134,6 +135,41 @@ BOOST_AUTO_TEST_CASE(AllowEmptyClusterTest)
...
+  // Make sure no assignments were changed.
+  for (size_t i = 0; i < assignments.n_elem; i++)
+    BOOST_REQUIRE_EQUAL(assignments[i], assignmentsOld[i]);
+
+  // Make sure no counts were changed for clusters that are not empty.
+  for (size_t i = 0; i < 2; i++)
+    BOOST_REQUIRE_EQUAL(counts[i], countsOld[i]);
+
+  // Make sure that counts contain one less element than old counts.
+  BOOST_REQUIRE_GT(countsOld.n_elem,counts.n_elem);
```
zoq: Another picky style comment, missing space between ``countsOld.n_elem`` and ``counts.n_elem``.

src/mlpack/methods/kmeans/kill_empty_clusters.hpp
Prabhat-IIT: It's all because of my IDE. I'm sorry for these small mistakes

rcurtin:

src/mlpack/methods/kmeans/kill_empty_clusters.hpp
```diff
@@ -51,13 +51,17 @@ class KillEmptyClusters
...
-      arma::Col<size_t>& /* clusterCounts */,
+      arma::Col<size_t>& clusterCounts,
       MetricType& /* metric */,
       const size_t /* iteration */)
   {
-    // Kill the empty cluster.
-    newCentroids.col(emptyCluster).fill(DBL_MAX);
-    return 0; // No points were changed.
+    // Remove the empty cluster.
+    if (emptyCluster < newCentroids.n_cols)
```
zoq: Looks like the brace placement and indentation is off, can you take a look at: https://github.com/mlpack/mlpack/wiki/DesignGuidelines#brace-placement.

Prabhat-IIT:
Done @zoq the style check is still failing due to other files!!

rcurtin:
Thanks for the contribution.  I made a few additional changes during merge in a94a1f5e just for simplification.

Prabhat-IIT:
Thanks for constant support rcurtin :)

PR #1240, opened 2018-02-12
Update CMakeLists.txt

LeonCHEN233:
The CMakeList does not support boost.1.66.0 on windows yet, so I added new Boost version to CMakeList.txt .


rcurtin:
Ok, with two approvals I will go ahead and merge this now.  Thank you for the contribution! :)

LeonCHEN233:
thanks for reviewing

Issue #1228, opened 2018-02-08
Missing common initializations in the ann module


sshkhr:
Xavier initialization (introduced in [Understanding the difficulty of training deep feedforward neural networks](http://proceedings.mlr.press/v9/glorot10a/glorot10a.pdf) ) and He initialization (introduced in [Delving Deep into Rectifiers: Surpassing Human-Level Performance on ImageNet Classification](https://arxiv.org/abs/1502.01852) ) are among the most common initializations used today in training deep neural nets. I found that mlpack is missing an implementation of these in the init_rules of the ann module.  

#939 was working on implementing these but no work has been done on it for some time now. Is this feature still being worked on ?

sshkhrnwbie:
Thanks for the update @zoq . I will try and work on it over the weekend. If I need any help I will take the discussion to the irc channel.

zoq:
Sounds good.

saurabhpal97:
@sshkhrnwbie any updates regarding the implementation? I also want to work on this issue so if you have not already implemented the initialization modules maybe we can split the work.

sshkhrnwbie:
That sounds like a great idea @saurabhpal97 . I am trying to get a VarianceScaling initializer implementation to work. You will find it much easier to implement the Xavier and He initializers using it once it is finished.

saurabhpal97:
Thanks @sshkhrnwbie  I will get right to it

kartikdutt18:
Hey @rcurtin, should we close this issue ?
He Initialization(#1280) and Glorot Initialization (Xavier #1272) are already present in mlpack. Thanks.

kartikdutt18:
Hey @rcurtin, @zoq should we close this issue ?
He Initialization and Glorot Initialization (Xavier) are already present in mlpack. Thanks.

rcurtin:
@kartikdutt18 you're right, thanks for pointing it out.  I think we can close it.  (If I missed something, let me know, we can reopen it. :))

PR #1241, opened 2018-02-13
Added VarianceScaling initializer 

sshkhrnwbie:
This PR tries to solve #1228 by implementing the VarianceScaling initializer based on the [Keras implementation](https://github.com/keras-team/keras/blob/master/keras/initializers.py#L155). The aim is to use this initializer to further implement other common initializers like Glorot and He. 

@zoq I could use your help in figuring out if this is a good (and correct) way to approach it. I haven't added the test I wrote based on this implementation as I wasn't sure if this was the right way to do it and also because the build was failing at my end due to the tests.


zoq:
This is what I had in mind, looks like Travis is able to build the code; if you post some additional information e.g. error message, I'm sure we can figure it out.

sshkhrnwbie:
@zoq I re-wrote the tests in same manner as GaussianInitTest and OivsInitTest and this time the build was passing at my end.

src/mlpack/methods/ann/init_rules/variance_scaling_init.hpp
```diff
@@ -0,0 +1,199 @@
...
+  template<typename eT, typename Mode = ModeType>
+  void Initialize(arma::Mat<eT>& W, 
+    const size_t rows, 
+    const size_t cols,
+    typename std::enable_if_t<std::is_same<Mode, FanOut>::value>* = 0)
+  {
+    double var = (scalingFactor / (double) (cols));
+    GaussianInitialization init(0, var);
+    init.Initialize(W, rows, cols);
+  }
```
rcurtin: To be honest, I am not sure that templates are really necessary here to split the functionality three ways.  Since `VarianceScalingNormalInit` is so simple, we might consider splitting it into three classes.  Alternately, I would suggest avoiding SFINAE for simplicity and simply make the `FanIn`, `FanOut`, or `FanAvg` classes implement an `Initialize()` method themselves that you can call from `VarianceScalingNormalInit::Initialize()` and `VarianceScalingUniformInit::Initialize()`.

src/mlpack/tests/init_rules_test.cpp
```diff
@@ -197,6 +198,56 @@ BOOST_AUTO_TEST_CASE(GaussianInitTest)
   BOOST_REQUIRE_EQUAL(weights3d.n_slices, slices);
 }
 
+/**
+ * Simple test of the VarianceScalingNormalInit class.
```
rcurtin: Can you clarify a little in your comment about what we are testing in this test please?

src/mlpack/methods/ann/init_rules/variance_scaling_init.hpp
zoq: I was also thinking about adding a single class for each, but since we would end up with 6 new classes, decided to go with SFINAE. We might be able to reduce the amount by using another template for ``FanIn``, ``FanOut`` and ``FanAvg`` that specifies the distribution.

rcurtin: Hmm, I guess you are right.  If we can't think of a better solution, I guess this one is fine.

src/mlpack/tests/init_rules_test.cpp
sshkhrnwbie: Okay. I will edit the test file in the next commit.

src/mlpack/tests/init_rules_test.cpp
```diff
@@ -197,6 +198,119 @@ BOOST_AUTO_TEST_CASE(GaussianInitTest)
...
+  for (size_t i = 0; i < rows; i++)
+    for (size_t j = 0; j < cols; j++)
+      BOOST_REQUIRE_SMALL(weights.at(i, j) - vsnWeights.at(i, j), 1e-3);
+
+  for (size_t k = 0; k < slices; k++)
+    for (size_t i = 0; i < rows; i++)
+      for (size_t j = 0; j < cols; j++)
+        BOOST_REQUIRE_SMALL(weights3d.slice(k).at(i, j) - 
+        					vsnWeights3d.slice(k).at(i, j), 1e-3);
+
```
sshkhrnwbie: I now realize the test I wrote (taken from OthogonalInit test without much thought) regarding how close the values should be to manually initializing the weights using the same approach doesn't make much sense since the range of numbers from the underlying distribution is much higher than 1e-3. 

@rcurtin Could you suggest me some other tests which should be useful in this case. Maybe I should add checks to see if the values are lying in the required range ? Any suggestions are most welcome.

zoq: Yeah coming up with a meaningful test here is tricky, testing if they are in the correct range is one approach.

sshkhrnwbie: Okay I will edit the tests to check the range which the initialized values lie in.

sshkhrnwbie:
@zoq I have added checks on the range of values initialized. The Travis-CI build seems to have timed out in the third job but passed on the other two. I was able to build locally as well.  

Please have a look and see if any other thing needs to be added. I think this should be okay for now,

src/mlpack/tests/init_rules_test.cpp
rcurtin: Just a note, this one is still not handled yet.  I wanted to make sure it does not fall off your radar. :)

src/mlpack/tests/init_rules_test.cpp
```diff
@@ -197,6 +198,103 @@ BOOST_AUTO_TEST_CASE(GaussianInitTest)
...
+  	  BOOST_REQUIRE_LE(weights.at(i, j), 3*stddev);
+  	}
+
+  for (size_t k = 0; k < slices; k++)
+    for (size_t i = 0; i < rows; i++)
+      for (size_t j = 0; j < cols; j++)
+      {
+        BOOST_REQUIRE_GE(weights3d.slice(k).at(i, j), - 3*stddev);
+    	BOOST_REQUIRE_LE(weights3d.slice(k).at(i, j), 3*stddev);
+      }
```
rcurtin: Some of the spacing and style does not seem to be right, especially in this section.  Could you go through the style guide and fix it please?  We normally have an automatic style checker but it is offline right now (hopefully it will be back online soon).

https://github.com/mlpack/mlpack/wiki/DesignGuidelines#style-guidelines

sshkhrnwbie: I will fix the style issues asap

zoq:

sshkhrnwbie:
@zoq Looks like the style checks that are failing now are in the logistic regression test, decision tree test and dbscan test. I can send in a separate PR for these if desired.

rcurtin:
You're right; the other style issues have been fixed in #1283, so you don't need to worry about those.

src/mlpack/methods/ann/init_rules/variance_scaling_init.hpp
```diff
@@ -0,0 +1,195 @@
...
+    double var = (scalingFactor / (double) (rows));
+    GaussianInitialization init(0, var);
+    init.Initialize(W, rows, cols);
+  }
+
+  template<typename eT, typename Mode = ModeType>
+  void Initialize(arma::Mat<eT>& W,
+    const size_t rows,
+    const size_t cols,
+    typename std::enable_if_t<std::is_same<Mode, FanOut>::value>* = 0)
```
rcurtin: A tiny style comment---you should either line up the extra parameters with the first `arma::Mat<eT>& W` or double-indent the wrapped lines for the parameters.  Either is fine.

src/mlpack/methods/ann/init_rules/variance_scaling_init.hpp
```diff
@@ -0,0 +1,195 @@
...
+    init.Initialize(W, rows, cols);
+  }
+
+  template<typename eT, typename Mode = ModeType>
+  void Initialize(arma::Mat<eT>& W,
+    const size_t rows,
+    const size_t cols,
+    typename std::enable_if_t<std::is_same<Mode, FanAvg>::value>* = 0)
+  {
+    double var = (2*scalingFactor / ((double) (rows) + (double) (cols)));
```
rcurtin: I guess the style checker does not catch this... can you use `2 * scalingFactor` please? (horizontal whitespace)

This appears a few more times in the PR, so if you can fix it in all places that would be great.

sshkhrnwbie:
@rcurtin This does not follow one particular paper as such. The initializer changes its scale to the shape of the weights for both the normal and the uniform case. It is based on the [Keras implementation](https://github.com/keras-team/keras/blob/master/keras/initializers.py#L155) for the same.

I implemented it mostly to enable a fix for #1228 as this will make it easier to implement a He, LeCun or Glorot initializer (both uniform and normal) as they are all just variations of this. @zoq  suggested on the IRC that this may be good to have for implementing the above mentioned ones. 

Since I raised the issue #1272 and #1280 have been opened which to try to fix the missing initializations individually. If this and the above two get merged I think we can go ahead and close #1228

rcurtin:
That sounds fine, all I am pointing out is that my review only touches on the implementation side of things, and I don't know the larger plan of action here. :)

sshkhrnwbie:
@rcurtin Got your point :) 
Anyway I think this is more or less done from my side and the checks are passing as well. Whenever you and @zoq get a chance to go through it please do and point out if anything more is required or if this is good to merge. Thanks.

src/mlpack/methods/ann/init_rules/variance_scaling_init.hpp
```diff
@@ -0,0 +1,202 @@
+/**
+ * @file variance_scaling_init.hpp
+ * @author Shashank Shekhar
+ *
+ * Intialization rule for the neural networks.
+ * VarianceScalingNormalInit is performed by assigning a gaussian matrix with
+ * zero mean and variance given by (scaling factor / n) to the weight matrix.
+ * VarianceScalingUniformInit is performed by assigning a random matrix between
```
zoq: Do you think an extra line here could improve the overall readability?

src/mlpack/methods/ann/init_rules/variance_scaling_init.hpp
```diff
@@ -0,0 +1,202 @@
...
+#include <mlpack/core/math/random.hpp>
+
+#include "gaussian_init.hpp"
+#include "random_init.hpp"
+
+
+namespace mlpack {
+namespace ann /** Artificial Neural Network. */ {
+
+class FanIn
```
zoq: Can you clarify for what each class is used for?

src/mlpack/methods/ann/init_rules/variance_scaling_init.hpp
```diff
@@ -0,0 +1,202 @@
...
+ * n = cols for FanOut mode (number of neurons feeding out)
+ * n = (rows + cols)/2 for FanAvg mode (average of neurons feeding in and out)
+ */
+template<typename ModeType = FanAvg>
+class VarianceScalingNormalInit
+{
+ public:
+  // Empty Constructor
+  // Scaling factor is the amount by which variance will be scaled.
+  VarianceScalingNormalInit(const size_t scalingFactor = 1):
```
zoq: It would be great if you could add a comment for each parameter, so that doxygen can pick up the information and generate the documentation. Ideally you can take a look at any existing class for an example.

src/mlpack/methods/ann/init_rules/variance_scaling_init.hpp
```diff
@@ -0,0 +1,202 @@
...
+  {}
+
+  template<typename eT, typename Mode = ModeType>
+  void Initialize(
+      arma::Mat<eT>& W,
+      const size_t rows,
+      const size_t cols,
+      typename std::enable_if_t<std::is_same<Mode, FanIn>::value>* = 0)
+  {
+    double var = (scalingFactor / (double) (rows));
```
zoq: The surrounding ``()`` is redundant.  Also, should we use ``const`` here?

src/mlpack/methods/ann/init_rules/variance_scaling_init.hpp
```diff
@@ -0,0 +1,202 @@
...
+      const size_t rows,
+      const size_t cols,
+      typename std::enable_if_t<std::is_same<Mode, FanAvg>::value>* = 0)
+  {
+    double limit = sqrt(6 * scalingFactor /
+        ((double) (rows) + (double) (cols)));
+    RandomInitialization init(-limit, limit);
+    init.Initialize(W, rows, cols);
+  }
+
```
zoq: Picky style comment, remove the extra line.

src/mlpack/tests/init_rules_test.cpp
```diff
@@ -197,6 +198,108 @@ BOOST_AUTO_TEST_CASE(GaussianInitTest)
...
+
+  BOOST_REQUIRE_EQUAL(weights3d.n_rows, rows);
+  BOOST_REQUIRE_EQUAL(weights3d.n_cols, cols);
+  BOOST_REQUIRE_EQUAL(weights3d.n_slices, slices);
+
+  // Check that the initialized values lie in the required range
+
+  double limit = sqrt(3.0 / ((rows + cols) / 2.0));
+
+  for (size_t i = 0; i < rows; i++)
```
zoq: We could use ``arma::all`` to simplify the test here. Let me know what you think.

sshkhrnwbie: arma:all does make sense. It should reduce all the loops.

src/mlpack/methods/ann/init_rules/variance_scaling_init.hpp
sshkhrnwbie: sure

src/mlpack/methods/ann/init_rules/variance_scaling_init.hpp
sshkhrnwbie: I will look into doxygen documentation and make the required additions.

src/mlpack/methods/ann/init_rules/variance_scaling_init.hpp
sshkhrnwbie: I had described the classes later on but I think it makes sense to remove the two descriptions below and add a single one here.

src/mlpack/methods/ann/init_rules/variance_scaling_init.hpp
sshkhrnwbie: yes. I will add it.

rcurtin:
@sshkhrnwbie: have you had a chance to look into this?  If you don't think that you will finish this we can close it for inactivity or maybe someone else will finish it, just let us know.

sshkhrnwbie:
@rcurtin Sorry I got busy and couldn't finish off the requested changes. I will try and make the required changes over the weekend.

rcurtin:
No worries, sounds good.  Just wanted to check on the status. :)

rcurtin:
Hi @sshkhrnwbie, I am really sorry, I think there has been an oversight.  We are already merging PRs #1342/#1280 and have merged #1272, which contain both the He and Glorot and Lecun initializations.  So I think it turns out this implementation is redundant.  I'm sorry about the misunderstanding, but since those are already in progress we can go with those instead.

In any case, I recognize that you've put in a lot of work to this PR.  So I would be happy to mail you some mlpack stickers.  If you'd like that, just send me an email at ryan@ratml.org with your mailing address and I will get the stickers in the mail for you.

sshkhrnwbie:
Hi @rcurtin That's quite alright. The end aim of the issue and the PR was to get those initializers into mlpack.  

I will drop you a mail and get started on some other issues. Cheers

PR #1242, opened 2018-02-13
Added zip_safe to setup options in setup.py.in

### LINKED ISSUE #1231: Python import takes too long.

akhandait:
Resolves #1231
Here is the documentation : http://setuptools.readthedocs.io/en/latest/setuptools.html#setting-the-zip-safe-flag
It now extracts the files in dist-packages and import time in python is drastically reduced.

### DIFF
--- a/COPYRIGHT.txt
+++ b/COPYRIGHT.txt
@@ -89,6 +89,7 @@ Copyright:
   Copyright 2017&2018, Sourabh Varshney <sourabhvarshney111@gmail.com>
   Copyright 2018, Projyal Dev <projyal@gmail.com>
   Copyright 2018, Nikhil Goel <nikhilgoel199797@gmail.com>
+  Copyright 2018, Atharva Khandait <akhandait45@gmail.com>
 
 License: BSD-3-clause
   All rights reserved.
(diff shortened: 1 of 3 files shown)


akhandait:
What can I do to pass the travis check?

rcurtin:
Don't worry about the failing travis build; I have seen this in some other PRs too.  I will be looking into it later today and I'll let you know what I find.  Also #1221 is open to try and help with the random Travis builds that are not starting correctly.

I think this change is just fine, but let me double-check that the zip_safe change isn't affecting the Python build by hand after I debug the other Python failure and then we can go ahead and merge.

akhandait:
It will be nice to have my name in the list of contributors. Will you add it or should I add it myself and commit it here?

rcurtin:
If you can just add it here as a commit, that would be great. :)

rcurtin:

PR #1246, opened 2018-02-14
DBSCAN Binding test

nikhilgoel1997:
DBSCAN Binding test as required from issue #1152

### DIFF
--- a/src/mlpack/methods/dbscan/dbscan_main.cpp
+++ b/src/mlpack/methods/dbscan/dbscan_main.cpp
@@ -127,8 +127,8 @@ static void mlpackMain()
   RequireParamValue<double>("epsilon", [](double x) { return x > 0; },
       true, "invalid value of epsilon specified");
 
-  // Value of min_size can't be negative.
-  RequireParamValue<int>("min_size", [](int y) { return y >= 0; },
+  // Value of min_size should be positive.
+  RequireParamValue<int>("min_size", [](int y) { return y > 0; },
       true, "invalid value of min_size specified");
 
   // Fire off naive search if needed.
--- a/src/mlpack/methods/dbscan/dbscan_main.cpp
+++ b/src/mlpack/methods/dbscan/dbscan_main.cpp
@@ -123,13 +123,20 @@ static void mlpackMain()
       "hilbert-r", "r-plus", "r-plus-plus", "ball" }, true,
       "unknown tree type");
 
+  // Value of epsilon should be positive.
+  RequireParamValue<double>("epsilon", [](double x) { return x > 0; },
+      true, "invalid value of epsilon specified");
+
+  // Value of min_size can't be negative.
+  RequireParamValue<int>("min_size", [](int y) { return y >= 0; },
+      true, "invalid value of min_size specified");
+
   // Fire off naive search if needed.
   if (CLI::HasParam("naive"))
   {
     RangeSearch<> rs(true);
     RunDBSCAN(rs);
-  }
-
+  } else {
   const string treeType = CLI::GetParam<string>("tree_type");
   if (treeType == "kd")
     RunDBSCAN<RangeSearch<>>();
@@ -149,4 +156,5 @@ static void mlpackMain()
     RunDBSCAN<RangeSearch<EuclideanDistance, arma::mat, RPlusPlusTree>>();
   else if (treeType == "ball")
     RunDBSCAN<RangeSearch<EuclideanDistance, arma::mat, BallTree>>();
+  }
 }
--- a/src/mlpack/methods/dbscan/dbscan_main.cpp
+++ b/src/mlpack/methods/dbscan/dbscan_main.cpp
@@ -136,25 +136,27 @@ static void mlpackMain()
   {
     RangeSearch<> rs(true);
     RunDBSCAN(rs);
-  } else {
-  const string treeType = CLI::GetParam<string>("tree_type");
-  if (treeType == "kd")
-    RunDBSCAN<RangeSearch<>>();
-  else if (treeType == "cover")
-    RunDBSCAN<RangeSearch<EuclideanDistance, arma::mat, StandardCoverTree>>();
-  else if (treeType == "r")
-    RunDBSCAN<RangeSearch<EuclideanDistance, arma::mat, RTree>>();
-  else if (treeType == "r-star")
-    RunDBSCAN<RangeSearch<EuclideanDistance, arma::mat, RStarTree>>();
-  else if (treeType == "x")
-    RunDBSCAN<RangeSearch<EuclideanDistance, arma::mat, XTree>>();
-  else if (treeType == "hilbert-r")
-    RunDBSCAN<RangeSearch<EuclideanDistance, arma::mat, HilbertRTree>>();
-  else if (treeType == "r-plus")
-    RunDBSCAN<RangeSearch<EuclideanDistance, arma::mat, RPlusTree>>();
-  else if (treeType == "r-plus-plus")
-    RunDBSCAN<RangeSearch<EuclideanDistance, arma::mat, RPlusPlusTree>>();
-  else if (treeType == "ball")
-    RunDBSCAN<RangeSearch<EuclideanDistance, arma::mat, BallTree>>();
+  }
+  else
+  {
+    const string treeType = CLI::GetParam<string>("tree_type");
+    if (treeType == "kd")
+      RunDBSCAN<RangeSearch<>>();
+    else if (treeType == "cover")
+      RunDBSCAN<RangeSearch<EuclideanDistance, arma::mat, StandardCoverTree>>();
+    else if (treeType == "r")
+      RunDBSCAN<RangeSearch<EuclideanDistance, arma::mat, RTree>>();
+    else if (treeType == "r-star")
+      RunDBSCAN<RangeSearch<EuclideanDistance, arma::mat, RStarTree>>();
+    else if (treeType == "x")
+      RunDBSCAN<RangeSearch<EuclideanDistance, arma::mat, XTree>>();
+    else if (treeType == "hilbert-r")
+      RunDBSCAN<RangeSearch<EuclideanDistance, arma::mat, HilbertRTree>>();
+    else if (treeType == "r-plus")
+      RunDBSCAN<RangeSearch<EuclideanDistance, arma::mat, RPlusTree>>();
+    else if (treeType == "r-plus-plus")
+      RunDBSCAN<RangeSearch<EuclideanDistance, arma::mat, RPlusPlusTree>>();
+    else if (treeType == "ball")
+      RunDBSCAN<RangeSearch<EuclideanDistance, arma::mat, BallTree>>();
   }
 }
--- a/src/mlpack/tests/main_tests/dbscan_test.cpp
+++ b/src/mlpack/tests/main_tests/dbscan_test.cpp
@@ -220,6 +220,8 @@ BOOST_AUTO_TEST_CASE(DBSCANDiffTreeTypeTest)
   if (!data::Load("iris.csv", inputData))
     BOOST_FAIL("Unable to load dataset iris.csv!");
 
+  // Tree type = kd tree.
+
   SetInputParam("input", inputData);
   SetInputParam("tree_type", std::string("kd"));
 
@@ -232,6 +234,8 @@ BOOST_AUTO_TEST_CASE(DBSCANDiffTreeTypeTest)
 
   CLI::GetSingleton().Parameters()["input"].wasPassed = false;
   CLI::GetSingleton().Parameters()["tree_type"].wasPassed = false;
+
+  // Tree Type = r tree.
 
   SetInputParam("input", inputData);
   SetInputParam("tree_type", std::string("r"));
@@ -246,6 +250,8 @@ BOOST_AUTO_TEST_CASE(DBSCANDiffTreeTypeTest)
   CLI::GetSingleton().Parameters()["input"].wasPassed = false;
   CLI::GetSingleton().Parameters()["tree_type"].wasPassed = false;
 
+  // Tree type = r-star tree.
+
   SetInputParam("input", inputData);
   SetInputParam("tree_type", std::string("r-star"));
 
@@ -258,6 +264,8 @@ BOOST_AUTO_TEST_CASE(DBSCANDiffTreeTypeTest)
 
   CLI::GetSingleton().Parameters()["input"].wasPassed = false;
   CLI::GetSingleton().Parameters()["tree_type"].wasPassed = false;
+
+  // Tree Type = x tree.
 
   SetInputParam("input", inputData);
   SetInputParam("tree_type", std::string("x"));
@@ -272,6 +280,8 @@ BOOST_AUTO_TEST_CASE(DBSCANDiffTreeTypeTest)
   CLI::GetSingleton().Parameters()["input"].wasPassed = false;
   CLI::GetSingleton().Parameters()["tree_type"].wasPassed = false;
 
+  // Tree Type = hilbert-r tree.
+
   SetInputParam("input", inputData);
   SetInputParam("tree_type", std::string("hilbert-r"));
 
@@ -284,6 +294,8 @@ BOOST_AUTO_TEST_CASE(DBSCANDiffTreeTypeTest)
 
   CLI::GetSingleton().Parameters()["input"].wasPassed = false;
   CLI::GetSingleton().Parameters()["tree_type"].wasPassed = false;
+
+  // Tree Type = r-plus tree.
 
   SetInputParam("input", inputData);
   SetInputParam("tree_type", std::string("r-plus"));
@@ -298,6 +310,8 @@ BOOST_AUTO_TEST_CASE(DBSCANDiffTreeTypeTest)
   CLI::GetSingleton().Parameters()["input"].wasPassed = false;
   CLI::GetSingleton().Parameters()["tree_type"].wasPassed = false;
 
+  // Tree Type = r-plus-plus tree.
+
   SetInputParam("input", inputData);
   SetInputParam("tree_type", std::string("r-plus-plus"));
 
@@ -311,6 +325,8 @@ BOOST_AUTO_TEST_CASE(DBSCANDiffTreeTypeTest)
   CLI::GetSingleton().Parameters()["input"].wasPassed = false;
   CLI::GetSingleton().Parameters()["tree_type"].wasPassed = false;
 
+  // Tree Type = cover tree.
+
   SetInputParam("input", inputData);
   SetInputParam("tree_type", std::string("cover"));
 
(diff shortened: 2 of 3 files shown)


src/mlpack/methods/dbscan/dbscan_main.cpp
```diff
@@ -123,6 +123,14 @@ static void mlpackMain()
       "hilbert-r", "r-plus", "r-plus-plus", "ball" }, true,
       "unknown tree type");
 
+  // Value of epsilon sohould be greater than 0.
+  RequireParamValue<double>("epsilon", [](double x) { return x > 0; },
+      true, "invalid value of epsilon specified");
```
manish7294: I think it would me more reasonable to write something like `Value of epsilon can't be negative `. Same with `min_size. `What do you think?

src/mlpack/tests/main_tests/dbscan_test.cpp
```diff
@@ -0,0 +1,476 @@
...
+
+  arma::Row<size_t> output2;
+  output2 = std::move(CLI::GetParam<arma::Row<size_t>>("assignments"));
+
+  int noisePointsOutput2 = 0;
+
+  for (size_t i = 0; i < output2.n_elem; ++i)
+  {
+    if (output2[i] == 18446744073709551615)
+    {
```
manish7294: It think it would be better use `arma::accu(A==B)` here. And moreover instead of mentioning this big dirty number here may be you can use some kind of numeric limit as noise may not always have this value . Please check this at other places too.

src/mlpack/tests/main_tests/dbscan_test.cpp
```diff
@@ -0,0 +1,476 @@
...
+  CheckMatrices(kdOutput, rPlusOutput);
+  CheckMatrices(kdOutput, rPlusPlusOutput);
+  CheckMatrices(kdOutput, coverOutput);
+  CheckMatrices(kdOutput, ballOutput);
+}
+
+/**
+ * Check that the assignment of cluster is same if
+ * naive search or single tree is used for search
+ */
```
manish7294: Just a minor comment, please end the comments with a  '.' :)

src/mlpack/tests/main_tests/dbscan_test.cpp
```diff
@@ -0,0 +1,476 @@
...
+
+  mlpackMain();
+
+  // Check that number of predicted labels is equal to the input test points.
+  BOOST_REQUIRE_EQUAL(CLI::GetParam<arma::Row<size_t>>("assignments").n_cols,
+                      inputSize);
+  BOOST_REQUIRE_EQUAL(CLI::GetParam<arma::Row<size_t>>("assignments").n_rows,
+                      1);
+  BOOST_REQUIRE_EQUAL(CLI::GetParam<arma::mat>("centroids").n_rows, 2);
+  BOOST_REQUIRE_EQUAL(CLI::GetParam<arma::mat>("centroids").n_cols, 4);
```
manish7294: May be you can write some small comments here about these numbers in centroids dimension check as it may vary with datasets. Your call?

src/mlpack/tests/main_tests/dbscan_test.cpp
nikhilgoel1997: I got the same number for noise points on all datasets I ran DBSCAN on. I'm not sure if noise can have a different value..

manish7294: Ya, you are right as garbage value depends on the datatype. So, it will remain same for a particular datatype.

nikhilgoel1997:
Hello
I don't understand why the naiveSearchTest is not working. Travis build is working fine without the naive search test. Brute force search is working in command line. (The total time is significantly more)
Am I missing something here?

src/mlpack/tests/main_tests/dbscan_test.cpp
manish7294: Since, you are using iris dataset so the assignments can't be greater than 3. So, may be you case use something like `noisePointOutput2 = arma::accu(output2 < 0 && output2 > 2)` to calculate `noisePoints` and add a small comment about the same. Let me know what you think.

nikhilgoel1997: I think there is a misunderstanding. This is unsupervised learning. Assignments can be as many as possible with suitable input options. DBSCAN just forms clusters. If you give min size of clusters to be 0 and radius of search(Epsilon) to be very less as well, you can get as many clusters as the number of initial points.
Something like `output2 > inputData.n_cols` would be more appropriate.

manish7294: You are right, that's what I am trying to specify that there will always be a upper limit, as in case of assignments it will be 3 and as you said in clusters it will be inputData.n_cols. Let me know if it clears anything.

manish7294:
I went through naiveSearchTest and found that it's giving empty output for `naive search`. I also verified the same using command :
`bin/mlpack_dbscan  -i iris.csv -N true -C centroids.csv`
So, I think there may be a bug in DBSCAN bindings but not sure. If you like you may try having a look at dbscan_main.cpp.

src/mlpack/tests/main_tests/dbscan_test.cpp
nikhilgoel1997: Assignments are the same as the number of clusters in DBSCAN.
I'll take care of it! I also didn't want to use the garbage value. Thanks for your help!

nikhilgoel1997:
I also think it's a bug with bindings because command line is working. I'm just waiting for confirmation before opening an issue. Since you're able to reproduce the problem, is it safe to assume it's a bug?

src/mlpack/tests/main_tests/dbscan_test.cpp
manish7294: Ya, you are right, I just wanted to reflect the idea of using arma::accu().

manish7294:
I don't think that command line is working quite as expected. As I got nothing using command-line. Moreover command line executes the command using bindings only, so it may be a problem with bindings.

nikhilgoel1997:
Are you sure? I'm getting the correct answer using command line
`mlpack_dbscan -i iris.csv -N true -C centroids.csv -a centroids.txt`
The output is the same as I got without `-N`
And since the execution time increased quite significantly, it's safe to assume that naive search is working.

manish7294:
Now it's weird as I am getting nothing using` -N true` while as on removing it I am getting perfect results.

src/mlpack/tests/main_tests/dbscan_test.cpp
nikhilgoel1997: It's absolutely fine! I was able to think of another test and remove those sad garbage values! Thank you for your help!

nikhilgoel1997:
let's wait for the owners and see if they're able to reproduce this bug and if naive search is working for them.

src/mlpack/tests/main_tests/dbscan_test.cpp
```diff
@@ -190,7 +212,7 @@ BOOST_AUTO_TEST_CASE(DBSCANEpsilonNoisePointsTest)
 
   for (size_t i = 0; i < output1.n_elem; ++i)
   {
-    if (output1[i] == 18446744073709551615)
+    if (arma::accu(output1[i] > inputData.n_cols))
     {
       noisePointsOutput1++;
```
manish7294: I think this will always set noisePointOutput to a max of ,1 as you are just checking the return value of accu in if statement. I think if we want number of noisePoints then something like `noisePoints = arma::accu(output1[i] > inputData.n_cols) `should be enough. And you may impose the condition for negative garbage values too.

rcurtin:
I can reproduce the issue when I compile with debugging symbols:

```
$ bin/mlpack_dbscan -i iris.csv -N -C centroids.csv -a centroids.txt -v
[DEBUG] Compiled with debugging symbols.
[INFO ] Loading 'iris.csv' as CSV data.  Size is 4 x 150.
[INFO ] Performing range search.
[INFO ] Range search complete.
[INFO ] 2 clusters found.
[INFO ] Performing range search.
[INFO ] Range search complete.

error: max(): object has no elements
  what():  max(): object has no elements
Aborted
```

Without the `-N` I get:

```
$ bin/mlpack_dbscan -i iris.csv -C centroids.csv -a centroids.txt -v
[DEBUG] Compiled with debugging symbols.
[INFO ] Loading 'iris.csv' as CSV data.  Size is 4 x 150.
[INFO ] Performing range search.
[INFO ] Range search complete.
[INFO ] 2 clusters found.
[INFO ] Saving raw ASCII formatted data to 'centroids.txt'.
[INFO ] Saving CSV data to 'centroids.csv'.
[INFO ] 
[INFO ] Execution parameters:
[INFO ]   assignments_file: centroids.txt
[INFO ]   centroids_file: centroids.csv
[INFO ]   epsilon: 1
[INFO ]   help: 0
[INFO ]   info: 
[INFO ]   input_file: iris.csv
[INFO ]   min_size: 5
[INFO ]   naive: 0
[INFO ]   single_mode: 0
[INFO ]   tree_type: kd
[INFO ]   verbose: 1
[INFO ]   version: 0
[INFO ] Program timers:
[INFO ]   loading_data: 0.000860s
[INFO ]   range_search/computing_neighbors: 0.008075s
[INFO ]   range_search/tree_building: 0.000262s
[INFO ]   saving_data: 0.000428s
[INFO ]   total_time: 0.011944s
```

So there is definitely some issue with the naive search.  If you'd like to dig into it more, I'd appreciate it, otherwise we can open an issue.  I would say running gdb and getting a backtrace when the exception is thrown from Armadillo should expose pretty clearly what is going on.  Let me know if you would like any help with that, if you'd like to do it.

src/mlpack/tests/main_tests/dbscan_test.cpp
```diff
@@ -0,0 +1,517 @@
...
+ * when min_size is equal to 0.
+ */
+BOOST_AUTO_TEST_CASE(DBSCANClusterNumberTest)
+{
+  arma::mat inputData;
+  if (!data::Load("iris.csv", inputData))
+    BOOST_FAIL("Unable to load dataset iris.csv!");
+
+  SetInputParam("input", inputData);
+  SetInputParam("min_size", (int) 0);
```
rcurtin: I'm a little confused by this---the check you added to `dbscan_main.cpp` forces `min_size` to be greater than 0.  So `mlpackMain()` should throw an exception here.  I'm also not sure I understand what you mean to test here; can you clarify please?

src/mlpack/tests/main_tests/dbscan_test.cpp
```diff
@@ -0,0 +1,517 @@
...
+  mlpackMain();
+
+  arma::Row<size_t> output1;
+  output1 = std::move(CLI::GetParam<arma::Row<size_t>>("assignments"));
+
+  int noisePointsOutput1 = 0;
+
+  for (size_t i = 0; i < output1.n_elem; ++i)
+  {
+    if (arma::accu(output1[i] > inputData.n_cols))
```
rcurtin: Why use `arma::accu()` here?  A noise assignment will be `SIZE_MAX` (I updated the documentation, which was incorrect, in d984e10e4; sorry about that), so you can just compare against that value.

src/mlpack/tests/main_tests/dbscan_test.cpp
```diff
@@ -0,0 +1,517 @@
...
+
+  for (size_t i = 0; i < output2.n_elem; ++i)
+  {
+    if (arma::accu(output2[i] > inputData.n_cols))
+    {
+      noisePointsOutput2++;
+    }
+  }
+
+  BOOST_REQUIRE_LE(noisePointsOutput1, noisePointsOutput2);
```
rcurtin: I am not sure this is the best strategy for testing here.  What we want to see is that our results are different as `epsilon` changes, but as we change `epsilon` we are not guaranteed that the number of points that are noise change (in the worst case, the assignments stay the same).  I would instead suggest sweeping the value of `epsilon` and `min_points` and making sure the resulting cluster assignments are different for at least one element; this lets us know that the values of `epsilon` and `min_points` are being taken into account by the binding.  So I think the other two tests you have written are fine for that case.

src/mlpack/methods/dbscan/dbscan_main.cpp
rcurtin: I would actually say `value of epsilon must be positive`, since we are not allowing `0` as a value.  (The same applies for `min_size`).  It is a subtle difference but still a difference... :)

src/mlpack/tests/main_tests/dbscan_test.cpp
```diff
@@ -0,0 +1,517 @@
...
+
+  mlpackMain();
+
+  // Check that number of predicted labels is equal to the input test points.
+  BOOST_REQUIRE_EQUAL(CLI::GetParam<arma::Row<size_t>>("assignments").n_cols,
+                      inputSize);
+  BOOST_REQUIRE_EQUAL(CLI::GetParam<arma::Row<size_t>>("assignments").n_rows,
+                      1);
+  BOOST_REQUIRE_EQUAL(CLI::GetParam<arma::mat>("centroids").n_rows, 4);
+  BOOST_REQUIRE_EQUAL(CLI::GetParam<arma::mat>("centroids").n_cols, 2);
```
rcurtin: We aren't necessarily guaranteed that DBSCAN with its default parameters will converge to two clusters.  But I think it is reasonable to require that we have at least one centroid.

src/mlpack/tests/main_tests/dbscan_test.cpp
manish7294: Sorry, It's on my part. I suggested Nikhil to either use `noisePointOutput2 = arma::accu(output2 > UPPER_LIMIT_OF_DATASET_POINTS)` instead of calculating using garbage value or to use some kind of numeric_limit here. I wasn't able to give enough points on part of numeric_limit.
Well, I think it would be better to use  `noisePointOutput2 = arma::accu(output2 == SIZE_MAX)` to calculate number of points.

src/mlpack/tests/main_tests/dbscan_test.cpp
nikhilgoel1997: I am sorry for that. It should be >= 0 in `dbscan_main.cpp`
Setting min_size = 0 ensures that there will be no noise points which will maximize the number of clusters.

src/mlpack/tests/main_tests/dbscan_test.cpp
nikhilgoel1997: Yes, it depends on initial centroids! Sorry about that

src/mlpack/tests/main_tests/dbscan_test.cpp
nikhilgoel1997: No, I had different test for different results if we change the value of epsilon. 
With this test, I wanted to check if noise points increases or in worst case remain the same.
I'll remove the test if it is not required

nikhilgoel1997:
I will look into the problem with naive search asap and open an issue afterwards!

rcurtin:
No need to open another issue---just fix it as part of this PR. :)

src/mlpack/tests/main_tests/dbscan_test.cpp
rcurtin: If I am not mistaken, `min_size = 1` should also guarantee there are no noise points.

src/mlpack/tests/main_tests/dbscan_test.cpp
rcurtin: Yep, sounds good.  This should already be tested in `DBSCANTest` (`src/mlpack/tests/dbscan_test.cpp`).  Keep in mind, the purpose of testing the bindings is just to make sure that the options being given get to the `DBSCAN` class correctly.  We have already tested the `DBSCAN` method itself in the other tests.  At the same time, if you see that there are some things about the `DBSCAN` class itself that are not being tested, then definitely feel free to also add tests there. :)

src/mlpack/tests/main_tests/dbscan_test.cpp
rcurtin: Sorry, I am a little confused as to what you mean---there are no initial centroids in DBSCAN.  Maybe I misunderstood?

src/mlpack/tests/main_tests/dbscan_test.cpp
nikhilgoel1997: Yes, you're right. I'm sorry about the confusion!

src/mlpack/tests/main_tests/dbscan_test.cpp
nikhilgoel1997: I added as many tests as I could think of.. 
Yes, these tests are not required here! I'll remove them
Am I missing any other parameter tests?

src/mlpack/methods/dbscan/dbscan_main.cpp
nikhilgoel1997: I've removed 0 as it was not required. Hope the comments are fine now

nikhilgoel1997:
Hi, I've fixed the naiveSearch problem.
Also, allowing 0 as a value for min_size may not be a good idea so I've decided to remove it. Is there anything else I need to do in this test?

src/mlpack/methods/dbscan/dbscan_main.cpp
```diff
@@ -123,13 +123,20 @@ static void mlpackMain()
...
+      true, "invalid value of min_size specified");
+
   // Fire off naive search if needed.
   if (CLI::HasParam("naive"))
   {
     RangeSearch<> rs(true);
     RunDBSCAN(rs);
-  }
-
+  } else {
```
rcurtin: I know the style checker is not currently working, but can you reformat this line as

```
}
else
{
```

and update the spacing for the code now inside the else block? Thanks!

src/mlpack/tests/main_tests/dbscan_test.cpp
```diff
@@ -0,0 +1,409 @@
...
+
+  Log::Fatal.ignoreInput = true;
+  BOOST_REQUIRE_THROW(mlpackMain(), std::runtime_error);
+  Log::Fatal.ignoreInput = false;
+}
+
+/**
+ * Number of clusters should be less than the input number of points
+ * when min_size is equal to 1.
+ */
```
rcurtin: Can you update this comment? I think it is not accurate.  This test actually appears to be testing that no points are labeled as noise when min_size is 1.

nikhilgoel1997:
Sorry for the delay! I was extremely busy in an event in my college.
Thank you for your help @manish7294 and rcurtin!

src/mlpack/tests/main_tests/dbscan_test.cpp
```diff
@@ -0,0 +1,409 @@
...
+    BOOST_FAIL("Unable to load dataset iris.csv!");
+
+  SetInputParam("input", inputData);
+  SetInputParam("tree_type", std::string("kd"));
+
+  mlpackMain();
+
+  arma::Row<size_t> kdOutput;
+  kdOutput = std::move(CLI::GetParam<arma::Row<size_t>>("assignments"));
+
```
manish7294: Maybe you can add some comments to differentiate the start/end for different trees. Sorry, for being so picky. It's your call if you like to, don't feel obligated.

rcurtin:
Since this has two approvals, I will go ahead and merge it now.  Thank you for the contribution! :+1:

PR #1249, opened 2018-02-14
Travis fixes

rcurtin:
This should fix some of the Travis build issues we have been seeing.


rcurtin:
Let's hope it helps some of the other builds too. :)

Issue #1243, opened 2018-02-13
A possible bug in neighborhood-based CF implementation


Wenhao-H:
Hi,

I noticed that in cf.cpp, when calculating the nearest neighbors of a certain user _users(i)_, the user _users(i)_ itself is returned as one of the nearest neighbors:

https://github.com/mlpack/mlpack/blob/5662319dc25d78107e470f4695678956c3f98a95/src/mlpack/methods/cf/cf.cpp#L82-L99

I suppose _users(i)_ should be excluded from the list of nearest neighbors in a neighborhood-based CF.  The code should be easy to fix. But as this implementation seems to be a combination of neighborhood-based method and latent factor based method, i'm not sure whether it's a bug here. If it's a bug I will work on it.

Wenhao-H:
I modified the code to print all pairs of neighbors:

    for (size_t j = 0; j < neighborhood.n_rows; ++j) {
      averages += w * h.col(neighborhood(j, i));
      std::cout << "(" << users(i) << "," << neighborhood(j ,i) << ")" << std::endl;
    }

and run a simple program:

```
include <mlpack/methods/cf/cf.hpp>

using namespace mlpack::cf;
using namespace mlpack::amf;
using namespace mlpack;

int main() {
  arma::mat dataset;
  data::Load("./data/GroupLensSmall.csv", dataset);
  CF cf(dataset, amf::NMFALSFactorizer(), 3);
  arma::Mat<size_t> recommendations;
  cf.GetRecommendations(5, recommendations);
  return 0;
}
```

And the printed result shows that each user is selected as the nearest neighbor of itself:
```
(0,0)
(0,58)
(0,6)
(1,1)
(1,73)
(1,191)
(2,2)
(2,3)
(2,126)
(3,3)
(3,2)
(3,126)
(4,4)
(4,76)
(4,24)
(5,5)
(5,9)
(5,84)
```

rcurtin:
Thanks for confirming that this is an issue, I appreciate that.  At this point if you want to fix it we can definitely merge a PR for that.  I would say the easiest way to fix this might be to search for `numUsersForSimilarity + 1`, then simply ignore the first row of results.  Modifying `NeighborSearch` for that would be, I think, a little too difficult and the runtime gain would likely be negligible at best.  But maybe you have another solution in mind, so don't feel restricted to what I said. :)

Wenhao-H:
I agree with your solution. That's what I had in mind too. I will fix it and make a PR:)

[...]

PR #1251, opened 2018-02-15
fix cf neighbor selection bug

Wenhao-H:
fix issue #1243

### DIFF
--- a/src/mlpack/methods/cf/cf.cpp
+++ b/src/mlpack/methods/cf/cf.cpp
@@ -78,10 +78,13 @@ void CF::GetRecommendations(const size_t numRecs,
   arma::Mat<size_t> neighborhood;
 
   // Calculate the neighborhood of the queried users.
+  // A queried user should be excluded from its list of neighbors,
+  // so numUsersForSimliarity + 1 neighbors will be selected, and then
+  // the queried user, which is the neareat neighbor, will be ignored. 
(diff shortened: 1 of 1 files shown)


rcurtin:
Hi Wenhao,

Thanks for the contribution!  I spent a little while thinking about it and I am actually not sure there is an issue to begin with.  I'm sorry I didn't realize this earlier.  Let me know what you think:

When we do the KNN to find similar users, the goal here is to normalize our predictions over similar users.  Since we have decomposed the sparse data matrix and reassembled (via NMF or whatever the decomposition we use is), we actually already have recommendations for we can use for the query user, since the reconstructed matrix will have values for the items that the query user did not rank.  However, these can be noisy, so the idea of performing the KNN to find similar users is to average the values over a number of users (including the query user) that the query user is similar to.

But in this, we should still be considering the query user in the search.  Does that make sense?  Let me know if I can clarify or if you disagree.

Wenhao-H:
Hi Ryan,

Yeah your explanation makes sense.  Actually I had this confusion when I started modifying the code. Thanks for clarification!

You can close the PR and issue :)

rcurtin:
Sounds good.  I clarified the comments in 796acc55 so hopefully we won't get confused about it again. :)  If you find any other problems, please do not hesitate to file an issue!

PR #1252, opened 2018-02-16
Mixin classes for optimizers, FunctionType API checks, and speedups

rcurtin:
This is a continuation of #1141; @shikharbhardwaj and @mentekid, if you find anything you don't like or that I took the wrong way from your ideas let me know and we can change it. :)

Basically, the problem I was originally trying to solve was that a lot of code in mlpack can be accelerated if the `Evaluate()` and `Gradient()` calls can be combined.  This is especially true in logistic regression and for training deep neural networks.  Therefore, I wanted users to have a way to implement `EvaluateWithGradient()` and a way for optimizers to depend on that function, but I didn't want to force it to be a part of the `FunctionType` API (I don't want that to be too large).

So instead I used template metaprogramming to create a wrapper `Function<FunctionType>` class that has "mixin" classes that add overloads of `EvaluateWithGradient()` if they don't exist.  Thus if the user writes `Evaluate()` and `Gradient()`, then viola, an overload of `EvaluateWithGradient()` will be available in the `Function<FunctionType>` class.  Similarly if a user writes `EvaluateWithGradient()` but not `Evaluate()` or `Gradient()`, then both `Evaluate()` and `Gradient()` will be available in `Function<FunctionType>` (though there may be some slowdown depending on the `FunctionType`).  If a user has implemented all three, the mixins do nothing.

The mixin framework allows an easy way to add more mixin functions should we need later.  For now I just add `EvaluateWithGradient()` when I can, and I adapted the existing optimizers to take advantage of that.  For logistic regression, this can give a 20-40% speedup per evaluation, which is nice.  There are also a few speedups for L-BFGS in here.

I also refactored the FunctionType checks from #1141 to fit with the mixin functionality that I wrote.

Lastly, I updated the new optimizer tutorial to talk about all the possible different kinds of FunctionTypes a user can implement.

I know it is a big set of changes... it turned out to be a lot more code than I was hoping (but I guess that is always true...).


src/mlpack/core/optimizers/cmaes/cmaes.hpp
```diff
@@ -166,8 +166,7 @@ class CMAES
 /**
  * Convenient typedef for CMAES approximation.
  */
-template <typename SelectionPolicyType = RandomSelection>
-using ApproxCMAES = CMAES<SelectionPolicyType>;
+typedef CMAES<RandomSelection> ApproxCMAES;
```
zoq: What I really like about ``using`` is that you can use ``ApproxCMAES``, ``ApproxCMAES<>`` or ``ApproxCMAES<..>``.

src/mlpack/methods/logistic_regression/logistic_regression_function_impl.hpp
```diff
@@ -235,6 +225,78 @@ void LogisticRegressionFunction<MatType>::PartialGradient(
...
+      predictors.cols(begin, begin + batchSize - 1).t() + regularization;
+
+  // Now compute the objective function using the sigmoids.
+  arma::rowvec respD = arma::conv_to<arma::rowvec>::from(responses.subvec(begin,
+      begin + batchSize - 1));
+  const double result = arma::accu(arma::log(1.0 - respD + sigmoids %
+      (2 * respD - 1.0)));
+
+  // Invert the result, because it's a minimization.
+  return -result + objectiveRegularization;
```
zoq: Should we write ``objectiveRegularization - result``?

src/mlpack/core/optimizers/lbfgs/test_functions.hpp
```diff
@@ -0,0 +1,175 @@
+/**
+ * @file test_functions.hpp
```
zoq: This function is now in the problems directory.

src/mlpack/core/optimizers/cmaes/cmaes.hpp
rcurtin: Ah, I had not thought about that.  Reverted.

src/mlpack/methods/logistic_regression/logistic_regression_function_impl.hpp
rcurtin: Good point, that is easier to read.  Fixed.

src/mlpack/core/optimizers/lbfgs/test_functions.hpp
rcurtin: Oops, I think this was an accidental change during the merge process.  I've removed the files.

rcurtin:

rcurtin:
Ok, since it passes all the tests (except the travis build that timed out, but I think that is random), I will go ahead and merge this in 7 days to leave time for any comments.

PR #1253, opened 2018-02-16
Fix random seed for tests

robertohueso:
I'm not sure if this is right but as far as I have tested seems fine, looking forward for your feedback :smile: 
Issue #1227

### DIFF
--- a/src/mlpack/tests/CMakeLists.txt
+++ b/src/mlpack/tests/CMakeLists.txt
@@ -143,6 +143,7 @@ add_executable(mlpack_test
   main_tests/random_forest_test.cpp
   main_tests/softmax_regression_test.cpp
   main_tests/sparse_coding_test.cpp
+  main_tests/hoeffding_tree_test.cpp
 )
 
 # Link dependencies of test executable.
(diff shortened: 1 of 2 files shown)


ShikharJ:
@robertohueso Welcome to mlpack! It looks like you have opened a PR for an issue that is already being worked upon by @Prabhat-IIT in #1236. I think it might be worthwhile to talk to him before continuing to work on this :)

Prabhat-IIT:
@robertohueso have you tested this fix? I've tried this already and it didn't work for me because `RandomSeed()` is not being called from `mlpack_preprocess_split` in this case. Please ensure that `RandomSeed()` is being called from `mlpack_preprocess_split` and not being called from `preprocess_split_test` either through debugging or by printing on the command line.

robertohueso:
@ShikharJ Thank you! I didn't notice :)
@Prabhat-IIT I have been testing today and it's true it's not being called from `mlpack_preprocess_split`. Seems like `mlpack_preprocess_split` "BINDING_TYPE" is "BINDING_TYPE_TEST".
1. It should be "BINDING_TYPE_CLI " right?
2. Should we open another issue?

Prabhat-IIT:
@robertohueso  there's a Macro conflict here!! I'll fix it after 3 days after my exams are over.

rcurtin:
Hi @robertohueso, can you reopen this PR?  I glanced at the code (it appears to be gone now) and if I am remembering the code correctly, this is the right way to solve the issue.  So if you can reopen this, we can figure out if it is working correctly and go from there.  Thanks!

PR #1255, opened 2018-02-17
minor changes in comment

luffy1996:
Changing the comments to go inline with the initialisation .

### DIFF
--- a/src/mlpack/core/optimizers/problems/sphere_function.cpp
+++ b/src/mlpack/core/optimizers/problems/sphere_function.cpp
@@ -22,7 +22,7 @@ SphereFunction::SphereFunction(const size_t n) :
 {
   initialPoint.set_size(n, 1);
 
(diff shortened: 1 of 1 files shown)


luffy1996:
I believe this commit doesn't have anything to do with build errors.

luffy1996:
@zoq , Can you please check the errors.

ShikharJ:
@luffy1996 As far as comment related changes are concerned, you can just reword the commit as 
```
[skip ci] minor changes....
```
That way, when you push, the CI builds will not be triggered.

zoq:
Personally I don't use ``[skip ci]`` to keep the commit log cleaner. Anyway, no worries if a build fails for some unrelate reasons, there is no restriction that we can only merge if every build is green.

luffy1996:
Thanks @rcurtin  , @zoq :)

zoq:
Ok, with two approvals I will go ahead and merge this now. Thank you for the contribution! :)

PR #1256, opened 2018-02-18
Rectify Decision Tree Bindings Test Implementation

manish7294:
Rectify error in DecisionTreeMinimumLeafSizeTest pointed out by @akhandait in #1174


src/mlpack/tests/main_tests/decision_tree_test.cpp
```diff
@@ -152,6 +152,11 @@ BOOST_AUTO_TEST_CASE(DecisionTreeMinimumLeafSizeTest)
 
   // Initialize an all-ones weight matrix.
   arma::mat weights(1, labels.n_cols, arma::fill::ones);
+ 
+  // Input training data.
+  SetInputParam("training", std::move(inputData));
```
akhandait: The test now throws a FATAL error because the input parameter training is supposed to be a tuple containing the info and the inputData.

manish7294: oops! I overlooked changes made by @haritha1313

zoq:

PR #1257, opened 2018-02-18
Fix serialization issue for boost v1.58

PlantsAndBuildings:
**TL,DR**: This PR backports `boost/serialization/vector.hpp` for boost 1.58 (and earlier) because of various failing tests related to different serialization behaviors in v1.58 and later versions.

In boost v1.58 when objects of type `std::vector<arma::mat>` are loaded, the vector gets appended to rather than being cleared and rewritten. See [this](https://stackoverflow.com/questions/34042933/boost-vector-serialization-append-issue) question on SO to understand the issue. As a result of this, several tests (`PerceptronSerializationTest`, `DecisionStumpSerializationTest`, `GMMLoadSaveTest` etc) failed while running mlpack_test with boost 1.58. Another problem that this behavior of v1.58 created (which I pointed out on IRC - [link](http://mlpack.org/irc/mlpack.20180207.html)) was that training an HMM using `mlpack_hmm_train` and then trying to use the remaining HMM utilities on the saved model (such as `mlpack_hmm_loglik`) gave errors -- which the tests did not report (maybe the tests need to be more rigorous?).

The first fix I considered was trying to find all instances where a vector of `arma::mat`s is trying to be loaded - and clear the vector before the loading step. This is tedious - and since the tests were not reporting all possible errors (as in the case of `mlpack_hmm_train` and subsequent `mlpack_hmm_loglik`) - not guaranteed to be foolproof.

Later, while working on the tests for the `mlpack_hmm_train` CLI binding, @rcurtin suggested backporting an armadillo function (for some other issue specific to that code). I found that `unordered_map` in `boost/serialization` was already backported for boost versions 1.56 and earlier. Using that as a template, I backported `boost/serialization/vector.hpp` from v1.62 for v1.58 and earlier.

Now,
1. All tests should pass for boost version v1.58 and v1.62 and armadillo 8.400 (thats what I'm running on my system)
2. The `hmm_train` followed by `hmm_loglik` should work without errors on v1.58

Sorry for the painfully detailed comment - I've been struggling with this for quite some time and telling the story is, well .. cathartic.

### DIFF
--- a/src/mlpack/core/boost_backport/boost_backport_serialization.hpp
+++ b/src/mlpack/core/boost_backport/boost_backport_serialization.hpp
@@ -24,5 +24,13 @@
   #include <boost/serialization/unordered_map.hpp>
 #endif
 
+#if BOOST_VERSION <= 105800
+  #include "mlpack/core/boost_backport/collections_load_imp.hpp"
+  #include "mlpack/core/boost_backport/collections_save_imp.hpp"
+  #include "mlpack/core/boost_backport/vector.hpp"
+#else
+  #include <boost/serialization/vector.hpp>
+#endif
+
 #endif // MLPACK_CORE_BOOST_BACKPORT_SERIALIZATION_HPP
 
--- a/src/mlpack/core/boost_backport/boost_backport_serialization.hpp
+++ b/src/mlpack/core/boost_backport/boost_backport_serialization.hpp
@@ -6,6 +6,7 @@
  * following boost functionality here:
  *
  *  * unordered_set serialization support (added in boost 1.56.0)
+ *  * vector serialization (changed after boost 1.58.0)
  *
  * If the detected boost version is greater than 1.58.0, we include the normal
  * serialization functions (not the backported ones).  For all older versions we
@@ -25,6 +26,14 @@
 #endif
 
 #if BOOST_VERSION <= 105800
+  /**
+   * Boost versions 1.58.0 and earlier have a different vector serialization
+   * behaivor as compared to later versions. Notably, loading a
+   * std::vector<arma::mat> does not clear the vector before the load
+   * in v1.58 and earlier; while in the later versions, the vector is cleared
+   * before loading. This causes some tests related to serialization to fail
+   * with versions 1.58. This backport solves the issue.
+   */
   #include "mlpack/core/boost_backport/collections_load_imp.hpp"
   #include "mlpack/core/boost_backport/collections_save_imp.hpp"
   #include "mlpack/core/boost_backport/vector.hpp"
--- a/src/mlpack/core/boost_backport/boost_backport_serialization.hpp
+++ b/src/mlpack/core/boost_backport/boost_backport_serialization.hpp
@@ -25,7 +25,7 @@
   #include <boost/serialization/unordered_map.hpp>
 #endif
 
-#if BOOST_VERSION <= 105800
+#if BOOST_VERSION == 105800
   /**
    * Boost versions 1.58.0 and earlier have a different vector serialization
    * behaivor as compared to later versions. Notably, loading a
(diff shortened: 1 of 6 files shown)


src/mlpack/core/boost_backport/boost_backport_serialization.hpp
```diff
@@ -24,5 +24,13 @@
   #include <boost/serialization/unordered_map.hpp>
 #endif
 
+#if BOOST_VERSION <= 105800
```
zoq: Form the file:

```
/**
 * @file boost_backport.hpp
 * @author Yannis Mentekidis
 *
 * Centralized control of what boost files to include. We have backported the
 * following boost functionality here:
 *
 *  * unordered_set serialization support (added in boost 1.56.0)
 *
 * If the detected boost version is greater than 1.58.0, we include the normal
 * serialization functions (not the backported ones).  For all older versions we
 * include the backported headers.
 */
```

would be great if you could add the same for the vector backport.

src/mlpack/prereqs.hpp
```diff
@@ -72,7 +72,6 @@ using enable_if_t = typename enable_if<B, T>::type;
 // defined, but we still need to define it (as nothing) so that the mlpack
 // serialization shim compiles.
 #include <boost/serialization/serialization.hpp>
-#include <boost/serialization/vector.hpp>
```
rcurtin: It's fine to remove this from here, but I think it would be useful to add a comment pointing out why we are including `vector.hpp` somewhere else.

src/mlpack/core/boost_backport/boost_backport_serialization.hpp
```diff
@@ -24,5 +25,21 @@
...
+   * Boost versions 1.58.0 and earlier have a different vector serialization
+   * behaivor as compared to later versions. Notably, loading a
+   * std::vector<arma::mat> does not clear the vector before the load
+   * in v1.58 and earlier; while in the later versions, the vector is cleared
+   * before loading. This causes some tests related to serialization to fail
+   * with versions 1.58. This backport solves the issue.
+   */
+  #include "mlpack/core/boost_backport/collections_load_imp.hpp"
+  #include "mlpack/core/boost_backport/collections_save_imp.hpp"
+  #include "mlpack/core/boost_backport/vector.hpp"
```
rcurtin: If a user happens to do something like this:

```
#include <boost/serialization/vector.hpp>
#include <mlpack/core.hpp>
```

and the Boost version is 1.58, then this will actually not fix the issue because `vector.hpp` is included before we have a chance to include it properly.  I think it would be useful to add a warning if the Boost version is 1.58 and `vector.hpp` is already included; that could say something a little like the Armadillo warning in `mlpack/prereqs.hpp`.  What do you think?

PlantsAndBuildings:
The `BOOST_VERSION <= 105800` was just to be on the safer side. I have not verified for myself whether the vector append behavior is unique only to 1.58 or is applicable to older versions as well (as you see on the launchpad issue, even the person who has posted has not tried it for versions 1.55 to 1.57). On the other hand, we know for sure that the behavior has definitely changed v1.59 onwards because a boost dev said so - so, yeah, it was a safe bet going for <= 105800.

I guess we will find out for sure when we do the hand-checking -- so I'll go ahead and change the <= to ==. In any case, does it really affect performance if we use "native" serialization instead of backported serialization?

src/mlpack/core/boost_backport/boost_backport_serialization.hpp
PlantsAndBuildings: Ah, you're right. I completely missed that. I don't suppose there is some way we can enforce the order of includes (as opposed to just giving a warning - which most people completely ignore anyway)? I ask because you've faced exactly the same issue with armadillo. I'll add the warning for now and look into whether there is some way to enforce the include of backported vector.hpp (and no other vector.hpp) if the detected boost version is 1.58.

rcurtin:
I hand-checked the code with Boost 1.57, 1.58, and 1.59, and in all cases the tests pass.  I don't think it affects performance if we use the backported vs. native serialization, but, I am not sure if the backported code will work with earlier Boost versions (like 1.50 or before); maybe something in the serialization code changed.  So it's probably better to be on the safe side and backport only for Boost 1.58.

rcurtin:
With two approvals I will go ahead and merge this.  Thanks for the contribution! :)

PR #1258, opened 2018-02-18
add non-negative check for nmf_test

Wenhao-H:
To make sure all elements in W and H matrices are non-negative.

### DIFF
--- a/src/mlpack/tests/nmf_test.cpp
+++ b/src/mlpack/tests/nmf_test.cpp
@@ -222,4 +222,65 @@ BOOST_AUTO_TEST_CASE(SparseNMFALSTest)
       1e-5);
 }
 
-BOOST_AUTO_TEST_SUITE_END();
+/**
+ * Check if all elements in W and H are non-negative.
+ * Default Case.
+ * Random Acol initilization and distance minimization update.
+ */
+BOOST_AUTO_TEST_CASE(NonNegNMFDefaultTest)
+{
+  mat w = randu<mat>(20, 12);
+  mat h = randu<mat>(12, 20);
+  mat v = w * h;
+  const size_t r = 12;
+
+  AMF<> nmf;
+  nmf.Apply(v, r, w, h);
+
+  BOOST_REQUIRE(arma::all(vectorise(std::move(w)) >= 0)
+              && arma::all(vectorise(std::move(h)) >= 0));
+}
+
+/**
+ * Check if all elements in W and H are non-negative.
+ * Random initialization divergence minimization update.
+ */
+BOOST_AUTO_TEST_CASE(NonNegNMFRandomDivTest)
+{
+  mat w = randu<mat>(20, 12);
+  mat h = randu<mat>(12, 20);
+  mat v = w * h;
+  const size_t r = 12;
+
+  // Custom tighter tolerance.
+  SimpleResidueTermination srt(1e-8, 10000);
+  AMF<SimpleResidueTermination,
+      RandomInitialization,
+      NMFMultiplicativeDivergenceUpdate> nmf(srt);
+  nmf.Apply(v, r, w, h);
+
+  BOOST_REQUIRE(arma::all(vectorise(std::move(w)) >= 0)
+              && arma::all(vectorise(std::move(h)) >= 0));
+}
+
+/**
+ * Check if all elements in W and H are non-negative.
+ * Random initialization, alternating least squares update.
+ */
+BOOST_AUTO_TEST_CASE(NonNegNMFALSTest)
+{
+  mat w = randu<mat>(20, 12);
+  mat h = randu<mat>(12, 20);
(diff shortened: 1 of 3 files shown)


src/mlpack/tests/nmf_test.cpp
```diff
@@ -222,4 +222,65 @@ BOOST_AUTO_TEST_CASE(SparseNMFALSTest)
...
+{
+  mat w = randu<mat>(20, 12);
+  mat h = randu<mat>(12, 20);
+  mat v = w * h;
+  const size_t r = 12;
+
+  AMF<> nmf;
+  nmf.Apply(v, r, w, h);
+
+  BOOST_REQUIRE(arma::all(vectorise(std::move(w)) >= 0)
```
zoq: No need to use ``move`` here, this should make the line shorter and more readable.

src/mlpack/tests/nmf_test.cpp
```diff
@@ -222,4 +222,65 @@ BOOST_AUTO_TEST_CASE(SparseNMFALSTest)
...
+BOOST_AUTO_TEST_CASE(NonNegNMFALSTest)
+{
+  mat w = randu<mat>(20, 12);
+  mat h = randu<mat>(12, 20);
+  mat v = w * h;
+  const size_t r = 12;
+
+  SimpleResidueTermination srt(1e-12, 50000);
+  AMF<SimpleResidueTermination, RandomAcolInitialization<>, NMFALSUpdate>
+        nmf(srt);
```
zoq: When wrapping a line, the next line should be tabbed twice from where the previous line began.

Wenhao-H:
@zoq style fixed! Thanks for reviewing :)

src/mlpack/tests/nmf_test.cpp
```diff
@@ -222,4 +222,66 @@ BOOST_AUTO_TEST_CASE(SparseNMFALSTest)
...
+  const size_t r = 12;
+
+  SimpleResidueTermination srt(1e-12, 50000);
+  AMF<SimpleResidueTermination,
+      RandomAcolInitialization<>,
+      NMFALSUpdate> nmf(srt);
+  nmf.Apply(v, r, w, h);
+
+  BOOST_REQUIRE(arma::all(vectorise(w) >= 0)
+      && arma::all(vectorise(h) >= 0));
```
rcurtin: Just a minor note---I don't think that `vectorise()` is needed; if I'm wrong and it is, could you write the more explicit `arma::vectorise()` please?  Some compilers won't accept the plain `vectorise()` since we do not have `using namespace arma;` specified.  Thanks!

src/mlpack/tests/nmf_test.cpp
```diff
@@ -222,4 +222,66 @@ BOOST_AUTO_TEST_CASE(SparseNMFALSTest)
...
+{
+  mat w = randu<mat>(20, 12);
+  mat h = randu<mat>(12, 20);
+  mat v = w * h;
+  const size_t r = 12;
+
+  // Custom tighter tolerance.
+  SimpleResidueTermination srt(1e-8, 10000);
+  AMF<SimpleResidueTermination,
+      RandomInitialization,
```
rcurtin: Any reason for not using `RandomAcolInitialization` here?

Wenhao-H: Because in nmf_main.cpp, the factorizer for `updateRules == "multdiv"` uses `RandomInitialization`.

src/mlpack/tests/nmf_test.cpp
Wenhao-H: `vectorise()` is needed because for a matrix `w`, `arma::all(w > 0)` returns a vector, as explained in armadillo document: http://arma.sourceforge.net/docs.html#all
I will change it to `arma::vectorise`.

Wenhao-H:
Hi Ryan, I've already made the changes as your comments suggest. And yes, I would like to get the mlpack stickers :) Thanks for reviewing!

src/mlpack/tests/nmf_test.cpp
zoq: ``all(mat)`` returns ``umat``, so in this case I think ``vectorise`` is an easy solution.

rcurtin: Yep, thank you for the clarification; I see now that the `vectorise()` is needed.

src/mlpack/tests/nmf_test.cpp
rcurtin: Ok, thanks for the clarification.  I should have looked in `nmf_main.cpp` to answer my own question, sorry about that.

Wenhao-H: don't worry, I'm happy to help!

rcurtin:

PR #1260, opened 2018-02-19
fix maxIterations=0 bug in SimpleResidueTermination

Wenhao-H:
In SimpleResidueTermination, there should be no limit on the number of iterations if maxIterations == 0:
https://github.com/mlpack/mlpack/blob/617a2acb52617e98f681a34137a7381e7c017db2/src/mlpack/methods/amf/termination_policies/simple_residue_termination.hpp#L31-L45
But the IsConverged() function ignores this in its return statement:
https://github.com/mlpack/mlpack/blob/617a2acb52617e98f681a34137a7381e7c017db2/src/mlpack/methods/amf/termination_policies/simple_residue_termination.hpp#L68-L86
which leads to nmf's immediate termination:
```
➜  build git:(master) mlpack_nmf --input_file test_data_3_1000.csv --rank 2 --verbose --max_iterations 0
[WARN ] Should pass either '--h_file (-H)' or '--w_file (-W)' or both; no output will be saved!
[INFO ] Loading 'test_data_3_1000.csv' as CSV data.  Size is 3 x 1000.
[INFO ] Performing NMF with multiplicative distance-based update rules.
[INFO ] Initialized W and H.
[INFO ] Iteration 2; residue inf.
[INFO ] AMF converged to residue of inf in 2 iterations.
```
So I used a simple modification to fix the bug.

### DIFF
--- a/src/mlpack/methods/amf/termination_policies/simple_residue_termination.hpp
+++ b/src/mlpack/methods/amf/termination_policies/simple_residue_termination.hpp
@@ -82,7 +82,9 @@ class SimpleResidueTermination
     Log::Info << "Iteration " << iteration << "; residue " << residue << ".\n";
 
     // Check if termination criterion is met.
-    return (residue < minResidue || iteration > maxIterations);
+    // If maxIterations == 0, there is no iteration limit.
+    return (residue < minResidue || 
+      ((maxIterations != 0) && (iteration > maxIterations)));
   }
 
   //! Get current value of residue
--- a/src/mlpack/methods/amf/termination_policies/simple_residue_termination.hpp
+++ b/src/mlpack/methods/amf/termination_policies/simple_residue_termination.hpp
@@ -83,7 +83,7 @@ class SimpleResidueTermination
 
     // Check if termination criterion is met.
     // If maxIterations == 0, there is no iteration limit.
-    return (residue < minResidue || iteration == maxIterations)
+    return (residue < minResidue || iteration == maxIterations);
   }
 
   //! Get current value of residue
--- a/src/mlpack/methods/amf/termination_policies/simple_residue_termination.hpp
+++ b/src/mlpack/methods/amf/termination_policies/simple_residue_termination.hpp
@@ -53,7 +53,7 @@ class SimpleResidueTermination
   {
     // Initialize the things we keep track of.
     residue = DBL_MAX;
-    iteration = 1;
+    iteration = 0;
     nm = V.n_rows * V.n_cols;
     // Remove history.
     normOld = 0;
@@ -84,7 +84,7 @@ class SimpleResidueTermination
     // Check if termination criterion is met.
     // If maxIterations == 0, there is no iteration limit.
     return (residue < minResidue || 
-        (maxIterations != 0 && iteration > maxIterations));
+        (maxIterations != 0 && iteration >= maxIterations));
   }
 
   //! Get current value of residue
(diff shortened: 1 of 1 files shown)


src/mlpack/methods/amf/termination_policies/simple_residue_termination.hpp
```diff
@@ -82,7 +82,9 @@ class SimpleResidueTermination
     Log::Info << "Iteration " << iteration << "; residue " << residue << ".\n";
 
     // Check if termination criterion is met.
-    return (residue < minResidue || iteration > maxIterations);
+    // If maxIterations == 0, there is no iteration limit.
+    return (residue < minResidue || 
+      ((maxIterations != 0) && (iteration > maxIterations)));
```
zoq: When wrapping a line, the next line should be tabbed twice from where the previous line began. Also we can remove the brackets around ``maxIterations != 0`` and ``iteration > maxIterations``.

Wenhao-H:
@zoq style fixed! Thanks for reviewing :)

rcurtin:
You can also simplify the condition a little to just write `iteration != maxIterations` and this will work if `maxIterations` is 0, assuming that `iteration` is initially set to 1.

Wenhao-H:
@rcurtin But it doesn't work when `maxIterations` is not 0 because it will cause `isConverged()` to return true after first iteration.

src/mlpack/methods/amf/termination_policies/simple_residue_termination.hpp
```diff
@@ -53,7 +53,7 @@ class SimpleResidueTermination
   {
     // Initialize the things we keep track of.
     residue = DBL_MAX;
-    iteration = 1;
+    iteration = 0;
```
Wenhao-H: I changed the initial `iteration` to 0, so the iteration number printed by `Log::Info` below will start from `1`. If `iteration` is initially set to 1, the starting iteration number would be `2`.

rcurtin: Ah, ok, I see.  The trick we do for other optimizers does not work in this case because `iteration++` must come before the return statement.  Another option might be to remove the `iteration++` from line 81 and embed it into the conditional:

```
return (residue < minResidue || iteration++ != maxIterations);
```

but I am not picky, I think whatever you choose is ok.

Wenhao-H: Do you mean `return (residue < minResidue || iteration++ >= maxIterations);`? Because like I mentioned in a previous comment, `iteration != maxIterations` doesn't work properly when `maxIterations != 0`.
And I think keeping `iteration++` in line 81 is more clear and readable.

rcurtin: Ack, sorry, the condition would actually be `iteration++ == maxIterations`.  Anyway, I can agree that `iteration++` in line 81 is more clear, so we can just keep it like that.

Wenhao-H: What if `iteration` overflows and becomes 0?
But it seems very unlikely...

rcurtin: Yep, we are making the assumption in the other code that the user isn't going to be patient enough to wait for 2^64 iterations (or 2^32 depending on architecture).

Wenhao-H: modified, should be ready to merge now.

Wenhao-H:
@zoq I added my name in another PR #1258 , and I have mailed Ryan for the stickers. Thanks :)

rcurtin:
With two approvals I will go ahead and merge this now. Thanks again. :)

PR #1261, opened 2018-02-19
add nmf binding test

Wenhao-H:
Added the nmf_test.cpp for nmf binding. (related issue: #1152)

I also found a bug in the process. I opened a dedicated PR(#1260) for that.

### DIFF
--- a/src/mlpack/methods/amf/init_rules/given_init.hpp
+++ b/src/mlpack/methods/amf/init_rules/given_init.hpp
@@ -56,17 +56,29 @@ class GivenInitialization
   {
     // Make sure the initial W, H matrices have correct size.
     if (w.n_rows != V.n_rows)
-      Log::Fatal << "The number of rows in given W (" <<  w.n_rows 
-          << ") doesn't equal the number of rows in V (" << V.n_rows << ") !" << std::endl;
+    {
+      Log::Fatal << "The number of rows in given W (" <<  w.n_rows
+          << ") doesn't equal the number of rows in V (" << V.n_rows
+          << ") !" << std::endl;
+    }
     if (w.n_cols != r)
-      Log::Fatal << "The number of columns in given W (" <<  w.n_cols 
-          << ") doesn't equal the rank of factorization (" << r << ") !" << std::endl;
+    {
+      Log::Fatal << "The number of columns in given W (" <<  w.n_cols
+          << ") doesn't equal the rank of factorization (" << r
+          << ") !" << std::endl;
+    }
     if (h.n_cols != V.n_cols)
+    {
       Log::Fatal << "The number of columns in given H (" <<  h.n_cols
-          << ") doesn't equal the number of columns in V (" << V.n_cols << ") !" << std::endl;
+          << ") doesn't equal the number of columns in V (" << V.n_cols
+          << ") !" << std::endl;
+    }
     if (h.n_rows != r)
+    {
       Log::Fatal << "The number of rows in given H (" <<  h.n_rows
-          << ") doesn't equal the rank of factorization (" << r << ") !" << std::endl;
+          << ") doesn't equal the rank of factorization (" << r
+          << ") !"<< std::endl;
+    }
 
     // Initialize to the given matrices.
     W = w;
--- a/src/mlpack/methods/amf/init_rules/given_init.hpp
+++ b/src/mlpack/methods/amf/init_rules/given_init.hpp
@@ -49,11 +49,25 @@ class GivenInitialization
    * @param H H matrix, to be filled with random noise.
    */
   template<typename MatType>
-  inline void Initialize(const MatType& /* V */,
-                         const size_t /* r */,
+  inline void Initialize(const MatType& V,
+                         const size_t r,
                          arma::mat& W,
                          arma::mat& H)
   {
+    // Make sure the initial W, H matrices have correct size.
+    if (w.n_rows != V.n_rows)
+      Log::Fatal << "The number of rows in given W (" <<  w.n_rows 
+          << ") doesn't equal the number of rows in V (" << V.n_rows << ") !" << std::endl;
+    if (w.n_cols != r)
+      Log::Fatal << "The number of columns in given W (" <<  w.n_cols 
+          << ") doesn't equal the rank of factorization (" << r << ") !" << std::endl;
+    if (h.n_cols != V.n_cols)
+      Log::Fatal << "The number of columns in given H (" <<  h.n_cols
+          << ") doesn't equal the number of columns in V (" << V.n_cols << ") !" << std::endl;
+    if (h.n_rows != r)
+      Log::Fatal << "The number of rows in given H (" <<  h.n_rows
+          << ") doesn't equal the rank of factorization (" << r << ") !" << std::endl;
+
     // Initialize to the given matrices.
     W = w;
     H = h;
--- a/src/mlpack/methods/nmf/nmf_main.cpp
+++ b/src/mlpack/methods/nmf/nmf_main.cpp
@@ -95,6 +95,8 @@ static void mlpackMain()
       "the rank of the factorization must be greater than 0");
   RequireParamInSet<string>("update_rules", { "multdist", "multdiv", "als" },
       true, "unknown update rules");
+  RequireParamValue<int>("max_iterations", [](int x) { return x >= 0; },
+      true, "max_iterations must be non-negative");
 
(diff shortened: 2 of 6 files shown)


Wenhao-H:
Yes I can add support for that. I understand why we want to give user the choice to specify initial `W, H `matrices. But can you clarify why we need to test that `min_residue` and `max_iterations` make a difference to the program execution?

rcurtin:
In these tests we basically just want to ensure that the user can specify each option, and that each option actually makes a difference.  So running NMF twice with different `min_residue` parameters can tell us if the binding is actually using the value of `min_residue`; the same is true for `max_iterations`.  There's no need to check the validity of the result, just that passing different parameters gave a different output.

Wenhao-H:
Understood. I am working on theses tests.

Wenhao-H:
Things added in these new commits are:

1. Add support for specifying initial `W, H` matrices in nmf_main.cpp. The corresponding cli parameters are `initial_w` and `initial_h`.
2. New param check function: `RequireNoneOrAllPassed()`. It requires that either none or all parameters in constraints set are passed. This is used in nmf binding to make sure that either none or both of `initial_w` and `initial_h` are passed.
3. Add matrices size check in `Initialize()` in `class GivenInitialization`. This is to make sure that the sizes of specified initial `W, H` matrices are valid.
4. Tests that make sure `min_residue` and `max_iterations` make difference to program execution.

Let me know where I can improve:)

src/mlpack/tests/main_tests/nmf_test.cpp
```diff
@@ -0,0 +1,294 @@
...
+
+/**
+ * Ensure min_residue is used, by testing that 
+ * min_resude makes a difference to the program.  
+ */
+BOOST_AUTO_TEST_CASE(NMFMinResidueTest)
+{
+  mat v = arma::randu(10, 10);
+  mat initial_w = arma::randu(10, 5);
+  mat initial_h = arma::randu(5, 10);
```
rcurtin: Sorry for the nitpicking, but could you name these `initialW` and `initialH` in accordance with the style guide?  (variable names shouldn't have underscores)

rcurtin:

src/mlpack/methods/amf/init_rules/given_init.hpp
```diff
@@ -49,11 +49,25 @@ class GivenInitialization
...
   template<typename MatType>
-  inline void Initialize(const MatType& /* V */,
-                         const size_t /* r */,
+  inline void Initialize(const MatType& V,
+                         const size_t r,
                          arma::mat& W,
                          arma::mat& H)
   {
+    // Make sure the initial W, H matrices have correct size.
+    if (w.n_rows != V.n_rows)
```
zoq: Do you mind to use ``{}`` for the multiline statement here, same for the following statements, to improve the readability?

src/mlpack/methods/nmf/nmf_main.cpp
```diff
@@ -111,28 +118,74 @@ static void mlpackMain()
...
     SimpleResidueTermination srt(minResidue, maxIterations);
-    AMF<> amf(srt);
-    amf.Apply(V, r, W, H);
+    if (CLI::HasParam("initial_w"))
+    {
+      // Initialization with given W, H matrices.
+      GivenInitialization ginit = GivenInitialization(
+          std::move(CLI::GetParam<arma::mat>("initial_w")),
+          std::move(CLI::GetParam<arma::mat>("initial_h"))
+          );
```
zoq: Should we close right after ``>("initial_h"))``?

src/mlpack/methods/nmf/nmf_main.cpp
```diff
@@ -111,28 +118,74 @@ static void mlpackMain()
...
+
     SimpleResidueTermination srt(minResidue, maxIterations);
-    AMF<SimpleResidueTermination,
+    if (CLI::HasParam("initial_w"))
+    {
+      // Initialization with given W, H matrices.
+      GivenInitialization ginit = GivenInitialization(
+          std::move(CLI::GetParam<arma::mat>("initial_w")),
+          std::move(CLI::GetParam<arma::mat>("initial_h"))
+          );
```
zoq: See comment above.

src/mlpack/methods/nmf/nmf_main.cpp
```diff
@@ -111,28 +118,74 @@ static void mlpackMain()
...
```
zoq: See comment above.

src/mlpack/tests/main_tests/nmf_test.cpp
```diff
@@ -0,0 +1,294 @@
...
+
+  // Rank should not be negative.
+  r = -1;
+  SetInputParam("input", std::move(v));
+  SetInputParam("rank", r);
+ 
+  Log::Fatal.ignoreInput = true;
+  BOOST_REQUIRE_THROW(mlpackMain(), std::runtime_error);
+  Log::Fatal.ignoreInput = false;
+
```
zoq: I think we can remove the extra line here.

src/mlpack/tests/main_tests/nmf_test.cpp
```diff
@@ -0,0 +1,294 @@
...
+  SetInputParam("input", v);
+  SetInputParam("rank", r);
+  SetInputParam("initial_w", initial_w);
+  SetInputParam("initial_h", initial_h);
+  
+  mlpackMain();
+
+  const mat w1 = CLI::GetParam<mat>("w");
+  const mat h1 = CLI::GetParam<mat>("h");
+
```
zoq: Looks like we can remove the extra line here.

src/mlpack/tests/main_tests/nmf_test.cpp
```diff
@@ -0,0 +1,294 @@
...
+
+  ResetSettings();
+
```
zoq: Looks like another extra line.

src/mlpack/tests/main_tests/nmf_test.cpp
```diff
@@ -0,0 +1,294 @@
...
+  SetInputParam("input", v);
+  SetInputParam("rank", r);
+  SetInputParam("initial_w", initial_w);
+  SetInputParam("initial_h", initial_h);
+
+  mlpackMain();
+
+  const mat w2 = CLI::GetParam<mat>("w");
+  const mat h2 = CLI::GetParam<mat>("h");
+
```
zoq: Should we remove the line here as well?

Wenhao-H:
style fixed! @rcurtin I would like to merge this PR first, and I will add support for allowing user to specify one of `H, W` in a later PR.

src/mlpack/tests/main_tests/nmf_test.cpp
```diff
@@ -0,0 +1,287 @@
...
+  SetInputParam("initial_w", initialW);
+  SetInputParam("initial_h", initialH);
+
+  mlpackMain();
+
+  const mat w2 = CLI::GetParam<mat>("w");
+  const mat h2 = CLI::GetParam<mat>("h");
+
+  // The resulting matrices should be different.
+  BOOST_REQUIRE(!arma::approx_equal(w1, w2, "absdiff", 1e-5)
```
zoq: ``approx_equal`` was added in armadillo 6.700 but the minimum version we require is 6.500.0; however ``CheckMatrices`` should work just fine.

rcurtin:
@Wenhao-H: that sounds fine to me.  @zoq is right, the `approx_equal()` issue is what is causing most of the checks to fail.  If you fix that I think all the builds will be successful. :)

src/mlpack/tests/main_tests/nmf_test.cpp
Wenhao-H: The `CheckMatrices` in test_tools.hpp tests that two matrices are approximately equal. Is there a function which tests two matrices are NOT approximately equal, as we need in this case?

rcurtin: Hm, you are right, there is not.  You could do something like this:

```
BOOST_REQUIRE_GT(arma::norm(w1 - w2), 1e-5)
```

(I haven't checked that but I think it would work.)

Wenhao-H: fixed. Thanks for helping:)

rcurtin:
Thanks for the contribution. :)

PR #1262, opened 2018-02-19
Add KNN binding tests

robertohueso:
Add first KNN binding test.
In reference to issue #1152

I will be adding more but since these are my first contributions to the project I want to make sure it's fine.


rcurtin:
Hi @robertohueso, I think you saw in #1152 that @akhandait was already working on the KNN bindings before this PR opened.  His PR #1263 actually has the same test in it that you have here, and since he "claimed" it before this PR was open, let's go ahead and resolve the issue by closing this PR and going with his tests.  Sorry for the confusion.  You could also write similar tests for `mlpack_krann`; this test here could be adapted for that as a starting point.  If that's what you'd like to do let me know and I'll update #1152 accordingly.

PR #1263, opened 2018-02-19
k-nearest(furthest)-neighbors binding tests

akhandait:
#1152
Here are the tests I have written so far.
However, knn_main has a lot of possible parameters and hence a lot more will be required.
Suggest me whatever tests I should add and I will add them as soon as possible.
@robertohueso Can you please find some other method to write tests for? :)

EDIT : 
With minimal changes in knn_test.cpp, I wrote kfn_test.cpp as the test requirements are almost the same. I will keep this updated along with additions in knn_test.

List of tasks to be completed:
- [x] Binding tests for knn_main
- [x] Binding tests for kfn_main
- [ ] Add a `dual-greedy` mode
- [x] Add `CheckMatricesNotEqual` function
- [x] True neighbors/distances bug fix


nikhilgoel1997:
Hello @akhandait! Thank you for your contribution. I'd like to suggest few functionalities that can be tested for knn:-

- Check that different search algorithm gives same result.
- Invalid search algorithm causes an error.
- Check that different type of trees in search algorithm gives same result.
- Invalid tree type causes an error.
- Invalid values of  `tau`, `rho` and `epsilon` causes and error.
- `Epsilon`, `tau` and `rho` should be checked to see if it has an effect on result.
- `leaf size` should be checked to see if it has an effect on result.

I might have missed a few tests, I'm not sure. I'll tell you if I remember any other functionalities that can be tested!

rcurtin:
Hi @akhandait, thanks for the contribution.  I think Nikhil's suggestions are accurate though like he pointed out they aren't necessarily comprehensive.  Would you like to implement those tests?  If so, I'll wait to review until we have those added.  Thanks!

akhandait:
@rcurtin I am working on the tests @nikhilgoel1997 suggested, you can review when I add those.
Thanks @nikhilgoel1997 for the suggestions. :)

akhandait:
Is epsilon >= 1 supposed to be allowed in knn?
It is mentioned in `kfn_main.cpp` that it has to be in the range `(0, 1]`, but even in kfn it is allowed with `>100` percent error.

akhandait:
According to the `KNNAllTreeTypesTest`, `spill` tree type gives wrong results. I am kind of reluctant to open so many issues for neighbor search, should I? Or should we just resolve this here?

akhandait:
Okay, so after looking into it, I found out that spill tree type is not an exact search, but what about those extremely large values it is throwing? the 14th entry has the first appearance of such values but they keep coming after that.

rcurtin:
> Is epsilon >= 1 supposed to be allowed in knn?
> It is mentioned in kfn_main.cpp that it has to be in the range (0, 1], but even in kfn it is allowed with >100 percent error.

Yes, the strategies for approximation are slightly different for KNN and KFN (if you sit down and write out the equations for what approximation means in each of these cases, it makes a little more sense).  For KNN, you can have any greater-than-0 epsilon (see `Relax()` in the sort policies for more information on how the approximate relaxation is applied; you might have to dig in the code a bit to see where `Relax()` is used so it makes sense).  For KFN though, the relaxation is `1 / (1 - epsilon) * distance`, so `epsilon` must be in the range `[0, 1)`.  It looks like the check in `kfn_main.cpp` is incorrect (I guess I copied the KNN checks but did not fully check it---this is why these tests are important! :))

> According to the KNNAllTreeTypesTest, spill tree type gives different results. Is this an issue?

Yes, this was mentioned to me last night by someone else.  I saw that the dual-tree spill tree search uses a defeatist search, which can be a little confusing for a user---all the other `dualtree` searches give exact results.  I might suggest that we modify the code a little bit and add a `dual-greedy` mode (or similarly named) which uses a dual-tree defeatist search (and possibly only works with spill trees), and force spill trees with `dualtree` search to use non-defeatist search.  There will be a little bit of code modification that has to happen for all that to work.  We can open that in a separate issue if you like and comment out the spill tree from KNNAllTreeTypesTest for now, if you prefer; just let me know.

src/mlpack/tests/main_tests/kfn_test.cpp
```diff
@@ -0,0 +1,378 @@
...
+  SetInputParam("query", queryData);
+  SetInputParam("k", (int) 10);
+
+  mlpackMain();
+
+  arma::Mat<size_t> neighbors;
+  arma::mat distances;
+  neighbors = std::move(CLI::GetParam<arma::Mat<size_t>>("neighbors"));
+  distances = std::move(CLI::GetParam<arma::mat>("distances"));
+
```
rcurtin: I think we need to call `CleanMemory()` before a second call of `mlpackMain()`, otherwise the program may have allocated a `KFNModel` that needs to be deleted.

src/mlpack/tests/main_tests/kfn_test.cpp
```diff
@@ -0,0 +1,378 @@
...
+ */
+BOOST_AUTO_TEST_CASE(KFNInvalidEpsilonTest)
+{
+  arma::mat referenceData;
+  referenceData.randu(3, 100); // 100 points in 3 dimensions.
+
+  // Random input, some k <= number of reference points.
+  SetInputParam("reference", std::move(referenceData));
+  SetInputParam("k", (int) 10);
+  SetInputParam("epsilon", (double) -1); // Invalid.
```
rcurtin: It's probably also worthwhile to either add another test or add another call to `mlpackMain()` where we check that `epsilon` is not 1 or greater.  (You are probably already thinking this based on the other comment I left about KFN approximation, but I thought it was a good idea to write it down so we don't forget about it. :))

src/mlpack/tests/main_tests/kfn_test.cpp
```diff
@@ -0,0 +1,378 @@
...
+  // {
+  //   CheckMatrices(neighbors[i], neighbors[i + 1]);
+  //   CheckMatrices(distances[i], distances[i + 1]);
+  // }
+  CheckMatrices(neighbors[0], neighbors[1]);
+  CheckMatrices(neighbors[1], neighbors[2]);
+  //CheckMatrices(neighbors[2], neighbors[3]);
+  CheckMatrices(distances[0], distances[1]);
+  CheckMatrices(distances[1], distances[2]);
+  //CheckMatrices(distances[2], distances[3]);
```
rcurtin: Yep, it's fine to avoid checking the greedy output here.  I think it might be useful just to make sure there are no `DBL_MAX`es in the `distances[3]` matrix though.

src/mlpack/tests/main_tests/kfn_test.cpp
```diff
@@ -0,0 +1,378 @@
...
+  CheckMatrices(distances, CLI::GetParam<arma::mat>("distances"));
+}
+
+/*
+ * Ensure that different search algorithms give same result.
+ */
+BOOST_AUTO_TEST_CASE(KFNAllAlgorithmsTest)
+{ 
+  string algorithms[] = {"dual_tree", "naive", "single_tree", "greedy"};
+  int nof_algorithms = sizeof(algorithms)/sizeof(algorithms[0]);
```
rcurtin: This is a minor style comment, but can you make sure that variable names don't have underscores in them like the style guide specifies?  https://github.com/mlpack/mlpack/wiki/DesignGuidelines#style-guidelines

So `nofAlgorithms`, for example.  I think there are other places in this code where the same issue applies, so if you can take a look through and fix it I'd appreciate it.  Also horizontal whitespace in this line: `sizeof(algorithms) / sizeof(algorithms[0])`.

akhandait:
Will it be okay if I make those changes in this PR as it will become much easier to write tests for the modified code? I am talking about the comment where you mentioned a possible `dual-greedy` mode.

akhandait:
And thank you for the review. Also sorry I didn't follow the style guide in some places, I will make sure to follow it next time. I will come up with the tests you suggested as soon as possible.

rcurtin:
That's just fine, feel free to fix it in this PR.  Don't worry about the style---there is usually a checker that can make the process easier. :)

akhandait:
When checking if an error is thrown when k is 0, knn test gives `memory access violation at address: 0x00000000: no mapping at fault address`. In the terminal, when running knn_main, it gives `segmentation fault(core dumped)`. Should I put a check in knn_main to throw an error when k = 0? Or, is the problem deeper and the algorithm needs to be checked?
When running kfn_main in the terminal and putting k = 0, the same segmentation fault is given but in the tests, memory access violation is NOT thrown and the test simply fails.
What is exactly happening here?

rcurtin:
Hmm, looks like a null pointer dereference.  Have you tried using a debugger like gdb and getting a stack trace where the segfault is happening? That could help make clear what the issue is.

rcurtin:

akhandait:
Really sorry for the delay on this one @rcurtin. I will come up with updates in a day.

rcurtin:
No worries---I just wanted Jenkins to build the PR since it is now back online. :)

akhandait:
@rcurtin Shouldn't we have a CheckMatricesNotequal as we have a CheckMatrices function for tests.
Because to check if two matrices are not equal, I see different methods used in the tests already present in the main_tests and some of them have possible bugs.

src/mlpack/tests/main_tests/kfn_test.cpp
akhandait: calling CleanMemory() throws a `memory access violation`. I am currently trying to dig into this and the other segmentation fault with a debugger.

akhandait:
I haven't added any tests in this commit, will add them in some time.

rcurtin:
I don't have any issues with adding a `CheckMatricesNotEqual()` function or similar.

While debugging you might also try valgrind; it can print exactly where invalid accesses happen and even better it can print when something is freed twice, and it will also print where that object was originally allocated.  I like to use it like this:

```
valgrind --leak-check=full --track-origins=yes bin/mlpack_test -t TestSuiteName/TestCaseName 2>&1 | less
```

(I redirect stderr to stdout and pipe to less to make reading the output easier.)

Hope this helps, let me know if I can clarify anything.

zoq:
For memory errors you can also check the results of the jenkins job: http://masterblaster.mlpack.org/job/pull-requests%20mlpack%20memory/410/valgrindResult/
We might have to adapt the failure report setting.

akhandait:
Thanks @rcurtin @zoq. Should I make a new PR for the `CheckMatricesNotEqual()` function, because I think I should use them in all the other already pushed tests. I will replicate the required changes in this PR as well.

rcurtin:
Sure, another PR is fine for me or if you want to do it here that is fine also.

akhandait:
I think I will go ahead and do it in this PR, the changes are minor.

akhandait:
I used valgrind and I think I have some clues as to where the memory errors are being thrown.
I commented out all the tests in `KNNMainTest` except the `KNNModelReuseTest` and saved the valgrind output:
[debug_specific.txt](https://github.com/mlpack/mlpack/files/1777454/debug_specific.txt)
Also, valgrind shows that there is definitely some memory in this as well as all the other tests as well, so basically all calls of `knn_main`.
I am still trying to figure out where exactly the memory error is occurring. Lines 380, 797, 518 of neighbor_search_impl.hpp seem to be of interest according to the output and I think maybe it has something to do with the `refrenceTree` pointer.
@zoq @rcurtin If you have some suggestions about how to better read the valgrind output for our errors, please tell.

akhandait:
Outputs of greedy spill tree don't change with changing rho, is this expected behavior or should I look for a possible bug somewhere?

rcurtin:
What values are you trying for changing rho?

src/mlpack/tests/main_tests/kfn_test.cpp
rcurtin: This is likely because of the line down below:

```
SetInputParam("input_model", 
      std::move(CLI::GetParam<KFNModel*>("output_model")));
```

Before calling `CleanMemory()` you'll need to copy the `output_model` result, since `CleanMemory()` will delete it.

akhandait:
I am trying to keep them as far apart as possible, I tried 0.2, 1 and they were still the same. Should I try using even bigger inputs, for the tau test, when I used 100 points in 3 dimensions, the outputs were the same but for 1000 points they were different.

src/mlpack/tests/main_tests/kfn_test.cpp
akhandait: I had a related doubt, after calling `CleanMemory()`, if we don't call `mlpackMain()`, why is a `memory access violation` thrown?
Also, I noticed that otherwise this error is only thrown whenever we set the input parameter `input_model` both before or after the CleanMemory() call.

rcurtin:
I realized that `tau` is not being set, but I think `tau` needs to be greater than 0 for `rho` to have any effect.  `tau` defines the width of the overlapping region for the spill tree---see the paper for more details about it.

akhandait:
The test is failing even after setting a non-zero of `tau`. 
About the `dual-greedy` mode, how do you think we should handle it? I think I would have to make this applicable in the `neighbor_search_impl` as well. Also, in the `dual-greedy` algorithm, should we just keep it defeatist(tau = 0)? Or should we allow all values of tau? and what about `singletree` algorithm with spill tress? It also has the same problem.
While forcing `dualtree` spill searches to use non-defeatist search, should we fix some high value of `tau`? Here too, what about the `singletree` algorithm?

akhandait:
Sorry for the confusion. I will implement this.
What about using spill trees with single_tree mode? We have the same problems then as throwing really large values in neighbors matrix. I think even that is using defeatist search by default, don't we need to make similar changes as you mentioned above for single_tree mode as well?

akhandait:
I think there were some additions to neighbor_search_impl.cpp which I didn't pull into my local branch. Hence those unintended changes, I will correct it.

akhandait:
So, I was setting `tau` to be 1 and it was too high for `rho` to have any effect. I have set it 0.3 now and the test passes. :)

rcurtin:
> What about using spill trees with single_tree mode? We have the same problems then as throwing really large values in neighbors matrix. I think even that is using defeatist search by default, don't we need to make similar changes as you mentioned above for single_tree mode as well?

Agreed, single_tree mode with spill trees should not be defeatist search.  Only greedy mode should be defeatist.

However regardless of defeatist or not, for single-tree search the invalid values in the neighbors matrix should be solved by #1234; are you saying there are still problems?

> Another doubt: Why is the style check giving a warning saying "Do not use variable length arrays." whenever I have used arma::Mat<size_t> neighbors[nofalgorithms]. Should I just replace nofalgorithms by the number or should I let this as it is?

I think that the style checker may be wrong here, because that is not a variable-length array.  Still, I think the code might be clearer if you just wrote `const int nofalgorithms = 3` or simply hardcoded `3` into the creation of the array.

akhandait:
I think even after #1234, when I tried `single_tree` algorithm with `spill` trees, the invalid values are being thrown.

rcurtin:
It looks to me as though both `single_tree` and `dual_tree` search with the spill tree are defined in such a way that they are defeatist search:

```
template<template<typename TreeMetricType,
                  typename TreeStatType,
                  typename TreeMatType> class TreeType = tree::SPTree>
using DefeatistKNN = NeighborSearch<
    NearestNeighborSort,
    metric::EuclideanDistance,
    arma::mat,
    TreeType,
    TreeType<metric::EuclideanDistance,
        NeighborSearchStat<NearestNeighborSort>,
        arma::mat>::template DefeatistDualTreeTraverser,
    TreeType<metric::EuclideanDistance,
        NeighborSearchStat<NearestNeighborSort>,
        arma::mat>::template DefeatistSingleTreeTraverser>;
```

```
typedef DefeatistKNN<tree::SPTree> SpillKNN;
```

We will need to ensure that when in `single_tree` and `dual_tree` mode, we don't use the greedy traversers.  So I suspect we will need to modify the typedef above, or at least not use `SpillKNN` in `ns_model.hpp` and `ns_model_impl.hpp`.

akhandait:
So, I changed the typedef to this:
```
template<template<typename TreeMetricType,
                  typename TreeStatType,
                  typename TreeMatType> class TreeType = tree::SPTree>
using NonDefeatistKNN = NeighborSearch<
    NearestNeighborSort,
    metric::EuclideanDistance,
    arma::mat,
    TreeType,
    TreeType<metric::EuclideanDistance,
        NeighborSearchStat<NearestNeighborSort>,
        arma::mat>::template DualTreeTraverser,
    TreeType<metric::EuclideanDistance,
        NeighborSearchStat<NearestNeighborSort>,
        arma::mat>::template SingleTreeTraverser>;

typedef NonDefeatistKNN<tree::SPTree> SpillKNN;
```
I will later define something like a Defeatist `DefSpillKNN`.
But now, when I tried tau `0.05`, it gives weird results. For some points it just repeats some of k nearest numbers, so actually there are less than k nearest neighbors. This happens in general for tau close to zero.

rcurtin:
Right, I remember now the way `NeighborSearchRules` are written, the tree should not have duplicate points in it.  In fact it is only reasonable to use a spill tree for defeatist search; it isn't going to give any advantages for exact search.  So I am fine if you want to modify the command-line program to require that if the tree type is `spill`, then the search mode must be `greedy` or `dual-greedy`.

akhandait:
When we set a `dual-greedy` mode, it will be `dual-tree` algorithm with `spill` trees and use the defeatist search, it will still throw those large invalid values. I think we need to sort that out.
I observed that the large invalid values are only thrown in `dual-tree` `spill` and `single-tree` `spill` tree mode when `tau` is exactly zero. For all other values of `tau`, it seems to work fine.
I don't understand conceptually, why setting the overlapping size(`tau`) to zero causes this problem. Maybe it has something to do with the `DefeatistDualTreeTraverser` and `DefeatistSingleTreeTraverser`.
Do we really need a `dual-greedy` mode here, or should we just not allow `tau` to be zero in the above cases.

rcurtin:
Right, so I think that this is a decision that was made by Marcos some time back.  I believe what he chose to do was have the search be implicitly approximate anytime the spill tree was chosen; this is done by setting the `SpillKNN` traversers to the defeatist traversers.

I think we can avoid adding a `dual-greedy` mode if we are sure to issue a message to `Log::Warn` pointing out that the search is going to be approximate whenever spill trees are used.

However both those things are separate from the issue where `tau != 0` gives invalid results, but I can't reproduce that issue.  The following two commands give good results:

```
$ bin/mlpack_knn -r test_data_3_1000.csv -v -t spill -a greedy -k 5 --leaf_size 1 -n n.csv -d d.csv --tau 0.5
$ bin/mlpack_knn -r test_data_3_1000.csv -v -t spill -a greedy -k 5 --leaf_size 1 -n n.csv -d d.csv
```

But the following two commands give bad results:

```
$ bin/mlpack_knn -r test_data_3_1000.csv -v -t spill -a single_tree -k 5 --leaf_size 1 -n n.csv -d d.csv -v
$ bin/mlpack_knn -r test_data_3_1000.csv -v -t spill -k 5 --leaf_size 1 -n n.csv -d d.csv -v
```

I'm not sure why `-a single_tree` gives different results than `-a greedy`---I think, according to the typedefs, that this should be executing exactly the same code.  That might be a good place to start debugging if you are willing to do that.  I do think that #1234 only fixed single-tree greedy search, so probably for dual-tree greedy search the same fix would need to be applied.

akhandait:
>However both those things are separate from the issue where tau != 0 gives invalid results, but I can't reproduce that issue. 

Sorry, Do you mean here that you can't reproduce invalid results when `tau != 0`? 
The issue is when `tau == 0` and not when `tau != 0`. Also, it is only for `dual-tree` and `single-tree` and not `greedy`. `greedy` never seems to give invalid results.

>I'm not sure why -a single_tree gives different results than -a greedy---I think, according to the typedefs, that this should be executing exactly the same code.

I will look into this.

rcurtin:
> Sorry, Do you mean here that you can't reproduce invalid results when tau != 0?
The issue is when tau == 0 and not when tau != 0. Also, it is only for dual-tree and single-tree and not greedy. greedy never seems to give invalid results.

Got it, I think maybe I misunderstood a little.  In any case, I think that the first thing to look into (which will lead to the solution for all these things) is to figure out why `single_tree` gives different results than `greedy` for the spill tree.

akhandait:
Really sorry for the inactivity here. This PR is on my mind and I will definitely get on this once I get the chance.

rcurtin:
No worries, whenever you have a chance.  If you don't get to it, I will---eventually. :)

Hsankesara:
Hey @rcurtin and @akhandait, Can I work on this PR and made changes which are required? Many of the tests are in there and I can add a few more which are required.

akhandait:
@Hsankesara Sure! I haven't been able to get back to this PR, I will be glad if someone completed it.

Hsankesara:
Hey @akhandait, I am wondering if there are some tests that haven't been implemented which you are intending to do. It would be helpful if you list them. Also if you have any suggestion, please also note it. I'll try to implement them as soon as I can.

akhandait:
Hi @Hsankesara, thanks for the PR. I am sorry I haven't been replying. I will take a look at it tomorrow and get back to you.

rcurtin:
Let's close this in favor of #1809. :+1:

Issue #1259, opened 2018-02-18
K-nearest neighbor search gives an arbitrary point


akhandait:
When we provide only the reference dataset to the knn search and put k equal to the number of points in the dataset, the last nearest point it provides for each of the points is apparently an arbitrary one.

Here is an example, 
` mlpack_knn -k 7 -r cov.csv -n neighbors.csv ` in the terminal
when cov has the following points:
4 | 2
7 | 1
8 | 7
6 | 3
4 | 5
7 | 7
9 | 0

gives following neighbors:
3 | 4 | 1 | 6 | 5 | 2 | 65
3 | 6 | 0 | 4 | 5 | 2 | 65
5 | 4 | 3 | 1 | 0 | 6 | 65
1 | 0 | 4 | 5 | 6 | 2 | 65
3 | 0 | 5 | 2 | 1 | 6 | 65
2 | 4 | 3 | 0 | 1 | 6 | 65
1 | 3 | 0 | 4 | 2 | 5 | 65

and following distances:
2.2360679775 | 3 | 3.1622776602 | 5.3851648071 | 5.8309518948 | 6.4031242374 | 1.797693134862E+308
2.2360679775 | 2.2360679775 | 3.1622776602 | 5 | 6 | 6.0827625303 | 1.797693134862E+308
1 | 4.472135955 | 4.472135955 | 6.0827625303 | 6.4031242374 | 7.0710678119 | 1.797693134862E+308
2.2360679775 | 2.2360679775 | 2.8284271247 | 4.1231056256 | 4.2426406871 | 4.472135955 | 1.797693134862E+308
2.8284271247 | 3 | 3.6055512755 | 4.472135955 | 5 | 7.0710678119 | 1.797693134862E+308
1 | 3.6055512755 | 4.1231056256 | 5.8309518948 | 6 | 7.2801098893 | 1.797693134862E+308
2.2360679775 | 4.2426406871 | 5.3851648071 | 7.0710678119 | 7.0710678119 | 7.2801098893 | 1.797693134862E+308

It works correctly as expected when the query dataset is provided and k is set equal to number of reference points.

Should k = number of reference points even be allowed when query dataset is not provided?

rcurtin:
Hi Atharva,

[...]

PR #1265, opened 2018-02-20
Fix neighbor search edge case

akhandait:
Solves #1259
Throw errors if k is set equal to the number of reference points when query data is not provided, in both kfn and knn.

### DIFF
--- a/src/mlpack/methods/neighbor_search/kfn_main.cpp
+++ b/src/mlpack/methods/neighbor_search/kfn_main.cpp
@@ -274,7 +274,7 @@ static void mlpackMain()
     // when query data has not been provided.
     if (!CLI::HasParam("query") && k == kfn->Dataset().n_cols)
     {
-      Log::Fatal << "Invalid k: " << k << "; must not be equal to the number of "
+      Log::Fatal << "Invalid k: " << k << "; must be less than the number of "
           << "reference points (" << kfn->Dataset().n_cols << ") "
           << "if query data has not been provided." << endl;
     }
--- a/src/mlpack/methods/neighbor_search/kfn_main.cpp
+++ b/src/mlpack/methods/neighbor_search/kfn_main.cpp
@@ -261,14 +261,22 @@ static void mlpackMain()
           << queryData.n_rows << "x" << queryData.n_cols << ")." << endl;
     }
 
-    // Sanity check on k value: must be greater than 0, must be less than the
-    // number of reference points.  Since it is unsigned, we only test the upper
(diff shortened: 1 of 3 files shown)


src/mlpack/methods/neighbor_search/kfn_main.cpp
```diff
@@ -261,15 +261,23 @@ static void mlpackMain()
...
           << "than or equal to the number of reference points ("
           << kfn->Dataset().n_cols << ")." << endl;
     }
+    // Sanity check on k value: must not be equal to the number of reference points
+    // when query data has not been provided.
+    if (!CLI::HasParam("query") && k == kfn->Dataset().n_cols)
+    {
+      Log::Fatal << "Invalid k: " << k << "; must not be equal to the number of "
+          << "reference points (" << kfn->Dataset().n_cols << ") "
+          << "if query data has not been provided." << endl;
```
rcurtin: I'm going to be a bit picky about the error message here if that's ok... this message tells the user what k cannot be but doesn't guide them for how to easily fix it.  So can I ask you to change this to something like "; must be less than the number of reference points (" please?

akhandait: Yes, that's much more appropriate. I will do it.

src/mlpack/methods/neighbor_search/neighbor_search_impl.hpp
```diff
@@ -850,6 +850,14 @@ DualTreeTraversalType, SingleTreeTraversalType>::Search(
         << "points in the reference set (" << referenceSet->n_cols << ")";
     throw std::invalid_argument(ss.str());
   }
+  if (k == referenceSet->n_cols)
+  {
+    std::stringstream ss;
+    ss << "requested value of k (" << k << ") is equal to the number of "
```
zoq: Another picky comment, this should start with a capital letter.

rcurtin:
Oops, forgot to do the merge.  Sorry for the delay. :)

akhandait:
Should I make some changes for the style check to pass for the merge?

rcurtin:
Sorry for the delay, I wanted to wait until the CI tests were done before merging, then I did not mark in my todo list that I had to respond to this once they were done.  I will fix all the issues during merge; the static code analysis issues aren't your fault, the style issues are easy, and the failing test is actually an invalid test that's easy to fix.  I'll do this after lunch today, so nothing needed from your end.

rcurtin:

PR #1271, opened 2018-02-23
Vanilla stochastic policy gradient implementation

chogba:
Have implemented vanilla stochastic policy gradient with Monte Carlo updates. 
The files compile without errors.
Have referred to Shangtong's [pull request](https://github.com/mlpack/mlpack/pull/934/files).

Reference - REINFORCE algorithm in slide no. 20 in David Silver's lecture slides - [here](http://www0.cs.ucl.ac.uk/staff/d.silver/web/Teaching_files/pg.pdf)

For the advantage function, I am using discounted state value.
The output layer is a softmax (defined in  src/mlpack/methods/ann/layer/policy_impl.hpp).
The network is getting updated at every step in an episode - no batch updates, as described in the pseudocode.

Instead of minimising a cross entropy loss for the softmax, we are using likelihood ratios, and maximising the score function (slide 18) with v~t~ as an unbiased estimate for r in the expression in the slides. The gradients have been correspondingly computed. \pi (s,\theta) is the softmax outputted by the network, and the gradients of the network parameters have been computed wrt it [reference](https://deepnotes.io/softmax-crossentropy)

Yet to make changes in the CMakeLists and other corresponding changes to incorporate these files.
Would help a lot if I can have these files reviewed first for any logical errors, from a vanilla implementation perspective.

Presently working on adding a Boost test suite for the files, will add it soon.


src/mlpack/methods/ann/layer/policy_impl.hpp
```diff
@@ -0,0 +1,58 @@
...
+    const arma::Mat<eT> &&prob,
+          arma::Mat<eT> &&advantage,
+          arma::Mat<eT> &&g)
+{
+  // Gradient with respect to nodes corresponding to actions other than the action taken is -p*advantage.
+  // Gradient with respect to node corresponding to taken action is (1-p)*advantage.
+  // The advantage vector has been correspondingly modified to accommodate this.
+  g = - (prob % advantage);
+  // We want to maiximise the objective function i.e take a positive step towards this gradient. Hence reverse the sign.
+  g = -g;
```
rcurtin: Why not just `g = (prob % advantage)`?  That will remove the unnecessary computation.

src/mlpack/methods/reinforcement_learning/policy_gradient_impl.hpp
```diff
@@ -0,0 +1,170 @@
+#ifndef MLPACK_METHODS_RL_POLICY_GRADIENT_IMPL_HPP
+#define MLPACK_METHODS_RL_POLICY_GRADIENT_IMPL_HPP
+
+#include <cstddef>
+#include <mlpack/methods/reinforcement_learning/training_config.hpp>
+#include <armadillo>
```
rcurtin: Ah, instead of including Armadillo directly you should instead include either `mlpack/core.hpp` or `mlpack/prereqs.hpp`.  This is because there is some special functionality that has to be added into Armadillo during the inclusion process.  That will probably also make it unnecessary to include `<cstddef>`.

src/mlpack/methods/ann/layer/policy.hpp
```diff
@@ -0,0 +1,99 @@
...
+ * @tparam InputDataType Type of the input data (arma::colvec, arma::mat,
+ *         arma::sp_mat or arma::cube).
+ * @tparam OutputDataType Type of the output data (arma::colvec, arma::mat,
+ *         arma::sp_mat or arma::cube).
+ */
+template<
+  typename InputDataType = arma::mat,
+  typename OutputDataType = arma::mat
+>
+class Policy
```
rcurtin: My initial thought about this layer is, would this layer be useful to anyone writing any generic neural network?  Or should we instead add this specifically to the RL codebase and namespace only, and then add it as a `CustomLayerTypes`?

I am a little concerned that someone will come across this class, but not understand where or why they might use it.  Ideally, a user should be able to quickly figure out what each class is for and how it is used.  (I realize that is an ideal and we will never truly achieve it, but hey, we should try... :))

PR #1274, opened 2018-02-24
SELU activation function layer implementation

dakshitagrawal97:
This PR implements the SELU activation function as proposed by this [NIPS paper](https://arxiv.org/abs/1706.02515).  It has been implemented as a layer due to the necessity of a particular weight initialization to be done with the SELU activation function as given in the paper.


dakshitagrawal97:
@zoq Could you please have a look at this PR?  Instead of implementing it in the activation_functions module as discussed in IRC, I did it in the layers module.  This is because:
1. the value of the parameters are fixed only if the input is normalized.  
2. a particular weight initialization has to be done.

Issue #1273, opened 2018-02-24
Implement Batch Normalisation Layer


Manthan-R-Sheth:
I would like to begin the task of implementing deep learning modules with the implementation of [Batch Normalisation Layer](https://arxiv.org/abs/1502.03167). This is due to the wide advantage of using BatchNorm for DL tasks. It can be used for FFN as well as Convolution Networks.

@zoq @lozhnikov @rcurtin, shall i go ahead with this task?

zoq:
BatchNorm would be a great to have, there is an open PR https://github.com/mlpack/mlpack/pull/955 which I think is a good start, so if you like, clone the code and resolve the remaining issues.

sourabhvarshney111:
@Manthan-R-Sheth I think we have completed our work for this. Can we close it now? What say?

zoq:
Right, let's close this one.

PR #1275, opened 2018-02-25
Implement BatchNorm Layer

Manthan-R-Sheth:
This PR implements Batch Normalisation Layer to the ann/layer module.
This is with regards to the issue #1273. I will be adding tests for the same in the next commits, once the implementation is approved.


zoq:
Do you mind to add a simple test first? This makes the review a lot easier since we have something you can test right away.

Manthan-R-Sheth:
I have added the tests, but the travis builds are failing due to errors in loading from xml archive (in the serialisation test). I think its unrelated to the addition of BatchNorm Layer. Nevertheless I am not sure about it.
Can you guide me through.

src/mlpack/methods/ann/layer/batch_norm_impl.hpp
```diff
@@ -0,0 +1,141 @@
...
+}
+
+template<typename InputDataType, typename OutputDataType>
+template<typename Archive>
+void BatchNorm<InputDataType, OutputDataType>::serialize(
+  Archive& ar, const unsigned int /* version */)
+{
+  ar & BOOST_SERIALIZATION_NVP(gamma);
+  ar & BOOST_SERIALIZATION_NVP(beta);
+  ar & BOOST_SERIALIZATION_NVP(stats.mean());
```
zoq: Not sure what ``stats.mean()``and ``stats.var(1)`` returns, but my guess is it#s not serializable out of the box. But I think it's safe to remove the two lines here, since each is set at every run.

Manthan-R-Sheth: Both return vectors. I think addition of `#include <boost/serialization/vector.hpp>` should make it serialisable. However, I am not sure if it will be safe to remove the lines. This is because if the model is used in deterministic mode, then stats() will be null.

src/mlpack/tests/ann_layer_test.cpp
```diff
@@ -1282,4 +1282,104 @@ BOOST_AUTO_TEST_CASE(SimpleMeanSquaredErrorLayerTest)
...
+  model.Deterministic() = true;
+  model.Forward(std::move(input), std::move(output));
+
+  result << 0.83504842 <<  0.15182699 << -0.53139445 << arma::endr
+         << 0.55589881 << -1.11179762 << -0.44471905 << arma::endr
+         << -0.39980015 << -0.39980015 << -1.29935049 << arma::endr
+         << -0.45175395 << -0.45175395 << -0.45175395 << arma::endr;
+
+  CheckMatrices(output, result);
+}
```
rcurtin: It might be a good idea to also add a test that uses the `CheckGradient()` function; this is a nice way to ensure that the implementation of the objective matches the implementation of the gradient.

Manthan-R-Sheth: I have added it in the latest commit.

src/mlpack/methods/ann/layer/batch_norm_impl.hpp
Manthan-R-Sheth: Actually I checked it and found that it is not serializable and so I have removed it as of now. However I am not able to convince myself about whether it is safe to do so.

zoq:
```c++
mean = arma::mean(input, 1);
variance = arma::var(input, 1, 1);
```
this is done in each forward pass, so if we load a model we don't use the previous value of ``mean`` and ``variance``.

src/mlpack/methods/ann/layer/batch_norm_impl.hpp
```diff
@@ -0,0 +1,140 @@
...
+    }
+  }
+
+  for (size_t i = 0; i < output.n_rows; i++)
+  {
+    output.row(i) = arma::as_scalar(gamma.row(i))*((input.row(i)
+                    - arma::as_scalar(mean.row(i)))
+                    / (arma::as_scalar(arma::sqrt(variance.row(i) + eps))))
+                    + arma::as_scalar(beta.row(i));
+  }
```
rcurtin: Armadillo typically does calculations much faster when you do then columnwise.  Can you figure out a way to do this by looping over columns instead of rows?

src/mlpack/methods/ann/layer/batch_norm_impl.hpp
```diff
@@ -0,0 +1,140 @@
...
+  else
+  {
+    mean = arma::mean(input, 1);
+    variance = arma::var(input, 1, 1);
+
+    for (size_t i = 0; i < output.n_cols; i++)
+    {
+      stats(input.col(i));
+    }
```
rcurtin: I have a question about this.  In the paper, the authors simplify for the mini-batch setting:

```
In the batch setting where each training step is based on the entire training set, we would use the whole set to normalize activations. However, this is impractical when using stochastic optimization. Therefore, we make the second simplification: since we use mini-batches in stochastic gradient training, each mini-batch produces estimates of the mean and variance of each activation.
```

How does this apply to what you have implemented here?  To me, it looks like `stats` collects statistics on all points that pass through `Forward()` before `Reset()` is called.  (I believe `Reset()` is actually only called when the network is created or serialized, so maybe we should be resetting the running statistics more often?)

It would be possible to reset the statistics every call to `Forward()` because `Forward()` may be called with mini-batches, but, it might also be useful to find a way to allow the user to specify how often the statistics should be reset.  Let me know what you think---or if I misunderstood the functioning of the code here. :)

ShikharJ:
Hi, sorry for commenting a bit late. I observed that this PR emulates #955 in a number of ways, should this be using the commits from there @rcurtin ?

src/mlpack/methods/ann/layer/batch_norm_impl.hpp
Manthan-R-Sheth: In this implementation, it is considered that each column is the training sample and so statistics across rows (which correspond to same dimension) are taken. I think transposing input first and finding statistics over columns, thereby looping over columns and transposing the result back should do the trick.

src/mlpack/methods/ann/layer/batch_norm_impl.hpp
Manthan-R-Sheth: `stats` object is used to get the mean and variance only during the testing time i.e. `deterministic = true`. So in the case of mini-batches at training, mean and variance are calculated from the samples in the mini-batch only, however they are stored in `stats` so that during testing, entire training statistics can be used.
Correct me if I misunderstood your question.

src/mlpack/methods/ann/layer/batch_norm_impl.hpp
Manthan-R-Sheth: so, shall I implement this, if it sounds fine?

rcurtin: No, transposing and then performing the calculation will certainly be slower because the cost of the transpose itself will almost certainly outweigh the slowness of the suboptimal access pattern.  I would instead suggest spending a little time to think about how you can perform this computation in batch for all rows at once.  i.e., instead of looping over each column, can you set the entire `output` matrix in a single Armadillo expression?  It should be possible.

src/mlpack/methods/ann/layer/batch_norm_impl.hpp
rcurtin: Ah, now it makes sense; I overlooked how `mean` and `variance` are set differently.  Thank you for the clarification!

rcurtin:
@ShikharJ: you're right, thank for pointing that out.  @Manthan-R-Sheth: do you think you could rebase your commits onto the existing commits for #955 so that @chvsp still gets credited correctly for those commits?

src/mlpack/methods/ann/layer/batch_norm_impl.hpp
Manthan-R-Sheth: Sure, I would like to think on this and come up with something nice. 
Thanks.

Manthan-R-Sheth:
@rcurtin I will perform the comparisons. If the results are positive, then it will be a good check/advantage of BatchNormLayer. 
However , I have some issues with `checkGradient()` that I am trying to solve. The screenshot below shows `estGradient(i)` and `orgGradient(i)`.
screenshot from 2018-03-01 21 37 01.
The scale parameter(gamma) which are initial the 10 rows for `10 X 1` input has some high differences for gradients.Any views on this will be helpful.

Also, I had preserved @chvsp 's name in the author list for the files thinking that it would be good enough. However, ya i can definitely rebase the commits, if you say so.

Manthan-R-Sheth:
The catch was that, input was declared as `randu()` but it should be `randn()`. Batch Normalisation assumes normal distribution of the input which is then scaled and shifted. Hence, choosing such an input did the trick. `checkGradient()` test should now pass.

Manthan-R-Sheth:
@rcurtin 
I tried the model comparison task on Kaggle-Digit-Recogniser and found that the training time did not reduce significantly Infact it was approximately the same before and after the inclusion of BatchNorm layers to the model. The reason could be that since it is a fully connected network, the benefits of BatchNorm are not visible significantly. I am planning to try it with a CNN network to establish the acceleration of training time.
However there was a significant increase the accuracy as observed during the process (I have used PReLU layer as the activation function). The mages below show the increase in accuracy :- 

Without BatchNorm,
withoutbatchnorm
  
With BatchNorm,
batchnorm

src/mlpack/methods/ann/layer/batch_norm_impl.hpp
Manthan-R-Sheth:
```
variance = variance + eps;
variance = arma::sqrt(variance);
variance = arma::repmat(variance,1,input.n_cols);
output = arma::repmat(gamma,1,input.n_cols)%((input - arma::repmat(mean,1,input.n_cols))%(1/variance)) + arma::repmat(beta,1,input.n_cols);
```

This transforms the input matrix directly to output matrix. This is should do the trick.
Any views @rcurtin

rcurtin:
> The catch was that, input was declared as randu() but it should be randn(). Batch Normalisation assumes normal distribution of the input which is then scaled and shifted. Hence, choosing such an input did the trick. checkGradient() test should now pass.

Yeah, I tested this also with different random seeds.  I forgot that I was running the tests and ended up running them 8500 times, but there were no failures.

```
i=0; while(true); do echo $i; bin/mlpack_test -t ANNLayerTest/GradientBatchNormLayerTest 2>&1 | grep -v '*** No errors' | grep -v '^[ ]*$' | grep -v '^Running '; sleep 1; i=$(($i + 1)); done
```

(I did that after I added `math::RandomSeed(std::time(NULL))` to the top of the `GradientBatchNormLayerTest`.)

However there is still a failure for the `BatchNormTest`:

```
Running 1 test case...
/home/ryan/src/mlpack-manthan/src/mlpack/tests/test_tools.hpp(36): fatal error: in "ANNLayerTest/BatchNormTest": difference{0.000189782} between a[i]{0.10997912794208686} and b[i]{0.11} exceeds 0.01%
```

The improvement for the Kaggle example looks promising; when I said that it should train "faster" all I really meant was that it would achieve better test accuracy in fewer epochs, which is exactly what it's doing here.  If you do want to test also with a CNN, I think it would be a good idea, and if you wanted we could adapt your code and add it to the models/ repo too.

Manthan-R-Sheth:
@rcurtin 
The failure in BatchNormTest is due to precision issues(rounding off error). This is the reason why travis build is failing too.
Ya i will test it for CNN and then we can add it to the models repo.
Any more code changes that I have to do as of now? I think the looping over column optimisation will be the only one.

rcurtin:
> The failure in BatchNormTest is due to precision issues(rounding off error). This is the reason why travis build is failing too.

If that's true, then should we adjust the tolerance?

I can't think of any other code issues at the moment other than fixing the static code analysis checks and the style checks.

Manthan-R-Sheth:
Ya, even I think increasing tolerance is a good workaround, as I locally tested and the values were just off due to precision errors.
And can I update the looping over columns with the code below,

```
variance = arma::repmat(arma::sqrt(variance+eps),1,input.n_cols);
output = arma::repmat(gamma,1,input.n_cols)%((input - arma::repmat(mean,1,input.n_cols))%(1/variance)) + arma::repmat(beta,1,input.n_cols);
```
This should prevent the looping over the matrix.

rcurtin:
Hmm, my only concern would be that the expression is complex enough that `repmat()` will actually create the copies.  I wonder if you could do something more like ...

```
// this is not correct, it is just the start of an idea...
output = input - arma::repmat(mean, 1, input.n_cols);
output.each_col() %= gamma / (variance + eps);
output.each_col() += beta;
```

I think maybe that might work but it will need a little thought.  Maybe `each_col()` can be used entirely instead of `repmat()`, and I would say that is far more likely to avoid any unnecessary creations of copied matrices.

Manthan-R-Sheth:
Agreed about the creation of copies issue. I guess the below should be good to go then,
```
output = input.each_col() - mean;
output.each_col() %= gamma/arma::sqrt(variance+eps);
output.each_col() += beta;
```

zoq:
Yes, this looks good.

Manthan-R-Sheth:
@zoq @rcurtin 
I have optimised calculations for `forward` `backward` `gradient` methods. There is no need to use looping over the rows. Instead `each_col()` is used thereby optimising the code.

rcurtin:
That is a nice improvement; thanks for working on it.  It looks like there is a compilation failure, can you fix that please?  I don't think there are many further comments from me (I'll take a final look once all the tests are passing).

Manthan-R-Sheth:
@rcurtin I ran into some rebase issues, can I make a fresh rebased PR for the same?

rcurtin:
Sure, no problem, feel free.

Manthan-R-Sheth:
@rcurtin @zoq 
I managed to restore the initial state back and rebase the branch as well.
Git history is clean now.
I think it should be fine now.

zoq:
Can you handle the static analysis job issues?

src/mlpack/methods/ann/layer/batch_norm_impl.hpp
```diff
@@ -0,0 +1,128 @@
...
+  // Nothing to do here.
+}
+
+template <typename InputDataType, typename OutputDataType>
+BatchNorm<InputDataType, OutputDataType>::BatchNorm(
+  const size_t size, const double eps) : size(size), eps(eps)
+{
+  weights.set_size(size + size, 1);
+}
+
```
zoq: Looks like we  can remove the extra line here.

src/mlpack/methods/ann/layer/batch_norm_impl.hpp
```diff
@@ -0,0 +1,128 @@
...
+  deterministic = false;
+  gamma.fill(1.0);
+  beta.fill(0.0);
+  stats.reset();
+}
+
+template<typename InputDataType, typename OutputDataType>
+template<typename eT>
+void BatchNorm<InputDataType, OutputDataType>::Forward(
+  const arma::Mat<eT>&& input, arma::Mat<eT>&& output)
```
zoq: This should be tabbed twice.

src/mlpack/methods/ann/layer/batch_norm_impl.hpp
```diff
@@ -0,0 +1,128 @@
...
+
+  output = input.each_col() - mean;
+  output.each_col() %= gamma/arma::sqrt(variance+eps);
+  output.each_col() += beta;
+}
+
+template<typename InputDataType, typename OutputDataType>
+template<typename eT>
+void BatchNorm<InputDataType, OutputDataType>::Backward(
+  const arma::Mat<eT>&& input, arma::Mat<eT>&& gy, arma::Mat<eT>&& g)
```
zoq: This should be tabbed twice.

src/mlpack/methods/ann/layer/batch_norm_impl.hpp
```diff
@@ -0,0 +1,128 @@
...
+    variance = arma::var(input, 1, 1);
+
+    for (size_t i = 0; i < output.n_cols; i++)
+    {
+      stats(input.col(i));
+    }
+  }
+
+  output = input.each_col() - mean;
+  output.each_col() %= gamma/arma::sqrt(variance+eps);
```
zoq: Another picky style comment, please use a space between operations.

src/mlpack/methods/ann/layer/batch_norm_impl.hpp
```diff
@@ -0,0 +1,128 @@
...
+  output.each_col() += beta;
+}
+
+template<typename InputDataType, typename OutputDataType>
+template<typename eT>
+void BatchNorm<InputDataType, OutputDataType>::Backward(
+  const arma::Mat<eT>&& input, arma::Mat<eT>&& gy, arma::Mat<eT>&& g)
+{
+  size_t n = input.n_cols;
```
zoq: I think we could just use ``input.n_cols``, since this is only used twice.

src/mlpack/methods/ann/layer/batch_norm_impl.hpp
```diff
@@ -0,0 +1,128 @@
...
+void BatchNorm<InputDataType, OutputDataType>::Gradient(
+  const arma::Mat<eT>&& input,
+  arma::Mat<eT>&& error,
+  arma::Mat<eT>&& gradient)
+{
+  arma::mat normalized(input.n_rows, input.n_cols);
+  gradient.reshape(size + size, 1);
+
+  normalized = input.each_col() - arma::mean(input, 1)
+                                  / arma::sqrt(arma::var(input, 1, 1) + eps);
```
zoq: This should be tabbed twice.

src/mlpack/methods/ann/layer/batch_norm_impl.hpp
```diff
@@ -0,0 +1,128 @@
...
+
+  gradient.submat(0, 0, gamma.n_elem - 1, 0) = arma::sum(normalized % error, 1);
+  gradient.submat(gamma.n_elem, 0, gradient.n_elem - 1, 0) =
+                                                          arma::sum(error, 1);
```
zoq: See comment, above.

src/mlpack/methods/ann/layer/layer_types.hpp
```diff
@@ -108,7 +111,8 @@ using LayerTypes = boost::variant<
     BaseLayer<IdentityFunction, arma::mat, arma::mat>*,
     BaseLayer<TanhFunction, arma::mat, arma::mat>*,
     BaseLayer<RectifierFunction, arma::mat, arma::mat>*,
-    BilinearInterpolation<arma::mat, arma::mat>*,
+    BatchNorm<arma::mat, arma::mat>*,
+//    BilinearInterpolation<arma::mat, arma::mat>*,
```
zoq: We should integrate the ``BilinearInterpolation`` class.

src/mlpack/tests/ann_layer_test.cpp
```diff
@@ -1317,4 +1318,129 @@ BOOST_AUTO_TEST_CASE(SimpleBilinearInterpolationLayerTest)
...
+ * Link to the implementation - http://cthorey.github.io./backpropagation/
+ */
+BOOST_AUTO_TEST_CASE(BatchNormTest)
+{
+  arma::mat input, output;
+  input << 5.1 << 3.5 << 1.4 << arma::endr
+        << 4.9 << 3.0 << 1.4 << arma::endr
+        << 4.7 << 3.2 << 1.3 << arma::endr;
+
+  size_t numUnits = input.n_rows;
```
zoq: This is just used once, so I guess we can also directly use ``input.n_rows``.

src/mlpack/tests/ann_layer_test.cpp
```diff
@@ -1317,4 +1318,129 @@ BOOST_AUTO_TEST_CASE(SimpleBilinearInterpolationLayerTest)
...
+  // Backward Pass Test.
+  arma::mat gy;
+  gy << 0.8402 << 0.9116 << 0.2778 << arma::endr
+     << 0.3944 << 0.1976 << 0.5540 << arma::endr
+     << 0.7831 << 0.3352 << 0.4774 << arma::endr;
+
+  model.Backward(std::move(input), std::move(gy), std::move(output));
+  result << -0.0780 << 0.1376 << -0.0596 << arma::endr
+         << 0.0602 << -0.1317 << 0.0715 << arma::endr
+         << 0.0835 << -0.1493 << 0.0658  << arma::endr;
```
zoq: Looks like an extra space at the end.

src/mlpack/tests/ann_layer_test.cpp
```diff
@@ -1317,4 +1318,129 @@ BOOST_AUTO_TEST_CASE(SimpleBilinearInterpolationLayerTest)
...
+  struct GradientFunction
+  {
+    GradientFunction()
+    {
+      input = arma::randn(10, 256);
+      arma::mat target;
+      target.ones(1, 256);
+
+      model = new FFN<NegativeLogLikelihood<>, NguyenWidrowInitialization>(
+        input, target);
```
zoq: This should be tabbed twice.

Manthan-R-Sheth:
@zoq Thanks for the suggestions. 
I have made the necessary changes and updated the code.

src/mlpack/methods/ann/layer/batch_norm_impl.hpp
```diff
@@ -0,0 +1,131 @@
...
+#include "batch_norm.hpp"
+
+namespace mlpack {
+namespace ann { /** Artificial Neural Network. */
+
+template<typename InputDataType, typename OutputDataType>
+BatchNorm<InputDataType, OutputDataType>::BatchNorm():
+  size(10),
+  eps(1e-7),
+  deterministic(false)
```
rcurtin: Sorry for all the really picky comments here... maybe we need to update the style checker's rules to make sure it catches everything... :)

In this bit there should be a space between `BatchNorm()` and `:`: `BatchNorm() :`, and the initialization list should be tabbed twice.  This happens a few other cases but I don't want to leave 10,000 comments so I'll just make a note of it once here.

Manthan-R-Sheth:
@rcurtin 
I have fixed those.
Thanks for approving. I had tested the math behind the implementation both through code as well as manually and I think works fine. 
I used [this](https://kratzert.github.io/2016/02/12/understanding-the-gradient-flow-through-the-batch-normalization-layer.html) for reference as well.

Manthan-R-Sheth:
@rcurtin  @zoq 
Any updates with this?

zoq:
I'll recheck the forward/backward pass one more time, let you know if anything isn't clear and I think after that this is ready to be merged.

Manthan-R-Sheth:
Sure, thanks.

src/mlpack/methods/ann/layer/batch_norm_impl.hpp
```diff
@@ -0,0 +1,131 @@
...
+    variance = arma::var(input, 1, 1);
+
+    for (size_t i = 0; i < output.n_cols; i++)
+    {
+      stats(input.col(i));
+    }
+  }
+
+  output = input.each_col() - mean;
+  output.each_col() %= gamma / arma::sqrt(variance+eps);
```
zoq: Picky style comment, use a space between operations; that also applies for other places e.g. line 94. Also, do you think we should write  ``gamma / (arma::sqrt(variance) + eps)`` here?

src/mlpack/methods/ann/layer/batch_norm_impl.hpp
```diff
@@ -0,0 +1,131 @@
...
+  output.each_col() += beta;
+}
+
+template<typename InputDataType, typename OutputDataType>
+template<typename eT>
+void BatchNorm<InputDataType, OutputDataType>::Backward(
+    const arma::Mat<eT>&& input, arma::Mat<eT>&& gy, arma::Mat<eT>&& g)
+{
+  g.reshape(input.n_rows, input.n_cols);
```
zoq: This isn't necessary, since we set ``g`` in line 94.

src/mlpack/methods/ann/layer/batch_norm_impl.hpp
```diff
@@ -0,0 +1,131 @@
...
+
+  mean = arma::mean(input, 1);
+  variance = arma::var(input, 1, 1);
+
+  arma::mat m = arma::sum(gy % (input.each_col() - mean), 1);
+  g = arma::repmat(m, 1, input.n_cols) % -(input.each_col() - mean);
```
zoq: Maybe ``(-input.each_col() + mean);`` is easier to read?

src/mlpack/methods/ann/layer/batch_norm_impl.hpp
```diff
@@ -0,0 +1,131 @@
...
+  g.each_col() %= 1.0 / (variance + eps);
+  g += (input.n_cols * gy - arma::repmat((arma::sum(gy, 1)), 1, input.n_cols));
```
zoq: Could we use ``each_col`` here as well?

src/mlpack/methods/ann/layer/batch_norm_impl.hpp
```diff
@@ -0,0 +1,131 @@
...
+
+template<typename InputDataType, typename OutputDataType>
+template<typename eT>
+void BatchNorm<InputDataType, OutputDataType>::Gradient(
+    const arma::Mat<eT>&& input,
+    arma::Mat<eT>&& error,
+    arma::Mat<eT>&& gradient)
+{
+  arma::mat normalized(input.n_rows, input.n_cols);
+  gradient.reshape(size + size, 1);
```
zoq: We should use ``set_size`` which should be faster.

src/mlpack/methods/ann/layer/batch_norm_impl.hpp
```diff
@@ -0,0 +1,131 @@
...
+}
+
+template<typename InputDataType, typename OutputDataType>
+template<typename eT>
+void BatchNorm<InputDataType, OutputDataType>::Gradient(
+    const arma::Mat<eT>&& input,
+    arma::Mat<eT>&& error,
+    arma::Mat<eT>&& gradient)
+{
+  arma::mat normalized(input.n_rows, input.n_cols);
```
zoq: We can just write ``arma::mat normalized =`` no need to allocate memory here.

src/mlpack/methods/ann/layer/batch_norm_impl.hpp
Manthan-R-Sheth: arma::sqrt(variance +eps) will be closer to variance as compared to (arma::sqrt(variance) + eps) for varainace > 1. For variance < 1, the latter will be closer but the difference will be very less only.
Technically both are fine, but I think going with first one is better as it will be closer to true variance when variance > 1 (which will be the case most of the times).

src/mlpack/methods/ann/layer/batch_norm_impl.hpp
Manthan-R-Sheth: since it is g += n*gy - repmat(sum(gy,1));
I updated it to 
g += gy.each_col() - sum(gy,1);  g+= (n-1)*gy;

Manthan-R-Sheth:
@zoq I have updated the code.

src/mlpack/methods/ann/layer/batch_norm_impl.hpp
```diff
@@ -0,0 +1,129 @@
...
+template<typename InputDataType, typename OutputDataType>
+template<typename eT>
+void BatchNorm<InputDataType, OutputDataType>::Backward(
+    const arma::Mat<eT>&& input, arma::Mat<eT>&& gy, arma::Mat<eT>&& g)
+{
+  mean = arma::mean(input, 1);
+  variance = arma::var(input, 1, 1);
+
+  arma::mat m = arma::sum(gy % (input.each_col() - mean), 1);
+  g = arma::repmat(m, 1, input.n_cols) % (mean - input.each_col());
```
zoq: I'm wondering could we write: ``(mean - input.each_col()) % m)`` here as well? Ideally I like to avoid repmat here.

Manthan-R-Sheth: It will give logic error, due to dimension mismatch.
However, 
```
g = (mean - input.each_col());
g.each_col() %= m;
```
should do the trick. I will check and update, if it works.
Nice catch though, I missed this one.

Manthan-R-Sheth:
@zoq and @rcurtin 
Good to go now?

zoq:
Thanks again for the great contribution.

Issue #1276, opened 2018-02-25
Word embeddings for string representation.


Manthan-R-Sheth:
This is a prototype of the numeric representation of strings.
```
embed = model.Embedding(Vocab_size, Dimen_size) // Embedding Layer
embed.weight = (pretrained_weight) // matrix datatype
```
`pretrained_weight` can be word vector, GLoVE or any other already trained dense representation. 
Also each word is indexed by the id, which allows for word embedding lookup.

Steps that I think would be needed :-
1. A parser has to be written/modified to load the csv for pretrained words. Once a `arma::mat` is formed, it can be used for all algorithms. (CLI, cpp programs as well as bindings).
2. Support for different dimensions representations.

@zoq @rcurtin  if you could give your inputs related to this, it would be very helpful.

rcurtin:
Can you clarify what you mean by 'each word is indexed by the id'?  I guess the most important thing (to me) is the representation we use in C++ for a string dataset---unless we instead convert directly and immediately to a numeric representation.  I think that once we have a clear picture of what to do there, the rest of the pieces will easily fall into place.

Manthan-R-Sheth:
What I mean by 'each word indexed by the id' is that each word in the input will have a unique id associated to it. This will be done by the `tokeniser`(this will be implemented) that identifies the unique words in the input samples. Suppose our vocab size is 50 words (which we find after tokeniser does first pass through the data). Each word will also be mapped to a unique id. For eg:- The sentence "This is the name of the first commit" will be converted to [ 17,12,45,32,43,45,27,29 ] after the tokeniser is executed.
This is the input for the Embedding Layer. Thus input is considered to be a matrix of word indices.
This associated id will be the row number in the embedding matrix that is learnt for the words in the vocab or used directly by pre-trained vectors.

Manthan-R-Sheth:
@rcurtin 
Any views on the above idea?

scrcro:
I am currently working on an Embedding layer. You can write your own layers now with the latest drop, therefore you will have control over the `Forward`, `Backward` (not so important as this will be the first layer usually) and `Gradient` function. There are a couple of issues I am facing too:
- The optimization right now will update all your weight vectors and not just the relevant ones. This could be a problem for very large vocabularies
- If you want to use RNN then it expects one output for every timestep. This is a bit of a problem for n-gram models. But if you stick to FFN you should be ok.

[...]

PR #1277, opened 2018-02-26
Add Embedding layer and test

Manthan-R-Sheth:
Main purpose of the` Embedding layer` is to have numeric representations of the words(string) inputs. This is with reference to #1276.

`Embedding<> model(vocabSize, dimensionSize, pretrain=false)
`
`
model.Reset()
`

This creates an `embeddingMatrix` of size `vocabSize x dimensionSize`. Input is considered to be a matrix of word indices. For eg:- If the vocabSize is 2 (which means there are only 2 words in the vocabulary), then a possible input is `[[0,1],[1,1]]`, which means that input has 2 sentences each of 2 words with the corresponding word indices as matrix elements.

The `embeddingMatrix` consists of either pretrained word representations (pretrain=true) or trainable word representations(pretrain=false).

For the above example with word dimension of 3, the output of embedding layer will be of the form `[[[0.2,0.3],[0.5,0.6]],[[0.5,0.6],[0.5,0.6]]]`, which is the numeric representation of each word of the input sentence.

TODO 
A tokenizer needs to be made which tokenizes and pads the input sentences thereby making the input compatible with the embedding layer.

### DIFF
--- a/src/mlpack/methods/ann/layer/embedding_impl.hpp
+++ b/src/mlpack/methods/ann/layer/embedding_impl.hpp
@@ -1,0 +1,105 @@
+/**
+ * @file max_pooling_impl.hpp
+ * @author Manthan-R-Sheth
+ *
+ * Implementation of the Embedding class.
+ *
+ * mlpack is free software; you may redistribute it and/or modify it under the
+ * terms of the 3-clause BSD license.  You should have received a copy of the
+ * 3-clause BSD license along with mlpack.  If not, see
+ * http://www.opensource.org/licenses/BSD-3-Clause for more information.
+ */
+#ifndef MLPACK_EMBEDDING_IMPL_HPP
+#define MLPACK_EMBEDDING_IMPL_HPP
+
+// In case it hasn't yet been included.
+#include "embedding.hpp"
+
+namespace mlpack {
+namespace ann /** Artificial Neural Network. */ {
+
+template<typename InputDataType, typename OutputDataType>
+Embedding<InputDataType, OutputDataType>::Embedding()
+{
+  // Nothing to do here.
+}
+
+template<typename InputDataType, typename OutputDataType>
+Embedding<InputDataType, OutputDataType>::Embedding(
+  const size_t vocabSize,
+  const size_t dimensionSize,
+  const bool pretrain) :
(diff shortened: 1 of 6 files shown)


Manthan-R-Sheth:
@zoq @rcurtin
I tried to add the support for more than one sentence, but the output of the layer will be `arma::cube` in that case, as a result of which many changes will have to be made for compatibility. So it is single instance learning as of now.
This made me to think about the support for batch learning in general for algorithms(as sentences are batches for the embedding layer). 
I am figuring out a way for this. Any suggestions regarding the same?

src/mlpack/methods/ann/layer/embedding_impl.hpp
```diff
@@ -0,0 +1,102 @@
...
+template<typename InputDataType, typename OutputDataType>
+template<typename eT>
+void Embedding<InputDataType, OutputDataType>::Forward(
+  const arma::Mat<eT>&& input, arma::Mat<eT>&& output)
+{
+
+  // Apply embeddingMatrix multiplication to the input and store the results.
+  for (arma::uword j = 0; j < input.n_cols; ++j)
+  {
+    arma::rowvec wordRepresentation = embeddingMatrix.row(input(0,j));
```
Manthan-R-Sheth: i have considered only single instance, due to which 0 is the row_number.

rcurtin:
Hi Manjan,

Let us leave this one for the actual GSoC project please.  Many students will submit proposals with different APIs and ideas, and therefore I don't want to settle on any abstractions until a student is working on the project (if we accept a student for that project this year, which we don't currently know).  Otherwise we may risk merging lots of code that will end up being unused or might need to be fully refactored.

sourabhvarshney111:
@Manthan-R-Sheth I would like to work on this as my GSoC project. Can I?

rcurtin:
I'd actually prefer to close it for now just because there are so many PRs open and they are hard to organize.  In the future we can reopen when we are sure it is the right direction to go.

We can discuss either on the mailing list or in #1276; I'll leave a response there next.

Manthan-R-Sheth:
Sure, thanks.
I shall close the PR then.

PR #1278, opened 2018-02-26
Add Alpha Dropout Layer

dakshitagrawal97:
This PR adds a variant of dropout - alpha_dropout - as proposed in this [NIPS Paper](https://arxiv.org/pdf/1706.02515.pdf).  
I will add the tests once the implementation is approved.

### DIFF
--- a/src/mlpack/methods/ann/layer/alpha_dropout.hpp
+++ b/src/mlpack/methods/ann/layer/alpha_dropout.hpp
@@ -37,12 +37,12 @@ namespace mlpack {
  *
  * @code
  * @article{Klambauer2017,
-+ *   author  = {Gunter Klambauer and Thomas Unterthiner and
-+ *              Andreas Mayr},
-+ *   title   = {Self-Normalizing Neural Networks},
-+ *   journal = {Advances in Neural Information Processing Systems},
-+ *   year    = {2017}
-+ * }
+ *   author  = {Gunter Klambauer and Thomas Unterthiner and
+ *              Andreas Mayr},
+ *   title   = {Self-Normalizing Neural Networks},
+ *   journal = {Advances in Neural Information Processing Systems},
+ *   year    = {2017}
+ * }
  * }
  * @endcode
  *
@@ -114,6 +114,12 @@ namespace mlpack {
 
             //! Value to be added to a*x for affine transformation.
             double B() const { return b; }
+
+            //! Value of alpha_dash.
+            double Alpha_Dash() const {return alpha_dash; }
+
+            //! Get the mask.
+            OutputDataType const& Mask() const {return mask;}
 
             //! Modify the probability of setting a value to alpha_dash.  As
             //! 'a' and 'b' depend on 'ratio', modify them as well.
--- a/src/mlpack/methods/ann/layer/alpha_dropout.hpp
+++ b/src/mlpack/methods/ann/layer/alpha_dropout.hpp
@@ -26,31 +26,31 @@ namespace mlpack {
 namespace ann /** Artificial Neural Network. */ {
 
 /**
-* The alpha - dropout layer is a regularizer that randomly with probability 'ratio'
-* sets input values to alpha_dash.  An affine transformation is applied to the inputs.
-* In the deterministic mode (during testing), the layer just gives the output.
-*
-* Note: During training you should set deterministic to false and during
-* testing you should set deterministic to true.
-*
-* For more information, see the following.
-*
-* @code
-* @article{Klambauer2017,
-*   author  = {Gunter Klambauer and Thomas Unterthiner and
-*              Andreas Mayr},
-*   title   = {Self-Normalizing Neural Networks},
-*   journal = {Advances in Neural Information Processing Systems},
-*   year    = {2017}
-* }
-* }
-* @endcode
-*
-* @tparam InputDataType Type of the input data (arma::colvec, arma::mat,
-*         arma::sp_mat or arma::cube).
-* @tparam OutputDataType Type of the output data (arma::colvec, arma::mat,
-*         arma::sp_mat or arma::cube).
-*/
+ * The alpha - dropout layer is a regularizer that randomly with probability 'ratio'
+ * sets input values to alpha_dash.  An affine transformation is applied to the inputs.
+ * In the deterministic mode (during testing), the layer just gives the output.
+ *
+ * Note: During training you should set deterministic to false and during
+ * testing you should set deterministic to true.
+ *
+ * For more information, see the following.
+ *
+ * @code
+ * @article{Klambauer2017,
+ *   author  = {Gunter Klambauer and Thomas Unterthiner and
+ *              Andreas Mayr},
+ *   title   = {Self-Normalizing Neural Networks},
+ *   journal = {Advances in Neural Information Processing Systems},
+ *   year    = {2017}
+ * }
+ * }
+ * @endcode
+ *
+ * @tparam InputDataType Type of the input data (arma::colvec, arma::mat,
+ *         arma::sp_mat or arma::cube).
+ * @tparam OutputDataType Type of the output data (arma::colvec, arma::mat,
+ *         arma::sp_mat or arma::cube).
+ */
 template <
         typename InputDataType = arma::mat,
         typename OutputDataType = arma::mat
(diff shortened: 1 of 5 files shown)


zoq:
Thanks, for the contribution, do you mind to add a simple test first, this makes the review process a lot easier, since we have something we can run right away.

dakshitagrawal97:
Sure @zoq I will add a test for the layer.  Thanks for the review.

dakshitagrawal97:
@zoq Please have a look.  I have added the implementation tests. :)

src/mlpack/methods/ann/layer/alpha_dropout.hpp
```diff
@@ -0,0 +1,182 @@
...
+    bool deterministic;
+
+    //! Value of alpha for normalized inputs (taken from SELU)
+    const double alpha = 1.6732632423543772848170429916717;
+
+    //! Value of lambda for normalized inputs (taken from SELU)
+    const double lambda = 1.0507009873554804934193349852946;
+
+    //! The low variance value of SELU activation function.
+    double alpha_dash = -alpha*lambda;
```
zoq: What do you think, could run ``Ratio(...)`` right after we initialized ``ratio`` instead, to avoid the redefinition.

src/mlpack/methods/ann/layer/alpha_dropout_impl.hpp
```diff
@@ -0,0 +1,87 @@
...
+    }
+    else
+    {
+        // Set values to alpha_dash with probability ratio.  Then apply affine
+        // transformation so as to keep mean and variance of outputs to their
+        // original values.
+
+        mask = arma::randu< arma::Mat<eT> >(input.n_rows, input.n_cols);
+        mask.transform( [&](double val) { return (val > ratio); } );
+        output = input % mask + alpha_dash * (1 - mask);
```
zoq: What about adding ``* a + b`` here.

src/mlpack/methods/ann/layer/alpha_dropout_impl.hpp
```diff
@@ -0,0 +1,87 @@
...
+template<typename InputDataType, typename OutputDataType>
+template<typename Archive>
+void Alpha_Dropout<InputDataType, OutputDataType>::serialize(
+        Archive& ar,
+        const unsigned int /* version */)
+{
+    ar & BOOST_SERIALIZATION_NVP(ratio);
+
+    // Reset values of a and b
+    a = pow((1 - ratio) * (1 + ratio * pow(alpha_dash, 2)), -0.5);
```
zoq: Either we should check if we are loading, or we could simply serialize ``a`` and ``b`` as well.

src/mlpack/methods/ann/layer/alpha_dropout.hpp
```diff
@@ -0,0 +1,182 @@
...
+* @tparam InputDataType Type of the input data (arma::colvec, arma::mat,
+*         arma::sp_mat or arma::cube).
+* @tparam OutputDataType Type of the output data (arma::colvec, arma::mat,
+*         arma::sp_mat or arma::cube).
+*/
+template <
+        typename InputDataType = arma::mat,
+        typename OutputDataType = arma::mat
+>
+class Alpha_Dropout
```
zoq: Please use camel casing for all names.

src/mlpack/methods/ann/layer/alpha_dropout.hpp
dakshitagrawal97: You mean to run the `Ratio(...)` function right after we initialize the class right?  If so, yes that would be better to do.

src/mlpack/methods/ann/layer/alpha_dropout_impl.hpp
dakshitagrawal97: Okay.  I was implementing it stepwise for ease of reading but this can be done.

src/mlpack/methods/ann/layer/alpha_dropout.hpp
dakshitagrawal97: I will keep that in mind.  Thanks for pointing it out.

src/mlpack/methods/ann/layer/alpha_dropout_impl.hpp
dakshitagrawal97: I guess better to serialize `a` and `b`.  Will change it,

dakshitagrawal97:
@zoq I have fixed the indentation and incorporated the changes you mentioned.  It should be good to go.  Let me know if any changes are required. :)

src/mlpack/methods/ann/layer/alpha_dropout.hpp
```diff
@@ -0,0 +1,183 @@
...
+  double Ratio() const { return ratio; }
+
+  //! Value to be multiplied with x for affine transformation.
+  double A() const { return a; }
+
+  //! Value to be added to a*x for affine transformation.
+  double B() const { return b; }
+
+  //! Value of alpha_dash.
+  double Alpha_Dash() const {return alpha_dash; }
```
rcurtin: Please don't use underscores in variables names---see the style guide for more information.

src/mlpack/methods/ann/layer/alpha_dropout.hpp
```diff
@@ -0,0 +1,183 @@
...
+ * @tparam InputDataType Type of the input data (arma::colvec, arma::mat,
+ *         arma::sp_mat or arma::cube).
+ * @tparam OutputDataType Type of the output data (arma::colvec, arma::mat,
+ *         arma::sp_mat or arma::cube).
+ */
+template <
+        typename InputDataType = arma::mat,
+        typename OutputDataType = arma::mat
+>
+class alphaDropout
```
rcurtin: The class should be named `AlphaDropout` not `alphaDropout` please.

src/mlpack/methods/ann/layer/alpha_dropout.hpp
```diff
@@ -0,0 +1,183 @@
...
+{
+ public:
+  /**
+   * Create the Alpha_Dropout object using the specified ratio.
+   *
+   * @param ratio The probability of setting a value to alpha_dash.
+   */
+  alphaDropout(const double ratio = 0.5);
```
rcurtin: Shouldn't we allow setting `alpha_dash` in the constructor?

src/mlpack/methods/ann/layer/alpha_dropout.hpp
```diff
@@ -0,0 +1,183 @@
...
+ * http://www.opensource.org/licenses/BSD-3-Clause for more information.
+ */
+
+#ifndef MLPACK_METHODS_ANN_LAYER_ALPHA_DROPOUT_HPP
+#define MLPACK_METHODS_ANN_LAYER_ALPHA_DROPOUT_HPP
+
+#include <mlpack/prereqs.hpp>
+
+// To use power function.
+#include <cmath>
```
rcurtin: This should already be included by `mlpack/prereqs.hpp`.

src/mlpack/methods/ann/layer/alpha_dropout.hpp
```diff
@@ -0,0 +1,183 @@
...
+
+namespace mlpack {
+namespace ann /** Artificial Neural Network. */ {
+
+/**
+ * The alpha - dropout layer is a regularizer that randomly with probability 'ratio'
+ * sets input values to alpha_dash.  An affine transformation is applied to the inputs.
+ * In the deterministic mode (during testing), the layer just gives the output.
```
rcurtin: I think that we should clarify this documentation somewhat:

 * There's not clear documentation on what alpha_dash is or why that value is being chosen.  In the paper, it is chosen specifically to keep the outputs normalized.
 * There is no note that this is intended to be used with self-normalizing neural networks.
 * The meaning of `An affine transformation is applied to the inputs` is not clear.
 * I am not sure the comment `During training you should set deterministic to false` is necessary, since the user does not actually do that; the `FNN` and `RNN` classes do.

src/mlpack/methods/ann/layer/alpha_dropout.hpp
```diff
@@ -0,0 +1,183 @@
...
+  bool deterministic;
+
+  //! Value of alpha for normalized inputs (taken from SELU)
+  const double alpha = 1.6732632423543772848170429916717;
+
+  //! Value of lambda for normalized inputs (taken from SELU)
+  const double lambda = 1.0507009873554804934193349852946;
+
+  //! The low variance value of SELU activation function.
+  double alpha_dash = -alpha*lambda;
```
rcurtin: I think you can declare each of these constexprs.

sourabhvarshney111:
@dakshitagrawal97 If you are busy with something else, I would like to work on this further.

src/mlpack/methods/ann/layer/alpha_dropout.hpp
sourabhvarshney111: @rcurtin 
Are we not setting it an initial value using alpha * lambda. May be I missed something. Please correct me if I am wrong.

src/mlpack/methods/ann/layer/alpha_dropout.hpp
sourabhvarshney111: Do you mean using #define for these?

dakshitagrawal97:
@sourabhvarshney111 Sure, feel free to work on the PR. :)

src/mlpack/methods/ann/layer/alpha_dropout.hpp
zoq: constexpr; see http://en.cppreference.com/w/cpp/language/constant_expression

src/mlpack/methods/ann/layer/alpha_dropout.hpp
zoq: Let's use ``-alpha*lambda`` as the default, but as Ryan pointed out a user should be able to overwrite the scaling parameter.

zoq:
Thanks @dakshitagrawal97 @sourabhvarshney111 merged with #1349.

PR #1280, opened 2018-02-26
Add He Initialization and Lecun Normal Initialization

dakshitagrawal97:
This PR adds the following initialization rules with regards to issue #1228:
1. He initialization as given in this [paper](https://www.cv-foundation.org/openaccess/content_iccv_2015/papers/He_Delving_Deep_into_ICCV_2015_paper.pdf). 
2. Lecun Normal initialization as given in this [paper](http://yann.lecun.com/exdb/publis/pdf/lecun-98b.pdf) and also used in [Self Normalizing Networks](https://arxiv.org/pdf/1706.02515.pdf). 

I will add the tests for the same after the implementation is approved.

### DIFF
--- a/src/mlpack/methods/ann/init_rules/he_init.hpp
+++ b/src/mlpack/methods/ann/init_rules/he_init.hpp
@@ -1,0 +1,100 @@
+/**
+ * @file he_init.hpp
+ * @author Dakshit Agrawal
+ *
+ * Intialization rule given by He et. al. for neural networks. The He
+ * initialization initializes weights of the neural network to better
+ * suit the rectified activation units.
+ *
+ * For more information, the following paper can be referred to:
+ *
+ * @code
+ * @article{He2015DelvingDI,
+ * title={Delving Deep into Rectifiers: Surpassing Human-Level Performance on ImageNet Classification},
+ * author={Kaiming He and Xiangyu Zhang and Shaoqing Ren and Jian Sun},
+ * journal={2015 IEEE International Conference on Computer Vision (ICCV)},
+ * year={2015},
+ * pages={1026-1034}}
+ * @endcode
+ *
+ * mlpack is free software; you may redistribute it and/or modify it under the
+ * terms of the 3-clause BSD license.  You should have received a copy of the
+ * 3-clause BSD license along with mlpack.  If not, see
+ * http://www.opensource.org/licenses/BSD-3-Clause for more information.
+ */
+
+#ifndef MLPACK_METHODS_ANN_INIT_RULES_HE_INIT_HPP
+#define MLPACK_METHODS_ANN_INIT_RULES_HE_INIT_HPP
+
+#include <mlpack/prereqs.hpp>
+#include <mlpack/core/math/random.hpp>
+
+using namespace mlpack::math;
+
+namespace mlpack {
+    namespace ann /** Artificial Neural Network. */ {
+
+/**
+ * This class is used to initialize weight matrix with the He initialization rule.
+ */
+        class HeInitialization
+        {
+        public:
+            /**
+             * Initialize the HeInitialization object.
+             *
+             */
+            HeInitialization()
+            {
+                // Nothing to do here.
+            }
+
+            /**
+             * Initialize the elements of the weight matrix with the He initialization
+             * rule.
+             *
+             * @param W Weight matrix to initialize.
+             * @param rows Number of rows.
+             * @param cols Number of columns.
+             */
+            void Initialize(arma::mat& W,
+                            const size_t rows,
+                            const size_t cols)
+            {
+                double_t variance = 2 / rows;
+
+                if (W.is_empty())
+                {
+                    W = arma::mat(rows, cols);
+                }
+
+                W.imbue( [&]() { return arma::as_scalar(RandNormal(0, variance)); } );
+            }
+
+            /**
+             * Initialize the elements of the specified weight 3rd order tensor
+             * with He initialization rule.
+             *
+             * @param W Weight matrix to initialize.
+             * @param rows Number of rows.
+             * @param cols Number of columns.
+             * @param slice Numbers of slices.
+             */
+            void Initialize(arma::cube & W,
+                            const size_t rows,
+                            const size_t cols,
+                            const size_t slices)
+            {
+                W = arma::cube(rows, cols, slices);
+
+                for (size_t i = 0; i < slices; i++) {
+                    Initialize(W.slice(i), rows, cols);
+                }
+            }
+
+        }; // class HeInitialization
+
+    } // namespace ann
+} // namespace mlpack
+
+#endif
--- a/src/mlpack/methods/ann/init_rules/he_init.hpp
+++ b/src/mlpack/methods/ann/init_rules/he_init.hpp
@@ -10,7 +10,8 @@
  *
  * @code
  * @article{He2015DelvingDI,
- * title={Delving Deep into Rectifiers: Surpassing Human-Level Performance on ImageNet Classification},
+ * title={Delving Deep into Rectifiers: Surpassing Human-Level Performance
+ * on ImageNet Classification},
  * author={Kaiming He and Xiangyu Zhang and Shaoqing Ren and Jian Sun},
  * journal={2015 IEEE International Conference on Computer Vision (ICCV)},
  * year={2015},
@@ -42,7 +43,8 @@ namespace ann /** Artificial Neural Network. */ {
  *
  * @code
  * @article{He2015DelvingDI,
- * title={Delving Deep into Rectifiers: Surpassing Human-Level Performance on ImageNet Classification},
+ * title={Delving Deep into Rectifiers: Surpassing Human-Level Performance
+ * on ImageNet Classification},
  * author={Kaiming He and Xiangyu Zhang and Shaoqing Ren and Jian Sun},
  * journal={2015 IEEE International Conference on Computer Vision (ICCV)},
  * year={2015},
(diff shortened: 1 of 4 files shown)


rcurtin:

src/mlpack/methods/ann/init_rules/he_init.hpp
```diff
@@ -0,0 +1,99 @@
...
+     *
+     * @param W Weight matrix to initialize.
+     * @param rows Number of rows.
+     * @param cols Number of columns.
+     */
+    void Initialize(arma::mat& W,
+                    const size_t rows,
+                    const size_t cols)
+    {
+        double_t variance = 2 / rows;
```
zoq: Are you sure this is correct? If I remember right, this should be ``sqrt(2/inputsize)``.

src/mlpack/methods/ann/init_rules/he_init.hpp
```diff
@@ -0,0 +1,99 @@
...
+
+        if (W.is_empty())
+        {
+            W = arma::mat(rows, cols);
```
zoq: Should we use ``set_size`` here?

src/mlpack/methods/ann/init_rules/he_init.hpp
```diff
@@ -0,0 +1,99 @@
...
+ * http://www.opensource.org/licenses/BSD-3-Clause for more information.
+ */
+
+#ifndef MLPACK_METHODS_ANN_INIT_RULES_HE_INIT_HPP
+#define MLPACK_METHODS_ANN_INIT_RULES_HE_INIT_HPP
+
+#include <mlpack/prereqs.hpp>
+#include <mlpack/core/math/random.hpp>
+
+using namespace mlpack::math;
```
rcurtin: We shouldn't use `using` declarations in header files.

src/mlpack/methods/ann/init_rules/he_init.hpp
```diff
@@ -0,0 +1,99 @@
...
+                    const size_t cols)
+    {
+        double_t variance = 2 / rows;
+
+        if (W.is_empty())
+        {
+            W = arma::mat(rows, cols);
+        }
+
+        W.imbue( [&]() { return arma::as_scalar(RandNormal(0, variance)); } );
```
rcurtin: I think you might be able to use Armadillo's `randn()` for this one and then multiply by the variance.

src/mlpack/methods/ann/init_rules/he_init.hpp
dakshitagrawal97: The standard deviation is  `sqrt(2/inputsize)`.

src/mlpack/methods/ann/init_rules/he_init.hpp
dakshitagrawal97: Sorry for that, I will fix it.

src/mlpack/methods/ann/init_rules/he_init.hpp
zoq: So, wouldn't it be more intuitive to use ``arma::randn`` so that we can use the formula from the paper?

src/mlpack/methods/ann/init_rules/he_init.hpp
dakshitagrawal97: The probability distribution will not exactly be Gaussian distribution with variance `tau` if we use a Gaussian distribution with variance 1 and multiply by `tau`.

dakshitagrawal97: The exponential factor in the probability term of Gaussian Distribution will not be taken into account if we take a Gaussian distribution with variance 1 and multiply it by variance `tau`.

src/mlpack/methods/ann/init_rules/he_init.hpp
dakshitagrawal97: Yes, that would be a better option.  Thanks for pointing it out.

src/mlpack/methods/ann/init_rules/he_init.hpp
dakshitagrawal97: `arma::randn` samples from a Gaussian distribution of mean 0 and variance 1.

We want samples from a Gaussian distribution of mean 0 and variance `2/rows`.

dakshitagrawal97:
A type casting error crept up, I will fix that.

src/mlpack/methods/ann/init_rules/he_init.hpp
zoq: Right, I was thinking about the xavier initalization which uses an uniform initialization. Okay, in this case, let's go with the current form, but do you mind to add a simple comment?

dakshitagrawal97: Sure, I will do that.

dakshitagrawal97:
@zoq @rcurtin I have incorporated the changes that you mentioned.

src/mlpack/methods/ann/init_rules/he_init.hpp
rcurtin: If I multiply a random variable X with variance V(X) by some factor c, then the variance V(cX) = c^2 V(X).  This means there should be no problem with using `randn()` here.

src/mlpack/methods/ann/init_rules/he_init.hpp
```diff
@@ -0,0 +1,103 @@
...
+     * @param cols Number of columns.
+     */
+    void Initialize(arma::mat& W,
+                    const size_t rows,
+                    const size_t cols)
+    {
+        // He initialization rule says to initialize weights with random
+        // values taken from a gaussian distribution with mean = 0 and
+        // standard deviation = sqrt(2/rows), i.e. variance = (2/rows).
+        double_t variance =  2.0 / ((double) rows);
```
rcurtin: Why double_t and not double?

dakshitagrawal97: There is no particular reason.  But on searching, I found that it is better to use double_t to make the code more adaptive to compile on both Windows and Unix.  However, it shouldn't make much difference.  Do let me know which one to use.

dakshitagrawal97:
@rcurtin The style is checking is passing.  Is there anything that I am missing?  Please let me know.

src/mlpack/methods/ann/init_rules/he_init.hpp
dakshitagrawal97: Thanks for the information!  So I should multiply by ` sqrt(variance) ` after getting the values from the `randn()` function, right?

dakshitagrawal97:
@rcurtin I found the design guidelines and will update accordingly.

src/mlpack/methods/ann/init_rules/he_init.hpp
rcurtin: >  But on searching, I found that it is better to use double_t to make the code more adaptive to compile on both Windows and Unix.

Can you dig into that and explain a little of what you found please?

In any case we should stick with `double` since it's what's used everywhere else.

dakshitagrawal97: I am sorry, but I could not find any concrete reason for using `double_t` over `double`.  I have changed `double_t` to `double` in the code.

dakshitagrawal97:
@rcurtin @zoq I have incorporated the changes you mentioned.  This should be good to go.  Let me know if any changes are required. :)

src/mlpack/methods/ann/init_rules/he_init.hpp
```diff
@@ -0,0 +1,122 @@
...
+ * @author Dakshit Agrawal
+ *
+ * Intialization rule given by He et. al. for neural networks. The He
+ * initialization initializes weights of the neural network to better
+ * suit the rectified activation units.
+ *
+ * For more information, the following paper can be referred to:
+ *
+ * @code
+ * @article{He2015DelvingDI,
```
zoq: Since you added this to the class itself, I think this is redundant.

src/mlpack/methods/ann/init_rules/he_init.hpp
```diff
@@ -0,0 +1,122 @@
...
+   * @param cols Number of columns.
+   */
+  void Initialize(arma::mat& W,
+                  const size_t rows,
+                  const size_t cols)
+  {
+    // He initialization rule says to initialize weights with random
+    // values taken from a gaussian distribution with mean = 0 and
+    // standard deviation = sqrt(2/rows), i.e. variance = (2/rows).
+    double variance =  2.0 / ((double) rows);
```
zoq: Looks like there is an extra space that can be removed, also should we use ``const``?

src/mlpack/methods/ann/init_rules/he_init.hpp
```diff
@@ -0,0 +1,122 @@
...
+
+    if (W.is_empty())
+    {
+      W.set_size(rows, cols);
+    }
+
+    // Multipling a random variable X with variance V(X) by some factor c,
+    // then the variance V(cX) = (c^2)* V(X).
+    W.imbue( [&]() { return sqrt(variance) * arma::randn();} );
```
zoq: Picky style comment, can you add a space at the end of `` arma::randn();``.

src/mlpack/methods/ann/init_rules/he_init.hpp
```diff
@@ -0,0 +1,122 @@
...
+                  const size_t rows,
+                  const size_t cols,
+                  const size_t slices)
+  {
+    if (W.is_empty())
+    {
+      W.set_size(rows, cols, slices);
+    }
+
+    for (size_t i = 0; i < slices; i++) {
```
zoq: Either remove the ``{}``here or put the opening on a new line; see https://github.com/mlpack/mlpack/wiki/DesignGuidelines#brace-placement for more information.

src/mlpack/tests/init_rules_test.cpp
```diff
@@ -197,6 +199,57 @@ BOOST_AUTO_TEST_CASE(GaussianInitTest)
...
+  initializer.Initialize(weights3d, rows, cols, slices);
+
+  BOOST_REQUIRE_EQUAL(weights.n_rows, rows);
+  BOOST_REQUIRE_EQUAL(weights.n_cols, cols);
+
+  BOOST_REQUIRE_EQUAL(weights3d.n_rows, rows);
+  BOOST_REQUIRE_EQUAL(weights3d.n_cols, cols);
+  BOOST_REQUIRE_EQUAL(weights3d.n_slices, slices);
+}
+
```
zoq: Another picky comment; Looks like an extra line that can be removed.

PR #1281, opened 2018-02-26
Added FlexibleReLU activation layer

aarushgupta:
Added FlexibleRelu activation layer as mentioned in this [paper](https://arxiv.org/abs/1706.08098). This activation function is particularly useful for adding a trainable bias term to each output value from a convolutional filter.

### DIFF
--- a/src/mlpack/methods/ann/layer/flexible_relu.hpp
+++ b/src/mlpack/methods/ann/layer/flexible_relu.hpp
@@ -1,0 +1,194 @@
+/**
+ * @file flexible_relu.hpp
+ * @author Aarush Gupta
+ *
+ * Definition of FlexibleReLU layer as described by
+ * Suo Qiu, Xiangmin Xu and Bolun Cai in
+ * "FReLU: Flexible Rectified Linear Units for Improving Convolutional
+ *   Neural Networks", 2018
+ *
+ * For more information, read the following paper:
+ *
+ * @code
+ * @article{
+ *  author  = {Suo Qiu, Xiangmin Xu and Bolun Cai},
+ *  title   = {FReLU: Flexible Rectified Linear Units for Improving
+ *             Convolutional Neural Networks}
+ *  journal = {arxiv preprint},
+ *  year    = {2018}
+ * }
+ * @endcode
+ *
+ * mlpack is free software; you may redistribute it and/or modify it under the
+ * terms of the 3-clause BSD license.  You should have received a copy of the
+ * 3-clause BSD license along with mlpack.  If not, see
+ * http://www.opensource.org/licenses/BSD-3-Clause for more information.
+ */
+#ifndef MLPACK_METHODS_ANN_LAYER_FLEXIBLERELU_HPP
+#define MLPACK_METHODS_ANN_LAYER_FLEXIBLERELU_HPP
+
+#include <mlpack/prereqs.hpp>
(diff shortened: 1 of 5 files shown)


zoq:
Can you add a test for the FReLU layer?

aarushgupta:
Sure

src/mlpack/methods/ann/layer/flexible_relu.hpp
```diff
@@ -0,0 +1,195 @@
...
+ *
+ * @code
+ * @article{
+ *  author  = {Suo Qiu, Xiangmin Xu and Bolun Cai},
+ *  title   = {FReLU: Flexible Rectified Linear Units for Improving
+ *             Convolutional Neural Networks}
+ *  journal = {arxiv preprint},
+ *  year    = {2018}
+ * }
+ * @endcode
```
rcurtin: Can you add this to the documentation for the class itself also please?  A user may not necessarily find it here.

aarushgupta: Sure.

aarushgupta:
@zoq @rcurtin I have made the changes. Please review it once.

zoq:
Do you think we could reuse the existing relu function and just add ``alpha``?

aarushgupta:
In that case, how would we differentiate between non-trainable(and zero) and trainable alpha?

zoq:
Not sure, I get the issue, I was thinking instead of using the ``Fn()`` from the class, we could use the ``RectifierFunction`` inside the forward pass and add ``alpha``. Maybe I missed something?

aarushgupta:
Yeah, I think we can do that. We won't require flexible_relu.hpp then?

aarushgupta:
Got it. I will remove it and do the necessary changes.

aarushgupta:
@zoq The function RectifierFunction::Fn() does not return a value for matrices and tensors whereas it returns a double type value for single double type values. How should I handle this? Maybe we could check the shape of the input and output variables before calling RectifierFunction::Fn(input, output)?

zoq:
I see, I guess as simple solution is to do something like ``Fn(input, output); output += alpha;``but I'm not sure if this is significantly slower as doing everything in a single line. So I guess what we should do here, is to go the direct way. But I think we can still simplify the code here, e.g. remove the extra methods and perform the operation inside the Forward function itself. Let me know what you think.

aarushgupta:
Sorry for the late reply. I guess in doing so, we will have to write separate cases for each of tensors, matrices, and scalers which is already there in the flexible_relu.hpp file.

zoq:
Just for the matrix object, the scalar method is never used, it's private anyway.

Manthan-R-Sheth:
@zoq 
@aarushgupta any updates on this? I can take this up if its fine.
Thanks.

aarushgupta:
Sure!

sourabhvarshney111:
@Manthan-R-Sheth If you have rebased this PR, can @aarushgupta please close this pr?

Manthan-R-Sheth:
I am fine with whatever @aarushgupta , @zoq and @rcurtin have to say in this.

zoq:
I'll close this one once the other PR is merged.

zoq:
@aarushgupta thanks again, merged with #1341.

PR #1284, opened 2018-02-27
Switch back to the Trusty image

zoq:
This commit does a switch back to the Trusty image, since the Xenial image is unstable.


zoq:
With two approvals I will go ahead and merge this now. Thanks!

PR #1283, opened 2018-02-27
Minor style fixes

zoq:
Minor style fixes.

PR #1285, opened 2018-02-27
Changed style issues which were causing every style build to fail

sourabhvarshney111:
I am fixing the style issues which were causing every build to fail


ShikharJ:
This is already done in #1283 @sourabhvarshney111.

zoq:
Correct, this is fixed in #1283, anyway thanks for looking into the issue.

sourabhvarshney111:
@ShikharJ @zoq Thanks for correcting me. I actually went into working after I saw that every build was failing due to them.

PR #1287, opened 2018-03-01
Add Lambda Parameter to ELU Layer for SELU activation function

dakshitagrawal97:
This PR adds the SELU activation function as introduced in this [NIPS paper](https://papers.nips.cc/paper/6698-self-normalizing-neural-networks.pdf).  The activation function is helpful in keeping the mean and variance of the input invariant.

This has been implemented by adding `lambda` parameter to the `elu.hpp` layer so as to reuse ELU, as ` SELU = lambda * ELU(alpha)` with constant `lambda` and `alpha` for normalized inputs.

Tests for the SELU activation function have been added as well.

**Note:** This function works well only when the input is normalized.  Hence care must be taken that the input is normalized.


src/mlpack/methods/ann/activation_functions/selu_function.hpp
```diff
@@ -0,0 +1,164 @@
...
+ public:
+    /**
+     * Computes the SELU activation function.
+     *
+     * @param x Input data.
+     * @return f(x).
+     */
+    static double Fn(const double x)
+    {
+        if (x < DBL_MAX) {
```
rcurtin: Why is this check necessary?

src/mlpack/methods/ann/activation_functions/selu_function.hpp
```diff
@@ -0,0 +1,164 @@
...
+    /**
+     * Computes the SELU activation function using a dense matrix as input.
+     *
+     * @param x Input data.
+     * @param y The resulting output activation.
+     */
+    template<typename eT>
+    static void Fn(const arma::Mat<eT>& x, arma::Mat<eT>& y)
+    {
+        y = x;
```
rcurtin: This seems unnecessary if you are then looping and setting y(i) = Fn(x(i)).  Maybe you meant to use `y.set_size()` instead?

src/mlpack/tests/activation_functions_test.cpp
```diff
@@ -325,6 +326,56 @@ void CheckPReLUGradientCorrect(const arma::colvec input,
...
+  arma::mat output;
+
+  SELUFunction::Fn(input, output);
+
+  BOOST_REQUIRE_LE(
+     arma::as_scalar(arma::abs(arma::mean(input) - arma::mean(output))), 0.01);
+
+  BOOST_REQUIRE_LE(
+     arma::as_scalar(arma::abs(arma::var(input) - arma::var(output))), 0.01);
+}
```
rcurtin: Would it make sense to also test with data that is not already normalized, and ensure that the output is changed sufficiently?

src/mlpack/methods/ann/activation_functions/selu_function.hpp
```diff
@@ -0,0 +1,164 @@
...
+ *
+ * @code
+ * @article{Klambauer2017,
+ *   author  = {Gunter Klambauer and Thomas Unterthiner and
+ *              Andreas Mayr},
+ *   title   = {Self-Normalizing Neural Networks},
+ *   journal = {Advances in Neural Information Processing Systems},
+ *   year    = {2017}
+ * }
+ * }
```
rcurtin: Can we also put the reference to the paper in the class documentation (below) also please?  A user may not find this documentation if they are, e.g., looking at Doxygen output or something.

dakshitagrawal97: Sure I will do that.

src/mlpack/methods/ann/activation_functions/selu_function.hpp
dakshitagrawal97: SELU does not have an upper bound, hence the need for a check.  The return value that I have set is a random value.

src/mlpack/tests/activation_functions_test.cpp
dakshitagrawal97: Yes, I can add a test in which the mean and variance of non-normalized inputs do not remain the same after passing through the SELU function.

src/mlpack/methods/ann/activation_functions/selu_function.hpp
dakshitagrawal97: Yes, I meant to make y the same size of x.  I will make the change.  Is ` y.set_size( size(X)) ` alright?

zoq: Yes, ``y.set_size( size(X))`` is just fine.

zoq:
What do you think should we reuse the existing ELU layer here? It's just ``lambda * ELU(alpha)``.

dakshitagrawal97:
@zoq The ELU layer has `alpha` as a hyperparameter.  But in a SELU, the `alpha` and `lambda` are constant for normalized inputs.  The ELU layer can be reused, but then its alpha value will have to be set constant, with the condition that the input is normalized.

zoq:
I think another option is to extend the ELU layer; something like:

```c++
ELU(double alpha) : lambda(1)
ELU () : alpha(1.0507009873554802), lambda(1.0507009873554802)
```

and add some comments about the specific parameter, we could even add an alias using ``using SELU = ELU;``, this is just an idea. Let me know what you think.

dakshitagrawal97:
Do you mean to say that we add the `lambda` parameter to the ELU layer and set it to `=1` in the ELU layer's case?  And when we need to use SELU layer, we define a different  `lambda`?  

It might go beyond intuition for the ELU function, but it should be okay if we add a comment of why the lambda parameter has been introduced in the ELU layer. 

If I have misunderstood, please let me know.

zoq:
Right, since we never expose the lambda parameter to the user I think this is fine. I don't mind to implement the function as an individual class, but on the other side, implementing a new function just to multiply with another parameter is probably not ideal. Anyway, let me know what you like to do here.

dakshitagrawal97:
You are right that implementing a layer just to multiply with a parameter is not ideal.  I guess what you have proposed should be done.  I will update the ELU layer accordingly and let you know if anything comes up.  :)

dakshitagrawal97:
@zoq @rcurtin I am not sure why Travis Build is unsuccessful. :(

ShikharJ:
@dakshitagrawal97 That's a server side error, nothing related to the changes here.

src/mlpack/methods/ann/activation_functions/CMakeLists.txt
```diff
@@ -8,7 +8,7 @@ set(SOURCES
   rectifier_function.hpp
   softplus_function.hpp
   swish_function.hpp
-)
+        )
```
zoq: Looks like we could remove the extra whitespace here.

src/mlpack/methods/ann/layer/elu.hpp
```diff
@@ -49,6 +73,56 @@ namespace ann /** Artificial Neural Network. */ {
...
+ *   \end{array}
+ * \right. \\
+ * f'(x) &=& \left\{
+ *   \begin{array}{lr}
+ *     lambda & : x > 0 \\
+ *     lambda * (y + alpha) & : x \le 0
+ *   \end{array}
+ * \right.
+ * @f}
+ *
```
zoq: Can you remove the extra line here.

src/mlpack/methods/ann/layer/elu.hpp
```diff
@@ -49,6 +73,56 @@ namespace ann /** Artificial Neural Network. */ {
...
+ *   author  = {Gunter Klambauer and Thomas Unterthiner and
+ *              Andreas Mayr},
+ *   title   = {Self-Normalizing Neural Networks},
+ *   journal = {Advances in Neural Information Processing Systems},
+ *   year    = {2017}
+ * }
+ * }
+ * @endcode
+ *
+ * NOTE:
```
zoq: Can you use ``Note Make sure to use SELU activation function with normalized inputs and weights initialized with Lecun Normal Initialization.`` here, I think in this case doxygen is able to pick this up and highlight the text.

src/mlpack/methods/ann/layer/elu.hpp
```diff
@@ -62,13 +136,25 @@ class ELU
 {
  public:
   /**
-   * Create the ELU object using the specified parameters. The non zero
+   * Create the ELU object.
+   *
+   * NOTE: Use this constructor for SELU activation function.
+   *
+   */
+
```
zoq: Can you remove the extra line here.

src/mlpack/methods/ann/layer/elu.hpp
```diff
@@ -62,13 +136,25 @@ class ELU
 {
  public:
   /**
-   * Create the ELU object using the specified parameters. The non zero
+   * Create the ELU object.
+   *
+   * NOTE: Use this constructor for SELU activation function.
```
zoq: See my comment above. Also, should we add a alias at the end called SELU something like : ``using SELU = ELU<arma::mat, arma::mat>;``.

src/mlpack/methods/ann/layer/elu.hpp
```diff
@@ -120,28 +209,29 @@ class ELU
...
+   * Computes the value of activation function.
    *
    * @param x Input data.
    * @return f(x).
    */
   double Fn(const double x)
   {
-    if (x < DBL_MAX)
-      return (x > 0) ? x : alpha * (std::exp(x) - 1);
+    if (x < DBL_MAX) {
```
zoq: Can you put the opening ``{`` on a new line; see https://github.com/mlpack/mlpack/wiki/DesignGuidelines#brace-placement for more information.

src/mlpack/tests/activation_functions_test.cpp
```diff
@@ -325,6 +327,92 @@ void CheckPReLUGradientCorrect(const arma::colvec input,
   BOOST_REQUIRE_CLOSE(gradient(0), target(0), 1e-3);
 }
 
+/*
+ * Simple SELU activation test to check whether the mean and variance remain
+ * invariant after passing normalized inputs through the function.
+ *
+ */
+
```
zoq: Can you remove the extra line here.

src/mlpack/tests/activation_functions_test.cpp
```diff
@@ -325,6 +327,92 @@ void CheckPReLUGradientCorrect(const arma::colvec input,
...
+/*
+ * Simple SELU activation test to check whether the mean and variance
+ * vary significantly after passing unnormalized inputs through the function.
+ *
+ */
+
+BOOST_AUTO_TEST_CASE(SELUFunctionUnnormalizedTest)
+{
+  const arma::colvec input("5.96402758 0.9966824 0.99975321 1 \
+                                         7.76159416 -0.76159416 0.96402758 8");
```
zoq: Can you align the line with the first value from the previous?

src/mlpack/tests/activation_functions_test.cpp
```diff
@@ -325,6 +327,92 @@ void CheckPReLUGradientCorrect(const arma::colvec input,
...
+  BOOST_REQUIRE_LE(
+      arma::as_scalar(arma::abs(arma::mean(derivatives) - selu.Lambda())),
+      10e-5);
+
+  input.fill(-1);
+
+  selu.Backward(std::move(input), std::move(error), std::move(derivatives));
+
+  BOOST_REQUIRE_LE(arma::as_scalar(arma::abs(arma::mean(derivatives) -
+                    selu.Lambda()*(selu.Alpha()-1))), 10e-5);
```
zoq: The next line should be tabbed twice; see https://github.com/mlpack/mlpack/wiki/DesignGuidelines#line-length-and-wrapping for more information.

PR #1289, opened 2018-03-02
Add binding tests for mean shift

yamidark:
Part of #1152.
Added mean_shift.cpp with binding tests for mean shift. Let me know if there are other tests that I should include in here.

Also, while testing `mlpack_mean_shift`, there seems to be a bug where if `max_iterations` is too low w.r.t the number of reference points, the program crashes with an `std::invalid_argument`. This is because we may not have found any centroids and run the `KNN.search` with an empty matrix. 
Should I fix this by ensuring we find at least 1 centroid despite the `max_iterations` before breaking from the loop?

### DIFF
--- a/src/mlpack/tests/main_tests/mean_shift_test.cpp
+++ b/src/mlpack/tests/main_tests/mean_shift_test.cpp
@@ -16,8 +16,8 @@ static const std::string testName = "Mean Shift";
 
 #include <mlpack/core.hpp>
 #include <mlpack/core/util/mlpack_main.hpp>
-#include "test_helper.hpp"
 #include <mlpack/methods/mean_shift/mean_shift_main.cpp>
+#include "test_helper.hpp"
 
 #include <boost/test/unit_test.hpp>
 #include "../test_tools.hpp"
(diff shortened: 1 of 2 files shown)


src/mlpack/tests/main_tests/mean_shift_test.cpp
```diff
@@ -0,0 +1,176 @@
...
+  // Input random data points.
+  SetInputParam("input", std::move(x));
+  SetInputParam("labels_only", true);
+
+  mlpackMain();
+
+  // Check number of output points are the same.
+  BOOST_REQUIRE_EQUAL(CLI::GetParam<arma::mat>("output").n_rows, 100);
+  // Check that there is only 1 column containing all the labels.
+  BOOST_REQUIRE_EQUAL(CLI::GetParam<arma::mat>("output").n_cols, 1);
```
rcurtin: Hm, it seems strange to me that this is transposed (I would expect one column and 100 rows).  If you are willing to dig a little deeper, I see that the issue is that `Cluster()` returns an `arma::Col<size_t>` for assignments, but really this should be an `arma::Row<size_t>`.  It will require a little refactoring, then we will need to change this test a bit and ensure that when we run `mlpack_mean_shift` from the command-line with `--output_file` and `--labels_only` set, we get a file with many rows and only one column.

I also tested `mlpack_mean_shift` from the command-line with `--output_file` and `--labels_only`, and it puts all the labels in a single line of an output CSV, which is something that should be fixed.

Don't feel obligated to fix the issue; I can also open an issue if you prefer.  Just let me know if you'd like to handle it here in this PR or not.

yamidark: @rcurtin 

I tried converting the `arma::Col<size_t>` to `arma::Row<size_t>` in `mean_shift_impl.hpp` and `mean_shift_main.cpp` but it doesn't seem to work.
Rather, I tried just removing the `.t()` from `assignments = resultingNeighbours.t()` at the end of `mean_shift_impl.hpp` and it seems to resolve this issue. 
So should I still refactor each `arma::Col<size_t>` to `arma::Row<size_t>` then?

yamidark:
Added test for `in_place` option and resolved the style check issues.

PR #1290, opened 2018-03-02
Enable ccache

zoq:
Let's see if ccache is able to accelerate the travis build.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,7 +10,7 @@ env:
 before_install:
   # boost serialization issue with clang 5; for more information
   # http://lists.llvm.org/pipermail/cfe-dev/2017-September/055571.html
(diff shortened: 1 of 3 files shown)


rcurtin:
Well it's certainly much faster but now there is some kind of serialization issue.  I don't see those problems in the docker nightly matrix build (http://masterblaster.mlpack.org/job/docker%20mlpack%20nightly%20build/) but the versions of clang and boost are different.  (I am still working on getting those containers built correctly...)

sourabhvarshney111:
@zoq Can you tell me what is the purpose of this PR?

zoq:
Acceleration of the travis build, to prevent build timeouts. But I think we can close this one for now.

PR #1291, opened 2018-03-02
Some copy overhead reduced

amanmehta-maniac:
As suggested, I have taken a function at a time and tried you implement the idea mentioned in #1021.
Kindly review and let me know what you think.

### DIFF
--- a/src/mlpack/methods/linear_regression/linear_regression.cpp
+++ b/src/mlpack/methods/linear_regression/linear_regression.cpp
@@ -122,10 +122,10 @@ void LinearRegression::Train(const arma::mat& predictors,
   }
 }
 
-// - Since predictions need modification we will directly pass using std::move
-// - Also since points need to be read-only, we will send it using mat&& instead of mat& (to reduce copy overhead)
 
-void LinearRegression::Predict(arma::mat&& points, arma::vec predictions)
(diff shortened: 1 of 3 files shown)


rcurtin:
Hi there,

This doesn't really work.  We can't pass by value for parameters that will be updated or modified, and passing an rvalue reference instead of a const lvalue reference is a little bit misleading from the function signature side: if it is a const lvalue reference, then the user can expect that nothing will be changed, but with an rvalue reference there is no such expectation.

amanmehta-maniac:
Hey,
For const consistency: We could change the definition of _method(const mat& var)_ to _method(mat var)_. Now during function call, _method(myvar)_, the copy overhead will only occur when myvar is of const datatype. However, unnecessary copy would be avoided. Correct me if I am wrong.

amanmehta-maniac:
Hey @rcurtin, please have a look at my updated PR.

rcurtin:
Hi there @amanmetha-maniac,

I'm sorry to say but this still doesn't really work, although I appreciate the effort that went into it.  I think that it would be useful if you spent some more time with C++ documentation to fully understand the issue, and then some more time with the mlpack codebase to see where the refactoring would actually help before opening a PR.  The comments I wrote in my previous response still apply to the updated code.

I will go ahead and close this one to try and keep the list of open PRs small.

Thanks,

Ryan

PR #1293, opened 2018-03-03
Tests for bindings hmm_loglik, hmm_generate and hmm_viterbi

PlantsAndBuildings:
There wasn't a lot to test since hmm_train tests have already been written -- please let me know if I've missed something and I'll update these tests.


src/mlpack/tests/main_tests/hmm_generate_test.cpp
```diff
@@ -0,0 +1,219 @@
...
+{
+  template<typename HMMType>
+  static void Apply(HMMType& hmm, vector<arma::mat>* trainSeq)
+  {
+    // For now, perform unsupervised (Baum-Welch) training
+    hmm.Train(*trainSeq);
+  }
+};
+
+BOOST_AUTO_TEST_CASE(HMMGenerateCheckDimensionsTest)
```
rcurtin: This looks good to me, but I think it would be worthwhile to check for the `GaussianHMM` and `GaussianMixtureModelHMM` types also, just to make sure the binding is loading and running them properly.

src/mlpack/tests/main_tests/hmm_loglik_test.cpp
```diff
@@ -0,0 +1,165 @@
...
+  {
+    const size_t states = 2;
+
+    // Create the initialized-to-zero model.
+    Create(hmm, *trainSeq, states);
+
+    // Initializing the emission distribution depends on the distribution.
+    // Therefore we have to use the helper functions.
+    RandomInitialize(hmm.Emission());
+  }
```
rcurtin: Since these appear in multiple files, I think it might make sense to declare a file `hmm_test_utils.hpp` in this test directory and move these structures there.

src/mlpack/tests/main_tests/hmm_loglik_test.cpp
```diff
@@ -0,0 +1,165 @@
...
+
+
+  // Set the params for the hmm_loglik invocation
+  SetInputParam("input_model", h);
+  SetInputParam("input", inp);
+
+  mlpackMain();
+
+  double loglik = CLI::GetParam<double>("log_likelihood");
+  BOOST_REQUIRE(loglik <= 0);
```
rcurtin: Are you sure it's guaranteed that the log-likelihood be less than 0?  I think any value that's not NaN or Inf here is fine.

Also, I think it might be worthwhile to check here also with `GaussianHMM` and `GaussianMixtureModelHMM`.

src/mlpack/tests/main_tests/hmm_viterbi_test.cpp
```diff
@@ -0,0 +1,166 @@
...
+  h->PerformAction<Train, std::vector<arma::mat>>(&trainSeq);
+
+  // Set the params for the hmm_viterbi invocation
+  SetInputParam("input_model", h);
+  SetInputParam("input", inp);
+
+  mlpackMain();
+
+  arma::Mat<size_t> out = CLI::GetParam<arma::Mat<size_t> >("output");
+  BOOST_REQUIRE(out.n_rows == inp.n_rows);
```
rcurtin: Are you sure this is right?  I thought `output` should have one row containing the states computed by the Viterbi alignment.  (Maybe `inp.n_rows` is 1, but still, we should probably change the test to be more clear about this.  This would also mean the `n_elem` check isn't necessary.)

src/mlpack/tests/main_tests/hmm_viterbi_test.cpp
```diff
@@ -0,0 +1,166 @@
...
+{
+  template<typename HMMType>
+  static void Apply(HMMType& hmm, vector<arma::mat>* trainSeq)
+  {
+    // For now, perform unsupervised (Baum-Welch) training
+    hmm.Train(*trainSeq);
+  }
+};
+
+BOOST_AUTO_TEST_CASE(HMMViterbiCheckDimensionsTest)
```
rcurtin: Just like the other tests, I think it would also be useful here to test `GaussianHMM` and `GaussianMixtureModelHMM`. :)

src/mlpack/tests/main_tests/hmm_loglik_test.cpp
PlantsAndBuildings: Erm.. Isn't the product of all scaling factors equal to the probability of a given observation sequence ([ref](https://en.wikipedia.org/wiki/Forward%E2%80%93backward_algorithm#Forward_probabilities))? And the log of a probability should be <= 0. I'm fairly certain <= 0 is correct -- let me know what you think.

PlantsAndBuildings: I feel like I should elaborate, just so I'm conceptually clear on what is happening in the algorithm. In
https://github.com/mlpack/mlpack/blob/486be65dab54e8f0fba4066827c8358ffdea3d4e/src/mlpack/methods/hmm/hmm_impl.hpp#L454-L464
The call to `Forward()` populates `forward` and `scales`. Product of all elements in `scales` is:
 
<a href="https://www.codecogs.com/eqnedit.php?latex=P(o_1,o_2&space;...&space;o_k&space;|&space;\pi)" target="_blank"></a>

And `LogLikelihood` returns the log of this probability.

rcurtin: You're right, thanks for the clarification.  I think I wasn't thinking on all cylinders. :)

rcurtin:
I see in the tests that a call to `chol()` has failed; my initial guess (and I haven't dug in) is that this is the Cholesky decomposition of one of the GMMs, and it's not sufficiently positive definite for the decomposition.  My guess is that when training the GMM HMM, some slightly more structured data is needed.  You might take a look at `hmm_test.cpp` and use some of the code there to generate a dataset that you can use.

rcurtin:
The failure here is actually pretty interesting.  I dove in with a debugger because I was so confused by the failing test, and when I step through the code, hmm_train_main.cpp actually appears to be jumping to the `Init` struct in `hmm_test_utils.hpp`.  So I think maybe the linker is doing something wrong here, and you might consider renaming the structs in `hmm_test_utils.hpp` so that the `HMMTrainMainTest` tests pass.

PlantsAndBuildings:
Sorry, I've been meaning to get this done but I've been really busy with work. Thanks a lot for looking into this - I will make the changes ASAP.

PlantsAndBuildings:
@rcurtin - You were absolutely right, changing the struct names fixed the build.

src/mlpack/tests/main_tests/hmm_viterbi_test.cpp
```diff
@@ -0,0 +1,160 @@
...
+  arma::Mat<size_t> out = CLI::GetParam<arma::Mat<size_t> >("output");
+
+  BOOST_REQUIRE_EQUAL(out.n_rows, 1);
+  BOOST_REQUIRE_EQUAL(out.n_cols, inp.n_cols);
+}
+
+BOOST_AUTO_TEST_CASE(HMMViterbiGaussianHMMCheckDimensionsTest)
+{
+  // Train an HMM
+  HMMModel * h = new HMMModel(GaussianHMM);
```
rcurtin: Can you make this read `HMMModel*` please?  (This occurs in a few other places)

src/mlpack/tests/main_tests/hmm_loglik_test.cpp
```diff
@@ -0,0 +1,69 @@
...
+    bindings::tests::CleanMemory();
+    CLI::ClearSettings();
+  }
+};
+
+BOOST_FIXTURE_TEST_SUITE(HMMLoglikMainTest, HMMLoglikTestFixture);
+
+BOOST_AUTO_TEST_CASE(HMMLoglikOutputNegativeTest)
+{
+  // Create an HMMModel
```
rcurtin: I know this is picky, but can you make sure that comments are complete sentences that end with periods?
https://github.com/mlpack/mlpack/wiki/DesignGuidelines#comments

rcurtin:
Thanks again for the contribution; it's nice to have these tests added. :)

PR #1297, opened 2018-03-07
CF decomposition using QuicSVD

haritha1313:
Enables using QuicSVD for decomposition in collaborative filtering.

### DIFF
--- a/src/mlpack/methods/cf/cf_impl.hpp
+++ b/src/mlpack/methods/cf/cf_impl.hpp
@@ -51,8 +51,37 @@ void ApplyFactorizer(FactorizerType& factorizer,
                          && FactorizerTraits<
                          FactorizerType>::UsesQuicSVD>* = 0)
 {
-  arma::mat sigma;
+ // Check if the given rank is valid.
+  size_t r;
+  if (rank > data.n_rows || rank > data.n_cols)
+  {
+    Log::Info << "Rank " << rank << ", given for decomposition is invalid."
+        << std::endl;
+
+    r = (data.n_rows > data.n_cols) ? data.n_cols : data.n_rows;
+    Log::Info << "Setting decomposition rank to " << rank << std::endl;
+  }
+  else
+  {
+    r = rank;
+  }
+
+  // Get svd factorization.
+  arma::vec sigma;
   factorizer.Apply(data, w, sigma, h);
+
+  // Remove the part of w and h depending upon the value of rank.
+  w = w.submat(0, 0, w.n_rows - 1, r - 1);
+  h = h.submat(0, 0, h.n_cols - 1, r - 1);
+
+  // Take only required eigenvalues.
+  sigma = sigma.subvec(0, r - 1);
+
+  // Eigenvalue matrix is multiplied to w.
+  w = w * arma::diagmat(sigma);
+
+  // Take transpose of the matrix h as required by CF module.
+  h = arma::trans(h);
 }
 
 // Apply the factorizer when coordinate lists are not used.
--- a/src/mlpack/methods/cf/cf_impl.hpp
+++ b/src/mlpack/methods/cf/cf_impl.hpp
@@ -22,7 +22,7 @@
 namespace mlpack {
 namespace cf {
 
-// Apply QuicSVD factorizer.
+// Apply the factorizer when coordinate lists are used.
 template<typename FactorizerType>
 void ApplyFactorizer(FactorizerType& factorizer,
                      const arma::mat& data,
@@ -38,7 +38,7 @@ void ApplyFactorizer(FactorizerType& factorizer,
   factorizer.Apply(data, rank, w, h);
 }
 
-// Apply the factorizer when coordinate lists are used.
+// Apply QuicSVD factorizer.
 template<typename FactorizerType>
 void ApplyFactorizer(FactorizerType& factorizer,
                      const arma::mat& data,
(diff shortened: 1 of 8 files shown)


Wenhao-H:
Hi haritha, thanks for adding support for quicSVD! :) 
I have one question here in function `ApplyFactorizer()` for quicSVD:
```
arma::mat sigma;
factorizer.Apply(data, w, sigma, h);
```
In factorization-based collaborative filtering, we compute the factorization `V = W * H` and minimizes reconstruction error, and the resulting matrices can be regarded as users' and items' latent vectors. So after computing SVD `(V = W * sigma * H)` using the dataset, in order to minimize the error, we can compute the square root of singular values in `sigma` and multiply them to `W` and `H` respectively. I haven't tried but I think this might work:
```
arma::mat sigma;
factorizer.Apply(data, w, sigma, h);
arma::mat sigmaSqrt = arma::sqrt(sigma);
w = w * sigmaSqrt;
v = sigmaSqrt * v;
```
Let me know if you disagree or if I can clarify anything :)

src/mlpack/methods/cf/cf_impl.hpp
```diff
@@ -22,7 +22,7 @@
 namespace mlpack {
 namespace cf {
 
-// Apply the factorizer when a coordinate list is used.
+// Apply QuicSVD factorizer.
```
Wenhao-H: misplaced comment?

haritha1313: Hey. Actually it is a new function for handling `QuicSVD` enabled using `UseQuicSVD` parameter in `FactorizerTraits`. The original function with the initial comment is defined beneath that.

Wenhao-H: I think the function here is the original one and the function beneath is the new function for `QuicSVD`.

haritha1313: oops. Thanks a lot for catching that. Must have messed it up when transferring from local copy :D .

haritha1313:
@Wenhao-H I really appreciate the analysis. Its clearer now and I understand you meant the converting of matrices to a form suitable for CF. It is a WIP PR and I haven't linked it to SVDwrapper yet, and I guess thats why you had the confusion.

src/mlpack/methods/quic_svd/quic_svd.hpp
```diff
@@ -67,10 +70,10 @@ class QUIC_SVD
...
    * @param delta Cumulative probability for Monte Carlo error lower bound.
    */
-  QUIC_SVD(const arma::mat& dataset,
+  void Apply(const arma::mat& dataset,
            arma::mat& u,
-           arma::mat& v,
            arma::mat& sigma,
+           arma::mat& v,
            const double epsilon = 0.03,
            const double delta = 0.1);
```
rcurtin: I agree, I think this is a better way to structure this class in general.  But if we have `void Apply()`, I think we can actually remove the `dataset` member from the class itself.  To match the other classes a little better, I might suggest having the constructor take `epsilon` and `delta` as parameters instead of `Apply()`.

Also, I think it would be useful to templatize `Apply()` to accept any `MatType`, so it can work with both dense and sparse matrices.

src/mlpack/methods/cf/cf_impl.hpp
```diff
@@ -31,11 +31,37 @@ void ApplyFactorizer(FactorizerType& factorizer,
...
+  // Get svd factorization.
+  arma::vec sigma;
+  factorizer.Apply(data, w, sigma, h);
+
+  // Sigma matrix is multiplied to w.
+  w = w * sigma;
+
+  // Take transpose of the matrix h as required by CF module.
+  h = arma::trans(h);
+}
```
rcurtin: The main problem I see here is that the rank of the decomposition isn't taken into account.  One way to do this might be the following:

1. Refactor `QUIC_SVD` and `CosineTree` a little so that it terminates when enough vectors are added (see `cosine_tree.cpp:97`), and then make sure `rank` can be passed to `QUIC_SVD`.

2. Add a simple "shim" class that has the function `Apply(data, rank, w, h)` and basically calls the result of this function.

3. Ensure that the CF code uses the shim class instead of QUIC_SVD directly.

That is just one idea; maybe there are some other ideas out there, let me know if you want to do it some other way or see some better solution.

src/mlpack/methods/quic_svd/quic_svd.hpp
```diff
@@ -94,4 +97,20 @@ class QUIC_SVD
...
+namespace mlpack {
+namespace cf {
+
+//! Factorizer traits of Quic SVD.
+template<>
+class FactorizerTraits<mlpack::svd::QUIC_SVD>
+{
+ public:
+  //! Data provided to QuicSVD need not be cleaned.
+  static const bool UsesCoordinateList = true;
```
rcurtin: I don't think QUIC_SVD supports coordinate lists---that means that the data is expected as a matrix filled with rows of `(row, col, value)`.  Basically a coordinate list is just another representation of a sparse matrix.

src/mlpack/methods/quic_svd/quic_svd.hpp
haritha1313: Even I didn't particularly think having dataset as a member is necessary, I just went with the existing changes. I'll make the changes :) .

Issue #1112, opened 2017-09-07
Algorithm yet to be implemented


Rajiv2605:
Hi there,
I am interested in implementing an algorithm or a feature in mlpack which hasn't been implemented yet. It would be great if you could suggest any :smile:

zoq:
Hallo, maybe you can propose something that you like to see in mlpack. Another idea is to take a look at the GSoC ideas list: https://github.com/mlpack/mlpack/wiki/SummerOfCodeIdeas maybe you can find something interesting.

Rajiv2605:
@zoq I'll look into it 😄

Rajiv2605:
@zoq I found this to be interesting: https://github.com/mlpack/mlpack/wiki/SummerOfCodeIdeas#fast-k-centers-algorithm--implementation

rcurtin:
I would strongly suggest finding a different project to pursue; that one is very difficult.  It would probably be best to select some machine learning algorithm that you do not see in mlpack that you think would be useful.

Rajiv2605:
@rcurtin I went through the mplack repository and I don't think Convolutional Neural Network has been implemented yet. Can I work on it?

zoq:
The Conv nets are implemented here: https://github.com/mlpack/mlpack/tree/master/src/mlpack/methods/ann; but if you interested in that direction, there is room for additions and improvements e.g. by implementing a faster conv layer. Let me know what you think.

Rajiv2605:
@zoq Yes, I am interested! Could you tell me more about it?

zoq:
One thing we could do is to remove unnecessary (temporary) copies inside the Forward and Backward function (https://github.com/mlpack/mlpack/blob/master/src/mlpack/methods/ann/layer/convolution_impl.hpp), like ``rotatedFilter`` in the ``Backward`` pass. Another idea is to implement a more efficient convolution operator by exploiting multicore CPU's e.g. http://proceedings.mlr.press/v70/budden17a.html could be interesting.

Rajiv2605:
:+1: I'll look into it.

zoq:
I saw your comment on the BigBatch SGD optimizer PR, if you are interested in implementing another optimizer, let me know there are a lot of interesting papers I really like to see implemented in the future like: YellowFin: https://arxiv.org/abs/1706.03471 or Natasha 2: https://arxiv.org/abs/1708.08694, etc.

Rajiv2605:
[...]

PR #1300, opened 2018-03-09
Saga optimizer added

Prabhat-IIT:
#1112 
As proposed by @zoq I've implemented SAGA optimizer


src/mlpack/core/optimizers/saga/saga_impl.hpp
```diff
@@ -0,0 +1,186 @@
+/**
+ * @file saga_impl.hpp
+ * @author Prabhat Sharma
+ *
+ * Implementation of stochastic variance reduced gradient (SVRG).
```
sourabhvarshney111: Implementation of SAGA.

src/mlpack/core/optimizers/saga/saga_update.hpp
```diff
@@ -0,0 +1,64 @@
+/**
+ * @file saga_update.hpp
+ * @author Prabhat Sharma
+ *
+ * Vanilla update for stochastic variance reduced gradient (SVRG).
```
sourabhvarshney111: Vanilla update for SAGA

src/mlpack/core/optimizers/saga/saga_update.hpp
```diff
@@ -0,0 +1,64 @@
...
+       * @param batchSize Batch size to be used for the given iteration.
+       * @param stepSize Step size to be used for the given iteration.
+       */
+      void Update(arma::mat& iterate,
+                  const arma::mat& avgGradient,
+                  const arma::mat& gradient,
+                  const arma::mat& gradient0,
+                  const double stepSize)
+      {
+        // Perform the vanilla SVRG update.
```
sourabhvarshney111: Perform the vanilla SAGA update

src/mlpack/core/optimizers/saga/saga_update.hpp
```diff
@@ -0,0 +1,64 @@
...
+
+/**
+ * Vanilla update policy for Stochastic variance reduced gradient (SVRG).
+ * The following update scheme is used to update SGD in every iteration:
+ */
+    class SAGAUpdate
+    {
+    public:
+      /**
+       * The Initialize method is called by SVRG Optimizer method before the start
```
sourabhvarshney111: called by SAGA Optimizer

src/mlpack/core/optimizers/saga/saga_update.hpp
```diff
@@ -0,0 +1,64 @@
...
+#ifndef MLPACK_CORE_OPTIMIZERS_SAGA_SAGA_UPDATE_HPP
+#define MLPACK_CORE_OPTIMIZERS_SAGA_SAGA_UPDATE_HPP
+
+#include <mlpack/prereqs.hpp>
+
+namespace mlpack {
+  namespace optimization {
+
+/**
+ * Vanilla update policy for Stochastic variance reduced gradient (SVRG).
```
sourabhvarshney111: Vanilla update for SAGA. I think the comments needs to be updated too. Check them out..

src/mlpack/core/optimizers/saga/saga.hpp
```diff
@@ -0,0 +1,234 @@
...
+ * \f[
+ * | f(A_{j + n}) - f(A_j) | < \epsilon.
+ * \f]
+ *
+ * The parameter \f$\epsilon\f$ is specified by the tolerance parameter to the
+ * constructor; \f$n\f$ is specified by the maxIterations parameter.
+ *
+ * This class is useful for data-dependent functions whose objective function
+ * can be expressed as a sum of objective functions operating on an individual
+ * point.  Then, SVRG considers the gradient of the objective function operating
```
sourabhvarshney111: SAGA considers ...

src/mlpack/core/optimizers/saga/saga.hpp
```diff
@@ -0,0 +1,228 @@
...
+ * such as NCA (see mlpack::nca::NCA), NumFunctions() should return the number
+ * of points in the dataset, and Evaluate(coordinates, 0) will evaluate the
+ * objective function on the first point in the dataset (presumably, the dataset
+ * is held internally in the DecomposableFunctionType).
+ *
+ * For more information, please refer to:
+ *
+ * @code
+ *@incollection{NIPS2014_5258,
+ * title     = {SAGA: A Fast Incremental Gradient Method With Support for Non-Strongly Convex Composite Objectives},
```
zoq: Lines should be no more than 80 characters wide.

src/mlpack/core/optimizers/saga/saga.hpp
```diff
@@ -0,0 +1,228 @@
...
+ *
+ * @tparam UpdatePolicyType update policy used by SAGA during the iterative
+ *     update process. By default vanilla update policy
+ *     (see mlpack::optimization::VanillaUpdate) is used.
+ * @tparam DecayPolicyType Decay policy used during the iterative update
+ *     process to adjust the step size. By default the step size isn't going to
+ *     be adjusted (i.e. NoDecay is used).
+ */
+template<typename UpdatePolicyType = SAGAUpdate,
+    typename DecayPolicyType = NoDecay>
```
zoq: Please align the second template parameter with the first one.

src/mlpack/core/optimizers/saga/saga_impl.hpp
```diff
@@ -0,0 +1,185 @@
...
+        currentFunction = 0;
+
+        // Determine order of visitation.
+        if (shuffle)
+          function.Shuffle();
+      }
+
+      b = math::RandInt(0, numBatches); // Random batch selected
+      // Find the effective batch size (the last batch may be smaller).
+      if (b == numBatches-1)
```
zoq: Please use a space between operations.

src/mlpack/core/optimizers/saga/saga_impl.hpp
```diff
@@ -0,0 +1,185 @@
...
+      {
+        effectiveBatchSize = numFunctions - b*batchSize;
```
zoq: I think we could avoid the condition here, and use something like:

https://github.com/mlpack/mlpack/blob/130918a5b3bf961eb1f88b251ad56035e9ad7e3c/src/mlpack/core/optimizers/sgd/sgd_impl.hpp#L106-L114

src/mlpack/core/optimizers/saga/saga_impl.hpp
```diff
@@ -0,0 +1,185 @@
...
+        currentFunction = b*batchSize;
+      }
+      else
+      {
+        effectiveBatchSize = batchSize;
+        currentFunction = b*effectiveBatchSize;
+      }
+
```
zoq: Picky style comment, remove the extra line.

src/mlpack/core/optimizers/saga/saga_impl.hpp
```diff
@@ -0,0 +1,185 @@
...
+    arma::mat avgGradient(iterate.n_rows, iterate.n_cols);
+
+    for (size_t f = 0, b=0; f < numFunctions;
+      /* incrementing done manually */)
+    {
+      // Find the effective batch size (the last batch may be smaller).
+      effectiveBatchSize = std::min(batchSize, numFunctions - f);
+
+      function.Gradient(iterate, f, gradient, effectiveBatchSize);
+      tableOfGradients.slice(b) = gradient;
```
zoq: We could avoid the extra copy here, if we use ``tableOfGradients.slice(b) `` inside the ``Gradient`` function instead of the ``gradient`` parameter.

src/mlpack/core/optimizers/saga/saga_impl.hpp
```diff
@@ -0,0 +1,185 @@
...
+      {
+        effectiveBatchSize = batchSize;
+        currentFunction = b*effectiveBatchSize;
+      }
+
+
+      // Calculate the gradient of a random function.
+      function.Gradient(iterate, currentFunction, gradient,
+                        effectiveBatchSize);
+      gradient0 = tableOfGradients.slice(b);
```
zoq: We should update the table at each iteration: ``All other entries in the table remain unchanged. `` see section 2.

src/mlpack/core/optimizers/saga/saga_impl.hpp
```diff
@@ -0,0 +1,185 @@
...
+
+      // Use the update policy to take a step.
+      updatePolicy.Update(iterate, avgGradient, gradient, gradient0,
+                          stepSize);
+      // update average
+      avgGradient += (gradient-tableOfGradients.slice(b))/numBatches;
```
zoq: Isn't the main difference between SAG and SAGA, that in case of SAG we take the average and for SAGA we don't?

src/mlpack/core/optimizers/saga/saga_impl.hpp
```diff
@@ -0,0 +1,185 @@
...
+      effectiveBatchSize = std::min(batchSize, numFunctions - f);
+
+      function.Gradient(iterate, f, gradient, effectiveBatchSize);
+      tableOfGradients.slice(b) = gradient;
+
+      f += effectiveBatchSize;
+      avgGradient += gradient;
+      b++;
+    }
+    avgGradient /= (double) numBatches;
```
zoq: We only do this once, and update the ``avgGradient`` over time in the inner loop.

src/mlpack/core/optimizers/saga/saga_impl.hpp
Prabhat-IIT: In both of them we take average gradients but there is a difference in their update step. In sage we divide the present and previous value of the gradient that is changed in present iteration with batchsize but in SAGA we don't divide them by batchsize..we just subtract them and then add average gradient to it. Although there's very little difference in both SAGA estimate turns out to be unbiased after convexity analysis

src/mlpack/core/optimizers/saga/saga_impl.hpp
Prabhat-IIT: we've to find the `avgGradient` only once after which we can simply do like this `avgGradient += (gradient-tableOfGradients.slice(b))/numBatches` which I've done in the inner loop. The idea is to just update the gradient changed

zoq: But ``avgGradient`` is reset in each outer loop, which I think isn't necessary.

Prabhat-IIT: Yup, It isn't necessary. Thanx for the catch :)

src/mlpack/core/optimizers/saga/saga_impl.hpp
Prabhat-IIT: Done

Prabhat-IIT:
@zoq  travis is failing due to time limit exceed. Should I make a minor commit to trigger it once more?

zoq:
If it's necessary we can rerun the travis job, but I think in this case it's fine.

src/mlpack/core/optimizers/saga/CMakeLists.txt
```diff
@@ -0,0 +1,12 @@
+set(SOURCES
+        saga.hpp
+        saga_impl.hpp
+        saga_update.hpp
+        )
```
zoq: This looks strange, can you take a look at the other ``CMakeLists.txt`` files and make some adjustments.

src/mlpack/tests/CMakeLists.txt
```diff
@@ -147,7 +148,7 @@ add_executable(mlpack_test
   main_tests/sparse_coding_test.cpp
   main_tests/hoeffding_tree_test.cpp
   main_tests/hmm_train_test.cpp
-)
+  )
```
zoq: No need to introduce spaces here.

src/mlpack/tests/saga_test.cpp
```diff
@@ -0,0 +1,50 @@
...
+ */
+BOOST_AUTO_TEST_CASE(SAGALogisticRegressionTest)
+{
+  arma::mat data, testData, shuffledData;
+  arma::Row<size_t> responses, testResponses, shuffledResponses;
+
+  LogisticRegressionTestData(data, testData, shuffledData,
+      responses, testResponses, shuffledResponses);
+
+  // Now run big-batch SGD with a couple of batch sizes.
```
zoq: This should be SAGA.

src/mlpack/tests/saga_test.cpp
```diff
@@ -0,0 +1,50 @@
...
+#include <mlpack/core/optimizers/saga/saga.hpp>
+
+#include <boost/test/unit_test.hpp>
+#include "test_tools.hpp"
+#include "test_function_tools.hpp"
+
+using namespace mlpack;
+using namespace mlpack::optimization;
+
+BOOST_AUTO_TEST_SUITE(SAGATest);
```
zoq: Do you think you could add two more tests, you could pick from: https://github.com/mlpack/mlpack/tree/master/src/mlpack/core/optimizers/problems.

src/mlpack/core/optimizers/saga/saga_impl.hpp
```diff
@@ -0,0 +1,180 @@
...
+                        effectiveBatchSize);
+      gradient0 = tableOfGradients.slice(b);
+
+      // Use the update policy to take a step.
+      updatePolicy.Update(iterate, avgGradient, gradient, gradient0,
+                          stepSize);
+      // update average
+      avgGradient += (gradient-tableOfGradients.slice(b))/numBatches;
+
+      // update table of gradients
```
zoq: It's helpful to always have comments that are complete sentences with proper grammar and punctuation; see https://github.com/mlpack/mlpack/wiki/DesignGuidelines#comments for more information.

src/mlpack/core/optimizers/saga/saga_impl.hpp
```diff
@@ -0,0 +1,180 @@
...
+    // Store current parameter for the calculation of the variance reduced
+    // gradient.
+    iterate0 = iterate;
+
+    for (size_t f = 0, b=0, currentFunction = 0; f < numFunctions;
+      /* incrementing done manually */)
+    {
+      b = math::RandInt(0, numBatches); // Random batch selected
+      // Find the effective batch size (the last batch may be smaller).
+      currentFunction = b*batchSize;
```
zoq: Please use a space between operations, there are more examples here, so please check the code for the pattern.

src/mlpack/core/optimizers/saga/saga_impl.hpp
Prabhat-IIT: Sorry, I'll keep that in mind for future!

src/mlpack/tests/saga_test.cpp
Prabhat-IIT: @zoq  I've added Rosenbrock test for both SAGA and SVRG. I think It'll be sufficient? One thing I've noticed in the process is that SAGA requires much smaller step size as compared to SVRG even if both are quite similar. Another thing is that these both cannot pass `sgd_test_function` I don't know why but I've tried experimenting with many values but both couldn't pass till the end

Prabhat-IIT:
@zoq SAGA test on Rosenbrock function is failing sometimes on Travis although it is running fine locally in my system. What can be cause of this?

zoq:
See https://github.com/mlpack/mlpack/issues/922 about more informations regarding failing test, perhaps increasing the number of iterations would be enough to get a stable travis build.

Prabhat-IIT:
@zoq  I think its ready now :) Please let me know if I can do anything else

src/mlpack/tests/saga_test.cpp
```diff
@@ -55,8 +55,9 @@ BOOST_AUTO_TEST_CASE(SAGALogisticRegressionTest)
 BOOST_AUTO_TEST_CASE(GeneralizedRosenbrockTest)
 {
   // Loop over several variants.
-  for (size_t i = 10; i < 20; i += 5)
+  for (size_t i = 5; i < 20; i += 2)
   {
+    mlpack::math::RandomSeed(std::time(NULL));
```
zoq: Please do not use RandomSeed inside the test; see https://github.com/mlpack/mlpack/pull/1306 for more information.

Prabhat-IIT: @zoq the success of SAGA closely depends on the generation of random numbers. Here, I'm setting the seed not to a fixed value but to `time(NULL)` which I think will change with the iteration. SAGA cannot cope up with various variants of Rosenbrock function without the change of `RandomSeed()` Its failing on some variants in the absence of this. But I tried setting the `RandomSeed(time(NULL))` inside SAGA but it didn't helped. Tests are still failing. What should I do?

rcurtin: If SAGA does not reliably converge, this suggests to me that there may be a bug or problem---or perhaps the algorithm parameters need to be tuned.  However, if it is like simulated annealing and it is known that sometimes it simply will not converge because the starting point is bad, I would suggest modifying the test so that you run the optimization multiple times and ensure that at least one of those runs is successful.  You can see the simulated annealing test (`sa_test.cpp`) for an example of that.

Prabhat-IIT: @rcurtin SAGA does not reliably converge only on some of the variants of Rosenbrock functions and that too in the absence of setting the seed. I don't think if there'll be any bug because then the SAGA will not converge at all and wouldn't even pass the `LogisticRegressionTest` I've seen simulated annealing test and I find it suitable for that because `sa` is very sensitive to initialization but I think it shouldn't be the case for SAGA as its a fast gradient based convergence method. The main issue here is that how can we justify SAGA coping up with all the variants if the seed is set and otherwise not? Let us see what @zoq has to say about it

src/mlpack/core/optimizers/saga/saga_impl.hpp
```diff
@@ -124,6 +124,7 @@ double SAGAType<UpdatePolicyType, DecayPolicyType>::Optimize(
     for (size_t f = 0, b=0, currentFunction = 0; f < numFunctions;
       /* incrementing done manually */)
     {
+      mlpack::math::RandomSeed(time(NULL));
```
zoq: math::RandInt should return different values for each step. So this isn't needed, as @rcurtin pointed out we could restart the test, something like https://github.com/mlpack/mlpack/blob/master/src/mlpack/tests/recurrent_network_test.cpp#L83. let me know if I should clarify the point.

Prabhat-IIT: You're trying to say that we should test that SAGA should succeed atleast one time but @zoq  I'm very much curious that why in the world setting just a RandomSeed() can help SAGA to converge on all the variants of `Rosenbrock function`

zoq: ``RandomSeed()`` doesn't help, see  https://travis-ci.org/mlpack/mlpack/builds/355241650?utm_source=github_status&utm_medium=notification each build failed.

zoq: Ideally we could find more robust settings, but if we can't the solution above should work.

Prabhat-IIT: @zoq I'm extremely sorry for the delay. I was not feeling well since yesterday :(

src/mlpack/tests/saga_test.cpp
```diff
@@ -65,10 +75,16 @@ BOOST_AUTO_TEST_CASE(GeneralizedRosenbrockTest)
...
 
-    BOOST_REQUIRE_SMALL(result, 1e-10);
     for (size_t j = 0; j < i; ++j)
-      BOOST_REQUIRE_CLOSE(coordinates[j], (double) 1.0, 1e-3);
+      if (abs(coordinates[j] - 1.0) < 1e-3)
+        indicator++;
+    if (result< 1e-10 && indicator == i)
+    {
+      success++;
+      break;
```
zoq: Maybe I missed something, but it looks like if the condition is true this will break, and we don't test further batches.

Prabhat-IIT: I think since we're testing that SAGA converges on atleast one variant of `Rosenbrock`function then there's no point to run the loop further. Same has been done in this https://github.com/mlpack/mlpack/blob/master/src/mlpack/tests/recurrent_network_test.cpp#L83

zoq: The recurrent network test is different, we only test a single setting, in this case we'd like to know if SAGA can handle different batch sizes, but we will stop after the first one (if successful).

Prabhat-IIT: @zoq I've modified the Test for testing batches. Is this the correct way to test batches? But still SAGA failed one time out of three on Travis. I couldn't even reproduce the results as the Test will always pass on my system. I couldn't figure it out what's wrong. I've again thoroughly checked the algorithm implemented in SAGA but I find it correct. I tried the implementation in python and applied to a dataset the results were pretty good. What do you think is going wrong?

zoq:
I did some tested with the existing code and made some modifications and also added a test for the Sphere function instead of the Rosenbrock function. Note I adjusted the update policy to provide the possibility to add SAG as well. let me know what you think.

https://gist.github.com/zoq/7cb998607946a26d28ac6919362cfae5
https://gist.github.com/zoq/93994b06aab98fb71c3d73dc982faf4d
https://gist.github.com/zoq/64623426c16dbf502654e3d1483efb0a

Prabhat-IIT:
@zoq Thanks for the modifications. You could have just suggested me and I would have made the changes :sweat_smile: . Literally, thanks for looking into it personally despite of busy Gsoc schedule. Just left one comment. Changing the update policy will enable us to support sag also. Should we make a template class specialization in this case for sag and saga or should we make them seperate as per user convinience?

zoq:
What do you think about using an alias:

```c++
using SAGA = AdamType<SagaUpdate>;
using SAG = AdamType<SagUpdate>;
```

let me know what you think.

Prabhat-IIT:
Yes, I think it'll be a good idea but we've had to make separate update policy for SAGA and SAG rather than following the modification you've done in the the update policy.

zoq:
Haven't done the math, but I think implementing another update method in the form of:

```c++
iterate -= stepSize * (avgGradient + (gradient - gradient0) / numBatches);
```

would do it; we have to pass ``numBatches``.

Prabhat-IIT:
@zoq  Sorry for the delay. I was watching so much awesome talks on optimization in deep learning :sweat_smile:

Prabhat-IIT:
@zoq  can you believe just by changing the order of tests leads to successful Travis build. It is beyond my comprehension but yeah now I think it is mergeable at least :)

rcurtin:
Changing the order of the tests will typically mean that the random data for the test is different.  So if the optimizer now fails with different random data, it seems likely to me that the test would fail with a lot of different random seeds.

Can you try running with many different random seeds and seeing if the test fails?  Typically I aim to see zero failures after ~a thousand or more test runs.

(Also if you are using `math::RandomSeed(std::time(NULL))` be sure to wait at least one second between each test run.)

Prabhat-IIT:
@rcurtin  The test passed for 1 to 1000 RandomSeed. Here are the results
https://drive.google.com/file/d/1ASBkJye45C2GPTULWnfPMkfd3ywsz9nm/view?usp=sharing

src/mlpack/core/optimizers/saga/sag_update.hpp
```diff
@@ -0,0 +1,65 @@
...
+   */
+  void Update(arma::mat& iterate,
+              const arma::mat& avgGradient,
+              const arma::mat& gradient,
+              const arma::mat& gradient0,
+              const double stepSize,
+              const size_t batches)
+  {
+    // Perform the vanilla SAG update.
+    iterate -= stepSize * (avgGradient + (gradient - gradient0)/batches);
```
zoq: Picky style comment, can you use a space between operations?

src/mlpack/core/optimizers/saga/saga_update.hpp
```diff
@@ -0,0 +1,65 @@
...
+   * @param gradient0 The old gradient matrix at time t - 1.
+   * @param stepSize Step size to be used for the given iteration.
+   * @param batches Total number of batches present.
+   */
+  void Update(arma::mat& iterate,
+              const arma::mat& avgGradient,
+              const arma::mat& gradient,
+              const arma::mat& gradient0,
+              const double stepSize,
+              const size_t batches)
```
zoq: Can you write ``const size_t /* batches */`` to avoid a compiler warning?

Prabhat-IIT:
@zoq  Is it ready now?

Prabhat-IIT:
@mlpack-jenkins  test this again

Prabhat-IIT:

src/mlpack/core/optimizers/saga/sag_update.hpp
```diff
@@ -0,0 +1,65 @@
+/**
+ * @file sag_update.hpp
+ * @authors Prabhat Sharma and Marcus Edel
+ *
+ * Vanilla update for SAG.
```
zoq: What do you think about ``SAG, a stochastic gradient method for optimizing the sum of a finite set
of smooth functions, where the sum is strongly convex.``

src/mlpack/core/optimizers/saga/sag_update.hpp
```diff
@@ -0,0 +1,65 @@
...
+#ifndef MLPACK_CORE_OPTIMIZERS_SAG_SAG_UPDATE_HPP
+#define MLPACK_CORE_OPTIMIZERS_SAG_SAG_UPDATE_HPP
+
+#include <mlpack/prereqs.hpp>
+
+namespace mlpack {
+namespace optimization {
+
+/**
+* Vanilla update policy for SAG.
```
zoq: See comment above, also do you think we could cite the paper here?

src/mlpack/core/optimizers/saga/saga_impl.hpp
```diff
@@ -0,0 +1,175 @@
...
+    resetPolicy(resetPolicy)
+{ /* Nothing to do. */ }
+
+//! Optimize the function (minimize).
+template<typename UpdatePolicyType, typename DecayPolicyType>
+template<typename DecomposableFunctionType>
+double SAGAType<UpdatePolicyType, DecayPolicyType>::Optimize(
+    DecomposableFunctionType& function,
+    arma::mat& iterate)
+{
```
rcurtin: Can you use the `Function<>` class to provide any methods that `DecomposableFunctionType` might not have?  Take a look at `sgd_impl.hpp`, specifically the lines at the top:

```
  typedef Function<DecomposableFunctionType> FullFunctionType;
  FullFunctionType& f(static_cast<FullFunctionType&>(function));
```

Then, you could call `EvaluateWithGradient()` to get both the objective and the gradient at the same time.  (This can make a big difference!)

src/mlpack/core/optimizers/saga/saga_impl.hpp
```diff
@@ -0,0 +1,175 @@
...
+    function.Gradient(iterate, f, tableOfGradients.slice(b),
+                      effectiveBatchSize);
+
+    f += effectiveBatchSize;
+    avgGradient += tableOfGradients.slice(b);
+    b++;
+  }
+  avgGradient /= (double) numBatches; // Calculate average gradient
+
+  for (size_t i = 0; i < actualMaxIterations; ++i)
```
rcurtin: In this case, is `actualMaxIterations` the number of passes over the training data?  We should be careful here, because with all other optimizers, `maxIterations` refers to the number of points that are seen---so `maxIterations / numFunctions` is the number of epochs.  This could be confusing to a user.

src/mlpack/core/optimizers/saga/saga_impl.hpp
rcurtin: This line also incurs an unnecessary copy; you might consider using the advanced Armadillo constructor to make an alias of this particular slice.

src/mlpack/core/optimizers/saga/saga_impl.hpp
```diff
@@ -0,0 +1,175 @@
...
+    lastObjective = overallObjective;
+
+    // Store current parameter for the calculation of the variance reduced
+    // gradient.
+    iterate0 = iterate;
+
+    for (size_t f = 0, b=0, currentFunction = 0; f < numFunctions;
+      /* incrementing done manually */)
+    {
+      b = math::RandInt(0, numBatches); // Select a random batch
```
rcurtin: Since we already called `function.Shuffle()`, why do we select a random batch here?  They are already in a random order.

src/mlpack/core/optimizers/saga/saga_impl.hpp
```diff
@@ -0,0 +1,175 @@
...
+
+  for (size_t i = 0; i < actualMaxIterations; ++i)
+  {
+    // Calculate the objective function.
+    overallObjective = 0;
+    for (size_t f = 0; f < numFunctions; f += batchSize)
+    {
+      const size_t effectiveBatchSize = std::min(batchSize, numFunctions - f);
+      overallObjective += function.Evaluate(iterate, f, effectiveBatchSize);
+    }
```
rcurtin: If we use `EvaluateWithGradient()` in the loops over batches below, we can cache the objective value of the model (or something very close to it) for way cheaper than this.  Typically looping over all batches to calculate the objective value over all data can be very very time-consuming.  If there is a reason we need the exact calculation here, that's fine with me, but I do want to point out the very large overhead involved.

src/mlpack/tests/saga_test.cpp
```diff
@@ -0,0 +1,125 @@
+/**
+ * @file saga_test.cpp
+ * @authors Prabhat Sharma and Marcus Edel
```
rcurtin: I think you should use this here instead:

```
@author Prabhat Sharma
@author Marcus Edel
```

src/mlpack/core/optimizers/saga/saga_impl.hpp
Prabhat-IIT: I don' think it'll be useful here as I've had to calculate `Gradient()` and `Evaluate()` separately because `gradient` of only a random function is to be computed during one epoch whereas objective is to be computed for every `function`

src/mlpack/core/optimizers/saga/saga_impl.hpp
Prabhat-IIT: Here `actualMaxIterations` is indeed number of passes over dataset. AFAIK the inner loop passes over all the `functions` or `batches` and the outer loop is for passing over all the `functions` repeatedly. Should, I name it something else? I've named it so because optimizers like `sgd` and `svrg` uses this name.

src/mlpack/core/optimizers/saga/saga_impl.hpp
Prabhat-IIT: I'm still not clear should we use `EvaluateWithGradient()` every time in place of `Evaluate()` for faster processing or whenever we want objective value and gradient together in an efficient way?

src/mlpack/core/optimizers/saga/saga_impl.hpp
rcurtin: All of the other optimizers take the `maxIterations` parameter to mean the number of points seen, so I would suggest adjusting this code here so that `maxIterations` means the same thing.

src/mlpack/core/optimizers/saga/saga_impl.hpp
rcurtin: The paper doesn't really mention how to extend support for the algorithm to larger batch sizes, so I think we have some freedom here.  Right now your implementation is kind of structured like this:

```
for (maxIterations)
{
  if (...)
  {
    // print output, check for convergence
  }

  for (batches)
  {
    // select batch randomly, update parameters
  }
}
```

But honestly I think that you could do it more cleanly like this:

```
for (iteration = 0, iteration < maxIterations; iteration += batchSize /* this will have to be handled more carefully */)
{
  if (iteration % numPoints == 0)
  {
    // check for convergence, print output
    // also shuffle the dataset if desired
  }

  // select a random batch and update the gradient
}
```

I am omitting a lot of details but the overall structure is a lot simpler and in my opinion it is easier to understand.

src/mlpack/core/optimizers/saga/saga_impl.hpp
rcurtin: `EvaluateWithGradient()` in many cases will provide you with both the objective and the gradient in less time than it takes to call both `Evaluate()` and `Gradient()`.  So if you can restructure the code to only call `EvaluateWithGradient()` instead of `Gradient()` and then cache the objective values it is returning, you don't have to loop over all of the batches and call `Evaluate()`, which can save a significant amount of processing time.

src/mlpack/core/optimizers/saga/saga_impl.hpp
Prabhat-IIT: I'm extremely sorry for the delay :( But don't you think this structure breaks the algorithm itself? In each `iteration` i.e. in one pass of dataset we've to go over all the batches . In your implementation we are doing only one iteration i.e. going over all the batches only once! Let me know if I can clarify my point!

src/mlpack/core/optimizers/saga/saga_impl.hpp
Prabhat-IIT: Thanks, now its clear to me but I couldn't think of refactoring the code much to use it because here the `gradient` of only a random function is calculated and the random function can repeat itself, so if I try to `Evaluate` and add to the `objectiveValue` then it'll be misleading. The only way I can use it is at the time of initializing gradients so that initial `objectiveValue` will be calculated at the same time using `EvaluateWithGradient()`. It just leads to one less looping over all the functions to calculate objective.

src/mlpack/core/optimizers/saga/saga_impl.hpp
Prabhat-IIT: I'm a little bit confused here. If maxiteration is the number of points seen then what is `numFunctions`. AFAIK after analysing the existing codebase `numFunctions`=`predictors.n_cols` which should be the number of points seen. Moreover, in almost every optimizer we calculate `actualMaxIteration` as `std::numeric_limits<size_t>::max()`whenever `maxiteration` is `0` by this logic it shouldn't be calculated in this way as we always had the number of points seen i.e. the number of points in dataset beforehend, we can then assign the number of points to `actualMaxiteration`.

src/mlpack/core/optimizers/saga/saga_impl.hpp
rcurtin: The choices are this:

 * We calculate the exact objective value at the end of an epoch and this means calling `Evaluate()` on every point in the dataset.  That might be very large and could take nearly as long as the epoch of training itself.
 * We sum the objectives as we perform the epoch, giving us an estimate of the objective value, not an exact value.

Given that the objective is used specifically for convergence, I think it is fine to go with the latter.  This is what Keras does.  If the parameters are not moving much between two epochs, then it is still true that the objective will converge.

Prabhat-IIT: Ohh!! I get it now. It does make sense and ultimately what we are after is convergence and not the objective value itself. I'll make the required changes if you could kindly clarify my above points. Thanks.

src/mlpack/core/optimizers/saga/saga_impl.hpp
```diff
@@ -90,67 +99,72 @@ double SAGAType<UpdatePolicyType, DecayPolicyType>::Optimize(
...
-      function.Gradient(iterate, currentFunction, gradient,
-                        effectiveBatchSize);
-      gradient0 = tableOfGradients.slice(b);
+    // Find the effective batch size; we have to take the minimum of three
+    // things:
+    // - the batch size can't be larger than the user-specified batch size;
+    // - the batch size can't be larger than the number of iterations left
+    //       before actualMaxIterations is hit;
+    // - the batch size can't be larger than the number of functions left.
+    effectiveBatchSize = std::min(
```
zoq: This should use ``batch``instead of ``i`` right?

src/mlpack/core/optimizers/saga/saga_impl.hpp
```diff
@@ -0,0 +1,188 @@
...
+    // Calculate the gradient of a random function.
+    function.Gradient(iterate, batchStart, gradient,
+                      effectiveBatchSize);
+
+    // Use the update policy to take a step.
+    updatePolicy.Update(iterate, avgGradient, gradient,
+                        tableOfGradients.slice(batch),
+                        stepSize, numBatches);
+    // Update the average gradient.
+    avgGradient += (gradient-tableOfGradients.slice(batch))/numBatches;
```
zoq: Not sure I get the reason, for removing the inner loop. Maybe I missed something, but dividing by ``numBatches`` in each iteration would result in a really small gradient.

src/mlpack/core/optimizers/saga/saga_impl.hpp
Prabhat-IIT: I don't think so! In implementations like `sgd` we use `i`

src/mlpack/core/optimizers/saga/saga_impl.hpp
Prabhat-IIT: Inner loop is removed as per the suggestion of @rcurtin

zoq: SAGA and SAG counts to the group of variance reduction optimization like SVRG and SARAH. So I think we should follow the SVRG structure instead of SGD (inner loop).

It'll be sum of gradient of all the batches divided by numbatches

I don't think that is implemented, right now you perform a single batch step, and divide the gradient by numBatches:

```c++
avgGradient += (gradient-tableOfGradients.slice(batch))/numBatches;
```

So, I think we should switch back to the inner loop, and incorporate the changes @rcurtin suggested.

Prabhat-IIT: @zoq Maybe I'm missing something but I think `avgGradient`=(`gradient0` + `gradient1` +....+ `gradientN`)/`numBatches` so suppose `i`th gradient changes where `i` is the random batch selected, then how will `avgGradient` change? We just have to subtract the old `i`th gradient and add the new `i` th gradient i.e. `avgGradient`+= new `i`th gradient/`numBatches` - old `i`th gradient  Let me know if I can clear my point further

Prabhat-IIT: @rcurtin suggested something like 
```for (iteration = 0, iteration < maxIterations; iteration += batchSize /* this will have to be handled more carefully */)
{
  if (iteration % numPoints == 0)
  {
    // check for convergence, print output
    // also shuffle the dataset if desired
  }

  // select a random batch and update the gradient
}
```
which is the present implementation
my previous implementation was like this

```for (maxIterations)
{
  if (...)
  {
    // print output, check for convergence
  }

  for (batches)
  {
    // select batch randomly, update parameters
  }
}```

So, should i finally switch to previous ones? IMO the present implementaion suggested by @rcurtin  should be correct. Initially I thought it was not suited to SAGA but it works just like the previous implementation just making it clean by incorporating the trick of `actualMaxIteration%numFuntions` which is just like completing one epoch.

Prabhat-IIT: @rcurtin  can you also take a look here please :) 
Moreover, @zoq  and @rcurtin I think you won't mind but my practical Vivas and Final semester examinations are going on. It'll soon be over and I'll be back on actively contributing. Sorry, for delaying everything :(

rcurtin:
This would be a part of ensmallen now, so if we like we can adapt and reopen it there. :+1:

Issue #150, opened 2014-12-29
implement KDE


rcurtin:
**Reported by nslagle on 26 Feb 41795516 04:51 UTC**
KDE requires work to compile under the new mlpack::core/logger/CLI scheme.  

Migrated-From: http://trac.research.cc.gatech.edu/fastlab/ticket/148

rcurtin:
A simple implementation can be found as a tree-independent dual-tree algorithm in the paper "Tree-Independent Dual-Tree Algorithms".

serendipity5497:
Hey I would like to work on this.

rcurtin:
Go ahead, I'd suggest starting with a single-tree implementation.  You can take a look at the paper referenced in the comment above and references (or my thesis "Improving Dual-Tree Algorithms" although I can't guarantee it's exciting to read), and at the implementations of single-tree and dual-tree algorithms already in mlpack as an example.

robertohueso:
I would like to start working on this, please @serendipity5497 let me know if you continue to work on this :smiley:

rcurtin:
@robertohueso great work, very happy to have this merged.  I will finally close this issue as fixed after so many years. :+1:

PR #1301, opened 2018-03-09
Implement KDE

robertohueso:
Add Kernel Density Estimation algorithm with dual-trees (Based on the paper "Tree-Independent Dual-Tree Algorithms")
Referring to issue #150 

**Work in progress PR...**


src/mlpack/methods/kde/kde_rules_impl.hpp
```diff
@@ -0,0 +1,129 @@
...
+    //std::cout << referenceNode.Point(0) << "\n";
+    arma::vec center = arma::vec();
+    referenceNode.Center(center);
+    for (size_t i = 0; i < queryNode.NumDescendants(); ++i)
+    {
+      densities(queryNode.Point(i)) +=
+        referenceNode.NumDescendants() *
+        kernel.Evaluate(metric.Evaluate(querySet.col(queryNode.Point(i)),
+                                        center));
+    }
```
rcurtin: We could approximate across the query nodes here too if you wanted, but I think the bound would need to be changed just a little bit or maybe another `if` conditional added.  Then it could just be something like

```
arma::vec queryCenter, referenceCenter;
queryNode.Center(queryCenter);
referenceNode.Center(referenceCenter);
const double kernel = kernel.Evaluate(metric.Evaluate(queryCenter, referenceCenter));
for (size_t i = 0; i < queryNode.NumDescendants(); ++i)
  densities(queryNode.Descendant(i)) += referenceNode.NumDescendants() * kernel;
```

That could be a slightly faster prune.  Also I think the prune you have here should use `densities(queryNode.Descendant(i))` not `densities(queryNode.Point(i))`.

robertohueso: I think I found a bug while working on this, it's more detailed in #1363

robertohueso:
@rcurtin Should we allow CLI / Python binding users to select different metric/kernels/trees?
When I use templates like this it must be defined on compile time right?
```C++
template<typename MetricType = mlpack::metric::EuclideanDistance,
         typename MatType = arma::mat,
         typename KernelType = kernel::GaussianKernel,
         template<typename TreeMetricType,
                  typename TreeStatType,
                  typename TreeMatType> class TreeType = tree::KDTree>
```

src/mlpack/methods/kde/kde.hpp
```diff
@@ -0,0 +1,63 @@
...
+
+  KernelType* kernel;
+
+  Tree* referenceTree;
+
+  double error;
+
+  double bandwidth;
+
+  const int leafSize;
```
rcurtin: We can omit this parameter since it's specific to only some kinds of trees.  If the user wants to use a custom leaf size, they can pass in a pre-built tree.

src/mlpack/methods/kde/kde.hpp
```diff
@@ -0,0 +1,63 @@
...
+ private:
+  const MatType& referenceSet;
+
+  KernelType* kernel;
+
+  Tree* referenceTree;
+
+  double error;
+
+  double bandwidth;
```
rcurtin: I think that this is unused.

src/mlpack/methods/kde/kde_impl.hpp
```diff
@@ -0,0 +1,110 @@
...
+    {
+      density(j) = this->kernel.Evaluate(query.col(i),
+                                         this->referenceSet.col(j));
+    }
+    result(i) = arma::trunc_log(arma::sum(density)) -
+      std::log(referenceSet.n_cols);
+    //this->kernel.Normalizer(query.n_rows);
+    //result(i) = (1/referenceSet.n_cols)*(accumulated);
+  }
+  return result;*/
```
rcurtin: We can later add support for different search modes: single-tree, dual-tree, and brute force, so that we can use all three of these.

src/mlpack/methods/kde/kde_rules.hpp
```diff
@@ -0,0 +1,105 @@
...
+  //! The reference set.
+  const arma::mat& referenceSet;
+
+  //! The query set.
+  const arma::mat& querySet;
+
+  //! Density values.
+  arma::vec& densities;
+
+  const double error;
```
rcurtin: I am understanding this to be the maximum allowed error for a single point.  It's probably best to use relative-value approximation, so, e.g., a user can set `error = 0.05` and this means that the returned KDE value must be within 5% of the true KDE value.

src/mlpack/methods/kde/kde_rules_impl.hpp
```diff
@@ -0,0 +1,130 @@
...
+//! Double-tree scoring function.
+template<typename MetricType, typename KernelType, typename TreeType>
+double KDERules<MetricType, KernelType, TreeType>::
+Score(TreeType& queryNode, TreeType& referenceNode)
+{
+  double score, bound;
+  bound = kernel.Evaluate(queryNode.MinDistance(referenceNode)) -
+          kernel.Evaluate(queryNode.MaxDistance(referenceNode));
+
+  if (bound <= error / referenceSet.n_cols)
```
rcurtin: I think this is absolute error pruning.  I guess we can use both (absolute and relative), and maybe it is easy to give the user the option of which to choose.  I think the "Tree-Independent Dual-Tree Algorithms" paper has relative-value approximation pruning rules---if they are not there let me know, I can dig them up or otherwise write them.  (It is not too hard a change.)

src/mlpack/methods/kde/kde_rules_impl.hpp
```diff
@@ -0,0 +1,130 @@
...
+  {
+    arma::vec queryCenter, referenceCenter;
+    referenceNode.Center(referenceCenter);
+    queryNode.Center(queryCenter);
```
rcurtin: Here we could actually use the tree traits for optimization.  For some tree types, the first point held in the tree is also the center of the node, so we could do

```
if (TreeTraits<TreeType>::FirstPointIsCentroid)
{
  // use queryNode.Dataset().col(queryNode.Point(0)) instead of queryCenter, and
  // referenceNode.Dataset().col(referenceNode.Point(0)) instead of referenceCenter.
}
else
{
  // use queryCenter and referenceCenter
}
```

This could help save a lot of time for some tree types.

src/mlpack/tests/kde_test.cpp
```diff
@@ -0,0 +1,59 @@
...
+                      { 0.4, -3.0},
+                      { 0.0,  0.0},
+                      {-2.1,  1.0} };
+  arma::inplace_trans(reference);
+  arma::inplace_trans(query);
+  arma::vec estimations = arma::vec(query.n_cols, arma::fill::zeros);
+  arma::vec estimations_result = {0.08323668699564207296148765635734889656305,
+                                  0.00167470061366603324010116082831700623501,
+                                  0.07658867126520703394465527935608406551182,
+                                  0.01028120384800740999553525512055784929544};
```
rcurtin: Are these hand-calculated values?

It would also be fine to write another test where you compare against results calculated with a brute force algorithm that's implemented in the test.

src/mlpack/methods/kde/kde_rules_impl.hpp
robertohueso: I think it just has an absolute error approach.
What do you think about something like this?

```C++
double score, bound, min;
bound = kernel.Evaluate(queryNode.MinDistance(referenceNode)) -
        kernel.Evaluate(queryNode.MaxDistance(referenceNode));
min_dist = kernel.Evaluate(queryNode.MinDistance(referenceNode));
if (bound <= (abs_error + rel_error * min_dist) / referenceSet.n_cols)
{...
```

src/mlpack/methods/kde/kde_rules_impl.hpp
robertohueso: I understand, but doesn't it make more sense that the tree handles this?
Currently cover tree (and its derivates) is the only one that has ```FirstPointIsCentroid``` and it does a good job at handling right?

```C++
 //! Get the center of the node and store it in the given vector.
  void Center(arma::vec& center) const
  {
    center = arma::vec(dataset->col(point));
  }
...
...
...
  //! Index of the point in the matrix which this node represents.
  size_t point;
```

Although it's easy to implement, maybe for future cases, but I just had this doubt :smile:

src/mlpack/methods/kde/kde_rules_impl.hpp
rcurtin: I think it would be very odd for someone to use a combined absolute error and relative error approach, but (assuming that prune is correct, and I *think* it is but I have not double-checked) if you are planning to force via warnings or errors that `abs_error = 0` when `rel_error > 0` and `rel_error = 0` when `abs_error > 0` I think that's fine.

src/mlpack/methods/kde/kde_rules_impl.hpp
rcurtin: Yes, but the problem is that when you call `CoverTree::Center()` it will copy the column of the dataset.  If you use `FirstPointIsCentroid`, we can avoid the copy, and this can be a noticeable acceleration.

src/mlpack/methods/kde/kde_rules_impl.hpp
```diff
@@ -0,0 +1,148 @@
...
+template<typename MetricType, typename KernelType, typename TreeType>
+double KDERules<MetricType, KernelType, TreeType>::
+Score(TreeType& queryNode, TreeType& referenceNode)
+{
+  const double maxKernel = kernel.Evaluate(queryNode.MinDistance(referenceNode));
+  const double minKernel = kernel.Evaluate(queryNode.MaxDistance(referenceNode));
+  const double bound = maxKernel - minKernel;
+  double score;
+
+  if (bound <= (absError + relError * minKernel) / referenceSet.n_cols)
```
robertohueso: I think handles relative error tolerance but I still have to re-check

robertohueso:
Work to do:
- [ ] Review KDERules
- [ ] Add more and better tests
- [ ] Implement KDE main
- [ ] Add documentation

rcurtin:
Looking good, let me know if you need any help or when it's ready for a detailed review. :+1:

robertohueso:
Thanks =)

robertohueso:
@rcurtin Does it make sense that I also add support for trees that have ```HasDuplicatedPoints = true```? Or should it just ```Log::Fatal``` when that happens?

robertohueso:
@rcurtin Thanks for your help =). I would like to finish this soon (maybe 1 week or less). For now I will avoid handling ```HasDuplicatedPoints``` (using ```static_assert```) but I would like to implement it in the future.

What I think is left right know:
- I will try to make use of that ```HasSelfChildren``` trick.
- Add documentation.
- Maybe add more options to ```kde_main```
- Maybe add a few more tests.

Anyway I will let you know the moment it's ready so you can take a deeper review.

rcurtin:
Sounds good.  Looking forward to it. :)

src/mlpack/methods/kde/kde_main.cpp
```diff
@@ -0,0 +1,165 @@
...
+using namespace mlpack::util;
+using namespace std;
+
+// Define parameters for the executable.
+PROGRAM_INFO("Kernel Density Estimation",
+    "This program performs a Kernel Density Estimation. KDE is a "
+    "non-parametric way of estimating probability density function. "
+    "For each query point the program will estimate its probability density "
+    "by applying a kernel function to each reference point. Computational "
+    " complexity is O(n^2) but it is optimized by making use of dual-trees. "
```
rcurtin: If you want to point this complexity issue out I might suggest this clarification: `The computational complexity of this is O(N^2) where there are N query points and N reference points, but this implementation will typically see better performance as it uses an approximate dual-tree algorithm for acceleration.`

src/mlpack/methods/kde/kde_main.cpp
```diff
@@ -0,0 +1,165 @@
...
+    "\n\n"
+    "For example, the following will run KDE using the points in "
+    "reference_set.csv and query_set.csv. It will apply an Epanechnikov kernel "
+    "with a 0.2 bandwidth to each reference point and use a KD-Tree for the "
+    "dual-tree optimization. The result will be stored in a densities.csv file "
+    "with a maximum error of 5%"
+    "\n\n"
+    "$ kde --reference reference_set.csv --query query_set.csv --bandwidth 0.2 "
+    "--kernel epanechnikov --tree kd-tree --rel_error 0.05 --output "
+    "densities.csv"
```
rcurtin: So that the Python documentation generates correctly, can you use the `PRINT_CALL()`, `PRINT_PARAM_DATASET()`, and `PRINT_PARAM_STRING()` macros used in the other `_main.cpp` programs?  You could see, for instance, https://github.com/mlpack/mlpack/blob/master/src/mlpack/methods/logistic_regression/logistic_regression_main.cpp#L90.

src/mlpack/methods/kde/kde_main.cpp
```diff
@@ -0,0 +1,165 @@
...
+    "\n\n"
+    "Dual-tree optimization allows to avoid lots of barely relevant "
+    "calculations (as kernel function values decrease with distance) if you "
+    "can afford a little error (you can define how much is the maximum you are "
+    "willing to afford) over the final result. This program runs using an "
```
rcurtin: It might be useful to simply say instead of `if you can afford a little error (...)`, `so it is an approximate computation.  "You can specify the maximum relative or absolute error with the " + PRINT_PARAM_STRING("rel_error") + " and " + PRINT_PARAM_STRING("abs_error") + " options."

src/mlpack/methods/kde/kde_main.cpp
```diff
@@ -0,0 +1,165 @@
...
+    "Euclidean metric. If no output file is specified then it will output the "
+    "result to standard output.");
```
rcurtin: Actually output to stdout will cause some difficulties for bindings to other languages, so I think we need to remove that functionality and for this documentation I'd suggest something more like `"the output density estimations will be stored in " + PRINT_PARAM_STRING("output")`.  (It's hard to write that carefully so it makes sense for both Python and CLI programs.  Maybe you can think of a better way to do it?)

src/mlpack/methods/kde/kde_main.cpp
```diff
@@ -0,0 +1,165 @@
...
+
+// Required options.
+PARAM_MATRIX_IN("reference", "Input dataset to KDE on.", "r");
+PARAM_MATRIX_IN_REQ("query", "Query dataset to KDE on.", "q");
```
rcurtin: Should we allow the user to pass only `reference` like in the `knn` and `kfn` programs?  In these cases the density estimate would be computed for one reference point using all other reference points.  (So it would need some slight changes in the `BaseCase()` rules to avoid using the same reference point as both query and reference in this case.)

src/mlpack/methods/kde/kde_main.cpp
```diff
@@ -0,0 +1,165 @@
...
+               "File containing pre-trained KDE model.",
+               "m");
+PARAM_MODEL_OUT(KDEModel,
+                "output_model",
+                "If specified, the KDE model will be saved to the given file.",
+                "M");
+
+// Configuration options
+PARAM_STRING_IN("kernel", "Kernel to use for the estimation"
+    "('gaussian', 'epanechnikov').", "k", "gaussian");
```
rcurtin: We could also include `laplacian` and `spherical` and `triangular` here.

src/mlpack/methods/kde/kde_main.cpp
```diff
@@ -0,0 +1,165 @@
...
+PARAM_STRING_IN("tree", "Tree to use for the estimation"
+    "('kd-tree', 'ball-tree').", "t", "kd-tree");
```
rcurtin: Up to you, maybe we could add other tree types?  Maybe that causes problems with the breadth-first traversal.

src/mlpack/methods/kde/kde_main.cpp
```diff
@@ -0,0 +1,165 @@
...
+PARAM_DOUBLE_IN("rel_error",
+                "Relative error tolerance for the result",
+                "e",
+                1e-8);
```
rcurtin: Do you think that 1e-8 is a really small value for the default here?  I guess I would think that someone would say, e.g., 0.01 to 0.1 (1% to 10%) for the error they might be willing to tolerate.  I may be misunderstanding, I am not sure if this relative error tolerance is for each query point.  (I think it is...)

src/mlpack/methods/kde/kde_main.cpp
```diff
@@ -0,0 +1,165 @@
...
+PARAM_DOUBLE_IN("abs_error",
+                "Relative error tolerance for the result",
+                "E",
+                0.0);
+PARAM_FLAG("breadth_first", "Use breadth-first traversal instead of depth"
+           "first.", "w");
+// Maybe in the future it could be interesting to implement different metrics.
```
rcurtin: Agreed, but for now I'd say the `KDE<>` C++ class is a good enough way for users to do this.

src/mlpack/methods/kde/kde_main.cpp
```diff
@@ -0,0 +1,165 @@
...
+    kde->BreadthFirst() = breadthFirst;
+
+    // Set KernelType
+    if (kernelStr == "gaussian")
+      kde->KernelType() = KDEModel::GAUSSIAN_KERNEL;
+    else if (kernelStr == "epanechnikov")
+      kde->KernelType() = KDEModel::EPANECHNIKOV_KERNEL;
+    else
+      Log::Fatal << "Input kernel is not valid or not supported yet."
+                 << std::endl;
```
rcurtin: This can cause a subtle memory leak---`Log::Fatal` will terminate, but we should still delete the `kde` object before calling that.  You could also use:

```
RequireParamInSet<string>("kernel", { "gaussian", "epanechnikov" });
```

src/mlpack/methods/kde/kde_model.hpp
```diff
@@ -0,0 +1,190 @@
...
+
+  TreeTypes& TreeType() { return treeType; }
+
+  KernelTypes KernelType() const { return kernelType; }
+
+  KernelTypes& KernelType() { return kernelType; }
+
+  void BuildModel(arma::mat&& referenceSet);
+
+  void Evaluate(arma::mat&& querySet, arma::vec& estimations);
```
rcurtin: I know it is tedious but if you'd like to add a quick documentation bit to each of these functions I'd appreciate it.  If not I can do it during merge, up to you. :)

src/mlpack/methods/kde/kde_model_impl.hpp
```diff
@@ -0,0 +1,195 @@
...
+  other.bandwidth = 1.0;
+  other.relError = 1e-6;
+  other.absError = 0;
+  other.breadthFirst = false;
+  other.kernelType = KernelTypes::GAUSSIAN_KERNEL;
+  other.treeType = TreeTypes::KD_TREE;
+  other.kdeModel = decltype(other.kdeModel)();
+}
+
+inline KDEModel& KDEModel::operator=(KDEModel other)
```
rcurtin: Just to check, I think this covers both the move and copy `operator=()`.  If so that is nice and probably a simplification we could (and should) make in other places too. :+1:

src/mlpack/methods/kde/kde_model_impl.hpp
```diff
@@ -0,0 +1,195 @@
...
+}
+
+inline void KDEModel::BuildModel(arma::mat&& referenceSet)
+{
+  // Clean memory, if necessary.
+  boost::apply_visitor(DeleteVisitor(), kdeModel);
+
+  if (kernelType == GAUSSIAN_KERNEL && treeType == KD_TREE)
+    kdeModel = new KDEType<kernel::GaussianKernel, tree::KDTree>
+      (bandwidth, relError, absError, breadthFirst);
```
rcurtin: Oops, tiny style issue---a new line should be tabbed in twice.  Also enclosing the multi-line inner part of the `if` in braces might be nice...

```
if (...)
{
  two line
      statement
}
```

src/mlpack/methods/kde/kde_rules_impl.hpp
```diff
@@ -0,0 +1,194 @@
...
+template<typename MetricType, typename KernelType, typename TreeType>
+inline force_inline
+double KDERules<MetricType, KernelType, TreeType>::BaseCase(
+    const size_t queryIndex,
+    const size_t referenceIndex)
+{
+  double distance = metric.Evaluate(querySet.col(queryIndex),
+                                    referenceSet.col(referenceIndex));
+  if (tree::TreeTraits<TreeType>::RearrangesDataset)
+    densities(oldFromNewQueries.at(queryIndex)) += kernel.Evaluate(distance);
```
rcurtin: I'm undecided on this one.  I think it may be quicker to do `densities(queryIndex) = ...` here, and then re-map all the indices after the computation is done.  However that is just my intuition, and that may not actually provide any speedup for the KDE problem.  Do you have any thoughts here?

src/mlpack/methods/kde/kde_rules_impl.hpp
```diff
@@ -0,0 +1,194 @@
...
+}
+
+//! Single-tree scoring function.
+template<typename MetricType, typename KernelType, typename TreeType>
+double KDERules<MetricType, KernelType, TreeType>::
+Score(const size_t /* queryIndex */, TreeType& /* referenceNode */)
+{
+  ++scores;
+  traversalInfo.LastScore() = 0.0;
+  return 0.0;
```
rcurtin: Ah, I think maybe this could cause a problem.  The `BinarySpaceTree` dual-tree traverser (which is used for `KDTree` and `BallTree`) uses a call to the single-tree `Score()` after calling the dual-tree `Score()`, once it hits a query leaf node.  But I can see that the tests are passing, so I am a little confused on why the code works despite this.  I'll have to look into it, unless you already know the reason.

src/mlpack/methods/kde/kde_rules_impl.hpp
```diff
@@ -0,0 +1,194 @@
...
+    {
+      kernelValue = EvaluateKernel(queryNode.Point(0), referenceNode.Point(0));
+    }
+    // If a child center is the same as its parent center.
+    else if (tree::TreeTraits<TreeType>::HasSelfChildren)
+    {
+      // Reference node.
+      if (referenceNode.Parent() != NULL &&
+          referenceNode.Point(0) == referenceNode.Parent()->Point(0))
+        referenceCenter = referenceNode.Parent()->Stat().Centroid();
```
rcurtin: This will actually cause a copy of `referenceCenter` instead of reseating the reference; is that what you intended to happen?  If not I guess a pointer should be used for `queryCenter` and `referenceCenter`.  It's also possible to have multiple self-children, so maybe a `while()` is needed here.  I am not sure it is a problem though, it depends on your intentions when writing it.

src/mlpack/methods/kde/kde_rules_impl.hpp
```diff
@@ -0,0 +1,194 @@
...
+        queryNode.Center(queryCenter);
+      }
+      // Compute kernel value.
+      kernelValue = EvaluateKernel(queryCenter, referenceCenter);
+    }
+    // Regular case.
+    else
+    {
+      referenceNode.Center(referenceCenter);
+      queryNode.Center(queryCenter);
```
rcurtin: I am not sure either here, would it be better to compute the center before the call to `Traverse()` entirely?  I think this will cause a lot of redundant computations, since each reference node will have `Score()` called on it many times during the traversal.

src/mlpack/methods/kde/kde_rules_impl.hpp
```diff
@@ -0,0 +1,194 @@
...
+      kernelValue = EvaluateKernel(queryCenter, referenceCenter);
+    }
+
+    #pragma omp for
```
rcurtin: Did you see good speedup with this?  Just curious.  I have played with using OpenMP at higher levels of the tree traversal and it seemed to work ok, so it would be interesting to know if it helps here or if the overhead of giving all the threads work is too much here.

src/mlpack/methods/kde/kde.hpp
```diff
@@ -0,0 +1,245 @@
...
+  void Train(Tree* referenceTree);
+
+  /**
+   * Estimate density of each point in the query set given the data of the
+   * reference set. The result is stored in an estimations vector.
+   *
+   * - Dimension of each point in the query set must match the dimension of each
+   *   point in the reference set.
+   *
+   * - Use std::move if the query set is no longer needed.
```
rcurtin: Do we need to add another overload so that works right?

src/mlpack/methods/kde/kde_impl.hpp
```diff
@@ -0,0 +1,474 @@
...
+                  typename TreeMatType> class TreeType>
+void KDE<MetricType, MatType, KernelType, TreeType>::
+Train(MatType referenceSet)
+{
+  // Check if referenceSet is not an empty set.
+  if (referenceSet.n_cols == 0)
+    throw std::invalid_argument("cannot train KDE model with an empty "
+                                "reference set");
+  this->ownsReferenceTree = true;
+  this->referenceTree = new Tree(referenceSet);
```
rcurtin: I think this should be `new Tree(std::move(referenceSet))`; correct me if I'm wrong. :)

src/mlpack/methods/kde/kde_impl.hpp
```diff
@@ -0,0 +1,474 @@
...
+  {
+    // DualTreeTraverser Depth-First
+    typename Tree::template DualTreeTraverser<RuleType> traverser(rules);
+    traverser.Traverse(*queryTree, *referenceTree);
+  }
+  estimations /= referenceTree->Dataset().n_cols;
+
+  // Normalize if required.
+  if (kernel::KernelTraits<KernelType>::IsNormalized)
+    estimations /= kernel->Normalizer(queryTree->Dataset().n_rows);
```
rcurtin: It might be possible to combine this code with the other overload of `Evaluate()` by simply calling this function from that one.

src/mlpack/methods/kde/kde_impl.hpp
```diff
@@ -0,0 +1,474 @@
...
+{
+  if (relError < 0 || relError > 1)
+    throw std::invalid_argument("Relative error tolerance must be a value "
+                                "between 0 and 1");
+  if (absError < 0)
+    throw std::invalid_argument("Absolute error tolerance must be a value "
+                                "greater or equal to 0");
+  if (relError > 0 && absError > 0)
+    Log::Warn << "Absolute and relative error tolerances will be sumed up"
+              << std::endl;
```
rcurtin: I don't think it's necessary to give a warning here if the documentation is clear about what happens in this case.

src/mlpack/methods/kde/kde.hpp
robertohueso: Solved in 15e0127

src/mlpack/methods/kde/kde_main.cpp
robertohueso: Solved in e97d1bf8c313736948425fa56b3d8de24cc4b4ec

src/mlpack/methods/kde/kde_model.hpp
robertohueso: Solved in aa2e85bff66ab8e80fabf8347bf79a360b039b03

src/mlpack/methods/kde/kde_main.cpp
robertohueso: ```laplacian``` included in 7dbaf03259eb169edb08ea9e299eef19b0ebe3d1

robertohueso: ```spherical``` included in e15ef141398fb43bb05f1e02d736cf2e324db799

robertohueso: ```triangular``` included in e7b7b57e3db03600e605a73d5096ba03f4f4e9e0

src/mlpack/methods/kde/kde_main.cpp
robertohueso: Solved in 89f11f89d48567ec952b2ffacf6bd4979e80a443

src/mlpack/methods/kde/kde_main.cpp
robertohueso: Yes, it's for each query point

src/mlpack/methods/kde/kde_main.cpp
robertohueso: Added Cover-tree, Octree and R-Tree in 9cff9c5b6d7ffc44d8b764652a808d4bda798bfc.
I took the decision not to include breadth-first traversal for the moment since you can do that from the C++ API.
I mark this as solved but I will probably add more trees in the future :smile:

src/mlpack/tests/kde_test.cpp
```diff
@@ -0,0 +1,517 @@
...
+
+  // Check whether results are equal.
+  for (size_t i = 0; i < query.n_cols; ++i)
+    BOOST_REQUIRE_CLOSE(bfEstimations[i], treeEstimations[i], relError);
+}
+
+/**
+ * Test BallTree dual-tree implementation results against brute force results.
+ */
+BOOST_AUTO_TEST_CASE(BallTreeGaussianKDETest)
```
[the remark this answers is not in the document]
robertohueso: Solved in 13de1a340d85766c9c4d94d0f06590488a231845

src/mlpack/methods/kde/kde_rules_impl.hpp
```diff
@@ -0,0 +1,218 @@
...
+      }
+      // Compute kernel value.
+      kernelValue = EvaluateKernel(queryStat.Centroid(),
+                                   referenceStat.Centroid());
+    }
+
+    // Can be paralellized but we avoid it for now because of a compilation
+    // error in visual C++ compiler.
+    // #pragma omp for
+    for (size_t i = 0; i < queryNode.NumDescendants(); ++i)
```
rcurtin: Just a note, the Visual C++ error can be resolved if you use `omp_size_t` in the for loop instead of `size_t`.  (That's a workaround we had to create.)

src/mlpack/tests/kde_test.cpp
```diff
@@ -0,0 +1,621 @@
...
+  BOOST_REQUIRE_EQUAL(kdeBinary.IsTrained(), true);
+
+  // Test if execution gives the same result.
+  arma::vec xmlEstimations = arma::vec(query.n_cols, arma::fill::zeros);
+  arma::vec textEstimations = arma::vec(query.n_cols, arma::fill::zeros);
+  arma::vec binEstimations = arma::vec(query.n_cols, arma::fill::zeros);
+
+  kde.Evaluate(query, xmlEstimations);
+  kde.Evaluate(query, textEstimations);
+  kde.Evaluate(query, binEstimations);
```
rcurtin: Shouldn't these be `kdeXml`, `kdeText`, and `kdeBinary`? :)

robertohueso: Exactly, thanks for pointing it out :)
Solved in 93c8191f5020acae898325e24bba4838671d0a05

src/mlpack/methods/kde/kde_rules_impl.hpp
robertohueso: Fixed in d023b8266be0d9e43d64dfd1f8d0e2b00ab0db4b
It worked before because all leaf nodes were visited.

src/mlpack/methods/kde/kde_model.hpp
```diff
@@ -0,0 +1,369 @@
...
+         template<typename TreeMetricType,
+                  typename TreeStatType,
+                  typename TreeMatType> class TreeType>
+using KDEType = KDE<metric::EuclideanDistance,
+                    arma::mat,
+                    KernelType,
+                    TreeType,
+                    TreeType<metric::EuclideanDistance,
+                             kde::KDEStat,
+                             arma::mat>::template DualTreeTraverser>;
```
rcurtin: What happens if you modify `KDEType` to require the `DualTreeTraverser` type as a third template parameter?  I wonder if this might help fix the Windows build.

robertohueso: Doesn't work either, plus there're other places in mlpack where it's done in the same way and it works (e.g ```neighbor_search/ns_model.hpp```)

rcurtin: Another idea might be to remove lots of types from the `boost::variant` type just to see if that helps.  Often with MSVC finding the actual problem involves a lot of commenting stuff out to pinpoint the problem...

robertohueso: I think I just found the error. Apparently this kind of partial specialization is not allowed by MSVC.
```
  //! DualBiKDE specialized on Gaussian Kernel KDEType.
  template<template<typename TreeMetricType,
                    typename TreeStatType,
                    typename TreeMatType> class TreeType>
  void operator()(KDETypeT<kernel::GaussianKernel, TreeType>* kde) const;
```
I can either fully specialize KDETypeT or not specialize at all.

rcurtin: Hmm, you might be able to sidestep the specialization a bit by not specializing on the kernel type and instead, inside the implementation of `operator()`, do something like...

```
if (std::is_same<KernelType, kernel::GaussianKernel>::value)
{
  // Gaussian kernel implementation...
}
else if (std::is_same<KernelType, ...>::value)
{
  ...
}
```

That might be kind of ugly but it would probably at least work on Windows.  What do you think?

robertohueso: The problem with that implementation is that for some kernels it calls ```kernel.Normalizer()``` and since not every kernel has that method, it won't compile either :(

rcurtin: Maybe you could do something like this?

```
template<typename KernelType>
void DoSomethingWithKernel()
{
  // default implementation
}

template<>
void DoSomethingWithKernel<GaussianKernel>()
{
  // for a kernel that has Normalizer()...
}

template<typename KernelType,
          template<typename TreeMetricType, typename TreeStatType, typename TreeMatType> class TreeType>
void operator()(KDETypeT<KernelType, TreeType>* kde) const
{
  // We don't specialize KernelType here, but instead call out to the correctly specialized version of
  // DoSomethingWithKernel<>()...
  ... DoSomethingWithKernel<GaussianKernel> ...
}
```

That's not tailored to the code you're writing but hopefully you get the idea---we can use a generic `operator()` which in turn calls out to some function that is specialized for certain kernel types.  I think this strategy could work as a workaround... let me know if I overlooked something.

robertohueso: Thanks for the idea! :) Solved in 170039423f97a2b065998ac330422ffa1fd11962

src/mlpack/methods/kde/kde_rules_impl.hpp
robertohueso: I've been playing with this and times are pretty much the same, can't see any big difference actually.

After executing a few quick tests and adding up the times for KDE computation I get this results:

* Without OMP: 369.532391s
* With OMP: 364.340302s

Almost in all tests the OMP execution gives a very slightly better result than the one without OMP, but this difference is so small that I'm not sure if this is enough to justify using OMP.

src/mlpack/tests/kde_test.cpp
```diff
@@ -0,0 +1,517 @@
...
+  kde.Train(referenceTree);
+  kde.Evaluate(queryTree, oldFromNewQueries, estimations);
+  for (size_t i = 0; i < query.n_cols; ++i)
+    BOOST_REQUIRE_CLOSE(estimations[i], estimationsResult[i], 1e-8);
+  delete queryTree;
+  delete referenceTree;
+}
+
+/**
+ * Test dual-tree implementation results against brute force results.
```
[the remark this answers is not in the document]
robertohueso: That would require to add support for single-tree traversals. Should I do it?

robertohueso:
I think this is ready for some review :)

src/mlpack/methods/kde/kde_rules_impl.hpp
rcurtin: I think the issue would be more that we would be accessing `densities()` in a mostly linear order which can really help with caching.  So `BaseCase()` is likely to be called multiple times with adjacent points... i.e. if we are at some node combination we'll loop over all query points and all reference points and call `BaseCase()`, and often these points will be sequential.

Like I said my intuition could be wrong here---it might not make much of a speed difference---but it's probably worth checking. :)

src/mlpack/methods/kde/kde_main.cpp
```diff
@@ -0,0 +1,189 @@
...
+  {
+    arma::mat query = std::move(CLI::GetParam<arma::mat>("query"));
+    kde->Evaluate(std::move(query), estimations);
+  }
+  else
+    kde->Evaluate(estimations);
+
+  // Output results if needed.
+  if (CLI::HasParam("output"))
+    CLI::GetParam<arma::mat>("output") = std::move(estimations);
```
rcurtin: We should transpose this---when I run this I get a big CSV with one line and all the predictions on that line.  Also maybe it would be better to call the output `predictions` or `estimates`?

robertohueso: Solved in 0bb023aba67d0550db02450e9dd59ec700072d5c

src/mlpack/tests/kde_test.cpp
```diff
@@ -0,0 +1,621 @@
...
+  GaussianKernel kernel(kernelBandwidth);
+  KDE<metric::EuclideanDistance,
+      arma::mat,
+      kernel::GaussianKernel,
+      tree::KDTree>
+    kde(metric, kernel, relError, 0.0);
+  kde.Train(reference);
+
+  // When evaluating using the query dataset matrix
+  BOOST_REQUIRE_NO_THROW(kde.Evaluate(query, estimations));
```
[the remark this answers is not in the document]
rcurtin: Ah, ok, thanks---I didn't see that `estimations` was cleared.

src/mlpack/methods/kde/kde_rules_impl.hpp
```diff
@@ -0,0 +1,279 @@
...
+    kernel.Evaluate(referenceNode.MaxDistance(queryPoint));
+  const double bound = maxKernel - minKernel;
+
+  if (tree::TreeTraits<TreeType>::FirstPointIsCentroid &&
+      lastQueryIndex == queryIndex &&
+      traversalInfo.LastReferenceNode() != NULL &&
+      traversalInfo.LastReferenceNode()->Point(0) == referenceNode.Point(0))
+  {
+    // Don't duplicate calculations.
+    newCalculations = false;
```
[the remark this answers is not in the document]
robertohueso: Solved in a3f101287c5c6ed5be50565a87467f63c6c16e19 :)

src/mlpack/methods/kde/kde_rules_impl.hpp
robertohueso: After testing this there's a slight speedup so it's implemented in bd2e970568452930cb2390a83581e4d80710383d :) Thanks for the idea!

src/mlpack/tests/kde_test.cpp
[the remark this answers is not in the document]
robertohueso: Added in 213c2078c3da340d30e2bb4f914ea5cc662565fe

src/mlpack/methods/kde/kde.hpp
```diff
@@ -0,0 +1,256 @@
...
+   *
+   * @param metric Instantiated metric object.
+   * @param kernel Instantiated kernel object.
+   * @param relError Relative error tolerance of the model.
+   * @param absError Absolute error tolerance of the model.
+   */
+  KDE(MetricType& metric,
+      KernelType& kernel,
+      const double relError = 0.05,
+      const double absError = 0);
```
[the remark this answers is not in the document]
robertohueso: Improved in 21e4b89c2ceac1df833f9cbf4a6e39a7624770dd :)

robertohueso:
I've been working today to fix the issues and adding new functionality :)

- Answering your question, yes, I'd like to add this to ```HISTORY.md``` but I'm not really sure how it works, if I should put a new vesion number and add KDE or...?
- I'll try to work on tests (specially on ```main_tests```) asap :)

I'll let you know when I think this is ready for review. Thanks again for the time you're taking with this PR and I hope that pothole gets fixed soon! :)

rcurtin:
> Answering your question, yes, I'd like to add this to HISTORY.md but I'm not really sure how it works, if I should put a new vesion number and add KDE or...?

Sure, you can do that, or I can handle it during merge.  This will definitely be part of a 3.1.0 release since it has major new added functionality.

> I'll try to work on tests (specially on main_tests) asap :)

Sounds good.  Really these shouldn't be too hard to do; you can take a look at the other tests in `main_tests/` to get an idea.  Remember that these tests are to ensure that the input is getting to the binding, not to ensure that KDE itself works---since the KDE's correctness is already tested in `src/mlpack/tests/kde_test.cpp`. :)

> I'll try to work on tests (specially on main_tests) asap :)

Hehe, I don't think that they will fix it anytime soon... I just learned to drive in the other lane when I go on that road. :)

robertohueso:
I think this is closer to be ready :)
@rcurtin Could you have a look at the changes?

src/mlpack/methods/kde/kde_rules_impl.hpp
rcurtin: Did you want to handle this before merge?  I'm not sure if it gave much speedup.

src/mlpack/tests/kde_test.cpp
```diff
@@ -0,0 +1,821 @@
...
+                                bfEstimations,
+                                kernel);
+
+  // Optimized KDE
+  metric::EuclideanDistance metric;
+  KDE<GaussianKernel,
+      metric::EuclideanDistance,
+      arma::mat,
+      tree::StandardCoverTree>
+    kde(relError, 0.0, kernel, KDEMode::DUAL_TREE_MODE, metric);
```
rcurtin: Oops, this should be doubly indented.

src/mlpack/tests/kde_test.cpp
```diff
@@ -0,0 +1,821 @@
...
+  // Dual-tree KDE
+  typedef KDTree<EuclideanDistance, kde::KDEStat, arma::mat> Tree;
+  std::vector<size_t> oldFromNewQueries, oldFromNewReferences;
+  Tree* queryTree = new Tree(query, oldFromNewQueries, 2);
+  Tree* referenceTree = new Tree(reference, oldFromNewReferences, 2);
+  KDE<GaussianKernel,
+      EuclideanDistance,
+      arma::mat,
+      KDTree>
+    kde(relError, 0.0, GaussianKernel(kernelBandwidth));
```
rcurtin: Same here with double indentation, and a few other places in this file too it looks like.

src/mlpack/tests/main_tests/kde_test.cpp
```diff
@@ -0,0 +1,408 @@
...
+ **/
+BOOST_AUTO_TEST_CASE(KDEMainInvalidKernel)
+{
+  arma::mat reference = arma::randu<arma::mat>(2, 10);
+  arma::mat query = arma::randu<arma::mat>(2, 5);
+
+  // Main params
+  SetInputParam("reference", reference);
+  SetInputParam("query", query);
+  SetInputParam("kernel", std::string("linux"));
```
rcurtin: :grinning:

src/mlpack/tests/kde_test.cpp
robertohueso: This is mostly fixed in 7bf036b0e87942fbbd9df7ef077fd31ccc25a6a9 :)

src/mlpack/methods/kde/kde_rules_impl.hpp
robertohueso: I tested both implementations but it didn't gave much speedup and mostly it made it slower so I removed it.

src/mlpack/methods/kde/kde_model.hpp
```diff
@@ -0,0 +1,381 @@
...
+  using KDETypeT = KDEType<KernelType, TreeType>;
+
+  //! Default DualBiKDE on some KDEType.
+  template<typename KernelType,
+           template<typename TreeMetricType,
+                    typename TreeStatType,
+                    typename TreeMatType> class TreeType>
+  void operator()(KDETypeT<KernelType, TreeType>* kde) const;
+
+  // TODO Implement specific cases where a leaf size can be selected.
```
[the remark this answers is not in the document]
rcurtin: I asked the same thing. :)  https://github.com/mlpack/mlpack/pull/1301#discussion_r245723816  I think we'll open an issue after the merge is done and it can get handled later.

src/mlpack/methods/kde/kde.hpp
```diff
@@ -0,0 +1,261 @@
+/**
+ * @file kde.hpp
+ * @author Roberto Hueso (robertohueso96@gmail.com)
```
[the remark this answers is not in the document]
robertohueso: True! Thanks :)

src/mlpack/methods/kde/kde.hpp
```diff
@@ -0,0 +1,261 @@
...
+//! KDEMode represents the ways in which KDE algorithm can be executed.
+enum KDEMode
+{
+  DUAL_TREE_MODE,
+  SINGLE_TREE_MODE
+};
+
+/**
+ * The KDE class is a template class for performing Kernel Density Estimations.
+ * In statistics, kernel density estimation, is a way to estimate the
```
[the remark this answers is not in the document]
robertohueso: Solved in e3a5eee8394fb19086e2633cf8ed5446fd30df44. Thanks :)

src/mlpack/methods/kde/kde.hpp
```diff
@@ -0,0 +1,261 @@
...
+                  typename TreeStatType,
+                  typename TreeMatType> class TreeType = tree::KDTree,
+         template<typename RuleType> class DualTreeTraversalType =
+             TreeType<MetricType,
+                      kde::KDEStat,
+                      MatType>::template DualTreeTraverser,
+         template<typename RuleType> class SingleTreeTraversalType =
+             TreeType<MetricType,
+                      kde::KDEStat,
+                      MatType>::template SingleTreeTraverser>
```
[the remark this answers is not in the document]
robertohueso: Thanks for pointing this out! Actually I forgot to update this. Updated in e3a5eee8394fb19086e2633cf8ed5446fd30df44.

src/mlpack/methods/kde/kde_model.hpp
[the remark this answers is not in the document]
robertohueso: I can handle the issue in the future :) The problem is I won't have time until mid-February probably

src/mlpack/methods/kde/kde_model.hpp
```diff
@@ -0,0 +1,381 @@
...
+                    arma::mat,
+                    TreeType,
+                    TreeType<metric::EuclideanDistance,
+                             kde::KDEStat,
+                             arma::mat>::template DualTreeTraverser,
+                    TreeType<metric::EuclideanDistance,
+                             kde::KDEStat,
+                             arma::mat>::template SingleTreeTraverser>;
+/**
+ * KernerlNormalizer holds a set of methods to normalize estimations applying
```
ShikharJ: I guess this is `KernelNormalizer`?

src/mlpack/methods/kde/kde_model.hpp
```diff
@@ -0,0 +1,381 @@
...
+ * in each case the appropiate kernel normalizer function.
+ */
+class KernelNormalizer
+{
+ private:
+  // SFINAE helper to check if has a Normalizer function.
```
ShikharJ: Please change this to `check if Normalizer function is present` or something of the similar style.

rcurtin:
@ShikharJ I'll wait to merge until you're done reviewing. :+1:

src/mlpack/methods/kde/kde_model.hpp
robertohueso: Exactly :) Fixed typo in 0d5bb4641cc0e1795dc1a17669b4f480275b8854

src/mlpack/methods/kde/kde_model.hpp
robertohueso: Done in 0d5bb4641cc0e1795dc1a17669b4f480275b8854

robertohueso:
Much faster! :+1:

src/mlpack/tests/main_tests/kde_test.cpp
```diff
@@ -0,0 +1,408 @@
...
+BOOST_AUTO_TEST_CASE(KDENoInputData)
+{
+  // No input data is not provided. Should throw a runtime error.
+  Log::Fatal.ignoreInput = true;
+  BOOST_REQUIRE_THROW(mlpackMain(), std::runtime_error);
+  Log::Fatal.ignoreInput = false;
+}
+
+/**
+  * Check that there're as many densities in the result as query points.
```
ShikharJ: I think `the are` is the correct way to say it.

robertohueso: Solved in 20230ae14408ae5655ad8f02be2d0a26b3b612b7.

src/mlpack/methods/kde/kde.hpp
```diff
@@ -0,0 +1,263 @@
...
+   */
+  void Train(MatType referenceSet);
+
+  /**
+   * Trains the KDE model. Sets the reference tree to an already created tree.
+   *
+   * - If TreeTraits<TreeType>::RearrangesDataset is False then it is possible
+   *   to use an empty oldFromNewReferences vector.
+   *
+   * @param referenceTree New already created reference tree.
```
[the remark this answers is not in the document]
robertohueso: Tried to improve it in 20230ae14408ae5655ad8f02be2d0a26b3b612b7 :)

src/mlpack/methods/kde/kde_model.hpp
```diff
@@ -0,0 +1,381 @@
...
+  //! Get the kernel type of the model.
+  KernelTypes KernelType() const { return kernelType; }
+
+  //! Modify the kernel type of the model.
+  KernelTypes& KernelType() { return kernelType; }
+
+  //! Get the mode of the model.
+  KDEMode Mode() const;
+
+  //! Modify de mode of the model.
```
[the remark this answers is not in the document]
robertohueso: Solved in 20230ae14408ae5655ad8f02be2d0a26b3b612b7.

src/mlpack/methods/kde/kde_model.hpp
```diff
@@ -0,0 +1,381 @@
...
+ * It performs a monochromatic KDE.
+ */
+class DualMonoKDE : public boost::static_visitor<void>
+{
+ private:
+  //! Vector to store the KDE results.
+  arma::vec& estimations;
+
+ public:
+  //! Alias template necessary for visual C++ compiler.
```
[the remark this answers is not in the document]
robertohueso: Solved in 20230ae14408ae5655ad8f02be2d0a26b3b612b7.

rcurtin:
:tada: Thanks again for the contribution @robertohueso.  It's great to have this merged and part of the library.

For the leaf size issue, if you like, you can just work on it when you are able.  Or, if you want, you can open an issue detailing what needs to be done and see if anyone is interested in doing it.  Up to you how to handle that. :+1:

PR #1303, opened 2018-03-09
Disallow compilation on clang 3.4 and older.

rcurtin:
mlpack actually compiles fine in those cases but when you try to run the tests there is a segfault that I believe to be a compiler error in initialization.

I spent a while debugging it and wasn't able to really find any reason for the failure, but I think that clang 3.4.2 is so old that we don't need to worry about it.  So I think it is reasonable to issue an error during configuration if clang is too old.

Example failure: http://masterblaster.mlpack.org/view/docker/job/docker%20mlpack%20nightly%20build/240/armadillo_version=armadillo-6.500.5,boost_version=boost_1_49_0,buildmode=debug,compiler_version=llvm-3.4.2,label=linux-amd64/

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,4 +1,4 @@
-cmake_minimum_required(VERSION 2.8.5)
+cmake_minimum_required(VERSION 2.8.10)
 project(mlpack C CXX)
 
 include(CMake/cotire.cmake)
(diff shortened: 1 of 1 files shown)


zoq:
Does e.g. ``mlpack_pca`` fail as well? I guess if only the test bin fails it would be fine to print a warning, but on the other side, we can't be sure it actually works since we can't run the tests.

rcurtin:
I thought the same.  Although the command line programs did appear to work when run by hand, I didn't test that the output was correct.  If clang is generating incorrect code in these cases during the initialization of some boost::serialization singletons (which appears to be where the failure is) I would have some concerns about serialized models.

I guess it is possible we could go with a warning but I would be concerned nobody will see it in the very long cmake output.  What do you think?  I don't have a huge preference either way.

zoq:
The serialization part is a good point, I agree a warning would go under, and it might be easier to upgrade anyway (clang 3.4 isn't super old but I think old enough, released on 02 Jan 2014).

rcurtin:
Ok, I will go ahead and merge immediately.

PR #1264, opened 2018-02-20
Fix random seed for tests

robertohueso:
Reopening #1253 (I deleted the branch)

Referring issue #1227

PR #1306, opened 2018-03-11
Allow setting a fixed random seed in binding tests

Wenhao-H:
This PR adds a `SetFixedRandomSeed()`function to allow tests to set a predefined random seed.

I am writing binding test for CF and found that it's hard to test whether a certain parameter is used when I cannot set a fixed random seed (setting random seed was disabled in #1264). What I was trying to do was to run `mlpack()` multiple times with different values set to the tested parameter each time, to see if the results are different. This is impossible or hard to be done in CF when random seed cannot be set, and it is probably the same in other binding tests in #1152 (like nmf).  So I added this function to allow tests programs to set a **predefined** random seed. In this way every run of `mlpack_test` in a single build still gives the same testing results.

So in a binding test case, the code may look like:
```
BOOST_AUTO_TEST_CASE(...)
{
  ...
  mlpack::math::SetFixedRandomSeed();
  mlpack();
  double result1 = ...
  ResetSettings();
  ...
  mlpack::math::SetFixedRandomSeed();
  mlpack();
  double result2 = ...
  BOOST_REQUIRE_LT(std::abs(result2 - result1), 1e-5);
}
```

### DIFF
--- a/src/mlpack/core/math/random.hpp
+++ b/src/mlpack/core/math/random.hpp
@@ -54,7 +54,7 @@ inline void RandomSeed(const size_t seed)
 #if (BINDING_TYPE == BINDING_TYPE_TEST)
 inline void SetFixedRandomSeed()
 {
-  const size_t seed = 54321;
+  const static size_t seed = rand();
   randGen.seed((uint32_t) seed);
   srand((unsigned int) seed);
   arma::arma_rng::set_seed(seed);
--- a/src/mlpack/core/math/random.hpp
+++ b/src/mlpack/core/math/random.hpp
@@ -49,7 +49,7 @@ inline void RandomSeed(const size_t seed)
 }
 
 /**
- * Set the random seed to a predefined seed.
+ * Set the random seed to a fixed number.
  */
 #if (BINDING_TYPE == BINDING_TYPE_TEST)
 inline void SetFixedRandomSeed()
--- a/src/mlpack/core/math/random.hpp
+++ b/src/mlpack/core/math/random.hpp
@@ -50,9 +50,13 @@ inline void RandomSeed(const size_t seed)
 
 /**
  * Set the random seed to a fixed number.
+ * This function is used in binding tests to set a fixed random seed before
+ * calling mlpack(). In this way we can test whether a certain parameter makes
+ * a difference to execution of CLI binding.
+ * Refer to pull request #1306 for discussion on this function.
  */
 #if (BINDING_TYPE == BINDING_TYPE_TEST)
-inline void SetFixedRandomSeed()
+inline void FixedRandomSeed()
 {
   const static size_t seed = rand();
   randGen.seed((uint32_t) seed);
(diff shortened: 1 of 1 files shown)


rcurtin:
For the sake of debugging, it's definitely useful to be able to set a random seed for tests, but I think in general a test should not be dependent on the random seed itself.  I think maybe for the particular problem you are dealing with with CF, maybe it is better to try and think of a strategy that does not affect the random seed, or some other way to get around the need for a random seed.  What do you think?

Wenhao-H:
I agree it's better if tests don't depend on setting random seeds. But for the problem i'm dealing with in CF binding test, I don't think there is an easy way to work around. All amf-based factorizers provided by CF cli binding use random initializations. Unlike in NMF binding test where we can provide the option of setting the initial matrices, in CF giving user the option to set initial matrices may not be a good idea. So I can't think of a way to test whether some parameters (like `min_residue`, `iteration_only_termination`) are actually used in the program execution. If we prefer not allowing tests to set random seed I will leave out the tests for these parameters at the moment. Is that alright?

rcurtin:
I see what you mean.  I agree that allowing users to specify initial matrices isn't a great idea, and I also can't think of a good way to test `min_residue` or `iteration_only_termination`.

Just for clarity, one of the reasons that it's not a great idea to set the random seed at all during the tests is for the matrix build:

http://masterblaster.mlpack.org/job/docker%20mlpack%20weekly%20build/
(that's one example, there are a couple other matrix builds)

For each of those builds, since the environment is different, the 'default' random seed ends up being different.  So it is easy to take a look at the matrix and get an idea of whether or not any tests will randomly fail for users.  If, somewhere in the tests, we set the random seed to a specific value, then any tests run after that test on different systems will have exactly the same results, so the matrix build will not show intermittently failing tests (which can sometimes be tolerances that are set wrong and can also sometimes be unexpected memory issues).

So, how about for now, we leave a TODO in the CF tests that `min_residue` and `iteration_only_termination` aren't fully tested?

Wenhao-H:
@rcurtin I think the new commit can solve the problem you clarified. Is there any other reason why it's still not a good idea to do this?

src/mlpack/core/math/random.hpp
```diff
@@ -48,6 +48,19 @@ inline void RandomSeed(const size_t seed)
   #endif
 }
 
+/**
+ * Set the random seed to a fixed number.
+ */
+#if (BINDING_TYPE == BINDING_TYPE_TEST)
+inline void SetFixedRandomSeed()
```
zoq: Not sure about the method name, ``Set`` is rarely used over the codebase. Perhaps ``FixedRandomSeed`` is just fine as well?

src/mlpack/core/math/random.hpp
```diff
@@ -48,6 +48,19 @@ inline void RandomSeed(const size_t seed)
   #endif
 }
 
+/**
+ * Set the random seed to a fixed number.
```
zoq: Perhaps it's a good idea to add a reference to the PR here, ideally we use the function only for specific cases.

Wenhao-H:
@zoq Modified to your suggestion. Thanks for reviewing!

rcurtin:
Thanks for the contribution---this is a nice improvement. :)

PR #1309, opened 2018-03-12
Reduced error pruning (Regularisation) for decision tree

Manthan-R-Sheth:
This is a work in progress PR for implementing the pruning of decision tree which acts like a regulariser for decision tree. This PR implements reduced error (minimum error) tree pruning algorithm on validation set. 
It begins pruning from leaf nodes and recursively prunes each node in the tree if the accuracy on the validation set increases, thereby giving minimum error tree as the end result.

TODO list :- 
- [x]  Include a `count` parameter in each node to calculate the exact `classProbabilities`and `dimensionTypeOrMajorityClass` for leaf nodes.

- [ ]  Reduce the number of local copies due to various recursive calls.

- [x] Write tests for testing the pruning algorithm.

- [ ] Make `ValidateScore` a template parameter to include smallest tree pruning algorithm (and many more) as well.

Suggestions on this first prototype would be very helpful, @rcurtin @zoq


Manthan-R-Sheth:
@zoq @rcurtin 
At the time of pruning, the current node should be assigned the majority class of the data that it holds and not the first child's majority class. Can you help me as to how can I implement that.
I am thinking of holding a count at each node and then using the count and classProbabilities, calculate the majority class.
Is this good enough?

sourabhvarshney111:
@Manthan-R-Sheth I would like to work with you. Can you provide me with the source you are using?

Manthan-R-Sheth:
I am not using one particular source, its a combination of multiple resources. Also, I think the basic pruning is in place, its just that the pruned node has to be associated to a correct class now and then the TODO list items. I will work on those parts once the proposal work is done.
Happy to collaborate with you.
@sourabhvarshney111

Manthan-R-Sheth:
[This](https://github.com/ryanmadden/decision-tree/blob/master/decision-tree.py) could be a good starting point. After getting hold of this, try to go through the code in this PR. That should be good enough to get you started.
@sourabhvarshney111

sourabhvarshney111:
Thanks @Manthan-R-Sheth. I would go through this asap.

sourabhvarshney111:
@Manthan-R-Sheth Can you provide me write access for your prune tree branch?

Manthan-R-Sheth:
@sourabhvarshney111 how about collaborating over pull requests. We can discuss it and then merge to the branch in case it looks fine to both of us?

sourabhvarshney111:
Yes we can also do that. But Can you please update the repo. It seems it is somewhat outdated.

sourabhvarshney111:
@Manthan-R-Sheth I tried to open a PR in your repo, but that doesn't seem to work. Can you help me out?

Manthan-R-Sheth:
sure, let me check once, i will get back to you.

Manthan-R-Sheth:
@sourabhvarshney111 You will have to fork my repo (which will create a remote) and then make a PR to it.

Manthan-R-Sheth:
@sourabhvarshney111 check out the changes I have made. This completed the first task of the TODO list.
We will have to work on tests now.

@zoq @rcurtin Can you review this please, now.

sourabhvarshney111:
I think we need to test for all the variants which are used in decision tree and check that pruning produces less or equal error on validation set than the decision tree without pruning. What say @Manthan-R-Sheth?

sourabhvarshney111:
@Manthan-R-Sheth I will try to do it asap.

sourabhvarshney111:
@Manthan-R-Sheth  I think we need not test for every test above because all the trees will not be pruned like the case when data set is perfect. What say?

Manthan-R-Sheth:
@sourabhvarshney111 
I have added a test to give you an idea.I will be adding on more tests to check the implementation. I will also debug my implementation in case the test fails.

sourabhvarshney111:
Thanks once again : )

sourabhvarshney111:
@Manthan-R-Sheth Looks like we need to pass wdcorrect as best Score otherwise if pruning is not done then best Score would be zero. What say?

Manthan-R-Sheth:
@zoq @rcurtin Pruning tests passes now.
@sourabhvarshney111  ya that can be done, however even if that is not done it won't be a problem as the decision tree has memorized the entire training set and so the first time it reaches the leaf(during pruning), new best score will be calculated and so it won't be zero.
Passing wdcorrect can fasten the process though.

src/mlpack/methods/decision_tree/decision_tree.hpp
```diff
@@ -371,6 +409,12 @@ class DecisionTree :
    */
   arma::vec classProbabilities;
 
+  // this stores the starting point in the dataset that belongs to this node.
+  size_t beginIndex;
+
+  // this stores the count of the data points stored in this node.
+  size_t countInNode;
```
rcurtin: It's really important that we keep the `DecisionTree` object as small as possible, so since these are only used for training, I'd suggest that we find a way to move them into some auxiliary training structure that isn't stored in the tree.

src/mlpack/methods/decision_tree/decision_tree.hpp
```diff
@@ -354,6 +359,39 @@ class DecisionTree :
...
+   */
+  template<bool UseWeights, typename MatType, typename LabelsType,
+      typename WeightsType>
+  void Prune(DecisionTree* root,
+             LabelsType&& labels,
+             const size_t numClasses,
+             WeightsType&& weights,
+             MatType&& validData,
+             LabelsType&& validLabels,
+             double& bestScore);
```
rcurtin: I'd also suggest that we find a way to templatize out the strategy for pruning, instead of including code in the DecisionTree itself that specifies a single pruning strategy.

Manthan-R-Sheth: Ya i agree with templatizing pruning() (and/or validascore() as well to include different validating strategies) . Infact that was my next step once the code was reviewed to be fine.

Manthan-R-Sheth:
@rcurtin 
I was implementing reduced error post tree pruning and so including it in the training step would change the meaning. Your suggestion is another kind of pruning method and can be implemented too. The split will only be made if the error on validation set due to the splitting is more than the present state.
Shall i implement that also?

src/mlpack/methods/decision_tree/decision_tree.hpp
sourabhvarshney111: We wanted to first check whether the implementation is correct. If so, we would make template for the tree. So, can you please review the implementation and the tests, so that we can move to the next step.

src/mlpack/methods/decision_tree/decision_tree.hpp
Manthan-R-Sheth: @rcurtin 
By addition of two variables (countinNode and beginIndex) the problem of finding the majority class in the node is solved. I understand that it will increase the size of the class and hence the overheads for building the tree and so I will think of storing them so that the overheads aren't increased.

rcurtin:
Let's handle the design issues before the implementation issues.  I say this because after the code is implemented, it is generally easy to make changes to the code so it functions correctly.  But it is much more time-consuming to redesign and refactor the code.  Therefore, let's first get the design to a good state, and then we can worry about whether or not it's exactly correct later.  (And really, the tests that are implemented should be showing you that anyway.)

If your pruning algorithm needs full access to the tree after building, then I think that perhaps the idea I proposed for how to templatize may not work.  So you may need to rethink that.  But at the end of the day any way we do it should roughly follow the following guidelines:

 * We cannot add any members to `DecisionTree`.
 * Any overhead of pruning should be minimal---tree construction time is very important.
 * The pruning strategy should be written as a separate class that plugs in as a template parameter.
 * Any pruning strategy that a user might implement should only need to implement a few well-documented methods.

Manthan-R-Sheth:
@rcurtin 
I understand the point. I will add it as a template parameter and figure out a way to do it. My only concern is that I think the two variables beginIndex and countInNode will be required to be added in DeisionTree class to get the maximum probability of any non-leaf node. (I will give it a thought again, however.)

sourabhvarshney111:
I think we should go with the @rcurtin idea because I think if we are providing type of pruning in template, then it does not matter whether it is pre pruning or post pruning. With pre-pruning, I think we might avoid beginIndex and CountInNode variables. What say @Manthan-R-Sheth?

Manthan-R-Sheth:
@sourabhvarshney111 
Adding pruning as a template parameter has been a part of the plan always. Also pre-pruning would be implemented with a little overhead but without any need of additional class members. What is the point of concern is the post-pruning implementation. I think we would need to add data members to the DecisionTree class (2 integers) for it to work even when it is a template parameter.

I will have to think of something to avoid the use of extra members and implement post-pruning.

rcurtin:
I have not thought through the design so I cannot tell you exactly how to do it.  What I do know is that the pruning information is not necessary once the tree is built.  So it is not a good idea to store it in the members of `DecisionTree`.  I am unsure of the exact solution that can be used for it, but a prerequisite for consideration of merging, the PR would have to satisfy the conditions I mentioned above.

Manthan-R-Sheth:
This should calculate the majority class of each non-leaf node without any need of additional data members. ```calculateProbabilities``` is modified to contain number of data points instead of probabilities. I will push the corresponding changes in sometime for tests to pass.

sourabhvarshney111:
Sorry @Manthan-R-Sheth I saw this too late. But are you checking each node to be leaf node or not in the pruning method itself?

Manthan-R-Sheth:
@sourabhvarshney111 
Pruning only happens if all the children of a particular node are leaves. In case they are not, current  node is not pruned and its children are looped over in DFS fashion. At each loop of the children of the current node, it is checked if the siblings are all leaves or not. If they are, then current node is pruned. 
Thus current node can be pruned even if all children are not leaves (which can happen if the non-leaf children are pruned from bottom-up).
I hope it clarifies the approach.

Manthan-R-Sheth:
@rcurtin @zoq 
I have updated the PR according to your suggestions and requirements. Can you have a look at the code now.

src/mlpack/methods/decision_tree/decision_tree_impl.hpp
```diff
@@ -1043,12 +1091,40 @@ void DecisionTree<FitnessFunction,
   }
 
   // Now normalize into probabilities.
-  classProbabilities /= UseWeights ? sumWeights : labels.n_elem;
+  // classProbabilities /= UseWeights ? sumWeights : labels.n_elem;
```
rcurtin: I'd really like if we could avoid doing this.  This means that when we do prediction, we'll have to do a division.  And I feel pretty strongly that we need to keep prediction as fast as possible.

Here is a different approach---when building the tree, allow the `PruningMethod` to store some state about each node.  Maybe a `std::map<DecisionTree*, SomeValuesStruct>` or something like this, where you can store the class counts for later pruning.  Then when training is done, the extra data held by the `PruningMethod` can be deleted, and pruning methods can collect any data they need during training.  I would say, then, that we could add a `CreateNode(...)` function to each `PruningMethod`, which could then update any internal state and variables that need to be known for training.

I know this involves a lot of refactoring, and I am sorry about that, but again, speed is really important here.

rcurtin:
I've closed this for inactivity---I think it is a long way from being merged and I'd like to make the list of open PRs smaller. :)  If you find time to work on this again, I'm happy to try and help out.

PR #1310, opened 2018-03-13
Fix bug in mean shift when max_iterations is too low w.r.t size of data

yamidark:
In mean shift, when max_iterations specified is too low w.r.t size of data (e.g. 2 max iterations with `iris_test.csv` and 9 max iterations with `test_data_3_1000.csv`), the program terminates after throwing an instance of `std::invalid_argument`.
>what(): Requested value of k (1) is greater than the number of points in reference set (0).
Aborted (core dumped)

This is because we may not have any centroids converged after max iterations of the mean shift algorithm and then run the `neighbor::KNN.search` with an empty matrix.

This PR fixes this by adding 1 centroid previously calculated if the `centroids` matrix is empty.

### DIFF
--- a/src/mlpack/methods/mean_shift/mean_shift.hpp
+++ b/src/mlpack/methods/mean_shift/mean_shift.hpp
@@ -84,6 +84,7 @@ class MeanShift
   void Cluster(const MatType& data,
                arma::Row<size_t>& assignments,
                arma::mat& centroids,
+               bool forceConvergence = true,
                bool useSeeds = true);
 
   //! Get the maximum number of iterations.
--- a/src/mlpack/methods/mean_shift/mean_shift.hpp
+++ b/src/mlpack/methods/mean_shift/mean_shift.hpp
@@ -33,8 +33,8 @@ namespace meanshift /** Mean shift clustering. */ {
  * extern arma::mat data; // Dataset we want to run mean shift on.
  * arma::Row<size_t> assignments; // Cluster assignments.
  * arma::mat centroids; // Cluster centroids.
- * bool forceConvergence; // Flag whether to force each centroid seed to
- * converge regardless of maxIterations.
+ * bool forceConvergence = true; // Flag whether to force each centroid seed
+ * to converge regardless of maxIterations.
  *
  * MeanShift<> meanShift();
  * meanShift.Cluster(dataset, assignments, centroids, forceConvergence);
--- a/src/mlpack/methods/mean_shift/mean_shift.hpp
+++ b/src/mlpack/methods/mean_shift/mean_shift.hpp
@@ -33,9 +33,11 @@ namespace meanshift /** Mean shift clustering. */ {
  * extern arma::mat data; // Dataset we want to run mean shift on.
  * arma::Row<size_t> assignments; // Cluster assignments.
  * arma::mat centroids; // Cluster centroids.
+ * bool forceConvergence; // Flag whether to force each centroid seed to
+ * converge regardless of maxIterations.
  *
  * MeanShift<> meanShift();
- * meanShift.Cluster(dataset, assignments, centroids);
+ * meanShift.Cluster(dataset, assignments, centroids, forceConvergence);
  * @endcode
  *
  * @tparam UseKernel Use kernel or mean to calculate new centroid.
@@ -80,6 +82,8 @@ class MeanShift
    * @param data Dataset to cluster.
    * @param assignments Vector to store cluster assignments in.
    * @param centroids Matrix in which centroids are stored.
+   * @param forceConvergence Flag whether to force each centroid seed to
+   * converge regardless of maxIterations.
    */
   void Cluster(const MatType& data,
                arma::Row<size_t>& assignments,
--- a/src/mlpack/methods/mean_shift/mean_shift_impl.hpp
+++ b/src/mlpack/methods/mean_shift/mean_shift_impl.hpp
@@ -278,6 +278,10 @@ inline void MeanShift<UseKernel, KernelType, MatType>::Cluster(
     }
     assignments.zeros();
   }
+  else if (centroids.n_cols == 1)
+  {
+    assignments.zeros();
+  }
   else
   {
     // Assign centroids to each point.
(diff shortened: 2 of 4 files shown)


rcurtin:
Thanks for opening this PR.  I see the problem that you've identified and agree that it is a problem we should fix.  I have a few comments---

 * Do you think it is better to also add some boolean option to the `MeanShift` class that will automatically increase the number of iterations so that at least one cluster converges?  This could maybe be called `forceConvergence`, and when it was set to true, termination would occur either at `maxIterations` if a cluster had converged, or after `maxIterations` when the first cluster converges.

 * If we have to extract a single cluster manually, there's no need for the `NeighborSearch` step---all the assignments will be for the single cluster that was chosen.  So there is a slight optimization we could make.

 * Is it better to take a random centroid, or should be pick the centroid that was the closest to convergence?  I am not sure what is the better option here.

yamidark:
@rcurtin 
> Do you think it is better to also add some boolean option to the MeanShift class that will automatically increase the number of iterations so that at least one cluster converges?

Yep, I think this is a good idea, especially if we set this `forceConvergence` boolean to be `true` by default. Will add this in. 

>If we have to extract a single cluster manually, there's no need for the NeighborSearch step

I see, will add this optimization in.

>Is it better to take a random centroid, or should be pick the centroid that was the closest to convergence?

I'm also not sure which would be better.
Maybe we can give a warning message if we reach this stage that none of the clusters converges so we give a random 1, and to either give a larger `max_iterations` or set `forceConvergence` to `true`?

rcurtin:
> Maybe we can give a warning message if we reach this stage that none of the clusters converges so we give a random 1, and to either give a larger max_iterations or set forceConvergence to true?

That sounds perfectly reasonable to me; let's go with that.

Thanks!

yamidark:
@rcurtin I have made the changes suggested.

src/mlpack/methods/mean_shift/mean_shift.hpp
```diff
@@ -84,6 +84,7 @@ class MeanShift
   void Cluster(const MatType& data,
                arma::Row<size_t>& assignments,
                arma::mat& centroids,
+               bool forceConvergence = true,
```
rcurtin: Could you add some documentation to the comments above the method describing what `forceConvergence` is please?

src/mlpack/methods/mean_shift/mean_shift_impl.hpp
```diff
@@ -216,8 +217,8 @@ inline void MeanShift<UseKernel, KernelType, MatType>::Cluster(
   {
     // Initial centroid is the seed itself.
     allCentroids.col(i) = pSeeds->unsafe_col(i);
-    for (size_t completedIterations = 0; completedIterations < maxIterations;
-         completedIterations++)
+    for (size_t completedIterations = 0; completedIterations < maxIterations
+      || forceConvergence; completedIterations++)
```
rcurtin: Ah, I guess maybe I should have clarified a bit.  I thought that a `forceConvergence` parameter would run the mean shift algorithm for at least the minimum number of iterations necessary to produce one cluster.  So we could also add some part to the end of the inner loop like

```
if (forceConvergence && centroids.n_cols >= 1)
  break;
```

What do you think?

src/mlpack/methods/mean_shift/mean_shift_impl.hpp
```diff
@@ -260,12 +261,32 @@ inline void MeanShift<UseKernel, KernelType, MatType>::Cluster(
...
 
-  // Assign centroids to each point.
-  neighbor::KNN neighborSearcher(centroids);
-  arma::mat neighborDistances;
-  arma::Mat<size_t> resultingNeighbors;
-  neighborSearcher.Search(data, 1, resultingNeighbors, neighborDistances);
-  assignments = resultingNeighbors;
+  // If no centroid has converged due to too little iterations and without
+  // forcing convergence, take 1 random centroid calculated.
+  if (centroids.empty())
```
rcurtin: We can also optimize for the case of `centroids.n_cols == 1` with another `else if` and call `assignments.zeros()` there to avoid the `KNN` search.

src/mlpack/methods/mean_shift/mean_shift_impl.hpp
yamidark: So if `forceConvergence` is passed, we should only force the 1st seed to converge but not the others? 
From what I understand from the code, the mean shift algorithm runs for each seed centroid generated earlier up to the given `maxIterations` or it converges, so if `forceConvergence` is given I think we should be forcing all these seed centroids to converge.

src/mlpack/methods/mean_shift/mean_shift.hpp
yamidark: Noted, will add it in.

src/mlpack/methods/mean_shift/mean_shift_impl.hpp
yamidark: Noted, will add this in.

src/mlpack/methods/mean_shift/mean_shift_impl.hpp
rcurtin: Ok, I can agree with that.  Let's leave it the way it is now then.

yamidark:
@rcurtin I have made the changes suggested again.

src/mlpack/methods/mean_shift/mean_shift.hpp
```diff
@@ -33,9 +33,11 @@ namespace meanshift /** Mean shift clustering. */ {
  * extern arma::mat data; // Dataset we want to run mean shift on.
  * arma::Row<size_t> assignments; // Cluster assignments.
  * arma::mat centroids; // Cluster centroids.
+ * bool forceConvergence; // Flag whether to force each centroid seed to
+ * converge regardless of maxIterations.
```
rcurtin: Just for clarity of documentation I'd suggest setting `forceConvergence` to `true` or `false` in the example here (doesn't matter which).

src/mlpack/methods/mean_shift/mean_shift_main.cpp
```diff
@@ -89,7 +92,8 @@ static void mlpackMain()
 
   Timer::Start("clustering");
   Log::Info << "Performing mean shift clustering..." << endl;
-  meanShift.Cluster(dataset, assignments, centroids);
+  meanShift.Cluster(dataset, assignments, centroids,
+   CLI::HasParam("force_convergence"));
```
rcurtin: Looks like the spacing is off here---a continued line should be indented four spaces from the previous line.

PR #1313, opened 2018-03-14
Miscellaneous test fixes

rcurtin:
I've fixed a number of tests that I found to be occasionally failing:

 * `NMFRandomDivTest`: allow multiple trials of the test to prevent the effects of bad random seeds.

 * Serialization tests: with Boost 1.49 and LLVM 5.0.1 (and possibly other compilers), serialization tests can sometimes fail.  After digging to the bottom of the issue I concluded that the best solution was a workaround, so now trees avoid serializing any null pointers.

 * `HMMTrainMainTest/HMMTrainReuseGaussianModelTest`: it looks like sometimes the retraining actually moves the HMM model too much, so I set the tolerance for the retraining to a very large value.

 * `XTreeTraverserTest`: in a very uncommon edge case, the number of descendants and the bound isn't properly reset.


zoq:
I think you are going to fix the 2 issues pointed out by the Static Code Analysis Check.

PR #1314, opened 2018-03-15
ApproxKFN Binding tests

Namrata96:
#1152  These are a few tests I have written for the ApproxKFN method so far as my first contribution. 
Suggest me whatever tests I should add and I will add them as soon as possible. 

These are some of the tests I plan to add:
- [x] Check that we can't pass both input_model and reference matrix.

- [x] Check that a saved model can be used again.


Namrata96:
Thanks @rcurtin for the review. I will add the tests suggested by you.

src/mlpack/tests/main_tests/approx_kfn_test.cpp
```diff
@@ -0,0 +1,116 @@
...
+{
+  arma::mat referenceData;
+  referenceData.randu(2, 80); // 80 points in 2 dimensions.
+
+  // Random input, any k <= reference points.
+  SetInputParam("reference", std::move(referenceData));
+  SetInputParam("k", (int) 10);
+
+  // Input pre-trained model.
+  SetInputParam("input_model", std::move(CLI::GetParam<ApproxKFNModel*>("output_model")));
```
akhandait: Lines should be less than 80 characters long. Do keep this in my mind for the tests that you must be writing now.

src/mlpack/tests/main_tests/approx_kfn_test.cpp
```diff
@@ -0,0 +1,116 @@
...
+}
+
+/**
+ * Check that we can't specify an invalid k.
+ */
+BOOST_AUTO_TEST_CASE(ApproxKFNInvalidKTest)
+{
+  arma::mat referenceData;
+  referenceData.randu(2, 80); // 80 points in 2 dimensions.
+
```
akhandait: A line shouldn't end in whitespace, so in this case, it means there must be no spaces in a line with no code.
Same for line 107.

src/mlpack/tests/main_tests/approx_kfn_test.cpp
```diff
@@ -0,0 +1,116 @@
...
+ * Check that we can't specify an invalid k.
+ */
+BOOST_AUTO_TEST_CASE(ApproxKFNInvalidKTest)
+{
+  arma::mat referenceData;
+  referenceData.randu(2, 80); // 80 points in 2 dimensions.
+  
+  SetInputParam("reference", std::move(referenceData));
+  // Random input, k > reference points.
+  SetInputParam("k", (int) 81); //Invalid
```
akhandait: There should be a space between // and the comment. Same for line 109.

src/mlpack/tests/main_tests/approx_kfn_test.cpp
```diff
@@ -0,0 +1,116 @@
...
+  
+  SetInputParam("reference", std::move(referenceData));
+  SetInputParam("algorithm", (string) "any_algo"); //Invalid
+
+  Log::Fatal.ignoreInput = true;
+  BOOST_REQUIRE_THROW(mlpackMain(), std::runtime_error);
+  Log::Fatal.ignoreInput = false;
+}
+
+BOOST_AUTO_TEST_SUITE_END();
```
akhandait: The file should end with a newline.

Namrata96:
thanks @akhandait for reviewing it. I'm going through the style guide, and will incorporate your changes. :)

Namrata96:
@rcurtin while implementing the test "Make sure that num_tables has an effect on the output by running twice with different num_tables parameters and ensuring the results are different." , I'm trying to ensure that the distances matrices returned are not the equal - in PR #1263, Atharva has implemented a CheckMatricesNotEqual() function but it hasn't been merged yet. Should I wait for the merge or submit my own in this PR?

akhandait:
That PR has a task which isn't done yet. Also, right now I am focusing on the gsoc proposal, I think it will take some time to get that merged. Maybe you can just copy those functions I defined and use it here, that way the PR that merges second can just merge without any conflicts.

src/mlpack/tests/main_tests/approx_kfn_test.cpp
```diff
@@ -0,0 +1,251 @@
...
+  arma::mat referenceData;
+  referenceData.randu(2, 80); // 80 points in 2 dimensions.
+
+  // Random input, any k <= reference points.
+  SetInputParam("reference", std::move(referenceData));
+  SetInputParam("k", (int) 10);
+
+  // Input pre-trained model.
+  SetInputParam("input_model", std::move(CLI::GetParam<ApproxKFNModel*>
+    ("output_model")));
```
rcurtin: I'm a little bit concerned about this---"output_model" isn't going to be set to anything useful, so I think it may be safer to create an `ApproxKFNModel` in the test then simply pass a pointer to it to `SetInputParam()`.

Also, I think this test means you can check off the first todo in the PR description.

src/mlpack/tests/main_tests/approx_kfn_test.cpp
```diff
@@ -0,0 +1,251 @@
...
+  SetInputParam("reference", std::move(referenceData));
+  // Random input, any k <= reference points.
+  SetInputParam("k", (int) 10);
+  SetInputParam("calculate_error", (bool) true);
+
+  // Random matrix specifying exact distances of each point to its k neighbors.
+  // Note that the values in the matrix do not matter as we are only concernec
+  // with the dimensions of the matrix passed.
+  arma::mat exactDistances;
+  exactDistances.randu(9, 90); // Wrong size (should be (10, 80)).
```
rcurtin: I think the size here should actually be `(1, 80)`, but in any case, the test still works, I am only saying that because I think the comment is incorrect.

Namrata96:
- Make sure different results are returned when the algorithm is ds or qdafn.
I may have missed something, so please let me know but are both strategies supposed to give different approx k furthest neighbors?

Namrata96:
@rcurtin Thanks for the review. I'll add the remaining tests, as well as incorporate the suggestions suggested by you.

src/mlpack/tests/main_tests/approx_kfn_test.cpp
Namrata96: Oh, I might have understood exact distances incorrectly. Doesn't the matrix contain values for each of the k furthest neighbors for each point in the reference set?

rcurtin: Hmm, I guess the program actually doesn't really care how many rows are in the `exact_distances` matrix as long as it is more than one, so I suppose it doesn't matter.  Still, I think `(1, 80)` might be a better comment in case anyone reads this test trying to figure out how to use the binding, since only that first row is needed.

rcurtin:
> I may have missed something, so please let me know but are both strategies supposed to give different approx k furthest neighbors?

Yeah, the strategies are very different so they should give different results.  In order to maximize the probability of different results, I'd suggest using a high-dimensional random dataset generated with `randu()`.

When you write the test, it's probably good to make sure that it doesn't fail sometimes.  Testing this is a little more tedious than it used to be.  First modify `RandomSeed()` in `mlpack/core/math/random.hpp` and comment out the `#if (BINDING_TYPE != BINDING_TYPE_TEST)`, `#else`, and `#endif` lines.  Then, add `mlpack::math::RandomSeed(std::time(NULL));` to the top of the test you wrote.  This means that every time you run the test, it will run with a different random seed.  Then, you can run the test many times and make sure it doesn't fail on any of the runs.  I like to use a loop like this:

```
i=0; while(true); do bin/mlpack_test -t TestSuiteName/TestCaseName 2>&1 | grep -v '^Running' | grep -v 'No errors' | grep -v '^[ ]*$'; i=$((i + 1)); echo $i; done
```

sourabhvarshney111:
@Namrata96 Are you still working on the PR?

Namrata96:
@sourabhvarshney111 Yes, I am. Sorry for the delay, I was travelling.

sourabhvarshney111:
No issues @Namrata96. Carry on with your work. I was just asking because it seemed inactive from a long time. : )

Namrata96:
Sorry for the delay in completing this PR. I have written the remaining tests, and made sure they do not fail when run multiple times with different seeds.

sourabhvarshney111:
@Namrata96 Looks like you have completed the PR before review and some tests are failing. Please take a look at these failing test logs, you will find it easy to make them pass straightaway. Let us know if you need any help.

Namrata96:
@sourabhvarshney111  sorry, I didn't mean that. My intention was to not specifically ask for a particular reviewer's time. 
Regarding the failing checks, I'll look into them right away.

sourabhvarshney111:
@Namrata96 Sorry I just misunderstood. I was just looking forward to see this merged. So, I offered help. Never mind please : (

Namrata96:
@sourabhvarshney111 no worries, thanks for offering help. :)

src/mlpack/bindings/python/mlpack/matrix_utils.py
```diff
@@ -58,7 +58,7 @@ def to_matrix(x, dtype=np.double, copy=False):
 def to_matrix_with_info(x, dtype, copy=False):
   """
   Given some array-like X (which should be either a numpy ndarray or a pandas
-  DataFrame, convert into a numpy matrix of the given dtype.
+  DataFrame), convert it into a numpy matrix of the given dtype.
```
rcurtin: Thanks, always nice to have documentation fixes like this. :)

src/mlpack/methods/approx_kfn/approx_kfn_main.cpp
```diff
@@ -263,6 +263,15 @@ static void mlpackMain()
       arma::mat exactDistances;
       if (CLI::HasParam("exact_distances"))
       {
+        // Check the exact distances matrix has the right dimensions.
+        RequireParamValue<arma::mat>("exact_distances",
+          [&](arma::mat exactDists){ return (exactDists.n_rows == k); }, true,
+          "the number of rows in exact distances matrix must be equal to k.");
```
rcurtin: I think it would be better to pass a reference to the lambda here to avoid copying the matrix:

```
[&](const arma::mat& exactDists) { ... }
```

src/mlpack/tests/main_tests/approx_kfn_test.cpp
rcurtin: Just to clarify on this one, can you change to something like:

```
ApproxKFNModel m;
SetInputParam("input_model", &m);
```

src/mlpack/tests/main_tests/approx_kfn_test.cpp
```diff
@@ -0,0 +1,396 @@
...
+  SetInputParam("k", (int) 81); // Invalid.
+
+  Log::Fatal.ignoreInput = true;
+  BOOST_REQUIRE_THROW(mlpackMain(), std::invalid_argument);
+  Log::Fatal.ignoreInput = false;
+}
+
+/**
+ * Make sure that the dimensions of neighbors and distances is correct given a
+   value of k.
```
rcurtin: I know this is really minor, but there is a missing `*` here, could you fix it please? :)

src/mlpack/tests/main_tests/approx_kfn_test.cpp
```diff
@@ -0,0 +1,396 @@
...
+  SetInputParam("reference", std::move(referenceData));
+  // Random input, any k <= reference points.
+  SetInputParam("k", (int) 10);
+  SetInputParam("algorithm", (string) "ds");
+
+  // First solution.
+  mlpackMain();
+
+  // Get the distances and neighbors matrix after first training.
+  arma::mat FirstOutputDistances =
```
rcurtin: As per the style guide, variables shouldn't have capitalized names, so could you please change this to `firstOutputDistances`?  (This applies a few other places.)

src/mlpack/tests/test_tools.hpp
```diff
@@ -66,6 +66,97 @@ inline void CheckMatrices(const arma::cube& a,
   }
 }
 
+// Check if two matrices are different.
+inline void CheckMatricesNotEqual(const arma::mat& a,
```
rcurtin: Why not templatize?

```
template<typename MatType>
inline void CheckMatricesNotEqual(const MatType& a, const MatType& b, const double tolerance);
```

PR #1315, opened 2018-03-15
RangeSearch: update methods and constructors to pass-by-value

yamidark:
Updated the methods and constructors in `range_search` to pass-by-value instead of the 2 overload methods as suggested in #1021.

### DIFF
--- a/src/mlpack/methods/range_search/range_search_impl.hpp
+++ b/src/mlpack/methods/range_search/range_search_impl.hpp
@@ -217,42 +217,7 @@ template<typename MetricType,
                   typename TreeStatType,
                   typename TreeMatType> class TreeType>
 void RangeSearch<MetricType, MatType, TreeType>::Train(
-    const MatType& referenceSet)
-{
-  // Clean up the old tree, if we built one.
-  if (treeOwner && referenceTree)
-    delete referenceTree;
-
-  // Rebuild the tree, if necessary.
-  if (!naive)
-  {
-    referenceTree = BuildTree<Tree>(const_cast<MatType&>(referenceSet),
-        oldFromNewReferences);
-    treeOwner = true;
-  }
-  else
-  {
-    treeOwner = false;
-  }
-
-  // Delete the old reference set, if we owned it.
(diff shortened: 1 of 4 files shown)


yamidark:
@rcurtin 
- There still seems to be places where `setOwner` is set to `false`, so I don't think it should be removed.
- I have tried removing and modifying 1 of the overloads of `BuildTree<>`, but it seems like both of them are still needed in `RangeSearch::Search`.

yamidark:
Sorry, I'm not too sure about what you just mentioned. 
Your earlier comment asked if:
>Is RangeSearch<>::setOwner still necessary? I think now it is always true that setOwner will always be either true or the same value as naive, which means it could be removed.

but in the code you mentioned a place where `setOwner` was previously set to `true`, not a place where `setOwner` was previously `false`.
 So there are still places like in line 82(of this PR) where `setOwner` is set to `false` and line 132(of this PR) where `setOwner` is set to `!other.referenceTree`, and I'm not sure if they should be changed to `naive` or something else as well.

rcurtin:
I hope this clarifies, let me know if not---

 * In line 82, `setOwner` is set to false and so is `naive`, because we are passing a tree to the constructor (which implies we are not in naive search mode).
 * In line 132, `!other.referenceTree` is actually equivalent to `other.naive`---if `other` is in naive search mode, `other` will have no reference tree built.

So I think there are no problems there.  Let me know if anything is still unclear; I hope this helps.

src/mlpack/methods/range_search/range_search_impl.hpp
```diff
@@ -265,13 +256,12 @@ void RangeSearch<MetricType, MatType, TreeType>::Train(
 
   if (treeOwner && referenceTree)
     delete this->referenceTree;
-  if (setOwner && referenceSet)
+  if (naive && referenceSet)
```
rcurtin: Here `naive` can't be true since the user is passing a prebuilt tree, so I think we can remove it as a condition.

rcurtin:
Thank you for the nice contribution! :)

PR #1317, opened 2018-03-17
Particle Swarm Optimization.

chintans111:
Added a basic PSO optimizer for the lbest variant. Also added some code structure to allow the use of gradient descent along with PSO forming a hybrid approach, but that is not implemented yet. It will follow in future commits.

Edit: Proper documentation is still pending within the code, it will follow in a future commit.


chintans111:
@zoq @rcurtin Should I close this PR? Not really looking to get this merged into the code at the moment, just wanted to get an idea of how everything works. Might pick up another small issue in the meanwhile and continue reading more about PSO.

zoq:
Your call, we might be able to combine with https://github.com/mlpack/mlpack/pull/1225. Let me know what you think.

chintans111:
That is what I initially had in mind, but if needed I can just push the changes again. Might just make it easier to merge without any conflicts as I can just pull #1225 and make any additional tweaks needed there.

zoq:
Let's keep this open until #1225 is merged afterwards we can incorporate the changes. Does this sounds good to you?

chintans111:
Yes that sounds perfect. I was also thinking of adding some kind of range for particle initialization, as right now it only spreads particles in the [0, 1] hypercube. Might be useful for initialization I think. Also, the Rastrigin test seems to be consistently failing on the Travis build for fewer iterations (~2000), something I have been trying to reproduce on my system. I have set the iterations ridiculously high for that right now, but I'm also planning to check if increasing the ```exploitationFactor``` helps. Any suggestions for that?

The lbest approach takes more iterations to converge than the gbest approach, but Rastrigin seems to be taking quite a lot more.

zoq:
If you run the tests make sure to use a different random seed for each run, see https://github.com/mlpack/mlpack/issues/922 for more information. About the initialization part, I think providing more options in that direction would be great to have.

chintans111:
#922 proved to be helpful, thank you. About the initialization part, was thinking of adding a lower limit and an upper limit for the initialization range ```[lower, upper]``` and providing the respective defaults arguments to make it ```[0, 1]```. Does this sound good?

zoq:
Sounds like a good start to me, ideally we can pass a class for the initialization step, that way a user can use his own strategy, what do you think?

chintans111:
I do not fully get this, but are we talking about having a user defined class which has methods for, say, initializing positions like: ```arma::mat particlePositions = initializerClass.GetParticlePositions();```? And the user will then pass this class instead as an argument as opposed to the various parameters?

I did have it on my mind to separate out the phases of the algorithm (initialization, iterations, maybe gradient descent), this might make it even better.

Please correct me if I'm wrong.

chintans111:
@zoq Just added some initialization options for starters, it brought down the number of iterations required while I ran some tests. Can you please confirm if the approach mentioned in the previous comment is correct?

zoq:
> I do not fully get this, but are we talking about having a user defined class which has methods for, say, initializing positions like: arma::mat particlePositions = initializerClass.GetParticlePositions();? And the user will then pass this class instead as an argument as opposed to the various parameters?

Agreed, that might be a good idea to do, makes the optimizer more extendable.

zoq:
We already used the policy pattern for the various optimization methods e.g. for the update step, take a look at the template parameter (``UpdatePolicyType``): https://github.com/mlpack/mlpack/blob/master/src/mlpack/core/optimizers/sgd/sgd.hpp.

chintans111:
@zoq okay I misunderstood it before. We have already used a policy based velocity update method, shouldn't be too difficult to add another policy for initialization as well. Thanks for the clarification.

chintans111:
@zoq sorry for the delay, should have informed you that practical exams are going on. Just completed the Initialization Policy as discussed earlier, please let me know if any changes are required. Currently waiting for #1372 to be merged as that is causing the style checks to fail.

Also looking towards completing the documentation next (mainly the description of the PSO algorithm and the mathematics behind it are incomplete) and tidying up a few minor things (variable names in the ```LBestUpdate``` policy for instance, along with its getters and setters), and then moving on to either another variant of PSO, or testing out the hybrid approach using gradient descent.

In case we go with the new variant implementation, which variant should I pick up? In my proposal I have tentatively mentioned that the 2 basic variants will be gbest and lbest PSO using constriction factor, but gbest has already been implemented in #1225, so it would be redundant. I am also browsing around for some interesting ones, but if you have any specific variant in mind, please let me know.

Some surrounding code structure is ready for the hybrid approach, but haven't actually implemented GD yet. Will be referring to the existing implementation under optimizers for that purpose. I had a doubt regarding the testing of the proposed method - how exactly do we quantify the achieved improvement (if any) over the PSO-only version? The Optimize function do not return the number of iterations taken in any way, so it might be difficult to write test cases around that. What I currently have in mind is to compare the quality of the solution (minimum value obtained) after running the optimizer for a smaller number of iterations than required for convergence and comparing the results of both, the PSO-only and the hybrid methods. Do you have any suggestions or improvements in mind?

I am also considering moving the ```exploitationFactor``` and ```explorationFactor``` arguments to the ```UpdatePolicy``` and completely removing them from the ```PSOType``` class. This is to eliminate redundancy, but the main reason being that the values of these parameters depend on the ```UpdatePolicy``` being used and thus makes it difficult to provide meaningful defaults across all policies in the ```PSOType``` constructor. The downside is that the user will have to write a few more lines to initialize the parameters for the required policy. Please let me know your thoughts about this.

PR #1319, opened 2018-03-17
Improvements and Speedups to Convolution Layer and Associated Rules

ShikharJ:
@zoq I couldn't find much ways to improve upon the current implementation, as far as the code is concerned. Did you have any other idea in mind regarding this?

### DIFF
--- a/src/mlpack/methods/ann/convolution_rules/fft_convolution.hpp
+++ b/src/mlpack/methods/ann/convolution_rules/fft_convolution.hpp
@@ -21,8 +21,8 @@ namespace ann /** Artificial Neural Network. */ {
 
 /**
  * Computes the two-dimensional convolution through fft. This class allows
- * specification of the type of the border type. The convolution can be compute
- * with the valid border type of the full border type (default).
+ * specification of the type of the border type. The convolution can be
+ * computed with the valid border type of the full border type (default).
  *
  * FullConvolution: returns the full two-dimensional convolution.
  * ValidConvolution: returns only those parts of the convolution that are
@@ -40,12 +40,12 @@ class FFTConvolution
   /*
    * Perform a convolution through fft (valid mode). This method only supports
    * input which is even on the last dimension. In case of an odd input width, a
-   * user can manually pad the imput or specify the padLastDim parameter which
+   * user can manually pad the input or specify the padLastDim parameter which
    * takes care of the padding. The filter instead can have any size. When using
-   * the valid mode the filters has to be smaller than the input.
+   * the valid mode the filter has to be smaller than the input.
    *
    * @param input Input used to perform the convolution.
-   * @param filter Filter used to perform the conolution.
+   * @param filter Filter used to perform the convolution.
    * @param output Output data that contains the results of the convolution.
    */
   template<typename eT, typename Border = BorderMode>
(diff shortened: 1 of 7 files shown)


src/mlpack/methods/ann/convolution_rules/fft_convolution.hpp
```diff
@@ -64,23 +64,23 @@ class FFTConvolution
...
 
-    output = arma::real(ifft2(arma::fft2(inputPadded) % arma::fft2(
+    arma::Mat<eT> temp = arma::real(ifft2(arma::fft2(inputPadded) % arma::fft2(
         filterPadded)));
 
     // Extract the region of interest. We don't need to handle the padLastDim in
     // a special way we just cut it out from the output matrix.
-    output = output.submat(filter.n_rows - 1, filter.n_cols - 1,
+    output = temp.submat(filter.n_rows - 1, filter.n_cols - 1,
         input.n_rows - 1, input.n_cols - 1);
```
rcurtin: I don't know if the switch to a `temp` matrix makes any difference here.

src/mlpack/methods/ann/convolution_rules/fft_convolution.hpp
```diff
@@ -208,8 +206,7 @@ class FFTConvolution
     for (size_t i = 1; i < input.n_slices; i++)
     {
       FFTConvolution<BorderMode>::Convolution(input.slice(i), filter,
-          convOutput);
-      output.slice(i) = convOutput;
+          output.slice(i));
```
rcurtin: This avoids a copy; it should be a nice improvement.

src/mlpack/methods/ann/convolution_rules/svd_convolution.hpp
```diff
@@ -87,13 +87,13 @@ class SVDConvolution
         NaiveConvolution<BorderMode>::Convolution(subOutput, U.unsafe_col(0),
             output);
 
+        arma::Mat<eT> temp;
```
rcurtin: I don't disagree with this change, but just as a note, the time it takes to construct the object probably pales in comparison to the time `Convolution()` takes, so this isn't likely to make any noticeable difference.  Still, it is not a bad idea to do. :)

src/mlpack/methods/decision_tree/decision_tree_impl.hpp
```diff
@@ -392,7 +392,7 @@ void DecisionTree<FitnessFunction,
                                       LabelsType&& labels,
                                       const size_t numClasses,
                                       const size_t minimumLeafSize,
-                                      const double minimumGainSplit)
+                                      const double /* minimumGainSplit */)
```
rcurtin: Ah, I think this is actually a bug.  @Manthan-R-Sheth: should the `minimumGainSplit` parameter be passed to the `Train<>` call below?

Manthan-R-Sheth: @rcurtin
Yes you are right, minimumGainSplit has to be passed toTrain<> call. I probably missed it. 
I will open a PR for the samw in 3 days as i am travelling now. Is it fine?

rcurtin: Sure, that's fine---enjoy your traveling. :)

Manthan-R-Sheth: Sure thanks.

src/mlpack/methods/ann/convolution_rules/fft_convolution.hpp
ShikharJ: It doesn't, but then it prevents us from changing the `output` matrix's dimensions, and hence we can directly pass matrix slices (preventing copies) like below.

ShikharJ:
@zoq This is ready for a review. Unless you have any additional suggestions?

src/mlpack/methods/decision_tree/decision_tree_impl.hpp
zoq: Sounds like a simple change to me, I guess if @ShikharJ and @Manthan-R-Sheth is fine with it we could incorporate the change in this PR. Let me know what you think.

src/mlpack/methods/ann/convolution_rules/fft_convolution.hpp
ShikharJ: Actually, the size of the passed matrix changes on assignment of the element wise operation above. But since we're directly passing slices, that can't be possible and gives a compilation error, and so the `temp` matrix is used here.

src/mlpack/methods/decision_tree/decision_tree_impl.hpp
Manthan-R-Sheth: Sorry for the late response owing to my travelling.
Thanks for making the change. It looks good to go now.
 :D

ShikharJ:
@zoq This is ready for merging.

zoq:
Thanks again for another great contribution.

PR #1324, opened 2018-03-19
Removed Reduntant Constructor and modified comments

sourabhvarshney111:
Remove unnecessary constructor requiring predictors and predictors which are again required in train method.

### DIFF
--- a/src/mlpack/tests/ann_layer_test.cpp
+++ b/src/mlpack/tests/ann_layer_test.cpp
@@ -240,10 +240,8 @@ BOOST_AUTO_TEST_CASE(GradientAddLayerTest)
       target = arma::mat("1");
 
       model = new FFN<NegativeLogLikelihood<>, NguyenWidrowInitialization>();
-      arma::mat& temp = model->Predictors();
-      arma::mat& temp1 = model->Responses();
-      temp = input;
-      temp1 = target;
+      model->Predictors() = input;
+      model->Responses() = target;
       model->Add<IdentityLayer<> >();
       model->Add<Add<> >(10);
       model->Add<LogSoftMax<> >();
@@ -464,10 +462,8 @@ BOOST_AUTO_TEST_CASE(GradientLinearLayerTest)
       target = arma::mat("1");
 
       model = new FFN<NegativeLogLikelihood<>, NguyenWidrowInitialization>();
-      arma::mat& temp = model->Predictors();
-      arma::mat& temp1 = model->Responses();
-      temp = input;
-      temp1 = target;
+      model->Predictors() = input;
+      model->Responses() = target;
       model->Add<IdentityLayer<> >();
       model->Add<Linear<> >(10, 2);
       model->Add<LogSoftMax<> >();
@@ -550,10 +546,8 @@ BOOST_AUTO_TEST_CASE(GradientLinearNoBiasLayerTest)
       target = arma::mat("1");
 
       model = new FFN<NegativeLogLikelihood<>, NguyenWidrowInitialization>();
-      arma::mat& temp = model->Predictors();
-      arma::mat& temp1 = model->Responses();
-      temp = input;
-      temp1 = target;
+      model->Predictors() = input;
+      model->Responses() = target;
       model->Add<IdentityLayer<> >();
       model->Add<LinearNoBias<> >(10, 2);
       model->Add<LogSoftMax<> >();
@@ -798,10 +792,8 @@ BOOST_AUTO_TEST_CASE(GradientLSTMLayerTest)
       const size_t rho = 5;
 
       model = new RNN<NegativeLogLikelihood<> >(rho);
-      arma::cube& temp = model->Predictors();
-      arma::cube& temp1 = model->Responses();
-      temp = input;
-      temp1 = target;
+      model->Predictors() = input;
+      model->Responses() = target;
       model->Add<IdentityLayer<> >();
       model->Add<Linear<> >(1, 10);
       model->Add<LSTM<> >(10, 3, rho);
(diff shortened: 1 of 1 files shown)


sourabhvarshney111:
I have not corrected tests. The build will error now. I am correcting this currently.

akhandait:
In the second definition of the `Train` function (the one which does not ask for the optimizer), I think we can replace line 114 - 125 with the `ResetData` data function just like in the first definition. Maybe you could throw in those changes too in this PR as it's quite minimal.

sourabhvarshney111:
I have implemented those changes. I would push them next.

sourabhvarshney111:
@zoq I am not getting clearly where the function type is defined. Can you help?

src/mlpack/methods/ann/ffn.hpp
```diff
@@ -179,26 +158,7 @@ class FFN
...
-   * separable objective function.  This just calls the overload of Evaluate()
-   * with deterministic = true.
-   *
-   * @param parameters Matrix model parameters.
-   * @param begin Index of the starting point to use for objective function
-   *        evaluation.
-   * @param batchSize Number of points to be passed at a time to use for
-   *        objective function evaluation.
-   */
-  double Evaluate(const arma::mat& parameters,
```
zoq: This function is need for the optimization process.

src/mlpack/methods/ann/rnn.hpp
```diff
@@ -193,25 +161,7 @@ class RNN
...
-   * function is usually called by the optimizer to train the model.  This just
-   * calls the other overload of Evaluate() with deterministic = true.
-   *
-   * @param parameters Matrix model parameters.
-   * @param begin Index of the starting point to use for objective function
-   *        evaluation.
-   * @param batchSize Number of points to be passed at a time to use for
-   *        objective function evaluation.
-   */
-  double Evaluate(const arma::mat& parameters,
```
zoq: See the comment for the FFN class.

sourabhvarshney111: I think this happened mistakenly.

sourabhvarshney111:
@zoq I tried to debug the code. But, I found there was no integer being divided by zero. Can you help?

Manthan-R-Sheth:
@sourabhvarshney111 
Can you try setting numFunctions to some non-zero value in the constructor and check if the same error persists? 
The failing tests aren't calling optimizer directly(which uses numFunctions) and so I am not sure if we can draw something useful out of it but still it should be worth a try. Your views?

sourabhvarshney111:
I think the issue lies in Check Gradient function. But in that only two divisions are performed and I don't think denominator were zeros in them.

Manthan-R-Sheth:
Ya, the issue seems to be in CheckGradient() but CheckGradient() calls the gradient (Evaluate() and then Gradient() ) pass on the model. It is possible that due to your changes, the backward or forward pass of any layer produced zero in the denominator (i think numFunctions to be a case).  So it is possible the issues lies in here indirectly.

sourabhvarshney111:
@Manthan-R-Sheth I tried to trace where the control is heading but I found that there was no division happening. But, I thought your idea was worth a try. So I am pushing it through a commit.

sourabhvarshney111:
@Manthan-R-Sheth you are correct. We missed the trick there. It was this thing which were throwing the exception. Thanks for the help. I would correct it soon.

Manthan-R-Sheth:
@sourabhvarshney111 
Sure :smile: 
However, it will be interesting to look up why GradientBatchNormLayerTest (and other such tests) didn't fail.

sourabhvarshney111:
@Manthan-R-Sheth I am still correcting the code. But I think something different was related to the model that were declared in them.

sourabhvarshney111:
@zoq @Manthan-R-Sheth The problem now lies that the tests are failing due to large error in test.

sourabhvarshney111:
@zoq The possible reason I could find is that Gradient function is passing wrong output or the inputs to it are themselves wrong. Am I right? Any suggestions.

src/mlpack/methods/ann/ffn_impl.hpp
```diff
@@ -111,16 +92,7 @@ template<typename OptimizerType>
...
-
-  this->predictors = std::move(predictors);
-  this->responses = std::move(responses);
-
-  this->deterministic = true;
-  ResetDeterministic();
-
-  if (!reset)
-    ResetParameters();
+  ResetData(std::move(predictors), std::move(responses));
```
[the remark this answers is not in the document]
sourabhvarshney111: These are just for optimization. The code in this can be done in a single line as suggested by @akhandait.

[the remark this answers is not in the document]
sourabhvarshney111: No problem @rcurtin. Can you look at my comment above and help me with that?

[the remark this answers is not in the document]
sourabhvarshney111: @rcurtin Looks like I found the issue. I figured out that input and target were assigned to responses and predictors incorrectly.

sourabhvarshney111:
@zoq @rcurtin I think this is now ready for merge. Review whenever you are free.

src/mlpack/tests/ann_layer_test.cpp
```diff
@@ -915,7 +932,11 @@ BOOST_AUTO_TEST_CASE(GradientGRULayerTest)
       target = arma::ones(1, 1, 5);
       const size_t rho = 5;
 
-      model = new RNN<NegativeLogLikelihood<> >(input, target, rho);
+      model = new RNN<NegativeLogLikelihood<> >(rho);
+      arma::cube& temp = model->Predictors();
```
zoq: We can use: ``model->Predictors() = input;`` and ``model->Responses() = target;``.

src/mlpack/methods/ann/ffn.hpp
```diff
@@ -53,8 +53,7 @@ class FFN
   using NetworkType = FFN<OutputLayerType, InitializationRuleType>;
 
   /**
-   * Create the FFN object with the given predictors and responses set (this is
-   * the set that is used to train the network).
+   * Create the FFN object.
```
zoq: Can you make a single block here?

sourabhvarshney111: Sorry I did not get what you said?

sourabhvarshney111: Done

src/mlpack/tests/ann_layer_test.cpp
sourabhvarshney111: Done

Issue #1187, opened 2018-01-02
Memory leaks detected by Boost tests


e-freiman:
When running tests on Windows machine (and compiled by Visual C++ compiler in Debug mode), Boost find many memory leaks. They aren't visible on Linux because of [this](http://www.boost.org/doc/libs/1_64_0/libs/test/doc/html/boost_test/utf_reference/rt_param_reference/detect_memory_leaks.html):

> The only platform which supports memory leak detection is Microsoft Visual Studio family of compilers in debug builds.

I guess this could be a hard issue to handle. As a first step I want to try running each test separately (may be through a python script) and post here which tests exactly reveal memory leaks. Does it make sense? Any ideas?

zoq:
Sounds like a good plan; we should reactive the existing memory test jenkins job (valgrind), btw. you can use: ``mlpack_test --list_content`` to get all test suites/cases; perhaps: https://gist.github.com/zoq/a7ab3d96f0872f03a7aa63a2bd38980d is helpful to clear the output.

evgfreyman:
I run tests separately by using [this script](https://gist.github.com/evgfreyman/fc60883150b2d215df52f418e0a439f2) on my machine - Windows 8.1, Visual C++ 2015. The results are following:

ANNLayerTest/SimpleAddMergeLayerTest - has memory leaks
ANNLayerTest/GradientGRULayerTest - failed
CFTest/CFPredictTest - failed
FastMKSTest/SimpleTrainKernelTest - has memory leaks
FastMKSTest/SerializationTest - has memory leaks
FastMKSTest/FastMKSModelLinearTest - has memory leaks
FastMKSTest/FastMKSModelPolynomialTest - has memory leaks
FastMKSTest/FastMKSModelCosineTest - has memory leaks
FastMKSTest/FastMKSModelGaussianTest - has memory leaks
FastMKSTest/FastMKSModelEpanTest - has memory leaks
FastMKSTest/FastMKSModelTriangularTest - has memory leaks
HoeffdingTreeTest/MultipleSerializationTest - has memory leaks
HoeffdingTreeTest/HoeffdingTreeModelSerializationTest - has memory leaks
LSHTest/CopyConstructorAndOperatorTest - has memory leaks
QDAFNTest/QDAFNUniformSet - failed - has memory leaks
RecurrentNetworkTest/SequenceClassificationTest - has memory leaks
RecurrentNetworkTest/GRURecursiveReberGrammarTest - failed
[...]

PR #1325, opened 2018-03-19
Fix AddMerge and Recurrent layer memory issue.

zoq:
Fix memory issues as pointed out in https://github.com/mlpack/mlpack/issues/1187.

### DIFF
--- a/src/mlpack/methods/ann/layer/add_merge.hpp
+++ b/src/mlpack/methods/ann/layer/add_merge.hpp
@@ -42,8 +42,15 @@ template<
 class AddMerge
 {
  public:
-  //! Create the AddMerge object.
-  AddMerge();
+  /**
+   * Create the AddMerge object using the specified parameters.
+   *
+   * @param model Expose the all network modules.
+   */
+  AddMerge(const bool model = false);
+
+  //! Destructor to release allocated memory.
+  ~AddMerge();
 
   /**
    * Ordinary feed forward pass of a neural network, evaluating the function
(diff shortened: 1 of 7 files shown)


src/mlpack/methods/ann/layer/add_merge.hpp
```diff
@@ -107,15 +114,33 @@ class AddMerge
...
+  //! Return the model modules.
+  // std::vector<LayerTypes<CustomLayers...> >& Model()
+  // {
+  //   if (model)
+  //   {
+  //     return network;
+  //   }
+
+  //   return empty;
+  // }
```
rcurtin: I think this commented code could be removed, but I'm not sure.  I guess maybe the layer should have a `Model()` function?

src/mlpack/methods/ann/layer/add_merge.hpp
```diff
@@ -107,15 +114,33 @@ class AddMerge
...
+
   /**
    * Serialize the layer.
    */
   template<typename Archive>
   void serialize(Archive& ar, const unsigned int /* version */);
 
  private:
+  //! Parameter which indicates if the modules should be exposed.
+  bool model;
```
rcurtin: Hm, I guess I am a little confused about `model`---if it is `true`, then we delete the layers in the destructor.  Maybe a better name is `ownsLayers` or something?  I am not sure if I have understood fully so correct me if I'm wrong.

src/mlpack/methods/ann/rnn.hpp
```diff
@@ -237,26 +237,18 @@ class RNN
...
-  template<typename LayerType>
-  void Add(const LayerType& layer) { network.push_back(new LayerType(layer)); }
-
-  /**
-   * Add a new layer to the model.
+  /*
+   * Add a new module to the model.
    *
-   * @param args The parameters to be passed to the layer constructor.
+   * @param layer The Layer to be added to the model.
```
rcurtin: Oops, it looks like the changes make the documentation backwards---I guess this method should have `@param args` and the other should have `@param layer`.

src/mlpack/methods/ann/layer/add_merge.hpp
zoq: Good point, changed to ``ownsLayers``.

PR #1327, opened 2018-03-21
Update backward step.

zoq:
Updated the backward step formula for the Batch Normalization layer, this should pass the Gradient Check test:
- http://masterblaster.mlpack.org/view/docker/job/docker%20mlpack%20nightly%20build/armadillo_version=armadillo-7.100.3,boost_version=boost_1_49_0,buildmode=debug,compiler_version=gcc-7.2.0,label=linux-amd64/

@Manthan-R-Sheth let me know what you think.

### DIFF
--- a/src/mlpack/methods/ann/layer/batch_norm_impl.hpp
+++ b/src/mlpack/methods/ann/layer/batch_norm_impl.hpp
@@ -59,9 +59,13 @@ void BatchNorm<InputDataType, OutputDataType>::Forward(
   // the forward pass when deterministic is set to true.
   if (deterministic)
   {
+    // Mini--batch mean using the stats object.
     mean = stats.mean();
+
+    // Mini--batch variance using the stats object.
     variance = stats.var(1);
 
+    // Normalize the input and scale and shift the output.
     output = input.each_col() - mean;
     output.each_col() %= gamma / arma::sqrt(variance + eps);
     output.each_col() += beta;
@@ -74,12 +78,14 @@ void BatchNorm<InputDataType, OutputDataType>::Forward(
     for (size_t i = 0; i < input.n_cols; i++)
       stats(input.col(i));
 
+    // Normalize the input.
     output = input.each_col() - mean;
     output.each_col() /= arma::sqrt(variance + eps);
 
     // Reused in the backward and gradient step.
     normalized = output;
 
+    // Scale and shift the output.
     output.each_col() %= gamma;
     output.each_col() += beta;
   }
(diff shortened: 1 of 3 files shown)


src/mlpack/methods/ann/layer/batch_norm_impl.hpp
```diff
@@ -55,64 +55,64 @@ template<typename eT>
...
+  const arma::mat inputMean = input.each_col() - mean;
+  const arma::mat stdInv = 1.0 / arma::sqrt(variance + eps);
+  const arma::mat norm = gy.each_col() % gamma;
+  const arma::mat var = arma::sum(norm % inputMean, 1) %
+      arma::pow(stdInv, 3.0) * 0.5;
+
+  g = (norm.each_col() % stdInv) + (inputMean.each_col() %
+      var * 2 / input.n_cols);
+  g.each_col() += arma::sum(norm.each_col() % -stdInv, 1) + var %
+      arma::mean(-2 * inputMean, 1) / input.n_cols;
```
rcurtin: Do you think it would be worthwhile to rewrite this expression to avoid the temporaries `inputMean`, `stdInv`, `norm`, and `var`, similar to how it was written before?

rcurtin: Sorry, I spent a lot more time with this this morning and I think that my comment above is misguided---since some of these variables are used multiple times, it doesn't make sense to avoid the temporaries.

I used the paper to try and understand this code as the equation for ∂l/∂xᵢ on page 4, but I am a little bit confused.  I can see that:

 * `(norm.each_col() % stdInv)` corresponds directly to (∂l/∂x̂) * (1 / sqrt(σ_B^2 + ϵ)).
 * `(inputMean.each_col() % var * 2 / input.n_cols)` corresponds directly to (∂l/∂σ_B^2) * 2 (xᵢ - µ_B) / m.
 * `(arma::sum(norm.each_col() % -stdInv, 1))` corresponds directly to (∂l/∂μ_B)... but I don't see the multiplication by 1/m that goes after that term.
 * I'm not sure what `var % arma::mean(-2 * inputMean, 1) / input.n_cols` refers to.

I'm pretty sure I've just misunderstood here; since it passes all the tests, my gut feeling is that it's right.  It might be helpful to add comments describing the conversion from the paper to the code.

zoq:
@Manthan-R-Sheth What do you like to do here, should we go with this one or do you like to work on a fix for the current implementation?

Manthan-R-Sheth:
I agree with @rcurtin that it should be merged now due to matrix build testing but i didn't quite understand the idea of having `normalized` matrix. We won't be able to independently call Backward() due to this. I understand that the test needs to be commented as a result of this. Correct me if i am wrong here. 
Otherwise I think the implementation is correct.
Sorry for the late reply, I was quite occupied in travelling and GSoC proposal and so could give less time here.

zoq:
I don't think there is a reason to call ``Backward()`` without calling ``Forward()`` first unless I missed something?

We could still test the output of ``Backward()``, but we would have to call ``Forward`` first, the reason I removed the test was that the GradientCheck already checks the Backward pass and I don't think the output we compared against was correct.

Manthan-R-Sheth:
@zoq 
Ya i understand that Backward() wouldnt be used standalone for practical problems. I meant only for testing purposes.
I believed each function is written independent of each other and so can be tested independently too.

I calculated the outputs manually and so yes there is a chance of precision errors as far as i think.

PR #1329, opened 2018-03-23
Acrobat game for Reinforcement learning Environments

luffy1996:
https://gym.openai.com/envs/Acrobot-v1/
Implementation of Acrobat using Runge-Kutta method for computing new states.
@zoq , @rcurtin you might want to have a look


zoq:
Can you test the new environment against an existing implementation?

src/mlpack/methods/reinforcement_learning/environment/acrobat.hpp
```diff
@@ -0,0 +1,319 @@
...
+    double& AngularVelocity2() { return data[3]; }
+
+    //! Encode the state to a column vector.
+    const arma::colvec& Encode() const { return data; }
+
+    //! Dimension of the encoded state.
+    static constexpr size_t dimension = 4;
+
+   private :
+    arma::colvec data;
```
zoq: Would be great if you could add a comment here.

src/mlpack/methods/reinforcement_learning/environment/acrobat.hpp
```diff
@@ -0,0 +1,319 @@
...
+   * @param link_mass1 mass of link 1.
+   * @param link_mass2 mass of link 2.
+   * @param link_com1 position of the center of mass of link 1.
+   * @param link_com2 position of the center of mass of link 2.
+   * @param link_moi moments of inertia for both link.
+   * @param max_vel1 max angular velocity of link1.
+   * @param max_vel2 max angular velocity of link2.
+   */
+  Acrobat(const double gravity = 9.81,
+          const double link_length1 = 1.0,
```
zoq: Use camel casing for all names; see https://github.com/mlpack/mlpack/wiki/DesignGuidelines#naming-conventions.

src/mlpack/methods/reinforcement_learning/environment/acrobat.hpp
```diff
@@ -0,0 +1,319 @@
...
+    * @param state_ Current State
+    * @param Action Action Taken
+    * @param nextState nextState 
+    */
+  double Torque(const Action& Action)
+  {
+    // Add noise to the Torque
+    std::default_random_engine generator;
+    std::uniform_real_distribution<double> distribution(-0.1, +0.1);
+    double torque = double(Action - 1) + distribution(generator);
```
zoq: Do you mind to use https://github.com/mlpack/mlpack/blob/793fe4abe47cf03872e77fd00369d1653b450758/src/mlpack/core/math/random.hpp#L79 here. I think that would make the expression cleaner.

src/mlpack/methods/reinforcement_learning/environment/acrobat.hpp
```diff
@@ -0,0 +1,319 @@
...
+  {
+    double theta1 = state.Theta1();
+    double theta2 = state.Theta2();
+    return bool ((-cos(theta1)-cos(theta1+theta2)) > 1.0);
+  }
+  /**
+   * @param state Current State
+   * @param torque Torque Applied 
+   */
+  arma::colvec dsdt(arma::colvec state,
```
zoq: Please use upper camel casing for method names.

src/mlpack/methods/reinforcement_learning/environment/acrobat.hpp
```diff
@@ -0,0 +1,319 @@
...
+  }
+  /**
+   *
+   * @param state The current State
+   */        
+  bool IsTerminal(const State& state) const
+  {
+    double theta1 = state.Theta1();
+    double theta2 = state.Theta2();
+    return bool ((-cos(theta1)-cos(theta1+theta2)) > 1.0);
```
zoq: Should we directly use ``state.Theta1()`` instead of make a copy first?

zoq: No need to make an explicit cast here.

src/mlpack/methods/reinforcement_learning/environment/acrobat.hpp
```diff
@@ -0,0 +1,319 @@
...
+ * 3-clause BSD license along with mlpack.  If not, see
+ * http://www.opensource.org/licenses/BSD-3-Clause for more information.
+ */
+
+#ifndef MLPACK_METHODS_RL_ENVIRONMENT_ACROBAT_HPP
+#define MLPACK_METHODS_RL_ENVIRONMENT_ACROBAT_HPP
+#define PIE 3.14159265358979323846
+
+#include <math.h>
+#include <random>
```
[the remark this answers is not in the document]
luffy1996: I missed out core.hpp . I have added the changes

luffy1996:
I tried to test the environment with existing q_learning_test.hpp file in mlpack/tests. It seems working fine. Should I write some specific code to test it ?

zoq:
If you could write a test that uses the new environment, that would be great, if you like you can add a new test to ``q_learning_test.hpp``.

luffy1996:
@zoq I have wriiten a sample test in q_learning_test.cpp

zoq:
Great, can you make the test part of the PR?

luffy1996:
@zoq done :)

src/mlpack/methods/reinforcement_learning/environment/acrobat.hpp
```diff
@@ -0,0 +1,324 @@
...
+  {
+    Rk4(state, action, nextState);
+    nextState.Theta1() = Wrap(nextState.Theta1(), -M_PI, M_PI);
+
+    nextState.Theta2() = Wrap(nextState.Theta2(), -M_PI, M_PI);
+    nextState.AngularVelocity1() = Bound(nextState.AngularVelocity1(),
+                                             -maxVel1 , maxVel1);
+    nextState.AngularVelocity2() = Bound(nextState.AngularVelocity2(),
+                                             -maxVel2 , maxVel2);
+    return -1;
```
zoq: Should this return a reward of 0 if the state is terminal?

src/mlpack/tests/q_learning_test.cpp
```diff
@@ -170,4 +171,74 @@ BOOST_AUTO_TEST_CASE(CartPoleWithDoubleDQN)
...
+      break;
+    }
+
+    /**
+     *
+     * I am using a thresold of -350.0 . While testing it my previous code 
+     * on Q learning https://github.com/luffy1996/rl-using-mlpack/blob/master/qlearning.cpp
+     * I witnessed that -350.0 is average return after 80 episodes.
+     * https://drive.google.com/open?id=1QzZP5egUWhcSJr_byceihU8rIF5sSqdU
+     * Hence I am using -350.0 to test convergence.
```
zoq: The 'goal' is to swing the end to a height of at least one link above the base, over a couple of runs, correct? I'm not sure -350.0 does represent that state. 

So not sure where 50 is coming from:

``size_t requirement = 50; // This variable checks if the game is converging or not .``

src/mlpack/methods/reinforcement_learning/environment/acrobat.hpp
```diff
@@ -0,0 +1,324 @@
...
+  }
+  /**
+   *
+   * @param state The current State
+   */        
+  bool IsTerminal(const State& state) const
+  {
+    return bool (-cos(state.Theta1())-cos(state.Theta1()
+      + state.Theta2()) > 1.0);
+  }
```
luffy1996: The goal of the game is to reach a height of one link as soon as possible. I have implemented this here. Now we cannot wait for long time for the convergence. I wanted a similar approach as of Shantao's to check convergence . 

    /**
     * Reaching running average return 35 is enough to show it works.
     * For the speed of the test case, I didn't set high criterion.
     */

https://github.com/mlpack/mlpack/blob/master/src/mlpack/tests/q_learning_test.cpp

zoq: For the test this is fine, but we should also make sure we have something that can solve the env, e.g. is able to get a reward of ... over 50 consecutive trials. Also, can you also add a test in the form of https://github.com/mlpack/mlpack/blob/master/src/mlpack/tests/rl_components_test.cpp.

luffy1996: @zoq my bad I missed the rl test previous time. However I have added this.
I am not clear about the 50 consecutive trails you are talking about.

zoq: What I mean with 50 consecutive trails is that we should make sure the method we test against (QLearning) is able to solve the env with different starting conditions x times in a row.

luffy1996: I have added it. I believe now it is ready to be merged :)

luffy1996: @zoq Done :).

luffy1996: @zoq Do you mind explaining me the reason to check for 50 consecutive trails. It takes huge amount of time to check for convergence. Ideally we do not have to give so much time for testing. Would you please explain me the where I am missing .

zoq: You are right, we should not implement this as a unit test, but we should test this at least once, e.g. modify the test script you already have. I just like to make sure, the environment works as expected, and this is just another test I think might be helpful.

luffy1996: @zoq it is ready to be merged.

zoq:
I'll have to check the internal logic, but in the meantime it would be great if you could take a look at the style guidelines, e.g. https://github.com/mlpack/mlpack/wiki/DesignGuidelines#line-length-and-wrapping.

src/mlpack/tests/q_learning_test.cpp
```diff
@@ -1,6 +1,6 @@
 /**
  * @file q_learning_test.hpp
- * @author Shangtong Zhang
+ * @author Shangtong Zhang and Rohan Raj
```
rcurtin: You can just add a second line `@author Rohan Raj` here; I think Doxygen will not parse multiple authors on one line correctly.

luffy1996: Should it be like?
@author Shangtong Zhang
@author Rohan Raj

rcurtin: Right, exactly.

luffy1996: I am waiting for this file to be reviewed . Once it is done, I will add all the changes accordingly. Thanks

src/mlpack/methods/reinforcement_learning/environment/acrobat.hpp
```diff
@@ -0,0 +1,317 @@
...
+   * 2 : positive torque
+   */
+  double Torque(const Action& action) const
+  {
+    // Add noise to the Torque
+    /*
+    * Torque is action number - 1.
+    * {0,1,2} -> {-1,0,1} 
+    */
+    double torque = double(action - 1) + mlpack::math::Random(-0.1, 0.1);
```
zoq: We could directly return the torque. This applies for the other functions as well.

src/mlpack/methods/reinforcement_learning/environment/acrobat.hpp
```diff
@@ -0,0 +1,317 @@
...
+
+    double ddtheta2 = (a + d2 / d1 * phi1 - m2 * l1 * lc2 * pow(dtheta1, 2) *
+     sin(theta2) - phi2) / (m2 * pow(lc2, 2) + I2 - pow(d2, 2) / d1);
+
+    double ddtheta1 = -(d2 * ddtheta2 + phi1) / d1;
+    arma::colvec returnValues = {dtheta1, dtheta2, ddtheta1, ddtheta2};
+    return returnValues;
+  };
+  /**
+   * @param value scalar value to wrap
```
zoq: Would be great if you could add a comment what this function does.

src/mlpack/methods/reinforcement_learning/environment/acrobat.hpp
```diff
@@ -0,0 +1,317 @@
...
+    double torque = double(action - 1) + mlpack::math::Random(-0.1, 0.1);
+    return torque;
+  }
+
+  /**
+    * @param state_ Current State
+    * @param Action Action Taken
+    * @param nextState nextState 
+    */
+  arma::colvec Rk4(const arma::colvec state,
```
zoq: Do you think we could use boost odeint here?

src/mlpack/methods/reinforcement_learning/environment/acrobat.hpp
```diff
@@ -0,0 +1,317 @@
...
+    if (value > maximum) value = value - diff;
+    else if (value < minimum) value = value + diff;
+    return value;
+  };
+  /**
+    * @param value scalar value to bound
+    * @param minimum minimum range of bound
+    * @param maximum maximum range of bound
+    */     
+  double  Bound(double value,
```
zoq: Not sure, an extra function is necessary here, since the actual implementation is quite simple and it's only used twice. Let me know what you think.

src/mlpack/methods/reinforcement_learning/environment/acrobat.hpp
```diff
@@ -0,0 +1,317 @@
...
+    double phi1 = - m2 * l1 * lc2 * pow(dtheta2, 2) * sin(theta2)
+      - 2 * m2 * l1 * lc2 * dtheta2 * dtheta1 * sin(theta2)
+      + (m1 * lc1 + m2 * l1) * g * cos(theta1 - M_PI / 2)
+      + phi2;
+
+    double ddtheta2 = (a + d2 / d1 * phi1 - m2 * l1 * lc2 * pow(dtheta1, 2) *
+     sin(theta2) - phi2) / (m2 * pow(lc2, 2) + I2 - pow(d2, 2) / d1);
+
+    double ddtheta1 = -(d2 * ddtheta2 + phi1) / d1;
+    arma::colvec returnValues = {dtheta1, dtheta2, ddtheta1, ddtheta2};
```
zoq: It might be faster, to initalize ``returnValues`` once, and use the parameter directly.

src/mlpack/methods/reinforcement_learning/environment/acrobat.hpp
```diff
@@ -0,0 +1,317 @@
...
+      + state.Theta2()) > 1.0);
+  }
+  /**
+   * @param state Current State
+   * @param torque Torque Applied 
+   */
+  arma::colvec Dsdt(arma::colvec state,
+               const double torque) const
+  {
+    double m1 = linkMass1;
```
zoq: Not sure creating an alias for each parameter helps to improve the readability. I guess, we could also just use ``linkLength1 `` instead of ``lc1``. Let me know what you think.

src/mlpack/methods/reinforcement_learning/environment/acrobat.hpp
luffy1996: Ideally I was hoping this review. I have a habit of naming the variable and then returning it. It looks clean and understandable to me. 

I have visited mlpack structure and it seems values are returned directly. I will add these changes .

src/mlpack/methods/reinforcement_learning/environment/acrobat.hpp
luffy1996: Sure I will add it once the code is reviewed.

zoq: The internal logic looks correct to me.

src/mlpack/methods/reinforcement_learning/environment/acrobat.hpp
luffy1996: I wanted the code to be readable and hence I used the name. I feel the code is more readable to me. However I will alter it as per your suggesstion

src/mlpack/methods/reinforcement_learning/environment/acrobat.hpp
luffy1996: I will add this in my final commit. Please convey me if the code looks correct to you.

src/mlpack/methods/reinforcement_learning/environment/acrobat.hpp
luffy1996: Do you think this is ready to be merged after adding all the reviews?

src/mlpack/methods/reinforcement_learning/environment/acrobat.hpp
luffy1996: I think it is needed here. There are chances that angle may go out of bound. To bring back we need a wrap function. Like 5pi/2 is ideally pi/2. If we not use the wrap feature state space will unnecessarily increase?

How do you see this?

src/mlpack/methods/reinforcement_learning/environment/acrobat.hpp
luffy1996: Do you mind giving me an example. I am not sure how you see this implementation .

src/mlpack/methods/reinforcement_learning/environment/acrobat.hpp
luffy1996: @zoq I do not get how you want this initialisation to take place.

luffy1996:
@zoq, @rcurtin I think this commit is ready to be merged.

src/mlpack/methods/reinforcement_learning/environment/acrobat.hpp
zoq: I was thinking about:

```c++
arma::colvec values(4);
values(0) = ...;
values(1) = ...;

return values;
```

src/mlpack/methods/reinforcement_learning/environment/acrobat.hpp
zoq: Agreed, but do you think we need an extra function, or could we directly use ``std::min(std::max(value, minimum), maximum);``.

src/mlpack/methods/reinforcement_learning/environment/acrobat.hpp
zoq: I haven't really looked into it yet, but https://www.boost.org/doc/libs/master/libs/numeric/odeint/doc/html/index.html might be helpful, this is just an idea, I don't mind to go with the current implementation.

luffy1996:
Dear @rcurtin , @zoq . I have been working on using boost odient library to integrate RK4 method in Acrobat system. To use such implementation, I would have to include boost odient library (https://www.boost.org/doc/libs/1_55_0/boost/numeric/odeint.hpp) . Now this can be done in two ways. I can directly include this header file in my Acrobat file and do away with it. But there might be some version issues that might come because of this (https://github.com/mlpack/mlpack/blob/master/src/mlpack/core/boost_backport/boost_backport_serialization.hpp) .

Please guide me how should I proceed. Is it required to add support for odient in mlpack to continue with runge kutta's method. If odient is already included in mlpack, please provide me sources of the file. Thanks

zoq:
Hello, can you clarify what you mean with "But there might be some version issues that might come because of this[...] ", I'm not sure I see the problem, I guess including ``boost/numeric/odeint.hpp`` inside the env is just fine.

luffy1996:
Dear @zoq , I think problem might be here. I am using boost for the first time. 
If a users boost_version < 105600, will it cause any problem to them? 

#if BOOST_VERSION < 105600
  // Backported unordered_map.
  #include "mlpack/core/boost_backport/unordered_map.hpp"
#else
  // Boost's version.
  #include <boost/serialization/unordered_map.hpp>
#endif

#if BOOST_VERSION == 105800
  /**
   * Boost versions 1.58.0 and earlier have a different vector serialization
   * behaivor as compared to later versions. Notably, loading a
   * std::vector<arma::mat> does not clear the vector before the load
   * in v1.58 and earlier; while in the later versions, the vector is cleared
   * before loading. This causes some tests related to serialization to fail
   * with versions 1.58. This backport solves the issue.
   */
  #ifdef BOOST_SERIALIZATION_VECTOR_HPP
  #pragma message "Detected Boost version is 1.58. Including\
  boost/serialization/vector.hpp before mlpack/core.hpp can cause problems. It\
  should only be necessary to include mlpack/core.hpp and not\
  boost/serialization/vector.hpp."
  #endif
  #include "mlpack/core/boost_backport/collections_load_imp.hpp"
  #include "mlpack/core/boost_backport/collections_save_imp.hpp"
  #include "mlpack/core/boost_backport/vector.hpp"
#else
  #include <boost/serialization/vector.hpp>
#endif

luffy1996:
Directly importing the library is the best option. But I am confused with the boost_backport implementation on mlpack.

rcurtin:
I'd like to avoid another Boost dependency if possible, so if you can check to make sure that there is not another way to solve the problem that would be great.

If we do have to use boost's odeint implementation, it is not a problem to simply `#include <...>` the file from your code.  It's a header-only library so we don't need to change any link options.  Be careful to make sure that you only include the library where you need it, since including boost headers can make compilation take a long time.

As long as the support you are using from the Boost library is available in the oldest version of Boost that we support (I believe 1.49, you can check the readme to see), then there is no need to backport anything.

zoq:
The extra dependency is a good point, let's go with the current implementation.

zoq:
If you can handle the style and comment issues, I think this is ready to be merged.

luffy1996:
@zoq I think now it is ready to be merged.

zoq:
Hello @luffy1996, can you make sure each function has a description.

luffy1996:
Dear @zoq, @rcurtin All the reviews have been added. I have tried to explain the all functions briefly. I'd like this game to be merged. Thanks

luffy1996:
Dear @zoq , @rcurtin Do you have any other suggestions ? I have added all reviews from my side. Thanks

luffy1996:
@zoq , Do I need to add something more here. I guess this is set to be merged. Thanks

PR #1331, opened 2018-03-26
A few more test adjustments

rcurtin:
I went through the list of unstable builds and adjusted these tests.  It should further reduce the test error rate.


rcurtin:
Sounds good, once I see this passes Travis I'll merge and restart the builds on masterblaster.

Issue #1332, opened 2018-03-27
Implement Reward Clipping for Reinforcement Learning games


chennakeshava1998:
If different games use different scales of rewards, then it becomes hard to train the same RL algorithm for different games. This issue will be solved if clipped rewards are used

sshkhr:
@chennakeshava1998  I think this can be marked as resolved now. Thanks for pointing out this issue

ShikharJ:
Thanks for the work @sshkr.

PR #1333, opened 2018-03-27
Add reward_clipping.hpp file

chennakeshava1998:
@zoq As per our discussion on IRC, I have tried to implement reward_clipping function here as described by Issue number #1332 

I have a few doubts in the file. I wanted to clarify them, before writing the test file for this method

1. Usually rewards are scalars in RL games. But should I use arma::mat instead? It might be helpful in the future though.
2. Should I return values in the GetReward() function? If we are handling large matrices, it might be better to return references. Since clippedReward is a member of RewardClipping class, the returned reference will be valid until the current object is in scope. 

But my doubt is,  since the function is returning a reference, we can accidentally (force)change the clipped reward by using the assignment operator; something like 

`` RewardClipping obj(-1, 1);
double reward = 10;
obj.GetReward(reward) = 2; // this is valid as per c++ syntax, but logically wrong
``
Please correct me if I am wrong anywhere.

### DIFF
--- a/src/mlpack/methods/reinforcement_learning/reward_clipping.hpp
+++ b/src/mlpack/methods/reinforcement_learning/reward_clipping.hpp
@@ -1,0 +1,87 @@
+/**
+ * @file reward_clipping.hpp
+ * @author Chenna Keshava B S
+ *
+ * Class for implementiong Reward clipping 
+ *
+ * mlpack is free software; you may redistribute it and/or modify it under the
+ * terms of the 3-clause BSD license.  You should have received a copy of the
+ * 3-clause BSD license along with mlpack.  If not, see
+ * http://www.opensource.org/licenses/BSD-3-Clause for more information.
+ */
+#ifndef MLPACK_SRC_MLPACK_METHODS_REINFORCEMENT_LEARNING_REWARD_CLIPPING_HPP
+
+#define MLPACK_SRC_MLPACK_METHODS_REINFORCEMENT_LEARNING_REWARD_CLIPPING_HPP
+
+
+namespace mlpack {
+namespace rl {
+
+/**
+ * Interface for clipping the reward to some value between the specified maximum and minimum value
+ * (Clipping here is implemented as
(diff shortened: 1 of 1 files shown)


rcurtin:
I was not a part of the IRC discussion, so I apologize if I am asking something that is already known, but can you provide some information about how we would plug this into the existing mlpack infrastructure?

chennakeshava1998:
1. Reward Clipping is (as of now) used mostly in RL. So, I was planning to place it here - lpack/src/mlpack/methods/reinforcement_learning/. Another option it to place it in - mlpack/src/mlpack/core/metrics/. Which makes it more generic to other techniques also. 
2. I am planning to model the API in a similar way to that of GradientClipping class.

zoq:
Can you think of a test, that uses the clipping technique?

chennakeshava1998:
One comprehensive way is: Run the algorithm with and without reward clipping on the same set of games, and then check various parameters like - stability of learning (is the Q-Values fluctuating a lot?), the accuracy, the average number of iterations (or Computation time) to reach a threshold accuracy etc

chennakeshava1998:
But the reward clipping functionality itself can be tested like that of Gradient Clipping -  mlpack/src/mlpack/tests/gradient_clipping_test.cpp . Just test if the clipped reward is equal to the predetermined values of the test case. But I don't think this is what you wanted.

zoq:
Thanks for the feedback, do you think we could implement both ideas as a simple test, e.g. training Q-Learning on a simple task with reward clipping and another simple test in the form of gradient_clipping_test?

chennakeshava1998:
Yes, I think we can do that. I was thinking of training q-learning on cartpole and mountain-car. Also, we will have to use the same neural network(if we are testing on multiple games) isn't it? Or are you suggesting a case of q-learning for only one game with reward clipping?

zoq:
Either one of those two is fine, thanks for looking into the tests.

rcurtin:
I'm going to go ahead and close this for inactivity.  Please feel free to reopen it if work continues. :+1:

PR #1335, opened 2018-03-29
More test fixes

rcurtin:
Just a couple more test fixes; they are all just little tolerance adjustments except for a modification to the randomized PCA dimensionality reduction test.  Sometimes the LU factorization can fail, so we simply catch the exception and try again, up to 3 times.


rcurtin:
Once Travis succeeds, I'll merge and restart the matrix builds.

PR #1336, opened 2018-03-29
Documentation update

rcurtin:
I went through the tutorials and the guide and made sure that everything is up to date.  After this is merged I'll have to double-check that doxygen renders everything properly.


doc/guide/build.hpp
```diff
@@ -46,61 +66,74 @@ system and have headers present:
  - Boost (math_c99, program_options, serialization, unit_test_framework, heap,
           spirit) >= 1.49
 
+For Python bindings, the following packages are required:
+
+ - setuptools
```
zoq: Not sure, if this is necessary to note here, that ``pytest-runner`` is also needed for the tests? I guess in most cases it's going to be installed via the setuptools?

PR #1337, opened 2018-03-29
mlpack_preprocess_describe CLI binding tests

PlantsAndBuildings:
Added tests for `mlpack_preprocess_describe` ( #1152 ) 

Currently, the `mlpack_preprocess_describe` utility prints output to console and doesn't give back a "results" matrix - I've added a `PARAM_MATRIX_OUT` in `preprocess_describe_main.cpp` for the same.

Also, checks for the input parameters were missing - those have been added in `preprocess_describe_main.cpp` as well.

Please let me know if I've missed important tests.

### DIFF
--- a/src/mlpack/methods/preprocess/preprocess_describe_main.cpp
+++ b/src/mlpack/methods/preprocess/preprocess_describe_main.cpp
@@ -246,7 +246,7 @@ static void mlpackMain()
         Skewness(feature, fStd, fMean, population),
         Kurtosis(feature, fStd, fMean, population),
         StandardError(feature.n_elem, fStd)
-    });
+    }).t();
   };
 
   // If the user specified dimension, describe statistics of the given
--- a/src/mlpack/methods/preprocess/preprocess_describe_main.cpp
+++ b/src/mlpack/methods/preprocess/preprocess_describe_main.cpp
@@ -61,6 +61,7 @@ PARAM_FLAG("population", "If specified, the program will calculate statistics "
 PARAM_FLAG("row_major", "If specified, the program will calculate statistics "
     "across rows, not across columns.  (Remember that in mlpack, a column "
     "represents a point, so this option is generally not necessary.)", "r");
+PARAM_MATRIX_OUT("output", "matrix to save output to", "o");
 
 /**
  * Calculates the sum of deviations to the Nth Power.
@@ -164,6 +165,24 @@ static void mlpackMain()
   // Load the data.
   arma::mat& data = CLI::GetParam<arma::mat>("input");
 
+  const size_t dimensions = rowMajor ? data.n_cols : data.n_rows;
+  RequireParamValue<int>("dimension", [](int x) { return x >= 0; }, true,
+      "dimension must be nonnegative");
+  std::ostringstream error;
+  error << "dimension must be less than the number of dimensions "
+      << "of the input data (" << dimensions << ")";
+  RequireParamValue<int>("dimension",
+      [dimensions](int x) { return size_t(x) < dimensions; }, true, error.str());
+
+  RequireParamValue<int>("precision", [](int x) { return x >= 0; }, true,
+      "precision must be nonnegative");
+  
+  RequireParamValue<int>("width", [](int x) { return x >= 0; }, true,
+      "width must be nonnegative");
+
+  // Create a mat to store the output in.
+  arma::mat outputMat(dimensions, 11);
+
   // Generate boost format recipe.
   const string widthPrecision("%-" + to_string(width) + "." +
       to_string(precision));
@@ -185,7 +204,7 @@ static void mlpackMain()
       % "range" % "skew" % "kurt" % "SE" << endl;
 
   // Lambda function to print out the results.
-  auto PrintStatResults = [&](size_t dim, bool rowMajor)
+  auto PrintStatResults = [&](size_t dim, bool rowMajor, arma::mat& outputMat)
   {
     arma::rowvec feature;
     if (rowMajor)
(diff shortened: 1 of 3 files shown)


src/mlpack/methods/preprocess/preprocess_describe_main.cpp
```diff
@@ -213,21 +232,39 @@ static void mlpackMain()
         % Kurtosis(feature, fStd, fMean, population)
         % StandardError(feature.n_elem, fStd)
         << endl;
+
+    // Store the output in the output arma::mat.
+    outputMat.row(dim) = arma::vec({
```
manish7294: @PlantsAndBuildings Would you mind, justifying the rationale behind the need for output matrix for data description? I think there is no necessity of having output matrix, in this case, just the print should do the required. Moreover, in the current state, the user will not able to comprehend what does each output cell signifies (As you will be getting a column of 11 floating point values in the output file), if he/she went looking for the results in the output file.

PlantsAndBuildings: Thanks for the review.

I'll start off by saying that - yes, you're absolutely right that the user will not be able to make sense of the output. Also, I know that what I'm about to describe is not a very clean way to do things - and I'm open to suggestions to do things differently. 

I added that param because I don't see any other way to access the results  programatically - which is a problem if you're trying to write tests for this binding. I'll elaborate - consider the following test which checks the working of --row_major command line flag:
``` c++
 BOOST_AUTO_TEST_CASE(PreprocessDescribeRowMajorTest)
 {
   arma::mat dataset;
   data::Load("trainSet.csv", dataset);

   SetInputParam("input", std::move(dataset));
   mlpackMain();

   arma::mat colMajorOutput = CLI::GetParam<arma::mat>("output");

   SetInputParam("row_major", true);
   mlpackMain();

   arma::mat rowMajorOutput = CLI::GetParam<arma::mat>("output");

   BOOST_REQUIRE_EQUAL(colMajorOutput.n_rows,
                       CLI::GetParam<arma::mat>("input").n_rows);
   BOOST_REQUIRE_EQUAL(rowMajorOutput.n_rows,
                       CLI::GetParam<arma::mat>("input").n_cols);
   BOOST_REQUIRE_EQUAL(colMajorOutput.n_cols, rowMajorOutput.n_cols);
 }
```
I'm not sure how we would able to get `colMajorOutput` and `rowMajorOutput` without having to add the `output` param. Let me know what you think.

manish7294:
Yeah, this seems to be a problem, if we want to carry out dimensional tests.
I spent some time thinking about it, and this is what I think, we may carry out  :

- As we currently don't have the individual functions tests for preprocess describe (in case I didn't miss them out), adding those will assure that we are printing 11 values for a particular dimension. Since row-major just makes a rowvec of different size, this could act as a replacement for:
`
BOOST_REQUIRE_EQUAL(colMajorOutput.n_cols, rowMajorOutput.n_cols);
`
- As far as number of dimensions are concerned, we may do something like:
```
// This test is just a kind of boundary check.
SetInputParam("row_major", true);
SetInputParam("dimension", CLI::GetParam<arma::mat>("input").n_cols - 1);

// Shouldn't throw any error.
mlpackMain()

SetInputParam("dimension", CLI::GetParam<arma::mat>("input").n_cols);

BOOST_REQUIRE_THROW(mlpackMain(), std::runtime_error);
```
in place of
```
BOOST_REQUIRE_EQUAL(rowMajorOutput.n_rows,
                       CLI::GetParam<arma::mat>("input").n_cols);
```

I myself not sure about the clearness of above ideas & also not sure whether to even consider them. So at this point, it would be the best to have @rcurtin @zoq opinion about the output matrix idea.

src/mlpack/methods/preprocess/preprocess_describe_main.cpp
rcurtin: This is definitely a hard binding to test.  I have to agree with Manish, it looks like the user will not be able to understand the output of the output matrix.  I think this will be a bit confusing.  I do have another idea, let me know what you think:

The output all goes to `Log::Info`.  But we can hijack the destination; this is done in the tests for `PrefixedOutStream`:

```
std::ostringstream oss;
Log::Info.destination = oss;

// Call mlpackMain() and the output will get filled up...
mlpackMain();

// Store the output.
const std::string output = oss.str();
oss.clear();

// Now change some parameters and run again...
...
mlpackMain();

// Now we can check that the output was different.
BOOST_REQUIRE_NE(oss.str(), output);
```

What I showed above only shows how to check if the output of two runs is different.  I think it would be possible and not too hard to actually parse the output, but honestly in most cases here I think just checking that the output is different is sufficient (like for the `row_major` test you have above).

PlantsAndBuildings:
```
std::ostringstream oss;
Log::Info.destination = oss;
```
Unfortunately, this is not allowed. Refer [this](http://en.cppreference.com/w/cpp/io/basic_ostream/operator%3D). I'm looking into this a bit more - but I don't see too many alternatives other than what Manish has suggested (the boundary checks approach).

rcurtin: You are right, sorry, `destination` is a reference.  Do you think the following might work?

```
PrefixedOutStream oldInfo = Log::Info;
std::ostringstream oss;
Log::Info = PrefixedOutStream(oss, ""); // no prefix needed for this.

// Call mlpackMain(), do other stuff...
mlpackMain();
...

// Now reset what Log::Info was.
Log::Info = oldInfo;
```

I think that will work but I am not sure...

PlantsAndBuildings: Sorry for the slow response on this. No, this too will not work. It says:
```
/root/mlpack/src/mlpack/core/util/prefixedoutstream.hpp:46:7: error: non-static reference member 'std::ostream& mlpack::util::PrefixedOutStream::destination', can't use default assignment operator
```

rcurtin: You are right, I did not think that through.  I can supply a patch that will internally change `destination` to `std::ostream*`; that could help.  Would you like me to do that?

PlantsAndBuildings: Umm .. okay. I do have a few concerns though.

It just seems like a lot of effort (on your end, since you're the one who will have to figure out how to write the patch) to go through for testing just one binding. Will doing this (ie. changing `destination` to `std::ostream*`) be useful for some other stuff as well? Rather, are there any more use-cases where we would need to hijack the INFO stream? If not, then I guess we should consider going with one of the other alternatives:

1. What Manish suggested initially - the boundary checks approach.
2. Writing explicitly to an output file from `preprocess_describe_main.cpp` binding - this way we can add column headings to the output file and the output would make sense to the user (see previous comments).
3. "Hiding" the output parameter -- something like `PARAM_MAT_OUT("__output")` and explicitly warning the user that this option is meant for internal use only.
4. This might be a bit drastic but - scrapping the tests that need to look at the output of the binding? - since the binding is obviously working fine.

Either way, if you think that providing the patch (or perhaps holding my hand through writing it) is relatively easy and worth the effort - then that is what we should go with. Let me know, I'm happy to go with any approach.

rcurtin: Sorry for the slow response on this one, it has been a busy past few days and I have fallen far behind.

Honestly I don't have a huge problem with option 4.  I agree that it is giving us output as expected.  Manish's suggestion to check the number of dimensions sounds just fine to me.

I do think the binding is more useful when it provides output on stdout instead of into a file.  If we want to change the output, I might think a better way to do it would be to use `PARAM_STRING_OUT()` or `PARAM_VECTOR_OUT()` or something like this to provide the output.  But we could open a separate issue for that if there was interest in solving it.  I don't think it's a hugely pressing need so I don't think we need to do that here.

So, up to you---if you'd rather take suggestions (4) and (1), that would be just fine with me.  But if you want to do something more complicated, we can do that also.

PR #1339, opened 2018-03-30
Quickstart guides for CLI and Python

rcurtin:
I've written some very quick "quickstart" guides.  I still have a revision or two I want to do, and I need to fix links inside of them, but otherwise I think this is just about ready.


doc/guide/cli_quickstart.hpp
```diff
@@ -0,0 +1,230 @@
...
+for i in `seq 1 10`; do
+    item=`cat recommendations.csv | awk -F',' '{ print $'$i' }'`;
+    head -n $(($item + 2)) movies.csv | tail -1 | \
+        sed 's/^[^,]*,[^,]*,//' | \
+        sed 's/\(.*\),.*$/\1/' | sed 's/"//g';
+done
+@endcode
+
+Here is some example output, showing that user 1 seems to have good taste in
+movies:
```
zoq: Agreed :)

doc/guide/cli_quickstart.hpp
```diff
@@ -0,0 +1,230 @@
...
+@endcode
+
+@section cli_quickstart_nextsteps Next steps with mlpack
+
+Now that you have done some simple work with mlpack, you have seen how it can
+easily plug into a data science production workflow for the command line.  A
+great thing to do next would be to look at more documentation for the mlpack
+command-line programs:
+
+ - <a href="http://www.mlpack.org/docs/mlpack-2.2.5/man.html">mlpack
```
zoq: Do we update the link after 3.0 is out?

zoq:
Looks still good to me.

rcurtin:
Ok, here is the result:

http://www.mlpack.org/docs/mlpack-git/doxygen/python_quickstart.html
http://www.mlpack.org/docs/mlpack-git/doxygen/cli_quickstart.html

(I hand-rebuilt the documentation on the website for testing.)

Let me know if you see any issues.

zoq:
Don't see any issues on the two pages.

PR #1340, opened 2018-03-30
Minor changes

luffy1996:
Removed unnecessary lines in worker for Asynchronous learning

### DIFF
--- a/src/mlpack/methods/reinforcement_learning/worker/n_step_q_learning_worker.hpp
+++ b/src/mlpack/methods/reinforcement_learning/worker/n_step_q_learning_worker.hpp
@@ -100,7 +100,6 @@ class NStepQLearningWorker
     episodeReturn += reward;
     steps++;
 
-    terminal = terminal || steps >= config.StepLimit();
     if (deterministic)
     {
       if (terminal)
--- a/src/mlpack/methods/reinforcement_learning/worker/n_step_q_learning_worker.hpp
+++ b/src/mlpack/methods/reinforcement_learning/worker/n_step_q_learning_worker.hpp
@@ -100,6 +100,7 @@ class NStepQLearningWorker
     episodeReturn += reward;
     steps++;
 
+    terminal = (terminal || steps >= config.StepLimit());
     if (deterministic)
     {
       if (terminal)
--- a/src/mlpack/methods/reinforcement_learning/worker/one_step_q_learning_worker.hpp
+++ b/src/mlpack/methods/reinforcement_learning/worker/one_step_q_learning_worker.hpp
@@ -100,6 +100,7 @@ class OneStepQLearningWorker
     episodeReturn += reward;
     steps++;
 
+    terminal = (terminal || steps >= config.StepLimit());
     if (deterministic)
     {
       if (terminal)
--- a/src/mlpack/methods/reinforcement_learning/worker/one_step_q_learning_worker.hpp
+++ b/src/mlpack/methods/reinforcement_learning/worker/one_step_q_learning_worker.hpp
@@ -100,7 +100,6 @@ class OneStepQLearningWorker
     episodeReturn += reward;
     steps++;
 
-    terminal = terminal || steps >= config.StepLimit();
     if (deterministic)
     {
       if (terminal)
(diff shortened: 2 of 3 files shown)


src/mlpack/methods/reinforcement_learning/worker/n_step_q_learning_worker.hpp
```diff
@@ -100,7 +100,6 @@ class NStepQLearningWorker
     episodeReturn += reward;
     steps++;
 
-    terminal = terminal || steps >= config.StepLimit();
```
zoq: In this case we remove to option to return if the user defined number of steps is reached, maybe I missed something?

luffy1996: I am still not getting the use of this line here. This condition does have any set of commands under its body. I believe this will always be true as assignment operation has a logic of 1.

I think there should be a if before this condition. That will explain the use over here. How do you see this?

zoq: The condition in this case is: ``terminal || steps >= config.StepLimit()``, so if either on of them is true, we return true otherwise false. Here is a simple example: http://cpp.sh/53hsa. Also note ``terminal || steps >= config.StepLimit()`` is evaluated first, and the assignment afterwards.

luffy1996: I believe this should be under if condition. Why this is executed as a statement I am not getting ? There is some implementation error , right?

luffy1996: I am speaking of this cpp.sh/9igeg . @zoq

zoq: The condition is in line: 105, 123, 182, 132. Let's take the first as an example:

```c++
if (deterministic)
{
  if (terminal)
  {
    totalReward = episodeReturn;
    Reset();
    // Sync with latest learning network.
    network = learningNetwork;
    return true;
  }
  state = nextState;
  return false;
}
```

if terminal is false we skip the final sync step and directly return the state.

luffy1996: Dear @zoq , Do you think this should be something like this?
```
    if (terminal|| steps >= config.StepLimit())
    {
        if (deterministic)
        {
          if (terminal)
            {
              totalReward = episodeReturn;
              Reset();
              // Sync with latest learning network.
              network = learningNetwork;
              return true;
            }
        state = nextState;
        return false;
        }
    }
```
I am pretty doubtful about the line 
```
terminal = terminal || steps >= config.StepLimit();
```

zoq: Sorry for the slow response, I'm not sure I get what you like to achieve here, in the deterministic case (inference/test), we just need the state, so there is no need to update the network. If we test ``if (terminal|| steps >= config.StepLimit())`` first, we would train the method. I'm not sure I missed something, would be great if you could elaborate on that one.

luffy1996: ```terminal = terminal || steps >= config.StepLimit();```
Take this line. It checks the condition if ```terminal``` is true or ```steps >=  config.StepLimit()```. Now we have a semicolon after this ```;```. I think now this entire expression becomes unnecessary because we did not gain any thing from this boolean check.

Take line 98 ```bool terminal = environment.IsTerminal(nextState);```. Here we have initialised terminal so another assignment operation wasn't required here. More over I am having confusion how this expression ```terminal = terminal || steps >= config.StepLimit();``` is helping us. I understand what is the logic of the statement but I do not get how this logic is used.

I guess now my doubt should be clear. Thanks

luffy1996: Hello @zoq , I understood the code. I was missing out one point. I am grateful to you for providing me assistance here. If no one has any problem I guess this PR could be closed.

zoq: Okay, good to hear that you figured it out.

zoq:
Closing; see the discussion on the first comment for more details.

PR #1341, opened 2018-03-30
Implement Flexible ReLU 

Manthan-R-Sheth:
This is in continuity to #1281.
Function call overheads have been removed and code is simplified to include the calculation in `Forward() `and `Backward()`.

### DIFF
--- a/src/mlpack/methods/ann/layer/flexible_relu.hpp
+++ b/src/mlpack/methods/ann/layer/flexible_relu.hpp
@@ -1,6 +1,7 @@
 /**
  * @file flexible_relu.hpp
  * @author Aarush Gupta
+ * @author Manthan-R-Sheth
  *
  * Definition of FlexibleReLU layer as described by
  * Suo Qiu, Xiangmin Xu and Bolun Cai in
@@ -15,6 +16,7 @@
  *  title   = {FReLU: Flexible Rectified Linear Units for Improving
  *             Convolutional Neural Networks}
  *  journal = {arxiv preprint},
+ *  URL     = {https://arxiv.org/pdf/1706.08098.pdf},
  *  year    = {2018}
  * }
  * @endcode
--- a/src/mlpack/methods/ann/layer/flexible_relu.hpp
+++ b/src/mlpack/methods/ann/layer/flexible_relu.hpp
@@ -1,0 +1,194 @@
+/**
+ * @file flexible_relu.hpp
+ * @author Aarush Gupta
+ *
+ * Definition of FlexibleReLU layer as described by
+ * Suo Qiu, Xiangmin Xu and Bolun Cai in
+ * "FReLU: Flexible Rectified Linear Units for Improving Convolutional
+ *   Neural Networks", 2018
+ *
+ * For more information, read the following paper:
+ *
+ * @code
+ * @article{
+ *  author  = {Suo Qiu, Xiangmin Xu and Bolun Cai},
+ *  title   = {FReLU: Flexible Rectified Linear Units for Improving
+ *             Convolutional Neural Networks}
+ *  journal = {arxiv preprint},
+ *  year    = {2018}
+ * }
+ * @endcode
+ *
+ * mlpack is free software; you may redistribute it and/or modify it under the
+ * terms of the 3-clause BSD license.  You should have received a copy of the
+ * 3-clause BSD license along with mlpack.  If not, see
+ * http://www.opensource.org/licenses/BSD-3-Clause for more information.
+ */
+#ifndef MLPACK_METHODS_ANN_LAYER_FLEXIBLERELU_HPP
+#define MLPACK_METHODS_ANN_LAYER_FLEXIBLERELU_HPP
+
+#include <mlpack/prereqs.hpp>
+
+namespace mlpack {
+namespace ann /**Artificial Neural Network*/ {
+
+/**
+ *The FlexibleReLU activation function, defined by
+ *
+ * @f{eqnarray*}{
+ * f(x) &=& \max(0,x)+alpha \\
+ * f'(x) &=& \left\{
+ * 	 \begin(array){lr}
+ * 	   1 & : x > 0 \\
+ * 	   0 & : x \le 0
+ * 	 \end{array}
+ * \right
+ *@f}
+ *
+ *@tparam InputDataType Type of the input data ( arma::colvec, arma::mar,
+ *        arma::sp_mat or arma::cube)
+ *
+ *@tparam OutputDataType Type of the output data (arma::colvec, arma::mat,
+ *        arma::sp_mat or arma::cube)
+ *
+ */
+
+
+template <
+	     typename InputDataType = arma::mat,
+	     typename OutputDataType = arma::mat
+>
+class FlexibleReLU
+{
+public:
+  /**
+   *
+   * Create the FlexibleReLU object using the specified parameters.
+   * The non zero parameter can be adjusted by specifying the parameter
+   * alpha which controls the range of the relu function. ( Default alpha = 0)
+   * This parameter is trainable. 
+   *@param alpha Parameter for adjusting the range of the relu function.
+   *
+   */
+  FlexibleReLU(const double alpha = 0);
+
+  /**
+   * Ordinary feed forward pass of a neural network, evaluating the function
+   * f(x) by propagating the activity forward through f.
+   *
+   * @param input Input data used for evaluating the specified function.
+   * @param output Resulting output activation.
+   */
+  template<typename InputType, typename OutputType>
+  void Forward(const InputType&& input, OutputType&& output);
+
+  /**
+   * Ordinary feed backward pass of a neural network, calculating the function
+   * f(x) by propagating x backwards through f. Using the results from the feed
+   * forward pass.
+   *
+   * @param input The propagated input activation.
+   * @param gy The backpropagated error.
+   * @param g The calculated gradient.
+   */
+  template<typename DataType>
+  void Backward(const DataType&& input, DataType&& gy, DataType&& g);
+
+  //! Get the input parameter.
+  InputDataType const& InputParameter() const { return inputParameter; }
+  //!Modify the input parameter.
+  InputDataType& InputParameter() { return inputParameter; }
+
+  //!Get the output parameter.
+  OutputDataType const& OutputParameter() const { return outputParameter; }
+  //Modify the output parameter.
+  OutputDataType& OutputParameter() { return outputParameter; }
+
+  //!Get the delta.
+  OutputDataType const& Delta() const { return delta; }
+  //!Modify the delta.
+  OutputDataType& Delta() { return delta;}
+  
+  //!Get the parameter controlling the range of the relu function.
+  double const& Alpha() const { return alpha; }
+  //!Modify the parameter controlling the range of the relu function.
+  double& Alpha() { return alpha; }
+
+  /**
+   * Serialize the layer.
+   */
+  template<typename Archive>
+  void serialize(Archive& ar, const unsigned int /* version*/);
+
+private:
+  /**
+   * Computes the FlexibleReLU function
+   *
+   * @param x Input data.
+   * @return f(x).
+   */
+ doubel Fn(const double x)
+ {
+ 	return (std::max(x,0) + alpha);
+ }
+
+ /**
+   * Computes the FlexibleReLU function using a dense matrix as input.
+   *
+   * @param x Input data.
+   * @param y The resulting output activation.
+   */
+ template<typename eT>
+ void Fn(const arma::Mat<eT>& x, arma::Mat<eT>& y)
+ {
+ 	y = arma::max(arma::zeros<arma::Mat<eT> >(x.n_rows, x.n_cols), x)+alpha;
+ }
+ 
+ /**
+  * Computes the first derivative of the LeakyReLU function.
+  *
+  * @param x Input data.
+  * @return f'(x)
+  */
+ double Deriv(const double x)
+ {
+   return x > 0;
+ }
+
+ /**
+   * Computes the first derivative of the FlexibleReLU function.
+   *
+   * @param y Input activations.
+   * @param The resulting dreivatives
+   */
+ template<typename InputType, typename, OutputType>
+ void Deriv(const InputParameter& x, OutputType& y)
+ {
+ 	y = x
+
+ 	for (size_t i = 0; i < x.n_elem; i++)
+	{
+	  y(i) = Deriv(x(i));
+	}
+ }
+
+ //! Locally-stored delta object.
+ OutputDataType delta;
+
+ //! Locally-stored input parameter object.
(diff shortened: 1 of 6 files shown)


sourabhvarshney111:
@Manthan-R-Sheth I think this is complete. Is it?

Manthan-R-Sheth:
Lets wait for @zoq and @rcurtin

src/mlpack/methods/ann/layer/flexible_relu.hpp
```diff
@@ -0,0 +1,146 @@
...
+ *
+ * Definition of FlexibleReLU layer as described by
+ * Suo Qiu, Xiangmin Xu and Bolun Cai in
+ * "FReLU: Flexible Rectified Linear Units for Improving Convolutional
+ *   Neural Networks", 2018
+ *
+ * For more information, read the following paper:
+ *
+ * @code
+ * @article{
```
zoq: This is missing a name, also do you mind to move this into the class description block, that way, doxygen can pick this up.

src/mlpack/methods/ann/layer/flexible_relu.hpp
```diff
@@ -0,0 +1,146 @@
...
+ *  author  = {Suo Qiu, Xiangmin Xu and Bolun Cai},
+ *  title   = {FReLU: Flexible Rectified Linear Units for Improving
+ *             Convolutional Neural Networks}
+ *  journal = {arxiv preprint},
+ *  URL     = {https://arxiv.org/pdf/1706.08098.pdf},
```
zoq: Can you link to the arxiv page instead using the pdf link here.

src/mlpack/methods/ann/layer/flexible_relu.hpp
```diff
@@ -0,0 +1,146 @@
...
+ * 	 \end{array}
+ * \right
+ *@f}
+ *
+ *@tparam InputDataType Type of the input data ( arma::colvec, arma::mar,
+ *        arma::sp_mat or arma::cube)
+ *
+ *@tparam OutputDataType Type of the output data (arma::colvec, arma::mat,
+ *        arma::sp_mat or arma::cube)
+ *
```
zoq: Looks like we could remove the extra line here.

src/mlpack/methods/ann/layer/flexible_relu.hpp
```diff
@@ -0,0 +1,146 @@
...
+ * 	 \begin(array){lr}
+ * 	   1 & : x > 0 \\
+ * 	   0 & : x \le 0
+ * 	 \end{array}
+ * \right
+ *@f}
+ *
+ *@tparam InputDataType Type of the input data ( arma::colvec, arma::mar,
+ *        arma::sp_mat or arma::cube)
+ *
```
zoq: Do you mind to remove the extra line and the extra space before ``arma::colvec``?

sourabhvarshney111:
@Manthan-R-Sheth Can you fix the style commit?

src/mlpack/methods/ann/layer/flexible_relu_impl.hpp
```diff
@@ -0,0 +1,71 @@
...
+  // Nothing to do here.
+}
+
+template<typename InputDataType, typename OutputDataType>
+template<typename InputType, typename OutputType>
+void FlexibleReLU<InputDataType, OutputDataType>::Forward(
+    const InputType&& input, OutputType&& output)
+{
+  output = arma::max(arma::zeros<InputType>(input.n_rows, input.n_cols), input)
+    + alpha;
```
rcurtin: A minor style issue---the second line here should be doubly indented (four spaces, not two).  Also I think you might be able to do this with a lambda passed to `.transform()`, which could make it a little faster.

src/mlpack/methods/ann/layer/flexible_relu_impl.hpp
```diff
@@ -0,0 +1,71 @@
...
+
+template<typename InputDataType, typename OutputDataType>
+template<typename DataType>
+void FlexibleReLU<InputDataType, OutputDataType>::Backward(
+    const DataType&& input, DataType&& gy, DataType&& g)
+{
+  DataType derivative;
+
+  //! Compute the first derivative of FlexibleReLU function.
+  derivative = input;
```
rcurtin: You can save a copy here if we do `derivative.set_size(input.n_rows, input.n_cols)`, which will just set the size.

src/mlpack/methods/ann/layer/flexible_relu_impl.hpp
```diff
@@ -0,0 +1,71 @@
...
+
+  for (size_t i = 0; i < input.n_elem; i++)
+  {
+    derivative(i) = input(i) > 0? 1 : 0;
+  }
```
rcurtin: You could also replace this with a call to `.transform()`.

Manthan-R-Sheth:
@rcurtin 
Yes your are right, since there is a trainable parameter, Gradient() and Parameters() should be implemented. I have done that and I have optimized the code too.
Do you think CheckGradient() test should be implemented here now?

src/mlpack/methods/ann/layer/flexible_relu_impl.hpp
```diff
@@ -0,0 +1,95 @@
...
+  const arma::Mat<eT>&& input, arma::Mat<eT>&& error,
+  arma::Mat<eT>&& gradient)
+{
+  if (gradient.n_elem == 0)
+  {
+    gradient = arma::zeros<arma::Mat<eT>>(1, 1);
+  }
+
+  arma::mat zeros = arma::zeros<arma::Mat<eT>>(input.n_rows, input.n_cols);
+  gradient(0) = arma::accu(error % arma::min(zeros, input)) / input.n_cols;
```
rcurtin: It would be a lot better here if we could avoid allocating `zeros`; that will be time-consuming.

src/mlpack/methods/ann/layer/flexible_relu_impl.hpp
```diff
@@ -0,0 +1,95 @@
...
+
+template<typename InputDataType, typename OutputDataType>
+template<typename InputType, typename OutputType>
+void FlexibleReLU<InputDataType, OutputDataType>::Forward(
+    const InputType&& input, OutputType&& output)
+{
+  int i = -1;
+  output = arma::zeros<InputType>(input.n_rows, input.n_cols);
+  output.transform([input, &i, this](double val) { ++i;
+      return (std::max(input(i), 0.0) + alpha(0)); } );
```
rcurtin: Ack, I did not realize that `transform()` is in-place.  Why not use something like `arma::clamp()` instead then?  i.e. `output = arma::clamp(input, 0.0, DBL_MAX) + alpha(0);`

src/mlpack/methods/ann/layer/flexible_relu_impl.hpp
```diff
@@ -0,0 +1,95 @@
...
+template<typename DataType>
+void FlexibleReLU<InputDataType, OutputDataType>::Backward(
+    const DataType&& input, DataType&& gy, DataType&& g)
+{
+  DataType derivative;
+  //! Compute the first derivative of FlexibleReLU function.
+  derivative.set_size(input.n_rows, input.n_cols);
+  int i = -1;
+  derivative.transform([input, &i](double val) { ++i;
+    return (input(i) > 0? 1 : 0); } );
```
rcurtin: Here I think we could just use a boolean expression: `derivation = (input > 0);`  (or some variant like that)

src/mlpack/methods/ann/layer/flexible_relu_impl.hpp
Manthan-R-Sheth: Yes we just needed a function like that

Manthan-R-Sheth:
@rcurtin 
I have updated according to the suggestions.

src/mlpack/methods/ann/layer/flexible_relu_impl.hpp
Manthan-R-Sheth: I have used this :
```derivative = arma::sign(input);```
```derivative.elem(arma::find(derivative < 0.0)) += 1;```

This should be good to go or the lambda was faster already?

**Comment by rcurtin** [comment]
Looks like the gradient test is failing; can you look into it please?

```
/home/travis/build/mlpack/mlpack/src/mlpack/tests/ann_layer_test.cpp(677): fatal error in "GradientFlexibleReLULayerTest": critical check CheckGradient(function) <= 1e-4 failed [0.45575305843508807 > 0.0001]
```

**Comment by Manthan-R-Sheth** [comment]
@rcurtin @zoq 
Can you have a look at the ```Backward()``` and ```Gradient()``` functions. It look right to me. 
Maybe understanding what is wrong can make ```CheckGradient()``` pass too.

**Comment by zoq** [comment]
I don't think that's the case here, have to take a closer look at the code, especially the ``Gradient`` step.

**Comment by Manthan-R-Sheth** [comment]
@zoq @rcurtin 
I gave the same input for these two tests :- 
1. ```bin/mlpack_test -t ANNLayerTest```
2. ```bin/mlpack_test -t ANNLayerTest/GradientFlexibleReLULayerTest```

and I found that the errors after `model->Evaluate() ` are different and so are gradients. This is why its passing in 1 but failing in 2. Infact on printing the values of parameters for both cases, it was found that 2 has estimated and original gradient 0. 
Any ideas on what could be a possible cause for this.

File: src/mlpack/methods/ann/layer/flexible_relu_impl.hpp [hunk 11]
```diff
@@ -0,0 +1,87 @@
...
+
+template<typename InputDataType, typename OutputDataType>
+template<typename DataType>
+void FlexibleReLU<InputDataType, OutputDataType>::Backward(
+    const DataType&& input, DataType&& gy, DataType&& g)
+{
+  DataType derivative;
+  //! Compute the first derivative of FlexibleReLU function.
+  derivative = arma::sign(input);
+  derivative.elem(arma::find(derivative < 0.0)) += 1;
```
zoq: We could do something like: ``derivative.elem(find(input > 0) ).ones();`` here, I think it's easier to get. Let me know what you think.

src/mlpack/methods/ann/layer/flexible_relu_impl.hpp
```diff
@@ -0,0 +1,87 @@
...
+template<typename eT>
+void FlexibleReLU<InputDataType, OutputDataType>::Gradient(
+  const arma::Mat<eT>&& input, arma::Mat<eT>&& error,
+  arma::Mat<eT>&& gradient)
+{
+  if (gradient.n_elem == 0)
+  {
+    gradient = arma::zeros<arma::Mat<eT>>(1, 1);
+  }
+  gradient(0) = arma::accu(error) / input.n_cols;
```
zoq: Shouldn't this be ``gradient(0) = arma::sum(error)``?

src/mlpack/methods/ann/layer/flexible_relu_impl.hpp
```diff
@@ -0,0 +1,87 @@
...
+
+template<typename InputDataType, typename OutputDataType>
+template<typename eT>
+void FlexibleReLU<InputDataType, OutputDataType>::Gradient(
+  const arma::Mat<eT>&& input, arma::Mat<eT>&& error,
+  arma::Mat<eT>&& gradient)
+{
+  if (gradient.n_elem == 0)
+  {
+    gradient = arma::zeros<arma::Mat<eT>>(1, 1);
```
zoq: Let's use ``set_size`` here, to make this step somewhat faster.

src/mlpack/methods/ann/layer/flexible_relu_impl.hpp
Manthan-R-Sheth: ```sum()``` would give a row vector (error for each instance in input), but we need a double values for ```alpha``` (trainable parameter) and so i used ```accu()```.
it is ```accu(error)``` as the derivative of the output with respect to ```alpha``` is 1 in case of flexible relu.

zoq: Ah, right, will run some tests later today.

src/mlpack/methods/ann/layer/flexible_relu_impl.hpp
Manthan-R-Sheth: i think i will have to initialize ```derivative``` with ```zeros()``` or ```input``` before using ```derivative.elem(find(input > 0) ).ones();```.
Tell me if i missed anything.

zoq: you are right

src/mlpack/methods/ann/layer/flexible_relu_impl.hpp
Manthan-R-Sheth: sure @zoq thanks.

zoq:
Sorry for the slow response on this one, the main issue is that the output is clipped (zeroed if < 0), so a really small perturbation in the positive/negative direction ends up with the same result and has no effect. An easy solution is to rely on positive weights, which isn't perfect since we don't cover the complete range of the frelu function. So, in addition, we could compare the gradient against a precomputed one with positive/negative weights. Let me know what you think.

Here is the modified test:

```c++
/**
 * Flexible ReLU layer numerically gradient test.
 */
BOOST_AUTO_TEST_CASE(GradientFlexibleReLULayerTest)
{
  // Add function gradient instantiation.
  struct GradientFunction
  {
    GradientFunction()
    {
      input = arma::randu(2, 1);
      target = arma::mat("1");

      model = new FFN<NegativeLogLikelihood<>, RandomInitialization>(
          NegativeLogLikelihood<>(), RandomInitialization(0.1, 0.5));

      model->Predictors() = input;
      model->Responses() = target;
      model->Add<LinearNoBias<> >(2, 5);
      model->Add<FlexibleReLU<> >(0.05);
      model->Add<LogSoftMax<> >();
    }

    ~GradientFunction()
    {
      delete model;
    }

    double Gradient(arma::mat& gradient) const
    {
      arma::mat output;
      double error = model->Evaluate(model->Parameters(), 0, 1);
      model->Gradient(model->Parameters(), 0, gradient, 1);
      return error;
    }

    arma::mat& Parameters() { return model->Parameters(); }

    FFN<NegativeLogLikelihood<>, RandomInitialization>* model;
    arma::mat input, target;
  } function;

}
```

src/mlpack/methods/ann/layer/flexible_relu_impl.hpp
rcurtin: It would be possible to also do `derivative = arma::clamp(arma::sign(input), 0.0, 1.0);` but I don't think this will currently give any acceleration. It may be slightly easier to read though.

Manthan-R-Sheth:
@rcurtin @zoq
I have done the changes. Thanks.

zoq:
@Manthan-R-Sheth thanks for the great contribution.

PR #1343, opened 2018-03-30
Remove deprecated functionality for mlpack 3

rcurtin:
In my view this is the last thing that is necessary before branching off mlpack 3.  I also added any necessary licenses.

Is there anything else that needs to be added to HISTORY.md?  There has been so much development I am sure I missed some things.


HISTORY.md
```diff
@@ -8,6 +8,21 @@
   * Add automatically generated Python bindings.  These have the same interface
     as the command-line programs.
 
+  * Add deep learning infrastructure in src/mlpack/methods/ann/.
```
zoq: and RL in rc/mlpack/methods/reinforcement_learning

HISTORY.md
```diff
@@ -8,6 +8,21 @@
...
+
+  * Add hyperparameter tuning infrastructure and cross-validation infrastructure
+    in src/mlpack/core/cv/ and src/mlpack/core/hpt/.
+
+  * Fix bug in mean shift.
+
+  * Add random forests (see src/mlpack/methods/random_forest).
+
+  * Numerous other bugfixes and testing improvements.
+
```
zoq: We could add randomized and block krylov svd as well.

src/mlpack/methods/pca/pca.hpp
```diff
@@ -40,8 +40,8 @@ class PCAType
    *
    * @param scaleData Whether or not to scale the data.
```
zoq: Should we remove the note from the file ``Note this class will be changed to have the name PCA in mlpack 3.0.0``?

rcurtin:
Thanks for the comments, I've addressed them now.

PR #1344, opened 2018-03-31
Space between compiler flags

zoq:
Fix issue where ``CMAKE_CXX_FLAGS`` is automatically set and doesn't start with a space.


rcurtin:
It looks like on some systems the space is actually there, so I made the change in 522c6f990, which should fix both situations.

PR #1348, opened 2018-03-31
Single output for multiple timestep input for RNN

scrcro:
Add support for multiple timestep -- one output for RNN


scrcro:
Hello 
I made the changes for supporting multiple timestep input and one output. 
The style check failed on a file that I did not change. Hence I think can be ignored.
The Linux build failed, but looks like some env issue. Please suggest.
-scrcro

scrcro:
I am done with the changes. It is ready for review.

src/mlpack/tests/recurrent_network_test.cpp
```diff
@@ -15,6 +15,10 @@
 #include <mlpack/methods/ann/layer/layer.hpp>
 #include <mlpack/methods/ann/rnn.hpp>
 #include <mlpack/core/data/binarize.hpp>
+#include <mlpack/core/optimizers/rmsprop/rmsprop.hpp>
+#include <cmath>
```
rcurtin: This should already be included by `prereqs.hpp` so I think there is no need to include it here.

src/mlpack/methods/ann/rnn_impl.hpp
```diff
@@ -201,14 +201,19 @@ double RNN<OutputLayerType, InitializationRuleType, CustomLayers...>::Evaluate(
   ResetCells();
 
   double performance = 0;
+  size_t respSeq = 0;
```
rcurtin: Do you think it would be useful to add a simple check on the shape of the `responses` if `single` is true or false?  Basically if `single == true`, we should ensure that the responses only have one slice.  And I think we should also be sure to point this out in the documentation.

src/mlpack/tests/recurrent_network_test.cpp
```diff
@@ -1085,4 +1089,140 @@ BOOST_AUTO_TEST_CASE(CustomRecursiveReberGrammarTest)
...
+ * @param gain    The gain on the amplitude
+ * @param freq    The frquency of the sine wave
+ * @param phase   The phase shift if any
+ * @param noisePercent  The percent noise to induce
+ * @param numCycles  How many full size wave cycles required. All the data
+ * points will be fit into these cycles.
+ * @param normalize Whether to normalise the data. This may be required for some
+ * layers like LSTM. Default is true.
+ */
+void GenerateNoisySinRNN(arma::cube& data, arma::cube& labels, size_t rho,
```
rcurtin: I think maybe we should rename this---it's not creating an `RNN`, it's creating data, so I think the method name should reflect that.

src/mlpack/tests/recurrent_network_test.cpp
```diff
@@ -1085,4 +1089,140 @@ BOOST_AUTO_TEST_CASE(CustomRecursiveReberGrammarTest)
...
+  }
+  arma::colvec    x(points);
+  int             i = 0;
+  double          interval = numCycles / freq / points;
+  std::mt19937_64 rng;
+  // initialize the random number generator with time-dependent seed
+  uint64_t timeSeed =
+      std::chrono::high_resolution_clock::now().time_since_epoch().count();
+  std::seed_seq ss{uint32_t(timeSeed & 0xffffffff), uint32_t(timeSeed >> 32)};
+  rng.seed(ss);
```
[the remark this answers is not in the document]
scrcro: Done

src/mlpack/tests/recurrent_network_test.cpp
scrcro: RMSProp is not included somehow, I was getting a compile error. Therefore I had to add it.
Removed `cmath`

src/mlpack/methods/ann/rnn_impl.hpp
scrcro: I need to go through the code for the output to do this. Let me check.

src/mlpack/tests/recurrent_network_test.cpp
scrcro: Well you are right on the data. At the end though I rearrange the data to make two cubes of input data and labels so that they can be used in the RNN training directly. Hence the name. I changed the documentation on the code also. Let me think of a better name..

scrcro:
I also noticed something strange. When I run my test in isolation the mean error I get is lot less. But when I run my test with the other test, it seems to generate error which is much higher. Please see below:
```
hduser_@ubuntu:~/git/mlpack/build$ bin/mlpack_test -t RecurrentNetworkTest
Running 13 test cases...
/home/hduser_/git/mlpack/src/mlpack/tests/recurrent_network_test.cpp(1226): fatal error: in "RecurrentNetworkTest/MultiTimestepTest": critical check err <= 1e-04 has failed [0.0025246568949844257 > 0.0001]

*** 1 failure is detected in the test module "mlpackTest"
hduser_@ubuntu:~/git/mlpack/build$ bin/mlpack_test -t RecurrentNetworkTest/MultiTimestepTest
Running 1 test case...

*** No errors detected
```
Is there a reason that the test should be affected by other tests?

Manthan-R-Sheth:
@scrcro 
I think one possible reason could be the random seed used in the test. Both the executions will have different random seeds used for the generation of the data and labels.

rcurtin:
Right, I think Manthan is correct here.  It would probably be good to run the test many times with different random seeds and ensure that it does not fail.  You may have to adjust the tolerance of the test.

scrcro:
Sorry I should have mentioned. This is happening consistently. Not one off.
Also the order of the mean error is 1e-5 when run in isolation. And it is
1e-02 when run with other tests.

src/mlpack/methods/ann/rnn_impl.hpp
```diff
@@ -201,15 +201,18 @@ double RNN<OutputLayerType, InitializationRuleType, CustomLayers...>::Evaluate(
...
```
zoq: Do you mind to use a more descriptive name for the parameter e.g. ``responsesSeq``?

src/mlpack/tests/recurrent_network_test.cpp
```diff
@@ -1085,4 +1088,135 @@ BOOST_AUTO_TEST_CASE(CustomRecursiveReberGrammarTest)
...
+  }
+  else
+  {
+    points += rho - r + outputSteps;
+  }
+  arma::colvec    x(points);
+  int             i = 0;
+  double          interval = numCycles / freq / points;
+
+  RandomSeed(20);
```
zoq: Not sure I get the idea here, why do we use a static random seed? Ideally this should work with every seed.

src/mlpack/tests/recurrent_network_test.cpp
```diff
@@ -1085,4 +1088,135 @@ BOOST_AUTO_TEST_CASE(CustomRecursiveReberGrammarTest)
...
+ * @param freq    The frquency of the sine wave
+ * @param phase   The phase shift if any
+ * @param noisePercent  The percent noise to induce
+ * @param numCycles  How many full size wave cycles required. All the data
+ * points will be fit into these cycles.
+ * @param normalize Whether to normalise the data. This may be required for some
+ * layers like LSTM. Default is true.
+ */
+void GenerateNoisySinRNN(arma::cube& data, arma::cube& labels, size_t rho,
+                         size_t outputSteps = 1, const int dataPoints = 100,
```
zoq: Can you use a single line for each parameter? See https://github.com/mlpack/mlpack/wiki/DesignGuidelines#method-declarations for more information, I think right now this isn't easy to read.

src/mlpack/tests/recurrent_network_test.cpp
```diff
@@ -1085,4 +1088,135 @@ BOOST_AUTO_TEST_CASE(CustomRecursiveReberGrammarTest)
...
+  int r = dataPoints % rho;
+  if (r == 0)
+  {
+    points += outputSteps;
+  }
+  else
+  {
+    points += rho - r + outputSteps;
+  }
+  arma::colvec    x(points);
```
zoq: Another picky style comment; can you remove the extra spaces?

scrcro:
I am done with the changes as suggested. I also made some adjustments to the tests to make them execute faster and report the errors more accurately.

scrcro:
Please review. The build was successful.

scrcro:
Hello
Do you think this can be merged now? If there are other review comments please let me know.
cheers

src/mlpack/methods/ann/rnn_impl.hpp
```diff
@@ -283,6 +286,14 @@ void RNN<OutputLayerType, InitializationRuleType, CustomLayers...>::Gradient(
     {
       error.zeros();
     }
+    else if (single && seqNum == 0)
+    {
+      outputLayer.Backward(std::move(boost::apply_visitor(
```
zoq: Picky style comment; When wrapping a line, the next line should be tabbed twice from where the previous line began.  See https://github.com/mlpack/mlpack/wiki/DesignGuidelines#line-length-and-wrapping for more information.

src/mlpack/tests/recurrent_network_test.cpp
```diff
@@ -1085,4 +1088,140 @@ BOOST_AUTO_TEST_CASE(CustomRecursiveReberGrammarTest)
...
+  else
+  {
+    points += rho - r + outputSteps;
+  }
+  arma::colvec x(points);
+  int i = 0;
+  double interval = numCycles / freq / points;
+  uint64_t timeSeed =
+      std::chrono::high_resolution_clock::now().time_since_epoch().count();
+  RandomSeed(timeSeed);
```
zoq: I don't think we should force a random seed here, this would avoid the possibility to set a static seed e.g. for debugging.

src/mlpack/tests/recurrent_network_test.cpp
```diff
@@ -1085,4 +1088,140 @@ BOOST_AUTO_TEST_CASE(CustomRecursiveReberGrammarTest)
...
+                         arma::cube& labels,
+                         size_t rho,
+                         size_t outputSteps = 1,
+                         const int dataPoints = 100,
+                         const double gain = 1.0,
+                         const int freq = 10,
+                         const double phase = 0,
+                         const int noisePercent = 20,
+                         const double numCycles = 6.0,
+                         const bool   normalize = true)
```
zoq: Another picky style comment, can you remove the extra spaces?

src/mlpack/tests/recurrent_network_test.cpp
```diff
@@ -1085,4 +1088,140 @@ BOOST_AUTO_TEST_CASE(CustomRecursiveReberGrammarTest)
...
+/**
+ * @brief Generates noisy sine wave and outputs the data and the labels that
+ * can be used directly for training and testing with RNN.
+ *
+ * @param data    The data points as output
+ * @param labels  The expected values as output
+ * @param rho     The size of the sequence of each data point
+ * @param outputSteps How many output steps to consider for every rho inputs
+ * @param dataPoints  The number of generated data points. The actual generated
+ * data points may be more than this to adjust to the outputSteps.
```
zoq: See the comment above.

src/mlpack/tests/recurrent_network_test.cpp
```diff
@@ -1085,4 +1088,140 @@ BOOST_AUTO_TEST_CASE(CustomRecursiveReberGrammarTest)
...
+    val = gain * ::sin(2 * M_PI * freq * t + phase) +
+          (noisePercent * gain / 100 * Random(0.0, 0.1));
+  });
+
+  arma::colvec y = x;
+  if (normalize)
+    y = arma::normalise(x);
+
+  // Now break this into columns of rho size slices.
+  size_t n_columns = y.n_elem / rho;
```
zoq: Can you use camel casing for all names, see https://github.com/mlpack/mlpack/wiki/DesignGuidelines#naming-conventions for more information.

src/mlpack/tests/recurrent_network_test.cpp
scrcro: I am not very familiar with the random number generation in mlpack. Do you want me to remove the `RandomSeed` call altogether?

zoq: Yeah, that would solve the issue.

scrcro:
Hi 
Review comments addressed. Please have a look.

zoq:
Thanks again for a great contribution.

PR #1355, opened 2018-04-05
CF decomposition using RandomizedSVD

haritha1313:


src/mlpack/methods/randomized_svd/randomized_svd.cpp
```diff
@@ -47,6 +47,91 @@ RandomizedSVD::RandomizedSVD(const size_t iteratedPower,
   /* Nothing to do here */
 }
 
+void RandomizedSVD::Apply(const arma::sp_mat& data,
```
zoq: What do you think about using a template parameter here, to reduce the code duplication? I guess something like:

```c++
template<typename DataType>
void RandomizedSVD::Apply(const arma::sp_mat& data, arma::mat& u, arma::mat& v, const size_t rank)
{
...
}

```

could work, and introduce a specialization for ``arma::sp_mat`` and ``arma::mat`` to set ``eps``. I guess, enable_if could also work to set the correct value for ``eps``.

haritha1313: Thanks for the review, although using template parameter was the reason i left it as WIP :D . Couldn't get to it soon enough. The current `eps` value is working for both as of now, but its a good point to try, maybe there could be an RMSE improvement.

src/mlpack/methods/cf/cf_impl.hpp
```diff
@@ -45,11 +45,36 @@ void ApplyFactorizer(FactorizerType& factorizer,
...
+                     const typename std::enable_if_t<!FactorizerTraits<
+                         FactorizerType>::UsesCoordinateList
+                         && FactorizerTraits<
+                         FactorizerType>::UsesRandomizedSVD>* = 0)
+{
+  arma::vec sigma;
+  factorizer.Apply(cleanedData, w, sigma, h, rank);
+
+  // Sigma matrix is multiplied to w.
+  w = w * sigma;
```
Wenhao-H: Should we use diagonal mat here? Something like:
```C++
w = w * arma::diagmat(sigma);
```

haritha1313: Yes, thanks for pointing it out. :)

src/mlpack/methods/randomized_svd/randomized_svd.cpp
haritha1313: @zoq I have a small doubt regarding what you mentioned. Won't using specialization or enable_if for `eps` both lead to the same function repeating with change in only one calculation, i.e similar code duplication as overloading? It might be trivial and I guess I am understanding something wrong. But any help would be great.

zoq: I had something like this in mind:

```c++
template<typename DataType>
void RandomizedSVD::Apply(const arma::mat& data, arma::mat& u, arma::mat& v, const size_t rank, const double eps)
{
...
}

void RandomizedSVD::Apply(const arma::sp_mat& data, arma::mat& u, arma::mat& v, const size_t rank)
{
  Apply(data, u, v, rank, 0);
}

void RandomizedSVD::Apply(const arma::mat& data, arma::mat& u, arma::mat& v, const size_t rank)
{
  Apply(data, u, v, rank, eps);
}
```

Let me know what you think.

haritha1313: Sorry for the late response. That did help understand what you meant. Thank you. :)

zoq: No worries, as you said this is a WIP PR, and I don't think 10 hours counts as late 😃

zoq:
I think it might be worth the effort to refactor the cf class so that it uses a policy based design for the factorization, that way we can implement new methods without modifying the main class; I guess that makes even more sense if we add more methods in the future. Let me know what you think.

The PCA class or the optimizer codebase is a good example.

haritha1313:
@zoq You are right, that would make much more sense. I'm working on this and will complete as soon as I can. I've got exams coming up and hence the delay.

zoq:
No worries, we don't have to rush anything; good luck on your exams .

src/mlpack/methods/randomized_svd/randomized_svd.hpp
```diff
@@ -149,4 +261,20 @@ class RandomizedSVD
...
+namespace cf {
+
+//! Factorizer traits of Regularized SVD.
+template<>
+class FactorizerTraits<mlpack::svd::RandomizedSVD>
+{
+ public:
+  //! Data provided to RandomizedSVD has to be cleaned.
+  static const bool UsesCoordinateList = false;
+  static const bool UsesRandomizedSVD = true;
```
rcurtin: I think that things will change with the refactoring, but if we still have `FactorizerTraits` afterwards, I think maybe we should call this `UsesSVDTypeDecomposition` or something like that.  The real difference here is just that the factorizer returns some `U`, `s`, `V` matrices where s is a vector, and `U * diagmat(s) * V.t()` is equal to the original matrix.

haritha1313:
To do: 
Debug CLI error when loading/saving model.
Any possible restructuring in cf_test.cpp.

zoq:
Looks like you solved the build issues; just some probably random test failures we should take a closer look into, in many cases this just means we have to adjust some tolerance (threshold) or increase the number of iterations.

src/mlpack/methods/cf/cf.hpp
```diff
@@ -73,10 +56,11 @@ struct FactorizerTraits
  * are in a matrix that holds doubles, should hold integer (or size_t) values.
  * The user and item indices are assumed to start at 0.
  *
- * @tparam FactorizerType The type of matrix factorization to use to decompose
+ * @tparam DecompositionPolicy The algorithm to use to decompose
  *     the rating matrix (a W and H matrix).  This must implement the method
  *     Apply(arma::sp_mat& data, size_t rank, arma::mat& W, arma::mat& H).
  */
+template<typename DecompositionPolicy = NMFPolicy>
 class CF
```
zoq: To allow backwards compatibility, we should call this CFType, and provide an alias that uses, ``amf::NMFALSFactorizer``. Something like:

```c++
using CF = CFType<amf::NMFALSFactorizer>;
```

that way, the old interface remains since I can just reuse the ``CF`` class without the template parameter.

Wenhao-H:
Hmm it seems that our work have some overlaps. I am working on CF data normalization at #1397. Basically what it does is to normalize rating before training, and denormalize predicted rating before returning the results. I think the conflicts are easy to resolve so it's not a big for the moment. But as we are working on the same module for the summer, it is better if we discuss on how to collaborate so as to avoid repetitive work. We can discuss over email or a new issue later:))

For `randomized_svd`, it seems that the rank of decomposition does not equal the rank specified in constructor. Would you mind having a look at that? A test can be added to check whether rank of decomposition is correct.

haritha1313:
@Wenhao-H I think this PR is some basic refactoring which could cause some conflicts between our work. But once this part is done, my work will be on another method of performing collaborative filtering and yours on existing CF class i suppose. So I guess in future conflicts would be less. But in case there is anything else, we can always discuss. :)

zoq:
@Wenhao-H if this doesn't make it easier for you, I guess we could delay the changes if @haritha1313 agrees, but if you think it makes things easier, let's merge this one as soon as it's done.

Wenhao-H:
@zoq Everything's fine! This is a nice improvement and it would be better if we can merge this soon so I can resolve the conflicts earlier.

src/mlpack/methods/cf/cf.hpp
```diff
@@ -88,78 +72,90 @@ class CF
...
    * @param numUsersForSimilarity Size of the neighborhood.
    * @param rank Rank parameter for matrix factorization.
+   * @param maxIterations Maximum number of iterations.
+   * @param minResidue Residue required to terminate.
+   * @param mit Whether to terminate only when maxIterations is reached.
    */
-  template<typename FactorizerType = amf::NMFALSFactorizer>
-  CF(const arma::mat& data,
-     FactorizerType factorizer = FactorizerType(),
+  template<typename mat_type>
```
zoq: Please use camel casing for all names; see https://github.com/mlpack/mlpack/wiki/DesignGuidelines#naming-conventions for more information.

src/mlpack/methods/cf/cf.hpp
```diff
@@ -88,78 +72,90 @@ class CF
...
+   * @param mit Whether to terminate only when maxIterations is reached.
    */
-  template<typename FactorizerType = amf::NMFALSFactorizer>
-  CF(const arma::sp_mat& data,
-     FactorizerType factorizer = FactorizerType(),
-     const size_t numUsersForSimilarity = 5,
-     const size_t rank = 0,
-     const typename std::enable_if_t<
-         !FactorizerTraits<FactorizerType>::UsesCoordinateList>* = 0);
+  template<typename mat_type>
```
zoq: See comment above.

src/mlpack/methods/cf/cf.hpp
```diff
@@ -88,78 +72,90 @@ class CF
...
+   * @param data Input dataset; dense matrix (coordinate lists).
+   * @param decomposition Instantiated DecompositionPolicy object.
+   * @param maxIterations Maximum number of iterations.
+   * @param minResidue Residue required to terminate.
+   * @param mit Whether to terminate only when maxIterations is reached.
    */
-  template<typename FactorizerType = amf::NMFALSFactorizer>
   void Train(const arma::mat& data,
-             FactorizerType factorizer = FactorizerType());
+             DecompositionPolicy decomposition = DecompositionPolicy(),
```
zoq: We should use a reference here if possible.

src/mlpack/methods/cf/cf.hpp
```diff
@@ -88,78 +72,90 @@ class CF
...
-   * @param data Sparse matrix data.
-   * @param factorizer Instantiated factorizer.
+   * @param data Input dataset; sparse matrix (user item table).
+   * @param decomposition Instantiated DecompositionPolicy object.
+   * @param maxIterations Maximum number of iterations.
+   * @param minResidue Residue required to terminate.
+   * @param mit Whether to terminate only when maxIterations is reached.
    */
-  template<typename FactorizerType = amf::NMFALSFactorizer>
   void Train(const arma::sp_mat& data,
```
zoq: Perhaps I missed something, but we should at least provide a Train method, that does use the decomposition method from the constructor, let me know what you think.

src/mlpack/methods/cf/cf.hpp
```diff
@@ -210,15 +206,15 @@ class CF
   /**
    * Generates the given number of recommendations for the specified users.
    *
-   * @param numRecs Number of Recommendations
-   * @param recommendations Matrix to save recommendations
-   * @param users Users for which recommendations are to be generated
+   * @param numRecs Number of Recommendations.
```
zoq: Nice catch!

src/mlpack/methods/cf/cf_impl.hpp
```diff
@@ -69,37 +37,59 @@ CF::CF(const arma::mat& data,
...
+            const arma::sp_mat& cleanedData,
+            const size_t rank,
+            arma::mat& w,
+            arma::mat& h,
+            const size_t maxIterations,
+            const double minResidue,
+            const bool mit)
+{
+  decomposition.Apply(data, cleanedData, rank, w,
+                      h, maxIterations, minResidue, mit);
```
zoq: The next line should be tabbed twice from where the previous line began, see https://github.com/mlpack/mlpack/wiki/DesignGuidelines#naming-conventions for more information.

src/mlpack/methods/cf/cf_impl.hpp
```diff
@@ -121,15 +111,18 @@ void CF::Train(const arma::mat& data, FactorizerType factorizer)
   // Decompose the data matrix (which is in coordinate list form) to user and
   // data matrices.
   Timer::Start("cf_factorization");
-  ApplyFactorizer(factorizer, data, cleanedData, this->rank, w, h);
+  decomposition.Apply(data, cleanedData, rank, w,
+                      h, maxIterations, minResidue, mit);
```
zoq: See comment above.

src/mlpack/methods/cf/cf_impl.hpp
```diff
@@ -69,37 +37,59 @@ CF::CF(const arma::mat& data,
...
- * Construct the CF object using an instantiated factorizer.
+ * Construct the CF object using an instantiated decomposition policy.
  */
-template<typename FactorizerType>
-CF::CF(const arma::sp_mat& data,
-       FactorizerType factorizer,
+template<typename DecompositionPolicy>
+template<typename mat_type>
+CF<DecompositionPolicy>::CF(mat_type const& data,
+       const DecompositionPolicy& decomposition,
```
zoq: Do you mind to align the parameter with the first one (``data``)?

src/mlpack/methods/cf/cf.hpp
zoq: You are right, we already know the decomposition policy at construction time.

src/mlpack/methods/cf/cf_main.cpp
```diff
@@ -101,8 +105,8 @@ PARAM_DOUBLE_IN("min_residue", "Residue required to terminate the factorization"
     " (lower values generally mean better fits).", "r", 1e-5);
 
 // Load/save a model.
-PARAM_MODEL_IN(CF, "input_model", "Trained CF model to load.", "m");
-PARAM_MODEL_OUT(CF, "output_model", "Output for trained CF model.", "M");
+PARAM_MODEL_IN(CF<>, "input_model", "Trained CF model to load.", "m");
```
haritha1313: @zoq What do you think would be the right way to mention the DecompositionPolicy for input_model and output_model? Right now they are being set to default policy NMF even for objects with other policy, causing error.

src/mlpack/tests/cf_test.cpp
```diff
@@ -28,7 +35,8 @@ using namespace std;
  * Make sure that correct number of recommendations are generated when query
  * set. Default case.
  */
-BOOST_AUTO_TEST_CASE(CFGetRecommendationsAllUsersTest)
+template<typename DecompositionPolicy>
+void GetRecommendationsAllUsers()
 {
   // Dummy number of recommendations.
   size_t numRecs = 3;
```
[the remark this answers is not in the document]
haritha1313: I'm sorry, but I didn't quite understand what you meant by not having to change anything at this point.

src/mlpack/methods/cf/cf_main.cpp
zoq: I was thinking we could do ssomething similar as done for the pca executable:

https://github.com/mlpack/mlpack/blob/095f784a30742ff64f4616c362475d6c8ec8ea15/src/mlpack/methods/pca/pca_main.cpp#L68-L70

What do you think?

haritha1313: Maybe I am missing something, but PCA doesn't have method to load/save model, and thus doesn't have a PARAM_MODEL_IN/OUT parameter, and this is what I am concerned about in CF.

rcurtin: In other cases where this has come up (it does in `LogisticRegression` for example) I separate the template parameter used for the training of the model.  So, e.g., maybe `CF` has no template parameters but `CF::Train()` does, since the decomposition is only used at training time.

If that doesn't work (and I haven't spent a long time thinking about it, so maybe it doesn't), you might try some wrapper model that can hold all `CF<...>` types that might be used.  See, e.g., `AdaBoostModel` in `src/mlpack/methods/adaboost/adaboost_model.hpp`, then you can just serialize that.

zoq: Good point, we don't have to templatize the complete class, ``Train()`` should be sufficient, what do you think @haritha1313 @Wenhao-H

Wenhao-H: I agree at the moment we don't need to templatize CF class. But for rating normalization that I'm working on, CF needs to be class template because the normalization object needs to be saved as a class member. I will write a class like `AdaBoostModel` to solve this.  So for now, I think templatizing only `CF::Train()` would be enough.

haritha1313: Yes, I have done the necessary edits to templatize only `CF::Train()`. I'll push them soon. @Wenhao-H I have taken a look into how `AdaBoostModel` has been written, so in case your immediate work needs CF class as templatized itself, I can do that too. Just let me know.

Wenhao-H: @haritha1313 The implementation of the model wrapper class (called `CFModel` maybe) is determined by the class template parameter (which would be `NormalizationType` in my case). So I think it would be easier for me to work on that. Thanks for offering the help though:)

zoq: Sounds like a good solution to me.

src/mlpack/methods/cf/cf_impl.hpp
```diff
@@ -148,16 +106,19 @@ void CF::Train(const arma::sp_mat& data,
     this->rank = rankEstimate;
   }
 
+  // Decompose the data matrix (which is in coordinate list form) to user and
+  // data matrices.
   Timer::Start("cf_factorization");
-  factorizer.Apply(cleanedData, this->rank, w, h);
+  decomposition.Apply(data, cleanedData, rank, w,
+                      h, maxIterations, minResidue, mit);
```
zoq: Picky style issue number two, this should be tabbed twice from the start.

src/mlpack/methods/cf/decomposition_policies/CMakeLists.txt
```diff
@@ -0,0 +1,19 @@
+# Define the files we need to compile
+# Anything not in this list will not be compiled into mlpack.
+set(SOURCES
+  batch_svd_method.hpp
+  nmf_method.hpp
+  randomized_svd_method.hpp
+  regularized_svd_method.hpp
+  SVD_complete_method.hpp
```
zoq: Do you mind to use ``svd_complete_method`` and ``svd_incomplete_method.hpp``, to be consistent with the rest of the codebase?

src/mlpack/methods/cf/decomposition_policies/SVD_complete_method.hpp
```diff
@@ -0,0 +1,81 @@
...
+#include <mlpack/methods/amf/amf.hpp>
+#include <mlpack/methods/amf/update_rules/nmf_als.hpp>
+#include <mlpack/methods/amf/termination_policies/max_iteration_termination.hpp>
+#include <mlpack/methods/amf/termination_policies/simple_residue_termination.hpp>
+
+namespace mlpack {
+namespace cf {
+
+/**
+ * Implementation of the SVD complete incremental policy
```
zoq: Missing ``.`` at the end.

src/mlpack/methods/cf/decomposition_policies/SVD_complete_method.hpp
```diff
@@ -0,0 +1,81 @@
...
+   * @param cleanedData item user table in form of sparse matrix.
+   * @param rank Rank parameter for matrix factorization.
+   * @param w First matrix formed after decomposition.
+   * @param h Second matrix formed after decomposition.
+   * @param maxIterations Maximum number of iterations.
+   * @param minResidue Residue required to terminate.
+   * @param mit Whether to terminate only when maxIterations is reached.
+   */
+  template<typename MatType>
+  void Apply(MatType const& /* data */,
```
zoq: See comment above.

src/mlpack/methods/cf/decomposition_policies/SVD_complete_method.hpp
zoq: Do you think you could add a short description about what this policy does, this should be useful for people who like to use the CF code.

src/mlpack/methods/cf/decomposition_policies/randomized_svd_method.hpp
```diff
@@ -0,0 +1,103 @@
...
+             arma::mat& w,
+             arma::mat& h,
+             const size_t maxIterations,
+             const double /* minResidue */,
+             const bool /* mit */)
+  {
+    arma::vec sigma;
+
+    // Do singular value decomposition using the randomized SVD algorithm.
+    svd::RandomizedSVD rsvd(rank+2, maxIterations);
```
zoq: I think this should be ``svd::RandomizedSVD rsvd(ìterationPower, maxIterations)``.

src/mlpack/methods/randomized_svd/randomized_svd.hpp
```diff
@@ -120,6 +137,101 @@ class RandomizedSVD
...
+   * using the randomized SVD.
+   *
+   * @param data Data matrix.
+   * @param u First unitary matrix.
+   * @param v Second unitary matrix.
+   * @param sigma Diagonal matrix of singular values.
+   * @param rank Rank of the approximation.
+   * @param rowMean Centered mean value matrix.
+   */
+  template<typename mat_type>
```
zoq: Another picky style comment, please use camel casing for all names.

src/mlpack/methods/randomized_svd/randomized_svd.hpp
```diff
@@ -120,6 +137,101 @@ class RandomizedSVD
...
+  void Apply(mat_type const& data,
```
zoq: I think we can simplify the class, and only provide two ``Apply`` methods:

```c++
template<typename MatType>
void RandomizedSVD::Apply(const MatType& data,
                          arma::mat& u,
                          arma::vec& s,
                          arma::mat& v,
                          const size_t rank)
{
  // Center the data into a temporary matrix.
  arma::vec rowMean = arma::sum(data, 1) / data.n_cols + eps;
  Apply(data, u, s, v, rank, rowMean);
}


template<typename MatType>
void RandomizedSVD::Apply(const MatType& data,
                          arma::mat& u,
                          arma::vec& s,
                          arma::mat& v,
                          const size_t rank,
                          const MatType& rowMean)
{
  ...
}
```

do you think that works? Also let's put the implementation inside the ``.cpp class``.

src/mlpack/methods/cf/decomposition_policies/SVD_incomplete_method.hpp
```diff
@@ -0,0 +1,81 @@
...
+          amf::SVDIncompleteIncrementalLearning> svdici(iter);
+
+      svdici.Apply(cleanedData, rank, w, h);
+    }
+    else
+    {
+      amf::SimpleResidueTermination srt(minResidue, maxIterations);
+
+      // Do singular value decomposition using incomplete incremental method.
+      amf::SVDIncompleteIncrementalFactorizer<arma::sp_mat> svdici(srt);
```
zoq: Should we use ``MatType`` instead of ``arma::sp_mat`` here?

src/mlpack/methods/randomized_svd/randomized_svd.hpp
zoq: On a second look, I would say you already did that, but it looks like the merge failed.

src/mlpack/methods/cf/decomposition_policies/SVD_incomplete_method.hpp
haritha1313: What I thought was that SVDIncomplete method needs data in cleaned form, and thus accepts the sparse matrix. MatType is template parameter for raw data, which could be of form arma::mat, which won't be useful for this method.

zoq: Agreed, it's not useful, do you think we should add a simple comment here to point that out?

src/mlpack/tests/cf_test.cpp
```diff
@@ -548,5 +669,395 @@ BOOST_AUTO_TEST_CASE(SerializationTest)
   }
 }
 
+/**
+ * Make sure that correct number of recommendations are generated when query
+ * set for randomized SVD.
+ */
+BOOST_AUTO_TEST_CASE(CFGetRecommendationsAllUsersRandSVDTest)
+{
+  GetRecommendationsAllUsers<RandomizedSVDPolicy>();
```
zoq: I guess, if we increase the ``iteratedPower`` and ``maxIterations`` and don't use the default values it might give us better results. In this case we would have to pass an instantization of the policy. Something like:

```c++
RandomizedSVDPolicy policy(10, 10);
```

might work.

haritha1313: I was getting different error log when testing locally and thus wasn't able to debug. The jenkins log made more sense, and it was a silly enough mistake. :) Guess its kind of solved now. I'll push the changes and if this change is still necessary, do let me know.

zoq:
Does ``CFTest/TrainRegSVDTest`` pass on your system?

haritha1313:
CFTest hasn't been running properly on my system. This particular error that jenkins has caught makes it get stuck indefinitely when running locally.

src/mlpack/methods/cf/cf.cpp
```diff
@@ -20,24 +20,24 @@
 namespace mlpack {
 namespace cf {
 
-// Default CF constructor.
-CF::CF(const size_t numUsersForSimilarity,
-       const size_t rank) :
+// Default CFType constructor.
+CFType::CFType(const size_t numUsersForSimilarity,
```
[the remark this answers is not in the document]
Wenhao-H: @zoq Sorry that I missed your comment here. Yes you are right that I need to templatize the class. It's easier for me if we go with `CFType`.

zoq:
Can you identify the test? On my system only the ``TrainRegSVDTest`` failed, but the rest passes.

haritha1313:
Yes, that one fails in jenkins too. It gives error related to sparse matrix assignment, but none has been used in the train function for `RegSVD`.

src/mlpack/tests/cf_test.cpp
```diff
@@ -451,8 +451,8 @@ template<>
 void Train<>(RegSVDPolicy& decomposition)
 {
   // Generate random data.
-  arma::mat randomData(100, 100);
-  randomData.randu();
+  arma::mat randomData = arma::zeros(100, 100);
```
zoq: Haven't thought about it before, but we could also just use ``arma::mat randomData(100, 100, arma::fill::eye);``, it's a little bit shorter.

zoq:
I think this is ready, before we merge this perhaps we could decrease the test time somehow. A simple solution is to see if we can decrease the number of training iterations. I'll take a closer look into the test cases, maybe you have an idea?

zoq:
Yeah, the timings are looking good to me, thanks for looking into the issue.

zoq:
Let me run the cf tests using different random seeds to make sure they are stable; for more information https://github.com/mlpack/mlpack/issues/922.

zoq:
Okay, not a single failure in 300 trials, do you like to make any other changes?

haritha1313:
Thats great! I have gone through it again, and tried running using CLI too. I can't think of anything to change in particular, unless you have any suggestion, of course.

zoq:
Thanks again for the great contribution, this should make things a lot easier in the future.

PR #1356, opened 2018-04-07
Add fastmks test

Prabhat-IIT:
#1152 
Here are the tests I've written for fastmks. As of now some tests are failing strangely and I'm trying to fix that. Any help is appreciated. Moreover, if anyone can suggest more test I'll be more than happy to add it.

### DIFF
--- a/src/mlpack/methods/fastmks/fastmks_main.cpp
+++ b/src/mlpack/methods/fastmks/fastmks_main.cpp
@@ -201,9 +201,9 @@ static void mlpackMain()
       arma::mat queryData = std::move(CLI::GetParam<arma::mat>("query"));
 
       if (CLI::HasParam("reference"))
-      	if (queryData.n_rows != referenceData.n_rows)
-      		Log::Fatal << "Dimension of Reference and Query dataset"
-											"must be equal" << endl;
+        if (queryData.n_rows != referenceData.n_rows)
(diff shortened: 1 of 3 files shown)


Prabhat-IIT:
One thing I've in mind is comparing the `indices` returned by Gaussian kernel and other shift-invariant kernels with `knn`. @rcurtin   can you confirm that both will always return the same result?

Prabhat-IIT:
@zoq I need some help with one thing that while writing tests I couldn't pass `kernels` correctly. Neither setting input param directly like SetInputParam("kernels", "hyptan") works nor it works like SetInputParam("kernels", kernalName) where `string kernelName`="hyptan" for ex. It leads to tests running forever. I've commented out the code as of now.

Prabhat-IIT:
> I dug into this one a little bit and I see there is a bug in the FastMKS program. Basically, we pass the reference set to the model, but then the reference set is deallocated before the model is saved. 
I've reported this bug on irc two times but you were busy that time. Thanks for the patch.I'll do the necessary changes soon!

rcurtin:
Yes, and I saw it each time you posted but had not had a chance to look into it until now.

rcurtin:
Since #1765 is merged, we can close this.  Thanks @Prabhat-IIT for your hard work on the initial PR. :+1:

PR #1360, opened 2018-04-10
Truncated Normal Initialization of Weights

PlantsAndBuildings:
#1228 

[Keras implementation](https://github.com/keras-team/keras/blob/af804d0a5db9a8f20fbb083b48655b2687ce89d9/keras/initializers.py#L122) creates a separate initialization class. I have only added an option in the existing implementation of Gaussian initialization.

Please close this PR if this is not required/too simple.

### DIFF
--- a/src/mlpack/methods/ann/init_rules/gaussian_init.hpp
+++ b/src/mlpack/methods/ann/init_rules/gaussian_init.hpp
@@ -102,6 +102,58 @@ class GaussianInitialization
   double stddev;
 }; // class GaussianInitialization
 
+/**
+ * This class initializes a given weight matrix with values sampled from a
+ * Gaussian Distribution of given mean and variance. It ensures that none of
+ * the sampled values are more than two standard deviations away from the mean.
+ */
+class TruncatedGaussianInitialization
+{
+ public:
+  /**
+   * Initialize the gaussian with given mean and variance.
+   *
+   * @param mean Mean of the gaussian.
+   * @param variance Variance of the gaussian.
+   */
+  TruncatedGaussianInitialization(const double mean = 0,
+                                  const double variance = 1.0) :
+     g(new GaussianInitialization(mean, variance))
+  {
+    // Nothing to do here.
+  }
+
+  /**
+   * Initialize the elements weight matrix using a Gaussian distribution.
+   *
+   * @param W Weight matrix to initialize.
+   * @param rows Number of rows.
+   * @param cols Number of columns.
+   */
+  void Initialize(arma::mat& W,
+                  const size_t rows,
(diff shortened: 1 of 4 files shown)


src/mlpack/methods/ann/init_rules/gaussian_init.hpp
```diff
@@ -49,13 +49,26 @@ class GaussianInitialization
    */
   void Initialize(arma::mat& W,
                   const size_t rows,
-                  const size_t cols)
+                  const size_t cols,
+                  bool truncated = false)
```
zoq: The issue I see here, is that the user can't pass ``truncated = true``, if the initialization rule is used inside the neural network codebase. So, I guess a simple solution would be to write an extra class. Let me know what you think.

PlantsAndBuildings: Yeah, you're right  - I've created a new class `TruncatedGaussianInitialization` now. A couple of points I'd like your thoughts on:

1. Would it be better to have a separate file `truncated_gaussian_init.hpp`, as opposed to placing both `GaussianInitialization` and `TruncatedGaussianInitialization` under the same roof? I don't see any file having multiple class definitions anywhere else in the codebase, so let me know if you think its better to pull `TruncatedGaussianInitialization` out into a new file.

2. I've let the `truncated` flag in the `Initialize` methods of `GaussianInitialization` be as it is; and used a `GaussianInitialization` object in the `TruncatedGaussianInitialization` class. I know its generally better to have loosely coupled class definitions, but I think these two are similar enough for us to get away with this?

rcurtin: Another idea would be to provide the flag as a template parameter, i.e., make a class `GaussianInitializationType<bool>`, then

```
/**
 * documentation for GaussianInitiazliation
 */
typedef GaussianInitializationType<false> GaussianInitialization;

/**
 * documentation for TruncatedGaussianInitialization
 */
typedef GaussianInitializationType<true> TruncatedGaussianInitialization;
```

Maybe that might help reduce the amount of code.  I am not picky though, I think it's fine to do it either way (two separate classes or one with a template parameter and typedefs).  I just thought it might be helpful to suggest it. :)

zoq: Agreed, that is also an option, @PlantsAndBuildings your decision.

zoq:
Looks like, there are some merge issues, let me know if I should fix them.

PlantsAndBuildings:
Sorry for going AWOL, work has been a bit hectic .. I'll fix this later tonight.

PR #1361, opened 2018-04-10
Header file : Adam  update to vanilla update

luffy1996:
Updated Header file in the test file. The optimizer used in the test is vanilla update. This will help the users to understand the code better.

### DIFF
--- a/src/mlpack/tests/async_learning_test.cpp
+++ b/src/mlpack/tests/async_learning_test.cpp
@@ -16,7 +16,7 @@
 #include <mlpack/methods/ann/layer/layer.hpp>
 #include <mlpack/methods/reinforcement_learning/async_learning.hpp>
 #include <mlpack/methods/reinforcement_learning/environment/cart_pole.hpp>
-#include <mlpack/core/optimizers/adam/adam_update.hpp>
+#include <mlpack/core/optimizers/sgd/update_policies/vanilla_update.hpp>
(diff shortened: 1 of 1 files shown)


src/mlpack/tests/async_learning_test.cpp
```diff
@@ -1,6 +1,7 @@
 /**
  * @file async_learning_test.hpp
  * @author Shangtong Zhang
+ * @author Rohan Raj
```
zoq: I think we should only add the ``@author`` tag if someone made a reasonable contribution to the file; every contribution is very much appreciated, but I think it would be fair if we add names according to the rule I have mentioned. Let me know what you think.

src/mlpack/tests/async_learning_test.cpp
```diff
@@ -16,7 +17,7 @@
 #include <mlpack/methods/ann/layer/layer.hpp>
 #include <mlpack/methods/reinforcement_learning/async_learning.hpp>
 #include <mlpack/methods/reinforcement_learning/environment/cart_pole.hpp>
-#include <mlpack/core/optimizers/adam/adam_update.hpp>
+#include <mlpack/core/optimizers/sgd/update_policies/vanilla_update.hpp>
```
zoq: Agreed, this improves the readability and should help to avoid confusions.

zoq:
This should avoid some confusions, thanks for the contribution.

PR #1294, opened 2018-03-03
Add Loss Functions Folder, KL Divergence Loss

dakshitagrawal97:
This PR adds a `loss_functions` folder inside the `ann` folder.  The KL Divergence loss has been implemented and put into this folder.

Tests for KL Divergence loss function have been added as well.

PR #1365, opened 2018-04-13
Add Loss Functions Folder, KL Divergence Loss

sourabhvarshney111:
This is towards completion of #1294. I have moved some of the loss functions in the loss functions folder. I have tried to refactor the code as much as possible.


sourabhvarshney111:
@zoq Async learning test has used mean squared error as its layer. I have a little idea about reinforcement learning. Can you help me deal with this?

zoq:
It's still fine, to move the layer/loss function, we just have to adjust the include path.

sourabhvarshney111:
I think I got it. I just have to make a loss_function.hpp and loss_function_types.hpp in loss_functions folder. Right? And make suitable changes.

zoq:
I'm not sure I get the issue with the ``MeanSquaredError`` method.

You are referring to:

```c++
  FFN<MeanSquaredError<>, GaussianInitialization> model(MeanSquaredError<>(),
```

right? If that is the case, I don't see the issue if we move the ``MeanSquaredError`` code into another folder.

sourabhvarshney111:
@zoq  I was thinking whether we should make a loss_functions.hpp like layer.hpp so that we can include it directly without needing to include the files separately. Am I right?

zoq:
I see, I guess in this case it's fine to include a specific layer, since in most cases only one loss function is used. A consolidated header file might inccrease the build time.

sourabhvarshney111:
@zoq I got your point. I have made the changes and waiting for the tests to complete.

src/mlpack/methods/ann/loss_functions/KLDivergence.hpp
```diff
@@ -0,0 +1,112 @@
+/**
+ * @file KLDivergence.hpp
+ * @author Dakshit Agrawal
+ *
+ * Definition of the KL Divergence error function.
```
zoq: Might be a good idea to call this Kullback–Leibler divergence.

src/mlpack/methods/ann/loss_functions/KLDivergence.hpp
```diff
@@ -0,0 +1,112 @@
...
+#ifndef MLPACK_METHODS_ANN_LOSS_FUNCTION_KLDIVERGENCE_HPP
+#define MLPACK_METHODS_ANN_LOSS_FUNCTION_KLDIVERGENCE_HPP
+
+#include <mlpack/prereqs.hpp>
+
+namespace mlpack {
+namespace ann /** Artificial Neural Network. */ {
+
+/**
+ * The KL divergence error function measures the network's
```
zoq: I think this one could be improved, e.g. The Kullback–Leibler divergence is often used  for continuous distributions (direct regression). Perhaps adding a reference  (paper) is useful as well.

src/mlpack/methods/ann/loss_functions/KLDivergence.hpp
```diff
@@ -0,0 +1,112 @@
...
+>
+class KLDivergence
+{
+ public:
+  /**
+   * Create the KLDivergence object with the specified parameters.
+   *
+   * @param takeMean Boolean variable to specify whether to take mean or not.
+   */
+  KLDivergence(const bool takeMean = false);
```
zoq: What do you think, should we just call it ``mean``?

src/mlpack/methods/ann/loss_functions/KLDivergence_impl.hpp
```diff
@@ -0,0 +1,73 @@
...
+  // Nothing to do here.
+}
+
+template<typename InputDataType, typename OutputDataType>
+template<typename InputType, typename TargetType>
+double KLDivergence<InputDataType, OutputDataType>::Forward(
+    const InputType&& input, const TargetType&& target)
+{
+  // Input and Target should be of same size.
+  double loss;
```
zoq: No need to define the parameter here, we can return right away.

src/mlpack/methods/ann/loss_functions/KLDivergence_impl.hpp
```diff
@@ -0,0 +1,73 @@
...
+
+  if (takeMean) {
```
zoq: Can you put the ``{`` on a new line.

src/mlpack/methods/ann/loss_functions/KLDivergence_impl.hpp
```diff
@@ -0,0 +1,73 @@
...
+}
+
+template<typename InputDataType, typename OutputDataType>
+template<typename InputType, typename TargetType, typename OutputType>
+void KLDivergence<InputDataType, OutputDataType>::Backward(
+    const InputType&& input,
+    const TargetType&& target,
+    OutputType&& output)
+{
+  if (takeMean) {
```
zoq: See the comment above.

src/mlpack/methods/ann/loss_functions/KLDivergence_impl.hpp
sourabhvarshney111: Done :)

src/mlpack/methods/ann/loss_functions/KLDivergence.hpp
sourabhvarshney111: I have changed the file name. Should I change class name too?

zoq: I think the class and filename is just fine, it's a pretty common alias, but we should mention at least once Kullback–Leibler divergence in the comments.

src/mlpack/methods/ann/loss_functions/KLDivergence.hpp
sourabhvarshney111: I think it would not specify it's real purpose. It is meant for specifying whether we have to take the mean or use the sum of individual loss.

zoq: I see the point, thanks for the clarification.

src/mlpack/methods/ann/loss_functions/KLDivergence.hpp
sourabhvarshney111: Should this reference be like:
@code
  @article{
    title={On Information and Sufficiency},
    author={S. Kullback, R.A. Leibler},
    url = {https://projecteuclid.org/euclid.aoms/1177729694},
    journal = {The Annals of Mathematical Statistics},
    year={1951}
  }
 @endcode

sourabhvarshney111:
@zoq I have made the requested changes.

src/mlpack/methods/ann/loss_functions/KLDivergence_impl.hpp
```diff
@@ -0,0 +1,74 @@
...
+    const TargetType&& target,
+    OutputType&& output)
+{
+  if (takeMean)
+  {
+    output = arma::mean(arma::mean(arma::log(input) - arma::log(target) + 1));
+  }
+  else
+  {
+    output = arma::accu(arma::log(input) - arma::log(target) + 1);
```
zoq: I'm not sure this is correct, I thought it should be ``log(input/target)`` maybe I missed something?

zoq: Ah, sorry I was talking about mean(log(input/target)), but I realized, the NLL, does also use accu, so I think we should stick to that metric.

sourabhvarshney111: So, is this ready for merge?

src/mlpack/methods/ann/loss_functions/CMakeLists.txt
```diff
@@ -0,0 +1,21 @@
+# Define the files we need to compile
+# Anything not in this list will not be compiled into mlpack.
+set(SOURCES
+  cross_entropy_error.hpp
+  cross_entropy_error_impl.hpp
+  KLDivergence.hpp
```
rcurtin: This is a minor comment, but could we rename this `kl_divergence.hpp` during the merge (or now) to match the other filenames in mlpack?

PR #1366, opened 2018-04-14
Ann input/output size (WIP)

### LINKED ISSUE #1267: ANN Input/Output Size
Currently, if we build a network we have to specify the input and output size for each layer, ideally a user should only specify the input/output size for the first layer, and only specify the output size for the upcoming layer. Here is a simple example:

```c++
model.Add<Linear<> >(inputSizeA, outputSizeA);
model.Add<SigmoidLayer<> >();
model.Add<Linear<> >(inputSizeB, outputSizeB);
model.Add<LogSoftMax<> >();
```

which should transforms to this:

```c++
model.Add<Linear<> >(inputSize, outputSize);
model.Add<SigmoidLayer<> >();
model.Add<Linear<> >(layerSize);
model.Add<LogSoftMax<> >();
```

There is something similar already in place to set the input and output width for the convolution operator:

https://github.com/mlpack/mlpack/blob/af3c00b164a9bc14b73e74af2fadc016ab7e9c27/src/mlpack/methods/ann/ffn_impl.hpp#L332-L369

What this code does is to read and set the input/output width in the first forward pass. So once we know the output size of the previous layer we know the input size of the current layer and so on.

A first step would be to look at the https://github.com/mlpack/mlpack/blob/master/src/mlpack/methods/ann/visitor/output_height_visitor.hpp and https://github.com/mlpack/mlpack/blob/master/src/mlpack/methods/ann/visitor/set_input_height_visitor_impl.hpp, which should be helpful to implement another visitor for the input/output size.

akhandait:
Resolves #1267.
Currently only works for Linear layer, if the approach is approved I will do it for all the other layers.
Sorry for the commit messages.

This can be done now:
```cpp
model.Add<Linear<> >(inputSize, outputSize);
model.Add<SigmoidLayer<> >();
model.Add<Linear<> >(layerSize);
model.Add<LogSoftMax<> >();
```


zoq:
Thanks for looking into the issue; this looks like a good start to me, would be great to cover the rest as well.

rcurtin:
Nice support, I think it is a good improvement.  What happens if I do the following?

```
model.Add<Linear<> >(outputSize);
model.Add<SigmoidLayer<> >();
model.Add<Linear<> >(layerSize);
model.Add<LogSoftMax<> >();
```

akhandait:
Sorry for the delay on this, I will try to continue on this soon. 
@rcurtin Currently it won't support it, but I think we can get that to work if in the `SetInputSize` function, we consider the size of the input data and set the `inSize` of the first layer equal to it.
The approach I used in this is that we call the `SetInputSize` before the first forward pass and initialize all the parameters. So, the input data will be given by then and we can easily do this.

akhandait:
I have rebased it on master and combined the two commits. I will resume working on this and get this merged.

akhandait:
@zoq Another definition for `ResetParameters()` wasn't needed. `ANNLayerTest`, `FeedForwardNetworkTest`, `LossFunctionsTest` pass so it's backward compatible. Now, we just need to make it work with all types of layers.

zoq:
Can you implement a simple test that uses the extended FFN + Linear layer class?

akhandait:
I will add it in a couple of days.

zoq:
Sounds good, thanks!

src/mlpack/methods/ann/ffn_impl.hpp
```diff
@@ -498,6 +535,7 @@ void FFN<OutputLayerType, InitializationRuleType, CustomLayers...>::serialize(
     Archive& ar, const unsigned int /* version */)
 {
   ar & BOOST_SERIALIZATION_NVP(parameter);
+  ar & BOOST_SERIALIZATION_NVP(size);
```
rcurtin: Probably we should do some special handling here since `size` is a newly serialized parameter, like this... uncomment `version`, then do

```
if (version > 1)
  ar & BOOST_SERIALIZATION_NVP(size);
else if (Archive::is_loading::value)
  size = 0; // Set size to a default value that will be inferred later.
```

then below the class definition, change the line `BOOST_STATIC_CONSTANT(int, value = 1)` to `BOOST_STATIC_CONSTANT(int, value = 2)`.

This will allow older saved models to be loaded even though they were saved without the `size` parameter.  (I think that 0 is a decent default value; maybe if there is a better way to figure it out you can do that.)

src/mlpack/methods/ann/init_rules/const_init.hpp
```diff
@@ -28,7 +28,7 @@ class ConstInitialization
   /**
    *  Create the ConstantInitialization object.
    */
-  ConstInitialization(const double initVal) : initVal(initVal)
+  ConstInitialization(const double initVal = 0.5) : initVal(initVal)
```
rcurtin: Should we still make this change?  It looks unrelated.

src/mlpack/methods/ann/layer/convolution.hpp
```diff
@@ -143,7 +143,7 @@ class Convolution
 
   //! Get the input width.
   size_t const& InputWidth() const { return inputWidth; }
-  //! Modify input the width.
+  //! Modify the input width.
```
rcurtin: Nice catch! :rocket:

src/mlpack/methods/ann/layer/linear.hpp
```diff
@@ -47,6 +47,13 @@ class Linear
    */
   Linear(const size_t inSize, const size_t outSize);
 
+  /**
+   * Create the Linear layer object using the specified layer size.
```
rcurtin: It might be worth adding `the input layer size will automatically be inferred` to this comment, so that users understand the differences between this and the previous overload. :)

src/mlpack/tests/ann_layer_test.cpp
```diff
@@ -424,6 +424,34 @@ BOOST_AUTO_TEST_CASE(GradientLinearLayerTest)
   BOOST_REQUIRE_LE(CheckGradient(function), 1e-4);
 }
 
+/**
+ * Simple linear layer with only output size test.
+ */
+BOOST_AUTO_TEST_CASE(SimpleLinearLayerOutputSizeTest)
+{
+  FFN<NegativeLogLikelihood<>, ConstInitialization> modelA, modelB;
```
rcurtin: Ah, I guess this is why you added the default parameter to `ConstInitialization`.  I guess, without setting that, you could do this instead:

```
FFN<NegativeLogLikelihood<>, ConstInitialization> modelA(NegativeLogLikelihood<>(), ConstInitialization(0.5));
FFN<NegativeLogLikelihood<>, ConstInitialization> modelB(NegativeLogLikelihood<>(), ConstInitialization(0.5));
```
```

**Comment by akhandait** [comment]
@rcurtin We should definitely get this merged.
> once we merge this we have the option of opening a good beginner issue where people can add this support to other layers and get more familiar with the neural network code in the codebase.

That sounds like a great idea.
Thanks for the review. I will get back in a couple of days.

**Comment by rcurtin** [comment]
Sounds good.  Sorry again the review took over a year :-O

**Comment by akhandait** [comment]
I'm a little confused on this one. The `SimpleLinearLayerOutputSizeTest` test is passing on some checks and failing on some. It's passing on my system. Can someone tell me what I'm missing here?
In the checks where it's failing, the problem seems to be the dimensions of the weight matrix in matrix multiplication.

**Comment by rcurtin** [comment]
@akhandait looks to me like the reason only some builds are failing is that it's those builds that are compiled with debugging symbols fail.  This is because if we compile with `-DDEBUG=OFF -DPROFILE=OFF` (the default), Armadillo bounds checks aren't applied.  So my guess is although the tests are passing on your system (and some of the builds), there may still be an invalid memory access.

I think maybe a quick way to solve it would be to compile locally with `-DDEBUG=ON`, and then use gdb or another debugger to find out where Armadillo is throwing an exception, and go from there.  Usually when I encounter this it's because of something like... I expected the `mat` to have `N` rows and `1` column but it actually had `1` row and `N` columns... or something like that.

Hope this helps, let me know if I can clarify any further. :)

File: src/mlpack/methods/ann/visitor/output_size_visitor.hpp [hunk 6]
```diff
@@ -0,0 +1,75 @@
...
+  size_t operator()(LayerType* layer) const;
+
+  size_t operator()(MoreTypes layer) const;
+
+ private:
+  //! Return 0 if the module doesn't implement the InputSize() or Model()
+  //! function.
+  template<typename T>
+  typename std::enable_if<
+      !HasInputSize<T, size_t&(T::*)()>::value &&
```
akhandait: I'm pretty sure that the builds are failing due to an incorrect function pointer here and the next four functions. The `Linear()` object has an `InputSize()` function which means that the second function should be called here. However, the first one is getting called and that's probably why the test is failing. I had written these for different function types when #1985 wasn't being worked on. Can someone help me with the correct function pointer here. I tried `size_t(T::*)()` but it's not working.

rcurtin: Could this be because the method is `const`?  Maybe try this signature: `size_t(T::*)() const`... I think that one will work; not sure though.

rcurtin: Hey @akhandait, have you had a chance to try that fix yet?  It would be great to get this merged, and it's so close. :)

akhandait: Hey, I'm really sorry about taking so long. I have been very busy with college work. I had tried what you suggested but it didn't quite work. I'll try it again tomorrow night and get back. This PR has been around for too long now.

rcurtin: No worries!  I know how busy things get.  If you don't get to it, it's okay, someone else can handle it too; just let us know if you want a hand. :)

akhandait: Thanks for understanding! I'll give it one more try and then let you know if I need any help. :)

akhandait: It was a very simple fix, should have figured it out a month ago. However, I had to remove the const from `InputSize()` and `OutputSize()` because I am changing the values through the visitor.

akhandait:
I think the failing checks are unrelated and Appveyor is going on forever.

src/mlpack/methods/ann/ffn_impl.hpp
rcurtin: Nice, looks better now, but I think we can remove the line this commented is connected to, otherwise `size` will be doubly serialized:

```
ar & BOOST_SERIALIZATION_NVP(size);
```

I can do that during merge if you like.

src/mlpack/methods/ann/visitor/output_size_visitor.hpp
rcurtin: Sounds good---awesome to see it worked out. :)

zoq:
Looks like one job is currently running and the other one is queued, but I'm not too worried about the Appveyor build, it's basically there to build the msi installer, which at some point is hopefully integrated into the azure build pipeline.

src/mlpack/methods/ann/visitor/output_size_visitor_impl.hpp
```diff
@@ -0,0 +1,99 @@
...
+OutputSizeVisitor::LayerOutputSize(T* layer) const
+{
+  for (size_t i = 0; i < layer->Model().size(); ++i)
+  {
+    size_t outSize = boost::apply_visitor(OutputSizeVisitor(),
+        layer->Model()[layer->Model().size() - 1 - i]);
+
+    if (outSize != 0)
+    {
+      return outSize;
```
zoq: I could be wrong but I think we want the last layer that is nonzero.

zoq:
Agreed, great work, makes it a lot easier to create a new network.

src/mlpack/methods/ann/visitor/output_size_visitor_impl.hpp
akhandait: I am indexing by `layer->Model().size() - 1 - i` in line 60. So, I think that's correct. Even I am a bit confused about what exactly I had written then.

akhandait:
> Awesome work @akhandait, to me it looks good. Just two more comments from my side:
> 
> * Should we change `InputSize()` and `OutputSize()` to be non-`const` for all layer types?
> * Should we make the same changes that we did to `FNN` to `RNN`?
> 
> Both those things are pretty boilerplate refactorings, so if you want, I'm happy to handle them.

It would be very helpful if you could make those changes. Sorry about that.

zoq:
@akhandait will push the adjustments to your fork, hope that's fine with you.

akhandait:
@zoq Of course, no issues. Thanks for the help.

zoq:
Turns out there are some layers that can't be supported without changing the interface, without breaking reverse compatibility. One example is the `LSTM` layer:

https://github.com/mlpack/mlpack/blob/b8db9ba0d97f361f211ec82607cf8dfd9d8361f2/src/mlpack/methods/ann/layer/lstm.hpp#L75-L77

(If I remove `inSize` the method call is ambiguous.)

Something similar applies to the Convolution layer, but as a workaround I added another  constructor that is similar to keras and uses `arma::vec`.

```c++
  AtrousConvolution(const size_t filter,
                    const arma::vec& kernelSize,
                    const arma::vec& strides = arma::vec({1, 1}),
                    const arma::vec& dilation = arma::vec({1, 1}),
                    const std::string& padding = "none");
```
I think we can remove the input size entirely with the current approach, if we break reverse compatibility. Personally thinking, I rather support all layers than only a subset. Meaning that we would have to delay this once more. But @akhandait @rcurtin let me know what you think.

rcurtin:
@zoq hmm, yeah, a bit tricky.  I guess it depends on whether we consider the LSTM layer the "internal" or "external" API (that's up to you, in my opinion).  If it's internal, then according to our policy we're allowed to do reverse-incompatible changes for the 3.3.0 release; otherwise we can wait until 4.0.0.

I thought of a way that we can avoid reverse-incompatibility; we can provide the constructor

```
LSTM(const size_t outSize);
```

and for this constructor, in the documentation, we could write that if the user wants to set `rho`, they should do so with `Rho()`, or use the existing constructor that also takes `inSize`, and set `inSize` to `0` (or some other value that signifies that it should be inferred).

What do you think?

zoq:
Hm, I guess we could see it as an internal API, let's see how far I can get this before the next release.

rcurtin:

jeffin143:
> Hm, I guess we could see it as an internal API, let's see how far I can get this before the next release.

I am totally in for this, Let me know @zoq , How are you paving this further and if you need any help in testing this. 

I will also take a look at it, I guess this really a great support

zoq:
Have to fix an issue with the GRU layer, will do that today, and will look into the GAN issues, but after that this should be ready.

src/mlpack/methods/ann/layer/atrous_convolution.hpp
```diff
@@ -85,7 +87,24 @@ class AtrousConvolution
...
+   * input maps will be inferred.
+   *
+   * @param outSize The number of output filter.
+   * @param kernelSize The height and width of the convolution window;
+   * @param strides The height and width of the stride.
+   * @param dilation The space between the cells of filters.
+   * @param padding The type of padding ('valid' or 'same', 'none').
+   *     Defaults to 'none'.
+   */
+  AtrousConvolution(const size_t filter,
```
rcurtin:
```suggestion
  AtrousConvolution(const size_t outSize,
```
(The parameter name doesn't seem to match the documentation; maybe `filter` or `filters` is a better name, I'm not sure.)

src/mlpack/methods/ann/layer/convolution.hpp
```diff
@@ -76,7 +77,23 @@ class Convolution
...
+   * filter size, stride and padding parameter. The number of input maps will
+   * be inferred.
+   *
+   * @param outSize The number of output filter.
+   * @param kernelSize The height and width of the convolution window;
+   * @param strides The height and width of the stride.
+   * @param padding The type of padding ('valid', 'same', 'none').
+   *     Defaults to 'none'.
+   */
+  Convolution(const size_t filter,
```
rcurtin:
```suggestion
  Convolution(const size_t outSize,
```

Same issue here, whichever way you want to solve it is fine with me. :+1:

src/mlpack/methods/ann/layer/linear_impl.hpp
```diff
@@ -41,13 +41,27 @@ Linear<InputDataType, OutputDataType, RegularizerType>::Linear(
...
 template<typename InputDataType, typename OutputDataType,
     typename RegularizerType>
 void Linear<InputDataType, OutputDataType, RegularizerType>::Reset()
 {
+  weights.set_size(outSize * inSize + outSize, 1);
+
   weight = arma::mat(weights.memptr(), outSize, inSize, false, false);
   bias = arma::mat(weights.memptr() + weight.n_elem,
-      outSize, 1, false, false);
+   outSize, 1, false, false);
```
rcurtin: I can't seem to get Github suggestions to do it, but I think that the `bias` definition can all fit on one line. :+1:

src/mlpack/methods/ann/layer/max_pooling.hpp
```diff
@@ -56,7 +56,7 @@ class MaxPooling
   MaxPooling();
 
   /**
-   * Create the MaxPooling object using the specified number of units.
+   * Create the MaxPooling object using the specified parameter.
```
rcurtin:
```suggestion
   * Create the MaxPooling object using the specified parameters.
```
Not sure, maybe this is an improvement? :)

src/mlpack/methods/ann/layer/mean_pooling.hpp
```diff
@@ -37,7 +37,7 @@ class MeanPooling
   MeanPooling();
 
   /**
-   * Create the MeanPooling object using the specified number of units.
+   * Create the MeanPooling object using the specified parameter.
```
rcurtin:
```suggestion
   * Create the MeanPooling object using the specified parameters.
```
(Same as last suggestion; up to you if you want to take it.)

src/mlpack/methods/ann/visitor/output_size_visitor_impl.hpp
rcurtin: So just to be clear, our requirement is now that if a layer does not implement `InputSize()` or `OutputSize()`, we expect that the output is the same size as whatever input it receives?  (That seems fine to me, I just want to check.)

src/mlpack/methods/ann/ffn.hpp
```diff
@@ -441,6 +448,9 @@ class FFN
   //! parameter.
   InitializationRuleType initializeRule;
 
+  //! The input size.
+  size_t size;
+
   //! The input width.
   size_t width;
```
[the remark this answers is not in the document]
zoq: Could be part of the PR.

zoq:
> Looking good to me. My guess is that the GAN tests and `BRNNReberGrammarTest` are failing because they need the same (or similar) modifications as `ffn.hpp` and `rnn.hpp`---but that _should_ be pretty simple.
> 
> Based on my understanding I think that we need `InputSize()` and `OutputSize()` implemented for any layer that changes the size of its input; correct me if I'm wrong on that. I made a list of layers that I didn't see in this PR:
> 
>     * BilinearInterpolation
> 
>     * Concatenate
> 
>     * Concat
> 
>     * ConcatPerformance
> 
>     * Constant
> 
>     * CReLU
> 
>     * Glimpse
> 
>     * GRU
> 
>     * Highway
> 
>     * Join (should this layer be removed?  I don't think it works right with batch size > 1)
> 
>     * LayerNorm ?
> 
>     * Lookup
> 
>     * MiniBatchDiscrimination
> 
>     * Padding
> 
>     * RecurrentAttention
> 
>     * Select
> 
>     * Subview
> 
>     * VirtualBatchNorm ?
> 
>     * WeightNorm ?
> 
> 

I think layers that don't have a number of input units don't have to be modified. Layers, that don't don't have a number of input units don't, but expose the Model() function, are fine as well because the InputSizeVistor checks for Model(), to set the input size of wrapped layers. That said:

Concatenate
Concat
ConcatPerformance
Constant
CReLU
Lookup
Select
VirtualBatchNorm
WeightNorm

should be fine.

> I'm not 100% sure on all of those. But it seems like those should be transitioned too. (If that makes this infeasible for 3.3.0, no worries from my end. I'm happy to help though.)

Actually, I'm not 100% sure this should/could go into 3.3.0, ideally, I like to make sure all of that works as expected, I added a test for each of the layers I modified so far, however not all of them are tested in combination with each other. So maybe this should go into the next release, so we don't delay the process any further?

> Also, should we mark the old constructors that took both input and output size as `mlpack_deprecated` and adapt the tests to use the new constructors, then remove those old constructors in mlpack 4.0.0?

Yes, good idea.

src/mlpack/methods/ann/visitor/output_size_visitor_impl.hpp
zoq: Yes, if the layer does not implement `OutputSize()` the input size for the next layer stays the same, one example is the `Dropout` layer which just sits on top of any layer and doesn't change the output size.

rcurtin:
After some discussion, I think that it's best to push this off to a future milestone.  Rushing this big of a change into a release is probably not the best idea. :smile:

rcurtin:
> I think layers that don't have a number of input units don't have to be modified. Layers, that don't don't have a number of input units don't, but expose the Model() function, are fine as well because the InputSizeVistor checks for Model(), to set the input size of wrapped layers. That said:
> 
> Concatenate
> Concat
> ConcatPerformance
> Constant
> CReLU
> Lookup
> Select
> VirtualBatchNorm
> WeightNorm
> 
> should be fine.

Sounds good.  Do you think that during the PR we can document somewhere where setting `InputSize()` is necessary and where it isn't?  I'm not sure I 100% understand it. :)

jeffin143:
ping @zoq , Any updated on this, I guess this one would be a good improvement, we can adapt one layer and then open issue for other layers and let community take on from there

PR #1367, opened 2018-04-14
NMF: allow user to specify one of initial W/H matrices

wenhaoh2:
This PR is follow-up of the discussion in #1261. It adds support to allow user to specify one of initial H/W matrices, and to randomly initialize the other one.

### DIFF
--- a/src/mlpack/methods/amf/init_rules/given_init.hpp
+++ b/src/mlpack/methods/amf/init_rules/given_init.hpp
@@ -28,25 +28,74 @@ class GivenInitialization
 {
  public:
   // Empty constructor required for the InitializeRule template.
-  GivenInitialization() { }
+  GivenInitialization() : wIsGiven(false), hIsGiven(false) { }
 
   // Initialize the GivenInitialization object with the given matrices.
-  GivenInitialization(const arma::mat& w, const arma::mat& h) : w(w), h(h) { }
+  GivenInitialization(const arma::mat& w, const arma::mat& h) :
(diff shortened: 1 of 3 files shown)


Wenhao-H:
@zoq Thanks for the suggestion:) Actually there is a small difference between this implementation and what you meant. By this implementation, one of `W/H` matrix will be initialized randomly each time `Initialize()` is called. So if the same `GivenInitialization` instance is used multiple times, the randomly initialized matrices are different.
And to ensure I correctly understand, by combination of rules, do you mean that for example, `W` is initialized with existing `GivenInitialization` and `H` is initialized with existing `RandomInitialization`?

zoq:
You are right, I thought about ``GivenInitialization`` as a one time initialization rule.

About the merge idea, you are right, my only concern is, right now each rule will initialize W and H,  maybe the performance in this case is negligible.

Wenhao-H:
@zoq that's a good idea. What I have in mind is to use a class which takes two Initialization Rules as template parameters, and initialize `W` with the first and `H` with the second.
```C++
template<typename WInitializationRuleType, typename HInitializationRuleType>
class MergeInitializaition
{
  // ...
};
```
I think the performance overhead is negligible, compared to the time used to compute the iterations.
But my only concern is that this might not be a common usage.

Wenhao-H:
I'm sorry for the late reply because I have been writing my bachelor thesis. I will refactor the code to  the `MergeInitialization` idea once I finish the thesis, which may take a week or so:)

zoq:
Sounds good, hopefully you have some fun to finish the last part of your thesis.

PR #1368, opened 2018-04-15
Added Continuous Mountain Car to Reinforcement Learning Environment

luffy1996:
This is an implementation of continuous mountain car game available on Openai Gym. 
Reference : https://gym.openai.com/envs/MountainCarContinuous-v0/

It is difficult to write a test file as reinforcement learning module doesn't support actor-critic methods presently . Thus for the coming week, I will be working to add DDPG support to mlpack. This algorithm can be then used to test the Continous Car environment. 

DDPG Paper : https://arxiv.org/abs/1509.02971

### DIFF
--- a/src/mlpack/methods/reinforcement_learning/environment/continuous_mountain_car.hpp
+++ b/src/mlpack/methods/reinforcement_learning/environment/continuous_mountain_car.hpp
@@ -77,7 +77,7 @@ class ContinuousMountainCar
    */
   struct Action
   {
-    double force[1];
+    double action[1];
     // Storing degree of freedom
     const int size = 1;
   };
@@ -114,7 +114,7 @@ class ContinuousMountainCar
                 State& nextState) const
   {
     // Calculate acceleration.
-    double force = std::min(std::max(action.force[0], -1.0), 1.0);
+    double force = std::min(std::max(action.action[0], -1.0), 1.0);
     nextState.Velocity() = state.Velocity() + 0.001 * force - 0.0025 *
         std::cos(3 * state.Position());
     nextState.Velocity() = std::min(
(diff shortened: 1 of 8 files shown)


zoq:
Thanks, from my side I think this is still valuable without DDPG. So, I'll take a closer look at the code and make comments if necessary, so that we can merge this one.

luffy1996:
@zoq. Please let me know if any changes are required. Thanks :)

luffy1996:
@zoq : for style error, merging #1372 will solve the problem.

src/mlpack/methods/reinforcement_learning/environment/continuous_mountain_car.hpp
```diff
@@ -0,0 +1,194 @@
...
+   * @param nextState The next state.
+   * @return reward, it's always -1.0.
+   */
+  double Sample(const State& state,
+                const Action& action,
+                State& nextState) const
+  {
+    // Calculate acceleration.
+    double force = std::min(std::max(action.action[0], -1.0), 1.0);
+    nextState.Velocity() = state.Velocity() + 0.001 * force - 0.0025 *
```
zoq: Should we make ``0.001`` and ``0.0025`` configurable?

src/mlpack/methods/reinforcement_learning/environment/continuous_mountain_car.hpp
```diff
@@ -0,0 +1,194 @@
...
+   * Initial position is randomly generated within [-0.6, -0.4].
+   * Initial velocity is 0.
+   *
+   * @return Initial state for each episode.
+   */
+  State InitialSample() const
+  {
+    State state;
+    state.Velocity() = 0.0;
+    state.Position() = arma::as_scalar(arma::randu(1)) * 0.2 - 0.6;
```
zoq: What about using:

https://github.com/mlpack/mlpack/blob/114137eec8c025d62e77e2c9379ce5f0f76943b9/src/mlpack/core/math/random.hpp#L71-L74

that way we could avoid ``arma::as_scalar``.

zoq:

PR #1369, opened 2018-04-16
Add tests for command-line and Python bindings (nca)

yaswagner:
Tests for the NCA binding, to insure its inputs and outputs work correctly. 
Added a nca_test.cpp file under maint_test.

### DIFF
--- a/src/mlpack/tests/main_tests/nca_test.cpp
+++ b/src/mlpack/tests/main_tests/nca_test.cpp
@@ -1,0 +1,445 @@
+/**
+ * @file nca_test.cpp
+ * @author Yasmine Dumouchel
+ *
+ * Test mlpackMain() of nca_main.cpp.
+ *
+ * mlpack is free software; you may redistribute it and/or modify it under the
+ * terms of the 3-clause BSD license.  You should have received a copy of the
+ * 3-clause BSD license along with mlpack.  If not, see
+ * http://www.opensource.org/licenses/BSD-3-Clause for more information.
+ */
+
+#include <string>
+
+#define BINDING_TYPE BINDING_TYPE_TEST
+static const std::string testName = "nca";
+
+#include <mlpack/core.hpp>
+#include <mlpack/core/util/mlpack_main.hpp>
+#include <mlpack/core/metrics/lmetric.hpp>
+#include "test_helper.hpp"
+#include <mlpack/methods/nca/nca_main.cpp>
+#include <mlpack/core/optimizers/lbfgs/lbfgs.hpp>
+
+#include <boost/test/unit_test.hpp>
+#include "../test_tools.hpp"
+
+using namespace mlpack;
+
+struct NCATestFixture
+{
+ public:
+  NCATestFixture()
+  {
+    // Cache in the options for this program.
+    CLI::RestoreSettings(testName);
+  }
+
+  ~NCATestFixture()
+  {
+    // Clear the settings.
+    bindings::tests::CleanMemory();
+    CLI::ClearSettings();
+  }
+};
+
+BOOST_FIXTURE_TEST_SUITE(NCAMainTest, NCATestFixture);
+
+/**
+ * Ensure that, when labels are implicitily given with input,
+ * the last column is treated as labels and that we get the
+ * desired shape of output.
+ */
+BOOST_AUTO_TEST_CASE(NCAExplicitImplicitLabelsTest)
+{
+  // Random dataset.
+  arma::mat x;
+  x.randu(3, 6);
+
+  SetInputParam("input", std::move(x));
+
+  mlpackMain();
+
+  // Check that last row was treated as label by checking that
+  // the output has 1 less row.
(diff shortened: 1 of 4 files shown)


src/mlpack/tests/main_tests/nca_test.cpp
```diff
@@ -0,0 +1,380 @@
...
+{
+  // Simple dataset with 6 points and two classes.
+  arma::mat x              = "-0.1 -0.1 -0.1  0.1  0.1  0.1;"
+                             " 1.0  0.0 -1.0  1.0  0.0 -1.0 ";
+  arma::Row<size_t> labels = " 0    0    0    1    1    1   ";
+
+  // Set parameters and set normalize to true.
+  SetInputParam("input", std::move(x));
+  SetInputParam("labels", std::move(labels));
+  SetInputParam("normalize", true);
```
rcurtin: So, there is a little bit of trickiness here.  NCA by default uses the SGD optimizer, and the SGD optimizer depends on the random seed since it chooses batches randomly.  In this case, we want to ensure that normalization makes a difference, but as the code currently is, it's likely to give different results simply by virtue of the fact that the order in which batches are visited for SGD will be different.

I think a workaround we can use is for this type of test, we can set `linear_scan` to `true`, so that we will always visit points in the same order.  (The `linear_scan` test is fine as-is, since there all we want to know is that we get a different result there.  In fact, it might be useful to add another test where we run the same data twice with `linear_scan` set, and we should expect the same result.)

yaswagner:
Perfect and I will fix the 2 issues as well as add my name.

And yes I would love some mlpack stickers! Here is my address:
	3959 Rue Clark, Montréal, Canada H2W 1W8

Thank you!

src/mlpack/tests/main_tests/nca_test.cpp
```diff
@@ -0,0 +1,380 @@
...
+* Ensure that output is different when batch_size is different.
+*/
+BOOST_AUTO_TEST_CASE(NCADiffferentBatchSizeTest)
+{
+  // Simple dataset with 6 points and two classes.
+  arma::mat x              = "-0.1 -0.1 -0.1  0.1  0.1  0.1;"
+                             " 1.0  0.0 -1.0  1.0  0.0 -1.0 ";
+  arma::Row<size_t> labels = " 0    0    0    1    1    1   ";
+
+  // Set parameters with a small bath_size.
```
zoq: Minor issue; small ``batch_size``.

src/mlpack/tests/main_tests/nca_test.cpp
```diff
@@ -0,0 +1,380 @@
...
+    // Clear the settings.
+    bindings::tests::CleanMemory();
+    CLI::ClearSettings();
+  }
+};
+
+BOOST_FIXTURE_TEST_SUITE(NCAMainTest, NCATestFixture);
+
+/**
+* Ensure that, when labels are implicitily given with input,
```
zoq: Pedantic style issue; missing space at the beginning, this should align with the first ``*`` from the first line.

src/mlpack/tests/main_tests/nca_test.cpp
```diff
@@ -0,0 +1,380 @@
...
+  // Reset settings.
+  CLI::ClearSettings();
+  CLI::RestoreSettings(testName);
+
+  // Input same dataset.
+  arma::mat y               = "-0.1 -0.1 -0.1  0.1  0.1  0.1;"
+                              " 1.0  0.0 -1.0  1.0  0.0 -1.0 ";
+  arma::Row<size_t> labels2 = " 0    0    0    1    1    1   ";
+
+  // Set parameters using the same input but set linear_scan flag to false
```
zoq: Pedantic style issue number 2, missing ``.`` at the end.

src/mlpack/tests/main_tests/nca_test.cpp
```diff
@@ -0,0 +1,380 @@
...
+
+  mlpackMain();
+
+  // Check that the output matrices are different.
+  BOOST_REQUIRE_GT(arma::accu(CLI::GetParam<arma::mat>("output") != output), 0);
+}
+
+/**
+* Ensure that using a different value of max_iteration
+* results in a different output matrix
```
zoq: See comment above.

rcurtin:
Thanks again for the contribution!  I fixed a handful of things during the merge:

 * f5bee6e79, 7d12b2bed: added an explicit test for the label size to `nca_main.cpp`---the previous test would only succeed when compiled in debug mode.
 * 93fa62891: fixed a memory error I found with valgrind while running the tests (though that wasn't related to any code you wrote, I just found it and fixed it)
 * 6ad2eaf8e: minor issue, the labels should have had 100 labels not 3 :)
 * e18d61dae: I applied `linear_scan` to the rest of the tests that used SGD

Let me know if I broke anything or changed it in a way you don't like; I think the changes are mostly pretty minor though. :+1:

PR #1370, opened 2018-04-16
Add terminal Reward .

luffy1996:
The previous implementation did not care about the terminal condition and its reward. Hence I am adding terminal condition to the environment, to take care of the terminal conditions .

### DIFF
--- a/src/mlpack/methods/reinforcement_learning/q_learning_impl.hpp
+++ b/src/mlpack/methods/reinforcement_learning/q_learning_impl.hpp
@@ -155,7 +155,7 @@ double QLearning<
       target(sampledActions[i], i) = sampledRewards[i];
     else
       target(sampledActions[i], i) = sampledRewards[i] + config.Discount() *
-          (isTerminal[i] ? 0.0 : nextActionValues(bestActions[i], i));
+          nextActionValues(bestActions[i], i);
   }
 
   // Learn form experience.
--- a/src/mlpack/methods/reinforcement_learning/q_learning_impl.hpp
+++ b/src/mlpack/methods/reinforcement_learning/q_learning_impl.hpp
@@ -144,10 +144,18 @@ double QLearning<
   // Compute the update target.
   arma::mat target;
   learningNetwork.Forward(sampledStates, target);
+  /**
+   * If the agent is at a terminal state, then we don't need to add the
+   * discounted reward. At terminal state, the agent wont perform any
+   * action.
+   */
   for (size_t i = 0; i < sampledNextStates.n_cols; ++i)
   {
-    target(sampledActions[i], i) = sampledRewards[i] + config.Discount() *
-        (isTerminal[i] ? 0.0 : nextActionValues(bestActions[i], i));
+    if (isTerminal[i])
+      target(sampledActions[i], i) = sampledRewards[i];
+    else
+      target(sampledActions[i], i) = sampledRewards[i] + config.Discount() *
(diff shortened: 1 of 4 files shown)


luffy1996:
Dear @zoq , I have added terminal case to the Q Learning Algorithm. This implementation was overlooked previously. Please let me know what do you think about this.

luffy1996:
Style : #1372 
Static Code Analysis : I am not sure why this error has occurred. I guess all the previous analysis on this file was successful. 

@zoq , Please let me know if any details are required.

sourabhvarshney111:
@luffy1996 I don't think there is an error which you have created. It is like if you have changed the file then every code of that file is checked for static code analysis. See static checks for https://github.com/mlpack/mlpack/pull/1309 , you can easily detect that these might not be errors too but instead a use of the statement in different manner.

zoq:
Don't worry about the Static Code Analysis issue, in this case it's just fine to use ``action`` inside the condition.

luffy1996:
@sourabhvarshney111 . I looked at the PR and it seems that this error has nothing to do with me changing few lines of the code. Thanks for pointing it out.

luffy1996:
@zoq , Do you think this implementation should be merged ?

zoq:
The issue I see is, that is, that this bias could influence the learning strategy. Let's say we have two runs with -10 and -11 reward and -11 is also a terminal state. With your code we end up with -10 and 90 as the reward which could be fine, but this would imply that the second run (90) is way better in terms of the reward. Let me know what you think. I guess, if someone used this before, we could make the ``done reward`` configurable.

zoq:
> For this game, the purpose is to reach the minimum height in a minimum number of steps. Now if we give a reward of -1 here, like the previous implementation, the agent doesn't learn the dynamics of the game.

I don't think this is true, it could learn the task since the agent might be able to correlate the reward over the time. However, I agree that returning another reward might be beneficial, but I'm not sure 100 is a good default value, why not just go with 0 as done for the Cartpole task?

luffy1996:
0 could also be given to the agent as the default value. I am going ahead with default values used by openai gym environment.

src/mlpack/methods/reinforcement_learning/q_learning_impl.hpp
```diff
@@ -144,10 +144,18 @@ double QLearning<
...
+  /**
+   * If the agent is at a terminal state, then we don't need to add the
+   * discounted reward. At terminal state, the agent wont perform any
+   * action.
+   */
   for (size_t i = 0; i < sampledNextStates.n_cols; ++i)
   {
-    target(sampledActions[i], i) = sampledRewards[i] + config.Discount() *
-        (isTerminal[i] ? 0.0 : nextActionValues(bestActions[i], i));
+    if (isTerminal[i])
```
zoq: The current code already covers that:

```c++
config.Discount() * (isTerminal[i] ? 0.0 : nextActionValues(bestActions[i], i))
```

if ``isTerminal[i]`` is true we multiply ``config.Discount()`` with ``0.0``.

luffy1996:
@rcurtin I do understand your point. The choice of doneReward is usually user dependent. I was planning for a higher reward, but @zoq pointed out that we should keep it less. I guess he was correct at this point because we should not introduce large bias from our end. Let the user determine the magnitude of doneReward for their game.

luffy1996:
@zoq, I think you can go ahead and add this. :)

zoq:
Thanks, happy to use this in some future tests.

PR #1372, opened 2018-04-17
Minor Style

luffy1996:
This is minor style fixes for acrobat game. The length of line 230 was 81, which created style issues.


zoq:
Thanks for the quick fix.

PR #1373, opened 2018-04-19
fix unused result warning

conradsnicta:
Recent versions of armadillo emit warnings when the result of iterator postfix increment/decrement is not used.  The void statement by itself isn't sufficient to suppress these warnings.


zoq:
Your choice, either way, is fine with me.

zoq:
Thanks again, for the fix.

PR #1374, opened 2018-04-20
fix unused result warning

conradsnicta:
As per https://github.com/mlpack/mlpack/pull/1373


conradsnicta:
@rcurtin - if the contribution is sufficient, then adding to the list of contributors sounds good (but please don't include my email address).

I'll have to make more contributions before deserving the stickers :)

rcurtin:
Sure, added in 46eb0a07 with no email.  I have international stamps so it is easy for me to mail stickers, so just let me know if you ever want any.  :)

PR #1378, opened 2018-04-21
Update NeighborSearch: Switching l-rvalue reference overload to pass-by-value

ThanasisMattas:
Switching lvalue reference and rvalue reference overload to pass-by-value argument in the Neighbor Search algorithm, as suggested at #1021 issue.

A user can pass the data by making a copy (Method(data)) or, avoiding copies if possible, by moving the data into the method (Method(std::move(data))).


rcurtin:
Thanks again! :)

PR #1379, opened 2018-04-22
robust_pca implementation

charan223:
Added robust pca implementation, which is pca using robust svd analysis

### DIFF
--- a/src/mlpack/methods/pca/decomposition_policies/robust_svd_method.hpp
+++ b/src/mlpack/methods/pca/decomposition_policies/robust_svd_method.hpp
@@ -1,0 +1,100 @@
+/**
+ * @file robust_svd_method.hpp
+ * @author Charan Reddy
+ *
+ * Implementation of the robust svd method for use in the Principal
+ * Components Analysis method.
+ *
+ * mlpack is free software; you may redistribute it and/or modify it under the
+ * terms of the 3-clause BSD license.  You should have received a copy of the
+ * 3-clause BSD license along with mlpack.  If not, see
+ * http://www.opensource.org/licenses/BSD-3-Clause for more information.
+ */
+
+#ifndef MLPACK_METHODS_PCA_DECOMPOSITION_POLICIES_ROBUST_SVD_METHOD_HPP
+#define MLPACK_METHODS_PCA_DECOMPOSITION_POLICIES_ROBUST_SVD_METHOD_HPP
+
+#include <mlpack/prereqs.hpp>
+#include <mlpack/methods/robust_svd/robust_svd.hpp>
+
+namespace mlpack {
+namespace pca {
+
+/**
+ * Implementation of the robust SVD policy.
+ */
+class RobustSVDPolicy
+{
+ public:
+  /**
+   * Use robust SVD method to perform the principal components analysis
+   * (PCA).
+   *
+   * @param iteratedPower Size of the normalized power iterations
+   *        (Default: rank + 2).
+   * @param maxIterations Number of iterations for the power method
+   *        (Default: 2).
+   */
+  RobustSVDPolicy(const size_t iteratedPower = 0,
+                      const size_t maxIterations = 2) :
+      iteratedPower(iteratedPower),
+      maxIterations(maxIterations)
+  {
+    /* Nothing to do here */
+  }
+
+  /**
+   * Apply Principal Component Analysis to the provided data set using the
+   * robust SVD.
+   *
+   * @param data Data matrix.
+   * @param centeredData Centered data matrix.
+   * @param transformedData Matrix to put results of PCA into.
+   * @param eigVal Vector to put eigenvalues into.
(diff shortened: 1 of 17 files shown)


src/mlpack/methods/pca/decomposition_policies/robust_svd_method.hpp
```diff
@@ -0,0 +1,100 @@
...
+#define MLPACK_METHODS_PCA_DECOMPOSITION_POLICIES_ROBUST_SVD_METHOD_HPP
+
+#include <mlpack/prereqs.hpp>
+#include <mlpack/methods/robust_svd/robust_svd.hpp>
+
+namespace mlpack {
+namespace pca {
+
+/**
+ * Implementation of the robust SVD policy.
```
rcurtin: Can you clarify some more in the documentation here about what robust SVD is, and can you add a reference to the paper you are implementing?

src/mlpack/methods/pca/decomposition_policies/robust_svd_method.hpp~
```diff
@@ -0,0 +1,100 @@
+/**
+ * @file robust_svd_method.hpp
```
rcurtin: I think this file was accidentally checked in, it looks like an editor-created file.

src/mlpack/methods/pca/pca.hpp
```diff
@@ -27,10 +27,11 @@ namespace pca {
  * common, widely-used technique that is often used for either dimensionality
  * reduction or transforming data into a better basis.  Further information on
  * PCA can be found in almost any statistics or machine learning textbook, and
- * all over the internet.
+ * all over the internet. Note this class will be changed to have the name PCA
+ * in mlpack 3.0.0
```
rcurtin: This reverts some changes that we have made for mlpack 3.0.0, can you undo your changes to this file please?

src/mlpack/methods/pca/pca_main.cpp
```diff
@@ -10,6 +10,8 @@
  * 3-clause BSD license along with mlpack.  If not, see
  * http://www.opensource.org/licenses/BSD-3-Clause for more information.
  */
+
+
```
rcurtin: No need to add extra lines.  (That also applies a few other places later in the file.)

src/mlpack/methods/pca/pca_main.cpp
```diff
@@ -101,6 +105,7 @@ void RunPCA(arma::mat& dataset,
 
 static void mlpackMain()
 {
+  Log::Info << "here" << endl;
```
rcurtin: This looks like extra debugging output left here by accident.

src/mlpack/methods/robust_svd/CMakeLists.txt~
```diff
@@ -0,0 +1,15 @@
+# Define the files we need to compile.
```
rcurtin: This looks like an extra file also.

src/mlpack/methods/robust_svd/check.cpp~
```diff
@@ -0,0 +1,11 @@
+#include <armadillo>
+#include <iostream>
```
rcurtin: This one also seems extra.

src/mlpack/methods/robust_svd/robust_svd.cpp
```diff
@@ -0,0 +1,85 @@
...
+	check2 = arma::zeros(size(X));
+	S = arma::sign(X) * arma::max((arma::abs(X) - mu_inv*lmbda), check2);
+	Y = Y + mu * (data - L - S);
+	err = arma::accu(arma::pow(arma::abs(data - L - S) , 2));
+	iter+=1;
+	if(iter%100 !=0 || iter ==1 || iter >max_iter || err<=tol)
+	{
+		std::cout<<"iteration: "<< iter << " error: "<<err<<std::endl;
+	} 
+  }
```
rcurtin: There is no documentation or comments at all in this block of code.  I think that will make it very hard to maintain.  If you can rewrite this completely in a well-commented and easier to understand way (with references to equations in the paper where needed) it would be really helpful.

Also we should not use output directly to `cout`---if you want to provide iteration-specific output, use `Log::Info` please.

src/mlpack/tests/CMakeLists.txt
```diff
@@ -149,13 +147,8 @@ add_executable(mlpack_test
   main_tests/random_forest_test.cpp
   main_tests/softmax_regression_test.cpp
   main_tests/sparse_coding_test.cpp
-  main_tests/kmeans_test.cpp
   main_tests/hoeffding_tree_test.cpp
-  main_tests/hmm_viterbi_test.cpp
   main_tests/hmm_train_test.cpp
-  main_tests/hmm_loglik_test.cpp
-  main_tests/hmm_generate_test.cpp
-  main_tests/hmm_test_utils.hpp
```
rcurtin: We shouldn't remove tests.

PR #1380, opened 2018-04-24
Add missing check to suppress compiler warning

WhoisZihan:
The variable 'model' may not be initialized, thus causing a compile warning in
src/mlpack/methods/hoeffding_trees/hoeffding_tree_main.cpp:251

### DIFF
--- a/src/mlpack/methods/hoeffding_trees/hoeffding_tree_main.cpp
+++ b/src/mlpack/methods/hoeffding_trees/hoeffding_tree_main.cpp
@@ -159,6 +159,10 @@ static void mlpackMain()
       model = new HoeffdingTreeModel(HoeffdingTreeModel::INFO_HOEFFDING);
     else if (CLI::HasParam("info_gain") && (numericSplitStrategy == "binary"))
       model = new HoeffdingTreeModel(HoeffdingTreeModel::INFO_BINARY);
+    else {
+      Log::Fatal << "model is not initialized" << std::endl;
+      return;
+    }
   }
 
   // Now, do we need to train?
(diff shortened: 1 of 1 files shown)


src/mlpack/methods/hoeffding_trees/hoeffding_tree_main.cpp
```diff
@@ -159,6 +159,10 @@ static void mlpackMain()
       model = new HoeffdingTreeModel(HoeffdingTreeModel::INFO_HOEFFDING);
     else if (CLI::HasParam("info_gain") && (numericSplitStrategy == "binary"))
       model = new HoeffdingTreeModel(HoeffdingTreeModel::INFO_BINARY);
+    else {
```
zoq: Avoid placing opening braces on the same line. See https://github.com/mlpack/mlpack/wiki/DesignGuidelines#brace-placement for more information; this might be irrelevant if we switch the ``else if`` to ``else``.

WhoisZihan:
@rcurtin @zoq Hi, thanks for the review.

> I think that it is not possible for the code to actually get to that point. So an easier fix would be to change the last else if to just an else.

Thanks for clarifying, I didn't know it because I just start to get myself familiar with machine learning and mlpack, so this is just a minor patch without knowing the context. I just update it with a new commit.

> I'll go ahead and merge this in 3 days to leave time for comments. If you like, we can add your name to the list of contributors in src/mlpack/core.hpp and COPYRIGHT.txt (since even though it is a small contribution it is still a contribution!). Also, if you like, I can mail you some mlpack stickers that you can put on your laptop. If you want some, send me your mailing address at ryan@ratml.org and I will put them in the mail.

That would be nice! Stickers will look cool. I think I can contribute more when I get familiar enough with ML and mlpack.

src/mlpack/methods/hoeffding_trees/hoeffding_tree_main.cpp
WhoisZihan: Thanks for clarifying, I just updated the pull request with a new commit.

zoq:
Sometimes, tests that have nothing to do with your code will fail. Don't worry about those; see https://github.com/mlpack/mlpack/issues/922 for more information. In this case we should see if we can track down the error, but  no need to delay the PR.

Issue #404, opened 2015-02-13
SDP should deal with linearly dependent constraints somehow


stephentu:
we need some strategy for dealing w/ a linearly dependent set of constraints. for LR-SDP, this isn't so big of a deal. but, this is especially key for the primal dual methods, which require this assumption so that various linear systems have unique solutions. we could either introduce a method which removes redundant constraints, or adapt the primal-dual method to handle degenerate linear systems.

stephentu:
note that, if we want the MVU SDP to work, then this will be necessary

palashahuja:
@stephentu , could you please give more information about this ?

rcurtin:
@palashahuja: are you familiar with SDPs and LRSDP?  If not you should take a look into that first.  mlpack provides some code for solving SDPs (in src/mlpack/core/optimizers/sdp/) but it doesn't handle linear constraints nicely (you have to represent your linear constraint as an entire dense or sparse matrix which is often unwieldy and in some cases not possible).  So we need to find a way to adapt that code to deal with linear constraints more nicely.

palashahuja:
@rcurtin , I have looked into LRSDP and SDP. What I am asking is how to deal with linear constraints ?

rcurtin:
I don't know what you mean when you say "some sort of string".  This is, in my interpretation, an exploratory ticket: we don't currently have a good way to deal with linear constraints, and one needs to be designed.  So the goal is to design a nice solution for handling linear constraints.  It's possible this is already being handled by someone else in #578, so it might be better to try and avoid duplicating work here.

stephentu:
@rcurtin: this task is slightly different than #578 (it took me a while to realize this). This tasks deals w/ the issue that currently the IP code assumes the nullspace of the A operator is trivial. This is not an easy task IMO, and I'm not sure what the current benefits are (nobody has ever complained about this).

rcurtin:
Oops, you're right, I misread---I read "we need some strategy for dealing w/ linear constraints" but it actually says "we need some strategy for dealing w/ a linearly dependent set of constraints".  Sorry about that!

rickllyxu:
@stephentu @rcurtin  Glad to tell you that, mvu based on primal-dual SDP solver **works**! I modify the file `mlpack/src/mlpack/tests/sdp_primal_dual_test.cpp` and the test case `SmallMvuSdp` works! I will commit codes after adjusting the code style.
[...]

PR #1381, opened 2018-04-26
add simple code to statisfy the independence and SmallMvuSdp works

rickllyxu:
Plz review the code. I try to solve the issue [#404](https://github.com/mlpack/mlpack/issues/404)

### DIFF
--- a/src/mlpack/tests/sdp_primal_dual_test.cpp
+++ b/src/mlpack/tests/sdp_primal_dual_test.cpp
@@ -14,7 +14,6 @@
 #include <mlpack/core/optimizers/sdp/primal_dual.hpp>
 #include <mlpack/methods/neighbor_search/neighbor_search.hpp>
 
-//#include <set>
 #include <boost/test/unit_test.hpp>
 #include "test_tools.hpp"
 
@@ -559,7 +558,7 @@ BOOST_AUTO_TEST_CASE(CorrelationCoeffToySdp)
   BOOST_REQUIRE_CLOSE(obj, 2 * (-0.978), 1e-3);
 }
 
- /**
+/**
  * Maximum variance unfolding (MVU) SDP to learn the unrolled gram matrix. For
  * the SDP formulation, see:
  *
@@ -576,9 +575,9 @@ BOOST_AUTO_TEST_CASE(CorrelationCoeffToySdp)
  * @param origData origDim x numPoints
  * @param numNeighbors
  */
- static inline SDP<arma::sp_mat> ConstructMvuSDP(const arma::mat& origData,
-                                                size_t numNeighbors)
- {
+static inline SDP<arma::sp_mat> ConstructMvuSDP(const arma::mat& origData,
+                                              size_t numNeighbors)
+{
   const size_t numPoints = origData.n_cols;
 
   assert(numNeighbors <= numPoints);
@@ -587,25 +586,30 @@ BOOST_AUTO_TEST_CASE(CorrelationCoeffToySdp)
   arma::mat distances;
   KNN knn(origData);
   knn.Search(numNeighbors, neighbors, distances);
+
   // how to construct the matrices A, b, and C? Plz refer to the above paper.
   std::set<std::pair<int, int>> neighbors_set;
   for (size_t i = 0; i < neighbors.n_cols; ++i)
   {
-     for (size_t j = 0; j < numNeighbors; ++j)
-     {
-       size_t x = i, y = neighbors(j, i);
-       if (x > y) {
-           x = neighbors(j, i);
-           y = i;
-       }
-       if (neighbors_set.find(std::make_pair(x, y)) != neighbors_set.end()) {
-           continue;
-       }
-       neighbors_set.insert(std::make_pair(x, y));
-     }
+    for (size_t j = 0; j < numNeighbors; ++j)
+    {
+      size_t x = i, y = neighbors(j, i);
+      if (x > y)
+      {
+        x = neighbors(j, i);
+        y = i;
+      }
+
+      if (neighbors_set.find(std::make_pair(x, y)) != neighbors_set.end())
+      {
+        continue;
+      }
+
+      neighbors_set.insert(std::make_pair(x, y));
+    }
   }
 
-//  SDP<arma::sp_mat> sdp(numPoints, numNeighbors * numPoints, 1);
+  // SDP<arma::sp_mat> sdp(numPoints, numNeighbors * numPoints, 1);
   SDP<arma::sp_mat> sdp(numPoints, neighbors_set.size(), 1);
   sdp.C().eye(numPoints, numPoints);
   sdp.C() *= -1;
@@ -628,14 +632,15 @@ BOOST_AUTO_TEST_CASE(CorrelationCoeffToySdp)
         x = neighbors(j, i);
         y = i;
       }
-      if (neighbors_set.find(std::make_pair(x, y)) != neighbors_set.end()) {
+      if (neighbors_set.find(std::make_pair(x, y)) != neighbors_set.end())
+      {
         continue;
       }
       neighbors_set.insert(std::make_pair(x, y));
 
       // This is the index of the constraint.
       const size_t index = _index;
-        _index++;
+      _index++;
       arma::sp_mat& aRef = sdp.SparseA()[index];
       aRef.zeros(numPoints, numPoints);
 
@@ -657,16 +662,16 @@ BOOST_AUTO_TEST_CASE(CorrelationCoeffToySdp)
   }
 
   return sdp;
- }
+}
 
- /**
+/**
  * Maximum variance unfolding
  *
  * Test doesn't work, because the constraint matrices are not linearly
  * independent.
  */
- BOOST_AUTO_TEST_CASE(SmallMvuSdp)
- {
+BOOST_AUTO_TEST_CASE(SmallMvuSdp)
+{
   const size_t n = 20;
 
   arma::mat origData(3, n);
@@ -689,7 +694,7 @@ BOOST_AUTO_TEST_CASE(CorrelationCoeffToySdp)
   const auto p = solver.Optimize(X, ysparse, ydense, Z);
   bool success = CheckKKT(sdp, X, ysparse, ydense, Z);
   BOOST_REQUIRE_EQUAL(success, true);
-//  BOOST_REQUIRE(p.first);
- }
+  // BOOST_REQUIRE(p.first);
+}
 
 BOOST_AUTO_TEST_SUITE_END();
--- a/src/mlpack/tests/sdp_primal_dual_test.cpp
+++ b/src/mlpack/tests/sdp_primal_dual_test.cpp
@@ -1,6 +1,7 @@
 /**
  * @file sdp_primal_dual_test.cpp
  * @author Stephen Tu
+ * @author Kaiqiang Xu
  *
  *
  * mlpack is free software; you may redistribute it and/or modify it under the
@@ -13,6 +14,7 @@
 #include <mlpack/core/optimizers/sdp/primal_dual.hpp>
 #include <mlpack/methods/neighbor_search/neighbor_search.hpp>
 
+//#include <set>
 #include <boost/test/unit_test.hpp>
 #include "test_tools.hpp"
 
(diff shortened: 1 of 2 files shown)


src/mlpack/tests/sdp_primal_dual_test.cpp
```diff
@@ -557,92 +558,141 @@ BOOST_AUTO_TEST_CASE(CorrelationCoeffToySdp)
...
+  // SDP<arma::sp_mat> sdp(numPoints, numNeighbors * numPoints, 1);
+  SDP<arma::sp_mat> sdp(numPoints, neighbors_set.size(), 1);
+  sdp.C().eye(numPoints, numPoints);
+  sdp.C() *= -1;
+  sdp.DenseA()[0].ones(numPoints, numPoints);
+  sdp.DenseB()[0] = 0;
+  neighbors_set.clear();
+  size_t _index = 0;
+
+  for (size_t i = 0; i < neighbors.n_cols; ++i)
```
rcurtin: It seems like you could just loop over the already-filled `neighbors_set` here instead of refiltering all the duplicate neighbor pairs.

src/mlpack/tests/sdp_primal_dual_test.cpp
```diff
@@ -557,92 +558,141 @@ BOOST_AUTO_TEST_CASE(CorrelationCoeffToySdp)
...
+{
+  const size_t n = 20;
+
+  arma::mat origData(3, n);
+
+  // sample n random points on 3-dim unit sphere
+  GaussianDistribution gauss(3);
+  for (size_t i = 0; i < n; i++)
+  {
+    // how european of them
```
rcurtin: Actually Conrad Sanderson (the Armadillo maintainer) is Australian. :)

src/mlpack/tests/sdp_primal_dual_test.cpp
```diff
@@ -557,92 +558,141 @@ BOOST_AUTO_TEST_CASE(CorrelationCoeffToySdp)
...
+  }
+
+  return sdp;
+}
+
+/**
+ * Maximum variance unfolding
+ *
+ * Test doesn't work, because the constraint matrices are not linearly
+ * independent.
```
rcurtin: It seems like the test does work now, so I think we can remove this comment.

src/mlpack/tests/sdp_primal_dual_test.cpp
```diff
@@ -557,92 +558,141 @@ BOOST_AUTO_TEST_CASE(CorrelationCoeffToySdp)
...
+  }
+
+  // SDP<arma::sp_mat> sdp(numPoints, numNeighbors * numPoints, 1);
+  SDP<arma::sp_mat> sdp(numPoints, neighbors_set.size(), 1);
+  sdp.C().eye(numPoints, numPoints);
+  sdp.C() *= -1;
+  sdp.DenseA()[0].ones(numPoints, numPoints);
+  sdp.DenseB()[0] = 0;
+  neighbors_set.clear();
+  size_t _index = 0;
```
rcurtin: A minor style issue---we should name this `index` not `_index`.

src/mlpack/tests/sdp_primal_dual_test.cpp
```diff
@@ -557,92 +558,141 @@ BOOST_AUTO_TEST_CASE(CorrelationCoeffToySdp)
...
+
+  assert(numNeighbors <= numPoints);
+
+  arma::Mat<size_t> neighbors;
+  arma::mat distances;
+  KNN knn(origData);
+  knn.Search(numNeighbors, neighbors, distances);
+
+  // how to construct the matrices A, b, and C? Plz refer to the above paper.
+  std::set<std::pair<int, int>> neighbors_set;
```
rcurtin: A minor style issue... we should call this `neighborsSet` not `neighbors_set` (i.e., no underscores in the variable names, we use camelcase naming instead).

src/mlpack/tests/sdp_primal_dual_test.cpp
rickllyxu: right! I planned to loop the already-filled set, but then I found I cannot determine `sdp.SparseB()[index] = distances(j, i);` without `j`. Though we can store the `distances(j, i)` into the set, I don't think it is sensible to do it for the distances are floating-point numbers.

src/mlpack/tests/sdp_primal_dual_test.cpp
rickllyxu: :) hhha

src/mlpack/tests/sdp_primal_dual_test.cpp
rcurtin: You could use `std::tuple<size_t, size_t, double>` instead of `std::pair<int, int>`.  (Actually it should also be `size_t` instead of `int` regardless of what you do.)

rickllyxu:
Note that the linear independence is satisfied as I say in #404. However the the SDP formulation of mvu still almost can not be solved. It can work when I set the number of samples to 20 (100) and `numNeighbors` to 5 (50), but not work (not converge) for 30, 50, 100 with all of their `numNeighbors` is 5. 

The codes around `mlpack/src/mlpack/tests/sdp_primal_dual_test.cpp(695)`  in my test are following:
```
  BOOST_REQUIRE_EQUAL(sdp.HasLinearlyIndependentConstraints(), true);

  PrimalDualSolver<SDP<arma::sp_mat>> solver(sdp);
  arma::mat X, Z;
  arma::vec ysparse, ydense;
  const auto p = solver.Optimize(X, ysparse, ydense, Z);
  bool success = CheckKKT(sdp, X, ysparse, ydense, Z);
  BOOST_REQUIRE_EQUAL(success, true);  // This is line 695.
```
Here is some test record for running shell command `time bin/mlpack_test -t SdpPrimalDualTest/SmallMvuSdp`:
**when the number of samples is 20, the number of neighbors is 5**, `const size_t n = 20, numneighbors = 5`:
*** No errors detected
real    0m0.232s
user    0m0.428s
sys     0m0.456s

**when the number of samples is 30, the number of neighbors is 5**, `const size_t n = 30, numneighbors = 5`:
mlpack/src/mlpack/tests/sdp_primal_dual_test.cpp(695): fatal error in "SmallMvuSdp": critical check success == true failed [false != true]
real    0m1.138s
user    0m1.839s
sys     0m2.570s

**when the number of samples is 50, the number of neighbors is 5**, `const size_t n = 50, numneighbors = 5`:
mlpack/src/mlpack/tests/sdp_primal_dual_test.cpp(695): fatal error in "SmallMvuSdp": critical check success == true failed [false != true] 
real    0m3.593s
user    0m4.631s
sys     0m2.838s

**when the number of samples is 100, the number of neighbors is 5**, `const size_t n = 100, numneighbors = 5`:
mlpack/src/mlpack/tests/sdp_primal_dual_test.cpp(695): fatal error in "SmallMvuSdp": critical check success == true failed [false != true]
real    0m59.322s
user    1m41.182s
sys     1m53.045s

**when the number of samples is 100, the number of neighbors is 50**, `const size_t n = 100, numneighbors = 50`:
*** No errors detected
real    11m53.673s
user    21m50.191s
sys     22m36.955s

I think this converage version is deserved to merged with good comment and its time cost is within 1 second.

src/mlpack/tests/sdp_primal_dual_test.cpp
rcurtin: It may be a starting point issue.  You might try running the test many times with different random seeds and seeing how different the results are, and how often it fails/succeeds.  (Note that if you use `math::RandomSeed()`, you'll have to comment out the #ifdef in src/mlpack/core/math/random.hpp.)

src/mlpack/tests/sdp_primal_dual_test.cpp
rickllyxu: good suggestion! I will update it, but I concern about the comparison of floating-point number. I will figure it out.

rickllyxu: overloading the comparabel function will solve the concerning:)

src/mlpack/tests/sdp_primal_dual_test.cpp
rickllyxu: What do you mean "comment out the #ifdef in src/mlpack/core/math/random.hpp"? Can I just use `mlpack::math::RandomSeed((size_t) std::time(NULL))` ?

rickllyxu:
I set the random seeds, and get same result as before.

rcurtin:
Hm, ok.  Well I think it is definitely a good fix to fix the linear independence problem with the constraints, but it looks like maybe that is not everything here.  Where inside `CheckKKT()` is the test failing?  The possibilities seem to be:

 * `X` is not p.s.d.
 * `Z` is not p.s.d.
 * At least one sparse constraint is not satisfied.
 * At least one dense constraint is not satisfied.
 * The infeasibility of the dual is too high.

If you can isolate which situation (or situations) is (or are) happening, I think that can help a lot in figuring out what is wrong. :+1:

src/mlpack/tests/sdp_primal_dual_test.cpp
rcurtin: Ah, sorry, this comment was incorrect on my end.  If we were writing a test for any of the command-line bindings, using different random seeds is disabled via the `#ifdef`s in `random.hpp` (since some bindings will always try to set a random seed, and we don't want that to happen for reproducibility).  But since you are not working with any test in `main_tests/`, the issue doesn't apply here---`sdp_primal_dual_test.cpp` is not built with `BINDING_TYPE` set to anything at all.

Sorry for the confusion!

rickllyxu:
How do you think of the following information? I run `bin/mlpack_test -t SdpPrimalDualTest/SmallMvuSdp -v` with debug_mode on. Note that the number of samples is `30` and neighbors is `5`. Obviously, it seems that this sdp problem does not have strong duality. And the failing of cholesky decomposition of `X` results in terminating optimization. 

Running 1 test case...
**[DEBUG]** iter=1, primal=-20.123, dual=21.0325, gap=-41.1554, ||XZ||=1.62474, primalInfeas=3.75796, dualInfeas=26.9144, mu=0.0133029
**[DEBUG]** iter=2, primal=-34.1739, dual=10.0865, gap=-44.2604, ||XZ||=9.65928, primalInfeas=3.72619, dualInfeas=25.6497, mu=0.151887
**[DEBUG]** iter=3, primal=-34.4576, dual=-18.1977, gap=-16.2599, ||XZ||=10.6204, primalInfeas=3.66979, dualInfeas=14.6628, mu=0.259062
**[DEBUG]** iter=4, primal=-34.4864, dual=-18.2381, gap=-16.2483, ||XZ||=10.6053, primalInfeas=3.65902, dualInfeas=14.6403, mu=0.0816745
**[DEBUG]** iter=5, primal=-34.4867, dual=-18.2385, gap=-16.2482, ||XZ||=10.6051, primalInfeas=3.65892, dualInfeas=14.6401, mu=0.081413
**[DEBUG]** iter=6, primal=-34.4867, dual=-18.2385, gap=-16.2482, ||XZ||=10.6051, primalInfeas=3.65892, dualInfeas=14.6401, mu=0.0814113
**[DEBUG]** iter=7, primal=-34.4867, dual=-18.2385, gap=-16.2482, ||XZ||=10.6051, primalInfeas=3.65892, dualInfeas=14.6401, mu=0.0814112
**[DEBUG]** iter=8, primal=-34.4867, dual=-18.2385, gap=-16.2482, ||XZ||=10.6051, primalInfeas=3.65892, dualInfeas=14.6401, mu=0.0814051
**[DEBUG]** iter=9, primal=-34.4867, dual=-18.2385, gap=-16.2482, ||XZ||=10.6051, primalInfeas=3.65892, dualInfeas=14.6401, mu=0.0816368

**warning:** solve(): system appears singular (rcond: 8.00499e-17); attempting approximate solution

**warning:** solve(): system appears singular (rcond: 8.00499e-17); attempting approximate solution
**[DEBUG]** iter=10, primal=-34.4867, dual=-18.2385, gap=-16.2482, ||XZ||=10.6051, primalInfeas=3.65892, dualInfeas=14.6401, mu=0.0863428

**warning**: solve(): system appears singular (rcond: 6.53551e-17); attempting approximate solution

**warning**: chol(): decomposition failed
**[WARN ]** primal_dual_impl.hpp::Alpha(): chol decomposition failed
**[WARN ]** PrimalDualSolver::Optimize(): cholesky decomposition of X failed!  Terminating optimization.
**[WARN ]** sdp_primal_dual_test.cpp::CheckKKT(): check KKT conditions, matrix X is not positive semidefinite.

mlpack/src/mlpack/tests/sdp_primal_dual_test.cpp(693): fatal error in "SmallMvuSdp": critical check success == true failed [false != true]

rcurtin:
Hmm, so looking at this particular output, what I see is that the system that `solve()` is being called on is becoming increasingly singular, and eventually (on iteration 10) this means that the Cholesky decomposition fails (because whatever matrix in `Alpha()`, I guess `X`, is no longer positive semidefinite).  So if you can look into what happens to `X` each iteration, I think this could be helpful for debugging.

Is this the way that it fails for every random seed?  It might be worth trying over, e.g., every 50 random seeds and seeing if the behavior is different.

rcurtin:
I've closed this for inactivity, but if you'd like to work on it again in the future we can definitely reopen the PR. :)

rickllyxu:
> I've closed this for inactivity, but if you'd like to work on it again in the future we can definitely reopen the PR. :)

OK, these days I am working on the problem. But I do not think there will be a solution. Just as you said, `X` is no longer positive semidefinite. I check the code (implement of primal-dual SDP solver) leveraging the paper, I did not find any bugs. Let me try again.

rcurtin:
Sounds good---if you have a breakthrough, we can definitely get it merged in. :+1:

Issue #1358, opened 2018-04-07
[SoftmaxRegression] munmap_chunk(): invalid pointer during smr.Train()


bqv:
See https://gist.github.com/bqv/ada8359fe79da1ede98012d9e5579712 for a minimal reproducible example.

Backtrace:

```#0  __GI_raise (sig=sig@entry=6) at ../sysdeps/unix/sysv/linux/raise.c:51
#1  0x00007f8759677cf7 in __GI_abort () at abort.c:90
#2  0x00007f87596b8f87 in __libc_message (action=action@entry=do_abort, fmt=fmt@entry=0x7f87597bec38 "%s\n") at ../sysdeps/posix/libc_fatal.c:181
#3  0x00007f87596bf27a in malloc_printerr (str=str@entry=0x7f87597c04f8 "munmap_chunk(): invalid pointer") at malloc.c:5354
#4  0x00007f87596bf72c in munmap_chunk (p=<optimized out>) at malloc.c:2845
#5  0x00000000004155c8 in arma::memory::release<double const> (mem=0x7fffffffe180) at /usr/include/armadillo_bits/memory.hpp:145
#6  arma::Mat<double>::~Mat (this=0x7fffffffd830) at /usr/include/armadillo_bits/Mat_meat.hpp:29
#7  0x0000000000419b05 in mlpack::regression::SoftmaxRegressionFunction::~SoftmaxRegressionFunction (this=0x7fffffffd6b0) at /usr/include/mlpack/methods/softmax_regression/softmax_regression_function.hpp:21
#8  0x0000000000414fbf in mlpack::regression::SoftmaxRegression::Train<mlpack::optimization::GradientDescent> (this=0x7fffffffde10, data=..., labels=..., numClasses=4, optimizer=...)
    at /usr/include/mlpack/methods/softmax_regression/softmax_regression_impl.hpp:64
#9  0x000000000040add6 in mlpackMain () at mlpack.cpp:33
#10 0x000000000040aa2a in main (argc=2, argv=0x7fffffffe448) at /usr/include/mlpack/core/util/mlpack_main.hpp:68```

**Comment by bqv** [comment]
Further digging shows this error can be changed to `Mat::init(): requested size is too large` by changing numClasses in the initialization call of smr to 0

**Comment by bqv** [comment]
Further yet digging shows uninitialized memory regions in valgrind when run on the following code with a much smaller dataset:
```
#include <mlpack/core.hpp>
[...]

PR #1382, opened 2018-04-26
Warn differences in OpenMP settings.

rcurtin:
If the user is compiling their application with OpenMP but mlpack was not compiled with OpenMP (or vice versa) we issue a warning.  This situation can cause disaster with recent versions of Armadillo.

This solves #1358.  (Or, at least, it tells the user what is wrong in situations that can lead to #1358.)

### DIFF
--- a/CMake/CreateArmaConfigInfo.cmake
+++ b/CMake/CreateArmaConfigInfo.cmake
@@ -47,6 +47,15 @@ else()
   set(ARMA_64BIT_WORD_DEFINE "#define MLPACK_ARMA_64BIT_WORD")
 endif()
 
+# Next we need to know if we are compiling with OpenMP support.
+# Other places in the CMake configuration should have already done the
+# find(OpenMP).
+if (OPENMP_FOUND)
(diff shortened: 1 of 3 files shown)


zoq:
I don't think there is anything wrong with the style, if you agree I'll go ahead and whitelist this file, to avoid any issue with further PR's.

rcurtin:
Agreed, I think we have to whitelist the file unfortunately.  I'll go ahead and merge this now then.

eyildiz-ugoe:
What is the solution to this problem? I keep getting it even after I used `cmake -DUSE_OPENMP=OFF ..`. 

```
/usr/local/include/mlpack/core/util/arma_config_check.hpp:55:23: note: #pragma message: mlpack was compiled with OpenMP support, but you are compiling without OpenMP support.  This will almost certainly cause irreparable disaster.  Either enable OpenMP support in your application (e.g., add -fopenmp to your compiler command line), or, recompile mlpack *without* OpenMP support.
       #pragma message "mlpack was compiled with OpenMP support, but you are \
                       ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~   

```

PR #1385, opened 2018-04-30
Add JOSS paper for submission.

rcurtin:
This adds the mlpack JOSS paper to the mlpack repository, which is required for the submission to JOSS:  https://github.com/openjournals/joss-reviews/issues/703

I think that we could either remove it or leave it in the repository after the submission is done.


rcurtin:
Ok, I will go ahead and merge and update the joss-reviews issue.

PR #1386, opened 2018-05-01
Transposed Convolutional Layer Implementation

ShikharJ:
This follows from the current implementation of `Convolution` class and a lot of changes need to be discussed (and made) here. Transposed Convolutional Layers are also called Deconvolutional Layers or Fractional Strided Convolutional Layers.


ShikharJ:
Yes, currently, we just need to make the changes to the `Forward`, `Backward` and `Gradient` routines, rest all should remain unchanged.

src/mlpack/tests/transposed_convolutional_network_test.cpp
```diff
@@ -0,0 +1,121 @@
...
+using namespace mlpack;
+using namespace mlpack::ann;
+using namespace mlpack::optimization;
+
+BOOST_AUTO_TEST_SUITE(TransposedConvolutionalNetworkTest);
+
+/**
+ * Train the vanilla network on a larger dataset.
+ */
+BOOST_AUTO_TEST_CASE(VanillaNetworkTest)
```
zoq: What do you think about adding a test that compares against some precomputed results? Just to test the ``TransposedConvolutional`` layer, also a gradient check test might be a good idea to add.

ShikharJ: @zoq Could you provide an example from the code where something like that is done? That would make it easier for me to implement the test and debug the current implementation, then I can go ahead and correct the `transposed_network_test.cpp` (which is currently just a copy of the `convolutional_network_test.cpp`).

zoq: gradient check for the linear layer:

https://github.com/mlpack/mlpack/blob/e37ceb380c4cfb7559916b795b66231f6e2f90f9/src/mlpack/tests/ann_layer_test.cpp#L540-L578

Convolution test against a precomputed result:

https://github.com/mlpack/mlpack/blob/e37ceb380c4cfb7559916b795b66231f6e2f90f9/src/mlpack/tests/convolution_test.cpp#L157-L189

ShikharJ:
@zoq Can you help me in reviewing this? I'm not really confident with the current implementation.

zoq:
Do you think you could add a test that checks the forward pass against a pre-computed result, that would be helpful in the review process.

ShikharJ:
Yeah sure, give me some time, I messed up my `Tensorflow` GPU installation on my system after updating `Ubuntu`.

ShikharJ:
Yeah sure, give me some time, I messed up my `Tensorflow` GPU installation on my system after updating `Ubuntu`.

ShikharJ:
@zoq Nice suggestion :) I was able to debug the `Forward` routine using the pre-computed result. I'll add more tests tomorrow. By the way, how can we obtain the pre-computed result for a `Backward` routine using `Tensorflow`?

zoq:
Awesome, ``tf.gradients()`` is probably the way to go: https://stackoverflow.com/questions/35226428/how-do-i-get-the-gradient-of-the-loss-at-a-tensorflow-variable we can also manually calculate the derivative, wolfram alpha is often useful as well. 

we can also manually calculate the

src/mlpack/methods/ann/layer/transposed_convolution_impl.hpp
```diff
@@ -0,0 +1,299 @@
...
+      inputTemp.n_cols, inputTemp.n_slices);
+  std::cout << "gTemp: " << std::endl << gTemp << std::endl;
+
+  for (size_t outMap = 0, outMapIdx = 0; outMap < outSize; outMap++)
+  {
+    for (size_t inMap = 0; inMap < inSize; inMap++, outMapIdx++)
+    {
+      arma::Mat<eT> output;
+      ForwardConvolutionRule::Convolution(mappedError.slice(outMap),
+          weight.slice(outMapIdx), output, 1, 1);
```
ShikharJ: @zoq I wasn't able to get `Tensorflow` to provide me with the values for the `Backward` routine. The implementation is most likely correct (Forward is completely debugged and tested), unless we need to rotate the weight matrix here again. Can you think of an alternative way of verifying?

ShikharJ: I am yet to check for Gradients on all the test cases. Should be done in a few hours :)

zoq: Sounds good 👍

ShikharJ:
@zoq This is ready for a review.

src/mlpack/methods/ann/layer/convolution.hpp
```diff
@@ -61,7 +61,7 @@ class Convolution
    * @param dH Stride of filter application in the y direction.
    * @param padW Padding width of the input.
    * @param padH Padding height of the input.
-   * @param inputWidth The widht of the input data.
+   * @param inputWidth The width of the input data.
```
zoq: Nice catch, thanks!

src/mlpack/methods/ann/layer/transposed_convolution.hpp
```diff
@@ -0,0 +1,345 @@
...
+           size_t wPad,
+           size_t hPad,
+           arma::Cube<eT>& output)
+  {
+    output = arma::zeros(input.n_rows + wPad * 2,
+        input.n_cols + hPad * 2, input.n_slices);
+
+    for (size_t i = 0; i < input.n_slices; ++i)
+    {
+      Pad<double>(input.slice(i), wPad, hPad, output.slice(i));
```
zoq: Should this be ``Pad<eT>``?

src/mlpack/methods/ann/layer/transposed_convolution_impl.hpp
```diff
@@ -0,0 +1,271 @@
...
+    OutputDataType
+>::Forward(const arma::Mat<eT>&& input, arma::Mat<eT>&& output)
+{
+  inputTemp = arma::cube(input.memptr(), inputWidth, inputHeight, inSize);
+
+  outputWidth = TransposedConvOutSize(inputWidth, kW, dW, padW);
+  outputHeight = TransposedConvOutSize(inputHeight, kH, dH, padH);
+
+  outputTemp = arma::zeros<arma::Cube<eT> >(outputWidth, outputHeight,
+                                            outSize);
```
zoq: This should be tabbed twice, I know there are some inconsistencies over the codebase, but I think this would align with the style guide.

src/mlpack/tests/ann_layer_test.cpp
```diff
@@ -1438,4 +1438,166 @@ BOOST_AUTO_TEST_CASE(GradientBatchNormLayerTest)
...
+
+  TransposedConvolution<> module1(1, 1, 3, 3, 1, 1, 0, 0, 4, 4);
+  // Test the Forward function.
+  input = arma::linspace<arma::colvec>(0, 15, 16);
+  module1.Parameters() = arma::mat(9, 1, arma::fill::zeros);
+  module1.Parameters()(0) = 1.0;
+  module1.Parameters()(8) = 2.0;
+  module1.Reset();
+  module1.Forward(std::move(input), std::move(output));
+  BOOST_REQUIRE_EQUAL(arma::accu(output), 360.0);
```
zoq: Perhaps we should mentioned that the expected value was manually calculated.

src/mlpack/tests/convolution_test.cpp
```diff
@@ -73,7 +73,7 @@ void Convolution3DMethodTest(const arma::cube input,
   arma::cube convOutput;
   ConvolutionFunction::Convolution(input, filter, convOutput);
 
-  // Check the outut dimension.
+  // Check the output dimension.
```
zoq: Another nice catch, thanks!

src/mlpack/methods/ann/layer/transposed_convolution.hpp
```diff
@@ -0,0 +1,345 @@
+/**
+ * @file transposed_convolution.hpp
+ * @author Marcus Edel
+ * @author Shikhar Jaiswal
```
zoq: I think you should put your name on the top.

src/mlpack/methods/ann/layer/transposed_convolution.hpp
ShikharJ: Ah, yes, didn't look much into it. This should be updated in `convolution.hpp` as well.

zoq: Absolutely, would be great if you could include this in the PR.

src/mlpack/methods/ann/layer/transposed_convolution_impl.hpp
```diff
@@ -0,0 +1,271 @@
...
+      BackwardConvolutionRule::Convolution(inputTemp.slice(inMap),
+          rotatedFilter, convOutput, 1, 1);
+
+      outputTemp.slice(outMap) += convOutput;
+    }
+
+    outputTemp.slice(outMap) += bias(outMap);
+  }
+
+  output = arma::Mat<eT>(outputTemp.memptr(), outputTemp.n_elem, 1);
```
rcurtin: Would it be better to do `output = std::move(outputTemp)` possibly followed by a reshape?  I think there's no need for a copy here, but if you use the advanced constructor that makes an alias, the memory owned by `outputTemp` will go out of scope as soon as the method returns.

src/mlpack/methods/ann/layer/transposed_convolution_impl.hpp
```diff
@@ -0,0 +1,271 @@
...
+      arma::Mat<eT> output;
+
+      ForwardConvolutionRule::Convolution(mappedError.slice(outMap),
+          weight.slice(outMapIdx), output, 1, 1);
+
+      gTemp.slice(inMap) += output;
+    }
+  }
+
+  g = arma::mat(gTemp.memptr(), gTemp.n_elem, 1);
```
rcurtin: Same comment applies here.

src/mlpack/methods/ann/layer/transposed_convolution_impl.hpp
zoq: Ahh,  nice catch, the same applies to the convolution layer.

ShikharJ: This wouldn't work as `output` is a `arma::mat` and `outputTemp` is a `arma::cube`.

zoq:
Thanks for a another great contribution.

src/mlpack/methods/ann/layer/transposed_convolution_impl.hpp
ShikharJ: @rcurtin Reshaping first worked for this. I'd prefer this method to aliasing, because of the difference in types.

PR #1391, opened 2018-05-11
Stochastic Optimization for PCA.

zoq:
Implementation of Stochastic PCA as described in "Stochastic Optimization for PCA and PLS", R. Arora et al.


src/mlpack/methods/pca/decomposition_policies/stochastic_method.hpp
```diff
@@ -0,0 +1,323 @@
...
+ * Implementation of the Stochastic Policy. The stochastic approximation
+ * algorithm is an iterative algorithm, where in each iteration a single
+ * sampled point is used to perform an update as done in Stochastic Gradient
+ * Descent. For PCA this means iteratively using vectors sampled from a
+ * distribution to update the subspace.
+ *
+ * For more information, see the following.
+ *
+ * @code
+ * @inproceedings{Musco2015,
```
rcurtin: Maybe this should be `arora2012`?  I met one of the Musco brothers at NIPS last year, but I can't remember which one (probably Cameron?).  A fun person to talk to. :)

src/mlpack/methods/pca/decomposition_policies/stochastic_method.hpp
```diff
@@ -0,0 +1,323 @@
...
+    }
+
+    /**
+     * Shuffle the order of function visitation. This may be called by the
+     * optimizer.
+     */
+    void Shuffle()
+    {
+      visitationOrder = arma::shuffle(
+          arma::linspace<arma::Row<size_t> >(0, data.n_cols - 1, data.n_cols));
```
rcurtin: It might give a nice speedup to shuffle the entire data matrix instead of keeping a `visitationOrder`.

zoq:
Agreed stochastic methods can provide faster solutions, in this case, the runtime highly depends on the number of optimization steps (MaxIterations), since there is no condition to stop early; so this has to be tailored down for each task.

Besides the paper, what I really like is the combination of the existing optimization framework with the PCA class.

Timings:

```
data size: 100x100 rank: 100
----------------------------
ExactSVDPolicy: 6.29743 ms
RandomizedBlockKrylovSVDPolicy: 14.3532 ms
RandomizedSVDPolicy: 12.5829 ms
StochasticAdamPolicy (Iterations 300): 33.0697 ms
StochasticAdamPolicy (Iterations 100): 16.1772 ms

data size: 1000x1000 rank: 100
----------------------------
ExactSVDPolicy: 1039.41 ms
RandomizedBlockKrylovSVDPolicy: 757.691 ms
RandomizedSVDPolicy: 143.457 ms
StochasticAdamPolicy (Iterations 300): 422.688 ms
StochasticAdamPolicy (Iterations 100): 225.264 ms

data size: 5000x5000 rank: 100
----------------------------
ExactSVDPolicy: 119041 ms
RandomizedBlockKrylovSVDPolicy: 10375.3 ms
RandomizedSVDPolicy: 3881.91 ms
StochasticAdamPolicy (Iterations 300): 6435.66 ms
StochasticAdamPolicy (Iterations 100): 2148.02 ms

data size: 4000x8000 rank: 100
----------------------------
ExactSVDPolicy: 154379 ms
RandomizedBlockKrylovSVDPolicy: 7940.06 ms
RandomizedSVDPolicy: 2792.8 ms
StochasticAdamPolicy (Iterations 300): 3306.98 ms
StochasticAdamPolicy (Iterations 100): 2605.95 ms

data size: 6000x12000 rank: 100
----------------------------
ExactSVDPolicy: 530332 ms
RandomizedBlockKrylovSVDPolicy: 14385.7 ms
RandomizedSVDPolicy: 5892.35 ms
StochasticAdamPolicy (Iterations 300): 5990.87 ms
StochasticAdamPolicy (Iterations 100): 4726.83 ms

data size: 7000x19000 rank: 100
----------------------------
ExactSVDPolicy: 839472 ms
RandomizedBlockKrylovSVDPolicy: 25140.6 ms
RandomizedSVDPolicy: 10154.6 ms
StochasticAdamPolicy (Iterations 300): 9784.8 ms
StochasticAdamPolicy (Iterations 100): 8343.91 ms
```

src/mlpack/methods/pca/decomposition_policies/stochastic_method.hpp
zoq: You are absolutely right.

src/mlpack/methods/pca/decomposition_policies/stochastic_method.hpp
```diff
@@ -0,0 +1,299 @@
...
+  template<typename T>
+  inline typename std::enable_if<
+      HasStepSize<T, double&(T::*)()>::value, void>::type
+  SetStepSize(T& optimizer)
+  {
+    if (optimizer.StepSize() > 0)
+    {
+      Log::Warn << "StepSize(): invalid value (> 0), automatically take the "
+          << "negative direction." << std::endl;
+      optimizer.StepSize() *= -1;
```
rcurtin: Hmm, would it be easier to simply negate the objective function?  The formulation of stochastic PCA here is a maximization, but the mlpack optimizers minimize.  So if you just use the negative objective (and derive the gradient accordingly) I think that removes the need for this.

src/mlpack/methods/pca/decomposition_policies/stochastic_method.hpp
```diff
@@ -0,0 +1,299 @@
...
+     */
+    void Shuffle()
+    {
+      // Generate ordering.
+      arma::uvec ordering = arma::shuffle(arma::linspace<arma::uvec>(0,
+          data.n_cols - 1, data.n_cols));
+      arma::mat newData = data.cols(ordering);
+
+      // If we are an alias, make sure we don't write to the original data.
+      math::ClearAlias(data);
```
rcurtin: Hm, I think that even if `data` is an alias, when we call `data = std::move(newData)` then `data` will no longer be an alias since it just takes the memory pointer from `newData` (which is not an alias).  I am not 100% sure but about 95% sure on that, based on looking at the Armadillo matrix `operator=` for rvalue references.

src/mlpack/methods/pca/decomposition_policies/stochastic_method.hpp
```diff
@@ -0,0 +1,299 @@
...
+     *
+     * @param coordinates The function coordinates.
+     * @param begin The first function.
+     * @param batchSize Number of points to process.
+     */
+    double Evaluate(const arma::mat& /* coordinates */,
+                    const size_t /* begin */,
+                    const size_t /* batchSize */)
+    {
+      return pl;
```
rcurtin: I'm a bit confused about the psuedoloss here... I think you are trying to avoid the computation of the objective entirely (so maybe my comment about negative the objective makes no sense, since I wrote it before I got here!), but I'd be concerned that the optimizer might terminate early because it detects convergence.

From equation 1, I guess that the objective is just `tr(U^T x x^T U)`.  If we have some `EvaluateWithGradient()`, then we only have to do the extra computation of pre-multiplying `U^T` to the term we take in the gradient, and then taking the trace of that.  So maybe it is not that much extra, but I could see that it might still make a difference.

What do you think, is there any problem computing the actual loss?  I am not opposed to avoiding the objective calculation if it's not any problem.

src/mlpack/methods/pca/decomposition_policies/stochastic_method.hpp
```diff
@@ -0,0 +1,299 @@
...
+   *
+   * @param iterate Parameters that minimize the function.
+   */
+  template<typename... Targs>
+  void Update(arma::mat& iterate, Targs... Fargs)
+  {
+    updatePolicy.Update(iterate, Fargs...);
+
+    arma::mat R;
+    arma::qr_econ(iterate, R, iterate);
```
rcurtin: Just to check that I understand correctly, I believe that this is the renormalization step that isn't necessary every iteration.  If that's the case, no need to respond, then I got it. :)

src/mlpack/methods/pca/decomposition_policies/stochastic_method.hpp
```diff
@@ -0,0 +1,299 @@
...
+  }
+
+  /**
+   * General update step, all parameters are forwarded to the actual
+   * update policy.
+   *
+   * @param iterate Parameters that minimize the function.
+   */
+  template<typename... Targs>
+  void Update(arma::mat& iterate, Targs... Fargs)
```
rcurtin: I guess if you want to be really picky it should be `fargs` (or `fArgs`?) not `Fargs`, but I dunno if you want to handle template pack parameters differently. :)

src/mlpack/methods/pca/decomposition_policies/stochastic_method.hpp
zoq: Right, at this point I copied:

https://github.com/mlpack/mlpack/blob/e3fe135459593bf5276b3a2794513a9de2ec1c51/src/mlpack/methods/logistic_regression/logistic_regression_function_impl.hpp#L68-L80

rcurtin: (Sorry for the slow response here.)  I see, I guess I originally wrote this code.  I think that maybe we could omit the `ClearAlias()` call in both places, but I suppose it's not a huge deal either way---up to you if you want to make the change.

src/mlpack/methods/pca/decomposition_policies/stochastic_method.hpp
```diff
@@ -271,7 +259,8 @@ class PCAUpdate
   template<typename... Targs>
   void Update(arma::mat& iterate, Targs... fArgs)
   {
-    updatePolicy.Update(iterate, Fargs...);
+    iterate *= -1;
```
rcurtin: I guess, I am still just confused about this bit---shouldn't this be unnecessary if we simply negate the results of `Evaluate()` and `EvaluateWithGradient()` and `Gradient()`?  Sorry if I am overlooking something---it's possible I've misunderstood.  It just seems like extra computation here.

src/mlpack/methods/pca/decomposition_policies/stochastic_method.hpp
```diff
@@ -0,0 +1,287 @@
...
+  /**
+   * General update step, all parameters are forwarded to the actual
+   * update policy.
+   *
+   * @param iterate Parameters that minimize the function.
+   */
+  template<typename... Targs>
+  void Update(arma::mat& iterate, Targs... fArgs)
+  {
+    iterate *= -1;
```
rcurtin: (Oops, I put this in the wrong place originally.)

I guess, I am still just confused about this bit---shouldn't this be unnecessary if we simply negate the results of Evaluate() and EvaluateWithGradient() and Gradient()? Sorry if I am overlooking something---it's possible I've misunderstood. It just seems like extra computation here.

rcurtin:
I think this needs transition to ensmallen before merge, but should be otherwise good to go. :+1:

src/mlpack/methods/pca/decomposition_policies/stochastic_method.hpp
rcurtin: Ah, sorry that I never responded to this.  I follow now, sorry for the confusion.

zoq:
Agreed, let's see if I can do this on the next days.

zoq:

zoq:

zoq:

rcurtin:
@zoq what are you thinking for this one?  I do think it would be nice to get merged, but not sure how much time you have to work on it.

zoq:
Let's see if I can do the necessary changes in the next days.

zoq:
This is ready, so let's keep this open.

birm:
Bumping this; is this still ready?

rcurtin:
Hey @zoq, this one got approved a long time ago but not merged---is everything still ready?  If so, if you want to merge master in we can go ahead and merge it in. :+1:

shrit:
Agreed this is a cool feature to merge before releasing mlpack 4, @zoq great work :100:

shrit:
I have marked this for mlpack 4. If you got a chance to merge it before that would be great :+1:

PR #1393, opened 2018-05-14
Convolution adjustements

zoq:
Adjustments:

- adjust the weights size inside the ann layer test, to avoid memory issue
- use advanced constructor inside the convolution method

``std::move`` should work as well, we could get back to the original solution if we decide that's cleaner.

### DIFF
--- a/src/mlpack/tests/ann_layer_test.cpp
+++ b/src/mlpack/tests/ann_layer_test.cpp
@@ -1448,7 +1448,7 @@ BOOST_AUTO_TEST_CASE(SimpleTransposedConvolutionLayerTest)
   TransposedConvolution<> module1(1, 1, 3, 3, 1, 1, 0, 0, 4, 4);
   // Test the Forward function.
   input = arma::linspace<arma::colvec>(0, 15, 16);
-  module1.Parameters() = arma::mat(9, 1, arma::fill::zeros);
+  module1.Parameters() = arma::mat(9 + 1, 1, arma::fill::zeros);
   module1.Parameters()(0) = 1.0;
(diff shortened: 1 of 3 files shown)


src/mlpack/tests/ann_layer_test.cpp
```diff
@@ -1448,7 +1448,7 @@ BOOST_AUTO_TEST_CASE(SimpleTransposedConvolutionLayerTest)
   TransposedConvolution<> module1(1, 1, 3, 3, 1, 1, 0, 0, 4, 4);
   // Test the Forward function.
   input = arma::linspace<arma::colvec>(0, 15, 16);
-  module1.Parameters() = arma::mat(9, 1, arma::fill::zeros);
+  module1.Parameters() = arma::mat(9 + 1, 1, arma::fill::zeros);
```
ShikharJ: I'm assuming this was done to account for the aliasing?

zoq: Actually to account for the bias term:

```c++
weight = arma::cube(weights.memptr(), kW, kH,
        outSize * inSize, false, false);
bias = arma::mat(weights.memptr() + weight.n_elem,
        outSize, 1, false, false);
```

for the test cases ``outSize`` is 1 so, we have to allocate for the bias term as well.

zoq:
It looks like, the static analysis job is confused with the class name ``Convolution``, not sure why, but I think this one should be ignored.

rcurtin:
Ah, right, agreed.  I had not looked closely at the issue, but it does indeed look like the static code analyzer is confused. :)

PR #1388, opened 2018-05-09
Added Pendulum continuous environment of OpenAI Gym

sshkhrnwbie:
This adds the Pendulum environment (https://gym.openai.com/envs/Pendulum-v0/) which is a continuous control environment with no terminal state. This should fix #1322 as once this is merged all the classic control environments from OpenAI Gym will be added to mlpack's reinforcement learning module.

PR #1394, opened 2018-05-15
Make Continuous Mountain Car Environment same as Gym implementation

sshkhrnwbie:
This PR make the contiunous Mountain Car Environment streamlined with the implementation in OpenAI Gym (https://github.com/openai/gym/blob/master/gym/envs/classic_control/continuous_mountain_car.py) . The changes are : 

- Added a goal position
- The terminal state is reached when car crosses goal position (earlier it was set as when car reaches maximum allowed position)
- The reward for reaching the terminal state is 100 - cost of action (earlier it was just 100)
- The next state calculation is more configurable via introducing the power member variable (This was requested by @zoq in #1368 but not addressed)

I have also added my name to list of contributors following #1388


src/mlpack/methods/reinforcement_learning/environment/continuous_mountain_car.hpp
```diff
@@ -96,13 +96,17 @@ class ContinuousMountainCar
    * @param velocityMax Maximum legal velocity.
    */
   ContinuousMountainCar(const double positionMin = -1.2,
-                        const double positionMax = 0.5,
+                        const double positionMax = 0.6,
```
zoq: Can you add a comment for each new parameter? Also I think you did some good additions to this file, so feel free to add yourself as another author.

src/mlpack/methods/reinforcement_learning/environment/continuous_mountain_car.hpp
```diff
@@ -120,24 +124,25 @@ class ContinuousMountainCar
...
-    // If it is a terminal state, reward is 100.0
+
+    // Calculate reward
+    double reward = 0.0;
+    // If it is a terminal state, add a reward of 100.0
     if (IsTerminal(nextState))
-      return 100.0;
-    return -pow(action.action[0], 2)*0.1;
+      reward = 100.0;
+    reward -= std::pow(action.action[0], 2) * 0.1;
```
zoq: Don't think this makes a huge difference, but I guess it's a good idea to get the same results as the gym env.

zoq:
Thanks again for the contribution 👍

PR #1397, opened 2018-05-17
CF Data Normalization

Wenhao-H:
This PR adds data normalization for the CF module. 

Things to do:
- [x] Add `OverallMeanNormalization`, `UserMeanNormalization`, `ItemMeanNormalization`, `ZNormalization`, `CombinedNormalization`...
- [x] Add tests. Debug. Update comments.
#


Wenhao-H:
I think this can be reviewed now. I will resolve the conflicts after #1297  is merged and add tests for normalization methods.

Wenhao-H:
@lozhnikov No worries. I am waiting for the merge of CF refactoring in another PR, so it will take some time before all codes are completely added.

zoq:
@Wenhao-H do you think you could build against Haritha's branch for now; I'd like to avoid the situation in where you are blocked by another PR. The interface isn't going to change for now, so I think it's somewhat save to work with the latest version.

src/mlpack/methods/cf/normalization/overall_mean_normalization.hpp
```diff
@@ -0,0 +1,111 @@
...
+   */
+  void Normalize(arma::sp_mat& cleanedData)
+  {
+    // Caculate mean of all non zero ratings.
+    if (cleanedData.n_nonzero != 0)
+      mean = arma::accu(cleanedData) / cleanedData.n_nonzero;
+    else
+      mean = 0;
+    // Subtract mean from all non zero ratings.
+    arma::sp_mat::iterator it = cleanedData.begin();
```
zoq: We could put this block into the first condition and remove the else case, since ``cleanedData`` is going to be the same if ``cleanedData.n_nonzero == 0``.

src/mlpack/methods/cf/normalization/user_mean_normalization.hpp
```diff
@@ -0,0 +1,132 @@
...
+  }
+
+  /**
+   * Denormalize computed rating by adding user mean.
+   *
+   * @param user User ID.
+   * @param item Item ID.
+   * @param rating Computed rating before denormalization.
+   */
+  double Denormalize(const int user,
```
zoq: Do you mind to go with ``size_t`` for the index (user), to be consistent with the rest of the codebase?

src/mlpack/methods/cf/normalization/z_score_normalization.hpp
```diff
@@ -0,0 +1,122 @@
...
+    // Caculate mean and stdev of all non zero ratings.
+    arma::vec ratings = arma::nonzeros(cleanedData);
+    mean = arma::mean(ratings);
+    stddev = arma::stddev(ratings);
+
+    // Subtract mean from existing rating and divide it by stddev.
+    arma::sp_mat::iterator it = cleanedData.begin();
+    arma::sp_mat::iterator it_end = cleanedData.end();
+    for (; it != it_end; it++)
+      *it = (*it - mean) / stddev;
```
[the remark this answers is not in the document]
zoq: I guess, if you keep the ``Fatal`` log, this is fine, since we never reach this point if ``stddev < 1e-14``, since ``Fatal``will break afterwards.

src/mlpack/methods/cf/normalization/overall_mean_normalization.hpp
Wenhao-H: Yes we should put the block into the first condition like you suggested. But we still need the else case to set `mean = 0` because `mean` is a class member.

src/mlpack/tests/cf_test.cpp
```diff
@@ -256,14 +299,16 @@ BOOST_AUTO_TEST_CASE(CFPredictTest)
   totalError = std::sqrt(totalError) / savedCols.n_cols;
 
   // The mean squared error should be less than one.
-  BOOST_REQUIRE_LT(totalError, 0.5);
+  BOOST_REQUIRE_LT(totalError, 0.6);
```
Wenhao-H: I don't think `0.6` is a good bound for the test here. If we use RMSE as metric, we would have
```
RMSE = totalError * sqrt(savedCols.n_cols) = totalError * sqrt(50)
````
Let's suppose `totalError = 0.6`, we can get `RMSE = 0.6 * sqrt(50) = 4.24`. For the dataset used in this test, rating range is `[1, 5]`, which means that, even with random prediction, RMSE still won't be larger than `4`.

Comment (zoq) [reply src/mlpack/tests/cf_test.cpp]: Nice catch! We should fix that.

Comment (zoq) [reply src/mlpack/tests/cf_test.cpp]: Should I comment on the CF refactoring PR, or do you like to do this inside this PR?

Comment (Wenhao-H) [reply src/mlpack/tests/cf_test.cpp]: @zoq I'd like to fix this in this PR

Comment (zoq) [reply src/mlpack/tests/cf_test.cpp]: great, thanks!

File: src/mlpack/methods/cf/cf_impl.hpp [hunk 5]
```diff
@@ -148,22 +130,271 @@ void CF::Train(const arma::sp_mat& data,
...
+    std::vector<Candidate> vect(numRecs, def);
+    typedef std::priority_queue<Candidate, std::vector<Candidate>, CandidateCmp>
+        CandidateList;
+    CandidateList pqueue(CandidateCmp(), std::move(vect));
+
+    // Look through the averages column corresponding to the current user.
+    for (size_t j = 0; j < averages.n_rows; ++j)
+    {
+      // Ensure that the user hasn't already rated the item.
+      if (cleanedData(j, users(i)) != 0.0)
```
Wenhao-H: After data normalization, an existing rating could be normalized to zero. (e.g. using OverallMeanNormalization and `r_ij == mean == 3`). Is it a good idea to use another sparse matrix (`arma::sp_umat`) to  save whether a rating is already given in training set?

src/mlpack/methods/cf/cf.cpp
```diff
@@ -1,267 +0,0 @@
...
-  // matrices.
-  arma::umat locations(2, data.n_cols);
-  arma::vec values(data.n_cols);
-  for (size_t i = 0; i < data.n_cols; ++i)
-  {
-    // We have to transpose it because items are rows, and users are columns.
-    locations(1, i) = ((arma::uword) data(0, i));
-    locations(0, i) = ((arma::uword) data(1, i));
-    values(i) = data(2, i);
-    if (values(i) == 0)
```
Wenhao-H: same issue as above.

Wenhao-H:
I modified the `rmse` bound to a very small value and run the test locally, so as to get how different factorizers and different normalization methods perform. Here's the error log:
```
fatal error: in "CFTest/CFPredictRandSVDTest": critical check rmse < 0.1 has failed [4.1485219667956947 >= 0.10000000000000001]
fatal error: in "CFTest/CFPredictRegSVDTest": critical check rmse < 0.1 has failed [1.1190737335013556 >= 0.10000000000000001]
fatal error: in "CFTest/CFPredictBatchSVDTest": critical check rmse < 0.1 has failed [1.1117962645429331 >= 0.10000000000000001]
fatal error: in "CFTest/CFPredictNMFTest": critical check rmse < 0.1 has failed [3.0978957591249361 >= 0.10000000000000001]
fatal error: in "CFTest/CFPredictSVDCompleteTest": critical check rmse < 0.1 has failed [3.0060347850737306 >= 0.10000000000000001]
fatal error: in "CFTest/CFPredictSVDIncompleteTest": critical check rmse < 0.1 has failed [3.1625067960546462 >= 0.10000000000000001]
fatal error: in "CFTest/CFPredictOverallMeanNormalization": critical check rmse < 0.1 has failed [1.035565051556222 >= 0.10000000000000001]
fatal error: in "CFTest/CFPredictUserMeanNormalization": critical check rmse < 0.1 has failed [1.2130446373595365 >= 0.10000000000000001]
fatal error: in "CFTest/CFPredictItemMeanNormalization": critical check rmse < 0.1 has failed [1.0033650716791893 >= 0.10000000000000001]
fatal error: in "CFTest/CFPredictCombinedNormalization": critical check rmse < 0.1 has failed [0.95827935621543858 >= 0.10000000000000001]
```
I have analyzed  in [one of my previous email](http://knife.lugatgt.org/pipermail/mlpack/2018-February/003527.html) why some factorizers (including `NMF, RandSVD, SVDComplete, SVDIncomplete`) give unsatisfactory results. In brief, these factorizers are trained to predict zero where ratings are missing (because matrix entry is implicitly zero in `sp_mat`).

So I think it would be better if the default factorizer of CF is changed to `RegSVD`.

The tests of normalization methods use `NMF`. Thus the results above also show the improvements brought by data normalization.

src/mlpack/methods/cf/cf_impl.hpp
lozhnikov: Hmm, that's an interesting question. I think you could denormalize the item here, however I am not sure it's the most cost-efficient way.

Wenhao-H: @lozhnikov Whether the item is in training set doesn't change the way it is denormalized. So I don't think it works.

lozhnikov: yeah, you're right. There is another option: we could replace every exact zero by `std::numerical_limits<double>::min()` at the normalization step. However this solution looks like a crutch.

Wenhao-H: That should be an efficient way to solve this. Let's use this solution for now and we can switch to the first solution if we run into any issue.

Wenhao-H:
@lozhnikov @zoq All data normalization methods and tests are added. I think this PR is ready to be merged. Can you please review this?

src/mlpack/methods/cf/normalization/combined_normalization.hpp
```diff
@@ -0,0 +1,215 @@
...
+#include <mlpack/prereqs.hpp>
+
+namespace mlpack {
+namespace cf {
+
+/**
+ * This normalization class performs a sequence of normalization methods on
+ * raw ratings.
+ */
+template<typename ...NormalizationTypes>
```
zoq: Can you attache the ellipsis to the typename, in this case we follow https://msdn.microsoft.com/en-us/library/dn439779.aspx.

src/mlpack/methods/cf/normalization/combined_normalization.hpp
```diff
@@ -0,0 +1,215 @@
...
+#define MLPACK_METHODS_CF_NORMALIZATION_COMBINED_NORMALIZATION_HPP
+
+#include <mlpack/prereqs.hpp>
+
+namespace mlpack {
+namespace cf {
+
+/**
+ * This normalization class performs a sequence of normalization methods on
+ * raw ratings.
```
zoq: Can you add a simple example in the comments here something like:

https://github.com/mlpack/mlpack/blob/6a59dd568f0b4c62af52e9813bbaaecc15d8a95d/src/mlpack/methods/quic_svd/quic_svd.hpp#L39-L51

src/mlpack/methods/cf/normalization/item_mean_normalization.hpp
```diff
@@ -0,0 +1,142 @@
...
+    data.each_col([&](arma::vec& datapoint) {
+      const size_t item = (size_t) datapoint(1);
+      const double rating = datapoint(2);
+      itemMean(item) += rating;
+      ratingNum(item) += 1;
+    });
+
+    // Calculate item mean and subtract item mean from ratings.
+    // Set item mean to 0 if the item has no rating.
+    // Should we use mean of all item means if an item has no rating?
```
zoq: I don't think that is necessary.

src/mlpack/methods/cf/normalization/item_mean_normalization.hpp
```diff
@@ -0,0 +1,142 @@
...
+    for (size_t i = 0; i < itemNum; i++)
+      if (ratingNum(i) != 0)
+        itemMean(i) /= ratingNum(i);
+    data.each_col([&](arma::vec& datapoint) {
```
zoq: I think adding a line break makes this easier to read.

src/mlpack/tests/cf_test.cpp
```diff
@@ -235,12 +273,17 @@ BOOST_AUTO_TEST_CASE(CFPredictTest)
...
-  CF::CleanData(dataset, cleanedData);
+  if (cleanData)
+  {
+    // Make data into sparse matrix.
+    arma::sp_mat cleanedData;
+    CFType<NormalizationType>::CleanData(dataset, cleanedData);
+
+    arma::sp_mat dataset;
+    dataset = cleanedData;
+  }
```
Wenhao-H: I'm not sure if I missed anything because it seems to me that this `if` block doesn't actually do anything.

src/mlpack/methods/cf/cf_impl.hpp
```diff
@@ -148,22 +132,271 @@ void CF::Train(const arma::sp_mat& data,
...
+        continue; // The user already rated the item.
+
+      // Is the estimated value better than the worst candidate?
+      if (averages[j] > pqueue.top().first)
```
lozhnikov: I think you should denormalize `avereges[j]` first since `pqueue.top().first` is a denormalized value.

src/mlpack/methods/cf/normalization/item_mean_normalization.hpp
```diff
@@ -0,0 +1,157 @@
...
+  void Normalize(arma::sp_mat& cleanedData)
+  {
+    itemMean = arma::vec(arma::mean(cleanedData, 1));
+
+    arma::sp_mat::iterator it = cleanedData.begin();
+    arma::sp_mat::iterator it_end = cleanedData.end();
+    for (; it != it_end; it++)
+    {
+      *it = *it - itemMean(it.row());
+      // The algorithm omits rating of zero. If normalized rating equals zero,
```
lozhnikov: Could you add the same comment to `GetRecommendations()` and `CleanData()`?

src/mlpack/tests/cf_test.cpp
lozhnikov: Yeah, I think you can remove it.

src/mlpack/methods/cf/normalization/z_score_normalization.hpp
```diff
@@ -0,0 +1,162 @@
...
+namespace mlpack {
+namespace cf {
+
+/**
+ * This normalization class performs z-score normalization on raw ratings.
+ *
+ * An example of how to use ZScoreNormalization in CF is shown below:
+ *
+ * @code
+ * extern arma::mat data; // (user, item, rating) table
```
zoq: Really picky comment, but do you mind to use complete sentences (https://github.com/mlpack/mlpack/wiki/DesignGuidelines#comments) on the example as well?

src/mlpack/methods/cf/normalization/z_score_normalization.hpp
```diff
@@ -0,0 +1,162 @@
...
+    {
+      Log::Fatal << "Standard deviation of all existing ratings is 0! "
+          << "This may indicate that all existing ratings are the same."
+          << std::endl;
+    }
+
+    data.row(2) = (data.row(2) - mean) / stddev;
+    // The algorithm omits rating of zero. If normalized rating equals zero,
+    // it is set to the smallest positive double value.
+    data.row(2).for_each([](double& x) {
```
zoq: This should be on a new line: https://github.com/mlpack/mlpack/wiki/DesignGuidelines#brace-placement.

src/mlpack/methods/cf/normalization/user_mean_normalization.hpp
```diff
@@ -0,0 +1,157 @@
...
+    {
+      const size_t user = combinations(0, i);
+      predictions(i) += userMean(user);
+    }
+  }
+
+  /**
+   * Return user mean.
+   */
+  arma::vec UserMean() const
```
zoq: Do you think we could name this ``mean`` as well? it might be easier if we use the same name for all methods, or do you think this might be confusing? Also what do you think about using:

```c++
const arma::vec& Mean() const { return userMean; }
```

src/mlpack/methods/cf/normalization/user_mean_normalization.hpp
```diff
@@ -0,0 +1,157 @@
...
+      // it is set to the smallest positive double value.
+      if (datapoint(2) == 0)
+        datapoint(2) = std::numeric_limits<double>::min();
+    });
+  }
+
+  /**
+   * Normalize the data by subtracting user mean from each of existing rating.
+   *
+   * @param cleanedData Sparse matrix data.
```
zoq: Perhaps it's a good idea to clarify the comment somewhat; what do you think about: ``Input dataset as a sparse matrix. ``

src/mlpack/methods/cf/normalization/no_normalization.hpp
```diff
@@ -0,0 +1,79 @@
...
+ public:
+  // Empty constructor.
+  NoNormalization() { }
+
+  /**
+   * Do nothing.
+   *
+   * @param data Input dataset in the form of coordinate list.
+   */
+  inline void Normalize(const arma::mat& /* data */) const { }
```
zoq: What about using a template here:

```c++
template<typename MatType>
inline void Normalize(constMatType& /* data */) const { }
```

that way we can remove the overload.

src/mlpack/methods/cf/normalization/no_normalization.hpp
```diff
@@ -0,0 +1,79 @@
...
+   */
+  inline void Denormalize(const arma::Mat<size_t>& /* combinations */,
+                          const arma::vec& /* predictions */) const
+  { }
+
+  /**
+   * Serialization.
+   */
+  template<typename Archive>
+  void serialize(Archive& /* ar */, const unsigned int /* version */) { }
```
zoq: No need to provide the function if it's empty (we use enable_if to detect if the function is available).

src/mlpack/methods/cf/normalization/item_mean_normalization.hpp
```diff
@@ -0,0 +1,157 @@
...
+    {
+      const size_t item = combinations(1, i);
+      predictions(i) += itemMean(item);
+    }
+  }
+
+  /**
+   * Return item mean.
+   */
+  arma::vec ItemMean() const
```
zoq: See the comment about the ``UserMean``.

src/mlpack/methods/cf/normalization/item_mean_normalization.hpp
```diff
@@ -0,0 +1,157 @@
...
+   *
+   * @param combinations User/Item combinations.
+   * @param predictions Predicted ratings for each user/item combination.
+   */
+  void Denormalize(const arma::Mat<size_t>& combinations,
+                   arma::vec& predictions) const
+  {
+    for (size_t i = 0; i < predictions.n_elem; i++)
+    {
+      const size_t item = combinations(1, i);
```
zoq: your choice, but I think we could also just write:

```c++
predictions += combinations.submat(...).
```

src/mlpack/methods/cf/normalization/item_mean_normalization.hpp
```diff
@@ -0,0 +1,157 @@
...
+
+    // Calculate item mean and subtract item mean from ratings.
+    // Set item mean to 0 if the item has no rating.
+    for (size_t i = 0; i < itemNum; i++)
+    {
+      if (ratingNum(i) != 0)
+        itemMean(i) /= ratingNum(i);
+    }
+
+    data.each_col([&](arma::vec& datapoint) {
```
zoq: Can you put the opening ``{`` on a new line: https://github.com/mlpack/mlpack/wiki/DesignGuidelines#brace-placement.

src/mlpack/methods/cf/normalization/combined_normalization.hpp
```diff
@@ -0,0 +1,231 @@
...
+
+  // Empty constructor.
+  CombinedNormalization() { }
+
+  /**
+   * Normalize the data by calling Normalize() in each normalization object.
+   *
+   * @param data Input dataset in the form of coordinate list.
+   */
+  void Normalize(arma::mat& data)
```
zoq: We could use a template here to combine the two methods, see comment below.

src/mlpack/methods/cf/normalization/combined_normalization.hpp
```diff
@@ -0,0 +1,231 @@
...
+  }
+
+  /**
+   * Normalize the data by calling Normalize() in each normalization object.
+   *
+   * @param cleanedData Sparse matrix data.
+   */
+  void Normalize(arma::sp_mat& cleanedData)
+  {
+    SequenceNormalize<0>(cleanedData);
```
zoq: Haven't really thought about it but it might be worth to take a look into ``boost::fusion`` ``for_each``.

src/mlpack/methods/cf/normalization/user_mean_normalization.hpp
Wenhao-H: Yeah it's better to name this `Mean()` and return const reference to the object. I didn't think this through. Thanks!

src/mlpack/methods/cf/normalization/item_mean_normalization.hpp
Wenhao-H: Added:)

src/mlpack/methods/cf/normalization/z_score_normalization.hpp
Wenhao-H: @zoq Should this lambda function be written like:
```C++
data.each_col([&](arma::vec& datapoint)
    {
      const size_t item = (size_t) datapoint(1);
      const double rating = datapoint(2);
      itemMean(item) += rating;
      ratingNum(item) += 1;
    });
```

src/mlpack/methods/cf/normalization/no_normalization.hpp
Wenhao-H:
```
/usr/local/include/boost/serialization/access.hpp:116:11: error: no member named 'serialize' in
      'mlpack::cf::NoNormalization'
        t.serialize(ar, file_version);
```
It generated a compilation error after I removed the `serialize()` function in `NoNormalization`. Any idea why this error occurs?

zoq: Ah, I guess since you call the serialize method inside the ``CombinedNormalization`` class, we can't just remove the function. I guess in this case let's leave the empty function for now.

src/mlpack/methods/cf/normalization/z_score_normalization.hpp
zoq: Like this:

```c++
data.each_col([&](arma::vec& datapoint)
{
  const size_t item = (size_t) datapoint(1);
  const double rating = datapoint(2);
  itemMean(item) += rating;
  ratingNum(item) += 1;
});
```

src/mlpack/methods/cf/normalization/item_mean_normalization.hpp
Wenhao-H: hmm I think we should do something like:
```C++
arma::Col<size_t> items = combinations.row(1);
predictions += itemMean.elem(items);
```
But `X.elem()` takes `arma::uvec` as input and I didn't find a easy conversion from `arma::Col<size_t>` to `arma::uvec`. So unless there is an easy conversion I missed, I think we can stay with the current solution.

src/mlpack/methods/cf/normalization/no_normalization.hpp
Wenhao-H: Right I forgot about that, thanks!

src/mlpack/methods/cf/normalization/combined_normalization.hpp
Wenhao-H: I think it could work but `boost::fusion` isn't one of mlpack's dependencies yet.

src/mlpack/methods/cf/normalization/item_mean_normalization.hpp
zoq: There is http://arma.sourceforge.net/docs.html#conv_to, but I think since this would cause an extra copy, let's stick with the current implementation, thanks for looking into it.

zoq:
Let me know if you need help to solve the travis issue.

src/mlpack/methods/cf/normalization/combined_normalization.hpp
```diff
@@ -0,0 +1,208 @@
...
+  //! Unpack normalizations tuple to serialize.
+  template<
+      int I, /* Which normalization in tuple to serialize */
+      typename Archive,
+      typename = std::enable_if_t<(I < std::tuple_size<TupleType>::value)>>
+  void SequenceSerialize(Archive& ar, const unsigned int version)
+  {
+    std::string tagName = "normalization_";
+    tagName += std::to_string(I);
+    ar & boost::serialization::make_nvp(
```
lozhnikov: Correct me if I'm mistaken, but why don't you write just `ar & BOOST_SERIALIZATION_NVP(normalizations);`? I thought we can serialize tuples.

src/mlpack/methods/cf/normalization/user_mean_normalization.hpp
```diff
@@ -0,0 +1,158 @@
...
+
+  /**
+   * Normalize the data by subtracting user mean from each of existing ratings.
+   *
+   * @param data Input dataset in the form of coordinate list.
+   */
+  void Normalize(arma::mat& data)
+  {
+    const size_t userNum = arma::max(data.row(0)) + 1;
+    userMean = arma::vec(userNum, arma::fill::zeros);
```
lozhnikov: You should replace the datatype by `arma::rowvec` since actually it is a row. That should fix the Travis issue.

src/mlpack/methods/cf/normalization/user_mean_normalization.hpp
```diff
@@ -0,0 +1,158 @@
...
+  }
+
+  /**
+   * Normalize the data by subtracting user mean from each of existing rating.
+   *
+   * @param cleanedData Input data as a sparse matrix.
+   */
+  void Normalize(arma::sp_mat& cleanedData)
+  {
+    userMean = arma::vec(arma::mean(cleanedData, 0));
```
lozhnikov: Same here

src/mlpack/methods/cf/normalization/user_mean_normalization.hpp
```diff
@@ -0,0 +1,158 @@
...
+    {
+      const size_t user = combinations(0, i);
+      predictions(i) += userMean(user);
+    }
+  }
+
+  /**
+   * Return user mean.
+   */
+  const arma::vec& Mean() const { return userMean; }
```
lozhnikov: here

src/mlpack/methods/cf/normalization/user_mean_normalization.hpp
```diff
@@ -0,0 +1,158 @@
...
+   */
+  template<typename Archive>
+  void serialize(Archive& ar, const unsigned int /* version */)
+  {
+    ar & BOOST_SERIALIZATION_NVP(userMean);
+  }
+
+ private:
+  //! User mean.
+  arma::vec userMean;
```
lozhnikov: and here.

src/mlpack/methods/cf/normalization/combined_normalization.hpp
Wenhao-H: Boost Serialization doesn't support tuple serialization (refer to [here](https://stackoverflow.com/questions/14744303/does-boost-support-serialization-of-c11s-stdtuple)), so I wrote an implementation to serialize it.

src/mlpack/methods/cf/cf_impl.hpp
```diff
@@ -148,22 +132,278 @@ void CF::Train(const arma::sp_mat& data,
...
+      // Ensure that the user hasn't already rated the item.
+      // The algorithm omits rating of zero. Thus, when normalizing original
+      // ratings in Normalize(), if normalized rating equals zero, it is set
+      // to the smallest positive double value.
+      if (cleanedData(j, users(i)) != 0.0)
+        continue; // The user already rated the item.
+
+      // Is the estimated value better than the worst candidate?
+      // Denormalize rating before comparison.
+      double realRating = normalization.Denormalize(i, j, averages[j]);
```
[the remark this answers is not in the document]
Wenhao-H: Yes you are right. My mistake. Thanks!!

src/mlpack/methods/cf/normalization/combined_normalization.hpp
lozhnikov: Yeah, you're right. Curious, Boost 1.66 still doesn't support tuples, I didn't know.

Wenhao-H: Yea boost 1.66 still doesn't support tuple serialization. I tried and it fails.

Wenhao-H:
@zoq @lozhnikov Thanks for reviewing! Do you have any further comment?

Wenhao-H:
@zoq @lozhnikov Thanks for reviewing. The travis issue has been solved with the help from Mikhail. Do you have any other comment?

PR #1401, opened 2018-05-20
embedding layer

haritha1313:
Just added an alias for the existing lookup layer, to use it as embedding layer.


zoq:
Thanks again for the contribution, should make the NCF code cleaner.

PR #1410, opened 2018-05-25
CF Interpolation Weights

Wenhao-H:
To do:

- [x] Add `AverageInterpolation`, `SimilarityInterpolation`,  `RegressionInterpolation`.
- [x] Add `LMetricSearch`, `CosineSearch`, `PearsonSearch`.
- [x] Modify methods in `class CF` to use interpolation policies.
- [x] Update tests.


src/mlpack/methods/cf/interpolation_policies/average_interpolation.hpp
```diff
@@ -0,0 +1,51 @@
...
+   * @param similarities similarites between query user and neighbors.
+   */
+  void GetWeights(arma::vec& weights, const arma::vec& similarities) const
+  {
+    if (similarities.n_elem == 0)
+    {
+      Log::Fatal << "Require: similarities.n_elem > 0. There should be at "
+          << "least one neighbor!" << std::endl;
+    }
+    weights.resize(similarities.n_elem);
```
zoq: ``.resize() are considerably slower than .set_size()``, since we fill the matrix anyway, let's use ``set_size(..)`` here.

src/mlpack/methods/cf/interpolation_policies/average_interpolation.hpp
```diff
@@ -0,0 +1,51 @@
...
+class AverageInterpolation
+{
+ public:
+  // Empty constructor.
+  AverageInterpolation() { }
+
+  /**
+   * Interoplation weights are identical and sum up to one.
+   *
+   * @param weights resulting interpolation weights.
```
zoq: Picky style comment, the comment after the parameter should be in capitals.

src/mlpack/methods/cf/interpolation_policies/similarity_interpolation.hpp
```diff
@@ -0,0 +1,56 @@
...
+   */
+  void GetWeights(arma::vec& weights, const arma::vec& similarities) const
+  {
+    if (similarities.n_elem == 0)
+    {
+      Log::Fatal << "Require: similarities.n_elem > 0. There should be at "
+          << "least one neighbor!" << std::endl;
+    }
+
+    double similaritiesSum = arma::sum(similarites);
```
zoq: I'm wondering if something like: `` weights = similarities / (arma::sum(similarites) + eps);`` where ``double eps = 1e-9`` works as well. This should avoid division by zero and probably has no effect on the result.

src/mlpack/methods/cf/neighbor_search_policies/lmetric_search.hpp
```diff
@@ -0,0 +1,70 @@
...
+    // are not larger than one.
+    similarites = 1.0 / (1.0 + similarities);
+  }
+
+ private:
+  //! NeighborSearch object.
+  NeighborSearchType neighborSearch;
+};
+
+using EuclideanSearch = LMetricSearch<2>;
```
zoq: Not sure what the idea behind this is, can you elaborate on this one?

src/mlpack/methods/cf/interpolation_policies/similarity_interpolation.hpp
Wenhao-H: @zoq Yes your implementation is better.

src/mlpack/methods/cf/neighbor_search_policies/lmetric_search.hpp
Wenhao-H: @zoq The current CF performs neighbor search with Euclidean Distance. I am working to modify that to policy-based design and provide alternatives like `EuclideanSearch`, `CosineSearch`, `PearsonSearch` (maybe others as well). The `Search()` method in these classes return `neighbors`, and `similarities` which is used in calculating interpolation weights (i.e. `GetWeights()` method in interpolation classes).
As to `LMetricSearch`, I guess it is possible that a user may want to try other L_p distance besides L_2 (Euclidean Distance), though I haven't met such cases.

zoq: Thanks for the clarification, that makes sense, I was just wondering if a user is going to use the alias and if it's a good idea to put it into the cf namespace.

src/mlpack/methods/cf/neighbor_search_policies/pearson_search.hpp
```diff
@@ -0,0 +1,107 @@
...
+   * the vector to unit length.
+   *
+   * @param Set of reference points.
+   */
+  PearsonSearch(const arma::mat& referenceSet)
+  {
+    // Normalize all vectors in referenceSet.
+    // For each vector x, first subtract mean(x) from each element in x.
+    // Then normalize the vector to unit length.
+    arma::mat normalizedSet(arma::size(referenceSet));
```
zoq: Can we use ``arma:: normalise`` here?

Wenhao-H: Yes we should use `normalise()` instead.

src/mlpack/methods/cf/interpolation_policies/similarity_interpolation.hpp
```diff
@@ -0,0 +1,52 @@
...
+   */
+  void GetWeights(arma::vec& weights, const arma::vec& similarities) const
+  {
+    if (similarities.n_elem == 0)
+    {
+      Log::Fatal << "Require: similarities.n_elem > 0. There should be at "
+          << "least one neighbor!" << std::endl;
+    }
+
+    similarties = similarities / (arma::sum(similarities) + 1e-9);
```
lozhnikov: I think the function should initialize `weights` here.
Moreover, I guess it is better to initialize `weights` to ones if `std::fabs(arma::sum(similarities)) < eps`.

Wenhao-H: @lozhnikov Right, `weights` should be initialized here. Thanks for pointing this out.
Do you think we should initialize `weights` to `1 / similarities.n_elem` if `std::fabs(arma::sum(similarities)) < eps`? Because we restrict similarity to be in `[0, 1]`, in this rare case, all neighbors can be regarded as equally distant from the query vector.

lozhnikov: yeah, sure.

src/mlpack/methods/cf/interpolation_policies/regression_interpolation.hpp
```diff
@@ -0,0 +1,124 @@
...
+   */
+  RegressionInterpolation() { }
+
+  /**
+   * 
+   */
+  RegressionInterpolation(const arma::sp_mat& cleanedData)
+  {
+    const size_t userNum = cleanedData.n_cols;
+    a.set_size(userNum, userNum);
```
zoq: You can also write: ``a.set_size(arma::size(cleanedData))``, just wanted to point that out.

Wenhao-H:
All neighbor search classes and weights interpolation classes are added. I might spend some time later to work on `NNRegressionInterpolation` which is an improvement of `RegressionInterpolation`.

I will templatize some functions in `CFType` to make make use of neighbor search policies and interpolation policies, and add tests for these classes. But as I don't want to introduce conflicts against my another PR on CF data normalization (#1397), I plan to do that after the other PR is merged (and I guess that PR is close to being merged).  At the same time I will try implementing `NNRegressionInterpolation` and run the CF algorithm with newly added data normalization, neighbor search policies and interpolation policies to check performance improvement.

Wenhao-H:
Benchmark results on newly-added policies in CF (including data normalization, neighbor search policies and interpolation policies) can be viewed [here](https://gist.github.com/Wenhao-H/8470c63f5063e18057af42e05c4549e0).

src/mlpack/methods/cf/interpolation_policies/regression_interpolation.hpp
```diff
@@ -0,0 +1,156 @@
...
+      return;
+    }
+
+    for (size_t i = 0; i < neighborNum; i++)
+    {
+      // Calculate coefficient.
+      arma::vec iPrediction;
+      for (size_t j = i; j < neighborNum; j++)
+      {
+        if (a(neighbors(i), a(neighbors(j))) != 0)
```
lozhnikov: Did you mean `a(neighbors(i), neighbors(j))`?

src/mlpack/methods/cf/interpolation_policies/regression_interpolation.hpp
```diff
@@ -0,0 +1,156 @@
...
+        {
+          // The coefficient has already been cached.
+          coeff(i, j) = a(neighbors(i), a(neighbors(j)));
```
lozhnikov: the same comment here

src/mlpack/methods/cf/interpolation_policies/regression_interpolation.hpp
Wenhao-H: Oops made a strange mistake... Thanks!

src/mlpack/tests/cf_test.cpp
```diff
@@ -37,6 +43,66 @@ using namespace mlpack;
...
+static void GetDatasets(arma::mat& dataset, arma::mat& savedCols)
+{
+  // Whether datasets have been initialized.
+  static bool datasetsInitialized = false;
+
+  static arma::mat initDataset;
+  static arma::mat initSavedCols(3, 50);
+
+  // If datasets have been initialized, just pass them to parameters.
+  if (datasetsInitialized)
```
lozhnikov: Are you sure this makes sense? I am not certain, but I think `ctest` runs each test as a separate process. Moreover, it isn't safe to use`static` variables with multi-threading, I guess it is better to avoid them at all even if we aren't going to use multi-threading at all.

src/mlpack/methods/cf/interpolation_policies/regression_interpolation.hpp
```diff
@@ -0,0 +1,156 @@
...
+  {
+    const size_t userNum = cleanedData.n_cols;
+    a.set_size(userNum, userNum);
+    b.set_size(userNum, userNum);
+  }
+
+  /**
+   * The regression-based interpolation problem can be solved by a linear
+   * system of equations. This method first calculates the coefficients and
+   * contant terms for the equations and then solve the equations to get
```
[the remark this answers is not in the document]
Wenhao-H: Yeah I meant `constant`, thanks for the careful review:)

Wenhao-H:
@lozhnikov That's a good solution. Correct me if I'm misunderstand. Let's say we use `DecompositionPolicy` to decompose the rating matrix. To compute a rating we need to call the function `decomposition.GetRating(user, item, ...)`. Therefore, in order to implement BiasSVD and SVD++, we only need to add `BiasSVDPolicy` and `SVDPlusPlusPolicy` as decomposition policies, and provide functions to train the model on given datasets,  to compute rating for a particular user-item pair, etc.

I have one more question about this solution. To refactor `DecompositionPolicy`, should model parameters (`W`, `H`, ...) be saved as members in CFType<> as they are now, or be saved as members in `DecompositionPolicy`?

src/mlpack/methods/cf/cf_impl.hpp
```diff
@@ -284,17 +299,26 @@ double CFType<NormalizationType>::Predict(const size_t user,
...
+      query, numUsersForSimilarity, neighborhood, similarities);
 
-  a.Search(query, numUsersForSimilarity, neighborhood, resultingDistances);
+  arma::vec weights;
 
-  double rating = 0; // We'll take the average of neighborhood values.
+  // Calculate interpolation weights.
+  InterpolationPolicy interpolation(cleanedData);
+  interpolation.GetWeights(weights, w, h, user,
+      neighborhood.col(0), similarities, cleanedData);
```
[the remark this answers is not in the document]
Wenhao-H: Yes you are right:)

src/mlpack/methods/cf/cf_impl.hpp
```diff
@@ -326,12 +351,27 @@ void CFType<NormalizationType>::Predict(const arma::Mat<size_t>& combinations,
...
+      queries, numUsersForSimilarity, neighborhood, similarities);
+
+  arma::mat weights(numUsersForSimilarity, users.n_elem);
+
+  // Calculate interpolation weights.
+  InterpolationPolicy interpolation(cleanedData);
+  for (size_t i = 0; i < users.n_elem; i++)
+  {
+    arma::vec weightVec;
+    interpolation.GetWeights(weightVec, w, h, users[i],
```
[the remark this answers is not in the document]
Wenhao-H: This generates an error:
```
candidate function [with VectorType = arma::subview_col<double>] not viable:
expects an l-value for 1st argument
void GetWeights(VectorType& weights,
```
So I think this doesn't work because `weights.col(i)` returns a r-value.

lozhnikov:
@Wenhao-H Yeah, exactly. So, I think `CFType` won't be just a wrapper since it contains a lot common code for each decomposition policy. Yes, I guess we should move `W` and `H` to `DecompositionPolicy`.

src/mlpack/methods/cf/cf_impl.hpp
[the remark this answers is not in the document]
lozhnikov: Hmm... In that case
```
template<typename VectorType>
void GetWeights(VectorType&& weights, /* other arguments */)
```
should work very well.

zoq:
Agreed, let's move ``W`` and ``H`` into the ``DecompositionPolicy``.

src/mlpack/methods/cf/cf_impl.hpp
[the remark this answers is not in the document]
Wenhao-H: It works, thanks! As we cannot use `set_size()` for `arma::subview_col<double>` object, I removed `weights.set_size(...)` from the body of `GetWeights()`, and set the size of weights before calling `GetWeights()`. Comments reminding user to initialize the size of weights are also added.

Wenhao-H:
Great! I will open another PR to refactor `DecompositionPolicy` and add `BiasSVDPolicy`, `SVDPlusPlusPolicy`.

src/mlpack/tests/cf_test.cpp
Wenhao-H: @lozhnikov I am not familiar with CTest so I searched for some relevant documents. CTest can run different tests in parallel, but all test cases in the same test suite still run in serial. And yes `static` variables are not safe in multithreading context. But as we are not using multithreading here, I think it would be fine to use static variables.
What do you think?

lozhnikov: Well, I measured the time of the test on my system. The whole `CFTest` takes 4.5 sec and 38.37 sec on my system in release and debug mode respectively. I tried to comment out the whole `if` statement and the difference is 0.3 sec and 1.22 sec in release and debug mode respectively. Thus, the difference is relatively small in comparison with the whole `CFTest`, so I suggest to remove static variables.

Wenhao-H: I didn't expect the time difference to be so small. Then we should remove static variables. Thanks for testing the time difference!

src/mlpack/methods/cf/neighbor_search_policies/pearson_search.hpp
```diff
@@ -0,0 +1,105 @@
...
+  /**
+   * Given a set of query points, find the nearest k neighbors, and return
+   * similarities. Similarities are non-negative and no larger than one.
+   *
+   * @param query A set of query points.
+   * @param k Number of neighbors to search.
+   * @param neighbors Nearest neighbors.
+   * @param similarites Similarities between query point and its neighbors.
+   */
+  void Search(const arma::mat& query, const size_t k,
```
zoq: Picky style comment, can you align the parameter with the first one?

src/mlpack/methods/cf/neighbor_search_policies/pearson_search.hpp
```diff
@@ -0,0 +1,105 @@
...
+      arma::Mat<size_t>& neighbors, arma::mat& similarities)
+  {
+    // Normalize all vectors in query.
+    // For each vector x, first subtract mean(x) from each element in x.
+    // Then normalize the vector to unit length.
+    arma::mat normalizedQuery(arma::size(query));
+    for (size_t i = 0; i < query.n_cols; i++)
+    {
+      // Subtract mean(x) from each element in x.
+      normalizedQuery.col(i) = query.col(i) - arma::mean(query.col(i));
```
zoq: It should be possible move this one outside the loop and write:

```c++
normalizedQuery = query - arma::mean(query);
```

I think the same is true for the normalize step.

src/mlpack/methods/cf/neighbor_search_policies/pearson_search.hpp
```diff
@@ -0,0 +1,105 @@
...
+      // Normalize the vector to unit length.
+      normalizedQuery.col(i) = arma::normalise(normalizedQuery.col(i), 2);
+    }
+
+    neighborSearch.Search(normalizedQuery, k, neighbors, similarities);
+
+    // Resulting similarities from Search() are Euclidean distance.
+    // For normalized vectors a and b, pearson(a, b) = 1 - dis(a, b) ^ 2 / 2,
+    // where dis(a, b) is Euclidean distance.
+    similarities = 1 - arma::pow(similarities, 2) / 2.0;
```
zoq: Maybe I missed something, but could we combine this line with the next? That way there is no need to subtract/add 1.

src/mlpack/methods/cf/neighbor_search_policies/cosine_search.hpp
```diff
@@ -0,0 +1,82 @@
...
+  {
+    // Normalize query vectors to unit length.
+    arma::mat normalizedQuery = arma::normalise(query, 2, 0);
+
+    neighborSearch.Search(normalizedQuery, k, neighbors, similarities);
+
+    // Resulting similarities from Search() are Euclidean distance.
+    // For unit vectors a and b, cos(a, b) = 1 - dis(a, b) ^ 2 / 2,
+    // where dis(a, b) is Euclidean distance.
+    similarities = 1 - arma::pow(similarities, 2) / 2.0;
```
zoq: The comment below.

src/mlpack/methods/cf/neighbor_search_policies/pearson_search.hpp
Wenhao-H: Yes you are right. After combining the two lines we have
```c++
similarities = 1 - arma::pow(similarities, 2) / 4.0;
```
But I thought it is better to show how the value is computed by not combining these two steps.

zoq: One option would be to extend the existing comment to make clear where this is coming from, do you think that is reasonable?

Wenhao-H: Yes sure!

src/mlpack/methods/cf/interpolation_policies/regression_interpolation.hpp
```diff
@@ -0,0 +1,171 @@
...
+          if (iPrediction.size() == 0)
+            // Avoid recalculation of iPrediction.
+            iPrediction = w * h.col(neighbors(i));
+          arma::vec jPrediction = w * h.col(neighbors(j));
+          coeff(i, j) = arma::dot(iPrediction, jPrediction) / itemNum;
+          if (coeff(i, j) == 0)
+            coeff(i, j) = std::numeric_limits<double>::min();
+          coeff(j, i) = coeff(i, j);
+          // Cache calcualted coefficient.
+          a(neighbors(i), neighbors(j)) = coeff(i, j);
```
lozhnikov: I am not sure that it is reasonable to cache the coefficients since armadillo stores sparse matrices in CSC format, and populating a sparse matrix via element access operators is inefficient. Moreover you don't take into account that the matrix is symmetric. I am not sure but I think `std::map` could handle that better. How do you think?

src/mlpack/tests/cf_test.cpp
```diff
@@ -1165,4 +1039,70 @@ BOOST_AUTO_TEST_CASE(SerializationCombinedNormalizationTest)
...
+ */
+BOOST_AUTO_TEST_CASE(CFPredictRegressionInterpolation)
+{
+  // The small Grouplens dataset we use is very sparse.
+  // RegressionInterpolation doesn't generalize well when the data
+  // is very sparse. Thus we set a very high rmse bound here.
+  CFPredict<NMFPolicy,
+            OverallMeanNormalization,
+            EuclideanSearch,
+            RegressionInterpolation>(12);
```
lozhnikov: How do you think, maybe it is reasonable to use another dataset for this test?

src/mlpack/methods/cf/interpolation_policies/regression_interpolation.hpp
Wenhao-H: Ah I understand. Thanks for pointing this out. I didn't know how sparse matrix is implemented and I will take a look at CSC format later. I think `std::unordered_map` may be a better choice because it is generally faster than `std::map`.

rcurtin: Actually, this has changed recently (for Armadillo 7+ I believe).  We implemented a 'cache' for Armadillo so that setting sparse matrix elements by, e.g., `x(a, b) = c` is faster.  Basically in the backend this fills out a `std::map<>`, which is synced with the actual CSC representation when necessary.  I don't have any particular opinions on how you do this here, but I thought I'd let you know since I saw the comment.

If you're interested in more details, here is a paper on how it works: http://www.ratml.org/pub/pdf/2018open.pdf

Wenhao-H: @rcurtin Thanks for the clear explanation:)
@lozhnikov In this case I think we can keep using sparse matrix to cache the values, right?

src/mlpack/tests/cf_test.cpp
Wenhao-H: Yeah that would be better. But the problem is I didn't find a dataset which is small and dense enough to be suitable for the test here.

Wenhao-H: @lozhnikov Actually I just found that changing from `NMFPolicy` to `RegSVDPolicy` can reduce `rmse` to around `1.50`. So we can just use `RegSVDPolicy` instead of `NMFPolicy` here.

src/mlpack/methods/cf/interpolation_policies/regression_interpolation.hpp
lozhnikov: @rcurtin Thanks for the clarification, it's interesting.
Actually, looks like the correct link is http://www.ratml.org/pub/pdf/2018user.pdf :)
@Wenhao-H Yeah, sure.

src/mlpack/tests/cf_test.cpp
lozhnikov: @Wenhao-H Let's wait for the test results. Actually, I just wanted to point out some interesting questions. So, we could merge the PR, open an issue and investigate the questions later.

src/mlpack/methods/cf/interpolation_policies/regression_interpolation.hpp
rcurtin: Oops, right, looked like I typed it wrong.  Thanks for the fix :)

src/mlpack/tests/cf_test.cpp
Wenhao-H: Sure!

src/mlpack/tests/cf_test.cpp
```diff
@@ -1165,4 +1039,70 @@ BOOST_AUTO_TEST_CASE(SerializationCombinedNormalizationTest)
...
+
+/**
+ * Make sure that Predict() is returning reasonable results for
+ * RegressionInterpolation.
+ */
+BOOST_AUTO_TEST_CASE(CFPredictRegressionInterpolation)
+{
+  // The small Grouplens dataset we use is very sparse.
+  // RegressionInterpolation doesn't generalize well when the data
+  // is very sparse. Thus we set a very high rmse bound here.
```
lozhnikov: Looks like the comment is out of date now. It's just a reminder. I can fix the comment as soon as I merge the PR.

PR #1412, opened 2018-05-31
Add shuffle parameter to KFoldCV constructor

### LINKED ISSUE #1409: Add shuffle parameter to KFoldCV constructor
(from an IRC discussion) It would be useful to add a `shuffle` flag to the `KFoldCV` constructor, which would shuffle the data before the k-fold cross-validation is done.  This can be important for situations where the data is ordered by class---a k-fold split as-is would fail horribly.

It'd probably be easiest to implement this by just calling `ShuffleData()` inside the constructor for `KFoldCV`.  We should also implement a test, which can basically just be running `KFoldCV` with and without the shuffle option and ensuring the results are close, but not exactly the same.

### LINKED ISSUE #1411: ShuffleData directly reads sparse matrix arrays
Looking at this
https://github.com/mlpack/mlpack/blob/8985af2fd6704dbcd3e8362bea996d5889c2ccd7/src/mlpack/tests/math_test.cpp#L634-L635
leads to:
https://github.com/mlpack/mlpack/blob/8985af2fd6704dbcd3e8362bea996d5889c2ccd7/src/mlpack/core/math/shuffle_data.hpp#L51
which directly reads sparse matrix arrays here:
https://github.com/mlpack/mlpack/blob/8985af2fd6704dbcd3e8362bea996d5889c2ccd7/src/mlpack/core/math/shuffle_data.hpp#L64-L65
This is problematic, as the caching in mechanism in Armadillo 8 means it's not safe to assume that the underlying storage arrays in SpMat are valid:

https://github.com/conradsnicta/armadillo-code/blob/d5a4278bfb919a7b5e3529ca522041010b643b7c/include/armadillo_bits/SpMat_bones.hpp#L48-L49

Possible solutions to fix mlpack's ShuffleData():
1. Instead of directly reading the underlying storage arrays in SpMat, use the [sparse matrix iterators](http://arma.sourceforge.net/docs.html#iterators_spmat).
2. Call SpMat::sync() before reading the underlying storage arrays.

Solution 1 is probably the safer option, as it avoid surprises in the future (eg. potential extensions/modifications to the sparse storage format in Armadillo).

rcurtin:
I've modified the KFoldCV class so that there is an optional 'shuffle' parameter, and made a few other changes also:

 * Better handling of sparse matrices to fix #1411 throughout the codebase.

 * Better handling of datasets where the number of points is not evenly divisible by `k` in `KFoldCV`.  @micyril, if you like, you can take a look at what I did and see if there are any issues.  Basically the modification amounts to holding a `lastBinSize`, since the last cross-validation bin may hold fewer points than the others, and then modifying the `GetTrainingSubset()` and `GetValidationSubset()` functions.

The shuffling is done at the beginning of any call to `KFoldCV::Evaluate()`.

This fixes #1409.

### DIFF
--- a/src/mlpack/core/cv/k_fold_cv.hpp
+++ b/src/mlpack/core/cv/k_fold_cv.hpp
@@ -40,6 +40,10 @@ namespace cv {
  * double softmaxAccuracy = cv.Evaluate(lambda);
  * @endcode
  *
+ * Before calling @c Evaluate(), it is possible to shuffle the data by calling
+ * the @c Shuffle() function.  Shuffling is performed at construction time if
+ * the parameter @c shuffle is set to @c true in the constructor.
+ *
  * @tparam MLAlgorithm A machine learning algorithm.
  * @tparam Metric A metric to assess the quality of a trained model.
  * @tparam MatType The type of data.
@@ -69,9 +73,7 @@ class KFoldCV
    * @param xs Data points to cross-validate on.
    * @param ys Predictions (labels for classification algorithms and responses
    *     for regression algorithms) for each data point.
-   * @param shuffle Whether or not to shuffle the data and predictions before
-   *     performing cross-validation.  Shuffling will be performed before every
-   *     call to Evaluate().
+   * @param shuffle Whether or not to shuffle the data during construction.
    */
   KFoldCV(const size_t k,
           const MatType& xs,
@@ -85,9 +87,7 @@ class KFoldCV
    * @param xs Data points to cross-validate on.
    * @param ys Labels for each data point.
    * @param numClasses Number of classes in the dataset.
-   * @param shuffle Whether or not to shuffle the data and predictions before
-   *     performing cross-validation.  Shuffling will be performed before every
-   *     call to Evaluate().
+   * @param shuffle Whether or not to shuffle the data during construction.
    */
   KFoldCV(const size_t k,
           const MatType& xs,
@@ -104,9 +104,7 @@ class KFoldCV
    * @param datasetInfo Type information for each dimension of the dataset.
    * @param ys Labels for each data point.
    * @param numClasses Number of classes in the dataset.
-   * @param shuffle Whether or not to shuffle the data and predictions before
-   *     performing cross-validation.  Shuffling will be performed before every
-   *     call to Evaluate().
+   * @param shuffle Whether or not to shuffle the data during construction.
    */
   KFoldCV(const size_t k,
           const MatType& xs,
@@ -124,9 +122,7 @@ class KFoldCV
    * @param ys Predictions (labels for classification algorithms and responses
    *     for regression algorithms) for each data point.
    * @param weights Observation weights (for boosting).
-   * @param shuffle Whether or not to shuffle the data and predictions before
-   *     performing cross-validation.  Shuffling will be performed before every
-   *     call to Evaluate().
+   * @param shuffle Whether or not to shuffle the data during construction.
    */
   KFoldCV(const size_t k,
           const MatType& xs,
@@ -143,9 +139,7 @@ class KFoldCV
    * @param ys Labels for each data point.
    * @param numClasses Number of classes in the dataset.
    * @param weights Observation weights (for boosting).
-   * @param shuffle Whether or not to shuffle the data and predictions before
-   *     performing cross-validation.  Shuffling will be performed before every
-   *     call to Evaluate().
+   * @param shuffle Whether or not to shuffle the data during construction.
    */
   KFoldCV(const size_t k,
           const MatType& xs,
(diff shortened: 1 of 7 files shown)


octogman:
Awesome Ryan, thanks for adding data shuffling!

src/mlpack/core/cv/k_fold_cv_impl.hpp
```diff
@@ -297,8 +365,13 @@ arma::Mat<ElementType> KFoldCV<MLAlgorithm,
     arma::Mat<ElementType>& m,
     const size_t i)
 {
-  return arma::Mat<ElementType>(m.colptr(binSize * i), m.n_rows,
-      trainingSubsetSize, false, true);
+  // If this is the last fold, we have to handle it a little bit differently,
+  // since the last fold may not contain 'binSize' points.
+  const size_t subsetSize = (i == k - 1) ? lastBinSize + (k - 2) * binSize :
+      trainingSubsetSize;
```
micyril: If I understand it right, the whole story with `lastBinSize` is to make use of all data when the number of the samples is not divisible by `k` (the previous code will miss some samples during each run in this case). If it's true, then I guess we need to use this "last bin" in a training subset for `i` from `1` to `k - 1` rather than just when `i == k - 1`.  We probably also should get rid of the variable `trainingSubsetSize`, since with this implementation it makes the code more confusing.

src/mlpack/core/cv/k_fold_cv_impl.hpp
```diff
@@ -269,6 +284,58 @@ double KFoldCV<MLAlgorithm,
...
+         typename WeightsType>
+template<bool Enabled, typename>
+void KFoldCV<MLAlgorithm,
+             Metric,
+             MatType,
+             PredictionsType,
+             WeightsType>::ShuffleData()
+{
+  MatType data = xs.cols(0, (k - 1) * binSize + lastBinSize - 1);
+  PredictionsType labels = ys.subvec(0, (k - 1) * binSize + lastBinSize - 1);
```
micyril: Probably it's better to use the name "predictions" instead of "labels" since k-fold cross-validation can be applied to such methods as `LinearRegression`. Also probably it's better to use the method `cols` rather than `subvec` if one day we want to use it with something like `FFN` (which in the current implementation uses `arma::mat` for `responses`).

src/mlpack/core/cv/k_fold_cv.hpp
```diff
@@ -69,10 +69,14 @@ class KFoldCV
    * @param xs Data points to cross-validate on.
    * @param ys Predictions (labels for classification algorithms and responses
    *     for regression algorithms) for each data point.
+   * @param shuffle Whether or not to shuffle the data and predictions before
+   *     performing cross-validation.  Shuffling will be performed before every
+   *     call to Evaluate().
```
micyril: When I was implementing `KFoldCV`, I also was thinking about the need to shuffle the passed data. But I was thinking about doing it once during construction of a `KFoldCV` object rather than before each call of `Evaluate()`. Is it a good practice to do it for each set of hyper-parameters? If so, can you provide some references?

src/mlpack/core/cv/k_fold_cv_impl.hpp
micyril: By reading your initial message for this PR I see that we are probably not on same page about the last cross-validation bin. I think that when we do something like this
```
binSize = source.n_cols / k;
trainingSubsetSize = binSize * (k - 1);
lastBinSize = source.n_cols - ((k - 1) * binSize);
```
`lastBinSize` is actually no less than `binSize`.

src/mlpack/core/cv/k_fold_cv_impl.hpp
rcurtin: Actually I'll just change them to `xsNew` and `ysNew`, there is no confusion there.  And you are right, `cols()` would be better.

src/mlpack/core/cv/k_fold_cv_impl.hpp
rcurtin: Since `source.n_cols / k` is integer math it's effectively `floor(source.n_cols / k)`, so actually I had thought about this a little wrong.  I had been thinking `lastBinSize` would be less than `binSize` but it will actually be up to `k` larger than `binSize`.  But I think that is no problem.

I agree that `trainingSubsetSize` is now confusing since it can change, so I've removed the variable and replaced it with the manual calculation of `binSize * (k - 1)` or `binSize * (k - 2) + lastBinSize`.

src/mlpack/core/cv/k_fold_cv.hpp
rcurtin: Agreed, I was not thinking of the hyper-parameter tuning usage.  So I changed the functionality---now if `shuffle` is specified to the constructor, it shuffles during object construction, and shuffling can be performed again manually with the `Shuffle()` function.

src/mlpack/core/cv/k_fold_cv_impl.hpp
rcurtin: You're right, thanks for the clarification.

rcurtin:
@micyril thanks for the review---I'll go ahead and merge this in 3 days to leave time for any other comments.

PR #1414, opened 2018-06-01
Extend support of Softplus function to matrices, cubes.

akhandait:
It only supported vectors previously.


akhandait:
@sumedhghaisas This is required for the sampling layer.

sumedhghaisas:
@akhandait Nice work :) No comments from side either. This will help us use Softplus in bounding the standard deviation.

zoq:
Thanks again!

PR #1417, opened 2018-06-01
Implemented DCGAN Tests

ShikharJ:
Currently this contains the commits from the GAN PR, which should be merged soon, and then this would be rebased. Just the parameters of the generator network need to be corrected for this to work, which would be done soon.

### DIFF
--- a/src/mlpack/methods/ann/layer/leaky_relu.hpp
+++ b/src/mlpack/methods/ann/layer/leaky_relu.hpp
@@ -46,7 +46,7 @@ class LeakyReLU
  public:
   /**
    * Create the LeakyReLU object using the specified parameters.
-   * The non zero gradient can be adjusted by specifying tha parameter
+   * The non zero gradient can be adjusted by specifying the parameter
    * alpha in the range 0 to 1. Default (alpha = 0.03)
    *
    * @param alpha Non zero gradient
--- a/src/mlpack/tests/dcgan_test.cpp
+++ b/src/mlpack/tests/dcgan_test.cpp
@@ -1,0 +1,280 @@
+/**
+ * @file dcgan_network_test.cpp
+ * @author Shikhar Jaiswal
+ *
+ * Tests the DCGAN network.
+ *
+ * mlpack is free software; you may redistribute it and/or modify it under the
+ * terms of the 3-clause BSD license.  You should have received a copy of the
+ * 3-clause BSD license along with mlpack.  If not, see
+ * http://www.opensource.org/licenses/BSD-3-Clause for more information.
+ */
+#include <mlpack/core.hpp>
+
+#include <mlpack/methods/ann/init_rules/gaussian_init.hpp>
+#include <mlpack/methods/ann/loss_functions/cross_entropy_error.hpp>
+#include <mlpack/methods/ann/loss_functions/sigmoid_cross_entropy_error.hpp>
+#include <mlpack/methods/ann/gan.hpp>
+#include <mlpack/methods/ann/ffn.hpp>
+#include <mlpack/methods/ann/layer/layer.hpp>
+#include <mlpack/methods/softmax_regression/softmax_regression.hpp>
+#include <mlpack/core/optimizers/adam/adam.hpp>
+
+#include <boost/test/unit_test.hpp>
+#include "test_tools.hpp"
+
+using namespace mlpack;
+using namespace mlpack::ann;
+using namespace mlpack::math;
+using namespace mlpack::optimization;
+using namespace mlpack::regression;
+using namespace std::placeholders;
+
+BOOST_AUTO_TEST_SUITE(DCGANNetworkTest);
+
+/*
+ * Tests the DCGAN implementation on the MNIST dataset.
+ * It's not viable to train on bigger parameters due to time constraints.
+ * Please refer mlpack/models repository for the tutorial.
+ */
+BOOST_AUTO_TEST_CASE(DCGANMNISTTest)
+{
+  size_t dNumKernels = 32;
+  size_t discriminatorPreTrain = 5;
+  size_t batchSize = 5;
+  size_t noiseDim = 100;
+  size_t generatorUpdateStep = 1;
+  size_t numSamples = 10;
+  double stepSize = 0.0003;
+  double eps = 1e-8;
+  size_t numEpoches = 1;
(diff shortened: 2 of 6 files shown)


zoq:
Sounds like a good plan to me, thanks!

zoq:
Okay, the ``Gradient``function is calling:

https://github.com/mlpack/mlpack/blob/6e0df8755b3c9ce3f1dccf8adda9d9747dd2874f/src/mlpack/methods/ann/ffn_impl.hpp#L218-L248

from the ``FFN`` class, which runs the Forward/Backward/Gradient function. I'll take a closer look into the issue tomorrow.

zoq:
On a first look

```c++
generator.Add<TransposedConvolution<> >(4 * dNumKernels, 2 * dNumKernels, 4, 4, 2, 2, 1, 1,
      3, 3);
```

returns the wrong size in the backward pass.

ShikharJ:
@zoq Silly me, I must have made a calculation mistake while setting the parameters. Sorry for bothering you on a weekend :)

ShikharJ:
@zoq For the `10,000` image and `20 epoch` run on `MNIST`, seems like `DCGAN` is a bit slower to converge, but atleast the results are nice. Probably we should spend more time looking for hyper-parameters:
mnist_dcgan_20_epoch

zoq:
Promising results, I think the main factor we are trying to do right now is to accelerate the training process right?

ShikharJ:
@zoq Can you also help me in making the Travis build passable?

src/mlpack/tests/dcgan_test.cpp
```diff
@@ -0,0 +1,285 @@
...
+#include "test_tools.hpp"
+
+using namespace mlpack;
+using namespace mlpack::ann;
+using namespace mlpack::math;
+using namespace mlpack::optimization;
+using namespace mlpack::regression;
+using namespace std::placeholders;
+
+// BOOST_AUTO_TEST_SUITE(DCGANNetworkTest);
```
rcurtin: The CMake/CTest configuration will try to run the `DCGANNetworkTest` when this file is added regardless, so you could either comment the file out from `CMakeLists.txt`, uncomment this line (and maybe at least one test will need to be uncommented), or uncomment the whole thing.  I think that will fix the current Travis issue.

ShikharJ: I commented out the file from CMakeLists.txt, it's better than letting the tests run, and potentially timeout the build.

src/mlpack/methods/ann/layer/leaky_relu.hpp
```diff
@@ -46,7 +46,7 @@ class LeakyReLU
  public:
   /**
    * Create the LeakyReLU object using the specified parameters.
-   * The non zero gradient can be adjusted by specifying tha parameter
+   * The non zero gradient can be adjusted by specifying the parameter
```
zoq: Nice catch!

rcurtin:
Should we uncomment the tests before merge?  Maybe you have already had a discussion about this and plan to implement them later or something, so it's possible I am just not aware of the plan. :)

src/mlpack/tests/dcgan_test.cpp
```diff
@@ -0,0 +1,288 @@
...
+  double stepSize = 0.0003;
+  double eps = 1e-8;
+  size_t numEpoches = 1;
+  double tolerance = 1e-5;
+  int datasetMaxCols = 40;
+  bool shuffle = true;
+  double multiplier = 10;
+  std::string output_dataset = "output_dcgan_mnist.csv";
+
+  std::cout << "output_dataset = '" << output_dataset << "'" << std::endl;
```
zoq: hm, either we remove the information or use Log::Info/Log::Debug.

src/mlpack/tests/dcgan_test.cpp
```diff
@@ -0,0 +1,288 @@
...
+  size_t generatorUpdateStep = 1;
+  size_t numSamples = 10;
+  double stepSize = 0.0003;
+  double eps = 1e-8;
+  size_t numEpoches = 1;
+  double tolerance = 1e-5;
+  int datasetMaxCols = 40;
+  bool shuffle = true;
+  double multiplier = 10;
+  std::string output_dataset = "output_dcgan_mnist.csv";
```
zoq: We should use camel casing for all names.

src/mlpack/tests/dcgan_test.cpp
```diff
@@ -0,0 +1,288 @@
...
+    samples = trainData.col(math::RandInt(0, trainData.n_cols));
+    samples.reshape(dim, dim);
+    samples = samples.t();
+
+    generatedData.submat(dim,
+        i * dim, 2 * dim - 1, i * dim + dim - 1) = samples;
+  }
+
+  std::cout << "Saving output to " << output_dataset << "..." << std::endl;
+  generatedData.save(output_dataset, arma::csv_ascii);
```
zoq: In an automated test case, I think we don't need to save the output, but we can comment the lines, and add a comment; I guess someone might find this helpful. Also, I'd like to note that we just test if the the code builds and runs, since we do not take a look at the output.

zoq: If we just use this one to see if the code builds, perhaps we can decrease the number of training iterations?

ShikharJ: The number of iterations is anyways down to the minimum (`1`).

zoq:
```c++
size_t numIterations = trainData.n_cols * numEpoches;
numIterations /= batchSize;
```

isn't ``numIterations > 0`` since ``trainData.n_cols = datasetMaxCols = 40``?

ShikharJ:
@zoq Ah, sorry, I got confused between number of epochs and number of iterations. The reason why I set it at 40 was because I could still get decent outputs of 4s within a single epoch of training (probably because the first 40 images must be images of 4s only). Hence if someone wants to save the generated output and visualize, he can easily do so. Also, it hardly affects the `Travis` runtime.

zoq:
I see, do you think we could reduce the number of iterations, just to slightly reduce the runtime?

ShikharJ:
@zoq This is good to go. The modifications made here would be helpful in implementing `WGAN`.

PR #1420, opened 2018-06-04
[WIP] Add a Sampling layer, implement KL divergence in it 

akhandait:
I will add the backward function for KL divergence soon. @sumedhghaisas


src/mlpack/methods/ann/layer/sampling_impl.hpp
```diff
@@ -0,0 +1,116 @@
...
+  if (inSize != 2 * outSize)
+  {
+    Log::Fatal << "The input size of Sampling layer should be 2 * output size!"
+        << std::endl;
+  }
+}
+
+template <typename InputDataType, typename OutputDataType>
+Sampling<InputDataType, OutputDataType>::Sampling(
+    const size_t sampleSize) :
```
sumedhghaisas: Could we rename this to latentSize instead? Name sampleSize could mean the number of samples as well.

src/mlpack/methods/ann/layer/sampling_impl.hpp
```diff
@@ -0,0 +1,116 @@
...
+namespace ann /** Artificial Neural Network. */ {
+
+template<typename InputDataType, typename OutputDataType>
+Sampling<InputDataType, OutputDataType>::Sampling()
+{
+  // Nothing to do here.
+}
+
+template <typename InputDataType, typename OutputDataType>
+Sampling<InputDataType, OutputDataType>::Sampling(
```
sumedhghaisas: Is there any specific case in which this constructor will be used? I think the constructor which accepts only the size should suffice.

src/mlpack/methods/ann/layer/sampling_impl.hpp
```diff
@@ -0,0 +1,116 @@
...
+void Sampling<InputDataType, OutputDataType>::Forward(
+    const arma::Mat<eT>&& input, arma::Mat<eT>&& output)
+{
+  if (input.n_rows != 2 * outSize)
+  {
+    Log::Fatal << "The output size of layer before the Sampling layer should "
+        << "be 2 * output size of the Sampling layer!" << std::endl;
+  }
+
+  arma::arma_rng::set_seed_random();
```
sumedhghaisas: Setting the seed inside core code should be avoided. This way we could test the 'Forward' function by putting a constant seed in test function. Its user's job to set the seed properly.

src/mlpack/methods/ann/layer/sampling.hpp
```diff
@@ -0,0 +1,168 @@
...
+  size_t& OutputSize() { return outSize; }
+
+  /**
+   * Serialize the layer
+   */
+  template<typename Archive>
+  void serialize(Archive& ar, const unsigned int /* version */);
+
+ private:
+  //! Locally-stored number of input units.
```
sumedhghaisas: Here we could just store 'latentSize' to reduce redundancy. 'inSize' would be 2 * latentSize and 'outSize' will be latentSize

akhandait:
Done ! :)
Do check my reply on the seed review though.

src/mlpack/methods/ann/layer/reparametrization_impl.hpp
```diff
@@ -0,0 +1,100 @@
...
+  output = mean + std::move(stdDeviation) % gaussianSample;
+}
+
+template<typename InputDataType, typename OutputDataType>
+template<typename eT>
+void Reparametrization<InputDataType, OutputDataType>::Backward(
+    const arma::Mat<eT>&& input, arma::Mat<eT>&& gy, arma::Mat<eT>&& g)
+{
+  arma::Mat<eT> softplusDer;
+  SoftplusFunction::Deriv((input - std::move(mean)) / gaussianSample,
```
sumedhghaisas: I think it should be 

SoftplusFunction::Deriv(stddev, softplusDeriv)

src/mlpack/methods/ann/layer/reparametrization_impl.hpp
```diff
@@ -0,0 +1,100 @@
...
+{
+  if (input.n_rows != 2 * latentSize)
+  {
+    Log::Fatal << "The output size of layer before the Reparametrization "
+        << "layer should be 2 * latent size of the Reparametrization layer!"
+        << std::endl;
+  }
+
+  mean = input.submat(latentSize, 0, 2 * latentSize - 1, input.n_cols - 1);
+  SoftplusFunction::Fn(input.submat(0, 0, latentSize - 1, input.n_cols - 1),
```
sumedhghaisas: we need to store the second vector as well

stddev = input.submat(...)

src/mlpack/methods/ann/layer/reparametrization_impl.hpp
```diff
@@ -0,0 +1,100 @@
...
+      stdDeviation);
```
sumedhghaisas: store it directly in 'output'

src/mlpack/methods/ann/layer/reparametrization_impl.hpp
```diff
@@ -0,0 +1,100 @@
...
+
+  gaussianSample = arma::randn<arma::Mat<eT>>(latentSize, input.n_cols);
+  output = mean + std::move(stdDeviation) % gaussianSample;
```
sumedhghaisas: output = mean + std::move(output) % gaussianSample

src/mlpack/methods/ann/layer/reparametrization_impl.hpp
akhandait: Oh sorry. I got confused thinking you wanted to use `submat` here somewhere. This seems correct.

src/mlpack/methods/ann/layer/reparametrization_impl.hpp
akhandait: Will do.

src/mlpack/methods/ann/layer/reparametrization_impl.hpp
akhandait: Yes.

akhandait:
You didn't mention the changes about the delta calculation as you did in irc. Did you forget it here or is mine correct?

src/mlpack/methods/ann/layer/reparametrization_impl.hpp
```diff
@@ -0,0 +1,100 @@
...
+}
+
+template<typename InputDataType, typename OutputDataType>
+template<typename InputType>
+double Reparametrization<InputDataType, OutputDataType>::klForward(
+    const InputType&& input)
+{
+  stdDeviation = input.submat(0, 0, latentSize - 1, input.n_cols);
+
+  return -0.5 * arma::accu(arma::log(stdDeviation) - stdDeviation - arma::pow(
```
sumedhghaisas: I think it should be arma::pow(stdDeviation) rather than just stdDeviation in the second term. Maybe that also affects the 'Backward'?

sumedhghaisas:
Which delta changes? Maybe I wasn't explaining properly on IRC :) You mean the delta coming from KL?

src/mlpack/methods/ann/layer/reparametrization_impl.hpp
sumedhghaisas: I think the 'klBackward' function also need some changes after this change.

akhandait:
I am talking about this g_stddev = gy / sample. I think I need to change that too.

src/mlpack/methods/ann/layer/reparametrization_impl.hpp
akhandait: Yeah we didn't discuss klBackward, what changes will it require?

src/mlpack/methods/ann/layer/reparametrization_impl.hpp
akhandait: I think it needs to be `stdDeviation`, this is when I store standard deviation in the first place, if we use `output` here then we will have to calculate standard deviation everytime.

src/mlpack/methods/ann/layer/reparametrization_impl.hpp
akhandait: Okay, so there has been some confusion. I am not storing it here because we don't consider that the standard deviation. The first part of the input vector after applying softplus gives the standard deviation, because the input can have negative values.

src/mlpack/methods/ann/layer/reparametrization_impl.hpp
akhandait: Understood, will do.

src/mlpack/methods/ann/layer/sampling_impl.hpp
rcurtin: Just FYI, if you do want to set the seed, we have a function `math::RandomSeed()` that will set it for both mlpack and Armadillo.

A bit more information on random seeds---if the user never calls `math::RandomSeed()` or equivalent, then the program that is compiled will have one hard-coded seed that's set at compile time.  A lot of the command-line programs have an option to set the random seed with the `--seed` option, or alternately they will call `math::RandomSeed(std::time(NULL))` to choose a random seed based on the time.

rcurtin:
> @rcurtin @zoq , Maybe the parameter name 'input', should be 'layerOutput' or something in function 'Backward'?

I don't have a huge problem either way.  There is some additional documentation here: http://mlpack.org/docs/mlpack-3.0.1/doxygen/anntutorial.html#layer_api_anntut but maybe it could be clarified a little?

src/mlpack/methods/ann/layer/reparametrization_impl.hpp
sumedhghaisas: The Softplus::Deriv function requires the input to the layer and stdDeviation is the output. We need to store the input and pass it to the Deriv function.

src/mlpack/methods/ann/layer/reparametrization_impl.hpp
```diff
@@ -0,0 +1,99 @@
...
+
+template<typename InputDataType, typename OutputDataType>
+template<typename eT>
+void Reparametrization<InputDataType, OutputDataType>::Backward(
+    const arma::Mat<eT>&& /* input */, arma::Mat<eT>&& gy, arma::Mat<eT>&& g)
+{
+  arma::Mat<eT> softplusDer;
+  SoftplusFunction::Deriv(std::move(stdDeviation), softplusDer);
+
+  g = join_cols(gy % std::move(gaussianSample) % std::move(softplusDer), gy);
```
sumedhghaisas: You need to also add the error signal from kl, 'klBackward' to 'g' here. The current error signal only contains the error from reconstruction, when we add the error from 'kl' we will complete the loss function. The error of 'kl' does not affect the decoder thus no need to add the error signal to 'decoder' layers.

zoq:
> @rcurtin @zoq , Maybe the parameter name 'input', should be 'layerOutput' or something in function 'Backward'?
> 
> I don't have a huge problem either way. There is some additional documentation here: http://mlpack.org/docs/mlpack-3.0.1/doxygen/anntutorial.html#layer_api_anntut but maybe it could be clarified a little?

Same here, if you like to change the parameter name please go ahead, either way we should clarify that in the documentation.

src/mlpack/methods/ann/layer/reparametrization_impl.hpp
akhandait: Yes, in the Deriv we can do that.

src/mlpack/methods/ann/layer/reparametrization_impl.hpp
akhandait: Yes, I will do that in the next commit.

src/mlpack/methods/ann/layer/reparametrization_impl.hpp
```diff
@@ -0,0 +1,104 @@
...
+  mean = input.submat(latentSize, 0, 2 * latentSize - 1, input.n_cols - 1);
+  preStdDev = input.submat(0, 0, latentSize - 1, input.n_cols - 1);
+
+  if (stochastic)
+    gaussianSample = arma::randn<arma::Mat<eT>>(latentSize, input.n_cols);
+  else
+    gaussianSample = arma::ones<arma::Mat<eT>>(latentSize, input.n_cols) * 0.7;
+
+  SoftplusFunction::Fn(preStdDev, output);
+  output = mean + output % gaussianSample;
```
sumedhghaisas: output %= gaussianSample; output += mean; Might save a copy :)

src/mlpack/methods/ann/layer/reparametrization_impl.hpp
```diff
@@ -0,0 +1,104 @@
...
+  SoftplusFunction::Deriv(std::move(preStdDev), g);
+  g = join_cols(gy % std::move(gaussianSample) % g, gy);
+}
+
+template<typename InputDataType, typename OutputDataType>
+template<typename InputType>
+double Reparametrization<InputDataType, OutputDataType>::klForward(
+    const InputType&& input)
+{
+  preStdDev = input.submat(0, 0, latentSize - 1, input.n_cols);
```
akhandait: @sumedhghaisas I think we cannot overwrite `preStdDev` in `klForward`, because in that case, `Backward` will use the overwritten value and not the original one we want it to use.

akhandait:
Sorry for the repeating commits, I need to check before making the commits from now on.

src/mlpack/methods/ann/layer/reparametrization_impl.hpp
```diff
@@ -0,0 +1,117 @@
...
+    g = join_cols(gy % std::move(gaussianSample) % g, gy) + std::move(klBack);
+  }
+  else
+    g = join_cols(gy % std::move(gaussianSample) % g, gy);
+}
+
+template<typename InputDataType, typename OutputDataType>
+template<typename InputType>
+double Reparametrization<InputDataType, OutputDataType>::klForward(
+    const InputType&& input)
```
rcurtin: Can we rename this KLForward() to match the style guide?  Alternately, it is such a small function that it could be inlined into `Loss()` since it is only used there.

The same applies for `klBackward()` too.

src/mlpack/methods/ann/layer/reparametrization_impl.hpp
```diff
@@ -0,0 +1,117 @@
...
+  output = join_cols((-1 / stdDev + stdDev) % output, mean);
+}
+
+template<typename InputDataType, typename OutputDataType>
+template<typename Archive>
+void Reparametrization<InputDataType, OutputDataType>::serialize(
+    Archive& ar, const unsigned int /* version */)
+{
+  ar & BOOST_SERIALIZATION_NVP(latentSize);
+  ar & BOOST_SERIALIZATION_NVP(stochastic);
```
rcurtin: Shouldn't we serialize `includeKl` also?  Maybe I am wrong, so ignore the comment if so.

src/mlpack/tests/ann_layer_test.cpp
```diff
@@ -1847,4 +1836,122 @@ BOOST_AUTO_TEST_CASE(SubviewIndexTest)
...
+  Reparametrization<> module(5, false);
+
+  input = join_cols(arma::ones<arma::mat>(5, 1),
+      arma::zeros<arma::mat>(5, 1));
+
+  // Test if two forward passes generate same output.
+  module.Forward(std::move(input), std::move(outputA));
+  module.Forward(std::move(input), std::move(outputB));
+
+  CheckMatrices(std::move(outputA), std::move(outputB));
```
rcurtin: Technically no need for the `std::move()` in the `CheckMatrices()` call here but it doesn't make a difference, it still works as-is.

src/mlpack/tests/ann_layer_test.cpp
```diff
@@ -1847,4 +1836,122 @@ BOOST_AUTO_TEST_CASE(SubviewIndexTest)
...
 
+/*
+ * Simple Reparametrization module test.
+ */
+BOOST_AUTO_TEST_CASE(SimpleReparametrizationLayerTest)
+{
+  arma::mat input, output, delta;
+  Reparametrization<> module(5);
+
+  // Test the Forward function.
```
rcurtin: Is there any chance I could ask to add some comments here on what the test is testing and what the expected output should be (and why)?  That can make maintenance a lot easier down the road, since whoever the maintainer is won't need to figure out why the test is right.

src/mlpack/tests/ann_layer_test.cpp
```diff
@@ -1847,4 +1836,122 @@ BOOST_AUTO_TEST_CASE(SubviewIndexTest)
...
+}
+
+/**
+ * Reparametrization module includeKl boolean test.
+ */
+BOOST_AUTO_TEST_CASE(ReparametrizationLayerIncludeKlTest)
+{
+  arma::mat input, output, gy, delta;
+  Reparametrization<> module(5, true, false);
+
```
rcurtin: If we can also add some comments here on what we are testing and why the output is what's expected, I think it could be really helpful.

akhandait:
@rcurtin Sorry I didn't follow the style guidelines in some places. I will make these changes soon and open a PR.

src/mlpack/methods/ann/layer/reparametrization_impl.hpp
akhandait: I will make the style changes, sorry for not keeping them in mind. Although, I wanted `klForward()` and `klBackward()` to be seperate from the `Forward()` and `Backward()` of the main Repar layer. For some other reasons I implemented them here. Is that okay? Will not having this inlined into Loss() cause significant performance issues? If yes then I will make the changes.

rcurtin:
No worries, it's ok! :)

For `klForward()` and `klBackward()`, any inlining won't cause enough of a performance difference to notice anywhere.  The compiler is likely to do whatever it needs to make it go faster---the `Loss()` function itself may be inlined entirely also.  So I wouldn't worry about it here, I don't think that any difference could be reasonably observed from our end.

src/mlpack/methods/ann/layer/reparametrization.hpp
```diff
@@ -0,0 +1,162 @@
...
+class Reparametrization
+{
+ public:
+  //! Create the Reparametrization object.
+  Reparametrization();
+
+  /**
+   * Create the Reparametrization layer object using the specified sample vector size.
+   *
+   * @param layerSize The number of output units.
```
zoq: Should this be ``latentSize``?

src/mlpack/methods/ann/layer/reparametrization_impl.hpp
zoq: Depending on the size I think it could be useful to inline the function and to avoid ``join_cols`` and the corresponding split afterwards.

Issue #1419, opened 2018-06-02
[JOSS] Issue in perceptron Python-binding documentation with mlpack class assignment


rasbt:
In the perceptron docs, at http://www.mlpack.org/docs/mlpack-3.0.1/python/perceptron.html

there are some issues in the code examples

```python
>>> from mlpack import perceptron

>>> perceptron(training=training_data, labels=training_labels)
>>> perceptron = output['output_model']

>>> perceptron(input_model=perceptron, test=test_data)
>>> predictions = output['output']
```

First there's the "output" issue referenced in #1418. For now, let's assume that's fixed:

```python
>>> output = perceptron(training=training_data, labels=training_labels)
>>> perceptron = output['output_model']

>>> output = perceptron(input_model=perceptron, test=test_data)
>>> predictions = output['output']
```

Now, the issue is this will raise an `TypeError: 'mlpack.perceptron.PerceptronModelType' object is not callable`. Presumably, because it's somehow trying to use the module `perceptron`. I would recommend not to overwrite the `perceptron` class imported from `mlpack` (earlier in the documentation) and modify it as follows (perceptron -> perceptron_model):

```python
>>> output = perceptron(training=training_data, labels=training_labels)
>>> perceptron_model = output['output_model']

>>> output = perceptron(input_model=perceptron_model, test=test_data)
>>> predictions = output['output']
```

Would probably be a good idea to double-check the other modules as well

rcurtin:
Good call---I have a fix in #1421.  This is one of the tricky bits with writing agnostic documentation: for the command-line, this makes sense.  The model just gets called `perceptron.bin`.  But in Python it goes south fast like you pointed out. :)

I checked all the other bindings too, and none have the same issue.

rcurtin:
Fixed with the merge of #1421.

Issue #1418, opened 2018-06-02
[JOSS] Documentation typo in Adaboost code example


rasbt:
In the Python Bindings docs of Adaboost (http://www.mlpack.org/docs/mlpack-3.0.1/python/adaboost.html), the following code example is provided:

    >>> adaboost(training=data, weak_learner='perceptron')
    >>> model = output['output']

I believe this should be 

    >>> adaboost(training=data, weak_learner='perceptron')
    >>> model = output['output_model']

based on the description at the very bottom: (`output_model (mlpack.AdaBoostModelType): Output trained AdaBoost model.`)?

rcurtin:
You're right about the lack of `output =`, this is a bug in the auto-generated documentation.  It always helps to have an extra set of eyes---thank you for pointing this out.  The `output` -> `output_model` is a specific error in the AdaBoost code.

rasbt:
>  It always helps to have an extra set of eyes

I agree :). Coincidentally, not too long ago, I submitted a ML toolkit library to JOSS as well (mlxtend) and was bumping into similar issues with the autogenerated docs that the reviewer found

rcurtin:
Fixed with #1421.

rcurtin:
#1421 is merged, so I will close this now.

Issue #1416, opened 2018-06-01
[JOSS] how to set up the mlpack Python bindings


rasbt:
I followed the installation instructions and everything went super smoothly thus far! :)

I can use the binaries from within the bin folder, and also the tests all pass:

```bash
~/mlpack-3.0.1/build$ bin/mlpack_test
Running 1351 test cases...

...
warning: solve(): system appears singular (rcond: 1.31214e-18); attempting approximate solution

warning: solve(): system appears singular (rcond: 1.31214e-18); attempting approximate solution

*** No errors detected
```

I am wondering now what step would be required to enable the Python bindings? Would I just need to add `~/mlpack-3.0.1/src/mlpack/bindings/python/mlpack` to my Python `PATH`? It would be nice if that could be briefly explained in the documentation (maybe in the "building from source" docs).

rcurtin:
Honestly I'd say `make install` is the easiest way to do it, like the build directions suggest.  If you really want to do it manually in the build directory it's hard enough that I wanted to avoid adding it to the documentation... from the build directory:

```
export LD_LIBRARY_PATH=lib/
export PYTHONPATH=src/mlpack/bindings/python/
```

If you `make install` it will put it in `/usr/local/lib/pythonX.Y/site-packages/` which should automatically be in the `PYTHONPATH`.

As I write this though, I do think you are right, it would be useful to add a section to the build documentation mentioning how to work with the Python bindings without installing.

rasbt:
> If you make install it will put it in /usr/local/lib/pythonX.Y/site-packages/ which should automatically be in the PYTHONPATH.

Hm, I used that, but for some reason, it didn't add anything to the site-packages directory (checked both the python 2.7 and 3.5 dirs), which is why I assumed a manual step might be necessary (I used `sudo make install`)

```python
>>> import mlpack
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
ImportError: No module named mlpack
```

> As I write this though, I do think you are right, it would be useful to add a section to the build documentation mentioning how to work with the Python bindings without installing. I'll do that when I have a chance.

Thanks, that would be nice.

rcurtin:
That's strange. Mind posting the output of `make install` (it will be long...)?  The Python bindings are pretty fiddly and I've found that different systems behave pretty differently.

(The Python bindings are also available via pip for manylinux platforms but where's the fun in that?? :))

rcurtin:
[...]

PR #1421, opened 2018-06-04
JOSS documentation fixes

rcurtin:
This is a collection of fixes for the issues brought up by @rasbt:

 * Clarify how to use the Python bindings without installation and mention when `LD_LIBRARY_PATH` might be needed. (#1416)
 * Print `output = ` before automatically-generated calls to Python bindings in documentation. (#1418)
 * Change name of `perceptron` variable in the perceptron code documentation. (#1419)
 * Rewrite the kRANN documentation since it was written specifically for command-line bindings and never updated.
 * Write the random forest documentation since it seems it never actually got written.

Since these fix some fairly important documentation issues from the Python angle, I would have no problem doing a release after this merge as mlpack 3.0.2.

### DIFF
--- a/README.md
+++ b/README.md
@@ -106,6 +106,11 @@ with the following command:
 
     $ sudo apt-get install libmlpack-dev
 
+Note: Older Ubuntu versions may not have the most recent version of mlpack
+available---for instance, at the time of this writing, Ubuntu 16.04 only has
+mlpack 2.0.1 available.  Options include upgrading your Ubuntu version, finding
+a PPA or other non-official sources, or installing with a manual build.
+
 There are some other useful pages to consult in addition to this section:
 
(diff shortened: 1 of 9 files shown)


rcurtin:
Ok, I will go ahead and merge this in three days.

rasbt:
That's great, as far as I can tell from the changelog, it addresses all the issues that came up during the JOSS review :).

Maybe I am too early, but in case this is not a temporary thing, the link's (e.g., under Python Bindings) don't work, yet:

```
Not Found

The requested URL /docs/mlpack-3.0.2/python/perceptron.html was not found on this server.
```

rcurtin:
Yep, sorry about that---I was in the process of doing the release and accidentally changed the links on the homepage before the documentation was built, and then was interrupted by dinner and a movie. :)

The documentation works now, and addresses the various issues that you pointed out.  Thank you again!

rasbt:
No worries, and it works fine now!

PR #1428, opened 2018-06-10
subview layer addition

haritha1313:


src/mlpack/methods/ann/layer/subview.hpp
```diff
@@ -0,0 +1,153 @@
...
+   * Ordinary feed forward pass of a neural network, evaluating the function
+   * f(x) by propagating the activity forward through f.
+   *
+   * @param input Input data used for evaluating the specified function.
+   * @param output Resulting output activation.
+   */
+  template<typename InputType, typename OutputType>
+  void Forward(InputType&& input, OutputType&& output)
+  {
+    // Check if input has been selected as required.
```
zoq: I was wondering if we could use one of armadillo's advanced constructors to reuse the memory, something like:

```c++
output = arma::Mat<ElemType>(input.memptr() + begin, 1, end - begin, false, false);
```

src/mlpack/methods/ann/layer/subview.hpp
```diff
@@ -0,0 +1,153 @@
...
+        input.shed_rows(end+1, limit-1);
+      }
+      else
+      {
+        input.shed_rows(0, begin-1);
+      }
+    }
+    arma::Mat<eT> derivative;
+    IdentityFunction::Deriv(input, derivative);
+    g = gy % derivative;
```
zoq: I think we can just forward the ``gy`` so ``g = gy``, that's what the identity function does anyway, maybe I missed anything?

src/mlpack/methods/ann/layer/subview.hpp
haritha1313: Ofcourse, I'll edit it. Just one doubt, shouldn't the `n_rows` parameter be `(end-begin)` and `n_cols` be `input.n_cols`, considering we are passing matrices and not just vectors?

src/mlpack/methods/ann/layer/subview.hpp
haritha1313: You are right, just took a look at `identityfunction.hpp`.

src/mlpack/methods/ann/layer/subview.hpp
zoq: You are right, just thought about vectors.

src/mlpack/methods/ann/layer/subview.hpp
```diff
@@ -0,0 +1,140 @@
...
+   * @param input Input data used for evaluating the specified function.
+   * @param output Resulting output activation.
+   */
+  template<typename InputType, typename OutputType>
+  void Forward(InputType&& input, OutputType&& output)
+  {
+    // Check if input has been selected as required.
+    if ((input.n_rows != (end-begin+1)) && (end != 0))
+    {
+      arma::uword cols = input.n_cols;
```
zoq: I think we could directly use ``input.n_cols``.

src/mlpack/methods/ann/layer/subview.hpp
```diff
@@ -0,0 +1,140 @@
...
+   * @param gy The backpropagated error.
+   * @param g The calculated gradient.
+   */
+  template<typename eT>
+  void Backward(arma::Mat<eT>&& input,
+                arma::Mat<eT>&& gy,
+                arma::Mat<eT>&& g)
+  {
+    // Check if input has been selected as required.
+    if ((input.n_rows != (end-begin+1)) && (end != 0))
```
zoq: Do you think we can remove the if block? Looks like the input isn't used anyway.

src/mlpack/tests/ann_layer_test.cpp
```diff
@@ -1798,4 +1798,23 @@ BOOST_AUTO_TEST_CASE(GradientLayerNormTest)
...
+ */
+BOOST_AUTO_TEST_CASE(SimpleSubviewLayerTest)
+{
+  arma::mat output, input, delta;
+  Subview<> module(10, 19);
+
+  // Test the Forward function.
+  input = arma::ones(20, 1);
+  module.Forward(std::move(input), std::move(output));
+  BOOST_REQUIRE_EQUAL(output.n_rows, 10);
```
zoq: It might be good to test the check the output as well. We could use http://arma.sourceforge.net/docs.html#linspace to create an input of the form ``[1, 2, 3, 4, 5, 6, 7, 8, 9, 10]``, we can also hardcode the input and check if the output does match with the given index. In this case it might be a good idea to check different start and end points.

zoq:

src/mlpack/methods/ann/layer/subview.hpp
```diff
@@ -0,0 +1,133 @@
...
+  //! Modify the delta.
+  OutputDataType& Delta() { return delta; }
+
+  /**
+   * Serialize the layer.
+   */
+  template<typename Archive>
+  void serialize(Archive& /* ar */, const unsigned int /* version */)
+  {
+    /* Nothing to do here */
```
rcurtin: I know I am late to the party here, but should we serialize `begin` and `end`?

src/mlpack/methods/ann/layer/subview.hpp
```diff
@@ -0,0 +1,133 @@
...
+    typename OutputDataType = arma::mat
+>
+class Subview
+{
+ public:
+  /**
+   * Create the Subview layer object using the specified range of input to accept.
+   *
+   * @param begin Start index.
+   * @param end End index.
```
rcurtin: Looks like this works with 1D inputs, but it might be useful to call this `Subview1D`, and perhaps provide some other time a `Subview2D` layer, in case the input is matrix-shaped (not vector-shaped).

zoq: Good point, it should be possible to support both in a single class, so I think we can still call this class ``Subview``.

src/mlpack/methods/ann/layer/subview.hpp
zoq: Right, nice catch, let me fix that.

src/mlpack/methods/ann/layer/subview.hpp
haritha1313: I think it does work for both vector and matrix shaped inputs. Maybe I understood something wrong, but I tried running it on matrices and it is being modified as required.

rcurtin: Right, it works with both, but let's suppose I have a matrix of size 5x5 and what I want to do is take the middle 3x3 elements out of it.  If we use the Armadillo advanced constructors we don't have a way to do that, since the middle 3x3 won't be contiguous in memory.  In that situation we would have no choice but to extract the submatrix with `submat()`.  (I guess it might be possible to pass an arma::subview type to the next layer as input, but I am not sure if that would actually work in the framework we have.  I don't think that avoiding the subview copy in this situation is so important that it has to be dealt with right now though.)

haritha1313: Right, then we can go on with `submat()`. I suppose. @zoq Unfortunately I deleted the branch I was working on, so just let me know if I should restore it and revert the merge to make these changes.

zoq: No need to revert the changes, we can open a new PR to do the modification.

PR #1430, opened 2018-06-13
Return the final objective of the trained model

xtelinco:
Hi,
It is sometimes useful to have the final objective of the trained model, especially if it failed to train properly.

It has been useful for me, letting you have this as it may well be useful for others.

Cheers,

Rob

### DIFF
--- a/src/mlpack/methods/logistic_regression/logistic_regression.hpp
+++ b/src/mlpack/methods/logistic_regression/logistic_regression.hpp
@@ -125,9 +125,10 @@ class LogisticRegression
    * @tparam OptimizerType Type of optimizer to use to train the model.
    * @param predictors Input training variables.
    * @param responses Outputs results from input training variables.
+   * @return The final objective of the trained model (NaN or Inf on error)
    */
   template<typename OptimizerType = mlpack::optimization::L_BFGS>
-  void Train(const MatType& predictors,
+  double Train(const MatType& predictors,
              const arma::Row<size_t>& responses);
 
   /**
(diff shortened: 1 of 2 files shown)


zoq:
Thanks for the contribution, I think it might make sense to return the final objective for every classifier, to be consistent with the rest of the codebase. If I remember right there was this idea to construct something around the optimizer framework which monitors different stats, which might be another solution.

Either way, if we agree on this change this is a good start.

rcurtin:
Hey Rob,

I agree, this could be useful, and I'd be happy to see it merged.  @zoq: do you think we should open an issue to adapt the rest of the classifiers in the same way?  I think there may be some classifiers that might not have any reasonable loss to return (for instance, if we ever implement a kNN classifier, there's no training loss to return).

Here's a test we could add to `src/mlpack/tests/logistic_regression_test.cpp` just as a sanity check:

```
/**
 * Test that LogisticRegression::Train() returns the expected final objective value.
 */
BOOST_AUTO_TEST_CASE(LogisticRegressionTrainReturnObjective)
{
  // Very simple fake dataset.
  arma::mat data("1 2 3;"
                 "1 2 3");
  arma::Row<size_t> responses("1 1 0");

  // The returned objective from training should be close to 0.
  LogisticRegression<> lr(data.n_rows, 0.5);
  BOOST_REQUIRE_SMALL(lr.Train<L_BFGS>(data, responses), 1e-5);

  // Also check with a pre-trained optimizer.
  L_BFGS l;
  BOOST_REQUIRE_SMALL(lr.Train(data, responses, l), 1e-5);
}
```

I haven't run that test, but I'm pretty sure that it should work fine.

zoq:
@rcurtin right, it doesn't make sense to return something in this cases; let's open an issue.

PR #1431, opened 2018-06-13
Batch Support for GANs

ShikharJ:


src/mlpack/methods/ann/gan_impl.hpp
```diff
@@ -124,40 +129,46 @@ double GAN<Model, InitializationRuleType, Noise>::Evaluate(
...
+  // Alias here
+  currentInput = this->predictors.cols(i, i + batchSize - 1);
+  currentTarget = this->responses.cols(i, i + batchSize - 1);
 
   discriminator.Forward(std::move(currentInput));
+  std::cout << "Fine?" << std::endl;
   double res = discriminator.outputLayer.Forward(
       std::move(boost::apply_visitor(
           outputParameterVisitor,
           discriminator.network.back())), std::move(currentTarget));
```
ShikharJ: @zoq I'm having a bit of trouble here. When the above function gets called, the `boost::apply_visitor` routine returns a `1 x 1` object and `currentTarget` is of size `1 x batchSize`, which leads to an error with element-wise multiplication inside `discriminator.outputLayer.Forward` which is the `SigmoidCrossEntropyError::Forward`. Could you help me with this?

ShikharJ: @zoq I'll think over and try out the second approach, as I think we'll have to change the parameters in all the layers for the first approach to work, which is something that would be quite complicated for a prospective user. I'll let you know.

zoq: Okay, I can make the necessary modifications if you like; we just have to adjust the conv layers, the rest already works with a ``batchSize > 1``.

zoq:
Looks like this introduced some index out of bounds error.

```c++
discriminator.responses.cols(numFunctions, numFunctions + batchSize - 1)
```

might be the issue here.

ShikharJ:
@zoq We're facing an issue [here](https://github.com/mlpack/mlpack/blob/master/src/mlpack/methods/ann/ffn_impl.hpp#L326) on iteration `1`. I'm not too comfortable with the visitor pattern implementations, can you help me in fixing this?

zoq:
Inside the convolution layer we use:

```c++
outputTemp.slice(outMap) += bias(outMap);
```

which failes if ``batchSize > 1`` since we iterate over all input slices.

zoq:
Also we have to use:

```c++
output = arma::Mat<eT>(outputTemp.memptr(), outputTemp.n_elem / input.n_cols, input.n_cols);
```

instead of

```c++
output = arma::Mat<eT>(outputTemp.memptr(), outputTemp.n_elem, 1);
```

inside the ``MeanPooling`` class.

src/mlpack/methods/ann/gan_impl.hpp
```diff
@@ -127,8 +129,9 @@ double GAN<Model, InitializationRuleType, Noise>::Evaluate(
   if (!reset)
     Reset();
 
-  currentInput = this->predictors.unsafe_col(i);
-  currentTarget = this->responses.unsafe_col(i);
+  // Alias here
+  currentInput = this->predictors.cols(i, i + batchSize - 1);
+  currentTarget = this->responses.cols(i, i + batchSize - 1);
```
ShikharJ: @zoq Is there any better way of aliasing the columns here? If not I will remove the above comment.

ShikharJ:
@zoq Hmm, it seems that the `Appveyor` build is failing due to some reason,

src/mlpack/methods/ann/gan_impl.hpp
zoq: hm, we could use ``memptr()`` (advanced constructor), what do you think?

src/mlpack/methods/ann/layer/transposed_convolution_impl.hpp
```diff
@@ -114,30 +114,39 @@ void TransposedConvolution<
...
   outputTemp = arma::Cube<eT>(output.memptr(), outputWidth, outputHeight,
-      outSize, false, false);
+      outSize * batchSize, false, false);
   outputTemp.zeros();
 
-  for (size_t outMap = 0, outMapIdx = 0; outMap < outSize; outMap++)
+  for (size_t outMap = 0, outMapIdx = 0, batchCount = 0; outMap <
+      outSize * batchSize; outMap++)
   {
+    if (outMap != 0 and outMap % outSize == 0)
```
zoq: My guess is that ``and`` isn't supported in MSVC, but that's easy to test.

ShikharJ:
@zoq This is ready for a review.

src/mlpack/methods/ann/layer/transposed_convolution_impl.hpp
```diff
@@ -114,30 +114,39 @@ void TransposedConvolution<
     OutputDataType
 >::Forward(const arma::Mat<eT>&& input, arma::Mat<eT>&& output)
 {
-  inputTemp = arma::cube(input.memptr(), inputWidth, inputHeight, inSize);
+  batchSize = input.n_cols;
+  inputTemp = arma::cube(input.memptr(), inputWidth, inputHeight,
```
zoq: hm, should we use ``copy_aux_mem = false`` here to avoid another copy?

src/mlpack/methods/ann/layer/bilinear_interpolation_impl.hpp
```diff
@@ -53,20 +54,22 @@ template<typename eT>
...
   }
 
   assert(inRowSize >= 2);
   assert(inColSize >= 2);
 
-  arma::cube inputAsCube(input.memptr(), inRowSize, inColSize, depth);
-  arma::cube outputAsCube(output.memptr(), outRowSize, outColSize, depth,
-                          false, true);
+  arma::cube inputAsCube(input.memptr(), inRowSize, inColSize,
+                         depth * batchSize);
```
zoq: See coment about ``copy_aux_mem``.

src/mlpack/methods/ann/layer/transposed_convolution_impl.hpp
ShikharJ: @zoq I don't think we can do that as that would cause it to cast away the constness of the input matrix.

ShikharJ:
@zoq The static analysis checks are superfluous in my opinion, though you may want to take a look for yourself to be certain.

zoq:
Agreed, I think it's still good practice to initialize ``inSize, outSize, batchSize`` (constructor list), the other issues are just false.

src/mlpack/methods/ann/layer/transposed_convolution_impl.hpp
zoq: Good point, I guess we could do the same as done here:

https://github.com/mlpack/mlpack/blob/7fead71e00d8a6c3b9b2a4708a13cc469f259ffb/src/mlpack/methods/logistic_regression/logistic_regression_function_impl.hpp#L26-L27

your call.

ShikharJ: I'll make the changes, since it would lead to a considerable speedup for larger batch-sizes.

ShikharJ:
@zoq This is ready for another review.

src/mlpack/methods/ann/layer/atrous_convolution_impl.hpp
```diff
@@ -175,16 +188,23 @@ void AtrousConvolution<
 >::Backward(
     const arma::Mat<eT>&& /* input */, arma::Mat<eT>&& gy, arma::Mat<eT>&& g)
 {
-  arma::cube mappedError(gy.memptr(), outputWidth, outputHeight, outSize,
-      false, false);
+  arma::cube mappedError(gy.memptr(), outputWidth, outputHeight,
+      outSize * batchSize, false, false);
```
rcurtin: Shouldn't it be possible to infer `batchSize` from the shape of `input`, `gy`, or `g`?

ShikharJ: Yes we can, but since we're anyways storing the `batchSize` in the forward pass, there's no need to call `input.n_cols` or `gy.n_cols` repeatedly.

ShikharJ:
The full dataset trained under just 10 hours!
mnist_gan_full

PR #1434, opened 2018-06-16
Remove redundant data members, move NegativeLogLikelihood to loss folder

akhandait:


zoq:
Looks like there are some merge conflicts, I can resolve the conflict, just let me know.

akhandait:
Done!
Just to inform, I left the `inputParameter` in some layers that can involve multiple layers in case they take Linear or Convolutional layers.

zoq:
Thanks again for taking a closer look into the issue.

PR #1427, opened 2018-06-08
Merge layer run parameter.

zoq:
This introduces the run parameter for the merge layer; if set the merge layer forwards the Forward/Backward/Gradient call. This enables us to use the FFN/RNN class to merge sub-networks.

Here is an example:

```c++
// Create a simple pseudo input/output pair.
arma::mat input = arma::randn(10, 1);
arma::mat target;
target.ones(1, 256);

// User sub-network.
Sequential<>* userModel = new Sequential<>();
userModel->Add<Linear<> >(10, 2);

// Item sub-network.
Sequential<>* itemModel = new Sequential<>();
itemModel->Add<Linear<> >(10, 2);

// Merge the user and item sub-network.
MultiplyMerge<> mergeModel(true, true);
mergeModel.Add(userModel);
mergeModel.Add(itemModel);

// Create the main network.
FFN<NegativeLogLikelihood<>, RandomInitialization> network;
network.Add<IdentityLayer<> >();
network.Add<MultiplyMerge<> >(mergeModel);
network.Add<IdentityLayer<> >();
network.Add<Linear<> >(2, 2);
network.Add<SigmoidLayer<> >();
network.Add<LogSoftMax<> >();

// Set some simple input/output data.
network.Predictors() = input;
network.Responses() = target;

// Gradient internally calls the Evaluate method.
arma::mat gradient;
// network.Evaluate(network.Parameters(), 0, 1);
network.Gradient(network.Parameters(), 0, gradient, 1);
```

PR #1435, opened 2018-06-16
Matrix subview support

haritha1313:


src/mlpack/methods/ann/layer/subview.hpp
```diff
@@ -57,9 +61,11 @@ class Subview
   void Forward(InputType&& input, OutputType&& output)
   {
     // Check if input has been selected as required.
-    if ((input.n_rows != (end - begin + 1)) && (end != 0))
+    if (((input.n_rows != (endRow - beginRow + 1)) ||
+        (input.n_cols != (endCol - beginCol +1)))
+        && !(endRow == 0 && endCol == 0))
     {
-      output = arma::mat(&input(begin), end - begin + 1, input.n_cols, false);
+      output = input.submat(beginRow, beginCol, endRow, endCol);
```
zoq: Thanks for looking into the issue, I guess it might be useful to combine the both ideas. So if the input is a vector ``n_cols = 1`` or if ``beginCol = 0`` and ``endCol = 0``, we can use the advanced constructor to avoid the copy. Let me know what you think.

haritha1313: You are right. I think `beginCol != 0` is the only  case where submatrix is required. Seeing that advanced constructor can work for matrices too if `beginCol` is 0.

rcurtin: I'm not sure if the condition is right to use the advanced constructor.  I agree that definitely if `n_cols == 1` then we can use the advanced constructor, using the memory from `&input(beginRow)` to `&input(endRow)`.  But I think the second condition should be `beginRow == 0` and `endRow == input.n_rows - 1`.

haritha1313: Based on a few trials this is what I have understood: For any matrix whose subview has `beginCol == 0`, advanced constructor can be used using the memory from `&input(beginRow)` and specifying the number of rows and number of columns, i guess we are able to do this because armadillo uses column major. When `beginCol != 0` we can't access the starting index using the same method, hence `submat` was needed. Please correct me if I am mistaken. I tried out this on a few cases and it is working. Thus I couldn't understand the need of the second condition you mentioned, it would be great if you could help me with the example of a subview.

zoq:
In the latest blog post you said:

>  I am facing a few issues when using the newly created subview layer is merged with existing neural network models, which I am trying to debug right now.

Can you comment on that point, perhaps we can provide some input on this one.

haritha1313:
@zoq It was a slight mistake. I thought it was the introduction of `subview` layer which caused the trouble, and was trying to debug that. But the issue was with the `inSize` and `outSize` parameters on introducing `Embedding` layer. It has been solved now. Thanks for offering help. I think the CreateModel() is complete now. I think I'll go on and make a PR with the implemented methods so that you can stay updated.

src/mlpack/methods/ann/layer/subview.hpp
rcurtin: Since Armadillo is column-major, this means that every point in every column is contiguous in memory.  So you are right that if `beginCol == 0`, then the beginning of the memory we need to use is just `&input(beginRow)` (or, alternately, `input.memptr() + beginRow`).

But let's consider the case where `beginRow == 0` and `endRow == input.n_rows - 1`.  In this situation, we are using all of the rows of the matrix.  So if we have any number of columns, they will be contiguous in memory, using the memory from `&input(beginCol * n_rows)` to `&input((endCol + 1) * n_rows - 1)`; or, alternately expressed, from `input.colptr(beginCol)` to `input.colptr(endCol) - 1`.  Does that make more sense?  Let me know if I can clarify further.

haritha1313: You are right, I see what you meant now. Thanks for the clarification. I'll make the necessary edits. :)

zoq:
Okay, #1427 is merged, there is just one last issue we should think about before merging this one as well.

Currently, this layer doesn't support batches, which I think would be great if we like to accelerate the training process. So in case of batches, the input is a matrix were each col represents another sample. An easy solution would be iterate over the input, but I don't see an easy way which allows us to differentiate between the batch and non-batch input which could be a matrix as well. One solution would be to let the user provide the input size, let me know what you think.

haritha1313:
Referring to other layers implemented, input size which user provides is of form `(inSize, outSize)`. So does this assume that each sample is a vector, making the net input a matrix? \n I assume that if there is batch input where each sample is of matrix form, the total input will be `arma::cube`. So we can add two parameters `num_rows`, `num_cols` (`num_cols = 1` for vector input) in the constructor for input size, and iterate over each sample as you said.

zoq:
If a single input sample is a matrix, it will be vectorized internally, so that each layer can expect that each col represents another sample. In case of the RNN class you can pass an ``arma::cube`` as input, were each slice represents another time step but in this case the input is also transformed as well.

So, we could say that a user has to specify the inSize, I think outSize isn't needed since we can derive the rest from the other parameters. In most of the cases ``inSize`` is ``1``, and if the input is a matrix ``inSize > 1``, this should be a rare case, and I don't think we will need it for the ann modules right now.

We could also call the parameter ``width`` and ``height`` in this case ``width`` is the same as ``inSize`` but I don't think we need ``height``. What do you think?

haritha1313:
Thanks for helping me understand. So we just have to iterate `input.n_rows/inSize` times and create subviews of samples using the existing method. Is that right?

zoq:
``input.n_cols / inSize`` times

here is a simple example for ``batchSize = 1`` and ``inSize = 1`` it would look like

``input = arma::ones(10, 1);``

``Subview(0, 4)``

would return the first half of the input vector of size:

``output = arma::ones(5, 1);``

if we use a ``batchSize > 1`` e.g. ``2``

the input looks like:

``input = arma::ones(10, 2);``

``inSize`` is still ``1``, so the output would be of size

``output = arma::ones(5, 2);``

if someone does use a matrix as input e.g.

``arma::mat(10, 10)``

it will be transformed to:

``arma::mat(10 * 10, 1)``

so the output of the Subview layer would be:


```c++
size_t batchSize = input.n_cols / inSize

output.set_size(... (depends on the layer parameter), batchSize);

for (size_t i = 0; i < batchSize; i++)
{
  if (beginRow == 0 && endRow == (input.n_rows - 1))
  {
    output.col(i) = arma::mat(&input(beginCol * input.n_rows), input.n_rows * endCol - beginCol + 1, 1, false);
  }
  else
  {
    output.col(i) = arma::vectorise(input.submat(beginRow, beginCol, endRow, endCol));
  }
}
```

I hope this is useful, let me know if I should clarify anything.

haritha1313:
Sorry. `input.n_cols` was  what I meant. Made a mistake while typing. :D . Thanks for the much useful explanation.

zoq:
Okay, just one more thing, do you think we could add a batch test?

src/mlpack/methods/ann/layer/subview.hpp
```diff
@@ -56,10 +68,42 @@ class Subview
...
+        (endCol - beginCol + 1))) || (input.n_cols != batchSize))
     {
-      output = arma::mat(&input(begin), end - begin + 1, input.n_cols, false);
+      for (size_t i = 0; i < batchSize; i++)
+      {
+        if (beginRow == 0 && endRow == (input.n_rows - 1))
+        {
+          // Matrix of contiguous elements.
+          output.col(i) = arma::mat(&input(batchBegin * input.n_rows),
+              (input.n_rows * (batchEnd - batchBegin + 1)), 1, false);
```
rcurtin: So, I don't think there's a need to try and use any aliasing anymore, since `output.col(i) = ...` will always perform a copy.  I agree that we need to do it because the batch size is usually more than 1, but probably now we can simplify.

haritha1313: Sorry for the late response. The notification just skipped me. Do you mean that we can just use submat() for both cases, in fact avoid having two cases? I didn't get what you meant by aliasing here.

zoq: Right, that would simplify the code.

rcurtin: Yeah, there's no need for the special constructor here, so we can always use `submat()`.

zoq:
Let me know if you need help with the merge conflicts.

zoq:
Thanks again for looking into all those issues.

PR #1436, opened 2018-06-18
Win build and app tutorials

gmanlan:
- Tutorial 1: Up-to-date (VS2017/mlpack3.0.2) Windows Build Guide (doc/guide/build_windows.hpp)
- Tutorial 2: Sample ML C++ App for Windows (doc/guide/sample_ml_app.hpp)
- VS project: Sample App for Windows (doc/examples/sample-ml-app)

Tested using Win10, VS2017, latest version of mlpack, armadillo 8.500, boost 1.66


doc/guide/build_windows.hpp
```diff
@@ -0,0 +1,95 @@
...
+
+The directories and paths used in this tutorial are just for reference purposes.
+
+@section build_prereqs Pre-requisites
+
+- Install CMake for Windows (win64-x64 version from https://cmake.org/download/)
+and make sure you can use it from the Command Prompt (may need to add to the PATH)
+
+- Download the latest mlpack release from here:
+<a href="http://www.mlpack.org/files/mlpack-3.0.2.tar.gz">mlpack-3.0.2</a>
```
zoq: I was thinking if it might be useful to provide an alias for the latest package so that we don't have to update this tutorial once we have a new release. @rcurtin what do you think?

doc/guide/build_windows.hpp
```diff
@@ -0,0 +1,95 @@
...
+
+@section build_instructions Windows build instructions
+
+- Unzip mlpack to "C:\mlpack\mlpack-3.0.2"
```
zoq: Is there any way to use an existing home directory? Something like ``~/``on linux.

doc/guide/sample_ml_app.hpp
```diff
@@ -0,0 +1,194 @@
...
+using namespace mlpack::cv;
+@endcode
+
+@section sample_load_dataset Loading the dataset
+
+First step is about loading the dataset. Here we load a CSV dataset 
+(i.e.: mlpack/tests/data/german.csv), assuming the labels don't require normalization.
+
+@code
+bool loaded = mlpack::data::Load("data/german.csv", dataset);
```
zoq: Should we define ``dataset, labels`` first? I guess people might copy each line, and end up with some errors.

doc/guide/sample_ml_app.hpp
```diff
@@ -0,0 +1,194 @@
...
+	numClasses, numTrees, minimumLeafSize);
+@endcode
+
+Now that the training is completed, we quickly compute the training accuracy:
+
+@code
+Row<size_t> predictions;
+rf.Classify(dataset, predictions);
+const size_t correct = arma::accu(predictions == labels);
+printf("\nTraining Accuracy: %f", (double(correct) / double(labels.n_elem)));
```
zoq: hm, should we use std::cout here? Usually we don't use ``printf`` in the codebase.

doc/guide/sample_ml_app.hpp
```diff
@@ -0,0 +1,194 @@
...
+double cvF1 = F1<Binary>::Evaluate(rf, dataset, labels);
+printf("\nF1: %f", cvF1);
+@endcode
+
+@section sample_save_model Saving the model
+
+Now that our model is trained and validated, we save it to a file so we can use it later.
+
+@code
+mlpack::data::Save("mymodel.xml", "model", rf, false,
```
zoq: Save should be able to derive the format from the filename.

doc/guide/build_windows.hpp
gmanlan: This is tricky - we could use %userprofile% in Windows, but that directory is not very friendly for the purpose of an easy-to-follow tutorial (and it changes depending on the windows version) - I think for simplicity we should use the most basic directory we can imagine (there is also a note that says this path is just for reference). Long paths or paths with spaces may produce issues in Windows so we are aiming for the safest option.

Let me know if you still want to change it.

doc/guide/sample_ml_app.hpp
gmanlan: Yep, added the missing dataset definition.

doc/guide/sample_ml_app.hpp
gmanlan: Changed all the printf to cout - thanks

doc/guide/sample_ml_app.hpp
gmanlan: Right, removed the unnecessary parameter

doc/guide/sample_ml_app.hpp
```diff
@@ -0,0 +1,197 @@
...
+double cvF1 = F1<Binary>::Evaluate(rf, dataset, labels);
+cout << "\nF1: " << cvF1;
+@endcode
+
+@section sample_save_model Saving the model
+
+Now that our model is trained and validated, we save it to a file so we can use it later.
+
+@code
+mlpack::data::Save("mymodel.xml", "model", rf, false);
```
rcurtin: It might be worth pointing out that we could also save as `mymodel.bin` which will be much smaller.  The XML saves are huge because of all the XML tags :(

doc/guide/build_windows.hpp
rcurtin: I already go through `doc/` when I do every release and increment any version numbers, so it's not a huge issue either way.  Aliases are nice if we can do it; do you have a suggested way to do it?

doc/guide/build_windows.hpp
```diff
@@ -0,0 +1,95 @@
+/*! @page build_windows Building mlpack From Source
+
+@section build_intro Introduction
```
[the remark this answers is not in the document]
gmanlan: 👍

doc/guide/build_windows.hpp
```diff
@@ -0,0 +1,95 @@
+/*! @page build_windows Building mlpack From Source
```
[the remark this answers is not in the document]
gmanlan: 👍

doc/guide/sample_ml_app.hpp
```diff
@@ -0,0 +1,197 @@
+/*! @page sample_ml_app Sample C++ ML App
```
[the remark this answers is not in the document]
gmanlan: 👍

doc/guide/sample_ml_app.hpp
```diff
@@ -0,0 +1,197 @@
...
+Row<size_t> predictions;
+rf.Classify(dataset, predictions);
+const size_t correct = arma::accu(predictions == labels);
+cout << "\nTraining Accuracy: " << (double(correct) / double(labels.n_elem));
+@endcode
+
+@section sample_crossvalidation Cross-Validating
+
+To evaluate the classifier, we use K-Fold cross-validation. We also define which metric to use in order
+to assess the quality of the trained model.
```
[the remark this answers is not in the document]
gmanlan: 👍

doc/guide/sample_ml_app.hpp
```diff
@@ -0,0 +1,197 @@
...
+
+@code
+const size_t k = 10;
+KFoldCV<RandomForest<GiniGain, RandomDimensionSelect>, Accuracy> cv(k, 
+	dataset, labels, numClasses);
+double cvAcc = cv.Evaluate(numTrees, minimumLeafSize);
+cout << "\nKFoldCV Accuracy: " << cvAcc;
```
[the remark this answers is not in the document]
gmanlan: Right. Added a note for that.

doc/guide/sample_ml_app.hpp
gmanlan: 👍

gmanlan:
I'm thinking that using AppVeyor for the example VS project would require to modify the project configuration (i.e. paths) so it works with the build system - however this would break the consistency between the example project and the 'sample_project_config' section of the 'sample_ml_app.hpp' tutorial ... what do you think?

zoq:
I guess we could use ``mlpack-latest`` instead of ``mlpack-3.0.2`` or something similar to get the same paths, do you think that would be reasonable, or is there anything else I missed? If the path is the main reason, we could also create an alias as part of the build script.

rcurtin:
I can set up an `mlpack-latest.tar.gz` link on mlpack.org for the sake of documentation, but I don't think it's a problem to hardcode `mlpack-3.0.2` and then update it after each release.  What do you think, would that work?

gmanlan:
To reduce the overhead of updating the doc each time a new release is available, I have changed the mlpack download path to "http://www.mlpack.org/download.html" so it always refers to the latest version.

rcurtin:
I talked to @gmanlan over email, I think maybe the best thing to do here is merge as-is, and create an issue for the VS build of the tutorial.

For that build, I think the best idea is to have a special VS project configuration that we can keep in some directory like `.appveyor/`.  Then in `appveyor.yml` we can just copy that configuration into place, overwriting the existing configuration, and then run the example build as an extra step.

zoq:
Sounds reasonable to me, no need to delay this really helpful tutorial any longer.

rcurtin:
Thanks again for the contribution!  I really appreciate it.  I forgot to add, if you'd like to add your name to `src/mlpack/core.hpp` and `COPYRIGHT.txt`, please feel free and I will merge it!  And if you'd like some mlpack stickers to put on your laptop, feel free to send Marcus or I an email with your mailing address and we will get them sent. :)

I opened #1463 for the build part.

gmanlan:
Great - I'm glad it helps. I just realized that we have not linked/updated the main doc/tutorials page at http://www.mlpack.org/docs/mlpack-3.0.2/doxygen/tutorials.html, so I will be updating this soon to make sure users can find both Linux and Windows tutorials.

PR #1437, opened 2018-06-19
[WIP] Implement Dual Optimizer Class

ShikharJ:


zoq:
Do you think we could implement this (``'DualOptimizer'``) class, so that it's not just for training GAN's?

It looks like it's already usable for other classes as well, and we should just use different names? Let me know what you think.

zoq:
I tested different ideas but none of them was general enough, that I would say it could be easily reused for another non-GAN related function. So I think, either we implement an dual optimizer that is specific for the GAN class or we implement the approach directly in the GAN class itself. Inside the GAN class means that we provide another Train method that takes two optimizers and handle the rest in the Train function. Do you think one of the ideas is reasonable?

AYESDIE:
@ShikharJ : Do you think that this should be implemented here, considering all the optimizers are moved to Ensmallen.

ShikharJ:
@AYESDIE You are correct, this should be implemented in ensmallen, but I'd keep this open here to serve as a reminder to-do.

saksham189:
Hi @ShikharJ , Can I work on this? I was thinking that we could implement this inside the GAN Train method as @zoq had suggested if it is going to be specific to a GAN.

ShikharJ:
@saksham189 I would rather have this implemented in ensmallen itself, even if it is unique to a `GAN`.

saksham189:
@ShikharJ   After rethinking about the design I can see advantages of having this in `ensmallen` if we were to have different `Train` methods for example and since we should be keeping the optimization code inside ensmallen anyways. 
I wanted your opinion regarding the implementation. I was thinking that we could keep a boolean inside GAN class and calculate the gradient for the generator or discriminator depending on its value once the pertaining is over.

rcurtin:
My concern is that we don't currently have any abstraction inside of ensmallen to handle optimizing two functions at once.  I'd suggest that for now we keep this in mlpack, and if we find this abstraction to also be useful outside of just our context here (and as the abstraction matures) we can figure out how to fit it into ensmallen.  So far I only really see it being useful to GANs right now, but someone correct me if it has clear and widespread applicability. :)

saksham189:
I don't know any other use of this as well other than GANs. So maybe we could have it inside the Train method only. What do you think @ShikharJ ?

zoq:
Agreed, since this is going to be GAN specific, it shouldn't be part of ensmallen, at least for now.

@saksham189 I was thinking the same, so let's put it into the ``Train`` method.

saksham189:
I am not sure if using 2 optimizers inside `Train` method of GAN would be feasible with the current `SGD` optimizer API. We want to train the discriminator over one batch of data using one optimizer and then train the generator using another optimizer over 1 batch of data and repeat this multiple times.

However,  calling `Optimize` by setting `maxIterations = batchSize` would result in the method calculating the final objective value over the entire training data at the end which we do not want. Also, there is no way to control which batch is used for training and it would keep training with the first batch. (We could swap the training data but that is expensive). 
We probably will need to define a method for training on just one batch of data. Let me know what you think.

zoq:
What if we don't use the SGD optimizer but instead use the update policy (https://github.com/mlpack/ensmallen/tree/master/include/ensmallen_bits/sgd/update_policies). In this case we would implement another SGD related class that could e.g. take two functions. Let me know what you think.

saksham189:
>  In this case we would implement another SGD related class that could e.g. take two functions

So this would be inside ensmallen? This would work but it would again be specific to the GAN class. 
Also do you think we should separate the `EvaluateWithGradient` method for discriminator and generator or add an extra boolean parameter to the function to denote which is to be trained?

walragatver:
Hi  @zoq, @ShikharJ, @rcurtin, @saksham189, 
* I am having an idea to resolve this. With this idea, we will `not need` to make any changes in `ensmallen` repo and it will also add some more functionality to the current implementation.
* Basically the idea is to use the `Train` function of `FFN` class i.e `Train` function of Generator and `Discriminator`.
* We will add two members to current `GAN` class namely `DiscriminatorOptimizer` and `GeneratorOptimizer`.
* The user will specify both of them and will be passed in the `Train` function of `GAN` class. However, the `important point` to note here is that none of the two `optimizer's` will have `actual iterations` for which the `GAN` will be trained. However, there `maxIterations` parameter will have the number of `iterations` `Generator` and `Discriminator` will iterate in `One` single iteration of `GAN`. For example, In Standard GAN in one `GAN` iteration if `GeneratorUpdateStep = 1` then `Discriminator` `trains for one iteration` and `Generator` network also updates it's parameter`one time (1 Iteration)`. Please let me know if this is not clear.
* We will take one more parameter in the `Train` function namely `numIterations` specifying the `Iterations` the `GAN` will train.
* The current `GAN` training takes place as follows.
```
GAN :: Train(Optimizer)  {  
   // Call to optimize.  
   for(int i = 0; i < actualIterations;)  
   {
        // Calculate Gradients of Generator and Discriminator Gradients simultaneously.  

       // Update the Parameters of Both the networks together.  
   }  
}`
```
* Now, the way I am proposing is something like this. 
```
GAN :: Train(Optimizer)  {  
   // Call to optimize.  
   for(int i = 0; i < actualIterations;)  
   {
        // Calculate Gradients of Discriminator and update it's parameters simultaneously.  

       // Calculate Gradients of Generator and update it's parameters simultaneously. 
   }  
}`
```
* In order to do this, we will use `FFN :: Train()` function with the above two `Optimizer's`. 
* We will not use `Optimizer API` functions for `GANS` but directly. but will use them with the  help of `FFN`.  
* Prototype of `GAN's Train` Function : 
```
GAN :: Train(arma mat :: TrainData, arma:: responses, Optimizer &DiscriminatorOptimizer, Optimizer &GeneratorOptimizer, size_t numIterations)
{
    //Explicit for- loop (Not of Optimizer API).
    for(int i = 0; i < numIterations;)
   {
        //Make Noise. 
        FakeImages = Generator.forward(Noise).
        arma :: mat predictorsDiscriminator = FakeImages + BatchofRealImages. // Memory sharing 
possible.
        arma :: mat responsesDiscriminator;
        // Train the discriminator.
        Discriminator.Train(predictorsDiscriminator, responsesDiscriminator, DiscriminatorOptimizer);

       // If we are training Generator.
      {
            // Generate Images.
            // Train the discriminator after flipping labels.
            Discriminator.Train(fakeImages, flippedResponses, DiscriminatorOptimizer);
            // Pass It's error to generator using 
            generator.error = generator.error = boost::apply_visitor(deltaVisitor,
        discriminator.network[1]);
            // makeNoise. 
           Generator.Train(Noise, GeneratorOptimizer);
     }
   }  
}
```
* As we use a `reference` to the optimizer in `FFN:: Train()` parameters during each training will be same. 
* However as mentioned in #1770. `Generators` are trained in an unsupervised way. So in the current `FFN` implementation following members will be added.
```
// Bool to know supervised or unsupervised mode.
bool supervised. 
// Overload for train function taking only predictors.
FFN:: Train(predictors, Optimizer); 
```
* There would be a change in EvaluateWithGradient() of FFN like below and also constructor's in `FFN` will be needed to modify. The Default value of supervised will be kept 1. For the `Generator` network it will be set to false.
At line 335 in `ffn_impl.hpp`:
```
// If supervised learning is set then calculate the error with the help of responses else skip.
if(supervised)
{
      outputLayer.Backward(
      std::move(boost::apply_visitor(outputParameterVisitor, network.back())),
      std::move(responses.cols(begin, begin + batchSize - 1)),
      std::move(error));
}
``` 
* However, `important point` to note here is that we will not follow `Optimizer API` in GAN class as we are not using that directly. Basically, `GAN` class would be a `wrapper` for implementing `GAN`
* With this idea we don't need to make any changes in `ensmallen`.

Please let me know about this idea. If everyone agrees I will start implementing it.

zoq:
Maybe I missed something, in:

```c++
Discriminator.Train(predictorsDiscriminator, responsesDiscriminator, DiscriminatorOptimizer);
```
and
```c++
Discriminator.Train(fakeImages, flippedResponses, DiscriminatorOptimizer);
```

we will do a full sweep over the samples at the end, here is the snippet from the SGD optimizer:

```c++
// Calculate final objective.
  overallObjective = 0;
  for (size_t i = 0; i < numFunctions; i += batchSize)
  {
    const size_t effectiveBatchSize = std::min(batchSize, numFunctions - i);
    overallObjective += f.Evaluate(iterate, i, effectiveBatchSize);
  }
  return overallObjective;
```

ideally we can avoid that, for which I think we would have to write a special optimizer for the GAN case. The optimizer would look similar to the SGD one, but will take two functions as input. As I said maybe I missed something. Let me know what you think.

walragatver:
Actually, I didn't get your doubt clearly.  
I think  you are worried  about the `overhead` with this. 
> ```c++
> // Calculate final objective.
>   overallObjective = 0;
>   for (size_t i = 0; i < numFunctions; i += batchSize)
>   {
>     const size_t effectiveBatchSize = std::min(batchSize, numFunctions - i);
>     overallObjective += f.Evaluate(iterate, i, effectiveBatchSize);
>   }
>   return overallObjective;
> ```

* I would agree about it, But it's not that overhead if we look at the current implementation.  At the end in the current implementation, we are giving a full sweep to calculate the `Loss` over the training set. which is quite equivalent to calculate `loss` with `batches` one by one. 

* If you are concerned about the `overhead` what we can do is add somewhat `modularity` to current `optimizers` such that it will only calculate `overall loss` only when it is specified. And it will `avoid` that calculation when not specified. An addition of a boolean parameter may suffice that like this.

```
bool calculateLoss; // True if we wish to calculate the loss.
// Calculate final objective if  mentioned.
overallObjective = 0;
if(calculateLoss)
{
   for (size_t i = 0; i < numFunctions; i += batchSize)
  {
    const size_t effectiveBatchSize = std::min(batchSize, numFunctions - i);
    overallObjective += f.Evaluate(iterate, i, effectiveBatchSize);
  }
}
 return overallObjective;
```
* I know this might be wierd.

Or,

* I think you are assuming that we are passing all `predictors` to the `Discriminator` in one iteration.  
* We will be passing `two batches` in one iteration in `predictorsDiscriminator` out of which one will be of `real predictors` and one will be of `fake predictors`.
* Similarly, `fakeImages` will be one batch of images from `Generator`. 

Or, 

I think that you are concerned about the Loss of `GAN`.
* But, as we know the loss of `GANs` doesn't determines its performance. So I thought It would be ideal not to calculate that loss.

Actually, I took the reference to implementation over here in Keras. 
https://github.com/eriklindernoren/Keras-GAN/blob/master/gan/gan.py

I thought we might be able to implement a `Train` function similar as implemented above.


Let me know if these are not your doubts. 
I would agree on the overhead thing but we can avoid that with that bool parameter if required.
Please let me know if my idea is going to scrap and thanks for the quick reply :+1:

saksham189:
```
Discriminator.Train(fakeImages, flippedResponses, DiscriminatorOptimizer);
```
If I am not wrong then this would incorrectly update the gradients of the discriminator.

walragatver:
> ```
> Discriminator.Train(fakeImages, flippedResponses, DiscriminatorOptimizer);
> ```
> If I am not wrong then this would incorrectly update the gradients of the discriminator.

* Actually, I have written that line based on the current implementation. Most probably you are correct as it was also done in a similar way in the link I have shared. I was knowing that but wasn't sure hence didn't mention while writing.

At line `316` in `gan_impl.hpp` currently: 
```
discriminator.responses.cols(numFunctions, numFunctions + batchSize - 1) =
        arma::ones(1, batchSize);
discriminator.Gradient(discriminator.parameter, numFunctions,
        noiseGradientDiscriminator, batchSize);
```
This is similar to what I have written.
* However, the way you want it would be: 
```
Discriminator.Forward(fakeImages, fakeResponses);
Discriminator.outputLayer.Backward(
  std::move(boost::apply_visitor(outputParameterVisitor, network.back())),
      std::move(flippedResponses.cols(begin, begin + batchSize - 1)),
      std::move(error));
Discriminator.Backward(); // To calculate gradient of activation of first layer in 
                        // Discriminator without any updates.

generator.error = boost::apply_visitor(deltaVisitor,
        discriminator.network[1]);

```
Doesn't make any significant change. But thanks for raising it as I also I wanted to mention that :+1: 
Let me know it that's wrong.

rcurtin:
> ideally we can avoid that, for which I think we would have to write a special optimizer for the GAN case.

My comment here isn't about GANs at all, but it might be nice to provide support for ensmallen optimizers so that the exact objective isn't computed at the end of the optimization; i.e., a user could specify `sgd.ExactObjective() = false` (and that would be the default) and it would return an approximation of the objective just like most other libraries.  This is a tangent from this issue, like I said, but if there is interest I can open an issue in the ensmallen repository for it; just let me know what you think.  (I've thought it was a good idea for a while, but have never done it.)

saksham189:
Thanks for the clarification @walragatver 
The best option IMO is to create two different functions for GAN and introduce a corresponding optimiser in ensmallen to stay consistent with the rest of the mlpack API as @zoq had suggested which I am already working on but got a bit busy due to upcoming final exams.

zoq:
> >  ideally we can avoid that, for which I think we would have to write a special optimizer for the GAN case.

> My comment here isn't about GANs at all, but it might be nice to provide support for ensmallen optimizers so that the exact objective isn't computed at the end of the optimization; i.e., a user could specify sgd.ExactObjective() = false (and that would be the default) and it would return an approximation of the objective just like most other libraries. This is a tangent from this issue, like I said, but if there is interest I can open an issue in the ensmallen repository for it; just let me know what you think. (I've thought it was a good idea for a while, but have never done it.)

Sounds like a good idea to me, would be helpful here as well.

walragatver:
> sgd.ExactObjective() = false


Yes, @rcurtin, it will be sort of like this. 
> bool calculateLoss; // True if we wish to calculate the loss.
> // Calculate final objective if  mentioned.
> overallObjective = 0;
> if(calculateLoss)
> {
>     for (size_t i = 0; i < numFunctions; i += batchSize)
>     {
>      const size_t effectiveBatchSize = std::min(batchSize, numFunctions - i);
>      overallObjective += f.Evaluate(iterate, i, effectiveB//.atchSize);
>    }
> }
>  return overallObjective;

* It would be a great modification. I am definitely interested to work on it, as it may also be important for other implementation's. 


@saksham189, regarding this, what we can do is keep `supervised` boolean as `private` so that users will not know about it and will be only for `internal working`.
> I am not in favor of introducing the `supervised` boolean variable in `FFN` because it is not general enough.

rcurtin:
Thanks for the input---see mlpack/ensmallen#108.

walragatver:
Hi everyone, Should I also start working on my proposed solution?

zoq:
@walragatver it might be a good idea to finish the work on mlpack/ensmallen#108 first, let me know what you think.

walragatver:
Okay sir, I just wanted a confirmation that whether it's correct or not. :+1:

rcurtin:

PR #1433, opened 2018-06-15
[WIP] Add NormalDistribution class and ReconstructionLoss

akhandait:
@sumedhghaisas I will add the tests for `ReconstructionLoss` soon.

PR #1439, opened 2018-06-21
Add NormalDistribution class

akhandait:
splitting #1433

### DIFF
--- a/src/mlpack/core/dists/normal_distribution.hpp
+++ b/src/mlpack/core/dists/normal_distribution.hpp
@@ -1,0 +1,123 @@
+/**
+ * @file normal_distribution.hpp
+ * @author Atharva Khandait
+ *
+ * Implementation of the Normal distribution.
+ *
+ * mlpack is free software; you may redistribute it and/or modify it under the
+ * terms of the 3-clause BSD license.  You should have received a copy of the
+ * 3-clause BSD license along with mlpack.  If not, see
+ * http://www.opensource.org/licenses/BSD-3-Clause for more information.
+ */
+#ifndef MLPACK_CORE_DISTRIBUTIONS_NORMAL_DISTRIBUTION_HPP
+#define MLPACK_CORE_DISTRIBUTIONS_NORMAL_DISTRIBUTION_HPP
+
+#include <mlpack/prereqs.hpp>
+
+namespace mlpack {
+namespace distribution {
+
+/**
+ * Multiple univariate Normal(Gaussian) distributions.
+ */
+template <typename DataType = arma::mat>
+class NormalDistribution
+{
+ private:
+  //! Means of the distributions.
+  DataType mean;
+  //! Standard deviations of the distributions.
+  DataType stdDeviation;
+  //! Pre standard deviation. After softplus this will give standard deviation.
+  DataType preStdDev;
+
+  //! log(2pi)
+  static const constexpr double log2pi = 1.83787706640934533908193770912475883;
+
+ public:
+  /**
+   * Default constructor, which creates a Normal distribution with zero
+   * dimension.
+   */
+  NormalDistribution()
+  {
+    // Nothing to do here.
+  }
+
+  /**
+   * Create multiple Normal distributions with the given means and Standard
+   * deviations.
+   */
+  NormalDistribution(const DataType& mean, const DataType& stdDeviation);
+
+  /*
+   * Create multiple Normal distribution with the given input matrix which
+   * has pre standard deviation in the upper half and mean in the lower half.
+   */
+  NormalDistribution(const DataType& input);
+
+  /**
+   * Return the probabilities of the given matrix of observations.
+   */
+  double Probability(const DataType& observation) const
+  {
+    return exp(LogProbability(observation));
+  }
+
+  /**
+   * Return the log probabilities of the given matrix of observations.
+   */
+  double LogProbability(const DataType& observation) const;
+
+  /**
+   * Stores the gradient of the log probability in the output matrix.
+   */
+  void LogProbBackward(const DataType& observation, DataType& output) const;
+
+  /**
+   * Return a matrix of randomly generated observations according to the
+   * probability distributions defined by this object.
+   *
+   * @return Matrix of random observations from this Normal distribution.
+   */
+  DataType Random() const;
+
+  /**
+   * Return the mean.
+   */
+  const DataType& Mean() const { return mean; }
+
+  /**
+   * Return a modifiable copy of the mean.
+   */
+  DataType& Mean() { return mean; }
+
+  /**
+   * Return the standard deviation.
+   */
+  const DataType& StdDeviation() const { return stdDeviation; }
+
+  /**
+   * Return a modifiable copy of the standard deviation.
+   */
+  DataType& StdDeviation() { return stdDeviation; }
+
+  /**
+   * Serialize the distribution.
+   */
+  template<typename Archive>
+  void serialize(Archive& ar, const unsigned int /* version */)
+  {
+    // We just need to serialize each of the members.
+    ar & BOOST_SERIALIZATION_NVP(mean);
+    ar & BOOST_SERIALIZATION_NVP(stdDeviation);
+  }
+}; // class NormalDistribution
+
+} // namespace distribution
+} // namespace mlpack
+
+// Include implementation.
+#include "normal_distribution_impl.hpp"
+
+#endif
--- a/src/mlpack/methods/ann/CMakeLists.txt
+++ b/src/mlpack/methods/ann/CMakeLists.txt
@@ -16,6 +16,7 @@ add_subdirectory(layer)
 add_subdirectory(loss_functions)
 add_subdirectory(convolution_rules)
 add_subdirectory(augmented)
+add_subdirectory(dists)
 
 # Add directory name to sources.
 set(DIR_SRCS)
--- a/src/mlpack/methods/ann/dists/normal_distribution.hpp
+++ b/src/mlpack/methods/ann/dists/normal_distribution.hpp
@@ -21,6 +21,9 @@ namespace ann /** Artificial Neural Network. */ {
 /**
  * Multiple univariate Normal(Gaussian) distributions.
  *
+ * Note: For multivariate distributions, where covariace is not diagonal, there is
+ * another class GaussianDistribution in core/dists.
+ *
  * @tparam DataType Type of the input data. (arma::colvec, arma::mat,
  *         arma::sp_mat or arma::cube).
  */
@@ -32,10 +35,7 @@ class NormalDistribution
    * Default constructor, which creates a Normal distribution with zero
    * dimension.
    */
-  NormalDistribution()
-  {
-    // Nothing to do here.
-  }
+  NormalDistribution();
 
   /**
    * Create multiple Normal distributions with the given parameters.
--- a/src/mlpack/methods/ann/dists/normal_distribution.hpp
+++ b/src/mlpack/methods/ann/dists/normal_distribution.hpp
@@ -2,20 +2,20 @@
  * @file normal_distribution.hpp
  * @author Atharva Khandait
  *
- * Implementation of the Normal distribution.
+ * Definition of the Normal distribution class.
  *
  * mlpack is free software; you may redistribute it and/or modify it under the
  * terms of the 3-clause BSD license.  You should have received a copy of the
  * 3-clause BSD license along with mlpack.  If not, see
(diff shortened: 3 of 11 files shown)


akhandait:
Sorry, I forgot to mention this class in the `GaussianDistribution` class, will do that in the next commit.

src/mlpack/core/dists/normal_distribution.hpp
```diff
@@ -0,0 +1,123 @@
...
+  NormalDistribution()
+  {
+    // Nothing to do here.
+  }
+
+  /**
+   * Create multiple Normal distributions with the given means and Standard
+   * deviations.
+   */
+  NormalDistribution(const DataType& mean, const DataType& stdDeviation);
```
sumedhghaisas: I reccommend removing this constructor. LogProbBackward function assumes the input was a single vector, thus this constructor will confuse the users.

src/mlpack/core/dists/normal_distribution.hpp
```diff
@@ -0,0 +1,123 @@
...
+   */
+  void LogProbBackward(const DataType& observation, DataType& output) const;
+
+  /**
+   * Return a matrix of randomly generated observations according to the
+   * probability distributions defined by this object.
+   *
+   * @return Matrix of random observations from this Normal distribution.
+   */
+  DataType Random() const;
```
sumedhghaisas: As we are shifting this dist inside ANN module, do you think name 'sample' will be more appropriate?

src/mlpack/core/dists/normal_distribution.hpp
```diff
@@ -0,0 +1,123 @@
...
+   * Create multiple Normal distributions with the given means and Standard
+   * deviations.
+   */
+  NormalDistribution(const DataType& mean, const DataType& stdDeviation);
+
+  /*
+   * Create multiple Normal distribution with the given input matrix which
+   * has pre standard deviation in the upper half and mean in the lower half.
+   */
+  NormalDistribution(const DataType& input);
```
sumedhghaisas: Rather that hardcoding the softplus transformation, it would better to control it with a boolean. This way if user is addingan7y specific non-linearity before this layer, we can accommodate that by setting the boolean False.

src/mlpack/core/dists/normal_distribution.hpp
akhandait: The `LogProbBackward` function takes care of this case as well. If this constructor is used, the `preStdDev` will be empty and it has been taken care of. Do you still feel we should remove it?

src/mlpack/methods/ann/CMakeLists.txt
```diff
@@ -16,6 +16,7 @@ add_subdirectory(layer)
 add_subdirectory(loss_functions)
 add_subdirectory(convolution_rules)
 add_subdirectory(augmented)
+add_subdirectory(dists)
```
rcurtin: Shouldn't we put this in `src/mlpack/core/dists/`?

src/mlpack/methods/ann/dists/normal_distribution.hpp
```diff
@@ -0,0 +1,135 @@
...
+namespace mlpack {
+namespace ann /** Artificial Neural Network. */ {
+
+/**
+ * Multiple univariate Normal(Gaussian) distributions.
+ *
+ * @tparam DataType Type of the input data. (arma::colvec, arma::mat,
+ *         arma::sp_mat or arma::cube).
+ */
+template <typename DataType = arma::mat>
```
rcurtin: Sorry, I'm a bit confused by this one, wouldn't this class only be used where the data type is an `arma::vec`?  (That is, a vector with one element for each dimension)

src/mlpack/methods/ann/dists/normal_distribution.hpp
```diff
@@ -0,0 +1,135 @@
...
+#ifndef MLPACK_METHODS_ANN_DISTRIBUTIONS_NORMAL_DISTRIBUTION_HPP
+#define MLPACK_METHODS_ANN_DISTRIBUTIONS_NORMAL_DISTRIBUTION_HPP
+
+#include <mlpack/prereqs.hpp>
+
+namespace mlpack {
+namespace ann /** Artificial Neural Network. */ {
+
+/**
+ * Multiple univariate Normal(Gaussian) distributions.
```
rcurtin: It would probably be a good idea to mention the `GaussianDistribution` class for the case where the covariance matrix is not diagonal.  (And it's probably also good to modify the `GaussianDistribution` comments too to mention this class.)

src/mlpack/methods/ann/dists/normal_distribution_impl.hpp
```diff
@@ -0,0 +1,98 @@
...
+}
+
+template<typename DataType>
+void NormalDistribution<DataType>::LogProbBackward(
+    const DataType&& observation, DataType&& output) const
+{
+  if (!applySoftplus)
+    output = -0.5 * join_cols((2 / stdDeviation - 2 *
+      arma::pow(mean - observation, 2) / arma::pow(stdDeviation, 3)), 2 *
+      (mean - observation) / arma::pow(stdDeviation, 2));
```
rcurtin: A little style issue, the wrapped lines here should be indented twice.  Also it would probably be nice to put braces around this if statement since it's more than one line.

src/mlpack/methods/ann/CMakeLists.txt
rcurtin: Ah, yeah, sorry I missed the discussion.  I could see it either way---I don't think it's a problem to add it to the `src/mlpack/core/dists/` folder as long as it has the same functionality as the other distributions there, but if you prefer to keep it here that's fine also.

src/mlpack/methods/ann/dists/normal_distribution.hpp
akhandait: As this is a class allowing multiple univariate normal distributions, it can be any data type. For example, two 5 x 5 matrices with means in one and standard deviations in the other. This will have 25 univariate distributions with each element of a collective sample matrix denoting a sample from the corresponding parameters.

src/mlpack/methods/ann/CMakeLists.txt
akhandait: If it's not a problem, I will discuss it with Sumedh and decide whether we should move it back to the core/dists folder.

rcurtin: Sure, that's fine, and I wouldn't have a particularly strong opinion here since I am not 100% involved in this bit.  My intuition is usually that if we can put code in more generic places, it's easier for reuse later and helps us keep the amount of code we have to maintain down.  Maybe that is possible here, maybe it doesn't make sense.

src/mlpack/methods/ann/dists/normal_distribution.hpp
akhandait: Okay, if it’s just changing the default template parameter then it’s not a problem. If I move it to core, then I will change it to vector.

src/mlpack/methods/ann/CMakeLists.txt
akhandait: @sumedhghaisas What do you think about this?

sumedhghaisas: @akhandait @rcurtin. I agree with @rcurtin, even my first choice is to put this in core/dists and we did plan that if I remember correctly. We are using NormalDistribution to define distribution over the inputs. The reason I suggested to put this distribution here (for now) is, a lot of ANN layer sensitive code was added to the distribution. For example, constructor accepting mean and variance together in 1 matrix of shape [batch_size, 2 * latent_dim], ability for the user to add SoftplusFunction over the variance to make sure its strictly positive, dealing with batching, and backward for LogProb. All these characteristic make this very ANN specific and putting it in core/dists would involved huge efforts to find the middle ground between the two.

src/mlpack/methods/ann/dists/normal_distribution.hpp
sumedhghaisas: I agree, users of both class should know the existence of the other class. :)

akhandait: I will add them in the commit after the next one.

src/mlpack/methods/ann/dists/normal_distribution_impl.hpp
akhandait: Done!

src/mlpack/methods/ann/CMakeLists.txt
akhandait: So, I guess for now we will keep it in the ann folder.

src/mlpack/methods/ann/dists/normal_distribution.hpp
akhandait: Is it okay if I keep it `arma::mat` as it will be in the ann module?

src/mlpack/methods/ann/CMakeLists.txt
rcurtin: Ah, I see what you mean.  Thanks for clarifying and sorry for making you type out the justification since you probably already wrote it in IRC earlier. :)

src/mlpack/methods/ann/dists/normal_distribution.hpp
rcurtin: Right, since it will stay in the ann folder, I see no need to make the change and my comment here can be ignored. :)

src/mlpack/methods/ann/dists/normal_distribution_impl.hpp
```diff
@@ -0,0 +1,106 @@
...
+    const DataType&& observation) const
+{
+  if (observation.size() != mean.size())
+  {
+    Log::Fatal << "NormalDistribution<>::NormalDistribution(): The size of the"
+        << "observation should be equal to the sizes of the mean and standard"
+        << "deviation." << std::endl;
+  }
+
+  return -0.5 * (arma::accu(2 * arma::log(stdDev) + arma::pow(
```
akhandait: @sumedhghaisas I think we made a mistake here, the error should be negative log probability and not log probability, shouldn't it?

src/mlpack/methods/ann/dists/normal_distribution_impl.hpp
```diff
@@ -0,0 +1,106 @@
...
+{
+  // Nothing to do here.
+}
+
+template<typename DataType>
+NormalDistribution<DataType>::NormalDistribution(
+    const DataType&& mean,
+    const DataType&& stdDev) :
+    mean(mean),
+    stdDev(stdDev)
```
rcurtin: We should probably initialize `applySoftplus` here, otherwise its value is unpredictable.

src/mlpack/methods/ann/dists/normal_distribution.hpp
```diff
@@ -0,0 +1,146 @@
...
+  /**
+   * Serialize the distribution.
+   */
+  template<typename Archive>
+  void serialize(Archive& ar, const unsigned int /* version */)
+  {
+    // We just need to serialize each of the members.
+    ar & BOOST_SERIALIZATION_NVP(mean);
+    ar & BOOST_SERIALIZATION_NVP(stdDev);
+    ar & BOOST_SERIALIZATION_NVP(preStdDev);
```
rcurtin: I think `applySoftplus` must also be serialized.

src/mlpack/methods/ann/dists/normal_distribution_impl.hpp
```diff
@@ -0,0 +1,106 @@
...
+  if (param.n_rows % 2 != 0)
+  {
+    Log::Fatal << "NormalDistribution<>::NormalDistribution(): The number of "
+        << "rows of param matrix should be even." << std::endl;
+  }
+
+  if (applySoftplus)
+    SoftplusFunction::Fn(preStdDev, stdDev);
+  else
+    stdDev = preStdDev;
```
rcurtin: Is this the only place `preStdDev` is used?  If so we can probably remove it as a class member and only use it locally here.

rcurtin: Ah, sorry, in this case you can ignore my comment. :)

akhandait:
@sumedhghaisas I think I figured out what we were doing wrong in this. :D

akhandait:
Please don't delete this. 
@rcurtin Is a comment enough for mlpack-bot to not delete this?

rcurtin:
@akhandait you can also add the 's: keep open' label and mlpack-bot will ignore it. :)

rcurtin:

PR #1444, opened 2018-06-23
Wasserstein GAN Implementation

ShikharJ:
TODO:

- [x] Implement standard WGAN policy
- [x] Implement gradient-penalized WGAN policy
- [x] Debug the entire PR

### DIFF
--- a/src/mlpack/methods/ann/gan.hpp
+++ b/src/mlpack/methods/ann/gan.hpp
@@ -14,6 +14,7 @@
 #include <mlpack/core.hpp>
 
 #include <mlpack/methods/ann/ffn.hpp>
+#include <mlpack/methods/ann/gan_policies.hpp>
 #include <mlpack/methods/ann/visitor/output_parameter_visitor.hpp>
 #include <mlpack/methods/ann/visitor/reset_visitor.hpp>
 #include <mlpack/methods/ann/visitor/weight_size_visitor.hpp>
@@ -31,7 +32,7 @@ namespace ann /** Artificial Neural Network. **/ {
  * technique can generate photographs that look at least superficially
  * authentic to human observers, having many realistic characteristics.
  *
- * For more information, see the following papers.
+ * For more information, see the following paper:
  *
  * @code
  * @article{Goodfellow14,
@@ -44,26 +45,17 @@ namespace ann /** Artificial Neural Network. **/ {
  *   eprint    = {1406.2661},
  * }
  * @endcode
- * 
- * @code
- * @article{Salimans16,
- *   author    = {Tim Salimans, Ian Goodfellow, Wojciech Zaremba,
- *                Vicki Cheung, Alec Radford and Xi Chen},
- *   title     = {Improved Techniques for Training GANs},
- *   year      = {2016},
- *   url       = {http://arxiv.org/abs/1606.03498},
- *   eprint    = {1606.03498},
- * }
- * @endcode
  *
  * @tparam Model The class type of Generator and Discriminator.
  * @tparam InitializationRuleType Type of Initializer.
  * @tparam Noise The noise function to use.
+ * @tparam PolicyType The GAN variant to be used (GAN, DCGAN, WGAN or WGANGP).
  */
 template<
   typename Model,
   typename InitializationRuleType,
-  class Noise
+  typename Noise,
+  typename PolicyType = StandardGAN
 >
 class GAN
 {
@@ -74,11 +66,13 @@ class GAN
    * @param trainData The real data.
    * @param generator Generator network.
    * @param discriminator Discriminator network.
-   * @param batchSize BatchSize to be used for training.
-   * @param generatorUpdateStep Number of steps of Discriminator training
-   *                            before updating generator.
-   * @param preTrainSize Num of preTraining step of Discriminator.
+   * @param batchSize Batch size to be used for training.
+   * @param generatorUpdateStep Number of steps to train Discriminator
+   *                            before updating Generator.
+   * @param preTrainSize Number of pre-training steps of Discriminator.
    * @param multiplier Ratio of learning rate of Discriminator to the Generator.
+   * @param clippingParameter Weight range for enforcing Lipschitz constraint.
+   * @param lambda Parameter for setting the gradient penalty.
    */
   GAN(arma::mat& trainData,
       Model& generator,
@@ -89,7 +83,9 @@ class GAN
       size_t batchSize,
       size_t generatorUpdateStep,
       size_t preTrainSize,
-      double multiplier);
+      double multiplier,
+      double clippingParameter = 0.01,
+      double lambda = 10.0);
 
   // Reset function.
   void Reset();
@@ -99,29 +95,103 @@ class GAN
   void Train(OptimizerType& Optimizer);
 
   /**
-   * Evaluate function for the GAN gives the performance of the GAN on the
+   * Evaluate function for the Standard GAN and DCGAN.
+   * This function gives the performance of the Standard GAN or DCGAN on the
    * current input.
    *
    * @param parameters The parameters of the network.
    * @param i Index of the current input.
+   * @param batchSize Variable to store the present number of inputs.
    */
-  double Evaluate(const arma::mat& parameters,
-                  const size_t i,
-                  const size_t batchSize);
+  template<typename Policy = PolicyType>
+  typename std::enable_if<std::is_same<Policy, StandardGAN>::value ||
+                          std::is_same<Policy, DCGAN>::value, double>::type
+  Evaluate(const arma::mat& parameters,
+           const size_t i,
+           const size_t batchSize);
 
   /**
-   * Gradient function for GAN.
+   * Evaluate function for the WGAN.
+   * This function gives the performance of the WGAN on the current input.
+   *
+   * @param parameters The parameters of the network.
+   * @param i Index of the current input.
+   * @param batchSize Variable to store the present number of inputs.
+   */
+  template<typename Policy = PolicyType>
+  typename std::enable_if<std::is_same<Policy, WGAN>::value,
+                          double>::type
+  Evaluate(const arma::mat& parameters,
+           const size_t i,
+           const size_t batchSize);
+
+  /**
+   * Evaluate function for the WGAN-GP.
+   * This function gives the performance of the WGAN-GP on the current input.
+   *
+   * @param parameters The parameters of the network.
+   * @param i Index of the current input.
+   * @param batchSize Variable to store the present number of inputs.
+   */
+  template<typename Policy = PolicyType>
+  typename std::enable_if<std::is_same<Policy, WGANGP>::value,
+                          double>::type
+  Evaluate(const arma::mat& parameters,
+           const size_t i,
+           const size_t batchSize);
+
+  /**
+   * Gradient function for Standard GAN and DCGAN.
    * This function passes the gradient based on which network is being
    * trained, i.e., Generator or Discriminator.
    * 
    * @param parameters present parameters of the network.
    * @param i Index of the predictors.
    * @param gradient Variable to store the present gradient.
+   * @param batchSize Variable to store the present number of inputs.
    */
-  void Gradient(const arma::mat& parameters,
-                const size_t i,
-                arma::mat& gradient,
-                const size_t batchSize);
+  template<typename Policy = PolicyType>
+  typename std::enable_if<std::is_same<Policy, StandardGAN>::value ||
+                          std::is_same<Policy, DCGAN>::value, void>::type
+  Gradient(const arma::mat& parameters,
+           const size_t i,
+           arma::mat& gradient,
+           const size_t batchSize);
+
+  /**
+   * Gradient function for WGAN.
+   * This function passes the gradient based on which network is being
+   * trained, i.e., Generator or Discriminator.
+   * 
+   * @param parameters present parameters of the network.
+   * @param i Index of the predictors.
+   * @param gradient Variable to store the present gradient.
+   * @param batchSize Variable to store the present number of inputs.
+   */
+  template<typename Policy = PolicyType>
+  typename std::enable_if<std::is_same<Policy, WGAN>::value, void>::type
+  Gradient(const arma::mat& parameters,
+           const size_t i,
+           arma::mat& gradient,
+           const size_t batchSize);
+
+  /**
+   * Gradient function for WGAN-GP.
+   * This function passes the gradient based on which network is being
+   * trained, i.e., Generator or Discriminator.
+   * 
+   * @param parameters present parameters of the network.
+   * @param i Index of the predictors.
+   * @param gradient Variable to store the present gradient.
+   * @param batchSize Variable to store the present number of inputs.
+   */
+  template<typename Policy = PolicyType>
+  typename std::enable_if<std::is_same<Policy, WGANGP>::value,
+                          void>::type
+  Gradient(const arma::mat& parameters,
+           const size_t i,
+           arma::mat& gradient,
+           const size_t batchSize);
 
   /**
    * Shuffle the order of function visitation. This may be called by the
@@ -167,7 +237,7 @@ class GAN
   //! Locally stored Discriminator network.
   Model& discriminator;
   //! Locally stored Initializer.
-  InitializationRuleType  initializeRule;
+  InitializationRuleType initializeRule;
   //! Locally stored Noise function
   Noise noiseFunction;
   //! Locally stored input dimension of the Generator network.
@@ -186,6 +256,10 @@ class GAN
   size_t preTrainSize;
   //! Locally stored learning rate ratio for Generator network.
   double multiplier;
+  //! Locally stored weight clipping parameter.
+  double clippingParameter;
+  //! Locally stored lambda parameter.
+  double lambda;
   //! Locally stored reset parameter.
   bool reset;
   //! Locally stored delta visitor.
@@ -208,6 +282,8 @@ class GAN
   arma::mat gradientDiscriminator;
   //! Locally stored gradient for noise data in the predictors.
   arma::mat noiseGradientDiscriminator;
+  //! Locally stored norm of the gradient of Discriminator.
+  arma::mat normGradientDiscriminator;
   //! Locally stored noise using the noise function.
   arma::mat noise;
   //! Locally stored gradient for Generator.
--- a/src/mlpack/methods/ann/gan/gan.hpp
+++ b/src/mlpack/methods/ann/gan/gan.hpp
@@ -8,13 +8,13 @@
  * 3-clause BSD license along with mlpack. If not, see
  * http://www.opensource.org/licenses/BSD-3-Clause for more information.
  */
-#ifndef MLPACK_METHODS_ANN_GAN_HPP
-#define MLPACK_METHODS_ANN_GAN_HPP
+#ifndef MLPACK_METHODS_ANN_GAN_GAN_HPP
+#define MLPACK_METHODS_ANN_GAN_GAN_HPP
 
 #include <mlpack/core.hpp>
 
 #include <mlpack/methods/ann/ffn.hpp>
-#include <mlpack/methods/ann/gan_policies.hpp>
+#include <mlpack/methods/ann/gan/gan_policies.hpp>
 #include <mlpack/methods/ann/visitor/output_parameter_visitor.hpp>
 #include <mlpack/methods/ann/visitor/reset_visitor.hpp>
 #include <mlpack/methods/ann/visitor/weight_size_visitor.hpp>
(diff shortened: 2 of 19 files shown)


src/mlpack/methods/ann/gan_policies/standard_gan.hpp
```diff
@@ -0,0 +1,180 @@
...
+ * Implementation of the standard GAN policies.
+ *
+ * mlpack is free software; you may redistribute it and/or modify it under the
+ * terms of the 3-clause BSD license.  You should have received a copy of the
+ * 3-clause BSD license along with mlpack.  If not, see
+ * http://www.opensource.org/licenses/BSD-3-Clause for more information.
+ */
+#ifndef MLPACK_METHODS_ANN_GAN_POLICIES_STANDARD_GAN_HPP
+#define MLPACK_METHODS_ANN_GAN_POLICIES_STANDARD_GAN_HPP
+
```
zoq: Not sure, which one is is going to be picked up first, so perhaps we have to include ``gan.hpp``here.

ShikharJ: I don't think it fixes our issue I tried doing the above to no avail.

zoq:
Really strange, another idea was that the parameter might be shadowed by the reset parameter from the FFN class; but that's not the case, the policy can't even access a parameter that's only available in the GAN class.

So here is another idea, what about using SFINAE + enable_if, something like:


```c++
gan.hpp
-------------------------------------------
class StandardGAN { /* Dummy GAN class */ };
class WGAN { /* Dummy GAN class */ };

template<typename PolicyType>
class GAN
{
  GAN(..);

  Gradient(...)
}


standard_gan.hpp
-------------------------------------------

template<typename Policy = PolicyType>
  static typename std::enable_if<
      std::is_same<Policy, StandardGAN>::value, void>::type
GAN::Gradient(...)

wgan.hpp
-------------------------------------------

template<typename Policy = PolicyType>
  static typename std::enable_if<
      std::is_same<Policy, WGAN>::value, void>::type
GAN::Gradient(...)
```

We did something similar for the convolution rules. Let me know what you think.

ShikharJ:
@zoq I have pushed in the new changes, still the build is failing due to some reason I'm not able to figure. Can you take a look?

zoq:
>   Cannot find source file:
>     /home/travis/build/mlpack/mlpack/src/mlpack/methods/ann/gan_policies.hpp

Looks like this file wasn't added, can you push the file as well?

ShikharJ:
@zoq Are you sure it works? I'm still facing the same issues.

zoq:
It's possible that I missed something, if you push the latest code, I can take another look, no problem.

ShikharJ:
@zoq @lozhnikov This is ready for a review.

ShikharJ:
Here are some results I get on the latest code for the full `70000` image `MNIST` dataset (all trained under 7 hours!):
`GAN`
output_gan
`DCGAN`
output_dcgan
`WGAN`
output_wgan
`WGAN-GP`
output_wgangp

src/mlpack/methods/ann/gan_impl.hpp
```diff
@@ -153,17 +180,134 @@ double GAN<Model, InitializationRuleType, Noise>::Evaluate(
...
+
+  currentTarget = arma::mat(discriminator.responses.memptr() + numFunctions,
+      1, batchSize, false, false);
+  res += discriminator.outputLayer.Forward(
+      std::move(boost::apply_visitor(
+      outputParameterVisitor,
+      discriminator.network.back())), std::move(currentTarget));
+
+  // Gradient Penalty is calculated here.
+  double epsilon = math::Random();
```
zoq: Should we use a static parameter if we aren't training?

src/mlpack/methods/ann/gan_impl.hpp
```diff
@@ -227,30 +462,128 @@ Gradient(const arma::mat& /*parameters*/, const size_t i, arma::mat& gradient,
...
+    generator.ResetGradients(gradientGenerator);
+    generator.Gradient(generator.parameter, 0, gradientGenerator, batchSize);
+
+    gradientGenerator *= multiplier;
+  }
+
+  counter++;
+  currentBatch++;
+
+  if (counter * batchSize >= numFunctions)
```
zoq: Might be a good idea to add a comment here, that this is used to control the pretrain step.

src/mlpack/methods/ann/loss_functions/earth_mover_distance.hpp
```diff
@@ -0,0 +1,84 @@
...
+ * @tparam InputDataType Type of the input data (arma::colvec, arma::mat,
+ *         arma::sp_mat or arma::cube).
+ * @tparam OutputDataType Type of the output data (arma::colvec, arma::mat,
+ *         arma::sp_mat or arma::cube).
+ */
+template <
+    typename InputDataType = arma::mat,
+    typename OutputDataType = arma::mat
+>
+class EarthMoverDistance
```
zoq: I know this is a super simple loss function, but do you mind to add a test to: https://github.com/mlpack/mlpack/blob/master/src/mlpack/tests/loss_functions_test.cpp.

src/mlpack/methods/ann/gan_impl.hpp
ShikharJ: I don't think it's necessary, but I don't mind adding it either, your call.

ShikharJ:
@zoq This is updated now.

src/mlpack/methods/ann/gan_impl.hpp
zoq: I'm not too sure either, I think at least it will not decrease the performance we currently see, so if you see a simple solution, I would go for it.

zoq:
I wonder if the test time increased, let's wait for the results.

src/mlpack/methods/ann/gan.hpp
```diff
@@ -99,15 +95,33 @@ class GAN
...
+  typename std::enable_if<std::is_same<Policy, StandardGAN>::value ||
+                          std::is_same<Policy, DCGAN>::value, double>::type
+  Evaluate(const arma::mat& parameters,
+           const size_t i,
+           const size_t batchSize);
+
+  template<typename Policy = PolicyType>
+  typename std::enable_if<std::is_same<Policy, WGAN>::value,
+                          double>::type
+  Evaluate(const arma::mat& parameters,
```
zoq: Can you add a comment for the other Evaluate forms as well, I know this is redundant, but doxygen will pick this one up but unfortunately doxygen isn't going to use the first comment.

src/mlpack/methods/ann/gan.hpp
```diff
@@ -117,11 +131,30 @@ class GAN
...
+   * @param batchSize Variable to store the present number of inputs.
    */
-  void Gradient(const arma::mat& parameters,
-                const size_t i,
-                arma::mat& gradient,
-                const size_t batchSize);
+  template<typename Policy = PolicyType>
+  typename std::enable_if<std::is_same<Policy, StandardGAN>::value ||
+                          std::is_same<Policy, DCGAN>::value, void>::type
+  Gradient(const arma::mat& parameters,
```
zoq: See comment above.

src/mlpack/methods/ann/gan_impl.hpp
ShikharJ: Since the original paper doesn't ask for that, nor have I seen any implementations for it, I'd prefer if we skip.

zoq: Fine for me.

ShikharJ:
@zoq This is updated again.

src/mlpack/tests/wgan_test.cpp
```diff
@@ -0,0 +1,290 @@
...
+      << " stepSize = " << stepSize << std::endl
+      << " numEpoches = " << numEpoches << std::endl
+      << " tolerance = " << tolerance << std::endl
+      << " shuffle = " << shuffle << std::endl;
+
+  arma::mat trainData;
+  trainData.load("mnist_first250_training_4s_and_9s.arm");
+  Log::Info << arma::size(trainData) << std::endl;
+
+  if (datasetMaxCols > 0)
```
zoq: We could remove the check here to solve the static code analysis issue, same for the other tests.

zoq:
This is just a thought but do you think it would make sense to split ``gan_impl.hpp`` into multiple files?

src/mlpack/methods/ann/gan/gan.hpp
```diff
@@ -0,0 +1,357 @@
...
+   * @param discriminator Discriminator network.
+   * @param batchSize Batch size to be used for training.
+   * @param generatorUpdateStep Number of steps to train Discriminator
+   *                            before updating Generator.
+   * @param preTrainSize Number of pre-training steps of Discriminator.
+   * @param multiplier Ratio of learning rate of Discriminator to the Generator.
+   * @param clippingParameter Weight range for enforcing Lipschitz constraint.
+   * @param lambda Parameter for setting the gradient penalty.
+   */
+  GAN(arma::mat& trainData,
```
rcurtin: Should this be `const`?

src/mlpack/methods/ann/gan/gan.hpp
```diff
@@ -0,0 +1,357 @@
...
+namespace mlpack {
+namespace ann /** Artificial Neural Network. **/ {
+
+/**
+ * The implementation of the standard GAN module. Generative Adversarial
+ * Networks (GANs) are a class of artificial intelligence algorithms used
+ * in unsupervised machine learning, implemented by a system of two neural
+ * networks contesting with each other in a zero-sum game framework. This
+ * technique can generate photographs that look at least superficially
+ * authentic to human observers, having many realistic characteristics.
```
rcurtin: It would be really useful to have some example usage here or in some tutorial somewhere.  I think just from the constructor I understand how I would use it, but examples are always great starting points for users.

src/mlpack/methods/ann/gan/gan.hpp
```diff
@@ -0,0 +1,357 @@
...
+      Model& discriminator,
+      InitializationRuleType initializeRule,
+      Noise noiseFunction,
+      size_t noiseDim,
+      size_t batchSize,
+      size_t generatorUpdateStep,
+      size_t preTrainSize,
+      double multiplier,
+      double clippingParameter = 0.01,
+      double lambda = 10.0);
```
rcurtin: It's a little pedantic but you could make all these inputs `const` also, just to match the rest of the codebase.

src/mlpack/methods/ann/gan/gan.hpp
```diff
@@ -0,0 +1,357 @@
...
+   * @param parameters The parameters of the network.
+   * @param i Index of the current input.
+   * @param batchSize Variable to store the present number of inputs.
+   */
+  template<typename Policy = PolicyType>
+  typename std::enable_if<std::is_same<Policy, StandardGAN>::value ||
+                          std::is_same<Policy, DCGAN>::value, double>::type
+  Evaluate(const arma::mat& parameters,
+           const size_t i,
+           const size_t batchSize);
```
rcurtin: The various `enable_if`s make me wonder if it's better to put the `Evaluate()` function in the `Policy` class, and have this just be a wrapper to call out to `Policy::Evaluate()`.  I know it is very late in the design process so maybe it is not worth revisiting, and I think this can be fine, but maybe it is an opportunity for improvement later or something.

src/mlpack/methods/ann/gan/gan_impl.hpp
```diff
@@ -276,9 +440,14 @@ Predict(arma::mat&& input, arma::mat& output)
...
+template<
+  typename Model,
+  typename InitializationRuleType,
+  typename Noise,
+  typename PolicyType
+>
 template<typename Archive>
-void GAN<Model, InitializationRuleType, Noise>::
+void GAN<Model, InitializationRuleType, Noise, PolicyType>::
 serialize(Archive& ar, const unsigned int /* version */)
```
rcurtin: Actually this will not work right; since the `GAN` class is holding a `Model&` for the discriminator and generator networks, we can't serialize this directly since there is no constructor that takes neither.  I might suggest instead holding a `Model` directly, then allowing the user to pass one with `std::move()` (to avoid a copy) or a `const` one (which will incur a copy).

src/mlpack/methods/ann/gan/gan.hpp
ShikharJ: We can make it to be `const`, however I had checked through all the layers, and wherever the data is passed, no changes would be done on it. I'll make the changes.

src/mlpack/methods/ann/gan/gan.hpp
ShikharJ: Agreed.

src/mlpack/methods/ann/gan/gan.hpp
ShikharJ: Actually, we started the implementation by doing that, but unfortunately the templatized classes were unable to access the `GAN` class' internal members. So we chose this approach instead.

src/mlpack/methods/ann/gan/gan_impl.hpp
rcurtin: A default constructor unfortunately won't work as the internal reference members must be set at construction time, and you'll have no models to set the discriminator or generator networks' references to.

src/mlpack/methods/ann/gan/gan.hpp
rcurtin: Right, another idea would be to make the templatized classes accept all the internal `GAN` class members that would be needed in their signatures.  I think there would be no issue with that, but it is up to you.  I'm not sure if it would be uglier or prettier.

ShikharJ: That's exactly the next approach we went to. But @zoq and I ended up agreeing that since `Evaluate()` and `Gradient()` tend to access all the members, it would look a lot worse and using it would be even more frustrating.

ShikharJ:
@rcurtin Thanks for your comments. This has been updated.

src/mlpack/methods/ann/gan/gan.hpp
rcurtin: Right, I guess you have already had this discussion then.  Sorry to make you write it down again :)

zoq:
Can you take a look at the failing tests?

src/mlpack/methods/ann/gan/gan_impl.hpp
```diff
@@ -23,28 +23,37 @@
...
     size_t preTrainSize,
-    double multiplier):
+    const double multiplier,
+    const double clippingParameter,
+    const double lambda):
     predictors(predictors),
-    generator(generator),
-    discriminator(discriminator),
+    generator(std::move(generator)),
+    discriminator(std::move(discriminator)),
```
ShikharJ: @zoq I'm not sure this implementation is correct. Might be the cause of the failure?

zoq:
Can you implement a move/copy constructor? See the FFN class for an example.

ShikharJ:
@rcurtin Could you advise regarding serialization issue? I have tried using `std::move`, but that seemingly leads to memory violation issues.

rcurtin:
Does whatever you are calling `std::move()` on have a correct move operator/move constructor?  If not, maybe some pointers are being copied in an invalid way.  If there's no move operator, the default move operator will be used, which may not be what is desired.

ShikharJ:
@rcurtin Model is basically the `FFN<>()` layer, and that does have a copy and move constructor. Still the issue arises.

zoq:
@ShikharJ looks like Travis is fine, do you see any issue on your local machine?

zoq:
So did you test:

```c++
GAN(arma::mat& trainData, Model generator, Model discriminator, ...)
```
```c++
GAN<...> gan(data, std::move(generator), std::move(discriminator))
````
this should solve the serialization issue. You can also use:
```c++
GAN<...> gan(data, generator, discriminator)
```
but as Ryan pointed out this will include a copy, which might be neglectable.

ShikharJ:
@zoq I did. But in either case, I'll have to change:
```
//! Locally stored Generator network.
  Model& generator;
  //! Locally stored Discriminator network.
  Model& discriminator;
```
to
```
//! Locally stored Generator network.
  Model generator;
  //! Locally stored Discriminator network.
  Model discriminator;
```
which was causing `Travis` to fail.

rcurtin:
I'm confused on the problem here---I don't see any failing Travis build so there's nothing for me to help debug with.  My advice would be, if it is segfaulting on Travis only, see if you can reproduce it.  Running with valgrind might expose some invalid memory accesses, which are likely to be the problem.

zoq:
I guess, @ShikharJ reverted the change and squashed the commits?

Can you remove the move constructor of the FFN class:

```c++
template<typename OutputLayerType, typename InitializationRuleType,
         typename... CustomLayers>
FFN<OutputLayerType, InitializationRuleType, CustomLayers...>::FFN(
    FFN&& network):
```

and test again?

ShikharJ:
@zoq I tried your suggestion, and that doesn't work as well.

rcurtin:
For what it's worth, it's really hard to try and help debug with input like "it didn't work"---at least for me, to be actually helpful, I need detailed information on what you tried, what the error was, and so forth.  I can chime in further helping with this one if you can reproduce the issue locally and try with valgrind and see what that yields.  Without good information I am just making guesses at what is wrong (and I have already guessed above, so I have no new guesses).

src/mlpack/methods/ann/gan/gan_impl.hpp
```diff
@@ -0,0 +1,523 @@
...
+{
+  // Insert IdentityLayer for joining the Generator and Discriminator.
+  discriminator.network.insert(
+      discriminator.network.begin(),
+      new IdentityLayer<>());
+
+  counter = 0;
+  currentBatch = 0;
+
+  discriminator.deterministic = generator.deterministic = true;
```
zoq: We should use ``this.discriminator`` or change the parameter name of to ``discriminatorModel`` or something like, that should fix the issue we have seen before.

zoq:
As for me, the main confusion was that the issue disappeared, without any visible commit. I think there is no issue if we have commits in the history that show the coding process, that way we could track at which time an issue/solution was introduced. We can always stash the commits at the merge step if we think it might be helpful e.g. to keep the history clean. Let me know what you think.

src/mlpack/methods/ann/gan/gan.hpp
```diff
@@ -0,0 +1,365 @@
...
+   * @param batchSize Batch size to be used for training.
+   * @param generatorUpdateStep Number of steps to train Discriminator
+   *                            before updating Generator.
+   * @param preTrainSize Number of pre-training steps of Discriminator.
+   * @param multiplier Ratio of learning rate of Discriminator to the Generator.
+   * @param clippingParameter Weight range for enforcing Lipschitz constraint.
+   * @param lambda Parameter for setting the gradient penalty.
+   */
+  GAN(arma::mat& trainData,
+      Model& generator,
```
zoq: This should be: ``Model generator`` and  ``Model discriminator``.

ShikharJ:
@rcurtin Apologies, actually I was in a hurry, and had to comment from my phone. Sure, I'll try running `valgrind` and see what result I obtain.

zoq:
@ShikharJ do you think that this is a random failure or is there something wrong with the GAN test case?

ShikharJ:
@zoq I really don't know. This particular test was written by Kris and Mikhail, but I'm not sure why changing the constructor to use move semantics would cause it to fail. I'm able to reproduce this in my system as well.

zoq:
We could set the same random seed and compare the results, with and without the move operations.

src/mlpack/tests/wgan_test.cpp
```diff
@@ -268,7 +268,7 @@ BOOST_AUTO_TEST_CASE(WGANGPMNISTTest)
     arma::mat samples;
     noise.imbue( [&]() { return noiseFunction(); } );
 
-    generator.Forward(noise, samples);
+    wgan_gp.Generator().Forward(noise, samples);
```
zoq: Please use camel casing for all names. See https://github.com/mlpack/mlpack/wiki/DesignGuidelines#naming-conventions for all information.

src/mlpack/methods/ann/gan/wgangp_impl.hpp
```diff
@@ -0,0 +1,302 @@
...
+  typename InitializationRuleType,
+  typename Noise,
+  typename PolicyType
+>
+template<typename GradType, typename Policy>
+typename std::enable_if<std::is_same<Policy, WGANGP>::value,
+                        double>::type
+GAN<Model, InitializationRuleType, Noise, PolicyType>::
+EvaluateWithGradient(const arma::mat& /* parameters */,
+         const size_t i,
```
zoq: Can you align the parameter with the first one?

src/mlpack/tests/wgan_test.cpp
ShikharJ: I think it is done exactly in the way specified? I don't see an issue here.

zoq: Instead of ``wgan_gp`` we should use ``wganGP``or something similar.

zoq:
Really nice contribution, thanks!

PR #1452, opened 2018-06-27
Make changes suggested in sampling PR

akhandait:
This is a follow-up PR to the sampling one where Ryan suggested some changes.

### DIFF
--- a/src/mlpack/methods/ann/layer/reparametrization_impl.hpp
+++ b/src/mlpack/methods/ann/layer/reparametrization_impl.hpp
@@ -74,7 +74,7 @@ void Reparametrization<InputDataType, OutputDataType>::Backward(
   if (includeKl)
   {
     arma::Mat<eT> klBack;
-    klBackward(std::move(klBack));
+    KLBackward(std::move(klBack));
     g = join_cols(gy % std::move(gaussianSample) % g, gy) + std::move(klBack);
   }
   else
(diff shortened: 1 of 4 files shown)


src/mlpack/methods/ann/layer/reparametrization_impl.hpp
```diff
@@ -83,19 +83,16 @@ void Reparametrization<InputDataType, OutputDataType>::Backward(
 
 template<typename InputDataType, typename OutputDataType>
 template<typename InputType>
-double Reparametrization<InputDataType, OutputDataType>::klForward(
-    const InputType&& input)
+inline double Reparametrization<InputDataType, OutputDataType>::KLForward(
+    const InputType&& mean, const InputType&& stdDev)
```
rcurtin: Oh, I meant actually inlining these by removing `KLForward()` and placing the implementation where you've used it in `Loss()`.  Sorry, I should have been clearer. :)

akhandait: No, it's not your fault. You were clear enough. :) Later @zoq commented on that PR and even he mentioned about inlining this function. I took that as making the function inline. Though, now I am not sure what he exactly meant. @zoq Can you please clarify what you meant by inline in this case? Did you mean inlining it into the `Loss()` function?

zoq: Sorry for the confusion, what I meant was to implement ``KLForward`` into the ``Loss`` function; that way we can avoid the extra split. Does this sound reasonable?

akhandait: I can't think of a case where the `KLForward()` and `KLBackward()` will be used besides the `Forward()` and `Backward()` functions of this layer. So, I guess you both are correct, there isn't any need to have them as seperate functions. I will inline this into the `Loss()` function.

rcurtin:
Thanks again! :+1:

PR #1457, opened 2018-07-02
Fix invalid error in local_coordinate_coding

rcurtin:
I have removed a check that would mean that the user must pass either 'initial_dictionary' or 'input_model' to any call of the program.  This makes it impossible to run LCC with a random initial dictionary.

I think this would be sufficient for releasing 3.0.3 as a bugfix, but we can also wait a little if there are more fixes or functionality.


rcurtin:
Agreed, let's give it a week and see if there is anything else.

PR #1458, opened 2018-07-02
[WIP] Refactor CF DecompositionPolicy and add BiasSVD, SVD++

Wenhao-H:
This PR refactors CF `DecompositionPolicy` and add `BiasSVDPolicy`, `SVDPlusPlusPolicy`.

To do:

- [x]  Refactor all existing decomposition policies.
- [ ] Add `BiasSVDPolicy`, `SVDPlusPlusPolicy`.
- [ ] Add tests and update comments.


src/mlpack/methods/cf/cf_model.hpp
```diff
@@ -0,0 +1,216 @@
...
+    REGULARIZED_SVD,
+    SVD_COMPLETE,
+    SVD_INCOMPLETE
+  };
+
+ private:
+  //! The type of decomposition policy.
+  size_t decompositionPolicy;
+  //! Non-NULL if using NMFPolicy.
+  CFType<NMFPolicy>* nmfCF;
```
lozhnikov: How do you think is it a good idea to refactor this in terms of `boost::variant` like `src/mlpack/methods/neighbor_search/ns_model.hpp`?

src/mlpack/methods/cf/decomposition_policies/batch_svd_method.hpp
```diff
@@ -73,6 +69,89 @@ class BatchSVDPolicy
       svdbatch.Apply(cleanedData, rank, w, h);
     }
   }
+
+  /**
+   * Return predicted rating given user ID and item ID.
```
lozhnikov: If I am not mistaken, this code duplicates six times. I think it is a good idea to add another abstraction layer. How do you think?

src/mlpack/methods/cf/cf_model.hpp
Wenhao-H: Yes I will work on that. Thanks!

src/mlpack/methods/cf/decomposition_policies/batch_svd_method.hpp
Wenhao-H: Hmm I understand what you mean but I am not sure how to add an abstraction layer. If we are going to add another layer (e.g. maybe a wrapper class), we also need to consider how to extend it to `BiasSVD` and `SVD++` which will be implemented.

lozhnikov: I think that should become clear as soon as `BiasSVD` and `SVD++` are implemented.
Right now I've got the following idea:
Actually, we have 2 policies: `CholevskyPolicy` and `SVDPolicy`. Meanwhile, `CholevskyPolicy` depends on matrices `W` and `H`, so it has another template argument which can be equal to `BatchSVDPolicy`, `NMFPolicy`, `RandomizedSVDPolicy` and so on.
For example:
```
template<typename DecompositionPolicy>
class CholevskyNeighborhoodPolicy
{
  DecompositionPolicy p;

  void Apply(.......)
  {
    p.Apply(............);
  }

  double GetRating(..........);
  void GetNeighbourhood(.........);
};
```
Thus, let's wait for `BiasSVD` and `SVD++`, then maybe I'll suggest something else.

Wenhao-H: Okay! Many thanks!

Wenhao-H:
I added the implementation of `BiasSVD` and a few tests. The implementation of `BiasSVD` is modification on the implementation of `RegularizedSVD`.

src/mlpack/methods/bias_svd/bias_svd_function_impl.hpp
```diff
@@ -0,0 +1,398 @@
...
+        for (size_t i = 0; i < rank - 1; ++i)
+        {
+          #pragma omp atomic
+          iterate(i, user) -= userVecUpdate(i);
+          #pragma omp atomic
+          iterate(i, item) -= itemVecUpdate(i);
+        }
+        #pragma omp atomic
+        iterate(rank, user) -= userBiasUpdate;
+        iterate(rank, item) -= itemBiasUpdate;
```
lozhnikov: You should add `#pragma omp atomic` here again.

src/mlpack/methods/bias_svd/bias_svd_function_impl.hpp
```diff
@@ -0,0 +1,398 @@
...
+                  parameters.col(item).subvec(0, rank - 1));
+    double ratingErrorSquared = ratingError * ratingError;
+
+    // Calculate the regularization penalty corresponding to the parameters.
+    double userVecNorm = arma::norm(parameters.col(user), 2);
+    double itemVecNorm = arma::norm(parameters.col(item), 2);
+    double regularizationError = lambda * (userVecNorm * userVecNorm +
+                                           itemVecNorm * itemVecNorm + 
+                                           userBias * userBias +
+                                           itemBias + itemBias);
```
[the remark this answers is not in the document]
Wenhao-H: Yes you are right. I missed that. Thanks!

src/mlpack/methods/bias_svd/bias_svd_function_impl.hpp
```diff
@@ -0,0 +1,398 @@
...
+    gradient.col(user).subvec(0, rank - 1) +=
+        2 * (lambda * parameters.col(user).subvec(0, rank - 1) -
+        ratingError * parameters.col(item).subvec(0, rank - 1));
+    gradient.col(item).subvec(0, rank - 1) +=
+        2 * (lambda * parameters.col(item).subvec(0, rank - 1) -
+        ratingError * parameters.col(user).subvec(0, rank - 1));
+    gradient(user, rank) +=
+        2 * (lambda * parameters(user, rank) - ratingError);
+    gradient(item, rank) +=
+        2 * (lambda * parameters(item, rank) - ratingError);
```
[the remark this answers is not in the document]
Wenhao-H: A good summary of `BiasSVD` algorithm can be found at the top of [this page](http://surprise.readthedocs.io/en/stable/matrix_factorization.html).

Wenhao-H:
`CFPredictSVDPPTest` passed but `RecommendationAccuracySVDPPTest` failed.  I think the hyperparameters of `SVDPlusPlus` need to be tuned. I will try to find better default hyperparameters for `SVDPlusPlus`.

lozhnikov:
@Wenhao-H Sorry for the slow response, I am still looking through the code. There are a lot of things to check, so it should take a while.

Wenhao-H:
@lozhnikov No problem. At the moment I am trying to correct the Travis Test failures, and add test cases and comments.

zoq:
I'm wondering did you test the code with ``OMP_NUM_THREADS=1``?

Wenhao-H:
@zoq No but I will do that soon. Does something go wrong?

src/mlpack/methods/svdplusplus/svdplusplus_function_impl.hpp
```diff
@@ -0,0 +1,561 @@
...
+              stepSize * (lambda *
+              iterate.col(implicitStart + it.row()).subvec(0, rank - 1) -
+              ratingError / std::sqrt(implicitCount) *
+              iterate.col(item).subvec(0, rank - 1));
+          implicitItems(implicitIndex) = it.row();
+        }
+
+        // Gradient is non-zero only for the parameter columns corresponding to
+        // the example.
+        for (size_t i = 0; i < rank; ++i)
```
lozhnikov: I think it is wise to vectorize the loops below and use
```
#pragma omp critical
{
  iterate(i, user) -= ...
  iterate(i, item) -= ...
}
```
instead of many `omp atomic`.

src/mlpack/methods/bias_svd/bias_svd_function_impl.hpp
```diff
@@ -0,0 +1,396 @@
...
+    const double rating = data(2, i);
+    const double userBias = parameters(rank, user);
+    const double itemBias = parameters(rank, item);
+    double ratingError = rating - userBias - itemBias -
+        arma::dot(parameters.col(user).subvec(0, rank - 1),
+                  parameters.col(item).subvec(0, rank - 1));
+
+    // Gradient is non-zero only for the parameter columns corresponding to the
+    // example.
+    for (size_t j = 0; j < rank; ++j)
```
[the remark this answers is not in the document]
Wenhao-H: That is how it was implemented at first. But the problem is that `subvec()` method is not supported when `GradType` is `sp_mat`. So I changed it to loop instead.

src/mlpack/methods/bias_svd/bias_svd_function_impl.hpp
```diff
@@ -0,0 +1,396 @@
...
+    const double itemBias = parameters(rank, item);
+    double ratingError = rating - userBias - itemBias -
+        arma::dot(parameters.col(user).subvec(0, rank - 1),
+                  parameters.col(item).subvec(0, rank - 1));
+
+    double lambda = function.Lambda();
+
+    // Gradient is non-zero only for the parameter columns corresponding to the
+    // example.
+    parameters.col(user).subvec(0, rank - 1) -= stepSize * (
```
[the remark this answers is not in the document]
Wenhao-H: Hmm I think it's better to just multiply the gradient by 2 to avoid confusion here.

src/mlpack/methods/svdplusplus/svdplusplus_function_impl.hpp
Wenhao-H: Yes but changing from `atomic` to `critical` may slow down the program execution.

rcurtin:
Don't worry about the failing LaTeX check, it'll be fixed with #1488.

Wenhao-H:
It seems that the travis build failed because the testing exceeds the time limit. Is that caused by that the newly added test cases take too much time?

lozhnikov:
@Wenhao-H I tried to run the test on my machine. `BiasSVD` took 81 sec. Could you reduce it somehow?

src/mlpack/methods/bias_svd/bias_svd_function.hpp
```diff
@@ -0,0 +1,170 @@
...
+ * An implementation of the BiasSVDFunction class.
+ *
+ * mlpack is free software; you may redistribute it and/or modify it under the
+ * terms of the 3-clause BSD license.  You should have received a copy of the
+ * 3-clause BSD license along with mlpack.  If not, see
+ * http://www.opensource.org/licenses/BSD-3-Clause for more information.
+ */
+
+#ifndef MLPACK_METHODS_BIAS_SVD_BIAS_FUNCTION_SVD_HPP
+#define MLPACK_METHODS_BIAS_SVD_BIAS_FUNCTION_SVD_HPP
```
lozhnikov: I think the name should correspond to the file name. So, I suggest to change that to `MLPACK_METHODS_BIAS_SVD_BIAS_SVD_FUNCTION_HPP`.

src/mlpack/methods/bias_svd/bias_svd_function_impl.hpp
```diff
@@ -0,0 +1,411 @@
...
+ *
+ * An implementation of the BiasSVDFunction class.
+ *
+ * mlpack is free software; you may redistribute it and/or modify it under the
+ * terms of the 3-clause BSD license.  You should have received a copy of the
+ * 3-clause BSD license along with mlpack.  If not, see
+ * http://www.opensource.org/licenses/BSD-3-Clause for more information.
+ */
+#ifndef MLPACK_METHODS_BIAS_SVD_BIAS_FUNCTION_SVD_IMPL_HPP
+#define MLPACK_METHODS_BIAS_SVD_BIAS_FUNCTION_SVD_IMPL_HPP
```
lozhnikov: Same comment.

src/mlpack/methods/bias_svd/bias_svd_function_impl.hpp
```diff
@@ -0,0 +1,411 @@
...
+    cost += (ratingErrorSquared + regularizationError);
+  }
+
+  return cost;
+}
+
+template <typename MatType>
+double BiasSVDFunction<MatType>::Evaluate(const arma::mat& parameters,
+                                                 const size_t start,
+                                                 const size_t batchSize) const
```
lozhnikov: Could you fix spaces here?

src/mlpack/methods/bias_svd/bias_svd_function_impl.hpp
```diff
@@ -0,0 +1,411 @@
...
+template<typename MatType>
+void BiasSVDFunction<MatType>::Shuffle()
+{
+  data = data.cols(arma::shuffle(arma::linspace<arma::uvec>(0, data.n_cols - 1,
+      data.n_cols)));
+}
+
+template <typename MatType>
+double BiasSVDFunction<MatType>::Evaluate(const arma::mat& parameters) const
+{
```
lozhnikov: Actually, we've got a lot of almost duplicated code such as `RegularizedSVD`, `BiasSVD` and `SVDPlusPlus`. Unfortunately, I don't see a proper way to merge them. However we could reduce some code as follows. I think we could replace the whole body of this function by
```
{
  return Evaluate(parameters, 0, data.n_cols);
}
```
What do you think?

src/mlpack/methods/bias_svd/bias_svd_function_impl.hpp
```diff
@@ -0,0 +1,411 @@
...
+
+    objective += (ratingErrorSquared + regularizationError);
+  }
+
+  return objective;
+}
+
+template <typename MatType>
+void BiasSVDFunction<MatType>::Gradient(const arma::mat& parameters,
+                                               arma::mat& gradient) const
```
lozhnikov: The indent is incorrect.

src/mlpack/methods/bias_svd/bias_svd_function_impl.hpp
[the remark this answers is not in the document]
lozhnikov: Ah, I see. I didn't think of that. Thanks for the clarification.

src/mlpack/methods/bias_svd/bias_svd_function_impl.hpp
```diff
@@ -0,0 +1,411 @@
...
+    const double rating = data(2, i);
+    const double userBias = parameters(rank, user);
+    const double itemBias = parameters(rank, item);
+    double ratingError = rating - userBias - itemBias -
+        arma::dot(parameters.col(user).subvec(0, rank - 1),
+                  parameters.col(item).subvec(0, rank - 1));
+
+    // Gradient is non-zero only for the parameter columns corresponding to the
+    // example.
+    gradient.col(user).subvec(0, rank - 1) +=
```
lozhnikov: Looks like the only reason of the function is vectorized sums. Do we really need this? I suggest to replace the body of the function by
```
{
  Gradient(parameters, 0, gradient, data.n_cols);
}
```

src/mlpack/methods/bias_svd/bias_svd_function_impl.hpp
```diff
@@ -0,0 +1,411 @@
...
+        2 * (lambda * parameters(rank, item) - ratingError);
+  }
+}
+
+template <typename MatType>
+template <typename GradType>
+void BiasSVDFunction<MatType>::Gradient(const arma::mat& parameters,
+                                               const size_t start,
+                                               GradType& gradient,
+                                               const size_t batchSize) const
```
lozhnikov: Could you fix the spaces here?

src/mlpack/methods/bias_svd/bias_svd_function_impl.hpp
```diff
@@ -0,0 +1,411 @@
...
+    const size_t user = data(0, i);
+    const size_t item = data(1, i) + numUsers;
+
+    // Calculate the squared error in the prediction.
+    const double rating = data(2, i);
+    const double userBias = parameters(rank, user);
+    const double itemBias = parameters(rank, item);
+    double ratingError = rating - userBias - itemBias -
+        arma::dot(parameters.col(user).subvec(0, rank - 1),
+                  parameters.col(item).subvec(0, rank - 1));
```
lozhnikov: I forgot to say that you could avoid doing every time `.subvec(0, rank - 1)`. In order to do that you can introduce another matrix `arma::mat userItemMatrix(parameter.memptr(), rank, numUsers + numItems, false, false)`. I just wanted to make sure that you know.

src/mlpack/methods/svdplusplus/svdplusplus_function_impl.hpp
lozhnikov: Perhaps you are right. I don't know exactly how `atomic` is implemented internally, but it is possible `atomic` doesn't deal with mutexes.

Wenhao-H:
@lozhnikov Are you sure that it is `BiasSVDTest` that takes up 81 secs? Because `BiasSVDTest` only takes ~7 secs when I run it on my computer.

src/mlpack/methods/bias_svd/bias_svd_function_impl.hpp
Wenhao-H: Yes I also noticed the problem but I don't have any thought on how to properly merge the duplicated code either. For the `Evaluate(parameters)` function you pointed out, yes I think we can do that.

lozhnikov:
@Wenhao-H Yeah, I argee. Probably it happened due to I ran `ctest -j4`. `CFTest` took 19 sec on my system. It is not too much in comparison with `RangeSearchTest` (26 secs).

src/mlpack/methods/bias_svd/bias_svd_function_impl.hpp
Wenhao-H: I am not totally sure but I think vectorized sum can be faster on certain hardwares that supports SIMD.

src/mlpack/methods/bias_svd/bias_svd_function_impl.hpp
Wenhao-H: I'm afraid this doesn't work, because in armadillo matrix is stored in column-major ordering.

lozhnikov: Indeed, `armadillo` stores matrices in a column-major order, I didn't think of that. However you could reshape the `parameters` matrix to `1 x ((rank + 1) * (numUsers + numItems))`. However, not sure that it is worth to do.

Wenhao-H:
@lozhnikov Thanks! The test that fails is `RecommendationAccuracySVDPPTest`. I am not able to identify the reason why `SVD++` can pass the `CFPredictSVDPPTest` but failed the one above. For now I think we can remove `RecommendationAccuracySVDPPTest` and leave a comment there.

src/mlpack/methods/bias_svd/bias_svd_function_impl.hpp
Wenhao-H: @lozhnikov I don't understand how reshaping `parameters` can solve this. Can you provide a code example?

lozhnikov:
@Wenhao-H Let me look in detail into `RecommendationAccuracySVDPPTest`. Maybe I could suggest something reasonable.

src/mlpack/methods/svdplusplus/svdplusplus_function_impl.hpp
```diff
@@ -0,0 +1,611 @@
...
+    gradient(rank, item) +=
+        2 * (lambda * parameters(rank, item) - ratingError);
+    // Calculate gradients for item implicit vector.
+    it = implicitData.begin_col(user);
+    it_end = implicitData.end_col(user);
+    for (; it != it_end; it++)
+    {
+      gradient.col(implicitStart + it.row()).subvec(0, rank - 1) +=
+          2 * (lambda *
+          parameters.col(implicitStart + it.row()).subvec(0, rank - 1) -
```
lozhnikov: I think `parameters.col(implicitStart + it.row()).subvec(0, rank - 1)` should be normalized i.e. you should divide it by `implicitCount`.

src/mlpack/methods/svdplusplus/svdplusplus_function_impl.hpp
```diff
@@ -0,0 +1,611 @@
...
+          ratingError / std::sqrt(implicitCount) *
```
lozhnikov: And I think it is worth to mention that `implicitCount != 0` here.

src/mlpack/methods/svdplusplus/svdplusplus_function_impl.hpp
```diff
@@ -0,0 +1,611 @@
...
+    // Calculate gradients for item implicit vector.
+    it = implicitData.begin_col(user);
+    it_end = implicitData.end_col(user);
+    for (; it != it_end; it++)
+    {
+      for (size_t j = 0; j < rank; ++j)
+      {
+        gradient(j, implicitStart + it.row()) +=
+            2 * (lambda *
+            parameters(j, implicitStart + it.row()) -
```
lozhnikov: Same comment: `parameters(j, implicitStart + it.row())` should be normalized.

src/mlpack/methods/svdplusplus/svdplusplus_function_impl.hpp
```diff
@@ -0,0 +1,611 @@
...
+    parameters(rank, item) -= stepSize * 2 * (
+        lambda * parameters(rank, item) - ratingError);
+    // Update item implicit vectors.
+    it = implicitData.begin_col(user);
+    it_end = implicitData.end_col(user);
+    for (; it != it_end; it++)
+    {
+      parameters.col(implicitStart + it.row()).subvec(0, rank - 1) -=
+          stepSize * 2 * (lambda *
+          parameters.col(implicitStart + it.row()).subvec(0, rank - 1) -
```
lozhnikov: Again `parameters.col(implicitStart + it.row()).subvec(0, rank - 1) / implicitCount`.

src/mlpack/methods/svdplusplus/svdplusplus_function_impl.hpp
```diff
@@ -0,0 +1,611 @@
...
+    }
+  }
+}
+
+template <typename MatType>
+template <typename GradType>
+void SVDPlusPlusFunction<MatType>::Gradient(const arma::mat& parameters,
+                                               const size_t start,
+                                               GradType& gradient,
+                                               const size_t batchSize) const
```
lozhnikov: Could you fix spaces here?

src/mlpack/methods/svdplusplus/svdplusplus_function_impl.hpp
```diff
@@ -0,0 +1,611 @@
...
+
+    objective += (ratingErrorSquared + regularizationError);
+  }
+
+  return objective;
+}
+
+template <typename MatType>
+void SVDPlusPlusFunction<MatType>::Gradient(const arma::mat& parameters,
+                                               arma::mat& gradient) const
```
lozhnikov: Could you fix the indent here?

src/mlpack/methods/svdplusplus/svdplusplus_function_impl.hpp
```diff
@@ -0,0 +1,611 @@
...
+    cost += (ratingErrorSquared + regularizationError);
+  }
+
+  return cost;
+}
+
+template <typename MatType>
+double SVDPlusPlusFunction<MatType>::Evaluate(const arma::mat& parameters,
+                                                 const size_t start,
+                                                 const size_t batchSize) const
```
lozhnikov: Incorrect indent.

src/mlpack/methods/bias_svd/bias_svd_function_impl.hpp
Wenhao-H: Ah I understand now. Thanks for the example. In my view both implementations are good, but the implementation you suggested is more clear. It could take some time to modify. I think for now we can stay with the current implementation, and write down some code comments as a reminder so we can improve this in another PR later. What do you think?

src/mlpack/methods/svdplusplus/svdplusplus_function_impl.hpp
```diff
@@ -0,0 +1,611 @@
...
+        arma::mat itemImplicitUpdate(rank, implicitCount);
+        arma::Col<size_t> implicitItems(implicitCount);
+        it = implicitData.begin_col(user);
+        it_end = implicitData.end_col(user);
+        size_t implicitIndex = 0;
+        for (; it != it_end; it++, implicitIndex++)
+        {
+          itemImplicitUpdate.col(implicitIndex) =
+              stepSize * 2 * (lambda *
+              iterate.col(implicitStart + it.row()).subvec(0, rank - 1) -
```
lozhnikov: `iterate.col(implicitStart + it.row()).subvec(0, rank - 1) / implicitCount`

src/mlpack/methods/bias_svd/bias_svd_function_impl.hpp
lozhnikov: Yeah, I think even the comment is unnecessary. Both implementations are good.

src/mlpack/methods/svdplusplus/svdplusplus_impl.hpp
```diff
@@ -0,0 +1,117 @@
...
+void SVDPlusPlus<OptimizerType>::Apply(const arma::mat& data,
+                                       const arma::mat& implicitData,
+                                       const size_t rank,
+                                       arma::mat& u,
+                                       arma::mat& v,
+                                       arma::vec& p,
+                                       arma::vec& q,
+                                       arma::mat& y)
+{
+  // batchSize is 1 in our implementation of Bias SVD.
```
lozhnikov: Could you update the comment? I mean `Bias SVD`.

src/mlpack/methods/cf/cf.hpp
```diff
@@ -99,9 +102,9 @@ class CFType
    * @param minResidue Residue required to terminate.
    * @param mit Whether to terminate only when maxIterations is reached.
    */
-  template<typename MatType, typename DecompositionPolicy = NMFPolicy>
+  template<typename MatType>
   CFType(const MatType& data,
-         DecompositionPolicy& decomposition = DecompositionPolicy(),
+         const DecompositionPolicy& decomposition = DecompositionPolicy(),
```
lozhnikov: Could you remove `@tparam DecompositionPolicy` from the comment above?

src/mlpack/methods/cf/cf.hpp
```diff
@@ -122,9 +125,8 @@ class CFType
    * @param minResidue Residue required to terminate.
    * @param mit Whether to terminate only when maxIterations is reached.
    */
-  template<typename DecompositionPolicy>
   void Train(const arma::mat& data,
-             DecompositionPolicy& decomposition,
+             const DecompositionPolicy& decomposition,
```
lozhnikov: same comment.

src/mlpack/methods/cf/cf.hpp
```diff
@@ -143,9 +145,8 @@ class CFType
    * @param minResidue Residue required to terminate.
    * @param mit Whether to terminate only when maxIterations is reached.
    */
-  template<typename DecompositionPolicy>
   void Train(const arma::sp_mat& data,
-             DecompositionPolicy& decomposition,
+             const DecompositionPolicy& decomposition,
```
lozhnikov: Same comment.

src/mlpack/methods/cf/cf_model.hpp
```diff
@@ -0,0 +1,153 @@
...
+#include <mlpack/methods/cf/decomposition_policies/regularized_svd_method.hpp>
+#include <mlpack/methods/cf/decomposition_policies/svd_complete_method.hpp>
+#include <mlpack/methods/cf/decomposition_policies/svd_incomplete_method.hpp>
+#include <mlpack/methods/cf/decomposition_policies/bias_svd_method.hpp>
+#include <mlpack/methods/cf/decomposition_policies/svdplusplus_method.hpp>
+
+namespace mlpack {
+namespace cf {
+
+class DeleteVisitor : public boost::static_visitor<void>
```
lozhnikov: Could you add a short description before each visitor?

src/mlpack/methods/cf/normalization/item_mean_normalization.hpp
```diff
@@ -89,11 +89,26 @@ class ItemMeanNormalization
...
+    for (size_t i = 0; i < itemMean.n_elem; i++)
+    {
+      if (ratingNum(i) != 0)
+        itemMean(i) /= ratingNum(i);
+    }
+
+    // Normalize the data.
+    it = cleanedData.begin();
+    it_end = cleanedData.end();
+    for (; it != it_end; it++)
```
lozhnikov: Looks like there is a bug here. Look at the valgrind output
```
==19199== Invalid read of size 8
==19199==    at 0x3DEA7C1: row (SpMat_bones.hpp:385)
==19199==    by 0x3DEA7C1: arma::SpMat<double>::const_iterator::operator!=(arma::SpMat<double>::const_iterator const&) const (SpMat_iterators_meat.hpp:239)
==19199==    by 0x4135BE7: mlpack::cf::ItemMeanNormalization::Normalize(arma::SpMat<double>&) (item_mean_normalization.hpp:111)
==19199==    by 0x41AA940: mlpack::cf::CFType<mlpack::cf::NMFPolicy, mlpack::cf::ItemMeanNormalization>::Train(arma::SpMat<double> const&, mlpack::cf::NMFPolicy const&, unsigned long, double, bool) (cf_impl.hpp:132)
==19199==    by 0x4194105: mlpack::cf::CFType<mlpack::cf::NMFPolicy, mlpack::cf::ItemMeanNormalization>::CFType<arma::SpMat<double> >(arma::SpMat<double> const&, mlpack::cf::NMFPolicy const&, unsigned long, unsigned long, unsigned long, double, bool) (cf_impl.hpp:72)
==19199==    by 0x41654F2: void CFTest::Serialization<mlpack::cf::NMFPolicy, mlpack::cf::ItemMeanNormalization>() (cf_test.cpp:441)
```
Looks like `it_end` becomes out of date as soon as you change the matrix. The following code should fix the issue
```
for (; it != cleanedData.end(); it++)
```

src/mlpack/methods/cf/normalization/user_mean_normalization.hpp
```diff
@@ -89,10 +89,26 @@ class UserMeanNormalization
...
+    }
+    for (size_t i = 0; i < userMean.n_elem; i++)
+    {
+      if (ratingNum(i) != 0)
+        userMean(i) /= ratingNum(i);
+    }
+
+    // Normalize the data.
+    it = cleanedData.begin();
+    it_end = cleanedData.end();
```
lozhnikov: Same issue here.

src/mlpack/methods/cf/normalization/item_mean_normalization.hpp
Wenhao-H: Ahh thanks a lot!! I spent some time but was not able to debug this. BTW, where can I read the valgrind output?

lozhnikov: I ran valgrind on my local computer:) It isn't integrated with CI.

src/mlpack/methods/svdplusplus/svdplusplus_function_impl.hpp
```diff
@@ -0,0 +1,561 @@
...
+    for (; it != it_end; it++)
+    {
+      userVec += parameters.col(implicitStart + it.row()).subvec(0, rank - 1);
+      if (implicitVecsNormSquare(itemRealIdx) < 0)
+      { 
+        implicitVecsNormSquare(itemRealIdx) = arma::dot(
+            parameters.col(implicitStart + it.row()).subvec(0, rank - 1),
+            parameters.col(implicitStart + it.row()).subvec(0, rank - 1));
+      }
+      regularizationError += lambda * implicitVecsNormSquare(itemRealIdx);
```
lozhnikov: I think, the code above is incorrect (you should use `it.row()` instead of `itemRealIdx`). The fix follows:
```
if (implicitVecsNormSquare(it.row()) < 0)
{
  implicitVecsNormSquare(it.row()) = arma::dot(
      parameters.col(implicitStart + it.row()).subvec(0, rank - 1),
      parameters.col(implicitStart + it.row()).subvec(0, rank - 1));
}
regularizationError += lambda * implicitVecsNormSquare(it.row());
```

src/mlpack/methods/svdplusplus/svdplusplus_function_impl.hpp
```diff
@@ -0,0 +1,561 @@
...
+    it = implicitData.begin_col(user);
+    it_end = implicitData.end_col(user);
+    for (; it != it_end; it++)
+    {
+      // Note that implicitCount != 0 if this loop is acutally executed.
+      gradient.col(implicitStart + it.row()).subvec(0, rank - 1) +=
+          2.0 / implicitCount * (lambda *
+          parameters.col(implicitStart + it.row()).subvec(0, rank - 1) -
+          ratingError / std::sqrt(implicitCount) *
+          parameters.col(item).subvec(0, rank - 1));
```
lozhnikov: Looks like you put `/ implicitCount` into the wrong place. I meant this:
```
      gradient.col(implicitStart + it.row()).subvec(0, rank - 1) +=
          2.0 * (lambda / implicitCount *
          parameters.col(implicitStart + it.row()).subvec(0, rank - 1) -
          ratingError / std::sqrt(implicitCount) *
          parameters.col(item).subvec(0, rank - 1));
```

src/mlpack/methods/svdplusplus/svdplusplus_function_impl.hpp
```diff
@@ -0,0 +1,561 @@
...
+    // Calculate gradients for item implicit vector.
+    it = implicitData.begin_col(user);
+    it_end = implicitData.end_col(user);
+    for (; it != it_end; it++)
+    {
+      // Note that implicitCount != 0 if this loop is acutally executed.
+      for (size_t j = 0; j < rank; ++j)
+      {
+        gradient(j, implicitStart + it.row()) +=
+            2.0 / implicitCount * (lambda *
```
lozhnikov: Same issue here.

src/mlpack/methods/svdplusplus/svdplusplus_function_impl.hpp
```diff
@@ -0,0 +1,561 @@
...
+    parameters(rank, item) -= stepSize * 2 * (
+        lambda * parameters(rank, item) - ratingError);
+    // Update item implicit vectors.
+    it = implicitData.begin_col(user);
+    it_end = implicitData.end_col(user);
+    for (; it != it_end; it++)
+    {
+      // Note that implicitCount != 0 if this loop is acutally executed.
+      parameters.col(implicitStart + it.row()).subvec(0, rank - 1) -=
+          stepSize * 2.0 / implicitCount * (lambda *
```
lozhnikov: Here

src/mlpack/methods/svdplusplus/svdplusplus_function_impl.hpp
```diff
@@ -0,0 +1,561 @@
...
+        // Update of item implicit vectors.
+        arma::mat itemImplicitUpdate(rank, implicitCount);
+        arma::Col<size_t> implicitItems(implicitCount);
+        it = implicitData.begin_col(user);
+        it_end = implicitData.end_col(user);
+        size_t implicitIndex = 0;
+        for (; it != it_end; it++, implicitIndex++)
+        {
+          itemImplicitUpdate.col(implicitIndex) =
+              stepSize * 2.0 / implicitCount * (lambda *
```
lozhnikov: And here.

src/mlpack/methods/svdplusplus/svdplusplus_function_impl.hpp
Wenhao-H: Yes you are right. I will fix it. Thanks!

Wenhao-H:
I am not able to reduce the number of failures in `RecommendationAccuracySVDPPTest` to below `17` either. I've tried different combinations of normalization methods, neighbor search methods and interpolation methods and tuned the hyperparameters. The lowest number of failures was around `25`. @lozhnikov Do you think we can remove this test? Because we have `CFPredictSVDPPTest` to test the rsme performance of svd++.

lozhnikov:
@Wenhao-H I looked cautiously through the `SVD++` implementation and I think it's correct. So, I suggest to comment out the test and open an issue (and point out the issue in the comment before the test) since it could be interesting to investigate why the test fails.
Actually, the PR looks good to me. I'd like to look through it again a couple of times to be sure. I hope I'll do that by weekend.

There are two problems:
1. The decomposition policies contain a lot of duplicated code. I suggest to split `DecompositionPolicy` into 2 classes: `DecompositionPolicy` and `NeighborhoodPolicy` in such a way `DecompositionPolicy` decomposes a matrix and `NeighborhoodPolicy` finds the neighbors. However I'd like to move it to another PR since I don't want to overcomplicate this PR. On the other side these changes vary API a bit so they should be done before the release. @rcurtin @zoq What do you think?
2. Currently `RegularizedSVD`, `SVD++` and `BiasSVD` are implemented in similar way. However I don't have any clear understanding how to implement it in a general way without code duplication.
Perhaps it is a good idea to open an issue.

Wenhao-H:
@lozhnikov I have opened an issue #1501 for the failing `RecommendationAccuracySVDPPTest`.
For the code duplication in `RegularizedSVD`, `SVD++` and `BiasSVD`, I will open an issue after this PR is merged if we cannot think of a way to solve this.

Wenhao-H:
@lozhnikov Thanks a lot! I will make the changes you proposed probably in the next two days. This is a large PR so let's wait for 5 days before merge.
There is one more thing. The travis build failure is caused by LMNN test. Can we ignore this failure and go ahead to merge this PR?

lozhnikov:
@Wenhao-H The test isn't connected with the PR. Perhaps the failure is fixed in the master branch.

Wenhao-H:
Hmm I'm not sure I understand the AppVeyor build failure. Can someone give me a hand?

src/mlpack/methods/cf/cf_model_impl.hpp
```diff
@@ -44,8 +45,9 @@ template<typename DecompositionPolicy>
 void PredictVisitor::operator()(CFType<DecompositionPolicy>* c) const
 {
   if (c)
-    return c->Predict(combinations, predictions);
-  throw std::runtime_error("no cf model initialized");
+    throw std::runtime_error("no cf model initialized");
```
lozhnikov: `if (!c)` of course

lozhnikov:
@Wenhao-H Yeah, you can apply the patch using the following command:
```
patch -p1 < path/to/patch
```
(the working dir should be the root of the repo)

Issue #1449, opened 2018-06-27
LMNN: optimization by bounding slack term


rcurtin:
This is an optimization for #1407.  This one is focused on the following term, which appears during the computation of the LMNNFunction objective function and gradient:

```
          eval = metric.Evaluate(transformedDataset.col(i),
                    transformedDataset.col(targetNeighbors(j, i))) -
                 metric.Evaluate(transformedDataset.col(i),
                    transformedDataset.col(impostors(l, i)));

        // Check bounding condition.
        if (eval < -1)
        {
          // update bound.
          bp = l;
          break;
        }
```

We're already avoiding the computation of extra slack terms with the `if (eval < -1)`, which is great, but we can do better!  Using a (not-yet-derived) bound depending on `|| L_{t + 1} - L_t ||_F^2`, we can bound what `eval` will be at this iteration using last iteration's value.  If it's not possible that `eval` is greater than `-1`, then we can avoid the `metric.Evaluate()` call, which is great.

The first step here would be deriving the bound, then we would need to probably find a way to cache `eval`s from previous iterations and implement the bound.  I think that we can get a large speedup here, since for very many points `eval` will be far less than `-1`.

(Actually, it might be interesting to run a quick experiment and see, typically, how many points have `eval` greater than -1, how many have `eval` greater than -2, etc., to get an idea of how much speedup this will get us.)

manish7294:
I ran some of the tests as you mentioned, and I think it is apparent that there's a lot of speedups.
Here are some of the figures:
Iris:
```
(eval  > -1) --- (-2 < eval < -1) --- (eval < -2)
1298 ------ 862------1590
944 ------ 656------2150
667 ------ 426------2657
921 ------ 647------2182
655 ------ 408------2687
898 ------ 636------2216
646 ------ 391------2713
866 ------ 633------2251
640 ------ 375------2735
```

vc2:
```
(eval  > -1) --- (-2 < eval < -1) --- (eval < -2)
1477 ------ 210------3488
1430 ------ 226------3519
1415 ------ 259------3501
1427 ------ 257------3491
1442 ------ 246------3487
1445 ------ 257------3473
1449 ------ 255------3471
1457 ------ 248------3470
```

letters:
```
(eval  > -1) --- (-2 < eval < -1) --- (eval < -2)
30240 ------ 88219------377372
15095 ------ 41007------443898
20475 ------ 60067------419458
24889 ------ 74262------400849
```

[...]

PR #1461, opened 2018-07-04
LMNN: Bounds for slack term

manish7294:
See #1449

### DIFF
--- a/src/mlpack/methods/lmnn/lmnn_function.hpp
+++ b/src/mlpack/methods/lmnn/lmnn_function.hpp
@@ -238,6 +238,15 @@ class LMNNFunction
   * uses batches.
   */
   inline void Precalculate();
+  //! Update cache transformation matrices.
+  inline void UpdateCache(const arma::mat& transformation,
+                          const size_t begin,
+                          const size_t batchSize);
+  //! Calculate norm of change in transformation.
+  inline void TransDiff(std::map<size_t, double>& transformationDiffs,
+                        const arma::mat& transformation,
+                        const size_t begin,
+                        const size_t batchSize);
 };
 
 } // namespace lmnn
--- a/src/mlpack/methods/lmnn/lmnn_function.hpp
+++ b/src/mlpack/methods/lmnn/lmnn_function.hpp
@@ -221,12 +221,14 @@ class LMNNFunction
   //! Holds the norm of each data point.
   arma::vec norm;
   //! Hold previous eval values for each datapoint.
-  std::vector<arma::mat> evalOld;
+  arma::cube evalOld;
+  //! Hold previous maximum norm of impostor.
+  arma::mat maxImpNorm;
   //! Holds previous transformation matrix. Used for L-BFGS like optimizer.
   arma::mat transformationOld;
   //! Holds previous transformation matrix for each point. Used for
   //! optimizers which operates over batches.
-  std::vector<arma::mat> transformationOldPoint;
+  arma::cube transformationOldPoint;
   /**
   * Precalculate the gradient part due to target neighbors and stores
   * the result as a matrix. Used for L-BFGS like optimizers which does not
--- a/src/mlpack/methods/lmnn/lmnn_function.hpp
+++ b/src/mlpack/methods/lmnn/lmnn_function.hpp
@@ -221,9 +221,12 @@ class LMNNFunction
   //! Holds the norm of each data point.
   arma::vec norm;
   //! Hold previous eval values for each datapoint.
-  arma::vec evalOld;
-  //! Holds previous transformation matrix.
+  std::vector<arma::mat> evalOld;
+  //! Holds previous transformation matrix. Used for L-BFGS like optimizer.
   arma::mat transformationOld;
+  //! Holds previous transformation matrix for each point. Used for
+  //! optimizers which operates over batches.
+  std::vector<arma::mat> transformationOldPoint;
   /**
   * Precalculate the gradient part due to target neighbors and stores
   * the result as a matrix. Used for L-BFGS like optimizers which does not
--- a/src/mlpack/methods/lmnn/lmnn_function.hpp
+++ b/src/mlpack/methods/lmnn/lmnn_function.hpp
@@ -63,6 +63,7 @@ class LMNNFunction
                size_t range,
                MetricType metric = MetricType());
 
+
   /**
    * Shuffle the points in the dataset. This may be used by optimizers.
    */
@@ -217,6 +218,12 @@ class LMNNFunction
   Constraints<MetricType> constraint;
   //! Holds pre-calculated cij.
   arma::mat pCij;
+  //! Holds the norm of each data point.
+  arma::vec norm;
+  //! Hold previous eval value.
+  double evalOld;
+  //! Holds previous transformation matrix.
+  arma::mat transformationOld;
   /**
   * Precalculate the gradient part due to target neighbors and stores
   * the result as a matrix. Used for L-BFGS like optimizers which does not
--- a/src/mlpack/methods/lmnn/lmnn_function.hpp
+++ b/src/mlpack/methods/lmnn/lmnn_function.hpp
@@ -226,9 +226,12 @@ class LMNNFunction
   arma::mat maxImpNorm;
   //! Holds previous transformation matrix. Used for L-BFGS like optimizer.
   arma::mat transformationOld;
-  //! Holds previous transformation matrix for each point. Used for
-  //! optimizers which operates over batches.
-  arma::cube transformationOldPoint;
+  //! Holds previous transformation matrices.
+  std::vector<arma::mat> oldTransformationMatrices;
+  //! Holds number of points which are using each transformation matrix.
+  std::vector<size_t> oldTransformationCounts;
+  //! Holds points to transformation matrix mapping.
+  arma::vec lastTransformationIndices;
   /**
   * Precalculate the gradient part due to target neighbors and stores
   * the result as a matrix. Used for L-BFGS like optimizers which does not
--- a/src/mlpack/methods/lmnn/lmnn_function.hpp
+++ b/src/mlpack/methods/lmnn/lmnn_function.hpp
@@ -220,8 +220,8 @@ class LMNNFunction
   arma::mat pCij;
   //! Holds the norm of each data point.
   arma::vec norm;
-  //! Hold previous eval value.
-  double evalOld;
+  //! Hold previous eval values for each datapoint.
+  arma::vec evalOld;
   //! Holds previous transformation matrix.
   arma::mat transformationOld;
   /**
--- a/src/mlpack/methods/lmnn/lmnn_function_impl.hpp
+++ b/src/mlpack/methods/lmnn/lmnn_function_impl.hpp
@@ -152,15 +152,7 @@ double LMNNFunction<MetricType>::Evaluate(const arma::mat& transformation)
 
           // Check if there is need to calculate exact eval value.
           if (eval <= -1)
-          {
             exactEval = false;
-          }
-          else
-          {
-            // Reset cacche max impostor norm.
-            maxImpNorm(l, i) = 0;
-            evalOld(l, j, i) = arma::datum::nan;
-          }
         }
 
         // Calculate exact eval value.
@@ -193,6 +185,14 @@ double LMNNFunction<MetricType>::Evaluate(const arma::mat& transformation)
         }
 
         cost += regularization * (1 + eval);
+
+        // Reset cache.
+        if (eval > -1)
+        {
+          // update bound.
+          evalOld(l, j, i) = arma::datum::nan;
+          maxImpNorm(l, i) = 0;
+        }
       }
     }
   }
@@ -303,7 +303,7 @@ double LMNNFunction<MetricType>::Evaluate(const arma::mat& transformation,
         {
           // update bound.
           evalOld(l, j, i) = arma::datum::nan;
-          maxImpNorm(l, i)=0;
+          maxImpNorm(l, i) = 0;
           lastTransformationIndices(i) = arma::datum::nan;
           --oldTransformationCounts[lastTransformationIndices(i)];
         }
@@ -352,6 +352,22 @@ template<typename GradType>
 void LMNNFunction<MetricType>::Gradient(const arma::mat& transformation,
                                         GradType& gradient)
 {
+  // Apply metric over dataset.
+  transformedDataset = transformation * dataset;
+
+  // Calculate norm of change in transformation.
+  double transformationDiff = 0;
+  if (transformationOld.n_elem != 0)
+  {
+    transformationDiff = arma::norm(transformation - transformationOld);
+  }
+
+  if (iteration++ % range == 0)
+  {
+    // Re-calculate impostors on transformed dataset.
+    constraint.Impostors(impostors, distance, transformedDataset, labels);
+  }
+
   gradient.zeros(transformation.n_rows, transformation.n_cols);
 
   // Calculate gradient due to target neighbors.
@@ -367,18 +383,63 @@ void LMNNFunction<MetricType>::Gradient(const arma::mat& transformation,
       // Bound constraints to avoid uneccesary computation.
       for (size_t l = 0, bp = k; l < bp ; l++)
       {
-        // Calculate gradient due to triplets.
-        double eval = metric.Evaluate(transformedDataset.col(i),
-                        transformedDataset.col(targetNeighbors(j, i))) -
-                    metric.Evaluate(transformedDataset.col(i),
-                        transformedDataset.col(impostors(l, i)));
+        // Calculate cost due to {data point, target neighbors, impostors}
+        // triplets.
+        double eval = 0;
+
+        // Flag to trigger exact eval calculation.
+        bool exactEval = true;
+
+        // Bounds for eval.
+        if (transformationOld.n_elem != 0 && !std::isnan(evalOld(l, j, i)))
+        {
+          // Update cache max impostor norm.
+          maxImpNorm(l, i) = std::max(maxImpNorm(l, i), norm(impostors(l, i)));
+
+          eval = evalOld(l, j, i) + transformationDiff *
+            (norm(targetNeighbors(j, i)) + maxImpNorm(l, i) +
+            2 * norm(i));
+
+          // Check if there is need to calculate exact eval value.
+          if (eval <= -1)
+            exactEval = false;
+        }
+
+        // Calculate exact eval value.
+        if(exactEval)
+        {
+          if (iteration - 1 % range == 0)
+          {
+            eval = metric.Evaluate(transformedDataset.col(i),
+                     transformedDataset.col(targetNeighbors(j, i))) -
+                 distance(l, i);
+          }
+          else
+          {
+            eval = metric.Evaluate(transformedDataset.col(i),
+                     transformedDataset.col(targetNeighbors(j, i))) -
+                   metric.Evaluate(transformedDataset.col(i),
+                       transformedDataset.col(impostors(l, i)));
+          }
+        }
+
+        // Update cache eval value.
+        evalOld(l, j, i) = eval;
 
         // Check bounding condition.
-        if (eval < -1)
+        if (eval <= -1)
         {
           // update bound.
           bp = l;
           break;
+        }
+
+        // Reset cache.
+        if (eval > -1)
+        {
+          // update bound.
+          evalOld(l, j, i) = arma::datum::nan;
+          maxImpNorm(l, i) = 0;
         }
 
         // Caculate gradient due to impostors.
@@ -393,6 +454,9 @@ void LMNNFunction<MetricType>::Gradient(const arma::mat& transformation,
 
   gradient = 2 * transformation * ((1 - regularization) * cij +
       regularization * cil);
+
+  // Update cache transformation matrix.
+  transformationOld = transformation;
 }
 
 //! Compute gradient over a batch of data points.
@@ -403,6 +467,16 @@ void LMNNFunction<MetricType>::Gradient(const arma::mat& transformation,
                                         GradType& gradient,
                                         const size_t batchSize)
 {
+  // Apply metric over dataset.
+  transformedDataset = transformation * dataset;
+
+  if (iteration++ % range == 0)
+  {
+    // Re-calculate impostors on transformed dataset.
+    constraint.Impostors(impostors, distance, transformedDataset, labels,
+        begin, batchSize);
+  }
+
   gradient.zeros(transformation.n_rows, transformation.n_cols);
 
   arma::mat cij = arma::zeros(dataset.n_rows, dataset.n_rows);
@@ -417,23 +491,77 @@ void LMNNFunction<MetricType>::Gradient(const arma::mat& transformation,
       cij += diff * arma::trans(diff);
     }
 
+    // Calculate norm of change in transformation.
+    double transformationDiff = 0;
+    if (arma::is_finite(lastTransformationIndices(i)))
+    {
+      transformationDiff = arma::norm(transformation -
+          oldTransformationMatrices[lastTransformationIndices(i)]);
+    }
+
     for (int j = k - 1; j >= 0; j--)
     {
       // Bound constraints to avoid uneccesary computation.
       for (size_t l = 0, bp = k; l < bp ; l++)
       {
-        // Calculate gradient due to triplets.
-        double eval = metric.Evaluate(transformedDataset.col(i),
-                        transformedDataset.col(targetNeighbors(j, i))) -
-                    metric.Evaluate(transformedDataset.col(i),
-                        transformedDataset.col(impostors(l, i)));
+        // Calculate cost due to {data point, target neighbors, impostors}
+        // triplets.
+        double eval = 0;
+
+        // Flag to trigger exact eval calculation.
+        bool exactEval = true;
+
+        // Bounds for eval.
+        if (arma::is_finite(lastTransformationIndices(i)) && !std::isnan(evalOld(l, j, i)))
+        {
+          // Update cache max impostor norm.
+          maxImpNorm(l, i) = std::max(maxImpNorm(l, i), norm(impostors(l, i)));
+
+          eval = evalOld(l, j, i) + transformationDiff *
+            (norm(targetNeighbors(j, i)) + maxImpNorm(l, i) +
+            2 * norm(i));
+
+          if (eval <= -1)
+            exactEval = false;
+        }
+
+        // Calculate exact eval value.
+        if(exactEval)
+        {
+          if (iteration - 1 % range == 0)
+          {
+            eval = metric.Evaluate(transformedDataset.col(i),
+                     transformedDataset.col(targetNeighbors(j, i))) -
+                 distance(l, i);
+          }
+          else
+          {
+            eval = metric.Evaluate(transformedDataset.col(i),
+                     transformedDataset.col(targetNeighbors(j, i))) -
+                   metric.Evaluate(transformedDataset.col(i),
+                       transformedDataset.col(impostors(l, i)));
+          }
+        }
+
+        // Update cache eval value.
+        evalOld(l, j, i) = eval;
 
         // Check bounding condition.
-        if (eval < -1)
+        if (eval <= -1)
         {
           // update bound.
           bp = l;
           break;
+        }
+
+        // Reset cache.
+        if (eval > -1 && arma::is_finite(lastTransformationIndices(i)))
+        {
+          // update bound.
+          evalOld(l, j, i) = arma::datum::nan;
+          maxImpNorm(l, i) = 0;
+          lastTransformationIndices(i) = arma::datum::nan;
+          --oldTransformationCounts[lastTransformationIndices(i)];
         }
 
         // Caculate gradient due to impostors.
@@ -448,6 +576,38 @@ void LMNNFunction<MetricType>::Gradient(const arma::mat& transformation,
 
   gradient = 2 * transformation * ((1 - regularization) * cij +
       regularization * cil);
+
+  // Update cache transformation matrices.
+  // Are there any empty transformation matrices?
+  size_t index = oldTransformationMatrices.size();
+  for (size_t i = 0; i < oldTransformationCounts.size(); ++i)
+  {
+    if (oldTransformationCounts[i] == 0)
+    {
+      index = i; // Reuse this index.
+      break;
+    }
+  }
+
+  // Did we find an unused matrix?  If not, we have to allocate new space.
+  if (index == oldTransformationMatrices.size())
+  {
+    oldTransformationMatrices.push_back(transformation);
+    oldTransformationCounts.push_back(0);
+  }
+  else
+  {
+    oldTransformationMatrices[index] = transformation;
+  }
+
+  // Update all the transformation indices.
+  for (size_t i = begin; i < begin + batchSize; ++i)
+  {
+    --oldTransformationCounts[lastTransformationIndices(i)];
+    lastTransformationIndices(i) = index;
+  }
+
+  oldTransformationCounts[index] += batchSize;
 }
 
 //! Compute cost & gradient over whole dataset.
@@ -517,15 +677,7 @@ double LMNNFunction<MetricType>::EvaluateWithGradient(
 
           // Check if there is need to calculate exact eval value.
           if (eval <= -1)
-          {
             exactEval = false;
-          }
-          else
-          {
-            // Reset cacche max impostor norm.
-            maxImpNorm(l, i) = 0;
-            evalOld(l, j, i) = arma::datum::nan;
-          }
         }
 
         // Calculate exact eval value.
@@ -558,6 +710,14 @@ double LMNNFunction<MetricType>::EvaluateWithGradient(
         }
 
         cost += regularization * (1 + eval);
+
+        // Reset cache.
+        if (eval > -1)
+        {
+          // update bound.
+          evalOld(l, j, i) = arma::datum::nan;
+          maxImpNorm(l, i) = 0;
+        }
 
         // Caculate gradient due to impostors.
         arma::vec diff = dataset.col(i) - dataset.col(targetNeighbors(j, i));
@@ -689,7 +849,7 @@ double LMNNFunction<MetricType>::EvaluateWithGradient(
         {
           // update bound.
           evalOld(l, j, i) = arma::datum::nan;
-          maxImpNorm(l, i)=0;
+          maxImpNorm(l, i) = 0;
           lastTransformationIndices(i) = arma::datum::nan;
           --oldTransformationCounts[lastTransformationIndices(i)];
         }
--- a/src/mlpack/methods/lmnn/lmnn_function_impl.hpp
+++ b/src/mlpack/methods/lmnn/lmnn_function_impl.hpp
@@ -107,18 +107,34 @@ double LMNNFunction<MetricType>::Evaluate(const arma::mat& transformation)
         // Calculate cost due to {data point, target neighbors, impostors}
         // triplets.
         double eval = 0;
-        if ((iteration - 1 % range == 0) && (iteration - 1 != 0))
+        if(transformationOld.n_elem != 0)
         {
-          eval = metric.Evaluate(transformedDataset.col(i),
-                    transformedDataset.col(targetNeighbors(j, i))) -
-                 distance(l, i);
+          eval = evalOld + arma::norm(transformation - transformationOld) *
+              (norm(targetNeighbors(j, i)) + norm(impostors(l, i)) + 2 * norm(i));
+          if (eval > -1)
+          {
+            // Calculate exact eval.
+            if ((iteration - 1 % range == 0))
+            {
+              eval = metric.Evaluate(transformedDataset.col(i),
+                         transformedDataset.col(targetNeighbors(j, i))) -
+                     distance(l, i);
+            }
+            else
+            {
+              eval = metric.Evaluate(transformedDataset.col(i),
+                       transformedDataset.col(targetNeighbors(j, i))) -
+                     metric.Evaluate(transformedDataset.col(i),
+                       transformedDataset.col(impostors(l, i)));
+            }
+          }
         }
         else
         {
           eval = metric.Evaluate(transformedDataset.col(i),
-                    transformedDataset.col(targetNeighbors(j, i))) -
+                     transformedDataset.col(targetNeighbors(j, i))) -
                  metric.Evaluate(transformedDataset.col(i),
-                    transformedDataset.col(impostors(l, i)));
+                     transformedDataset.col(impostors(l, i)));
         }
 
         // Check bounding condition.
@@ -128,6 +144,10 @@ double LMNNFunction<MetricType>::Evaluate(const arma::mat& transformation)
           bp = l;
           break;
         }
+
+        // Update cache.
+        evalOld = eval;
+        transformationOld = transformation;
 
         cost += regularization * (1 + eval);
       }
@@ -173,18 +193,34 @@ double LMNNFunction<MetricType>::Evaluate(const arma::mat& transformation,
         // Calculate cost due to {data point, target neighbors, impostors}
         // triplets.
         double eval = 0;
-        if ((iteration - 1 % range == 0) && (iteration - 1 != 0))
+        if(transformationOld.n_elem != 0)
         {
-          eval = metric.Evaluate(transformedDataset.col(i),
-                    transformedDataset.col(targetNeighbors(j, i))) -
-                 distance(l, i);
+          eval = evalOld + arma::norm(transformation - transformationOld) *
+              (norm(targetNeighbors(j, i)) + norm(impostors(l, i)) + 2 * norm(i));
+          if (eval > -1)
+          {
+            // Calculate exact eval.
+            if ((iteration - 1 % range == 0))
+            {
+              eval = metric.Evaluate(transformedDataset.col(i),
+                        transformedDataset.col(targetNeighbors(j, i))) -
+                     distance(l, i);
+            }
+            else
+            {
+              eval = metric.Evaluate(transformedDataset.col(i),
+                       transformedDataset.col(targetNeighbors(j, i))) -
+                     metric.Evaluate(transformedDataset.col(i),
+                       transformedDataset.col(impostors(l, i)));
+            }
+          }
         }
         else
         {
           eval = metric.Evaluate(transformedDataset.col(i),
-                    transformedDataset.col(targetNeighbors(j, i))) -
+                     transformedDataset.col(targetNeighbors(j, i))) -
                  metric.Evaluate(transformedDataset.col(i),
-                    transformedDataset.col(impostors(l, i)));
+                     transformedDataset.col(impostors(l, i)));
         }
 
         // Check bounding condition.
@@ -194,6 +230,10 @@ double LMNNFunction<MetricType>::Evaluate(const arma::mat& transformation,
           bp = l;
           break;
         }
+
+        // Update cache.
+        evalOld = eval;
+        transformationOld = transformation;
 
         cost += regularization * (1 + eval);
       }
@@ -351,18 +391,34 @@ double LMNNFunction<MetricType>::EvaluateWithGradient(
         // Calculate cost due to {data point, target neighbors, impostors}
         // triplets.
         double eval = 0;
-        if ((iteration - 1 % range == 0) && (iteration - 1 != 0))
+        if(transformationOld.n_elem != 0)
         {
-          eval = metric.Evaluate(transformedDataset.col(i),
-                    transformedDataset.col(targetNeighbors(j, i))) -
-                 distance(l, i);
+          eval = evalOld + arma::norm(transformation - transformationOld) *
+              (norm(targetNeighbors(j, i)) + norm(impostors(l, i)) + 2 * norm(i));
+          if (eval > -1)
+          {
+            // Calculate exact eval.
+            if ((iteration - 1 % range == 0))
+            {
+              eval = metric.Evaluate(transformedDataset.col(i),
+                        transformedDataset.col(targetNeighbors(j, i))) -
+                     distance(l, i);
+            }
+            else
+            {
+              eval = metric.Evaluate(transformedDataset.col(i),
+                       transformedDataset.col(targetNeighbors(j, i))) -
+                     metric.Evaluate(transformedDataset.col(i),
+                       transformedDataset.col(impostors(l, i)));
+            }
+          }
         }
         else
         {
           eval = metric.Evaluate(transformedDataset.col(i),
-                    transformedDataset.col(targetNeighbors(j, i))) -
+                     transformedDataset.col(targetNeighbors(j, i))) -
                  metric.Evaluate(transformedDataset.col(i),
-                    transformedDataset.col(impostors(l, i)));
+                     transformedDataset.col(impostors(l, i)));
         }
 
         // Check bounding condition.
@@ -372,6 +428,10 @@ double LMNNFunction<MetricType>::EvaluateWithGradient(
           bp = l;
           break;
         }
+
+        // Update cache.
+        evalOld = eval;
+        transformationOld = transformation;
 
         cost += regularization * (1 + eval);
 
@@ -439,18 +499,34 @@ double LMNNFunction<MetricType>::EvaluateWithGradient(
         // Calculate cost due to {data point, target neighbors, impostors}
         // triplets.
         double eval = 0;
-        if ((iteration - 1 % range == 0) && (iteration - 1 != 0))
+        if(transformationOld.n_elem != 0)
         {
-          eval = metric.Evaluate(transformedDataset.col(i),
-                    transformedDataset.col(targetNeighbors(j, i))) -
-                 distance(l, i);
+          eval = evalOld + arma::norm(transformation - transformationOld) *
+              (norm(targetNeighbors(j, i)) + norm(impostors(l, i)) + 2 * norm(i));
+          if (eval > -1)
+          {
+            // Calculate exact eval.
+            if ((iteration - 1 % range == 0))
+            {
+              eval = metric.Evaluate(transformedDataset.col(i),
+                        transformedDataset.col(targetNeighbors(j, i))) -
+                     distance(l, i);
+            }
+            else
+            {
+              eval = metric.Evaluate(transformedDataset.col(i),
+                       transformedDataset.col(targetNeighbors(j, i))) -
+                     metric.Evaluate(transformedDataset.col(i),
+                       transformedDataset.col(impostors(l, i)));
+            }
+          }
         }
         else
         {
           eval = metric.Evaluate(transformedDataset.col(i),
-                    transformedDataset.col(targetNeighbors(j, i))) -
+                     transformedDataset.col(targetNeighbors(j, i))) -
                  metric.Evaluate(transformedDataset.col(i),
-                    transformedDataset.col(impostors(l, i)));
+                     transformedDataset.col(impostors(l, i)));
         }
 
         // Check bounding condition.
@@ -460,6 +536,10 @@ double LMNNFunction<MetricType>::EvaluateWithGradient(
           bp = l;
           break;
         }
+
+        // Update cache.
+        evalOld = eval;
+        transformationOld = transformation;
 
         cost += regularization * (1 + eval);
 
@@ -483,9 +563,11 @@ template<typename MetricType>
 inline void LMNNFunction<MetricType>::Precalculate()
 {
   pCij.zeros(dataset.n_rows, dataset.n_rows);
+  norm.zeros(dataset.n_cols);
 
   for (size_t i = 0; i < dataset.n_cols; i++)
   {
+    norm(i) = arma::norm(dataset.col(i));
     for (size_t j = 0; j < k ; j++)
     {
       // Calculate gradient due to target neighbors.
--- a/src/mlpack/methods/lmnn/lmnn_function_impl.hpp
+++ b/src/mlpack/methods/lmnn/lmnn_function_impl.hpp
@@ -41,11 +41,15 @@ LMNNFunction<MetricType>::LMNNFunction(const arma::mat& dataset,
   // Initialize transformed dataset to base dataset.
   transformedDataset = dataset;
 
+  // Initialize cache.
   evalOld.set_size(k, k, dataset.n_cols);
   evalOld.fill(arma::datum::nan);
 
   maxImpNorm.set_size(k, dataset.n_cols);
   maxImpNorm.fill(0);
+
+  lastTransformationIndices.set_size(dataset.n_cols);
+  lastTransformationIndices.fill(arma::datum::nan);
 
   // Initialize target neighbors & impostors.
   targetNeighbors = arma::Mat<size_t>(k, dataset.n_cols, arma::fill::zeros);
@@ -66,7 +70,7 @@ void LMNNFunction<MetricType>::Shuffle()
   arma::mat newDataset = dataset;
   arma::Mat<size_t> newLabels = labels;
   arma::cube newEvalOld = evalOld;
-  arma::cube newTransformationOldPoint = transformationOldPoint;
+  arma::vec newlastTransformationIndices = lastTransformationIndices;
   arma::mat newMaxImpNorm = maxImpNorm;
 
   // Generate ordering.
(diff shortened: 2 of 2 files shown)


manish7294:
@rcurtin  Implementing this bound has caused a bit increase in the runtime of the program, please have a look at it.

rcurtin:
What have you tried to debug the lack of speedup?

src/mlpack/methods/lmnn/lmnn_function_impl.hpp
```diff
@@ -107,28 +107,49 @@ double LMNNFunction<MetricType>::Evaluate(const arma::mat& transformation)
...
         {
-          eval = metric.Evaluate(transformedDataset.col(i),
-                    transformedDataset.col(targetNeighbors(j, i))) -
-                 distance(l, i);
+          eval = evalOld + arma::norm(transformation - transformationOld) *
+              (norm(targetNeighbors(j, i)) + norm(impostors(l, i)) +
+              2 * norm(i));
+          if (eval > -1)
+          {
+            // Calculate exact eval.
```
manish7294: I think this may be the reason for lack of speedup. Here each time` eval > -1` we are forced to do computation two times, one for eval bound and one for exact eval value.

rcurtin: Yes, but your simulations in #1449 showed that a large number of the `eval` calculations could be skipped.  Have you tried finding out if we are skipping the `eval` calculation the expected number of times?

manish7294: I think I made some serious mistake while doing those simulations, I can't seem to reproduce those result. I even tried with previous 12  days versions of LMNN but all are giving same results, and are totally opposite to the simulations in #1449. According to current simulation, there are very few points on which eval < -1.

rcurtin: I'm virtually certain your simulations were correct---so, under that assumption, what's different between what you did there and what's implemented here?

manish7294: I think I had done the same thing like I have initiated three count variables and after each eval calculation I have set three if else if conditions to increase the count based on eval value and just before return statement I am printing the values.

rcurtin: Right, but you should be able to compute the number of evals as you did in #1449 by simply doing the same counting at the final check `if (eval <= -1) { bp = l; break; }`.

manish7294: that's what strange that the result is not the same. I even printed the eval values.

rcurtin: I'd suggest starting the debugging from square one.  From my side I am quite confident that the original theory and your simulations in #1449 were correct (and this is supported by the observations in the JMLR paper).  So start from the beginning, make sure that you can get the original observations you were seeing in #1449 with unmodified code again, and then take a look at your implementation here and make sure that you still can.

Looking at the code my strong belief is that simply counting the number of `eval`s less than -2, less than -1, and greater than -1 in the code you have here should provide the same result.  But I believe that the bound implementation itself is wrong.  However, that is just my intuition, which is no substitute for digging into the problem and figuring out what is wrong.

manish7294: I couldn't get those results on iris and vc2 but with letters dataset, it is much better
```
(eval  > -1) --- (-2 < eval < -1) --- (eval < -2)
[DEBUG] Compiled with debugging symbols.
30240---28381---69168
[DEBUG] L-BFGS iteration 0; objective 124341, gradient norm 141545, 0.
13051---20816---77796
[DEBUG] L-BFGS iteration 1; objective 93106.6, gradient norm 44531.9, 0.2512.
5137---23661---75860
[DEBUG] L-BFGS iteration 2; objective 49849.1, gradient norm 19805.7, 0.464602.
30---22888---77110
[DEBUG] L-BFGS iteration 3; objective 22247.1, gradient norm 17154.7, 0.55371.
0---636---99364
[DEBUG] L-BFGS iteration 4; objective 9315.77, gradient norm 14277.2, 0.58126.
0---8---99992
[DEBUG] L-BFGS iteration 5; objective 5669.07, gradient norm 10134.7, 0.391455.
0---0---100000
```
I think I probably made some mistake earlier.

rcurtin: It may be a little time consuming but I think it's very important to understand what is going on here.  So I think it is worthwhile to take the time to dig in further and see for sure if you made a mistake earlier or what.  Since I can't see the code you are using, I can't review it myself.  And with the code you just ran it looks a little suspicious to me since there are 0 evals greater than -1 by the fifth L-BFGS iteration.  That doesn't seem right to me.

During this whole process I would say it's really important to keep rigorous and detailed notes to avoid confusion like this, since if we want to eventually publish on it we need to be 100% sure that what we're writing is correct.

manish7294: Ideally, it should be like this: 
```
(eval  > -1)----(-2 < eval < -1)----(eval < -2)
1298----862----1590
944----656----2150
667----426----2657
921----647----2182
655----408----2687
898----636----2216
646----391----2713
866----633----2251
640----375----2735
```
But using these new bounds results are:
```
(eval  > -1)----(-2 < eval < -1)----(eval < -2)
0----0----3750
0----0----3750
0----0----3750
0----0----3750
0----0----3750
0----3750----0
0----3750----0
0----3750----0
0----3750----0
```
For cross - checking if I bypass our bounds by setting `if (transformationOld.n_elem != 0)` to false, then we are getting ideal results.

manish7294:
@rcurtin I think now the implementation is fully refined. runtime has also improved w.r.t current LMNN implementation.

manish7294:
The thing I am gonna write will contradict my above comment, Sorry about that. 
The runtimes are good and on par with earlier timings but I can't strongly say that they are much better.
Here's the reason why I think it is so. Our bounds algorithm efficiently prune most of the points on which `eval < -1`, means these point never encounters the bound checking condition `eval = evalOld + transformationDiff * (norm(targetNeighbors(j, i)) + norm(impostors(l, i)) + 2 * norm(i))` and all those points that pass through it are majorly those with ` eval > -1` leading to double calculation (one we are calculating upper bound for eval and then we are forced to calculate exact eval as well). The yesterday's eval simulations which were done without commenting out 'if (eval < -1) { bp = l; break;}' also support this.
In general for every few thousands `(eval > -1)`, we are just encountering few hundred's of` (eval < -1)`, all others are being pruned.

src/mlpack/methods/lmnn/lmnn_function_impl.hpp
```diff
@@ -439,28 +531,46 @@ double LMNNFunction<MetricType>::EvaluateWithGradient(
...
         // triplets.
         double eval = 0;
-        if ((iteration - 1 % range == 0) && (iteration - 1 != 0))
+        if (transformationOld.n_elem != 0)
         {
-          eval = metric.Evaluate(transformedDataset.col(i),
-                    transformedDataset.col(targetNeighbors(j, i))) -
-                 distance(l, i);
+          eval = evalOld + transformationDiff * (norm(targetNeighbors(j, i)) +
+              norm(impostors(l, i)) + 2 * norm(i));
```
rcurtin: Right, caching the norms for this bound can make a huge difference for speedup.

src/mlpack/methods/lmnn/lmnn_function.hpp
```diff
@@ -217,6 +218,12 @@ class LMNNFunction
   Constraints<MetricType> constraint;
   //! Holds pre-calculated cij.
   arma::mat pCij;
+  //! Holds the norm of each data point.
+  arma::vec norm;
+  //! Hold previous eval value.
+  double evalOld;
```
rcurtin: This confuses me a little bit---shouldn't we cache one `evalOld` for each query point / impostor combination?  I don't think it should just be one `eval` for the whole dataset.

src/mlpack/methods/lmnn/lmnn_function_impl.hpp
manish7294: Ya, these are cached norms (just have same name as arma::norm) :)

src/mlpack/methods/lmnn/lmnn_function.hpp
manish7294: Right, It seems like a big mistake. I think it's just a coincidence that result didn't differ much  :)

rcurtin: No worries, if you want to fix it and push the change we can see if there is a big improvement. :+1:

manish7294: I have made the changes. Probably results are most consistent but sadly, there is no improvement in the runtimes :(

rcurtin: Then probably something is not right in the code.  I also don't know what "no improvement" means, since you've already pointed out that the runtime is improved (but I don't know how much).  I'd suggest that we start debugging here, so some things to check:

 * How tight *should* the bound be?  Remember that the bound term is ||L_{t - 1} - L_t|| * (||x_a|| + ||x_b|| + 2 ||x_i||)---is that typically large or small?  Norm-centering the dataset can also help reduce the norms on average.

 * If the step size is sufficiently small, then || L_{t - 1} - L_t || should be so small that basically all evaluations should be bounded.  So you should be able to see, as the number of iterations increase (and the steps get smaller), that more and more pruning can happen.  If that's not happening, there's likely to be a bug.

 * The results may be different for different datasets.  So we should always be careful to run the benchmarks against a handful of datasets instead of just one or two, before saying that there is no improvement.

I think that in this case there is likely to be some bug though.  I'll take a look at the code when I have a chance.

manish7294: Just for the confirmation, have you read https://github.com/mlpack/mlpack/pull/1461#issuecomment-403009285 ?

rcurtin: Yes, I read it, but at that time the code was only storing one `eval` for the entire dataset, which is definitely not right.  In fact I have found that the code never actually prunes anything.  Let me write up a longer comment shortly.

src/mlpack/methods/lmnn/lmnn_function.hpp
```diff
@@ -217,6 +218,12 @@ class LMNNFunction
   Constraints<MetricType> constraint;
   //! Holds pre-calculated cij.
   arma::mat pCij;
+  //! Holds the norm of each data point.
+  arma::vec norm;
+  //! Hold previous eval values for each datapoint.
+  arma::vec evalOld;
```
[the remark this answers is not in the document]
manish7294: First of all, Thanks a lot for doing this. That's a whole new level of debugging from my view. It has helped in clearing a lot of things :)

I think I can at least slightly see what you mean here, While doing eval derivation we started taking a particular `x_i, x_a`, and `x_b`. So, the new eval bound `eval_new <= eval_old + || L_{t + 1} - L_t ||_F^2 * ( ||x_a||^2 + 2 * ||x_i||^2 + ||x_b||^2 )` should be for those particular points only. If it's right, then do you think we need to store eval for each triplet `(x_i, x_a, x_b)`?
 Which might be a bit inconvenient though.

src/mlpack/methods/lmnn/lmnn_function_impl.hpp
```diff
@@ -107,33 +117,54 @@ double LMNNFunction<MetricType>::Evaluate(const arma::mat& transformation)
...
-        if (eval < -1)
+        if (eval <= -1)
         {
           // update bound.
           bp = l;
           break;
         }
 
+        // Update cache eval value.
+        evalOld[i] = eval;
```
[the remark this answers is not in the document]
manish7294: I always had my doubts regarding this, and since I was always neglecting and considering the marginal improvement as system noise. After some more confusing thoughts, I end up putting it here.

So basically this should be done at least before the `break` statement, right?

manish7294:
I forgot to mention this, It strike me today while working over bounds for impostors. I think we may not be able to use this for optimizers using batches as it would be a lot inconvenient to keep track of transformationOld for points of each batch. So I think in this code that part is wrong.

rcurtin:
Right, so I think maybe holding a bunch of transformation matrices and then a list for each point of which transformation matrix was last used could be a way to solve this.

Alternately, we could store only one transformation matrix when `Shuffle()` is called, and then be careful about how we compute `evalOld` for each point (since it would not be computed with the exact same transformation matrix).

manish7294:
I think we may have to store a transformation matrix for each point as apart from shuffle(), we have some optimizers having variable batchSize as well. So that could cause some extra problems. In terms of memory, it will be of the same order as of `evalOld`.

src/mlpack/methods/lmnn/lmnn_function_impl.hpp
```diff
@@ -439,22 +619,45 @@ double LMNNFunction<MetricType>::EvaluateWithGradient(
...
+        // Flag to trigger exact eval calculation.
+        bool exactEval = true;
+
+        // Bounds for eval.
+        if (transformationOld.n_elem != 0 && !std::isnan(evalOld[i](j, l)))
         {
-          eval = metric.Evaluate(transformedDataset.col(i),
-                    transformedDataset.col(targetNeighbors(j, i))) -
-                 distance(l, i);
+          eval = evalOld[i](j, l) + transformationDiff *
```
manish7294: This is one of the biggest problems. Each time` impostors()` is called mostly the triplets {i, j, l} changes and since `evalOld` contains values for different triples {i, j ,l}, it gets all wrong. 
One solution may be to reset evalOld on every `impostors()` call but then we will lose the significance of this optimization.

src/mlpack/methods/lmnn/lmnn_function_impl.hpp
```diff
@@ -439,22 +619,45 @@ double LMNNFunction<MetricType>::EvaluateWithGradient(
...
+          {
+            eval = metric.Evaluate(transformedDataset.col(i),
+                     transformedDataset.col(targetNeighbors(j, i))) -
+                   metric.Evaluate(transformedDataset.col(i),
+                       transformedDataset.col(impostors(l, i)));
+          }
         }
 
+        // Update cache eval value.
+        evalOld[i](j, l) = eval;
```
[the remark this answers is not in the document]
rcurtin: Is it actually a problem if we store an approximate bounded value in evalOld?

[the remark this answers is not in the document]
manish7294: I think it can be when approximate eval is very close to -1 than actual eval or vice-versa.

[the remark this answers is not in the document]
manish7294: At the very least we won't be getting same results.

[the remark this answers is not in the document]
rcurtin: But does your idea here match the results I was showing in my simulation?  Note that I was comparing the output transformation matrices and received the exact same results.  To put it more clearly: suppose that we take `evalOld[i](j, l) = eval + x` for some x greater than 0; now, the bound for that particular `evalOld` is loose.  What will happen the next iteration?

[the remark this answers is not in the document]
manish7294: Sorry, that was just a thought because of which earlier I was updating it after break statement. If the results don't divert then I think we shouldn't worry about this. Truly, I also haven't seen any deviation but was fearing it might happen with some different dataset.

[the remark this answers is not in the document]
rcurtin: Don't worry, it's definitely ok and a good thing to have the thought, but it's worthwhile to follow through on the thinking and make sure we're on the same page.  Do you mind working out what happens the next iteration if `evalOld` is loose like I wrote above?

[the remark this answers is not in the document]
manish7294: I think I haven't got that part properly. Just to be sure: loose like - we don't calculate exact value if we have calculated approximate one?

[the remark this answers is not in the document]
manish7294: Or do you mean just any random x > 0 to be placed there?

src/mlpack/methods/lmnn/lmnn_function_impl.hpp
manish7294: Please have a look at this one too. It worries me the most.

src/mlpack/methods/lmnn/lmnn_function_impl.hpp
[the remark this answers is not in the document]
rcurtin: Right, it turns out these are the same things, but I would say it's easier to think about if you just say that we've taken `evalOld[i](j, l) = eval + x` for some `x` greater than 0.  Later we can generalize to what happens if we've taken some approximate value.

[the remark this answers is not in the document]
manish7294: Without x:

Iris:
```
Evaluate(): 0 pruned of 2383, transformation diff 0.  Active: 1949, close 420, inactive 14.
Evaluate(): 0 pruned of 1782, transformation diff 0.998596.  Active: 1245, close 251, inactive 286.
Evaluate(): 0 pruned of 1655, transformation diff 0.492509.  Active: 1074, close 258, inactive 323.
Evaluate(): 0 pruned of 1339, transformation diff 1.34492.  Active: 685, close 244, inactive 410.
Evaluate(): 0 pruned of 1475, transformation diff 0.672462.  Active: 847, close 249, inactive 379.
Evaluate(): 0 pruned of 1364, transformation diff 0.47076.  Active: 710, close 243, inactive 411.
Evaluate(): 0 pruned of 1411, transformation diff 0.23538.  Active: 771, close 244, inactive 396.
Evaluate(): 233 pruned of 1450, transformation diff 0.11769.  Active: 818, close 324, inactive 308.
Evaluate(): 78 pruned of 1460, transformation diff 0.058845.  Active: 829, close 312, inactive 319.
Evaluate(): 321 pruned of 1467, transformation diff 0.0294225.  Active: 838, close 355, inactive 274.
Evaluate(): 370 pruned of 1472, transformation diff 0.0147113.  Active: 844, close 294, inactive 334.
Evaluate(): 443 pruned of 1473, transformation diff 0.00735563.  Active: 845, close 304, inactive 324.
Evaluate(): 485 pruned of 1473, transformation diff 0.00367782.  Active: 845, close 319, inactive 309.
Evaluate(): 532 pruned of 1473, transformation diff 0.00183891.  Active: 845, close 317, inactive 311.
Evaluate(): 560 pruned of 1475, transformation diff 0.000919454.  Active: 847, close 321, inactive 307.
Evaluate(): 579 pruned of 1475, transformation diff 0.000459727.  Active: 847, close 321, inactive 307.
Evaluate(): 607 pruned of 1475, transformation diff 0.000229863.  Active: 847, close 321, inactive 307.
Evaluate(): 612 pruned of 1475, transformation diff 0.000114932.  Active: 847, close 322, inactive 306.
Evaluate(): 622 pruned of 1475, transformation diff 5.74659e-05.  Active: 847, close 322, inactive 306.
Evaluate(): 623 pruned of 1475, transformation diff 2.87329e-05.  Active: 847, close 322, inactive 306.
Evaluate(): 626 pruned of 1475, transformation diff 1.43665e-05.  Active: 847, close 322, inactive 306.
Evaluate(): 627 pruned of 1475, transformation diff 7.18323e-06.  Active: 847, close 322, inactive 306.
Evaluate(): 626 pruned of 1475, transformation diff 3.59162e-06.  Active: 847, close 322, inactive 306.
Evaluate(): 628 pruned of 1475, transformation diff 1.79581e-06.  Active: 847, close 322, inactive 306.
Evaluate(): 628 pruned of 1475, transformation diff 8.97904e-07.  Active: 847, close 322, inactive 306.
Evaluate(): 628 pruned of 1475, transformation diff 4.48952e-07.  Active: 847, close 322, inactive 306.
```
vc2:
```
Evaluate(): 0 pruned of 4223, transformation diff 0.0583577.  Active: 3931, close 157, inactive 135.
Evaluate(): 0 pruned of 4417, transformation diff 0.0457136.  Active: 4187, close 207, inactive 23.
Evaluate(): 0 pruned of 4457, transformation diff 0.041658.  Active: 4212, close 187, inactive 58.
Evaluate(): 0 pruned of 4450, transformation diff 0.0165774.  Active: 4224, close 192, inactive 34.
Evaluate(): 0 pruned of 4484, transformation diff 0.00613733.  Active: 4265, close 190, inactive 29.
Evaluate(): 0 pruned of 4520, transformation diff 0.00481902.  Active: 4316, close 178, inactive 26.
Evaluate(): 0 pruned of 4501, transformation diff 0.00240951.  Active: 4284, close 191, inactive 26.
Evaluate(): 27 pruned of 4491, transformation diff 0.00120476.  Active: 4273, close 217, inactive 1.
Evaluate(): 52 pruned of 4487, transformation diff 0.000602378.  Active: 4267, close 199, inactive 21.
Evaluate(): 92 pruned of 4484, transformation diff 0.000301189.  Active: 4265, close 206, inactive 13.
Evaluate(): 118 pruned of 4483, transformation diff 0.000150594.  Active: 4264, close 206, inactive 13.
Evaluate(): 143 pruned of 4484, transformation diff 7.52972e-05.  Active: 4265, close 208, inactive 11.
Evaluate(): 181 pruned of 4484, transformation diff 3.76486e-05.  Active: 4265, close 209, inactive 10.
Evaluate(): 194 pruned of 4484, transformation diff 1.88243e-05.  Active: 4265, close 210, inactive 9.
Evaluate(): 202 pruned of 4484, transformation diff 9.41215e-06.  Active: 4265, close 210, inactive 9.
Evaluate(): 211 pruned of 4484, transformation diff 4.70608e-06.  Active: 4265, close 210, inactive 9.
Evaluate(): 212 pruned of 4484, transformation diff 2.35304e-06.  Active: 4265, close 210, inactive 9.
Evaluate(): 213 pruned of 4484, transformation diff 1.17652e-06.  Active: 4265, close 210, inactive 9.
Evaluate(): 217 pruned of 4484, transformation diff 5.8826e-07.  Active: 4265, close 210, inactive 9.
Evaluate(): 219 pruned of 4484, transformation diff 2.9413e-07.  Active: 4265, close 210, inactive 9.
Evaluate(): 219 pruned of 4484, transformation diff 1.47065e-07.  Active: 4265, close 210, inactive 9.
Evaluate(): 219 pruned of 4484, transformation diff 7.35324e-08.  Active: 4265, close 210, inactive 9.
Evaluate(): 219 pruned of 4484, transformation diff 3.67662e-08.  Active: 4265, close 210, inactive 9.
Evaluate(): 219 pruned of 4484, transformation diff 1.83831e-08.  Active: 4265, close 210, inactive 9.
Evaluate(): 219 pruned of 4484, transformation diff 9.19156e-09.  Active: 4265, close 210, inactive 9.
Evaluate(): 219 pruned of 4484, transformation diff 4.59578e-09.  Active: 4265, close 210, inactive 9.
```

with a random x:
Iris:
```
Evaluate(): 0 pruned of 3008, transformation diff 0.  Active: 2697, close 311, inactive 0.
Evaluate(): 0 pruned of 1956, transformation diff 0.998204.  Active: 1457, close 234, inactive 265.
Evaluate(): 0 pruned of 1877, transformation diff 0.231619.  Active: 1364, close 230, inactive 283.
Evaluate(): 0 pruned of 1795, transformation diff 0.254781.  Active: 1259, close 238, inactive 298.
Evaluate(): 0 pruned of 1135, transformation diff 5.42563.  Active: 436, close 184, inactive 515.
Evaluate(): 0 pruned of 1271, transformation diff 2.71282.  Active: 588, close 226, inactive 457.
Evaluate(): 0 pruned of 1276, transformation diff 0.633136.  Active: 597, close 227, inactive 452.
Evaluate(): 0 pruned of 1348, transformation diff 0.52932.  Active: 688, close 239, inactive 421.
Evaluate(): 63 pruned of 1365, transformation diff 0.178184.  Active: 711, close 305, inactive 349.
Evaluate(): 106 pruned of 1347, transformation diff 0.148173.  Active: 691, close 345, inactive 311.
Evaluate(): 153 pruned of 1356, transformation diff 0.0740863.  Active: 701, close 273, inactive 382.
Evaluate(): 277 pruned of 1361, transformation diff 0.0370431.  Active: 707, close 368, inactive 286.
Evaluate(): 271 pruned of 1363, transformation diff 0.0185216.  Active: 708, close 354, inactive 301.
Evaluate(): 336 pruned of 1364, transformation diff 0.00926078.  Active: 709, close 342, inactive 313.
Evaluate(): 393 pruned of 1364, transformation diff 0.00463039.  Active: 709, close 399, inactive 256.
Evaluate(): 371 pruned of 1364, transformation diff 0.0023152.  Active: 709, close 398, inactive 257.
Evaluate(): 380 pruned of 1365, transformation diff 0.0011576.  Active: 710, close 381, inactive 274.
Evaluate(): 401 pruned of 1365, transformation diff 0.000578799.  Active: 711, close 363, inactive 291.
Evaluate(): 408 pruned of 1365, transformation diff 0.000289399.  Active: 711, close 377, inactive 277.
Evaluate(): 398 pruned of 1365, transformation diff 0.0001447.  Active: 711, close 380, inactive 274.
```

vc2:
```
Evaluate(): 0 pruned of 4535, transformation diff 0.0519045.  Active: 4351, close 147, inactive 37.
Evaluate(): 0 pruned of 4470, transformation diff 0.00897.  Active: 4259, close 165, inactive 46.
Evaluate(): 0 pruned of 4485, transformation diff 0.00883794.  Active: 4276, close 175, inactive 34.
Evaluate(): 0 pruned of 4486, transformation diff 0.00396619.  Active: 4272, close 180, inactive 34.
Evaluate(): 0 pruned of 4505, transformation diff 0.003486.  Active: 4295, close 180, inactive 30.
Evaluate(): 6 pruned of 4497, transformation diff 0.001743.  Active: 4285, close 185, inactive 27.
Evaluate(): 25 pruned of 4490, transformation diff 0.000871501.  Active: 4276, close 196, inactive 18.
Evaluate(): 39 pruned of 4488, transformation diff 0.00043575.  Active: 4274, close 194, inactive 20.
Evaluate(): 43 pruned of 4486, transformation diff 0.000217875.  Active: 4272, close 199, inactive 15.
Evaluate(): 60 pruned of 4486, transformation diff 0.000108938.  Active: 4272, close 199, inactive 15.
Evaluate(): 64 pruned of 4486, transformation diff 5.44688e-05.  Active: 4272, close 199, inactive 15.
Evaluate(): 64 pruned of 4486, transformation diff 2.72344e-05.  Active: 4272, close 199, inactive 15.
Evaluate(): 54 pruned of 4486, transformation diff 1.36172e-05.  Active: 4272, close 190, inactive 24.
Evaluate(): 62 pruned of 4486, transformation diff 6.8086e-06.  Active: 4272, close 199, inactive 15.
Evaluate(): 70 pruned of 4486, transformation diff 3.4043e-06.  Active: 4272, close 203, inactive 11.
Evaluate(): 60 pruned of 4486, transformation diff 1.70215e-06.  Active: 4272, close 200, inactive 14.
Evaluate(): 54 pruned of 4486, transformation diff 8.51075e-07.  Active: 4272, close 192, inactive 22.
Evaluate(): 67 pruned of 4486, transformation diff 4.25537e-07.  Active: 4272, close 201, inactive 13.
Evaluate(): 77 pruned of 4486, transformation diff 2.12769e-07.  Active: 4272, close 197, inactive 17.
Evaluate(): 61 pruned of 4486, transformation diff 1.06384e-07.  Active: 4272, close 195, inactive 19.
Evaluate(): 65 pruned of 4486, transformation diff 5.31922e-08.  Active: 4272, close 194, inactive 20.
Evaluate(): 77 pruned of 4486, transformation diff 2.65961e-08.  Active: 4272, close 199, inactive 15.
Evaluate(): 66 pruned of 4486, transformation diff 1.3298e-08.  Active: 4272, close 201, inactive 13.

```
I don't think there is much deviation here.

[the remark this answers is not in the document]
rcurtin: So I can see from the results that fewer points are pruned, but what I was hoping to have you think about is theoretically what should happen when we add some `x` to the `eval` value.  Will the bound still be valid the next iteration?  Consider taking a look at the derivation of this bound and plugging in the change of adding `x` to it.

manish7294:
Here's the theoretical essence as per my thought -
We have eval < evalOld + pointSpecificConstantTerm
So, if we are increasing evalOld by adding a positive x,  we are basically
increasing upper bound for new eval. It will be eval + x. Hence the
calculation for eval bound which we are doing will give eval increased by a
factor of x, hence pulling it towards positive axis or just reducing its
probability of being negative. Which all together means less number of
prunes.

Please tell me if it sounds a bit confusing.

src/mlpack/methods/lmnn/lmnn_function_impl.hpp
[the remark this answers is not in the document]
manish7294: Here's the theoretical essence as per my thought -
We have `eval < evalOld + pointSpecificConstantTerm` So, if we are increasing evalOld by adding a positive x, we are basically increasing upper bound for new eval. It will be `eval + x`. Hence the calculation for eval bound which we are doing will give `eval` increased by a factor of `x`, hence pulling it towards positive axis or just reducing its probability of being negative. Which all together means less number of prunes.
Please tell me if it sounds a bit confusing. I can try to clarify further.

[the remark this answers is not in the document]
rcurtin: The explanation makes sense to me.  I agree---increasing `x` would result in fewer prunes.  Would you then conclude that for any positive `x`, even if we take `evalOld[i](j, l) = eval + x`, the bound will still be correct?  (As in, it will still be a valid bound.)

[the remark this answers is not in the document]
manish7294: No doubt, It will surely be correct. Just the prunes will decrease.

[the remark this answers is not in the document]
manish7294: Having loose bounds definitely shouldn't be wrong.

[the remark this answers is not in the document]
rcurtin: Ok, I agree.  Now we can get back to your original question about the placement of how `evalOld` is assigned.  Even if we assign `evalOld` before the prune, the worst case is that we assign a bound instead of an inexact value (so, this is equivalent to saying `evalOld[i](j, l) = eval + x` where `x` is the bounding term || L_{t + 1} - L_t || * (|| x_j || + || x_l || + 2 || x_i ||).  Since we both agree that it is still a valid bound, just possibly loose, there is no problem with doing this.  In fact, in the next iteration, if the exact calculation of `eval` is not done and we end up pruning based on another bound, our stored `evalOld` value will be loose---but not invalid.  So we will still get some prunes, but it will not be perfect.

The other option, as you pointed out, is to assign `evalOld` after the pruning.  But in this case, this means that we will never set any `evalOld` value for a lot of pruned points, and then it won't be possible to prune them in the following iteration.  You can see this in the simulations I ran yesterday---without setting `evalOld` before the pruning, we get zero prunes at all throughout the entire algorithm (and no speedup).

So, given these two choices, it's best to go with the approach that will give some prunes despite the bound being looser than it could possibly be.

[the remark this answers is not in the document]
manish7294: Totally Agree :) Thanks for clearing this out.

src/mlpack/methods/lmnn/lmnn_function_impl.hpp
rcurtin: Yes, I wanted to wait to discuss this one until we took care of the other part.  I thought about it this morning and because the bound depends on `|| x_i ||`, `|| x_j ||`, and `|| x_l ||`, I believe it is possible that the impostors changing could cause a problem.  But let's talk a little bit more about what the problem is.  As always please read carefully, it's possible there may be an error in what I've written.

Our whole goal is to bound `eval` for the next iteration.  We've done this by computing the *maximum* possible `eval` value for the next iteration (not the minimum).  Implicitly our derivation does this by bounding how much closer the impostor `x_l` could get given `|| L_{t + 1} - L_t ||`.  So even if the impostor changes, the bound will not.  I haven't been very clear in my description here (but I am about to go to lunch so I want to keep it short), but I'd encourage you to sit down and try to work this out if you like.

That said there is still another problem: each iteration, we compute this bounding term || L_{t + 1} - L_t || * (|| x_j || + || x_l || + 2 || x_i ||).  But if `x_l` changes, then suddenly you are right, it is possible that the bound could be incorrect---because we need to be using `x_l` with respect to the original impostor `x_l` that the `eval` was calculated with, not a new impostor point.

I thought about it for a while, and I think the problem could be solved if you did this: every time we perform a direct computation of `eval`, store the quantity || x_j || + || x_l || + 2 || x_i ||.  Then when we compute the bound, just use that stored quantity.  Now, even if the impostor changes, the bound will still be valid.  As soon as the bound is no longer able to prune (i.e. once it becomes greater than `-1`), then we'll perform another direct computation of `eval` and store the new quantity with the new impostors.

What do you think, does that make sense?

rcurtin: Right, I think that I agree with the derivation here, so definitely if `|| x_k ||^2 > || x_{k + 1} ||^2`, we can use the previous bound as-is.  In the other case, where `|| x_k ||^2 <= || x_{k + 1} ||^2` (that is, the new impostor `x_{k + 1}` has larger norm than the old impostor `x_k`), a valid bound is

```
d(L_t x_i, L_t x_k) - A || x_{k + 1} ||^2 - B,  // call this (1)
```

although that bound would actually be somewhat looser than the bound

```
d(L_t x_i, L_t x_{k + 1}) - A || x_{k + 1} ||^2 - B.  // call this (2)
```

It's possible to put these together to make a nice bound that uses `max(|| x_k_t ||^2, || x_k_{t + 1}} ||^2)`, where `x_k_t` (sorry for the ugly notation) is impostor `k` at iteration `t` and `x_k_{t + 1}` is impostor `k` at iteration `t + 1`.  I think it might be possible to make that tighter by exploiting that impostors can only change when `iteration - 1 % range == 0`, but I think the result would be really complex.  For that `max()`, note that when impostors don't change, `x_k_t == x_k_{t + 1}` and the bound is the same as before.

manish7294: So, our new condition is basically,
`eval_new <= eval_old + || L_{t + 1} - L_t ||_F^2 * ( ||x_a||^2 + 2 * ||x_i||^2 + max(||x_b||^2,  ||x_b_p||^2))`
Here,
x_a : target neighbor
x_b : current impostor
x_b_p : previous impostor
x_i : current data point

As per the implementation technicalities, I think we will have to hold the last maximum impostors norm used and will be solving - 
`eval_new <= eval_old + || L_{t + 1} - L_t ||_F^2 * ( ||x_a||^2 + 2 * ||x_i||^2 + max(||x_b||^2,  previousMaxNorm))`

So basically we will have to hold another cube.

manish7294: Or, maybe just a 2D matrix will do.

manish7294:
Hopefully, this will be the last problem in the course of this PR -  
I have no idea why by BigBatchSGD after a number of iterations just keeps on running without printing any other iteration output. As if it is just stuck at a point. I have observed this several time but couldn't find an answer.

src/mlpack/methods/lmnn/lmnn_function_impl.hpp
```diff
@@ -137,15 +137,32 @@ double LMNNFunction<MetricType>::Evaluate(const arma::mat& transformation)
...
-        if (transformationOld.n_elem != 0 && !std::isnan(evalOld[i](j, l)))
+        if (transformationOld.n_elem != 0 && !std::isnan(evalOld(j, l, i)))
         {
-          eval = evalOld[i](j, l) + transformationDiff *
-            (norm(targetNeighbors(j, i)) + norm(impostors(l, i)) +
+          double maxNorm;
+          if (!std::isnan(maxImpNorm(l, i)))
+            maxNorm = std::max(norm(impostors(l, i)), maxImpNorm(l, i));
+          else
+            maxNorm = norm(impostors(l, i));
```
rcurtin: We could avoid the `if` statement here if we initialize `maxImpNorm` elements to 0.  (Correct me if I am wrong---but I think that will work.)  You could even just do

```
maxImpNorm(l, i) = std::max(maxImpNorm(l, i), norm(impostors(l, i)));
```

in that case.

src/mlpack/methods/lmnn/lmnn_function_impl.hpp
```diff
@@ -137,15 +137,32 @@ double LMNNFunction<MetricType>::Evaluate(const arma::mat& transformation)
...
+
+          // Update cache max impostor norm.
+          maxImpNorm(l, i) = maxNorm;
+
+          eval = evalOld(j, l, i) + transformationDiff *
```
rcurtin: This is a minor point, but since Armadillo is column-major, points in the same column end up being contiguous in memory.  Since we're looping over `i`, then `j`, then `l`, the best access pattern for `evalOld` will be if we are accessing as `evalOld(l, j, i)` (i.e. `l` corresponds to the row, `j` to the column, `i` to the slice).

src/mlpack/methods/lmnn/lmnn_function_impl.hpp
```diff
@@ -196,6 +213,14 @@ double LMNNFunction<MetricType>::Evaluate(const arma::mat& transformation,
 {
   double cost = 0;
 
+  // Ensure cache transformation cube has correct size.
+  if (transformationOldPoint.n_elem == 0)
+  {
+    transformationOldPoint.set_size(transformation.n_rows,
+        transformation.n_cols, dataset.n_cols);
+    transformationOldPoint.fill(arma::datum::nan);
+  }
```
rcurtin: I think this means your approach is that you will store one transformation matrix per point?  That will be very large in memory.  I'd suggest instead a `std::vector<arma::mat>` of old transformation matrices, and then a `arma::Row<size_t>` to track the index that is used for each point.  (I know that earlier I suggested changing from a `std::vector<arma::mat>` to an `arma::cube`, but that was for a case where the number of slices in the cube wouldn't change.  Here, I think it would be better for memory to only store as many transformation matrices as we need.  Also, I guess each call to Shuffle() we can check for unused transformation matrices and either remove them or just call `clear()`.)

rcurtin:
> I have no idea why BigBatchSGD after a number of iterations just keeps on running without printing any other iteration output. As if it is just stuck at a point. I have observed this several time but couldn't find an answer.

Hmm, let's look into this once we resolve the couple of other comments and testing.  If you can tell me how to reproduce it, I can help guide you through the debugging.

src/mlpack/methods/lmnn/lmnn_function_impl.hpp
manish7294: Definitely, that's a better way of doing this.

src/mlpack/methods/lmnn/lmnn_function_impl.hpp
manish7294: I have some doubts regarding this -
This is mainly done for batch support, the reason why I have done this because of the changing batch size, Let''s suppose in next iteration batch size changed, then I am not able to see how will we keep  track of the transformationOld now. Please explain a little bit more, maybe I am missing something here.

rcurtin: Right, so the idea is this.  Hold an `arma::Row<size_t>` called `lastTransformationIndices` (or something), and hold a `std::vector<arma::mat>` called `oldTransformationMatrices` (or something).  Then, when we do a batch:

```
oldTransformationMatrices.push_back(transformationMatrix);
lastTransformationIndices.subvec(begin, begin + batchSize - 1) = oldTransformationMatrices.size();
```

Then, when you need to get the transformation matrix for a point `i`, do `oldTransformationMatrices[lastTransformationIndices[i]]`.  Do you see what I mean?  There will be a little bit of extra complexity here but that is the basic idea.

rcurtin:
Ok, and what have you tried to produce output that could help pinpoint what the problem is?
(Feel free to wait to approach this issue until the other issues are fixed, that's fine with me.  It's possible that some of the other changes you will make may solve this.)

src/mlpack/methods/lmnn/lmnn_function_impl.hpp
manish7294: The update process is nasty and will require some extra computation time at each Evaluate() call
```cpp
  // Update & remove unused cache matrices.
  oldTransformationMatrices.push_back(transformation);
  lastTransformationIndices.cols(begin, begin + batchSize - 1) =
      oldTransformationMatrices.size() - 1;

  size_t maxIndex = lastTransformationIndices.max();
  arma::uvec matricesUsed = arma::unique(lastTransformationIndices);
  for (size_t i = 0, j = 0, l = 0; i <= maxIndex; i++)
  {
    if (i != matricesUsed(j))
    {
      oldTransformationMatrices.erase(i - l);
      l++;
    }
    else
      j++;
  }
  // Normalize lastTransformationIndices.
  Maybe NormalizeLabels() can be used for this.
```


So, Do you think it is worth the tradeoff for space? And I think we are using the same amount of space for `evalOld` as well.

manish7294:
It looks like making these new changes solved the problem. As I am not seeing it anymore. The optimization is successfully converging. Maybe something was wrong with `maxImpNorm` implementation earlier.

manish7294:
Here is the gist for correctness of distance - https://gist.github.com/manish7294/7f939c3ed80be5af7ea3f2b216304efb
The results are overall same, just a little difference with BigBatchSGD.

manish7294:
Here are some simulations https://gist.github.com/manish7294/4e7285e680f47ca5190ef6869215072f

rcurtin:
Thanks for the simulations.  It looks to me like there may be some problem---this pruning should be exact and should not affect the results at all.  But I can see for L-BFGS for the iris dataset and for BBSGD, the results are a little different.  Are you sure that any randomness is removed from BBSGD?

src/mlpack/methods/lmnn/lmnn_function_impl.hpp
rcurtin: Right, let's consider using a slightly different strategy, holding the following auxiliary variables:

```
// It might be useful to come up with shorter names, these are really long...

// This will hold each of the transformation matrices.
std::vector<arma::mat> oldTransformationMatrices;
// This will hold how many points are using each transformation matrix.
std::vector<size_t> oldTransformationCounts;
// This will hold which transformation matrix each point is using.
arma::Row<size_t> lastTransformationIndices;
```

Then, whenever we call `Evaluate(..., begin, batchSize)`, we can do the following:

```
// Are there any empty transformation matrices?
size_t index = oldTransformationMatrices.size();
for (size_t i = 0; i < oldTransformationCounts.size(); ++i)
{
  if (oldTransformationCounts[i] == 0)
  {
    index = i; // Reuse this index.
    break;
  }
}

// Did we find an unused matrix?  If not, we have to allocate new space.
if (index == oldTransformationMatrices.size())
{
  oldTransformationMatrices.push_back(transformationMatrix);
  oldTransformationCounts.push_back(0);
}
else
{
  oldTransformationMatrices[index] = transformationMatrix;
}
    
// Update all the transformation indices.
for (size_t i = begin; i < begin + batchSize; ++i)
  --oldTransformationCounts[lastTransformationIndices[i]];
lastTransformationIndices.cols(begin, begin + batchSize - 1) = index;
oldTransformationCounts[index] += batchSize;
```

During initialization for the whole procedure I'd suggest setting `oldTransformationMatrices` to have one empty matrix and set all the `lastTransformationIndices` to `0` and the counts accordingly.

And, when shuffling is done, we just need to modify `lastTransformationIndices`, which will be very similar to the existing shuffling code.

This is an improvement over the idea above, because we don't need to call `arma::unique()` every call, which is very expensive.  Instead the cost of this approach, at worst, is one transformation matrix copy followed by `2 * batchSize + 1` assignments, which is pretty trivial compared to the rest of the computation.

src/mlpack/methods/lmnn/lmnn_function_impl.hpp
```diff
@@ -225,10 +341,25 @@ void LMNNFunction<MetricType>::Gradient(const arma::mat& transformation,
...
-                          transformedDataset.col(targetNeighbors(j, i))) -
-                      metric.Evaluate(transformedDataset.col(i),
-                          transformedDataset.col(impostors(l, i)));
+        double eval = 0;
+
+        // Flag to trigger exact eval calculation.
+        bool exactEval = true;
+
+        // Use eval calualated during Evaluate.
+        if (!std::isnan(evalOld(l, j, i)))
```
manish7294: I think I have at least pinpointed the cause of difference in result - 

1.  In this PR we have breaking condition as `if (eval <= -1)` whereas in the current LMNN we have if `(eval < -1)`. Doing this made solved the iris deviation.

2.  There is something going wrong here while using cache evalOld for Gradient() function. If I change if (!std::isnan(evalOld(l, j, i))) to false, BBSGD deviation problem solves as well.

rcurtin: > 1. In this PR we have breaking condition as if (eval <= -1) whereas in the current LMNN we have if (eval < -1). Doing this made solved the iris deviation. So, I think we don't need to worry about this.

Got it, so you mean that when you run the current PR with `if (eval < -1)`, you get the exact same results?

> 2. There is something going wrong here while using cache evalOld for Gradient() function. If I change if (!std::isnan(evalOld(l, j, i))) to false, BBSGD deviation problem solves as well.

Sorry, do you mean you change to `if (false)` or `if (std::isnan(evalOld(l, j, i)) == false)` or something else?  I don't know if I understood.

manish7294: > Got it, so you mean that when you run the current PR with if (eval < -1), you get the exact same results with L-BFGS on the iris dataset as the current git master branch?

Yes

> Sorry, do you mean you change to if (false) or if (std::isnan(evalOld(l, j, i)) == false) or something else? I don't know if I understood.Sorry, do you mean you change to if (false) or if (std::isnan(evalOld(l, j, i)) == false) or something else? I don't know if I understood.

Ya, I wrote a confusing statement. Sorry, about that. I meant  if (false)

manish7294: Here's is my thought about eval <= -1 
Our original condition was max(0, 1 + eval), and 0 doesn't contribute anything so skipping it shouldn't be any problem. And at eval = -1, we are basically  dealing with 0. Does it sounds right?

Ya, I will try that.

rcurtin: You're right, when `eval == -1`, there is no change to the cost function.  But there is a change to the gradient:

```
         // Caculate gradient due to impostors.
         arma::vec diff = dataset.col(i) - dataset.col(targetNeighbors(j, i));
         cil += diff * arma::trans(diff);
 
         diff = dataset.col(i) - dataset.col(impostors(l, i));
         cil -= diff * arma::trans(diff);
```

In the git master branch, those terms get added to `cil` even if `eval == -1`.  When I review the paper it's not fully clear whether we should include those terms in our gradient computation.  The paper writes:

"Let us define a set of triples N_t, such that (i, j,l) ∈ N_t if and only if the indices (i, j,l) trigger the hinge loss in the second part of Eq. (18)."

I think we can go either way here, and I like the tighter condition of `if (eval <= -1)` since that will allow a little bit of extra pruning.  So I would say, let's leave it as it is with `<=`, but just in order to test correctness with BigBatchSGD, let's make sure we do that test with `<` to match the git master code.

manish7294:
Ah! finally! It looks like we are done here :)
Here's updated distance correctness simulations- https://gist.github.com/manish7294/7f939c3ed80be5af7ea3f2b216304efb

src/mlpack/methods/lmnn/lmnn_function_impl.hpp
manish7294: Thanks!  That's a really good algorithm to handle the situation :)

rcurtin:
Looks great! Thank you for the hard work.  I'll give this a final review shortly but I am pretty sure any comments will be very minor.

The last thing to check would be the speedup---can you run some timing simulations to checkwhat the speedup is on different datasets?

Thanks again for the hard work here.  I'm really happy to see this merged soon. :)

manish7294:
Timings Simulations:
Timings:
```
               With Change ----- Without Change
Iris -          0.841357 ------- 0.878832
vc2 -           3.255247 ------- 3.341104
balance_scale - 9.823953 ------- 12.755052
ecoli -         5.727747 ------- 5.909457
diabetes -      18.633512 ------ 18.724131
```

Pruning output -
https://gist.github.com/manish7294/cbafa9f0d1ba1370c5ef0160ec4b2430

src/mlpack/methods/lmnn/lmnn_function_impl.hpp
```diff
@@ -161,45 +227,122 @@ double LMNNFunction<MetricType>::Evaluate(const arma::mat& transformation,
...
       cost += (1 - regularization) * eval;
     }
 
+    // Calculate norm of change in transformation.
+    double transformationDiff = 0;
+    if (arma::is_finite(lastTransformationIndices(i)))
+    {
+      transformationDiff = arma::norm(transformation -
+          oldTransformationMatrices[lastTransformationIndices(i)]);
+    }
```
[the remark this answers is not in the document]
manish7294: I think it is possible to do this for amsgrad and sgd (change can be made in EvaluatewithGradient()) but will have to keep it like this for bbsgd.

src/mlpack/methods/lmnn/lmnn_function_impl.hpp
```diff
@@ -161,45 +227,122 @@ double LMNNFunction<MetricType>::Evaluate(const arma::mat& transformation,
...
+  }
+
+  // Update all the transformation indices.
+  for (size_t i = begin; i < begin + batchSize; ++i)
+  {
+    --oldTransformationCounts[lastTransformationIndices(i)];
+    lastTransformationIndices(i) = index;
+  }
+
+  oldTransformationCounts[index] += batchSize;
```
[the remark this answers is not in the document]
manish7294: Sure, I will make separate method for this update part.

src/mlpack/methods/lmnn/lmnn_function_impl.hpp
```diff
@@ -476,16 +867,52 @@ double LMNNFunction<MetricType>::EvaluateWithGradient(
...
+
+
   return cost;
 }
 
 template<typename MetricType>
 inline void LMNNFunction<MetricType>::Precalculate()
 {
   pCij.zeros(dataset.n_rows, dataset.n_rows);
+  norm.zeros(dataset.n_cols);
```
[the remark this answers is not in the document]
manish7294: Done in #1466

src/mlpack/methods/lmnn/lmnn_function_impl.hpp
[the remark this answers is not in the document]
rcurtin: Right, you can leave it like this and I think it's fine.  What I was thinking was more along the lines of something like this:

```
std::set<size_t> uniqueIndices;
std::map<size_t, double> transformationDiffs;
for (size_t i = begin; i < begin + batchSize; ++i)
{
  if (transformationDiffs.count(lastTransformationIndices[i]) == 0)
  {
    // If you make the change so that oldTransformationMatrices[0] is empty, this will work.
    if (lastTransformationIndices[i] == 0)
    {
      transformationDiffs[0] = 0.0; // This won't be used anyway...
    }
    else
    {
      transformationDiffs[lastTrasformationIndices[i]] = arma::norm(transformation -
          oldTransformationMatrices[lastTransformationIndices(i)]);
    }
  }
}
```

I think that will work, but I haven't tried or debugged it.  Then, later, instead of using `transformationDiff` for the bound calculation, you'd use `transformationDiff[oldTransformationIndices[i]]`.

src/mlpack/methods/lmnn/lmnn_function_impl.hpp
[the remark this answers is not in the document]
rcurtin: Ah, sorry, forgot about that.  :+1:

src/mlpack/methods/lmnn/lmnn_function_impl.hpp
```diff
@@ -57,16 +67,28 @@ LMNNFunction<MetricType>::LMNNFunction(const arma::mat& dataset,
...
-  labels = std::move(newLabels);
+  dataset = newDataset.cols(ordering);
+  labels = newLabels.cols(ordering);
+  maxImpNorm = newMaxImpNorm.cols(ordering);
+  lastTransformationIndices = newlastTransformationIndices.elem(ordering);
+
+  for (size_t i = 0; i < ordering.n_elem; i++)
+  {
+    evalOld.slice(i) = newEvalOld.slice(ordering(i));
+  }
```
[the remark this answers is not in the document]
manish7294: Done!

src/mlpack/methods/lmnn/lmnn_function_impl.hpp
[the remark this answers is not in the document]
manish7294: Done!

manish7294:
I think all the comments have been handled :)

manish7294:
> As a side note, I was a little surprised that so many of the constraints for many of the datasets were active. In many cases it seemed like only 10% or fewer of the constraints were less than -1 in the last iterations of the optimization. The text in the JMLR paper on page 240 under equation (19) suggests that most of the constraints won't be "active" (i.e. greater than -1), but this seems to not be what we are seeing. I see no problems in our code or anything, I just thought that was an interesting observation to point out.

Currently we are not visiting each and every point as `if (eval < -1) {bp = l; break;}` help us to avoid visiting major portion of the points having `eval < -1`.
Let's take the case of iris itself. Ideally, for `k = 5`, we should be having, `150 (dataset.n_cols) * 5 (Number of target neighbors) * 5 (Number of impostors) = 3750` constraints but as apparent from the result we actually visits around `1402` points and out of them `753` usually remains active. So the active to constraints percentage come down to `20.08%` for iris, similarly for balance scale - `57.49%`, ecoli - `67.55%`, vc2 - `76.50%`, diabetes - `95.40%` and for letters it is `0%`.

So, I think the text may be referring to the percentage w.r.t to total number of constraints. But still it seems, this figure is highly dataset dependent, and at least looking at these results, the statement of the authors - **most of the constraints won't be "active"** doesn't seems quite right.

rcurtin:
Oh, you are right about the number of constraints. I had not considered all the skipped ones.  Thank you for the clarification!

manish7294:
I'm the one who should be thanking you. Thanks for always having my back. I couldn't have done it without you. All of your's innovative ideas were Awesome. :)

src/mlpack/methods/lmnn/lmnn_function_impl.hpp
```diff
@@ -107,35 +223,65 @@ double LMNNFunction<MetricType>::Evaluate(const arma::mat& transformation)
         // Calculate cost due to {data point, target neighbors, impostors}
         // triplets.
         double eval = 0;
-        if ((iteration - 1 % range == 0) && (iteration - 1 != 0))
+
+        // Bounds for eval.
+        if (transformationOld.n_elem != 0 && evalOld(l, j, i) < -1)
```
zoq: Perhaps ``.is_empty()`` is more descriptive?

src/mlpack/methods/lmnn/lmnn_function_impl.hpp
```diff
@@ -107,35 +223,65 @@ double LMNNFunction<MetricType>::Evaluate(const arma::mat& transformation)
...
         {
-          eval = metric.Evaluate(transformedDataset.col(i),
-                    transformedDataset.col(targetNeighbors(j, i))) -
-                 distance(l, i);
+          // Update cache max impostor norm.
+          maxImpNorm(l, i) = std::max(maxImpNorm(l, i), norm(impostors(l, i)));
+
+          eval = evalOld(l, j, i) + transformationDiff *
+              (norm(targetNeighbors(j, i)) + maxImpNorm(l, i) +
+              2 * norm(i));
```
zoq: Not sure, I get ``norm(i)`` here, it's always ``i`` if ``i >= 0``. Did I miss something? Perhaps the same applies to ``targetNeighbors``.

manish7294: It looks like you got `norm(i)` confused with `arma::norm(i)`(Though I guess arma::norm expects a vector or matrix to passed not a scalar) :)
Actually it is a vector which stores norm of each data point.
`norm(i) = arma::norm(dataset.col(i));`

src/mlpack/methods/lmnn/lmnn_function_impl.hpp
manish7294: Done!

src/mlpack/methods/lmnn/lmnn_function_impl.hpp
zoq: Ahh, this makes sense, thanks for the clarification.

rcurtin:
Thanks again. Very happy to see this come together. :+1:

PR #1462, opened 2018-07-04
Fix math::RandomSeed() BINDING_TYPE macro condition

manish7294:
Check whether the BINDING_TYPE is NULL or not. The order of mlpack_main.cpp inclusion is also corrected all over the codebase.

### DIFF
--- a/src/mlpack/core/math/random.hpp
+++ b/src/mlpack/core/math/random.hpp
@@ -39,7 +39,8 @@ extern MLPACK_EXPORT std::normal_distribution<> randNormalDist;
  */
 inline void RandomSeed(const size_t seed)
 {
-  #if (BINDING_TYPE == NULL || BINDING_TYPE != BINDING_TYPE_TEST)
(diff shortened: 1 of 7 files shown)


src/mlpack/core/math/random.hpp
```diff
@@ -39,7 +39,7 @@ extern MLPACK_EXPORT std::normal_distribution<> randNormalDist;
  */
 inline void RandomSeed(const size_t seed)
 {
-  #if (BINDING_TYPE != BINDING_TYPE_TEST)
+  #if BINDING_TYPE == BINDING_TYPE_UNKNOWN || BINDING_TYPE != BINDING_TYPE_TEST
```
rcurtin: I think it's probably better to use `defined(BINDING_TYPE)` instead of `BINDING_TYPE == BINDING_TYPE_UNKNOWN`.

manish7294:
Right, that's exactly why I kept `!defined(BINDING_TYPE)` there :)

rcurtin:
Thanks again :+1:

PR #1464, opened 2018-07-06
Swats optimizer

zoq:
Implementation of "Improving Generalization Performance by Switching from Adam to SGD" by Nitish Shirish Keskar and Richard Socher.

### DIFF
--- a/src/mlpack/core/optimizers/swats/swats_update.hpp
+++ b/src/mlpack/core/optimizers/swats/swats_update.hpp
@@ -120,7 +120,10 @@ class SWATSUpdate
       sgdRate = sgdLamda / biasCorrection2;
 
       if (std::abs(sgdRate - rate) < epsilon && iteration > 1)
+      {
         phaseSGD = true;
+        sgdV.zeros();
+      }
     }
   }
 
--- a/src/mlpack/core/optimizers/swats/swats_update.hpp
+++ b/src/mlpack/core/optimizers/swats/swats_update.hpp
@@ -1,0 +1,177 @@
+/**
+ * @file swats_update.hpp
+ * @author Marcus Edel
+ *
+ * SWATS update rule for Switches from Adam to SGD.
+ *
+ * mlpack is free software; you may redistribute it and/or modify it under the
+ * terms of the 3-clause BSD license.  You should have received a copy of the
+ * 3-clause BSD license along with mlpack.  If not, see
+ * http://www.opensource.org/licenses/BSD-3-Clause for more information.
+ */
+#ifndef MLPACK_CORE_OPTIMIZERS_SWATS_SWATS_UPDATE_HPP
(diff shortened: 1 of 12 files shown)


ShikharJ:
I took a quick look, and everything seems nice. I'll take a detailed look later during the day.

src/mlpack/core/optimizers/swats/swats_update.hpp
```diff
@@ -0,0 +1,177 @@
...
+ * condition is satisfied. The condition relates to the projection of Adam steps
+ * on the gradient subspace.
+ *
+ * For more information, see the following.
+ *
+ * @code
+ * @article{Keskar2017,
+ *   author  = {Nitish Shirish Keskar and Richard Socher},
+ *   title   = {Improving Generalization Performance by Switching from Adam to
+ *              {SGD}},
```
ShikharJ: Are the extra parentheses really required here?

src/mlpack/core/optimizers/swats/swats_update.hpp
```diff
@@ -0,0 +1,177 @@
...
+    const double deltaGradient = arma::dot(delta, gradient);
+    if (deltaGradient != 0)
+    {
+      const double rate = arma::dot(delta, delta) / deltaGradient;
+      sgdLamda = beta2 * sgdLamda + (1 - beta2) * rate;
+      sgdRate = sgdLamda / biasCorrection2;
+
+      if (std::abs(sgdRate - rate) < epsilon && iteration > 1)
+        phaseSGD = true;
+    }
```
ShikharJ: Has the implementation been deviated from the standard algorithm above?

src/mlpack/core/optimizers/swats/swats_update.hpp
zoq: The ``{{`` will prevent BibTeX from adjusting the capitalization, so instead of ``Sgd`` we get ``SGD``.

src/mlpack/core/optimizers/swats/swats_update.hpp
zoq: If I don't missed anything this should match with the algorithm from the paper, so a combination of SGD and Adam.

ShikharJ: @zoq Shouldn't the steps 19 to 23 from the paper also be included here?

zoq: We don't have to set ``v_k = 0`` , since ``sgdV`` is already 0, but I can change that to make it more clear. About line 23, you are right we have to reset the value for the next iteration.

zoq: On a second thought, we don't have to save ``v_(k-1)`` if the condition is false, ``v_k`` just stays the same, since we use the value from the previous iteration.

zoq:
Moved to: https://github.com/mlpack/ensmallen/pull/42

PR #1465, opened 2018-07-07
Fix/add references to Win tutorials and sample app

gmanlan:
Minor changes to link core.hpp and tutorials.txt to the recently added Windows build tutorial and sample app.

Let me know if I missed any other ref location.

### DIFF
--- a/doc/guide/build.hpp
+++ b/doc/guide/build.hpp
@@ -24,8 +24,7 @@ mlpack uses CMake as a build system and allows several flexible build
 configuration options.  One can consult any of numerous CMake tutorials for
 further documentation, but this tutorial should be enough to get mlpack built
 and installed on most Linux and UNIX-like systems (including OS X).  If you want
-to build mlpack on Windows, see <a
-href="https://keon.io/mlpack-on-windows/">Keon's excellent tutorial</a>.
+to build mlpack on Windows, see \ref build_windows.
 
 You can download the latest mlpack release from here:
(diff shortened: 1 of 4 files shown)


doc/guide/build.hpp
```diff
@@ -24,8 +24,7 @@ mlpack uses CMake as a build system and allows several flexible build
 configuration options.  One can consult any of numerous CMake tutorials for
 further documentation, but this tutorial should be enough to get mlpack built
 and installed on most Linux and UNIX-like systems (including OS X).  If you want
-to build mlpack on Windows, see <a
-href="https://keon.io/mlpack-on-windows/">Keon's excellent tutorial</a>.
+to build mlpack on Windows, see \ref build_windows.
```
zoq: I guess we could still link to Keon's tutorial, what do you think?

gmanlan: My understanding is that it is outdated (mlpack 2.0, VS2015). I think that was the main issue we discussed with @rcurtin, and the motivation to write the recent Win tutorial and sample app. Happy to add it again. Thoughts?

rcurtin: Maybe it's better to re-add in parentheses or something but mention that it is old?  Some of the descriptions and explanations there might still be useful, I think.

gmanlan: Right - I will re-add it with a note in case somebody is using VS2015. Thanks!

zoq:
With two approvals, I'll go ahead and merge this now, thanks again for the contribution, and as Ryan said happy to drop some stickers into the mail if you like.

Issue #1448, opened 2018-06-27
LMNN: optimization by avoiding impostor recalculation using bounds


rcurtin:
This is related to #1407 and is an optimization that we could do.  It focuses on the recalculation of impostors that is performed at each iteration (which happens every `range` iterations).  Using the bounds in this paper, we can compute when we do not need to recalculate impostors:

http://www.ratml.org/misc/lmnn_bounds.pdf

Essentially to implement this, we will have to do the following:

 * Perform norm-centering before the entire LMNN computation to reduce the average norm of points (and make the bound more effective).  Likely that's a good step to take anyway that can help resultant kNN accuracy.
 * Cache the transformation matrix each time `Impostors()` is called.
 * At each iteration's call to `Impostors()`, compute the Frobenius norm of the difference between the current transformation matrix and the cached transformation matrix.
 * If the Frobenius norm is below the bound, don't calculate impostors!

Note that the bound in the paper can apply for each point individually, so it is possible to do the following:

 * In the small-batch setting (`Evaluate(..., begin, batchSize)`), use single-tree search to only search for query points where the bound requires recomputation of impostors.
 * In the full-batch setting, if query trees are being cached, the computation of a new and smaller query tree would take time.  Therefore, only build a custom smaller query tree if, e.g., 30% or less of query points need impostors recalculated.

manish7294:
I tried these bound by creating an overload of Impostors() to calculate impostors for some specific points of the dataset which violates the bounds (there is no tree caching as of now). 
The results were pretty good but are dataset dependent, like on iris, vc2 and balance_scale the new runtimes are around 65 - 70% of the previous runtimes whereas on letters and covertype-5k, the runtimes are almost same.

rcurtin:
Right, I would suspect that the speedup here would be dataset-dependent.  Would you like to open a PR with what you've got implemented?

manish7294:
Sure, I will open one shortly.

rcurtin:
This is done since #1466 is merged.  @manish7294: great work! :+1:

PR #1466, opened 2018-07-09
LMNN: Bounds for avoiding Impostors recalculation

manish7294:
see #1448


src/mlpack/methods/lmnn/lmnn_function_impl.hpp
```diff
@@ -82,10 +82,33 @@ double LMNNFunction<MetricType>::Evaluate(const arma::mat& transformation)
...
+    for (size_t i = 0; i < dataset.n_cols; i++)
+    {
+      if (transformationDiff * (2 * norm(i) + norm(impostors(k - 1, i)) +
+          norm(impostors(k, i))) > distance(k, i) -  distance(k - 1, i))
+      {
+        tempPoints.push_back(i);
+      }
+    }
+
+    arma::uvec points = arma::conv_to<arma::uvec>::from(tempPoints);
```
rcurtin: Another potential opportunity for speedup would be to allocate a "large-enough" array here, which could be done in a couple ways (I am not sure which would be better, feel free to try each!):

1. Allocate `points` to have length equal to `dataset.n_cols`, then simply fill it and track how full it is.  Then have `Impostors()` also take a parameter indicating how many elements of `points` should be used.

2. Allocate `points` as above, but with length `dataset.n_cols * 0.75` (or some tunable parameter, I am not sure which will be best).  If, during the pruning check process, `points` becomes full, then simply shortcut and run `Impostors()` on everything.

3. Take a two-pass approach, where in the first pass you compute whether or not the point is pruned, then allocate the array with the correct size.  In the second pass, fill the elements.  I'm hesitant to say this will be faster here, but it could be.  The `conv_to<>::from()` operation could take a little time.

rcurtin:
Ah, I see also that this (like others) has a potential problem with the storage of the old transformation matrix for different batches.  Let's talk about that specifically in #1461 so we (or more specifically I) don't get confused during the discussion because it is in multiple places. :)

manish7294:
Here are some simulations. https://gist.github.com/manish7294/b1efd856ba3013803be29801bfa4629e
I have used mean-centering here. Have a doubt about the correctness of it tough:
```
for (size_t i = 0; i < data.n_rows; i++)
  {
    data.row(i) =  data.row(i) - (double)arma::mean(data.row(i));
  }
```
It seems points change too much on covertype.

rcurtin:
I believe that's correct, looking at the implementation.  If you really want to quickly check if it is correct, just ensure that the starting kNN accuracy with and without mean-centering is the same (kNN should be invariant to mean-centering).  The real check would be if the exact kNN results including distances are the same, so you could do that too.

The results look great.  I can see that in the later iterations for vc2 and iris, very few impostors are recalculated.  I'm not too surprised about the covertype dataset---I've found that the points have very large norm and so as a result the bounds don't seem to be as effective there.  Still the mean centering should help.

When #1461 is done let's return to this and work out the batch size issue, do some more timing simulations to convince ourselves the speedup is good, and then merge it in. :+1:

manish7294:
Just have to figure out one issue - we are obliged to calculate `k + 1` target neighbors and impostors instead of k. Now, let say we have only 6 points in one of the class. Ideally, we should be able to run simulations up to k = 5, but now we are restricted to up to k = 4. That's the reason one of the unit test is failing. 
I think If we see this problem from the point of real datasets, we generally don't have this low number of points in any class and shouldn't cause much of a problem. We may just output an error message to the user if this is happening with his/her parameters. What do you think?

rcurtin:
Right, that won't typically happen in real-world datasets, but if it does you can do this... search for `k` neighbors, resize the distances matrix to have `k + 1` rows, then fill that last row with `DBL_MAX`.  Then the bounds will still work correctly.

manish7294:
Ya, that would solve the problem, just an extra check like this inside LMNNFunction() will be required

```
  size_t minCount = arma::min(arma::histc(labels, arma::unique(labels)));
  if (minCount < k + 1)
  {
    // use DBL_MAX
  }
```

rcurtin:
Couldn't you just check the size of each of the training sets right before the `Search()` call instead of calling `histc()`?

manish7294:
@rcurtin I think it's now ready for review.

manish7294:
@rcurtin It looks like when we calculate `targetNeighbors` for `k + 1`, then the result of first k rows is a bit different than when we calculate `targetNeighbors` for `k`. Ideally, the first k rows of result from `k + 1` targetNeighbors calculation should be the same as the result when `k` targetNeighbors are calculated. Similarly, `Impostors` faces the same issue.

To cross-check this thing, I ran `mlpack_knn` twice, one for `k = 3` and one for ` k = 4`, and it seems entries of some of the columns gets a kind of shuffling / re-ordering effect on changing `k`.

rcurtin:
Can you file an issue for the KNN bug so that I can try and reproduce it?  If the distances are exactly the same, the results may come back in a slightly different order, but the distance should always be exactly the same.

manish7294:
Right, I am filing a bug issue.

src/mlpack/methods/lmnn/lmnn_function_impl.hpp
```diff
@@ -191,17 +207,44 @@ double LMNNFunction<MetricType>::Evaluate(const arma::mat& transformation)
...
+      // Re-calculate impostors on transformed dataset.
+      constraint.Impostors(impostors, distance, transformedDataset, labels);
+    }
+  }
+  else if (iteration++ % range == 0)
   {
-    // Re-calculate impostors on transformed dataset.
-    constraint.Impostors(impostors, distance, transformedDataset, labels);
+      // Re-calculate impostors on transformed dataset.
+      constraint.Impostors(impostors, distance, transformedDataset, labels);
```
rcurtin: I think the extra indent wasn't necessary here.

src/mlpack/methods/lmnn/lmnn_function_impl.hpp
```diff
@@ -395,17 +468,44 @@ void LMNNFunction<MetricType>::Gradient(const arma::mat& transformation,
   // Apply metric over dataset.
   transformedDataset = transformation * dataset;
 
-  // Calculate norm of change in transformation.
   double transformationDiff = 0;
-  if (!transformationOld.is_empty())
+  if (!transformationOld.is_empty() && iteration++ % range == 0)
```
rcurtin: `transformationDiff` is used every iteration, so shouldn't we avoid adding the `&& iteration++ % range == 0` here?

manish7294:
One of the Accuracy tests over 6 points simple dataset using amsgrad is failing somehow, and it's weird that I am not getting that on my system. Debbuging this led me #1470 earlier, as there is some deviation in final results because of that. Other than that, I don't think I am still up to anything else.

manish7294:
I performed some simulations and I think having a dataset having different distances to all neighbors solves the issue. Even vc2 works perfectly.

manish7294:
I think the deviation could be expected as we are actually training over a different permutation of points. Here a point may be such that it has a greater impact over some other dimension than the other point, and still having same distance from the query point.

rcurtin:
I see.  Give me some time to think about it, and you can think about it also, and let's see if we can figure out a way to handle it.  The answer may be requiring that NearestNeighborSearch gives us back deterministic orderings of points, or the answer may be that we specifically look through our kNN results and reorder neighbor indices when distances are the same.

manish7294:
On a second thought, Though being a bit different, the distance learned will still be correct, as we are just taking a different set of constraints which satisfy the conditions of LMNN.

src/mlpack/methods/lmnn/lmnn_function_impl.hpp
```diff
@@ -191,17 +207,44 @@ double LMNNFunction<MetricType>::Evaluate(const arma::mat& transformation)
...
+      for (size_t i = 0; i < dataset.n_cols; i++)
+      {
+        if (transformationDiff * (2 * norm(i) + norm(impostors(k - 1, i)) +
+            norm(impostors(k, i))) > distance(k, i) -  distance(k - 1, i))
+        {
+          points(numPoints++) = i;
+        }
+      }
+
+      points.resize(numPoints);
```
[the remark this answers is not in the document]
manish7294: Done!

src/mlpack/methods/lmnn/lmnn_function_impl.hpp
manish7294: Done!

src/mlpack/methods/lmnn/lmnn_function_impl.hpp
manish7294: Done!

rcurtin:
Unfortunately I've done a little bit of thinking about the exact-same-distance issue and I've come to two conclusions:

1.  Right now all bounds we are computing are with respect to the k'th impostor.  In essence these bounds are saying, "the distance between the point and its k'th impostor next iteration cannot be greater than (our bound quantity)".  But remember that this depends on the norm of the impostor.  So what if the (k - 1)'th impostor has huge norm but the k'th impostor has very small norm?  In this case it is possible that the (k - 1)'th impostor will be further away next iteration than the k'th impostor.  That means that, with respect to our bounds, it's possible that our current bounds could say "it's not possible that the (k + 1)'th impostor and the k'th impostor could switch places in the distance order" and be correct---but at the exact same time it could be possible that the (k - 1)'th and (k + 1)'th impostor would switch places.  (Or something like this.)  At the end of my long explanation here, it's clear that the only solution is that we have to rework the bounds a little, to take the max over all `k` possible bounds.  (Also note that for `k == 1`, there is no problem, and I think the bounds being incorrect would be very rare in practice.)  Let's worry about this issue in another PR or something since I don't believe it is causing any of our issues.

2.  The ordering of the results does matter because it will affect the gradient direction.  If the impostor ordering is switching every iteration, this will cause trouble for the gradient (I believe).  In any case, the order in which we want our results to be is in order of increasing norm.  So, e.g., if the k'th and (k + 1)'th impostor have the same distance to the query point, then we'd prefer to have it in the order that the k'th impostor has smaller norm than the (k + 1)'th impostor.  But since this is problem-specific to our LMNN optimizations, it doesn't make sense to me to do this computation in the `NeighborSearch` code.  So I believe the best way to do this will be to update `Impostors()` so that it runs a loop after calling `Search()` to ensure that the ordering is correct.  I worked out what I think is some code that will do this, but I did not have a chance to fully integrate this into the different `Constraints::Impostors()` overloads.  On the laptop I'm using I don't actually have a way to middle-click, so I can't copy-paste it, but I'll email the function to you separately.

manish7294:
> Right now all bounds we are computing are with respect to the k'th impostor. In essence these bounds are saying, "the distance between the point and its k'th impostor next iteration cannot be greater than (our bound quantity)". But remember that this depends on the norm of the impostor. So what if the (k - 1)'th impostor has huge norm but the k'th impostor has very small norm? In this case it is possible that the (k - 1)'th impostor will be further away next iteration than the k'th impostor. That means that, with respect to our bounds, it's possible that our current bounds could say "it's not possible that the (k + 1)'th impostor and the k'th impostor could switch places in the distance order" and be correct---but at the exact same time it could be possible that the (k - 1)'th and (k + 1)'th impostor would switch places. (Or something like this.) At the end of my long explanation here, it's clear that the only solution is that we have to rework the bounds a little, to take the max over all k possible bounds. (Also note that for k == 1, there is no problem, and I think the bounds being incorrect would be very rare in practice.) Let's worry about this issue in another PR or something since I don't believe it is causing any of our issues.

Agreed, this definitely is something that had been overlooked. It seems doing this will be a bit tricky.

> The ordering of the results does matter because it will affect the gradient direction. If the impostor ordering is switching every iteration, this will cause trouble for the gradient (I believe). In any case, the order in which we want our results to be is in order of increasing norm. So, e.g., if the k'th and (k + 1)'th impostor have the same distance to the query point, then we'd prefer to have it in the order that the k'th impostor has smaller norm than the (k + 1)'th impostor. But since this is problem-specific to our LMNN optimizations, it doesn't make sense to me to do this computation in the NeighborSearch code. So I believe the best way to do this will be to update Impostors() so that it runs a loop after calling Search() to ensure that the ordering is correct. I worked out what I think is some code that will do this, but I did not have a chance to fully integrate this into the different Constraints::Impostors() overloads. On the laptop I'm using I don't actually have a way to middle-click, so I can't copy-paste it, but I'll email the function to you separately.

Right, that will be the best option compared to making changes to Neighbor Search. I saw the code, and I think good to me. I think computation cost won't be an issue as compared to other parts of LMNN.

rcurtin:
I think everything is working here but the random test failure is due to #1469 not being merged yet.  Could you run the same simulations we've run on other PRs to see the runtimes with and without this optimization?  (It would be useful to calculate the accuracy too because it is easy to do, just as a sanity check to make sure nothing is wrong.  Ideally it should be the same.)

rcurtin:
It's also not clear to me why the build should take any longer than before---usually the build/install step takes roughly 2850s but in the Travis build here it's more like 3200, so the build runs out of time.  I think that too is random and we don't need to worry about it.  There's nothing in this PR that would cause an additional 350s of build time.

manish7294:
I guess I am not in position to make a comment over train runtime as I haven't debugged much here. But intiutively I think this may be due to bbsgd getting stuck, as I have seen it doing so earlier as well, though it doesn't make any sense. Probably #1469 will solve this.

rcurtin:
Sorry, I'm not sure I understand what your comment is referring to. Can you clarify?

rcurtin:
No worries, thanks for the clarification. I turn autocorrect off but then I just get lots of spelling errors. :)

I don't think we should worry about the Travis build time---I think we are just unlucky.  I'll try the build again a few times tomorrow and see if that helps; otherwise I (or you) could just time how long it takes to build the master branch and also this branch and make sure there is no significant difference.

manish7294:
I think I may have a reason for Travis timeout.

In BBSGD we update iterate using below lines - 
https://github.com/mlpack/mlpack/blob/455b00973168c0ac2403befc6858546aa83c63d2/src/mlpack/core/optimizers/bigbatch_sgd/adaptive_stepsize.hpp#L89-L90

but in case we have the `gradient` as a zero matrix, `iterate` will never be updated and same will continue forever. So basically, we will be calling evaluate on the same `iterate` matrix again & again.

Does this sound reasonable?

rcurtin:
The timeout does not seem to have to do with the testing phase.  The Travis build is split into a build phase---which usually takes ~2800 seconds, and a test phase.  But for this PR the Travis build step is taking ~3200-3500 seconds.  That time does not include the time to run the tests.  Here's an example successful build, where the build step for the first configuration takes only 2820 seconds:

https://travis-ci.org/mlpack/mlpack/builds/407688072

So I don't believe the issue is in the tests.  I believe it is some strange Travis artifact and I don't see any reason that that behavior should continue after the merge, so I would be happy to merge it in spite of these strange results.

Next, about BBSGD: you are right that the situation could occur that the gradient matrix is identically zero and in this case no step will be taken.  However, remember that BBSGD works on batches of points---so the next batch is likely to have a different gradient that will not be zero, and so a step will be taken.  Also, note that BBSGD does have a tolerance so if the gradient becomes zero, it will terminate at the end of the epoch.

So I don't expect that we are seeing any issues as a result of zero gradients in BBSGD.  If you've noticed cases where that indeed does happen, then let's discuss it, but I don't think it's happening here.

Anyway, that aside:

> Could you run the same simulations we've run on other PRs to see the runtimes with and without this optimization? (It would be useful to calculate the accuracy too because it is easy to do, just as a sanity check to make sure nothing is wrong. Ideally it should be the same.)

Have you had a chance to run these numbers yet? :)

manish7294:
Here are some simulations - https://gist.github.com/manish7294/01c192dee660bb09baf27841c84c9b70
Sorry for the delay.

manish7294:
Do these results seem meaningful to you?

rcurtin:
I am not sure, I think you have only run it for one optimizer and it's important to run it for all three because each `Evaluate()` and `EvaluateWithGradient()` use these bounds differently.  I was thinking that I would simply devote some time and run some experiments myself to try and understand why there is no speedup in the covertype and letters cases.

rcurtin:
Just an update, I have not collated my results entirely yet but I have seen promising results on large covertype datasets and MNIST:

```
covertype-5k lbfgs: 66 iterations

        total     computing_neighbors     acc before      acc after
old:    80.758s   21.672s                 75.56%          79.98%
new:    75.456s   15.992s                 75.56%          79.98%

covertype-5k amsgrad: 50 passes, batch size 256

        total     computing_neighbors     acc before      acc after
old:    136.093s  16.157s                 75.56%          75.54%
new:    137.484s  16.813s                 75.56%          77.26%


        total     computing_neighbors     building    acc before      acc after
old:    249.279s  128.153s                5.587s      90.11%          91.796%
new:    620.516s  340.776s                ~100s       90.11%          91.862%
```

The first MNIST run, on the 10k test points (784d) with L-BFGS and 25 maximum iterations, took ~3670s before this optimization and ~3697s afterwards.  So I guess that there is not much optimization possible here, but I want to investigate that more.  Alternately, if you would like to determine whether or not any impostors are being pruned with the MNIST dataset I'd be interested to see the results.

Overall I'm fairly convinced this is good and we should merge it, but I just want to make sure and see a few more results (mostly with large datasets) first.  I will be busy this weekend though, so I won't really have a chance to run anything, so if you're able to get to it before me great, otherwise feel free to wait.

rcurtin:
Note also, I believe the accuracy difference in some cases is due to our slight changes in the gradient calculation.

manish7294:
Right, the results seem good enough. Sure, I will perform some simulation for pruning like you did earlier. Can I still use slack?

rcurtin:
I think that should be fine, I am not sure what the long-term status of that system will be, I have not heard back yet.  But it is up for now.

manish7294:
@rcurtin  Here are some simulations mainly focusing over full covertype and mnist test(10k) dataset - https://gist.github.com/manish7294/f1f478d06ea91c91a42f059e3fabde42
Sorry, but the actual iteration number is off by one, I added the print statement after the iteration update. We can see the effects of impostors bounds on covertype starting from iteration 40, where as I don't think mnist is going to react anytime sooner as after even 50 iterations `transformationDiff` is still quite big.

manish7294:
This time I used mean centering. But it doesn't look like pruning happened at least for first 50 passes. Just to make sure everything is fine, I ran a simulation on iris as well for which pruning started from very early stages. https://gist.github.com/manish7294/baecea1fcf0553f73dfe7c67d1eb6e58

PR #1467, opened 2018-07-11
Add EvaluateWithGradient To FFN and RNN

ShikharJ:


src/mlpack/methods/ann/ffn_impl.hpp
```diff
@@ -225,6 +225,89 @@ double FFN<OutputLayerType, InitializationRuleType, CustomLayers...>::Evaluate(
...
+  }
+
+  outputLayer.Backward(
+      std::move(boost::apply_visitor(outputParameterVisitor, network.back())),
+      std::move(responses.cols(begin, begin + batchSize - 1)),
+      std::move(error));
+
+  Backward();
+  ResetGradients(gradient);
+  Gradient(std::move(predictors.cols(begin, begin + batchSize - 1)));
```
zoq: https://github.com/mlpack/mlpack/blob/4c19146cb7b0566226fb16dc49051af664c74b79/src/mlpack/methods/ann/ffn_impl.hpp#L248

This call in the ``Gradient(..)`` method can be avoided, since it's what we already did in line 277-285. So, I guess we can just implement our own Gradient function here, without the extra ``Evaluate``call, let me know what you think.

src/mlpack/methods/ann/ffn_impl.hpp
```diff
@@ -225,6 +225,89 @@ double FFN<OutputLayerType, InitializationRuleType, CustomLayers...>::Evaluate(
...
+    res += EvaluateWithGradient(parameters, i, gradient, 1, false);
+
+  return res;
+}
+
+template<typename OutputLayerType, typename InitializationRuleType,
+         typename... CustomLayers>
+template<typename GradType>
+double FFN<OutputLayerType, InitializationRuleType, CustomLayers...>::
+EvaluateWithGradient(const arma::mat& /* parameters */,
```
zoq: Do you think the existing ``Gradient`` method could use the ``EvaluateWithGradient`` as well?

ShikharJ: @zoq It could, but there would be practically no speedup, as the same functions would be required to get called.

zoq: Right, but we could remove some code duplication.

ShikharJ:
@zoq Thanks for the comments! This is ready for review.

src/mlpack/methods/ann/ffn_impl.hpp
```diff
@@ -227,11 +227,34 @@ double FFN<OutputLayerType, InitializationRuleType, CustomLayers...>::Evaluate(
...
+
+template<typename OutputLayerType, typename InitializationRuleType,
+         typename... CustomLayers>
+template<typename GradType>
+double FFN<OutputLayerType, InitializationRuleType, CustomLayers...>::
+EvaluateWithGradient(const arma::mat& parameters, GradType& gradient)
+{
+  double res = 0;
+  for (size_t i = 0; i < predictors.n_cols; ++i)
+    res += EvaluateWithGradient(parameters, i, gradient, 1, false);
```
rcurtin: Almost certainly a larger batch size will help this be faster.  I'd suggest a default of 64, 128, or 256 (those seem to be common values).

src/mlpack/methods/ann/ffn_impl.hpp
```diff
@@ -227,11 +227,34 @@ double FFN<OutputLayerType, InitializationRuleType, CustomLayers...>::Evaluate(
...
+
+template<typename OutputLayerType, typename InitializationRuleType,
+         typename... CustomLayers>
+template<typename GradType>
+double FFN<OutputLayerType, InitializationRuleType, CustomLayers...>::
+EvaluateWithGradient(const arma::mat& /* parameters */,
+                     const size_t begin,
+                     GradType& gradient,
+                     const size_t batchSize,
+                     const bool deterministic)
```
rcurtin: Do we (or will we) ever call with `deterministic = true`?  I think that when `determinstic = true`, we are just evaluating the network so there is only a need for a forward pass, in which case we'd call `Evaluate()` instead.  So could the parameter be removed?  It's possible there's something I've overlooked here, so let me know if so. :)

src/mlpack/methods/ann/ffn_impl.hpp
ShikharJ: This default seems to originate from the time when only stochastic models were there. I'll choose the lowest default of 64, as 256 might be too much.

src/mlpack/methods/ann/ffn_impl.hpp
ShikharJ: @rcurtin Even I am not too sure of this, I think @zoq might be able to help with this?

ShikharJ:
@rcurtin I'll find what the duration per call of individual functions is. That should give us a good idea of the speedup to be expected.

src/mlpack/methods/ann/ffn_impl.hpp
zoq: Correct, there is no need for ``deterministic`` in ``EvaluateWithGradient``.

src/mlpack/methods/ann/ffn_impl.hpp
ShikharJ: @rcurtin On another note, I'm not too sure if this should be changed or not. Many old tests would have been implemented keeping stochastic training in mind, and we anyways provide the option for `batchSize` now. So I don't see a point meddling with the old tests.

rcurtin: Shouldn't the result be exactly the same?  Or do you mean that there are models for which simply the computation of the gradient or objective function changes the parameters of the model?

ShikharJ: I just want this to follow from the `Evaluate()` implementation which does a stochastic evaluation only. Choosing another default would be a break from that design.

ShikharJ:
@zoq A final review?

src/mlpack/methods/ann/gan/wgan_impl.hpp
```diff
@@ -111,38 +111,20 @@ EvaluateWithGradient(const arma::mat& /* parameters */,
...
-  discriminator.Forward(std::move(currentInput));
-  double res = discriminator.outputLayer.Forward(
-      std::move(boost::apply_visitor(
-      outputParameterVisitor,
-      discriminator.network.back())), std::move(currentTarget));
-
   // Get the gradients of the Discriminator.
-  discriminator.Gradient(discriminator.parameter, i, gradientDiscriminator,
-      batchSize);
+  double res = discriminator.EvaluateWithGradient(discriminator.parameter,
```
zoq: Nice, that makes the file even shorter.

zoq:
Thanks, really great contribution.

PR #1468, opened 2018-07-17
Overload Forward() function.

akhandait:
This allows users to forward pass partially through the network.

### DIFF
--- a/src/mlpack/methods/ann/ffn.hpp
+++ b/src/mlpack/methods/ann/ffn.hpp
@@ -285,6 +285,22 @@ class FFN
   void Forward(arma::mat inputs, arma::mat& results);
 
   /**
+   * Perform a partial forward pass of the data.
+   *
+   * This function is meant for the cases when users require a forward pass only
+   * through certain layers and not the entire network.
+   *
+   * @param inputs The input data for the specified first layer.
+   * @param results The predicted results from the specified last layer.
+   * @param begin The index of the first layer.
+   * @param end The index of the last layer.
+   */
+  void Forward(arma::mat inputs,
+               arma::mat& results,
+               const size_t begin,
+               const size_t end);
+
+  /**
    * Perform the backward pass of the data in real batch mode.
    *
    * Forward and Backward should be used as a pair, and they are designed mainly
--- a/src/mlpack/tests/feedforward_network_test.cpp
+++ b/src/mlpack/tests/feedforward_network_test.cpp
@@ -629,19 +629,19 @@ BOOST_AUTO_TEST_CASE(PartialForwardTest)
   model.Add<Linear<> >(5, 10);
 
   // Add a new Add<> module which adds a constant term to the input.
-  Add<>* AddModule = new Add<>(10);
-  model.Add(AddModule);
+  Add<>* addModule = new Add<>(10);
+  model.Add(addModule);
 
-  LinearNoBias<>* LinearNoBiasModule = new LinearNoBias<>(10, 10);
-  model.Add(LinearNoBiasModule);
+  LinearNoBias<>* linearNoBiasModule = new LinearNoBias<>(10, 10);
+  model.Add(linearNoBiasModule);
 
   model.Add<Linear<> >(10, 10);
 
   model.ResetParameters();
   // Set the parameters of the Add<> module to a matrix of ones.
-  AddModule->Parameters() = arma::ones(10, 1);
+  addModule->Parameters() = arma::ones(10, 1);
   // Set the parameters of the LinearNoBias<> module to a matrix of ones.
-  LinearNoBiasModule->Parameters() = arma::ones(10, 10);
+  linearNoBiasModule->Parameters() = arma::ones(10, 10);
 
   arma::mat input = arma::ones(10, 1);
   arma::mat output;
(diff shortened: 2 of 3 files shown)


src/mlpack/methods/ann/ffn.hpp
```diff
@@ -284,6 +284,19 @@ class FFN
...
+   *
+   * This function is meant for the cases when users require a forward pass only
+   * through certain layers and not the entire network.
+   *
+   * @param inputs The input data for the specified first layer.
+   * @param results The predicted results from the specified last layer.
+   * @param begin The index of the first layer.
+   * @param end The index of the last layer.
+   */
+  void Forward(arma::mat inputs, arma::mat& results, size_t begin, size_t end);
```
zoq: Can we use ``const size_t``, to be consistent with the rest of the codebase?

zoq:
Ah, good point, yeah that would be tremendously helpful.

akhandait:
@rcurtin I think that's a good idea. I will add a test in `FeedForwardNetworkTest`, or should I add it in `ANNLayerTest`?

rcurtin:
No preference from my end, I think anywhere is fine. Thanks! :)

akhandait:
@rcurtin I understand, I will add a better test for this soon. I see that both of you have approved, but I think we should merge this after I add a better test.

src/mlpack/tests/feedforward_network_test.cpp
```diff
@@ -620,4 +620,47 @@ BOOST_AUTO_TEST_CASE(CustomLayerTest)
...
+
+  // As we only forward pass through Add module, input and output should
+  // differ by a matrix of ones.
+  CheckMatrices(input, output - 1);
+
+  // Forward pass only through the Add module and the Identity layer.
+  model.Forward(input,
+                output,
+                1 /* Index of the Add module */,
+                2 /* Index of the Identity layer */);
```
zoq: Do you mind to add another test, that includes the linear layer as well into the output, we can set the weights to ones as well; what I'd like to see is that the output changed. Let's suppose the end index doesn't work as expected the output is still the same as the first one since the second layer is an Identity layer.

akhandait:
@rcurtin What is the comment on memory usage?

src/mlpack/tests/feedforward_network_test.cpp
```diff
@@ -620,4 +620,51 @@ BOOST_AUTO_TEST_CASE(CustomLayerTest)
...
+{
+  FFN<NegativeLogLikelihood<>, RandomInitialization> model;
+  model.Add<Linear<> >(5, 10);
+
+  // Add a new Add<> module which adds a constant term to the input.
+  Add<>* AddModule = new Add<>(10);
+  model.Add(AddModule);
+
+  LinearNoBias<>* LinearNoBiasModule = new LinearNoBias<>(10, 10);
+  model.Add(LinearNoBiasModule);
```
rcurtin: Weird, the comment didn't seem to post.  I wonder what I did wrong.

Anyway: whenever you use `new`, you have to also use `delete` otherwise there is a memory leak.  So at the end of this test you'd need to add that.  But I would say, in this case, you can actually do something a little easier and avoid calling `new` at all:

```
Add<> addModule(10);
model.Add(addModule);
```

(Also note the lowercase variables, please do that to match the style guide. :))

Sorry this didn't post earlier!

akhandait:
@rcurtin I don't it's possible here to do this
```cpp
Add<> addModule(10);
model.Add(addModule);
```
Add() expects a pointer and throws a error if we try the above.

zoq:
Did you test ``model.Add<Add<> >(addModule);`` as well?

rcurtin:
I opened #1489 to discuss that issue, so I think it's fine to leave it as a layer created with `new` for now.

src/mlpack/tests/feedforward_network_test.cpp
```diff
@@ -620,4 +620,55 @@ BOOST_AUTO_TEST_CASE(CustomLayerTest)
   arma::mat prediction = arma::zeros<arma::mat>(1, predictionTemp.n_cols);
 }
 
+/**
+ * Test the overload of Forward function which allows partial forward pass.
+ */
+BOOST_AUTO_TEST_CASE(PartialForwardTest)
+{
+  FFN<NegativeLogLikelihood<>, RandomInitialization>* model =
+      new FFN<NegativeLogLikelihood<>, RandomInitialization>();
```
rcurtin: Ok, now I understand what's happened here.  Changing `model` to be a manually-allocated pointer to an `FFN` isn't necessary, my concern originally was that I saw that `addModule` and `linearNoBiasModule` were being allocated without any `delete` later.  So I thought that the memory that was being allocated would never be freed.

However after digging into it, the particular way this code currently works (see #1489), `model` will actually delete `addModule` and `linearNoBiasModule` when its destructor is called.  That destructor gets called regardless of whether `model` is allocated with `new` or not---when using `new`, the destructor is called when we call `delete`.  But if we use a local variable for `model`, then the destructor is called when it goes out of scope (at the end of the function) and the memory is freed there too.

So you can revert to a local `FFN` for `model` if you like, but this code works either way and doesn't leak right now.  Depending on what we do for #1489, this may need to be adjusted later, but I don't think we need to worry about it.

tl;dr: I was incorrect about the memory leak, sorry about that. :)

akhandait: Okay, I guess it's okay for me to revert to the second last commit then. Yeah, even I forgot that the FFN class deletes everything.

zoq:
Let me restart the travis build, just to see the test pass.

zoq:
Thanks again for the great work!

PR #1469, opened 2018-07-19
Resolve random failing of LMNN Test

manish7294:
It seems LMNNOptimizerTest from main tests is randomly failing. This PR is for dealing with this issue.

### DIFF
--- a/src/mlpack/core/optimizers/bigbatch_sgd/adaptive_stepsize.hpp
+++ b/src/mlpack/core/optimizers/bigbatch_sgd/adaptive_stepsize.hpp
@@ -90,7 +90,7 @@ class AdaptiveStepsize
     iterate -= stepSize * gradient;
 
     // TODO: Develop an absolute strategy to deal with stepSizeDecay updates in
-    // case we reaches arrive at local minima. See #1469 for more details.
+    // case we arrive at local minima. See #1469 for more details.
     double stepSizeDecay = 0;
     if (gradientNorm && sampleVariance && batchSize)
     {
--- a/src/mlpack/core/optimizers/bigbatch_sgd/adaptive_stepsize.hpp
+++ b/src/mlpack/core/optimizers/bigbatch_sgd/adaptive_stepsize.hpp
@@ -89,15 +89,20 @@ class AdaptiveStepsize
     // Update the iterate.
     iterate -= stepSize * gradient;
 
+    // TODO: Develop an absolute strategy to deal with stepSizeDecay updates in
+    // case we reaches arrive at local minima. See #1469 for more details.
     double stepSizeDecay = 0;
-    if (batchSize < function.NumFunctions())
+    if (gradientNorm && sampleVariance && batchSize)
     {
-      stepSizeDecay = (1 - (1 / ((double) batchSize - 1) * sampleVariance) /
-          (batchSize * gradientNorm)) / batchSize;
-    }
-    else
-    {
-      stepSizeDecay = 1 / function.NumFunctions();
+      if (batchSize < function.NumFunctions())
+      {
+        stepSizeDecay = (1 - (1 / ((double) batchSize - 1) * sampleVariance) /
+            (batchSize * gradientNorm)) / batchSize;
+      }
+      else
+      {
+        stepSizeDecay = 1 / function.NumFunctions();
+      }
     }
 
     // Stepsize smoothing.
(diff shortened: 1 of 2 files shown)


manish7294:
The issue in not fully debugged yet but it seems, it has to do something with shuffle() as it's only visible if  linear_scan is set to true.

rcurtin:
I agree that the change you've made will solve a problem, but I'm not sure about the BBSGD issue.  It's worth noting that the BBSGD optimizer is the only one that you have been using that uses the `Evaluate()` and `Gradient()` functions instead of `EvaluateWithGradient()`.

One way you could debug this more is to try and use gdb conditional breakpoints:
https://www.fayewilliams.com/2011/07/13/gdb-conditional-breakpoints/
So, for instance, you could set a breakpoint at the end of `Evaluate()` and then your condition would just be something like `if cost != cost` (remember NaN is not equal to itself).  Or possibly you could check for when any given `eval` was NaN, and then inspect all of the local variables to see if you can figure out what is going on.

(Alternately it's possible to avoid gdb and simply do a lot of printing to debug this.  I think either way is valid.)

rcurtin:
Also, it's worth considering the possibility that some element of `transformation` is NaN, which would mean probably that some part of the gradient being returned in an iteration is NaN---or very large.  Depending on the step size it is possible that BBSGD is simply diverging and needs a smaller step size.  So maybe that is a better first thing to check.

manish7294:
I think I have pinpointed the origin of error : https://github.com/mlpack/mlpack/blob/3b7bbf0f14172cdb00fd16cbf12918b07c888b96/src/mlpack/core/optimizers/bigbatch_sgd/adaptive_stepsize.hpp#L95-L96
Here in case `sampleVariance` or `gradientNorm` is 0 then stepSizeDecay will become `nan`.

I clearly don't know whether ideally gradientNorm can become 0 or not. Please tell me your views on this?

rcurtin:
Nice work digging in. :+1:

In general if the gradient norm is 0, then we are at a local minimum.  But if we are only operating on a batch of points, then we are at a local minimum _only for those points in the batch_, not in general.  So it's not clear to me what the best choice is here.

@zoq: what do you think in this situation?  My quick intuition is, maybe in this case we do not decay the step size at all.  Also don't the two `batchSize`s on the second line cancel out?  In this case I guess the result should be that `stepSize = 0` after the update, but that is probably a bad idea to do.

zoq:
You are right ``batchSize`` cancels out ``batchSize`` in ``(batchSize * gradientNorm)``. Skiping the decay step sounds reasonable to me; still have to think about the issue.

rcurtin:
Ok, @manish7294, if you'd like to modify the BBSGD optimizer to skip the decay step on iterations like this, I think that is fine for now.

manish7294:
So, shall we be keeping `stepSizeDecay` to 0 itself for iterations having `sampleVariance` or `gradientNorm` as 0, or do we just bypass the `if` condition and move to `else` condition to calculate `stepSizeDecay` on the basis of `NumFunctions`.

https://github.com/mlpack/mlpack/blob/3b7bbf0f14172cdb00fd16cbf12918b07c888b96/src/mlpack/core/optimizers/bigbatch_sgd/adaptive_stepsize.hpp#L93-L101

manish7294:
@rcurtin Do you think we shall merge this now. It will at least fix the failing build. Though, a better solution need to developed. I have added a comment regarding that.

manish7294:
I just have a quick look over the bigbatchSGD paper and I think that the last `batchSize` which was creating the cancellation issue is probably `function.NumFunctions()`

https://github.com/mlpack/mlpack/blob/3b7bbf0f14172cdb00fd16cbf12918b07c888b96/src/mlpack/core/optimizers/bigbatch_sgd/adaptive_stepsize.hpp#L96

Here's the reference - https://arxiv.org/pdf/1610.05792.pdf page 10,  **Algorithm 3 Big batch SGD: with BB stepsizes**, Line 14 - 18

zoq:
I'm not following, can you clarify:

> I think that the last batchSize which was creating the cancellation issue is probably function.NumFunctions()

manish7294:
@zoq  I was referring to this statement of yours and Ryan over the computation of `stepSizeDecay`

> You are right batchSize cancels out batchSize in (batchSize * gradientNorm).

Please let me know if this helps clarifying anything or you were expecting something else.

zoq:
I see thanks for the clarification.

rcurtin:
Thanks! :+1:

Issue #1445, opened 2018-06-27
LMNN: optimization of kNN computation by tree caching


rcurtin:
This is a follow-up to the LMNN support added in #1407.  One of the biggest parts of the computation we are performing is the repeated calculation of the impostors with a kNN search.  Looking at current benchmarks, the algorithm spends half or more of its time performing these searches.  Depending on the dataset, a lot of this time can be the tree building process (which at the current moment has no timer running for it when the reference tree is built). But the points in the tree don't change, although they are stretched.  So if we can cache the tree structure between iterations, we can completely save that time.

As a side note, for the SGD-like optimizers, we run kNN with small batches as the query set.  It's not worthwhile to cache the trees built on these batches, so let's not worry about that.  But it is worthwhile to cache the reference trees---which in the case of an L_BFGS-like optimizer uses the reference sets also as query sets.

However, the thing that makes caching these trees tricky is that each iteration, we change the transformation matrix L.  This means that all the bounds and quantities in our tree are wrong.  However, we could update all the quantities inside the tree according to the new transformation matrix L, and then we would be able to use it for kNN search in the new space correctly.  Over time, if L changes too much, the kNN search will get slower and slower (the trees will be able to prune less and less), but I am not sure if this will be enough of a problem that we'll need to care about it.  We can figure that out during simulations.

It would probably be worthwhile to look through the KD-tree code (which is built with the `BinarySpaceTree` class) in `src/mlpack/core/tree/binary_space_tree/` to see how these trees are built.  I will give a decent description here though that I hope will be helpful.  Given a dataset, we repeatedly split it on the dimension of maximum variance, in order to organize our points.  Each node in the tree, then, contains points that are localized in space.  For the `BinarySpaceTree` structure, there are a couple of important quantities:

 * `Bound()` (of type `HRectBound`): this specifies the bounding hyper-rectangle of each node.  Any point held in a node (and any point held in any descendant of that node) will be contained within the bound.

 * `FurthestDescendantDistance()`: this is the distance between the center of the bound and the furthest descendant point from the center.  (A descendant point is any point held in the node or any descendant of the node.  In our `BinarySpaceTree` class, only leaf nodes in the tree hold any points, so the set of descendant points of a given node N turns out to be the points held in all leaf nodes that are descendants of the node N.)

 * `MinimumBoundDistance()`: this is half the minimum width of the bound.

 * `ParentDistance()`: this is the distance between the center of the bound and the center of the parent's bound.

[...]

PR #1472, opened 2018-07-20
LMNN: Tree optimization

manish7294:
See #1445 

As of now, the changes are made for just lbfgs optimizer as the plan was to adapt the same for others once it is successfully done for lbfgs. So, basically one of the Impostors() function is modified.

### DIFF
--- a/src/mlpack/core/tree/binary_space_tree/binary_space_tree.hpp
+++ b/src/mlpack/core/tree/binary_space_tree/binary_space_tree.hpp
@@ -390,9 +390,6 @@ class BinarySpaceTree
   //! Return the minimum distance from the center of the node to any bound edge.
   ElemType MinimumBoundDistance() const;
 
-  //! Modify the minimum distance from the center of the node to any bound edge.
-  ElemType& MinimumBoundDistance() { return minimumBoundDistance; }
-
   //! Return the distance from the center of this node to the center of the
   //! parent node.
   ElemType ParentDistance() const { return parentDistance; }
--- a/src/mlpack/core/tree/binary_space_tree/binary_space_tree.hpp
+++ b/src/mlpack/core/tree/binary_space_tree/binary_space_tree.hpp
@@ -384,8 +384,14 @@ class BinarySpaceTree
    */
   ElemType FurthestDescendantDistance() const;
 
+  //! Modify the furthest possible descendant distance.
+  ElemType& FurthestDescendantDistance() { return furthestDescendantDistance; }
+
   //! Return the minimum distance from the center of the node to any bound edge.
   ElemType MinimumBoundDistance() const;
+
+  //! Modify the minimum distance from the center of the node to any bound edge.
+  ElemType& MinimumBoundDistance() { return minimumBoundDistance; }
 
   //! Return the distance from the center of this node to the center of the
   //! parent node.
--- a/src/mlpack/methods/lmnn/constraints_impl.hpp
+++ b/src/mlpack/methods/lmnn/constraints_impl.hpp
@@ -42,8 +42,8 @@ template<typename TreeType>
 inline void UpdateTree(TreeType& node)
 {
   double minWidth = DBL_MAX;
-  arma::Col<double> mins(node.Dataset().n_rows);
-  arma::Col<double> maxs(node.Dataset().n_rows);
+  arma::Col<typename TreeType::ElemType> mins(node.Dataset().n_rows);
+  arma::Col<typename TreeType::ElemType> maxs(node.Dataset().n_rows);
   for (size_t d = 0; d < node.Bound().Dim(); ++d)
   {
     mins[d] = node.Dataset().col(node.Descendant(0))[d];
@@ -214,7 +214,7 @@ void Constraints<MetricType>::Impostors(arma::Mat<size_t>& outputMatrix,
                                         arma::mat& outputDistance,
                                         const arma::mat& dataset,
                                         const arma::Row<size_t>& labels,
-                                        const arma::mat& diffTransform)
+                                        const arma::mat& transformation)
 {
   // Perform pre-calculation. If neccesary.
   Precalculate(labels);
@@ -228,12 +228,10 @@ void Constraints<MetricType>::Impostors(arma::Mat<size_t>& outputMatrix,
     // set and same class points as query set.
     if (knnObjects.size() == uniqueLabels.n_elem)
     {
-      auto Tree = knnObjects[i].ReferenceTree();
+      knnObjects[i].ReferenceTree().Dataset() =
+        transformation * origDatasets[i];
+      UpdateTree(knnObjects[i].ReferenceTree());
 
-      Tree.Dataset() = diffTransform * origDatasets[i];
-      UpdateTree(Tree);
-
-      knnObjects[i].Train(Tree);
       knnObjects[i].Search(dataset.cols(indexSame[i]), k, neighbors, distances);
     }
     else
--- a/src/mlpack/methods/lmnn/constraints_impl.hpp
+++ b/src/mlpack/methods/lmnn/constraints_impl.hpp
@@ -37,6 +37,52 @@ Constraints<MetricType>::Constraints(
   }
 }
 
+template<typename TreeType>
+inline void UpdateTree(TreeType& node, const arma::mat& dataset)
+{
+  double minWidth = DBL_MAX;
+  arma::Col<double> mins(min(dataset, 1));
+  arma::Col<double> maxs(max(dataset, 1));
+
+    // Update each of the HRectBound dimensions...
+  for (size_t d = 0; d < node.Bound().Dim(); ++d)
+  {
+    // Basically we had a previous bound and we want to "stretch" it for each dimension...
+    node.Bound()[d] |= math::RangeType<double>(mins[d], maxs[d]);
+
+    // Calculate minWidth.
+    double width = node.Bound()[d].Width();
+      minWidth = width;
+  }
+
+  node.Bound().MinWidth() = minWidth;
+
+  // We also have to update the properties of the tree.  Some are easier after recursion.
+  node.MinimumBoundDistance() = node.Bound().MinWidth() / 2.0;
+  // Technically this is loose but it is what the BinarySpaceTree already does.
+  node.FurthestDescendantDistance() = 0.5 * node.Bound().Diameter();
+
+  // Recurse into the children.
+  if (node.NumChildren() > 0)
+  {
+    UpdateTree(node.Child(0), dataset);
+    UpdateTree(node.Child(1), dataset);
+
+    // Recompute the parent distance for the left and right child.
+    arma::vec center, leftCenter, rightCenter;
+    node.Center(center);
+    node.Child(0).Center(leftCenter);
+    node.Child(1).Center(rightCenter);
+
+    const double leftParentDistance = node.Metric().Evaluate(center, leftCenter);
+    const double rightParentDistance = node.Metric().Evaluate(center,
+        rightCenter);
+
+    node.Child(0).ParentDistance() = leftParentDistance;
+    node.Child(1).ParentDistance() = rightParentDistance;
+  }
+}
+
 // Calculates k similar labeled nearest neighbors.
 template<typename MetricType>
 void Constraints<MetricType>::TargetNeighbors(arma::Mat<size_t>& outputMatrix,
@@ -145,7 +191,7 @@ void Constraints<MetricType>::Impostors(arma::Mat<size_t>& outputMatrix,
 // Calculates k differently labeled nearest neighbors. The function
 // writes back calculated neighbors & distances to passed matrices.
 template<typename MetricType>
-void Constraints<MetricType>::Impostors(arma::Mat<size_t>& outputNeighbors,
+void Constraints<MetricType>::Impostors(arma::Mat<size_t>& outputMatrix,
                                         arma::mat& outputDistance,
                                         const arma::mat& dataset,
                                         const arma::Row<size_t>& labels)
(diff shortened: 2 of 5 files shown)


manish7294:
>  // TODO: here is our bug.  The tree's dataset does not hold the whole dataset---remember, it holds dataset.cols(indexDiff[i]).  But when the call knn.Train(dataset.cols(indexDiff[i])) happens, that submatrix is copied *and rearranged*.  So, it's likely the easiest thing to do here  is pass (transformation - oldTransformation) and multiply that with Tree.Dataset(), to incrementally update the values in the dataset here.

Right, this is definitely a bug. Thanks for pointing this out.
Unfortunately we can't do that here as we apply `transformation` on original dataset whereas `Tree.Dataset()` already contains last iteration transformed dataset. So, it will be like transformation2 * (transformation1 * dataset). Need to figure out something else.

manish7294:
@rcurtin Is it possible to stop rearranging the dataset when tree is build?

manish7294:
I think this will work -

`Tree.Dataset() = transformation * inverseOf (oldTransformation) * Tree.Dataset();`

But the problem here will be finding the `inverseOf (oldTransformation)` as in our case, it is not necessarily a square matrix.

What do you think?

rcurtin:
I thunk maybe storing the original reordered data matrices after tree building into some `std::vector<arma::mat>`could work here.  So then you could do

```
tree.Dataset() = transformation * origTreeData[i];
```

or something to this effect.

manish7294:
I think it still won't be the original dataset for optimizer. It will depend on the initial learning point and even if initial learning point is identity matrix, it won't work for optimizers using batches.

rcurtin:
Have you tried the idea?  Why do you think it won't work for optimizers using batches?

manish7294:
Like just after first batch is processed, transformation will be updated and hence for the second batch it will a differently transformed dataset.

rcurtin:
Why will the transformation be updated?  What I am saying is that we have a set of reference trees that we built when we created the `Constraints()` objects.  We can store the datasets on which we build the trees---these are untransformed.  Then, whenever we want to do a search in `Impostors()`, we simply set the tree's dataset to the current transformation multiplied with the original untransformed dataset, update the tree, and search.  This should work for both batches and full-dataset computations.  If I can clarify what I mean, just let me know.

manish7294:
Oh! Are you referring to the reference trees being built in the LMNN function constructor. If that's the case I think that will work, just a bit of tweaking will be needed to pass them to constraints and adapt according to the search.

rcurtin:
Right, those are the important ones to cache.  If we can cache query trees for the whole-dataset evaluations that is great too but it's not likely to be possible or useful to cache a query tree for the small batch case.  Do you think you have enough info and ideas to finish the implementation from here?

manish7294:
I think it's atleast a good starting point, I will first try doing it from here and if by any chance I face some problem, I will let you know.

src/mlpack/core/tree/binary_space_tree/binary_space_tree.hpp
```diff
@@ -384,9 +384,15 @@ class BinarySpaceTree
    */
   ElemType FurthestDescendantDistance() const;
 
+  //! Modify the furthest possible descendant distance.
+  ElemType& FurthestDescendantDistance() { return furthestDescendantDistance; }
```
manish7294: It seems like this change is making a number of tests to fail.

manish7294:
@rcurtin  Can you please check and verify that this is what you were referring to earlier today.

src/mlpack/core/tree/binary_space_tree/binary_space_tree.hpp
rcurtin: What have you tried to debug this?

src/mlpack/core/tree/binary_space_tree/binary_space_tree.hpp
```diff
@@ -390,9 +390,6 @@ class BinarySpaceTree
   //! Return the minimum distance from the center of the node to any bound edge.
   ElemType MinimumBoundDistance() const;
 
-  //! Modify the minimum distance from the center of the node to any bound edge.
-  ElemType& MinimumBoundDistance() { return minimumBoundDistance; }
```
rcurtin: Right, it looks like the other version of this overload calls `bound.MinWidth()`.  So the `minimumBoundDistance` member appears to be entirely unused in `BinarySpaceTree` and we can remove it.

manish7294: And this sloves the failing tests issue as well :)

src/mlpack/methods/lmnn/constraints_impl.hpp
```diff
@@ -37,6 +37,71 @@ Constraints<MetricType>::Constraints(
   }
 }
 
+// We assume the dataset held in the tree has already been stretched.
+template<typename TreeType>
+inline void UpdateTree(TreeType& node)
+{
+  double minWidth = DBL_MAX;
+  arma::Col<double> mins(node.Dataset().n_rows);
+  arma::Col<double> maxs(node.Dataset().n_rows);
```
rcurtin: Technically this should actually be `arma::Col<typename TreeType::ElemType>` (or something like this) to make it maximally generic.  So if you can change it it would be great, but it won't affect any functionality.

src/mlpack/methods/lmnn/constraints_impl.hpp
```diff
@@ -145,33 +210,50 @@ void Constraints<MetricType>::Impostors(arma::Mat<size_t>& outputMatrix,
 // Calculates k differently labeled nearest neighbors. The function
 // writes back calculated neighbors & distances to passed matrices.
 template<typename MetricType>
-void Constraints<MetricType>::Impostors(arma::Mat<size_t>& outputNeighbors,
+void Constraints<MetricType>::Impostors(arma::Mat<size_t>& outputMatrix,
                                         arma::mat& outputDistance,
                                         const arma::mat& dataset,
-                                        const arma::Row<size_t>& labels)
+                                        const arma::Row<size_t>& labels,
+                                        const arma::mat& diffTransform)
```
rcurtin: I think the name `diffTransform` is misleading here, shouldn't it just be `transformation`?

src/mlpack/methods/lmnn/constraints_impl.hpp
```diff
@@ -145,33 +210,50 @@ void Constraints<MetricType>::Impostors(arma::Mat<size_t>& outputMatrix,
...
-    // set and  same class points as query set.
-    knn.Train(dataset.cols(indexDiff[i]));
-    knn.Search(dataset.cols(indexSame[i]), k, neighbors, distances);
+    // set and same class points as query set.
+    if (knnObjects.size() == uniqueLabels.n_elem)
+    {
+      auto Tree = knnObjects[i].ReferenceTree();
+
+      Tree.Dataset() = diffTransform * origDatasets[i];
+      UpdateTree(Tree);
```
rcurtin: If you want to make it even shorter you can do:

```
knnObjects[i].ReferenceTree().Dataset() = diffTransform * origDatasets[i];
UpdateTree(knnObjects[i].ReferenceTree());
```

In fact I am not sure if the `auto` will automatically infer a reference here (I don't remember exactly how auto works), so it's probably safer to do what I suggested unless `auto` is guaranteed to always infer a non-const reference instead of a copy or anything.

src/mlpack/methods/lmnn/constraints_impl.hpp
```diff
@@ -145,33 +210,50 @@ void Constraints<MetricType>::Impostors(arma::Mat<size_t>& outputMatrix,
...
+
+      knnObjects[i].Train(Tree);
```
rcurtin: This line actually shouldn't be necessary---we've already updated the tree internally (assuming `UpdateTree()` is called on a reference).

src/mlpack/methods/lmnn/constraints_impl.hpp
manish7294: Done!

src/mlpack/methods/lmnn/constraints_impl.hpp
manish7294: Ahh! forgot to change this one. Thanks for pointing that out! It's corrected now.

src/mlpack/methods/lmnn/constraints_impl.hpp
manish7294: Agreed, that's much better :)

src/mlpack/methods/lmnn/constraints_impl.hpp
manish7294: Done!
Sorry for not being available yesterday, I was shifting to my new dorm. It took some paperwork and a lot of hard work and almost took half of today's time :(

rcurtin: No worries, I know how much work a move can be!  I hope the heavy lifting is all done at this point. :)

manish7294:
Here's the simulation, though results are not good - https://gist.github.com/manish7294/106c16e82cc4c7bf44375f6ea51b6334

rcurtin:
Which optimizers was this for?  What were the values of k?  Did you time how long the tree update part took?  If the results are not good we must debug them.

manish7294:
I have done some timing simulations. Apparently, `UpdateTree()` is taking quite some time and is the reason for the increase in runtimes.

rcurtin:
Can you give full details of what you find please?  In general if I am trying to help debug something like this, two sentences don't give me enough clarity to give suggestions or help out.  This applies in general, not just to this particular PR debugging, by the way.

And which optimizers were you using for these simulations, and how do others perform?

manish7294:
I am using lbfgs for this, as currently, it's the only one for which optimization code is there. Others work over batches, so need to figure out the details for that. 

Here's a simple comparison of `impostors() ` calculation timing -
master branch lmnn takes 4.205126s for  k = 10 whereas the tree optimized version takes 5.882666s to run and moreover just `UpdateTree()` takes 1.587435s.

This is all in terms of runtime simulations. I understand this doesn't make much sense for debugging purpose but earlier I was just mentioning the relative timings, Sorry if that created confusion. So basically, I meant that UpdateTree() is taking a significant amount of time, maybe iteratively updating all the descendants of every node is just too costly.

rcurtin:
The results you're giving are all really inexact and it's hard for me to provide any in-depth guidance other than "have you tried to figure out why `UpdateTree()` is slow?"

Unfortunately the reality is that as a mentor there is no way I can devote the kind of time to this project that you are able to.  I can try to do my best, but if you don't provide very in-depth observations about what is going on and why you think it is going on, then I have to make the in-depth observations myself, which can take many hours.  To be honest, I don't really have the time to do that, so if you can't dig deeper here we're not likely to make any progress.  If we hope to publish something on our implementation here, the publishability of our work to some extent depends on the optimization strategies being non-obvious; so, for instance, simply implementing it in C++ is not sufficient, instead we must provide additional speedup by using trees, pruning unneeded computations, and so forth.  It's my belief that caching trees should be an optimization that should provide us some speedup.

Certainly you are right that iterating over all the descendants is costly, so you should be able to devise a way to update the bound values directly.  Probably this means that instead of caching `origDatasets`, you'll have to store the trees instead in some `origTrees` vector.

manish7294:
Right, I understand! Don't worry, I am already working on this. I just meant to update you that as of now implementation is slow.

rcurtin:
Ok, that is fine---but please be sure, when you write up an update, to be really clear about what you've done, what the simulation results show, why you think it was the right thing to do, etc.  Not only does this help me out, but it helps us both out later when we go back to write up the full set of accelerations we found for LMNN.

manish7294:
https://github.com/mlpack/mlpack/pull/1472/commits/c3d4c9b84133331ff08e6b73fa6088d1bb8e3927 This change have helped in making a significant improvement in the runtimes of this PR. I think there is still a lot that can be improved. 

I think I may have an idea to get some more speedup - 
Currently we are calculating mins and maxs for every node using below lines of code -
```
arma::Col<typename TreeType::ElemType> mins(min(node.Dataset().cols(
      node.Begin(), node.Begin() + node.Count() - 1), 1));
arma::Col<typename TreeType::ElemType> maxs(max(node.Dataset().cols(
     node.Begin(), node.Begin() + node.Count() - 1), 1));
```
I think instead of recalculating this multiple times for each node, why don't we just calculate this ones (by iteratively calulating mins and maxs for every dimension starting from last datapoint coming to the first one ), this will take just O(dataset.n_cols * d) time & O(dataset.n_cols * d) space and then carefully fetching the mins and maxs for a particular node from the earlier calculated result. Though some more details need to be figure out. Do you think it's a  reasonable idea.

manish7294:
@rcurtin Let's say we have a node and it's uses dataset begining from `node.Begin() `index, so is it like that it will be using all the data points between `node.Begin() ` & `dataset.n_cols` making ` node.Count()  =  dataset.n_cols  -  node.Begin()  + 1` or the `node.Count()` value can be anything?

manish7294:
@rcurtin  My above idea to compute `mins` and `maxs` only once didn't work out quite as expected since I assumed that each node utilizes the data starting from `node.Begin()` to end of dataset.  Other than this, I don't think I am getting any other to get speedup.

One strange thing I have noticed is that (I am really sorry if it doesn't seem quite clear, It's something that is hard to explain for me) -> I am seeing an average increase in ` number of node combinations were scored`. & `base cases were calculated`. when using `updateTree() ` w.r.t to when not using ` updateTree()`, though final distance matrix is same for both cases. And somewhere I feel this is the reason of lack of speedup.

for instance here's a comparison on iris of above two parameters:
without updateTree()


```
[INFO ] 215 node combinations were scored.
[INFO ] 550 base cases were calculated.
[INFO ] 244 node combinations were scored.
[INFO ] 1164 base cases were calculated.
[INFO ] 208 node combinations were scored.
[INFO ] 1686 base cases were calculated.
[DEBUG] L-BFGS iteration 1; objective 531.336, gradient norm 324.273, 0.239603.
[INFO ] 211 node combinations were scored.
[INFO ] 550 base cases were calculated.
[INFO ] 263 node combinations were scored.
[INFO ] 1100 base cases were calculated.
[INFO ] 208 node combinations were scored.
[INFO ] 1789 base cases were calculated.
[DEBUG] L-BFGS iteration 2; objective 498.662, gradient norm 188.14, 0.0614949.
[INFO ] 264 node combinations were scored.
[INFO ] 550 base cases were calculated.
[INFO ] 266 node combinations were scored.
[INFO ] 1035 base cases were calculated.
[INFO ] 230 node combinations were scored.
[INFO ] 1789 base cases were calculated.
[DEBUG] L-BFGS iteration 3; objective 498.097, gradient norm 52.4243, 0.00113241.

```
**Average Node combinations : 234.33
Average base case : 1134.778**


without updateTree()
```

[INFO ] 159 node combinations were scored.
[INFO ] 750 base cases were calculated.
[INFO ] 283 node combinations were scored.
[INFO ] 946 base cases were calculated.
[INFO ] 150 node combinations were scored.
[INFO ] 1505 base cases were calculated.
[DEBUG] L-BFGS iteration 1; objective 531.336, gradient norm 324.273, 0.239603.
[INFO ] 112 node combinations were scored.
[INFO ] 750 base cases were calculated.
[INFO ] 262 node combinations were scored.
[INFO ] 974 base cases were calculated.
[INFO ] 148 node combinations were scored.
[INFO ] 1653 base cases were calculated.
[DEBUG] L-BFGS iteration 2; objective 498.662, gradient norm 188.14, 0.0614949.
[INFO ] 142 node combinations were scored.
[INFO ] 750 base cases were calculated.
[INFO ] 268 node combinations were scored.
[INFO ] 1165 base cases were calculated.
[INFO ] 170 node combinations were scored.
[INFO ] 1505 base cases were calculated.
[DEBUG] L-BFGS iteration 3; objective 498.097, gradient norm 52.4243, 0.00113241.

```
**Average Node combinations :  188.22
Average base case : 1110.888**

And in my above idea of pre-computing mins and maxs, I saw the same thing like getting correct final distance matrix but a high increase in node combinations and base calculation leading to bad runtime.

rcurtin:
Right---as we update the tree, it's no longer likely to be the best tree for search.  So the number of prunes will slightly decrease (and in turn the number of scores and base cases will go up, which is exactly what you saw).  However the idea here is that the cost of rebuilding the tree is far higher than the slight degradation of reusing the tree.  But this only works if we can update the tree quickly.

Let me play with this one in the context of the branch I'm working in, and let's see if I can make it any better.  I'll use your code as a base idea---thank you for the hard work, it will be easy to plug new ideas in because of how you've structured the code.  :+1:

I'll post an update here if I can get an improvement. :)

PR #1474, opened 2018-07-25
Make use of OpenMP optional

### LINKED ISSUE #1473: How do I explicit disable OpenMP using cmake in mlpack?
How do I explicit disable OpenMP using cmake in mlpack?

rcurtin:
This fixes #1473, and I'll release mlpack 3.0.3 with this and some other fixes once this is merged.

Basically this just adds a CMake configuration option `USE_OPENMP` that allows the user to specify whether or not to use OpenMP.

So you can now configure as

```
$ cmake -DUSE_OPENMP=OFF ..
```

and OpenMP won't be used.

I also noticed the mlpack paper citation in the readme is out of date, so I updated that too. :)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -21,6 +21,7 @@ option(MATHJAX
     "Use MathJax for HTML Doxygen output (disabled by default)." OFF)
 option(FORCE_CXX11
     "Don't check that the compiler supports C++11, just assume it.  Make sure to specify any necessary flag to enable C++11 as part of CXXFLAGS." OFF)
+option(USE_OPENMP "If available, use OpenMP for parallelization." ON)
 enable_testing()
 
 # Currently Python bindings aren't known to build successfully on Windows, so
@@ -348,7 +349,10 @@ add_definitions(-DBOOST_TEST_DYN_LINK)
 #   ... openMP code here ...
 # }
 # #endif
-find_package(OpenMP)
+if (USE_OPENMP)
+  find_package(OpenMP)
+endif ()
+
 if (OPENMP_FOUND)
   add_definitions(-DHAS_OPENMP)
   set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} ${OpenMP_C_FLAGS}")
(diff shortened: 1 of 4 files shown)


rcurtin:
@gmanlan---I also made some simple documentation changes to link to your Windows tutorial from the main README.md, and added links from the Windows tutorial to the others that are around.  Let me know if you see any issues from that (commit 4b33dbb).

CMakeLists.txt
```diff
@@ -348,7 +349,10 @@ add_definitions(-DBOOST_TEST_DYN_LINK)
 #   ... openMP code here ...
 # }
 # #endif
-find_package(OpenMP)
+if (USE_OPENMP)
+  find_package(OpenMP)
+endif ()
+
 if (OPENMP_FOUND)
```
zoq: I'm wondering what's ``OPENMP_FOUND`` if ``USE_OPENMP`` is ``OFF``; undefined? So there is no need to do something like?:

```
if (USE_OPENMP  AND OPENMP_FOUND)
...
```

README.md
```diff
@@ -61,15 +61,18 @@ documentation. The website should be consulted for further information:
...
-    @article{mlpack2013,
-      title     = {{mlpack}: A Scalable {C++} Machine Learning Library},
-      author    = {Curtin, Ryan R. and Cline, James R. and Slagle, Neil P. and
-                   March, William B. and Ram, P. and Mehta, Nishant A. and Gray,
-                   Alexander G.},
-      journal   = {Journal of Machine Learning Research},
-      volume    = {14},
-      pages     = {801--805},
-      year      = {2013}
+    @article{mlpack2018,
```
zoq: Nice!

CMakeLists.txt
rcurtin: `OPENMP_FOUND` should be an unset variable, and CMake's `if()` operator will evaluate to false when a `if(<variable>)` is called for an undefined variable:

https://cmake.org/cmake/help/v3.0/command/if.html

README.md
rcurtin: Yeah, I forgot to do this earlier, sorry about that. :)

rcurtin:
Sounds good, since it is approved and it is such a simple fix I will merge it now.

PR #1476, opened 2018-07-27
Update LMNN low-rank tests to use all three optimizers.

rcurtin:
@manish7294 ---take a look at this and tell me what you think.

Basically all I have done is reworked the low-rank LMNN tests so that they can be included in the test suite.  It's true that sometimes the low-rank LMNN may randomly perform poorly (or at least very differently) based on initial starting point, so I've made it so that it can take three trials (or up to five with BBSGD since it seems to perform worse).

### DIFF
--- a/src/mlpack/tests/lmnn_test.cpp
+++ b/src/mlpack/tests/lmnn_test.cpp
@@ -17,6 +17,7 @@
 #include <mlpack/methods/lmnn/lmnn.hpp>
 #include <mlpack/core/optimizers/lbfgs/lbfgs.hpp>
 #include <mlpack/methods/neighbor_search/neighbor_search.hpp>
+#include <mlpack/core/optimizers/bigbatch_sgd/bigbatch_sgd.hpp>
 
 #include <boost/test/unit_test.hpp>
 #include "test_tools.hpp"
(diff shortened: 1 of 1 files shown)


src/mlpack/tests/lmnn_test.cpp
```diff
@@ -442,53 +443,171 @@ BOOST_AUTO_TEST_CASE(LMNNAccuracyTest)
...
+
+    // Learn a low rank matrix.
+    outputMatrix = arma::randu(4, 5);
+    lmnn.LearnDistance(outputMatrix);
+
+    double acc2 = KnnAccuracy(outputMatrix * dataset, labels, 1);
+
+    // We keep the tolerance very high.  We need to ensure the accuracy drop
+    // isn't any more than 10%.
+    success = ((acc1 - acc2) <= 10.0);
```
manish7294: Why we are only considering  accuracy drop case? I think it accuracy may increase as well and in that case this will always result in success no matter how much the difference. So, shouldn't we use `abs(acc1 - acc2)` here. What do you think?

rcurtin: If accuracy increases, even by a lot, I don't see this as an issue at all.  Mostly I want to check that accuracy doesn't drop way off, because to me that could indicate a big bug.  (Like if accuracy goes from 50% to 0%, for instance.)

manish7294: hmm, Then it's all right. I think it's good to merge :)

PR #1477, opened 2018-07-27
[WIP] Add SSRBM CIFAR-10 Test

ShikharJ:
Relevant: #1207

### DIFF
--- a/src/mlpack/tests/rbm_network_test.cpp
+++ b/src/mlpack/tests/rbm_network_test.cpp
@@ -226,13 +226,8 @@ BOOST_AUTO_TEST_CASE(SpikeSlabRBMCIFARTest)
   double slabPenalty = 10.5;
   double lambdaBias = 10;
 
-  if (!CLI::HasParam("dataset"))
-    Log::Fatal << "Input dataset is undefined!" << std::endl;
-  if (!CLI::HasParam("label"))
-    Log::Fatal << "Input lable  is undefined!" << std::endl;
-
-  std::string datafile = CLI::GetParam<std::string>("dataset");
-  std::string labelfile = CLI::GetParam<std::string>("label");
+  std::string datafile = "";
+  std::string labelfile = "";
 
   std::cout << "dataset = '" << dataset << "'" << std::endl;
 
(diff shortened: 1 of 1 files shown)


src/mlpack/tests/rbm_network_test.cpp
```diff
@@ -208,6 +208,182 @@ BOOST_AUTO_TEST_CASE(ssRBMClassificationTest)
...
+  size_t batchSize = 10;
+  double stepSize = 0.0001;
+  size_t numEpoches = 10;
+  double tolerance = 1e-5;
+  bool shuffle = true;
+  size_t poolSize = 4;
+  double slabPenalty = 10.5;
+  double lambdaBias = 10;
+
+  if (!CLI::HasParam("dataset"))
```
zoq: For the test we can just hardcode the dataset, no need to use the cli methods.

Rajiv2605:
@ShikharJ @zoq  I'll work on this issue :+1:

Rajiv2605:
@ShikharJ @zoq I am able to import data into `arma::mat`. According to the CIFAR-10 documentation,
The binary version contains the files formatted as follows:
`<1 x label><3072 x pixel>`
The dimension of the armadillo matrix after `.load()` is: `3841250 x 1`
I tried this code:
`std::cout<<testData(0, 0)<<std::endl;`
and also this,
`std::cout<<testData(3073, 0)<<std::endl;`
but both are not giving me the class number.
I also tried reshaping and extracting the first row thinking that it would be the set of labels:
```
trainData = arma::reshape(trainData, 3073, 1250);
trainLabels = trainData.col(0);
```
but even that is not giving me the class labels. What mistake am I making?

zoq:
Sorry for the slow response, which format (binary,matlab,python,csv) of the dataset do you use?

heisenbuug:
I was actually thinking of taking this as my GSoC project. Can you tell me what all work needs to be done so I can create an effective proposal.

Rajiv2605:
> Sorry for the slow response, which format (binary,matlab,python,csv) of the dataset do you use?

@zoq I am using binary format for the dataset.

khizirsiddiqui:
Hey, I was thinking to complete the PR. While building the PR, cmake produces error: `make[3]: *** No rule to make target '../src/mlpack/core/arma_extend/fn_ccov.hpp', needed by 'src/mlpack/cotire/mlpack_CXX_prefix.hxx.gch'.  Stop.` This filename doesn't exist in master branch at least, was this file renamed or something?

rcurtin:
Clearing your build directory and reconfiguring CMake from scratch should fix that, or alternately, maybe `fn_ccov.hpp` is still referenced on this branch somewhere in some `CMakeLists.txt` file?

khizirsiddiqui:
@rcurtin I couldn't find `fn_ccov.hpp` anywhere in `tests/CMakeLists.txt` or any other cmake file. Isn't this strange?
Honestly, I do not know much about writing CMake files, so I will give it a try after reading through some relevant tutorials. :+1:

rcurtin:
It may be leftover from a previous build; we removed `fn_ccov.hpp` some time ago.  Did you try clearing the build directory?  It's also possible that `fn_ccov.hpp` is referenced somewhere in the codebase on this branch, so be sure to search through every file; for instance with grep or something.

rcurtin:

PR #1479, opened 2018-07-28
[WIP] Add Bernoulli distribution and support for beta, conditional VAEs.

akhandait:


src/mlpack/methods/ann/dists/bernoulli_distribution.hpp
```diff
@@ -0,0 +1,132 @@
...
+{
+ public:
+  /**
+   * Default constructor, which creates a Bernoulli distribution with zero
+   * dimension.
+   */
+  BernoulliDistribution();
+
+  /**
+   * Create multiple Bernoulli distributions with the given parameters.
```
sumedhghaisas: multiple Bernoulli? This is quite vague, we need to specify how the distributions are defined. For example, we have to mention they are independent distributions and how many are defined? We also need to mention how does the shape of params affect the distribution shape.

src/mlpack/methods/ann/dists/bernoulli_distribution.hpp
```diff
@@ -0,0 +1,132 @@
...
+#define MLPACK_METHODS_ANN_DISTRIBUTIONS_BERNOULLI_DISTRIBUTION_HPP
+
+#include <mlpack/prereqs.hpp>
+#include "../activation_functions/logistic_function.hpp"
+
+namespace mlpack {
+namespace ann /** Artificial Neural Network. */ {
+
+/**
+ * Multiple Bernoulli distributions.
```
sumedhghaisas: We should give a very small discription of Bernoulli distribution, how its defined here, for example we define the distributions by their expected value, p, given to the class by 'params' parameter. And the independent assumption between the distributions.

src/mlpack/methods/ann/layer/reparametrization.hpp
```diff
@@ -46,10 +46,12 @@ class Reparametrization
    * @param latentSize The number of output latent units.
    * @param stochastic Whether we want random sample or constant.
    * @param includeKl Whether we want to include KL loss in backward function.
+   * @param beta Hyperparameter for constrained variational frameworks.
```
sumedhghaisas: 'beta' parameter needs to be described properly with reference to the paper. I suggest a small paragraph in the class description and mention here in the function.

akhandait:
I just realized that I forgot to make some changes that you had suggested in the ReconstructionLoss PR. I will make them tomorrow.

sumedhghaisas:
@rcurtin @zoq PR looks good to me. No more comments from my side. I will go ahead and merge it in 3 days, leaving time for some other comments.

src/mlpack/methods/ann/dists/bernoulli_distribution_impl.hpp
```diff
@@ -0,0 +1,85 @@
...
+        logits.n_cols, false, false);
+  }
+}
+
+template<typename DataType>
+DataType BernoulliDistribution<DataType>::Sample() const
+{
+  DataType sample;
+  sample = arma::conv_to<DataType>::from(arma::randu<DataType>
+      (probability.n_rows, probability.n_cols) <= probability);
```
rcurtin: If this turns out to be a bottleneck, it might be possible to optimize by looping directly over the elements of `sample`.  (i.e. the conversion to `DataType` from `arma::uword` won't be optimized out at compile time, I don't think)

src/mlpack/methods/ann/loss_functions/mean_squared_error_impl.hpp
```diff
@@ -29,7 +29,7 @@ template<typename InputType, typename TargetType>
 double MeanSquaredError<InputDataType, OutputDataType>::Forward(
     const InputType&& input, const TargetType&& target)
 {
-  return arma::mean(arma::mean(arma::square(input - target)));
+  return arma::accu(arma::square(input - target)) / target.n_cols;
```
rcurtin: Nice, much cleaner in my opinion. :+1:

src/mlpack/methods/ann/dists/bernoulli_distribution_impl.hpp
```diff
@@ -0,0 +1,85 @@
...
+  {
+    probability = arma::mat(logits.memptr(), logits.n_rows,
+        logits.n_cols, false, false);
+  }
+}
+
+template<typename DataType>
+DataType BernoulliDistribution<DataType>::Sample() const
+{
+  DataType sample;
```
zoq: We could remove the extra line an directly return the result.

src/mlpack/methods/ann/dists/bernoulli_distribution_impl.hpp
```diff
@@ -0,0 +1,85 @@
...
+      (probability.n_rows, probability.n_cols) <= probability);
+
+  return sample;
+}
+
+template<typename DataType>
+double BernoulliDistribution<DataType>::LogProbability(
+    const DataType&& observation) const
+{
+  return arma::accu(arma::log(probability + eps) % observation +
```
zoq: We could use ``trunc_log``.

akhandait: Yeah we can, I will change it, but I will still need eps for the denominator.

src/mlpack/methods/ann/dists/bernoulli_distribution_impl.hpp
akhandait: Oh, so do you mean something like this:
```cpp
Datatype sample = Datatype(size(probability));
for (int i = 0; i < probability.n_elem; i++)
{
  sample(i) = (int)randu(i)  < probability(i);
}
```
Also, can you please explain what exactly you mean when you say it won't be optimized at compile time? Can you point me to some source where I can read about it a bit?

src/mlpack/methods/ann/dists/bernoulli_distribution_impl.hpp
zoq: We can just leave the the ``eps`` parameter as a part of the constructor.

akhandait: Okay then, I will use `trunc_log` and `eps` for denominator then.

src/mlpack/methods/ann/dists/bernoulli_distribution_impl.hpp
akhandait: @rcurtin Sorry for reminding you, I am sure you would have gotten back on this, but I need to get this merged soon. Can you please answer the above doubt?

rcurtin: Hey Atharva,

Sorry, it was a really busy end-of-week and weekend and I didn't get to this.  I think the code is fine but you should cast to `Datatype` not `int`.

For the compile-time optimizations, Armadillo is a library that uses delayed evaluation.  So, basically, if I write 'matrix + matrix', the type returned is not a matrix, but instead a placeholder type that indicates that two matrices need to be added.

Section 8 of this paper has some more details:
http://citeseerx.ist.psu.edu/viewdoc/download?doi=10.1.1.739.504&rep=rep1&type=pdf

Also, here are the slides from a talk I gave that has more details:
http://conradsanderson.id.au/pdfs/sanderson_curtin_armadillo_pasc_2017.pdf

I hope this helps, let me know if I can clarify anything.  And sorry for the slow response. :)

akhandait: Thanks for the response!

sumedhghaisas:
@akhandait @zoq @rcurtin Is the PR ready to be merged? :)

rcurtin:
No other comments from my side, sorry again about the slowness.

akhandait:
@sumedhghaisas I need to push one final commit, after that it's ready.

PR #1482, opened 2018-07-31
Enable mean-centering

manish7294:
This PR includes code for performing mean-centering on the dataset.

### DIFF
--- a/src/mlpack/methods/lmnn/lmnn_main.cpp
+++ b/src/mlpack/methods/lmnn/lmnn_main.cpp
@@ -89,7 +89,9 @@ PROGRAM_INFO("Large Margin Nearest Neighbors (LMNN)",
     PRINT_PARAM_STRING("batch_size") + "), and the maximum number of passes "
     "(specified with " + PRINT_PARAM_STRING("passes") + ").  In "
     "addition, a normalized starting point can be used by specifying the " +
-    PRINT_PARAM_STRING("normalize") + " parameter. "
+    PRINT_PARAM_STRING("normalize") + " parameter. Furthermore, " +
+    "mean-centering can be performed on the dataset by specifying the " +
+    PRINT_PARAM_STRING("center") + "parameter. "
     "\n\n"
     "The L-BFGS optimizer, specified by the value 'lbfgs' for the parameter " +
     PRINT_PARAM_STRING("optimizer") + ", uses a back-tracking line search "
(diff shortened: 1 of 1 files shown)


src/mlpack/methods/lmnn/lmnn_main.cpp
```diff
@@ -89,7 +89,9 @@ PROGRAM_INFO("Large Margin Nearest Neighbors (LMNN)",
     PRINT_PARAM_STRING("batch_size") + "), and the maximum number of passes "
     "(specified with " + PRINT_PARAM_STRING("passes") + ").  In "
     "addition, a normalized starting point can be used by specifying the " +
-    PRINT_PARAM_STRING("normalize") + " parameter. "
+    PRINT_PARAM_STRING("normalize") + " parameter. Furthermore, " +
+    "mean-centering can be performed on the dataset by specifying the " +
+    PRINT_PARAM_STRING("center") + "parameter. "
```
rcurtin: Actually I realized that there's one more thing, which is that if we learn a metric using a mean-centered dataset, we can't then apply that metric to the non-mean-centered dataset.  So we should also add an output parameter `centered_data` or something like this, so that a user can have access to their mean-centered data for later use.  What do you think?

src/mlpack/methods/lmnn/lmnn_main.cpp
```diff
@@ -136,6 +138,8 @@ PARAM_INT_IN("rank", "Rank of distance matrix to be optimized. ", "A", 0);
 PARAM_FLAG("normalize", "Use a normalized starting point for optimization. It"
     "is useful for when points are far apart, or when SGD is returning NaN.",
     "N");
+PARAM_FLAG("center", "Perform mean-centering on the dataset. It is useful "
+    "when points have large norm values.", "C");
```
rcurtin: I guess technically it would be `when the centroid of the data is far from the origin`.

src/mlpack/methods/lmnn/lmnn_main.cpp
manish7294: Fair enough!

src/mlpack/methods/lmnn/lmnn_main.cpp
manish7294: Right, that's should be the one. Thanks for pointing this out.

manish7294: Right, that should be the one. Thanks for pointing this out.

Issue #365, opened 2014-12-31
Build windows installer package


rcurtin:
**Reported by speet3 on 8 Aug 2011**
Either NSIS or MSI, NSIS has a nice linux compiler, and MSI is the Microsoft supported installer format. In either case, the result needs to be published, probably in a similar location to the source tarball.

thirdwing:
Do you have the requirement for MSI or exe format? Maybe I can give a try this weekend.

rcurtin:
On Thu, Jan 08, 2015 at 03:40:46PM -0800, Qiang Kou (KK) wrote:
> Do you have the requirement for MSI or exe format? Maybe I can give a try this weekend.

Doesn't matter to me.  The goal would just be a Windows installer of
some sort that installs the mlpack executables (allknn, emst, etc.) and
the headers to somewhere on the system.  This will help a lot of people
who want to use mlpack on Windows.  Let me know if I can help out or
throw any hardware at it or something. :)

-- 
Ryan Curtin    | "Where we're going, we won't need eyes to see."
ryan@ratml.org |   - Dr. Weir

thirdwing:
OK. I will try this or next weekend. I have used Inno setup to package software several times before. There shouldn't be much problem.

II-Matto:
I was wondering if there are windows installers now?

thirdwing:
You can find an old version from https://dl.dropboxusercontent.com/u/81699687/mlpack-setup-x86_64-1_0_12.exe

II-Matto:
Many thanks! I just installed it, and found there is only *.a file. I guess this is compiled with mingw and cannot be used in visual studio?

thirdwing:
Yes. This is compiled using MinGW-w64.

If you want to use it with VS, you probably need to compile it by yourself.

On Wed, Sep 2, 2015 at 10:44 AM, II-Matto <notifications@github.com> wrote:

> Many thanks! I just installed it, and found there is only *.a file. I
> guess this is compiled with mingw and cannot be used in visual studio?
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/mlpack/mlpack/issues/365#issuecomment-137108232>.
>



-- 
Qiang Kou
qkou@umail.iu.edu
School of Informatics and Computing, Indiana University

II-Matto:
@thirdwing I just found the *.lib files which possibly can be used in visual studio. though they target only x86 platform. I will try to compile a x86 version with visual studio 2013.

[...]

PR #1485, opened 2018-08-02
Windows WiX Installer

gmanlan:
MSI Installer project for Windows, including appveyor script for staging/build/artifacts

Refers to #365


dist/win-installer/mlpack-win-installer/Product.wxs
```diff
@@ -0,0 +1,50 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<Wix xmlns="http://schemas.microsoft.com/wix/2006/wi">
+  <!-- DO NOT CHANGE the product GUID! It is forever -->
+	<Product Id="02A00C77-197D-4E91-B7D9-5836220E92E9"
+           UpgradeCode="6C2D7EC0-6F10-40CB-9703-1DC160A62662"
+           Name="MLPACK"
```
rcurtin: Ah, just a minor issue, can we change `MLPACK` to `mlpack` to match the rest of the codebase? :)

gmanlan: yep, done

gmanlan:
Ok, so I got rid of the .sln by just using raw heat/candle/light commands in appveyor.yml
Also switched to jpg instead of bmp (although wix reduces the quality even more).
Finally, added MLPACK_VERSION as env. var so the installer reads that for each build.

1) Do we want to add any other file in the installer? At the moment is the full /release directory and the includes. Also, do we need to distribute all the files in the /release dir (.exp, .lib. .dll .exe)?

2) Now the only problem I see is that we don't have a certificate to sign the installer (so Windows will refuse to install unless you select "run anyway")... purchasing a certificate is not that affordable... are you aware of any other option?

zoq:
> 1. Do we want to add any other file in the installer? 

Is it common to include the generated documentation as well? Also, including the complete release folder sounds like a good idea to me.

> 2. Now the only problem I see is that we don't have a certificate to sign the installer (so Windows will refuse to install unless you select "run anyway")... purchasing a certificate is not that affordable... are you aware of any other option?

I think there is no authority like let's encrypt which provides a certificate to sign the code, and if I remember right windows doesn't support gpg. Haven't checked the prices, so not sure buying one is an option at this point.

rcurtin:
> I think there is no authority like let's encrypt which provides a certificate to sign the code, and if I remember right windows doesn't support gpg. Haven't checked the prices, so not sure buying one is an option at this point.

Agreed.  I did a little reading and it seems like this is a pretty complex subject.  I didn't see any way to get a free certificate.  So I think it's no problem to distribute an unsigned installer for now, and if we find a way later to sign it, great, but if not I don't think it's any huge priority.

gmanlan:
Hi there,

Yes I will add the doc.
For the installer certificate, I couldn't find any reasonable option. There is only one company that provides 'kind of free' certificates for open source projects, but they still charge you for the hardware device required to use it (which I believe may have an annual fee later).

zoq:
> So I think it's no problem to distribute an unsigned installer for now, and if we find a way later to sign it, great, but if not I don't think it's any huge priority.

Agreed, don't think at this point this is a huge issue.

rcurtin:
Ok, for the documentation I set up a project on Jenkins:

http://ci.mlpack.org/job/mlpack%20-%20doxygen%20build/

I think that you could use this to get the most recent up-to-date Doxygen documentation pre-built.  I've set it up so every time there is a commit, the job kicks off.  It takes only about two minutes to run, so by the time the AppVeyor job runs, whatever documentation there should be up to date.  I tried to have it set the name of the build to the git commit hash so that we could ensure that the documentation being used by AppVeyor is the same as that being downloaded from ci.mlpack.org, but I did not see an easy way to do this---and I think it should be relatively rare that the documentation goes out of date anyway.

So, in any case, a few minutes after a git push happens, the Doxygen build job should be done, and you can access the artifact here:

http://ci.mlpack.org/job/mlpack%20-%20doxygen%20build/lastSuccessfulBuild/artifact/

or more specifically download it all as a zip:

http://ci.mlpack.org/job/mlpack%20-%20doxygen%20build/lastSuccessfulBuild/artifact/*zip*/archive.zip

It's not always guaranteed it will be there (maybe there is a failure?) so probably checking for a 404 or other errors is worthwhile, then I guess we can just build without documentation or fail the build (either is fine in my book).

Hope this helps... let me know if you want me to try and tune the Jenkins job at all.

gmanlan:
- Because of the nature of this project, including the dependencies is somehow optional but I agree we should add the dependencies for a smoother UX. So next question is, can we redistribute the dependencies based on the licenses? (I quickly checked openblas and boost, it seems we can, but I would appreciate a double check) - and in that case I think we need to include the third party licenses disclaimer in an installer's readme file.

- Moving the .exe to a different directory is not a bad idea, but then the .exe will not find the dll (i.e. mlpack.dll) so it's kind of breaking the consistency there. Alternatively, we can add the /lib directory to the PATH env var (pretty common but sometimes not recommended) so the binaries in a different path can find the libraries. And a third option would be to place both exe/dlls in the install root directory (like Anaconda does for example) and keep .lib/.exp in the /lib dir.

- Finally, if we are doing including the dependencies, then we should aim for the full solution and install the VC++ runtime redistributable as well?

rcurtin:
Ah, right, let me check the licenses (I'll do that tomorrow).  I think maybe the Anaconda solution to put it all in the root directory seems best, so if you concur, maybe we should do that?

Do we need the VC++ runtime? I thought that was only if we were using any managed C++ or anything.  It seems like the redistributable can be installed as a prerequisite, but I'm not 100% sure if this is helpful:

https://www.advancedinstaller.com/forums/viewtopic.php?t=29284

If it's too much work to set that up, I think it's fine to include the non-VC dependencies since I would imagine many users will have it anyway (would be nice though).

Thanks again for looking into this.  It is exciting to be able to provide a good package for Windows users finally. :)

rcurtin:
For the licenses, I think that all we need to do is include the licenses of the dependencies (Armadillo and Boost) with the distribution.  So for Armadillo that's the ASL 2.0 (https://gitlab.com/conradsnicta/armadillo-code/blob/unstable/LICENSE.txt) and for Boost that's the BSL (https://www.boost.org/LICENSE_1_0.txt).  I think it would be fine to include those as text files in the distribution, and if it's easy to have the users click 'agree' during installation it might be nice to add that there, but I also think it's not entirely necessary given that the BSD, ASL, and BSL licenses all have really easy-to-agree-to terms that I wouldn't ever expect anyone to reject.

gmanlan:
Thanks for double checking. Changes:
+ Included the dependencies licenses in License.rtf which is shown during the install wizard.
+ Moved .exe/.dll files to install root and keeping .lib/.exp in /lib
+ Included dependencies .dll in the install root as well

- Not adding VC++ runtime (feasible, but maybe not required at this time)

.appveyor.yml
```diff
@@ -19,6 +19,9 @@ install:
   - ps: nuget install boost_serialization-vc140 -o "${env:APPVEYOR_BUILD_FOLDER}" -Version 1.60.0
   - ps: nuget install boost_math_c99-vc140 -o "${env:APPVEYOR_BUILD_FOLDER}" -Version 1.60.0
   - ps: nuget install OpenBLAS -o "${env:APPVEYOR_BUILD_FOLDER}"
+  - set path=C:\Program Files (x86)\WiX Toolset v3.11\bin;%path%
+  - ps: $ver = (Get-Content "${env:APPVEYOR_BUILD_FOLDER}\src\mlpack\core\util\version.hpp" | where {$_ -like "*MLPACK_VERSION*"}); $env:MLPACK_VERSION = ($ver[0].substring($ver[0].length-1,1)+'.'+$ver[1].substring($ver[1].length-1,1)+'.'+$ver[2].substring($ver[2].length-1,1))
```
rcurtin: Hmm, the only thing is, if this is a git version instead of an official release, this will detect the most recently released version.  After the mlpack CMake configuration is done, the file `\src\mlpack\core\util\gitversion.hpp` will exist if this is the git master branch.  Internally will be just the single line

```
return "mlpack git-919dff2";
```

(Or whatever the git revision is.)  I don't speak powershell very well, but do you think it's possible to modify this to check if gitversion.hpp exists and extract `$ver` from there if so?

.appveyor.yml
```diff
@@ -35,11 +38,38 @@ build_script:
...
+  
+  # Preparing installer staging
+  - cd C:\projects\mlpack\dist\win-installer\staging && mkdir lib
+  - ps: cp C:\projects\mlpack\build\Release\*.lib C:\projects\mlpack\dist\win-installer\staging\lib\
+  - ps: cp C:\projects\mlpack\build\Release\*.exp C:\projects\mlpack\dist\win-installer\staging\lib\
+  - ps: cp C:\projects\mlpack\build\Release\*.dll C:\projects\mlpack\dist\win-installer\staging\
+  - ps: cp C:\projects\mlpack\build\Release\*.exe C:\projects\mlpack\dist\win-installer\staging\
+  - ps: cp C:\projects\mlpack\OpenBLAS.0.2.14.1\lib\native\bin\x64\*.dll C:\projects\mlpack\dist\win-installer\staging\
+  - ps: cp C:\projects\mlpack\boost_libs\boost_unit_test_framework-vc*.dll C:\projects\mlpack\dist\win-installer\staging\
+  - ps: cp C:\projects\mlpack\build\include\mlpack C:\projects\mlpack\dist\win-installer\staging -recurse
```
rcurtin: It might be worthwhile to put this into `staging\include`, but I don't think it's a huge deal either way.  To me it makes it clearer it's an include directory, but at the same time I know that Boost is distributed with its include directory just as `boost/` in the root directory.  Up to you. :+1:

.appveyor.yml
```diff
@@ -35,11 +38,38 @@ build_script:
...
+  - ps: cp C:\projects\mlpack\doc\examples C:\projects\mlpack\dist\win-installer\staging -recurse
+  
+  # Building MSI installer
+  - cd C:\projects\mlpack\dist\win-installer\mlpack-win-installer
+  - heat dir ..\staging -cg HeatGenerated -dr INSTALLFOLDER -sreg -srd -var var.HarvestPath -ag -sfrag -out HeatGeneratedFileList.wxs
```
rcurtin: Is it possible to wrap lines in the AppVeyor config?  These are super long. :)  I took a look, it looks like it can be done but it might be tricky.

https://help.appveyor.com/discussions/questions/1096-multiple-line-cmd-in-appveyoryml

I can try it in a separate PR after merging this if you like.

zoq:
> Do the images need to be .bmp? We could probably make them smaller with compression to .png or .jpg but I don't know if WiX requires .bmp images.

Actually, I was wondering if it's possible to use a higher resolution since the mlpack logo looks somewhat blurry. Don't think it would have a huge effect on the size since the most part is white anyway.

gmanlan:
@zoq, yes that's why I originally used the bmp files for slightly higher quality. However, the specification forces you to provide low-res bmp or jpg, and MSI installers reduce the quality of the image anyway (seems to be an old/known issue).

.appveyor.yml
rcurtin: Sure, that sounds good to me.  It is low priority anyway. :+1:

.appveyor.yml
rcurtin: Right, I trust your expertise here.  I am not so knowledgeable on Windows, so let's leave it as-is.  Thanks for the clarification. :+1:

rcurtin:
I had an opinion for smaller images, but you are right, it's not really that large anyway.  I'm fine switching to something larger if we like.  I guess my hope would be that maybe we can keep it under 50kB but if more is needed to make it looks good it's fine.  After all the test data directory is 3.7MB currently, so even 200kB is not much towards that.

Let me know when you're ready for merge and I can hit the button. :)

gmanlan:
Ok great, go ahead with the merge. My next two weeks are quite overloaded so I will not have time to work on this branch anymore. I will be back for more later (maybe we can gather feedback from users and optimize the installer with all the small changes all at once). Thanks!

.appveyor.yml
rcurtin: Ah, oops, I forgot to respond to this.  The gitversion.hpp is generated at build time, so it will be available on AppVeyor.  We can handle this in a follow-up issue if needed.

rcurtin:
Really happy to merge this.  Thanks so much for the hard work. :+1:  I opened #1506 to handle the few little remaining things.

PR #1487, opened 2018-08-02
Implementation of BoostMetric Distance Learning

manish7294:
An implementation of boostmetric distance learning technique.

### DIFF
--- a/src/mlpack/methods/boostmetric/boostmetric.hpp
+++ b/src/mlpack/methods/boostmetric/boostmetric.hpp
@@ -9,8 +9,8 @@
  * 3-clause BSD license along with mlpack.  If not, see
  * http://www.opensource.org/licenses/BSD-3-Clause for more information.
  */
-#ifndef MLPACK_METHODS_BOOSTMETRIC_BOOSTMETRIC_HPP
-#define MLPACK_METHODS_BOOSTMETRIC_BOOSTMETRIC_HPP
+#ifndef MLPACK_METHODS_BoostMetric_BoostMetric_HPP
+#define MLPACK_METHODS_BoostMetric_BoostMetric_HPP
 
 #include <mlpack/prereqs.hpp>
 #include <mlpack/core/metrics/lmetric.hpp>
@@ -49,11 +49,11 @@ namespace boostmetric /** BoostMetric Distance Learning. */ {
  * @tparam MetricType The type of metric to use for computation.
  */
 template<typename MetricType = metric::SquaredEuclideanDistance>
-class BOOSTMETRIC
+class BoostMetric
 {
  public:
   /**
-   * Initialize the BOOSTMETRIC object, passing a dataset (distance metric
+   * Initialize the BoostMetric object, passing a dataset (distance metric
    * is learned using this dataset) and labels. Initialization will copy
    * both dataset and labels matrices to internal copies.
    *
@@ -61,7 +61,7 @@ class BOOSTMETRIC
    * @param labels Input dataset labels.
    * @param k Number of targets to consider.
    */
-  BOOSTMETRIC(const arma::mat& dataset,
+  BoostMetric(const arma::mat& dataset,
        const arma::Row<size_t>& labels,
        const size_t k);
 
@@ -139,7 +139,7 @@ class BOOSTMETRIC
 
   //! Constraints instance.
   lmnn::Constraints<MetricType> constraint;
-}; // class BOOSTMETRIC
+}; // class BoostMetric
 
 } // namespace boostmetric
 } // namespace mlpack
--- a/src/mlpack/methods/boostmetric/boostmetric.hpp
+++ b/src/mlpack/methods/boostmetric/boostmetric.hpp
@@ -1,0 +1,150 @@
+/**
+ * @file boostmetric.hpp
+ * @author Manish Kumar
+ *
+ * Declaration of BoostMetric class.
+ *
+ * mlpack is free software; you may redistribute it and/or modify it under the
+ * terms of the 3-clause BSD license.  You should have received a copy of the
+ * 3-clause BSD license along with mlpack.  If not, see
+ * http://www.opensource.org/licenses/BSD-3-Clause for more information.
+ */
+#ifndef MLPACK_METHODS_BOOSTMETRIC_BOOSTMETRIC_HPP
+#define MLPACK_METHODS_BOOSTMETRIC_BOOSTMETRIC_HPP
+
+#include <mlpack/prereqs.hpp>
+#include <mlpack/core/metrics/lmetric.hpp>
+#include <mlpack/methods/lmnn/constraints.hpp>
+
+namespace mlpack {
+namespace boostmetric /** BoostMetric Distance Learning. */ {
+
+/**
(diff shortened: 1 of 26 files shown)


src/mlpack/methods/boostmetric/boostmetric.hpp
```diff
@@ -0,0 +1,150 @@
...
+ *   publisher = {Curran Associates, Inc.},
+ *   url = {http://papers.nips.cc/paper/
+ *       3658-positive-semidefinite-metric-learning-with-boosting.pdf}
+ * }
+ * @endcode
+ *
+ * @tparam MetricType The type of metric to use for computation.
+ */
+template<typename MetricType = metric::SquaredEuclideanDistance>
+class BOOSTMETRIC
```
rcurtin: Seems to me like we should call it `BoostMetric`.

src/mlpack/methods/boostmetric/boostmetric_impl.hpp
```diff
@@ -0,0 +1,179 @@
...
+    labels(labels),
+    k(k),
+    maxIter(500),
+    tolerance(1e-7),
+    wTolerance(1e-5),
+    constraint(dataset, labels, k)
+{ /* nothing to do */ }
+
+// Calculate inner product of two matrices.
+inline double innerProduct(arma::mat& Ar, arma::mat& Z)
```
rcurtin: Why not just use `arma::accu(Ar % Z)`?  (Actually `arma::dot()` might do the same thing also, but it's likely you might need a `vectorise()` call there too.)

src/mlpack/methods/boostmetric/boostmetric_impl.hpp
```diff
@@ -0,0 +1,179 @@
...
+  for (size_t i = 0; i < maxIter; i++)
+  {
+    // Initialize Acap.
+    Acap.zeros(dim, dim);
+    for (size_t j = 0; j < N; j++)
+      Acap += u(j) * A.slice(j);
+
+    arma::vec eigval;
+    arma::mat eigvec;
+    arma::eig_sym(eigval, eigvec, (Acap + trans(Acap) / 2));
```
rcurtin: I'd suggest timing this specific step also.  If we need we can modify Armadillo to use ARPACK to only get the largest eigenvalue, which can be a lot faster.

src/mlpack/methods/boostmetric/boostmetric_impl.hpp
```diff
@@ -0,0 +1,179 @@
...
+      if (wHigh - wLow < wTolerance || std::abs(lhs) < wTolerance)
+        break;
+    }
+
+    // Update u.
+    for (size_t j = 0; j < N; j++)
+      u(j) = u(j) * std::exp(-H(j) * w);
+
+    // Normalize u so that sum(u_i) = 1.
+    u /= arma::sum(u);
```
rcurtin: Do we end up with cases where `u` is either extremely small or 0?  In that case we could avoid adding it to the sum `Acap`.

src/mlpack/methods/boostmetric/boostmetric_impl.hpp
manish7294: I think it doesn't require any change, as we will be needing eigvector as well in very next calculation.
`arma::mat Z = eigvec.col(maxIndex(0)) * arma::trans(eigvec.col(maxIndex(0)));`

src/mlpack/methods/boostmetric/boostmetric_impl.hpp
```diff
@@ -0,0 +1,179 @@
...
+            << "; " << "terminating optimization." << std::endl;
+      break;
+    }
+
+    // Output current maximum eigen value.
+    Log::Info << "BoostMetric: iteration " << i << ", maximum eigen value "
+        <<  maxEig << "." << std::endl;
+
+    // Get index of maximum element.
+    arma::uvec maxIndex = arma::find(eigval == maxEig);
```
manish7294: I think we can at least avoid this calculation if we have something like index_max() but armadillo supports it starting from v 7.2000

src/mlpack/methods/boostmetric/boostmetric_impl.hpp
manish7294: Here's the simulation supporting this - https://gist.github.com/manish7294/01bb5b6d2f5c4cbdb60dc5af541e39dc
Let me know your thoughts on this.

src/mlpack/methods/boostmetric/boostmetric_impl.hpp
manish7294: That's much better.

src/mlpack/methods/boostmetric/boostmetric.hpp
manish7294: Sure

src/mlpack/methods/boostmetric/boostmetric_impl.hpp
rcurtin: It'll also take much longer for datasets with large dimensionality, so I'd be interested to see how long it takes for covertype (55d) and MNIST (784d).

src/mlpack/methods/boostmetric/boostmetric_impl.hpp
rcurtin: I see, it looks like this could be a good pruning opportunity if we have time (but it's not clear if this will provide much speedup to me, I'm not sure how much computation we save).

rcurtin:
Thanks for the hard work on this; I haven't had a time to review deeper yet, but do you have any comparisons with the current LMNN code?  (sorry to ask twice but I think that is one of the more important things we need to consider here)

manish7294:
No, worries! But I remember doing some simulations earlier, at least on a small dataset. Though I think I will do it again. I think this is what I was referring to - https://gist.github.com/manish7294/2388267666b1159ce261ce7b95dc923c

rcurtin:
Right, so just FYI, I'm waiting on more accurate comparisons for both small and large datasets with the current LMNN code, to get a better idea of how this performs before I do a more in-depth review.

manish7294:
I am really sorry for the delay, I just got way to busy with college work mainly due to industrial internships interviews. Hopefully, everything will be over by next week. After that, I will be able to work at full pace. Maybe I am requesting too much but do you think we can keep it on hold till then or I can try to take out some time?

manish7294:
Thanks a lot, I will be back as soon as everything is over :)

manish7294:
@rcurtin Here are some simulations I performed today - https://gist.github.com/manish7294/fc50da22451fd676d0ab0f4ccb4bc2a0

They might not be sufficient but I think they can atleast help in getting some insights.

While performing simulations on letters dataset I have noticed that max eigen value continuously kept on oscillating instead of getting minimized.

manish7294:
Ahh, finally after a whole year of wait, bot finally took it's call and marked this one as stale :)
So, I think we should also decide something for this @rcurtin.

rcurtin:
Yeah, this one fell off my list too, but I do think we should merge it in when we can.  Let me try and do a review in the near future.  I'll mark it 'keep open'. :+1:

PR #1492, opened 2018-08-13
Automatically Generated Go Bindings

yaswagner:
This pull request is an attempt at adding a Go binding generator to mlpack's codebase. The generated bindings include a C with C++ utility functions which can then be used by Go's cgo tool who lets Go packages call C code. The matrix used for the go bindings are from the gonum package (https://github.com/gonum).

The implementation of two parameters, precisely the matrix with dataset info parameter and the vectors of strings parameter, are still work in progress and are yet to be implemented.

GENERATING THE BINDING

First, to generate the bindings get mlpack in your go workspace:

	$  go get https://github.com/yaswagner/mlpack/

Then in your terminal go to the mlpack folder create your build directory and build cmake as desired.
For example:

	$ cd ${GOPATH}/src/mlpack.
	$ mkdir build
	$ cd build
	$ cmake -D BUILD_GO_BINDINGS=ON -D BUILD_PYTHON_BINDINGS=OFF ../
	$ make 

You can generate a specific binding to generate by using the mlpack_go_{method} target.
For instance, for the pca method:

	$ make mlpack_go_pca

USING THE BINDINGS

To use the bindings in Go, you must import the mlpack package and gonum package.

For example:

	import (
		"mlpack/build/src/mlpack/bindings/go/mlpack"
		"gonum.org/v1/gonum/mat"
	)

Optional parameters need to be set in the methods config struct and then we can set our desired parameters.
For instance, for pca, we can initialize the config struct called param and set verbosity to true by doing:

	param := mlpack.InitializePca()
	param.Verbose = true

We can then pass our optional parameters and call the Pca method as such:

	output := mlpack.Pca(input, param)

For output parameters we do not wish to use, we can use an underscore. 
For example, for the perceptron, if we wish to only use the 'output_model' and not 'output'"

	_, output_model := mlpack.Perceptron(param)

In order to get documentation about a method, we use 'godoc'.
For example, for the pca method we would type on the command line:

	godoc mlpack/build/src/mlpack/bindings/go/mlpack Pca

TESTING THE BINDINGS

Finally, to test the binding, Go's 'testing' is used. Simply go to the go test file following directory as such:

	$ cd ${GOPATH}/src/mlpack/src/bindings/go/tests/gofile

Then test by using the 'go test' command tool, as such:

	$ go test -v


src/mlpack/bindings/go/mlpack/cli_util.go
```diff
@@ -0,0 +1,150 @@
...
+func GetParamVecString(identifier string) []string {
+	e := int(C.MLPACK_VecStringSize(C.CString(identifier)))
+
+	var v MLPACK_VectorType
+	v.allocVecStringPtr(identifier)
+	runtime.GC()
+	time.Sleep(time.Second)
+
+	// Convert pointer to slice of data, to then pass it to a gonum matrix.
+	data := (*[1<<30 - 1]string)(v.mem)
```
rcurtin: I don't easily follow the `*[1<<30 - 1]string` here, is the idea that you are casting this to a very long string array?  If you say it's safe I trust you (since I am not a Go expert), but maybe it is better to convert it to a `double` pointer of the expected length?  Like `(*[v.n_elem]double)(v.mem)`?  I have no idea if that would work.

src/mlpack/bindings/go/print_cpp.cpp
```diff
@@ -0,0 +1,103 @@
...
+  }
+
+  for (ParamIter it = parameters.begin(); it != parameters.end(); ++it)
+  {
+    const util::ParamData& d = it->second;
+    if (d.input && !d.required &&
+        d.name != "help" && d.name != "info" &&
+        d.name != "version")
+      inputOptions.push_back(it->first);
+  }
```
rcurtin: Actually it looks like you are not using `inputOptions` or `outputOptions`, so I think you can remove this code.

src/mlpack/bindings/go/print_defn_output.hpp
```diff
@@ -0,0 +1,93 @@
...
+ * @param output Unused parameter.
+ */
+template<typename T>
+void PrintDefnOutput(const util::ParamData& d,
+                           const void* /* input */,
+                           void* /* output */)
+{
+
+  PrintDefnOutput<typename std::remove_pointer<T>::type>(d);
+
```
rcurtin: I think we could remove the extra empty lines here.  (I know it is a tiny style issue.)

src/mlpack/bindings/go/print_doc_functions_impl.hpp
```diff
@@ -0,0 +1,328 @@
...
+template<typename... Args>
+std::string ProgramCall(const std::string& programName, Args... args)
+{
+  std::string result = "";
+  std::string goProgramName = programName;
+  goProgramName[0] = std::toupper(goProgramName[0]);
+
+  // Initialize the method parameter structure
+  std::ostringstream oss;
+  oss << ">>> param := Initialize" << goProgramName << "()\n";
```
rcurtin: The `>>>` here was meant to imitate the Python REPL, but I don't know if Go has any equivalent thing.  You might consider just having a couple spaces indentation, i.e. `"  param := Initialize"` instead, but I'm not picky on how to do it.

src/mlpack/bindings/go/print_doc_functions_impl.hpp
```diff
@@ -0,0 +1,328 @@
...
+  return goModelName;
+}
+
+/**
+ * Given the name of a matrix, print it.  Here we do not need to modify
+ * anything.
+ */
+inline std::string PrintDataset(const std::string& datasetName)
+{
+  return "" + datasetName + "";
```
rcurtin: I guess this could just be `return datasetName`. :)

src/mlpack/bindings/go/print_doc_functions_impl.hpp
```diff
@@ -0,0 +1,328 @@
...
+/**
+ * Given the parameter name and an argument, return what should be written as
+ * documentation when referencing that argument.
+ */
+template<typename T>
+inline std::string ParamString(const std::string& paramName, const T& value)
+{
+  std::ostringstream oss;
+    oss << paramName << "=" << value;
+  return oss.str();
```
rcurtin: I think that this function will actually never be called, judging by all the calls to `PRINT_PARAM_STRING()` that I find in the `src/mlpack/methods/` folders.  So maybe it can be removed?

src/mlpack/bindings/go/print_go.cpp
```diff
@@ -0,0 +1,298 @@
...
+      inputOptions.push_back(it->first);
+  }
+
+  // First, we must generate the mlpack package name.
+  cout << "package mlpack" << endl;
+  cout << endl;
+
+  // Now we must print the cgo's import libraries and files.
+  cout << "/*" << endl;
+  cout << "#cgo CFLAGS: -I/capi -I. -g -Wall" << endl;
```
rcurtin: Should this be `-I./capi`?  Otherwise it would reference the `/capi` folder from the root of the filesystem.  Also we should probably remove `-g` at some point.

src/mlpack/bindings/go/print_go.cpp
```diff
@@ -0,0 +1,298 @@
...
+  cout << "#cgo LDFLAGS: -L/usr/local/lib/ -lm "
+      << "-lmlpack -L. -lmlpack_go_" << functionName << endl;
```
rcurtin: It may also be better to generate this automatically later from the CMake configuration, but that will make things more complex.  What do you think?

src/mlpack/bindings/go/print_go.cpp
```diff
@@ -0,0 +1,298 @@
...
+  for (ParamIter it = parameters.begin(); it != parameters.end(); ++it)
+  {
+    const util::ParamData& d = it->second;
+    if (d.input)
+      CLI::GetSingleton().functionMap[d.tname]["PrintModelUtilGo"](d, NULL, NULL);
+  }
+
+  // Print the comment describing the function and its parameters.
+  cout << "/*" << endl;
+  cout << "  " << HyphenateString(programInfo.documentation(), 2) << endl;
```
rcurtin: It's really awesome that the documentation printing works great!  That is not an easy thing given how different it is than the Python printing.  I did find one slight bug though; in the perceptron documentation, we get this:

```
  >>> param := InitializePerceptron()
  >>> param.Training = training_data
  >>> param.Labels = training_labels
  >>> perceptron := Perceptron(param)
~~
  Then, this model can be re-used for classification on the test data test_data.
   The example below does precisely that, saving the predicted classes to
  predictions.
~~
  >>> param := InitializePerceptron()
  >>> param.Input_model = perceptron
  >>> param.Test = test_data
  >>> predictions := Perceptron(param)
```

But I think, we also have to ignore the non-used return values, so it should be `_, perceptron := Perceptron(param)` then `predictions, _ := Perceptron(param)`.  I think it should not be too hard, just need to loop over the output options and see which ones aren't being used when printing the output documentation.  (Maybe it is harder than I think.  I am not sure, I haven't thought very hard about it...)

In any case the autogenerated documentation looks great otherwise!  That is one of the trickiest parts I had a lot of problems with for Python.  :rocket: * :+1: * :tada:

src/mlpack/bindings/go/print_h.cpp
```diff
@@ -0,0 +1,99 @@
...
+  }
+
+  for (ParamIter it = parameters.begin(); it != parameters.end(); ++it)
+  {
+    const util::ParamData& d = it->second;
+    if (d.input && !d.required &&
+        d.name != "help" && d.name != "info" &&
+        d.name != "version")
+      inputOptions.push_back(it->first);
+  }
```
rcurtin: Here also I think we are not using `inputOptions` or `outputOptions`.

src/mlpack/bindings/go/print_method_config.hpp
```diff
@@ -0,0 +1,148 @@
...
+    const util::ParamData& d,
+    const size_t indent,
+    const typename boost::disable_if<arma::is_arma_type<T>>::type* = 0,
+    const typename boost::enable_if<data::HasSerialize<T>>::type* = 0)
+{
+  const std::string prefix(indent, ' ');
+
+  std::string def = "nil";
+  if (std::is_same<T, bool>::value)
+    def = "false";
```
rcurtin: This is only enabled if the type has a `serialize()` function, so `bool` can't match here.  Also this seems to be the same code as the overload above, so you could change the SFINAE and use only one function (assuming they are exactly the same, I think they are):

```
template<typename T>
void PrintMethodConfig(
    const util::ParamData& d,
    const size_t indent,
    const typename boost::enable_if<arma::is_arma_type<T>>::type* = 0,
    const typename boost::enable_if<data::HasSerialize<T>>::type* = 0)
```

src/mlpack/bindings/go/print_method_init.hpp
```diff
@@ -0,0 +1,170 @@
...
+#include <mlpack/prereqs.hpp>
+#include "get_go_type.hpp"
+#include "strip_type.hpp"
+
+namespace mlpack {
+namespace bindings {
+namespace go {
+
+/**
+ * Print parameter with it's default value for a standard option type.
```
rcurtin: Tiny typo that jumped out at me, this should be `its`. Sorry for being pedantic. :)

src/mlpack/bindings/go/tests/gofile/go_binding_test.go
```diff
@@ -0,0 +1,351 @@
...
+
+import (
+	"mlpack/build/src/mlpack/bindings/go/mlpack"
+	"testing"
+
+	"gonum.org/v1/gonum/mat"
+)
+
+func TestRunBindingNoFlag(t *testing.T) {
+	t.Log("Test that when we run the binding correctly (with correct input parameters), we get the expected output.")
```
rcurtin: Can you make sure the lines are 80 characters or less please?  (I guess the automatic style checker doesn't check .go files.)

src/mlpack/bindings/go/tests/test_go_binding_main.cpp
```diff
@@ -0,0 +1,168 @@
+/**
```
rcurtin: I can't comment on the right file, so this comment is kinda off in nowhere, but do we need to include the file `src/mlpack/bindings/go/tests/gofile/gofile`?  It seems like it's a binary file, maybe a build artifact or something?

src/mlpack/bindings/go/mlpack/capi/arma_util.cpp
```diff
@@ -0,0 +1,251 @@
...
+  void *ptr = GetMemory(output);
+  return ptr;
+}
+
+/**
+ * Return the memory pointer of an Armadillo umat object.
+ */
+void *MLPACK_ArmaPtr_umat(const char *identifier)
+{
+  arma::Mat<double> output = arma::conv_to<arma::Mat<double>>::from(CLI::GetParam<arma::Mat<size_t>>(identifier));
```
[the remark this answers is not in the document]
Yashwants19: Hi @rcurtin right now on I am working on Go bindings. Here I am allocating space to `output` directly without passing it from `GetMemory()` is it a correct way or i am some blunder here. If you could help me with this comment. It would be great help to me. Thank You

rcurtin:
Closing in favor of #1884 (that's a logical continuation of this work).  Thanks again @yaswagner!

yaswagner:
Hi,

Yes this is the logical continuation. Sorry for not having finished them like i said i was going too. 
Hope all is well and that you had a nice year at your new job!

Best,

Yasmine

PR #1491, opened 2018-08-10
Improve Convolution Gradient() Method

ShikharJ:
I'll need to test for the outputs of the GANs. Then this can be merged.

PR #1493, opened 2018-08-13
Improve TransposedConvolution layer.

akhandait:


src/mlpack/methods/ann/layer/transposed_convolution.hpp
```diff
@@ -143,7 +153,7 @@ class TransposedConvolution
   OutputDataType& Gradient() { return gradient; }
 
   //! Get the input width.
-  size_t const& InputWidth() const { return inputWidth; }
+  size_t const& InputWidth () const { return inputWidth; }
```
ShikharJ: I think this change can be undone.

src/mlpack/tests/ann_layer_test.cpp
```diff
@@ -1448,32 +1448,30 @@ BOOST_AUTO_TEST_CASE(SimpleTransposedConvolutionLayerTest)
...
 
-  // Test the Backward function.
+  // Test the backward function.
   module1.Backward(std::move(input), std::move(output), std::move(delta));
-  BOOST_REQUIRE_EQUAL(arma::accu(delta), 720);
+  // Value calculated using tensorflow.nn.conv2d()
+  BOOST_REQUIRE_EQUAL(arma::accu(delta), 690.0);
 
-  TransposedConvolution<> module2(1, 1, 4, 4, 1, 1, 2, 2, 5, 5);
+  TransposedConvolution<> module2(1, 1, 3, 3, 1, 1, 1, 1, 6, 6, 6, 6);
```
ShikharJ: Please keep the original parameters here. These tests follow from the same Convolutional Aithmetic report, figure 4.1 and onwards.

src/mlpack/tests/ann_layer_test.cpp
```diff
@@ -1448,32 +1448,30 @@ BOOST_AUTO_TEST_CASE(SimpleTransposedConvolutionLayerTest)
...
-  input = arma::linspace<arma::colvec>(0, 24, 25);
+  input = arma::linspace<arma::colvec>(0, 35, 36);
   module2.Parameters() = arma::mat(16 + 1, 1, arma::fill::zeros);
   module2.Parameters()(0) = 1.0;
-  module2.Parameters()(3) = 1.0;
-  module2.Parameters()(6) = 1.0;
-  module2.Parameters()(9) = 1.0;
-  module2.Parameters()(12) = 1.0;
-  module2.Parameters()(15) = 2.0;
+  module2.Parameters()(8) = 2.0;
```
ShikharJ: Same here, we're testing for different variations.

ShikharJ:
@akhandait You can rebase to master now. #1491 is merged.

src/mlpack/tests/ann_layer_test.cpp
akhandait: @ShikharJ As far as I could figure out, `tf.nn.conv2d_transpose` doesn't allow size to go down(As it only allows VALID or SAME padding on the output size). In the above example, using the current formula for output size, it should be 4. Can you tell me what parameters(in the `tf.nn.conv2d_transpose` function) along with the below weights did you use to arrive at the value 2100?

ShikharJ: @akhandait I don't think the size should decrease, take a look at Figure 4.2, aren't you able to obtain an output size of 8? I don't remember the parameters from the top of my head, though I can derive them, just give me sometime, I need to reinstall tensorflow. I have sort of moved on to Pytorch these days.

akhandait: Sorry for the super late response. I understand the approach that was used. The output size calculated this way includes the padding added to the input of the equivalent convolution operation. However, most of the deep learning frameworks don't do that.(I tried Tensorflow, PyTorch and Thenao is exactly the same as our new implementation.) So, we will have to change the parameters. For example, in this case, the output size should actually be 6(as shown in the figure, 1 + 1 is the padding and 8 is not the size of the original input. I tested it in PyTorch.) In PyTorch and our new implementation, this case should have parameters like this:
kernel size = 4
stride = 1
padding = 1(what all deep learning frameworks do is take the padding of the equivalent convolution operation, so here padding of 1 was used which changed the actual input size 6x6 to 8x8. We should output 6 and not 8.)
inputHeight and Width = 5
outputHight and Width = 6
So, 5 = (6 + 2(1) - 4) / 1 + 1
But, as you correctly mentioned, we should test all the cases from that paper. I am going to shift to PyTorch for calculating the values for testing as Tensorflow unfortunately only allows 'SAME' and `VALID` padding which means we can't decrease size using transposed conv.

akhandait: @zoq Please verify this.

akhandait: @ShikharJ @zoq Can you please comment on this?

zoq: @akhandait Sorry for the slow response, I don't see any reason against the change; I did also check PyTorch and as you already said it's the same parameter set.

akhandait: @zoq Do you have any further comments on the PR? Can you give it a final review?

ShikharJ: @akhandait Your approach sounds reasonable to me, I was actually referring to the decreased kernel sizes which have now been fixed. I'll need to run the tests for the GANs and test their output, and cross check the tests as well. I'm planning on doing these over the weekend. I'll let you know what I find.

akhandait: @ShikharJ Sounds good. Thanks!

zoq:
Looks like the Transposed layer test failed.

akhandait:
@zoq It should pass now.

src/mlpack/tests/ann_layer_test.cpp
```diff
@@ -1347,63 +1350,66 @@ BOOST_AUTO_TEST_CASE(SimpleTransposedConvolutionLayerTest)
...
   BOOST_REQUIRE_EQUAL(arma::accu(output), 6000.0);
 
   // Test the backward function.
   module4.Backward(std::move(input), std::move(output), std::move(delta));
-  BOOST_REQUIRE_EQUAL(arma::accu(delta), 86208);
+  // Value calculated using torch.nn.functional.conv2d()
+  BOOST_REQUIRE_EQUAL(arma::accu(delta), 86208.0);
 
-  TransposedConvolution<> module5(1, 1, 3, 3, 2, 2, 0, 0, 5, 5);
+  TransposedConvolution<> module5(1, 1, 3, 3, 2, 2, 0, 0, 2, 2, 5, 5);
```
ShikharJ: Can you use the same input sizes here as well from the paper as done above?

ShikharJ:
I think these changes also need to be made in `gan_test.cpp` file.

ShikharJ:
@akhandait Sorry if I'm being unreasonable, but I'm unable to understand the rationale behind inserting the zeros on the input first and then applying convolution operation. Could you explain a bit on that? Also, I think since both the `ForwardConvolutionRule` and `BackwardConvolutionRule` are `valid`, I think it'd be better to use them in `Forward` and `Backward` functions respectively.

akhandait:
@zoq @ShikharJ Sorry for the inactivity on this one, I got busy with some other things. I will get back on this in a couple of days.

src/mlpack/tests/ann_layer_test.cpp
akhandait: Maybe there's been some confusion. They already are according to the paper. The above module 5 has input size 2 and output size 5 which follows 4.5 in the paper, similarly the modules 6 and 7 follow 4.6 and 4.7 respectively.

akhandait:
@ShikharJ Inserting zeros has been done to implement fractionally strided convolutions with the current convolution rules. Another way could have been to implement another convolution rule for fractionally strided convolution which performs it without actually adding zeros in the input. But this way for easier for the time being.

akhandait:
@ShikharJ Any updates here?

ShikharJ:
@akhandait Apologies, I'll get back on reviewing this and give you an update soon.

zoq:
@akhandait looks like there is a merge conflict, I can resolve this one if you like, just let me know. @ShikharJ Any update on this one from your side?

ShikharJ:
@zoq I had been primarily busy with CycleGAN recently, but I think this is close to completion, just need to check the output on the GANs.

akhandait:
I resolved it, let's check the tests now.

ShikharJ:
@akhandait This seems to have a bit of style issues, do you think you can rectify them?

akhandait:
Hey @ShikharJ, Jenkins isn't showing me where the checks have failed, it instead shows a Stack trace. Can you please restart the Jenkins build as I am not sure how to?

zoq:

akhandait:
@zoq @ShikharJ I can't understand the static checks warning here:
"The expression `kW - padW - 1 > 0` will work as `kW - padW != 1`." - line 211, 233
Why is it `!= 1`?
If they had said `> 1`, I wouldn't have worried about it, but I can't figure out this case.

rcurtin:
@akhandait I think both `kW` and `padW` are `size_t`, so this has to do with the fact that negative numbers can't be represented.  If `kW < padW` (which hopefully should not be the case) then `kW - padW` is some giant number greater than 0.  So, the only situation for which we can make the condition `kW - padW - 1 > 0` false is when `kW - padW - 1 == 0`, or, when `kW - padW == 1` (hence, the suggested condition of `kW - padW != 1`.  I think this is right, hope it helps. :)

akhandait:
@rcurtin Ahh I missed that, silly error. I will just cast it to `int`.

src/mlpack/methods/ann/layer/transposed_convolution.hpp
```diff
@@ -222,18 +214,23 @@ class TransposedConvolution
    * @param input The input to be padded.
    * @param wPad Padding width of the input.
    * @param hPad Padding height of the input.
+   * @param wExtra The number of extra zeros to the right.
+   * @param hExtra The number of extra zeros to the bottom.
    * @param output The padded output data.
    */
   template<typename eT>
   void Pad(const arma::Mat<eT>& input,
```
saksham189: Maybe we could introduce a `Padding` layer and use that inside the convolution layers to avoid redundant code.
[Pytorch already has support](https://pytorch.org/docs/stable/nn.html#padding-layers) for quite a lot of padding layers. This would also make the implementation more clear.

akhandait: @saksham189 Has there been any discussion on this on irc after you posted this? I think this might be good idea and should be given some thought.

zoq: Sorry for the slow response, makes totally sense to me.

saksham189: @akhandait I will work on adding the Padding layer and remove the redundant code in a new PR so that it is easy to review.

zoq:
Sorry for the slow response, happy to resolve the merge conflict, just let me know.

src/mlpack/methods/ann/layer/transposed_convolution.hpp
```diff
@@ -246,20 +243,81 @@ class TransposedConvolution
...
+          input.n_cols * dH - dH + 1);
+    }
+
+    for (size_t i = 0; i < output.n_rows; i += dH)
+    {
+      for (size_t j = 0; j < output.n_cols; j += dW)
+      {
+        // TODO: Use [] instead of () for speedup after this is completely
+        // debugged and approved.
+        output(i, j) = input(i / dH, j / dW);
```
zoq: Not sure using ``[]`` will bring us any speedup here, but maybe I missed something.

src/mlpack/methods/ann/layer/transposed_convolution.hpp
```diff
@@ -246,20 +243,81 @@ class TransposedConvolution
...
+   * Note: This function should be used before the Pad() function.
+   *
+   * @param input The input to be padded.
+   * @param dW Stride of filter application in the x direction.
+   * @param dH Stride of filter application in the y direction.
+   * @param output The padded output data.
+   */
+  template<typename eT>
+  void InsertZeros(const arma::Mat<eT>& input,
+                   size_t dW,
```
zoq: Maybe for consistency we should use ``const size_t``.

src/mlpack/methods/ann/layer/transposed_convolution.hpp
```diff
@@ -246,20 +243,81 @@ class TransposedConvolution
...
    * @param hPad Padding height of the input.
+   * @param wExtra The number of extra zeros to the right.
+   * @param hExtra The number of extra zeros to the bottom.
    * @param output The padded output data.
    */
   template<typename eT>
   void Pad(const arma::Cube<eT>& input,
            size_t wPad,
            size_t hPad,
+           size_t wExtra,
```
zoq: See comment below, looks like this applies to ``wPad, hPad`` as well.

src/mlpack/methods/ann/layer/transposed_convolution.hpp
```diff
@@ -246,20 +243,81 @@ class TransposedConvolution
...
+   * Note: This function should be used before the Pad() function.
+   *
+   * @param input The input to be padded.
+   * @param dW Stride of filter application in the x direction.
+   * @param dH Stride of filter application in the y direction.
+   * @param output The padded output data.
+   */
+  template<typename eT>
+  void InsertZeros(const arma::Cube<eT>& input,
+                   size_t dW,
```
zoq: See comment above.

src/mlpack/methods/ann/layer/transposed_convolution_impl.hpp
```diff
@@ -60,19 +60,21 @@ TransposedConvolution<
...
     dH(dH),
-    padW(padW),
-    padH(padH),
+    padW(kW - padW - 1),
+    padH(kH - padH - 1),
     inputWidth(inputWidth),
     inputHeight(inputHeight),
-    outputWidth(0),
-    outputHeight(0)
+    outputWidth(outputWidth),
```
zoq: We should update the serialisation as well.

src/mlpack/methods/ann/layer/transposed_convolution.hpp
akhandait: Isn't `[]` an optimized version of `()` without bounds check? It's mentioned in the Armadillo docs.

zoq: Right, this should only matter in debug mode:

ARMA_NO_DEBUG |   | Disable all run-time checks, such as bounds checking. This will result in faster code, but you first need to make sure that your code runs correctly! We strongly recommend to have the run-time checks enabled during development, as this greatly aids in finding mistakes in your code, and hence speeds up development. We recommend that run-time checks be disabled only for the shipped version of your program (ie. final release build).
-- | -- | --

akhandait: Oh, I didn't know that. I guess I'll leave it with `()` then, might actually be useful in debug mode if something goes wrong.

src/mlpack/methods/ann/layer/transposed_convolution_impl.hpp
akhandait: Sorry, but I didn't really get what you meant here.

src/mlpack/methods/ann/layer/transposed_convolution.hpp
```diff
@@ -247,6 +352,12 @@ class TransposedConvolution
   //! Locally-stored padding height.
   size_t padH;
 
+  //! Locally-stored number of zeros added to the right of input.
+  size_t aW;
+
+  //! Locally-stored number of zeros added to the right of input.
```
walragatver: Instead of `right`, I think it should be `top`.

src/mlpack/methods/ann/layer/transposed_convolution_impl.hpp
```diff
@@ -60,19 +60,21 @@ TransposedConvolution<
...
+    outputHeight(outputHeight)
 {
   weights.set_size((outSize * inSize * kW * kH) + outSize, 1);
   padding = new Padding<>(padW, padW, padH, padH);
```
walragatver: Over here @saksham189 has added a `padding` layer. But he has not used it anywhere. Can you use it while making these changes?

src/mlpack/methods/ann/layer/transposed_convolution_impl.hpp
```diff
@@ -119,8 +121,35 @@ void TransposedConvolution<
...
+
+  if (dW > 1 || dH > 1)
+  {
+    InsertZeros(inputTemp, dW, dH, inputExpandedTemp);
+
+    if (padW != 0 || padH != 0 || aW != 0 || aH != 0)
+      Pad(inputExpandedTemp, padW, padH, aW, aH, inputPaddedTemp);
+    else
+    {
+      inputPaddedTemp = arma::Cube<eT>(inputExpandedTemp.memptr(),
```
walragatver: I think we can optimize this code. But this will change after using the `padding` layer.

src/mlpack/methods/ann/layer/transposed_convolution_impl.hpp
rcurtin: Agreed, should we also serialize the members `aH` and `aW` in `serialize()`?  (Also, should we initialize them here?)

src/mlpack/methods/ann/layer/transposed_convolution.hpp
rcurtin: Looks like the Padding layer is done and merged; should we incorporate those changes here, or maybe in another PR after this is merged?  It may be worth opening an issue so we don't forget. :)

akhandait:
@rcurtin I have a very busy week ahead of me. I don't think I can incorporate the Padding layer before the release. Should I make the minor changes so that we can get the PR merged for the release, or just wait out this release and then merge the PR with the Padding layer?

rcurtin:
@akhandait no need to make a stressful week worse---maybe can you make the minor changes so we can merge this, then we can open an issue for the refactoring with the `Padding` layer, and that can happen later?  Let me know if that would work for you. :+1:

akhandait:
@rcurtin I guess that would be much better for me right now, thanks! I'll push the changes tomorrow.

src/mlpack/methods/ann/layer/transposed_convolution.hpp
```diff
@@ -65,6 +71,8 @@ class TransposedConvolution
    * @param padH Padding height of the input.
```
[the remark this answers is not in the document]
akhandait: I had added two notes above the @param listing to clarify this thing. Do you think that's enough? Or should we mention it for individual parameters?

src/mlpack/methods/ann/layer/transposed_convolution_impl.hpp
akhandait: I think we should do both of those things. Thanks for pointing out.

src/mlpack/methods/ann/layer/transposed_convolution.hpp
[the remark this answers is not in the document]
walragatver: I will say. Let's write it. If we look at some other layers then their @param listing is quite similar to this layer. So people (`like me`) may skip reading the upward note. I don't know whether my view is correct or not so it's totally up to you.

rcurtin:
@walragatver pointed out in IRC today that we shouldn't wait on the release any longer, as mlpack 3.1.1 downloads `ensmallen-latest.tar.gz`---which as of ensmallen 2.10.0 no longer works!  I agreed, so I've gone ahead and released 3.2.0 tonight (which doesn't have the same issue as 3.1.1 did with respect to the ensmallen version), and I've moved this to the 3.2.1 milestone.  Once this is finished and merged we can easily issue a patch release. :+1:

src/mlpack/methods/ann/layer/transposed_convolution_impl.hpp
```diff
@@ -280,9 +340,16 @@ void TransposedConvolution<
...
   ar & BOOST_SERIALIZATION_NVP(outputWidth);
   ar & BOOST_SERIALIZATION_NVP(outputHeight);
+  ar & BOOST_SERIALIZATION_NVP(aW);
+  ar & BOOST_SERIALIZATION_NVP(aH);
 
   if (Archive::is_loading::value)
+  {
     weights.set_size((outSize * inSize * kW * kH) + outSize, 1);
+
+    aW = (outputWidth + kW - 2 * padW - 2) % dW;
```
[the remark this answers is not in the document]
rcurtin: Ah, actually, this is a good point.  If we are serializing `aW` and `aH`, then we have to handle reverse compatibility.  (That's not hard but it's just a little bit tedious.)  Since as far as I can tell `aW` and `aH` are always computed, I'd suggest that we just remove the two lines `ar & BOOST_SERIALIZATION_NVP(aw);` and `ar & BOOST_SERIALIZATION_NVP(aH);` above.

rcurtin:
It looks like there is still a travis test failing, also.  If nobody has any time, I can try and debug it.

akhandait:
Sorry for stalling here. I am having issues building mlpack on my system, so it became harder to work on this as I need to wait for travis to build and then debug through those errors. I'll still try completing this today.

zoq:
Hopefully the build issues you see, are fixed with the latest ensmallen + mlpack version.

rcurtin:
@akhandait no problem, and I'm happy to help try and debug your system if you like, just let me know.  I think @zoq is right that the update of ensmallen might be the thing that makes a difference here; if you make sure the branch is rebased against master and then delete/remake your build directory, that should work out any issues (I hope).

akhandait:
@zoq It didn't :(. I am using ensmallen 2.10.3 and mlpack latest master(so rebase isn't an issue). I just posted the entire error on irc. Can you please see if it seems familiar? If not I will post the build config so that someone can reproduce it.

akhandait:
Hmm, one of the jobs in travis is still failing due to some reason. I don't think it's due to this PR though.

rcurtin:
You're right, it just timed out.  Let me try running it again and we'll see what happens...

akhandait:
I forgot to update `HISTORY.md` before the merge, could someone please do it? Or should I commit directly to the repository?

zoq:
Update ``HISTORY.md``in https://github.com/mlpack/mlpack/commit/7d677d0cd0460242bab28ee14aca1cd88def395d.

Issue #1490, opened 2018-08-03
LMNN: don't recompute impostors during an SGD batch


rcurtin:
I debugged the sometimes-hanging `LMNNTest/LMNNLowRankAccuracyBBSGDTest` today.  What I found was that it sometimes hangs because the objective function that is being optimized is actually changing during a single step, because `Evaluate()` and `Gradient()` may be called multiple times, and as a result the impostors may be recalculated and get different results, resulting in a different objective function.

In order to solve this I see a couple options:

1. Only recompute impostors when `Shuffle()` is called.
2. Track which points have had impostors recalculated since the last call to `Shuffle()`, and only recalculate impostors for a point if not.
3. Use some traits class to know at compile time which optimizers may make multiple calls to `Evaluate()` and `Gradient()` during each step.

@manish7294: what do you think, would you be willing to make these changes?  If we don't fix it somehow, techniques like BBSGD won't work with LMNN.

rcurtin:
I also commented out a test in 69ea3599 that suffers from the same issue.

manish7294:
Thanks for looking into this. I have few questions, maybe you can help me out here.

>  What I found was that it sometimes hangs because the objective function that is being optimized is actually changing during a single step, because `Evaluate()` and `Gradient()` may be called multiple times, and as a result the impostors may be recalculated and get different results, resulting in a different objective function.

I don't think I get it completely. Though there are multiple calls to `Evaluate()` and `Gradient()` in a single step but we always update iterate just once during the entire call. So, for one iteration impostors should be recalculated just once. At other times when iterate is not updated bounds should avoid recalculation.

> Only recompute impostors when Shuffle() is called.

I agree this could have been a good option, but I fear this would make learning while using linear scan ineffective. Moreover there is one more reason for avoiding recalulation of impostors during shuffle(), I will write it down in next lines.

> Track which points have had impostors recalculated since the last call to Shuffle(), and only recalculate impostors for a point if not.

In BBSGD, `Shuffle() `is called at very starting of an iteration and the update of iterate takes place after that, so recalculating impostors during `Shuffle()` is unproductive as we won't see any upadates in impostors.

> Use some traits class to know at compile time which optimizers may make multiple calls to Evaluate() and Gradient() during each step.

This could help if we find out a complementary solution for the problem.

Taking from your thoughts, I think problem could be shuffling of those points as well for which impostors recalculation is getting pruned by bounds. And as of now there is no code to handle that in `Shuffle()`

[...]

PR #1494, opened 2018-08-14
Recompute impostors only during Shuffle() call

manish7294:
see #1490


src/mlpack/methods/lmnn/lmnn_function_impl.hpp
```diff
@@ -96,34 +99,39 @@ LMNNFunction<MetricType>::LMNNFunction(const arma::mat& dataset,
...
-  math::ClearAlias(labels);
-
-  dataset = newDataset.cols(ordering);
-  labels = newLabels.cols(ordering);
-  maxImpNorm = newMaxImpNorm.cols(ordering);
-  lastTransformationIndices = newlastTransformationIndices.elem(ordering);
-  norm = newNorm.elem(ordering);
-
-  for (size_t i = 0; i < ordering.n_elem; i++)
+  if(!linearScan)
```
rcurtin: Tiny style issue---don't forget a space here please. :)

src/mlpack/methods/lmnn/lmnn_function_impl.hpp
```diff
@@ -351,37 +364,42 @@ double LMNNFunction<MetricType>::Evaluate(const arma::mat& transformation,
...
-        transformedDataset, labels, norm, points, numPoints);
-  }
-  else if (iteration++ % range == 0)
-  {
-    // Re-calculate impostors on transformed dataset.
-    constraint.Impostors(impostors, distance, transformedDataset, labels,
-        norm, begin, batchSize);
+      // Re-calculate impostors on transformed dataset.
+      constraint.Impostors(impostors, distance,
+          transformedDataset, labels, norm, points, numPoints);
```
rcurtin: If I am understanding right, I think this is only calculating the impostors for the individual batch.  But we could calculate impostors for _all_ points in the dataset since it is the beginning of the epoch.

I think that to this end, we could move the impostor recalculation code to its own method (since I think it should be the same across all `Evaluate()`, `Gradient()`, and `EvaluateWithGradient()` overloads... correct me if I am wrong). And then in each of these functions we could have just a simple snippet:

```
if (recalculate)
{
  RecalculateImpostors(...); // this function sets 'recalculate' to false internally
}
```

What do you think, could that work?

src/mlpack/methods/lmnn/lmnn_main.cpp
```diff
@@ -344,10 +344,12 @@ static void mlpackMain()
     LMNN<LMetric<2>> lmnn(data, labels, k);
     lmnn.Regularization() = regularization;
     lmnn.Range() = range;
+    lmnn.LinearScan() = shuffle;
     lmnn.Optimizer().StepSize() = stepSize;
     lmnn.Optimizer().MaxIterations() = passes * data.n_cols;
     lmnn.Optimizer().Tolerance() = tolerance;
-    lmnn.Optimizer().Shuffle() = shuffle;
+    // Bypass LinearScan. It will be handle in shuffle().
+    lmnn.Optimizer().Shuffle() = true;
```
rcurtin: I think it's `true` by default actually, so this line may not be necessary.  But it doesn't hurt anything.

src/mlpack/methods/lmnn/lmnn_function_impl.hpp
rcurtin: Actually `Shuffle()` is only called once, at the beginning of the epoch.  And typically it is also (slightly) faster to compute many impostors at once instead of doing them in small batches, due to the nature of the dual-tree nearest neighbor search algorithm.  Also if we compute them all at the beginning of the epoch, we can compute impostors with only one transformation matrix, which makes the handling of both the last transformation matrix easier and requires less computation to stretch the dataset (and the trees).  Also it makes the tree-related code I am working on a thousand times easier to write. ;)

So unless there's something I've overlooked, I think it would be a lot better to recompute all the impostors at once.  Let me know what you think.

manish7294: Then it's all good, earlier I thought that ``Shuffle()`` is being called during every batch. I will make the required changes.

manish7294:
@rcurtin I think it's ready for a review. Please check this out as you find some time.

rcurtin:
Thought before I go to bed (I think it is right): the use of `distance(l, i)` as a substitute for `metric.Evaluate(transformedDataset.col(i), transformedDataset.col(impostors(l, i)))` is currently used under the condition `iteration - 1 % range == 0`.  But this may not work right anymore, since impostors are only calculated in `Shuffle()`!  We can only use `distance()` as a speedup when the transformation matrix is the same as it was when impostors were calculated---so I wonder, maybe a replacement condition that could work is `if (transformationDiff == 0.0)`?  Let me know what you think.

rcurtin:
Here's a patch that seems to work: https://gist.github.com/rcurtin/ecde14f3f9e459e4fb5d756052738466

It's possible it could be cleaned up a little.  But it fixes the issue I described above.  Let me know what you think.

manish7294:
I applied the above patch you provided (It should definitely solve the problem of using distance()), but as you earlier mentioned bbsgd still seems to fail. Looks like there could some other problem behind it, probably something wrong is going on with pruning in the case of shuffling and that too just with bbsgd. Need to do some more dig up.

manish7294:
@rcurtin Sorry but still I couldn't find as why the tests are still failing. But I have observed that the `stepSize` is quickly reducing to 0 and as a result `bbsgd` is getting stuck into an infinite loop, though I strongly believe this is the cause of our problem but can't get as to why this is happening.

rcurtin:
I'll see if I can check into this tonight or soon.  This week has been extremely busy for me so I wasn't able to even run LMNN simulations... :(

On my system the patch was working, so either I'm not running the same examples as you or maybe I prepared the patch wrong.

rcurtin:
I think you're right.  I see in my copy of the code, I added `math::RandomSeed(std::time(NULL))` to the L-BFGS test, not the BBSGD test (oops!), and when I add it to the BBSGD test, I am able to get the test to hang.  Let me have a little while to think about what might be going wrong...

rcurtin:
I see what the issue is---when we compute the bounds `evalOld[...]`, we are currently using the difference between the current transformation matrix and the last transformation matrix used to compute impostors.  Instead, we have to use the difference between the current transformation matrix and the last transformation matrix used to compute the `evalOld` bounds!

I am working on a patch now.

manish7294:
Hi @rcurtin, Hope you are doing good. It's great to see all the activity and improvements over the past month. 
I was wondering did you happen to work over this. If not, then maybe I can work over this weekend (just please give me some insights over your finding).

rcurtin:
Hey @manish7294, sorry for the lack of update on this.  That night when I posted the comment, I actually found that my patch did not work.  I went to dig into why, but then I became consumed with various other issues and have not had a chance to return to this.  If you'd like to try and pick up where I left off debugging, here is the patch that I had:

https://gist.github.com/rcurtin/f9b511fc54f2cf1065e9ca4d54a3d1fa

The patch was getting pretty comprehensive---since we only recalculated impostors at the beginning of each epoch, I was able to remove some code.  However there is still something wrong in what I am doing, and I have not dug to the bottom of what yet.  Right now there is a random seed hardcoded into the test, but it's likely on your system you might need to find a new seed that fails (failure is obvious---it's just an infinite loop) and then hardcode that one.

Unfortunately since it's been a month I don't have much other memory of exactly where I was.  You could also ignore the patch and start debugging from the code as it is now on your branch, and maybe use the patch to give some ideas.  Anything could work :)

I've been hoping to have some time to return to this for quite a while now but it hasn't happened yet.

manish7294:
Hi Ryan, After a long period, I finally got a chance to rework this PR. 
Probably, since it's been so long, you might as well have faint idea of the issue (at least it was in my case :) ) 
So I am writing down a gist of the issue, we were facing:
The issue was related to working of `bigbatch_sgd` with LMNN. And as per your findings in #1490 , It was because the objective function being optimized was actually changing during single step of `bbsgd` due to multiple `Evaluate()` and `Gradient()` calls, which in result was changing objective function because of possible multiple recalculation of impostors. So, as a solution to this problem, It was decided to call recalculate impostors only once during a complete epoch (mainly during a shuffle() call, which takes place at start of an epoch). Though this PR implements the same but still in the end we were facing the issue of `bbsgd` getting stuck while optimization. Hopefully this will help in recalling things a bit :)

Now, here is the current update on the situation:
The `bbsgd` is actually getting stuck in the backtracking line search. The thing is we are doing recalculation of impostors only once at start of the epoch, that means our objective function remains same but because of this we are always calling ```function.Evaluate(iterateUpdate, offset,    backtrackingBatchSize)``` on same set of impostors, and in case the stepsize get too small (as per my observation, it eventually becomes 0 after some number of steps) -- we will just be calling Evaluate() on the same `iterateUpdate` i.e. we are stuck inside an infinte loop. And this condition has great chance of occurrence as after every step, `stepSize` will be reduced by a factor of `backtrackStepSize` leading to negligible change in  `overallObjectiveUpdate` value after each step i.e there will be a high possiblity of  ` (overallObjectiveUpdate > (overallObjective + searchParameter * stepSize * gradientNorm))` being always true, resulting in an infinite loop.

   ```cpp
 while (overallObjectiveUpdate >
        (overallObjective + searchParameter * stepSize * gradientNorm))
    {
      stepSize *= backtrackStepSize;


      iterateUpdate = iterate - (stepSize * gradient);
      overallObjectiveUpdate = function.Evaluate(iterateUpdate, offset,
          backtrackingBatchSize);
    }
```
 Moreover, I think that when we are Evaluating the function value, we should be doing that on transformed space (i.e. re-calculating the impostors for transformed space).

In short, as per my findings, we should avoid recalculation of impostors during `Gradient()`(to avoid change of objective function) and should recalculate impostors during `Evaluate()` (to get optimal iterate value while backtracking). 

Let me know what you think of this.

**Comment by rcurtin** [comment]
Oh wow, thanks for the detailed investigation. :+1: It has been a while since I thought about this but I do remember the issue and it is still on my list (unfortunately it just got pushed down a bunch).  Let me read through this in the next day or two and respond.

**Comment by manish7294** [comment]
Hi Ryan, did you get any chance to check this out?

**Comment by rcurtin** [comment]
Hey @manish7294, actually I did not get to it just now because I was spending all my time working on mlpack.org#6 (feel free to take a look at the new website and let me know what you think or if you find any problems, but don't feel obligated :)).

Do you want to try recalculating the impostors during `Evaluate()`, just to see if it works?

I have some concern that it won't help though, because the backtracking line search in BBSGD should depend on the function having two properties:

 (a) f(x) is differentiable everywhere
 (b) f(x) is continuous

but I feel that if we recompute impostors for different x, the function f(x) is actually discontinuous (its value changes at those points where impostors change).  So actually I would expect the recalculation of the impostors to make the problem worse (or at least not fix it)... but I could be wrong in my thinking here.  I haven't worked out the math to be sure.  Personally for this issue I had been suspecting something more along the lines of an implementation bug or something like this, but last time I looked through it I wasn't able to immediately find one.

Actually now that I read through this code that you posted... something doesn't quite add up in `adaptive_stepsize.hpp`.  Take a look through the Wikipedia definition of backtracking line search: https://en.wikipedia.org/wiki/Backtracking_line_search .  That's basically what is implemented in `adaptive_stepsize.hpp`.

To convert notation, see that `\alpha` in the Wikipedia page corresponds to `stepSize`.  Then, our search direction `p` is exactly the negative gradient, so therefore `m` would be equal to `-gradientNorm`.  And, `c` would correspond to `searchParameter`.  Thus, in the language of the Wikipedia page, we would expect `t = -cm`, or, rather, `t = -(searchParameter * -gradientNorm) = searchParameter * gradientNorm`.

The condition listed in the code then, is to terminate when

```
overallObjectiveUpdate <= (overallObjective + searchParameter * stepSize * gradientNorm)
```

That's equivalent to this:

```
(overallObjective - overallObjectiveUpdate) >= (-searchParameter * gradientNorm) * stepSize
```

and then converting to the language of the Wikipedia page, that gives:

```
f(x) - f(x + \alpha_j p) >= -(t) * \alpha
```

but... that's backwards!  It should be `t * \alpha` not `-t * \alpha`.  So... here's something to try (maybe the first thing to try): make line 156 of `adaptive_stepsize.hpp` read `overallObjective - searchParameter * stepSize * gradientNorm` (instead of `+`) and see if that fixes convergence.  If it does, we can talk to Marcus and see if he agrees it's an issue, and if so, we can incorporate it as a fix into ensmallen.

As always be sure to double-check my work... I make no guarantees and my weakness is forgetting +/- signs. ;)

manish7294:
I agree with you that on changing the impostors, function may not remain continuous & differentiable. And that will break very concept of line search. And you are also right about +/- sign. Moreover on digging deeper into big batch sgd paper, I found out there were several other issues as well with the implementation. For which I will be opening a PR shortly. After that we can check again.

manish7294:
After digging in I reached to this - https://hastebin.com/okeluriban.coffeescript
from this, it can bee seen that objective function is continuously increasing instead of decreasing. So, it makes me wonder whether there is some kind of error with our objective function implementation.
Let me know your thoughts on this?

rcurtin:
That just looks like a step size that's too large... what happens if you drop the step size by an order of magnitude (or more)?  Often with SGD-type optimizers that might have to be tuned.

manish7294:
For this case (on SGD) things work for at most order of 1e-4 & for bbsgd at most 1e-6, Moreover after optimization accuracy get decrease instead of increasing. Do you think it something that can be expected?

rcurtin:
It's not surprising that there is a maximum step size for which SGD will work.  But I am not sure what you mean with the rest of the comment.  Can you clarify or provide any output?

manish7294:
Like for bbsgd I got:
[INFO ] Accuracy on initial dataset: 75.8454%
[INFO ] Accuracy on transformed dataset: 63.285%

Similarly on SGD, I got
[INFO ] Accuracy on initial dataset: 75.8454%
[INFO ] Accuracy on transformed dataset: 75.3623%

So, I mean to say that it's lot dependent on what stepSize and initial transformation matrix (set  by seed) we are choosing.

>  Do you think it something that can be expected?

  By this I mean, I am not able to decide whether this is something  to be considered as a bug or not. I hope it clear things up.

rcurtin:
Hey @manish7294, I was thinking about it---I wonder if we could get better results here just by switching away from BBSGD?  Specifically, maybe we could try SPaLERA: http://ensmallen.org/docs.html#spalera-stochastic-gradient-descent-spalerasgd

I feel kind of stupid---I think I should have suggested this when we initially encountered the problem.  The idea was to use BBSGD because it uses adaptive batch sizes and is thus "smarter" than SGD in some way.  That was the reason I chose it.  But this is also true of lots of other optimizers, so SPALeRA could work too.  (And if that doesn't, maybe we could try some other ones.)

I'm no longer sure that anything is wrong in the code---I think maybe the code is just fine, but BBSGD does not work with it for some reason.  In any case, I think that neither of us have the time to really dig into this, so maybe we can switch it out for something that works?  Let me know what you think. :+1:

manish7294:
Yeah, I will be more than happy to try this out. Yeah, you are right, everytime we try finding out time for this something comes up and we find ourselves get busy with that and I think this will always be the case. :)
So, I will try to find some time to do this over the next 3-4 days.

rcurtin:
This one is still important enough for me to try and get figured out, but I just haven't had a chance yet.  So I'll mark it "keep open".

PR #1495, opened 2018-08-14
[WIP] Add Concatenate layer.

akhandait:


src/mlpack/methods/ann/layer/concatenate.hpp
```diff
@@ -0,0 +1,122 @@
...
+>
+class Concatenate
+{
+ public:
+  /**
+   * Create the Concatenate object using the specified number of output units.
+   *
+   * @param conat The matrix to be concatenated to the input.
+   */
+  Concatenate();
```
zoq: I think it might be a good idea, to provide a constructor that takes the concat matrix. What do you think?

src/mlpack/tests/ann_layer_test.cpp
```diff
@@ -1083,6 +1083,56 @@ BOOST_AUTO_TEST_CASE(GradientConcatLayerTest)
   BOOST_REQUIRE_LE(CheckGradient(function), 1e-4);
 }
 
+/**
+ * Concatenate layer numerical gradient test.
+ */
+BOOST_AUTO_TEST_CASE(GradientConcatenateLayerTest)
```
zoq: Can we add another simple test that, that takes a matrix and just checks the output?

akhandait: Yes, I had uploaded just one test because I wanted to open this PR before the deadline, I will add a test soon.

src/mlpack/methods/ann/layer/concatenate.hpp
akhandait: I was thinking about that, but I did this because every batch will have a different set of labels. We can't just have one matrix for this layer. This way, we can use train this using partial Forward(), Backward(), and Update() functions and change the labels during each pass. But I get it now, for other use cases, we should have a constructor for that as well.

src/mlpack/methods/ann/layer/concatenate.hpp
```diff
@@ -0,0 +1,122 @@
...
+    typename InputDataType = arma::mat,
+    typename OutputDataType = arma::mat
+>
+class Concatenate
+{
+ public:
+  /**
+   * Create the Concatenate object using the specified number of output units.
+   *
+   * @param conat The matrix to be concatenated to the input.
```
zoq: Looks like we are going with the ``Concat()``function, so we can remove the ``@param conat The matrix to be concatenated to the input.``.

src/mlpack/methods/ann/layer/concatenate_impl.hpp
```diff
@@ -0,0 +1,59 @@
...
+  if (concat.is_empty())
+    Log::Warn << "The concat matrix has not been provided." << std::endl;
+
+  if (input.n_cols != concat.n_cols)
+  {
+    Log::Fatal << "The number of columns of the concat matrix should be equal "
+        << "to the number of columns of input matrix." << std::endl;
+  }
+
+  inRows = input.n_rows;
```
zoq: Should we serialize the ``inRows`` parameter?

src/mlpack/methods/ann/layer/concatenate.hpp
akhandait: Oh, sorry I forgot to change this when you mentioned last time.

src/mlpack/methods/ann/layer/concatenate_impl.hpp
akhandait: Don't we serialize only properties of a layer which remain constant for all input? `inRows` varies for each forward pass, it's just something we store because it's needed in the subsequent backward pass.

zoq: Ah, you a right, this could change over time, I thought it remains the same.

akhandait:
@zoq Any further comments on this?

PR #1497, opened 2018-08-18
CycleGAN Module

ShikharJ:


src/mlpack/methods/ann/gan/cyclegan_impl.hpp
```diff
@@ -0,0 +1,613 @@
...
+
+  counterX = counterY = 0;
+  currentBatch = 0;
+
+  this->discriminatorX.deterministic = this->generatorY.deterministic = true;
+  this->discriminatorY.deterministic = this->generatorX.deterministic = true;
+
+  responsesX.set_size(1, trainDataX.n_cols);
+  responsesY.set_size(1, trainDataY.n_cols);
+  // TODO: Initialize responses
```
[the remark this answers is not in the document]
ShikharJ: @zoq I'm unsure how should we set the responses. For example in the case of `GAN`, we set the real data to 1 and generated data to 0.

src/mlpack/methods/ann/gan/cyclegan_impl.hpp
```diff
@@ -0,0 +1,613 @@
...
+    generatorY.Gradient(generatorY.parameter, 0, generatedGradientGY, batchSize);
+    generatorX.Gradient(generatorX.parameter, 0, generatedGradientGX, batchSize);
+
+    gradientGeneratorY += generatedGradientGY;
+    gradientGeneratorX += generatedGradientGX;
+    gradientGeneratorY *= multiplier;
+    gradientGeneratorX *= multiplier;
+  }  
+
+  counterX++;
```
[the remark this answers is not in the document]
ShikharJ: @zoq This is not required internally anywhere, but it helps in keeping track of the mini-batch number being processed at the current time. I don't think it would lead to a wrong result, because if the batch size changes, the `counterX * batchSize >= numFunctionsX` will regulate itself.

src/mlpack/tests/cyclegan_test.cpp
```diff
@@ -0,0 +1,187 @@
...
+using namespace std::placeholders;
+
+BOOST_AUTO_TEST_SUITE(CycleGANNetworkTest);
+
+/*
+ * Tests the CycleGAN implementation on MNIST to SVHN conversion.
+ * It's not viable to train on bigger parameters due to time constraints.
+ * Please refer mlpack/models repository for the tutorial.
+ *
+BOOST_AUTO_TEST_CASE(CycleGANMNISTToSVHNTest)
```
zoq: Do you think we could add a really simple test for the Cycle GAN?

ShikharJ: I will have to search up a bit, but as far as I have seen, this really is the simplest and least resource intensive test that I have encountered.

zoq: Do you have the travis timings for the test? Perhaps we can comment this one in?

ShikharJ: I'm yet to test this out on the full dataset, I got busy with formulating the ideas for the paper, so this had to wait.

zoq: Sure, no worries.

zoq:
I see you pushed something, but I'm not sure what the actual state of this is, do you need any help, I think it would be great to merge this one at some point.

ShikharJ:
Yeah, I'm setting up the datasets and running the tests, so every once in a while, I find the need to update this PR in a certain way. Shouldn't take too long.

rcurtin:

PR #1500, opened 2018-08-24
Replace linear regression implementation with something faster.

rcurtin:
Basically what I have done here is that following Anisa's email, I identified that linear regression actually creates an NxN matrix.  I think the old implementation scaled as O(N^2 d).  I then spent a little time at a whiteboard thinking about how to avoid that, and only use O(d^2) memory.  That resulted in this implementation, which scales as O(Nd^2 + d^3) where the O(Nd^2) comes from the covariance matrix computation and the O(d^3) comes from the `arma::solve()` call.

On a simple low-dimensional example with lots of points, the implementation scales much better.  I tested with a 200k point subset of the higgs dataset (28d), and the actual computation was 0.04s (whereas the loading time was like 1.4s).

Maybe for very high-dimensional datasets this implementation could be somewhat problematic, but at that point I think that using the optimization framework to do SGD would be better, or something like that.


rcurtin:
Ok, I will go ahead and merge this then.

PR #1502, opened 2018-08-24
Use 2 neighbors for LMNNAccuracyTest because 1 is unstable.

rcurtin:
It can cause unexpected changes in the impostor, which inverts the gradient; or, by only choosing one true neighbor, the gradient will be biased towards that one neighbor.  This can cause LMNN to fail to converge to a better solution.  If we use two impostors and neighbors it is much more robust.

So, with this fix, the LMNNAccuracyTest should not randomly fail anymore.  I tested with ~200 different random seeds and saw no failures, whereas before it would fail with ~half of all random seeds.

@manish7294: let me know if I overlooked anything.


manish7294:
Sounds fair. Thanks for looking into this :)

PR #1503, opened 2018-09-01
Added reward clipping wrapper

sshkhr:
Adds a fix for #1332 by introducing a wrapper around existing environments for reward clipping

TODO (as suggested in #1333 by @zoq ) 

- [x] Add test to show training converges Q-Learning on a simple task with reward clipping
- [x] Add test to check if the clipped reward is equal to the predetermined values of the test case (as in gradient_clipping_test)

### DIFF
--- a/src/mlpack/methods/reinforcement_learning/environment/reward_clipping.hpp
+++ b/src/mlpack/methods/reinforcement_learning/environment/reward_clipping.hpp
@@ -76,8 +76,8 @@ class RewardClipping
   }
 
   /**
-   * Dynamics of Continuous Mountain Car. Get reward and next state based 
-   * on current state and current action.
+   * Dynamics of Environment. The rewards returned from the base environment
+   * are clipped according the maximum and minimum values specified.
    *
    * @param state The current state.
    * @param action The current action.
@@ -88,12 +88,10 @@ class RewardClipping
                 const Action& action,
                 State& nextState) const
   {
-    // Get original unclipped reward from base environment
+    // Get original unclipped reward from base environment.
     double unclippedReward =  environment.Sample(state, action, nextState);
-    // Clip rewards according to the min and max limit
-    double clippedReward = std::min(
-        std::max(unclippedReward, minReward), maxReward);
-    return clippedReward;
+    // Clip rewards according to the min and max limit and return.
+    return std::min(std::max(unclippedReward, minReward), maxReward);
   }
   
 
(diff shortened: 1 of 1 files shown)


sshkhr:
@zoq could you help out regarding how to integrate this into the existing q-learning framework ?

The q-learning constructor expects an environment without any arguments but I need to pass an object of the base environment class to the wrapper in it's current form. I think changing the q-learning code may not be the right idea. Any help is appreciated

src/mlpack/methods/reinforcement_learning/environment/reward_clipping.hpp
```diff
@@ -0,0 +1,150 @@
...
+   * @param state The current state.
+   * @param action The current action.
+   * @param nextState The next state.
+   * @return clippedReward, Reward clipped between [minReward, maxReward].
+   */
+  double Sample(const State& state,
+                const Action& action,
+                State& nextState) const
+  {
+    // Get original unclipped reward from base environment
```
zoq: Picky style comment; missing point at the end.

src/mlpack/methods/reinforcement_learning/environment/reward_clipping.hpp
```diff
@@ -0,0 +1,150 @@
...
+    double unclippedReward =  environment.Sample(state, action, nextState);
+    // Clip rewards according to the min and max limit
```
zoq: See comment above.

src/mlpack/methods/reinforcement_learning/environment/reward_clipping.hpp
```diff
@@ -0,0 +1,150 @@
...
+    double clippedReward = std::min(
+        std::max(unclippedReward, minReward), maxReward);
+    return clippedReward;
```
zoq: If you like, we could directly return the clipped reward:

```c++
return std::min(std::max(unclippedReward, minReward), maxReward);
```

src/mlpack/methods/reinforcement_learning/environment/reward_clipping.hpp
```diff
@@ -0,0 +1,150 @@
...
+ private:
+  //! An instance of the UpdatePolicy used for actual optimization.
+  EnvironmentType environment;
+
+  //! Minimum possible value of clipped reward.
+  double minReward;
+
+  //! Maximum possible value of clipped reward.
+  double maxReward;
+
```
zoq: Looks like we could remove the empty lines here.

src/mlpack/methods/reinforcement_learning/environment/reward_clipping.hpp
```diff
@@ -108,12 +106,7 @@ class RewardClipping
   double Sample(const State& state, const Action& action) const
   {
     State nextState;
-    // Get original unclipped reward from base environment
-    double unclippedReward =  environment.Sample(state, action, nextState);
-    // Clip rewards according to the min and max limit
-    double clippedReward = std::min(
-        std::max(unclippedReward, minReward), maxReward);
-    return clippedReward;
+    return Sample(state, action, nextState);
```
zoq: Always nice to reduce existing code.

Issue #1504, opened 2018-09-04
BUG: The BatchNorm layer is not deserialized well.


zsogitbe:
I was saving and then loading a trained ann network while testing and I have noticed that the network is serialized well but when deserialized the BatchNorm layer is not loaded well. After loading, the beta and gamma vector sizes were still the default (10) and the values were not loaded.

zsogitbe:
I think that you will most probably need to do 'if (Archive::is_loading::value)' and 'set_size' for 'beta' and 'gamma' in the BatchNorm layer serialization! They are not deserialized/loaded well because the size is still the default... (the serialization/save is OK!). I am not sure how to get the right size at that place in the code because I am still new to mlpack but I am sure that more experienced users will know it immediately. I hope this helps!

zsogitbe:
This pull request does not work either. I would like to politely suggest that you always test the code before committing a pull request in order to prevent the accidental degradation of the library.
I have also noticed that you have removed the serialization of beta and gamma entirely. Are you sure that you know what you are doing? 
I do not have time to help you further with this issue because I am new to the library and I am just testing it.

ShikharJ:
Hmm, weird. @rcurtin Can you think of a reason why the serialization might not be working?

rcurtin:
> Are you sure that you know what you are doing?

I'm sure that @ShikharJ knows what he is doing---he has made substantial contributions to the library in the past year.

In this situation Shikhar is right: the `gamma` and `bias` matrices are simply aliases on top of the `weights` matrix.  The ANN code is complex so it's not necessarily easy to read, and the way that the network interacts with each layer is also complex.  For the sake of serialization, it turns out that the network itself holds all the weights in each layer as a contiguous matrix and that will be serialized, so we don't need to do that in the layer itself.

However it seems like maybe there are some other elements that aren't being serialized: `size`, `eps`, and `stats` jump out at me as possibilities.  Another possibility is that `stats` is being reset in `Reset()`, but for the batch normalization layer we should not be resetting that information when we serialize/deserialize the matrix.  Instead I think that should only happen at initialization time.

@zsogitbe: you're right that the PR that Shikhar proposed is not tested.  (If it was, you would not have encountered the bug at all.  But perfect testing coverage is not necessarily possible with such complex software...)  I think that a nice thing to do might be to work in your failing code as a test case, if you can supply simple code to build the network and then the code after serialization that fails.

zsogitbe:
[...]

PR #1505, opened 2018-09-04
Improvements to BatchNorm

ShikharJ:
Relevant: #1504 
@zsogitbe Can you confirm if this fixes your issue?

### DIFF
--- a/src/mlpack/methods/ann/layer/atrous_convolution_impl.hpp
+++ b/src/mlpack/methods/ann/layer/atrous_convolution_impl.hpp
@@ -96,10 +96,8 @@ void AtrousConvolution<
     OutputDataType
 >::Reset()
 {
-    weight = arma::cube(weights.memptr(), kW, kH,
-        outSize * inSize, false, false);
-    bias = arma::mat(weights.memptr() + weight.n_elem,
-        outSize, 1, false, false);
+  weight = arma::cube(weights.memptr(), kW, kH, outSize * inSize, false, false);
+  bias = arma::mat(weights.memptr() + weight.n_elem, outSize, 1, false, false);
 }
 
 template<
--- a/src/mlpack/methods/ann/layer/batch_norm_impl.hpp
+++ b/src/mlpack/methods/ann/layer/batch_norm_impl.hpp
@@ -139,8 +139,10 @@ template<typename Archive>
 void BatchNorm<InputDataType, OutputDataType>::serialize(
     Archive& ar, const unsigned int /* version */)
 {
-  ar & BOOST_SERIALIZATION_NVP(gamma);
-  ar & BOOST_SERIALIZATION_NVP(beta);
+  ar & BOOST_SERIALIZATION_NVP(size);
+
+  if (Archive::is_loading::value)
+    weights.set_size(size + size, 1);
 }
 
 } // namespace ann
--- a/src/mlpack/methods/ann/layer/convolution_impl.hpp
+++ b/src/mlpack/methods/ann/layer/convolution_impl.hpp
@@ -91,10 +91,8 @@ void Convolution<
     OutputDataType
 >::Reset()
 {
-    weight = arma::cube(weights.memptr(), kW, kH,
-        outSize * inSize, false, false);
-    bias = arma::mat(weights.memptr() + weight.n_elem,
-        outSize, 1, false, false);
+  weight = arma::cube(weights.memptr(), kW, kH, outSize * inSize, false, false);
+  bias = arma::mat(weights.memptr() + weight.n_elem, outSize, 1, false, false);
 }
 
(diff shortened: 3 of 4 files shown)


rcurtin:
Here's a test case we can add:

```
/**
 * Simple serialization test for batch normalization layer.
 */
BOOST_AUTO_TEST_CASE(BatchNormSerializationTest)
{
  arma::mat input(5, 100, arma::fill::randu);
  arma::mat output(5, 100, arma::fill::randu);

  FFN<NegativeLogLikelihood<>, RandomInitialization> model;
  model.Add<Linear<>>(input.n_rows, 10);
  model.Add<BatchNorm<>>();
  model.Add<ReLULayer<>>();
  model.Add<Linear<>>(10, 10);
  model.Add<ReLULayer<>>();
  model.Add<Linear<>>(10, output.n_rows);
  model.Add<LogSoftMax<>>();

  optimization::StandardSGD opt(0.1, 1, 5, -100, false);
  model.Train(input, output, opt);

  arma::mat originalOutput;
  model.Predict(input.col(0), originalOutput);

  // Now serialize the model.
  FFN<NegativeLogLikelihood<>, RandomInitialization> xmlModel, textModel,
      binaryModel;
  SerializeObjectAll(model, xmlModel, textModel, binaryModel);

  // Ensure that predictions are the same.
  arma::mat modelOutput, xmlOutput, textOutput, binaryOutput;
  model.Predict(input.col(0), modelOutput);
  xmlModel.Predict(input.col(0), xmlOutput);
  textModel.Predict(input.col(0), textOutput);
  binaryModel.Predict(input.col(0), binaryOutput);

  CheckMatrices(originalOutput, modelOutput, 1e-5);
  CheckMatrices(originalOutput, xmlOutput, 1e-5);
  CheckMatrices(originalOutput, textOutput, 1e-5);
  CheckMatrices(originalOutput, binaryOutput, 1e-5);
}
```

I think it could be generalized so that we check every layer, but I haven't done it here.  Sure enough in the master branch the test fails:

```
$ bin/mlpack_test -t ANNLayerTest
Running 60 test cases...
unknown location(0): fatal error: in "ANNLayerTest/BatchNormSerializationTest": memory access violation at address: 0x00000000: no mapping at fault address
/home/ryan/src/mlpack/src/mlpack/tests/serialization.hpp(235): last checkpoint

*** 1 failure is detected in the test module "mlpackTest"
```

zsogitbe:
Ryan, I have invested some more time in this issue this Sunday morning and here are my findings about this problem and connected issues:

1. The same problem also exists in LayerNorm.

2. I have tried to serialize 'size' and 'eps' and use weights.set_size when loading the archive, I have also tried to set deterministric = false and reinitialize gamma and beta, but nothing helps. The code still crashes. You will need to dig deeper into this problem. The 'deterministic' parameter (and the stats) may be one of the reasons but I am not completely sure about it because I am still discovering your code.

3. General remark: The compilation of the test cases takes too much time. I think that it would be much better to distribute the tests into separate projects.

4. Your above test, when added to serialization_test.cpp, refuses to compile. It gives an ambitious call to function size() in elu.hpp (line 197 with me), but I guess that you will also see this if you try to compile it.

I hope this helps. Br, Z.

zsogitbe:
Maybe take also a look at running_stat_vec_aux::update_stats(). Would it be possible to serialize a small sample and then construct a running_stat_vec and use the deserialized sample to update the stats?

rcurtin:
We should consider also submitting a patch to Conrad upstream that has a new constructor for `running_stats` that can take in the necessary info (mean, variance, number of samples).  But in the mean time, yeah, replacing `running_stats` with something that's basically identical could work...

I put the test I wrote into `ann_layer_test.cpp` and there were no compilation issues there, though I did also need to `#include "serialization.hpp"` to make the correct symbols available.

zoq:
> 1. The same problem also exists in LayerNorm.

https://github.com/mlpack/mlpack/pull/1508 should fix the the layer norm and batch norm issue; it does pass on the test Ryan suggested above, will test some other cases as well.

> 3. General remark: The compilation of the test cases takes too much time. I think that it would be much better to distribute the tests into separate projects.

You are absolutely right, most of the time I build with ``make -j4`` to use multiple cores for the build. Also, there is an option to discard the tests from the build ``cmake -DBUILD_TESTS=OFF``. Personally, I don't see a benefit of distributing the tests into a separate project, at least not at this point, but if anybody likes the idea and is willing to give it a shot, I'm open for a discussion.

zsogitbe:
Great work Marcus! The generalization of the test is also clever well done!
I confirm that this version passes my own test.
Br, -Z

ShikharJ:
I'll close this pull request as it has been taken up in #1508.

PR #1507, opened 2018-09-14
Correct and add examples on how to use the policies in CF

Wenhao-H:
This PR corrects and adds examples in the code comments on how to use the policies in CF.


zoq:
Ok, with two approvals I will go ahead and merge the code now; Thanks again. 👍

PR #1508, opened 2018-09-16
Layer serialization

zoq:
- batch normalization layer serialization test
- layer normalization layer serialization test

### DIFF
--- a/src/mlpack/methods/ann/layer/batch_norm_impl.hpp
+++ b/src/mlpack/methods/ann/layer/batch_norm_impl.hpp
@@ -20,23 +20,28 @@
 namespace mlpack {
 namespace ann { /** Artificial Neural Network. */
 
-template<typename InputDataType, typename OutputDataType>
+template <typename InputDataType, typename OutputDataType>
 BatchNorm<InputDataType, OutputDataType>::BatchNorm() :
     size(10),
     eps(1e-8),
-    deterministic(false)
+    loading(false),
+    deterministic(false),
+    count(0)
 {
   // Nothing to do here.
 }
-
 template <typename InputDataType, typename OutputDataType>
 BatchNorm<InputDataType, OutputDataType>::BatchNorm(
     const size_t size, const double eps) :
     size(size),
     eps(eps),
-    deterministic(false)
+    loading(false),
+    deterministic(false),
+    count(0)
 {
   weights.set_size(size + size, 1);
+  runningMean.zeros(size, 1);
+  runningVariance.zeros(size, 1);
 }
 
 template<typename InputDataType, typename OutputDataType>
@@ -44,10 +49,15 @@ void BatchNorm<InputDataType, OutputDataType>::Reset()
 {
   gamma = arma::mat(weights.memptr(), size, 1, false, false);
   beta = arma::mat(weights.memptr() + gamma.n_elem, size, 1, false, false);
+
+  if (!loading)
+  {
+    gamma.fill(1.0);
+    beta.fill(0.0);
+  }
+
   deterministic = false;
-  gamma.fill(1.0);
-  beta.fill(0.0);
-  stats.reset();
+  loading = false;
 }
 
 template<typename InputDataType, typename OutputDataType>
@@ -59,15 +69,9 @@ void BatchNorm<InputDataType, OutputDataType>::Forward(
   // the forward pass when deterministic is set to true.
   if (deterministic)
   {
-    // Mini--batch mean using the stats object.
-    mean = stats.mean();
-
-    // Mini--batch variance using the stats object.
-    variance = stats.var(1);
-
     // Normalize the input and scale and shift the output.
-    output = input.each_col() - mean;
-    output.each_col() %= gamma / arma::sqrt(variance + eps);
+    output = input.each_col() - runningMean;
+    output.each_col() %= gamma / arma::sqrt(runningVariance / count + eps);
     output.each_col() += beta;
   }
   else
@@ -75,12 +79,19 @@ void BatchNorm<InputDataType, OutputDataType>::Forward(
     mean = arma::mean(input, 1);
     variance = arma::var(input, 1, 1);
 
-    for (size_t i = 0; i < input.n_cols; i++)
-      stats(input.col(i));
-
     // Normalize the input.
     output = input.each_col() - mean;
     output.each_col() /= arma::sqrt(variance + eps);
+
+    // Use Welford method to compute the sample variance and mean.
+    for (size_t i = 0; i < input.n_cols; i++)
+    {
+      count += 1;
+
+      OutputDataType delta = input.col(i) - runningMean;
+      runningMean = runningMean + delta / count;
+      runningVariance += delta % (input.col(i) - runningMean);
+    }
 
     // Reused in the backward and gradient step.
     normalized = output;
@@ -139,8 +150,20 @@ template<typename Archive>
 void BatchNorm<InputDataType, OutputDataType>::serialize(
     Archive& ar, const unsigned int /* version */)
 {
+  ar & BOOST_SERIALIZATION_NVP(size);
+
+  if (Archive::is_loading::value)
+  {
+    weights.set_size(size + size, 1);
+    loading = false;
+  }
+
+  ar & BOOST_SERIALIZATION_NVP(eps);
   ar & BOOST_SERIALIZATION_NVP(gamma);
   ar & BOOST_SERIALIZATION_NVP(beta);
+  ar & BOOST_SERIALIZATION_NVP(count);
+  ar & BOOST_SERIALIZATION_NVP(runningMean);
+  ar & BOOST_SERIALIZATION_NVP(runningVariance);
 }
 
 } // namespace ann
--- a/src/mlpack/methods/ann/layer/batch_norm_impl.hpp
+++ b/src/mlpack/methods/ann/layer/batch_norm_impl.hpp
@@ -22,7 +22,6 @@ namespace ann { /** Artificial Neural Network. */
 
 template<typename InputDataType, typename OutputDataType>
 BatchNorm<InputDataType, OutputDataType>::BatchNorm() :
-    size(10),
     eps(1e-8),
     loading(false),
     deterministic(false),
--- a/src/mlpack/methods/ann/layer/batch_norm_impl.hpp
+++ b/src/mlpack/methods/ann/layer/batch_norm_impl.hpp
@@ -20,7 +20,7 @@
 namespace mlpack {
 namespace ann { /** Artificial Neural Network. */
 
-template <typename InputDataType, typename OutputDataType>
+template<typename InputDataType, typename OutputDataType>
 BatchNorm<InputDataType, OutputDataType>::BatchNorm() :
     size(10),
     eps(1e-8),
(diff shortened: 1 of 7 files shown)


rcurtin:

rcurtin:
The doxygen checks failed because one of the systems didn't have texlive installed.  I fixed that, hopefully it will work now.

zsogitbe:
One more very small typo in the LayerNormSerializationTest. The size should be 10.

src/mlpack/tests/serialization_test.cpp
```diff
@@ -1784,4 +1789,60 @@ BOOST_AUTO_TEST_CASE(ssRBMTest)
...
+  BatchNorm<> layer(10);
+  ANNLayerSerializationTest(layer);
+}
+
+/**
+ * Simple serialization test for layer normalization layer.
+ */
+BOOST_AUTO_TEST_CASE(LayerNormSerializationTest)
+{
+  LayerNorm<> layer(1);
```
ShikharJ: The size here can be set to 10 as suggested.

zoq: If I remember right, LN does work over samples so we compute the mean/variance for every single sample, and BN works over all samples/features; correct me if I'm wrong; if we change the value we would have to adjust the layer to use ``each_col``instead of ``each_row``.

zsogitbe: I did not look at your implementation but I expected to input the number of input nodes (10) (this is what I expect logically). In case the value is always 1 then it does not have any sense to have the size parameter and also does not have any sense to serialize it. I am still discovering your code, therefore it is your judgement what is important here but in any case please make sure that the code can not be misinterpreted (confusion).

zsogitbe: I was just looking at the LayerNormTest in ann_layer_test.cpp and there you have:

LayerNorm<> model(input.n_cols);

So be careful with this size parameter. I think you better remove it.

zoq: The size still depends on the input dimension, it's just 1 if the input is a vector; so I'm not sure we can remove the term right away, however, we should make clear that there is a difference between BN and LN, and that they are not interchangeable. @ShikharJ any thoughts on this?

zoq: @zsogitbe thanks for the input

zsogitbe: Marcus, I am sorry but I do not understand you. The input in the LayerNormSerializationTest is not a vector but it is a 5x100 matrix:

arma::mat input(5, 100, arma::fill::randu);

One more question. While further discovering your work I was looking at the BatchNorm layer implementation and I do not really see anywhere a batch size or batch calculation. Am I missing here something?
Thank you, -Z

rcurtin: Hey Zoltan, each point as input is one column in that matrix, and the batch size is specified by the optimizer---so from the perspective of each layer, the matrix it receives has `n_cols` points and each point has `n_rows` dimensions.  There is some documentation here on the structure of the ANN code: http://mlpack.org/docs/mlpack-3.0.3/doxygen/anntutorial.html.  Maybe that could help out?

zsogitbe: Yes, that is clear now with this explanation. Thank you Ryan!

The only thing which remains is the size for the LayerNorm. According to Marcus: "The size still depends on the input dimension, it's just 1 if the input is a vector". This is not clear from my point of view and I am not sure about this size=1. I could understand that it is always 1 but if it is not always 1 then it is always 'input.n_cols' which would mean 100 in case of the new ANNLayerSerializationTest. Could you please look at this and find some kind of clear explanation?
Thank you, -Z

ShikharJ: @zoq You are correct about the size being set above, kindly ignore my first comment, the input size should be `input.n_cols` in case of `LayerNorm` and `input.n_rows` in case of `BatchNorm`.

zsogitbe: Shikhar, in that case should the size not be 10 in the new example instead of 1? That is the n_cols in the former layer or am I missing here something?

ShikharJ: @zsogitbe This might be helpful (http://mlexplained.com/2018/01/13/weight-normalization-and-layer-normalization-explained-normalization-in-deep-learning-part-2/).

zsogitbe: Thank you Shikhar! I will look at it. I think that you should make one or two examples where you show how to set the size parameter in different situations (in a network) or remove completely this 'size' and compute it automatically (the best way) as you have suggested above.
For me it is not clear anymore how to set the size in both BatchNorm and LayerNorm in a real neural network (several layers). I will let you think about this and I will not interfere further. Thank you! -Z

ShikharJ: @zsogitbe I agree for bigger networks, you would have to compute the shape of the output matrix before applying these layers to that output, which would be frustrating.

zoq: As you already pointed out we would have to call ``Forward`` first, but at this point we already initialized the whole network parameters/weights, so ``gamma = arma::mat(weights.memptr(), 1, size, false, false);`` the ``weights`` would be empty. There is an open issue https://github.com/mlpack/mlpack/issues/1267 to deduce the layer input/output size from the previous layer, which would also simplify the BN/LN layer. So, if you agree we improve the current layer documentation (e.g. add some examples) and work on #1267 aferwards, to simplify the usage. Let me know what you think.

zoq: #1267 is definitely ambitious if we like to cover all settings, input/output size should be straightforward though. Also, #1267 is on top of the current codebase, so there is no need to touch any layer or layer method Forward, Backward, Gradient, Reset included.

Adding on this point, a simple Excel sheet is a nice idea, but that would also mean that you have to reimplement at least the input/output size logic for each layer; which I think is unnecessary if you could just call Forward for each layer to get the necessary information for free.

rcurtin: I always appreciate input but there are a couple of assumptions here that I think are worth clarifying.

> February (a long time ago = it will probably never happen)

For us this is not true; even mlpack bugs that are years old often do eventually get resolved.  We just recently got a fix in for a Windows installer, and that issue had been open since 2011.  That's only one of many examples.

> the column-row switch because of armadillo is twisting things already quite a bit

Note that a column-major representation of data is quite standard in C++, since that is also what is done in FORTRAN by LAPACK and BLAS, the tools that Armadillo is built on top of.  Eigen does the same thing.  So it only makes sense to have observations as columns in this case.  You're right that this is not like Python, which is now the standard for data science.

> Even you make mistakes with the size

I'm not convinced there is any problem with this code; see Marcus' comments.

There is a difference between "complexity" meaning that something is actually complex and "complexity" meaning that something is simply different than what you're used to.  mlpack is written for speed and this induces some actual complexity, which is why #1267 isn't completely trivial to solve (input/output size detection for instance would have to be done with the visitor paradigm at the start of training due to how the code is structured).  Certainly it's of importance to resolve the issue of automatic layer size detection when we can, but I am pretty sure we are all on the same page about that.

zsogitbe: Yes, sorry for the February remark. I meant it well. I know that you are all doing a great job on mlpack and it is great that it is available to us! I am impressed with the very clever architecture of it! If this would not be the case then I would not be here and I would not give so many remarks! With the complexity I meant that it is easy to make a mistake (like I mentioned in the test case where the size should be 100 instead of 1). I find it very important to work on simplification as much as you work on adding complexity, and this is more important today than ever before because complexity in machine learning is exponentially growing!
In a neural network architecture there can be many parameters which you have to set yourself and where you can make a mistake very easily (convolution, pooling, batchnorm, layernorm, etc.). For me having an automatic parameter selection is of huge importance but I understand that you are doing your best. In the mean time it would be a good idea to have an explanation and examples on the calculation of the size parameter in different situations (different combination of layers) or an Excel sheet. The variable name and explanation in the code are also important in order to reduce confusion because 'size' in one type of layer is not the same 'size' than in an other type of layer. 
I hope this helps. 
Br, -Z

ShikharJ:
Let's wait till #1512 gets merged.

src/mlpack/methods/ann/layer/batch_norm_impl.hpp
```diff
@@ -20,34 +20,44 @@
 namespace mlpack {
 namespace ann { /** Artificial Neural Network. */
 
-template<typename InputDataType, typename OutputDataType>
+template <typename InputDataType, typename OutputDataType>
```
ShikharJ: I guess we can remove the extra space here.

src/mlpack/methods/ann/layer/batch_norm_impl.hpp
```diff
@@ -20,34 +20,44 @@
...
     eps(1e-8),
-    deterministic(false)
+    loading(false),
+    deterministic(false),
+    count(0)
 {
   // Nothing to do here.
 }
-
 template <typename InputDataType, typename OutputDataType>
```
ShikharJ: Same here.

src/mlpack/methods/ann/layer/layer_norm_impl.hpp
```diff
@@ -19,10 +19,12 @@
 namespace mlpack {
 namespace ann { /** Artificial Neural Network. */
 
-template<typename InputDataType, typename OutputDataType>
+
+template <typename InputDataType, typename OutputDataType>
```
ShikharJ: Here as well.

src/mlpack/methods/ann/layer/layer_norm_impl.hpp
```diff
@@ -19,10 +19,12 @@
...
 LayerNorm<InputDataType, OutputDataType>::LayerNorm() :
-    size(10),
-    eps(1e-8)
+    size(1),
```
ShikharJ: I think we can revert back to the default size of 10 here now.

src/mlpack/tests/serialization_test.cpp
ShikharJ: This can be set to 10 now, since we've made the changes.

src/mlpack/methods/ann/layer/layer_norm_impl.hpp
zoq: Don't think that's necessary, the value has to be tailored to the task anyway, and I guess using ``LayerNorm<> layer();`` instead of ``LayerNorm<> layer(10);`` might create the wrong impression.

zsogitbe: Yes I agree with Marcus but I think that Shikhar meant to add the 10 which is also the default value. And here is another point where you could do a better job. Never design an example or test case where the default values are present because that may cause confusion (for the users) and mistakes (for you). I have seen default values in many of your examples. You especially like the values 10 and 1 for some reason :) lucky numbers for one of you?

ShikharJ: @zoq Please feel free to remove the default values for `BatchNorm` and `LayerNorm` parameter.

ShikharJ:
Thanks for the great contribution.

Issue #1446, opened 2018-06-27
LMNN: optimization of kNN computation by bound relaxation


rcurtin:
Here's another nice way that I think we can have a lot of speedup with the kNN search (not the tree building).

The kNN search stores bounds in the query tree during search.  The algorithm itself is quite complex, but in essence, when we start the search we have a bound set at DBL_MAX that says, "the k-nearest-neighbor of any descendant of this query node is no further than DBL_MAX away".  But we have a better bound on the k-nearest-neighbor---we know that the k-nearest-neighbor at iteration (t + 1) is going to be bounded by `d_{L_t}(x_i, x_j) + f(|| L_t - L_{t + 1} ||)` where `f(.)` is some quantity I still have to derive.

If we start the kNN search with a much closer bound than `DBL_MAX`, search can be *significantly* faster (think order of magnitude).  So to do this, we will have to:

 1. Derive the correct bound to use.
 2. Use cached query trees (#1445), or single-tree mode for small-batch computation.
 3. Set the bound to the bounded kNN distance before calling `knn.Search()`.
 4. If we are not caching query trees, store each query point's k'th NN distance for use next iteration.  (I think the code already does this.)

manish7294:
This sounds like another pretty good idea. I believe the bounds here will be somewhat similar to the ones we will be using in #1449.

manish7294:
I think this may be correct:
` f(|| L_t - L_{t + 1} ||)  =   || L_{t + 1} - L_t ||_F^2 *  max( ||x_a||^2 +  ||x_i||^2)`

x_a : a point among kth nearest neighbors of all data points

rcurtin:
Right, I think this is close to correct.  Here's a derivation, supposing x_i is the query point and x_a is the k-nearest neighbor (although the notation doesn't actually matter):

```
d(L_{t + 1} x_i, L_{t + 1} x_a) <= d(L_t x_i, L_t x_a) + d(L_t x_i, L_{t + 1} x_i) + d(L_t x_a, L_{t + 1} x_a)
 <= d(L_t x_i, L_t x_a) + || L_t - L_{t + 1} ||_F^2 ( || x_i ||^2 + || x_a ||^2).
```

Maybe I overlooked something?  I'm not sure where the `max()` came from in your solution, and it seems like maybe it is missing a second argument?

manish7294:
Ah! That was just a intutional expression I had written earlier :)

manish7294:
[...]

PR #1509, opened 2018-09-18
LMNN optimization: use bounds from previous iterations in tree-based kNN search

rcurtin:
This PR is to solve #1445 and #1446.  It implements the following:

 * Reuse trees from previous calls to `Impostors()`.
 * Use bounded values of impostors from previous calls to set bounds for this iteration's call to `Impostors()`.
 * Prune points based on impostor bounds from #1448 so they are not recomputed with the tree-based search.
 * Allow rebuilding of trees when the transformation is different enough from the one when the trees were built.

It took me a long time of playing with this to get it to work reasonably.  I'll add benchmarking times in the next little while.  I don't yet know if rebuilding the trees is helpful, but I think I have seen reasonable speedup (think 20-30% or more) from the tree optimizations overall.


rcurtin:
Note that this is dependent on #1494 being merged, since it won't work with the old style of SGD impostor calculation which only computed impostors for small batches at a time.  When that's merged I'll adapt the code here accordingly.

rcurtin:
@manish7294: sorry, I forgot to tag you in this!  Take a look and let me know what you think when you have a chance.

manish7294:
Hi Ryan, Thanks for working over it, It's good to see all this work. I am extremely sorry for delay from my side, just few more days and my mid-semester exams will be over.

manish7294:
Great! This is a lot of intelligent work, It good to hear that you got such significant speedups. Sorry for asking this, but do you have any numbers on accuracy and timing? Please don't bother to run simulations in case you don't have them saved somewhere. Now, code looks much cleaner, specially lmnn_function_impl.hpp & constraints_impl.hpp :) 
I guess I should hurry up with #1494 as well.

PR #1512, opened 2018-09-20
Correct Scaling Parameters in LayerNorm

ShikharJ:


zoq:
Ok, with two approvals I will go ahead and merge the code now.

zsogitbe:
Congratulations! It looks more logical now (in line with the other modules) and I suspect that it is faster also because of the removal of several reshape operations! Hopefully you had some time to test it with real data.

PR #1519, opened 2018-10-09
Replace copy and move overloads with pass-by-value for RASearch class

chigur:
For RASearch constructors and the class method Train, replace copy and move overloads with single pass-by-value methods. The user can use `std::move(referenceSet)` to prevent copying of the dataset. Implements the guidelines set in #1021 .


chigur:
Thank you so much Ryan. I'll add my name to the contributors list in some other other, more difficult, pull request. It just doesn't feel earned right now :)

rcurtin:
Thanks again! :+1:

Issue #1522, opened 2018-10-09
packaging of python extensions fail


marcespie:
The build system mistakenly assumes you can install stuff directly. This breaks havoc with any modern packaging system.

At the very least, support setting DESTDIR during configuration. It would be better to do that during the install stage, but it allows things to proceed correctly.
[patch.txt](https://github.com/mlpack/mlpack/files/2460366/patch.txt)

rcurtin:
You can get the same effect by just configuring CMake with `-DCMAKE_INSTALL_PREFIX=/your/preferred/install/root/`. :)

rcurtin:
(I already responded to this via a personal email, but probably better to respond here publicly just so all the discussion is in one place.)

Thanks for the clarification---I understand the difference now.  Are you sure that DESTDIR doesn't already work?  i.e.

`make DESTDIR=/path/to_staging_area/ install`?

The CMake documentation suggests this already should work on UNIX-like CMake systems:

https://cmake.org/cmake/help/latest/envvar/DESTDIR.html

rcurtin:
Fixed in #1523.

Issue #1521, opened 2018-10-09
Python bindings setup assumes too many things


marcespie:
Around line 40 of setup.py.in:

                  # CMAKE_CXX_FLAGS seems to have an extra space.
                  extra_compile_args=('-DBINDING_TYPE=BINDING_TYPE_PYX ' + \
                      '-std=c++11${CMAKE_CXX_FLAGS}').split(' '),

nope, it doesn't. Not always.  Funny seeing my compiler trying to build with
-std=c++11-O2

and right after that:
                 extra_link_args=['${OpenMP_CXX_FLAGS}'],

turns into a magnificent NOT FOUND while building extensions. I have no idea (yet) whether OpenMP is required for python bindings, or whether a specific build configuration accidentally has it.

If it's required, the main CMakeLists.txt logic should then proceed to disable python extensions...

marcespie:
I can confirm that the python bindings build just fine without ${OpenMP_CXX_FLAGS} on a !OpenMP installation...

rcurtin:
Ah, thanks.  I think that the two issues you've encountered here are related to #1478 and #1425, respectively.  In both of those cases, it was problematic to come up with fixes because we couldn't directly reproduce them.

Can you try the patch in #1478 and see if that works for you?  If so, then I'd be happy to commit that.

And for the OpenMP CXXFLAGS bit, I bet this is what was going wrong in #1425, so thanks for pointing it out.  We can probably do the following... in `CMakeLists.txt`, change the block at line 360 from:

```
if (OPENMP_FOUND)
  add_definitions(-DHAS_OPENMP)
  set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} ${OpenMP_C_FLAGS}")
  set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} ${OpenMP_CXX_FLAGS}")
else ()
  # Disable warnings for all the unknown OpenMP pragmas.
  set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wno-unknown-pragmas")
endif ()
```

to

```
if (OPENMP_FOUND)
  add_definitions(-DHAS_OPENMP)
  set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} ${OpenMP_C_FLAGS}")
  set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} ${OpenMP_CXX_FLAGS}")
else ()
[...]

Issue #1520, opened 2018-10-09
missing manpage dependency


marcespie:
Depending on the build order, manpage will be generated... or not.
[patch1.txt](https://github.com/mlpack/mlpack/files/2459454/patch1.txt)

rcurtin:
Nice catch, thank you!

rcurtin:
Fixed in #1523.

Issue #1518, opened 2018-10-08
execman linux'isms


marcespie:
- those scripts are not bash. They're perfectly portable bourne shell
- find -executable is a GNUism, the same thing is easy to do with POSIX find.
- compressing manpages  shouldn't be hard-coded, as that's a system preference...

[patch2.txt](https://github.com/mlpack/mlpack/files/2456615/patch2.txt)
[patch1.txt](https://github.com/mlpack/mlpack/files/2456616/patch1.txt)

rcurtin:
Agreed on all three counts, thanks.  Do you want to open a PR?  I'd be happy to commit these with your name on the patches and add your name to the list of contributors if you like.  (And, we also mail stickers you could put on your laptop or wherever to people who have contributed!  So if you want I can get some in the mail too.)

marcespie:
On Mon, Oct 08, 2018 at 07:31:12PM -0700, Ryan Curtin wrote:
> 
>    Agreed on all three counts, thanks. Do you want to open a PR? I'd be
>    happy to commit these with your name on the patches and add your name
>    to the list of contributors if you like. (And, we also mail stickers
>    you could put on your laptop or wherever to people who have
>    contributed! So if you want I can get some in the mail too.)
> 
>    --
>    You are receiving this because you authored the thread.
>    Reply to this email directly, [1]view it on GitHub, or [2]mute the
>    thread.

Please go ahead and commit directly.

It's small enough that I don't think it needs any attribution yet.

rcurtin:
Fixed in #1523.

PR #1523, opened 2018-10-09
Fixes from marcespie

rcurtin:
This is a collection of fixes for issues @marcespie has opened recently.  It includes the following changes:

 * (#1516) stop using -lrt since it isn't needed
 * add linking against -pthread for std::thread usage when using g++
 * (#1518) use /bin/sh not /bin/bash
 * (#1518) don't gzip man pages
 * (#1520) refactor dependencies on `man` target so they're automatically added when `add_cli_executable()` is called
 * (#1521, #1478) fix bad handling of spaces in CXXFLAGS in Python `setup.py`
 * (#1521, #1425?) fix bad usage of `OpenMP_CXX_FLAGS` in Python `setup.py`
 * (#1522) handle DESTDIR correctly for Python bindings (the CMake glue previously didn't)

@marcelotrevisani I'm not sure why you closed #1478, but I've pulled that branch in here and we can merge it as part of this one.  In my basic testing it seems to work.

@marcespie---if you'd like to test and see if this fixes the problems that you had, it would be great.  If it works, we can get it merged. :)

### DIFF
--- a/CMake/allexec2man.sh
+++ b/CMake/allexec2man.sh
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/bin/sh
 
 set -e
 
@@ -23,5 +23,4 @@ for program in $(find . -type f -executable -iname 'mlpack_*' | \
                 grep -v '_test$'); do
   echo "Generating man page for $program...";
   "$exec2man" "$program" "$outdir/$program.1"
-  gzip -f "$outdir/$program.1"
 done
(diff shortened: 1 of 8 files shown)


marcespie:
I don't see how the travis timeout could be related ?  Especially since it only occurs in ctest.

rcurtin:
Right, this seems like a random failure to me (sometimes this happens).  I think we can merge despite it.

rcurtin:
I'll go ahead and merge this in 3 days to leave time for any comments if anyone has any.

PR #1529, opened 2018-10-13
neural network regression example

zoq:
neural network regression contributed by @zsogitbe, I think before we can merge this in we should:

- integrate this into the existing ci, to make sure this doesn't go out of date
- write a simple guide for the example that provides some additional information about the method, usage, etc.
- reference this in existing tutorials


zsogitbe:
Thank you for posting it Marcus!

1. I have integrated a simple guide into the source code and can be extracted with Doxypress/Doxygen. It gives a nice web page with a description. If you want more then I invite someone to contribute this part. I think that the best would be if it is integrated into the source code that it can be extracted automatically. Feel free to add further author names to the example if the guide is expanded.. 

2. As it is mentioned in the example (TODO) the new FeatureNormalization class should go into 'mlpack/core/data/feature_normalization.hpp'. But the example will then only work if feature_normalization.hpp exist (+ #include <mlpack/core/data/feature_normalization.hpp> in the code) so the two should be synchronized.

doc/examples/nn_regression/nn_regression.cpp
```diff
@@ -0,0 +1,409 @@
...
+						// skip single-valued attribute
+						if ((*itemMin)(i) != (*itemMax)(i)) {
+							if (val == (*itemMin)(i))
+								val = scale_lower;
+							else if (val == (*itemMax)(i))
+								val = scale_upper;
+							else
+								val = scale_lower + Dx * (val - (*itemMin)(i)) / ((*itemMax)(i) - (*itemMin)(i));
+						}
+						datapoint(i) = val;
```
rcurtin: What about an implementation like this?

```
datapoint = arma::clamp((datapoint - itemMin) ./ ranges, 0.0, 1.0) * Dx + scale_lower;
```

Basically this vectorized expression requires us to have a precomputed vector `ranges = itemMax - itemMin`.  This should be faster to run too, since Armadillo should be able to make use of SIMD instructions and loop unrolling where possible.

doc/examples/nn_regression/nn_regression.cpp
```diff
@@ -0,0 +1,409 @@
...
+				// Use the itemMin, itemMax vectors if provided (e.g. for subsequent normalization with the same range)
+				// or make them if not provided
+				if (itemMin->size() < 1 || itemMax->size() < 1) {
+					*itemMin = arma::vec(data.n_rows, arma::fill::zeros);
+					*itemMax = arma::vec(data.n_rows, arma::fill::zeros);
+					for (int i = 0; i < data.n_rows; i++)
+					{
+						(*itemMin)(i) = arma::min(data.row(i));
+						(*itemMax)(i) = arma::max(data.row(i));
+					}
```
rcurtin: Because Armadillo is column-major, I think it will be much faster to do this in a vectorized operation also (which will iterate over columns not rows):

```
itemMin = arma::min(data, 1);
itemMax = arma::max(data, 1);
```

doc/examples/nn_regression/nn_regression.cpp
zsogitbe: Ryan, I am not yet very much used to armadillo, would you mind to change the code as you think that it is faster when you have some spare time (and please test it also, my code is tested and it works well)? Of course add your name as co-author.
My only remark is that do not use the clamp function because that would cause an error in **subsequent** normalization (the values may be outside the 0-1 limit.)!

rcurtin: Sure, I can go over and rewrite the Armadillo expressions.  I see what you mean about the clamp; I actually misread this as `if (val <= (*itemMin)(i))`, but I see the code is different now.  (Actually I think in this case it would be equivalently fast or faster than the current implementation to simply remove the `if`/`else if`/`else` and just use the body of the `else`.  But in nay case a one-liner with Armadillo should be nicer for maintenance so we can go with that.)

zsogitbe: OK, just be careful with the division by 0 (I do not see yet how you can prevent that with your armadillo one liner)! I am definitely interested in learning that from you :) . I thought that the if/else statement is faster than evaluating the expression (it may occur a lot that the min or max values are present).

zsogitbe: Yes, you are right Ryan. I have already corrected the code on my computer a few days ago. That part looks like now this:

```
if(!is_close_to_zero((*itemMax)(i) - (*itemMin)(i)))
{
	if (val == (*itemMin)(i))
		val = scale_lower;
	else if (val == (*itemMax)(i))
		val = scale_upper;
	else
		val = scale_lower + Dx * (val - (*itemMin)(i)) / ((*itemMax)(i) - (*itemMin)(i));
}
else {
	//in case all data points are the same then set all to 1 
	val = 1.0;
}
datapoint(i) = val;
```

zoq:
Let's keep this open.

kartikdutt18:
Hey @zoq, Do you think this should be shifted to the new examples repo?

kartikdutt18:
Hey @birm, I think we can close this now.

birm:
Good point; resolved via https://github.com/mlpack/examples/pull/73

kartikdutt18:
Awesome!
Thanks a lot.

PR #1531, opened 2018-10-16
Replace copy and move overloads with pass-by-value for NeighborSearch

chigur:
Replace copy and move overloads of the method Train in class NeighborSearch with pass-by-value. This pull request implements the guidelines suggested in #1021.

### DIFF
--- a/src/mlpack/methods/neighbor_search/neighbor_search_impl.hpp
+++ b/src/mlpack/methods/neighbor_search/neighbor_search_impl.hpp
@@ -331,46 +331,7 @@ template<typename SortPolicy,
          template<typename> class DualTreeTraversalType,
          template<typename> class SingleTreeTraversalType>
 void NeighborSearch<SortPolicy, MetricType, MatType, TreeType,
-DualTreeTraversalType, SingleTreeTraversalType>::Train(
-    const MatType& referenceSet)
-{
-  // Clean up the old tree, if we built one.
-  if (treeOwner && referenceTree)
-  {
-    oldFromNewReferences.clear();
-    delete referenceTree;
-  }
-
-  // Delete the old reference set, if we owned it.
-  if (setOwner && this->referenceSet)
-    delete this->referenceSet;
-
-  // We may need to rebuild the tree.
-  if (searchMode != NAIVE_MODE)
-  {
-    referenceTree = BuildTree<Tree>(referenceSet, oldFromNewReferences);
-    treeOwner = true;
-    this->referenceSet = &referenceTree->Dataset();
-  }
-  else
-  {
-    treeOwner = false;
-    this->referenceSet = &referenceSet;
-  }
-
-  setOwner = false; // We don't own the set in either case.
-}
-
-template<typename SortPolicy,
-         typename MetricType,
-         typename MatType,
-         template<typename TreeMetricType,
-                  typename TreeStatType,
-                  typename TreeMatType> class TreeType,
-         template<typename> class DualTreeTraversalType,
-         template<typename> class SingleTreeTraversalType>
-void NeighborSearch<SortPolicy, MetricType, MatType, TreeType,
-DualTreeTraversalType, SingleTreeTraversalType>::Train(MatType&& referenceSetIn)
+DualTreeTraversalType, SingleTreeTraversalType>::Train(MatType referenceSetIn)
 {
   // Clean up the old tree, if we built one.
   if (treeOwner && referenceTree)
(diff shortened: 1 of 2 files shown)


src/mlpack/methods/neighbor_search/neighbor_search_impl.hpp
```diff
@@ -331,46 +331,7 @@ template<typename SortPolicy,
...
-  if (searchMode != NAIVE_MODE)
-  {
-    referenceTree = BuildTree<Tree>(referenceSet, oldFromNewReferences);
-    treeOwner = true;
-    this->referenceSet = &referenceTree->Dataset();
-  }
-  else
-  {
-    treeOwner = false;
-    this->referenceSet = &referenceSet;
```
rcurtin: There is a slight behavior change here.  In the old code, you could pass a const reference and the matrix would not be copied and `setOwner` would be false in naive mode.  But now that will be copied, and it's not clear to me how that copy could be avoided in any case.  I don't think we need to change anything in this PR but it is something to think about.

rcurtin:
Still one travis failure but it seems random.  I'll go ahead and merge this in 3 days.  Thanks! :+1:

src/mlpack/methods/neighbor_search/neighbor_search_impl.hpp
chigur: You're right. On thinking more about this, isn't `RASearch` class similar to this one from the perspective of tree and dataset ownership management? If yes, then should we make the API of `NeighborSearch` class' `Train` method match that of `RASearch`? They're different right now and which is probably why `setOwner` is not needed in `RASearch`.

Also, should we do something about the erratically failing build? A quick google search suggests trying out [multiple ](https://github.com/scitran/core/pull/783/files#diff-3254677a7917c6c01f55212f86c57fbfR35) keyservers, using [timeouts](https://github.com/TechEmpower/FrameworkBenchmarks/pull/1033/files#diff-f79779b600de9619f774461033807b97R310), or using the `travis_retry` [option](https://docs.travis-ci.com/user/common-build-problems/#travis_retry).

rcurtin: I think you mean `RangeSearch` not `RASearch`?  If you want to make things match I'm just fine with that.  I think in `RangeSearch` the observation is that `setOwner == naive` always, so we only need to hold `naive` and the `setOwner` member can be removed.  If that's now true for `NeighborSearch` feel free to make the change.

Mostly I think nobody has had the time to debug the failing Travis build.  A long time ago we spent a while trying to upgrade the distro to xenial but we had all kinds of problems with that.  If you want to try making a quick configuration change, we can merge it in and see if it helps. :)

chigur: Yes, I meant `RangeSearch`. First, after merging this pull request, `setOwner == naive` will hold across all three classes, `NeighborSearch`, `RASearch`, and `RangeSearch`. Thus, we can remove it from all of them. Second, is it better to pass `referenceTree` as pass-by-value (`NeighborSearch::Train` in current pull request) or as a pointer (`RangeSearch::Train`)?

Okay. I'll try something.

rcurtin: Hmm, I think for passing pre-trained trees, we should actually do `NeighborSearch::Train(TreeType tree)` which allows a user to do `std::move()` to avoid a copy.  So we should mark the `RangeSearch::Train(TreeType*)` overload as deprecated (with `mlpack_deprecated`); we can't remove it until mlpack 4.0 because it would break reverse compatibility.

If a user wants to avoid the cost of copying trees I guess they can do, e.g.,

```
TreeType tree(data); // manually built tree
KNN knn;
knn.Train(std::move(tree)); // now 'tree' is empty but no copy
knn.Search(...);

// now we want the tree back...
tree = std::move(knn.ReferenceTree());
```

chigur: The new `std::move` semantics never cease to surprise me! Okay, so I'll mark the pointer methods with `mlpack_deprecated` and since I don't want to take up any more of your time, two final questions:
1. With the above changes to `NeighborClass`, only in naive mode will the class be the owner of the data. In any other mode the class is going to own the tree which in turn will own the dataset, right?
2. If above is true, then we don't even `treeOwner`. So, should I get rid of it in the current pull request itself?

rcurtin: 1. Yep, exactly.
2. Yes, if we change `Train()` to take just a `TreeType` instead of a pointer, then I guess we always own the tree if it exists.  So that is another nice simplification! :)

Don't worry about asking questions and taking my time---that is what I'm here for. :)  Looking forward to seeing the resulting changes!  It'll be a nice improvement. :+1:

chigur: Great! Thanks.

src/mlpack/methods/neighbor_search/neighbor_search_impl.hpp
```diff
@@ -274,16 +267,20 @@ NeighborSearch<SortPolicy,
...
+  if (referenceTree)
+  {
     delete referenceTree;
-  if (setOwner && referenceSet)
+    referenceTree = NULL;
+  }
+  else
+  {
     delete referenceSet;
+    referenceSet = NULL;
```
rcurtin: Technically setting the pointers to NULL isn't necessary anywhere in the code since either the object is being destructed or the pointers will then be set to something else, but it's not necessarily a problem to leave these lines in.

chigur: You're right. I had thought the same initially, but then `oldFronNewReferences` wasn't getting cleared correctly in the `Train` method when switching from non-naive to naive mode. Specifically, the `MoveTrainTest` test was failing. Maybe, I can get rid of setting pointers to NULL and still have everything work correctly. I was thinking that it might be a good practice to set deleted pointers to NULL to avoid double delete crashes (but, again, my understanding of these topics very basic). Do you want me to give it one more go?

rcurtin: Nah, I'll just remove the extra lines during merge. :)

rcurtin:
Thanks again for the contribution.  Like the last PR, if you want to add your name to the list of contributors or receive stickers in the mail, just let me know. :+1:

Issue #1532, opened 2018-10-17
Error running ANN code


HelioStrike:
I'm trying to create an FFN to predict a continuous value. The code compiles but I get the following error when calling model.Train():
```
error: Mat::operator(): index out of bounds
terminate called after throwing an instance of 'std::logic_error'
  what():  Mat::operator(): index out of bounds
Aborted (core dumped)
```
Here's the network:
```
FFN<> model;
model.Add<Linear<> > (train_x.n_rows, 20);
model.Add<Linear<> > (20, 30);
model.Add<Linear<> > (30, train_y.n_rows);
model.train(train_x, train_y); //train_x and train_y are arma::mat-s of shapes (8, 2000) and  (1, 2000)
```
I found some issues similar to this, like #1525 and #1415, but couldn't solve it.
Any help would be appreciated. Thank you.

rcurtin:
Hi there,

By default the `FFN<>` class uses `NegativeLogLikelihood<>` as a loss function, which is for classification problems.  Since you're trying to predict a continuous value, this is now regression, so I think you'll need to use a loss function like mean squared error:

```
FFN<MeanSquaredError<>> model;
```

Hope this helps. :)

HelioStrike:
That did it. Thanks again @rcurtin ! :D

rcurtin:
Glad to help :)

Issue #1525, opened 2018-10-11
Can't compile ANN code from tutorial


GLmontanari:
Hi there, I'm trying to compile the source code from the ANN tutorial, the link is http://mlpack.org/docs/mlpack-3.0.3/doxygen/anntutorial.html

I'm using windows 10 with MSVS 2017, mlpack 3.0.3

I just put the code in my main() function and it looks like this:
```c++
int main() {
	// Load the training set.
	arma::mat dataset;
	mlpack::data::Load("thyroid_train.csv", dataset, true);
	// Split the labels from the training set.
	arma::mat trainData = dataset.submat(0, 0, dataset.n_rows - 4,
		dataset.n_cols - 1);
	// Split the data from the training set.
	arma::mat trainLabels = dataset.submat(dataset.n_rows - 3, 0,
		dataset.n_rows - 1, dataset.n_cols - 1);
	// Initialize the network.
	FFN<> model;
	model.Add<Linear<> >(trainData.n_rows, 8);
	model.Add<SigmoidLayer<> >();
	model.Add<Linear<> >(8, 3);
	model.Add<LogSoftMax<> >();
	// Train the model.
	model.Train(trainData, trainLabels);
	// Use the Predict method to get the assignments.
	arma::mat assignments;
	model.Predict(trainData, assignments);
	return 0;
        }
```
when trying to compile i get the following error:
E0442 too few arguments for class template "mlpack::ann::Linear"

I read the API documentation, but I'm not as expert as to tell what's wrong. Can someone help?

zoq:
Strange, I can't test this right now, so if you are willing to run some tests that would be great:
- can you compile if you just use a Linear layer:
```c++
Linear<> layer(10, 10);
```
- what if we provide the extra template arguments:
```c++
// Load the training set.
arma::mat dataset;
mlpack::data::Load("thyroid_train.csv", dataset, true);
// Split the labels from the training set.
arma::mat trainData = dataset.submat(0, 0, dataset.n_rows - 4,
	dataset.n_cols - 1);
// Split the data from the training set.
arma::mat trainLabels = dataset.submat(dataset.n_rows - 3, 0,
	dataset.n_rows - 1, dataset.n_cols - 1);
// Initialize the network.
FFN<> model;
model.Add<Linear<arma::mat, arma::mat> >(trainData.n_rows, 8);
model.Add<SigmoidLayer<> >();
model.Add<Linear<> >(8, 3);
model.Add<LogSoftMax<> >();
[...]

Issue #1415, opened 2018-06-01
Problem with the sample code in ann::ffn. trying to fit the data ("thyroid_train.csv") mlpack-3.0.1.


IshanSeidenberg:
Compilation:
g++ -DDEBUG -g fnn.cpp -o fnn -O2  -larmadillo -lmlpack -std=c++11 -fopenmp -lboost_serialization

Output:
24, 3772
21, 3772
3, 3772
Training...

error: Mat::operator(): index out of bounds
terminate called after throwing an instance of 'std::logic_error'
  what():  Mat::operator(): index out of bounds
Aborted (core dumped)

Code:
`
#include <fstream>
#include <cmath>
#include <mlpack/core.hpp>

#include <mlpack/methods/ann/layer/layer.hpp>
#include <mlpack/methods/ann/layer/layer_types.hpp>
#include <mlpack/methods/ann/loss_functions/mean_squared_error.hpp>
#include <mlpack/methods/ann/loss_functions/sigmoid_cross_entropy_error.hpp>
#include <mlpack/methods/ann/init_rules/random_init.hpp>
#include <mlpack/methods/ann/init_rules/const_init.hpp>
#include <mlpack/methods/ann/init_rules/nguyen_widrow_init.hpp>
#include <mlpack/methods/ann/ffn.hpp>
#include <mlpack/methods/ann/rnn.hpp>

using namespace mlpack;
using namespace std;
using namespace ann;
using namespace mlpack::ann;
using namespace optimization;

int main(int argc,char **argv)
{
 
    // Load the training set.
    arma::mat dataset;
    data::Load("thyroid_train.csv", dataset, true);
    // Split the labels from the training set.
    arma::mat trainData = dataset.submat(0, 0, dataset.n_rows - 4,
        dataset.n_cols - 1);
    // Split the data from the training set.
    arma::mat trainLabels = dataset.submat(dataset.n_rows - 3, 0,
        dataset.n_rows - 1, dataset.n_cols - 1);
    // Initialize the network.
    FFN<> model;
    model.Add<Linear<> >(trainData.n_rows, 8);
    model.Add<SigmoidLayer<> >();
    model.Add<Linear<> >(8, 3);
    model.Add<LogSoftMax<> >();
    // Train the model.
    StandardSGD opt(0.1, 1, trainData.n_cols /* 1 epoch */, -100);

    cout << dataset.n_rows << ", " << dataset.n_cols << std::endl;
[...]

PR #1534, opened 2018-10-17
Change thyroid dataset labels to NegativeLogLikelihood required encoding

rcurtin:
Previously the data in `tests/data/thyroid_train.csv` and `tests/data/thyroid_test.csv` had one-hot encoded labels as the last three columns.  This led to some confusion in #1525 and possibly #1532.  So I was going to update the documentation as suggested in #1415, but then I realized even the mlpack tests make this transformation.  So the easiest thing to do is just to make the thyroid dataset have its last column be `{1, 2, 3}` (i.e. not-one-hot-encoded labels) which can be used by `FFN<>` directly.

I also fixed a compilation warning that I found. :)


rcurtin:
I also added `layer.hpp` as a default include for both `ffn.hpp` and `rnn.hpp`.

rcurtin:
Ok, I will go ahead and merge this in 3 days then.

PR #1535, opened 2018-10-17
Explicitly serialize the 'reset' parameter.

rcurtin:
Previously, models would always re-initialize their weights after loading.  But we have to increment the serialization version for backwards compatibility.  I opened this because a user sent me an email that pointed out the problem.

@ShikharJ: does the GAN code need to be correspondingly updated?  I see that there is a serialize function there, but it seems very different in its behavior.

### DIFF
--- a/src/mlpack/methods/ann/ffn.hpp
+++ b/src/mlpack/methods/ann/ffn.hpp
@@ -462,6 +462,24 @@ class FFN
 
 } // namespace ann
 } // namespace mlpack
+
+//! Set the serialization version of the FFN class.  Multiple template arguments
+//! makes this ugly...
+namespace boost {
+namespace serialization {
+
+template<typename OutputLayerType,
+         typename InitializationRuleType,
+         typename... CustomLayer>
+struct version<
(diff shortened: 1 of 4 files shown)


ShikharJ:
@rcurtin This part of the GAN code has remained unchanged from Kris' original implementation, which followed from `FFN` class' implementation. I'll check and get back to you on this.

rcurtin:
@ShikharJ: just curious, have you had any chance to look into this?  I'd like to get this merged soon if we can, so I can release 3.0.4.  Just looking at the GAN code, it seems like a lot of variables held in the GAN class aren't serialized at all (i.e. `noiseDim`, `noiseFunctions`, `batchSize`, ...).

ShikharJ:
@rcurtin Please feel free to merge this, I'll upload a new PR with the changes for the `GAN`class, so that we won't have to serialize all of the variables. Though it follows a similar philosophy as the `FFN` class, as in the discriminator and generator would always re-initialize their parameters.

rcurtin:
Ok, I'll go ahead and merge this in 3 days.  We can handle the GAN class later. :)

ShikharJ:
Thanks for the contribution!

Issue #1526, opened 2018-10-11
Mean shift isolated single point issue;


alexxucui:

alexxucui:
@ShangtongZhang Thanks for implementing mean-shift. 

I have a question about one of my use case:
I have three input points of:

(-8.03939 -24.0719) 
(-8.03939 -24.0719) 
(-5.54168 -16.5931)

I set the radius to 0.3. Obviously there should be 2 clusters but the lib gives only 1 cluster. The centroid are at (-8.039389, -24.071899) which makes sense. Even I removed the last point the centroid is still the same. I have tried turn on and off the force convergence but the results is the same. It seems the algorithm will always force one single points to some centroid.

rcurtin:
Hey there @alexxucui, what is `min_points` set to?  Is the third point being labeled as 'noise' instead of a cluster?

alexxucui:
Hi @rcurtin I cannot find `min_points` you mentioned. Do you mean `minFreq` in the `GenSeeds`? And I cannot find 'noise'? I only get assignments and centroids as output.

alexxucui:
@rcurtin Thanks a lot.

rcurtin:
Hi @alexxucui, sorry my response took so long.  It took until today to find a few minutes to look into this one.  I found that the existing mean shfit implementation disallows clusters that hold only a single point, but I don't know any reason for this.  (Perhaps it could be a parameter or something.)

In any case, I opened PR #1536 which should fix the issue.  I can't think of any easy workaround with the existing codebase, so you'll either have to copy the patch into your code manually, wait until the PR is merged, or wait until the next release that incorporates the patch.  (I have been meaning to release 3.0.4 for quite a while now, so I will have to actually find the time to do that in the next few weeks.)

Hope this helps!  Let me know if I overlooked anything.  And thanks for the bug report. :+1:

rcurtin:
This should be fixed now that #1536 is merged.  Thanks again for pointing out the bug and feel free to open more issues if you encounter more problems. :)

PR #1536, opened 2018-10-18
Mean shift: allow single-point clusters

rcurtin:
The previous mean shift code would disallow a single cluster if it only held one point.  This fix allows single-point clusters.  @ShangtongZhang: I wouldn't mind if you could take a quick look at this and ensure it doesn't break anything unforeseen. :)

This should be a fix for issue #1526.

### DIFF
--- a/src/mlpack/methods/mean_shift/mean_shift_impl.hpp
+++ b/src/mlpack/methods/mean_shift/mean_shift_impl.hpp
@@ -68,7 +68,7 @@ EstimateRadius(const MatType& data, double ratio)
   arma::rowvec maxDistances = max(distances);
 
   // Calculate and return the radius.
-  return sum(maxDistances) / (double) data.n_cols;
+  return arma::sum(maxDistances) / (double) data.n_cols;
 }
 
 // Class to compare two vectors.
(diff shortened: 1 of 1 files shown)


ShangtongZhang:
I think the change is fine, thanks for helping this!

src/mlpack/methods/mean_shift/mean_shift_impl.hpp
```diff
@@ -265,8 +265,9 @@ inline void MeanShift<UseKernel, KernelType, MatType>::Cluster(
   // forcing convergence, take 1 random centroid calculated.
   if (centroids.empty())
   {
-    Log::Warn << "No clusters converge, setting 1 random centroid calculated. "
-    "Try a larger max_iterations or pass force_convergence flag." << std::endl;
+    Log::Warn << "No clusters converged; setting 1 random centroid calculated. "
```
zoq: The error message is really command line specific, perhaps a more general approach like ''increase the number of iterations or set force converge`` does work as well?

rcurtin: Sure, that can work too.  I guess the parameter may not always be called `max_iterations` or `force_convergence`, depending on the language being used.  And a simple approach like what you suggested should work fine.  Fixed in bf2e7ba73.

rcurtin:
The AppVeyor failure doesn't seem to be a build failure so I'll go ahead and merge.

Issue #1530, opened 2018-10-16
cotire + armadillo config + ninja = disaster


marcespie:
The cmake system must be missing some dependency somewhere.

I see the custom target that runs cmake -P CMake/CreateArmaConfigInfo.cmake
firing up, then erroring out because the preprocessed headers depend on armaconfig.

errors like this:

fatal error: file '/pobj/mlpack-3.0.3/mlpack-3.0.3/src/mlpack/core/util/arma_config.hpp' has been modified since the precompiled header '/pobj/mlpack-3.0.3/build-amd64/src/mlpack/tests/cotire/mlpack_test_CXX_prefix.hxx.pch' was built
note: please rebuild precompiled header '/pobj/mlpack-3.0.3/build-amd64/src/mlpack/tests/cotire/mlpack_test_CXX_prefix.hxx.pch'

To reproduce, you just need to run cmake -DCMAKE_MAKE_PROGRAM=ninja
and then have bad luck (Ninja does builds in random order in the absence of inter-dependence, so sometimes it builds, and sometimes not).

Scenario that breaks:
- ninja fires up a cotire rule first, such as
[1/533] cd /pobj/mlpack-3.0.3/build-amd64 && /usr/local/bin/cmake -DCOTIRE_BUILD_TYPE:STRING=Release -P /pobj/mlpack-3.0.3/mlpack-3.0.3/CMake/cotire.cmake combine /pobj/mlpack-3.0.3/build-amd64/src/mlpack/mlpack_CXX_cotire.cmake /pobj/mlpack-3.0.3/build-amd64/src/mlpack/cotire/mlpack_CXX_prefix.cxx /pobj/mlpack-3.0.3/mlpack-3.0.3/src/mlpack/prereqs.hpp
- ninja fires up the armaconfig rule:
[8/533] cd /pobj/mlpack-3.0.3/mlpack-3.0.3 && /usr/local/bin/cmake -D ARMADILLO_INCLUDE_DIR="/usr/local/include" -D ARMADILLO_VERSION_MAJOR="9" -D OPENMP_FOUND="FALSE" -P CMake/CreateArmaConfigInfo.cmake

- the first compile rule that uses cotire breaks, because the precompiled header depends on the empty armaconfig that was generated before armaconfig was rebuilt.

For once, I don't know exactly where to add the dependency.

Having ninja builds work is rather important, because ninja can actually parallelize over subdirectories, and thus can be MUCH faster than say, gmake builds...

[...]

PR #1537, opened 2018-10-18
Fix dependency ordering for mlpack_gitversion and mlpack_arma_config targets

rcurtin:
This is a fix for #1530, verified by @marcespie.  It fixes an intermittent problem when using Ninja to build mlpack.


rcurtin:
Ok, I'll go ahead and merge this in 3 days.

PR #1540, opened 2018-10-20
Fixed bug in the derivative of the exponential linear unit.

aruniyer:
The derivative of the exponential linear unit is discussed correctly in the comments, but the implementation had a bug.

### DIFF
--- a/src/mlpack/methods/ann/layer/elu.hpp
+++ b/src/mlpack/methods/ann/layer/elu.hpp
@@ -210,7 +210,7 @@ class ELU
    */
   double Deriv(const double y)
   {
-    return (y > 0) ? lambda : lambda * (Fn(y) + alpha);
+    return (y > 0) ? lambda : (Fn(y) + lambda * alpha);
   }
 
   /**
--- a/src/mlpack/methods/ann/layer/elu.hpp
+++ b/src/mlpack/methods/ann/layer/elu.hpp
@@ -210,7 +210,7 @@ class ELU
    */
   double Deriv(const double y)
   {
-    return (y > 0) ? lambda : lambda * (y + alpha);
+    return (y > 0) ? lambda : lambda * (Fn(y) + alpha);
   }
 
   /**


src/mlpack/methods/ann/layer/elu.hpp
```diff
@@ -210,7 +210,7 @@ class ELU
    */
   double Deriv(const double y)
   {
-    return (y > 0) ? lambda : lambda * (y + alpha);
+    return (y > 0) ? lambda : lambda * (Fn(y) + alpha);
```
rcurtin: My derivation suggests that `lambda * (Fn(y) + alpha)` is incorrect.  Instead, for x < 0,

```
f(x) = lambda * alpha * (e^x - 1);

f'(x) = lambda * alpha * e^x;
       = lambda * alpha * e^x - lambda * alpha + lambda * alpha;
       = f(x) + lambda * alpha;
```

which is a bit different than what you've got here.  But I do agree that the original implementation was not correct (and actually I think the LaTeX at the top of the class is also not correct).

rcurtin:
Hey @aruniyer, thanks for the contribution! :+1:  It looks like it doesn't pass the travis tests, so I think maybe something is wrong.  I definitely agree the original implementation was not right, so I am not sure yet if it passed the tests because the tests were also wrong, or if it passed the tests because it was lucky. :)

src/mlpack/methods/ann/layer/elu.hpp
aruniyer: *Duh* I should have derived it instead of going with the comments. But yes, as you said the earlier implementation was incorrect.

aruniyer:
> Hey @aruniyer, thanks for the contribution! +1 It looks like it doesn't pass the travis tests, so I think maybe something is wrong. I definitely agree the original implementation was not right, so I am not sure yet if it passed the tests because the tests were also wrong, or if it passed the tests because it was lucky. :)

Hi, I just checked the tests, they do seem to be wrong. Although, given the values in the tests folder, they probably should have failed even for the earlier implementation ... strange :-?

aruniyer:
> Hey @aruniyer, thanks for the contribution! +1 It looks like it doesn't pass the travis tests, so I think maybe something is wrong. I definitely agree the original implementation was not right, so I am not sure yet if it passed the tests because the tests were also wrong, or if it passed the tests because it was lucky. :)

The derivatives in the test look correct.

src/mlpack/methods/ann/layer/elu.hpp
zoq: For the ELU function ``lambda`` is set to 1: 

https://github.com/mlpack/mlpack/blob/713b7c678fce8dafab4a01905a8e4b4a56ca1bac/src/mlpack/methods/ann/layer/elu.hpp#L115-L123

see the note "Use this constructor for ELU activation function.".

so the ``f(x)' = f(x) + alpha;``, also note y is already the value of ``f(x)``.

aruniyer: @zoq then the condition y > 0 in the derivative function makes no sense.

rcurtin: Oh, I see, if `y == f(x)` then the original implementation seems to be correct.  But I don't see what will call `Deriv()` with `f(x)` instead of `x`... it doesn't look like `Deriv(const InputType& x, OutputType& y)` does this.  Maybe I overlooked something?

zoq: The ELU ``Backward`` method will call the function with: ``Deriv(f(x), OutputType& y)`` to avoid an unnecessary recalculation.

> @zoq  then the condition y > 0 in the derivative function makes no sense.
True, we should definitely fix that.

rcurtin: Ah, got it.  My opinion is that we should add some documentation to the `Deriv()` function then to point that out. :)

rcurtin:
@aruniyer: have you had a chance to look into the failing tests here?  Based on the discussion here, I think that the original implementation was correct so I'm not sure there's any reason to change it, but, it might be worthwhile to clarify the documentation.

ShikharJ:
@aruniyer I don't think either implementations are correct given their context of use. I have opened another PR for this issue, as it is a bit deep rooted. Closing this for now.

PR #1543, opened 2018-10-22
Decision tree pass by value

robertohueso:
Replace arguments of decision trees with pass by value arguments so the user can choose whether to copy or move the data. This makes reference to #1021.

### DIFF
--- a/src/mlpack/methods/decision_tree/decision_tree_main.cpp
+++ b/src/mlpack/methods/decision_tree/decision_tree_main.cpp
@@ -181,21 +181,29 @@ static void mlpackMain()
       arma::Row<double> weights =
           std::move(CLI::GetParam<arma::Mat<double>>("weights"));
       if (CLI::HasParam("print_training_error"))
+      {
         model->tree = DecisionTree<>(trainingSet, model->info, labels,
             numClasses, std::move(weights), minLeafSize, minimumGainSplit);
+      }
       else
+      {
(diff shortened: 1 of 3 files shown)


robertohueso:
Makes sense. Sorry I should have checked better too.

So should I fix it and open a new PR?

rcurtin:
Either updating this PR or a new PR is fine with me; whatever works best for you.  (Sorry for the slow response on this one.)

src/mlpack/methods/decision_tree/decision_tree_main.cpp
```diff
@@ -180,13 +180,22 @@ static void mlpackMain()
     {
       arma::Row<double> weights =
           std::move(CLI::GetParam<arma::Mat<double>>("weights"));
-      model->tree = DecisionTree<>(trainingSet, model->info, labels,
-          numClasses, weights, minLeafSize, minimumGainSplit);
+      if (CLI::HasParam("print_training_error"))
```
zoq: Picky style comment do you mind to use ``{}`` around the ``if/else`` condition, to improve the readability.

robertohueso: Fixed in 2b9d68a59ea6ab64e673a3eaa7fca6f4aebb42b7

ShikharJ:
Thanks for the contribution!

Issue #1538, opened 2018-10-18
Wrong HMM predict (state sequence) and LogLikelihood


aabghari:
Hi,
I have a HMM model. Sometime the state sequence estimation is wrong which all the states are zero for  example. I have checked the output with a python hmm implementation hmmlearn. hmmlearn seems to give the right answers for the same model and parameters. I have attached a cpp file which uses MLPACK and a python file which uses hmmlearn. Both have the same parameters and observation. I have put enough comment in the cpp file to follow. The version of the MLPACK is 3.0.3.

Thanks,

Arash
[hmm_test.zip](https://github.com/mlpack/mlpack/files/2493314/hmm_test.zip)

aabghari:
I think I know why it happens. The problem is the numerical instability when calculating logStateProb. I am going to provide you with a patch.

rcurtin:
Ah, great, thanks for digging in!  I have not had time to look into it---I was hoping I could maybe tonight or tomorrow.  But if you can provide a patch, it accelerates the whole bug fixing process :)

aabghari:
How do you want me to provide you with the patch?

rcurtin:
A pull request is fine, or if you prefer, you can just post the diff here and I can create the PR.  Whatever is most convenient for you. :+1:

aabghari:
Hi, 
Please find the attached diff file. Let me explained what I have done. I think there are two cases which are the root of the problem.
1. Using the emission probability  instead of the logprob directly. For example in Predict() there is this line which uses the log of the emission probability like this log(emission.probability()). If you look at the emission prob calculation, you see that it's actually calculated from the logprob by taking an exp(). So we have log()->exp()->log() which causes numerical instability.
2. In all probability calculation either Forward() or Backward() for example, All the values are getting multiplied with each other which makes them very small at the end.

To fix the above mentioned causes, I replaced all the prob calculations with the log version. Thus, instead of multiply, it's all addition now. It can even reduce the CPU consumption since everything is now addition and subtraction. To achieve this I have to add to new functionalities which are AddLog() and AccuLog(). I have put this functions under core/math/log_add.{cpp,hpp}. Feel free to restructure them according to mlpack policy. 
I have to add a LogProbability function to GMM distribution as well so it can work with the new log version of the HMM.
I have also added a test case. This is the one originally failed on my application and that's how I discovered the issue at the first place.
All the tests are passed with this patch.

Please let me know when you have accepted this patch and merged it into the main branch.

[...]

PR #1547, opened 2018-10-24
Patch from aabghari to work in log-space for probabilities.

rcurtin:
This is from #1538.  The probability computations for the HMM and GMM are now done entirely in log-space instead of the normal space, which helps with low-probability precision and correctness issues.

It seems that when using OpenMP (my system has 8 cores) this can result in a significant acceleration:

```
# OLD
$ time bin/mlpack_test -t HMMTest
Running 18 test cases...

*** No errors detected

real	0m6.463s
user	0m26.263s
sys	0m21.803s

# NEW
$ time bin/mlpack_test -t HMMTest
Running 19 test cases...

*** No errors detected

real	0m1.225s
user	0m2.020s
sys	0m2.788s
```

However, when setting the number of threads to 1, it's a slight slowdown:

```
# OLD
$ time OMP_NUM_THREADS=1 bin/mlpack_test -t HMMTest
Running 18 test cases...

*** No errors detected

real	0m0.808s
user	0m0.779s
sys	0m0.029s

# NEW
$ time OMP_NUM_THREADS=1 bin/mlpack_test -t HMMTest
Running 19 test cases...

*** No errors detected

real	0m1.050s
user	0m1.022s
sys	0m0.028s
```

It's possible that the code could be tuned so the singlethreaded performance matched the old singlethreaded performance, but I'm happy to leave it as-is because the multithreaded performance is so much better.

### DIFF
--- a/src/mlpack/core.hpp
+++ b/src/mlpack/core.hpp
@@ -264,6 +264,7 @@
 #include <mlpack/core/data/save.hpp>
 #include <mlpack/core/data/normalize_labels.hpp>
 #include <mlpack/core/math/clamp.hpp>
+#include <mlpack/core/math/log_add.hpp>
 #include <mlpack/core/math/random.hpp>
 #include <mlpack/core/math/random_basis.hpp>
 #include <mlpack/core/math/lin_alg.hpp>
--- a/src/mlpack/core/math/CMakeLists.txt
+++ b/src/mlpack/core/math/CMakeLists.txt
@@ -7,6 +7,8 @@ set(SOURCES
   lin_alg.hpp
   lin_alg_impl.hpp
   lin_alg.cpp
+  log_add.hpp
+  log_add_impl.hpp
   make_alias.hpp
   random.hpp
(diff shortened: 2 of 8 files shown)


aabghari:
The slight difference in the running time with one thread could be due to the number of test cases. In the old one there is 18 tests while the new one has 19 test cases.

aabghari:
Hi, 

I am sorry but I noticed that I had uploaded rather a little bit older patch. I now attached the new patch. The only difference is that in the new patch I have added LogEstimate() function to hmm.hpp so Estimate() still returns the probability not the logprob and inside the Train() and Smooth()
  LogEstimate() gets called instead of Estimate(). The advantage of this is that HMM API remains unchanged and the new version does not break the compatibility.

[hmm2.patch.zip](https://github.com/mlpack/mlpack/files/2512832/hmm2.patch.zip)

rcurtin:
No problem, I updated the code and opened #1551.  Let me know if there are any issues. :)

Issue #1544, opened 2018-10-22
More dependency issues in CMakelists


marcespie:
found by a co-developer on OpenBSD...  custom commands should have BYPRODUCTS when required so that dependencies actually work.

(2 patches attached)
[patch2.txt](https://github.com/mlpack/mlpack/files/2501647/patch2.txt)
[patch1.txt](https://github.com/mlpack/mlpack/files/2501648/patch1.txt)

This makes ninja build break half-randomly again.

rcurtin:
Thanks for the fixes!  The patches look good to me, but before I prepare to merge them I have to figure out what to do about CMake reverse compatibility.  Our minimum version is 2.8 but `BYPRODUCTS` appears in something like 3.3 or so.  So I just need to test that there's not a failure on 2.8---or bump the minimum version.  I'll open a PR shortly, but it may be a little bit until I am able to find the right reverse compatibility fix.

marcespie:
On Tue, Oct 23, 2018 at 11:50:22AM -0700, Ryan Curtin wrote:
> 
>    Thanks for the fixes! The patches look good to me, but before I prepare
>    to merge them I have to figure out what to do about CMake reverse
>    compatibility. Our minimum version is 2.8 but BYPRODUCTS appears in
>    something like 3.3 or so. So I just need to test that there's not a
>    failure on 2.8---or bump the minimum version. I'll open a PR shortly,
>    but it may be a little bit until I am able to find the right reverse
>    compatibility fix.

Err, cmake 3.3 is over 3 years old by this point.

I surmise about everyone has updated already ?

rcurtin:
It's true, but I want to make sure the software can still be built on RHEL6 and Ubuntu 14.04, since a lot of people especially at universities with slow-moving IT processes can be stuck on that.  When I checked yesterday, I saw that the cmake package in the trusty repositories was 2.8.12.2... but what I did not see until I double-checked just now is that there is a cmake3 package that is sufficiently new, and this is also the situation for EPEL6.

So, I'll just update the minimum required CMake version in the PR.  :)

rcurtin:
Thanks again, PR opened as #1550.

rcurtin:
#1550 was merged a little while ago, so I'll go ahead and close this now.

PR #1550, opened 2018-10-25
Upgrade CMake minimum to 3.3.2 and a few Ninja fixes

rcurtin:
This is an extension of the patches provided by @marcespie in #1544.  Since the `BYPRODUCTS` argument didn't exist until CMake 3.3.2, I also made some changes to bump the minimum version of CMake to 3.3.2 and simplified the CMake configuration a little bit.

CMake 3.3.2 or newer is available in even the oldest places I looked that we should bother to support (Ubuntu 14.04 and RHEL 6) as `cmake3`.


rcurtin:

PR #1551, opened 2018-10-26
Patch from aabghari to work in log-space for probabilities.

rcurtin:
This is from #1538. The probability computations for the HMM and GMM are now done entirely in log-space instead of the normal space, which helps with low-probability precision and correctness issues.

This is an updated version of #1547.


src/mlpack/core/math/log_add.hpp
```diff
@@ -0,0 +1,40 @@
+/**
+ * @file log_add.hpp
+ * @author Arash Abghari
+ *
+ * Functions for logarithmic addition.
+ */
+#ifndef MLPACK_CORE_LOG_ADD_HPP
+#define MLPACK_CORE_LOG_ADD_HPP
```
zoq: Picky style comment, I think it would be a good idea to use adapt the header guard here, just to be conform with the existing code.

src/mlpack/core/math/log_add_impl.hpp
```diff
@@ -0,0 +1,70 @@
...
+ */
+#ifndef MLPACK_CORE_MATH_LOG_ADD_IMPL_HPP
+#define MLPACK_CORE_MATH_LOG_ADD_IMPL_HPP
+
+#include "log_add.hpp"
+
+namespace mlpack {
+namespace math {
+
+ /** Internal log-addition
```
zoq: This looks off.

src/mlpack/core/math/log_add_impl.hpp
```diff
@@ -0,0 +1,70 @@
...
+ *
+ * e^z = e^x + e^y
```
zoq: Should we use ``@f`` to make this more doxygen friendly?

src/mlpack/core/math/log_add_impl.hpp
```diff
@@ -0,0 +1,70 @@
...
+ *    when x > y, z = x + log(1 + e^{y-x})
+ * 
+ * @param x log value
+ * @param y log value
+ * @return log(e^x + e^y)
+ */
+template<typename T>
+T LogAdd(T x, T y)
+{
+  double d, r;
```
zoq: Looks like this should be ``T`` as well.

rcurtin:
Ok---I'll go ahead and merge this in 3 days. :)

Issue #1524, opened 2018-10-10
Fail in building MLPACK with MS Visual Studio 2015


jbensabat:
Hello 
I have tried to build mlpack following the instructions. 
I managed to download the packages (openblas and boost libs)
however CMake fails in the configuration phase of mlpack 
one of the errors is that boost libs are missing but they are there 
is there any consistent and complete guide on how to install mlapck from MSVC ?
thanks
jac

rcurtin:
Did you try reading http://www.mlpack.org/docs/mlpack-git/doxygen/build_windows.html ?

jbensabat:
Hi
Thanks for your message
This is where I started from
Best
Jac


Jacob Bensabat, PH.D.
CTO
EWRE Ltd.
Tel: +972-4-8383919.
Fax: +972-4-8387621.
Mobile: +972-544-417511.

rcurtin:
Ok, well, I can try and help out if you provide some details on what you have done.  That tutorial page I linked you to was just created and tested recently, so it should work.

jbensabat:
Hi
This is what I have done


1.       Creating the mlpack directory , then mlpack-3.0.3

2.       Unpacked the mlpack-master.zip to mlpack-3.0.3

3.       Created a MSVC solution from code – named mlpack in directory mlpack-3.0.3

4.       Used NuGet to download openblas and boost (version 1.68.00)

5.       Put all the compile boost libs inside a directory called boost_libs within the packages directory

6.       Put all the compiled libs into one lib

7.       create a directory armadillo inside mlpack

8.       download and unpacked armadillo into the armadillo directory

9.       created a build directory inside armadillo

tried to run the following bacth

cmake -G "Visual Studio 14 2015 Win64" ^
-DBLAS_LIBRARY:FILEPATH="C:/software/mlpack/mlpack-3.0.3/packages/OpenBLAS.0.2.14.1/lib/native/lib/x64/libopenblas.dll.a" ^
-DLAPACK_LIBRARY:FILEPATH="C:/software/mlpack/mlpack-3.0.3/packages/OpenBLAS.0.2.14.1/lib/native/lib/x64/libopenblas.dll.a" ^
-DCMAKE_PREFIX:FILEPATH="C:/software/mlpack/armadillo" ^
-DBUILD_SHARED_LIBS=OFF ..

I got an error – Cmakelists.txt is not present in the build directory.
Then I moved the batch file to the root directory of armadillo and I was able to build armadillo

[...]

PR #1553, opened 2018-10-31
Update Windows tutorial and build options

rcurtin:
This is to improve on the issues encountered in #1524.  It adds some description on debugging Boost configuration failures, and also disables autolinking with Boost which apparently causes some problems.

I have not tried these steps directly on Windows but instead on Linux and they seem to work.  If someone would like to verify (and update if needed on what's wrong), I can update the Boost configuration steps.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -327,6 +327,14 @@ find_package(Boost 1.49
 
 link_directories(${Boost_LIBRARY_DIRS})
 
+# In Visual Studio, automatic linking is performed, so we don't need to worry
+# about it.  Clear the list of libraries to link against and let Visual Studio
+# handle it.
+if (MSVC)
+  link_directories(${Boost_LIBRARY_DIRS})
+  set(Boost_LIBRARIES "")
+endif ()
+
 set(MLPACK_INCLUDE_DIRS ${MLPACK_INCLUDE_DIRS} ${Boost_INCLUDE_DIRS})
 set(MLPACK_LIBRARIES ${MLPACK_LIBRARIES} ${Boost_LIBRARIES})
 set(MLPACK_LIBRARY_DIRS ${MLPACK_LIBRARY_DIRS} ${Boost_LIBRARY_DIRS})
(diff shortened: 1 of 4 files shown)


rcurtin:
Ah, good idea.  Updated.

CMakeLists.txt
```diff
@@ -325,12 +327,10 @@ find_package(Boost 1.49
...
 link_directories(${Boost_LIBRARY_DIRS})
 
-# In Visual Studio, automatic linking is performed, so we don't need to worry
-# about it.  Clear the list of libraries to link against and let Visual Studio
-# handle it.
+# Disable automatic linking in Visual Studio.
 if (MSVC)
   link_directories(${Boost_LIBRARY_DIRS})
-  set(Boost_LIBRARIES "")
+  add_definitions(-DBOOST_ALL_NO_LIB)
```
zoq: Do you have to define ``BOOST_ALL_DYN_LINK`` as well?

rcurtin: I'm not 100% sure.  My impression was that by defining `BOOST_ALL_NO_LIB`, then Visual Studio would not automatically link the libraries, but CMake would configure the project to manually link the Boost libraries.  If you know more than me on this one, I'm happy to add `BOOST_ALL_DYN_LINK` also.

zoq: I just noticed the build failed and the only difference I can see between #1524 and this PR is the ``BOOST_ALL_DYN_LINK`` option. Not sure this will fix the windows build, maybe we have to adapt the build steps as well.

rcurtin: Oh, I see, I didn't realize the build was failing.  I'm not sure why I didn't notice that...

I'll try some debugging and see where it goes. :)

rcurtin:
I ended up removing the changes to Boost's autolinking because I couldn't get it to build right on Windows.  So now this just adds newer versions of Boost to `Boost_ADDITIONAL_VERSIONS`, and makes a couple of changes to what gets linked against.  The big change is the Windows configuration documentation, though.  I have no more changes to make here, so from my end this is ready.

PR #1556, opened 2018-11-08
CMake doesn't allow overloading of an option.

rcurtin:
So we can't "re-set" the default value.  This change makes the option defined with its default conditionally.  No actual functionality change here, other than that now Python bindings are properly disabled on Windows.


rcurtin:
I realized while debugging that our code currently forces Python to be found incorrectly, so I committed d44ef5d which removes this.  This PR will have to be merged before 3.0.4 is released then.

PR #1557, opened 2018-11-08
Add the Cauchy kernel

laivtom:
Added an implementation and tests for the Cauchy kernel described by Basak (2008) in the article _A Least Square Kernel Machine With Box Constraints_.

### DIFF
--- a/src/mlpack/core/kernels/cauchy_kernel.hpp
+++ b/src/mlpack/core/kernels/cauchy_kernel.hpp
@@ -1,0 +1,90 @@
+/**
+ * @file cauchy_kernel.hpp
+ * @author Tommi Laivamaa
+ *
+ * Implementation of the Cauchy kernel (CauchyKernel),
+ *
+ * mlpack is free software; you may redistribute it and/or modify it under the
+ * terms of the 3-clause BSD license.  You should have received a copy of the
+ * 3-clause BSD license along with mlpack.  If not, see
+ * http://www.opensource.org/licenses/BSD-3-Clause for more information.
+ */
+#ifndef MLPACK_CORE_KERNELS_CAUCHY_KERNEL_HPP
+#define MLPACK_CORE_KERNELS_CAUCHY_KERNEL_HPP
+
+#include <mlpack/prereqs.hpp>
+#include <mlpack/core/metrics/lmetric.hpp>
+#include <mlpack/core/kernels/kernel_traits.hpp>
+
+namespace mlpack {
+namespace kernel {
+
+/**
+ * The Cauchy kernel. Given two vector @f$ x @f$, @f$ y @f$, and a bandwidth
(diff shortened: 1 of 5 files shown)


src/mlpack/core/kernels/cauchy_kernel.hpp
```diff
@@ -0,0 +1,90 @@
...
+namespace mlpack {
+namespace kernel {
+
+/**
+ * The Cauchy kernel. Given two vector @f$ x @f$, @f$ y @f$, and a bandwidth
+ * @f$ \sigma @f$ (set in the constructor),
+ *
+ * @f[
+ * K(x, y) = \frac{1}{1 + (\frac{|| x - y ||}{\sigma})^2}.
+ * @f]
```
rcurtin: Want to add a BibTeX reference here to the paper the kernel is from?

src/mlpack/core/kernels/cauchy_kernel.hpp
```diff
@@ -0,0 +1,90 @@
...
+ */
+class CauchyKernel
+{
+ public:
+  /**
+   * Default constructor; sets bandwidth to 1.0.
+   */
+  CauchyKernel(): bandwidth(1.0)
```
rcurtin: Tiny picky style comment: can you change `CauchyKernel():` to `CauchyKernel() : `?

src/mlpack/core/kernels/cauchy_kernel.hpp
```diff
@@ -0,0 +1,90 @@
...
+   * @tparam VecTypeB Type of second vector (arma::vec, arma::sp_vec).
+   * @param a First vector.
+   * @param b Second vector.
+   * @return K(a, b).
+   */
+  template<typename VecTypeA, typename VecTypeB>
+  double Evaluate(const VecTypeA& a, const VecTypeB& b)
+  {
+    return 1 / (1 + (
+        pow(metric::EuclideanDistance::Evaluate(a, b) / bandwidth, 2)));
```
rcurtin: I think this needs to be `std::pow()` to be properly portable.

laivtom:
Thank you for the prompt review, @rcurtin :) Everything mentioned should be fixed now.

src/mlpack/core/kernels/cauchy_kernel.hpp
```diff
@@ -0,0 +1,104 @@
...
+  /**
+   * Default constructor; sets bandwidth to 1.0.
+   */
+  CauchyKernel() : bandwidth(1.0)
+  { }
+
+  /**
+   * Construct the Cauchy kernel with a custom bandwidth.
+   */
+  CauchyKernel(double bandwidth) : bandwidth(bandwidth)
```
zoq: We could probably combine both constructors, and use ``bandwidth = 1.0`` as the default value.

PR #1558, opened 2018-11-12
Improve GAN Serialization

ShikharJ:
I'll need to test this extensively if the tests pass. @rcurtin @zoq I'd like to keep this within the next release, so reviews / suggestions are welcome :)

### DIFF
--- a/src/mlpack/methods/ann/gan/gan_impl.hpp
+++ b/src/mlpack/methods/ann/gan/gan_impl.hpp
@@ -30,7 +30,6 @@ template<
   typename PolicyType
 >
 GAN<Model, InitializationRuleType, Noise, PolicyType>::GAN(
-    arma::mat& predictors,
     Model generator,
     Model discriminator,
     InitializationRuleType& initializeRule,
@@ -42,7 +41,6 @@ GAN<Model, InitializationRuleType, Noise, PolicyType>::GAN(
     const double multiplier,
     const double clippingParameter,
     const double lambda):
-    predictors(predictors),
     generator(std::move(generator)),
     discriminator(std::move(discriminator)),
     initializeRule(initializeRule),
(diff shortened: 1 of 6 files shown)


rcurtin:
@ShikharJ: I actually realized that the GAN code is not in 3.0.3 and I wasn't planning on releasing it in 3.0.4.  So when this is integrated, I think that the release that would include it would be 3.1.0.  Basically a patch version bump should only have bugfixes and really minor changes, but a minor version bump we can add new functionality.  So is it ok if I do a 3.0.4 release without waiting on this, and then we can do 3.1.0 which will include this?

src/mlpack/methods/ann/gan/gan_impl.hpp
```diff
@@ -425,6 +439,15 @@ serialize(Archive& ar, const unsigned int /* version */)
...
   ar & BOOST_SERIALIZATION_NVP(noiseFunction);
+  ar & BOOST_SERIALIZATION_NVP(noiseDim);
+  ar & BOOST_SERIALIZATION_NVP(numFunctions);
+  ar & BOOST_SERIALIZATION_NVP(numDimensions);
+  ar & BOOST_SERIALIZATION_NVP(batchSize);
+  ar & BOOST_SERIALIZATION_NVP(generatorUpdateStep);
+  ar & BOOST_SERIALIZATION_NVP(preTrainSize);
+  ar & BOOST_SERIALIZATION_NVP(multiplier);
+  ar & BOOST_SERIALIZATION_NVP(clippingParameter);
+  ar & BOOST_SERIALIZATION_NVP(lambda);
```
rcurtin: Should we call `Reset()` when we are loading?  Otherwise I have some concerns that `predictors` and `responses` could be out of date.

If you want to check whether or not the archive is loading or saving you can use the constexpr boolean values `Archive::is_loading::value` and `Archive::is_saving::value`.

ShikharJ:
@rcurtin Okay, I'll anyways need some time to debug and optimize this. Please go ahead with the release.

src/mlpack/methods/ann/gan/gan_impl.hpp
ShikharJ: @rcurtin I have changed the code in such a way that whenever the model gets loaded, and `Train()` is executed, internally it gives a call to `ResetData()` which initializes the matrix sizes for the predictors and responses, and later calls `Reset()` which initializes the weight of the parameters. So I think, if need be, we can add a call to `ResetData()` here in place of `Reset()`?

rcurtin: That seems reasonable, but a user may also call `Predict()` or any other methods right after loading, so we should be sure that the GAN is in a ready-to-use state after serialization.  If all those methods do work, then no problem. :)

You can add a pretty easy serialization test where you train a GAN, then serialize/deserialize and do predictions with it, and make sure the predictions are the same as the original.  Also you could do the same except then also do a little training after deserializing.  Let me know if you want me to clarify, I know I am not being totally precise with what I am writing here :)

shardulparab97:
@ShikharJ have begun working on this and will get back to you soon! :)

shardulparab97:
@ShikharJ @rcurtin I am unable to get a clear picture of the issue. What is expected in order to improve the GAN serialization? Thanks! :)

ShikharJ:
@shardulparab97 The issue with the current GAN implementation is that serialization doesn't happen for a number of parameters, and hence if someone wishes to load a pre-initialized model and continue further training on it, we don't have a support for that. Hence, it is expected to add that support in a way that whenever someone loads a pre-trained model, the required checks and initialization is done, so that there's practically no difference in a saved and loaded model, and a model solely trained using our framework. Let me know if I should clarify.

walragatver:
Hi @ShikharJ,
  Can I try to fix this? I have seen the code in `gan` folder. Also I am keeping `FFN` serialization as a reference.

ShikharJ:
@walragatver Sure, take it up for a spin, and see if you can get it to work. I'll be mostly pre-occupied with other PRs for a few days.

ShikharJ:
Closing this in favour of #1770.

PR #1561, opened 2018-11-13
decision_stump_impl.hpp has a bug.

rickllyxu:
[Here we should split a dimension relying on its values.](https://github.com/mlpack/mlpack/commit/9c5c0e53c3854b41f76c4df88937275cd3a4699a) 
 please refer above link to say what difference.

As comments saying, We need to split a dimension relying on its values. For example, if the values of a dimension are 1, 1, 1, 2, 2, 3, 3, 3, then we split it to 3 bins.

Please review code, thanks.


rcurtin:
Closed in favor of #1562 (I think that is the PR you meant to open).

PR #1562, opened 2018-11-13
decision_stump has a bug. Here we should split a dimension relying on its values.

rickllyxu:
[Here we should split a dimension relying on its values](https://github.com/mlpack/mlpack/commit/3f23fe66ca059cecfda67aaaaa816dd33cba0bdc).
please refer above link to say what difference.

As comments saying, We need to split a dimension relying on its values. For example, if the values of a dimension are 1, 1, 1, 2, 2, 3, 3, 3, then we split it to 3 bins.

Please review code, thanks.


rcurtin:
Hi @rickllyxu, thanks for the contribution.

I'm not sure if the changes you've suggested are correct---they cause lots of failures in the different decision stump tests.  I'm also not sure that there was a problem in the first place.  If you're sure that something was wrong with the previous code, can you clarify more what the issue was and why it was wrong?  Thanks! :+1:

rickllyxu:
@rcurtin sorry for being unfamiliar with git...  
In fact, I fix some points including: 
1. spliting a dimension relying on its values. just as [last commit did](https://github.com/mlpack/mlpack/pull/1562/commits/3f23fe66ca059cecfda67aaaaa816dd33cba0bdc)
2. entropy calculating. original code did not multiple by  `-1`, it is not correct. I fix it.
3. `decision_stump_test.cpp` has some errors. I modified it and add some comment.

Issue #1564, opened 2018-11-13
inappropriate perceptron option warning


barak:
````
$ mlpack_perceptron --input_model_file mlpack_perceptron.bin --test_file test.csv --output_file test-labels1.csv
[WARN ] '--output_file (-o)' ignored because '--test_file (-T)' is specified!
````
Hey, this is exactly like on `mlpack_perceptron(1)`!
````
$ man mlpack_perceptron | egrep 'test_file.*output_file'
       $ perceptron --input_model_file perceptron_model.bin --test_file test_data.csv --output_file predictions.csv
````

rcurtin:
Opened #1566.  Thanks for pointing this out. :+1:

rcurtin:
Fixed since #1566 is merged.  Thanks again for the report. :)

PR #1566, opened 2018-11-13
Fix inappropriate warning Barak pointed out in #1564.

rcurtin:
The warning should be issued if `output` is specified but `test` is not... not the other way around.


rcurtin:
It's been 3 days so I'll go ahead and merge this now.

PR #1568, opened 2018-11-14
Add the cosine distance

laivtom:


rcurtin:
We already have `src/mlpack/core/kernels/cosine_distance.hpp` (although technically that implements the cosine similarity).  Could you use that instead?

I'm a bit hesitant to put the cosine similarity in `metrics/` because it doesn't satisfy the triangle inequality and thus can't be used with a lot of the tree data structures.  (That doesn't mean we can't do it, we just need to discuss what the best option is a bit. :))

laivtom:
Definitely bit of an oversight on my part, adding the cosine distance definitely does not make much sense since the cosine similarity is available :)

About these kinds of distance/similarity measures that are not proper metrics generally; are there plans to add a separate namespace for them? I know I have personally worked on problems where they can be useful and as such might be nice to have available if others feel they have value as well. It definitely makes a lot of sense to conserve the metric-namespace for proper metrics only.

PR #1570, opened 2018-11-15
ensmallen (optimizer transition)

zoq:
Strip out the existing optimizer framework and use ensmallen.


src/mlpack/core/optimizers/CMakeLists.txt
```diff
@@ -1,41 +1,12 @@
...
-  ada_delta
-  ada_grad
-  adam
-  aug_lagrangian
-  cmaes
-  bigbatch_sgd
-  cne
-  fw
   gradient_descent
   grid_search
```
rcurtin: These are both a part of ensmallen also; we could probably remove them from here too.  I think there was a slight change to the APIs of each, but I can adapt that code if you like.

CMakeLists.txt
```diff
@@ -275,7 +275,8 @@ if (WIN32)
 endif ()
 
 # Include directories for the previous dependencies.
-set(MLPACK_INCLUDE_DIRS ${MLPACK_INCLUDE_DIRS} ${ARMADILLO_INCLUDE_DIRS})
+set(MLPACK_INCLUDE_DIRS ${MLPACK_INCLUDE_DIRS} ${ARMADILLO_INCLUDE_DIRS}
+    ${CMAKE_SOURCE_DIR}/src/mlpack/core/optimizers/ensmallen)
```
rcurtin: Do you think it might be possible to use CMake to check if ensmallen is available on the target system and then, e.g., use a git submodule or download and unpack the code and something like this if it isn't available?  I can probably write some code to do this in CMake, the trick is just figuring out the right way to do it.

Keeping ensmallen in the repo like this is a bit of a maintenance burden that ideally I'd like to avoid. :)

We can probably expect ensmallen to become readily available in different package managers soon, but it will probably take a while.

src/mlpack/core/optimizers/CMakeLists.txt
zoq: Right, both optimizer make use of mlpack's ``DatasetMapper`` class, which isn't part of ensmallen; I guess it would be nice to have some option to extend the optimizer class in the same way it is possible for armadillo, but I guess if we can adapt the existing code to use the class, there is no need.

CMakeLists.txt
zoq: The main issue with git submodule or a download step, is that you can't just download mlpack, and build it without an internet connection. Maybe the case doesn't really exist in the real world.

rcurtin: That's true.  Over time I think we can expect that people have ensmallen installed, but at the moment it would be another dependency that might not be available in people's package managers.  So maybe we could do the following:

 - For a release, package `ensmallen-x.y.z.tar.gz` (the newest version) in `deps/`.  CMake can try to find ensmallen on the system but if it is not available then it will unpack and use the one in `deps/`.
 - For a git repository, CMake can try to download the newest version of ensmallen if it isn't available from the website.

What do you think, would that be more reasonable?

zoq: That sounds like a good idea to me.

>  I can probably write some code to do this in CMake, the trick is just figuring out the right way to do it.

If you have the time, I guess we could just open a new PR and merge this one after the CMake modification?

src/mlpack/core/optimizers/CMakeLists.txt
rcurtin: https://gist.github.com/rcurtin/bbb9aa2ff5648626d8bd373ec3227e5f should fix it.  It incorporates a patch that I just opened for ensmallen that hasn't been merged yet.

CMakeLists.txt
rcurtin: Sure, I think that I can write this CMake script by tomorrow.

rcurtin: Here is an attempt:

https://gist.github.com/acd87095d44d73d1355a81a89dd9b87d 

It is not fully debugged but I am about to be on a really long flight with no internet so I wanted to get it uploaded. :) Let me know if there are problems and I will try to debug when I am able.

src/mlpack/tests/hpt_test.cpp
```diff
@@ -218,10 +218,19 @@ BOOST_AUTO_TEST_CASE(GridSearchTest)
...
   arma::mat actualParameters;
-  double actualObjective =
-      optimizer.Optimize(cvFun, actualParameters, datasetInfo);
+
+  std::vector<bool> categoricalDimensions(datasetInfo.Dimensionality());
+  arma::Row<size_t> numCategories(datasetInfo.Dimensionality());
+  for (size_t d = 0; d < datasetInfo.Dimensionality(); d++)
+  {
+    numCategories[d] = datasetInfo.NumMappings(d);
+    categoricalDimensions[d] = datasetInfo.Type(d) != mlpack::data::Datatype::categorical;
```
rcurtin: I think this should be `==` not `!=` and that should fix the failing test.

src/mlpack/core/hpt/hpt_impl.hpp
```diff
@@ -87,12 +87,22 @@ void HyperParameterTuner<MLAlgorithm,
 {
   static const size_t totalArgs = std::tuple_size<ArgsTuple>::value;
 
+  std::vector<bool> categoricalDimensions(datasetInfo.Dimensionality());
+  arma::Row<size_t> numCategories(datasetInfo.Dimensionality());
+  for (size_t d = 0; d < datasetInfo.Dimensionality(); d++)
+  {
+    numCategories[d] = datasetInfo.NumMappings(d);
+    categoricalDimensions[d] = datasetInfo.Type(d) !=
+        mlpack::data::Datatype::categorical;
```
rcurtin: I think this too should be `==` not `!=`.

src/mlpack/tests/hpt_test.cpp
```diff
@@ -320,8 +329,8 @@ BOOST_AUTO_TEST_CASE(HPTGradientDescentTest)
   // We pass LARS just because some ML algorithm should be passed. We pass MSE
   // to tell HyperParameterTuner that the objective function (QuadraticFunction)
   // should be minimized.
-  HyperParameterTuner<LARS, MSE, QuadraticFunction, GradientDescent>
-      hpt(a, b, c, d, xMin, yMin, zMin);
+  HyperParameterTuner<LARS, MSE, QuadraticFunction,
+      GradientDescent> hpt(a, b, c, d, xMin, yMin, zMin);
```
rcurtin: It looks like the wrong `categoricalDimensions` values end up getting passed to `ens::GradientDescent` here, but I haven't figured out why.  I'll try to take a look into it later tonight.

rcurtin: Actually, this is fixed by the `!=` -> `==` changes, but there is one more issue: `CVFunction` doesn't properly unmap values using the `DatasetInfo` object.  This used to be done inside of `GridSearch`, but now that `GridSearch` is a separate part of ensmallen that doesn't know anything about `DatasetInfo`, we have to put that logic inside `CVFunction`.  I'll write up a patch tomorrow.  (Probably while waiting in the registration line for NeurIPS...)

rcurtin: Well the registration line was short so I didn't get it done there but here is a patch:

https://gist.github.com/rcurtin/81294d3266175005cd59f4ecdb06e0b7

zoq: hm, I guess the test passed on your system, so maybe I missed something.

rcurtin: Ah, I did not include the `!=` -> `==` changes in my patch---it looks like that is the issue.

zoq: Yeah, let me fix that.

rcurtin:
Really happy to have this merged. :+1:

Issue #1572, opened 2018-11-18
Incorrect formula for derivative of logistic_function on vectors


ajaysub110:
template<typename InputVecType, typename OutputVecType>
  static void Deriv(const InputVecType& y, OutputVecType& x)
  {
    x = y % (1.0 - y);
  }
This is the current function for calculating the derivative of x = 1/(1+e^-y) (logistic function used between layers in anns). I think it is incorrect and should be changed to x = y*(1.0-y)

ajaysub110:
Can I take it up?

rcurtin:
This is not an error. Please read the Armadillo documentation. `%` is elementwise multiplication.

PR #1574, opened 2018-11-18
added fix for #1572

ajaysub110:


rcurtin:
See my comment in #1572

PR #1577, opened 2018-11-19
Decision stump fix

rickllyxu:
Hope this version convenience for you.

Just as said in [pull 1562](https://github.com/mlpack/mlpack/pull/1562), the `decision_stump_impl.cpp` has bugs. Please code review.
Also, the corresponding test case has some errors. I fix it.

### DIFF
--- a/src/mlpack/methods/decision_stump/decision_stump_impl.hpp
+++ b/src/mlpack/methods/decision_stump/decision_stump_impl.hpp
@@ -266,7 +266,7 @@ double DecisionStump<MatType>::SetupSplitDimension(
           sortedLabels.subvec(begin, end), sortedWeights.subvec(begin, end));
       i++;
     }
-    else if (sortedLabels(i) != sortedLabels(i + 1))
+    else if (dimension(sortedIndexDim(i)) != dimension(sortedIndexDim(i + 1)))
     {
       // If we're not at the last element of sortedLabels, then check whether
       // count is less than the current bucket size.
@@ -344,7 +344,7 @@ void DecisionStump<MatType>::TrainOnDim(const VecType& dimension,
 
       i++;
     }
-    else if (sortedLabels(i) != sortedLabels(i + 1))
+    else if (sortedSplitDim(i) != sortedSplitDim(i + 1))
     {
       if (count < bucketSize)
       {
@@ -508,7 +508,7 @@ double DecisionStump<MatType>::CalculateEntropy(
     }
   }
 
-  return entropy / std::log(2.0);
+  return -1.0 * entropy / std::log(2.0);
 }
 
 } // namespace decision_stump
--- a/src/mlpack/tests/decision_stump_test.cpp
+++ b/src/mlpack/tests/decision_stump_test.cpp
@@ -203,7 +203,9 @@ BOOST_AUTO_TEST_CASE(MultiClassSplit)
 
   mat testingData;
   testingData << -6.1 << -5.9 << -2.1 << -0.7 << 2.5 << 4.7 << 7.2 << 9.1;
-
+  // After training, we will get 6 bins before merging.
+  // bins: x < -4; -4 <= x < -1; -1 <= x < 2; 2 <= x < 5; 5 <= x < 8; 8 <= x
+  // labels:   0;            1;            0;          1;          2;      2
   DecisionStump<> ds(trainingData, labelsIn.row(0), numClasses, inpBucketSize);
 
   Row<size_t> predictedLabels;
@@ -212,7 +214,7 @@ BOOST_AUTO_TEST_CASE(MultiClassSplit)
   BOOST_CHECK_EQUAL(predictedLabels(0, 0), 0);
   BOOST_CHECK_EQUAL(predictedLabels(0, 1), 0);
   BOOST_CHECK_EQUAL(predictedLabels(0, 2), 1);
-  BOOST_CHECK_EQUAL(predictedLabels(0, 3), 1);
+  BOOST_CHECK_EQUAL(predictedLabels(0, 3), 0);
   BOOST_CHECK_EQUAL(predictedLabels(0, 4), 1);
   BOOST_CHECK_EQUAL(predictedLabels(0, 5), 1);
   BOOST_CHECK_EQUAL(predictedLabels(0, 6), 2);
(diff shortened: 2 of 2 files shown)


src/mlpack/methods/decision_stump/decision_stump_impl.hpp
```diff
@@ -266,7 +266,7 @@ double DecisionStump<MatType>::SetupSplitDimension(
           sortedLabels.subvec(begin, end), sortedWeights.subvec(begin, end));
       i++;
     }
-    else if (sortedLabels(i) != sortedLabels(i + 1))
+    else if (dimension(sortedIndexDim(i)) != dimension(sortedIndexDim(i + 1)))
```
rcurtin: I think that you are correct that we need to ensure the values `dimension(sortedIndexDim(i))` and `dimension(sortedIndexDim(i + 1))` are different; otherwise it's possible we might select an invalid split.  However the previous condition `sortedLabels(i) != sortedLabels(i + 1)` is the condition on which we should split, not `dimension(sortedIndexDim(i)) != dimension(sortedIndexDim(i + 1))`.

I would suggest instead that you modify the value of `end` inside this `else if` block to correctly choose a value where `dimension(sortedIndexDim(end)) != dimension(sortedIndexDim(end + 1))`.

Let me know if I can explain that better.

rickllyxu: The previous condition `sortedLabels(i) != sortedLabels(i+1)` is not correct.  We shall split dataset according to a certain feature rather than its label. If we split on its label, then how to do prediction? In other words, if we decide which bins a sample should locate in relying on its label when training, how to make decision when a new sample coming in during testing stage? Obviously the new sample during testing stage does not contain its label.

rcurtin: I looked through the book and it seems that the decision stump implemented here is supposed to be like a one-level ID3 decision tree, according to the book.  So I think maybe you are correct that there are some errors.

However a long-standing goal has been to remove the `DecisionStump` class entirely and replace it with `DecisionTree<..., true>` (where the last `true` template parameter indicates that only one level of the tree will be built).  So I think maybe it would be better to make `DecisionTree<>` work right than to keep patching this `DecisionStump` code.

In order to do that, we'd need to add a constructor to `DecisionTree` of the same form as this one from `DecisionStump:

```
  /**
   * Alternate constructor which copies the parameters bucketSize and classes
   * from an already initiated decision stump, other. It appropriately sets the
   * weight vector.
   *
   * @param other The other initiated Decision Stump object from
   *      which we copy the values.
   * @param data The data on which to train this object on.
   * @param labels The labels of data.
   * @param weights Weight vector to use while training. For boosting purposes.
   */
  DecisionStump(const DecisionStump<>& other,
                const MatType& data,
                const size_t numClasses,
                const arma::rowvec& weights);
```

This constructor is used by the `AdaBoost` class and allows `DecisionStump<>` to be used as a weak learner.  If you add a constructor like that to `DecisionTree`, then I think you can add this code:

```
typedef DecisionTree<GiniGain, BestBinaryNumericSplit, AllCategoricalSplit, AllDimensionSplit, true> ID3DecisionStump;
```

and then in `adaboost.hpp`, you can replace the default `WeakLearnerType` template argument for `AdaBoost` to `ID3DecisionStump`, and then the tests could be updated.

Do you think you would be willing to do that?

rickllyxu: Of course I am willing to do it.  But I have a question.

Now the `DecisionTree` is implemented by `CART` algorithm, which builds a binary decision tree and decides optimal feature and optimal split point by Gini. The formula `Gini(D, A)` is to measure the uncertainty of dataset `D` after splitting by `A=a`. The smaller, the better. We will select best pair as the optimal feature and optimal split point.

However, the ID3 firstly decide optimal feature by information gain, then for each possible value of the optimal feature (suppose we select feature A, its value marks as a_i), data set D is divided into several non-empty subsets Di according to A=a_i. 

So my question is, which splitting metric or fitness function should we use when implement `ID3DecisionStump` ? What's your suggestion? I tend to use entropy metric, then the code will like this:
```
typedef DecisionTree<InformationGain, BestBinaryNumericSplit, AllCategoricalSplit, AllDimensionSplit, true> ID3DecisionStump;
```

Then the root of one-level decision tree will have two or more children. But if we use GiniGain as metric, it will have only two children.

What do you think of my thoughts?

rcurtin: Actually we are abusing terminology a little bit when we talk about ID3 decision trees built on continuous/numeric features.  The original ID3 paper by Quinlan only discusses categorical features, not numeric features.  So I think we can decide what we think is best for numeric features. :)

You're right that ID3 is built on information gain, so the typedef you gave looks fine to me:

```
typedef DecisionTree<InformationGain, BestBinaryNumericSplit, AllCategoricalSplit, AllDimensionSplit, true> ID3DecisionStump;
```

But since we are using `BestBinaryNumericSplit`, we will end up splitting a numeric feature into only two children.  I think this would be perfectly okay for AdaBoost though, so it is not a problem.  (Also `BestBinaryNumericSplit` is reasonably fast; should be O(n log n) for splitting O(n) points, and it just does one sort of the data and then a single pass over it.)  If you wanted, we could implement some other algorithm than `BestBinaryNumericSplit`, but I don't see it as fully necessary.  I am actually not sure where the splitting algorithm Udit originally implemented for the `DecisionStump` code comes from, but I'd strongly suspect `BestBinaryNumericSplit` will give similar or better results when used in AdaBoost.

(Categorical features are split into each category, according to `AllCategoricalSplit`.  Note that the old `DecisionStump` class doesn't have support at all for categorical features.)

rickllyxu: Actually I looked through the chapter about Adaboost. It gives an example about AdaBoost. And the weak learner it uses is a decision stump with `BestBinaryNumericSplit` on numeric feature. So you are right and `BestBinaryNumericSplit` is OK.

I will follow the idea to implement it.

rickllyxu: Hello, I am back! 

I have read the code of `decision_tree` and `adaboost` carefully. I restate my task: adapting decision tree to replace decision stump. 

I want to confirm that if the implement of `decision tree` in MLPack is `CART` or `ID3`. It seems that `ID3`. So it is not binary decision tree?

rcurtin: Right---the mlpack decision tree is ID3.  By default it uses binary splits for numeric variables but non-binary splits for categorical variables.  The `NoRecursion` template variable, if set to true, will create a decision stump (i.e. it will build the first level of nodes, but not the second level).  That should make it a lot easier to use the decision tree as a decision stump. :)

PR #1581, opened 2018-11-26
Fixed excess brackets and added some Documentation

atulim:
Sir,

(1) I have removed excess brackets in the program, which resulted in decrease in the size of the program.
(2) I have also added documentation.

### DIFF
--- a/src/mlpack/methods/decision_tree/decision_tree_main.cpp
+++ b/src/mlpack/methods/decision_tree/decision_tree_main.cpp
@@ -149,15 +149,14 @@ static void mlpackMain()
   arma::mat trainingSet;
   arma::Row<size_t> labels;
 
+  // Returns true, if the training parameter is present 
   if (CLI::HasParam("training"))
   {
     model = new DecisionTreeModel();
     model->info = std::move(std::get<0>(CLI::GetParam<TupleType>("training")));
     trainingSet = std::move(std::get<1>(CLI::GetParam<TupleType>("training")));
     if (CLI::HasParam("labels"))
-    {
       labels = std::move(CLI::GetParam<arma::Row<size_t>>("labels"));
-    }
     else
     {
       // Extract the labels as the last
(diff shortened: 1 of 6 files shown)


src/mlpack/methods/decision_tree/decision_tree_main.cpp
```diff
@@ -180,30 +179,26 @@ static void mlpackMain()
     {
       arma::Row<double> weights =
           std::move(CLI::GetParam<arma::Mat<double>>("weights"));
+      // Returns true ,if the print_training_error parameter is present
       if (CLI::HasParam("print_training_error"))
-      {
-        model->tree = DecisionTree<>(trainingSet, model->info, labels,
+      model->tree = DecisionTree<>(trainingSet, model->info, labels,
```
zoq: Actually, using

```c++
if (CLI::HasParam("print_training_error"))
{
	model->tree = DecisionTree<>(trainingSet, model->info, labels,
	    numClasses, minLeafSize, minimumGainSplit);
}
```
instead of

```c++
if (CLI::HasParam("print_training_error"))
      model->tree = DecisionTree<>(trainingSet, model->info, labels,
            numClasses, minLeafSize, minimumGainSplit);
```

is intentional; I think it's easier to read and might avoid some confusions. What do you think?

rcurtin: Right, agreed, I think this is easier to read.  Also I am concerned that comments like `Returns true if the print_training_error parameter is present` are somewhat unnecessary for lines like `if (CLI::HasParam("print_training_error"))`.

atulim: Sir,

We can add space above and below line of model->tree but to think of it, '{' increase the size of file and also the cluster.
In order to be used, the size of the MLPACK should be less as possible.
Adding too much of brackets along with white spaces increases the size of the file, and if you want I can mention the difference in size of files.
Although I agree with adding a white space. But adding brackets seems to be increasing the size of file a bit which isn't required.

Issue #1578, opened 2018-11-23
non-configurable install directory locations


marcespie:
I still have a patch in OpenBSD because we install our man pages straight under /usr/local/man and not /usr/local/share/man

Actually, there's a module in cmake that ought to take care of that:
`
include(GNUInstallDirs)`

but this is likely to affect a bit more than just manpages...

Please consider it.

rcurtin:
:+1: I can work up a patch for that.  I think I have meant to use `GNUInstallDirs` for some time now.

rcurtin:
#1583 includes the patch. :)

rcurtin:
#1583 is merged so I will close this now.  Thanks again for the report and if you find more please don't hesitate to open more issues. :+1:

Issue #1576, opened 2018-11-19
nitpicks post 3.0.4


marcespie:
- comment 
# We can probably use FindPythonInterp when we require CMake 3.0.
in src/mlpack/bindings/python/CMakeLists.txt    is obsolete

- allexec2man still contains one linuxism.
find ..  -executable
is gnu find only, the portable idiom would read:
find . -perm -u+x

rcurtin:
See #1583 for the fixes---thanks again for pointing these out.  Always happy to make further changes if more make your life easier. :+1:

rcurtin:
#1583 is merged so I will close this.  Thanks again for reporting the issues. :)

PR #1583, opened 2018-11-28
CMake fixes and GNUInstallDirs

rcurtin:
These changes are for #1578 and #1576:

 * Use `GNUInstallDirs` from CMake, which lets the user configure things like LIBDIR, INCLUDEDIR, etc.  On Windows, we just set these to their existing defaults, but they can be configured there also.
 * A few minor formatting fixes.
 * Fix another Linux-ism for the `find` call in `allexec2man.sh`.

### DIFF
--- a/CMake/allexec2man.sh
+++ b/CMake/allexec2man.sh
@@ -18,7 +18,7 @@ exec2man="$1"
 outdir="$2"
 
 mkdir -p "$outdir"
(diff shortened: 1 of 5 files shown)


rcurtin:
This is ready from my end---the failure was a timeout on AppVeyor, which actually shouldn't happen anymore now that ensmallen is fully pulled out of the mlpack repository.

Issue #1455, opened 2018-07-01
Is there a way to output loss value for every iteration


scrcro:
If I need to understand how the loss value behaves with every iteration of the training, is there a way to output the loss value to a file? I know there is some output in the form of LOG::Info but I just need the raw values without the texts. Is there a good way to accomplish this?

rcurtin:
Hmm, there is not an easy way to do it (and maybe there should be).  Right now the three ways I can think of are:

 * Capture the `Log::Info` output that you already get during training and process it a little bit to recover the raw values only with awk and sed (or whatever tool).

 * Modify the SGD optimizer output in `src/mlpack/core/optimizers/sgd/` to get what you need and print it to `std::cout`.

 * Run the training in a single-epoch loop and print what you need, i.e.,

```
// I'm assuming opt is whatever optimizer you're using.
opt.MaxIterations() = data.n_cols; // so it will take one pass over the data
for (size_t i = 0; i < epochs; ++i)
{
  model.Train(trainingData, trainingResponses, opt);
  std::cout << CalculateLoss(model); // for some CalculateLoss() function
}
```

If you have ideas on how we could improve the overall abstraction to do things like this, I certainly have no opposition to that.  Maybe a template parameter to optimizers to perform some behavior at the end of each epoch or something?

zoq:
Another idea would be to forward the function calls, something like:

```c++
template<class FunctionType>
class WrapperFunction
{
 public:
  WrapperFunction(FunctionType& function) : function(function) { }

  arma::mat& Coordinates() { return coordinates; }

  void Shuffle() { function.Shuffle(); }

  size_t NumFunctions() const { return function.NumFunctions(); }

  arma::mat GetInitialPoint() const { return function.GetInitialPoint(); }

  double Evaluate(const arma::mat& coordinates,
                  const size_t begin,
                  const size_t batchSize)
  {
    const double result = function.Evaluate(coordinates, begin, batchSize);

    // Save the output here.

    return result;
  }

  double Evaluate(const arma::mat& coordinates)
  {
    const double result = function.Evaluate(coordinates);

    // Save the output here.

    return result;
  }

  void Gradient(const arma::mat& coordinates,
                const size_t begin,
                arma::mat& gradient,
                const size_t batchSize)
  {
    function.Gradient(coordinates, begin, gradient, batchSize);
  }

  void Gradient(const arma::mat& coordinates, arma::mat& gradient)
  {
    function.Gradient(coordinates, gradient);
[...]

PR #1587, opened 2018-12-03
Add control of Reset Policy to Adam optimizers

dtimson:
Seems to be needed to do "single-epoch loop" training (see issue #1455) without resetting each time.


rcurtin:
Thanks for the contribution!  Actually I think we should merge this into the `mlpack/ensmallen` repository (https://github.com/mlpack/ensmallen) which is the same optimizer framework, just separated from mlpack.  In future releases mlpack will use that.  Would you like to open a PR there?  If not I can port your commit, just let me know.

dtimson:
OK, Ryan, I've done that.

zoq:
See https://github.com/mlpack/ensmallen/pull/60 for further updates.

PR #1588, opened 2018-12-04
Version fixes

rcurtin:
This updates the documentation to use the latest released version number, and updates `version.hpp` to have a version that's after the latest stable release.

### DIFF
--- a/src/mlpack/core/util/version.hpp
+++ b/src/mlpack/core/util/version.hpp
@@ -18,7 +18,7 @@
 // with higher number than the most recent release.
 #define MLPACK_VERSION_MAJOR 3
 #define MLPACK_VERSION_MINOR 0
-#define MLPACK_VERSION_PATCH 3
+#define MLPACK_VERSION_PATCH 5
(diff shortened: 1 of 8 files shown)


src/mlpack/core/util/version.hpp
```diff
@@ -18,7 +18,7 @@
 // with higher number than the most recent release.
 #define MLPACK_VERSION_MAJOR 3
 #define MLPACK_VERSION_MINOR 0
-#define MLPACK_VERSION_PATCH 3
+#define MLPACK_VERSION_PATCH 5
```
ShikharJ: Again a minor query, shouldn't this be `4`?

rcurtin: The documentation says we will keep it one version newer than the latest stable version in the git repository.  Honestly I don't have too much of an opinion here so we could change it if we wanted.

zoq: I think this is just fine.

PR #1591, opened 2018-12-06
Try multiple different keyservers for reliability.

rcurtin:
This is based on the comments here: https://github.com/tianon/gosu/issues/17

I hope it will fix the unreliable travis builds; we'll see.


rcurtin:
Ok... let's see if it works long-term. :)

Issue #1592, opened 2018-12-09
Occasional failures for OneStepSarsaTest and OneStepQLearningTest


rcurtin:
I found that very occasionally, the tests `AsyncLearningTest/OneStepSarsaTest` and `AsyncLearningTest/OneStepQLearningTest` can fail.  Here are some examples (you will need to log into ci.mlpack.org before these links work):

http://ci.mlpack.org/job/docker%20mlpack%20weekly%20build/armadillo_version=armadillo-8.300.3,boost_version=boost_1_49_0,buildmode=debug,compiler_version=gcc-6.3.0,label=docker/lastCompletedBuild/testReport/mlpackTest.AsyncLearningTest._home_jenkins_workspace_docker.mlpack.weekly/build_armadillo_version_armadillo-8/OneStepSarsaTest/

http://ci.mlpack.org/job/docker%20mlpack%20weekly%20build/armadillo_version=armadillo-8.300.3,boost_version=boost_1_49_0,buildmode=debug,compiler_version=gcc-6.3.0,label=docker/lastCompletedBuild/testReport/mlpackTest.AsyncLearningTest._home_jenkins_workspace_docker.mlpack.weekly/build_armadillo_version_armadillo-8/OneStepQLearningTest/

At first I thought this was a memory error, but valgrind doesn't report anything wrong.  So I kept digging further and I realized that the problem is in this bit:

```
  auto measure = [&rewards, &pos, &testEpisodes](double reward)
  { 
    size_t maxEpisode = 10000;
    if (testEpisodes > maxEpisode)
      BOOST_REQUIRE(false);
    testEpisodes++;
    rewards[pos++] = reward;
    pos %= rewards.n_elem;
    // Maybe underestimated.
    double avgReward = arma::mean(rewards);
    Log::Debug << "Average return: " << avgReward
        << " Episode return: " << reward << std::endl;
    if (avgReward > 60)
      return true;
    return false;
  };
```

The failure comes because `testEpisodes > maxEpisode`.  When I found a random seed that would fail, I found that even if I ran for a million or more episodes, it would never converge.  In fact, `avgReward` stayed around 8.5 and the reward of each episode was generally less than 10, for `OneStepSarsaTest`.  Typically, `OneStepSarsaTest` will converge to `avgReward > 60` in roughly 2000 episodes (based on the few trials I looked at).  It's about the same for `OneStepQLearningTest`, but convergence seems to happen in more like 1000 episodes.

[...]

PR #1595, opened 2018-12-13
Update tests for more robustness

rcurtin:
Many of these tests would occasionally fail, so I have gone through and modified tolerances or made multiple runs as necessary to prevent (or reduce) random failures.

This solves #1592 and many other intermittent failures, and I think it will make the nightly and weekly builds all succeed again. :)

For the CF tests, I noticed that the RMSE of most decomposition methods is *significantly* lower if we use `OverallMeanNormalization` instead of `NoNormalization` for the `CFPredict<>()` test function.  Specifically, the `SVDComplete` and `SVDIncomplete` decomposition methods would often find RMSEs of greater than 3.5 (and note that the scale of ratings is 1 to 5...) with `NoNormalization`, but compete with `RegSVD` with RMSEs as low as 1.0 or lower when using `OverallMeanNormalization`.  @Wenhao-H: I'd like to check if this change is okay with you, or if this breaks the testing strategy.  Another normalization strategy may work better, but mostly I wanted to ensure that this is okay with you.

@ShangtongZhang: I also had to use multiple trials for `QLearningTest/AcrobatWithDQN`, because I found it failed very rarely.  I think this should be fine, correct me if you think the issue is something else.

@ShikharJ: for the `BinaryRBMClassificationTest` and `ssRBMClassificationTest`, I found that sometimes it is not able to match the performance of a regular model, so I added some tolerance there, so the tests are now of the form "make sure the accuracy is no worse than X% worse than the standard model".  But if there actually should be a guarantee that BinaryRBM or ssRBM should do better, let me know.


ShikharJ:
@rcurtin I'm not aware of any such guarantee as well, maybe we can try searching for better hyper-parameters later on to make these tests more robust, but for the time being, think the changes are good enough.

rcurtin:
Ok, this all seems to work right.  So I will merge it in 3 days to leave time for any other comments.

@Wenhao-H actually I forgot to tag you earlier, but if you have a moment and are interested it might be worth glancing at how I changed the CF tests for better convergence.  If you disagree with anything I can change it. :)

PR #1600, opened 2018-12-15
kernel_pca_test.cpp: Add tests

saksham189:
Add tests for kernel_pca command line.

### DIFF
--- a/src/mlpack/tests/main_tests/kernel_pca_test.cpp
+++ b/src/mlpack/tests/main_tests/kernel_pca_test.cpp
@@ -1,0 +1,356 @@
+/**
+ * @file kernel_pca_test.cpp
+ * @author Saksham Bansal
+ *
+ * Test mlpackMain() of kernel_pca_main.cpp.
+ *
+ * mlpack is free software; you may redistribute it and/or modify it under the
+ * terms of the 3-clause BSD license.  You should have received a copy of the
+ * 3-clause BSD license along with mlpack.  If not, see
+ * http://www.opensource.org/licenses/BSD-3-Clause for more information.
+ */
+#include <mlpack/core.hpp>
+
+#define BINDING_TYPE BINDING_TYPE_TEST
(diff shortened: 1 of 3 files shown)


saksham189:
Thanks for the review. I have made the required changes. Please do see if any other tests are required.

Also, I wanted to point out that when finding the kernel matrix this [line](https://github.com/mlpack/mlpack/blob/master/src/mlpack/methods/nystroem_method/nystroem_method_impl.hpp#L89) can cause the entire output to be NaN if a singular value is 0 and then the output dimension is incorrect. Is this supposed to be the expected output or should the NaN values be ignored?

src/mlpack/tests/main_tests/kernel_pca_test.cpp
```diff
@@ -0,0 +1,356 @@
...
+  SetInputParam("new_dimensionality", (int) 3);
+  SetInputParam("kernel", (std::string) "badName");
+
+  Log::Fatal.ignoreInput = true;
+  BOOST_REQUIRE_THROW(mlpackMain(), std::runtime_error);
+  Log::Fatal.ignoreInput = false;
+}
+
+/**
+ * Make sure if 0 dimensions is specified, we get a dataset with same
```
ShikharJ: I'd prefer if it were something like `Ensure for zero dimensionality, ...`

rcurtin: Done during merge; 1171da89

rcurtin:
Thanks again! :+1:  If you do want to add your name to the list of contributors, you can just open a separate PR, and as before, if you'd like any stickers just send an email.

I've updated the list of completed tests in #1152. :)

src/mlpack/tests/main_tests/kernel_pca_test.cpp
saksham189: Thanks. I was traveling for a few days (without laptop) and couldn't make the changes.

PR #1614, opened 2018-12-26
[WIP]Prioritized Experience Replay

robotcator:
As discussed [here](https://github.com/mlpack/mlpack/pull/1001), I am going to implement the prioritized experience replay. 

Currently, I add the skeleton for this module, I will commit to this PR.

[The authors](https://arxiv.org/abs/1511.05952) proposed to increase data efficiency by training on transitions from replay buffer, that were "hard" (e.g. resulted in high TD-error) for the agent last time it observes them. For those, who are familiar with supervised learning, the idea of Prioritized Experience Replay is somewhat similar to the "Online Hard Example Mining" (OHEM).


robotcator:
> This looks nice so far, looking forward to merge this once this is ready.

Thanks, I will continuously commit to this PR.

mirraaj:
@robotcator , @zoq Please let me know the status of this issue.

robotcator:
> @robotcator, @zoq Please let me know the status of this issue.

Hi, @mirraaj, this project is being actively developing currently, I am going to add a document for the PR.

src/mlpack/methods/reinforcement_learning/replay/prioritized_replay.hpp
```diff
@@ -0,0 +1,212 @@
...
+    }
+    weights /= weights.max();
+  }
+
+  void update_priorities(arma::ucolvec& indices, arma::colvec& priorities)
+  {
+//    update priorities of sampled transitions.
+      for (size_t i = 0; i < indices.n_rows; ++i)
+      {
+        idxSum.set(indices[i], alpha * priorities[i]);
```
manish7294: Hi @robotcator, great work on this :)
Looking this made me wonder if we can improve this step and do the complete update in a single call. So, here's my vague idea:
1. Create a sum tree for `alpha * priorities[i]`, let's name it as priTree.
2. and instead of `idxSum.set(indices[i], alpha * priorities[i]);` for each indices. make a binary call(means traverse each node) to each node of `idxSum` (that will be `O(n)`) and simultaneously update the values of `idxSum` with corresponding value of `priTree`. 
Or since we are making a full update everytime & are just replacing old values with new one, we can just replace the `idxSum` with `priTree`.

It may able to reduce time complexity from `O(nlog(n))` to` O(n)`.

This may not even be a reasonable idea, as it is just a random thought. But do let me know your views on this.

robotcator: Hi, @manish7294, great to see your detailed comments. Your idea is great and seems can reduce time complexity. But I am a little bit confused about the step2. 

The `sumtree` data structure will maintain the `prefix-sum` of samples' priorities,  it is like the [BIT structure](https://en.wikipedia.org/wiki/Fenwick_tree) actually.  Each time we call `idxSum.set(idx, value)`, the corresponding nodes will also be updated. 

From your description, I can not figure out how do you deal with this when you make a binary call( traverse each node) ? Can you give more details?

manish7294: > The sumtree data structure will maintain the prefix-sum of samples' priorities, it is like the BIT structure actually. Though using BIT also could have been a better option.
 
I don't think the implementation is that of a binary indexed tree, it's just like a segment tree.

> Each time we call idxSum.set(idx, value), the corresponding nodes will also be updated.

Right, as we will be traversing the `idxSum` tree we will simultaneously picking up the corresponding values from `priTree` to be used as updates. Moreover since we are doing a complete update of the tree each time, we can even just replace the `idxSum` with `priTree`, I hope this may clear things up a little bit. Let me know if you still have any confusion.

robotcator: > I don't think the implementation is that of a binary indexed tree, it's just like a segment tree.

Yes, you are right. This is my mistake. Both of them can maintain the prefix-sum, but it is segment-tree like data structure here. I add a document for this part. :)   
 
> Right, as we will be traversing the `idxSum` tree we will simultaneously picking up the corresponding values from `priTree` to be used as updates. Moreover since we are doing a complete update of the tree each time, we can even just replace the `idxSum` with `priTree`, I hope this may clear things up a little bit. Let me know if you still have any confusion.

Oh, sounds reasonable. Please let me clarify my understanding to make sure I did not miss what you said. 
1  Instead of calling `idxSum.set` repeatively. We update the leaf node of the sumtree for all the sampled indices.
2 Then update the inner node with bottom-up techniques. 

Oh, yeah, in this way, I think we do not need priTree at all, right?

manish7294: Yes, Exactly, it's basically building a new sum-tree.

robotcator: Hi, @manish7294, I added [`batchUpdate()`](https://github.com/mlpack/mlpack/blob/fc33c5b03ba9c51cc13e91a44baf2ab8db24dedf/src/mlpack/methods/reinforcement_learning/replay/sumtree.hpp#L80) method in `SumTree` class to support batch updating the priorities in `PrioritizedReplay` class.  I also added a simple test case for this method. Please feel free for pointing out if you have any suggestions.

zoq:
@robotcator Some of the mlpack tests are probabilistic, so they may randomly fail. We try to keep the failure probabilities low, but sometimes our tolerances are still too small. Looks like in the last check everything passed.

src/mlpack/methods/reinforcement_learning/replay/sumtree.hpp
```diff
@@ -0,0 +1,193 @@
...
+   *
+   * @param _start The starting position of subsequence
+   * @param _end The end position of subsequence
+   * @param node Reference position
+   * @param node_start Starting position of reference segment
+   * @param node_end End position of reference segment
+   */
+  T sumHelper(size_t _start, size_t _end, size_t node, size_t node_start, size_t node_end)
+  {
+    if (_start == node_start && _end == node_end)
```
manish7294: Variable name should start with a lowercase letter. So please avoid underscore. Same for all other places.

src/mlpack/methods/reinforcement_learning/replay/prioritized_replay.hpp
```diff
@@ -0,0 +1,269 @@
...
+    sampledIndices = sampleProportional();
+    betaAnneal();
+
+    sampledStates = states.cols(sampledIndices);
+    sampledActions = actions.elem(sampledIndices);
+    sampledRewards = rewards.elem(sampledIndices);
+    sampledNextStates = nextStates.cols(sampledIndices);
+    isTerminal = this->isTerminal.elem(sampledIndices);
+
+//  calculate the weights of sampled transitions
```
manish7294: Please check the indentation here.

src/mlpack/methods/reinforcement_learning/replay/prioritized_replay.hpp
```diff
@@ -0,0 +1,269 @@
...
+   * @param indices The indices of sample to be updated.
+   * @param priorities Their corresponding priorities.
+   */
+  void update_priorities(arma::ucolvec& indices, arma::colvec& priorities)
+  {
+      arma::colvec alphaPri(indices.n_rows);
+      for (size_t i = 0; i < indices.n_rows; ++i)
+      {
+        alphaPri = alpha * priorities[i];
+        max_priority = std::max(max_priority, priorities[i]);
```
manish7294: This is rather picky comment. But I guess we could replace this block with : 
```
alphaPri = alpha * priorities;
max_priority = std::max(max_priority, arma::max(alphaPri));
```
Can't say surely, whether it will make any difference in terms of speed. For that need to verify with armadillo `max` implementation.

src/mlpack/methods/reinforcement_learning/replay/prioritized_replay.hpp
```diff
@@ -0,0 +1,269 @@
...
+  {
+    beta = beta + (1 - initial_beta) * 1.0 / replay_beta_iters;
+  }
+
+private:
+  //! How much prioritization is used.
+  //  (0 - no prioritization, 1 - full prioritization)
+  double alpha;
+
+  double max_priority;
```
manish7294: Please avoid underscore in the names of variables. I would recommend going through [style guidelines](https://github.com/mlpack/mlpack/wiki/DesignGuidelines) once.

src/mlpack/methods/reinforcement_learning/replay/sumtree.hpp
```diff
@@ -0,0 +1,193 @@
...
+  SumTree(size_t capacity):
+      capacity(capacity)
+  {
+    element = std::vector<T> (2 * capacity);
+  }
+
+  /**
+   * Set the data array with idx
+   *
+   * @param idx  The array idx to be changed
```
manish7294: I suppose each line should end with a `.`

src/mlpack/methods/reinforcement_learning/replay/prioritized_replay.hpp
```diff
@@ -0,0 +1,269 @@
...
+      nextStates(dimension, capacity),
+      isTerminal(capacity),
+      full(false),
+      max_priority(1.0),
+      initial_beta(0.6),
+      replay_beta_iters(10000)
+  {
+    arma::arma_rng::set_seed(seed);
+    int size = 1;
+    while (size < capacity) {
```
manish7294: Here we are making a comparison between a signed & unsigned type. Though it may not matter much as `size` will be positive, but I guess we could always avoid unnecessary warnings.

robotcator:
> @robotcator Some of the mlpack tests are probabilistic, so they may randomly fail. We try to keep the failure probabilities low, but sometimes our tolerances are still too small. Looks like in the last check everything passed.

@zoq  Thank you for pointing out. The CI test is passed now.

src/mlpack/methods/reinforcement_learning/replay/prioritized_replay.hpp
robotcator: You are right, make the type consistent when comparing is much better.

src/mlpack/methods/reinforcement_learning/replay/prioritized_replay.hpp
robotcator: Nice catch.

src/mlpack/methods/reinforcement_learning/replay/prioritized_replay.hpp
robotcator: You are right, the vectorized code is much better. So I prefer using:
```
max_priority = std::max(max_priority, arma::max(alphaPri));
```
I think we can add a `TODO` tag for this block.

src/mlpack/methods/reinforcement_learning/replay/prioritized_replay.hpp
manish7294: Hi @robotcator, I suppose you were mentioning me here, but somehow accidentally tagged @mirraaj. Anyway in this case, I would say either is fine. You may use whatever fits you best. It hardly makes any difference.
Though I would ask @rcurtin @zoq If they have any preference in this regard and would like to pick one for future uses?

src/mlpack/methods/reinforcement_learning/q_learning.hpp
```diff
@@ -52,7 +53,8 @@ template <
   typename NetworkType,
   typename UpdaterType,
   typename PolicyType,
-  typename ReplayType = RandomReplay<EnvironmentType>
+  typename ReplayType = RandomReplay<EnvironmentType>,
```
zoq: Maybe I missed something, but I think it's possible to use the existing Interface and implement PrioritizedReplay as another policy, so a user can just pass ``PrioritizedReplay<EnvironmentType>`` instead of ``RandomReplay<EnvironmentType>`` does this make sense?

robotcator: > Maybe I missed something, but I think it's possible to use the existing Interface and implement PrioritizedReplay as another policy, so a user can just pass `PrioritizedReplay<EnvironmentType>` instead of `RandomReplay<EnvironmentType>` does this make sense?

Hi, @zoq, thank for your comments and sorry for the delay because of the Spring Festival. 

Yes, there may many different implementations of this. Changing that Interface is one of my implementations.  Let me clarify my thoughts on this. In my understanding, the policy outputs the agent's action to tell the agent what need to when facing the current state. The replay buffer is the place where the agent to store the experience. The offline RL algorithm can update the parameter by sampling experience from the replay buffer. 

`PrioritizedReplay` is the same as `RandomReplay ` except that the sampling behavior.  Currently, I implement two different `constructor` function, so the user can choose what type of replay buffer to use.  

You mentioned `implement PrioritizedReplay as another policy`, do you mean we put the `replay buffer`'s part into the policy?  I saw that the existing policy in current `policy` folder, all the interface are sampling action. Can you let me know more about your idea?

zoq: Right, ideally we could put everything into the policy, that way I could use prioritized replay for other classes as well. What about adding another method to the replay class, something like ``Update(...)`` that is called right after ``  updater.Update(learningNetwork.Parameters(), config.StepSize(), gradients);``. For the existing replay classes we do nothing but for the prioritized replay  class we update the priorities. Do you think that could work?

robotcator: > Right, ideally we could put everything into the policy, that way I could use prioritized replay for other classes as well. What about adding another method to the replay class, something like `Update(...)` that is called right after ` updater.Update(learningNetwork.Parameters(), config.StepSize(), gradients);`. For the existing replay classes we do nothing but for the prioritized replay class we update the priorities. Do you think that could work?

Yes, you are right. `... adding another method to the replay class, something like Update(...) that is called right after ...`  it will look like more elegent. 

Idealy, I hope that there is only one `replayMethod` variable instead of one `replayMethod` and one `prioritizedReplayMethod`. If the user pass `random_replay` object, we can sample experience randomly. and sampling with priorities  when pass `prioritized_replay` object. But class `random_replay` and `prioritized_replay` have no inheritence relation, I not sure that they can cast with each other. Do you have any idea about it?

robotcator: > Right, ideally we could put everything into the policy, that way I could use prioritized replay for other classes as well. What about adding another method to the replay class, something like `Update(...)` that is called right after ` updater.Update(learningNetwork.Parameters(), config.StepSize(), gradients);`. For the existing replay classes we do nothing but for the prioritized replay class we update the priorities. Do you think that could work?

Hi, @zoq, I update the code that adding an `update` method for `prioritized_replay` Class to update the priorities. I think the code can satisfy all the functionality of `prioritized_replay`.

BTW, I found that there is a way to make the code more elegant, but it needs C++14 support. 
We can use `boost:: variant` to achieve different behavior when using different replay buffer Class.
 
```
using Obj = boost::variant<RandomReplay, PrioritizedReplay>;
```
If we use RandomReplay, the code can be writen in this way:
```
Obj v = RandomReplay(...);
boost::apply_visitor([](auto const& obj) { obj.Sample(); }, v);
```
or PrioritizedReplay
```
Obj v = PrioritizedReplay(...);
boost::apply_visitor([](auto const& obj) { obj.Sample(); }, v);
```

the reference link: [boost variant simple call to common methods](https://stackoverflow.com/questions/37504740/boost-variant-simple-call-to-common-methods)

zoq: ``boost::variant``  is a really versatile construct, but I don't think we should switch to C++14 yet, but we should keep that in mind. Will take a look at the updates, thanks for looking into the issue.

zoq:
@robotcator can you take a look at the static code analysis checks, once you get a chance?

robotcator:
> @robotcator can you take a look at the static code analysis checks, once you get a chance?

Yes, of course. Sorry for the delay. I am on the way back to school last few days.

robotcator:
> @robotcator can you take a look at the static code analysis checks, once you get a chance?

@zoq Hi, is there anything need to update or fix before merging? Please let me know.

Also, I found that two distributions ( Categorical and  MultivariateNormal ) which often used in RL algorithm seems not been implemented in the ANN module. If I didn't missing something, such it is already in the code base, I will begin to contribute to that part.

zoq:
Thanks for take a look into the static analysis issues, I'll take a look at the changes tomorrow.


> Also, I found that two distributions ( Categorical and MultivariateNormal ) which often used in RL algorithm seems not been implemented in the ANN module. If I didn't missing something, such it is already in the code base, I will begin to contribute to that part.

You are right, if you like to work on the distributions please feel free, happy to review the code.

src/mlpack/methods/reinforcement_learning/replay/sumtree.hpp
```diff
@@ -0,0 +1,194 @@
...
+ *
+ * Build a Segment Tree like data structure.
+ * https://en.wikipedia.org/wiki/Segment_tree
+ *
+ * Used to maintain prefix-sum of an array.
+ *
+ */
+
+template<typename T>
+class SumTree
```
zoq: I wonder if it makes sense to implement the Segment Tree following: https://github.com/mlpack/mlpack/wiki/OldSummerOfCodeIdeas#implement-tree-types let me know what you think. @rcurtin might be able to provide some more insight.

src/mlpack/methods/reinforcement_learning/q_learning_impl.hpp
```diff
@@ -82,14 +125,16 @@ template <
...
+  typename ReplayType,
+  typename PrioritizedReplayType
 >
 double QLearning<
   EnvironmentType,
   NetworkType,
   UpdaterType,
   BehaviorPolicyType,
-  ReplayType
+  ReplayType,
```
zoq: Maybe I missed something but can you use the ReplayType and PrioritizedReplayType at the same time?

src/mlpack/methods/reinforcement_learning/replay/sumtree.hpp
rcurtin: I remember discussing the segment tree sometime years back, and my conclusion was that it wouldn't really fit in to the `TreeType` abstraction since it's not a multidimensional tree.  Another problem might be (I think) that the segment tree holds segments, not individual points, so it could be a little bit hard to fit that in to the current abstraction.  Maybe there are some ways it could be made to fit, but since segment trees (or sum trees) aren't used directly for distance-like computations in the same ways the existing trees are, I don't know how much overlap there is.

I'm not saying it's not possible of course, just that it might be too tricky and not worth it here.  I don't see a problem with having a separate `SumTree` structure if it's needed (assuming it doesn't already exist in some form in the STL or something). :)

src/mlpack/methods/reinforcement_learning/q_learning_impl.hpp
robotcator: > Maybe I missed something but can you use the ReplayType and PrioritizedReplayType at the same time?

Of course, we can not two type of replay method simultaneously. But I didn't get what you say about this part. Can you be more specific or give an example?

zoq: I thought we made all the changes to the interface to use experienced replay as the ReplayType.

robotcator: > I thought we made all the changes to the interface to use experienced replay as the ReplayType.

I am so sorry for my poor English :(. I still can not get the main idea. Do you mean we should keep the original template interface rather than add two kind of replay type  in here ?
```
template <
  typename EnvironmentType,
  typename NetworkType,
  typename UpdaterType,
  typename BehaviorPolicyType,
  typename ReplayType,
  typename PrioritizedReplayType
>
```

zoq: No worries, keeping the current interface sounds like a possible solution to me, but maybe I missed an important part.

src/mlpack/tests/q_learning_test.cpp
```diff
@@ -102,6 +102,72 @@ BOOST_AUTO_TEST_CASE(CartPoleWithDQN)
   BOOST_REQUIRE(converged);
 }
 
+//! Test DQN in Cart Pole task.
```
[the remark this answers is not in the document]
robotcator: > Should we mention this uses prioritized experience replay?

Yes, now,  it is fixed.

src/mlpack/methods/reinforcement_learning/q_learning_impl.hpp
robotcator: > No worries, keeping the current interface sounds like a possible solution to me, but maybe I missed an important part.

Yes, I hope the discussion may help me to write the code more consistent with the codebase. I also learn many things from your comments. Thanks.

robotcator:
@zoq  Hi, what is the status of PR?

zoq:
Sorry for the slow response, will take a look.

src/mlpack/methods/reinforcement_learning/q_learning_impl.hpp
```diff
@@ -121,8 +174,20 @@ double QLearning<
...
+  arma::rowvec weights;
+
+  if (!prioritized_replay)
+  {
+    replayMethod.Sample(sampledStates, sampledActions, sampledRewards,
+        sampledNextStates, isTerminal);
+  }
+  else
+  {
+    prioritizedReplayMethod.Sample(sampledStates, sampledActions,
```
zoq: What I had in mind was to remove the ``prioritizedReplayMethod`` template parameter and use the ``policy`` parameter instead; since the prioritized replay sample method takes the sampled indicies and weights, we would have to pass that to the other replay method as well. So,

https://github.com/mlpack/mlpack/blob/85e00107fe3e2bb8d12f331a725cdfe11acb290b/src/mlpack/methods/reinforcement_learning/replay/random_replay.hpp#L111-L115

becomes:

```c++
void Sample(arma::mat& sampledStates,
              arma::icolvec& sampledActions,
              arma::colvec& sampledRewards,
              arma::mat& sampledNextStates,
              arma::icolvec& isTerminal,
              arma::ucolvec& /*sampledIndices */,
              arma::rowvec& /*weights*/)
```

we just pass the two extra parameter, but we don't use them.

src/mlpack/methods/reinforcement_learning/q_learning_impl.hpp
```diff
@@ -161,8 +226,24 @@ double QLearning<
...
+  if (prioritized_replay)
+  {
+    gradients = arma::mean(weights) * gradients;
+  }
+
   updater.Update(learningNetwork.Parameters(), config.StepSize(), gradients);
 
+  if (prioritized_replay)
+  {
+    prioritizedReplayMethod.Update(target, sampledActions,
```
zoq: Do you think we could move this before the ``updater.update`` step? That way we could pass the gradient to the ``replayMethod.Update`` method and do the gradient step inside the ``replayMethod.Update`` method itself. If not we could introduce a new method something like ``GradientUpdate`` that just does nothing for the random replay method. Let me know what you think.

src/mlpack/methods/reinforcement_learning/q_learning_impl.hpp
robotcator: sorry for late reply, I am not feeling well recently.  I will take a look, thanks for the review.

zoq: No worries, hope you are feeling better soon.

robotcator: > No worries, hope you are feeling better soon.

sorry for being away for so long, I am back and will take a look.

zoq: Great, thanks!

robotcator: > What I had in mind was to remove the `prioritizedReplayMethod` template parameter and use the `policy` parameter instead; since the prioritized replay sample method takes the sampled indicies and weights, we would have to pass that to the other replay method as well. So,
> 
> https://github.com/mlpack/mlpack/blob/85e00107fe3e2bb8d12f331a725cdfe11acb290b/src/mlpack/methods/reinforcement_learning/replay/random_replay.hpp#L111-L115
> 
> becomes:
> 
> ```c++
> void Sample(arma::mat& sampledStates,
>               arma::icolvec& sampledActions,
>               arma::colvec& sampledRewards,
>               arma::mat& sampledNextStates,
>               arma::icolvec& isTerminal,
>               arma::ucolvec& /*sampledIndices */,
>               arma::rowvec& /*weights*/)
> ```
> 
> we just pass the two extra parameter, but we don't use them.

Hi, @zoq, please let me clarify to make sure I didn't miss anything.  
You mean remove the 'prioritizedReplayMethod' parameter 
```
arma::ucolvec& sampledIndices,
arma::rowvec& weights
```

But do you mean move those two parameters into `policy` class? I found that we can move them into `prioritizedReplayMethod`, then we can access them as a class parameter. What do you think of it?

src/mlpack/methods/reinforcement_learning/q_learning_impl.hpp
robotcator: Yes, you are more considerate. I agree with you, thanks.

robotcator: > Do you think we could move this before the `updater.update` step? That way we could pass the gradient to the `replayMethod.Update` method and do the gradient step inside the `replayMethod.Update` method itself. If not we could introduce a new method something like `GradientUpdate` that just does nothing for the random replay method. Let me know what you think.

If we pass the gradient to the `replayMethod.Update` method and do the gradient step inside the replayMethod.Update method, should we add the same interface for random replay?

zoq: I think that is a good idea.

robotcator: Hi, @zoq, I didn't create another new policy class, just remove the two parameters `sampledIndices` and `weights` to `PrioritizedReplay` as class attribution. It seems that more convenient.  So we can pass the same parameter to `sample` function. 

I also move the line which used to update the gradient into the `PrioritizedReplay` class's function `Update(...)`.
```
  gradients = arma::mean(weights) * gradients;
```

zoq: That sounds reasonable to me.

src/mlpack/methods/reinforcement_learning/replay/prioritized_replay.hpp
```diff
@@ -0,0 +1,285 @@
...
+   */
+  void BetaAnneal()
+  {
+    beta = beta + (1 - initialBeta) * 1.0 / replayBetaIters;
+  }
+
+  void Update(arma::mat target, arma::icolvec sampledActions,
+              arma::mat nextActionValues, arma::mat& gradients)
+  {
+    arma::colvec td_error(target.n_cols);
```
zoq: Minor style issue; can you use camel casing for all names; see https://github.com/mlpack/mlpack/wiki/DesignGuidelines#naming-conventions for more informations.

src/mlpack/methods/reinforcement_learning/replay/prioritized_replay.hpp
```diff
@@ -0,0 +1,285 @@
...
+
+  /**
+   *  Annealing the beta.
+   */
+  void BetaAnneal()
+  {
+    beta = beta + (1 - initialBeta) * 1.0 / replayBetaIters;
+  }
+
+  void Update(arma::mat target, arma::icolvec sampledActions,
```
zoq: Minor style issue, can you put the parameters after ``target`` on a newline?

zoq: We should add a description for each parameter.

robotcator: > Minor style issue, can you put the parameters after `target` on a newline?

fixed.

robotcator: > We should add a description for each parameter.

Yes, updated.

robotcator: Question about indexing for arma data type, such as `arma::colvec`, `arma::mat`. 

Is there any recommendation for using `[]` or `()` to index?

manish7294: If you are using arma data types, it is recommended to use () for excessing elements.

robotcator: > If you are using arma data types, it is recommended to use () for excessing elements.

Okay, I will update to make it consistent.

zoq:
Looks like this is ready from your side?

robotcator:
> Looks like this is ready from your side?

Yes, I think it is ready to merge. What do you think of it?

src/mlpack/methods/reinforcement_learning/q_learning_impl.hpp
manish7294: It's look like a discussion from way back. I doubt it may not even be active now.
@zoq Sorry, but I can't seem to get, what do you mean by removing `prioritizedReplayMethod` template parameter here? If this discussion is still applicable, can you please explain it a bit?

src/mlpack/methods/reinforcement_learning/q_learning.hpp
```diff
@@ -157,6 +169,8 @@ class QLearning
 
   //! Locally-stored flag indicating training mode or test mode.
   bool deterministic;
+
+  bool prioritized_replay;
```
zoq: Missing description and please use camel casing for all names.

src/mlpack/methods/reinforcement_learning/q_learning_impl.hpp
```diff
@@ -58,14 +61,54 @@ template <
...
+  prioritizedReplayMethod(std::move(prioritizedReplayMethod)),
+  environment(std::move(environment)),
+  totalSteps(0),
+  deterministic(false),
+  prioritized_replay(true)
+{
+  if (learningNetwork.Parameters().is_empty())
+    learningNetwork.ResetParameters();
+  this->updater.Initialize(learningNetwork.Parameters().n_rows,
+                           learningNetwork.Parameters().n_cols);
```
zoq: Can we tab this twice (4 spaces) to be consistent with the rest of the codebase.

src/mlpack/methods/reinforcement_learning/q_learning_impl.hpp
zoq: What I meant was to switch from:

```c++
template <
  typename EnvironmentType,
  typename NetworkType,
  typename UpdaterType,
  typename PolicyType,
  typename ReplayType = RandomReplay<EnvironmentType>,
  typename PrioritizedReplayType = PrioritizedReplay<EnvironmentType>
>
class QLearning
```

to what we already have

```c++
template <
  typename EnvironmentType,
  typename NetworkType,
  typename UpdaterType,
  typename PolicyType,
  typename ReplayType = RandomReplay<EnvironmentType>>
>
class QLearning
```

so instead of adding another template parameter (``PrioritizedReplayType``) why not reuse ``ReplayType``, which we could use like:


```c++
// Set up the network.
FFN<MeanSquaredError<>, GaussianInitialization> model(MeanSquaredError<>(),
model.Add<Linear<>>(4, 128);
model.Add<ReLULayer<>>();
model.Add<Linear<>>(128, 128);
model.Add<ReLULayer<>>();
model.Add<Linear<>>(128, 2);

// Set up the policy and replay method.
GreedyPolicy<CartPole> policy(1.0, 1000, 0.1, 0.99);
PrioritizedReplay<CartPole> replayMethod(10, 10000);	// Was RandomReplay<CartPole> replayMethod(10, 10000);

TrainingConfig config;
config.StepSize() = 0.01;
config.Discount() = 0.9;
config.TargetNetworkSyncInterval() = 100;
config.ExplorationSteps() = 100;
config.DoubleQLearning() = false;
config.StepLimit() = 200;

// Set up DQN agent.
QLearning<CartPole, decltype(model), AdamUpdate, decltype(policy)>
  agent(std::move(config), std::move(model), std::move(policy),
  std::move(replayMethod));
```

in the example above I just switch the type of the replay method from ``RandomReplay`` to ``PrioritizedReplay``.

The ``PrioritizedReplay`` class expects some additional parameters, but in some cases the interface is already the same, e.g.:


```c++
// Store the transition for replay.
if (prioritized_replay)
{
prioritizedReplayMethod.Store(state, action, reward,
    nextState, environment.IsTerminal(nextState));
}
else
{
replayMethod.Store(state, action, reward,
    nextState, environment.IsTerminal(nextState));
}
```

can just be:

```c++
replayMethod.Store(state, action, reward,
    nextState, environment.IsTerminal(nextState));
```

the same applies to the ``Sample`` and ``Update``. So to sum this up, In don't see a reason to add another template parameter for Prioritized Experience Replay. Does that sound reasonable?

manish7294: Thanks for thorough explanation. Right, that totally make sense. It would be best if we can preserve the template structure and avoid adding extra if else conditions. And your idea completely make up for it.
@robotcator please do have a look at this, and if you like you can make the required changes here itself.

src/mlpack/methods/reinforcement_learning/q_learning_impl.hpp
robotcator: > Can we tab this twice (4 spaces) to be consistent with the rest of the codebase.

Of course, I will fix that.

src/mlpack/methods/reinforcement_learning/q_learning_impl.hpp
robotcator: > What I meant was to switch from:
> 
> ```c++
> template <
>   typename EnvironmentType,
>   typename NetworkType,
>   typename UpdaterType,
>   typename PolicyType,
>   typename ReplayType = RandomReplay<EnvironmentType>,
>   typename PrioritizedReplayType = PrioritizedReplay<EnvironmentType>
> >
> class QLearning
> ```
> 
> to what we already have
> 
> ```c++
> template <
>   typename EnvironmentType,
>   typename NetworkType,
>   typename UpdaterType,
>   typename PolicyType,
>   typename ReplayType = RandomReplay<EnvironmentType>>
> >
> class QLearning
> ```
> 
> so instead of adding another template parameter (`PrioritizedReplayType`) why not reuse `ReplayType`, which we could use like:
> 
> ```c++
> // Set up the network.
> FFN<MeanSquaredError<>, GaussianInitialization> model(MeanSquaredError<>(),
>   GaussianInitialization(0, 0.001));
> model.Add<Linear<>>(4, 128);
> model.Add<ReLULayer<>>();
> model.Add<Linear<>>(128, 128);
> model.Add<ReLULayer<>>();
> model.Add<Linear<>>(128, 2);
> 
> // Set up the policy and replay method.
> GreedyPolicy<CartPole> policy(1.0, 1000, 0.1, 0.99);
> PrioritizedReplay<CartPole> replayMethod(10, 10000);	// Was RandomReplay<CartPole> replayMethod(10, 10000);
> 
> TrainingConfig config;
> config.StepSize() = 0.01;
> config.Discount() = 0.9;
> config.TargetNetworkSyncInterval() = 100;
> config.ExplorationSteps() = 100;
> config.DoubleQLearning() = false;
> config.StepLimit() = 200;
> 
> // Set up DQN agent.
> QLearning<CartPole, decltype(model), AdamUpdate, decltype(policy)>
>   agent(std::move(config), std::move(model), std::move(policy),
>   std::move(replayMethod));
> ```
> 
> in the example above I just switch the type of the replay method from `RandomReplay` to `PrioritizedReplay`.
> 
> The `PrioritizedReplay` class expects some additional parameters, but in some cases the interface is already the same, e.g.:
> 
> ```c++
> // Store the transition for replay.
> if (prioritized_replay)
> {
> prioritizedReplayMethod.Store(state, action, reward,
>     nextState, environment.IsTerminal(nextState));
> }
> else
> {
> replayMethod.Store(state, action, reward,
>     nextState, environment.IsTerminal(nextState));
> }
> ```
> 
> can just be:
> 
> ```c++
> replayMethod.Store(state, action, reward,
>     nextState, environment.IsTerminal(nextState));
> ```
> 
> the same applies to the `Sample` and `Update`. So to sum this up, In don't see a reason to add another template parameter for Prioritized Experience Replay. Does that sound reasonable?

But I have a question about that,  how do we know which `replay buffer` passed into the q_learning class?  If we don't use inheritance, how to deal with the situation when we have two constructors?

```
QLearning(TrainingConfig config,
            NetworkType network,
            PolicyType policy,
            ReplayType replayMethod,
            UpdaterType updater = UpdaterType(),
            EnvironmentType environment = EnvironmentType());

  QLearning(TrainingConfig config,
            NetworkType network,
            PolicyType policy,
            PrioritizedReplayType prioritizedReplayMethod,
            UpdaterType updater = UpdaterType(),
            EnvironmentType environment = EnvironmentType());
```

manish7294: > But I have a question about that, how do we know which replay buffer passed into the q_learning class? 

Sorry, I don't seem to understand the question. Isn't the type of replay you pass to the Qlearning agent will be the one, it will be using. Or am I missing something here?

> If we don't use inheritance, how to deal with the situation when we have two constructors?

We actually don't need to have two separate constructors, if we reuse the ReplayType template parameter instead of having a separate template parameter of prioritized replay. Does it clear things up?

zoq: Right, in the example above, ``PrioritizedReplayType`` would  just be ``ReplayType`, and ass @manish7294 already pointed out this makes the second constructor obsolete.

robotcator: > Right, in the example above, `PrioritizedReplayType` would just be ``ReplayType`, and ass @manish7294 already pointed out this makes the second constructor obsolete.

Sorry, it seems that I missing something. I do want to reuse the `ReplayType`, but I didn't realize that we can reuse the template parameter.

Now, things are all cleared. But if we use `prioritizedReplayMethod`, we should use the Q-learning class like this:
```
  QLearning<CartPole, decltype(model), AdamUpdate, decltype(policy),
      decltype(replayMethod)>
    agent(std::move(config), std::move(model), std::move(policy),
          std::move(replayMethod));
```
because the default replaymethod is `RandomReplay`.

zoq: True, we have to pass the correct type for the replay method.

src/mlpack/methods/reinforcement_learning/replay/prioritized_replay.hpp
```diff
@@ -0,0 +1,296 @@
...
+    while (size < capacity) {
+      size *= 2;
+    }
+
+    beta = initialBeta;
+    idxSum = SumTree<double>(size);
+  }
+
+  /**
+   * Store the given experience.
```
zoq: We might want to change the wording here, what about "Store the given experience and set priorities"?

src/mlpack/methods/reinforcement_learning/replay/prioritized_replay.hpp
```diff
@@ -0,0 +1,296 @@
...
+   * Sample some experience according to their priorities.
+   *
+   * @return The indices to be chosen.
+   */
+  arma::ucolvec SampleProportional()
+  {
+    arma::ucolvec idxes(batchSize);
+    double totalSum = idxSum.Sum(0, (full ? capacity : position));
+    double sumPerRange = totalSum / batchSize;
+    for (size_t bt = 0; bt < batchSize; bt++) {
```
zoq: Picky style comment, can you put the opening ``{`` on a new line.

src/mlpack/methods/reinforcement_learning/replay/prioritized_replay.hpp
```diff
@@ -0,0 +1,296 @@
...
+  /**
+   * Sample some experience according to their priorities.
+   *
+   * @param sampledStates Sampled encoded states.
+   * @param sampledActions Sampled actions.
+   * @param sampledRewards Sampled rewards.
+   * @param sampledNextStates Sampled encoded next states.
+   * @param isTerminal Indicate whether corresponding next state is terminal
+   *        state.
+   * @param sampledIndices Sampled indices.
```
zoq: Looks like we can remove the two parameters.

src/mlpack/methods/reinforcement_learning/replay/prioritized_replay.hpp
```diff
@@ -0,0 +1,296 @@
...
+   * @param weights Corresponding weight for updating the loss.
+   */
+  void Sample(arma::mat& sampledStates,
+              arma::icolvec& sampledActions,
+              arma::colvec& sampledRewards,
+              arma::mat& sampledNextStates,
+              arma::icolvec& isTerminal)
+  {
+//    size_t upperBound = full ? capacity : position;
```
zoq: Looks like this can be removed.

src/mlpack/methods/reinforcement_learning/replay/prioritized_replay.hpp
```diff
@@ -0,0 +1,296 @@
...
+    weights /= weights.max();
+  }
+
+  /**
+   * Update priorities of sampled transitions.
+   *
+   * @param indices The indices of sample to be updated.
+   * @param priorities Their corresponding priorities.
+   */
+  void UpdatePriorities(arma::ucolvec& indices, arma::colvec& priorities)
```
zoq: Can we use ``const`` on both parameter.

src/mlpack/methods/reinforcement_learning/replay/prioritized_replay.hpp
```diff
@@ -0,0 +1,296 @@
...
+  /**
+   *  Annealing the beta.
+   */
+  void BetaAnneal()
+  {
+    beta = beta + (1 - initialBeta) * 1.0 / replayBetaIters;
+  }
+
+  /**
+    *  Update the priorities of transitions and Update the gradients.
```
zoq: All comments below are one space off.

src/mlpack/methods/reinforcement_learning/replay/random_replay.hpp
```diff
@@ -135,6 +142,22 @@ class RandomReplay
     return full ? capacity : position;
   }
 
+  /**
+    *  Update the priorities of transitions and Update the gradients.
+    *
+    *  @param target The learned value
```
zoq: Missing stop at the end.

src/mlpack/methods/reinforcement_learning/replay/sumtree.hpp
```diff
@@ -0,0 +1,194 @@
...
+#include <mlpack/prereqs.hpp>
+
+namespace mlpack {
+namespace rl {
+
+/**
+ * Implementation of SumTree.
+ *
+ * Build a Segment Tree like data structure.
+ * https://en.wikipedia.org/wiki/Segment_tree
```
zoq: Do we have some paper we can reference here?

src/mlpack/methods/reinforcement_learning/replay/sumtree.hpp
```diff
@@ -0,0 +1,194 @@
...
+ *
+ * Used to maintain prefix-sum of an array.
+ *
+ */
+
```
zoq: We can remove the empty line here. Also can you add a comment for the template parameter?

src/mlpack/methods/reinforcement_learning/replay/sumtree.hpp
```diff
@@ -0,0 +1,194 @@
...
+template<typename T>
+class SumTree
+{
+ public:
+  /**
+   * Default constructor.
+   */
+  SumTree():
```
zoq: This should be:

``SumTree :  capacity(0)``

src/mlpack/methods/reinforcement_learning/replay/sumtree.hpp
```diff
@@ -0,0 +1,194 @@
...
+      capacity(0)
+  { /* Nothing to do here. */ }
+
+  /**
+   * Construct an instance of SumTree class.
+   *
+   * @param capacity Size of data.
+   */
+  SumTree(size_t capacity):
```
zoq: See comment above.

src/mlpack/methods/reinforcement_learning/replay/sumtree.hpp
```diff
@@ -0,0 +1,194 @@
...
+      capacity(capacity)
+  {
+    element = std::vector<T> (2 * capacity);
+  }
+
+  /**
+   * Set the data array with idx.
+   *
+   * @param idx The array idx to be changed.
+   * @param value The data that array with idx to be.
```
zoq: Not sure I get the comment.

src/mlpack/methods/reinforcement_learning/replay/sumtree.hpp
```diff
@@ -0,0 +1,194 @@
...
+    idx += capacity;
+    element[idx] = value;
+    idx /= 2;
+    while (idx >= 1)
+    {
+      element[idx] = element[2 * idx] + element[2 * idx + 1];
+      idx /= 2;
+    }
+  }
+
```
zoq: We can remove the empty line here.

src/mlpack/methods/reinforcement_learning/replay/sumtree.hpp
```diff
@@ -0,0 +1,194 @@
...
+    element = std::vector<T> (2 * capacity);
+  }
+
+  /**
+   * Set the data array with idx.
+   *
+   * @param idx The array idx to be changed.
+   * @param value The data that array with idx to be.
+   */
+  void Set(size_t idx, T value)
```
zoq: We might want to use ``const`` if possible that applies to the other methods as well.

src/mlpack/methods/reinforcement_learning/replay/sumtree.hpp
```diff
@@ -0,0 +1,194 @@
...
+   * @param indices The indices of data to be changed.
+   * @param data  The data that array with indices to be.
+   */
+  void BatchUpdate(arma::ucolvec indices, arma::Col<T> data)
+  {
+    for (size_t i = 0; i < indices.n_rows; i++)
+    {
+      element[indices[i] + capacity] = data[i];
+    }
+    // update the total tree with bottom-up technique
```
zoq: Missing stop at the end.

src/mlpack/methods/reinforcement_learning/replay/sumtree.hpp
```diff
@@ -0,0 +1,194 @@
...
+  T Sum()
+  {
+    return Sum(0, capacity);
+  }
+
+  /**
+   * Find the highest index `idx` in the array such that
+   * sum(arr[0] + arr[1] + ... + arr[idx]) <= mass.
+   *
+   * @param mass
```
zoq: This is missing a description.

src/mlpack/methods/reinforcement_learning/replay/sumtree.hpp
```diff
@@ -0,0 +1,194 @@
...
+          SumHelper(mid+1, end, 2 * node + 1, mid + 1 , nodeEnd);
+      }
+    }
+  }
+
+  /**
+   * Calculate the sum of contiguous subsequence of the array.
+   *
+   * @param start The starting position of subsequence.
+   * @param _end The end position of subsequence.
```
zoq: I think you meant to write ``end``.

src/mlpack/tests/q_learning_test.cpp
```diff
@@ -102,6 +102,73 @@ BOOST_AUTO_TEST_CASE(CartPoleWithDQN)
   BOOST_REQUIRE(converged);
 }
 
+//! Test DQN in Cart Pole task with Prioritized Replay
```
zoq: Missing stop at the end.

src/mlpack/tests/q_learning_test.cpp
```diff
@@ -102,6 +102,73 @@ BOOST_AUTO_TEST_CASE(CartPoleWithDQN)
...
+  config.Discount() = 0.9;
+  config.TargetNetworkSyncInterval() = 100;
+  config.ExplorationSteps() = 100;
+  config.DoubleQLearning() = false;
+  config.StepLimit() = 200;
+
+  // Set up DQN agent.
+  QLearning<CartPole, decltype(model), AdamUpdate, decltype(policy),
+      decltype(replayMethod)>
+    agent(std::move(config), std::move(model), std::move(policy),
```
zoq: This should be tabbed twice (4 spaces).

src/mlpack/methods/reinforcement_learning/replay/prioritized_replay.hpp
```diff
@@ -0,0 +1,295 @@
...
+   *
+   * @param batchSize Number of examples returned at each sample.
+   * @param capacity Total memory size in terms of number of examples.
+   * @param alpha How much prioritization is used.
+   * @param dimension The dimension of an encoded state.
+   */
+  PrioritizedReplay(const size_t batchSize,
+                    const size_t capacity,
+                    const double alpha,
+                    int seed = 1024,
```
zoq: No need to introduce a seed here, we can use the global state, which can be set with: `mlpack::math::RandomSeed(mused);`

zoq:
If you can fix the comment I left I think this is ready, I think there are two or three style issue, but I can fix those during the merge process.

robotcator:
> If you can fix the comment I left I think this is ready, I think there are two or three style issue, but I can fix those during the merge process.

Thanks, I will fix that, it seems that there are lots of code style issue, I will read the guideline carefully again.

zoq:
If you can remove the seed from the code, I think this is ready.

src/mlpack/methods/reinforcement_learning/replay/prioritized_replay.hpp
robotcator: > If you can remove the seed from the code, I think this is ready.

Sorry for missing that.

robotcator: > > If you can remove the seed from the code, I think this is ready.
> 
> Sorry for missing that.

@zoq, Do you mean passing a `seed` parameter or using ` math::RandomSeed(std::time(NULL))` like in `hmm_test.cpp`.

zoq: I was talking about using: ``math::RandomSeed(std::time(NULL))``, looks like you already removed the seed, so I don't think there is anything else to do.

robotcator: > I was talking about using: `math::RandomSeed(std::time(NULL))`, looks like you already removed the seed, so I don't think there is anything else to do.

Great.

zoq:
Excited to use this in the future.

zoq:
I fixed some style issue here: https://github.com/mlpack/mlpack/commit/cf1b948ece5cd1ab7efe34b6bac6835c99fff742

robotcator:
> I fixed some style issue here: [cf1b948](https://github.com/mlpack/mlpack/commit/cf1b948ece5cd1ab7efe34b6bac6835c99fff742)

Thanks for helping me fix the code style problem. How can I check those minor problem thoroughly, such as the [unused space](https://github.com/mlpack/mlpack/blob/cf1b948ece5cd1ab7efe34b6bac6835c99fff742/src/mlpack/methods/reinforcement_learning/replay/random_replay.hpp#L158)?

PR #1615, opened 2018-12-26
ffn.hpp and rnn.hpp: Documentation fix

saksham189:
Documentation fix for EvaluateWithGradient.

### DIFF
--- a/src/mlpack/methods/ann/rnn_impl.hpp
+++ b/src/mlpack/methods/ann/rnn_impl.hpp
@@ -304,13 +304,10 @@ EvaluateWithGradient(const arma::mat& /* parameters */,
       responseSeq = seqNum;
     }
 
-    if (!deterministic)
+    for (size_t l = 0; l < network.size(); ++l)
     {
-      for (size_t l = 0; l < network.size(); ++l)
-      {
-        boost::apply_visitor(SaveOutputParameterVisitor(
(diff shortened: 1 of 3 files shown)


src/mlpack/methods/ann/rnn_impl.hpp
```diff
@@ -304,13 +304,10 @@ EvaluateWithGradient(const arma::mat& /* parameters */,
       responseSeq = seqNum;
     }
 
-    if (!deterministic)
```
rcurtin: I agree with the documentation changes, but I'm really not sure about this one.  Why should we remove the `if (!deterministic)` here?

saksham189: I think on this https://github.com/mlpack/mlpack/pull/1615/files#diff-0e560b4ae5054e7fe4c029ec9eb0f1a7R275
we set `deterministic` to false.

saksham189: `EvaluateWithGradient` is only being used for training otherwise we could call `Evaluate`.

rcurtin: Thanks for the clarification; that makes sense to me.

ShikharJ:
@saksham189 Thanks a lot for the contribution.

Issue #1565, opened 2018-11-13
option inconsistency --output_file vs --predictions_file


barak:
Why is it
* `mlpack_perceptron --output_file` and
* `mlpack_logistic --output_file` but
* `mlpack_decision_tree --predictions_file`?

rcurtin:
Oversight. :)

I think that we should change all to `predictions_file`, but we have to do this in a reverse-compatible way until mlpack 4.0.0 when we can remove the `output_file` options.

If anyone is looking for a new task to do, I think this is a good one:

 1. Find all bindings in `src/mlpack/methods/` that have some line like `PARAM_MATRIX_OUT("output", ...)` where the output matrix is some output predictions.  That's probably just `mlpack_perceptron` and `mlpack_logistic_regression` but there may be some others so we should check them all.
 2. Add an option `PARAM_MATRIX_OUT("predictions", ...)` that does the same thing.
 3. Add a comment pointing out that `output` can be removed in mlpack 4.0.0, and modify the documentation for the `output` option to point out that it is deprecated.
 4. Ensure that, for now, the output matrix for `predictions` and `output` is identical.
 5. Update any tests in `src/mlpack/tests/main_tests/` to make sure that `predictions` and `output` are the same.

e-aakash:
@rcurtin Can i take up this issue?

Hemal-Mamtora:
I've found all occurences of PARAM_MATRIX_OUT in `src/mlpack/methods/` which are as follows:

kmeans/kmeans_main.cpp:PARAM_MATRIX_OUT
kmeans/kmeans_main.cpp:PARAM_MATRIX_OUT
lmnn/lmnn_main.cpp:PARAM_MATRIX_OUT
lmnn/lmnn_main.cpp:PARAM_MATRIX_OUT
lmnn/lmnn_main.cpp:PARAM_MATRIX_OUT
decision_tree/decision_tree_main.cpp:PARAM_MATRIX_OUT
det/det_main.cpp:PARAM_MATRIX_OUT
det/det_main.cpp:PARAM_MATRIX_OUT
det/det_main.cpp:PARAM_MATRIX_OUT
radical/radical_main.cpp:PARAM_MATRIX_OUT
radical/radical_main.cpp:PARAM_MATRIX_OUT
rann/krann_main.cpp:PARAM_MATRIX_OUT
pca/pca_main.cpp:PARAM_MATRIX_OUT
[...]

PR #1616, opened 2018-12-26
Option consistency in logistic regression #1565

Hemal-Mamtora:
With respect to issue #1565 : 
I have made changes to the `logistic_regression_main.cpp`
I have tested the binding for renamed parameters/options.

@rcurtin kindly review the changes. Let me know if any updates are needed. 

Thank you

### DIFF
--- a/src/mlpack/methods/logistic_regression/logistic_regression_main.cpp
+++ b/src/mlpack/methods/logistic_regression/logistic_regression_main.cpp
@@ -38,7 +38,7 @@ PROGRAM_INFO("L2-regularized Logistic Regression and Prediction",
     "those things at once.  In addition, this program allows classification on "
     "a test dataset (specified with the " + PRINT_PARAM_STRING("test") + " "
     "parameter) and the classification results may be saved with the " +
-    PRINT_PARAM_STRING("output") + " output parameter.  The trained logistic "
+    PRINT_PARAM_STRING("predictions") + " predictions parameter.  The trained logistic "
     "regression model may be saved using the " +
     PRINT_PARAM_STRING("output_model") + " output parameter."
     "\n\n"
@@ -75,7 +75,9 @@ PROGRAM_INFO("L2-regularized Logistic Regression and Prediction",
     "so long as an existing logistic regression model is given with the " +
     PRINT_PARAM_STRING("input_model") + " parameter.  The output predictions "
     "from the logistic regression model may be saved with the " +
-    PRINT_PARAM_STRING("output") + " parameter."
+    PRINT_PARAM_STRING("predictions") + " parameter." + 
+    "The" + PRINT_PARAM_STRING("output") + "parameter is depracated" + 
+    " and to be removed in mlpack-4.0.0 "
     "\n\n"
     "This implementation of logistic regression does not support the general "
     "multi-class case but instead only the two-class case.  Any labels must "
@@ -123,11 +125,19 @@ PARAM_MODEL_OUT(LogisticRegression<>, "output_model", "Output for trained "
 
 // Testing.
 PARAM_MATRIX_IN("test", "Matrix containing test dataset.", "T");
+//The PARAM_UROW_OUT("output"..) is depracated and can be removed in mlpack4.0.0
 PARAM_UROW_OUT("output", "If test data is specified, this matrix is where "
     "the predictions for the test set will be saved.", "o");
+PARAM_UROW_OUT("predictions", "If test data is specified, this matrix is where "
+    "the predictions for the test set will be saved.", "P");
+//PARAM_MATRIX_OUT("output_probabilities"..) is depracated 
+//and it can be removed in mlpack 4.0.0
 PARAM_MATRIX_OUT("output_probabilities", "If test data is specified, this "
     "matrix is where the class probabilities for the test set will be saved.",
     "p");
+PARAM_MATRIX_OUT("probabilities", "If test data is specified, this "
+    "matrix is where the class probabilities for the test set will be saved.",
+    "x");
 PARAM_DOUBLE_IN("decision_boundary", "Decision boundary for prediction; if the "
     "logistic function for a point is less than the boundary, the class is "
     "taken to be 0; otherwise, the class is 1.", "d", 0.5);
@@ -154,11 +164,18 @@ static void mlpackMain()
         "be saved");
   }
 
-  RequireAtLeastOnePassed({ "output_model", "output", "output_probabilities" },
+  // options "output" and "output_probabilites" are depracated and replaced by 
+  //"predictions" and "probabilties" respectively 
+  // options "output" and "output_probabilites" can be removed in mlpack 4.0.0
+  RequireAtLeastOnePassed({ "output_model", "output", "output_probabilities",
+  "predictions", "probabilities"},
       false, "no output will be saved");
 
+  //"output" and "output_probabilities" lines can be removed in mlpack 4.0.0
   ReportIgnoredParam({{ "test", false }}, "output");
   ReportIgnoredParam({{ "test", false }}, "output_probabilities");
+  ReportIgnoredParam({{ "test", false }}, "predictions");
+  ReportIgnoredParam({{ "test", false }}, "probabilites");
 
   // Max Iterations needs to be positive.
   RequireParamValue<int>("max_iterations", [](int x) { return x >= 0; },
@@ -317,6 +334,8 @@ static void mlpackMain()
           << trainingDimensionality << ")!" << endl;
     }
 
+    // The CLI param "output" is depracated and replaced by "predictions"
+    // The following if statement can be removed in mlpack 4.0.0
     // We must perform predictions on the test set.  Training (and the
     // optimizer) are irrelevant here; we'll pass in the model we have.
     if (CLI::HasParam("output"))
@@ -328,6 +347,19 @@ static void mlpackMain()
       CLI::GetParam<arma::Row<size_t>>("output") = std::move(predictions);
     }
 
+    // We must perform predictions on the test set.  Training (and the
+    // optimizer) are irrelevant here; we'll pass in the model we have.
+    if (CLI::HasParam("predictions"))
+    {
+      Log::Info << "Predicting classes of points in '"
+          << CLI::GetPrintableParam<arma::mat>("test") << "'." << endl;
+      model->Classify(testSet, predictions, decisionBoundary);
+
+      CLI::GetParam<arma::Row<size_t>>("predictions") = std::move(predictions);
+    }
+
+    //"output_probabilities" parameter is depracated and replaced by 
+    //"probabilities". The following if statement can be removed in mlpack 4.0.0
     if (CLI::HasParam("output_probabilities"))
     {
       Log::Info << "Calculating class probabilities of points in '"
@@ -336,6 +368,17 @@ static void mlpackMain()
       model->Classify(testSet, probabilities);
 
       CLI::GetParam<arma::mat>("output_probabilities") =
+          std::move(probabilities);
+    }
+
+    if (CLI::HasParam("probabilities"))
+    {
+      Log::Info << "Calculating class probabilities of points in '"
+          << CLI::GetPrintableParam<arma::mat>("test") << "'." << endl;
+      arma::mat probabilities;
+      model->Classify(testSet, probabilities);
+
+      CLI::GetParam<arma::mat>("probabilities") =
           std::move(probabilities);
     }
   }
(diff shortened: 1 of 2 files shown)


ShikharJ:
I'm not sure if these changes are necessary. @rcurtin What do you think?

src/mlpack/methods/logistic_regression/logistic_regression_main.cpp
```diff
@@ -38,8 +38,8 @@ PROGRAM_INFO("L2-regularized Logistic Regression and Prediction",
     "those things at once.  In addition, this program allows classification on "
     "a test dataset (specified with the " + PRINT_PARAM_STRING("test") + " "
     "parameter) and the classification results may be saved with the " +
-    PRINT_PARAM_STRING("output") + " output parameter.  The trained logistic "
-    "regression model may be saved using the " +
+    PRINT_PARAM_STRING("predictions") + " predictions parameter."
```
rcurtin: Actually we should leave this as `output parameter` and not change to `predictions parameter`.  Here we are referring to the fact that the parameter `predictions` is an output parameter (as opposed to an input parameter).

src/mlpack/methods/logistic_regression/logistic_regression_main.cpp
```diff
@@ -75,7 +75,9 @@ PROGRAM_INFO("L2-regularized Logistic Regression and Prediction",
     "so long as an existing logistic regression model is given with the " +
     PRINT_PARAM_STRING("input_model") + " parameter.  The output predictions "
     "from the logistic regression model may be saved with the " +
-    PRINT_PARAM_STRING("output") + " parameter."
+    PRINT_PARAM_STRING("predictions") + " parameter." + 
+    "The" + PRINT_PARAM_STRING("output") + "parameter is depracated" +
```
rcurtin: deprecated :)

Also I would suggest putting this in its own paragraph (so a `\n\n` before the `The`), and we should point out that `output_probabilities` is also deprecated.

src/mlpack/methods/logistic_regression/logistic_regression_main.cpp
```diff
@@ -75,7 +75,9 @@ PROGRAM_INFO("L2-regularized Logistic Regression and Prediction",
     "so long as an existing logistic regression model is given with the " +
     PRINT_PARAM_STRING("input_model") + " parameter.  The output predictions "
     "from the logistic regression model may be saved with the " +
-    PRINT_PARAM_STRING("output") + " parameter."
+    PRINT_PARAM_STRING("predictions") + " parameter." + 
+    "The" + PRINT_PARAM_STRING("output") + "parameter is depracated" + 
+    " and to be removed in mlpack-4.0.0 "
```
rcurtin: `to be` -> `will be`, `mlpack-4.0.0` -> `mlpack 4.` (minor grammatical fixes)

src/mlpack/methods/logistic_regression/logistic_regression_main.cpp
```diff
@@ -154,11 +164,18 @@ static void mlpackMain()
         "be saved");
   }
 
-  RequireAtLeastOnePassed({ "output_model", "output", "output_probabilities" },
+  // options "output" and "output_probabilites" are depracated and replaced by 
+  //"predictions" and "probabilties" respectively 
+  // options "output" and "output_probabilites" can be removed in mlpack 4.0.0
+  RequireAtLeastOnePassed({ "output_model", "output", "output_probabilities",
+  "predictions", "probabilities"},
```
rcurtin: A wrapped line should be indented twice.

src/mlpack/methods/logistic_regression/logistic_regression_main.cpp
```diff
@@ -328,6 +347,19 @@ static void mlpackMain()
       CLI::GetParam<arma::Row<size_t>>("output") = std::move(predictions);
```
rcurtin: This `CLI::GetParam<...>` call is the only line that needs to be in the `if` statement.  Can you condense this (and the following bit of code) a little accordingly please?

src/mlpack/methods/logistic_regression/logistic_regression_main.cpp
Hemal-Mamtora: should I also mention the respective replacements for the deprecated parameters in the paragraph ?

Hemal-Mamtora:
>  but we should also test in src/mlpack/tests/main_tests/logistic_regression_test.cpp that the predictions and probabilities outputs work correctly, instead of the old output and output_probabilities names.

@rcurtin, What I understand is, at this point of time, the tests for all 4:  `output`,`predictions` , `output_probabilities` and `probabilities` should work.

Or is it the case that the 2 new options tests should work `instead of` the 2 old options. 
Please clarify.

And please review the further changes that I have done.

src/mlpack/methods/logistic_regression/logistic_regression_main.cpp
rcurtin: I mean this...

```
if (CLI::HasParam("output") || CLI::HasParam("predictions"))
{
  Log::Info << "Predicting classes of points in '"
      << CLI::GetPrintableParam<arma::mat>("test") << "'." << endl;
  model->Classify(testSet, predictions, decisionBoundary);

  // 'output' and 'predictions' are the same, so output to both.
  CLI::GetParam<arma::Row<size_t>>("output") = predictions;
  CLI::GetParam<arma::Row<size_t>>("predictions") = std::move(predictions);
}
```

What do you think, did I overlook anything?

Hemal-Mamtora: I am not sure on how this code would behave in case when param `output` is not defined and we try to store values in `output` as well as parameters. It might give error: param `output` not defined.
I'll still try this code. Thanks for your reply :)

rcurtin: So actually the code should not have an error, since we can call `CLI::GetParam<...>("output")` even if `CLI::HasParam("output")` is false---the only requirement is that there is a `PARAM_*("output", ...)` somewhere.

But, now that I think about it, we shouldn't set those outputs if the user didn't ask for them---`CLI::HasParam("output")` will only be true if the user has asked for the output.  So we can make a simple change---

```
if (CLI::HasParam("output"))
  CLI::GetParam<arma::Row<size_t>>("output") = predictions;
if (CLI::HasParam("predictions"))
  CLI::GetParam<arma::Row<size_t>>("predictions") = std::move(predictions);
```

Hemal-Mamtora: That looks good, will update it :)

Hemal-Mamtora:
@rcurtin , I don't understand why this build fails: `continuous-integration/travis-ci/pr — The Travis CI build failed`

Since the changes had built successfully on my local pc.

The details of the failed build are : 
`/home/travis/build/mlpack/mlpack/src/mlpack/tests/main_tests/logistic_regression_test.cpp(120): fatal error in "LRPredictionSizeCheck": critical check testY.n_cols == M failed [0 != 15]
unknown location(0): fatal error in "LRResponsesRepresentationTest": memory access violation at address: 0x7ffd5e51c000: no mapping at fault address
/home/travis/build/mlpack/mlpack/src/mlpack/tests/main_tests/logistic_regression_test.cpp(142): last checkpoint`

It looks like `predictions` is not getting set.

Could you guide me on how to solve this issue?
Please also have a look at `src/mlpack/tests/main_tests/logistic_regression_test.cpp`

Thank you for your guidance !

rcurtin:
You can't do `std::move()` twice---once you call `std::move()` the first time, the object no longer has any data in it.  That's why in the example snippet I posted I only used `std::move()` in the second call.  If you fix that, I think that the tests will succeed.

Hemal-Mamtora:
@rcurtin , I updated the std::move related code. Still one fatal error remains which is as follows:

`unknown location(0): fatal error in "LRResponsesRepresentationTest": memory access violation at address: 0x7ffd0615f000: no mapping at fault address`

Memory access violation, Please let me know what I have missed.

Thank you !

rcurtin:
Hi @Hemal-Mamtora, what have you tried to debug the issue?

Hemal-Mamtora:
@rcurtin I was trying to debug the issue of no mapping at fault at fault address.
I tried keeping only 'predictions' in place of output in logistic_regre

Hemal-Mamtora:
Regarding the debugging of the issue,
I tried :
1. Keeping separate test cases for 'predictions' and 'output' parameter
2. Then keeping only the 'predictions' parameter for certain test cases
3. Replacing all the 'output' parameter by 'predictions' parameter.

All these changes can be found on the above commits

Some new fatal errors have cropped up. It says: 
[FATAL] Parameter '--probabilites' does not exist in this program.

But according to me the parameter 'probabilities' has no role in the logistic_regression_test.cpp
Since it's replacement 'output_probabilities' parameter is not seen in the file : logistic_regression_test.cpp

Please let me know on how to tackle this issue.

Thanks !

rcurtin:
I think you have a typo in your code :)

```
[FATAL] Parameter '--probabilites' does not exist in this program.
```

Shouldn't that be `probabilities` not `probabilites`?  The problem seems to be on line 183.

Hemal-Mamtora:
OK, I am working on it! 
Thank you for your review !

Hemal-Mamtora:
@rcurtin , I have added the 2 test cases to verify that the output from the old and the new parameters is the same. Please review. And please let me know if any changes are needed.
Thank you !

src/mlpack/methods/logistic_regression/logistic_regression_main.cpp
```diff
@@ -75,7 +75,15 @@ PROGRAM_INFO("L2-regularized Logistic Regression and Prediction",
     "so long as an existing logistic regression model is given with the " +
     PRINT_PARAM_STRING("input_model") + " parameter.  The output predictions "
     "from the logistic regression model may be saved with the " +
-    PRINT_PARAM_STRING("output") + " parameter."
+    PRINT_PARAM_STRING("predictions") + " parameter." + 
+    "\n\n"
+    "Note : The following parameters are deprecated and "
```
rcurtin: This is a tiny issue but we can remove the space here so it reads `Note:` :)

src/mlpack/methods/logistic_regression/logistic_regression_main.cpp
```diff
@@ -75,7 +75,15 @@ PROGRAM_INFO("L2-regularized Logistic Regression and Prediction",
...
     PRINT_PARAM_STRING("input_model") + " parameter.  The output predictions "
     "from the logistic regression model may be saved with the " +
-    PRINT_PARAM_STRING("output") + " parameter."
+    PRINT_PARAM_STRING("predictions") + " parameter." + 
+    "\n\n"
+    "Note : The following parameters are deprecated and " 
+    "will be removed in mlpack 4: " + PRINT_PARAM_STRING("output") + 
+    ", " + PRINT_PARAM_STRING("output_probabilities") + 
+    "\nUse " + PRINT_PARAM_STRING("predictions") + " instead of " + 
+    PRINT_PARAM_STRING("output") + "\nUse " +
```
rcurtin: We should add a period to the end of each of these lines, so instead of `"\nUse "`, maybe `".\nUse "`. :)

src/mlpack/tests/main_tests/logistic_regression_test.cpp
```diff
@@ -634,10 +634,76 @@ BOOST_AUTO_TEST_CASE(LRDecisionBoundaryTest)
...
+  const arma::Row<size_t> testY2 =
+      std::move(CLI::GetParam<arma::Row<size_t>>("output"));
+
+  // Both solutions must be equal.
+  BOOST_REQUIRE_EQUAL_COLLECTIONS(testY1.begin(), testY1.end(),
+                                  testY2.begin(), testY2.end());
+}
+
+/**
+  * Ensuring that the parameter 'output_probabilities' and the parameter
```
rcurtin: This is also pretty trivial but there should only be one space before the `*` here.

rcurtin:
Thanks for the contribution!  I have another bug to work out with mlpack-bot... it is supposed to offer you stickers after merge.  If it did, it would say this:

```
Hello there!  Thanks for your contribution.  I see that this is your first contribution to mlpack.  If you'd like to add your name to the list of contributors in `src/mlpack/core.hpp` and `COPYRIGHT.txt` and you haven't already, please feel free to push a change to this PR---or, if it gets merged before you can, feel free to open another PR.

In addition, if you'd like some stickers to put on your laptop, I'd be happy to help get them in the mail for you.  Just send an email with your physical mailing address to stickers@mlpack.org, and then one of the mlpack maintainers will put some stickers in an envelope for you.  It may take a few weeks to get them, depending on your location. :+1:
```

PR #1622, opened 2018-12-27
Fixes for intermittently failing tests

rcurtin:
Looking through the jobs on ci.mlpack.org, I took a list of tests that were failing and tried to adjust tolerances on each of them to reduce the failure probability.

@saksham189: some of the kernel PCA main tests were failing intermittently.  For instance, the sampling scheme test would fail because the new dimensionality was only 1, meaning that the Nystroem method would select only one point.  Since the dataset only had 5 points, the probability of failure was the probability that the random sampling scheme would select the same point as the ordered sampling scheme, which in this case was 20%.  A larger dataset and a higher new dimensionality appear to have fixed the problem.  One other test needed a bit of changes too.  Take a look at the changes and let me know if I broke anything (or if I missed anything) please. :)


rcurtin:
Thanks for the comments everyone. Hopefully we can get a green matrix build again. If not I will debug more tests and open another PR. :)

Issue #1580, opened 2018-11-24
Multidimensional RNN in mlpack ?


kvcuong:
Hi,
I've read about CRNN net papers for word-spotting, they reported really good result. Does Mlpack support Multi-Dim RNN like Bi-LSTM or MD-LSTM and CTC layers in order to build CRNN ?

I saw that mlpack has lstm, how can i stack 2 lstm to build Bi-LSTM ? 

thanks in advance.

zoq:
Currently, there is no support for MD-LSTM or B-LSTM; in case of B-LSTM it could be simulated by using two instances of the RNN class that are merged after each iteration.

shravankumar9892:
@zoq  So is it worth creating a blstm.hpp separately in the methods/ann/layer directory? Does it apply to brnn as well?

zoq:
What I had envisioned is another class similar to the ``RNN`` class, so there is no need to adjust or rewrite any layer. The basic idea would be to do two passes at the "same" time and merge the output at the end. Perhaps we could even reuse the RNN class internally. Let me know what you think.

sourabhvarshney111:
@zoq @shravankumar9892 I think first we can try to make that BRNN class similar to RNN class, which I took the responsibility this summer, but couldn't complete due to my tight schedule. I think that would be a good start.

shravankumar9892:
Yeah can work on the issue @sourabhvarshney111

zoq:
I think the easiest thing would be to start with the Forward pass, so right now we perform a single Forward step:

https://github.com/mlpack/mlpack/blob/823c34d46c67c6ff652af0bdf3ebbaddc96bc2b8/src/mlpack/methods/ann/rnn_impl.hpp#L208-L211

to get this to work, we also have to do another Forward step with the data from the other direction.

shravankumar9892:
Yes @zoq, we'll have to do it from other direction as well.

zoq:
Okay, great, let me know if I should clarify anything.

PR #1626, opened 2019-01-01
Add Bidirectional RNN

saksham189:
TODOS

- [x] Add test

Refer https://github.com/mlpack/mlpack/issues/1580

### DIFF
--- a/src/mlpack/methods/ann/brnn_impl.hpp
+++ b/src/mlpack/methods/ann/brnn_impl.hpp
@@ -329,10 +329,10 @@ EvaluateWithGradient(const arma::mat& /* parameters */,
   if (backwardGradient.is_empty())
   {
     backwardGradient = arma::zeros<arma::mat>(
-        parameter.n_rows/2,
+        parameter.n_rows/ 2,
         parameter.n_cols);
     forwardGradient = arma::zeros<arma::mat>(
-        parameter.n_rows/2,
+        parameter.n_rows/ 2,
         parameter.n_cols);
   }
   if (this->deterministic)
@@ -650,7 +650,6 @@ void BRNN<OutputLayerType, MergeLayerType, MergeOutputType,
   forwardRNN.ResetDeterministic();
   backwardRNN.ResetDeterministic();
 }
-
 
 template<typename OutputLayerType, typename MergeLayerType,
          typename MergeOutputType, typename InitializationRuleType,
--- a/src/mlpack/methods/ann/brnn_impl.hpp
+++ b/src/mlpack/methods/ann/brnn_impl.hpp
@@ -248,16 +248,6 @@ double BRNN<OutputLayerType, MergeLayerType, MergeOutputType,
         predictors.slice(rho - seqNum - 1).colptr(begin),
         predictors.n_rows, batchSize, false, true)));
 
-    if (!deterministic)
-    {
-      for (size_t l = 0; l < forwardRNN.network.size(); ++l)
-      {
-        boost::apply_visitor(SaveOutputParameterVisitor(
-            std::move(forwardRNNOutputParameter)), forwardRNN.network[l]);
-        boost::apply_visitor(SaveOutputParameterVisitor(
-            std::move(backwardRNNOutputParameter)), backwardRNN.network[l]);
-      }
-    }
     boost::apply_visitor(SaveOutputParameterVisitor(
         std::move(results1)), forwardRNN.network.back());
     boost::apply_visitor(SaveOutputParameterVisitor(
--- a/src/mlpack/methods/ann/brnn_impl.hpp
+++ b/src/mlpack/methods/ann/brnn_impl.hpp
@@ -1,0 +1,670 @@
+/**
+ * @file brnn_impl.hpp
+ * @author Saksham Bansal
+ *
+ * Definition of the BRNN class, which implements bidirectional recurrent
+ * neural networks.
+ *
+ * mlpack is free software; you may redistribute it and/or modify it under the
+ * terms of the 3-clause BSD license.  You should have received a copy of the
+ * 3-clause BSD license along with mlpack.  If not, see
+ * http://www.opensource.org/licenses/BSD-3-Clause for more information.
+ */
+#ifndef MLPACK_METHODS_ANN_BRNN_IMPL_HPP
+#define MLPACK_METHODS_ANN_BRNN_IMPL_HPP
+
+// In case it hasn't been included yet.
+#include "brnn.hpp"
+
+#include "visitor/load_output_parameter_visitor.hpp"
+#include "visitor/save_output_parameter_visitor.hpp"
+#include "visitor/forward_visitor.hpp"
+#include "visitor/backward_visitor.hpp"
+#include "visitor/reset_cell_visitor.hpp"
+#include "visitor/gradient_set_visitor.hpp"
+#include "visitor/gradient_visitor.hpp"
+#include "visitor/weight_set_visitor.hpp"
+#include "visitor/run_set_visitor.hpp"
+
+#include <boost/serialization/variant.hpp>
+
+namespace mlpack {
+namespace ann /** Artificial Neural Network. */ {
+
+template<typename OutputLayerType, typename MergeLayerType,
+         typename MergeOutputType, typename InitializationRuleType,
+         typename... CustomLayers>
+BRNN<OutputLayerType, MergeLayerType, MergeOutputType,
+    InitializationRuleType, CustomLayers...>::BRNN(
+    const size_t rho,
+    const bool single,
+    OutputLayerType outputLayer,
+    MergeLayerType mergeLayer,
+    MergeOutputType mergeOutput,
+    InitializationRuleType initializeRule) :
+    rho(rho),
+    outputLayer(std::move(outputLayer)),
+    mergeLayer(new MergeLayerType(mergeLayer)),
+    mergeOutput(new MergeOutputType(mergeOutput)),
+    initializeRule(std::move(initializeRule)),
+    inputSize(0),
+    outputSize(0),
+    targetSize(0),
+    reset(false),
+    single(single),
+    numFunctions(0),
+    deterministic(true),
+    forwardRNN(rho, single, outputLayer, initializeRule),
+    backwardRNN(rho, single, outputLayer, initializeRule)
+{
+  /* Nothing to do here. */
+}
+
+template<typename OutputLayerType, typename MergeLayerType,
+         typename MergeOutputType, typename InitializationRuleType,
+         typename... CustomLayers>
+template<typename OptimizerType>
+void BRNN<OutputLayerType, MergeLayerType, MergeOutputType,
+    InitializationRuleType, CustomLayers...>::Train(
+    arma::cube predictors,
+    arma::cube responses,
+    OptimizerType& optimizer)
+{
+  numFunctions = responses.n_cols;
+
+  this->predictors = std::move(predictors);
+  this->responses = std::move(responses);
+
+  this->deterministic = true;
+  ResetDeterministic();
+
+  if (!reset)
+  {
+    ResetParameters();
+  }
+
+  // Train the model.
+  Timer::Start("BRNN_optimization");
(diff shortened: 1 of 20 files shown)


saksham189:
I am not too sure about the merge operation of the two RNNs. Currently I concatenate the outputs and pass that through `LogSoftmax` layer. But I think that we might need to have a fully connected layer before that for the bidirectional RNN to be useful. 
Should we use FFN to perform the merge operation? @zoq

saksham189:
I can't find any tests specific for bidirectional RNN. I have used one of the tests for RNN to test my implementation.

zoq:
If I remember right Alex Graves used the Reber Grammar test (already implemented) for BLSTM as well.

zoq:
About the merge operation, we could directly merge the results inside the BRNN class; let me know what you think.

> Currently I concatenate the outputs and pass that through LogSoftmax layer

Shouldn't we take the mean over the two?

saksham189:
> Shouldn't we take the mean over the two?

There are multiple ways in which the outputs can be combined ie. summing, average and concatenating. Concatenate is used as the default in most places. Then we might apply a softmax layer or other fully connected layers to produce the final output.

zoq:
Do you think we could introduce another template parameter something like: ``typename MergeLayerType = Concat<>()``?

saksham189:
I would like to do that but the concat layer does a horizontal concatenation as compared to a vertical. 
https://github.com/mlpack/mlpack/blob/master/src/mlpack/methods/ann/layer/concat_impl.hpp#L64

Should I add a parameter to the layer?

saksham189:
Also, I think that we would be required to add a `run` parameter to `Concat` as in `AddMerge`. 
I think in the forward pass of the concat we want `run` to be set to false since `input` to each layer inside concat would be different and set to true when calling Backward. So, we should create a visitor to set value of `run` parameter.

zoq:
> Should I add a parameter to the layer or modify the current implementation? @zoq 

By modify the current implementation you mean the concat layer? Might be the best option here.

> Also, I think that we would be required to add a run parameter to Concat as in AddMerge.

Agreed, do you like to do this as part of the PR?

saksham189:
@zoq this is ready for review. 

> If I remember right Alex Graves used the Reber Grammar test (already implemented) for BLSTM as well.

The paper does not have any results on BLSTM over reber grammar test. But I was able to find that BLSTMs can converge over reber grammar test in 1000 iterations as compared to LSTMs which take 1000 iterations and so, I have implemented a test on that.

Also, I have followed the BRNN algorithm described in Alex Graves paper for implementing the brnn class.
screenshot 2019-01-20 at 7 29 43 pm
screenshot 2019-01-20 at 7 29 49 pm

src/mlpack/methods/ann/visitor/run_set_visitor_impl.hpp
```diff
@@ -0,0 +1,83 @@
+/**
+ * @file run_set_visitor_impl.hpp
+ * @author Marcus Edel
```
zoq: Since you wrote the code, you should adjust the name here :)

src/mlpack/tests/recurrent_network_test.cpp
```diff
@@ -581,6 +653,125 @@ void ReberGrammarTestNetwork(const size_t hiddenSize = 4,
   BOOST_REQUIRE_GE(successes, 1);
 }
 
+/**
+ * Train the specified BRNN network and the construct a Reber grammar dataset.
+ */
+template<typename RecurrentLayerType>
+void ReberGrammarTestNetworkBRNN(const size_t hiddenSize = 4,
```
zoq: Do you think we could pass a model instantiation to make the test code more general? In this case we could we don't have to write a special version for each network. Let me know what you think.

saksham189: Yes that does reduce quite a bit code duplication. Although I had to add one if condition as well since the layers in the model were different.

src/mlpack/methods/ann/brnn_impl.hpp
```diff
@@ -0,0 +1,675 @@
...
+    forwardGradient.zeros();
+    for (size_t l = 0; l < networkSize; ++l)
+    {
+      boost::apply_visitor(LoadOutputParameterVisitor(
+          std::move(forwardRNNOutputParameter)),
+          forwardRNN.network[networkSize - 1 - l]);
+    }
+    boost::apply_visitor(BackwardVisitor(std::move(
+        boost::apply_visitor(outputParameterVisitor, forwardRNN.network.back())),
+        std::move(allDelta[rho - seqNum - 1]), std::move(delta)), mergeLayer);
```
saksham189: I call `Backward()` here on `mergeLayer` passing input from `forwardRNN` and similarly I have to do this while passing the input from `backwardRNN` (because mergeLayer can only take 1 input). This would call backward for each layer inside mergeLayer with the given input. 
Do you think we could add a parameter to `Backward()` inside mergeLayer so that it calls `Backward()` on a specific layer inside the merge layers? @zoq

zoq: Or provide another ``Backward`` overload that takes another input?

saksham189: Yes I have added the overload. Please let me know what you think.

zoq: Looks good to me 👍 ; will take another look at the brrn class tomorrow.

saksham189: okay thanks!

saksham189:
Hey @zoq ,could you do a quick review.

src/mlpack/methods/ann/layer/concat_impl.hpp
```diff
@@ -86,62 +69,70 @@ template<typename eT>
...
     arma::mat delta;
-    if (gy.n_cols == 1)
-    {
-      delta = gy.submat(j, 0, j + elements - 1, 0);
-    }
-    else
+    for (size_t i = 0; i < network.size(); ++i)
     {
-      delta = gy.submat(0, i, elements - 1, i);
+      size_t rows = boost::apply_visitor(outputParameterVisitor, network[i]).n_rows;
```
zoq: Do you think we should add a comment here, to provide some insight what this code does?

src/mlpack/methods/ann/visitor/backward_visitor_impl.hpp
```diff
@@ -21,20 +21,48 @@ namespace ann {
...
 {
   layer->Backward(std::move(input), std::move(error), std::move(delta));
 }
 
+template<typename T>
+inline typename std::enable_if<
+    HasRunCheck<T, bool&(T::*)(void)>::value, void>::type
+BackwardVisitor::LayerBackward(T* layer, arma::mat& /* input */) const
+{
+  if (index  == -1)
```
zoq: We compare against an unsigned integer (``size_t``) so this is never true.

src/mlpack/methods/ann/layer/concat_impl.hpp
saksham189: Yes, would be better.

src/mlpack/methods/ann/visitor/backward_visitor_impl.hpp
saksham189: i have changed to int now

zoq:
@saksham189 Just a note, if you update the PR, I think it's useful that the changes you made are visible, this makes the review a lot easier since the reviewer can see right away what has changed (between updates). Let me know what you think.

saksham189:
> @saksham189 Just a note, if you update the PR, I think it's useful that the changes you made are visible, this makes the review a lot easier since the reviewer can see right away what has changed (between updates). Let me know what you think.

Yup. sorry about the inconvenience. Will push changes in new commits from now on.

zoq:

saksham189:
@zoq this is ready for review now.

src/mlpack/methods/ann/visitor/gradient_visitor.hpp
```diff
@@ -30,7 +30,7 @@ class GradientVisitor : public boost::static_visitor<void>
  public:
   //! Executes the Gradient() method of the given module using the input and
   //! delta parameter.
-  GradientVisitor(arma::mat&& input, arma::mat&& delta);
+  GradientVisitor(arma::mat&& input, arma::mat&& delta, int layer = -1);
```
zoq: Do you think instead of using ``int layer = -1`` we could provide a second constructor:

```c++
GradientVisitor(arma::mat&& input, arma::mat&& delta) : hasIndex(false), index(0)
{
}

GradientVisitor(arma::mat&& input, arma::mat&& delta, const size_t index) : hasIndex(true), index(index)
{
}
```

Let me know what you think.

saksham189: alright sounds good to me

saksham189: @zoq i have made the changes

ShikharJ:
Nice work @saksham189. I'll review in the next few days and then we can merge this in.

src/mlpack/methods/ann/visitor/backward_visitor_impl.hpp
```diff
@@ -24,17 +24,58 @@ inline BackwardVisitor::BackwardVisitor(arma::mat&& input,
...
+  {
+    layer->Backward(std::move(input), std::move(error),
+        std::move(delta));
+  }
+  else
+  {
+    layer->Backward(std::move(input), std::move(error),
+        std::move(delta), index);
+  }
+
```
ShikharJ: Extra newline here.

src/mlpack/methods/ann/visitor/gradient_visitor_impl.hpp
```diff
@@ -34,13 +46,33 @@ inline void GradientVisitor::operator()(LayerType* layer) const
...
+  {
+    layer->Gradient(std::move(input), std::move(delta),
+        std::move(layer->Gradient()));
+  }
+  else
+  {
+    layer->Gradient(std::move(input), std::move(delta),
+        std::move(layer->Gradient()), index);
+  }
+
```
ShikharJ: Same here.

zoq:
One more note, if you do commit something, it would be helpful if the commit message does reflect the changes you did. That makes searching through the commit history a lot easier.

src/mlpack/methods/ann/brnn_impl.hpp
```diff
@@ -0,0 +1,671 @@
...
+    gradient.zeros();
+  }
+
+  if (backwardGradient.is_empty())
+  {
+    backwardGradient = arma::zeros<arma::mat>(
+        parameter.n_rows/2,
+        parameter.n_cols);
+    forwardGradient = arma::zeros<arma::mat>(
+        parameter.n_rows/2,
```
ShikharJ: Can you add a space between the ` \ 2` here and above?

src/mlpack/methods/ann/brnn_impl.hpp
```diff
@@ -0,0 +1,671 @@
...
+void BRNN<OutputLayerType, MergeLayerType, MergeOutputType,
+    InitializationRuleType, CustomLayers...>::ResetDeterministic()
+{
+  forwardRNN.deterministic = this->deterministic;
+  backwardRNN.deterministic = this->deterministic;
+  forwardRNN.ResetDeterministic();
+  backwardRNN.ResetDeterministic();
+}
+
+
```
ShikharJ: Extra newline used here.

saksham189:
> One more note, if you do commit something, it would be helpful if the commit message does reflect the changes you did. That makes searching through the commit history a lot easier.

I will remember to follow this from now onwards. I have changed the previous commit message as well. @zoq

zoq:
Great, @saksham189 do you like to add anything else here, or is this ready?

saksham189:
This is ready from my side. Unless you have anything to add. @zoq

zoq:
Thanks again for the great contribution.

Issue #1625, opened 2018-12-30
DBSCAN isn't using `PointSelectionPolicy`


rcurtin:
It was pointed out in IRC today that the DBSCAN class, which has a template parameter `PointSelectionPolicy`, isn't actually using that template parameter class.  We should fix that.  This would be a good first issue for someone looking to contribute.  Here are the steps to fix the issue...

1. Read and understand what DBSCAN is and how it works.  Wikipedia or the original paper should suffice here, but in either case you should be familiar with DBSCAN.

2. Take a look at the code at the bottom of dbscan_impl.hpp, at line 193:

```
  // Now loop over all points.
  for (size_t i = 0; i < data.n_cols; ++i)
  {
    // Union to all neighbors.
    for (size_t j = 0; j < neighbors[i].size(); ++j)
      uf.Union(i, neighbors[i][j]);
  }
```

That happens after the range search is done, and basically we are building the clusters at this point by using a UnionFind structure.  The original intention of the `PointSelctionPolicy` was that instead of selecting points linearly (i.e. index 0, then 1, then 2, like the for loop above does), that another strategy could be used.  We should aim to replace that code with this:

```
  // Now loop over all points.
  for (size_t i = 0; i < data.n_cols; ++i)
  {
    // Get the next index.
    const size_t index = pointSelector.Select(i, data);
    for (size_t j = 0; j < neighbors[index].size(); ++j)
      uf.Union(index, neighbors[index][j]);
  }
```

3.  Take a look at `random_point_selection.hpp`.  You'll note that its signature does not match what I just proposed above, so it will need to be adapted.  Additionally, the class will need to internally hold the list of which points have been selected and which haven't, so that when `Select()` is called it returns a random index that has not yet been selected.

4.  Implemented `ordered_point_selection.hpp`, which when `Select(i, data)` is called just returns `i`---so this will imitate the functionality from before.

5.  Set the default `PointSelectionPolicy` to `OrderedPointSelection` from the previous step, so that it still behaves the same as previous versions of mlpack.

6.  Modify `dbscan_main.cpp` to add an option like `PARAM_FLAG("random_selection", ...)` that will control whether or not `OrderedPointSelection` or `RandomPointSelection` is used.

7.  Modify `HISTORY.md` to point out this change.

8.  Add a test for DBSCAN using `OrderedPointSelection` to `src/mlpack/tests/dbscan_test.cpp`.  Also make sure that `RandomPointSelection` is properly tested in that file.

[...]

PR #1627, opened 2019-01-02
Fixed DBSCAN isn't using PointSelectionPolicy issue #1625

KimSangYeon-DGU:
Hi, I fixed the issue #1625 .
Could you review this code?

### DIFF
--- a/HISTORY.md
+++ b/HISTORY.md
@@ -1,3 +1,7 @@
+### mlpack 3.0.5
+###### 2019-01-02
+  * Fix DBSCAN isn't using PointSelectionPolicy issue (#1625).
+  
 ### mlpack 3.0.4
 ###### 2018-11-13
   * Bump minimum CMake version to 3.3.2.
--- a/src/mlpack/methods/dbscan/dbscan_impl.hpp
+++ b/src/mlpack/methods/dbscan/dbscan_impl.hpp
@@ -192,9 +192,10 @@ void DBSCAN<RangeSearchType, PointSelectionPolicy>::BatchCluster(
   // Now loop over all points.
   for (size_t i = 0; i < data.n_cols; ++i)
   {
-    // Union to all neighbors.
-    for (size_t j = 0; j < neighbors[i].size(); ++j)
-      uf.Union(i, neighbors[i][j]);
+    // Get the next index.
+    const size_t index = pointSelector.Select(i, data);
+    for (size_t j = 0; j < neighbors[index].size(); ++j)
+      uf.Union(index, neighbors[index][j]);
   }
 }
 
--- a/src/mlpack/methods/dbscan/dbscan_main.cpp
+++ b/src/mlpack/methods/dbscan/dbscan_main.cpp
@@ -15,7 +15,7 @@
 #include <mlpack/core/tree/binary_space_tree.hpp>
 #include <mlpack/core/tree/rectangle_tree.hpp>
 #include <mlpack/core/tree/cover_tree.hpp>
-
+#include <mlpack/methods/dbscan/random_point_selection.hpp>
 #include "dbscan.hpp"
 
 using namespace mlpack;
@@ -76,6 +76,8 @@ PARAM_FLAG("single_mode", "If set, single-tree range search (not dual-tree) "
     "will be used.", "S");
 PARAM_FLAG("naive", "If set, brute-force range search (not tree-based) "
     "will be used.", "N");
+PARAM_FLAG("random_selection", "If set, random point selection (not ordered) "
+    "will be used.", "R");
 
 // Actually run the clustering, and process the output.
 template<typename RangeSearchType>
@@ -86,13 +88,26 @@ void RunDBSCAN(RangeSearchType rs = RangeSearchType())
 
   // Load dataset.
   arma::mat dataset = std::move(CLI::GetParam<arma::mat>("input"));
-
+  
   const double epsilon = CLI::GetParam<double>("epsilon");
   const size_t minSize = (size_t) CLI::GetParam<int>("min_size");
 
-  DBSCAN<RangeSearchType> d(epsilon, minSize, !CLI::HasParam("single_mode"),
-      rs);
+  
+  // Check if random selection is used.
+  if (!CLI::HasParam("random_selection"))
+  {
+    DBSCAN<RangeSearchType> d(epsilon, minSize, !CLI::HasParam("single_mode"),
+        rs);   
+  }
+  else
+  {
+    DBSCAN<RangeSearchType, RandomPointSelection> d(epsilon, minSize,
+        !CLI::HasParam("single_mode"), rs);
+  }
 
+  DBSCAN<RangeSearchType, RandomPointSelection> d(epsilon, minSize,
+        !CLI::HasParam("single_mode"), rs);
+  
   // If possible, avoid the overhead of calculating centroids.
   arma::Row<size_t> assignments;
   if (CLI::HasParam("centroids"))
--- a/src/mlpack/methods/dbscan/ordered_point_selection.hpp
+++ b/src/mlpack/methods/dbscan/ordered_point_selection.hpp
@@ -1,0 +1,44 @@
+/**
+ * @file ordered_point_selection.hpp
+ * @author Ryan Curtin
+ *
+ * sequentially select the next point for DBSCAN.
+ *
+ * mlpack is free software; you may redistribute it and/or modify it under the
+ * terms of the 3-clause BSD license.  You should have received a copy of the
+ * 3-clause BSD license along with mlpack.  If not, see
+ * http://www.opensource.org/licenses/BSD-3-Clause for more information.
+ */
+#ifndef MLPACK_METHODS_DBSCAN_ORDERED_POINT_SELECTION_HPP
+#define MLPACK_METHODS_DBSCAN_ORDERED_POINT_SELECTION_HPP
+
+#include <mlpack/prereqs.hpp>
+#include <boost/dynamic_bitset.hpp>
+
+namespace mlpack {
+namespace dbscan {
+
+/**
+ * This class can be used to sequentially select the next point to use for DBSCAN.
+ */
+class OrderedPointSelection
+{
+ public:
+  /**
+   * Select the next point to use, sequentially.
+   *
+   * @param unvisited Bitset indicating which points are unvisited.
+   * @param data Unused data.
+   */
+  template<typename MatType>
+  static size_t Select(const size_t point,
+                       const MatType& /* data */)
+  { 
+    return point; // Just return point.
+  }
+};
+
+} // namespace dbscan
+} // namespace mlpack
+
+#endif
--- a/src/mlpack/methods/dbscan/random_point_selection.hpp
+++ b/src/mlpack/methods/dbscan/random_point_selection.hpp
@@ -27,16 +27,24 @@ class RandomPointSelection
   /**
    * Select the next point to use, randomly.
    *
-   * @param unvisited Bitset indicating which points are unvisited.
-   * @param data Unused data.
+   * @param point Unused data.
+   * @param data Dataset to cluster.
    */
   template<typename MatType>
-  static size_t Select(const boost::dynamic_bitset<>& unvisited,
-                       const MatType& /* data */)
+  size_t Select(const size_t /* point */,
+                       const MatType& data /* data */)
   {
+    // Initialize the length of the unvisited bitset
+    size_t size = data.n_cols; // Get the size of points.
+    if (unvisited.size() != size) 
+    {
+      unvisited.resize(size); // Resize & Set bitset to one
+      unvisited.set();
+    }
+
     const size_t max = unvisited.count();
     const size_t index = math::RandInt(max);
-
+    
     // Select the index'th unvisited point.
     size_t found = 0;
     for (size_t i = 0; i < unvisited.size(); ++i)
@@ -45,13 +53,19 @@ class RandomPointSelection
         ++found;
 
       if (found > index)
+      {
+        unvisited[i].flip(); // Set unvisited point to visited point
         return i;
+      }
     }
-
     return 0; // Not sure if it is possible to get here.
   }
-};
 
+ private:
+  // Bitmask for unvisited points, If true, mean unvisited.
+  boost::dynamic_bitset<> unvisited;
+};
+  
 } // namespace dbscan
 } // namespace mlpack
 
--- a/src/mlpack/tests/dbscan_test.cpp
+++ b/src/mlpack/tests/dbscan_test.cpp
@@ -11,11 +11,13 @@
  */
 #include <mlpack/core.hpp>
 #include <mlpack/methods/dbscan/dbscan.hpp>
+#include <mlpack/methods/dbscan/random_point_selection.hpp>
 
 #include <boost/test/unit_test.hpp>
 #include "test_tools.hpp"
 
 using namespace mlpack;
+using namespace mlpack::range;
 using namespace mlpack::dbscan;
 using namespace mlpack::distribution;
 
@@ -269,6 +271,42 @@ BOOST_AUTO_TEST_CASE(GaussiansSingleModeTest)
     BOOST_REQUIRE_NE(assignments(i), matches(0));
     BOOST_REQUIRE_NE(assignments(i), matches(1));
   }
+}
+
+/**
+ * Check that OrderedPointSelection works correctly.
+ */
+BOOST_AUTO_TEST_CASE(OrderedPointSelectionTest)
+{
+  arma::mat points(10, 200, arma::fill::randu);
+
+  DBSCAN<> d(2.0, 2);
+
+  arma::Row<size_t> assignments;
+  const size_t clusters = d.Cluster(points, assignments);
+
+  BOOST_REQUIRE_EQUAL(clusters, 1);
+
+  // The number of assignments returned should be the same as points's
+  BOOST_REQUIRE_EQUAL(assignments.n_elem, points.n_cols);
+}
+
+/**
+ * Check that RandomPointSelection works correctly.
+ */
+BOOST_AUTO_TEST_CASE(RandomPointSelectionTest)
+{
+  arma::mat points(10, 200, arma::fill::randu);
+  
+  DBSCAN<RangeSearch<>, RandomPointSelection> d(2.0, 2);
+
+  arma::Row<size_t> assignments;
+  const size_t clusters = d.Cluster(points, assignments);
+
+  BOOST_REQUIRE_EQUAL(clusters, 1);
+
+  // The number of assignments returned should be the same as points's
+  BOOST_REQUIRE_EQUAL(assignments.n_elem, points.n_cols);
 }
 
 BOOST_AUTO_TEST_SUITE_END();
--- a/src/mlpack/tests/main_tests/dbscan_test.cpp
+++ b/src/mlpack/tests/main_tests/dbscan_test.cpp
@@ -429,4 +429,27 @@ BOOST_AUTO_TEST_CASE(DBSCANNaiveSearchTest)
   CheckMatrices(output, naiveOutput);
 }
 
+/**
+ * Check that the random selection is same 
+ */
+BOOST_AUTO_TEST_CASE(DBSCANRandomSelectionFlagTest)
+{
+  arma::mat inputData;
+  if (!data::Load("iris.csv", inputData))
+    BOOST_FAIL("Unable to load dataset iris.csv!");
+
+  size_t inputSize = inputData.n_cols;
+
+  SetInputParam("input", inputData);
+  SetInputParam("random_selection", true);
+
+  mlpackMain();
+
+  // Check that number of predicted labels is equal to the input test points.
+  BOOST_REQUIRE_EQUAL(CLI::GetParam<arma::Row<size_t>>("assignments").n_cols,
+                      inputSize);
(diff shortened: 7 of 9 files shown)


HISTORY.md
```diff
@@ -1,3 +1,7 @@
+### mlpack 3.0.5
+###### 2019-01-02
+  * Fix DBSCAN isn't using PointSelectionPolicy issue (#1625).
```
rcurtin: I know this is a minor comment, but can we make this say something more active, like `Change DBSCAN to use PointSelectionPolicy and add RandomSelectionPolicy (#1625)`.  Also we can change the date to `????-??-??`, since I am not sure when the release will be. :)

src/mlpack/methods/dbscan/random_point_selection.hpp
```diff
@@ -45,14 +53,20 @@ class RandomPointSelection
...
+ private:
+  // Bitmask for unvisited points, If true, mean unvisited.
+  boost::dynamic_bitset<> unvisited;
+};
+  
 } // namespace dbscan
 } // namespace mlpack
 
-#endif
+#endif
```
rcurtin: This is a really tiny issue, but can you add the newline back to the end of the file please?  Otherwise `cat random_point_selection.hpp` will give weird results when it prints the prompt right after the `#endif`. :)

src/mlpack/tests/main_tests/dbscan_test.cpp
```diff
@@ -429,4 +429,27 @@ BOOST_AUTO_TEST_CASE(DBSCANNaiveSearchTest)
   CheckMatrices(output, naiveOutput);
 }
 
+/**
+ * Check that the random selection is same
```
rcurtin: Actually I think we should make this test a little bit more challenging.  Here the goal should be to ensure that `random_selection` affects the results.  So we should find a dataset that, when we set `random_selection = false`, the results are different than when we set `random_selection = true` (for the vast majority of random seeds).  Maybe you could make this as a synthetic dataset, or see if different parameters of `epsilon` or `minPoints` for the iris dataset can work.

src/mlpack/methods/dbscan/ordered_point_selection.hpp
```diff
@@ -0,0 +1,44 @@
+/**
+ * @file ordered_point_selection.hpp
+ * @author Ryan Curtin
```
rcurtin: I'm not the author here; feel free to add your name. :)

src/mlpack/methods/dbscan/ordered_point_selection.hpp
```diff
@@ -0,0 +1,44 @@
...
+ * mlpack is free software; you may redistribute it and/or modify it under the
+ * terms of the 3-clause BSD license.  You should have received a copy of the
+ * 3-clause BSD license along with mlpack.  If not, see
+ * http://www.opensource.org/licenses/BSD-3-Clause for more information.
+ */
+#ifndef MLPACK_METHODS_DBSCAN_ORDERED_POINT_SELECTION_HPP
+#define MLPACK_METHODS_DBSCAN_ORDERED_POINT_SELECTION_HPP
+
+#include <mlpack/prereqs.hpp>
+#include <boost/dynamic_bitset.hpp>
```
rcurtin: `dynamic_bitset.hpp` probably isn't needed here.

src/mlpack/methods/dbscan/random_point_selection.hpp
```diff
@@ -45,14 +53,20 @@ class RandomPointSelection
...
+      }
     }
-
     return 0; // Not sure if it is possible to get here.
   }
-};
 
+ private:
+  // Bitmask for unvisited points, If true, mean unvisited.
+  boost::dynamic_bitset<> unvisited;
```
rcurtin: I don't remember why I used `dynamic_bitset<>`, but you can probably change this with `std::vector<bool>` to avoid all the overhead of including boost parts (that can actually be a significant compiler slowdown since Boost is so large).

KimSangYeon-DGU: Yes, I will do it!!

src/mlpack/tests/main_tests/dbscan_test.cpp
KimSangYeon-DGU: In fact, I needed a lot of advice on this part. Okay. I'll try to figure it out.

rcurtin: Give it a shot and let me know if I can help out.  This may be a little tricky to get right.

HISTORY.md
rcurtin: The style you used is great overall, so thank you for adhering to the style guidelines. :)

src/mlpack/tests/main_tests/dbscan_test.cpp
KimSangYeon-DGU: Yes, I'm currently working on this part. Thanks for the consideration!! :+1:

KimSangYeon-DGU: Hi, @rcurtin . I found the optimal parameter for maximizing difference between two using iris dataset. I used 'Grid Search algorithm' to find the optimal parameter, epsilon and min_size. 
The range of values I checked is as follows.
Epsilon: 0.001 ~ 2 (interval: 0.001)
Min_size: 1 ~ 10 (interval: 1)
So I examined a total of 20,000 parameters, the maximum difference is 0.98 and is when epsilon is 0.358 and min_size is 1. And I think the A shows the same result always in iris dataset that I used. Thus, I will try to write a test code comparing A's result to B's. I want to ask you if this is a good approach. I'll contact you again when I'm done. 
Thanks. :)

rcurtin: Yes, definitely `OrderedPointSelection` will show the same result in most places.  If you are finding that even with different random seeds, the results are always different, then that will work for the test.  Note that for anything in `main_tests/` (or more specifically where `BINDING_TYPE` is defined as `BINDING_TYPE_TEST`), the `math::RandomSeed()` function does nothing.  (This is because many mlpack bindings have an option for the random seed, and we don't want to set the random seed over and over again in the test for reproducibility.)

You can set the random seed by doing this:

```
    const size_t seed = std::time(NULL); // or some specific seed if you want
    randGen.seed((uint32_t) seed);
    srand((unsigned int) seed);
    arma::arma_rng::set_seed(seed);
```

That's just an adaptation of `math::RandomSeed()`.

KimSangYeon-DGU: Oh, Thank you. I will test it.

KimSangYeon-DGU:
Hi, @rcurtin. I updated the previous codes.
Could you please review this??

KimSangYeon-DGU:
Hi, @rcurtin. I updated the previous codes according to the feedback.
Could you please review this??

src/mlpack/methods/dbscan/dbscan_main.cpp
```diff
@@ -76,6 +76,8 @@ PARAM_FLAG("single_mode", "If set, single-tree range search (not dual-tree) "
     "will be used.", "S");
 PARAM_FLAG("naive", "If set, brute-force range search (not tree-based) "
     "will be used.", "N");
+PARAM_FLAG("random_selection", "If set, random point selection (not ordered) "
```
zoq: Should we allow the user to select the selection policy, something like: https://github.com/mlpack/mlpack/blob/376c17e11d2797e258b9c1c34aa393d953eccd22/src/mlpack/methods/kernel_pca/kernel_pca_main.cpp#L112-L113

So if we add another strategy we don't have to update the cmd API?

src/mlpack/methods/dbscan/dbscan_main.cpp
```diff
@@ -86,26 +88,49 @@ void RunDBSCAN(RangeSearchType rs = RangeSearchType())
 
   // Load dataset.
   arma::mat dataset = std::move(CLI::GetParam<arma::mat>("input"));
-
+
```
zoq: Looks like we can remove the extra whitespace here.

src/mlpack/methods/dbscan/ordered_point_selection.hpp
```diff
@@ -0,0 +1,43 @@
+/**
+ * @file ordered_point_selection.hpp
+ * @author Kim Sang-Yeon
+ *
+ * sequentially select the next point for DBSCAN.
```
zoq: Pedantic style issue; this should start with a capital letter.

src/mlpack/methods/dbscan/ordered_point_selection.hpp
```diff
@@ -0,0 +1,43 @@
...
+{
+ public:
+  /**
+   * Select the next point to use, sequentially.
+   *
+   * @param unvisited Bitset indicating which points are unvisited.
+   * @param data Unused data.
+   */
+  template<typename MatType>
+  size_t Select(const size_t point,
```
zoq: Should we use a static method here, to make this more readable/accessible  ``PointSelectionPolicy::Select(...)``?

src/mlpack/methods/dbscan/random_point_selection.hpp
```diff
@@ -27,16 +27,24 @@ class RandomPointSelection
...
-   * @param unvisited Bitset indicating which points are unvisited.
-   * @param data Unused data.
+   * @param point Unused data.
+   * @param data Dataset to cluster.
    */
   template<typename MatType>
-  static size_t Select(const boost::dynamic_bitset<>& unvisited,
-                       const MatType& /* data */)
+  size_t Select(const size_t /* point */,
+                       const MatType& data /* data */)
```
zoq: Another picky style comment, the second parameter should align with the first one. Also no need to write ``/* data */`` after the parameter, the comment above should be enough.

src/mlpack/methods/dbscan/random_point_selection.hpp
```diff
@@ -27,16 +27,24 @@ class RandomPointSelection
...
+    if (unvisited.size() != size) 
+    {
+      unvisited.resize(size, true); // Resize & Set bitset to one
+    }
+    
+    // Count the unvisited points and generate n-th index randomly
+    const size_t max = std::count(unvisited.begin(), unvisited.end(), true);
     const size_t index = math::RandInt(max);
-
+
```
zoq: Looks like we can remove the extra whitespace here as well.

src/mlpack/methods/dbscan/random_point_selection.hpp
```diff
@@ -27,16 +27,24 @@ class RandomPointSelection
...
-                       const MatType& /* data */)
+  size_t Select(const size_t /* point */,
+                       const MatType& data /* data */)
   {
-    const size_t max = unvisited.count();
+    // Initialize the length of the unvisited bitset
+    size_t size = data.n_cols; // Get the size of points.
+    if (unvisited.size() != size) 
+    {
+      unvisited.resize(size, true); // Resize & Set bitset to one
```
zoq: Missing point at the end of the comment, line 37 as well.

src/mlpack/methods/dbscan/dbscan_main.cpp
KimSangYeon-DGU: Hi, @zoq.
**Q1) Should we allow the user to select the selection policy**
-> In my opinion, currently, there are only two selection policies in DBSCAN, so it is okay to use a binary flag format, but if a policy is added in the future, it would be better to change the parameter flag format like `kernel_pca_main.cpp`.

**Q2) So if we add another strategy we don't have to update the cmd API?**
-> Because I edited the `dbscan_main.cpp` assuming there are only two possible selection policies, If another selection policy is added in the future, in addition to the parameter flag for user to choose the added selection policy, we should update the `dbscan_main` code.

src/mlpack/methods/dbscan/dbscan_main.cpp
KimSangYeon-DGU: Oh, Thank you!!

src/mlpack/methods/dbscan/ordered_point_selection.hpp
KimSangYeon-DGU: Yes, I will do!!

src/mlpack/methods/dbscan/ordered_point_selection.hpp
KimSangYeon-DGU: Okay, I will do.

src/mlpack/methods/dbscan/random_point_selection.hpp
KimSangYeon-DGU: Yes!! I will edit it

src/mlpack/methods/dbscan/random_point_selection.hpp
KimSangYeon-DGU: Yes, I will do.

src/mlpack/methods/dbscan/random_point_selection.hpp
KimSangYeon-DGU: Thank you for pointing that out!!

src/mlpack/methods/dbscan/dbscan_main.cpp
rcurtin: I don't have much of an opinion here either way.  The string option means we don't need to break any reverse compatibility when adding new options, though, which is nice.  Whatever you and @zoq choose to do is fine with me. :+1:

KimSangYeon-DGU:
First of all, I'm really sorry, I will be careful not to notify you twice in the future.
I edited it according to feedback. Thank you all so much for reviewing it. :)

HISTORY.md
```diff
@@ -1,3 +1,7 @@
+### mlpack 3.0.5
+###### ????-??-??
+  * Change DBSCAN to use PointSelectionPolicy and add RandomSelectionPolicy (#1625).
```
ShikharJ: If I'm not wrong, haven't you implemented `OrderedPointSelection` policy instead of `RandomSelectionPolicy` which was already implemented?

src/mlpack/methods/dbscan/dbscan_impl.hpp
```diff
@@ -192,13 +192,14 @@ void DBSCAN<RangeSearchType, PointSelectionPolicy>::BatchCluster(
...
+    for (size_t j = 0; j < neighbors[index].size(); ++j)
+      uf.Union(index, neighbors[index][j]);
   }
 }
 
 } // namespace dbscan
 } // namespace mlpack
 
-#endif
+#endif
```
ShikharJ: Please keep the newline here.

src/mlpack/methods/dbscan/dbscan_main.cpp
```diff
@@ -86,26 +88,48 @@ void RunDBSCAN(RangeSearchType rs = RangeSearchType())
...
+      d.Cluster(dataset, assignments);
+    }
   }
-  else
+  else // Ordered Selection
   {
-    d.Cluster(dataset, assignments);
+    DBSCAN<RangeSearchType> d(epsilon, minSize,
+        !CLI::HasParam("single_mode"), rs);
+
```
ShikharJ: Since only the call `DBSCAN<RangeSearchType, ...>` has to be different, I think we can save a bit on code duplication here.

src/mlpack/methods/dbscan/dbscan_main.cpp
```diff
@@ -86,26 +88,48 @@ void RunDBSCAN(RangeSearchType rs = RangeSearchType())
...
+      CLI::GetParam<arma::mat>("centroids") = std::move(centroids);
+    }
+    else
+    {
+      d.Cluster(dataset, assignments);
+    }
   }
-  else
+  else // Ordered Selection
```
ShikharJ: Please remove this comment from here. Also, if possible try to accomodate the comments on the line above the code, to be consistent with rest of the codebase.

src/mlpack/methods/dbscan/random_point_selection.hpp
```diff
@@ -27,14 +26,22 @@ class RandomPointSelection
...
   {
-    const size_t max = unvisited.count();
+    // Initialize the length of the unvisited bitset.
+    size_t size = data.n_cols; // Get the size of points.
+    if (unvisited.size() != size) 
+    {
+      unvisited.resize(size, true); // Resize & Set bitset to one.
+    }
+
+    // Count the unvisited points and generate n-th index randomly.
```
ShikharJ: Just use `nth`

src/mlpack/tests/main_tests/dbscan_test.cpp
```diff
@@ -429,4 +429,39 @@ BOOST_AUTO_TEST_CASE(DBSCANNaiveSearchTest)
   CheckMatrices(output, naiveOutput);
 }
 
+/**
+ * Check that the assignment of cluster is different if
+ * different point selection policies.
```
ShikharJ: Change `different point selection policies` to `point selection policies are different`.

src/mlpack/tests/main_tests/dbscan_test.cpp
```diff
@@ -429,4 +429,39 @@ BOOST_AUTO_TEST_CASE(DBSCANNaiveSearchTest)
...
+ */
+BOOST_AUTO_TEST_CASE(DBSCANRandomSelectionFlagTest)
+{
+  arma::mat inputData;
+  if (!data::Load("iris.csv", inputData))
+    BOOST_FAIL("Unable to load dataset iris.csv!");
```
ShikharJ: Please add a set of braces here for aesthetics.

src/mlpack/tests/dbscan_test.cpp
```diff
@@ -271,4 +273,40 @@ BOOST_AUTO_TEST_CASE(GaussiansSingleModeTest)
...
+  arma::mat points(10, 200, arma::fill::randu);
+
+  DBSCAN<> d(2.0, 2);
+
+  arma::Row<size_t> assignments;
+  const size_t clusters = d.Cluster(points, assignments);
+
+  BOOST_REQUIRE_EQUAL(clusters, 1);
+
+  // The number of assignments returned should be the same as points's
```
ShikharJ: Just change `points's` to `points`.

src/mlpack/tests/dbscan_test.cpp
```diff
@@ -271,4 +273,40 @@ BOOST_AUTO_TEST_CASE(GaussiansSingleModeTest)
...
+  arma::mat points(10, 200, arma::fill::randu);
+  
+  DBSCAN<RangeSearch<>, RandomPointSelection> d(2.0, 2);
+
+  arma::Row<size_t> assignments;
+  const size_t clusters = d.Cluster(points, assignments);
+
+  BOOST_REQUIRE_EQUAL(clusters, 1);
+
+  // The number of assignments returned should be the same as points's
```
ShikharJ: Same here.

HISTORY.md
KimSangYeon-DGU: Hi there, thank you for reviewing it.
Ah!, you're right. The `OrderedPointSelectionPolicy` is newly implemented, and the `RandomSelectionPolicy` is edited for use because it is not used before. The existing `src/mlpack/methods/dbscan/dbscan_impl.hpp` have selected points sequentially. This is why I selected the `RandonSelectionPolicy`, not `OrderedSelectionPolicy`.

But, after hearing what you said, It is better to change `RandomSelectionPolicy` to `OrderedSelectionPolicy` because It is actually new. Thanks!!

src/mlpack/methods/dbscan/dbscan_main.cpp
KimSangYeon-DGU: I totally agree. I will make new commit.
Could you please review it later??

ShikharJ: On a second thought, I feel code deduplication will have to come at a cost of using the assignment operation, and in that case it's better to leave it as it is, if only two policies ae going to require the code below. @rcurtin Is there a possibility that another policy might require the same piece of code as below? I'm not sure how many different selection policies are developed for DBSCAN. If they are indeed numerous, we can implement the below as a function.

ShikharJ: @KimSangYeon-DGU I don't think you ever need to request for reviews. For every commit you push or comment you make, we get notification emails since we're watching this repository. And we do check our mail :) Only in case where you PR has been left unchecked for let's say more than 3 days, you should hit one of us up.

KimSangYeon-DGU: > On a second thought, I feel code deduplication will have to come at a cost of using the assignment operation, and in that case it's better to leave it as it is, if only two policies are going to require the code below. @rcurtin Is there a possibility that another policy might require the same piece of code as below? I'm not sure how many different selection policies are developed for DBSCAN. If they are indeed numerous, we can implement the below as a function.

First of all, I've modified to reflect your feedback for extensibility.

ShikharJ: I see, I like the current changes, they're minimal.

src/mlpack/methods/dbscan/dbscan_main.cpp
zoq: I would go with the kernal pca strategy, but if @KimSangYeon-DGU likes the current solution, that fine with me as well.

KimSangYeon-DGU: I will try it. :)

KimSangYeon-DGU: @zoq I changed the code like `mlpack/src/mlpack/methods/kernel_pca/kernel_pca_main.cpp`.

PR #1628, opened 2019-01-02
Feature-dbscan point selection

Yugandhartripathi:
Fix for issue #1625 .
Please review the code.

### DIFF
--- a/src/mlpack/methods/dbscan/random_point_selection.hpp
+++ b/src/mlpack/methods/dbscan/random_point_selection.hpp
@@ -58,7 +58,7 @@ class RandomPointSelection
     return 0; // Not sure if it is possible to get here.
   }
  private:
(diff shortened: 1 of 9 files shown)


src/mlpack/methods/dbscan/random_point_selection.hpp
```diff
@@ -45,11 +50,15 @@ class RandomPointSelection
         ++found;
 
       if (found > index)
+      {
+        unvisited[i]=false;
```
rcurtin: Here we should be careful with the style: `unvisited[i] = false`.

Yugandhartripathi: @rcurtin Thanks for taking the time and going through my code.
I will go through the style and contributions guide and take better care of this next time :+1: .

Issue #1611, opened 2018-12-23
Issues with MLPACK on Windows using Mingw


lhedjazi:
Dear Mlpack community,

I am struggling since few days to compile Mlpack in the aim to use it in one of my projects but I am facing some issues according to CMake building configs. I tried everything possible and looked on the past issues submitted by Mlpack users. One of the issues is similar to the one we can find here : https://github.com/mlpack/mlpack/issues/813 

I am using:
- Mingw version: g++ (x86_64-posix-seh-rev0, Built by MinGW-W64 project) 8.1.0
-  CMake 3.13.1 version. 
- Mlpack version 3.0.4 with Lapack and Blas enabled in the config file (OpenBlas 0.2.14)
- Armadillo version 8.500.1
- Boost version 1.66.0

I have tried serveral opitons but faced different issues:

**Option 1**:  When I select to build both "CLI_EXECUTABLES" and "SHARED_LIBS"  in the CMake GUI it fails with the following error:

[ 25%] Linking CXX executable ..\..\..\..\mlpack_adaboost.exe
......
CMakeFiles\mlpack_adaboost.dir/objects.a(adaboost_main.cpp.obj):adaboost_main.cpp:(.text$_ZN6mlpack8bindings3cli8GetParamIPNS_8adaboost13AdaBoostModelEEEvRKNS_4util9ParamDataEPKvPv[_ZN6mlpack8bindings3cli8GetParamIPNS_8adaboost13AdaBoostModelEEEvRKNS_4util9ParamDataEPKvPv]+0x51): undefined reference to `mlpack::adaboost::AdaBoostModel::AdaBoostModel()'
collect2.exe: error: ld returned 1 exit status
mingw32-make.exe[2]: *** [src\mlpack\methods\adaboost\CMakeFiles\mlpack_adaboost.dir\build.make:94: mlpack_adaboost.exe] Error 1
mingw32-make.exe[1]: *** [CMakeFiles\Makefile2:1501: src/mlpack/methods/adaboost/CMakeFiles/mlpack_adaboost.dir/all] Error 2
mingw32-make.exe: *** [Makefile:140: all] Error 2

**Option 2**: When I select to build each of them seprately it works but I face one issue when I try to link to the so generated libraries in my c++ project:

[...]

PR #1629, opened 2019-01-02
CMake installation fixes

rcurtin:
This should solve #1621.  I had to refactor the CMake installation scripts for Python a bit to make it work right.

I also set BUILD_SHARED_LIBS=OFF by default on Windows as suggested in #1611.


rcurtin:
:+1: thanks!

PR #1632, opened 2019-01-06
Auto Test Binding for Range Search

niteya-shah:
Includes
Declaration of namespace
Added Basic Structure to test

I was writing a test for checking Min and Max values , and there seems to be no check to ensure that min<max . Executing it gave no error but all outputs came out to be null (Kind of expected), is this intentional or should I change this in the range_test_main.cpp file ?

### DIFF
--- a/COPYRIGHT.txt
+++ b/COPYRIGHT.txt
@@ -108,6 +108,7 @@ Copyright:
   Copyright 2018, Ayush Chamoli
   Copyright 2018, Tommi Laivamaa <tommi.laivamaa@protonmail.com>
   Copyright 2019, Kim SangYeon <sy0814k@gmail.com>
+  Copyright 2019, Niteya Shah <niteya.56@gmail.com>
   
 License: BSD-3-clause
   All rights reserved.
--- a/src/mlpack/core.hpp
+++ b/src/mlpack/core.hpp
@@ -254,6 +254,7 @@
  *   - Ayush Chamoli
  *   - Tommi Laivamaa <tommi.laivamaa@protonmail.com>
  *   - Kim SangYeon <sy0814k@gmail.com>
+ *   - Niteya Shah <niteya.56@gmail.com>
  */
 
 // First, include all of the prerequisites.


rcurtin:
I'd say if the user has specified a range where max > min, this is still technically a valid range search even though the result set will always be empty.  I wouldn't mind having some warning output for that case, but I don't see a reason to issue any fatal error.

As you work on this, I'll get emails when you push new commits, but I won't know when you're done, so please just post a comment when it's ready for review.  Thanks for taking on the task! :+1:

niteya-shah:
The Appveyor build failure seems to be because of a bug https://stackoverflow.com/questions/33605632/error-83-error-c2398-conversion-from-double-to-float-requires-a-narrowing-c

src/mlpack/tests/main_tests/range_search_test.cpp
```diff
@@ -0,0 +1,292 @@
...
+  cout<<"Leaf Value Tests"<<endl;
+  //Testing 3 different leaf values - default 20, 15 and 25
+  //Ensure that results match for different leaf values
+  arma::mat x={{0, 3, 3, 4, 3, 1},{4, 4, 4, 5, 5, 2},{0, 1, 2, 2, 3, 3}};
+  std::string distance_file = "distances.csv";
+  std::string neighbors_file = "neighbors.csv";
+  double min_v = 0, max_v = 3;
+  vector<std::vector<int>> neighbor_val = {{}, {2, 3, 4},{1, 3, 4, 5},{1, 2, 4},
+                                         {1, 2, 3},{2}};
+  vector<std::vector<float>> distance_val = {{},{1, 1.73205, 2.23607},
```
rcurtin: Why are you using `vector<vector<float>>` here and not `vector<vector<double>>` like `RangeSearch` returns?  This should fix your AppVeyor problem if you use `vector<vector<double>>`.

src/mlpack/methods/range_search/rs_model_impl.hpp
```diff
@@ -271,7 +288,7 @@ void MonoSearchVisitor::operator()(RSType* rs) const
 }
 
 //! Save parameters for bichromatic range search.
-BiSearchVisitor::BiSearchVisitor(const arma::mat& querySet,
+inline BiSearchVisitor::BiSearchVisitor(const arma::mat& querySet,
```
rcurtin: Thanks for these fixes.  Can you adjust the spacing for the other parameters to these methods to make them match where the `const arma::mat& querySet` parameter starts also please?

src/mlpack/tests/main_tests/range_search_test.cpp
```diff
@@ -0,0 +1,292 @@
...
+};
+
+BOOST_FIXTURE_TEST_SUITE(RangeSearchMainTest, RangeSearchTestFixture);
+/*
+* Check that the correct output is returned for a small synthetic input
+* case
+*/
+BOOST_AUTO_TEST_CASE(SyntheticRangeSearch)
+{
+  cout<<"Synthetic Test 1"<<endl;
```
rcurtin: Can you remove this debugging line please?

src/mlpack/tests/main_tests/range_search_test.cpp
```diff
@@ -0,0 +1,292 @@
...
+  //Matrix Input is expected in this format
+  arma::mat x={{0, 3, 3, 4, 3, 1},{4, 4, 4, 5, 5, 2},{0, 1, 2, 2, 3, 3}};
```
rcurtin: For formatting we should follow the style guide:

```
arma::mat x = {{0, 3, 3, 4, 3, 1},
               {4, 4, 4, 5, 5, 2},
               {0, 1, 2, 2, 3, 3}};
```

I did that because I think this will be over 80 columns if we put it all on one line.  Note the spaces around the `=`.  If you can take care of that all through the code it would be great.  The style guide is here: https://github.com/mlpack/mlpack/wiki/DesignGuidelines#style-guidelines

src/mlpack/tests/main_tests/range_search_test.cpp
```diff
@@ -0,0 +1,292 @@
...
+  std::string distance_file="distances.csv";
+  std::string neighbors_file="neighbors.csv";
+  double min_v=0, max_v=3;
+  vector<std::vector<size_t>> neighbor_val={{},{2, 3, 4},{1, 3, 4, 5},{1, 2, 4},
```
rcurtin: Why `vector<std::vector<size_t>>` and not the simpler `vector<vector<size_t>>`?  You're already `using namespace std` so it would make sense to make the type name shorter.

src/mlpack/tests/main_tests/range_search_test.cpp
```diff
@@ -0,0 +1,292 @@
...
+* case
+*/
+BOOST_AUTO_TEST_CASE(SyntheticRangeSearch)
+{
+  cout<<"Synthetic Test 1"<<endl;
+  //Matrix Input is expected in this format
+  arma::mat x={{0, 3, 3, 4, 3, 1},{4, 4, 4, 5, 5, 2},{0, 1, 2, 2, 3, 3}};
+  std::string distance_file="distances.csv";
+  std::string neighbors_file="neighbors.csv";
+  double min_v=0, max_v=3;
```
rcurtin: No underscores in parameter names please; use `distanceFile`, `neighborsFile`, `minV`, `maxV` please.

src/mlpack/tests/main_tests/range_search_test.cpp
```diff
@@ -0,0 +1,292 @@
...
+  SetInputParam("neighbors_file", neighbors_file);
+  cout<<" 1.Parameters Set"<<endl;
+
+  mlpackMain();
+
+  cout<<" 2.Model Created"<<endl;
+  math::Range r(min_v, max_v);
+  vector<vector<size_t>> neighbors;
+  vector<vector<double>> distances;
+  CLI::GetParam<RSModel*>("output_model")->Search(r, neighbors, distances);
```
rcurtin: The `mlpackMain()` function has already performed the search for you and put the results in `distances_file` and `neighbors_file`.  You should instead open those files and look at the results to ensure that the results are the same, instead of using the output model to perform a search.

Specifically the reason that I say this is that what you're doing here is testing that the `RSModel` trains correctly and can be used for searching; but what we really should be testing is that the `mlpackMain()` function does everything: trains `RSModel` correctly, and *also* calls `Search()` with the range specified by `min` and `max` and then *also* stores the data correctly in `distances_file` and `neighbors_file`.

src/mlpack/tests/main_tests/range_search_test.cpp
```diff
@@ -0,0 +1,292 @@
...
+  cout<<" 3.Search Executed"<<endl;
+
+  CheckMatrices(neighbors, neighbor_val, 1e-5);
+  CheckMatrices(distances, distance_val);
+  cout<<" 4.Results Verified"<<endl;
+
+  cout<<"Passed Synthetic Test 1"<<endl<<endl;
+}
+
+BOOST_AUTO_TEST_CASE(ParameterTesting)
```
rcurtin: Can you tell me a bit more about what you are testing here?  I am not sure of the difference between what this test is testing and what the previous test is testing.

src/mlpack/tests/main_tests/range_search_test.cpp
```diff
@@ -0,0 +1,292 @@
...
+  CheckMatrices(neighbors, neighbor_val, 1e-5);
+  CheckMatrices(distances, distance_val);
+
+  cout<<" 4.Results Verified"<<endl;
+
+  cout<<"Passed Synthetic Test 2"<<endl<<endl;
+}
+
+BOOST_AUTO_TEST_CASE(ModelCheck)
+{//check if an output model can be used again for a different usage
```
rcurtin: Can you put the comment up above the `BOOST_AUTO_TEST_CASE()` please?

src/mlpack/tests/main_tests/range_search_test.cpp
```diff
@@ -0,0 +1,292 @@
...
+
+  cout<<" 3.Model Reinserted"<<endl;
+  if (!( output_model == CLI::GetParam<RSModel*>("output_model") ))
+  {
+    BOOST_FAIL("Models are not Equal");
+  }
+  else
+  {
+    cout<<"Model Checking Test Passed"<<endl<<endl;
+  }
```
rcurtin: It would be worth also testing that the results of the range search are the same for both the first and second runs of `mlpackMain()`.

src/mlpack/tests/main_tests/range_search_test.cpp
```diff
@@ -0,0 +1,292 @@
...
+      //To prevent warning for lack of definition
+    SetInputParam("min", min_v);
+    SetInputParam("max", max_v);
+    SetInputParam("distances_file", distance_file);
+    SetInputParam("neighbors_file", neighbors_file);
+
+    mlpackMain();
+
+    RSModel* output_model2 = std::move(CLI::GetParam<RSModel*>("output_model"));
+    output_model2->Search(r, neighbors_temp, distances_temp);
```
rcurtin: Again you should load the result files and test them here instead.

src/mlpack/tests/main_tests/range_search_test.cpp
```diff
@@ -0,0 +1,292 @@
...
+
+  mlpackMain();
+
+  RSModel* output_model1 = std::move(CLI::GetParam<RSModel*>("output_model"));
+
+  output_model1->Search(r, neighbors, distances);
+
+  bindings::tests::CleanMemory();
+
+  for(size_t i = 1 ; i < arr.size() ; i++)
```
rcurtin: I need to think about this one, but changing the leaf size will produce the exact same results for each search.  However, it will produce a different tree.  So, perhaps one thing you could do would be to serialize the trained `RSModel` to a string (by creating a `TextArchive` or something wrapped around a `std::ostringstream` and serializing to that in the usual way specified by boost::serialization documentation), and ensure that each string is different.  That would test that the resulting models contain trees built with different leaf sizes.

src/mlpack/tests/main_tests/range_search_test.cpp
```diff
@@ -0,0 +1,292 @@
...
+
+    mlpackMain();
+
+    CLI::GetParam<RSModel*>("output_model")->Search(std::move(query_data), r,
+                                                neighbors_temp, distances_temp);
+    CheckMatrices(neighbors, neighbors_temp);
+    CheckMatrices(distances, distances_temp);
+    cout<<"    Successful"<<endl;
+  }
+}
```
rcurtin: We could also check here that the trained models are different after serialization, just to ensure that the `tree_type` parameter is not ignored.

src/mlpack/tests/main_tests/range_search_test.cpp
niteya-shah: The Search function has two modes , one with query and without . The first ensures that we get the output without the query while the other ensures that we get the correct one with query passed . I felt like I should keep them different but I dont mind merging them into one . Sorry for the confusing name Ill change it .

src/mlpack/tests/main_tests/range_search_test.cpp
niteya-shah: @rcurtin 
The reason I am not loading those files is reading them is hard. Not only are the values jagged , we have cases where rows can be NULL , which will automatically exit most read functions. Is there any way to load them still . I dont think armadillo can do this as while I tried loading the matrice , It didnt read anything (got a 0x0 matrice) for my distance file 

1, 1.73205, 2.23607
1, 1.41421, 1.41421, 3
1.73205, 1.41421, 1.41421
2.23607, 1.41421, 1.41421
3
(The space is intentional )
Is there any way to properly read them ?

rcurtin: Yes, Armadillo will fail at doing this.  A custom function would probably be the way to go.  It shouldn't be too hard to do this with `std::fstream`.

src/mlpack/tests/main_tests/range_search_test.cpp
rcurtin: I see, and I agree, we should keep the tests different, but a more descriptive name and documentation for what the test does would help. :)

rcurtin:
Let me know when this is ready for another review. :+1:

src/mlpack/methods/range_search/rs_model_impl.hpp
```diff
@@ -59,6 +59,23 @@ inline RSModel::RSModel(RSModel&& other) :
   other.rSearch = decltype(other.rSearch)();
 }
 
+inline bool RSModel::operator==(RSModel other)
+{
+  if ( (this->treeType == other.treeType) && (this->leafSize == other.leafSize)
```
rcurtin: This doesn't conform with the style guide, it should be `if ((this->treeType ...` (and the same with the closing parentheses).

src/mlpack/methods/range_search/rs_model_impl.hpp
```diff
@@ -59,6 +59,23 @@ inline RSModel::RSModel(RSModel&& other) :
...
+                                     && (this->randomBasis == other.randomBasis)
+                                     && (this->rSearch == other.rSearch)
+                                     && (this->q.n_cols ==other.q.n_cols)
+                                     && (this->q.n_rows == other.q.n_rows) )
+  {
+    for (size_t i = 0; i < this->q.n_elem ; i++)
+      if (this->q[i] != other.q[i] )
```
rcurtin: Weird spacing here too, please change to:

```
for (size_t i = 0; i < this->q.n_elem; i++)
  if (this->q[i] != other.q[i])
...
```

Also, the `this->` is not really necessary.

src/mlpack/methods/range_search/rs_model_impl.hpp
```diff
@@ -59,6 +59,23 @@ inline RSModel::RSModel(RSModel&& other) :
...
+        return false;
+    return true;
+  }
+  else
+    return false;
```
rcurtin: We should enclose this in `{}` to make the grouping more clear.

```
else
{
  return false;
}
```

src/mlpack/methods/range_search/rs_model_impl.hpp
rcurtin: The spacing is still not right here; why is it different for the `std::vector<...` lines?

src/mlpack/methods/range_search/rs_model_impl.hpp
```diff
@@ -351,7 +368,7 @@ void BiSearchVisitor::SearchLeaf(RSType* rs) const
 }
 
 //! Save parameters for Train.
-TrainVisitor::TrainVisitor(arma::mat&& referenceSet,
+inline TrainVisitor::TrainVisitor(arma::mat&& referenceSet,
```
rcurtin: The second parameter is not aligned now that you have added `inline`.

src/mlpack/tests/main_tests/range_search_test.cpp
```diff
@@ -0,0 +1,389 @@
...
+* Check that the correct output is returned for a small synthetic input
+* case
+*/
+BOOST_AUTO_TEST_CASE(RangeSearchTest)
+{
+  //Matrix Input is expected in this format
+  arma::mat x = {{0, 3, 3, 4, 3, 1},
+                 {4, 4, 4, 5, 5, 2},
+                 {0, 1, 2, 2, 3, 3}};
+  std::string distancefile = "distances.csv";
```
rcurtin: I would prefer camelcase names, i.e. `distanceFile`, like the rest of the code.

src/mlpack/tests/main_tests/range_search_test.cpp
```diff
@@ -0,0 +1,389 @@
...
+    // Clear the settings.
+    bindings::tests::CleanMemory();
+    CLI::ClearSettings();
+  }
+};
+
+BOOST_FIXTURE_TEST_SUITE(RangeSearchMainTest, RangeSearchTestFixture);
+/*
+* Check that the correct output is returned for a small synthetic input
+* case
```
rcurtin: Please make sure all comments are complete sentences (including punctuation).

src/mlpack/tests/main_tests/range_search_test.cpp
```diff
@@ -0,0 +1,389 @@
...
+  std::string distancefile = "distances.csv";
+  std::string neighborsfile = "neighbors.csv";
+  double minv = 0, maxv = 3;
+  vector<vector<size_t>> neighborval = {{},
+                                        {2, 3, 4},
+                                        {1, 3, 4, 5},
+                                        {1, 2, 4},
+                                        {1, 2, 3},
+                                        {2}};
+  vector<vector<double>> distanceval = {{},{1, 1.73205, 2.23607},
```
rcurtin: Why no line break between `{},` and `{1, 1.73205, 2.23607}`?

src/mlpack/tests/main_tests/range_search_test.cpp
```diff
@@ -0,0 +1,389 @@
...
+  SetInputParam("distances_file", distancefile);
+  SetInputParam("neighbors_file", neighborsfile);
+
+  mlpackMain();
+
+  neighbors = ReadData<size_t>(neighborsfile);
+  distances = ReadData<double>(distancefile);
+
+  CheckMatrices(neighbors, neighborval, 1e-5f);
+  CheckMatrices(distances, distanceval);
```
rcurtin: We should also remove the files once the test is done:

```
remove(neighborsfile);
remove(distancefile);
```

src/mlpack/tests/main_tests/range_search_test.cpp
```diff
@@ -0,0 +1,389 @@
...
+}
+//Perform a Test for Search with a Query
```
rcurtin: Please put an empty line between the end of the previous test and the beginning of this one.  And a more comprehensive comment describing this test would be nice also.

src/mlpack/tests/main_tests/range_search_test.cpp
```diff
@@ -0,0 +1,389 @@
...
+BOOST_AUTO_TEST_CASE(RangeSeachTestwithQuery)
+{
+  arma::mat querydata = {{5, 3, 1},{4, 2, 4},{3, 1, 7}};
```
rcurtin: There should be spaces after the `,`s.

src/mlpack/tests/main_tests/range_search_test.cpp
```diff
@@ -0,0 +1,389 @@
...
+  SetInputParam("neighbors_file", neighborfile);
+  SetInputParam("query", querydata);
+
+  mlpackMain();
+
+  neighbors = ReadData<size_t>(neighborfile);
+  distances = ReadData<double>(distancefile);
+
+  RSModel* outputmodel = move(CLI::GetParam<RSModel*>("output_model"));
+  CLI::GetSingleton().Parameters()["reference"].wasPassed=false;
```
rcurtin: There should be spaces around the `=` here.

src/mlpack/tests/main_tests/range_search_test.cpp
```diff
@@ -0,0 +1,389 @@
...
+
+  neighborstemp = ReadData<size_t>(neighborfile);
+  distancetemp = ReadData<double>(distancefile);
+
+  CheckMatrices(neighbors, neighborstemp, 1e-5f);
+  CheckMatrices(distances, distancetemp);
+
+  if (!( outputmodel == CLI::GetParam<RSModel*>("output_model") ))
+  {
+
```
rcurtin: Why is there an empty line here?

src/mlpack/tests/main_tests/range_search_test.cpp
```diff
@@ -0,0 +1,389 @@
...
+  mlpackMain();
+
+  neighborstemp = ReadData<size_t>(neighborfile);
+  distancetemp = ReadData<double>(distancefile);
+
+  CheckMatrices(neighbors, neighborstemp, 1e-5f);
+  CheckMatrices(distances, distancetemp);
+
+  if (!( outputmodel == CLI::GetParam<RSModel*>("output_model") ))
```
rcurtin: Why is there a space after the opening `(!(`?

src/mlpack/tests/main_tests/range_search_test.cpp
```diff
@@ -0,0 +1,389 @@
...
+  {
+
+    BOOST_FAIL("Models are not Equal");
+  }
+
```
rcurtin: There is an extra line here too.

src/mlpack/tests/main_tests/range_search_test.cpp
```diff
@@ -0,0 +1,389 @@
...
+}
+
+BOOST_AUTO_TEST_CASE(LeafValueTesting)
+{
+  //Testing 3 different leaf values - default 20, 15 and 25
+  //Ensure that results match for different leaf values
```
rcurtin: We should probably put this comment up above the test definition.

src/mlpack/methods/range_search/rs_model.hpp
```diff
@@ -385,6 +386,10 @@ class RSModel
    */
   void CleanMemory();
 };
+std::string LoadModel(std::string s);
+void SaveModel(RSModel* model,std::string s);
+//Serialize 2 models and then check their equality
+bool CheckModelSerial(RSModel* , RSModel* );
```
rcurtin: These are only used in the tests, so I would suggest that any test-specific functionality be put there and not here.

src/mlpack/methods/range_search/rs_model_impl.hpp
```diff
@@ -487,6 +504,29 @@ inline bool& RSModel::Naive()
...
+  return buffer.str();
+}
+//Compare two RSModels by serialising them and then comparing their documents
+inline bool CheckModelSerial(RSModel* model1, RSModel* model2)
+{
+  std::string strmodel1="model1",strmodel2="model2";
+  SaveModel(model1,strmodel1);
+  SaveModel(model2,strmodel2);
+  return !LoadModel(strmodel1).compare(LoadModel(strmodel2));
+}
```
rcurtin: I would suggest, instead of making this check function, to just make a function that returns a string.  Your `SaveModel()` function could be adapted:

```
std::string ModelToString(RSModel* model)
{
  std::ostringstream oss;
  boost::archive::text_oarchive oa(oss);
  oa << model;
  return oss.str();
}
```

Then you can compare two strings and make sure they are not the same. :+1:

src/mlpack/tests/main_tests/range_search_test.cpp
```diff
@@ -0,0 +1,389 @@
...
+  SetInputParam("distances_file", distancefile);
+  SetInputParam("neighbors_file", neighborsfile);
+  SetInputParam("reference", inputdata);
+  SetInputParam("random_basis",true);
+
+  mlpackMain();
+
+  BOOST_REQUIRE_EQUAL(CheckModelSerial(outputmodel,
+                                    CLI::GetParam<RSModel*>("output_model")),0);
+}
```
rcurtin: An empty line between tests and a comment at the top of the next test would be appreciated.

src/mlpack/tests/main_tests/range_search_test.cpp
```diff
@@ -0,0 +1,389 @@
...
+  SetInputParam("max", maxv);
+  SetInputParam("distances_file", distancefile);
+  SetInputParam("neighbors_file", neighborsfile);
+  SetInputParam("reference", inputdata);
+  SetInputParam("random_basis",true);
+
+  mlpackMain();
+
+  BOOST_REQUIRE_EQUAL(CheckModelSerial(outputmodel,
+                                    CLI::GetParam<RSModel*>("output_model")),0);
```
rcurtin: With the changes I suggested to `CheckModelSerial()` you could change this to `BOOST_REQUIRE_NE(ModelToString(outputmodel), ModelToString(CLI::GetParam<RSModel*>("output_model")));`.

src/mlpack/tests/main_tests/range_search_test.cpp
```diff
@@ -0,0 +1,389 @@
...
+  SetInputParam("reference", inputdata);
+
+  mlpackMain();
+
+  RSModel* outputmodel = move(CLI::GetParam<RSModel*>("output_model"));
+
+  SetInputParam("min", minv);
+  SetInputParam("max", maxv);
+  SetInputParam("distances_file", distancefile);
+  SetInputParam("neighbors_file", neighborsfile);
```
rcurtin: It would be nice to check that the resulting distances and neighbors are the same also.

src/mlpack/tests/main_tests/range_search_test.cpp
```diff
@@ -0,0 +1,389 @@
...
```
rcurtin: Same here, we should check that the distances and neighbors are the same.

src/mlpack/tests/test_tools.hpp
```diff
@@ -171,5 +171,39 @@ inline std::string FilterFileName(const std::string& inputString)
...
+    std::sort(vec2[i].begin(), vec2[i].end());
+    for (size_t j = 0 ; j < vec1[i].size() ; j++)
+    {
+      BOOST_REQUIRE_CLOSE(static_cast<float>(vec1[i][j]), static_cast<float>(vec2[i][j]), tolerance);
+    }
+  }
+}
+//Load A csv file into a vector of vector of templated datatype (code strips ',')
+template<typename T>
+std::vector<std::vector<T>> ReadData(std::string const& path)
```
rcurtin: I would suggest putting this at the top of `range_search_test.cpp` since it's only used there.

The parameter should also be `const std::string& path` not `std::string const& path`.

src/mlpack/tests/test_tools.hpp
```diff
@@ -171,5 +171,39 @@ inline std::string FilterFileName(const std::string& inputString)
...
+    std::ifstream ifs(path);
+    std::vector<std::vector<T>> table;
+    std::string line;
+    while (std::getline(ifs, line))
+    {
+      std::vector<T> numbers ;
+      T n ;
+      std::replace(line.begin(), line.end(), ',', ' ');
+      std::istringstream stm(line) ;
+      while( stm >> n ) numbers.push_back(n) ;
```
rcurtin: Please put the statement of the loop on a new line.

src/mlpack/tests/test_tools.hpp
```diff
@@ -171,5 +171,39 @@ inline std::string FilterFileName(const std::string& inputString)
...
 
+  BOOST_REQUIRE_EQUAL( vec1.size() , vec2.size() );
+  for ( size_t i = 0; i < vec1.size() ; i++)
+  {
+    BOOST_REQUIRE_EQUAL( vec1[i].size(), vec2[i].size() );
+    std::sort(vec1[i].begin(), vec1[i].end());
+    std::sort(vec2[i].begin(), vec2[i].end());
+    for (size_t j = 0 ; j < vec1[i].size() ; j++)
+    {
+      BOOST_REQUIRE_CLOSE(static_cast<float>(vec1[i][j]), static_cast<float>(vec2[i][j]), tolerance);
```
rcurtin: Casting every element to a `float` sort of breaks the purpose of the template parameter here.  Since `size_t` needs to be compared with `BOOST_REQUIRE_EQUAL()` and `double` needs to be compared with `BOOST_REQUIRE_CLOSE()` I would suggest making two overloads specific to those cases.

It's up to you, but I'd also consider putting this at the top of `range_search_test.cpp` since it is only used there.

niteya-shah:
@rcurtin I have fixed the issues and added the things that you had asked for , can you review my PR?

src/mlpack/tests/main_tests/range_search_test.cpp
```diff
@@ -0,0 +1,572 @@
...
+* and are checked with pre-calculated neighbor and distance values
+*/
+BOOST_AUTO_TEST_CASE(RangeSearchTest)
+{
+  //Matrix Input is expected in this format
+  arma::mat x = {{0, 3, 3, 4, 3, 1},
+                 {4, 4, 4, 5, 5, 2},
+                 {0, 1, 2, 2, 3, 3}};
+  std::string distanceFile = "distances.csv";
+  std::string neighborsFile = "neighbors.csv";
```
rcurtin: Since you have `using namespace std` you can actually just call this `string` not `std::string`. :)

src/mlpack/tests/main_tests/range_search_test.cpp
```diff
@@ -0,0 +1,572 @@
...
+  vector<vector<double>> distanceVal = {{},
+                                        {1, 1.73205, 2.23607},
+                                        {1, 1.41421, 1.41421, 3},
+                                        {1.73205, 1.41421, 1.41421},
+                                        {2.23607, 1.41421, 1.41421},
+                                        {3}};
+  vector<vector<size_t>> neighbors;
+  vector<vector<double>> distances;
+  SetInputParam("reference", move(x));
+  //To prevent warning for lack of definition
```
rcurtin: `// To prevent warning for lack of definition.`

(I know it is pedantic, but all of our comments should be complete sentences.)

src/mlpack/methods/range_search/rs_model_impl.hpp
```diff
@@ -59,6 +59,25 @@ inline RSModel::RSModel(RSModel&& other) :
   other.rSearch = decltype(other.rSearch)();
 }
 
+inline bool RSModel::operator==(RSModel other)
+{
+  if ((this->treeType == other.treeType) && (this->leafSize == other.leafSize)
+                                     && (this->randomBasis == other.randomBasis)
+                                     && (this->rSearch == other.rSearch)
```
rcurtin: I realized that this approach has a major flaw.  `rSearch` is a `boost::variant` that internally will hold a pointer to an object.  When you make this comparison, what's likely to happen is that the pointer values will be compared, and those won't be the same unless the object is exactly the same.  That means that `operator==` will return `false` even when the objects are holding copies of the same `rSearch` model!

So I think that instead of using `operator==()` in your tests later on, we should remove this method and always do the comparison based on `ModelToString()`, which doesn't have this problem since the entire model is serialized to string and there are no pointer issues there.

src/mlpack/tests/main_tests/range_search_utils.hpp
```diff
@@ -0,0 +1,103 @@
...
+    for (size_t j = 0; j < vec1[i].size(); j++)
+    {
+      BOOST_REQUIRE_EQUAL(vec1[i][j], vec2[i][j]);
+    }
+  }
+}
+
+/*
+* Load A csv file into a vector of vector of templated datatype (code strips ',')
+* by splitting on '\n' for lines and spaces for parts of a line
```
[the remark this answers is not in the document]
niteya-shah: This comment was exceeding 80 chars so I moved part of it to the next line , it is one sentence .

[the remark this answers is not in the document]
rcurtin: Sorry, I should have clarified---by 'complete sentence' I also mean punctuation and capitalization, and we should make sure the meaning is clear.  So for this one I might suggest something instead like this:

```
/**
 * Load a CSV file into a vector of vector with a templated datatype.  Any ','
 * characters are stripped from the input; lines are split on '\n' and elements
 * of each line are split on spaces.
 */
```

src/mlpack/tests/main_tests/range_search_utils.hpp
```diff
@@ -0,0 +1,103 @@
...
+
+/*
+* Check for 2 matrices of type vector<vector<double>> to ensure that their
+* values dont differ by more than tolerance , default is 0.001%
+* @param vec1 - vector 1 to be checked
+* @param vec2 - vector 2 to be checked
+* @param tolerance - difference in values in allowed
+*/
+inline void CheckMatrices(std::vector<std::vector<double>> vec1,
+                          std::vector<std::vector<double>> vec2,
```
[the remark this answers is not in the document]
niteya-shah: Using const reference throws an error because we have to sort the vectors later to compare them , so I made them reference the original , but they are not const .

niteya-shah:
@rcurtin I removed the == operator and made the fixes .

rcurtin:
@niteya-shah thanks for the contribution!  I made some style and comment changes in 64e64c4, aimed at making the comments clearer and there are a few style fixes also.  I hope this gives an idea of the type of comments and style we are looking for. :+1:

Issue #1635, opened 2019-01-07
Incorrect function signature for SparseSVMFunction::Gradient()


AYESDIE:
While going through #1633, I found the signature for function `SparseSVMFunction::Gradient()` in it's declaration and definition doesn't match.

`sparse_svm_function.hpp`
```
/**
   * Evaluate the gradient the gradient of the hinge loss function, following
   * the SparseFunctionType requirements on the Gradient function.
   *
   * @param parameters The parameters of the SVM.
   * @param id Index of the datapoint to use for the gradient evaluation.
   * @param gradient Sparse matrix to output the gradient into.
   */
  template <typename GradType>
  void Gradient(const arma::mat& parameters, size_t id, GradType& gradient);
```

`sparse_svm_function_impl.hpp`
```
template <typename GradType>
void SparseSVMFunction::Gradient(
    const arma::mat& parameters,
    const size_t firstId,
    GradType& gradient,
    const size_t batchSize)
```

rcurtin:
Fixed via #1635.  Thanks for pointing out the issue. :+1:

Issue #1633, opened 2019-01-06
Add binding and tests for sparse_svm


rcurtin:
We have a nice implementation of a sparse SVM function from @shikharbhardwaj in `src/mlpack/methods/sparse_svm/`, but we actually don't have a nice binding to use it and it's not tested either.  This issue would be a nice task for someone who wants to learn more about the binding system and how to write new bindings.  To work on this, familiarity with SVMs will be necessary.  I'll try and write a basic description of what's involved here, but don't feel obligated to follow it to the letter---it is just a guide.

1.  Take a look at `src/mlpack/methods/sparse_svm/sparse_svm_function.hpp` (and `sparse_svm_function_impl.hpp`) to understand the function that is being optimized.  Basically this is just the linear SVM objective function.  Now take a look at the code in `src/mlpack/methods/softmax_regression/`---that directory has `softmax_regression_function.hpp` just like `sparse_svm_function.hpp`, but it also has a class `SoftmaxRegression` in `softmax_regression.hpp` which we lack for sparse SVMs.

2. Take a look at the way `SoftmaxRegression` is tested in `src/mlpack/tests/softmax_regression_test.cpp` (and also `LogisticRegression` and `DecisionTree` have relevant tests, since these are all classifiers).

3. The first step of coding here will actually be to create a `SparseSVM` class that is a lot like the `SoftmaxRegression` class (i.e. has basically the same method signatures), except that it will use `SparseSVMFunction` internally instead of `SoftmaxRegressionFunction`.  Note that you'll need to use ensmallen optimizers for this; specifically, you'll want to use a sparse differentiable separable optimizer type (http://ensmallen.org/docs.html#sparse-differentiable-separable-functions) by default, but the user should be able to choose the optimizer type they want.  It may also be useful to provide a function `SparseSVMFunction::EvaluateWithGradient()`---that can help make things faster.

4. Once we have a `SparseSVM` class, we'll need to write tests for it in the same way that the other classifiers have tests written.  Create a file `sparse_svm_test.cpp` with these tests in it, using the other tests that you looked at earlier as a guide.

[...]

PR #1636, opened 2019-01-07
Fixed incorrect signature for SparseSVMFunction::Gradient().

AYESDIE:
Fix for the issue #1635.

Updated the decalaration of `Gradient()` inside `class SparseSVMFunction` to match the definition inside `sparse_svm_function_impl.hpp`.


rcurtin:
Ah, ok, I guess that I didn't check my email right. :)  If you don't get them after about a month after I originally shipped them, we can try again.

I'd say for #1633, you might want to talk to the other person working on it and see if you can split the task.  I think many of the steps could be handled by different people.  (i.e. one person could write the sparse SVM tests while another writes the SparseSVM class, etc.)

rcurtin:
Since there are three approvals, I'll go ahead and merge now. :+1:

Issue #1506, opened 2018-09-06
Minor Windows installer fixes/changes


rcurtin:
These are just a few leftover changes from #1485.  They aren't high priority but I don't think any of them are too hard.

1. Higher resolution image that looks less blurry.  https://github.com/mlpack/mlpack/pull/1485#issuecomment-417945989
2. Wrapped lines for AppVeyor config.  (Not totally sure if this will work.)  https://github.com/mlpack/mlpack/pull/1485#discussion_r214467616
3. Use git version in build number by parsing gitversion.hpp if it exists.  https://github.com/mlpack/mlpack/pull/1485#discussion_r214466275

This could be a nice intro task for someone with some Windows knowledge who wants to contribute.

gmanlan:
Another improvement might be to remove the *_test.exe and the test dependency (boost_unit_test_framework-vc140-mt-1_60.dll). Having the test in the installer may not be that useful (and it fails anyways because it can't find many test files).

KimSangYeon-DGU:
Hi @rcurtin. In work list, the 1 and 2 have been done and the 3 almost has been done. To use the git version in build, the version should be `x.x.x.x` or `x.x.x` format(x is integer from 0 to 655). However, the git version is 7 hexadecimal string, so we should convert this string to x.x.x.x format. So, I have a idea on converting it. **Combine two words into a integer number**. Below is example. Let's assume the git version is `abcd207`.
`
1. User's git version is **abcd207**.
2. Add the end of git version to '0' (like padding).
3. So changed git version is **abcd2070**
4. A binary of **a**, and **b** is 1010 and 1011 respectively, so it can be 1010 1011 (171)
5. Likewise, the rest of pair can be converted to 205, 32, and 112.
6. As a result, the git version abcd207 can be converted to 171.205.32.112.
`
Thus, we can distinguish between `git version` and `official mlpack version` using the number of `x`. (if the number of `x` is 3, it is `official mlpack version`, otherwise it is 'git version')
So, is it proper way to converting it??

rcurtin:
[...]

PR #1642, opened 2019-01-09
Fix/Change Windows installer

KimSangYeon-DGU:
This is fix/change related to #1506 .

- [x] 1. Higher resolution image that looks less blurry.
--> You can check the result design at [this](http://sylaboratory.tistory.com/24) or AppVeyor.

- [x] 2. Wrapped lines under 80 for AppVeyor config.
--> One of them is 82.(Of course, we can reduce the line.)

- [x] 3. Use git version in build number by parsing gitversion.hpp if it exists.
--> The product version's format should be x.x.x.x (x is integer from 0 to 65534), but the gitversion's format is a string like `mlpack git-ab34cd2`. So the string is converted to 4 integers.


ShikharJ:
Hi @KimSangYeon-DGU, do you think you can squash these commits into one, so that it becomes easier to review them later on? I'd suggest since this PR has three parts, maybe you can try pushing a commit each for each sub-part?

KimSangYeon-DGU:
> Hi @KimSangYeon-DGU, do you think you can squash these commits into one, so that it becomes easier to review them later on? I'd suggest since this PR has three parts, maybe you can try pushing a commit each for each sub-part?

That's good idea. :) I'll do that.

ShikharJ:
There's no need to make another PR, just simply force-push to this branch, after you have squashed the commits.

KimSangYeon-DGU:
@ShikharJ 
I'm sorry... In the middle of force-push, this PR is closed automatically...

zoq:
Looks like with the changes we run into the appveyor build timeout. Do you think we could reduce the build/test time? Another idea might be to setup a windows machine for jenkins. I could do that if there is interest in doing so.

KimSangYeon-DGU:
> Looks like with the changes we run into the appveyor build timeout. Do you think we could reduce the build/test time? Another idea might be to setup a windows machine for jenkins. I could do that if there is interest in doing so.

I think a possible way to reduce time. According to @gmanlan in #1506, our test case is always failed in windows build. So if we remove `the *_test.exe and the test dependency (boost_unit_test_framework-vc140-mt-1_60.dll)`, we could reduce the time by about 5 minutes. However, I'm not sure it is proper way.

gmanlan:
A clarification: in #1506 I have mentioned that the test binaries and dependencies are included in the Windows installer, and therefore are failing to execute on the end user system because the test data is not bundled in the installer. I was not referring to tests failing in the build server environment. That's why my suggestion was to exclude the test binaries from the installer files because it has little use for the end user that installs mlpack.

KimSangYeon-DGU:
> A clarification: in #1506 I have mentioned that the test binaries and dependencies are included in the Windows installer, and therefore are failing to execute on the end user system because the test data is not bundled in the installer. I was not referring to tests failing in the build server environment. That's why my suggestion was to exclude the test binaries from the installer files because it has little use for the end user that installs mlpack.

I apologize for misunderstanding. I'm really sorry... and thanks for your clarification.
May I ask you a question?? Do you think we can reduce the build time??

KimSangYeon-DGU:
While in use mobile phone, I pushed the closed button by mistake.

.appveyor.yml
```diff
@@ -20,34 +20,106 @@ install:
...
-  - ps: try{(Add-Type -AssemblyName System.IO.Compression.FileSystem); [System.IO.Compression.ZipFile]::ExtractToDirectory('C:\projects\mlpack\dist\win-installer\jenkinsdoc.zip', 'C:\projects\mlpack\dist\win-installer\staging\doc')}catch{Write-Output "Unable to add doc to installer, skipping!"}
+  # Pulling documentation for the installer.
+  - ps: >
+      try{(new-object net.webclient).DownloadFile('http://ci.mlpack.org/job/mlpack%20-%20doxygen%20build/lastSuccessfulBuild/artifact/build/doc/html/*zip*/html.zip',
+      'C:\projects\mlpack\dist\win-installer\jenkinsdoc.zip')}catch{Write-Output "Unable to pull jenkins doc, skipping!"}
+  - ps: >
+      try{(Add-Type -AssemblyName System.IO.Compression.FileSystem);
+      [System.IO.Compression.ZipFile]::ExtractToDirectory('C:\projects\mlpack\dist\win-installer\jenkinsdoc.zip',
+      'C:\projects\mlpack\dist\win-installer\staging\doc')}catch{Write-Output "Unable to add doc to installer, skipping!"}
```
rcurtin: It would be really nice if we could wrap these to under 80 characters, but given the long URLs I think it may not be possible.  Nice to know about the `>` character though for sure. :+1:

KimSangYeon-DGU:
> For the build time issue, @gmanlan is right, he was referring to the WiX installer, not what happens with AppVeyor when it is testing the build. 

Yes, I agree.

> It looks like actually that you have significantly reduced the build time given the last AppVeyor build, but I'm not sure what you did to do that. Do you have any thoughts on that? 

Actually, I just wrapped the lines. This is the same as before in an aspect of function. However, it is clear that the build time depends on the situation. Sometimes it is fast and sometimes it is slow. I don't know exactly, however, judging from that, I assumed that the threads might be the problem in the Appveyor virtual machine.

> The test time itself is not typically too long; it takes about 10 minutes on Travis. We might be able to reduce individual test times, but we could open another issue for that if we like.

Yes, I think it's better to manage this problem with a new issue, because we need to look into it more closely.

rcurtin:
> Sometimes it is short and sometimes it is long. I don't know exactly, however, judging from that, I assumed that the threads might be the problem in our Appveyor virtual machine.

Ah, ok, I see.  We should open a separate issue then.  Would you like to do that?

I think the only thing left to happen here is to handle the versioning scheme I mentioned in a comment at the bottom of #1506, and then I would be happy to merge this.

KimSangYeon-DGU:
> Ah, ok, I see. We should open a separate issue then. Would you like to do that?

I think the only thing left to happen here is to handle the versioning scheme I mentioned in a comment at the bottom of #1506, and then I would be happy to merge this.

Yes, I'll do that. :)

.appveyor.yml
```diff
@@ -130,6 +100,36 @@ build_script:
...
+        $gitversion = ($ver.Split('"')[1].Split(' ')[1].Split('-')[1]) + '0';
+        for ($i=0; $i -lt 4; $i++)
+        {
+          $env:MLPACK_VERSION += ([convert]::toint16($gitversion.substring($i*2, 2), 16)).ToString();
+          if ($i -lt 3)
+          {
+            $env:MLPACK_VERSION += '.'
+          }
+        }
+      }
```
KimSangYeon-DGU: This code extracts the git version and modifies it to `x.x.x.x` format, if the automatically generated file `C:\projects\mlpack\src\mlpack\core\util\gitversion.hpp` exists.

rcurtin: @KimSangYeon-DGU I wrote a clarification in #1506, can you try implementing that please instead?  I've pasted the comment below:

"Hehe, @KimSangYeon-DGU, thanks for the extensive analysis. But I was thinking something much simpler, just like we do for existing Linux builds. :) Basically we just use the git hash... so if the hash is b23d5ce then the version is mlpack git-b23d5ce. CMake will already generate ${CMAKE_BINARY_DIR}/src/mlpack/core/util/gitversion.hpp, which will have this text in it, so I think the only thing to do would be to use a line in the AppVeyor config in order to extract that text and modify the WiX configuration accordingly."

rcurtin:
Thanks again! :+1:

PR #1648, opened 2019-01-09
Updated SparseSVMFunction to LinearSVMFunction and added an initial LinearSVM class.

AYESDIE:
A PR for issue: #1633 

**Latest Changes:**
  - Added the reference to both the research papers.
  - Added a `fitIntercept` to consider `b_i` terms while evaluation.
    - Added test for `fitIntercept`.
  - Removed redundant test case `LinearSVMClassifyTest`.
  - Implemented `LinearSVMFunctionGradient` test in another way which doesn't fail (cannot guarantee this, but hope so).
  - Refactors:
    - `Evaluate()` and `EvaluateWithGradient()` returns `loss + regularization` directly.
    - `loss` is evaluated in a single line.
    - Added comments for the implementation of `Gradient()`.
    - `scores` is evaluated in a single line.

Change Logs:
- [10/01/2019]:
  - Moved the code for `sparse_svm_function.hpp` and `sparse_svm_function_impl.hpp`under the mlpack namespace.
  - Included missing headers for `MakeAlias` and `ShuffleData` in `sparse_svm_function_impl.hpp`.
  - Added `EvaluateWithGradient()` for `SparseSVMFunction` class.
  - Added a basic starter code for `SparseSVM` class.

- [03/02/2019]:
  - Updated to adapt multi-class classification using a `groundTruth` matrix. 
  - Added L2-Regularization parameter. 

- [04/02/2019]:
  - Added tests for `SparseSVMFunction` for `Evaluate()` and `Gradient()`.

- [05/02/2019]:
  - Changed everything from SparseSVM to LinearSVM.
  - Updates for class `LinearSVMFunction`:
    - Adding L2-Regularization Parameter.
    - Making the SVM function to work as a `multi class` classifier.
    - Added `EvaluateWithGradient()`.
  - `LinearSVM`:
    - A starter code that stores the parameters for the model. (Similar to `SoftmaxRegression`).
  - `Tests`:
    - `LinearSVMFunction::Evaluate()` is tested on variety of data with and without regularization parameter.
    - `LinearSVMFunction::Gradient()` is tested for a random dataset.

- [11/02/2019]
  - Added tests for `LinearSVM` Class.
  - Fixed the functions for proper working of `LinearSVM`.
  - Rebased this branch on `mlpack:master` for fixing the build issues.

- [21/02/2019]
  - Addressed these changes to HISTORY.md because I think this might be an important update.

- [05/03/2019]
  - Added `Δ` Hyperparameter and updated tests according to it.


AYESDIE:
`sparse_svm.cpp` and `sparse_svm.hpp` are not included in the `CMakeLists.txt` yet.

AYESDIE:
The parameters for `arma::max()` inside `SparseSVMFunction::Evaluate()` are not being passed properly which is leading the build to fail.
I think `arma::max()` has been updated since this function was written.
I will update it soon.

AYESDIE:
@rcurtin I added the changes to `SparseSVMFunction::Evaluate()` by clamping dots from `0.0` to `dots.max()`.
There was also an issue with `SparseSVM::Classify()` that is addressed.

rcurtin:
@AYESDIE good to hear you got it worked out. Do you want to add tests for `SparseSVM` in this PR or in a future PR?

meemeeseekuu:
I will add the tests for `SparseSVM` is this PR.

Is it fine if I make another PR for bindings and tests?

AYESDIE:
I will add the tests for `SparseSVM` in this PR.

Is it fine if I make another PR for bindings and tests?

AYESDIE:
@rcurtin working on the test and completing the tests will take a bit more time than I expected, I'll try to add the tests as soon as possible and then start working on adding bindings for it.

src/mlpack/methods/sparse_svm/sparse_svm.cpp
```diff
@@ -0,0 +1,85 @@
...
+#define MLPACK_METHODS_SPARSE_SVM_SPARSE_SVM_CPP
+
+// In case it hasn't been included yet.
+#include "sparse_svm.hpp"
+
+namespace mlpack {
+namespace svm {
+
+template <typename OptimizerType>
+SparseSVM::SparseSVM(
```
rcurtin: Since this is a templated function, we'd actually need to put it in a header file.  You could move it (and the other templated methods in this file) to `sparse_svm_impl.hpp`, which you can then include from `sparse_svm.hpp`.

src/mlpack/methods/sparse_svm/sparse_svm.cpp
```diff
@@ -0,0 +1,85 @@
...
+  Log::Info << "SparseSVM::SparseSVM(): final objective of "
+            << "trained model is " << out << "." << std::endl;
+
+  return out;
+}
+
+} // namespace regression
+} // namespace mlpack
+
+#endif
```
rcurtin: Can you make sure there's a newline at the end of each file please?

src/mlpack/methods/sparse_svm/sparse_svm.hpp
```diff
@@ -0,0 +1,104 @@
...
+
+#include <mlpack/prereqs.hpp>
+#include <ensmallen.hpp>
+
+#include "sparse_svm_function.hpp"
+
+namespace mlpack {
+namespace svm {
+
+class SparseSVM
```
rcurtin: Can we add some class documentation here describing what a sparse SVM is (briefly), any relevant paper references, and maybe point to where more information can be found?

src/mlpack/methods/sparse_svm/sparse_svm.hpp
```diff
@@ -0,0 +1,104 @@
...
+   * @param labels Labels associated with the feature data.
+   * @param optimizer Desired optimizer.
+   */
+  template <typename OptimizerType = ens::ParallelSGD<>>
+  SparseSVM(const arma::mat& data,
+            const arma::Row<size_t>& labels,
+            OptimizerType optimizer = OptimizerType());
+
+  /**
+   * Classify the given points, returning  predicted class label ({-1, 1})
```
rcurtin: `size_t` actually won't hold `-1` well, so we should make the class labels `0` or `1`.  You might need to rework the implementation for that a little bit too.

src/mlpack/methods/sparse_svm/sparse_svm_function.hpp
```diff
@@ -57,6 +79,41 @@ class SparseSVMFunction
...
+  double EvaluateWithGradient(const arma::mat& parameters,
+                              GradType& gradient) const;
+
+  /**
+   * Evaluate the gradient of the hinge loss function, following
+   * the SparseFunctionType requirements on the Gradient function
+   * followed by evaluation of the hinge loss function on the specified
+   * datapoints.
+   *
+   * @tparam GradType
```
rcurtin: Oops, I think you forgot documentation for this.  I guess `Type of the gradient matrix` could work, for instance.

src/mlpack/methods/sparse_svm/sparse_svm_function_impl.hpp
```diff
@@ -13,13 +13,19 @@
...
+namespace mlpack {
+namespace svm {
+
 SparseSVMFunction::SparseSVMFunction(
-    const arma::sp_mat& dataset, const arma::vec& labels) :
-    dataset(dataset),
-    labels(math::MakeAlias(const_cast<arma::vec&>(labels), false))
+  const arma::sp_mat& dataset, const arma::vec& labels) :
+  dataset(dataset),
+  labels(math::MakeAlias(const_cast<arma::vec&>(labels), false))
```
rcurtin: Hmm, should we make labels have type `arma::Row<size_t>` like for `Classify()`?

src/mlpack/tests/sparse_svm_test.cpp
```diff
@@ -0,0 +1,128 @@
...
+  const size_t dimension = 10;
+  const size_t points = 1000;
+  const size_t trials = 50;
+
+  // Initialize a random dataset.
+  arma::sp_mat data;
+  data = arma::sprandu(dimension, points, 0.2);
+
+  // Create random response.
+  arma::vec responses(points);
```
rcurtin: Here also I think we should use `arma::Row<size_t>` (and the labels `0` and `1`) for consistency.

src/mlpack/methods/sparse_svm/sparse_svm_function_impl.hpp
```diff
@@ -36,44 +42,124 @@ void SparseSVMFunction::Shuffle()
   labels = std::move(newLabels);
 }
 
+double SparseSVMFunction::Evaluate(const arma::mat& parameters)
+{
+  arma::mat params = parameters.head_cols(parameters.n_cols - 1);
```
rcurtin: This actually will perform a copy, so it would be faster to use `parameters.head_cols().t()` directly in the expression of `dots`.

src/mlpack/methods/sparse_svm/sparse_svm.hpp
AYESDIE: I think I missed this and probably didn't notice it since I haven't made any tests on this class. But thanks for pointing this out, I will update the way it works around labels `0` and `1` rather than `-1` and `1`.

AYESDIE:
@rcurtin I updated the implementation and these are the stuff from your comments that have been addressed:
- The `SparseSVMFunction` is templatized and we can now use it as a regular SVM if we use a dense matrix or a Sparse SVM when using sparse matrix.
- Added the L2-Regularization Parameter.
- I was not able to extend to Kernel SVMs yet since it requires Lagrange Multipliers for its evaluation and I was not able to make that in a *vectorized* form, instead what I did was that I extended it to be used as a multi-class SVM rather than classifying only 2 classes(0, 1).
- Labels are fixed from (-1, 1) and it can accept only labels as (0, 1, ... numClasses - 1) now.
- Made new tests for `SparseSVMFunction` (multi-class). 
- Tests for `SparseSVM` are still underway.
- All the issues in the *review* have also been addressed

src/mlpack/methods/linear_svm/linear_svm_impl.hpp
```diff
@@ -0,0 +1,163 @@
...
+
+  // Prepare necessary data
+  labels.zeros(data.n_cols);
+  double maxScore = 0;
+
+  // For each test input.
+  for (size_t i = 0; i < data.n_cols; ++i)
+  {
+    // For each class.
+    for (size_t j = 0; j < numClasses; ++j) {
```
rcurtin: Small style issue---please make sure the `{` is on the following line. :+1:

src/mlpack/methods/linear_svm/linear_svm_impl.hpp
```diff
@@ -0,0 +1,163 @@
...
+
+    // Set maximum probability to zero for next input.
+    maxScore = 0;
+  }
+}
+
+template <typename MatType>
+void LinearSVM<MatType>::Classify(
+        const MatType& data,
+        arma::Row<size_t> &labels,
```
rcurtin: The `&` should be grouped with the `arma::Row<size_t>`.

src/mlpack/methods/linear_svm/linear_svm_impl.hpp
```diff
@@ -0,0 +1,163 @@
...
+            << "trained model is " << out << "." << std::endl;
+
+  return out;
+}
+
+template <typename MatType>
+void LinearSVM<MatType>::Classify(
+    const MatType& data,
+    arma::Row<size_t> &labels)
+const
```
rcurtin: Could you call the other implementation of `Classify()` here instead of reimplementing?

AYESDIE: I followed these signature and the way they were implemented from Softmax Regression.
Should a similar thing be done for `LinearSVMFunction::Gradient()`, `LinearSVMFunction::Evaluate()` and `LinearSVMFunction::EvaluateWithGradient()` (Separable and non-separable) since the implementation is almost similar (I think not).

rcurtin: There may be an efficiency improvement for `EvaluateWithGradient()` if any of the computation is shared.  You can take a look at the implementation to see.  If not, there's no need for `EvaluateWithGradient()`.

More details here: http://ensmallen.org/docs.html#differentiable-functions

src/mlpack/methods/linear_svm/linear_svm_impl.hpp
```diff
@@ -0,0 +1,143 @@
...
+{
+  arma::Row<size_t> labels;
+
+  // Get predictions for the provided data.
+  Classify(testData, labels);
+
+  // Increment count for every correctly predicted label.
+  size_t count = 0;
+  for (size_t i = 0; i < labels.n_elem ; i++)
+    if (testLabels(i))
```
AYESDIE: I was looking for the bug in my `Classify()` when I found this gem :laughing:

AYESDIE:
The tests for `LinearSVM` are complete.

rcurtin:
@ayesdie are you thinking that you will finish the binding entirely?  If so I will wait until then to review it.

AYESDIE:
I have my exams coming up, so I'll be a bit busy with that for the next week. I'll continue on the bindings after that if it sounds good.

AYESDIE:
By the way, how many options should this binding offer.
I'm thinking of adding these two missing options:
- optimizer type
- matrix type ( dense or sparse)

Is there anything else that could be included into this?

rcurtin:
Ok, sounds good.  If you want to merge this without the bindings, we can review just that part and handle the bindings in another PR, but that's up to you.

I think the optimizer type and the matrix type are good to add; we should also allow access to `lambda` but I can't think of anything else.

AYESDIE:
The binding had the option to access `lambda`.

Well, I've removed the bindings in this PR, so you may review it.

AYESDIE:
Hi @rcurtin , I removed the bindings from this PR (some work still needs to be done on them, so I will add those in another PR once they are ready) and added a few more tests for `LinearSVM` using `ParallelSGD` and `GradientDescent`.
I also moved the `LinearSVMTest` under `parallel_tests`.
Could you please review this?

src/mlpack/methods/linear_svm/linear_svm.hpp
```diff
@@ -0,0 +1,195 @@
...
+  LinearSVM(const MatType& data,
+            const arma::Row<size_t>& labels,
+            const size_t numClasses = 2,
+            const double lambda = 0.0001,
+            OptimizerType optimizer = OptimizerType());
+
+  /**
+   * Initialize the SoftmaxRegression without performing training.  Default
+   * value of lambda is 0.0001.  Be sure to use Train() before calling
+   * Classify() or ComputeAccuracy(), otherwise the results may be meaningless.
```
rcurtin: You can also just add `lambda` as a parameter here, but I think it's okay either way.

src/mlpack/methods/linear_svm/linear_svm.hpp
```diff
@@ -0,0 +1,195 @@
...
+  size_t Classify(const VecType& point) const;
+
+  /**
+   * Computes accuracy of the learned model given the feature data and the
+   * labels associated with each data point. Predictions are made using the
+   * provided data and are compared with the actual labels.
+   *
+   * @param testData Matrix of data points using which predictions are made.
+   * @param testLabels Vector of labels associated with the data.
+   * @return Accuracy of the model.
```
rcurtin: Can you point out whether this will be between 0 and 1 or 0 and 100?  (I think 0 and 1 would be better, but I don't have a strong opinion.)

src/mlpack/methods/linear_svm/linear_svm_function.hpp
```diff
@@ -0,0 +1,210 @@
...
+ * 3-clause BSD license along with mlpack.  If not, see
+ * http://www.opensource.org/licenses/BSD-3-Clause for more information.
+ */
+#ifndef MLPACK_METHODS_LINEAR_SVM_LINEAR_SVM_FUNCTION_HPP
+#define MLPACK_METHODS_LINEAR_SVM_LINEAR_SVM_FUNCTION_HPP
+
+#include <mlpack/prereqs.hpp>
+
+namespace mlpack {
+namespace svm {
```
rcurtin: Picky comment, but can you leave an empty line between the namespace opening and the comment for `LinearSVMFunction`?

src/mlpack/methods/linear_svm/linear_svm_function_impl.hpp
```diff
@@ -0,0 +1,365 @@
...
+
+// In case it hasn't been included yet.
+#include "linear_svm_function.hpp"
+
+namespace mlpack {
+namespace svm {
+
+template <typename MatType>
+LinearSVMFunction<MatType>::LinearSVMFunction(
+  const MatType& dataset,
```
rcurtin: This should be doubly-indented (i.e. four spaces not two). :+1:

src/mlpack/methods/linear_svm/linear_svm_function_impl.hpp
```diff
@@ -0,0 +1,365 @@
...
+template <typename MatType>
+void LinearSVMFunction<MatType>::InitializeWeights(
+        arma::mat &weights,
+        const size_t featureSize,
+        const size_t numClasses)
+{
+  // Initialize values to 0.005 * r. 'r' is a matrix of random values taken from
+  // a Gaussian distribution with mean zero and variance one.
+  weights.randn(numClasses, featureSize);
+  weights *= 0.005;
```
rcurtin: Fiven the simplicity of the initialization here, I would suggest entirely removing the `InitializeWeights()` method and its two overloads, and inline it into the constructor where it's only used once.

src/mlpack/methods/linear_svm/linear_svm_function_impl.hpp
```diff
@@ -0,0 +1,365 @@
...
+  double loss, regularization, cost;
+
+  arma::mat scores = parameters * dataset;
+  arma::sp_mat correctScores = scores % groundTruth;
+  arma::mat margin = scores
+      - arma::repmat(arma::ones(numClasses).t() * correctScores, numClasses, 1)
+      + 1 - groundTruth;
+
+  // The Hinge Loss Function
+  loss = arma::accu(arma::clamp(margin, 0.0, margin.max()));
```
rcurtin: You could just use `DBL_MAX` or similar for the upper bound; it'll take non-constant time to compute `margin.max()`.

src/mlpack/methods/linear_svm/linear_svm.hpp
```diff
@@ -0,0 +1,195 @@
...
+  double Train(const MatType& data,
+               const arma::Row<size_t>& labels,
+               const size_t numClasses = 2,
+               OptimizerType optimizer = OptimizerType());
+
+
+  //! Sets the number of classes.
+  size_t& NumClasses() { return numClasses; }
+  //! Gets the number of classes.
+  size_t NumClasses() const { return numClasses; }
```
rcurtin: Typically for an SVM, the way to do classification for more than two classes is to build `numClasses` all-vs-one classifiers, and then for a test point, to select the class with the greatest margin.  Are you doing something different here?  Or is this restricted to two classes?

rcurtin: In my view, either we should restrict `LinearSVMFunction` to two classes and then internally hold multiple SVMs to do multiclass; or, if you are doing something different, we should make it clear what strategy we are using for multiclass linear SVM in the comments.  Let me know if I overlooked anything. :+1:

AYESDIE: I'm using [Weston and Watkins](https://www.elen.ucl.ac.be/Proceedings/esann/esannpdf/es1999-461.pdf) approach for this problem (which distinguish all classes in one single optimization processing).
Here, I evaluate the loss as
image
This doesn't restrict `LinearSVMFunction` to two classes. 
So, this implementation designs a single objective function for training all `numClasses`-binary SVMs simultaneously and maximize the margins from each class to the remaining ones.

rcurtin: Perfect, thank you for the clarification.  Can you link to that in the documentation somewhere (and possibly document that this is an extension of binary linear SVM to multiclass)?  I'll review the paper soon and then I'll be able to check the implementation fully. :+1:

AYESDIE: I have linked the document in the comment, but here it is again (https://www.elen.ucl.ac.be/Proceedings/esann/esannpdf/es1999-461.pdf).

src/mlpack/methods/linear_svm/linear_svm.hpp
```diff
@@ -0,0 +1,211 @@
...
+ * model, and supports training with multiple optimizers and classification.
+ * The class supports different observation types via the MatType template
+ * parameter; for instance, support vector classification can be performed
+ * on sparse datasets by specifying arma::sp_mat as the MatType parameter.
+ *
+ * Linear SVM can be used for general classification tasks which will work
+ * on multiclass classification. More technical details about
+ * the model can be found on the following webpage:
+ *
+ * https://link.springer.com/content/pdf/10.1007%2FBF00994018.pdf
```
rcurtin: It doesn't seem like this details the multiclass linear SVM extension you gave me in https://www.elen.ucl.ac.be/Proceedings/esann/esannpdf/es1999-461.pdf .  Should we reference both of those here?  (Also we should use BibTeX for the references, since direct URL links may not always work.)

src/mlpack/methods/linear_svm/linear_svm_function_impl.hpp
```diff
@@ -0,0 +1,352 @@
...
+  // calculated over all the training examples.
+
+  // Calculate the loss and regularization terms.
+  double loss, regularization, cost;
+
+  arma::mat scores = parameters * dataset;
+  arma::sp_mat correctScores = scores % groundTruth;
+  arma::mat margin = scores
+      - arma::repmat(arma::ones(numClasses).t() * correctScores, numClasses, 1)
+      + 1 - groundTruth;
```
rcurtin: Hmm, I'm not easily following this (that doesn't mean it's wrong, just that I don't 100% understand yet).  Working from eq. (2) in https://www.elen.ucl.ac.be/Proceedings/esann/esannpdf/es1999-461.pdf, I found that the optimization is equivalent to

```
min 0.5 Σ_m (w_m * w_m) + C Σ_i Σ_m max(0, 2 + (w_m x_i + b_m) - (w_{y_i} x_i + b_{y_i}))
```

so there is a `2` instead of a `1`.  This matches your derivation provided in an earlier comment but I am not sure what the value of Δ is in what you gave there.  Can you clarify a little bit please?  I may have done something wrong.  Also, if you could work that explanation into the comment here, it would make it a lot easier for future maintainers. :+1:  It's currently a little hard to see how to go from equation (2) given in that paper to this formulation here, which uses sparse matrices and sparse linear algebra.

Thanks!

src/mlpack/methods/linear_svm/linear_svm.hpp
rcurtin: Ah, now I realize what you mean.  For some reason I didn't notice that "Weston and Watkins" was linked, sorry!

src/mlpack/methods/linear_svm/linear_svm.hpp
AYESDIE: I'll make the changes here, thanks for pointing this out.

src/mlpack/methods/linear_svm/linear_svm_function_impl.hpp
AYESDIE: Thanks for pointing this, turns out, I've been using `1` as the value of the hyperparameter `Δ` while evaluating the Loss Function. You were right in deriving this. 
```
min 0.5 Σ_m (w_m * w_m) + C Σ_i Σ_m max(0, Δ + (w_m x_i + b_m) - (w_{y_i} x_i + b_{y_i}))
```


However, since I used Δ = 1, my loss function over there was 
```
min 0.5 Σ_m (w_m * w_m) + C Σ_i Σ_m max(0, 1 + (w_m x_i + b_m) - (w_{y_i} x_i + b_{y_i}))
```

The way I'm programmed it over there is: 
```
  // this simply evaluates the score of each class 
  arma::mat scores = parameters * dataset;

  // for each column, only the row(which indicates the correct label)
  // holds the value of the score.
  arma::sp_mat correctScores = scores % groundTruth;

  // this step does alot of things in a single line. 
  // first 
  // arma::repmat(arma::ones(numClasses).t() * correctScores, numClasses, 1)
  // this gives us a matrix where each row of any specific column is having the 
  // value of correct class score for that column.
  // Now,
  arma::mat margin = scores - 
  		arma::repmat(arma::ones(numClasses).t() * 
  		correctScores, numClasses, 1) + 1 - groundTruth;

  // Ignore the '1's here, the Δ is assumed to be 1 here (as of this review).
  // Now
  // the operation above now slightly corresponds to 		
  // Σ_i Σ_m max(0, 1 + (w_m x_i + b_m) - (w_{y_i} x_i + b_{y_i}))
  //
  // The reason I subtracted groundTruth in the end is to make the 
  // margin of the correct label in each column to be `0`, which is because 
  //  Σ_m is taken over every m != y_i. (Equation 2 of the given paper)
  // whereas I added `1` ( Δ Hyperparamtere ) to every element of the margin.
```

I made a small code that shows this operation(just by printing each matrix after every step, nothing more than that) and you can check it out here: https://github.com/AYESDIE/random-code/tree/master/mlpack-svmtest

I'll address these changes pretty soon.

AYESDIE: Also, I used http://cs231n.github.io/linear-classify/#svm for reference.

src/mlpack/methods/linear_svm/linear_svm.hpp
AYESDIE: I'm not sure how I should reference these both using BibTeX, I looked at `gan.hpp` and `dtree.hpp` (for a reference) but it seems like both of those represented in different manner since the later one was published in a book. Could you help me out in this?

AYESDIE:
@rcurtin Thanks for reviewing this and pointing out the missing `Δ Hyperparameter`. I have added the changes and added comments in the implementation for better understanding of what is happening there. I also updated the tests according to it and added one test that checks if the `LinearSVM` works for another value of `Δ`.

src/mlpack/methods/linear_svm/linear_svm_function_impl.hpp
rcurtin: Ah, ok.  I think it is fine to go with Δ = 2, since that keeps it closer to the original formulation.  The link you gave (to the cs231n site) is correct that Δ can take different values and the result will still be generally the same, but there is no direct mapping between Δ and λ; i.e., we aren't guaranteed that we can optimize the same objective function with Δ = 1 and Δ = 2.

src/mlpack/methods/linear_svm/linear_svm_impl.hpp
```diff
@@ -0,0 +1,165 @@
...
+      if (scores(j, i) > maxScore)
+      {
+        maxScore = scores(j, i);
+        labels(i) = j;
+      }
+    }
+
+    // Set maximum probability to zero for next input.
+    maxScore = 0;
+  }
```
rcurtin: Could `arma::index_max()` be used here?  Note that that is only available on Armadillo 7.300.0 and newer, but you can use things like `#if ARMA_VERSION_MAJOR > 7 || ARMA_VERSION_MAJOR == 7 && ARMA_VERSION_MINOR >= 300` to choose which to use.  Eventually as the minimum version of Armadillo increases we can remove the code that doesn't use `index_max()`.

src/mlpack/methods/linear_svm/linear_svm.hpp
rcurtin: `0 and 100` is fine, but can you mark that in the documentation?  `LogisticRegression::ComputeAccuracy()` also functions the same way.

src/mlpack/methods/linear_svm/linear_svm_impl.hpp
```diff
@@ -0,0 +1,165 @@
...
+  Classify(testData, labels);
+
+  // Increment count for every correctly predicted label.
+  size_t count = 0;
+  for (size_t i = 0; i < labels.n_elem ; i++)
+    if (testLabels(i) == labels(i))
+      count++;
+
+  // Return the accuracy.
+  return (double)count / labels.n_elem;
```
rcurtin: Are you sure this is between 0 and 100?  This looks like it is between 0 and 1.

src/mlpack/tests/linear_svm_test.cpp
```diff
@@ -0,0 +1,1129 @@
...
+      // Compute numerical gradients using the costs calculated above.
+      numGradient1 = (costPlus1 - costMinus1) / (2 * epsilon);
+      numGradient2 = (costPlus2 - costMinus2) / (2 * epsilon);
+
+      // Restore the parameter value.
+      parameters(i, j) += epsilon;
+
+      // Compare numerical and backpropagation gradient values.
+      BOOST_REQUIRE_SMALL(numGradient1 - gradient1(i, j), 1e-2);
+      BOOST_REQUIRE_SMALL(numGradient2 - gradient2(i, j), 1e-2);
```
rcurtin: Ah, I see you already have a numeric gradient test. :+1:  This is nice, but do you think we can reduce the tolerance to something more like `1e-3` or `1e-4`?

src/mlpack/tests/linear_svm_test.cpp
```diff
@@ -0,0 +1,1129 @@
...
+    //  Corresponding labels
+    arma::Row<size_t> labels = "1 0 1";
+
+    // Create a linear svm object using L-BFGS optimizer.
+    LinearSVM<arma::mat> lsvm(dataset, labels, numClasses, lambda,
+        delta, ens::L_BFGS());
+
+    // Compare training accuracy to 1.
+    const double acc = lsvm.ComputeAccuracy(dataset, labels);
+    BOOST_REQUIRE_CLOSE(acc, 1.0, 0.5);
```
rcurtin: Looks like the spacing is a little off here.

src/mlpack/tests/linear_svm_test.cpp
```diff
@@ -0,0 +1,1129 @@
...
+    labels[i] = size_t(0.0);
+  for (size_t i = 500; i < 1000; ++i)
+    labels[i] = size_t(1.0);
+
+  LinearSVM<arma::mat> lsvm(dataset.n_rows, 2);
+  LinearSVM<arma::mat> lsvm2(dataset.n_rows, 2);
+  lsvm.Parameters() = lsvm2.Parameters();
+  ens::L_BFGS optimizer;
+  lsvm.Train(dataset, labels, 2, std::move(optimizer));
+  lsvm2.Train(dataset, labels, 2, std::move(optimizer));
```
rcurtin: I'm not sure what this test is testing, can you clarify please?

src/mlpack/tests/linear_svm_test.cpp
```diff
@@ -0,0 +1,1129 @@
...
+  for (size_t i = (4 * points) / 5; i < points; i++)
+  {
+    data.col(i) = g5.Random();
+    labels(i) = 4;
+  }
+
+  arma::Row<size_t> predictions;
+  lsvm.Classify(data, predictions);
+
+  BOOST_REQUIRE_GE((double) arma::accu(predictions == labels), 450);
```
rcurtin: In the test above we require every prediction to be right, but here we only require 90% accuracy.  It seems like the same thing is being done in both cases though... can you clarify the reason for the difference please?

src/mlpack/methods/linear_svm/linear_svm_function_impl.hpp
```diff
@@ -0,0 +1,363 @@
...
+      { val = (val > 0) ? 1: 0; });
+
+  arma::mat difference = groundTruth
+      % (-arma::repmat(arma::sum(mask), numClasses, 1)) + mask;
+
+  gradient = difference * dataset.t();
+  gradient /= dataset.n_cols;
+
+  // Adding the regularization contribution to the gradient.
+  gradient += lambda * parameters;
```
rcurtin: Some optimizers, like Hogwild!, depend on the gradient actually being sparse: http://ensmallen.org/docs.html#sparse-differentiable-separable-functions.  But in this case, whenever `lambda > 0`, the gradient won't be sparse.  I wonder if this means `LinearSVMFunction` is not a great candidate for parallelization in that manner.

(However, it doesn't really matter for your implementation.  I am just pointing it out.)

src/mlpack/methods/linear_svm/linear_svm_function_impl.hpp
```diff
@@ -0,0 +1,363 @@
...
+
+  // Make a matrix where each column holds the value of the
+  // correct class Score.
+  arma::mat correctScores = arma::repmat(arma::ones(numClasses).t()
+      * (scores % groundTruth), numClasses, 1);
+
+  // Evaluates the margin and makes (y_i) term 0 so that
+  // they don't have any contribution in the summation.
+  arma::mat margin = scores - correctScores + delta -
+      (delta * groundTruth);
```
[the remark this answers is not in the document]
AYESDIE: I can do that but it will result in the statement being hard to understand (I'll probably write about it in the comments in a bit more details).

src/mlpack/methods/linear_svm/linear_svm_function_impl.hpp
```diff
@@ -0,0 +1,363 @@
...
+  arma::mat scores = parameters * dataset;
+  arma::mat correctScores = arma::repmat(arma::ones(numClasses).t()
+      * (scores % groundTruth), numClasses, 1);
+  arma::mat margin = scores - correctScores + delta
+      - (delta * groundTruth);
+
+  // The Hinge Loss Function
+  loss = arma::accu(arma::clamp(margin.cols(firstId, lastId),
+      0.0, DBL_MAX));
+  loss /= batchSize;
```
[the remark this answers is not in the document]
AYESDIE: Using the columns `firstId` through `lastId` was having several issues,but I cannot recall those. However, I'll go through this again.

src/mlpack/methods/linear_svm/linear_svm_function_impl.hpp
```diff
@@ -0,0 +1,363 @@
...
+ */
+template <typename MatType>
+void LinearSVMFunction<MatType>::InitializeWeights(
+    arma::mat &weights,
+    const size_t featureSize,
+    const size_t numClasses)
+{
+  // Initialize values to 0.005 * r. 'r' is a matrix of random values taken from
+  // a Gaussian distribution with mean zero and variance one.
+  weights.randn(numClasses, featureSize);
```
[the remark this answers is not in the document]
AYESDIE: Thanks for this tip. I think if we use `weights.randn(featureSize, numClasses)`, we will have to use `weights.t()` everywhere else.

src/mlpack/methods/linear_svm/linear_svm_impl.hpp
AYESDIE: I updated it to `0` and `1` since the last time you pointed this out (also updated the tests accordingly).

src/mlpack/tests/linear_svm_test.cpp
AYESDIE: I've been using `BOOST_REQUIRE_CLOSE` and `BOOST_REQUIRE_SMALL`with several values for tolerance including `1e-5`,`1e-4`,`1e-3`,`1e-2` and they are listed in decreasing order of the test failure (ultimately resulting in build failures). For the tolerance = `1e-2`, the build passes. 
(I think the tests on the commit history got removed since I had to rebase this with `master` after #1713). 
I will re-evaluate this but do you have a strategy that might help?

src/mlpack/tests/linear_svm_test.cpp
AYESDIE: It tests the stability of LinearSVM. It checks whether the parameters of both `lsvm` and `lsvm2` are close enough or not for the same `dataset` and `labels`.

src/mlpack/methods/linear_svm/linear_svm_function_impl.hpp
```diff
@@ -0,0 +1,363 @@
...
+template <typename MatType>
+double LinearSVMFunction<MatType>::Evaluate(
+    const arma::mat& parameters)
+{
+  // The objective function is the hinge loss function and it is
+  // calculated over all the training examples.
+
+  // Calculate the loss and regularization terms.
+  // L_i = Σ_i Σ_m max(0, Δ + (w_m x_i + b_m) - (w_{y_i} x_i + b_{y_i}))
+  // where (m != y_i)
```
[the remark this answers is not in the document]
AYESDIE: I think we could also do something like a `fitIntercept` for this (Similar to that in`SoftmaxRegressionFunction`). But I'll also go through your suggestion about it.

src/mlpack/methods/linear_svm/linear_svm_function_impl.hpp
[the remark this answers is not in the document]
AYESDIE: Ok, I just recalled the issue. On using `cols()` on a `arma::sp_mat`, everything turns to 0. 

I had tried to use the way you suggested way back but had to do the current implementation because of this.

```
  std::cout << "Before:\n" << arma::conv_to<arma::mat>::from(groundTruth);
  std::cout << "After:\n" << arma::conv_to<arma::mat>::from(groundTruth.cols(0,2));
```

```
Before:
        0   1.0000        0
   1.0000        0   1.0000
After:
        0        0        0
        0        0        0
```

There is no issue with `conv_to`. I've checked that.

src/mlpack/methods/linear_svm/linear_svm_function_impl.hpp
```diff
@@ -0,0 +1,363 @@
...
+  arma::mat margin = scores - correctScores + delta
+      - (delta * groundTruth);
+
+  // For each sample, find the total number of classes where
+  // ( margin > 0 ).
+  arma::mat mask = margin.for_each([](arma::mat::elem_type& val)
+      { val = (val > 0) ? 1: 0; });
+
+  arma::mat difference = groundTruth
+      % (-arma::repmat(arma::sum(mask), numClasses, 1)) + mask;
```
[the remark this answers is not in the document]
AYESDIE: The gradient is evaluated as follows:
- Add `x_i` to `w_j` if `margin_i_j`is positive.
- Subtract `x_i` from `w_y_i` for each positive `margin_i_j`.
- Take the average over the size of dataset.
- Add the regularization parameter.

Also, I've updated the comments to address this.

src/mlpack/methods/linear_svm/linear_svm_impl.hpp
AYESDIE: I didn't go through this but I tried the program that was available in the documentation of `arma::index_max()`. And it basically gave the linear index of the maximum element inside the matrix (could have realized that if I had just read the documentation, but I kinda skipped that, and ran the example code :rofl:). 
I have not given this a complete thought but I think we will have to use `arma::index_max()` on subsets of scores to find the max index of each column of the `dataset`. I think we can have this code for now, but it can definately be implemented in the way you suggested (Also, since `SoftmaxRegression` uses similar code, we can use `arma::index_max()` over there too).

src/mlpack/tests/linear_svm_test.cpp
AYESDIE: Seems like I added a redundant test case here.

src/mlpack/methods/linear_svm/linear_svm_function_impl.hpp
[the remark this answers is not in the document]
AYESDIE: I'll be honest with you, it was not as easy as replacing every `parameter` with `parameter.t()`. But it was fun looking for the reasons of failure of the tests. It works fine now.

AYESDIE:
@rcurtin Thanks for the last review, I went through every comment you made and tried to fix and improve most of the things you had pointed out.
I will list out the changes from the time of you last review:
  - Added the reference to both the research papers.
  - Added a `fitIntercept` to consider `b_i` terms while evaluation.
    - Added test for `fitIntercept`.
  - Removed redundant test case `LinearSVMClassifyTest`.
  - Implemented `LinearSVMFunctionGradient` test in another way which doesn't fail (cannot guarantee this, but hope so).
  - Refactors:
    - `Evaluate()` and `EvaluateWithGradient()` returns `loss + regularization` directly.
    - `loss` is evaluated in a single line.
    - Added comments for the implementation of `Gradient()`.
    - `scores` is evaluated in a single line.

There were some stuff that you suggested but I didn't address them:
  - Use `arma::index_max()` for `ComputeAccuracy()`.
  - Using subset of `dataset` while evaluation of Loss and Gradient in separable functions.

src/mlpack/tests/linear_svm_test.cpp
rcurtin: I guess I'm not sure why they would not be the same; since you're starting from the same initialization (I think), I wouldn't expect any part of the result to be different, since there's no randomness in the way you have it set up here.  So do you think the test is necessary?

src/mlpack/methods/linear_svm/linear_svm_impl.hpp
rcurtin: I'm pretty sure the standalone `index_max()` will, on matrices, give the index of the maximum element for each row or column: http://arma.sourceforge.net/docs.html#index_min_and_index_max_standalone

(Note that's different than the member `.index_max()` function.)

src/mlpack/methods/linear_svm/linear_svm_function_impl.hpp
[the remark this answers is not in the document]
rcurtin: I can't reproduce this.  I'm using this test program:

```
#include <armadillo>

using namespace arma;

int main()
{
  arma::sp_mat x(2, 3);
  x(1, 0) = 1;
  x(0, 1) = 2;
  x(1, 2) = 3;

  std::cout << "Sparse form:\n";
  x.print();

  std::cout << "\nConstructed into a dense matrix:\n";
  arma::mat z(x.cols(0, 2));
  z.print();

  std::cout << "\nconv_to<arma::mat>::from():\n";
  std::cout << arma::conv_to<arma::mat>::from(x.cols(0, 2));
}
```

That gives 

```
Sparse form:
[matrix size: 2x3; n_nonzero: 3; density: 50.00%]

     (1, 0)         1.0000
     (0, 1)         2.0000
     (1, 2)         3.0000


Constructed into a dense matrix:
        0   2.0000        0
   1.0000        0   3.0000

conv_to<arma::mat>::from():
        0   2.0000        0
   1.0000        0   3.0000
```

What happens if you build and run that test program?

src/mlpack/methods/linear_svm/linear_svm_function_impl.hpp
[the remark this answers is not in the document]
rcurtin: Thanks for the clarification.  Did you double-check that my derivation is right?  It might not be. :)  Also, with the new implementation, were you able to reduce the tolerance of the gradient norm check successfully?

src/mlpack/methods/linear_svm/linear_svm_function_impl.hpp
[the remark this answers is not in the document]
AYESDIE: This is strange, I'll try to run it myself.

src/mlpack/methods/linear_svm/linear_svm_function_impl.hpp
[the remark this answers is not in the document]
AYESDIE: I made a new test to evaluate the Gradient. So I was able to reduce tolerance to `1e-5` on the new test case, but the old test case was failing sometimes for smaller values of tolerance.

src/mlpack/tests/linear_svm_test.cpp
AYESDIE: Well, since it was the first time I'm making all the tests for anything, I took reference from many existing tests, and this was inspired from one of the `SoftmaxRegression` test. They do start from same initialization, so I think this one is redundant.

src/mlpack/methods/linear_svm/linear_svm_impl.hpp
AYESDIE: Oh, ok, sorry for my misinterpretation. I think I read about `.index_max()` member function.

src/mlpack/methods/linear_svm/linear_svm_function_impl.hpp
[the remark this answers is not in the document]
AYESDIE: The code that you sent me gave the same result as yours.

But when I run the same code for `groundTruth` it gives `0` for every value.
```
#include <iostream>
#include <armadillo>

void GetGroundTruthMatrix(
        arma::sp_mat& groundTruth,
        size_t numClasses)
{
  arma::uvec rowPointers(labels.n_elem);
  arma::uvec colPointers(labels.n_elem + 1);

  for (size_t i = 0; i < labels.n_elem; i++)
  {
    rowPointers(i) = labels(i);
    colPointers(i + 1) = i + 1;
  }

  arma::vec values;
  values.ones(labels.n_elem);

  groundTruth = arma::sp_mat(rowPointers, colPointers, values, numClasses,
                             labels.n_elem);
}


int main()
{
  arma::Row<size_t> labels = "1 0 1";

  arma::sp_mat x;
  GetGroundTruthMatrix(labels, x, 2);

  std::cout << "Sparse form:\n";
  x.print();

  std::cout << "\nConstructed into a dense matrix:\n";
  arma::mat z(x.cols(0, 2));
  z.print();

  std::cout << "\nconv_to<arma::mat>::from():\n";
  std::cout << arma::conv_to<arma::mat>::from(x.cols(0, 2));

  return 0;
}
```
This gives
```
Sparse form:
[matrix size: 2x3; n_nonzero: 3; density: 50.00%]

     (1, 0)         1.0000
     (0, 1)         1.0000
     (1, 2)         1.0000


Constructed into a dense matrix:
        0        0        0
        0        0        0

conv_to<arma::mat>::from():
        0        0        0
        0        0        0
```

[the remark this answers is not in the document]
AYESDIE: Could you test this one out?

src/mlpack/tests/linear_svm_test.cpp
AYESDIE: Removed this test case as of now, and replaced it with a casual test that compares the value evaluated without matrix operations to that calculated with `Gradient()`. If I am able to fix this, I'll include this test.

src/mlpack/methods/linear_svm/linear_svm_function_impl.hpp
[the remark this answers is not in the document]
rcurtin: What version of Armadillo are you using?  When I run that exactly as you gave it, I see

```
Using GroundTruth values

Sparse form:
[matrix size: 2x3; n_nonzero: 3; density: 50.00%]

     (1, 0)         1.0000
     (0, 1)         1.0000
     (1, 2)         1.0000


Constructed into a dense matrix:
        0   1.0000        0
   1.0000        0   1.0000

conv_to<arma::mat>::from():
        0   1.0000        0
   1.0000        0   1.0000


Using Custom values
Sparse form:
[matrix size: 2x3; n_nonzero: 3; density: 50.00%]

     (1, 0)         1.0000
     (0, 1)         2.0000
     (1, 2)         3.0000


Constructed into a dense matrix:
        0   2.0000        0
   1.0000        0   3.0000

conv_to<arma::mat>::from():
        0   2.0000        0
   1.0000        0   3.0000
```

I compiled as `g++ -o arma_test arma_test.cpp -larmadillo`, with four different versions of Armadillo: 8.500.1, 9.200.7, 7.400.3, and 6.700.6, and none of those gave the wrong output for converting to a dense matrix.  They all gave the output above.  I'm using gcc 8.2.0.  I even tried with clang and can't reproduce it.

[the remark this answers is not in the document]
AYESDIE: Uhh, I'm using Armadillo: 9.100.5 and gcc 8.2.0. 
Thanks for checking this out on so many versions of Armadillo and looking into it, I'll look for the issue since it is clear that there is some issue with something on my side.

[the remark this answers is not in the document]
AYESDIE: I might not include any changes to this as of now. I am still having this issue and I'm not able to test this out( the test fails because all the `sp_mat` where I use gets converted to `0`). I've tried reinstalling several versions of Armadillo and this is super strange imo.

src/mlpack/methods/linear_svm/linear_svm_function_impl.hpp
AYESDIE: In the latest commits (after your last evaluation), I've added `fitIntercept` and one of the the function I'm using is `.insert_rows` for the evaluation of `Gradient`. Since `.insert_rows` is not a member function of `arma::sp_mat`, as of now, we cannot use `Hogwild` or any other optimizer that would require gradient to be `sparse`.

I'll still be looking for alternate ways so as we can use Gradient matrix as `arma::sp_mat`.

src/mlpack/methods/linear_svm/linear_svm_function_impl.hpp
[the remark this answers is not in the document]
rcurtin: We cannot merge this until this issue is fixed.  This will be orders of magnitude slowdown for SGD-like optimizers.  So we need to figure out what is happening on your system that is different.

src/mlpack/methods/linear_svm/linear_svm_function_impl.hpp
rcurtin: Yeah, we should avoid `insert_rows()`.  In fact we should avoid that anyway because it will copy the original `gradient` matrix into a newly allocated larger one, which is inefficient.  Try allocating `gradient` with the correct size once and then using, e.g., `gradient.submat(a, b, c, d) = dataset * difference.t()` (you'll have to compute what a, b, c, and d should be).

AYESDIE: So for this, we would have to allocate size for `Gradient` so that we can add that `submat` and then the gradient of `bias terms` in the last row. Am I right?

src/mlpack/methods/linear_svm/linear_svm_function_impl.hpp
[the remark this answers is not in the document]
AYESDIE: I'll try to work it out.

src/mlpack/methods/linear_svm/linear_svm_function_impl.hpp
rcurtin: Yes, you can do that, and you can set the size with `gradient.set_size()`.

src/mlpack/methods/linear_svm/linear_svm_function_impl.hpp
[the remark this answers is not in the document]
rcurtin: Maybe do you want to try in a VirtualBox Linux container or something like this?  Actually, are you on OS X or Linux?  If the OS is different I suppose that might be the issue.  Still extremely strange, I haven't managed to think of anything that could possibly cause this.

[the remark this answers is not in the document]
AYESDIE: @rcurtin I probably looked alot about this issue. Here is everything I have done in details. 
I have a system with Ubuntu 18.04 on it. I have tried using Armadillo version `9.100.5` and version `8.400`. I was wondering if there were some issues with me using `OpenBLAS` or `LAPACK` so I tried them both and couldn't get a proper result (I may have done something wrong here, so it could also be the case). At this point, I pretty much messed up a lot with those libraries and dependencies, and I could not debug anything (I was also confused at this point). So I did a factory reset, installed a fresh copy of Ubuntu and tried it again. As soon as my system was ready, I got the Armadillo from the package manager `$ sudo apt-get install libarmadillo-dev` (so this time, the dependencies for Armadillo were installed by the package manager). At this point the version of Armadillo I have is `8.400`. On running the above program, I did not get the desired results.
After this, I tried to install `Armadillo` on my friend's system which runs on `Linux Arch`. Here, I got the package from the package manager (https://aur.archlinux.org/packages/armadillo/) and all the dependencies were managed by the package manager. Then, I ran the above program and got the same problem.

After all this, I made some changes **only** to separable `Evaluate()`, and checked them by committing them one after the other (by checking the result of the Travis Build).
  - Test Case 1:
    - In the first commit, I made the changes **(only to separable `Evaluate()`)** that were required, like using `dataset.cols(firstId, lastId)` and similar changes to get it working.
    - One important change in this was `arma::conv_to<arma::mat>::from(groundTruth).cols(firstId, lastId)`, which was basically giving me what `groundTruth.cols(firstId, lastId)` was supposed to give me, but which it wasn't.
    - I also made a test case (LinearSVMTestCase1) where I test separable `Evaluate()` and the overall `Evaluate()` on a very small dataset which can be evaluated by hand.
    - Result:
      - Commit: 7613bf9
      - It passes all the tests on my system.
      - Travis build is successful with no errors in test (https://travis-ci.org/mlpack/mlpack/builds/506116717)

  - Test Case 2:
    - Keeping everything else same, I replaced `arma::conv_to<arma::mat>::from(groundTruth).cols(firstId, lastId)` with `groundTruth.cols(firstId, lastId)`.
    - Result:
      - Commit: 828b07a
      - Test `LinearSVMTestCase1` and `LinearSVMFunctionSeparableEvaluate` fails on my system, since they use that `Evaluate()`.
      - These tests also fail in the Travis Build (https://travis-ci.org/mlpack/mlpack/jobs/506178863). Also, they fail with the same values of `hingeLoss` (comparing that of my system and the one showing in logs).

I'm not sure if it was the right way to reproduce this error, but since I got this issue on 2 system 
 (my freshly reinstalled `Ubuntu 18.04` and `Arch Linux`), which had no `Armadillo` installed before me testing it out (and also, got `Armadillo` from the Package Manager in these cases). 

If I did anything wrong here, or if I did any mistake here, please let me know.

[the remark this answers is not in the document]
AYESDIE: @rcurtin could you take a look at this?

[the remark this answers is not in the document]
rcurtin: Yes, it's on my todo list, I haven't had a chance yet.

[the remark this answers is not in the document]
AYESDIE: Ah, alright, I thought that you missed my last comment there, so I just pinged you again. 
Thanks for letting me know :)

src/mlpack/methods/linear_svm/linear_svm_function_impl.hpp
```diff
@@ -0,0 +1,511 @@
...
+  arma::uvec rowPointers(labels.n_elem);
+  arma::uvec colPointers(labels.n_elem + 1);
+
+  // Row pointers are the labels of the examples, and column pointers are the
+  // number of cumulative entries made uptil that column.
+  for (size_t i = 0; i < labels.n_elem; i++)
+  {
+    rowPointers(i) = labels(i);
+    colPointers(i + 1) = i + 1;
+  }
```
rcurtin: Here's the problem.  `colPointers[0]` is never set to anything.  It should actually be set to 0, so you can add `colPointers[0] = 0` before the loop.  And now here is the long story of how I got here.

I could see that the most recent commit could definitely cause `LinearSVMTest` to fail.  I thought a good first step towards solving this would be simply to print what `groundTruth.cols(firstId, lastId)` is in `Evaluate()`.  So I did that, and strangely found that `groundTruth.cols(firstId, lastId)` (with `firstId == lastId == 0`) said it was empty even though if I printed the whole `groundTruth` I saw

```
[matrix size: 2x3; n_nonzero: 3; density: 50.00%]

     (1, 0)         1.0000
     (0, 1)         1.0000
     (1, 2)         1.0000
```

Immediately I thought that perhaps there is a bug in Armadillo.  Specifically, this problem would be most likely in the sparse subview iterators, since the matrix printing functionality should simply be looping over the subview with a `const_iterator` or something.  So I opened up `armadillo_bits/SpSubview_iterators_meat.hpp`, which is one of my least favorite files in the world because it turns out that iterating over a subview of a sparse matrix stored in CSC format is kind of a disaster of complex code.  I added `cout` calls all over the place just so I could get an idea of what was going on, but to my surprise, there was actually no output when I ran the code!

So at this point I pulled out `gdb` and stepped into the call of `groundTruth.cols(firstId, lastId)`.  This call to `cols()` creates an `SpSubview`, so I stepped through the first constructor in `armadillo_bits/SpSubview_meat.hpp`.  That file has the following code:

```
  // There must be a O(1) way to do this
  uword lend     = m.col_ptrs[in_col1 + in_n_cols];
  uword lend_row = in_row1 + in_n_rows;
  uword count   = 0;

  for(uword i = m.col_ptrs[in_col1]; i < lend; ++i)
    {
    const uword m_row_indices_i = m.row_indices[i];

    const bool condition = (m_row_indices_i >= in_row1) && (m_row_indices_i < lend_row);

    count += condition ? uword(1) : uword(0);
    }

  access::rw(n_nonzero) = count;
```

So, when I stepped through this with gdb, I found that `count` ended up being 0 for our single-column submatrix.  But we can clearly see from the full `groundTruth` matrix that it should be 1---the code just counts the number of elements in the subview.  So I looked at the loop initialization variable `m.col_ptrs[in_col1]`---where `in_col1` was 0, and I found that actually `m.col_ptrs[0]` was some gigantic uninitialized value!  But the rest of `m.col_ptrs` were fine.  (The column pointers are part of the CSC format, and `m` here is the full sparse matrix, or `groundTruth` in our case.)

I took a look through the different `SpMat` constructors, to see if I could pinpoint any clear issue or something to look further into---but I saw they all correctly set every `col_ptrs` element to 0 whenever the matrix was constructed or resized.

So next I thought that a possibility was that `m.col_ptrs[0]` could give uninitialized memory if `col_ptrs` was never set to a size (i.e. if the matrix had 0 columns).  This made me look at where `groundTruth` was created, but I saw in the calls to `arma::sp_mat()` (like below) that the size was set reasonably.

Then I realized that we were setting the row indices and column pointers manually in this function, and suddenly it became clear as soon as I looked at this loop what was going on. :)

Once you add the line I suggested to the top of the loop, now the tests pass. :+1:

AYESDIE: Thanks for digging into this. I added the changes that you suggested and added them, and they worked perfectly. 
The fact that printing the entire `groundTruth` was not giving any issue but using `cols` over it was giving weird issues maybe tricked us into thinking that, maybe there was something wrong with Armadillo, but the actual issue was with the function where we manually initialised `groundTruth`.
I'll read your comment a few more times, to understand it properly, since you did alot of stuff here (And the way you debugged was cool :smile_cat:).

**P.S.**: Also, I made the same changes to the same function inside `SoftmaxRegression`, since I took the function from there (and it is not causing issue right now), and it could prevent similar issue from happening from the future if any changes are made to `SoftmaxRegression` or if anyone uses the same function in their implementation.

AYESDIE:
@rcurtin not sure why the Travis-CI is not showing update here, even though the build is successful. 
I added those changes, please review this when you are free :smile_cat:.

rcurtin:
Yeah, looks like it didn't update the label but it did build successfully.  Anyway, thanks for the changes, I'll take a look shortly.

src/mlpack/methods/linear_svm/linear_svm_function_impl.hpp
rcurtin: Wonderful, thanks for making the changes there also. :+1:

src/mlpack/methods/linear_svm/linear_svm.hpp
```diff
@@ -0,0 +1,250 @@
...
+ * Linear SVM can be used for general classification tasks which will work
+ * on multiclass classification. More technical details about
+ * the model can be found from the following:
+ *
+ * @code
+ * @inproceedings{weston1999support,
+ * title        = {Support vector machines for multi-class pattern
+ *                 recognition.},
+ * author       = {Weston, Jason and Watkins, Chris and others},
+ * booktitle    = {Esann},
```
rcurtin: If you can fix the ESANN conference description here, it would be great; otherwise I can do it during merge.

AYESDIE: I'm not sure what needs to be changed here. In a previous review, you said ESANN is not a conference. Do I need to replace it with `The 7th European Symposium on Artificial Neural Networks`?
If any other change is needed here, I could make them but it is also fine if you could do these changes while merging.

rcurtin:
Not sure why mlpack-bot isn't doing its job and marking this as approved... maybe it has to do with the Travis CI build not being finished, so let me restart that...

AYESDIE:
@rcurtin I agree that it may have to do something with the Travis build label. Also, it seems like that Travis build has not started (It has not started at the time on this comment, but it shows queued). 

Also, could you give a response to [this comment](https://github.com/mlpack/mlpack/pull/1648#discussion_r268474821)?

src/mlpack/methods/linear_svm/linear_svm.hpp
rcurtin: Yeah technically it would be "Proceedings of The 7th European Symposiun on Artifical Neural Networks". Anyway I'll handle it during merge.

AYESDIE:
@rcurtin seems like mlpack-bot still didn't approve this PR. Could you take a look at it?

rcurtin:
Since mlpack-bot should have approved it quite a while ago (meaning the PR should have two approvals), I'll just go ahead and merge it now.  Not sure why mlpack-bot didn't approve, but if it happens again with other PRs then it's probably worth digging into.

rcurtin:
Thanks again. :+1: Really happy to see this merged.

rcurtin:
I'm not sure why this isn't marked as merged, but it's merged into master now, so I'll go ahead and close the PR.

AYESDIE:
Thanks :) I will make a PR on Bindings and Benchmark soon.

PR #1649, opened 2019-01-10
Edited mountain_car.hpp to bound to goal position.

KimSangYeon-DGU:
This is a initial PR.

- [x] Edit code
- [ ] Performance check in OpenAI gym.


zoq:
What about using https://github.com/mlpack/mlpack/blob/master/src/mlpack/tests/q_learning_test.cpp  for a simple check?

src/mlpack/methods/reinforcement_learning/environment/mountain_car.hpp
```diff
@@ -90,16 +90,19 @@ class MountainCar
    *
    * @param positionMin Minimum legal position.
    * @param positionMax Maximum legal position.
+   * @param positionGoal Final target position.
    * @param velocityMin Minimum legal velocity.
    * @param velocityMax Maximum legal velocity.
    */
   MountainCar(const double positionMin = -1.2,
               const double positionMax = 0.5,
+              const double positionGoal = 0.45,
```
manish7294: openai uses default value of `goal_position` as 0.5 & `max_position` as 0.6, so I suggest we should be consistent with it.
https://github.com/openai/gym/blob/12e8b763d5dcda4962cbd17887d545f0eec6808a/gym/envs/classic_control/mountain_car.py#L24

KimSangYeon-DGU:
> What about using https://github.com/mlpack/mlpack/blob/master/src/mlpack/tests/q_learning_test.cpp for a simple check?
@zoq Thanks, I'll give it a shot. :)

src/mlpack/methods/reinforcement_learning/environment/mountain_car.hpp
```diff
@@ -183,7 +186,7 @@ class MountainCar
    */
   bool IsTerminal(const State& state) const
   {
-    return std::abs(state.Position() - positionMax) <= 1e-5;
+    return bool(state.Position() >= positionGoal);
```
zoq: Picky comment, I think there is no need to explicitly cast the output.

zoq:
Okay, fine for me.

KimSangYeon-DGU:
If the PR(#1654) is merged, I'll commit the test code for `Mountain Car` to avoid a conflict.

manish7294:
That works for me, I will wait for #1654 to get merged.

KimSangYeon-DGU:
Can you please rebuild the Travis and AppVeyor?? It works on my setup.

src/mlpack/tests/q_learning_test.cpp
```diff
@@ -251,4 +251,69 @@ BOOST_AUTO_TEST_CASE(AcrobotWithDQN)
...
+
+    if (episodes > 1000)
+    {
+      Log::Debug << "Mountain Car with DQN failed." << std::endl;
+      converged = false;
+      break;
+    }
+
+    /**
+     * Set a threshold of -370 to check a convergence.
```
ShikharJ: I think it is simply `check convergence`.

KimSangYeon-DGU:
Hmm... I should try the Travis build with improved version.

src/mlpack/tests/q_learning_test.cpp
ShikharJ: I meant we should just remove the 'a' :)

KimSangYeon-DGU: Haha!! Thanks. :)

Issue #1511, opened 2018-09-20
Buildable and testable documentation and tutorials


rcurtin:
Here are a couple of problems that we currently have:

 * Our documentation and tutorials often go out of date because parameters change and the tutorials don't get updated.
 * Our documentation doesn't have complete programs that a reader can just paste into a terminal, or Python interpreter, or .cpp file and run.
 * Our documentation doesn't have a good introduction to the kinds of problems mlpack can solve, and then a "flowchart" directing users to the kinds of algorithms they might want to solve a certain problem.
 * There's no guarantee that our code examples actually give the output that we claim for them to give.

I am planning to resolve this, so I wanted to open an issue for any discussion if anyone has any opinions.  (If not I'll just do it! ;))  I'd like to use the automatic bindings system so that we write tutorials with generic code snippets for calls to the command-line programs or Python bindings or other bindings in them.  These snippets:

 * will be written in C++, using the `CLI::GetParam()` function to set parameters
 * will define some tests (hopefully using the Boost Unit Test Framework) for the output
 * will have the ability to provide actual output values that can be substituted into the compiled documentation
 * will generate code usage examples for all binding languages that mlpack support, like the .NET documentation: https://docs.microsoft.com/en-us/dotnet/api/microsoft.win32.commondialog?view=netframework-4.7.2 (you can select the language in the upper right hand corner and it updates the code)
 * will be transformed from snippet to code via some CMake compilation into a program that outputs each of the snippets

I'm currently in the process of designing what the actual documentation will look like (in its "raw" form).  There will also need to be some support for C++-only examples.  When I have some examples, I'll post them here for discussion.  I'm hopeful that some better documentation like this will (a) help people understand what mlpack does and how it can help them, (b) ensure that users don't get confused by out-of-date documentation, and (c) drive more traffic to mlpack since the documentation and tutorials contain ready-to-use examples that someone can just copy-paste and use (and then maybe modify a little bit).

It might be worth thinking about notebooks too at some point so someone can play with mlpack entirely in a browser without downloading anything or setting it up, but I want to keep the scope somewhat limited so this doesn't take me a year to do nicely.  Maybe that can happen later. :)

If anyone's interested in helping (whether that's helping design, writing part of the code, or writing the revised documentation), I am sure that we can collaborate and work together on it! :)

rcurtin:
This is also related to #175.

conradsnicta:
[...]

PR #1653, opened 2019-01-12
Markdown bindings

rcurtin:
This is a big PR and I've been working on it since November.  Essentially I set out to make the mockup in #1511 (http://www.mlpack.org/docs/experimental/python.html) into an automatically-generated reality.  I used the experience I got working with Markdown and Jekyll for the ensmallen docs and decided a way I can do this is to create an automatic binding type that will output Markdown, which can then be dumped into a Jekyll site and turned into usable HTML.

Unfortunately in order to actually make this work right, I had to output a good amount of HTML and Kramdown-specific syntax in the Markdown.  So the generated Markdown isn't exactly human-readable without a bit of effort.  But it should render correctly with most Markdown-to-HTML tools.

Some important notes:

1. `PROGRAM_INFO()` now takes four arguments: name, short description, long description, and a list of `SEE_ALSO()`s.  I've adapted all the existing bindings accordingly.

2. The Markdown bindings can be built with `-DBUILD_MARKDOWN_BINDINGS=ON` and then `make markdown`, which will put everything in `build/doc/`.

3. I had to add a good amount of functionality to the CLI and Python binding types so that everything needed for the Markdown printing is there.

4. Example output after processing with Markdown (ignore the ensmallen header, that will get changed): http://www.ratml.org/mlpack_test/docs.html

5. I'll probably rewrite the mlpack website using Markdown and Jekyll using the same base as ensmallen.org, and then wrap this step into the build process.  I also plan to use some Javascript there to inject a second drop-down menu in the top left, which will allow selection of the version.


zoq:
Is the color choice intended?

cli_test

rcurtin:
Nope, that's just CSS on the Jekyll side that I forgot to fix.  I'll definitely have a better version of that for the website, but I imagined that most of the CSS will probably live in the mlpack.org repo.

PR #1654, opened 2019-01-12
Changed the file name `acrobat.hpp` to `acrobot.hpp`.

KimSangYeon-DGU:
Please refer to [this](https://gym.openai.com/envs/#classic_control). The file name `acrobat.hpp` should be changed to `acrobot.hpp`.


zoq:
Nice catch, to provide backward compatibility, we should add an alias, either by using ``typedef`` or ``using``.

KimSangYeon-DGU:
> Nice catch, to provide backward compatibility, we should add an alias, either by using `typedef` or `using`

Agreed. but which file should we add it to?

ShikharJ:
Thanks for the contribution!

PR #1656, opened 2019-01-13
Improve ELU Implementation

ShikharJ:
Follows from #1540.

### DIFF
--- a/src/mlpack/methods/ann/layer/elu.hpp
+++ b/src/mlpack/methods/ann/layer/elu.hpp
@@ -90,6 +90,10 @@ namespace ann /** Artificial Neural Network. */ {
  * }
  * @endcode
  *
+ * In the deterministic mode, there is no computation of the derivative.
+ *
+ * @note During training deterministic should be set to false and during
+ *       testing/inference deterministic should be set to true.
  * @note Make sure to use SELU activation function with normalized inputs and
  *       weights initialized with Lecun Normal Initialization.
  *
@@ -137,7 +141,7 @@ class ELU
    * f(x) by propagating x backwards through f. Using the results from the feed
    * forward pass.
    *
-   * @param input The propagated input activation.
+   * @param input The propagated input activation f(x).
    * @param gy The backpropagated error.
    * @param g The calculated gradient.
    */
@@ -205,28 +209,30 @@ class ELU
   /**
    * Computes the first derivative of the activation function.
    *
+   * @param x Input data.
    * @param y Propagated data f(x).
    * @return f'(x)
    */
-  double Deriv(const double y)
+  double Deriv(const double x, const double y)
   {
-    return (y > 0) ? lambda : y + lambda * alpha;
+    return (x > 0) ? lambda : y + lambda * alpha;
   }
 
   /**
    * Computes the first derivative of the activation function.
    *
-   * @param x Input activations.
-   * @param y The resulting derivatives.
+   * @param x Input data.
+   * @param y Output activations f(x).
+   * @param z The resulting derivatives.
    */
   template<typename InputType, typename OutputType>
   void Deriv(const InputType& x, OutputType& y)
   {
-    y = x;
+    derivative.set_size(arma::size(x));
 
     for (size_t i = 0; i < x.n_elem; i++)
     {
-      y(i) = Deriv(x(i));
+      derivative(i) = Deriv(x(i), y(i));
     }
   }
 
@@ -235,6 +241,9 @@ class ELU
 
   //! Locally-stored output parameter object.
   OutputDataType outputParameter;
+
+  //! Locally stored first derivative of the activation function.
+  arma::mat derivative;
 
   //! ELU Hyperparameter (0 < alpha)
   //! SELU parameter fixed to 1.6732632423543774 for normalized inputs.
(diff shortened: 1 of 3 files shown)


src/mlpack/methods/ann/layer/elu.hpp
```diff
@@ -205,19 +205,19 @@ class ELU
...
    * Computes the first derivative of the activation function.
    *
-   * @param x Input data.
+   * @param y Propagated data f(x).
    * @return f'(x)
    */
   double Deriv(const double y)
   {
-    return (y > 0) ? lambda : lambda * (y + alpha);
+    return (y > 0) ? lambda : y + lambda * alpha;
```
ShikharJ: The issue here is that `Deriv()` gets called from within the `Backward()` function, where `y` aliases `f(x)`. Hence the implementation goes slightly wrong by getting evaluated as `(f(x) > 0) ? lambda : f(x) + lambda * alpha;` whereas it should ideally be evaluated as `(x > 0) ? lambda : f(x) + lambda * alpha;`. I'm not sure how we can pass the original input `x` through `Backward()`. Suggestions are welcome.

ShikharJ: Also I think the previous implementation got lucky as `lambda` is only about `1.05...` numerically, and hence did not give a failure.

KimSangYeon-DGU: I think we can get x from f(x) using f^{-1}=ln(\frac{x}{\lambda\cdot \alpha}+1).

ShikharJ: That would be computationally expensive, especially when we're using a large matrix for storing the numbers and for each number, the inverse would have to be computed. For an `N x N` matrix, that means `O(N^2)` extra computations, practically anything greater than `N = 10^4` should be noticeably slower. I was thinking more along the lines of locally storing the latest input of the `Forward()` pass, since `Backward()` would ideally not be called without the `Forward()` routine. @rcurtin ?

zoq: Another solution might be to check the ``deterministic`` parameter and compute the derivative right away if necessary, that way we could avoid the extra copy. The solution also works for recurrent networks.

ShikharJ: @zoq Do you mean calling and saving the derivative from within the `Forward()` routine?

zoq: Right if we are in train mode, I think you did something similar for the dropout class.

ShikharJ:
Hmm, this would also require a bit of change to the test infrastructure of activation functions. I'll incorporate the changes here itself.

ShikharJ:
This is ready for review.

src/mlpack/methods/ann/layer/elu_impl.hpp
```diff
@@ -49,15 +52,18 @@ void ELU<InputDataType, OutputDataType>::Forward(
     const InputType&& input, OutputType&& output)
 {
   Fn(input, output);
+
+  if (not deterministic)
```
rcurtin: We don't use the `not` keyword anywhere else (actually I didn't even know until just now it was supported in C++), so maybe for consistency we should go with `!` here instead?

src/mlpack/methods/ann/layer/elu.hpp
```diff
@@ -206,27 +210,29 @@ class ELU
    * Computes the first derivative of the activation function.
    *
    * @param x Input data.
+   * @param y Propagated data.
```
rcurtin: Can you make this comment just a little more clear, in that the propagated data is actually equal to `f(x)`?

src/mlpack/methods/ann/layer/elu_impl.hpp
rcurtin: I didn't know that `not` was even a valid C++ keyword. :)  Since we use `!` everywhere else in the code, do you want to make this consistent with that and use `if (!deterministic)`?

ShikharJ: Yeah sure. Even I came across this recently. `C++` even has support for `or`, `and`, `xor` and so on. It's quite useful in technical interviews, makes the actual code look like pseudocode.

src/mlpack/methods/ann/layer/elu.hpp
```diff
@@ -90,6 +90,10 @@ namespace ann /** Artificial Neural Network. */ {
  * }
  * @endcode
  *
+ * In the deterministic mode, there is no computation of the derivative.
+ *
+ * @note During training you should set deterministic to false and during
```
zoq: Perhaps we should adjust the wording here:

``During training deterministic should be set to false and during testing/inference deterministic should be set to true``

Let me know what you think.

src/mlpack/methods/ann/layer/elu.hpp
```diff
@@ -206,27 +210,29 @@ class ELU
...
-   * @param x The resulting derivatives.
+   * @param x Input data.
+   * @param y Output activations f(x).
+   * @param z The resulting derivatives.
    */
   template<typename InputType, typename OutputType>
   void Deriv(const InputType& x, OutputType& y)
   {
-    y = x;
+    derivative = x;
```
zoq: Not sure if ``derivative.set_size(arma::size(y))`` is faster as to copy the value, we did go with the copy strategy for some other functions as well, so perhaps it's interesting to take a closer look into this. No need to include this in the PR.

ShikharJ:
Thanks for the reviews!

PR #1657, opened 2019-01-13
Various test fixes

rcurtin:
I saw the Jenkins build was unstable so I decided to debug the tests.  This contains fixes for five intermittently failing tests:

 * `GMMTrainEMOneGaussian`: instead of checking each element of the trained mean and covariances, we can just check the l2-norm of the differences.
 * `GMMTrainEMMultipleGaussians`: I overhauled this test so that (1) Gaussians in the mixture have a minimum weight of roughly 0.2; (2) multiple trials are allowed to prevent failures due to GMM converging to some local minimum.
 * `WeightedCategoricalLearningTest` and `UnweightedCategoricalLearningTest`: random forests aren't always guaranteed to be better than decision trees, so adjust the tolerance more.
 * `KernelPCASamplingSchemeTest`: here I actually found an issue in `NystreomMethod`, where we divide by the square root of the singular values.  But the singular values can sometimes be zero, giving `nan`s everywhere.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -27,7 +27,7 @@ before_install:
   - sudo cp .travis/config.hpp /usr/include/armadillo_bits/config.hpp
 
 install:
-  - mkdir build && cd build && cmake $CMAKE_OPTIONS .. && make -j2
+  - mkdir build && cd build && cmake $CMAKE_OPTIONS .. && travis_wait 30 make -j2
 
 script:
   - CTEST_OUTPUT_ON_FAILURE=1 travis_wait 30 ctest -j2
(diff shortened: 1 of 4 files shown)


rcurtin:
Yeah, it is true.  I am sick right now, so there's not much outside I can do today.  So I figured, might as well debug some tests. :)

KimSangYeon-DGU:
> I am sick right now

@rcurtin 
I'm so sorry to hear the news.

rcurtin:
> I'm so sorry to hear the news.

@KimSangYeon-DGU thanks.  It seems like a bad cold or something but I am sure I will get over it soon.

zoq:
Not sure if ``travis_wait`` solves the stalled build issue, but we could test it out:

``mkdir build && cd build && cmake $CMAKE_OPTIONS .. && travis_wait 30 make -j2``

PR #1663, opened 2019-01-15
Fix NMF saving and loading of matrices.

### LINKED ISSUE #1646: [NMF] Python binding inverting H and W
The documentation of Non-negative matrix factorization Python binding states:

>  This program performs non-negative matrix factorization on the given dataset,
>  storing the resulting decomposed matrices in the specified files.  For an
>   input dataset V, NMF decomposes V into two matrices W and H such that 
>    V = W * H
>    where all elements in W and H are non-negative.  If V is of size (n x m), then
>    W will be of size (n x r) and H will be of size (r x m), where r is the rank
>    of the factorization (specified by the 'rank' parameter).

However, this is contradicted by the execution of the following code.
```Python
import numpy as np
import mlpack as ml

#Setting dimensions
n = 10
m = 6
r = 2

#Creating W_0, H_0 and V
W_0 = np.random.uniform(low = 0.0, high = 1.0, size = (n, r))
H_0 = np.random.uniform(low = 0.0, high = 1.0, size = (r, m))
V_0 = W_0.dot(H_0)

#Running NMF with the expected input parameter
output_mlpack = ml.nmf(input = V_0, rank = r,
                       initial_w = W_0, initial_h = H_0,
                       seed = 20190101,
                       verbose = True)

>>> Traceback (most recent call last):
>>>
>>> File "<ipython-input-13-f731a0019f9d>", line 4, in <module>
>>> verbose = True)
>>>
>>> File "mlpack/nmf.pyx", line 156, in mlpack.nmf.nmf
>>>
>>> RuntimeError: fatal error; see Log::Fatal output
>>>

#Running NMF with the SWITCHED input parameters
output_mlpack = ml.nmf(input = V_0, rank = r,
                       initial_w = H_0, initial_h = W_0,
                       seed = 20190101,
                       verbose = True)

H = output_mlpack['h']
H.shape
>>> (10, 2)

W = output_mlpack['w']
W.shape
>>> (2, 6)

#Checking Frobenius norms ||W, H_0||_{F} and ||H, W_0||_{F}
numpy.linalg.norm(W_0 - H)
>>> 3.296522366252803e-16
numpy.linalg.norm(H_0 - W)
>>> 2.790263751007313e-16
```

rcurtin:
This fixes #1646, which pointed out that NMF returns transposed W/H matrices from the Python bindings (and this is true of the command-line bindings too).  The solution was to detect whether or not the binding type is going to transpose the matrices (via the new macro BINDING_MATRIX_TRANSPOSED) and use that to control the behavior in `nmf_main.cpp`.

I also made the `NMFMainTest` a little more robust by using non-square matrices for decomposition.

### DIFF
--- a/src/mlpack/methods/nmf/nmf_main.cpp
+++ b/src/mlpack/methods/nmf/nmf_main.cpp
@@ -105,7 +105,13 @@ static void mlpackMain()
   RequireAtLeastOnePassed({ "h", "w" }, false, "no output will be saved");
   RequireNoneOrAllPassed({"initial_w", "initial_h"}, true);
 
-  // Load input dataset.
+  // Load input dataset.  Note that this dataset will typically be transposed on
+  // load, since we are likely receiving it from a row-major language, but we
+  // get it in a column-major form.  Therefore, we're actually decomposing V^T =
+  // W^T * H^T.  Effectively this means we are solving, for the user, V = H*W.
+  // Therefore, we actually have to switch what we are saving, so we will save
+  // the W we get from amf.Apply() as H, and vice versa.  We know if the data is
+  // transposed based on the BINDING_MATRIX_TRANSPOSED macro.
   arma::mat V = std::move(CLI::GetParam<arma::mat>("input"));
 
   arma::mat W;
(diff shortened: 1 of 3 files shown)


src/mlpack/methods/nmf/nmf_main.cpp
```diff
@@ -121,9 +127,15 @@ static void mlpackMain()
     if (CLI::HasParam("initial_w"))
     {
       // Initialization with given W, H matrices.
+#ifdef BINDING_MATRIX_TRANSPOSED
+      GivenInitialization ginit = GivenInitialization(
```
zoq: Do you think it's worth to restructure the code to avoid the duplication of:

```c++
#ifdef BINDING_MATRIX_TRANSPOSED
      GivenInitialization ginit = GivenInitialization(
          std::move(CLI::GetParam<arma::mat>("initial_h")),
          std::move(CLI::GetParam<arma::mat>("initial_w")));
#else
      GivenInitialization ginit = GivenInitialization(
          std::move(CLI::GetParam<arma::mat>("initial_w")),
          std::move(CLI::GetParam<arma::mat>("initial_h")));
#endif
```

e.g. by introducing a RunAMF method?

zoq: Looks like a good improvement to me.

ShikharJ:
Thanks for the contribution!

PR #1664, opened 2019-01-15
Try to use xenial since it is stable.

rcurtin:
Let's see if this works...


rcurtin:
Ok, let's see how it works in practice. :)  I am sure it is only a matter of time until we unearth some problem...

Issue #1658, opened 2019-01-14
Implement wrapper for diagonally-constrained GMM HMMs


rcurtin:
This is a follow-up to issue #1499.  Specifically, there I wrote

> I can write up a nice wrapper class that will do this, and I can also incorporate the precalculation of forward and backward probabilities. We can see how much this improves things. :)

But I was too optimistic, and I don't think I have time to do this.  However I thought it might be a nice task for someone looking to contribute.  So I will try to detail the steps to do this below.  Overall, our goal is to make it easy for users to train Hidden Markov Models with Gaussian Mixture Models where each Gaussian has the constraint that the covariance matrix is diagonal.  Right now, we provide (through the bindings) `discrete`, `gaussian`, and `gmm` but `gmm` has a full covariance matrix, not a diagonal one.  So you could see this proposed `diagonal_gmm` as something that lives "between" `gaussian` and `gmm` in terms of expressivity.

1. Make sure you're at least reasonably familiar with HMMs and GMMs.  I'd suggest playing with the `mlpack_hmm_train`, `mlpack_hmm_viterbi`, `mlpack_hmm_loglik` and `mlpack_hmm_generate` command-line programs (or corresponding Python bindings) to see how a user might use these.  Also familiarize yourself with the `mlpack::hmm::HMM` class and `mlpack::gmm::GMM` class.  It's worth looking at the documentation in `src/mlpack/methods/hmm/hmm.hpp` and `src/mlpack/methods/gmm/gmm.hpp`, as well as the tests in `src/mlpack/tests/hmm_test.cpp` and `src/mlpack/tests/gmm_test.cpp`.

2. Note that we can train a Gaussian mixture model with a diagonal constraint with the following code:

```
GMM gmm(gaussians, dimensionality); // for some value of gaussians and dimensionality
gmm.Train<EMFit<kmeans::KMeans<>, DiagonalConstraint>>(data);
```

But that template parameter `FittingType` to `GMM::Train()` is not a class member.  In addition, because of `EMFit::ArmadilloGMMWrapper()`, when `GMM::Train()` is called with `DiagonalConstraint`, we internally use the Armadillo `gmm_diag` support.  We should write a new class `DiagonalGMM` that entirely wraps the `gmm_diag` class, but provides the same API as the existing `GMM` class.

[...]

PR #1666, opened 2019-01-15
Implement wrapper for diagonally-constrained GMM HMMs.

KimSangYeon-DGU:
Implement wrapper for diagonally-constrained GMM HMMs related to #1658.

**TODO** 

**DONE**
[GMM part]
- [x] Make `gmm_diag.hpp`.
- [x] Make `gmm_diag_impl.hpp`.
- [x] Make `gmm_diag.cpp`.
- [x] Make a test case for `DiagonalGMM`.
- [x] Add some test cases for `DiagonalGMM` in `gmm_test.cpp`.

[HMM part]
- [x] Modify `hmm_model.hpp` to support `DiagonalGMM`.
- [x] Adapt `hmm_train_main.cpp`, `hmm_viterbi_main.cpp`, `hmm_loglik_main.cpp`, and `hmm_generate_main.cpp` to support `DiagonalGMM`.
- [x] Adapt a test case in `hmm_generate_test.cpp`, `hmm_loglik_test.cpp`, `hmm_train_test.cpp` and `hmm_viterbi_test.cpp` to support `DiagonalGMM`.
- [x] Make some test cases for DiagonalGMM HMMs in `hmm_test.cpp`.
- [x] Compare a training performance between `diagonal` and `regular` GMM HMMs.
- [x] Add the test cases `DiagonalGMMHMMOneGaussianLabeledTrainingTest`, and `DiagonalGMMHMMMultipleGaussiansUnlabeledTrainingTest`.

[Editing & Refactoring]
- [x] Use `InitialClusteringType` instead of `FittingType` in `gmm_diag_impl.hpp`.
- [x] Change `r * r.t()` to `r` at line 282 in `hmm_train_main.cpp`.
- [x] Create `DiagCovGaussianDistribution` class that internally can hold `arma::vec` for the mean, an `arma::mat` for the diagonal covariance entries, and the inverse of each covariance entry.
- [x] Add test cases for `DiagCovGaussianDistribution` class.
- [x] Edit `DiagonalGMM` and `EMFit` classes to support `DiagCovGaussianDistribution`.
- [x] Edit and refactor the codes according to the reviews.
- [x] Add the comment about `DiagonalConstraint` will be deprecated in `HISTORY.md`.
- [x] Change the class name `DiagCovGaussianDistribution` to `DiagonalGaussianDistribution`.


src/mlpack/methods/gmm/gmm_diag_impl.hpp
```diff
@@ -0,0 +1,219 @@
...
+ */
+template<typename FittingType>
+double DiagonalGMM::Train(const arma::mat& observations,
+                          const size_t trials,
+                          const bool useExistingModel,
+                          FittingType fitter)
+{
+  // To use armadillo gmm_diag class, we create fitter for DiagonalConstraint.
+  EMFit<typename std::remove_reference<decltype(fitter.Clusterer())>::type,
+        DiagonalConstraint> armaFitter;
```
rcurtin: If this is restricted to using Armadillo's fitter, I'd be fine with using a template parameter `InitialClusteringType` instead of `FittingType`; then you don't need to extract it from `FittingType`. :+1:

src/mlpack/methods/hmm/hmm_train_main.cpp
```diff
@@ -225,6 +260,29 @@ struct Init
...
+      // Random means and covariances.
+      for (int g = 0; g < CLI::GetParam<int>("gaussians"); ++g)
+      {
+        const size_t dimensionality = e[i].Component(g).Mean().n_rows;
+        e[i].Component(g).Mean().randu();
+
+        // Generate random diagonal covariance.
+        arma::mat r = arma::diagmat(arma::randu<arma::mat>(dimensionality,
+            dimensionality));
+        e[i].Component(g).Covariance(r * r.t());
```
rcurtin: In other models the `r * r.t()` was to guarantee that the covariance generated was positive definite, but if we use a diagonal matrix, we're already guaranteed that---so no need for `r * r.t()` here (just `r` is fine).

src/mlpack/methods/gmm/gmm_diag.hpp
```diff
@@ -0,0 +1,299 @@
...
+class DiagonalGMM
+{
+ private:
+  //! The number of Gaussians in the model.
+  size_t gaussians;
+  //! The dimensionality of the model.
+  size_t dimensionality;
+
+  //! Vector of Gaussians
+  std::vector<distribution::GaussianDistribution> dists;
```
rcurtin: So there are some really nice optimizations and speedups that can be made here---we don't need to hold `distribution::GaussianDistribution` here, because that class assumes a full covariance (and thus computes the Cholesky decomposition for faster probability computations).  But since we have a diagonal covariance, we can create a separate `distribution::DiagonalGaussianDistribution` class (or something like that), which internally can hold an `arma::vec` for the mean and an `arma::vec` for the diagonal covariance entries.  It could even hold the inverse of each diagonal covariance entry, so as to make the `Probability()` computation faster.

I remember that there was some other work with GANs where a Gaussian distribution with diagonal covariance was needed, but if I am remembering right at that time it was chosen that we would put their diagonal-specific code in the NN code.  Maybe the creation of a new `DiagonalGaussianDistribution` class could let us refactor that code a bit. :+1:

I'm not tied to the name `DiagonalGaussianDistribution`, by the way, it's just an idea.

src/mlpack/methods/gmm/gmm_diag_impl.hpp
KimSangYeon-DGU: As you said, `InitialClusteringType` would be better. Thanks. :)

src/mlpack/methods/gmm/gmm_diag.hpp
KimSangYeon-DGU: Thank you for letting me know, I'm so happy to know that way to optimize our codes. I'll look into it. :)

KimSangYeon-DGU:
Hey @rcurtin. This is ready for a review and below is a list of updates. :)

**[distribution_test.cpp]**
- In `distribution_test.cpp`, I calculated the ground truth values by using R libraries, `dmvnorm`, and `cov.wt`. 

**[diag_cov_gaussian_distribution.hpp/cpp]**
-  For the unbiased estimator in the weighted sample case, I applied a formula to `Train()` that has `probabilities` parameter. Below is the formula. And this is implemented in [R library](https://stat.ethz.ch/R-manual/R-devel/library/stats/html/cov.wt.html) as well.
<p align="center">
<a href="https://www.codecogs.com/eqnedit.php?latex=\mu^{*}&space;=&space;\frac{\sum_{i=1}^{N}w_{i}\textup{x}_{i}}{\sum_{i=1}^{N}w_{i}}" target="_blank"></a>
</p>
<p align="center">
<a href="https://www.codecogs.com/eqnedit.php?latex=\Sigma&space;=&space;\frac{\sum_{i=1}^{N}w_{i}(\textbf{\textup{x}}_{i}-\mu^{*})^{T}(\textbf{\textup{x}}_{i}-\mu^{*})}{1-V_{2}}" target="_blank"></a>
</p>

-  <a href="https://www.codecogs.com/eqnedit.php?latex=w_{i}:" target="_blank"></a> each weight 
<a href="https://www.codecogs.com/eqnedit.php?latex=\textup{x}_i:" target="_blank"></a> each observation row vector 
<a href="https://www.codecogs.com/eqnedit.php?latex=\mu^{*}:" target="_blank"></a> weighted mean
<a href="https://www.codecogs.com/eqnedit.php?latex=V_{2}:" target="_blank"></a> the sum of each weight squared 
For more details, please refer to [Wikipedia](https://en.wikipedia.org/wiki/Weighted_arithmetic_mean)

**[em_fit.hpp/cpp]**
- I added a template paramter `DistributionType` to the `EMFit` class. And I set the default to `distribution::GaussianDistribution` for the backward compatibility.


**TODO**
Now, although I make some test cases in `main_tests/hmm_*_test.cpp`, I noticed I should make some test cases in `hmm_test.cpp` and then I'll try to compare the performance between `regular` and `digonal` GMM HMMs.

Please, let me know if I should clarify anything. :)

KimSangYeon-DGU:

src/mlpack/tests/gmm_test.cpp
```diff
@@ -894,4 +895,45 @@ BOOST_AUTO_TEST_CASE(DiagonalGMMTrainTest)
...
+
+    arma::vec actualMean = arma::mean(data, 1);
+    arma::mat actualCovar = arma::diagmat(
+        arma::ccov(data, 1 /* biased estimator */));
+
+    // Check the model to see that it is correct.
+    CheckMatrices(gmm.Component(0).Mean(), actualMean);
+    CheckMatrices(gmm.Component(0).Covariance(), actualCovar);
+
+    BOOST_REQUIRE_CLOSE(gmm.Weights()[0], 1.0, 1e-5);
```
rcurtin: I've had a lot of problems in the past with the GMM and HMM tests because they aren't robust to different random seeds (since the data is generated randomly).  You can test against different random seeds by adding `math::RandomSeed(std::time(NULL));` at the top of the test, then compiling and running from a shell a loop like this:

```
$ i=0; while(true); do echo $i; bin/mlpack_test -t GMMTest/DiagonalGMMTrainEMOneGaussian 2>&1 | grep 'fatal error'; sleep 1; i=$(($i + 1)); done
```

I'll often try to make sure that runs for something like 1000 iterations without any failures.  (That can take a little while, so I will start it over lunch in a screen or tmux session or something so I can look at the output when I come back. :))

src/mlpack/tests/main_tests/hmm_generate_test.cpp
```diff
@@ -171,6 +171,54 @@ BOOST_AUTO_TEST_CASE(HMMGenerateGMMHMMCheckDimensionsTest)
...
+
+  // Call to hmm_generate_main
+  mlpackMain();
+
+  // Get the generated observation sequence. Ensure that the generated sequence
+  // has the correct length (as provided in the input).
+  arma::mat obsSeq = CLI::GetParam<arma::mat>("output");
+  BOOST_REQUIRE_EQUAL(obsSeq.n_cols, (size_t) length);
+  BOOST_REQUIRE_EQUAL(obsSeq.n_rows, (size_t) 2);
+  BOOST_REQUIRE_EQUAL(obsSeq.n_elem, (size_t) (length*2));
```
rcurtin: Small spacing issue---this should be `(length * 2)`.

src/mlpack/tests/main_tests/hmm_train_test.cpp
```diff
@@ -165,6 +165,40 @@ inline void ApproximatelyEqual(HMMModel& h1,
...
+
+    BOOST_REQUIRE_EQUAL(d1.size(), d2.size());
+
+    // Check if gaussian, mean, covariance and weights are equal.
+    size_t states = d1.size();
+    for (size_t i=0; i<states; i++)
+    {
+      BOOST_REQUIRE_EQUAL(d1[i].Gaussians(), d2[i].Gaussians());
+      size_t gaussians = d1[i].Gaussians();
+      for (size_t j=0; j<gaussians; j++)
```
rcurtin: Some spacing issues here too.

src/mlpack/methods/hmm/hmm_train_main.cpp
```diff
@@ -62,8 +63,8 @@ PROGRAM_INFO("Hidden Markov Model (HMM) Training",
         "@doxygen/classmlpack_1_1hmm_1_1HMM.html"));
 
 PARAM_STRING_IN_REQ("input_file", "File containing input observations.", "i");
-PARAM_STRING_IN("type", "Type of HMM: discrete | gaussian | gmm.", "t",
-    "gaussian");
+PARAM_STRING_IN("type", "Type of HMM: discrete | gaussian | gmm | gmm_diag.",
+    "t", "gaussian");
```
rcurtin: Maybe we should use the name `diag_gmm` instead?  Also, I'd suggest putting `diag_gmm` before `gmm`, since that means they are labeled in order of ascending complexity/training time.

src/mlpack/methods/gmm/gmm_diag_impl.hpp
```diff
@@ -0,0 +1,221 @@
...
+void DiagonalGMM::serialize(Archive& ar, const unsigned int /* version */)
+{
+  ar & BOOST_SERIALIZATION_NVP(gaussians);
+  ar & BOOST_SERIALIZATION_NVP(dimensionality);
+
+  // Load (or save) the gaussians.  Not going to use the default std::vector
+  // serialize here because it won't call out correctly to serialize() for each
+  // Gaussian distribution.
+  if (Archive::is_loading::value)
+    dists.resize(gaussians);
```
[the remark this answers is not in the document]
KimSangYeon-DGU: > have you tried any serialization test without this call to `resize()`?

Not yet, but I'll try it. If there is any problem, I'll let you know.

src/mlpack/tests/gmm_test.cpp
KimSangYeon-DGU: It was really really helpful to me, thank you so much. :)

src/mlpack/core/dists/diag_cov_gaussian_distribution.hpp
```diff
@@ -0,0 +1,210 @@
...
+  // Now, we only want to calculate the diagonal elements of (diffs' * cov^-1 *
+  // diffs).  We just don't need any of the other elements.  We can calculate
+  // the right hand part of the equation (instead of the left side) so that
+  // later we are referencing columns, not rows -- that is faster.
+  const size_t k = observations.n_rows;
+  const arma::mat rhs = -0.5 * arma::diagmat(invCov) * diffs;
+  arma::vec logExponents(diffs.n_cols);
+
+  for (size_t i = 0; i < diffs.n_cols; i++)
+    logExponents(i) = accu(diffs.unsafe_col(i) % rhs.unsafe_col(i));
```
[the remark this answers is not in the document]
rcurtin: I think that can work, but I'd suggest holding `invCov` as an `arma::vec`, then doing

```
arma::vec logExponents = -0.5 * arma::pow(diffs, 2).t() * arma::diagmat(invCov);
```

src/mlpack/core/dists/diag_cov_gaussian_distribution.cpp
```diff
@@ -0,0 +1,170 @@
...
+    mean += normalizedProbs[i] * observations.col(i);
+
+  // Now find the covariance.
+  for (size_t i = 0; i < observations.n_cols; i++)
+  {
+    arma::vec obsNoMean = observations.col(i) - mean;
+    covariance.diag() += arma::diagvec(
+        normalizedProbs[i] * (obsNoMean * trans(obsNoMean)));
+    v2 += normalizedProbs(i) * normalizedProbs(i);
+  }
```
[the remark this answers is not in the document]
rcurtin: Actually it may be still faster here to avoid instantiating `onesRow`; Armadillo can optimize the `arma::ones<>` call, so you could do

```
arma::mat obsNoMean = observations - mean * arma::ones<arma::rowvec>(observations.n_cols);
```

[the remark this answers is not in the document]
KimSangYeon-DGU: Aha~, Thanks!

src/mlpack/core/dists/diag_cov_gaussian_distribution.hpp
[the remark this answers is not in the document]
KimSangYeon-DGU: Sorry for confusing you. The `Equation` is a example about a single observation like `arma::vec`, but the `Code` is example about multiple observations like `arma::mat`. So, in case of multiple observation(column-wise), Isn't the code correct??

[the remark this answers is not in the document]
rcurtin: Oh, right, sorry!  If you're looking to do elementwise multiplication, shouldn't we use the `%` operator instead of `*`?  i.e. `arma::pow(diffs, 2).t() % invCov`.

KimSangYeon-DGU:
Hi @rcurtin. I edited this PR according to some reviews. This is an update on what has been mainly changed since the second review. Please check this if you have a chance.

**[diag_cov_gaussian_distribution.hpp/cpp]**

- Changed the `arma::mat` to `arma::vec` for diagonal covariance.

**[diagonal_gmm.hpp/cpp]**
- Added three functions `Estimate()`, `InitialClustering()`, and `ArmadilloGMMWrapper()` for fast diagonal covariance computation without any changes in `EMFit` class.

**[gmm_test.cpp]**
- Added some test cases as mentioned in the [above review](https://github.com/mlpack/mlpack/pull/1666#pullrequestreview-201926104).

**[hmm_test.cpp]**
- Added some test cases as mentioned in the [above review](https://github.com/mlpack/mlpack/pull/1666#pullrequestreview-201926104). (I'm not sure I completed the test cases you intended.)

Finally, Here are the speed comparison results.
[HMMs with `diag_gmm` and `gmm`]
I used `mlpack_hmm_train` and stock time series data getting from [Kaggle](https://www.kaggle.com/szrlee/stock-time-series-20050101-to-20171231) (I converted the categorical to numeric manually). I also made a label file by hand for training.

[GMMs]
I tested the speed of `Train()` with probabilities. (without using the Armadillo `gmm_diag` class.). The data that I used was generated randomly using `arma::randu`.
result2


[HMMs]
**States: 5**, **Gaussians: 10**
`diag_gmm`: 0.0604 sec
`gmm`: 1.353 sec
Approximate speed-up 22.4 times.
result

**States: 10**, **Gaussians: 15**
`diag_gmm`: 0.0911 sec
`gmm`: 2.8491 sec
Approximate speed-up 31.3 times.

Please let me know, if I should clarify anything.

KimSangYeon-DGU:
Hi there @rcurtin. I thought about some possible refactoring options for `DiagonalGMM`. The `DiagonalGMM` uses EM algorithm to estimate parameters, and it was implemented in `Diagonal::Estimate()` in `diagonal_gmm.hpp`, not separate class like a `EMFit` class.  But I wonder if that was a proper way.
So, I thought of three possible options.

1. The EM algorithm that was implemented in `DiagonalGMM` moves to `EMFit` class by using flag, e.g. `std::is_same<Distribution, DiagonalGaussianDistribution>` at diagonal covariance calculation parts in `em_fit_impl.hpp`, however, because currently the `EMFit` class uses `GaussianDistribution` class as a distribution, we add a template parameter, e.g. `Distribution` to support various distribution types in `EMFit` class.

2. The second option is to make a `DiagonalEMFit` class for diagonal covariance calculation. If we choose the option, we can have an independent class without reforming `EMFit` class.

3. The last option is to keep the existing changes having an EM algorithm in `diagonal_gmm.hpp`.

Please let me know what you think. :)

src/mlpack/core/dists/diagonal_gaussian_distribution.cpp
```diff
@@ -0,0 +1,155 @@
...
+    LogDeterminant();
+    return;
+  }
+
+  // If their sum is not 1, divide the weights by them to normalize.
+  if (v1 != 1)
+  {
+    normalizedProbs = probabilities / v1;
+    v1 = arma::accu(normalizedProbs);
+  }
```
rcurtin: Actually reading through all this code, it would be faster to do this:

```
double v1 = arma::accu(probabilities);
if (v1 == 0)
{
  covariance += 1e-50;
  InvertCovariance();
  LogDeterminant();
  return;
}

arma::vec normalizedProbs = probabilities / v1;
```

(This avoids a copy of `probabilities` and a call to `arma::accu(normalizedProbs)`.)

src/mlpack/methods/gmm/diagonal_em_fit_impl.hpp
```diff
@@ -0,0 +1,387 @@
...
+    for (size_t j = 0; j < condProb.n_rows; j++)
+    {
+      // Avoid dividing by zero.
+      const double probSum = accu(condProb.row(j));
+      if (probSum != 0.0)
+        condProb.row(j) /= probSum;
+    }
+
+    // Store the sum of responsibilities over all the observations.
+    arma::vec N = arma::trans(arma::sum(condProb));
```
rcurtin: Can you change the name of this so it doesn't start with a capital letter please? :+1:

src/mlpack/methods/gmm/diagonal_em_fit_impl.hpp
```diff
@@ -0,0 +1,387 @@
+/**
+ * @file diagonal_em_fit_impl.hpp
+ * @author Kim SangYeon
+ *
+ * Implementation of EM algorithm for fitting DiagonalGMMs.
```
rcurtin: Can you point out that there are a couple of additional speedups over the usual EMFit here?  It took me a minute to compare the code and see what was different.  Additionally, it's up to you, but you could probably merge the `EMFit` and `DiagonalEMFit` class, using only the template type of the distribution to determine what computation to make for the update steps.

src/mlpack/methods/gmm/diagonal_em_fit_impl.hpp
```diff
@@ -0,0 +1,387 @@
...
+{
+  ar & BOOST_SERIALIZATION_NVP(maxIterations);
+  ar & BOOST_SERIALIZATION_NVP(tolerance);
+  ar & BOOST_SERIALIZATION_NVP(clusterer);
+  ar & BOOST_SERIALIZATION_NVP(constraint);
+}
+
+// Armadillo uses uword internally as an OpenMP index type, which crashes
+// Visual Studio.
+#ifndef _WIN32
```
rcurtin: Hmm, a bigger issue here is whether `ARMA_USE_OPENMP` is enabled on Windows.  If this does still crash Visual Studio, then we could modify `CMakeLists.txt` or `prereqs.hpp` or `core.hpp` to make sure that `ARMA_USE_OPENMP` is not defined when we are on Windows using Visual Studio.

If `ARMA_USE_OPENMP` is not defined, Windows should have no problems, meaning that we could always use the Armadillo implementation.  Unfortunately that would mean there's no reason to keep the handcrafted implementation you put in `DiagonalEMFit`... unless you can show it's faster than Armadillo's, in which case we should always use it. :)

src/mlpack/tests/distribution_test.cpp
```diff
@@ -1141,4 +1141,283 @@ BOOST_AUTO_TEST_CASE(RegressionDistributionTest)
...
+
+  DiagonalGaussianDistribution d;
+
+  // Estimate
+  d.Train(observations, probs);
+
+  BOOST_REQUIRE_CLOSE(d.Mean()(0), 4.5000000000000000, 1e-5);
+  BOOST_REQUIRE_CLOSE(d.Mean()(1), 4.4000000000000004, 1e-5);
+  BOOST_REQUIRE_CLOSE(d.Mean()(2), 3.5000000000000000, 1e-5);
+  BOOST_REQUIRE_CLOSE(d.Mean()(3), 6.7999999999999998, 1e-5);
```
rcurtin: You can probably just use `4.5`, `4.4`, `3.5`, and `6.8` here. :)

src/mlpack/tests/gmm_test.cpp
```diff
@@ -794,43 +795,169 @@ BOOST_AUTO_TEST_CASE(UseExistingModelTest)
   }
 }
 
+/********************************************************/
+/** Diagonal Gaussian Mixture Model(DiagonalGMM) Tests **/
+/********************************************************/
+
+//! Make sure Diagonal::Probability() works correctly in single observation.
```
rcurtin: `DiagonalGMM::Probability()` :)  This seems to happen a couple of times.

src/mlpack/tests/gmm_test.cpp
```diff
@@ -794,43 +795,169 @@ BOOST_AUTO_TEST_CASE(UseExistingModelTest)
...
+  for (size_t iterations = 0; iterations < 4; iterations++)
+  {
+    // Determine random covariance and mean.
+    arma::vec mean;
+    mean.randu(2);
+    arma::vec covar;
+    covar.randu(2);
+
+    arma::mat data;
+    data.randn(2 /* dimension */, 150 * pow(10, (iterations / 3.0)));
```
rcurtin: You can simplify here if you like:

```
arma::vec mean(2, arma::fill::randu);
arma::vec covar(2, arma::fill::randu);
arma::mat data(2, 150 * pow(10, (iterations / 3.0)), arma::fill:randn);
```

src/mlpack/tests/hmm_test.cpp
```diff
@@ -1228,5 +1229,598 @@ BOOST_AUTO_TEST_CASE(DiscreteHMMLoadSaveTest)
...
+  arma::vec actualCovar = arma::diagvec(
+      arma::ccov(observations[0], 1 /* biased estimator */));
+
+  // Check the model to see that it is correct.
+  CheckMatrices(hmm.Emission()[0].Component(0).Mean(), actualMean);
+  CheckMatrices(hmm.Emission()[0].Component(0).Covariance(), actualCovar);
+}
+
+/**
+ * Make sure the unlabeled training works reasonably given a sigle distribution
```
rcurtin: `single` :)

src/mlpack/tests/main_tests/hmm_generate_test.cpp
```diff
@@ -161,7 +161,55 @@ BOOST_AUTO_TEST_CASE(HMMGenerateGMMHMMCheckDimensionsTest)
...
+
+  // Now that we have a trained HMM model, we can use it to generate a sequence
+  // of states and observations - using the hmm_generate utility.
+  // Load the input model to be used for inference and the length of sequence
+  // to be generated.
+  int length = 3;
+  SetInputParam("model", h);
+  SetInputParam("length", length);
+
+  // Call to hmm_generate_main
```
rcurtin: Don't forget a `.` at the end of the sentence. :) (sorry for the pedantic comment)

src/mlpack/methods/gmm/diagonal_em_fit_impl.hpp
KimSangYeon-DGU: Ah, really good idea. :) I'll check it now! :)

src/mlpack/core/dists/diagonal_gaussian_distribution.cpp
```diff
@@ -0,0 +1,155 @@
...
+  const size_t k = observation.n_elem;
+  const arma::vec diff = observation - mean;
+  const arma::vec logExponent = (diff.t() * arma::diagmat(invCov) * diff);
+  return -0.5 * k * log2pi - 0.5 * logDetCov - 0.5 * logExponent(0);
+}
+
+void DiagonalGaussianDistribution::InvertCovariance()
+{
+  // Calculate the inverse of the diagonal covariance.
+  invCov = 1/covariance;
```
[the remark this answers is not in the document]
KimSangYeon-DGU: Sure :)

src/mlpack/core/dists/diagonal_gaussian_distribution.cpp
KimSangYeon-DGU: Thanks for letting me know!! :+1:

src/mlpack/methods/gmm/diagonal_em_fit_impl.hpp
KimSangYeon-DGU: Yeah, I'm going to merge the `EMFit` and `DiagonalEMFit` :)

src/mlpack/methods/gmm/diagonal_em_fit_impl.hpp
KimSangYeon-DGU: Hi @rcurtin, there is an issue, even though we don't define `ARMA_USE_OPENMP`. The ['gmm_diag'](https://gitlab.com/conradsnicta/armadillo-code/blob/9.300.x/include/armadillo_bits/gmm_diag_meat.hpp) class of Armadillo uses `uword` as an index type in `for` loop, regardless of using OpenMP. Visual Studio doesn't allow 'unsigned type index' in `for` loop (error: [C3016](https://docs.microsoft.com/en-us/cpp/error-messages/compiler-errors-2/compiler-error-c3016?view=vs-2017)). So I'm not sure we can always depend on Armadillo's `gmm_diag`.

I'll be merging the `EMFit` and `DiagonalEMFit`, please let me know what you think.

KimSangYeon-DGU:
Hi @rcurtin. I edited the code according to the reviews, especially I merged the `DiagonalEMFit` into `EMFit`. I think it is ready to review. :)

CMakeLists.txt
```diff
@@ -412,16 +412,18 @@ set(MLPACK_LIBRARY_DIRS ${MLPACK_LIBRARY_DIRS} ${Boost_LIBRARY_DIRS})
 add_definitions(-DBOOST_TEST_DYN_LINK)
 
 # Detect OpenMP support in a compiler. If the compiler supports OpenMP, flags
-# to compile with OpenMP are returned and added and the HAS_OPENMP definition is
-# added for compilation.
+# to compile with OpenMP are returned and added and the HAS_OPENMP definition
+# is added for compilation.  Visual Studio doesn't support unsigned type index
+# in for loop, while OpenMP uses uword in for loop.  So to prevent build crash
+# on Windows, disable OpenMP, even though its support was detected.
```
rcurtin: So to be more specific, the problem here actually arises only in Armadillo.  I don't see a need to disable OpenMP on Windows in general, but instead maybe we could have something like this?

```
// or _MSVC?  Other Windows compilers should be fine.
#ifdef _WIN32 && ARMA_USE_OPENMP

use arma::gmm_diag ...

#else 

Log::Warn << "Cannot use arma::gmm_diag on Visual Studio due to OpenMP compilation issues!"
    << " Using slower EMFit::Estimate() instead..." << std::endl;
// use EMFit::Estimate() to estimate a GMM using the current code with DiagonalConstraint

#endif 
```

What do you think?  I think that keeps the codebase relatively simple, and I don't mind if we have to use the slower implementation in just the Visual Studio case.  We can wait for Visual Studio to fix the issue (we will probably wait forever though).  I have some more comments about this below, so take a look at those and that might clarify what I mean a bit. :)

src/mlpack/methods/gmm/em_fit_impl.hpp
```diff
@@ -33,23 +35,20 @@ EMFit<InitialClusteringType, CovarianceConstraintPolicy>::EMFit(
...
 {
-  // Shortcut: if the user is using the DiagonalConstraint, then we will call
-  // out to Armadillo.  But Armadillo uses uword internally as an OpenMP index
-  // type, which crashes Visual Studio, so don't do this on Windows.
-  #ifndef _WIN32
   if (std::is_same<CovarianceConstraintPolicy, DiagonalConstraint>::value)
   {
     ArmadilloGMMWrapper(observations, dists, weights, useInitialModel);
     return;
   }
```
rcurtin: To expand on my earlier thought, here, I think we could do this:

```
if (std::is_same<Distribution, DiagonalGaussianDistribution>::value)
{
  // or _MSVC?
  #ifdef _WIN32 and ARMA_USE_OPENMP
    Log::Warn << "Cannot use arma::gmm_diag on Visual Studio!" ...
  #else 
    ArmadilloGMMWrapper(observations, dists, weights, useInitialModel);
    return;
  #endif 
}
```

The fallback here will just use the rest of this function.

src/mlpack/methods/gmm/em_fit_impl.hpp
```diff
@@ -361,7 +399,13 @@ ArmadilloGMMWrapper(const arma::mat& observations,
...
     {
       means.col(i) = dists[i].Mean();
-      covs.col(i) = dists[i].Covariance().diag();
+
+      // DiagonalGaussianDistribution has diagonal covariance as an arma::vec.
+      if (std::is_same<Distribution,
+          distribution::DiagonalGaussianDistribution>::value)
+        covs.col(i) = dists[i].Covariance();
+      else
+        covs.col(i) = dists[i].Covariance().diag();
```
rcurtin: With the code change I suggested above, this would always be called with `DiagonalGaussianDistribution`, so we could simplify this code a little bit.

rcurtin:
By the way, I did get the website done (or done enough to open a PR).  It's at mlpack.org#6, feel free to take a look and leave any comments if you like.  (But don't feel obligated. :))

src/mlpack/methods/gmm/diagonal_gmm.hpp
```diff
@@ -0,0 +1,308 @@
...
+  size_t Gaussians() const { return gaussians; }
+  //! Return the dimensionality of the model.
+  size_t Dimensionality() const { return dimensionality; }
+
+  /**
+   * Return a const reference to a component distribution.
+   *
+   * @param i index of component.
+   */
+  const distribution::DiagonalGaussianDistribution& Component(size_t i) const {
```
zoq: Can you put the opening ``{``on a new line, just to be consistent with the rest of the codebase.

src/mlpack/methods/gmm/diagonal_gmm.hpp
```diff
@@ -0,0 +1,308 @@
...
+      return dists[i]; }
+  /**
+   * Return a reference to a component distribution.
+   *
+   * @param i index of component.
+   */
+  distribution::DiagonalGaussianDistribution& Component(size_t i) {
```
zoq: See comment above.

CMakeLists.txt
KimSangYeon-DGU: Agreed, It looks good to me. As you mentioned, it keeps the codebase relatively simple. :)

src/mlpack/methods/gmm/em_fit_impl.hpp
KimSangYeon-DGU: Agreed :+1

src/mlpack/core/dists/diagonal_gaussian_distribution.cpp
```diff
@@ -0,0 +1,138 @@
...
+  this->logDetCov = arma::accu(log(covariance));
+  this->covariance = std::move(covariance);
+}
+
+double DiagonalGaussianDistribution::LogProbability(
+    const arma::vec& observation) const
+{
+  const size_t k = observation.n_elem;
+  const arma::vec diff = observation - mean;
+  const arma::vec logExponent = (diff.t() * arma::diagmat(invCov) * diff);
```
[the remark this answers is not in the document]
KimSangYeon-DGU: Ah, Thanks :)

src/mlpack/core/dists/diagonal_gaussian_distribution.cpp
```diff
@@ -0,0 +1,138 @@
...
+    mean.zeros(0);
+    covariance.zeros(0);
+    return;
+  }
+
+  // Calculate the mean.
+  mean = arma::sum(observations, 1);
+
+  // Normalize the mean.
+  mean /= observations.n_cols;
```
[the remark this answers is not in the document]
KimSangYeon-DGU: Thanks for pointing it out.

src/mlpack/core/dists/diagonal_gaussian_distribution.cpp
```diff
@@ -0,0 +1,138 @@
...
+  // If you want to know more detailed description,
+  // please refer to https://en.wikipedia.org/wiki/Weighted_arithmetic_mean
+  double v1 = arma::accu(probabilities);
+
+  // If their sum is 0, there is nothing in this Gaussian.
+  // At least, set the covariance so that it's invertible.
+  if (v1 == 0)
+  {
+    covariance += 1e-50;
+    invCov = 1 / covariance;
```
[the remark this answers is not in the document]
KimSangYeon-DGU: Thanks :)

KimSangYeon-DGU:
Hi @rcurtin, @zoq, thanks for reviewing this and I left a question.

Here is the speed comparison between Armadillo's `gmm_diag` class and `EMFit` class.
Data size: 3x20000
Armadillo's gmm_diag class: 0.1192 sec
EMFit class: 0.5627 sec
The `gmm_diag` was approximately **4.7** times faster than `EMFit`.
result3

Please let me know, if I should clarify anything :)

rcurtin:
Nice, thanks for the comparison. That is about what I expected. So we should always use Armadillo's implementation when we can. :)

KimSangYeon-DGU:
Oops... Sorry, I pushed the close button by mistake on my phone...

> So we should always use Armadillo's implementation when we can. :)

Yeah, the Armadillo's implementation was really fast. :)

src/mlpack/methods/gmm/em_fit_impl.hpp
rcurtin: I have a suggestion above, see what you think of that. :+1:

rcurtin:
Ah, I just thought, do you think we should modify the `gmm_train` program so that when the user specifies `diagonal_covariance`, we use `DiagonalGMM` for training then convert it back to a regular `GMM`?  We should probably do that in a separate PR, this one is so close I don't want to add a bunch more work to it. :)

src/mlpack/methods/gmm/em_fit_impl.hpp
```diff
@@ -33,23 +35,26 @@ EMFit<InitialClusteringType, CovarianceConstraintPolicy>::EMFit(
...
+Estimate(const arma::mat& observations,
+         std::vector<Distribution>& dists,
+         arma::vec& weights,
+         const bool useInitialModel)
 {
-  // Shortcut: if the user is using the DiagonalConstraint, then we will call
-  // out to Armadillo.  But Armadillo uses uword internally as an OpenMP index
-  // type, which crashes Visual Studio, so don't do this on Windows.
-  #ifndef _WIN32
   if (std::is_same<CovarianceConstraintPolicy, DiagonalConstraint>::value)
```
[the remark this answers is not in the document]
KimSangYeon-DGU: Sounds great, I'll do that :)

KimSangYeon-DGU:
> Do you think we should modify the gmm_train program so that when the user specifies diagonal_covariance, we use DiagonalGMM for training then convert it back to a regular GMM?

Yes, we should. I'll open a new PR to work it. :)

src/mlpack/methods/gmm/diagonal_gmm.hpp
```diff
@@ -0,0 +1,312 @@
...
+   * @endcode
+   *
+   * @param observations List of observations to classify.
+   * @param labels Object which will be filled with labels.
+   */
+  void Classify(const arma::mat& observations,
+                arma::Row<size_t>& labels) const;
+
+  /**
+   * Serialize the DigonalGMM.
```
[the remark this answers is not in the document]
KimSangYeon-DGU: Thanks for pointing it out!!

src/mlpack/methods/gmm/diagonal_constraint.hpp
```diff
@@ -30,6 +30,13 @@ class DiagonalConstraint
     covariance = arma::diagmat(arma::clamp(covariance.diag(), 1e-10, DBL_MAX));
   }
 
+  //! Force a covariance vector to be non-zero.
```
rcurtin: Can you clarify the documentation here to point out that the parameter here is just the diagonal of the covariance matrix?  Also I think we need to make these changes to `EigenvalueRatioConstraint` and `PositiveDefiniteConstraint`.  `PositiveDefiniteConstraint` should be easy:

```
/**
 * Apply the positive definiteness constraint to the given diagonal covariance matrix
 * (which is represented as a vector), and ensure each value on the diagonal is at least 1e-50.
 */
static void ApplyConstraint(arma::vec& diagCovariance)
{
  // No need to eigendecompose, since the diagonal is already the eigenvalues.
  arma::uvec eigvalOrder = arma::sort_index(diagCovariance);
  arma::vec eigval = diagCovariance[eigvalOrder];

    // If the matrix is not positive definite or if the condition number is
    // large, we must project it back onto the cone of positive definite
    // matrices with reasonable condition number (I'm picking 1e5 here, not for
    // any particular reason).
    double maxEigval = -DBL_MAX;
    for (size_t i = 0; i < diagCovariance.n_elem; ++i)
      if (diagCovariance[i] > maxEigval)
        maxEigval = diagCovariance[i];

  for (size_t i = 0; i < diagCovariance.n_elem; ++i)
  {
    if ((diagCovariance[i] < 0.0) || ((maxEigval / diagCovariance[i]) > 1e5) ||
        (maxEigval < 1e-50))
    {
      diagCovariance[i] = std::max(maxEigval / 1e5, 1e-50);
    }
  }
}
```

For `EigenvalueRatioConstraint` it's a little more tricky but not much:

```
/**
 * Apply the eigenvalue ratio constraint to the given diagonal covariance matrix (represented as a vector).
 */
void ApplyConstraint(arma::vec& diagCovariance) const
{
  // The matrix is already eigendecomposed but we need to sort the elements.
  arma::uvec eigvalOrder = arma::sort_index(diagCovariance);
  arma::vec eigvals = diagCovariance(eigvalOrder);

  // Change the eigenvalues to what we are forcing them to be.  There
  // shouldn't be any negative eigenvalues anyway, so it doesn't matter if we
  // are suddenly forcing them to be positive.  If the first eigenvalue is
  // negative, well, there are going to be some problems later...
  eigvals = (eigvals[0] * ratios);

  // Reassemble the matrix.
  for (size_t i = 0; i < eigvalOrder.n_elem; ++i)
    diagCovariance[eigvalOrder[i]] = eigval[i];
}
```

Actually I also found a bug with `EigenvalueRatioConstraint`.  I'll have to post a fix in another PR.

KimSangYeon-DGU: Oh! Thanks for clarification.

KimSangYeon-DGU: Hey @rcurtin, in the `PositiveDefiniteConstraint`, do we need the `eigvalOrder` and `eigval`? I'm curious because it is not used in the code.

rcurtin: You're right, I forgot to remove it in my draft.  I've updated the code above now.  :+1:

rcurtin: That code is not guaranteed to be right so it's good that you double-checked it. :)

KimSangYeon-DGU: Sure :)

src/mlpack/methods/gmm/eigenvalue_ratio_constraint.hpp
```diff
@@ -70,12 +70,33 @@ class EigenvalueRatioConstraint
...
+  {
+    // The matrix is already eigendecomposed but we need to sort the elements.
+    arma::uvec eigvalOrder = arma::sort_index(diagCovariance);
+    arma::vec eigvals = diagCovariance(eigvalOrder);
+
+    // Change the eigenvalues to what we are forcing them to be.  There
+    // shouldn't be any negative eigenvalues anyway, so it doesn't matter if we
+    // are suddenly forcing them to be positive.  If the first eigenvalue is
+    // negative, well, there are going to be some problems later...
+    eigvals = eigvals.each_col() % ratios;
```
KimSangYeon-DGU: In the `EigenvalueRatioConstraint`, is it right the bug you said?

rcurtin: The code I wrote doesn't have the bug but basically the eigenvalues are ordered backwards for the full covariance matrix implementation.

KimSangYeon-DGU: When I tested it, `eigvals[0] * ratios` doesn't work.

KimSangYeon-DGU: Well, I'll revert the change at line 73 in `eigenvalue_ratio_constraint.hpp`

src/mlpack/methods/gmm/eigenvalue_ratio_constraint.hpp
```diff
@@ -89,7 +90,7 @@ class EigenvalueRatioConstraint
     // shouldn't be any negative eigenvalues anyway, so it doesn't matter if we
     // are suddenly forcing them to be positive.  If the first eigenvalue is
     // negative, well, there are going to be some problems later...
-    eigvals = eigvals % ratios;
+    eigvals %= ratios;
```
rcurtin: Actually sorry, that should be `eigvals = (eigvals[0] * ratios)` like the diagonal case.  Up to you if you want to fix it here, if not I'll handle it in another PR since `EigenvalueRatioConstraint` is broken anyway.

KimSangYeon-DGU: Oh.. sorry. I'll change it right away. Thanks for pointing it out. :)

KimSangYeon-DGU:
Hi @rcurtin. I'm really happy. :) Without your help, it couldn't be implemented.

zoq:
I'm almost through the PR, really nice work indeed, will look at the rest (tomorrow) as well and comment if I see something.

ShikharJ:
Hmm, good job @KimSangYeon-DGU! I'll take a detailed look as well.

rcurtin:
No problem, I'm happy to wait to merge until you guys think it's good. :+1:

src/mlpack/methods/hmm/hmm_model.hpp
```diff
@@ -39,14 +41,17 @@ class HMMModel
   HMM<distribution::GaussianDistribution>* gaussianHMM;
   //! Not used if type is not GaussianMixtureModelHMM.
   HMM<gmm::GMM>* gmmHMM;
+  //! Not used if type is not DiagonalGaussianMixtureModelHMM.
+  HMM<gmm::DiagonalGMM>* diagGMMHMM;
 
  public:
   //! Construct an uninitialized model.
   HMMModel() :
```
zoq: What about we use a default value for the second constructor and remove this one?

src/mlpack/methods/gmm/em_fit_impl.hpp
```diff
@@ -231,16 +283,28 @@ InitialClustering(const arma::mat& observations,
...
   weights.zeros();
   for (size_t i = 0; i < dists.size(); ++i)
   {
     means[i].zeros(dists[i].Mean().n_elem);
-    covs[i].zeros(dists[i].Covariance().n_rows,
-                  dists[i].Covariance().n_cols);
+    if (isDiagGaussDist)
+      covs[i].zeros(dists[i].Covariance().n_elem);
+    else
+      covs[i].zeros(dists[i].Covariance().n_rows,
```
zoq: Do you mind to wrap this with ``{}`` to improve the readability?

src/mlpack/tests/hmm_test.cpp
```diff
@@ -1228,5 +1229,598 @@ BOOST_AUTO_TEST_CASE(DiscreteHMMLoadSaveTest)
...
+    for (size_t i = 0; i < 1000; i++)
+    {
+      if (predictions[i] != states[i])
+      {
+        success = false;
+        break;
+      }
+    }
+
+    if (success)
```
zoq: Looks like this test is missing a ``BOOST_REQUIRE_EQUAL(success, true);`` check at the end.

src/mlpack/tests/main_tests/hmm_viterbi_test.cpp
```diff
@@ -174,4 +174,69 @@ BOOST_AUTO_TEST_CASE(HMMViterbiGMMHMMCheckDimensionsTest)
...
+  // Make some observations.
+  arma::mat observations(2, 50);
+  arma::Row<size_t> states(50);
+
+  states[0] = 0;
+  observations.col(0) = gmms[0].Random();
+
+  for (size_t i = 1; i < 50; ++i)
+  {
+    double randValue = (double) rand() / (double) RAND_MAX;
```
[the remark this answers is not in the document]
KimSangYeon-DGU: Thanks for pointing it out!!

src/mlpack/methods/gmm/em_fit_impl.hpp
KimSangYeon-DGU: Of course not, I'll wrap this with `{}`

src/mlpack/tests/hmm_test.cpp
KimSangYeon-DGU: Oh.. Thanks!!

KimSangYeon-DGU:
Thanks Marcus, Shikhar! :)

KimSangYeon-DGU:
@ShikharJ. Can you please review this? Actually, after this PR gets merged, I'll have a plan to work conversion between `DiagonalGMM` and `GMM` when training using `mlpack_gmm_train`.

src/mlpack/methods/gmm/em_fit.hpp
```diff
@@ -118,6 +120,8 @@ class EMFit
   //! Modify the covariance constraint policy class.
   CovarianceConstraintPolicy& Constraint() { return constraint; }
 
+  //! Get the distribution.
```
ShikharJ: I guess this comment is indicative of the rest of the code that follows? Or is it something that can be removed?

src/mlpack/methods/gmm/em_fit_impl.hpp
```diff
@@ -33,23 +35,36 @@ EMFit<InitialClusteringType, CovarianceConstraintPolicy>::EMFit(
...
+      return;
+    #endif
+  }
+  else if (std::is_same<CovarianceConstraintPolicy, DiagonalConstraint>::value
+      && std::is_same<Distribution, distribution::GaussianDistribution>::value)
+  {
+    // EMFit::Estimate() using DiagonalConstraint with GaussianDistribution
+    // uses slow implementation.
+    Log::Warn << "EMFit::Estimate() using DiagonalConstraint with "
+        << "GaussianDistribution uses slow implementation, so DiagonalGMM is "
```
ShikharJ: I think it should be written as `makes use of slower implementation ...` in place of `uses slow implementation`.

src/mlpack/core/dists/diagonal_gaussian_distribution.hpp
```diff
@@ -0,0 +1,180 @@
...
+   * Create a Guassian distribution with the given mean and diagonal
+   * covariance.
+   * 
+   * @param mean Mean of distribution.
+   * @param covariance Covariance of distribution.
+   */
+  DiagonalGaussianDistribution(const arma::vec& mean,
+                               const arma::vec& covariance);
+
+  //! Return the dimensionalty of this distribution.
```
ShikharJ: I think it is `dimensionality`.

src/mlpack/core/dists/diagonal_gaussian_distribution.hpp
```diff
@@ -0,0 +1,180 @@
...
+  //! Return the mean.
+  const arma::vec& Mean() const { return mean; }
+
+  //! Return a modifiable copy of the mean.
+  arma::vec& Mean() { return mean; }
+
+  //! Return the covariance matrix.
+  const arma::vec& Covariance() const { return covariance; }
+
+  //! Set the covariance.
```
ShikharJ: Better to add the term `matrix` at the end.

src/mlpack/core/dists/diagonal_gaussian_distribution.hpp
```diff
@@ -0,0 +1,180 @@
...
+  void Covariance(const arma::vec& covariance);
+
+  //! Set the covariance using move assignment.
```
ShikharJ: Same here.

src/mlpack/methods/gmm/diagonal_gmm.cpp
```diff
@@ -0,0 +1,185 @@
...
+
+#include "diagonal_gmm.hpp"
+#include <mlpack/core/math/log_add.hpp>
+
+namespace mlpack {
+namespace gmm {
+
+/**
+ * Create a DiagonalGMM with the given number of Gaussians, each of which have
+ * the specified dimensionality.  The means and covariances will be set to 0.
```
ShikharJ: Extra stray space here after the end of sentence.

ShikharJ: I'd prefer it so, but I'm not sure if other reviewers would.

KimSangYeon-DGU:
Thanks for your advice and comments @ShikharJ. I definitely agree with you.

src/mlpack/methods/gmm/diagonal_gmm.cpp
KimSangYeon-DGU: > but I'm not sure if other reviewers would.

@rcurtin, @zoq Can I get some comments about this? If everyone is fine, I'm good as well.

rcurtin: Sorry for the slow response.  Typically I use two spaces; I think this can make it a little easier to read.  However that's just my personal preference and I don't feel strongly enough about that to codify it.  So, I guess, all that to say, I don't have a strong opinion here. :)

zoq: Same, if we are all fine with it, I guess we should leave it, since it's used all over the place.

KimSangYeon-DGU:
I edited comments and noticed the [conversation](https://github.com/mlpack/mlpack/pull/1749#discussion_r264014523). So, the inline function `LogProbability()` was moved to class itself in the `DiagonalGaussianDistribution`.

src/mlpack/methods/gmm/diagonal_gmm.cpp
ShikharJ: Okay, so this is resolved now.

src/mlpack/tests/hmm_test.cpp
```diff
@@ -1228,5 +1229,600 @@ BOOST_AUTO_TEST_CASE(DiscreteHMMLoadSaveTest)
...
+  {
+    BOOST_REQUIRE_LT(arma::norm(hmm.Emission()[dist].Mean() -
+        hmm2.Emission()[dist].Mean()), 0.1);
+    BOOST_REQUIRE_LT(arma::norm(hmm.Emission()[dist].Covariance() -
+        hmm2.Emission()[dist].Covariance()), 0.2);
+  }
+}
+
+/**
+ * Make sure the unlabeled 1-state training works reasonably given a sigle
```
ShikharJ: It should be `single` here.

src/mlpack/tests/hmm_test.cpp
```diff
@@ -1228,5 +1229,600 @@ BOOST_AUTO_TEST_CASE(DiscreteHMMLoadSaveTest)
...
+ * distribution with diagonal covariance.
+ */
+BOOST_AUTO_TEST_CASE(DiagonalGMMHMMOneGaussianOneStateTrainingTest)
+{
+  // Create a gaussian distribution with diagonal covariance.
```
ShikharJ: Same thing about capitalization of `gaussian`.

src/mlpack/tests/hmm_test.cpp
```diff
@@ -1228,5 +1229,600 @@ BOOST_AUTO_TEST_CASE(DiscreteHMMLoadSaveTest)
...
+        gmms[i].Component(0).Mean()), 0.2);
+
+  // Check the estimated covariances of the each emission distribution.
+  for (size_t i = 0; i < 2; i++)
+    BOOST_REQUIRE_LT(arma::norm(hmm.Emission()[i].Component(0).Covariance() -
+        gmms[i].Component(0).Covariance()), 0.5);
+}
+
+/**
+ * Make sure the labeled training works reasonably given a sigle distribution
```
ShikharJ: Same thing about `single` here.

src/mlpack/tests/hmm_test.cpp
```diff
@@ -1228,5 +1229,600 @@ BOOST_AUTO_TEST_CASE(DiscreteHMMLoadSaveTest)
...
+    boost::archive::xml_iarchive ar(ifs);
+    ar >> BOOST_SERIALIZATION_NVP(hmm2);
+  }
+
+  // Remove clutter.
+  remove("test-hmm-save.xml");
+
+  for (size_t j = 0; j < hmm.Emission().size(); j++)
+  {
+    // Check the number of gaussians.
```
ShikharJ: Here as well.

src/mlpack/tests/main_tests/hmm_train_test.cpp
```diff
@@ -240,6 +274,25 @@ BOOST_AUTO_TEST_CASE(HMMTrainGaussianTest)
   Log::Fatal.ignoreInput = false;
 }
 
+// Make sure that the number of gaussians cannot be less than 0.
```
ShikharJ: And here.

ShikharJ:
Holy cow, it took me more than three hours just to read and understand the code, let alone reason its correctness. I'll need some more time on this.

src/mlpack/methods/gmm/positive_definite_constraint.hpp
```diff
@@ -62,6 +62,32 @@ class PositiveDefiniteConstraint
...
+   * each value on the diagonal is at least 1e-50.
+   */
+  static void ApplyConstraint(arma::vec& diagCovariance)
+  {
+    // If the matrix is not positive definite or if the condition number is
+    // large, we must project it back onto the cone of positive definite
+    // matrices with reasonable condition number (I'm picking 1e5 here, not for
+    // any particular reason).
+    double maxEigval = -DBL_MAX;
+    for (size_t i = 0; i < diagCovariance.n_elem; ++i)
```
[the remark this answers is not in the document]
KimSangYeon-DGU: For the readability, Thanks! :)

src/mlpack/core/dists/diagonal_gaussian_distribution.hpp
```diff
@@ -0,0 +1,156 @@
...
+   */
+  DiagonalGaussianDistribution(const size_t dimension) :
+      mean(arma::zeros<arma::vec>(dimension)),
+      covariance(arma::ones<arma::vec>(dimension)),
+      invCov(arma::ones<arma::vec>(dimension)),
+      logDetCov(0)
+  { /* nothing to do. */ }
+
+  /**
+   * Create a Guassian distribution with the given mean and diagonal
```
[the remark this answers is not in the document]
KimSangYeon-DGU: Oops... Thanks!!

src/mlpack/tests/distribution_test.cpp
```diff
@@ -1141,4 +1141,283 @@ BOOST_AUTO_TEST_CASE(RegressionDistributionTest)
...
+  arma::mat observations("3 5 2 7;"
+                         "2 6 8 3;"
+                         "1 4 2 7;"
+                         "6 8 4 7");
+
+  arma::vec probs("0.3 0.4 0.1 0.2");
+
+  DiagonalGaussianDistribution d;
+
+  // Estimate
```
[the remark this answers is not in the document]
KimSangYeon-DGU: Thanks!!

KimSangYeon-DGU:
@ShikharJ I edited this PR according to the reviews. :)

src/mlpack/core/dists/diagonal_gaussian_distribution.cpp
```diff
@@ -0,0 +1,148 @@
...
+    const arma::mat& observations,
+    arma::vec& logProbabilities) const
+{
+  const size_t k = observations.n_rows;
+
+  // Column i of 'diffs' is the difference between observations.col(i) and
+  // the mean.
+  arma::mat diffs = observations.each_col() - mean;
+
+  // Calculates log of exponent equation in multivariate gaussian
```
ShikharJ: `gaussian` should be capitalized here.

KimSangYeon-DGU:
@ShikharJ Thanks for reviewing this. :)

> Next time, please try breaking your PRs into smaller ones, so that it is easier to provide reviews.

Agreed, I'll do that next time. Thanks for your hard work again.

rcurtin:
Really happy to see this merged. :+1: Thanks again for the hard work on this!

KimSangYeon-DGU:
Thanks, @rcurtin @zoq @ShikharJ :)

Issue #1644, opened 2019-01-09
Compilation Error of "sample-ml-app" in VisualStudio2017 (C++)


NoSaintNick:
I tried to compile the basic "sample-ml-app" (doc/examples) that processes "german.csv" (src/mlpack/tests/data) on my system:

Windows 7 64bit
Visual Studio 2017 for C++ (SDK-Version 10.0.17763.0)
MLPack 3.0.4
Boost 1.66 (used the installer "boost_1_66_0-msvc-14.1-64.exe", the compiled source version wouldn't work)
Armadillo 9.200.6

Because I copy/pasted the tutorial code from http://mlpack.org/docs/mlpack-3.0.4/doxygen/sample_ml_app.html (which is essentially the same) I could narrow down the error to line 79 in "sample-ml-app.cpp" and the usage of K-Fold Cross Validation:

`
KFoldCV<RandomForest<GiniGain, RandomDimensionSelect>, Accuracy> cv(k,
		dataset, labels, numClasses);
`

When this line (and the two following that depend on this declaration) is commented the program works fine, else I get an error message:

>Severity:	Error
>Code:		C3546
>Description:	'...': there are no parameter packs available to expand
>Project:		sample-ml-app
>File:		d:\mlpack\mlpack-3.0.4\build\include\mlpack\core\cv\meta_info_extractor.hpp
>Line:		51

The same line in the code from the website produces two additional errors:

>Severity:	Error
>Code:		C2338
>Description:	The given MLAlgorithm is not supported by MetaInfoExtractor
>Project:		MLPACK_Test_Console
>File:		d:\mlpack\mlpack-3.0.4\build\include\mlpack\core\cv\cv_base.hpp
>Line:		102

>Severity:	Error
>Code:		C2664
>Description:	`
mlpack::cv::KFoldCV<mlpack::tree::RandomForest<mlpack::tree::GiniGain,mlpack::tree::RandomDimensionSelect,mlpack::tree::BestBinaryNumericSplit,mlpack::tree::AllCategoricalSplit,double>,mlpack::cv::Accuracy,arma::mat,void *,void *>::KFoldCV(mlpack::cv::KFoldCV<mlpack::tree::RandomForest<mlpack::tree::GiniGain,mlpack::tree::RandomDimensionSelect,mlpack::tree::BestBinaryNumericSplit,mlpack::tree::AllCategoricalSplit,double>,mlpack::cv::Accuracy,arma::mat,void *,void *> &&)
[...]

PR #1667, opened 2019-01-15
MSVC C3546 workaround

zoq:
MSVC C3546 (there are no parameter packs available to expand) workaround.

### DIFF
--- a/src/mlpack/core/cv/meta_info_extractor.hpp
+++ b/src/mlpack/core/cv/meta_info_extractor.hpp
@@ -38,8 +38,14 @@ template<typename MatType,
          bool NumClasses>
 struct TrainForm;
 
-template<typename PT, typename WT, typename... SignatureParams>
-struct TrainFormBase
+// Due to an internal MSVC compiler bug we can't use two parameter packs.
+// So we have to write multiple TrainFormBase forms.
+//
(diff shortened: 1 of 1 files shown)


KimSangYeon-DGU:
It works on my setup as well. 👍

zoq:
@NoSaintNick tested the patch as well see https://github.com/mlpack/mlpack/issues/1644#issuecomment-455204475

src/mlpack/core/cv/meta_info_extractor.hpp
```diff
@@ -38,8 +38,14 @@ template<typename MatType,
          bool NumClasses>
 struct TrainForm;
 
-template<typename PT, typename WT, typename... SignatureParams>
-struct TrainFormBase
+// Due to an internal MSVC compiler bug we can't use two parameter packs.
+// So we have to write multiple TrainFormBase forms.
```
rcurtin: If Microsoft issues a patch update, we might consider putting this in an `#ifdef` for the Visual Studio compiler version.  The only reason I say that is because if Microsoft is using mlpack as a test for each release of Visual Studio, then we want to make sure that the things that expose bugs get tested so that future versions of Visual Studio won't break on valid C++ code. :+1:

I'm ok with the code either way though, this is just a suggestion.

PR #1669, opened 2019-01-15
Update contribution policy.

rcurtin:
Previously, we used to say that we would merge after one approval and a few days, or after two approvals immediately.  This was to ensure high code quality.  But as time has gone on, it's become kind of clear that this has also made merges pretty slow.

So, I think that we should change the policy so that as soon as something has one approval from someone on the Contributors team it can be merged.  And if others want to review it before it gets merged, it seems like the best idea would just be to post a comment asking to wait until they can review it before merge.

I sent an email about this to some in the contributors team and there was no objection (actually there was not much response at all), so I'll take that to mean that basically everyone would be okay with this. :)

I'll let this sit for a week or two to gather any discussion, and if there's no discussion (or we all agree we should make the change), I'll merge this in and change the repository settings so that one approving review is needed before merge. :+1:

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -20,4 +20,7 @@ mlpack's Contributors team, to ensure that (if applicable):
  * any new functionality is tested and working
 
 Once the pull request is approved by one member of the Contributors team, it can
-be merged.
+be merged.  Members of the Contributors team are encouraged to review pull
+requests that have already been reviewed, and pull request contributors are
+encouraged to seek multiple reviews.  Reviews from anyone not on the
+Contributors team are always appreciated.
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -19,9 +19,5 @@ mlpack's Contributors team, to ensure that (if applicable):
    [Style Guide](http://github.com/mlpack/mlpack/wiki/DesignGuidelines)
  * any new functionality is tested and working
 
-Once the pull request is approved by one member of the Contributors team, it may
-be merged between 3 and 7 days after approval.  This allows other contributors
-and maintainers to have time to also review the PR.  If a pull request has at
-least two approvals from members of the Contributors team, then the pull
-request may be immediately merged.  This applies even if the submitter of the
-PR is a member of the Contributors team.
+Once the pull request is approved by one member of the Contributors team, it can
+be merged.
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -11,7 +11,7 @@ interested in participating in Google Summer of Code, see
 
 ## Pull request process
 
-Once a pull request is submitted, it must be reviewed by at least one member of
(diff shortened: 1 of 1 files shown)


mirraaj:
Hello @rcurtin , I have been contributing to Mlpack for a long time. I think it would be better if we give encourage contributors to take responsibility for different sections. Example, reinforcement learning could be handled by some contributor, neural networks section by someone and so on. If we distribute the work evenly, things will go smoothly and at a quick pace. 

@zoq , @rcurtin , @ShikharJ Please let me know your views about this idea.

rcurtin:
@ShikharJ: I updated the language a little bit to indicate that multiple reviews are encouraged.  Does that capture what you were thinking?

@mirraaj: I think maybe the 'suggested reviewers' feature on Github might be helpful for that.  Or maybe I don't understand quite what you mean?  I don't think that we have parts of the code that are "owned" by specific contributors, although different folks in the community are certainly familiar with different parts of the code.

mirraaj:
@rcurtin, Yeah, I was referring to  'suggested reviewers' feature where people are assigned to review the work as per their familiarity with the topic. That would be a better approach to the problem.

zoq:
@mirraaj Github already implements that feature: https://help.github.com/articles/requesting-a-pull-request-review/ I checked a couple of PR's and the suggestion looks reasonable to me.

rcurtin:
I thought about how we could implement that as a requirement.  And I think what we could do is require two approvals for a PR, and then have a bot come auto-approve any PRs that received one approval 24 hours ago.  We could also have the bot auto-merge at that point, unless it's told not to (or there are merge conflicts, or other comments have been left)?

The reason I'm suggesting a bot here is that right now, when I approve a PR and say "I'll merge it in X days", I then have to write it down on my todo list and remember to come back to it in X days to do the merge.  It would be a little less overhead if I could just approve, and then the PR would be automatically merged if there were no problems or other comments.

What do you think?  Too much overhead, or a reasonable solution?  I can agree that waiting 24 hours is reasonable, but I don't know if we need to codify it.  We could also just put in the contribution policy that contributors should wait 24 hours after approval to merge, but that doesn't solve the todo list problem. :)

zoq:
I think that's a neat idea, personally, the current modification of the merge policy looks too vague to me and could result in more work after something is merged. Automating the process sounds like a good solution, no extra work for the reviewer and everyone has time to join the discussion. Also, I guess if someone does see a reason to merge something right away, that's fine too.

mirraaj:
@zoq, @rcurtin Sounds completely fine to me. It will make contribution efficient for the contributors.

rcurtin:
Ok, I updated the language a little bit to mention that mlpack-bot provides the second review after 24 hours.  I'll merge this tomorrow, and then set the repository settings to require two approvals before merge.  Thanks again everyone for the input! :+1:

PR #1655, opened 2019-01-13
Added test for gmm_train

gauravcr7rm:
This is the test for gmm_train() . I am facing problems for creating test for using kmeans in the same.

PR #1673, opened 2019-01-18
Added test for gmm probability

gauravcr7rm:
Added test for gmm probability


gauravcr7rm:
@rcurtin please see and tell if sample gmm model that i have created is valid or not.
If not that please tell me the way through which i can invoke gmm_train and make a sample gmm model.
Thanks in advance.

manish7294:
@gauravcr7rm Please add the test to cmake file (you can find it at `mlpack/tests/CMakeLists.txt`), so that it can be build. Same goes for #1655.

gauravcr7rm:
I have updated CMakeList to build the tests.
@zoq @rcurtin please tell if some changes are required in the above test.
I have also opened  a pr #1689 for gmm_generate.Please see that as well.

src/mlpack/tests/main_tests/gmm_probability.cpp
```diff
@@ -0,0 +1,88 @@
...
+            << 0 << 4 << 3 << 4 << 8 << 9 << 2 << 5 <<endr;
+
+  SetInputParam("input" , std::move(InputData));
+
+  Log::Fatal.ignoreInput = true;
+  BOOST_REQUIRE_THROW(mlpackMain(), std::runtime_error);
+  Log::Fatal.ignoreInput = false;
+}
+  
+//Make sure that input points are provided
```
zoq: Picky style comment, missing space after ``//`` and missing ``.`` at the end, for more details see https://github.com/mlpack/mlpack/wiki/DesignGuidelines#comments.

src/mlpack/tests/main_tests/gmm_probability_test.cpp
```diff
@@ -0,0 +1,97 @@
...
+{
+  int g = 3;
+  int trials = 2;
+
+
+  arma::mat inputData;
+  if (!data::Load("vc2.csv", inputData))
+  BOOST_FAIL("Unable to load train dataset vc2.csv!");
+  
+  gmm = new GMM((size_t)g , inputdata.n_rows);
```
zoq: Maybe I missed something but I there a reason to use ``new`` here?

gauravcr7rm: Here im trying to initialise a gmm model for which i have used new.

src/mlpack/tests/main_tests/gmm_probability.cpp
gauravcr7rm: I will do the changes suggested

rcurtin:
Oh, sorry, I forgot two things:

1. Can you make sure to add this test to `CMakeLists.txt`?  Right now it is not being tested.
2. Don't worry about the current CI test failure---that doesn't seem to have anything to do with your code.

gauravcr7rm:
@rcurtin what changes are supposed to be done here.

rcurtin:
This is also solved by #1899.

PR #1674, opened 2019-01-18
Dice loss class implementation

Rajiv2605:
@zoq I have added code for the dice loss class implementaton... Please let me know what you think :)


zoq:
Thanks for contribution, can you add a test for the new loss function?

Rajiv2605:
@zoq Is there a way to run the tests locally before pushing?

zoq:
If you make a test suite called ``"TestSuite" (BOOST_AUTO_TEST_SUITE(TestSuite))``, and then build ``mlpack_test`` (``make mlpack_test``), you can run only the tests in that test suite with ``bin/mlpack_test -t TestSuite``.  A specific
  test case called ``'TestCase' (BOOST_AUTO_TEST_CASE(TestCase))`` could be run with ``bin/mlpack_test -t TestSuite/TestCase``.

Rajiv2605:
@zoq @rcurtin Please let me know if anything has to be added :)

src/mlpack/methods/ann/loss_functions/dice_loss.hpp
```diff
@@ -0,0 +1,95 @@
...
+#ifndef MLPACK_METHODS_ANN_LOSS_FUNCTIONS_DICE_LOSS_HPP
+#define MLPACK_METHODS_ANN_LOSS_FUNCTIONS_DICE_LOSS_HPP
+
+#include <mlpack/prereqs.hpp>
+
+namespace mlpack {
+namespace ann /** Artificial Neural Network. */ {
+
+/**
+ * The dice loss performance function measures the network's
```
zoq: Do you think you could add/cite a reference here like: http://campar.in.tum.de/pub/milletari2016Vnet/milletari2016Vnet.pdf?

src/mlpack/methods/ann/loss_functions/dice_loss.hpp
```diff
@@ -0,0 +1,95 @@
...
+   * Serialize the layer.
+   */
+  template<typename Archive>
+  void serialize(Archive& ar, const unsigned int /* version */);
+
+ private:
+  //! Locally-stored output parameter object.
+  OutputDataType outputParameter;
+
+  //! The parameter to avoid overfitting
```
zoq: Missing ``.`` at the end.

src/mlpack/methods/ann/loss_functions/dice_loss_impl.hpp
```diff
@@ -0,0 +1,63 @@
...
+}
+
+template<typename InputDataType, typename OutputDataType>
+template<typename InputType, typename TargetType, typename OutputType>
+void DiceLoss<InputDataType, OutputDataType>::Backward(
+    const InputType&& input,
+    const TargetType&& target,
+    OutputType&& output)
+{
+  output = -2 * (target * (arma::accu(input % input) +
```
zoq: Did you you the expression from the paper?

Rajiv2605:
@zoq I have made all the suggested changes 👍 Please let me know if anything else has to be modified :)

Rajiv2605:
@zoq For implementing dense block and uncertainty map, should I open a new PR or should I use this itself?

zoq:
Let's open a new one. Have to check the derivative, but once that is checked I think this is ready to be merged.

src/mlpack/methods/ann/loss_functions/dice_loss_impl.hpp
zoq: Maybe I missed some clever trick; do we have to take the square of the second term: 

``arma::pow(arma::accu(target % target) + arma::accu(input % input), 2)``

Rajiv2605: Yeah you are right... My bad :sweat_smile:

Rajiv2605:
@zoq I have made the changes :+1: Please let me know if anything else has to be changed :)

src/mlpack/methods/ann/loss_functions/dice_loss_impl.hpp
zoq: Just to make sure we are talking about the same:

dice_loss


So, I think it should be:

``std::pow(arma::accu(target % target) + arma::accu(input % input), 2.0)``?

Rajiv2605: I'm sorry... Silly error on my part... I'll avoid such mistakes in the future :+1:

Rajiv2605:
@zoq I have made the changes... I hope I'm right this time :sweat_smile: Sorry for being careless...

src/mlpack/tests/loss_functions_test.cpp
```diff
@@ -354,4 +355,45 @@ BOOST_AUTO_TEST_CASE(GradientReconstructionLossTest)
...
+  // Test the Forward function. Loss should be 0 if input = target.
+  input1 = arma::ones(10, 1);
+  target = arma::ones(10, 1);
+  loss = module.Forward(std::move(input1), std::move(target));
+  BOOST_REQUIRE_SMALL(loss, 0.00001);
+
+  // Test the Forward function. Loss should be 0.185185185.
+  input2 = arma::ones(10, 1) * 0.5;
+  loss = module.Forward(std::move(input2), std::move(target));
+  BOOST_REQUIRE_SMALL(loss - 0.185185185, 0.00001);
```
zoq: Do you mind to use ``BOOST_REQUIRE_CLOSE`` here, I think the check is easier to get.

src/mlpack/tests/loss_functions_test.cpp
```diff
@@ -354,4 +355,45 @@ BOOST_AUTO_TEST_CASE(GradientReconstructionLossTest)
...
+  }
+  BOOST_REQUIRE_EQUAL(output.n_rows, input1.n_rows);
+  BOOST_REQUIRE_EQUAL(output.n_cols, input1.n_cols);
+
+  // Test the Backward function.
+  module.Backward(std::move(input2), std::move(target), std::move(output));
+  for (double el : output)
+  {
+    // For the 0.5 constant vector we should get -1.2 everywhere.
+    BOOST_REQUIRE_SMALL(el + 0.096, 0.00001);
```
zoq: Do you mind to use ``BOOST_REQUIRE_CLOSE`` here, I think the check is easier to get.

src/mlpack/methods/ann/loss_functions/dice_loss.hpp
```diff
@@ -0,0 +1,107 @@
...
+#define MLPACK_METHODS_ANN_LOSS_FUNCTIONS_DICE_LOSS_HPP
+
+#include <mlpack/prereqs.hpp>
+
+namespace mlpack {
+namespace ann /** Artificial Neural Network. */ {
+
+/**
+ * The dice loss performance function measures the network's
+ * performance according to the dice coeffecient
```
zoq: I think this should be coefficient.

src/mlpack/tests/loss_functions_test.cpp
```diff
@@ -354,4 +355,45 @@ BOOST_AUTO_TEST_CASE(GradientReconstructionLossTest)
...
+  }
+  BOOST_REQUIRE_EQUAL(output.n_rows, input1.n_rows);
+  BOOST_REQUIRE_EQUAL(output.n_cols, input1.n_cols);
+
+  // Test the Backward function.
+  module.Backward(std::move(input2), std::move(target), std::move(output));
+  for (double el : output)
+  {
+    // For the 0.5 constant vector we should get -1.2 everywhere.
+    BOOST_REQUIRE_CLOSE(el, -0.096, 0.00001);
```
zoq: Since you adapted the derivative the expected output changed, see ``-1.2 everywhere ``.

zoq:
Thanks, I found one last issue.

Rajiv2605:
@zoq Do you have any other suggestions? Is it ready to merge?

zoq:
> @zoq I have made the changes 👍 Just curious, how do you find such minute errors manually? 🤔 It's a challenging thing to do I guess :)

Normally, I go through each line :)

src/mlpack/methods/ann/loss_functions/dice_loss.hpp
```diff
@@ -0,0 +1,107 @@
...
+/**
+ * The dice loss performance function measures the network's
+ * performance according to the dice coefficient
+ * between the input and target distributions.
+ *
+ * For more information see the following.
+ *
+ * @inproceedings{milletari2016v,
+ *   title = {V-net: Fully convolutional neural networks
+ *            for volumetric medical image segmentation},
```
ShikharJ: Please use uppercase where necessary.

src/mlpack/methods/ann/loss_functions/dice_loss.hpp
```diff
@@ -0,0 +1,107 @@
...
+ *   author = {Milletari, Fausto and Navab, Nassir and Ahmadi, Seyed-Ahmad},
```
ShikharJ: Please use the same format as given on the `arxiv` [page](https://arxiv.org/abs/1606.04797)

src/mlpack/methods/ann/loss_functions/dice_loss.hpp
```diff
@@ -0,0 +1,107 @@
...
+ *   booktitle = {3D Vision (3DV), 2016 Fourth International Conference on},
```
ShikharJ: I guess this is incomplete?

src/mlpack/methods/ann/loss_functions/dice_loss.hpp
```diff
@@ -0,0 +1,107 @@
...
+ *   pages = {565--571},
```
ShikharJ: Hmm, not sure if two dashes have to be here.

src/mlpack/methods/ann/loss_functions/dice_loss_impl.hpp
```diff
@@ -0,0 +1,63 @@
...
+    arma::accu(target % target)) - 2 * input *
+    (arma::accu(target % input))) / std::pow(arma::accu(target % target)
+    + arma::accu(input % input), 2.0);
+}
```
ShikharJ: I think it also might be useful to include the smoothness parameter in the derivative.

src/mlpack/methods/ann/loss_functions/dice_loss.hpp
Rajiv2605: @ShikharJ I had used the format given in [Google Scholar](https://scholar.googleusercontent.com/scholar.bib?q=info:3hH5LL0SWYsJ:scholar.google.com/&output=citation&scisig=AAGBfm0AAAAAXFFNOY1z9qdSgLlNzWMXUwX0utxRcDfO&scisf=4&ct=citation&cd=-1&hl=en)

zoq: I would go with the ``IEEE`` publication as well, but up to you.

src/mlpack/methods/ann/loss_functions/dice_loss_impl.hpp
ShikharJ: @zoq I'm not sure I understand. What does 'check the derivative' mean? My understanding is that if you are using a smooth factor to prevent the coefficient from getting to zero, and you don't use the same for its derivative, then there's no guarantee that the derivative wouldn't vanish or explode. Hence the suggestion, also it seems plausible to use in this situation, since the denominator is practically the same as the coefficient, just raised to the power of two.

Rajiv2605: Yeah it makes sense to add the `smooth` parameter to prevent the gradient from blowing up... Also the main purpose of the parameter is to prevent overfitting so it might help in the gradient as well. You check it out [here](https://github.com/pytorch/pytorch/issues/1249#issuecomment-337999895).

@zoq @ShikharJ  I have updated the gradient in the latest commit. Please let me know if it is fine. I checked  the tests and it seemed to work fine.

src/mlpack/methods/ann/loss_functions/dice_loss.hpp
Rajiv2605: I have replaced the Google Scholar version with the `arxiv BibTex` version as suggested by @zoq . Please let me know if I should keep it or return back to the Google Scholar version.

src/mlpack/methods/ann/loss_functions/dice_loss.hpp
Rajiv2605: I have updated this with `BibTeX` version from `arxiv`. Please let me know if it is fine.

src/mlpack/methods/ann/loss_functions/dice_loss_impl.hpp
Rajiv2605: @zoq Yeah I checked, it's approximately the same.

Rajiv2605: The only issue is that the gradient isn't `0` when the `input=target` as the `smooth` keeps the gradient `non-zero`.

ShikharJ: @Rajiv2605 In that case, would you mind adding an `eps` parameter (and include it only for the denominators) for the coefficient and the derivative? This is done in other classes as well, you may set the default value from there. That should take care of the issue.

zoq: The derivative could be wrong, wolframalpha might be helpful here; example https://www.wolframalpha.com/input/?i=d%2Fdb+a(b%2Bx)%5E(-1%2Fc)

Rajiv2605: @zoq The derivative is right... I got the same value in wolframalpha

zoq: Do you mind to share the link to the wolframalpha link?

Rajiv2605: > Do you mind to share the link to the wolframalpha link?

Link: https://www.wolframalpha.com/input/?i=d%2Fdb+2ab%2F(b%5E2+%2B+a%5E2)

Rajiv2605: @ShikharJ The tests were made just by referring to the other tests in the file.

Rajiv2605: @zoq @ShikharJ Why do you feel the derivative might be wrong?

zoq: @Rajiv2605 thanks for the link, will take a look at the PR tomorrow, but if the derivative is right, not saying it's not, I think there is nothing else to do here. Don't think we have to check some other implementation to be sure the method is correct.

ShikharJ: @Rajiv2605 I meant a source for the numerical tests. What made you think that the expected value of the result is the one that is mentioned in the comments?

Rajiv2605: I tried calculating manually and also checked by printing the test outputs... Both were approximately the same, maybe differing in the second or the third decimal place... That's how I got those values.

zoq: So I added the smooth factor here: https://www.wolframalpha.com/input/?i=d%2Fdb+(2ab+%2B+s)+%2F+(b%5E2+%2B+a%5E2+%2B+s), and the denominator is the same, but the numerator is different, @Rajiv2605 do you see any issue on my side e.g. regarding the input function?

Rajiv2605: @zoq It's right... The numerator looks different as it doesn't have the summation symbol.

Rajiv2605: @ShikharJ @zoq I checked in Pytorch and I get similar values for gradients. Link: https://colab.research.google.com/drive/15XJBFRV8tzyD147nVVMwqm_h9QvuiEIz

zoq: But if I set ``N = 1`` there is no summation.

zoq: Just checked the notebook, ``-0.0878`` is somewhat different from the value (``-0.08230452``) we compare against.

Rajiv2605: @zoq @ShikharJ Please check the latest commit... The `Backward` function is right now :+1: I'm getting the output I got in Pytorch.

Rajiv2605:
@zoq @ShikharJ The output of the `Backward` function is correct now... Could you help me out with the static checks? Please let me know if any more changes have to be made :)

ShikharJ:
Thanks for this contribution, this looks good to me. I'll let @zoq merge this in in case he has further comments.

src/mlpack/methods/ann/loss_functions/dice_loss_impl.hpp
zoq: Personally, I don't use the results from another implementation as the main check, there is always the possibility the implementation isn't correct or they do something different as we do; so I think checking the derivative is more insightful. Anyway, I rechecked the updated backward step and now the numerator matches as well: https://www.wolframalpha.com/input/?i=(a+*+(b%5E2+%2B+a%5E2+%2B+s)+-+b+*+(2+*+a+*+b+%2B+s))

zoq:
Will merge this one once the static analysis check is fixed (hopefully I can fix the check today).

src/mlpack/methods/ann/loss_functions/dice_loss_impl.hpp
Rajiv2605: > Personally, I don't use the results from another implementation as the main check, there is always the possibility the implementation isn't correct or they do something different as we do; so I think checking the derivative is more insightful. Anyway, I rechecked the updated backward step and now the numerator matches as well: https://www.wolframalpha.com/input/?i=(a+*+(b%5E2+%2B+a%5E2+%2B+s)+-+b+*+(2+*+a+*+b+%2B+s))

Yeah, in the numerator, I had to make a small correction... The `2` should be inside the paranthesis but it was outside before. Hence, it multiplied `smooth` by `2` as well and that was the mistake :)

Rajiv2605:
> Will merge this one once the static analysis check is fixed (hopefully I can fix the check today).

@zoq @ShikharJ Thanks for your reviews :)

Rajiv2605:
@zoq I'll start working on the other two features once this is merged :+1:

zoq:

ShikharJ:
Looks like the check passed. Merging this in.

Issue #1483, opened 2018-07-31
Adapt classifiers to return final objective value from `Train()`


rcurtin:
This is an extension of the PR #1430.  Basically, we need to look through all the mlpack classifiers (anything that implements `Train()`) and see if it is appropriate to return some kind of objective value.  Then we also have to update the documentation, and where applicable, the tests.

This could be a nice first issue for someone looking to contribute.

starlord1311:
@rcurtin can i take up this issue?

rcurtin:
@starlord1311 sure, you are welcome to. :)

yashodeepm:
@rcurtin can you be more specific as to what is an objective value that needs to be returned? Can it something as simple as returning the vector of weight(with reference to a trained model using linear regression) ?

rcurtin:
Each model is optimizing some objective function, and this is what should be returned.  Linear regression is actually a little tricky since the optimization there does not explicitly calculate the objective (which should be the mean squared error I believe), so there it may be needed to explicitly compute it (assuming the computation does not take too long).

vaibhavb007:
Seems like there hasn't been much activity over this bug. Can I take this up?

kss682:
Hi, I am new to mlpack and getting to know the code base.Is this issue taken and I would like to get some help to resolve it.

rcurtin:
Hi @kss682, there is not a PR open for this issue so you are welcome to look at it.  I would suggest first trying to use the library's classifiers (like logistic regression, decision trees, etc.) from C++ to solve some simple machine learning tasks and then this should provide enough background to understand what needs to change in the code.

walragatver:
Hi Can I work on this issue? I am new to mlpack looking forward to contribute in GSOC 19.
 I have seen code of logistic regression.

zoq:
Hello @walragatver the comment above still applies, so if you like to work on the issue, please feel free.

rcurtin:
Since #1678 is merged, I think this issue can be closed.  Thanks again @walragatver! :+1:

PR #1678, opened 2019-01-19
Adapted Classifiers

walragatver:
Hi, 
This is pull Request for #1483 Issue.
I have changed three classifiers 
1) adaboost - returning ztProduct as it's upper bound for training error
2) decision stump - returning the entropy after splitting.
3) decision tree - returning final entropy of the tree . Added some code to get final entropy in recursive decision   tree.
Let me know if I have returned something wrong.

Update One:
Changed Two Classifiers 
1) HMM - In this returning final log likelihood obtained after training using data sequences.
2) Random Forest  - Returning average entropy of all the decision trees trained under the forest
                                 (Not Sure of this)

Update Two:
1) LARS : Returning final absolute maximum correlation after training
2) Linear_regression : Returning least squares error after training.
3) sparse_coding : Returning final minmized objective function value.
4) ANN : Returning objective function from GANS, FNN, RNN, RBM.  

Update Three:
1) local_coordinate_coding : returning final objective value

Also I went through some other classifiers and drawn following conclusions
1) Hoeffding Trees : In this we are using gain to find best split hence returning entropy is not good.
2) Naive Bayes Classifier : In this we are updating the probability and distribution curve and thus we don't need to return anything in this.
3) Perceptron : In this we are doing simple weight update without any objective function hence no need to return anything from it.
4) Softmax Regression : Train function is already modified.
5) GMM : Train function is already modified

I am quite new to these algorithms, please correct me if I have made a mistake .

### DIFF
--- a/src/mlpack/methods/adaboost/adaboost.hpp
+++ b/src/mlpack/methods/adaboost/adaboost.hpp
@@ -140,6 +140,7 @@ class AdaBoost
    * @param data Dataset to train on.
    * @param labels Labels for each point in the dataset.
    * @param learner Learner to use for training.
+   * @return ztProduct an upper bound for training error.
    */
   double Train(const MatType& data,
                const arma::Row<size_t>& labels,
--- a/src/mlpack/methods/adaboost/adaboost.hpp
+++ b/src/mlpack/methods/adaboost/adaboost.hpp
@@ -140,7 +140,7 @@ class AdaBoost
    * @param data Dataset to train on.
    * @param labels Labels for each point in the dataset.
    * @param learner Learner to use for training.
-   * @return ztProduct an upper bound for training error.
+   * @return The upper bound for training error.
    */
   double Train(const MatType& data,
                const arma::Row<size_t>& labels,
--- a/src/mlpack/methods/adaboost/adaboost.hpp
+++ b/src/mlpack/methods/adaboost/adaboost.hpp
@@ -141,7 +141,7 @@ class AdaBoost
    * @param labels Labels for each point in the dataset.
    * @param learner Learner to use for training.
    */
-  void Train(const MatType& data,
+  double Train(const MatType& data,
              const arma::Row<size_t>& labels,
              const size_t numClasses,
              const WeakLearnerType& learner,
--- a/src/mlpack/methods/adaboost/adaboost.hpp
+++ b/src/mlpack/methods/adaboost/adaboost.hpp
@@ -142,11 +142,11 @@ class AdaBoost
    * @param learner Learner to use for training.
    */
   double Train(const MatType& data,
-             const arma::Row<size_t>& labels,
-             const size_t numClasses,
-             const WeakLearnerType& learner,
-             const size_t iterations = 100,
-             const double tolerance = 1e-6);
+               const arma::Row<size_t>& labels,
+               const size_t numClasses,
+               const WeakLearnerType& learner,
+               const size_t iterations = 100,
+               const double tolerance = 1e-6);
 
   /**
    * Classify the given test points.
--- a/src/mlpack/methods/adaboost/adaboost_impl.hpp
+++ b/src/mlpack/methods/adaboost/adaboost_impl.hpp
@@ -62,7 +62,7 @@ AdaBoost<WeakLearnerType, MatType>::AdaBoost(const double tolerance) :
 
 // Train AdaBoost.
 template<typename WeakLearnerType, typename MatType>
-void AdaBoost<WeakLearnerType, MatType>::Train(
+double AdaBoost<WeakLearnerType, MatType>::Train(
     const MatType& data,
     const arma::Row<size_t>& labels,
     const size_t numClasses,
@@ -198,6 +198,7 @@ void AdaBoost<WeakLearnerType, MatType>::Train(
     // Accumulate the value of zt for the Hamming loss bound.
     ztProduct *= zt;
   }
+  return ztProduct;
 }
 
 /**
--- a/src/mlpack/methods/adaboost/adaboost_impl.hpp
+++ b/src/mlpack/methods/adaboost/adaboost_impl.hpp
@@ -57,7 +57,8 @@ template<typename WeakLearnerType, typename MatType>
 AdaBoost<WeakLearnerType, MatType>::AdaBoost(const double tolerance) :
     tolerance(tolerance)
 {
-  // Nothing to do.
+  numClasses = 0;
+  ztProduct = 1.0;
 }
 
 // Train AdaBoost.
--- a/src/mlpack/methods/adaboost/adaboost_impl.hpp
+++ b/src/mlpack/methods/adaboost/adaboost_impl.hpp
@@ -55,10 +55,11 @@ AdaBoost<WeakLearnerType, MatType>::AdaBoost(
 // Empty constructor.
 template<typename WeakLearnerType, typename MatType>
 AdaBoost<WeakLearnerType, MatType>::AdaBoost(const double tolerance) :
-    tolerance(tolerance)
+    tolerance(tolerance),
+    numClasses(0),
+    ztProduct(1.0)
 {
-  numClasses = 0;
-  ztProduct = 1.0;
+  // Nothing to do.
 }
 
 // Train AdaBoost.
--- a/src/mlpack/methods/ann/ffn.hpp
+++ b/src/mlpack/methods/ann/ffn.hpp
@@ -97,6 +97,7 @@ class FFN
    * @param predictors Input training variables.
    * @param responses Outputs results from input training variables.
    * @param optimizer Instantiated optimizer used to train the model.
+   * @return final objective value.
    */
   template<typename OptimizerType>
   double Train(arma::mat predictors,
@@ -118,6 +119,7 @@ class FFN
    * @tparam OptimizerType Type of optimizer to use to train the model.
    * @param predictors Input training variables.
    * @param responses Outputs results from input training variables.
+   * @return final objective value.
    */
   template<typename OptimizerType = ens::RMSProp>
   double Train(arma::mat predictors, arma::mat responses);
--- a/src/mlpack/methods/ann/ffn.hpp
+++ b/src/mlpack/methods/ann/ffn.hpp
@@ -100,8 +100,8 @@ class FFN
    */
   template<typename OptimizerType>
   double Train(arma::mat predictors,
-             arma::mat responses,
-             OptimizerType& optimizer);
+               arma::mat responses,
+               OptimizerType& optimizer);
 
   /**
    * Train the feedforward network on the given input data. By default, the
--- a/src/mlpack/methods/ann/ffn.hpp
+++ b/src/mlpack/methods/ann/ffn.hpp
@@ -97,7 +97,7 @@ class FFN
    * @param predictors Input training variables.
    * @param responses Outputs results from input training variables.
    * @param optimizer Instantiated optimizer used to train the model.
-   * @return final objective value.
+   * @return The final objective of the trained model (NaN or Inf on error).
    */
   template<typename OptimizerType>
   double Train(arma::mat predictors,
@@ -119,7 +119,7 @@ class FFN
    * @tparam OptimizerType Type of optimizer to use to train the model.
    * @param predictors Input training variables.
    * @param responses Outputs results from input training variables.
-   * @return final objective value.
+   * @return The final objective of the trained model (NaN or Inf on error).
    */
   template<typename OptimizerType = ens::RMSProp>
   double Train(arma::mat predictors, arma::mat responses);
--- a/src/mlpack/methods/ann/ffn.hpp
+++ b/src/mlpack/methods/ann/ffn.hpp
@@ -99,7 +99,7 @@ class FFN
    * @param optimizer Instantiated optimizer used to train the model.
    */
   template<typename OptimizerType>
-  void Train(arma::mat predictors,
+  double Train(arma::mat predictors,
              arma::mat responses,
              OptimizerType& optimizer);
 
@@ -120,7 +120,7 @@ class FFN
    * @param responses Outputs results from input training variables.
    */
   template<typename OptimizerType = ens::RMSProp>
-  void Train(arma::mat predictors, arma::mat responses);
+  double Train(arma::mat predictors, arma::mat responses);
 
   /**
    * Predict the responses to a given set of predictors. The responses will
--- a/src/mlpack/methods/ann/ffn_impl.hpp
+++ b/src/mlpack/methods/ann/ffn_impl.hpp
@@ -70,7 +70,7 @@ void FFN<OutputLayerType, InitializationRuleType, CustomLayers...>::ResetData(
 template<typename OutputLayerType, typename InitializationRuleType,
          typename... CustomLayers>
 template<typename OptimizerType>
-void FFN<OutputLayerType, InitializationRuleType, CustomLayers...>::Train(
+double FFN<OutputLayerType, InitializationRuleType, CustomLayers...>::Train(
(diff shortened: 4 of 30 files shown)


walragatver:
Hi @rcurtin, @zoq,
    Sorry for disturbance. But it looks like I have covered all classifiers. Are the conclusions drawn by me correct?

rcurtin:
Hey @walragatver, sorry that I haven't had a chance to take a look yet.  I'll review it when I'm able to (hopefully in the next few days). :+1:

zoq:
Sorry for the slow response, will take a look at the PR over the weekend.

src/mlpack/methods/decision_tree/decision_tree_impl.hpp
```diff
@@ -656,9 +656,24 @@ void DecisionTree<FitnessFunction,
...
             weights, minimumLeafSize, minimumGainSplit);
+        newChildGain = -newChildGain;
+        if(newChildGain>prevChildGain)
+        {
+          double prevContribute = prevChildGain * childCounts[i] / count ;
+          double newBestGain = bestGain - prevContribute;
+          double newContribute = newChildGain * childCounts[i] / count;
+          newBestGain = newBestGain + newContribute;
+          bestGain = std::max(bestGain,newBestGain);
+        }
```
rcurtin: It looks like the strategy you're using here is to check if the entropy of the child is different than the entropy of the child set.  That might happen if the child node makes any additional splits.  But when we call `child->Train()`, it will return either exactly the value of `prevChildGain`, if no further splits were done (i.e. if that child was a leaf), or something better, if further splits were done.  So there should be no need to compute or store `prevChildGain`.

Also, it looks like your strategy for computing `bestGain` is to subtract the contribution of the child points from the original `bestGain` (computed at the beginning of the method), and then add back in the contribution of the newly-built child node.  I think we can ignore the original `bestGain` computation though and essentially do something like this:

```
// Outside of the loop, initialize bestGain.
bestGain = 0;

// This is the loop over the children.
for (size_t i = 0; i < numChildren; ++i)
{
  ...
  double childGain = child->Train(...);
  bestGain += double(childCounts[i]) / double(count) * childGain;
}
```

(The `double()` is to make sure we don't do integer division.  I think it's necessary but I'm not 100% sure.)

Let me know if I can clarify this idea further.

src/mlpack/methods/lars/lars.cpp
```diff
@@ -129,7 +129,7 @@ void LARS::Train(const arma::mat& matX,
   {
     lambdaPath[0] = lambda1;
     Timer::Stop("lars_regression");
-    return;
+    return maxCorr;
```
rcurtin: I'm not sure that `maxCorr` is the objective minimized by LARS.  I think that objective is Equation 1.3 from https://projecteuclid.org/download/pdfview_1/euclid.aos/1083178935 .  Maybe I overlooked how `maxCorr` is equivalent to that value?

src/mlpack/methods/linear_regression/linear_regression.cpp
```diff
@@ -85,6 +85,7 @@ void LinearRegression::Train(const arma::mat& predictors,
       lambda * arma::eye<arma::mat>(p.n_rows, p.n_rows);
 
   parameters = arma::solve(cov, p * r.t());
+  return ComputeError(predictors,responses);
```
rcurtin: Tiny style issue---there should be a space between `predictors,` and `responses`.

src/mlpack/methods/random_forest/random_forest_impl.hpp
```diff
@@ -431,6 +431,7 @@ void RandomForest<
 {
   // Train each tree individually.
   trees.resize(numTrees); // This will fill the vector with untrained trees.
+  double avgGain = 0.0;
```
rcurtin: I think we might want to use an OpenMP reduce or something like this.  As it's written, `avgGain` is shared across all threads in the parallel loop for training each tree, so there may be locking issues.

src/mlpack/methods/sparse_coding/sparse_coding_impl.hpp
```diff
@@ -106,6 +106,7 @@ void SparseCoding::Train(
   }
 
   Timer::Stop("sparse_coding");
+  return lastObjVal;
```
rcurtin: Should this be `curObjVal`?

src/mlpack/methods/lars/lars.cpp
walragatver: Actually, I was not sure of that. 
I am returning that because in section 2 fig 3(right) of the paper they have plotted absolute current correlations as function of LARS step . In that maximum correlation decreases as we keep training. So I thought it's quite similar to decreasing error hence I thought returning that might be reasonable. Also I went through code and it was training on maxCorr so I thought that returning maxCorr is correct. Also in our code there is no function or something which is calculating that error. 
If there are modifications let me know

src/mlpack/methods/sparse_coding/sparse_coding_impl.hpp
walragatver: No actually, curObjVal is only in the scope of for loop , so we cannot return that. But lastObjVal is in the function scope and in the for loop we can see that we are copying curObjVal in lastObjVal (line 105). So I thought instead of declaring curObjVal outside for loop. It's better to return lastObjVal as it's getting updated with curObjVal at the end of for loop.

src/mlpack/methods/adaboost/adaboost.hpp
```diff
@@ -141,7 +141,7 @@ class AdaBoost
    * @param labels Labels for each point in the dataset.
    * @param learner Learner to use for training.
    */
-  void Train(const MatType& data,
+  double Train(const MatType& data,
```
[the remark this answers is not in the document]
walragatver: Okay Fixed it !

src/mlpack/methods/adaboost/adaboost_impl.hpp
```diff
@@ -198,6 +198,7 @@ void AdaBoost<WeakLearnerType, MatType>::Train(
     // Accumulate the value of zt for the Hamming loss bound.
     ztProduct *= zt;
   }
+  return ztProduct;
```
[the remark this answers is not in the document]
walragatver: Yes, It looks like many of our tests can be modified after this modification. I am thinking to work on this in a separate PR

src/mlpack/methods/decision_tree/decision_tree_impl.hpp
walragatver: Yes, I got that thanks for the suggestion

src/mlpack/methods/linear_regression/linear_regression.cpp
walragatver: Okay fixed it!

src/mlpack/methods/random_forest/random_forest_impl.hpp
walragatver: Done !

walragatver:
Hi @rcurtin ,
I have fixed and answered the things you have said. Just let me know if I have made a mistake.
Also I have modified local_coordinate_coding.
Umm, actually I have not adapted searching algorithms like range_search and collabarative filtering. I will adapt them in 2 days and will ping you for final review.

rcurtin:
Range search doesn't have a reasonable objective so I am not sure what we would return there.  If you can think of something reasonable, go ahead, but I'm not sure what.  (Similar situation for collaborative filtering.)

walragatver:
Hi @rcurtin ,
I have checked all the above algorithms and it looks like we cannot return anything reasonable. It looks like most of them use matrices as it's output. Also your thesis on dual tree algorithms helped me a lot in clearing the basics of dual and single-tree algorithms. 
So, this PR is now ready for it's final review.

src/mlpack/methods/adaboost/adaboost_impl.hpp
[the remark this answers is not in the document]
rcurtin: Ok, that's fine, but could you please open an issue so that we don't forget it?  You can assign it to yourself if you like.

src/mlpack/methods/decision_tree/decision_tree_impl.hpp
```diff
@@ -624,7 +624,7 @@ void DecisionTree<FitnessFunction,
     }
 
     // Figure out counts of children.
-    arma::Row<size_t> childCounts(numClasses, arma::fill::zeros);
+    arma::Row<size_t> childCounts(numChildren, arma::fill::zeros);
```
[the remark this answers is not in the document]
rcurtin: Yeah.  I think we happen to be guaranteed that `numClasses` is always greater than or equal to `numChildren` in the code we have, so it was a dormant bug, but still a bug.  Nice catch :sunglasses:

src/mlpack/methods/decision_tree/decision_tree_impl.hpp
```diff
@@ -624,10 +626,16 @@ void DecisionTree<FitnessFunction,
...
     for (size_t i = begin; i < begin + count; ++i)
       childCounts[childAssignments[i - begin]]++;
 
+    //Initialize bestGain if recursive split is allowed
+    if(!NoRecursion)
```
rcurtin: Just a few style comments---can you make sure that each comment is a fully punctuated sentence?  (i.e. period at the end).  Also we can put a space between `//` and `Initialize`, and a space between `if` and `(!NoRecursion)` to match the style guide.  These comments apply throughout, by the way, even though I've only pointed them out here.

src/mlpack/methods/decision_tree/decision_tree_impl.hpp
```diff
@@ -675,6 +685,7 @@ void DecisionTree<FitnessFunction,
         numClasses,
         UseWeights ? weights.subvec(begin, begin + count - 1) : weights);
   }
+  return -bestGain;
```
rcurtin: Do we need to negate a second time here?

src/mlpack/methods/lars/lars.cpp
rcurtin: I think there is a relation between the two---we keep adding variables to the solution until `maxCorr` is sufficently small, so in some sense `maxCorr` captures the amount of `Y` that is not captured by `X*beta`.  There may be some way to convert to something much closer to the mean-squared-error of the regression, but if not I think `maxCorr` can work.  If you'd like to take the time to look into it further, please feel free.

src/mlpack/methods/random_forest/random_forest_impl.hpp
rcurtin: :+1:

src/mlpack/methods/sparse_coding/sparse_coding_impl.hpp
rcurtin: Ok, but in this case take a look at line 101---if it converges, then `lastObjVal` is not up to date.  So I guess we need to update it there.

src/mlpack/methods/decision_tree/decision_tree_impl.hpp
walragatver: Yes, if as we can see we are maximizing entropy. Objective function always minimizes something. So returning negative entropy of final tree is reasonable.

rcurtin: Sorry, I thought that it was avoidable since we're already negating `childGain` earlier, but I realize that the code you have here also handles the case where no split is made (in which case `bestGain` needs to be negated).

walragatver:
Hi @rcurtin, @zoq,
I have made the final changes and fixed some static errors. It looks there is some glitch in the static check because the error it is pointing is not an error. There is no local `activeSet` in the train function of `lars.cpp`. 
Can you look into it?

rcurtin:
Can you try rebasing the master branch into your branch?  #1713 had some fixes for the static code analysis checks that might help here.

walragatver:
@rcurtin: I have already rebased my branch but it looks like a new error. I tried to clear it but looks like it's not an error.

rcurtin:
Ah, ok, sorry.  Yeah, I took a look at the code and I have to agree, the static code analyzer must be wrong here.  I can't see why it would think that `activeSet` is local.  So let's just not worry about that. :+1:

walragatver:
Hi @rcurtin, @zoq,
Can you review this PR today? I am thinking to move towards writing and test cases and finish it as fast as possible. 
Also I went through this year's GSOC ideas they are quite fascinating. Which one will you recommend for me?

rcurtin:
It's in my queue to do soon, but this week is a paper deadline week so there's not too much time for me.  Sorry about that...

zoq:
@walragatver regarding GSoC, difficult to recommend something at this point, personally, I would go with the idea I'm most passionate about. Also, you are free to propose something that's not listed.

src/mlpack/methods/ann/ffn_impl.hpp
```diff
@@ -84,12 +84,13 @@ void FFN<OutputLayerType, InitializationRuleType, CustomLayers...>::Train(
 
   Log::Info << "FFN::FFN(): final objective of trained model is " << out
       << "." << std::endl;
+  return out;
```
zoq: Might be more intuitive to call this loss, or objective, what do you think?

src/mlpack/methods/ann/ffn.hpp
```diff
@@ -97,11 +97,12 @@ class FFN
    * @param predictors Input training variables.
    * @param responses Outputs results from input training variables.
    * @param optimizer Instantiated optimizer used to train the model.
+   * @return final objective value.
```
zoq: Picky style comment, this should start with a capital letter.

src/mlpack/methods/ann/ffn_impl.hpp
walragatver: Yes, correct but it was already declared or defined. So I thought it's better not to change it's name. Also at every optimizer function it was written like that so I decided to keep it as it is.

src/mlpack/methods/ann/ffn.hpp
walragatver: Done

walragatver:
Hi @zoq, 
I have made the required changes.

src/mlpack/methods/ann/ffn_impl.hpp
zoq: I see, I don't mind to leave it like this.

src/mlpack/methods/ann/rbm/rbm_impl.hpp
```diff
@@ -50,6 +50,7 @@ RBM<InitializationRuleType, DataType, PolicyType>::RBM(
     reset(false)
 {
   numFunctions = this->predictors.n_cols;
+  steps = 0;
```
zoq: Do you mind to use the member initializer list for the ``steps`` parameter for consistency?

src/mlpack/methods/linear_regression/linear_regression.hpp
```diff
@@ -87,11 +88,12 @@ class LinearRegression
    * @param responses y, the responses to the data points.
    * @param intercept Whether or not to fit an intercept term.
    * @param weights Observation weights (for boosting).
+   * @return ordinary least squares error.
```
rcurtin: Tiny style issue; we should start the comment capitalized, i.e. `Ordinary` :)

rcurtin:
Hehe, always bugs in code... I wonder why mlpack-bot offered stickers twice...

walragatver:
Hi @rcurtin,
I have updated History.md. However was not able to preview the change. So if something goes wrong you handle that during merge.
Thanks for approval :+1:.

rcurtin:
Thanks for the contribution! :+1:

PR #1686, opened 2019-01-19
test bot change #2

ecurtin:
in which @rcurtin hijacks a Github account to test his bot creation.If this works, the bot should label the issue, post a comment, and then when I approve it from my real account, should post a comment about stickers. Let's see what happens...


rcurtin:
I would call this a success.  Now new contributors will get offered stickers. :+1:

PR #1689, opened 2019-01-23
Added test for gmm_generate

gauravcr7rm:
Added test for gmm_generate

### DIFF
--- a/src/mlpack/tests/main_tests/gmm_generate_test.cpp
+++ b/src/mlpack/tests/main_tests/gmm_generate_test.cpp
@@ -1,0 +1,109 @@
+/**
+ * @file gmm_train_test.cpp
+ * @author Gaurav Tripathi
+ *
+ * Test mlpackMain() of gmm_train_main.cpp.
+ *
+ * mlpack is free software; you may redistribute it and/or modify it under the
+ * terms of the 3-clause BSD license.  You should have received a copy of the
+ * 3-clause BSD license along with mlpack.  If not, see
+ * http://www.opensource.org/licenses/BSD-3-Clause for more information.
+ */
+#define BINDING_TYPE BINDING_TYPE_TEST
+
+#include <mlpack/core.hpp>
+static const std::string testName = "Gmm_generate"
+
+#include <mlpack/core/util/mlpack_main.hpp>
+#include <mlpack/methods/gmm/gmm_generate_main.cpp>
+#include <mlpack/methods/gmm/gmm_train_main.cpp>
+#include "test_helper.hpp"
+
+#include <boost/test/unit_tests.hpp>
+#include "../test_tools.hpp"
+
+using namespace mlpack;
+
+struct GmmGenerateTestFixture
+{
+  public:
+   GmmGenerateTestFixture()
+   {
+  	 CLI::RestoreSettings(testname);
+   }
+
+   ~GmmGenerateTestFixture()
+   {
+  	 bindings::tests::CleanMemory();
+  	 CLI::ClearSettings();
+   }
+};
+
+BOOST_FIXTURE_TEST_SUITE(GmmGenerateMainTest,
+                         GmmGenerateTestFixture);
+
+// Making sure samples are provided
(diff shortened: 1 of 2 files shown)


src/mlpack/tests/main_tests/gmm_generate_test.cpp
```diff
@@ -0,0 +1,109 @@
...
+
+BOOST_FIXTURE_TEST_SUITE(GmmGenerateMainTest,
+                         GmmGenerateTestFixture);
+
+// Making sure samples are provided.
+BOOST_AUTO_TEST_CASE(GmmGenerateSamples)
+{
+  arma::mat inputData;
+  if (!data::Load("vc2.csv", inputData))
+  BOOST_FAIL("Unable to load train dataset vc2.csv!");
```
rcurtin: This should be indented.

src/mlpack/tests/main_tests/gmm_generate_test.cpp
```diff
@@ -0,0 +1,109 @@
...
+  SetInputParam("input_model", std::move(gmm));
+
+  Log::Fatal.ignoreInput = true;
+  BOOST_REQUIRE_THROW(mlpackMain(), std::runtime_error);
+  Log::Fatal.ignoreInput = false;
+}
+
+// Checking dimensionality of output.
+BOOST_AUTO_TEST_CASE(GmmGenerateDimensionality)
+{ int samples =10;
```
rcurtin: Why is this on the same line as the opening brace?

src/mlpack/tests/main_tests/gmm_generate_test.cpp
```diff
@@ -0,0 +1,109 @@
...
+  
+  GMM gmm(1, 2);
+  gmm.Train(data, 10);
+
+  SetInputParam("input_model", std::move(gmm));
+  SetInputParam("samples", samples);
+
+  mlpackMain();
+
+  arma::mat output = CLI::GetParam<arma::mat>("output");
```
rcurtin: No need for a copy here; you can use a reference like `const arma::mat& output = ...`, or move the matrix with `arma::mat output = std::move(CLI::GetParam<arma::mat>("output"));`.

src/mlpack/tests/main_tests/gmm_generate_test.cpp
```diff
@@ -0,0 +1,109 @@
...
+
+  BOOST_REQUIRE_EQUAL(output.n_rows, gmm->Dimensionality());
+  BOOST_REQUIRE_EQUAL(output.n_cols, samples);
+ 
+}
+
+// Samples passed should be positive.
```
rcurtin: There is a related test, which is to see what happens when the number of samples passed is 0.

gauravcr7rm: ok ill do the changes

gauravcr7rm: @rcurtin i was able to build all these successfully in my local system and  i have made all the changes specified.
Please have a look

src/mlpack/tests/main_tests/gmm_generate_test.cpp
```diff
@@ -0,0 +1,81 @@
...
+// Making sure samples are provided.
+BOOST_AUTO_TEST_CASE(GmmGenerateSamples)
+{
+  arma::mat inputData;
+  if (!data::Load("vc2.csv", inputData))
+    BOOST_FAIL("Unable to load train dataset vc2.csv!");
+  
+  GMM gmm(1, 2);
+  gmm.Train(inputData, 2);
+  SetInputParam("input_model", std::move(gmm));
```
rcurtin: I think you need to call this with a pointer to the model, so `SetInputParam("input_model", &gmm);`.  If you fix that, I believe the build will pass.

rcurtin:
With #1899 merged this is handled so I'll close this.  Thanks again for the contribution. :+1:

PR #1691, opened 2019-01-24
Add issue templates

rcurtin:
Here is a basic idea for issue templates.  This should help us classify issues that people open more quickly, and hopefully help get the information we need to take care of the issues more quickly.

It seems like we can automatically apply labels to some of these, which is great, so I have done that where applicable (but the "question" template could be a feature request or a question or something else, so we can just let mlpack-bot label it as `unlabeled`).

### DIFF
--- a/.github/ISSUE_TEMPLATE/0-bug.md
+++ b/.github/ISSUE_TEMPLATE/0-bug.md
@@ -1,6 +1,7 @@
 ---
 name: bug report
 about: Use this template for reporting a bug that you have found in mlpack.
+labels: "t: bug report", "s: unanswered"
 
 ---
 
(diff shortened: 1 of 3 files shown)


ShikharJ:
Please feel free to merge this in if you think it is complete.

rcurtin:
Thanks for the reviews---hopefully these will be helpful. :)

zoq:
Looks like only one issue template is shown, I can see in the settings one can also add issue template, perhaps we have to active the other as well?

rcurtin:
@zoq: Looks like I had the `labels` bit wrong.  Fixed in a couple of commits: cfa363080, 5d9270d7c, 34f48152b, and 7d448192c.

PR #1692, opened 2019-01-24
Can Travis build for ICC?

rcurtin:
I found a Github project that sets up ICC for Travis.  So I tried modifying the Travis configuration... let's see if it builds correctly...


rcurtin:
Actually I'm going to conclude no for this one.  Travis doesn't seem to make secret environment variables available to PRs, so we wouldn't be able to test whether new code works with icc.  The better solution here would probably be a Docker container for Jenkins or something like this with icc set up in it.  The `icc-travis` script could still be useful for that, but setting that up on Jenkins is a bit more work so I'll leave it for some other time.

PR #1702, opened 2019-02-03
[WIP] Binding tests for Local Coordinate Coding

bhavya01:
Added basic dimensionality test (#1152)


bhavya01:
@rcurtin can you review this?

bhavya01:
Thanks for the comments. I will implement the changes.

bhavya01:
@rcurtin  I think this is ready for a review.

src/mlpack/methods/local_coordinate_coding/local_coordinate_coding_main.cpp
```diff
@@ -149,6 +149,22 @@ static void mlpackMain()
         matX.col(i) /= norm(matX.col(i), 2);
     }
 
+    if (CLI::GetParam<int>("atoms") < 1)
+    {
+      Log::Fatal << "The number of atoms should be atleast one!" << endl;
+    }
```
rcurtin: Can you use the functions in `src/mlpack/core/util/param_checks.hpp` for these three parameter checks?  Example:

```
RequireParamValue<int>("atoms", [](int x) { return x > 0; },
    "number of atoms must be greater than 1");
```

src/mlpack/tests/main_tests/local_coordinate_coding_test.cpp
```diff
@@ -0,0 +1,417 @@
...
+  SetInputParam("training", std::move(x));
+  SetInputParam("test", t);
+  SetInputParam("atoms", (int) 10);
+  SetInputParam("max_iterations", (int) 2);
+
+  mlpackMain();
+
+  // Get the encoded output and dictionary after training.
+  arma::mat initCodes = CLI::GetParam<arma::mat>("codes");
+  arma::mat initDict = CLI::GetParam<arma::mat>("dictionary");
```
rcurtin: You can use `std::move()` with these two to avoid a copy. :+1:

src/mlpack/tests/main_tests/local_coordinate_coding_test.cpp
```diff
@@ -0,0 +1,417 @@
...
+  SetInputParam("test", std::move(t));
+
+  Log::Fatal.ignoreInput = true;
+  BOOST_REQUIRE_THROW(mlpackMain(), std::runtime_error);
+  Log::Fatal.ignoreInput = false;
+}
+
+/*
+* Ensure that the program throws error for negative number of atoms.
+*/
```
rcurtin: Here's another one---what happens when the number of atoms is greater than the number of points in the dataset?

src/mlpack/tests/main_tests/local_coordinate_coding_test.cpp
```diff
@@ -0,0 +1,417 @@
...
+
+  bindings::tests::CleanMemory();
+  CLI::Parameters()["normalize"].wasPassed = false;
+
+  // Normalize the datasets.
+  for (size_t i = 0; i < x.n_cols; i++)
+    x.col(i) /= norm(x.col(i), 2);
+
+  for (size_t i = 0; i < t.n_cols; i++)
+    t.col(i) /= norm(t.col(i), 2);
```
rcurtin: Actually for this one, we should avoid normalizing here and then test that the dictionary and codes from the two different runs are different.  The reason is that although you've specified `normalize` as false for the second run and manually normalized `x` and `t`, it's still possible that the binding might be normalizing anyway---which wouldn't change the data at all (and thus give the same results).

bhavya01:
@rcurtin I have made the changes. Can you have a look?

src/mlpack/tests/main_tests/local_coordinate_coding_test.cpp
bhavya01: I could not infer from the paper that the number of atoms has to be less than the number of data points but in general, we expect that it should be the case.

rcurtin:
Thanks again! :+1:

PR #1707, opened 2019-02-06
Refactor HMM tests to reduce errors

rcurtin:
I noticed looking at Jenkins that the HMM test was failing, and I decided to investigate how robust these tests were.  So I ran them over and over again to identify tolerances that were bad, and in some cases this meant that I needed to change tests that looked at individual elements of matrices and replace these with tests that look at the norm of the difference of matrices, which tends to be much more robust and still give a similar test.

I left a loop running the HMM tests overnight and I'm now satisfied that no tests should regularly fail.


rcurtin:

src/mlpack/tests/hmm_test.cpp
```diff
@@ -204,10 +204,10 @@ BOOST_AUTO_TEST_CASE(SimpleBaumWelchDiscreteHMM)
...
   observations.push_back("0 1 0 1 0 1 0 1 0 1 0 1");
   observations.push_back("0 0 0 0 0 0 1 1 1 1 1 1");
-  observations.push_back("1 1 1 1 1 1 0 0 0 0 0 0");
-  observations.push_back("1 1 1 0 0 0 1 1 1 0 0 0");
-  observations.push_back("0 0 1 1 0 0 0 0 1 1 1 1");
-  observations.push_back("1 1 1 0 0 0 1 1 1 0 0 0");
+  observations.push_back("1 1 1 1 1 0 1 0 0 0 0 0");
+  observations.push_back("1 1 1 0 0 1 0 1 1 0 0 0");
+  observations.push_back("0 0 1 1 0 0 0 1 0 1 1 1");
+  observations.push_back("1 1 1 0 0 1 0 1 1 0 0 0");
```
rcurtin: This was for the static code analyzer, which didn't like the duplicated observations.  But changing the order of observations doesn't affect the result.

zoq: interesting, looks like after the update the static code analyzer is becoming picky

src/mlpack/tests/hmm_test.cpp
```diff
@@ -904,66 +859,84 @@ BOOST_AUTO_TEST_CASE(GaussianHMMPredictTest)
...
 BOOST_AUTO_TEST_CASE(GMMHMMPredictTest)
 {
-  // We will use two GMMs; one with two components and one with three.
-  std::vector<GMM> gmms(2);
-  gmms[0] = GMM(2, 2);
-  gmms[0].Weights() = arma::vec("0.75 0.25");
+  // It's possible, but extremely unlikely, that this test can fail.  So we are
+  // willing to do three trials in case the first two fail.
+  bool success = false;
+  for (size_t trial = 0; trial < 2; ++trial)
```
zoq: "do three trails", but we only do 2 here, maybe I missed something? I guess since this is unlikely anyway, 2 are just fine.

rcurtin: Oh, thanks.  Let me change that to `3` for extra paranoia.

PR #1708, opened 2019-02-06
Static analysis check

zoq:
Just a simple test to see if the Static analysis check works as expected.


zoq:

zoq:
Let's see if the threshold is correct.

Issue #1706, opened 2019-02-04
Avoid copying Pandas DataFrames for Python bindings when possible


rcurtin:
This issue is specific to the Python bindings and came out of the review of #1700 and the related issue #1694.  mlpack Python bindings support Pandas DataFrames, as in the following snippet:

```
import pandas as pd
import mlpack

d = pd.read_csv("my_data.csv")
output = mlpack.pca(input=d, verbose=True)
```

The Pandas DataFrame `d` gets automatically converted to a numpy matrix, and then the underlying data pointer of that numpy matrix is passed to the binding itself.

The code flow is that the method `mlpack.pca()` (or any other mlpack Python method) will call `to_matrix()` with an input matrix, which will then "convert" whatever was given to a numpy `ndarray`.  (Later, outside the scope of `to_matrix()`, the returned `ndarray` is converted to an Armadillo matrix without copying.  We don't need to worry about those details for this issue.)

Currently the implementation of `to_matrix()` operates basically by checking if what was given to `to_matrix()` is already an `ndarray` that is row-major (i.e. the flag `C_CONTIGUOUS` is set).  If that's true (and we are not explicitly told to copy the matrix), we return the matrix that was passed with no copy or modification.  But if the input to `to_matrix()` is *not* an `ndarray`, we end up copying no matter what.

So what we should aim to look into this issue is the following:

1. Does Pandas store an underlying DataFrame in the same memory layout as an `ndarray`?  If so, then...
2. Adapt `to_matrix()` to handle Pandas DataFrames (and possibly Pandas series too) without copying, when possible.

This issue can make a big difference, because for large datasets even just copying the memory itself can have significant cost.

mirraaj:
@rcurtin @zoq  Could you please provide the link where function ```to_matrix``` is defined?

rcurtin:
Oh!  Forgot to do that.  It's in `src/mlpack/bindings/python/mlpack/matrix_utils.py`.  Thanks for pointing that out. :)

mirraaj:
@rcurtin Let me see if I could contribute to this issue. Thanks

Yashwants19:
Hi @rcurtin please review #1711

rcurtin:
@Yashwants19 I saw the PR, I just haven't had a chance to respond to it yet.  Don't worry---I get an email for all activity, so I will respond when I'm able to.

g-laz77:
Is this issue still open?

abernauer:
[...]

Issue #1710, opened 2019-02-07
ValueError: The truth value of a DataFrame is ambiguous. Use a.empty, a.bool(), a.item(), a.any() or a.all().


Nimishkhurana:
#### Issue description

```
>>>import pandas as pd
>>> import numpy as np
>>> import mlpack
>>> 
>>> df = pd.read_csv('housingdata.csv',header=None)
>>> 
>>> X = df.iloc[:,5]
>>> y = df.iloc[:,13]
>>> 
>>> output= mlpack.linear_regression(X,y)
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "mlpack/linear_regression.pyx", line 119, in mlpack.linear_regression.linear_regression
  File "/home/nimish/anaconda3/lib/python3.6/site-packages/pandas/core/generic.py", line 1121, in __nonzero__
    .format(self.__class__.__name__))
ValueError: The truth value of a Series is ambiguous. Use a.empty, a.bool(), a.item(), a.any() or a.all().
>>> X = pd.DataFrame(X)
>>> y = pd.DataFrame(y)
>>> output= mlpack.linear_regression(X,y)
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "mlpack/linear_regression.pyx", line 119, in mlpack.linear_regression.linear_regression
  File "/home/nimish/anaconda3/lib/python3.6/site-packages/pandas/core/generic.py", line 1121, in __nonzero__
    .format(self.__class__.__name__))
ValueError: The truth value of a DataFrame is ambiguous. Use a.empty, a.bool(), a.item(), a.any() or a.all().
```

#### Your environment

 * version of mlpack:3.0.4
 * operating system:Ubuntu

I have build mlpack from source.

bhavya01:
Hi @Nimishkhurana 
It seems that there is an issue in the ordering of parameters to the function in the generated binding
I looked at the generated binding for linear regresssion in build/src/mlpack/bindings/python/mlpack/linear_regression.py and tried this
```
output = mlpack.linear_regression(training=X, training_output=y)
```
This solves the ValueError but is throwing a segmentation fault for me.

Nimishkhurana:
@bhavya01 I also tried the same you said and goit segmentation fault.
Also if I try to run this in jupyter-notebook then kernel keeps on dying again and again.

bhavya01:
[...]

PR #1711, opened 2019-02-07
FIX copying Pandas DataFrames for Python bindings

Yashwants19:
Address #1706

### DIFF
--- a/src/mlpack/bindings/python/mlpack/matrix_utils.py
+++ b/src/mlpack/bindings/python/mlpack/matrix_utils.py
@@ -58,7 +58,7 @@ def to_matrix(x, dtype=np.double, copy=False):
     else:
       return np.ndarray(x.shape,buffer=x.flatten(),dtype=dtype, order = 'C'),False
   else:
-    if isinstance(x, pd.core.series.Series):
+    if isinstance(x, pd.core.series.Series) or isinstance(x, pd.DataFrame):
       y = x.to_numpy()
     if copy: # Copy the matrix if required.
       return np.ndarray(y.shape,buffer=y.flatten(),dtype=dtype,order = 'C').copy("C"),True
--- a/src/mlpack/bindings/python/mlpack/matrix_utils.py
+++ b/src/mlpack/bindings/python/mlpack/matrix_utils.py
@@ -60,10 +60,10 @@ def to_matrix(x, dtype=np.double, copy=False):
   else:
     if isinstance(x, pd.core.series.Series) or isinstance(x, pd.DataFrame):
       y = x.to_numpy()
-    if copy: # Copy the matrix if required.
-      return np.ndarray(y.shape,buffer=y.flatten(),dtype=dtype,order = 'C').copy("C"),True
-    else:
-      return np.ndarray(y.shape,buffer=y.flatten(),dtype=dtype, order = 'C'),False
+      if copy: # Copy the matrix if required.
+        return np.ndarray(y.shape,buffer=y.flatten(),dtype=dtype,order = 'C').copy("C"),True
+      else:
+        return np.ndarray(y.shape,buffer=y.flatten(),dtype=dtype, order = 'C'),False
     
 
 def to_matrix_with_info(x, dtype, copy=False):
--- a/src/mlpack/bindings/python/mlpack/matrix_utils.py
+++ b/src/mlpack/bindings/python/mlpack/matrix_utils.py
@@ -52,13 +52,19 @@ def to_matrix(x, dtype=np.double, copy=False):
       return x.copy("C"), True
     else:
       return x, False
+  elif (isinstance(x, np.ndarray) and x.dtype == dtype and x.flags.f_contiguous):
+    if copy: # Copy the matrix if required.
+      return np.ma.array(x,copy = True,order= 'C'),True
+    else:
+      return np.ma.array(x,order= 'C'),False
   else:
     if isinstance(x, pd.core.series.Series):
-      x = pd.DataFrame(x)
+      y = x.to_numpy()
     if copy: # Copy the matrix if required.
-      return np.array(x, copy=True, dtype=dtype, order='C'), True
+      return np.ma.array(y,copy = True, dtype=dtype, order='C'), True
     else:
-      return np.array(x, dtype=dtype), False
+      return np.ma.array(y, dtype=dtype,order = 'C'), False
+    
 
 def to_matrix_with_info(x, dtype, copy=False):
   """
--- a/src/mlpack/bindings/python/mlpack/matrix_utils.py
+++ b/src/mlpack/bindings/python/mlpack/matrix_utils.py
@@ -54,16 +54,16 @@ def to_matrix(x, dtype=np.double, copy=False):
       return x, False
   elif (isinstance(x, np.ndarray) and x.dtype == dtype and x.flags.f_contiguous):
     if copy: # Copy the matrix if required.
-      return np.ndarray(x.shape, buffer=x.flatten(), dtype=dtype, order='C').copy("C"), True
+      return np.ndarray(x.shape, buffer=x.flatten(), dtype=x.dtype, order='C').copy("C"), True
     else:
-      return np.ndarray(x.shape, buffer=x.flatten(), dtype=dtype, order='C'), False
+      return np.ndarray(x.shape, buffer=x.flatten(), dtype=x.dtype, order='C'), False
   else:
     if isinstance(x, pd.core.series.Series) or isinstance(x, pd.DataFrame):
       y = x.values
       if copy: # Copy the matrix if required.
-        return np.ndarray(y.shape, buffer=y.flatten(), dtype=dtype, order='C').copy("C"), True
+        return np.ndarray(y.shape, buffer=y.flatten(), dtype=y.dtype, order='C').copy("C"), True
       else:
-        return np.ndarray(y.shape, buffer=y.flatten(), dtype=dtype, order='C'), False
+        return np.ndarray(y.shape, buffer=y.flatten(), dtype=y.dtype, order='C'), False
     else:
       return np.array(x, copy=True, dtype=dtype, order='C'), True
     
--- a/src/mlpack/bindings/python/mlpack/matrix_utils.py
+++ b/src/mlpack/bindings/python/mlpack/matrix_utils.py
@@ -54,9 +54,9 @@ def to_matrix(x, dtype=np.double, copy=False):
       return x, False
   elif (isinstance(x, np.ndarray) and x.dtype == dtype and x.flags.f_contiguous):
     if copy: # Copy the matrix if required.
-      return np.ma.array(x,copy = True,order= 'C'),True
+      return np.ma.array(x,copy = True, dtype=dtype ,order= 'C'),True
     else:
-      return np.ma.array(x,order= 'C'),False
+      return np.ma.array(x, dtype=dtype,order= 'C'),False
   else:
     if isinstance(x, pd.core.series.Series):
       y = x.to_numpy()
--- a/src/mlpack/bindings/python/mlpack/matrix_utils.py
+++ b/src/mlpack/bindings/python/mlpack/matrix_utils.py
@@ -54,16 +54,16 @@ def to_matrix(x, dtype=np.double, copy=False):
       return x, False
   elif (isinstance(x, np.ndarray) and x.dtype == dtype and x.flags.f_contiguous):
     if copy: # Copy the matrix if required.
-      return np.ma.array(x,copy = True, dtype=dtype ,order= 'C'),True
+      return np.ndarray(x.shape,buffer=x.flatten(),dtype=dtype,order = 'C').copy("C"),True
     else:
-      return np.ma.array(x, dtype=dtype,order= 'C'),False
+      return np.ndarray(x.shape,buffer=x.flatten(),dtype=dtype, order = 'C'),False
   else:
     if isinstance(x, pd.core.series.Series):
       y = x.to_numpy()
     if copy: # Copy the matrix if required.
-      return np.ma.array(y,copy = True, dtype=dtype, order='C'), True
+      return np.ndarray(y.shape,buffer=y.flatten(),dtype=dtype,order = 'C').copy("C"),True
     else:
-      return np.ma.array(y, dtype=dtype,order = 'C'), False
+      return np.ndarray(y.shape,buffer=y.flatten(),dtype=dtype, order = 'C'),False
     
 
 def to_matrix_with_info(x, dtype, copy=False):
--- a/src/mlpack/bindings/python/mlpack/matrix_utils.py
+++ b/src/mlpack/bindings/python/mlpack/matrix_utils.py
@@ -59,7 +59,7 @@ def to_matrix(x, dtype=np.double, copy=False):
       return np.ndarray(x.shape,buffer=x.flatten(),dtype=dtype, order = 'C'),False
   else:
     if isinstance(x, pd.core.series.Series) or isinstance(x, pd.DataFrame):
-      y = x.to_numpy()
+      y = x.values
       if copy: # Copy the matrix if required.
         return np.ndarray(y.shape,buffer=y.flatten(),dtype=dtype,order = 'C').copy("C"),True
       else:
--- a/src/mlpack/bindings/python/mlpack/matrix_utils.py
+++ b/src/mlpack/bindings/python/mlpack/matrix_utils.py
@@ -54,16 +54,18 @@ def to_matrix(x, dtype=np.double, copy=False):
       return x, False
   elif (isinstance(x, np.ndarray) and x.dtype == dtype and x.flags.f_contiguous):
     if copy: # Copy the matrix if required.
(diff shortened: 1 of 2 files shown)


src/mlpack/bindings/python/mlpack/matrix_utils.py
```diff
@@ -55,7 +55,10 @@ def to_matrix(x, dtype=np.double, copy=False):
   else:
     if isinstance(x, pd.core.series.Series):
       x = pd.DataFrame(x)
-    return np.array(x, copy=True, dtype=dtype, order='C'), True
+    if copy: # Copy the matrix if required.
+      return np.array(x, copy=True, dtype=dtype, order='C'), True
+    else:
+      return np.array(x, dtype=dtype), False
```
rcurtin: What happens if we pass a Pandas Dataframe where the order is `F` instead of `C`?  Try this code:

```
import pandas as pd
import numpy as np
d = pd.read_csv("some_data.csv")
x = np.array(d)
x.flags
```

You can see that `F_CONTIGUOUS` is true, but we need `C_CONTIGUOUS` to be true when we pass directly to mlpack.  So I think we need a way to handle this.

In addition I think this still copies the Pandas dataframe, since when I look at `x.flags` I see

```
  C_CONTIGUOUS : False                                                                                                                                          
  F_CONTIGUOUS : True                                                                                                                                           
  OWNDATA : True                                                                                                                                                
  WRITEABLE : True                                                                                                                                              
  ALIGNED : True                                                                                                                                                
  WRITEBACKIFCOPY : False                                                                                                                                       
  UPDATEIFCOPY : False   
```

and `OWNDATA : True` implies that the matrix was copied.

Also, we should probably add a test case to `src/mlpack/bindings/python/tests/test_python_binding.py`.  Take a look at the test `testNumpyMatrix(self)`, and we can adapt this to use a Pandas dataframe.  That would be a good way to ensure this works.

src/mlpack/bindings/python/mlpack/matrix_utils.py
```diff
@@ -55,7 +55,10 @@ def to_matrix(x, dtype=np.double, copy=False):
   else:
     if isinstance(x, pd.core.series.Series):
       x = pd.DataFrame(x)
```
rcurtin: Do you think there is any way we could avoid copying the `Series` object?  Also, another thing to notice is that when we do `x = pd.DataFrame(x)`, we are already copying the DataFrame, so if `copy=True` we don't need to copy a second time.

Yashwants19: hi @rcurtin can we use `y = x.to_numpy()`

rcurtin: Have you tried it to see if it causes a copy?

Yashwants19: Hi @rcurtin it doesn't make copy. if we run this 
`import pandas as pd
import numpy as np
d = pd.read_csv("some_data.csv")
x = d.to_numpy()
x.flags`

its gives 
`  C_CONTIGUOUS : False                                                                                                                                          
  F_CONTIGUOUS : True                                                                                                                                           
  OWNDATA : False                                                                                                                                               
  WRITEABLE : True                                                                                                                                              
  ALIGNED : True                                                                                                                                                
  WRITEBACKIFCOPY : False                                                                                                                                       
  UPDATEIFCOPY : False`


Here   `F_CONTIGUOUS`  is True we can resolve this by ` np.ma.array(x,order= 'C')`

rcurtin: I see---but why are you using the masked array support, and not just `np.array(x, order='C')`?  It seems to me like the mask support isn't needed.

In any case, whatever we do, we should add a test case ensuring that passing a Pandas Dataframe (or Series) to an mlpack binding works.  You could imitate the tests in `src/mlpack/bindings/python/tests/test_python_binding.py` for regular numpy matrices.  I'd prefer to ensure that this code works before we merge it, because it's really easy to introduce segfaults, etc., unless we have thoroughly tested it.

Yashwants19: As you told me `OWNDATA = True` implies that the matrix was copied. when i was using `np.array(x,order= 'C').flags`  it gives me 
  `
  C_CONTIGUOUS : True                                                                                                                                          
  F_CONTIGUOUS : False                                                                                                                                           
  OWNDATA : True                                                                                                                                                
  WRITEABLE : True                                                                                                                                              
  ALIGNED : True                                                                                                                                                
  WRITEBACKIFCOPY : False                                                                                                                                       
  UPDATEIFCOPY : False
`
hence it implies that the matrix was copied.
 But When I was using `np.ma.array(x,order ='C').flags`
its gives me 
  `
  C_CONTIGUOUS : True                                                                                                                                          
  F_CONTIGUOUS : False                                                                                                                                           
  OWNDATA : False                                                                                                                                                
  WRITEABLE : True                                                                                                                                              
  ALIGNED : True                                                                                                                                                
  WRITEBACKIFCOPY : False                                                                                                                                       
  UPDATEIFCOPY : False
`
Hence it implies that the matrix was not copied.

Yashwants19: And I m working on those test cases.

rcurtin: Let's not use the masked array support unless we can understand why that does not cause a copy.  It would be better if we could use `np.array()` and avoid a copy.  You may have to dig into the internals of numpy somewhat to learn more here.

Yashwants19: Hi we can use `y = np.ndarray(x.shape,buffer=x.flatten(),order = 'C')`  where x is  `x = d.to_numpy()`  when i use `y.flags` it gives me
 ```

  C_CONTIGUOUS : True
  F_CONTIGUOUS : False
  OWNDATA : False
  WRITEABLE : True
  ALIGNED : True
  WRITEBACKIFCOPY : False
  UPDATEIFCOPY : False
```

and its type is `numpy.ndarray`

**Comment by Yashwants19** [comment]
Hi @rcurtin please review this.

**Comment by rcurtin** [comment]
Hi @Yashwants19, don't worry, it is on my todo list.  I have been busy and have not had a chance yet.  I get emails for every bit of activity on the project, so don't worry, I did see your updates on this. :)

File: src/mlpack/bindings/python/mlpack/matrix_utils.py [hunk 3]
```diff
@@ -52,10 +52,19 @@ def to_matrix(x, dtype=np.double, copy=False):
...
+    if copy: # Copy the matrix if required.
+      return np.ndarray(x.shape,buffer=x.flatten(),dtype=dtype,order = 'C').copy("C"),True
+    else:
+      return np.ndarray(x.shape,buffer=x.flatten(),dtype=dtype, order = 'C'),False
   else:
-    if isinstance(x, pd.core.series.Series):
-      x = pd.DataFrame(x)
-    return np.array(x, copy=True, dtype=dtype, order='C'), True
+    if isinstance(x, pd.core.series.Series) or isinstance(x, pd.DataFrame):
+      y = x.to_numpy()
```
rcurtin: This only exists in Pandas 0.24 or newer, so we will need to handle cases where the Pandas version is older too.

src/mlpack/bindings/python/mlpack/matrix_utils.py
```diff
@@ -52,10 +52,19 @@ def to_matrix(x, dtype=np.double, copy=False):
       return x.copy("C"), True
     else:
       return x, False
+  elif (isinstance(x, np.ndarray) and x.dtype == dtype and x.flags.f_contiguous):
+    if copy: # Copy the matrix if required.
+      return np.ndarray(x.shape,buffer=x.flatten(),dtype=dtype,order = 'C').copy("C"),True
```
rcurtin: Doesn't this implicitly transpose the matrix?  I think that we should test this by making another test for `test_python_binding.py` where we pass a numpy matrix with `F_CONTIGUOUS` set.  (You can just make a random matrix with `order='F'`.)

src/mlpack/bindings/python/mlpack/matrix_utils.py
Yashwants19: We can use `x.values`

rcurtin: I think you are right, I did some playing with it and `x.values` does not seem to copy the matrix unless it is needed.

src/mlpack/bindings/python/mlpack/matrix_utils.py
Yashwants19: No I think it doesn't transpose the matrix?
```
df = pd.read_csv('rann_test_q_3_100.csv',header=None)
a = pd.Series(np.random.rand(100))
print(df.values.flags)
```
it gives
```
  C_CONTIGUOUS : False
  F_CONTIGUOUS : True
  OWNDATA : False
  WRITEABLE : True
  ALIGNED : True
  WRITEBACKIFCOPY : False
  UPDATEIFCOPY : False
```
when I compare `df.values` and `to_matrix(df.values)[0]`
as `np.array_equal(df.values,to_matrix(df.values)[0])` it results `True`

rcurtin: Huh, this is very strange to me.  Take a look at this Python session:

```
>>> import numpy as np
>>> x = np.random.rand(5, 5)
>>> x = np.array(x, order='F', copy=True)
>>> x
array([[0.4374804 , 0.30758627, 0.97045243, 0.35728207, 0.95584896],
       [0.41457004, 0.82433351, 0.44260369, 0.80797501, 0.78434397],
       [0.10598124, 0.98438937, 0.13709334, 0.24600426, 0.20042625],
       [0.73949954, 0.54399577, 0.64043765, 0.52684565, 0.89393217],
       [0.61319107, 0.67340923, 0.37922999, 0.32466414, 0.91222045]])
>>> x.flags
  C_CONTIGUOUS : False
  F_CONTIGUOUS : True
  OWNDATA : True
  WRITEABLE : True
  ALIGNED : True
  WRITEBACKIFCOPY : False
  UPDATEIFCOPY : False
>>> y = np.ndarray(x.shape, buffer=x.flatten(), order='C')
>>> y
array([[0.4374804 , 0.30758627, 0.97045243, 0.35728207, 0.95584896],
       [0.41457004, 0.82433351, 0.44260369, 0.80797501, 0.78434397],
       [0.10598124, 0.98438937, 0.13709334, 0.24600426, 0.20042625],
       [0.73949954, 0.54399577, 0.64043765, 0.52684565, 0.89393217],
       [0.61319107, 0.67340923, 0.37922999, 0.32466414, 0.91222045]])
>>> y.flags
  C_CONTIGUOUS : True
  F_CONTIGUOUS : False
  OWNDATA : False
  WRITEABLE : True
  ALIGNED : True
  WRITEBACKIFCOPY : False
  UPDATEIFCOPY : False
>>> z = np.ndarray(y.shape, buffer=y.flatten(), order='F')
>>> z
array([[0.4374804 , 0.41457004, 0.10598124, 0.73949954, 0.61319107],
       [0.30758627, 0.82433351, 0.98438937, 0.54399577, 0.67340923],
       [0.97045243, 0.44260369, 0.13709334, 0.64043765, 0.37922999],
       [0.35728207, 0.80797501, 0.24600426, 0.52684565, 0.32466414],
       [0.95584896, 0.78434397, 0.20042625, 0.89393217, 0.91222045]])
>>> z.flags
  C_CONTIGUOUS : False
  F_CONTIGUOUS : True
  OWNDATA : False
  WRITEABLE : True
  ALIGNED : True
  WRITEBACKIFCOPY : False
  UPDATEIFCOPY : False
```

Basically I created a matrix `x` that is random with `F_CONTIGUOUS`.  Then I used `np.ndarray()` with `buffer=x.flatten()` to convert to order `C_CONTIGUOUS`.  But somehow this does not copy the matrix or transpose it (which does not make sense to me---if it is `F_CONTIGUOUS` it is column-major; if it is `C_CONTIGUOUS` it is row-major; converting between the two by flattening should result in different element orderings).  But then, when I do the same thing to convert back to `F_CONTIGUOUS`, now the matrix is transposed (but not copied).

So, that's very strange to me, but it doesn't matter, since you have written a test case with `F_CONTIGUOUS` matrices, so we know it will work either way. :+1:

src/mlpack/bindings/python/mlpack/matrix_utils.py
```diff
@@ -52,10 +52,19 @@ def to_matrix(x, dtype=np.double, copy=False):
...
   else:
-    if isinstance(x, pd.core.series.Series):
-      x = pd.DataFrame(x)
-    return np.array(x, copy=True, dtype=dtype, order='C'), True
+    if isinstance(x, pd.core.series.Series) or isinstance(x, pd.DataFrame):
+      y = x.values
+      if copy: # Copy the matrix if required.
+        return np.ndarray(y.shape,buffer=y.flatten(),dtype=dtype,order = 'C').copy("C"),True
+      else:
+        return np.ndarray(y.shape,buffer=y.flatten(),dtype=dtype, order = 'C'),False
```
rcurtin: What about the `else` for this `if` block?  We should probably fall back on the original `return np.array(x, copy=True, dtype=dtype, order='C'), True` that was there.

src/mlpack/bindings/python/tests/test_python_binding.py
```diff
@@ -140,6 +140,135 @@ def testNumpyMatrixForceCopy(self):
     for j in range(100):
       self.assertEqual(2 * x[j, 2], output['matrix_out'][j, 2])
 
+
+  def testNumpyFContiguousMatrix(self):
+    """
+    The matrix with F_CONTIGUOUS set we pass in, we should get back with the third 
+    dimension doubled and the fifth forgotten.
+    """
+    x = np.array(np.random.rand(100, 5),order = 'F');
```
rcurtin: Can you fix the spacing here (and in general in the PR)?  It should be, e.g.,

```
x = np.array(np.random.rand(100, 5), order='F')
```

The keys here are that there are spaces between each argument, but not spaces between keyword arguments.

src/mlpack/bindings/python/tests/test_python_binding.py
```diff
@@ -140,6 +140,135 @@ def testNumpyMatrixForceCopy(self):
...
+    """
+    Test a Pandas Series input paramter
+    """
+    x =  pd.Series(np.random.rand(100))
+    z = copy.copy(x)
+
+    output = test_python_binding(string_in='hello',
+                                 int_in=12,
+                                 double_in=4.0,
+                                 col_in=z)
```
rcurtin: We should test a Pandas series also to be passed as `matrix_in`.  But that may require the fix for one-dimensional objects which is another part of #1710.  If so, we can wait to add that.

rcurtin: If you can let me know what you think here before we merge, I'd appreciate that. :+1:

Yashwants19: hi @rcurtin when we pass Pandas Series as `matrix_in` it gives Segmentation Fault as I believe this is because the function `numpy_to_mat_d()`, expects a `numpy.ndarray[numpy.double_t, ndim=2]`, but we are passing one with `ndim=1`

rcurtin: Ok, that's what I figured but I just wanted to check.  We can handle it in a different PR then. :)

src/mlpack/bindings/python/mlpack/matrix_utils.py
```diff
@@ -52,10 +52,21 @@ def to_matrix(x, dtype=np.double, copy=False):
...
   else:
-    if isinstance(x, pd.core.series.Series):
-      x = pd.DataFrame(x)
-    return np.array(x, copy=True, dtype=dtype, order='C'), True
+    if isinstance(x, pd.core.series.Series) or isinstance(x, pd.DataFrame):
+      y = x.values
+      if copy: # Copy the matrix if required.
+        return np.ndarray(y.shape, buffer=y.flatten(), dtype=y.dtype, order='C').copy("C"), True
+      else:
+        return np.ndarray(y.shape, buffer=y.flatten(), dtype=y.dtype, order='C'), False
```
rcurtin: Are you sure about the `y.dtype` here?  It's possible that this may not be the desired `dtype`.

Yashwants19: No, I thing we should use `dtype` only, I am Sorry I was testing Pandas series so I have used `y.dtype` . But i realize `dtype` gives desired result correctly. Should I make another commit to correct this.??

rcurtin: Yes, absolutely---we should not commit incorrect code.

Yashwants19: Resolved `dtype`

src/mlpack/bindings/python/tests/test_python_binding.py
Yashwants19: Yes we can test this on different PR :)

rcurtin:
Thanks again for the contribution! :+1:

PR #1713, opened 2019-02-09
Edited to comply with the Style Checks

KimSangYeon-DGU:
I noticed these codes fail to comply with the `Design Guidelines`, and trigger the `Style Checks` failure.


src/mlpack/methods/range_search/rs_model_impl.hpp
```diff
@@ -271,11 +271,12 @@ void MonoSearchVisitor::operator()(RSType* rs) const
 }
 
 //! Save parameters for bichromatic range search.
-inline BiSearchVisitor::BiSearchVisitor(const arma::mat& querySet,
-                                        const math::Range& range,
-                                        std::vector<std::vector<size_t>>& neighbors,
-                                        std::vector<std::vector<double>>& distances,
-                                        const size_t leafSize):
+inline BiSearchVisitor::
+BiSearchVisitor(const arma::mat& querySet,
```
zoq: This looks a little bit strange to me, what do you think about:

```c++
inline BiSearchVisitor::BiSearchVisitor(
    const arma::mat& querySet,
    const math::Range& range,
    std::vector<std::vector<size_t>>& neighbors,
    std::vector<std::vector<double>>& distances,
    const size_t leafSize):
```

zoq:
Agreed, thanks for the great work.

src/mlpack/methods/range_search/rs_model_impl.hpp
KimSangYeon-DGU: It looks so good to me. :) Thanks, I'll edit it.

KimSangYeon-DGU:
There are some `Static Code Analysis` issues, I'll fix them.

src/mlpack/methods/ann/rbm/spike_slab_rbm_impl.hpp
```diff
@@ -119,6 +119,9 @@ RBM<InitializationRuleType, DataType, PolicyType>::Phase(
 
   visiblePenaltyGrad = -0.5 * arma::dot(input, input)
       / std::pow(input.n_cols, 2);
+
+  // To avoid 'unused variable' warning.
+  (void) visiblePenaltyGrad;
```
rcurtin: If it's unused, I'd suggest just plain removing `visiblePenaltyGrad`.  But this raises the question of why it's even computed in the first place.  However, I don't know anything about `SpikeSlabRBM`, so I can't say. :(  @ShikharJ: any thoughts here?

src/mlpack/methods/logistic_regression/logistic_regression_main.cpp
```diff
@@ -286,6 +298,7 @@ static void mlpackMain()
 
       Log::Fatal << "Can't get responses from training data since it has less "
           << "than 2 rows." << endl;
+      return;
```
rcurtin: Technically the `return` here is unnecessary since `Log::Fatal` will throw an exception, but I don't mind putting it in to satisfy the static code analysis checker.

KimSangYeon-DGU: > Technically the `return` here is unnecessary since `Log::Fatal` will throw an exception

Agreed, I was worried about this part. After checking if the `checker` can detect `Log::Fatal` one more, I'll look into another solution to avoid unnecessary things.

src/mlpack/methods/ann/rbm/spike_slab_rbm_impl.hpp
KimSangYeon-DGU: Yes, we can remove `visiblePenaltyGrad` by doing this:
```cpp
gradient(weightGrad.n_elem + spikeBiasGrad.n_elem) = -0.5 * arma::dot(input, input)
      / std::pow(input.n_cols, 2);
```

ShikharJ: It's merely an alias for a part of the `gradient` matrix. I'm fine with removing it, but it might reduce the readability.

src/mlpack/methods/logistic_regression/logistic_regression_main.cpp
```diff
@@ -270,7 +270,10 @@ static void mlpackMain()
     {
       // Clean memory if needed.
       if (!CLI::HasParam("input_model"))
+      {
         delete model;
+        model = nullptr;
```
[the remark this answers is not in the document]
KimSangYeon-DGU: I decided to restore the changes. :)

KimSangYeon-DGU:
Thanks for your great opinion. After conversation with you, I decided to revert the changes in the `logistic_regression_main.cpp` and `spike_slab_rbm_impl.hpp`, because it is not a real problem even fail to pass the `Static Code Analysis Checks`. We can whitelist them.

zoq:

zoq:

zoq:
Thanks again for the great contribution!

src/mlpack/methods/ann/rbm/spike_slab_rbm_impl.hpp
rcurtin: Ah, sorry, I did not recognize it was an alias.  I should have looked more closely. :)

AYESDIE:
This was a great fix :)

PR #1714, opened 2019-02-09
Minor Dice Loss adjustments

zoq:
- include Dice loss into the build routine
- minor style fixes

### DIFF
--- a/src/mlpack/methods/ann/loss_functions/dice_loss.hpp
+++ b/src/mlpack/methods/ann/loss_functions/dice_loss.hpp
@@ -24,21 +24,16 @@ namespace ann /** Artificial Neural Network. */ {
  *
  * For more information see the following.
  *
- * @article{DBLP:journals/corr/MilletariNA16,
- *   author    = {Fausto Milletari and
- *                Nassir Navab and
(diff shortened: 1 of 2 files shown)


zoq:
Since the style check is working again, there are some issues we missed, will fix those in another PR.

src/mlpack/methods/ann/loss_functions/dice_loss.hpp
```diff
@@ -56,7 +51,7 @@ class DiceLoss
   /**
    * Create the DiceLoss object.
    *
-   * @param smooth The parameter to avoid overfitting.
+   * @param smooth The laplace smoothing parameter.
```
rcurtin: Should this be `Laplace`?  I know that is really pedantic though. :)

zoq:
Looks like https://github.com/mlpack/mlpack/pull/1713 was faster, great.

src/mlpack/methods/ann/loss_functions/dice_loss.hpp
zoq: Nice catch!

PR #1716, opened 2019-02-12
Implementation of dense block class

Rajiv2605:


memoiry:
It seems that there is no implementation?

Rajiv2605:
> It seems that there is no implementation?

I'm working on it now :sweat_smile: I'll push the code soon! :)

zoq:
Great, let me know if you need any help with.

Rajiv2605:
Hey @zoq, I had issues with my laptop so couldn't do much work in the past few days... I looked into the implementation and documentation of other `layers`. What is the difference between the functions `Backward()` and `Gradient()`? From the documentation, I understood the `Gradient()` function but I'm not able to understand the purpose of `Backward()` function. It would be great if you could help me out :)

zoq:
The ``Gradient `` function will calculate the gradient (update) of the output using the error from the backward pass. An easy example is the linear layer, let me know if I should describe what each function does in a more detail.

Rajiv2605:
@zoq Yeah, it would be very helpful if you could give an example for the `Backward()` and `Gradient()` functions :)  
I have added the `Forward()` function for the `Dense` class. Since `Dense` class is a combination of layers, I am not able to figure out on how to implement the `Backward()` and `Gradient()` functions. It would be great if you could guide me here.

Rajiv2605:
The `GradientTransposedConvolutionLayerTest` seems to be failing.

Rajiv2605:
@zoq I have implemented dense blocks class but I'm not sure about it's correctness... What test should I use? I had referred the Keras implementation to make it... So can I use it as a benchmark or do you suggest any other testing method?

zoq:
As a start I would implement the gradient check test, see:

https://github.com/mlpack/mlpack/blob/2c91431f2ae10eee336dfbd944dbbb4ba14cc705/src/mlpack/tests/ann_layer_test.cpp#L84-L121

for an example.

src/mlpack/methods/ann/layer/dense.hpp
```diff
@@ -0,0 +1,197 @@
...
+    const double& dropout_rate, const double& weight_decay);
+
+  /**
+   * Ordinary feed forward pass of the dense layer.
+   *
+   * @param input Input data used for evaluating the specified function.
+   * @param output Resulting output activation.
+   */
+  template<typename eT>
+  void Forward(const arma::Cube<eT>&& input, arma::Cube<eT>&& output);
```
zoq: Since the visitor (e.g. parameter) does expect ``arma::Mat`` instead of ``arma::Cube`` it returns with a bunc hof errors, so either we add the ability to handle ``arma::Cube`` as well or we use ``arma::Mat`` here and use the advanced constructor of ``arma::Cube`` to transform ``arma::Mat`` into a ``arma::Cube``. Let me know what you think.

Rajiv2605: I'm not sure if I understood the visitor part, are you talking about the files in `ann/visitor`? I feel that it would be better to add the ability to handle `arma::Cube` as it will be useful for other future implementations as well. Building an advanced constructor here might lead to code redundancies if some other layer also wants to use `arma::Cube`. I'm not sure if I'm correct :sweat_smile: , please let me know what you think :)

zoq: Right I was talking about ``ann/visitor``, if we go with the ``ann/visitor`` adaption we have to think about a way to pass between ``arma::cube`` and ``arma::mat``, e.g. a linear layer has to pass the output to the next layer that will take an ``arma::cube`` as input.

Rajiv2605: Yes :+1: I checked the files `add_visitor.hpp` and `add_visitor_impl.hpp` but could not find `arma::mat` explicitly used anywhere (All the functions are using templates); am I missing something? I checked `ffn.hpp` and it has it's own `Add()`... What exactly is the purpose of the `visitor` classes?

zoq: The visitor paradigma enables us to call different functions on ``boost::variant`` the data structure that is used to hold the layer.

So, since the visitor classes do work on templates, I think the only class we have to adapt is FFN and RNN, since they internally expect that the input/output parameter for each layer doesn't change.

Perhaps it makes more sense to use ``arma::mat`` instead of ``arma::cube`` for the Dense layer as well.

Rajiv2605: How can I concatenate layers in `arma::mat`? The reason I used `arma::cube` is that it had the `join_slices()` function which I couldn't find for the `arma::mat` class. Did I miss some function?

zoq: Sorry for the slow response, you could interpret the cols as slice, you can also transform an ``arma::mat`` into an ``arma::cube`` if you like to use ``join_slices()``:

```c++
arma::cube transformedMat = arma::cube(mat.memptr(), rows, cols, slices, false, false);
```

Rajiv2605: > Sorry for the slow response, you could interpret the cols as slice, you can also transform an `arma::mat` into an `arma::cube` if you like to use `join_slices()`:
> 
> ```c++
> arma::cube transformedMat = arma::cube(mat.memptr(), rows, cols, slices, false, false);
> ```

:+1: I'll try this method and let you know.

PR #1717, opened 2019-02-13
Fix Type Error (Check type of Every parameter)

Yashwants19:
Address #1710

TODOS
- [x] `Casting 2-d to 1-d array or 1-d to 2-d array `
- [x] `Type check for every parameter`
- [x] `Add Test for Thrown Exceptions`

### DIFF
--- a/src/mlpack/bindings/python/mlpack/matrix_utils.py
+++ b/src/mlpack/bindings/python/mlpack/matrix_utils.py
@@ -41,38 +41,31 @@ def to_matrix(x, dtype=np.double, copy=False):
   """
   Given some array-like X, return a numpy ndarray of the same type.
   """
-  # Check the Type of input X
-  if isinstance(x, np.ndarray) or isinstance(x, pd.DataFrame) or isinstance(x,
-      pd.Series) or isinstance(x, list) or isinstance(x, tuple):
+  # Make sure it's array-like at all.
+  if not hasattr(x, '__len__') and \
+      not hasattr(x, 'shape') and \
+      not hasattr(x, '__array__'):
+    raise TypeError("given argument is not array-like")
 
-    # Make sure it's array-like at all.
-    if not hasattr(x, '__len__') and \
-        not hasattr(x, 'shape') and \
-        not hasattr(x, '__array__'):
-      raise TypeError("given argument is not array-like")
-
-    if (isinstance(x, np.ndarray) and x.dtype == dtype and x.flags.c_contiguous):
-      if copy: # Copy the matrix if required.
-        return x.copy("C"), True
-      else:
-        return x, False
-    elif (isinstance(x, np.ndarray) and x.dtype == dtype and x.flags.f_contiguous):
-      if copy: # Copy the matrix if required.
-        return np.ndarray(x.shape, buffer=x.flatten(), dtype=dtype, order='C').copy("C"), True
-      else:
-        return np.ndarray(x.shape, buffer=x.flatten(), dtype=dtype, order='C'), False
+  if (isinstance(x, np.ndarray) and x.dtype == dtype and x.flags.c_contiguous):
+    if copy: # Copy the matrix if required.
+      return x.copy("C"), True
     else:
-      if isinstance(x, pd.core.series.Series) or isinstance(x, pd.DataFrame):
-        y = x.values
-        if copy: # Copy the matrix if required.
-          return np.ndarray(y.shape, buffer=y.flatten(), dtype=dtype, order='C').copy("C"), True
-        else:
-          return np.ndarray(y.shape, buffer=y.flatten(), dtype=dtype, order='C'), False
-      else:
-        return np.array(x, copy=True, dtype=dtype, order='C'), True
+      return x, False
+  elif (isinstance(x, np.ndarray) and x.dtype == dtype and x.flags.f_contiguous):
+    if copy: # Copy the matrix if required.
+      return np.ndarray(x.shape, buffer=x.flatten(), dtype=dtype, order='C').copy("C"), True
+    else:
+      return np.ndarray(x.shape, buffer=x.flatten(), dtype=dtype, order='C'), False
   else:
-    raise TypeError("'x' must have type '(np.ndarray,pd.DataFrame,pd.Series,list,tuple)'!")
-
+    if isinstance(x, pd.core.series.Series) or isinstance(x, pd.DataFrame):
+      y = x.values
+      if copy: # Copy the matrix if required.
+        return np.ndarray(y.shape, buffer=y.flatten(), dtype=dtype, order='C').copy("C"), True
+      else:
+        return np.ndarray(y.shape, buffer=y.flatten(), dtype=dtype, order='C'), False
+    else:
+      return np.array(x, copy=True, dtype=dtype, order='C'), True
     
 
 def to_matrix_with_info(x, dtype, copy=False):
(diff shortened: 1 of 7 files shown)


src/mlpack/bindings/python/print_pyx.cpp
```diff
@@ -87,7 +87,7 @@ void PrintPYX(const ProgramDoc& programInfo,
   cout << "cimport numpy as np" << endl;
   cout << endl;
   cout << "from libcpp.string cimport string" << endl;
-  cout << "from libcpp cimport bool" << endl;
+  cout << "from libcpp cimport bool as bl" << endl;
```
rcurtin: Could we call this `cbool` instead?  I know I am being a bit picky.

src/mlpack/bindings/python/print_pyx.cpp
```diff
@@ -186,10 +186,14 @@ void PrintPYX(const ProgramDoc& programInfo,
...
-  cout << "  if copy_all_inputs:" << endl;
-  cout << "    SetParam[bool](<const string> 'copy_all_inputs', "
+  cout << "  if isinstance(copy_all_inputs, bool):" << endl;
+  cout << "    if copy_all_inputs:" << endl;
+  cout << "      SetParam[bl](<const string> 'copy_all_inputs', "
       << "copy_all_inputs)" << endl;
-  cout << "    CLI.SetPassed(<const string> 'copy_all_inputs')" << endl;
+  cout << "      CLI.SetPassed(<const string> 'copy_all_inputs')" << endl;
+  cout << "  else:" << endl;
+  cout << "    raise TypeError(" <<"\"copy_all_inputs must have type 'bool'!\""
```
rcurtin: Can you fix the style here?  `<< "\'copy_all_inputs\' must have type 'bool'!\""`

Yashwants19:
You suggest should I use on this PR or go for another PR ??

rcurtin:
Up to you, no difference from my side

Yashwants19:
I will go for new PR.

Yashwants19:
Hi @rcurtin Please review this. If I have to change something or not

rcurtin:
Hi @Yashwants19, didn't I already say somewhere else that I will review everything when I have a chance?  You can see there are nearly 60 PRs open.

Yashwants19:
Sorry my mistake @rcurtin

src/mlpack/bindings/python/print_input_processing.hpp
```diff
@@ -122,21 +122,48 @@ void PrintInputProcessing(
   std::cout << prefix << "# Detect if the parameter was passed; set if so."
       << std::endl;
   if (!d.required)
-  {
-    std::cout << prefix << "if " << d.name << " is not None:" << std::endl;
+  { 
+    if (GetArmaType<T>() == "row" || GetArmaType<T>() == "col")
```
rcurtin: I'd suggest not calling out to a separate function here (since its functionality may change) but instead use the members `T::is_col` and `T::is_row` (they are boolean).  I am more confident that those won't change over time.

src/mlpack/bindings/python/print_input_processing.hpp
```diff
@@ -122,21 +122,48 @@ void PrintInputProcessing(
...
+    {
+      std::cout << prefix << "if " << d.name << " is not None:" << std::endl;
```
rcurtin: You could move this little bit outside the `if` since it appears in both the body and the body of the `else`. :)

src/mlpack/bindings/python/print_input_processing.hpp
```diff
@@ -122,21 +122,48 @@ void PrintInputProcessing(
...
+	  << d.name << "_tuple" << "[0].ravel()" << std::endl;
+      std::cout << prefix << "  " << d.name << "_mat = arma_numpy.numpy_to_"
+          << GetArmaType<T>() << "_" << GetNumpyTypeChar<T>() << "(" << d.name
+          << "_tuple[0], " << d.name << "_tuple[1])" << std::endl;
+      std::cout << prefix << "  SetParam[" << GetCythonType<T>(d) << "](<const "
+          << "string> '" << d.name << "', dereference(" << d.name << "_mat))"
+          << std::endl;
+      std::cout << prefix << "  CLI.SetPassed(<const string> '" << d.name 
+	  << "')" << std::endl;
+      std::cout << prefix << "  del " << d.name << "_mat";
```
rcurtin: This is hard to read (and I think that's kind of unavoidable)---can you update the comment block up above (the one that says `this gives code like`) to help make it a little easier?  It took me a little while to figure out what was going on here... :fearful:

src/mlpack/bindings/python/print_input_processing.hpp
```diff
@@ -122,21 +122,48 @@ void PrintInputProcessing(
...
+      std::cout << prefix << "  " << d.name << "_tuple = to_matrix(" << d.name
+          << ", dtype=" << GetNumpyType<typename T::elem_type>() << ", "
+          << "copy=CLI.HasParam('copy_all_inputs'))" << std::endl;
+      std::cout << prefix << "  " << "if len(" << d.name << "_tuple[0].shape"
+	  << ") > 1:" << std::endl;
+      std::cout << prefix << "  " << prefix << "if " << d.name << "_tuple[0]"
+	  << ".shape[0] == 1 or " << d.name << "_tuple[0].shape[1] == 1:" 
+	  << std::endl;
+      std::cout << prefix << "  " << prefix << "  " << d.name << "_tuple[0] = " 
+	  << d.name << "_tuple" << "[0].ravel()" << std::endl;
```
rcurtin: I think you can put these two bits together (`"_tuple" << "[0].ravel()"`).  And actually I think you can simplify the code overall if only lines 135-139 are in the body of the `if`---everything else appears to be the same.

src/mlpack/bindings/python/print_input_processing.hpp
```diff
@@ -122,21 +122,48 @@ void PrintInputProcessing(
...
+    else
+    {
+      std::cout << prefix << "if " << d.name << " is not None:" << std::endl;
+
+      std::cout << prefix << "  " << d.name << "_tuple = to_matrix(" << d.name
+          << ", dtype=" << GetNumpyType<typename T::elem_type>() << ", "
+          << "copy=CLI.HasParam('copy_all_inputs'))" << std::endl;
+      std::cout << prefix << "  " << d.name << "_mat = arma_numpy.numpy_to_"
+          << GetArmaType<T>() << "_" << GetNumpyTypeChar<T>() << "(" << d.name
+          << "_tuple[0], " << d.name << "_tuple[1])" << std::endl;
```
rcurtin: What about the other way (if we get a 1-dimensional numpy array but here we need it to be 2-d)?

Yashwants19:
We can add task list here. Can do both task on this PR only.

src/mlpack/bindings/python/print_input_processing.hpp
Yashwants19: Hi @rcurtin we can solve this by `np.reshape()`

Yashwants19:
Hello @rcurtin, I'm would like to work on the fastmks test for command-line and Python bindings

src/mlpack/bindings/python/print_input_processing.hpp
```diff
@@ -108,35 +108,82 @@ void PrintInputProcessing(
...
+    {
+  /**
+   * This gives us code like:
+   *
+   * # Detect if the parameter was passed; set if so.
+   * if param_name is not None:
+   *   param_name_tuple = to_matrix(param_name)
+   *   if len(param_name_tuple[0].shape) < 2:
+   *     param_name_tuple[0] = np.reshape((param_name_tuple[0]),
+   *	     (param_name_tuple[0].shape[0] , -1))
```
rcurtin: I think you could use `1` instead of `-1` here; there's no need to infer the second dimension when we already know it. :+1:

Yashwants19:
Hi @rcurtin  I was waiting for your review. I will complete these implementation as soon as possible. Thank You :)

Yashwants19:
Hi @rcurtin This is ready for review.

Yashwants19:
Hi @rcurtin Suggest me should I add more type for typecheck and I will add them as soon as possible

src/mlpack/bindings/python/print_input_processing.hpp
```diff
@@ -54,35 +55,132 @@ void PrintInputProcessing(
    *
    * # Detect if the parameter was passed; set if so.
    * if param_name is not None:
+   *   if isinstance(param_name, int):
    *   SetParam[int](<const string> 'param_name', param_name)
    *   CLI.SetPassed(<const string> 'param_name')
```
rcurtin: Oops, for this comment to be correct Python we should indent this bit.  (I'm being picky about this comment because this code below is really hard to read otherwise.  I don't think there's much we can do about that though. :))

src/mlpack/bindings/python/print_input_processing.hpp
```diff
@@ -54,35 +55,132 @@ void PrintInputProcessing(
...
+      std::cout << prefix << "if isinstance(" << name << ", bool):"
+        << std::endl;
+      std::cout << prefix << "  if " << name << " is not " << def << ":"
+        << std::endl;
+    }
+    else
+    {
+      std::cout << prefix << "if " << name << " is not " << def << ":"
+        << std::endl;
+    }
```
rcurtin: It looks like the output here is different for `bool`s... why?  Maybe I missed something.

src/mlpack/bindings/python/print_input_processing.hpp
```diff
@@ -54,35 +55,132 @@ void PrintInputProcessing(
...
+    else if (GetCythonType<T>(d) == "double")
+    {
+      std::cout << prefix << "  if isinstance(" << name << ", float):"
+        << std::endl;
+    }
+    else if (GetCythonType<T>(d) == "int")
+    {
+      std::cout << prefix << "  if isinstance(" << name << ", int):"
+        << std::endl;
+    }
```
rcurtin: Actually I think you can use `GetPrintableType()` here... take a look (`get_printable_type.hpp`) and tell me what you think.

```
std::cout << prefix << "  if isinstance(" << name << ", " << GetPrintableType<T>(d)
    << "):" << std::endl;
```

That could simplify this code quite a lot.

src/mlpack/bindings/python/print_input_processing.hpp
```diff
@@ -54,35 +55,132 @@ void PrintInputProcessing(
...
+      std::cout << "    else:" << std::endl;
+      std::cout << "      raise TypeError(" <<"\"'"<< name
+          <<"' must have type \'float'!\")" << std::endl;
+    }
+    else if (GetCythonType<T>(d) == "int")
+    {
+      std::cout << "    else:" << std::endl;
+      std::cout << "      raise TypeError(" <<"\"'"<< name
+          <<"' must have type \'int'!\")" << std::endl;
+    }
```
rcurtin: Same here, I think you can use `GetPrintableType<>()` to simplify.

src/mlpack/bindings/python/print_input_processing.hpp
```diff
@@ -91,15 +189,70 @@ void PrintInputProcessing(
...
  * Print input processing for a matrix type.
  */
+
+/**
+ * This gives us code like:
+ *
+ * # Detect if the parameter was passed; set if so.
+ * if param_name is not None:
+ *   if isinstance(param_name, np.ndarray) or isinstance(param_name,
+ *       pd.DataFrame) or isinstance(param_name, pd.Series):
```
rcurtin: I think we can accept `list` here too (really any "arraylike" works).  Actually I would be okay with encoding the check in `to_matrix()` directly, since `to_matrix()` is the method that actually controls whether or not the type can be accepted.

src/mlpack/bindings/python/print_input_processing.hpp
```diff
@@ -244,47 +497,72 @@ void PrintInputProcessing(
...
+   *   if isinstance(param_name, np.ndarray) or isinstance(param_name,
+   *       pd.DataFrame) or isinstance(param_name, pd.Series):
+   *    param_name_tuple = to_matrix_with_info(param_name)
+   *    param_name_mat = arma_numpy.numpy_to_matrix_d(param_name_tuple[0])
+   *    SetParamWithInfo[mat](<const string> 'param_name',
+   *        dereference(param_name_mat), &param_name_tuple[1][0])
+   *    CLI.SetPassed(<const string> 'param_name')
+   *   else:
+   *     raise TypeError("'param_name' must have type 
+   *         '(np.ndarray,pd.DataFrame,pd.Series)'!")
```
rcurtin: Same comment as earlier about `to_matrix()`, but applied here to `to_matrix_with_info()`.  Let me know what you think.

src/mlpack/bindings/python/tests/test_python_binding_main.cpp
```diff
@@ -30,6 +30,7 @@ PARAM_DOUBLE_IN_REQ("double_in", "Input double, must be 4.0.", "d");
 PARAM_FLAG("flag1", "Input flag, must be specified.", "f");
 PARAM_FLAG("flag2", "Input flag, must not be specified.", "F");
 PARAM_MATRIX_IN("matrix_in", "Input matrix.", "m");
+PARAM_MATRIX_IN("smatrix_in", "Input matrix.", "");
```
rcurtin: It seems to me like we could use the regular `matrix_in` parameter here.  Or is there some reason we need `smatrix_in` and `smatrix_out`?

src/mlpack/bindings/python/tests/test_python_binding.py
```diff
@@ -608,5 +648,142 @@ def testModelForceCopy(self):
...
+      self.assertEqual(output['col_out'][i], x[i] * 2)
+
+
+
+    
+  def testThrownException(self):
+
+    """
+    Test that we pass wrong type and get back TypeError
+    """
```
rcurtin: Nice, thanks for testing this. :+1:

src/mlpack/bindings/python/tests/test_python_binding.py
```diff
@@ -608,5 +648,142 @@ def testModelForceCopy(self):
...
+                                                              int_in=12,
+                                                              double_in=4.0,
+                                                              flag1=True,
+                                                              matrix_and_info_in = 10.0))
+
+    self.assertRaises(TypeError, lambda : test_python_binding(string_in='hello',
+                                                              int_in=12,
+                                                              double_in=4.0,
+                                                              flag1=True,
+                                                              copy_all_inputs = np.random.randn(10)))
```
rcurtin: I think this line is too long (and maybe others too), can you make sure these are all 80 characters?  Or I can do it during merge, up to you.  Our style checker doesn't run on Python code. :)

src/mlpack/bindings/python/print_input_processing.hpp
Yashwants19: If we use `GetPrintableType()` at the place of `int` or `float` or `str` it is not recognized by python as there is not datatype name `double` or `string`

src/mlpack/bindings/python/tests/test_python_binding_main.cpp
Yashwants19: I have to test pandas.Series as matrix which is 1 dim so i use another parameter name `smatrix`

src/mlpack/bindings/python/print_input_processing.hpp
Yashwants19: Here first I check the parameter as it is bool and then I pass the parameter as True or False

Yashwants19:
Hi @rcurtin can you please help me with these issues.

src/mlpack/bindings/python/tests/test_python_binding.py
Yashwants19: Thank you :)

Yashwants19:
Hi @rcurtin This PR is set to be merged

src/mlpack/bindings/python/print_input_processing.hpp
rcurtin: I see, but the code is not doing the `isinstance()` check if the default value is not `False`.

I think you are doing the `def == "False"` check as another way if checking that the type `T` is `bool`, but honestly I think we can avoid the check entirely if we use a function like `GetPrintableType<>()` for the type part of the `isinstance()` call.

src/mlpack/bindings/python/print_input_processing.hpp
rcurtin: That's a fair point.  The `GetPrintableType<>()` functions are used solely for printing documentation, so I would be okay adapting the typenames returned by that function to be the correct Python type.  Specifically I can see in the code that `double` should change to `float` and `string` to `str`.  But this still does not fully solve the problem for `list of ...` and the matrix types.

I think that for list types, we should make another overload of this function `PrintInputProcessing()` that is specifically for vector types.  This would mean using SFINAE to make the following function:

```
template<typename T>
void PrintInputProcessing(
    const util::ParamData& d,
    const size_t indent,
    const typename boost::enable_if<util::IsStdVector<T>::type* = 0,
    const typename boost::disable_if<arma::is_arma_type<T>>::type* = 0,
    const typename boost::disable_if<data::HasSerialize<T>>::type* = 0,
    const typename boost::disable_if<std::is_same<T,
        std::tuple<data::DatasetInfo, arma::mat>>>::type* = 0)
```

and the implementation of that function would be basically the same as the one I am commenting on here, except that we would need to test two things: `isinstance(name, list)` and, if the length of the list is greater than 0, `isinstance(name[0], PT)` where `PT` is the Python type of the list element (you could access this type in C++ as `T::value_type`, so I think you could call `GetPrintableType<T::value_type>()` to get the right result here.

In addition, because of the SFINAE signature, you'll have to add the following parameter to all of the other `PrintInputProcessing()` overloads (except the one at the bottom of the file):

```
const typename boost::disable_if<util::IsStdVector<T>>::type* = 0
```

Do let me know if I can clarify any of this.  SFINAE code can be ugly and confusing...

src/mlpack/bindings/python/tests/test_python_binding_main.cpp
```diff
@@ -105,6 +109,24 @@ static void mlpackMain()
...
+    out *= 2.0;
+
+    CLI::GetParam<arma::mat>("smatrix_out") = move(out);
+  }
+  // An input matrix(pandas.Series) should have all elements multiplied by two.
+  if (CLI::HasParam("s_umatrix_in"))
+  {
+    arma::Mat<size_t> out =
+        move(CLI::GetParam<arma::Mat<size_t>>("s_umatrix_in"));
+    out.row(0)*= 2.0;
```
rcurtin: You should be able to just do `out *= 2` here, assuming your intention is to multiply every element by 2 (I think that all inputs to `smatrix_in` and `s_umatrix_in` have only one row).

src/mlpack/bindings/python/tests/test_python_binding_main.cpp
rcurtin: Ah, okay, understood.

Yashwants19:
Sorry @rcurtin  just ignore that I accidentally closed this Pull request while typing and had to reopen it

Yashwants19:
Hi @rcurtin I have made the changes as suggested

src/mlpack/bindings/python/tests/test_python_binding.py
```diff
@@ -713,6 +749,308 @@ def testModel(self):
...
+    """
+    x = pd.DataFrame(np.random.rand(10), columns=list('a'))
+    x['e'] = pd.Series(['a', 'b', 'c', 'd', 'a', 'b', 'e', 'c', 'a', 'b'],
+        dtype='category')
+    z = copy.copy(x)
+
+    output = test_python_binding(string_in='hello',
+                                 int_in=12,
+                                 double_in=4.0,
+                                 matrix_and_info_in=z)
```
rcurtin: Hmm, technically I think the idea would be that we pass `x['e']` directly here?

Yashwants19: Here I pass `x ` i.e. 1-D and it is without column from `matrix_and_info_in`

Yashwants19:
Hi @rcurtin This is ready for review. I have made the changes as suggested

src/mlpack/bindings/python/mlpack/matrix_utils.py
```diff
@@ -41,129 +41,142 @@ def to_matrix(x, dtype=np.double, copy=False):
...
-      y = x.values
+  # Check the Type of input X
+  if isinstance(x, np.ndarray) or isinstance(x, pd.DataFrame) or isinstance(x,
+      pd.Series) or isinstance(x, list) or isinstance(x, tuple):
+
+    # Make sure it's array-like at all.
+    if not hasattr(x, '__len__') and \
+        not hasattr(x, 'shape') and \
+        not hasattr(x, '__array__'):
+      raise TypeError("given argument is not array-like")
```
rcurtin: Ah, sorry, I meant that the check is already in these functions, and it's this one right here.  So I think we can leave `to_matrix()` and `to_matrix_with_info()` as they were before the changes you made, since they already throw `TypeError`s if they are not array-like.  (A nice side-effect of this way of checking, just for `__len__`, `shape`, and `__array__`, means that other array-like things can also be passed to Python bindings, not just numpy and Pandas matrices.)

src/mlpack/bindings/python/tests/test_python_binding.py
rcurtin: Well, now I'm a little bit confused.  I thought that this test was testing that passing a Pandas series worked, but it seems like you're just passing a 2-D Pandas dataframe with one column.

src/mlpack/bindings/python/tests/test_python_binding.py
```diff
@@ -187,51 +187,42 @@ def testNumpyFContiguousMatrixForceCopy(self):
     for j in range(100):
       self.assertEqual(2 * x[j, 2], output['matrix_out'][j, 2])
 
-  def testPandasDataFrameMatrix(self):
+  def testPandasSeriesMatrix(self):
```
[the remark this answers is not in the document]
rcurtin: Thanks for adding that and the other tests. :+1:

src/mlpack/bindings/python/tests/test_python_binding.py
Yashwants19: Before I was testing that if we can pass 1-D matrix after casting it  to 2-D (as `matrix_and_info_in_reshape`) from  `arma_numpy.numpy_to_mat_d()`

Yashwants19:
Hi @rcurtin I have made the changes as suggested.

src/mlpack/bindings/python/tests/test_python_binding.py
rcurtin: Ah, okay, but I don't think that what we're passing is actually a 1-d Pandas object.  For that we'd need to pass a `Series` object, so I guess we'd need to do `z[0]` here in the call for `matrix_and_info_in`.

rcurtin: By the way, everything else in the PR looks good to me, so I'm ready to approve it once we handle this last comment.

Yashwants19:
Hi @rcurtin This PR is set to be merged

src/mlpack/bindings/python/mlpack/matrix_utils.py
```diff
@@ -81,7 +81,10 @@ def to_matrix_with_info(x, dtype, copy=False):
 
   if isinstance(x, np.ndarray):
     # It is already an ndarray, so the vector of info is all 0s (all numeric).
-    d = np.zeros([x.shape[1]], dtype=np.bool)
+    if len(x.shape) < 2:
+      d = np.zeros(0, dtype=np.bool)
```
rcurtin: Shouldn't this just be 1 in this case though? Correct me if I'm wrong.  This might work and run, but I suspect that there is an invalid memory access from the mlpack side if `d` really has length 0.  This comment applies to the other `d = np.zeros(0, dtype=np.bool)` too.

Yashwants19: Hi @rcurtin I was also bit confused about this. Thank you for correcting me. :)

Yashwants19:
Hi @rcurtin if no one is working on #1492(go bindings). I would like to continue this.

Yashwants19:
Thanks @rcurtin without your help, it couldn't be implemented. I will start working on Go bindings as soon as my exams are over. :)

rcurtin:
@Yashwants19 thanks so much for the hard work on this.  It's really great to have this merged and it fixes some very big problems.  I'll resolve all the related tickets and hopefully nobody will need to reopen them. :)

Issue #1695, opened 2019-01-30
Cannot use saved network outputs rnn


saksham189:
In `rnn` class we can call `Evaluate` with `deterministic = false` which saves the output parameters from each layer of rnn. 
https://github.com/mlpack/mlpack/blob/7d448192cc045555e36ee76a2dca51ce9637ac66/src/mlpack/methods/ann/rnn_impl.hpp#L217-L224

But I don't think there is any way to use these saved parameters for training the rnn. 

- `EvaluateWithGradient` will save its own parameters during forward pass. 
- `Forward()`, `Gradient()`, `Backward()` are all private methods.

Please let me know if I am missing something. @zoq

mirraaj:
@saksham189 I am trying to understand the issue. I believe we are saving the parameters in the ```Evaluate``` function. Please elaborate on what is meant by, ```I don't think there is any way to use these saved parameters for training the rnn.``` I would like to try to answer the question as per my knowledge.

saksham189:
> Please elaborate on what is meant by, `I don't think there is any way to use these saved parameters for training the rnn.`
These parameters are saved but it is not possible to use these for training. @mirraaj 
@zoq Please let me know what you think.

saksham189:
@zoq Please let me know what you think. 
One easy solution would be to remove the overload of `Evaluate()` that allows to save the parameters.

zoq:
Just to be clear you are talking about the network weights? In this case, you could use the serialization method: https://www.mlpack.org/docs/mlpack-git/doxygen/anntutorial.html#model_saving_loading_anntut or:

```c++
// Get network weights.
arma::mat weights = model.Parameters();

// Set network weights.
model.Parameters() = weights;
model.ResetParameters();
```

Let me know what you think.

saksham189:
Sorry, I think I used the wrong terminology. I meant to say the network outputs (rather than parameters) for each layer that are stored in the `moduleOutputParameter` vector in the `Evaluate` method.

saksham189:
@zoq @ShikharJ I do think this is a valid issue. There is no way to use the output parameters stored in `moduleOutputParameter` vector from `Evaluate` method for training.

zoq:
I see, do you think looping over all layer of the model could work to get the output? Something like:

```c++
for (size_t = 0; i < model.Model().size(); i++)
{
	model.Model()[i].OutputParameter();
}
```

saksham189:
[...]

PR #1720, opened 2019-02-15
Remove saving parameters in Evaluate

saksham189:
Closes https://github.com/mlpack/mlpack/issues/1695


zoq:
Thanks for opening the PR, did you look into other places where ``LoadOutputParameterVisitor`` and ``SaveOutputParameterVisitor`` is used as well?

saksham189:
yes, apart from `rnn`, `LoadOutputParameterVisitor` and `SaveOutputParameterVisitor` are used in `recurrent_attention_impl` as well but there is no such problem in the implementation of the layer.

zoq:

Issue #1715, opened 2019-02-10
Adapting test cases after modifying train functions.


walragatver:
This issue is a next step of #1483 Issue. 
After returning final objective from train function's it looks like some of the `tests` can be restructured.Also some `tests` are needed to be returned to ensure that always a reasonable `value` is returned. 

This is a nice issue for newbie's who want to understand the code base. Currently I am working on this, but if you are interested please feel free.

shardulparab97:
@walragatver we could work on this together?

walragatver:
> @walragatver we could work on this together?

@shardulparab97 Sorry for late reply. I think it will be better if only one of us work on it as it is not that big issue. So if you want to work on it just let me know.

ShikharJ:
Fixed in #1724.

PR #1724, opened 2019-02-19
Adapting testcases

walragatver:
Hi, this is a PR for #1715 issue.

Following are the things I have done.
1. Modified `Adaboost` Testcases and removed `ztProduct` from `adaboost` algorithm.
2. Added Test cases to other algorithms to make sure that value returned by `Train()` function is finite.
3. Also Fixed one or two warnings and bugs.
4. I have not added separate test case to some ANN's as they were quite big. But modified previous one's to check whether a finite value is returned or not.

`Imp` : After modifying GAN's test case it looks like `Train()` function is returning `NAN`. I have kept it as it is for now. It looks like a big test case. How to resolve it? It will be great help if you look at it soon as possible.


src/mlpack/tests/adaboost_test.cpp
```diff
@@ -62,7 +65,9 @@ BOOST_AUTO_TEST_CASE(HammingLossBoundIris)
       countError++;
   double hammingLoss = (double) countError / labels.n_cols;
 
-  BOOST_REQUIRE_LE(hammingLoss, a.ZtProduct());
+  // Check that ztProduct is finite.
+  BOOST_REQUIRE_EQUAL(boost::math::fpclassify(ztProduct), FP_NORMAL);
```
rcurtin: Probably easier to use `std::isfinite()` here, then we don't need to bring in the Boost libraries. :+1:

src/mlpack/methods/adaboost/adaboost_impl.hpp
```diff
@@ -246,7 +245,6 @@ void AdaBoost<WeakLearnerType, MatType>::serialize(Archive& ar,
 {
   ar & BOOST_SERIALIZATION_NVP(numClasses);
   ar & BOOST_SERIALIZATION_NVP(tolerance);
-  ar & BOOST_SERIALIZATION_NVP(ztProduct);
```
rcurtin: Ah, there is a little bit of trickiness here... we need to preserve reverse compatibility: if a user has an AdaBoost model saved with an old version, and we try to load with this version, it will crash.  Here's some code to do that:

```
template<typename WeakLearnerType, typename MatType>
template<typename Archive>
void AdaBoost<WeakLearnerType, MatType>::serialize(Archive& ar,
const unsigned int version)
{
  ar & BOOST_SERIALIZATION_NVP(numClasses);
  ar & BOOST_SERIALIZATION_NVP(tolerance);
  if (version == 0 && Archive::is_loading::value)
  {
    // Load unused ztProduct double and forget it.
    double tmpZtProduct = 0.0;
    ar & BOOST_SERIALIZATION_NVP(ztProduct);
  }
  ar & BOOST_SERIALIZATION_NVP(alpha);
}
```

Then, outside the class, you can add this (adapted from `src/mlpack/methods/ffn/ffn.hpp`):

```
//! Set the serialization version of the FFN class.  Multiple template arguments
//! makes this ugly...
namespace boost {
namespace serialization {

template<typename WeakLearnerType,
         typename MatType>
struct version<
    mlpack::adaboost::AdaBoost<WeakLearnerType, MatType>>
{
  BOOST_STATIC_CONSTANT(int, value = 1);
};

} // namespace serialization
} // namespace boost
```

src/mlpack/tests/random_forest_test.cpp
```diff
@@ -281,8 +282,8 @@ BOOST_AUTO_TEST_CASE(WeightedCategoricalLearningTest)
   arma::Row<size_t> fullLabels = arma::join_rows(trainingLabels, randomLabels);
 
   // Build a random forest and a decision tree.
-  RandomForest<> rf(fullData, di, fullLabels, 5, 15 /* 15 trees */, 5);
-  DecisionTree<> dt(fullData, di, fullLabels, 5, 5);
+  RandomForest<> rf(fullData, di, fullLabels, 5, weights, 15 /* 15 trees */, 5);
+  DecisionTree<> dt(fullData, di, fullLabels, 5, weights, 5);
```
rcurtin: Oh!  Nice catch!  Thank you! :+1:

src/mlpack/tests/gan_test.cpp
```diff
@@ -215,7 +216,10 @@ BOOST_AUTO_TEST_CASE(GANMNISTTest)
       discriminatorPreTrain, multiplier);
 
   Log::Info << "Training..." << std::endl;
-  gan.Train(optimizer);
+  double objVal = gan.Train(optimizer);
+
+  // Test that objective value returned by GAN::Train() is finite.
+  BOOST_REQUIRE_EQUAL(boost::math::fpclassify(objVal), FP_NAN);
```
rcurtin: Hmm, if this gave a loss of NaN, try maybe reducing `numIterations` to something like 100 or 1000, and maybe making the step size smaller.

src/mlpack/methods/adaboost/adaboost_impl.hpp
walragatver: Okay. 
With this I also got a chance to know boost serialization. Thanks :+1:

src/mlpack/tests/adaboost_test.cpp
walragatver: Yes, actually I saw everything in testcases is made with `boost` so decided to go with `boost`. Later on noticed the unwanted overhead due to inclusion of boost library as `Std` is already included .Sorry for that.

src/mlpack/tests/gan_test.cpp
walragatver: Okay, I have digged in somewhat and found following things.
1. We are only training on 10 training samples.
2. Number of iterations(`numIterations`) is only two.

Also I found that the layers of `generator` and `discriminator` are  same in `DCGAN`, `WGAN` and `WGANP` while  they are different in `StandardGan` testcase. 
Whereas , other parameters like `multiplier` , `iterations` etc. are same in all four. So I copied the
`generator` and `discriminator` layers of `DCGAN` in `standardGAN` testcase. And surprisingly, the error got eliminated. 
So there can be any of the following,
a) `parameters` of `StandardGan` are wrong.
b) `Layers` of `StandardGan`  are wrong.

However, I also read comments on top of testcase. It read `oreilly test on gan implementation` in `StandardGan` , whereas there were no such comments in other three. Hence I am not sure that the changes I have made is  a prefect solution.
If that's wrong solution just let me know. If I am going wrong it will be great help if you provide good resource link `oreilly test`, so that I can go more deeper.

walragatver:
Hi @rcurtin ,
I have made the corrections.  
 Also I have modified `DCGAN` , `WGAN` , and `CNN` testcases.

src/mlpack/tests/gan_test.cpp
rcurtin: Looks like you got it working, thanks for digging in.  It seems like nothing else is actually tested in that test case, which is a bit strange and leads me to wonder if it could be removed, but we can leave it there for now I guess.

walragatver:
Hi @rcurtin ,
Thanks for the approval :+1: . I have removed the merge conflicts which were occuring.
Also I have gone through #1735 issue. If no one works on it I will fix it after some time.

src/mlpack/tests/convolutional_network_test.cpp
```diff
@@ -129,4 +129,79 @@ BOOST_AUTO_TEST_CASE(VanillaNetworkTest)
...
+  model.Add<Linear<> >(10, 2);
+  model.Add<LogSoftMax<> >();
+
+  // Train for only 8 epochs.
+  ens::RMSProp opt(0.001, 1, 0.88, 1e-8, 8 * nPoints, -1);
+
+  double objVal = model.Train(X, Y, opt);
+
+  // Test that objective value returned by FFN::Train() is finite.
+  BOOST_REQUIRE_EQUAL(std::isfinite(objVal), true);
```
zoq: I wonder if it makes sense to include the check into the existing cnn test just to keep the test time low. Let me know what you think.

walragatver: Yes, you are correct. Actually I forgot to run `time` on `CNN` tests, sorry. I have made the changes.

walragatver:
Hi @zoq,
I have made the required changes. It looks like a new `Bidirectional RNN` module is added in mlpack but it's train function is not returning it's `final objective`. Should I modify that?

zoq:
> I have adapted the BRNN module

Great, thanks.

src/mlpack/tests/lars_test.cpp
```diff
@@ -352,4 +352,51 @@ BOOST_AUTO_TEST_CASE(TrainingConstructorWithNonDefaultsTest)
...
+
+  // Test with Cholesky decomposition and with lasso.
+  LARS lars1(true, lambda1, 0.0);
+  arma::vec betaOpt1;
+  maxCorr = lars1.Train(X, y, betaOpt1);
+
+  BOOST_REQUIRE_EQUAL(std::isfinite(maxCorr), true);
+
+  // Test without Cholesky decomposition and with lasso.
+
```
zoq: Do you mind to remove the extra line here.

src/mlpack/tests/local_coordinate_coding_test.cpp
```diff
@@ -165,4 +165,29 @@ BOOST_AUTO_TEST_CASE(SerializationTest)
...
+BOOST_AUTO_TEST_CASE(LocalCoordinateCodingTrainReturnObjective)
+{
+  double lambda1 = 0.1;
+  uword nAtoms = 10;
+
+  mat X;
+  X.load("mnist_first250_training_4s_and_9s.arm");
+  uword nPoints = X.n_cols;
+
+  // normalize each point since these are images
```
zoq: This should start with a capital letter and end with a ``.``.

walragatver:
Hi @zoq,
I have made the changes. Such silly mistakes will not happen again. :+1:

src/mlpack/tests/gan_test.cpp
```diff
@@ -173,34 +173,43 @@ BOOST_AUTO_TEST_CASE(GANMNISTTest)
...
   generator.Add<ReLULayer<> >();
-  generator.Add<BilinearInterpolation<> >(28, 28, 56, 56, noiseDim / 4);
-  generator.Add<Convolution<> >(noiseDim / 4, 1, 3, 3, 2, 2, 1, 1, 56, 56);
+  generator.Add<TransposedConvolution<> >(2 * dNumKernels, dNumKernels, 8, 8,
+      1, 1, 1, 1, 7, 7);
+  generator.Add<BatchNorm<> >(6272);
+  generator.Add<ReLULayer<> >();
+  generator.Add<TransposedConvolution<> >(dNumKernels, 1, 15, 15, 1, 1, 1, 1,
+      14, 14);
   generator.Add<TanHLayer<> >();
```
ShikharJ: I am -1 for this change. When you change the layers for the `StandardGAN`, and use the ones used in `DCGAN`, `WGAN` or `WGAN-GP`, it doesn't stay the same `StandardGAN` model implemented by Goodfellow et al.

src/mlpack/tests/gan_test.cpp
ShikharJ: @rcurtin That test is primarily an indicator of the working of the `GAN` as a whole. The one above it checks for the logical correctness. This is a generative network, so if you want to check the results, you can just save the output from this test, and see for yourself. Ideally, this should have been moved to the models repository, but if we had actually done that, we wouldn't have caught the NaN error if we simply merged this PR right? That would have lead to further confusion.

src/mlpack/tests/gan_test.cpp
walragatver: Hi @ShikharJ, thanks for reviewing this PR.
I went through previous PR's which you made. It looks like you have commented that test case #1204  and removed those comments in #1417. So I thought it's a normal test case. I am highly sorry for that. What is the correct thing to do about the given test case?

It looks like in #1417 you reduced `numEpochs`, `batchSize`, `discriminatorPretrain`, parameters that might be one of the reasons for the model failing to converge. I am not sure of this but that  can be one of the possibilities.

Also can you provide me the link where Goodfellow and others implemented this model. I tried to search it on internet but failed to found anything. It will be a great help.

rcurtin: Just FYI, @ShikharJ, when you and @walragatver get this bit figured out, I think this PR is ready to merge from my end, so feel free to do so. :+1:  I don't have an opinion about the GAN tests (and sorry if I left any comments that led us in the wrong direction before); I'm not 100% knowledgeable about GANs. :)

walragatver:
Hi @ShikharJ ,

Following is the output after using `previous` Layers with the given `parameters`.
prevLayerslessParamCropped

Following is the output after using `Changed` Layers with the given `parameters`.
newLayerLessParamCropped

Following is the output after using `Previous` layers with the `commented parameters` in #1204.
commentParamPrevLayersCropped

Following is the output after using `Changed` Layers with the `commented parameters` in #1204.
commentParamNewLayers


These outputs are quite strange so I have decided to comment the test case with `previous layers` for now.
I will be going in more depth soon.
Actually frankly speaking I am stuck with exams in the next five days, but will try to be in touch with you regarding this.

Following are some other things to note.
1. I have used `g++` to test the test case  on my system. 
2. Running time of any of the `four cases` mentioned above didn't exceed 30 minutes.(May be an indication to error).
3. I am training on `mnist_first250_training_4s_and_9s.arm`.


Please let me know if I am going in wrong direction.
Also I have tried to `improve gan serialization`. I will be making a PR about it soon. Please review it as well.

ShikharJ:
@walragatver This is quite weird, since I did get a reasonable output even on the GAN test (I think I commented them in one of the later PRs as well). If somehow changes made to the underlying API altered the results, the. I'll need sometime to review. But thanks for letting me know. I'll get to it later this week.

walragatver:
Hi @ShikharJ, 
 `Update`: When working on #1770 I have found that the test in `gan_test.cpp` is throwing an error during serialization which is `signal: integer divide by zero; address of failing instruction: 0x
55ffd3854bd1`. 
I also tried to change hyperparameters but didn't worked as well. So I think there may be small glitch in `layers` of the test. I will go in more depth soon.
All other test cases in `DCGAN`,`WGAN` are serializing properly.

walragatver:
Hi @ShikharJ, @zoq, @rcurtin,
I went through the layers, parameters in it are looking correct.
But, I also went through `ensmallen's adam` code. It is using `sgd optimiser`. So, After looking in `sgd_impl.hpp`  I found that on `line 86` the training gets terminated if it hits `nan`. So this may be the reason for not complete training of the model.

Also I searched somewhat on Internet, and I found this.
https://github.com/goodfeli/adversarial
This repo contain experiments conducted by `goodfellow` mentioned in the `research paper` of `GANS` in this they have implemented a simple model  with `few layers` for MNIST dataset.
But it has a limitation that it is using `MaxOut` layer in the model of `discriminator`. It looks we don't have that layer in `mlpack`. So I thought it's better to consult with you before going ahead. Also there are no other parameters it is using like `multiplier` which we are having, so we may need to write those `parameters` by ourselves. 

What to do next? Should we write the test mentioned in the link or should continue with the given test?

walragatver:
Hi @ShikharJ,
I went through this https://www.youtube.com/watch?v=myGAju4L7O8 and have found one error.

As we are adding the `TanHLayer` at the end of generator our output from generator is `normalized`  from -1 to 1. But in our test case implementation, input from `real images` is `not normailized`, they are still `rgb values` from `0 to 255`. So we need to normalize them. This can be the reason for failure of test case. 

Do we have any module to normalize the inputs? 

I will verify it if that's the reason and will push the changes soon.
 
Also thanks for the video it involved quite cool tricks.

ShikharJ:
@walragatver Read [this](https://github.com/mlpack/mlpack/pull/1813#pullrequestreview-217921686) thread. A similar point was raised by @saksham189 as well. I highly doubt that mere normalization would fix this issue. I have previously run the code more than a dozen times, with and without `BatchNorm`, but it didn't lead to any significant difference in output quality. I think the core reason behind the failure to converge is something that he fixed in #1813, which sadly I didn't had the time to test lately, but I'm planning on doing so soon. If I remember right, we did have a PR which we merged regarding setting the correct random seed, so maybe those changes could've triggered the convergence failure, as it is totally possible that we didn't set the seed correctly, and somehow got lucky with the values.

ShikharJ:
I think what you can try doing is maybe you can revert the changes, just for the standard `GAN`. I don't think these changes should wait for a fix on `GAN`. Let me know if its okay to you, and I can give a final review and merge this in.

walragatver:
Hi @ShikharJ ,
I am not taking about `BatchNorm`. I am talking about normalizing input's between `-1 to 1`.
I will make changes what I intend to say in one or two days. Hopefully that fixes it.
BTW, I have reviewed that thread already. 
I am fine with delay. Let's fix this and then only merge this.
Thanks for the concern 👍 .

rcurtin:
Hey @ShikharJ and @walragatver, there was some discussion in IRC about merging this PR before 3.1.0 is released, but we'll be pulling the GAN support since it is not ready yet. With that in mind, do we want to merge this as-is and handle the GAN issues later?

walragatver:
Hi @rcurtin , @ShikharJ,
I have reverted the changes in `gan_test.cpp`.
This is now ready to merge.
After this we can also merge #1813 as it contains some important changes in `gans`
I am working on fixing the test case. It will be fixed soon.

src/mlpack/tests/feedforward_network_test.cpp
```diff
@@ -579,4 +579,38 @@ BOOST_AUTO_TEST_CASE(PartialForwardTest)
...
+
+  arma::mat testData;
+  data::Load("thyroid_test.csv", testData, true);
+
+  arma::mat testLabels = testData.row(testData.n_rows - 1);
+  testData.shed_row(testData.n_rows - 1);
+
+  // Vanilla neural net with logistic activation function.
+  // Because 92 percent of the patients are not hyperthyroid the neural
+  // network must be significant better than 92%.
```
ShikharJ: I think it should be `significantly`.

src/mlpack/tests/decision_stump_test.cpp
```diff
@@ -391,4 +391,37 @@ BOOST_AUTO_TEST_CASE(IntTest)
...
+  Mat<size_t> labelsIn;
+  labelsIn << 0 << 1 << 0 << 1 << 0 << 1;
+
+  arma::Row<double> weights = arma::ones<arma::Row<double>>(labelsIn.n_elem);
+
+  double gain;
+
+  // Train a simple decision stump without weights.
+  DecisionStump<> ds;
+  gain = ds.Train(trainingData, labelsIn.row(0), numClasses, inpBucketSize);
```
ShikharJ: You can combine this line an `411`.

src/mlpack/tests/decision_tree_test.cpp
```diff
@@ -1126,4 +1126,60 @@ BOOST_AUTO_TEST_CASE(RegularisedDecisionTree)
...
+  weights.ones();
+
+  for (size_t i = 0; i < 1000; ++i)
+    labels[i] = i % 3; // 3 classes.
+
+  double entropy;
+
+  // Train a simpe tree on numeric dataset.
+  DecisionTree<> d(3);
+  entropy = d.Train(dataset, labels, 3, 50);
```
ShikharJ: Please combine this and `1142`.

src/mlpack/tests/decision_tree_test.cpp
```diff
@@ -1126,4 +1126,60 @@ BOOST_AUTO_TEST_CASE(RegularisedDecisionTree)
...
+
+  arma::Row<double> weights = arma::ones<arma::Row<double>>(l.n_elem);
+
+  double entropy;
+
+  // Train a simple tree on categorical dataset.
+  DecisionTree<> dtree(5);
+  entropy = dtree.Train(d, di, l, 5, 10);
+
+  BOOST_REQUIRE_EQUAL(std::isfinite(entropy), true);
```
ShikharJ: Same here.

src/mlpack/tests/lars_test.cpp
```diff
@@ -352,4 +352,50 @@ BOOST_AUTO_TEST_CASE(TrainingConstructorWithNonDefaultsTest)
...
+  arma::rowvec y = Y.row(0);
+
+  double maxCorr;
+  double lambda1 = 0.1;
+  double lambda2 = 0.1;
+
+  // Test with Cholesky decomposition and with lasso.
+  LARS lars1(true, lambda1, 0.0);
+  arma::vec betaOpt1;
+  maxCorr = lars1.Train(X, y, betaOpt1);
```
ShikharJ: Same here.

src/mlpack/tests/random_forest_test.cpp
```diff
@@ -395,4 +395,96 @@ BOOST_AUTO_TEST_CASE(SerializationTest)
...
+  for (size_t i = 0; i < dataset.n_cols; ++i)
+    weights[i] = math::Random(0.9, 1.0);
+  for (size_t i = dataset.n_cols; i < dataset.n_cols + 1000; ++i)
+    weights[i] = math::Random(0.0, 0.01); // Low weights for false points.
+
+  double entropy;
+
+  // Test random forest on unweighted numeric dataset.
+  RandomForest<GiniGain, RandomDimensionSelect> rf;
+  entropy = rf.Train(dataset, labels, 3, 10, 5);
```
ShikharJ: Same here.

src/mlpack/tests/random_forest_test.cpp
```diff
@@ -395,4 +395,96 @@ BOOST_AUTO_TEST_CASE(SerializationTest)
...
+    weights[i] = math::Random(0.0, 0.001);
+
+  arma::mat fullData = arma::join_rows(d, randomNoise);
+  arma::Row<size_t> fullLabels = arma::join_rows(l, randomLabels);
+
+  double entropy;
+
+  // Test random forest on unweighted categorical dataset.
+  RandomForest<> rf;
+  entropy = rf.Train(fullData, di, fullLabels, 5, 15 /* 15 trees */, 5);
```
ShikharJ: And here.

walragatver:
@ShikharJ, I have fixed the issues.

walragatver:
@mlpack-jenkins test this please.

ShikharJ:
Thanks for the contribution @walragatver

PR #1679, opened 2019-01-19
Add LogProbability for Discrete multi-input case.

mirraaj:
#1660 Added LogProbability function for Discrete Distribution.

PR #1726, opened 2019-02-20
Extend1679

mirraaj:
Returns the log probability of one observation (x) for one of the Gamma's dimensions.

### DIFF
--- a/src/mlpack/core/dists/gamma_distribution.hpp
+++ b/src/mlpack/core/dists/gamma_distribution.hpp
@@ -1,6 +1,7 @@
 /**
  * @file gamma_distribution.hpp
  * @author Yannis Mentekidis
+ * @author Rohan Raj
  *
  * Implementation of a Gamma distribution of multidimensional data that fits
  * gamma parameters (alpha, beta) to data.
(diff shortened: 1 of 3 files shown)


mirraaj:
@rcurtin @zoq This is an extension of #1679 . Do we need a test for this as well?

rcurtin:
If you added new functionality, then yes.

mirraaj:
@rcurtin If we merge #1679, this PR becomes just an extension of introducing function in Gamma Distribution. Let me know if you require justification/rectification if needed. Thanks

rcurtin:
Right, and if you added new functions to `GammaDistribution` then they should also be tested. :+1:

mirraaj:
@ShikharJ, @rcurtin, @zoq This is a small PR to extend #1679. I have also added the test for this addition. Please let me know your views about this. Thanks

src/mlpack/core/dists/gamma_distribution.hpp
```diff
@@ -171,6 +172,14 @@ class GammaDistribution
      */
     void LogProbability(const arma::mat& observations,
                         arma::vec& LogProbabilities) const;
+    /*
```
ShikharJ: Please include a newline here. Also another `*` I think.

mirraaj: @ShikharJ I have included the changes.

mirraaj:
@rcurtin @zoq Any comments?

rcurtin:
The only distributions left for this support now are `LaplaceDistribution` and `RegressionDistribution`.  In both cases the changes should be really easy, just like this PR.

mirraaj:
Yeah, I am working on it. I will push a PR soon. Thanks

PR #1728, opened 2019-02-21
Doc fixes

dtimson:
A small change to return the optimization objective function value (loss) from FFN::Train. Useful so progress can be tracked dynamically, for example plotting.

Also, a minor doc correction.


zoq:
Do you like to add the option to the allowed option list in https://github.com/mlpack/mlpack/blob/master/README.md as well?

zoq:
Also, maybe we should rename the PR to avoid any confusion?

dtimson:
OK. I've also filled out the list in build.hpp, and referred to it in the readme.

dtimson:
Also, the values returned from Train() look good to me, thanks. Good job walragatver!

ShikharJ:
@dtimson Thanks for the contribution!

PR #1730, opened 2019-02-21
Fix typo in Bernoulli distribution

robotcator:
Fix typo in Bernoulli distribution


robotcator:
I found that there two folder contains distribution. One is under `/mlpack/src/mlpack/core/dists`, the other is under `/mlpack/src/mlpack/methods/ann/dists`. So can anyone tell the difference of those two folder?

robotcator:
> Thanks! Always good to fix typos. :) The reason that this exists is that the ANN module (or parts of it) use distributions with a slightly different set of methods; so this `BernoulliDistribution` does not have the same API as everything in `src/mlpack/core/dists/`. It might be nice to add the missing functions to `BernoulliDistribution` and move it (and updated the ANN code that uses it), but at the time this was implemented we chose not to do that.

Thank for your clarification, I got it. I will look into deeper because RL methods have a strong relationship with this part.

Issue #1709, opened 2019-02-06
Make models more accessible in Python


rcurtin:
This issue has to do with mlpack's Python bindings.  Ideally, if you're interested in solving this, you should be familiar with how these bindings are used from Python, and have some knowledge of the automatic bindings system that generates these Python bindings.  If you don't but you're still interested in this, consider looking at the [Python quickstart](http://mlpack.org/docs/mlpack-git/doxygen/python_quickstart.html) and the [automatic bindings documentation](http://mlpack.org/docs/mlpack-git/doxygen/bindings.html) first.  Also, this is an open-ended issue---I don't know the right way to solve it, and some exploration of possibilities would be needed.

One of the major drawbacks of the Python bindings is that when we build a machine learning model, it's not really inspectable by a Python user.  In fact if I run the following code:

```
>>> import numpy as np
>>> from mlpack import logistic_regression
>>> x = np.random.rand(10, 10)
>>> y = [0 0 0 0 0 1 1 1 1 1]
>>> output = logistic_regression(training=x, labels=y, verbose=True)
```

then the output logistic regression model (accessed with `output['output_model']`) is not an easy to understand model but instead a Python object of type `LogisticRegressionModel` which internally just holds a pointer to the C++ memory that represents the `mlpack::regression::LogisticRegression` class.  A Python user can't do anything with this except pass it to another call to `logistic_regression()` or pickle it for later use:

```
>>> import pickle
>>> pickle.dumps(output['output_model'])
```

If you take a look at the generated `logistic_regression.pyx` file in the `src/mlpack/bindings/python/mlpack/` directory under your build directory, you can see that the model classes in Python (like `LogisticRegressionModel`) provide a `__getstate__()` and `__setstate__()` method for pickling, and these in turn call `SerializeIn()` and `SerializeOut()`.  Those functions both use boost::serialization to output the C++ model in a binary format (perfect for pickling).

But as you can see that's not very useful for people who might want to know what the parameters of their model is.  (And, the whole impetus for this issue was #1703, where I wanted to do exactly that.)  Now, the logistic regression model is actually really simple---it just holds a vector of parameters.  So it would be really nice if I could do something like this:

```
>>> output = logistic_regression(training=x, labels=y, verbose=True)
>>> model = transform_to_python(output['output_model'])
>>> model
{'parameters': [0.5, 0.6, 0.1, 0.4]}
```

[...]

PR #1734, opened 2019-02-23
 Make models more accessible in Python #1709 

MuLx10:
Make models more accessible in Python #1709 . Added SerializeParamOut with boost xml_archive. The xml is converted to json and passed via Serialize. The json is loaded in python to recover the parameters of the model.

```python
>>> import numpy as np
>>> from mlpack import pythonize_model 
>>> from mlpack import logistic_regression

>>> x = np.random.rand(10,10)
>>> y = [1,0,1,0,1,0,1,0,0,0]

>>> output = logistic_regression(training=x,labels=y)
>>> pythonize_model.transform(output['output_model'])

{'parameters': array([  7.14841847,  72.40319186, -22.58314159,  40.70852354,
       -27.78725555, -75.40210697,  29.39998803,  14.44999113,
       -60.09246953,  18.63193182, -17.18063739]), 'lambda': 0}
```

Also added the test for pythonize_model.py


MuLx10:
@rcurtin I added xml2json dependency in C++ because it is faster. But,  I will add the dependency in python.

MuLx10:
@rcurtin  I have created a PR #1771 . My git crashed. Hence, I am unable to access this PR.

Issue #1727, opened 2019-02-20
Windows install with vcpkg


kereslas:
I have made a [PR](https://github.com/Microsoft/vcpkg/pull/5409) to add mlpack 3.0.4 to vcpkg library manager. Right now it's open, but I hope it gets merged to master. If you have vcpkg on your system, installing mlpack should be as easy as executing:

.\vcpkg install mlpack:x64-windows

This should download, build and install mlpack and all its dependencies. After that, if you have vcpkg user-wide integration enabled, you can create a project on Visual Studio and it will manage automatically linking with the library. So, you only need to write code and compile, without worrying about include directories or libraries. If you don't have user-wide integration you can manually set up the project and use as usual.

It would be nice if some people can try the PR and check if it works fine. vcpkg was created for Windows, but now it works on Linux and Mac. I have not tested on this systems.

rcurtin:
@kereslas awesome!  Thank you for taking the time to do that.  I don't personally have Windows so I can't test it there, but maybe some other people watching this repository can do that.  Making the Windows installation process easier is definitely helpful.

I took a look at the installation scripts... do you think it would be good to install the header files (from `${CMAKE_BINARY_DIR}/include/` to somewhere also?  Maybe I overlooked something but I don't see that currently being done (maybe it's implicit?).  It may also be interesting to build the Python bindings but that would probably be hard to set up right...

kereslas:
The header files are copied to vcpkg\installed\<architecture>\include by the build process. This is the place later VS will try to found it, so you can use all the includes in any project without setting anything manually. In vcpkg\installed\<architecture> are also a lib and bin folder with libs and dlls.

Usually the people at vcpkg project recomend to only compile the libraries and not the execs of samples or tests. Because of that I have made an optional exe compilation flag "tools" that builds this assets also:

.\vcpkg install mlpack[tools]:x64-windows

If you use the build command without the flag, only the library is build. If I have some time I will check if the script can build the python bindings with this or similar flag.

kereslas:
The PR is officially merged to master branch, so anybody can install mlpack from now with vcpkg. The tests made by Microsoft include Mac OSX and seems it works on this system too.

ShikharJ:
@kereslas Amazing, thanks for the work! Do you think we should include this in the documentation as well? If you wish to do it, let me know, else I can open a PR.

rcurtin:
[...]

PR #1736, opened 2019-02-24
Improve Installation Documentation on Windows

ShikharJ:
Relevant: #1727

### DIFF
--- a/doc/guide/build_windows.hpp
+++ b/doc/guide/build_windows.hpp
@@ -1,6 +1,8 @@
 /**
  * @file build_windows.hpp
  * @author German Lancioni
+ * @author Miguel Canteras
+ * @author Shikhar Jaiswal
 
 @page build_windows Building mlpack From Source on Windows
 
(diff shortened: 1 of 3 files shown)


doc/guide/build_windows.hpp
```diff
@@ -10,11 +12,41 @@ This document discusses how to build mlpack for Windows from source, so you can
...
+now available for Windows installation through vcpkg:
+
+- Install Git (https://git-scm.com/downloads and execute setup)
+
+- Install CMake (https://cmake.org/ and execute setup)
+
+- Install vcpkg (execute in PowerShell):
+
+@code
+> git clone https://github.com/Microsoft/vcpkg.git
```
zoq: Not sure we should list the install routine for vcpkg here; does the PowerShell come with git?

rcurtin: Agreed with that---we can point readers to vcpkg's documentation for that.

ShikharJ: Okay, updated accordingly. Thanks.

Issue #1733, opened 2019-02-23
Ensmallen documentation yet to be removed


favre49:
In mlpack/doc/tutorials, there exists tutorials for CNE and Optimizers, both of which have now been shifted to the ensmallen repository. Shouldn't they be removed?

zoq:
Right, I think it makes sense to remove those files.

rcurtin:
Agreed, if you want to open a simple PR to do that I think we should merge it. :+1:

PR #1739, opened 2019-02-25
Removed ensmallen docs

favre49:
Re: issue #1733


doc/tutorials/optimizer/optimizer.txt
```diff
@@ -1,494 +0,0 @@
-/*!
-
-@file optimizer.txt
```
zoq: Can you remove the reference from https://github.com/mlpack/mlpack/blob/master/doc/tutorials/tutorials.txt?

favre49: yup, i've removed the references

zoq:
Great, thanks!

PR #1743, opened 2019-02-26
Add emfit_test.cpp for testing EMFit class

KimSangYeon-DGU:
Created `emfit_test.cpp` to verify the `EMFit` class.

### DIFF
--- a/src/mlpack/tests/emfit_test.cpp
+++ b/src/mlpack/tests/emfit_test.cpp
@@ -1,0 +1,223 @@
+/**
+ * @file emfit_test.cpp
+ * @author Kim SangYeon
+ * 
+ * Test for the EMFit class.
+ * 
+ * mlpack is free software; you may redistribute it and/or modify it under the
+ * terms of the 3-clause BSD license.  You should have received a copy of the
+ * 3-clause BSD license along with mlpack.  If not, see
+ * http://www.opensource.org/licenses/BSD-3-Clause for more information.
+ */
+#include <mlpack/core.hpp>
+
+#include <mlpack/methods/gmm/em_fit.hpp>
+
+#include <boost/test/unit_test.hpp>
+#include "test_tools.hpp"
+
+using namespace mlpack;
+using namespace mlpack::gmm;
+using namespace mlpack::kmeans;
+using namespace mlpack::distribution;
+
+BOOST_AUTO_TEST_SUITE(EMFitTest);
+
+//! Make sure the constructor of EMFit class works correctly.
+BOOST_AUTO_TEST_CASE(EMFitConstructorTest)
+{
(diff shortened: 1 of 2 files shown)


KimSangYeon-DGU:
I will increase the tolerance for covariance in the `EMFitEstimateTest`, because the bad result was detected in 12,736th loop.

KimSangYeon-DGU:
Would it be better to add this test cases in `gmm_test.cpp` without creating a new file `emfit_test.cpp`?? And some test cases of GMM and HMM seem to be failed. I'll look into them to find robust cases.

src/mlpack/tests/emfit_test.cpp
```diff
@@ -0,0 +1,223 @@
...
+  EMFit<> fitter;
+
+  // Check the default template parameters of EMFit class.
+  // The EMFit class's default clusterer and constraint for covariance are
+  // KMeans and PositiveDefiniteConstraint classes respectively.
+  if (std::is_same<decltype(fitter.Clusterer()), KMeans<>&>::value)
+    sameCluster = true;
+
+  if (std::is_same<decltype(fitter.Constraint()),
+    PositiveDefiniteConstraint&>::value)
```
rcurtin: This is a little pedantic but this should be indented twice. :+1:

src/mlpack/tests/emfit_test.cpp
```diff
@@ -0,0 +1,223 @@
...
+  // The EMFit class's default max-iteration is 300, and tolerance is 1e-10.
+  BOOST_CHECK_EQUAL(fitter.MaxIterations(), 300);
+  BOOST_CHECK_CLOSE(fitter.Tolerance(), 1e-10, 1e-3);
+  BOOST_CHECK_EQUAL(sameCluster, true);
+  BOOST_CHECK_EQUAL(sameConstraint, true);
+}
+
+/**
+ * Make sure EMFit can estimate parameters reasonably given observations
+ * from the Gaussian distributions.
```
rcurtin: Would you say that this is already tested via the `GMMTrainEMMultipleGaussians` test?

KimSangYeon-DGU: That makes sense. I thought it would be good to make some unit tests exclusively for `EMFit` class.

KimSangYeon-DGU:
> But I'm not sure if the second test is kind of a duplicate of an existing test in `gmm_test.cpp`, and it looks like the first test is just checking that the default template types of `EMFit<>` are `KMeans<>` and `PositiveDefiniteContraint`?

Hi @rcurtin. That makes sense, I would like to make some unit test exclusively for `EMFit` class. but it seems to be unnecessary.

Issue #1699, opened 2019-01-31
Implementation of Siamese Networks and related Loss function


shardulparab97:
@rcurtin  @zoq I am really interested in contributing to MLPack. I feel that MLPack can grow further in the computer vision domain too.

However being new to the codebase, I wanted to begin with the implementation of the following:
- [ ] Contrastive Loss
- [ ] A basic Siamese Network module 
- [ ] Triplet Loss

Please let me know your thoughts about the same and any other ideas in computer vision/image domain which can be worked upon. Thanks a lot! :smile:

zoq:
Hello, glad you find the project interesting. Contributions are always welcome, and I think the computer vision sector could be a great fit for mlpack.

Personally I like the idea to start with a loss function to get familiar with the codebase, so as you suggested Triplet /Contrastive Loss might be a good candidate. About a basic Siamese Network module, not sure there is need to write an extra module/layer do you think it would be possible to implement the module with the existing codebase and to implement an example similar to what was done here: https://github.com/mlpack/models?

shardulparab97:
Thank you @zoq. I will begin with the loss functions and then go towards implementing the Siamese Network as per your suggestion. :smiley:

zoq:
Sounds great, let me know if I should clarify anything.

shardulparab97:
@zoq I have opened Pull Request for the implementation here: #1704  
Please let me know what you think :)

zoq:
Great, will take a look at the PR in the next days.

Hemal-Mamtora:
Hello ! @zoq @shardulparab97
I am also interested in developing the Siamese network. May I assist you in developing it ?
I would like to start with the triplet loss function 
Please let me know. Thanks.

zoq:
I'll let @shardulparab97 answer this one.

shardulparab97:
Hi @Hemal-Mamtora  sorry for the late reply, was caught up with exams.
Yes please go ahead with the triplet loss function, it will only make the process faster! :smiley:  
Let me know if you need any help.
At the same time, we could work on the implementation of the Siamese Model.

Hemal-Mamtora:
Hi @shardulparab97 , I'll start right away on the triplet loss function.
Meanwhile if you have any reference material like technical papers about the topic(Siamese, triplet) please forward the link
This will help us stay on the same page and we could implement it faster.

johnsoncarl:
[...]

PR #1744, opened 2019-02-26
Triplet Loss Function for Siamese Network (issue #1699)

Hemal-Mamtora:
@zoq @shardulparab97 Please have a look at the implementation code of the Triplet Loss Function, Please revert if modification is required.
Thank you


zoq:
Thanks for the contribution, before I'll take a closer look, can you write a test for the loss function, see https://github.com/mlpack/mlpack/blob/master/src/mlpack/tests/loss_functions_test.cpp for some examples.

zoq:

Hemal-Mamtora:
@zoq I had some compile-time errors when I wrote the tests. I will start working again on them from the next week.

zoq:

src/mlpack/methods/ann/loss_functions/triplet_loss_impl.hpp
```diff
@@ -0,0 +1,75 @@
...
+    const InputType&& negative)
+{
+  arma::mat positive_distance = 
+      metric::SquaredEuclideanDistance::Evaluate(anchor, positive);
+  
+  arma::mat negative_distance = 
+       metric::SquaredEuclideanDistance::Evaluate(anchor, negative);
+
+  double triplet_loss = 
+      std::max(positive_distance - negative_distance + margin, 0);
```
zoq: `std::max` doesn't work on `area::mat` (`positive_distance, negative_distance`)`, not sure if that is right, but you could take sum (`area::accu`) over `positive_distance - negative_distance + margin` to solve the issue.

johnsoncarl:
Hey @zoq @hemam-mamtora , I found this too, while I was trying to write loss function test for this. 

I've reported this issue in the IRC channel yesterday, in the main channel and to you in DM.

Though I found it was not a compile time error but was encountered while running an example. I tried using colvec instead of mat, that thing didn't work too. 

I didn't give it much time as the build process eats so much time, despite turning of few not required builts. 

Would you like me to have a look at it too. Or I shall leave it to @hemam-mamtora ?

src/mlpack/methods/ann/loss_functions/triplet_loss_impl.hpp
johnsoncarl: Yeah, I realised this too, that accumulation function would help here. 

I will be doing this today later.

Hemal-Mamtora:
There was an issue on my local machine, so i pulled from mlpack/mlpack and then merged my branch with it. Apparently this has caused all other pull request to be included in my pull request, so I have closed this pull request, I will open a new one, with the required changes.

PR #1745, opened 2019-02-27
Update 1-documentation.md

Anupam-tripathi:
The grammar and punctuation of the documentation are corrected in this pull request.  This delivers a better experience to the reader.

### DIFF
--- a/.github/ISSUE_TEMPLATE/1-documentation.md
+++ b/.github/ISSUE_TEMPLATE/1-documentation.md
@@ -9,16 +9,16 @@ assignees: ''
 
 <!--
 
-Welcome!  Unfortunately not all documentation is perfect, and if you're opening
+Welcome!Unfortunately, not all documentation is perfect, and if you're opening
 a documentation issue we are interested in fixing it.  Please fill out the
-template below so that we can solve the problem more quickly; or, alternately,
+template below so that we can solve the problem more quickly; or, alternatively,
 open a PR with a fix, if you like.
 
 -->
 
 #### Problem location
 
-<!-- Link to incorrect website or location of source file with bad
+<!-- Link to incorrect website or location of the source file with bad
 documentation. -->
 
 #### Description of problem


.github/ISSUE_TEMPLATE/1-documentation.md
```diff
@@ -9,16 +9,16 @@ assignees: ''
 
 <!--
 
-Welcome!  Unfortunately not all documentation is perfect, and if you're opening
+Welcome!Unfortunately, not all documentation is perfect, and if you're opening
```
rcurtin: There's not a need for the comma after `Unfortunately`, and I'm not sure why you've removed the space between `Welcome!` and `Unfortunately`.

.github/ISSUE_TEMPLATE/1-documentation.md
```diff
@@ -9,16 +9,16 @@ assignees: ''
...
 a documentation issue we are interested in fixing it.  Please fill out the
-template below so that we can solve the problem more quickly; or, alternately,
+template below so that we can solve the problem more quickly; or, alternatively,
```
rcurtin: Either alternately or alternatively is valid here but I am indifferent, so changing to `alternatively` is fine with me.

.github/ISSUE_TEMPLATE/1-documentation.md
```diff
@@ -9,16 +9,16 @@ assignees: ''
...
 open a PR with a fix, if you like.
 
 -->
 
 #### Problem location
 
-<!-- Link to incorrect website or location of source file with bad
+<!-- Link to incorrect website or location of the source file with bad
```
rcurtin: Actually technically this should be "the source file(s)" if we want to be really pedantic about it.

ShikharJ: I think this is the only viable change, the rest should be reverted back @Anupam-tripathi.

jeffin143:
@Anupam-tripathi , It would be good if you could just look into the comments and resolve them.

rcurtin:
I'm not sure why mlpack-bot didn't close this for inactivity, but that's what I'll do here.  We can reopen it if needed.

Issue #1269, opened 2018-02-23
Nearest neighbors with spill trees in dualtree mode does not always return neighbors


vasiloglou:
I have noticed that if the k is big the dual tree algorithm will return max float distances and invalid indices. This probably happens because some leaf nodes do not have enough points

akhandait:
Yes, you are right, I was righting tests for nearest neighbors bindings in #1263 when I noticed this.
As discussed on that pull request, I am working on the solution. :)

rcurtin:
For anyone looking to contribute, this is a semi-challenging issue that requires some knowledge of how nearest neighbor works with spill trees.  You can reproduce @vasiloglou's original issue by running the following command once you've built mlpack:

```
mlpack_knn -i test_data_3_1000.csv --leaf_size 5 -t spill -k 10 -n n.csv -d d.csv -v -a single_tree
```

This runs k-nearest-neighbor search in single-tree mode using spill trees with a leaf size of 5.  (So, every leaf in the spill tree contains 5 or fewer points.)  But you can see in `n.csv` and `d.csv` (the result files), that many of the indices of nearest neighbors are invalid!  The reason for this is that the single-tree traverser for spill trees (in `src/mlpack/core/tree/spill_tree/spill_single_tree_traverser.hpp`), which will be run with `Defeatist = true`, never actually encounters enough points to add to the results!  This is because the algorithm will traverse to the nearest leaf, then run the base cases to find the nearest neighbor.  But since `leaf_size` is only 5, and we are looking for 10 neighbors, we are guaranteed that will find at most 5 nearest neighbors!

The task here, then, is to modify the code such that enough results are always returned for the defeatist spill tree single-tree traverser, and the defeatist spill tree dual-tree traverser---the problem also happens when using `-a dual_tree`.

UTpH:
Hello!
I'm looking to contribute to this project. Could I work on this if it is not already taken?

zoq:
Hello @UTpH , no one is working on this one right now, so please feel free to pick this up.

shashank-007:
@rcurtin 
```
mlpack_knn -i test_data_3_1000.csv --leaf_size 5 -t spill -k 10 -n n.csv -d d.csv -v -a single_tree
```
this command throws error 
```
[FATAL] Unable to detect type of 'test_data_3_1000.csv'; incorrect extension?
```

rcurtin:
@shashank-007: does the file exist?

shashank-007:
[...]

PR #1750, opened 2019-03-01
Fix Spill tree Throwing garbage value

Yashwants19:
Address #1269
Fix Spill Tree as it throwing some garbage value when `Defeatist = true`


Yashwants19:
As given in Spill Tree paper when `tau = 0` it must be considered as Metric Tree. And in Hybrid SP Tree when `rho*(node*NumChildren())` is more than either of node's children than we must go for overlapping nodes else must go for do overlapping nodes.

Yashwants19:
Hi @rcurtin as given in paper [http://www.cs.cmu.edu/~agray/approxnn.pdf](http://www.cs.cmu.edu/~agray/approxnn.pdf) in hybrid spill tree(3.2) if either of node's children contains more than ρ fraction of the total points in node(rho * all points at node), we undo the overlapping splitting. Instead, a conventional metric-tree partition (without overlapping) is used. I think so we do not apply this in our code.

PR #1752, opened 2019-03-03
Created default value for ConstInitialization constructor

RMaron:
Added a default value to the ConstInitialization constructor that allows it to be used as a template argument.
Initializes the weight matrix to 0.

### DIFF
--- a/src/mlpack/methods/ann/init_rules/const_init.hpp
+++ b/src/mlpack/methods/ann/init_rules/const_init.hpp
@@ -25,6 +25,14 @@ namespace ann /** Artificial Neural Network. */ {
 class ConstInitialization
 {
  public:
+
+  /**
+   *  Create the ConstantInitialization object.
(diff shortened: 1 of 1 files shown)


src/mlpack/methods/ann/init_rules/const_init.hpp
```diff
@@ -25,6 +25,14 @@ namespace ann /** Artificial Neural Network. */ {
 class ConstInitialization
 {
  public:
+
+  /**
+   *  Create the ConstantInitialization object.
+   *  Allow it to be used as a template argument.
+   */
+  ConstInitialization() : initVal(0.0) {};
```
zoq: What about we set a default value for the existing constructor:

``ConstInitialization(const double initVal = 0) : initVal(initVal)``

RMaron: Sounds good.

PR #1758, opened 2019-03-05
fix typo in negative_log_likelihood

robotcator:
As mentioned, fix typo in negative_log_likelihood. 

The consistency of interface, I found that there are two interfaces in negative_log_likelihood.hpp
```c++
  //! Get the input parameter.
  InputDataType& InputParameter() const { return inputParameter; }
  //! Modify the input parameter.
  InputDataType& InputParameter() { return inputParameter; }
```
But in some other loss function, they have not. Shall we just keep it or remove?  Literally, it didn't affect the functionality.

### DIFF
--- a/src/mlpack/methods/ann/loss_functions/negative_log_likelihood.hpp
+++ b/src/mlpack/methods/ann/loss_functions/negative_log_likelihood.hpp
@@ -40,7 +40,7 @@ class NegativeLogLikelihood
    */
   NegativeLogLikelihood();
 
-  /*
+  /**
    * Computes the Negative log likelihood.
    *
    * @param input Input data used for evaluating the specified function.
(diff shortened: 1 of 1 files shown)


zoq:
For consistency I would add the functions to the other loss functions as well.

rcurtin:
Oh, actually somehow I only looked at the code here and didn't answer the question.  Sorry about that @robotcator.

robotcator:
> For consistency I would add the functions to the other loss functions as well.

Okay, I get it.

robotcator:
> Oh, actually somehow I only looked at the code here and didn't answer the question. Sorry about that @robotcator.

Oh, no worries. There are many issues and PRs you need to answer and review. It is easy to miss some of the details. I just review this part of the code for implementing a self-defined loss function. Also, zoq answered this question. : )

PR #1759, opened 2019-03-05
Implemented Isomap

riaash04:
Implemented the Isomap algorithm which is part of Manifold Learning algorithms used for dimensionality reduction. This implemented can be extended to include variations in algorithms for constructing neighbourhood graphs and calculating all pair shortest distance in graphs.


.gitignore
```diff
@@ -6,3 +6,4 @@ src/mlpack/core/util/arma_config.hpp
 .idea
 cmake-build-*
 *.pyc
+.vscode*
```
mirraaj: You need a newline character here.

riaash04: Added thanks :)

src/mlpack/methods/isomap/isomap_main.cpp
```diff
@@ -0,0 +1,77 @@
...
+    PRINT_PARAM_STRING("neighbours") + " parameter to specify the number "
+    "of nearest neighbours to consider for constructing the neighbourhood "
+    "graph (make sure it is enough so that a connected graph is formed).\n "
+);
+
+// Parameters for the program
+PARAM_MATRIX_IN_REQ("input", "Input dataset to perform Isomap on.", "i");
+PARAM_MATRIX_OUT("output", "Matrix to save modified dataset to.", "o");
+PARAM_INT_IN("neighbours", "The number of neighbours to consider for "
+    "k-nearest.", "n", 3);
```
rcurtin: I'd say that the short-parameter `k` would be more appropriate here, just based on how things are used in other programs.  (Like, e.g., `knn` takes `k` as a parameter.)

Also, this is a minor point, but can you use the spelling `neighbors` instead?  That's what the rest of the library does.  I don't have a huge opinion either way, I just want to be consistent. :)

src/mlpack/methods/isomap/isomap_main.cpp
```diff
@@ -0,0 +1,77 @@
...
+
+static void mlpackMain()
+{
+  // Load input dataset.
+  arma::mat& input = CLI::GetParam<arma::mat>("input");
```
rcurtin: Consider doing `std::move()` here, otherwise the matrix is likely to be copied unnecessarily.

src/mlpack/methods/isomap/isomap_main.cpp
```diff
@@ -0,0 +1,77 @@
...
+
+  // The number of neighbours to consider for k-nearest
+  RequireParamValue<int>("neighbours", [](int x) { return x > 0; },
+      true, "number of neighbours must be greater than zero");
+  std::ostringstream error;
+  error << "cannot be greater than or equal to number of data points ("
+      << input.n_cols << ")";
+  RequireParamValue<int>("neighbours",
+      [input](int x) { return x < (int) input.n_cols; }, true,
+      error.str());
```
rcurtin: Nice, thanks for checking the parameters like this. :+1:

src/mlpack/methods/isomap/isomap_main.cpp
```diff
@@ -0,0 +1,77 @@
...
+
+  size_t n_neighbours = CLI::GetParam<int>("neighbours");
```
rcurtin: Another little style issue---can you avoid using variables with underscores in their name?  I'd suggest that `k` would be a fine name to use here, but `neighbors` or `numNeighbors` or whatever is fine too.

src/mlpack/methods/isomap/k_nearest.hpp
```diff
@@ -0,0 +1,115 @@
...
+
+#include <mlpack/prereqs.hpp>
+#include <queue>
+
+namespace mlpack {
+namespace isomap {
+/**
+ * This class implements the K-Nearest Neighbours function to construct
+ * a neighbourhood graph for the given distance matrix.
+*/
```
rcurtin: This comment is spaced strangely and away from the class; can you fix that please?

src/mlpack/methods/isomap/k_nearest.hpp
```diff
@@ -0,0 +1,115 @@
...
+{
+ public:
+  /**
+   * Function to make the neighbourhood graph for the given distance matrix
+   * and number of neighbours.
+   * 
+   * @param disMat -distance matrix to construct the graph
+   * @param n_neighbours -number of neighbours to consider
+   */
+  void MakeNeighbourhoodGraph(arma::mat &disMat,
```
rcurtin: Spacing issue---this should be `arma::mat& disMat`.

src/mlpack/methods/isomap/k_nearest.hpp
```diff
@@ -0,0 +1,115 @@
...
+                              const int n_neighbours)
+  {
+    // go through every row
+    for (size_t i = 0; i < disMat.n_rows; i++)
```
rcurtin: Armadillo stores matrices in column-major format, so it can be orders of magnitude faster to iterate over columns instead of iterating over rows.  To be more specific, this means that in memory, points in a single column are stored contiguously.

So, whereever you are iterating over matrices, I would advise you to do it column-by-column instead, even if this means storing the matrix in what seems like a transposed form.  It can have major speed benefits to do that.

src/mlpack/methods/isomap/k_nearest.hpp
```diff
@@ -0,0 +1,115 @@
...
+      // set distance to inf for every neighbour farther that n_neighbours
+      for (size_t j = 0; j < temp.size()-n_neighbours-1; j++)
+        disMat(i, temp(j)) = DBL_MAX;
+    }
+
+    // check if neighbourhood graph is connected
+    if (!IsConnected(disMat))
+      Log::Fatal << "Constructed neighbourhood graph is not connected. "
+                  << "Increase the number of neighbours (Default is 3).\n";
+  }
```
rcurtin: Actually this method can be greatly simplified or removed entirely if we use mlpack's nearest neighbor search infrastructure, since the `Search()` method will return only the k nearest neighbors instead of everything, and they will be in sorted order.

src/mlpack/methods/isomap/k_nearest.hpp
```diff
@@ -0,0 +1,115 @@
...
+      }
+    }
+
+    // checking if a node is not visited
+    for (size_t i = 0; i < tempMat.n_rows; i++)
+    {
+      if (!visited(i))
+      {
+        flag = 0; // means disconnected graph
+        break;
```
rcurtin: You could avoid `flag` entirely and just use `return false;` here, and then `return true` later.

src/mlpack/methods/isomap/mds.hpp
```diff
@@ -0,0 +1,121 @@
+/**
+ * @file mds.hpp
```
rcurtin: Should we put this in its own `mds/` directory and create a binding for that also?

src/mlpack/methods/isomap/isomap.hpp
```diff
@@ -0,0 +1,129 @@
...
+    arma::mat disMat(input.n_cols, input.n_cols);
+
+    // calculating distance matrix from the given input matrix
+    CalcDistanceMatrix(input, disMat);
+
+    // constructing neighbourhood graph (K_nearest is used)
+    neighbourhood.MakeNeighbourhoodGraph(disMat, n_neighbours);
+
+    // finding all pair shortest path in the neighbourhood graph created
+    shortestPath.FindShortestPath(disMat);
```
rcurtin: The goal of the entire algorithm up to this point is to construct a matrix that contains the shortest geodesic path between each pair of points.  So Dijkstra's algorithm is used here, but what you could actually equivalently use is just a minimum spanning tree.  We already have code that can compute that (as part of the nearest neighbor search) in the `DualTreeBoruvka` class.  So I would suggest that perhaps we should use that instead of the hand-implemented Dijkstra's algorithm here.

src/mlpack/methods/isomap/isomap.hpp
```diff
@@ -0,0 +1,129 @@
...
+
+    // making the shortest distance matrix symmetric (required for cMDS)
+    disMat = arma::min(disMat, disMat.t());
```
rcurtin: This particular way of doing this is likely to instantiate the transpose into a temporary matrix and then copy everything.  Probably the better thing to do is use `symmatu()` or `symmatl()`.

src/mlpack/methods/isomap/isomap.hpp
```diff
@@ -0,0 +1,129 @@
...
+
+    // mds object to perform classical multidimensional scaling
+    MDS md;
+
+    // performing classical multidimensional scaling
+    md.Apply(disMat);
```
rcurtin: The API of overwriting the distance matrix (which here is not the distance matrix but the geodesic distance matrix or dissimilarity matrix, right?) is a little bit weird.  Why not provide a different matrix for the output, like, e.g., `md.Apply(disMat, output)`?

src/mlpack/methods/isomap/isomap.hpp
```diff
@@ -0,0 +1,129 @@
...
+    input = disMat.t();
```
rcurtin: You're using `input` as an `output` here, so if you'd like to do that, then we should perhaps change the name...

src/mlpack/methods/isomap/mds.hpp
```diff
@@ -0,0 +1,121 @@
...
+  {
+    // check to see if the incoming matrix is valid for cMDS
+    IsValidMatrix(disMat);
+
+    size_t n = disMat.n_rows;
+    arma::mat temp(n, n);
+
+    // Spectral decomposition
+
+    arma::mat H = arma::eye<arma::mat>(n, n) - temp.fill((double) (1.0/n));
```
rcurtin: I don't think you need `temp`; you should be able to just subtract by `1.0 / n`.

src/mlpack/methods/isomap/mds.hpp
```diff
@@ -0,0 +1,121 @@
...
+    arma::mat B = H * (-0.5 * (disMat % disMat) ) * H;
+    arma::mat eigen_vecs;
+    arma::vec eigen_vals;
+
+    // guard eigen values against round-off errors
+    arma::eig_sym(eigen_vals, eigen_vecs, (B + B.t())/2);
```
rcurtin: This is a clever little trick for stability, but honestly I think that it is a bit costly because the size of `B` may be quite large.  I also suspect that it should not be a huge issue, so I don't think we need to do it.

src/mlpack/methods/isomap/mds.hpp
```diff
@@ -0,0 +1,121 @@
...
+
+    arma::uvec index = arma::sort_index(eigen_vals, "descend");
+    eigen_vals = arma::sort(eigen_vals, "descend");
```
rcurtin: From the Armadillo documentation for `eig_sym()`:

```
The eigenvalues are in ascending order
```

so this step is unnecessary.

src/mlpack/methods/isomap/mds.hpp
```diff
@@ -0,0 +1,121 @@
...
+    if (keep.is_empty())
+    {
+      // if none of the eigen values are positive, vector of zeros is returned
+      disMat = arma::zeros<arma::vec>(n);
+    }
+    else
+    {
+      index = index(keep);
+      disMat =
+      eigen_vecs.cols(index) * arma::diagmat(arma::pow(eigen_vals(keep), 0.5));
```
rcurtin: The wrapping on this line is very weird; can you put it on one line?  If you have to wrap it, the new line should be indented twice.

src/mlpack/methods/isomap/mds.hpp
```diff
@@ -0,0 +1,121 @@
...
+   */
+  void IsValidMatrix(arma::mat& disMat)
+  {
+    bool flag = 1;
+    double minDiff = 10*DBL_EPSILON;
+    // checking if matrix is a square matrix
+    flag = disMat.is_square();
+    if (flag)
+    {
+      // Checking for symmetric. could not find is_symmetric in armadillo
```
rcurtin: http://arma.sourceforge.net/docs.html#is_symmetric

src/mlpack/methods/isomap/mds.hpp
```diff
@@ -0,0 +1,121 @@
...
+  /**
+   * Fucntion to implement classical Multidimensional Scaling on a given
+   * dissimilarity matrix.
+   * 
+   * @param disMat -dissimilarity matrix (distance matrix)
+   */
+  void Apply(arma::mat& disMat)
+  {
+    // check to see if the incoming matrix is valid for cMDS
+    IsValidMatrix(disMat);
```
rcurtin: This check does take some amount of time to perform.  I would be okay if we omitted it and instead just required it in the documentation.  We are already sure that `Isomap` will be passing us a square and symmetric dissimilarity matrix.

src/mlpack/tests/main_tests/isomap_test.cpp
```diff
@@ -0,0 +1,78 @@
...
+#include <mlpack/core.hpp>
+#include <mlpack/core/util/mlpack_main.hpp>
+#include "test_helper.hpp"
+#include <mlpack/methods/isomap/isomap_main.cpp>
+
+#include <boost/test/unit_test.hpp>
+
+using namespace mlpack;
+
+struct IsoampFixture
```
rcurtin: Misspelling--- `IsoampFixture` -> `IsomapFixture`. :)

src/mlpack/methods/isomap/isomap.hpp
riaash04: As far as I know MST is a good approximation for all pair shortest paths in cases of sparse graphs as then it includes most of the useful edges. But I am not sure how accurate it would be in case of dense graphs (the type of graph would depend on the number of neighbours). If you think the approximation is reasonable we can use MST for this. Also, the complexity would reduce a little with MST.

src/mlpack/methods/isomap/mds.hpp
riaash04: Yes, I wasn't sure about that earlier. Next commit would resolve this.

src/mlpack/methods/isomap/mds.hpp
riaash04: Okay. I verified this on multiple data and you are right. Would remove the operation in next commit.

src/mlpack/methods/isomap/mds.hpp
riaash04: Oh yes. I missed this line in the documentation. The sorting is unnecessary. Would remove it in next commit.

src/mlpack/methods/isomap/mds.hpp
riaash04: Can't put it in one line, it oversteps the 80 char limit. Would be resolved in the next commit.

src/mlpack/methods/isomap/mds.hpp
riaash04: For some reason is_symmetric does is not present in my armadillo, the build fails saying is_symmetric is not part of mat. I am using Armadillo 8.400.0. Not sure what's wrong.

src/mlpack/methods/isomap/mds.hpp
riaash04: I put this check thinking MDS could be used separately. I think it's better to have a separate mds/ for this file as you suggested. I will do that.

src/mlpack/tests/main_tests/isomap_test.cpp
riaash04: Next commit would resolve this.

src/mlpack/methods/isomap/dijkstra.hpp
```diff
@@ -0,0 +1,97 @@
...
+ * This class implements the Dijkstra Algorithm for all pair shortest path.
+ */
+class Dijkstra
+{
+ public:
+  /**
+   * Function to supply new source nodes to Dijkstra's Algorithm, so that all
+   * pair shortest path is calculated.
+   * 
+   * @param disMat -neighbourhood graph from input dataset
```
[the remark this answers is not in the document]
riaash04: Next commit would resolve this.

src/mlpack/methods/isomap/dijkstra.hpp
```diff
@@ -0,0 +1,97 @@
...
+  */
+  void FindShortestPath(arma::mat& disMat)
+  {
+    for (size_t i = 0; i < disMat.n_rows; i++)
+    {
+      // to store the calculated shortest distances for a source
```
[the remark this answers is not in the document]
riaash04: Will resolve this in the next commit.

src/mlpack/methods/isomap/isomap.hpp
```diff
@@ -0,0 +1,129 @@
...
+ * Lawrence Cayton, 2005, 17, Algorithms for Manifold Learning
+ */
+
+template <typename NeighbourhoodFunction = K_Nearest,
+          typename ShortestPathAlgo = Dijkstra>
+class Isomap
+{
+ public:
+  /**
+   * Create the Isomap obejct and set the n_neighbours parameter to
```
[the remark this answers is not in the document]
riaash04: Next commit will resolve this.

src/mlpack/methods/isomap/isomap.hpp
```diff
@@ -0,0 +1,129 @@
...
+namespace isomap {
+
+/**
+ * This class implements the Isomap algorithm which is part of Manifold
+ * Learning algorithms, and is used for dimensionality reduction. This
+ * algorithm converts the dataset into a lower dimensional space which
+ * it is assumed to belong from the higher dimensional space it is
+ * currently embedded into.
+ * This implementation is based on the paper:
+ * Lawrence Cayton, 2005, 17, Algorithms for Manifold Learning
```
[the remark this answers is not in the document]
riaash04: Yes, surely. Next commit will resolve this.

src/mlpack/methods/isomap/isomap_main.cpp
```diff
@@ -0,0 +1,77 @@
+/**
+ * @file isomap_main.cpp
+ * @author Rishabh Ranjan
+ * 
+ * Main executable to run Isomap
```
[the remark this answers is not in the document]
riaash04: Sure.:) Next commit would resolve this.

src/mlpack/methods/isomap/isomap_main.cpp
```diff
@@ -0,0 +1,77 @@
...
+using namespace mlpack::isomap;
+using namespace mlpack::util;
+using namespace std;
+
+PROGRAM_INFO("Isomap",
+    // Short Description
+    "An implementation of Isomap algorithm for dimensionality reduction. "
+    "Given an input dataset the algorithm fits the dataset into a lower "
+    "dimensional space, assuming the data is embedded into a "
+    "higher dimensional space but belongs to a lower dimensional space.",
```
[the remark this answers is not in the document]
riaash04: I think this is a good idea. I will try and make it better. It can also include classical multidimensional scaling like, 

"... using the k-nearest-neighbor graph for finding geodesic distances and classical multidimensional scaling for spectral decomposition."

I will try and make comments better.

src/mlpack/methods/isomap/isomap_main.cpp
```diff
@@ -0,0 +1,77 @@
...
+    // Long Description
+    "An implementation of Isomap algorithm which is part of Manifold Learning "
+    "algorithms, which are used for dimensionality reduction. "
+    "Given a dataset, which is assumed to have data points embedded in a"
+    "higher dimesional space, it gives as output the dataset in a lower "
+    "dimensional space, thus reducing dimensions of the data.\n\n"
+    "Use the " + PRINT_PARAM_STRING("input") + " parameter to specify the "
+    "dataset to perform Isomap on.\n Use the " +
```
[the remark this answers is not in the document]
riaash04: Yes, surely. Next commit will resolve this.

src/mlpack/methods/isomap/isomap_main.cpp
```diff
@@ -0,0 +1,77 @@
...
+    PRINT_PARAM_STRING("neighbours") + " parameter to specify the number "
+    "of nearest neighbours to consider for constructing the neighbourhood "
+    "graph (make sure it is enough so that a connected graph is formed).\n "
```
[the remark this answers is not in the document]
riaash04: Yes, surely. I will add more and clearer comments in the next commit.

src/mlpack/methods/isomap/isomap_main.cpp
riaash04: Yes, next commit will resolve this.

src/mlpack/methods/isomap/isomap_main.cpp
riaash04: Thanks, lambda functions sure are helpful.

src/mlpack/methods/isomap/isomap_main.cpp
riaash04: Yes, surely. Next commit will resolve this.

src/mlpack/methods/isomap/k_nearest.hpp
riaash04: Yes surely. Next commit will resolve this.

src/mlpack/methods/isomap/k_nearest.hpp
riaash04: Next commit will resolve this.

src/mlpack/methods/isomap/k_nearest.hpp
riaash04: Oh, nice. I wasn't aware of that. Next commit will resolve this.

src/mlpack/methods/isomap/k_nearest.hpp
riaash04: Yes that is better, should have done that. Next commit will resolve this.

src/mlpack/methods/isomap/mds.hpp
riaash04: Yes, I think that is better. I will do that.

src/mlpack/methods/isomap/isomap.hpp
riaash04: Also, K-Nearest would return a directed graph. For applying Boruvka's algorithm we would have to change it to undirected graph, which could increase the number of nearest neighbours present in the graph for a node. But I think that would not affect the working of Isomap as anyway we make the geodesic distance matrix symmetric before cMDS.

src/mlpack/methods/isomap/isomap.hpp
riaash04: I wanted to have the smaller of two distances in the distance matrix. '<symmatu()>' or '<symmatl()>' both copy the upper triangular or lower triangular matrix. If we use MST for finding all pair shortest distance then it won't matter.

src/mlpack/methods/isomap/isomap.hpp
riaash04: I wanted to have result stored in the same space. But I don't think that will make much difference. And this is definitely better for understanding and reading. Also, since we are making a separate mds/ for mds, including an output is better.
Next commit will resolve this.

src/mlpack/methods/isomap/isomap.hpp
riaash04: I was trying to have the input matrix space store the output, and not have a separate space for output matrix.
Would it be better if I just returned the `disMat` matrix to main file and stored it in the output file?

riaash04:
> Hey @riaash04, thanks for the time you've put into this. It would be quite nice to incorporate Isomap into mlpack. But I think that there are several things that we could address to make this faster. Adding a new algorithm to mlpack is one of the more time-consuming and difficult things to do, so there may still be a lot of work left. I see two major main things that should also be done before we incorporate this into mlpack (in addition to the comments I've left):
> 
>     * Add tests for `Isomap` and `MDS` and the other utility classes you've made in `src/mlpack/tests/`.  I like the tests you added in `src/mlpack/tests/main_tests/isomap_test.cpp`, but we also have to test the class's functionality itself.
> 
>     * Ensure that the implementation we have here is faster and scales better than other implementations.  Some simple comparisons and simulations in the PR should be enough to show that.
> 
> 
> Let me know if I can clarify any of the comments I've left. Thanks again! +1

I am having a problem adding functionality tests. The MATLAB implementation of classical multidimensional scaling returns an extra dimension (for some examples), whose values are very close to `0` with the rest of the dimension columns having same value as the output `mds.hpp` produces. I found this is happening because of some difference in eigen values calculation (very little difference of the oreder of 1e-5). 

So, I am thinking about including an extra parameter `newDimensionality` that can be provided to limit the number of dimensions the output will have. Earlier Isomap was calculating number of dimensions in output itself (all the dimensions having positive eigen values). This would make the implementation more aligned with sklearn's and the orignal paper. I followed the MATLAB's approach earlier.

rcurtin:
> So, I am thinking about including an extra parameter newDimensionality that can be provided to limit the number of dimensions the output will have. Earlier Isomap was calculating number of dimensions in output itself (all the dimensions having positive eigen values). This would make the implementation more aligned with sklearn's and the orignal paper. I followed the MATLAB's approach earlier.

That seems like a reasonable approach to me; it could be nice to allow users to force a certain dimensionality for the embedding. :+1:

src/mlpack/methods/isomap/dijkstra.hpp
```diff
@@ -0,0 +1,97 @@
...
+ * 3-clause BSD license along with mlpack.  If not, see
+ * http://www.opensource.org/licenses/BSD-3-Clause for more information.
+ */
+
+#ifndef MLPACK_METHODS_ISOMAP_DIJKSTRA
+#define MLPACK_METHODS_ISOMAP_DIJKSTRA
+
+#include <mlpack/prereqs.hpp>
+#include <queue>
+#include <utility>
```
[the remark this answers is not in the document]
riaash04: I included `<queue>` in the `<mlpack/prereqs.hpp>`.

src/mlpack/methods/isomap/isomap.hpp
```diff
@@ -0,0 +1,129 @@
...
+   * matrix is required and it is safe to pass it as reference.
+   * 
+   * @param input -the input dataset to perform Isomap on.
+   */
+  void Apply(arma::mat& input)
+  {
+    arma::mat disMat(input.n_cols, input.n_cols);
+
+    // calculating distance matrix from the given input matrix
+    CalcDistanceMatrix(input, disMat);
```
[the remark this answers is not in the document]
riaash04: Ok got it. I am now using `NeighborSearch` class to find the k-nearest-neighbor graph.

src/mlpack/methods/isomap/isomap_main.cpp
riaash04: Yes, done.

src/mlpack/methods/isomap/isomap.hpp
riaash04: I am using `symmatu()` now, on finding that taking minimum wasn't making much difference.

src/mlpack/methods/isomap/mds.hpp
riaash04: I removed the this check and included in the `PROGRAM_INFO` that if the input to cMDS is dissimilarity matrix, then it should be symmetric, since this would save some time.

riaash04:
@rcurtin 
Sorry, it's took such a long time to make this commit. I was unable to work on this for quite a long time. But now I will try to be more consistent.

Added a new /cmds for Classical Multidimensional Scaling in methods as suggested.
Also, included the extra parameter in cMDS to specify the number of dimensions for output. This also makes isomap to have this parameter, as isomap uses cmds as the last step.
I will add tests for functionality and to measure performance against other implementations in a few days.
 
I have doubts regarding some of the things:
1. You said for finding shortest distances we could use minimum spanning trees, I have posted my doubt in your comment [https://github.com/mlpack/mlpack/pull/1759#discussion_r266254296](https://github.com/mlpack/mlpack/pull/1759#discussion_r266254296). Basically I want to know if it's ok to have the increased time complexity of finding all pair shortest paths using MST than the Dijkstra's.
2. In one of your comment you said we could remove the `IsConnected` function [https://github.com/mlpack/mlpack/pull/1759#discussion_r266216744](https://github.com/mlpack/mlpack/pull/1759#discussion_r266216744). Actually, this function is to check whether the k-nearest-neighbor graph formed is connected or not (weakly connected since it would be a directed graph). So even if we use the `neighbor_search` method for constructing the graph (which now we are), we would still need to perform this check.

PR #1762, opened 2019-03-06
minor fix up

robotcator:
minor fix up the comment of loss function in the `loss_functions` folder.


robotcator:
> Hmm, not sure it is a fixup, but the consistency eases my OCD.

Exactly, it is the OCD. Lol

PR #1765, opened 2019-03-08
Add fastmks_test for command-line and python bindings

Yashwants19:
Address #1152

### DIFF
--- a/src/mlpack/core/metrics/ip_metric_impl.hpp
+++ b/src/mlpack/core/metrics/ip_metric_impl.hpp
@@ -48,6 +48,29 @@ IPMetric<KernelType>::~IPMetric()
 }
 
 template<typename KernelType>
+IPMetric<KernelType>::IPMetric(const IPMetric& other) :
+  kernel(other.kernel),
+  kernelOwner(other.kernelOwner)
+{
+  // Nothing to do.
+}
+
+template<typename KernelType>
+IPMetric<KernelType>& IPMetric<KernelType>::operator=(const IPMetric& other)
+{
+  if (this == &other)
+    return *this;
+
+  if(kernelOwner)
+    delete kernel;
+
+  kernel = NULL;
+  kernel = new KernelType(*other.kernel);
+  kernelOwner = true;
+  return *this;
+}
+
+template<typename KernelType>
 template<typename Vec1Type, typename Vec2Type>
 inline typename Vec1Type::elem_type IPMetric<KernelType>::Evaluate(
     const Vec1Type& a,
--- a/src/mlpack/core/metrics/ip_metric_impl.hpp
+++ b/src/mlpack/core/metrics/ip_metric_impl.hpp
@@ -61,7 +61,7 @@ IPMetric<KernelType>& IPMetric<KernelType>::operator=(const IPMetric& other)
   if (this == &other)
     return *this;
 
-  if(kernelOwner)
+  if (kernelOwner)
     delete kernel;
 
   kernel = NULL;
--- a/src/mlpack/core/metrics/ip_metric_impl.hpp
+++ b/src/mlpack/core/metrics/ip_metric_impl.hpp
@@ -55,8 +55,8 @@ inline typename Vec1Type::elem_type IPMetric<KernelType>::Evaluate(
 {
   // This is the metric induced by the kernel function.
   // Maybe we can do better by caching some of this?
-  return sqrt(abs(kernel->Evaluate(a, a) + kernel->Evaluate(b, b) -
-      2 * kernel->Evaluate(a, b)));
+  return sqrt(kernel->Evaluate(a, a) + kernel->Evaluate(b, b) -
+      2 * kernel->Evaluate(a, b));
 }
 
 // Serialize the kernel.
--- a/src/mlpack/core/metrics/ip_metric_impl.hpp
+++ b/src/mlpack/core/metrics/ip_metric_impl.hpp
@@ -13,7 +13,7 @@
 #define MLPACK_METHODS_FASTMKS_IP_METRIC_IMPL_HPP
 
 // In case it hasn't been included yet.
-#include "ip_metric_impl.hpp"
+#include "ip_metric.hpp"
 
 #include <mlpack/core/metrics/lmetric.hpp>
 #include <mlpack/core/kernels/linear_kernel.hpp>
--- a/src/mlpack/core/metrics/ip_metric_impl.hpp
+++ b/src/mlpack/core/metrics/ip_metric_impl.hpp
@@ -55,8 +55,8 @@ inline typename Vec1Type::elem_type IPMetric<KernelType>::Evaluate(
 {
   // This is the metric induced by the kernel function.
   // Maybe we can do better by caching some of this?
-  return sqrt(kernel->Evaluate(a, a) + kernel->Evaluate(b, b) -
-      2 * kernel->Evaluate(a, b));
+  return sqrt(abs(kernel->Evaluate(a, a) + kernel->Evaluate(b, b) -
+      2 * kernel->Evaluate(a, b)));
 }
 
 // Serialize the kernel.
--- a/src/mlpack/methods/fastmks/fastmks_model_impl.hpp
+++ b/src/mlpack/methods/fastmks/fastmks_model_impl.hpp
@@ -32,7 +32,7 @@ void BuildFastMKSModel(FastMKS<KernelType>& f,
 
   if (f.Naive())
   {
-    f.Train(referenceData, k);
+    f.Train(std::move(referenceData), k);
   }
   else
   {
--- a/src/mlpack/methods/fastmks/fastmks_model_impl.hpp
+++ b/src/mlpack/methods/fastmks/fastmks_model_impl.hpp
@@ -27,7 +27,7 @@ void BuildFastMKSModel(FastMKS<KernelType>& f,
   // Do we need to build the tree?
   if (base <= 1.0)
   {
-    throw std::invalid_argument("base must be greater than 2");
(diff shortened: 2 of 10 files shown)


src/mlpack/tests/main_tests/fastmks_test.cpp
```diff
@@ -0,0 +1,658 @@
...
+ */
+BOOST_AUTO_TEST_CASE(FastMKSModelReuseTest)
+{
+  // 100 points in 3 dimensions.
+  arma::mat referenceData(3, 100, arma::fill::randu);
+  // 90 points in 3 dimensions.
+  arma::mat queryData(3, 90, arma::fill::randu);
+
+  // Random input, some k <= number of reference points.
+  SetInputParam("reference", referenceData);
```
rcurtin: You could use `std::move()` here if you liked, but no big deal either way.

src/mlpack/tests/main_tests/fastmks_test.cpp
```diff
@@ -0,0 +1,658 @@
...
+  kernel = std::move(CLI::GetParam<arma::mat>("kernels"));
+
+  bindings::tests::CleanMemory();
+
+  CLI::GetSingleton().Parameters()["reference"].wasPassed = false;
+  CLI::GetSingleton().Parameters()["query"].wasPassed = false;
+
+
+  SetInputParam("reference", referenceData);
+  SetInputParam("query", queryData);
```
rcurtin: I'm a bit confused here, you've passed the same dataset twice.  I thought the idea was to check that we get the same results when we leave `query` unset, so the query set is the reference set.

Also, technically, the results will not be the same when that is done: if we specify `'reference'` as `referenceSet only, then each point cannot appear in its own results.  But if we specify `'reference'` as `referenceSet` and also `'query'` as `referenceSet`, then, e.g., point 0 can appear (and is likely to) in the top-k max-kernel results for point 0.  However, you could run the test, and then in the comparison of any results, account for this by skipping over any cases where the point appears in its own results.  Let me know if I can clarify that point.

src/mlpack/tests/main_tests/fastmks_test.cpp
```diff
@@ -0,0 +1,658 @@
...
+  newkernel = std::move(CLI::GetParam<arma::mat>("kernels"));
+
+  CheckMatrices(indices, newindices);
+  CheckMatrices(kernel, newkernel);
+}
+
+/**
+  * Check that we can't specify base less than 2.
+  */
+BOOST_AUTO_TEST_CASE(FastMKSBaseTest)
```
rcurtin: I don't think this should be true.  Please let me know why you added this test (maybe I overlooked something?).  I think that any base > 1 should be fine, but correct me if I'm wrong.

src/mlpack/tests/main_tests/fastmks_test.cpp
```diff
@@ -0,0 +1,658 @@
...
+    // Same random inputs, different algorithms.
+    SetInputParam("reference", referenceData);
+    SetInputParam("query", queryData);
+    SetInputParam("kernel", kerneltypes[i]);
+
+    mlpackMain();
+
+    if (i == 0)
+    {
+      indices_compare =
```
rcurtin: Tiny style issue: please don't use variables with `_` in the name; instead call this `indicesCompare` or similar. :+1:

src/mlpack/tests/main_tests/fastmks_test.cpp
```diff
@@ -0,0 +1,658 @@
...
+  CLI::GetSingleton().Parameters()["bandwidth"].wasPassed = false;
+
+  SetInputParam("reference", referenceData);
+  SetInputParam("bandwidth", 4.0);
+
+  mlpackMain();
+
+  CheckMatricesNotEqual(triKernel,
+      CLI::GetParam<arma::mat>("kernels"));
+}
```
rcurtin: We should add a blank line between the end of this method and `BOOST_AUTO_TEST_SUITE_END()`. :+1:

src/mlpack/core/metrics/ip_metric_impl.hpp
```diff
@@ -55,8 +55,8 @@ inline typename Vec1Type::elem_type IPMetric<KernelType>::Evaluate(
 {
   // This is the metric induced by the kernel function.
   // Maybe we can do better by caching some of this?
-  return sqrt(kernel->Evaluate(a, a) + kernel->Evaluate(b, b) -
-      2 * kernel->Evaluate(a, b));
+  return sqrt(abs(kernel->Evaluate(a, a) + kernel->Evaluate(b, b) -
+      2 * kernel->Evaluate(a, b)));
```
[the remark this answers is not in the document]
Yashwants19: It leads to tests running forever for hyptan kernel.

src/mlpack/methods/fastmks/fastmks_model_impl.hpp
```diff
@@ -25,6 +25,11 @@ void BuildFastMKSModel(FastMKS<KernelType>& f,
                        const double base)
 {
   // Do we need to build the tree?
+  if (base < 2.0)
+  {
+    throw std::invalid_argument("base must be greater than 2");
+  }
```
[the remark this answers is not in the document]
Yashwants19: I add this because we must check that `base >1`

src/mlpack/tests/main_tests/fastmks_test.cpp
Yashwants19: Please let me know that if I have done some correct or I making some big blunder here. :)

Yashwants19:
Hi @rcurtin  I have made change changes as suggested.

src/mlpack/tests/main_tests/fastmks_test.cpp
```diff
@@ -0,0 +1,658 @@
...
+BOOST_AUTO_TEST_CASE(FastMKSRefModelTest)
+{
+  // 100 points in 3 dimensions.
+  arma::mat referenceData(3, 100, arma::fill::randu);
+
+  // Random input, some k <= number of reference points.
+  SetInputParam("reference", std::move(referenceData));
+  SetInputParam("k", (int) 10);
+
+  mlpackMain();
```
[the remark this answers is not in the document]
rcurtin: It doesn't look like this is fixed, can you call `SetInputParam()` to set `reference` before the second call to `mlpackMain()` below?

src/mlpack/tests/main_tests/fastmks_test.cpp
rcurtin: Now this looks better, thanks for the fix. :+1:

src/mlpack/tests/main_tests/fastmks_test.cpp
```diff
@@ -0,0 +1,657 @@
...
+  Log::Fatal.ignoreInput = false;
+}
+
+/**
+  * Ensure that different kernels returns different results.
+  */
+BOOST_AUTO_TEST_CASE(FastMKSKernelTest)
+{
+  std::string kerneltypes[] = {"polynomial", "cosine", "gaussian",
+      "epanechnikov", "triangular", "hyptan"};
```
rcurtin: If this is the test that is running forever for `hyptan`, I'd suggest a different strategy for `hyptan`.  I think you should be able to use the `vc2` dataset and that is more likely to work:

```
  arma::mat inputData;
  if (!data::Load("vc2.csv", inputData))
    BOOST_FAIL("Cannot load test dataset vc2.csv!");
```

Then you can use the `vc2` dataset for both the reference and query set.

Yashwants19:
Hi @rcurtin I have made the changes as suggested. Can you please help with the FastMKSTest issue.

src/mlpack/tests/fastmks_test.cpp
```diff
@@ -341,9 +341,9 @@ BOOST_AUTO_TEST_CASE(FastMKSModelLinearTest)
   FastMKSModel mNaive(FastMKSModel::LINEAR_KERNEL);
   FastMKSModel mSingle(FastMKSModel::LINEAR_KERNEL);
 
-  m.BuildModel(referenceData, lk, false, false, 2.0);
-  mNaive.BuildModel(referenceData, lk, false, true, 2.0);
-  mSingle.BuildModel(referenceData, lk, true, false, 2.0);
+  m.BuildModel(std::move(referenceData), lk, false, false, 2.0);
+  mNaive.BuildModel(std::move(referenceData), lk, false, true, 2.0);
+  mSingle.BuildModel(std::move(referenceData), lk, true, false, 2.0);
```
rcurtin: When you call `std::move()`, the internal parts of the object are transferred to the `BuildModel()` method, and so after you call `std::move(referenceData)`, then `referenceData` is an empty object.  Therefore, in situations where you need to use `referenceData` more than once, don't use `std::move()`.  I think this will fix the failing tests.

Yashwants19: Thanks for the help. Updated :)

Yashwants19:

Yashwants19:
Thank you for reviewing this PR and help. :)  Sure we can wait for the static code analysis checks. 
I want to work on Extra Tree classifier so if you can guide me. I can start with Extra Tree classifier after completing Go bindings PR.

rcurtin:
Thanks for the contribution! :+1: I fixed a few style issues in 710b6658.

Yashwants19:
Hi @rcurtin can I delete this branch. We can also close #1356. Can i also work on Extremely randomized tree.

rcurtin:
I'm not sure why Github did not mark this as merged, but yeah, you can go ahead and delete the branch---the commits are a part of master now.

You're more than welcome to work on extremely randomized trees (or anything you like); the only thing I can't guarantee is how quickly I'll be able to review the PR. :)

PR #1749, opened 2019-02-28
Extend1726

mirraaj:
Add feature to Laplace Distribution

PR #1766, opened 2019-03-08
Add braces to for loop.

mirraaj:
Add braces to loop as suggested by @ShikharJ in #1749 . Small Changes

### DIFF
--- a/src/mlpack/core/dists/discrete_distribution.hpp
+++ b/src/mlpack/core/dists/discrete_distribution.hpp
@@ -181,9 +181,7 @@ class DiscreteDistribution
   {
     probabilities.set_size(x.n_cols);
     for (size_t i = 0; i < x.n_cols; i++)
-    {
       probabilities(i) = Probability(x.unsafe_col(i));
-    }
   }
 
   /**
@@ -259,9 +257,7 @@ inline void DiscreteDistribution::LogProbability(
 {
   logProbabilities.set_size(x.n_cols);
   for (size_t i = 0; i < x.n_cols; i++)
-  {
     logProbabilities(i) = log(Probability(x.unsafe_col(i)));
-  }
 }
 
 } // namespace distribution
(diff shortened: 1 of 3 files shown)


zoq:
Personally, I think it's fine to not wrap a single line if the line fits in the 80 character limit. So

```c++
for (...)
 I_think = this_is_fine;
```
is just fine, but

```c++
for (...)
{
  I_think = this_should_be_wraped_since_this_command(
    does_exceed_the_80_character_limit);
    
}
```

should be wraped.

mirraaj:
@zoq I completely agree with you. This change in style is quite redundant. However, I think it's better to be consistent in implementation. Since I have added such changes in #1749, I went ahead and added these changes to other distributions as well. Let me know if you are willing to add these changes. Personally, I would like this to be merged. However, I rest the decision upon the members of the organization.

Thanks

zoq:
I don't mind to merge this in, but I think it would be nice if we can agree on something. Either we don't really bother and it's fine to do both (I think this is the current "rule") or we go with a specific style.

mirraaj:
@zoq I am grateful to you for sharing your thoughts. I think you are correct about the rules to be followed for such Pull Requests. I think it is a great opportunity to discuss how to tackle such style issues in future. For current, I believe it is fine to get this PR merged.

Thanks

rcurtin:
Personally I like the style that Marcus suggested where if it's a single-line non-wrapped loop (or if) then there's no need for braces.  It can help keep the code a little bit shorter.  But definitely multi-line (or wrapped) statements should have braces.

rcurtin:
So, I guess, what should we do with this one?  Personally I would be in favor of leaving the code the way it is.  Could we clarify the style guide any?

mirraaj:
How about I go ahead and remove the braces changes and just keep the impementation of Probability function? @rcurtin

mirraaj:
@rcurtin How about adding these changes?

ShikharJ:
@mirraj Can you include this in #1775? We can close this then.

PR #1771, opened 2019-03-10
Make models more accessible in Python (#1709)

MuLx10:
Make models more accessible in Python #1709  . Added SerializeParamOut with boost xml_archive. 
This PR removes xml2json dependency in #1734 . I will close #1734 .

```python
>>> import numpy as np
>>> from mlpack import pythonize_model 
>>> from mlpack import logistic_regression

>>> x = np.random.rand(10,10)
>>> y = [1,0,1,0,1,0,1,0,0,0]

>>> output = logistic_regression(training=x,labels=y)
>>> pythonize_model.transform(output['output_model'])

{'parameters': array([  7.14841847,  72.40319186, -22.58314159,  40.70852354,
       -27.78725555, -75.40210697,  29.39998803,  14.44999113,
       -60.09246953,  18.63193182, -17.18063739]), 'lambda': 0}
```

### DIFF
--- a/src/mlpack/bindings/python/mlpack/pythonize_model.py
+++ b/src/mlpack/bindings/python/mlpack/pythonize_model.py
@@ -16,92 +16,92 @@ from collections import defaultdict
 from xml.etree import cElementTree as ET
 
 def etree_to_dict(t, attrib = False):
-    """
-    Convert ElementTree to python dict.
-    """
-    d = {t.tag: {} if t.attrib else None}
-    children = list(t)
-    if children:
-        dd = defaultdict(list)
-        for dc in map(etree_to_dict, children):
-            for k, v in dc.items():
-                dd[k].append(v)
-        d = {t.tag: {k:v[0] if len(v) == 1 else v for k, v in dd.items()}}
-    if t.attrib and attrib:
-        d[t.tag].update(('@' + k, v) for k, v in t.attrib.items())
-    if t.text:
-        text = t.text.strip()
-        if children or t.attrib:
-            if text:
-              d[t.tag]['#text'] = text
-        else:
-            d[t.tag] = text
-    return d
+  """
+  Convert ElementTree to python dict.
+  """
+  d = {t.tag: {} if t.attrib else None}
+  children = list(t)
+  if children:
+    dd = defaultdict(list)
+    for dc in map(etree_to_dict, children):
+      for k, v in dc.items():
+        dd[k].append(v)
+    d = {t.tag: {k:v[0] if len(v) == 1 else v for k, v in dd.items()}}
+  if t.attrib and attrib:
+    d[t.tag].update(('@' + k, v) for k, v in t.attrib.items())
+  if t.text:
+    text = t.text.strip()
+    if children or t.attrib:
+      if text:
+        d[t.tag]['#text'] = text
+    else:
+      d[t.tag] = text
+  return d
 
 # Vector state mapping.
 vec_state = {0: "matrix", 1: "column vector", 2: "row vector"}
 
 def recursive_transform(json):
-    """
-    Recursiely traverse and convert array to numpy
-    std::pair<> to python tuple().
-    """
-    # Check if the parameter is dict.
-    if type(json).__name__ == 'dict':
-        keys = list(json.keys())
+  """
+  Recursiely traverse and convert array to numpy
+  std::pair<> to python tuple().
+  """
+  # Check if the parameter is dict.
+  if type(json).__name__ == 'dict':
+    keys = list(json.keys())
 
-        # Check if serialized object was a std::pair<>, std::unordered_map.
-        if any(key in keys for key in ['first', 'second']):
-            pair = (recursive_transform(json['first']), recursive_transform(json['second']))
-            for key in ['first', 'second']:
-                del json[key]
-            return pair
+    # Check if serialized object was a std::pair<>, std::unordered_map.
+    if any(key in keys for key in ['first', 'second']):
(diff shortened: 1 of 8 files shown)


MuLx10:
@rcurtin [In response to](https://github.com/mlpack/mlpack/pull/1734#pullrequestreview-212582146) for complex models
  
The parameter structure should be known .
Eg. sparse_autoencoder follows the following structure.

		//          vSize   1
		//       |        |  |
		//  hSize|   w1   |b1|
		//       |________|__|
		//       |        |  |
		//  hSize|   w2'  |  |
		//       |________|__|
		//      1|   b2'  |  |

Somehow I need to exploit the above information so that the inference would be more user friendly.

MuLx10:
@rcurtin [In response to](https://github.com/mlpack/mlpack/pull/1734#pullrequestreview-212582146) for complex models
  
The parameter structure should be known .
Eg. sparse_autoencoder follows the following structure.

		          vSize   1
		       |        |  |
		  hSize|   w1   |b1|
		       |________|__|
		       |        |  |
		  hSize|   w2'  |  |
		       |________|__|
		      1|   b2'  |  |

Somehow I need to exploit the above information so that the inference would be more user friendly.

rcurtin:
The sparse autoencoder doesn't have an associated `sparse_autoencoder_main.cpp`, so we never build a binding for it; so I'm not sure why you mentioned that one in particular.

However you are right that there are complex models for which we may not be able to infer the structure that we would ideally present to the user.  But in these cases, the structure and meaning of the parameters may be unknown to the bindings.  So I'm glad you pointed this out, but for what we are trying to do here I think we should focus on doing our best to provide something that's reasonable.

Let's take another example, `perceptron`.  The definition of `Perceptron::serialize()` is this:

```
//! Serialize the perceptron.
template<typename LearnPolicy,
         typename WeightInitializationPolicy,
         typename MatType>
template<typename Archive>
void Perceptron<LearnPolicy, WeightInitializationPolicy, MatType>::serialize(
    Archive& ar,
    const unsigned int /* version */)
{
  // We just need to serialize the maximum number of iterations, the weights,
  // and the biases.
  ar & BOOST_SERIALIZATION_NVP(maxIterations);
  ar & BOOST_SERIALIZATION_NVP(weights);
  ar & BOOST_SERIALIZATION_NVP(biases);
}
```

So from boost::serialization we will get some XML with the keys `maxIterations`, `weights`, and `biases`.  The best we can provide to a Python user, then, is a Python dict with exactly those three keys, and appropriate types held for each.

Basically, what I am saying is, we will have to do the best with the information we receive from boost::serialization, which will be entirely:

 * keys for each serialized object
 * values of each serialized object
 * types of each serialized object (although these may need to be inferred from the structure!)

Let me know if I can clarify anything.

MuLx10:
@rcurtin That's exactly what I wanted to say.  We cannot get a lot of information from the ` serialize() `. How I chose sparse_autoencoder jus because I came across its weights structure first.

src/mlpack/bindings/python/mlpack/pythonize_model.py
```diff
@@ -0,0 +1,72 @@
...
+        if type(model).__name__ != 'str':
+            model = model.__getparams__()
+    except Exception as e:
+        raise e
+    
+    model_dict = xml_to_dict(model)
+
+    # network parameters
+    try:
+        result["parameters"] = model_dict["parameters"]["item"]
```
rcurtin: Right, so to respond to this:

> @rcurtin That's exactly what I wanted to say. We cannot get a lot of information from the serialize(). However I tried to extract as many information as possible in the PR.

I agree completely, but what I think that we should do here in `transform()` is try and take a look at what we have in the dict, and label it accordingly.  For instance, if we detect the Armadillo serialization members (like `n_rows`, `n_cols`, `n_elem`, etc.) we should replace that part of the dict with a numpy `ndarray`.  You can take a look at `src/mlpack/core/arma_extend/Mat_extra_meat.hpp` to see the matrix serialization details.  (There is also serialization for sparse matrices in `src/mlpack/core/arma_extend/SpMat_extra_meat.hpp` and we can use that to convert the corresponding dict element to a scipy sparse matrix or something.)

MuLx10: Apologies for the delay. I have incorporated the changes

src/mlpack/bindings/python/mlpack/pythonize_model.py
```diff
@@ -0,0 +1,107 @@
...
+
+
+import numpy as np
+from collections import defaultdict
+from xml.etree import cElementTree as ET
+
+def etree_to_dict(t, attrib = False):
+    """
+    Convert ElementTree to python dict.
+    """
```
rcurtin: Looks like this file is done with four spaces per tab, but can we do it to match the rest of the mlpack style instead?

src/mlpack/bindings/python/mlpack/serialization.hpp
```diff
@@ -30,6 +30,17 @@ std::string SerializeOut(T* t, const std::string& name)
   return oss.str();
 }
 
+template<typename T>
+std::string SerializeParamsOut(T* t, const std::string& name)
```
rcurtin: Would a better name be `SerializeToXML()` or similar?

src/mlpack/bindings/python/mlpack/pythonize_model.py
MuLx10: Apologies for the delay. I 'll do the needful.

rcurtin:
I'm still intending to get to this, just haven't had a chance yet, so I've marked it 'keep open'.

PR #1773, opened 2019-03-10
Refactor core.hpp and CONTRIBUTORS.md.

rcurtin:
These work a little bit better for the new website (I think).  I noticed that a lot of the old doxygen documentation in `src/mlpack/core.hpp` made sense back when doxygen was the main entrypoint to the documentation... but it's not anymore.  So now we can extract the list of contributors from `COPYRIGHT.txt` or the Github contributors list, and new contributors don't need to add their name in two places.

If everyone agrees with this, I'll go ahead and merge it after mlpack/mlpack.org#6 is merged.

### DIFF
--- a/src/mlpack/core.hpp
+++ b/src/mlpack/core.hpp
@@ -21,7 +21,7 @@
  * bindings to other languages.  It is meant to be a machine learning analog to
  * LAPACK, and aims to implement a wide array of machine learning methods and
  * function as a "swiss army knife" for machine learning researchers.  The
- * mlpack website can be found at http://mlpack.org.
+ * mlpack website can be found at https://mlpack.org.
  *
  * @section howto How To Use This Documentation
  *
@@ -59,7 +59,7 @@
  * @section remarks Final Remarks
  *
  * For the list of contributors to mlpack, see
- * http://www.mlpack.org/contributors.html.  This library would not be possible
+ * https://www.mlpack.org/contributors.html.  This library would not be possible
  * without everyone's hard work and contributions!
  */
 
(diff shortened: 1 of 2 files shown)


src/mlpack/core.hpp
```diff
@@ -21,13 +21,7 @@
...
  * LAPACK, and aims to implement a wide array of machine learning methods and
  * function as a "swiss army knife" for machine learning researchers.  The
- * mlpack development website can be found at http://mlpack.org.
- *
- * mlpack uses the Armadillo C++ matrix library (http://arma.sourceforge.net)
- * for general matrix, vector, and linear algebra support.  mlpack also uses the
- * program_options, math_c99, and unit_test_framework components of the Boost
- * library, and optionally uses libbfd and libdl to give backtraces when
- * compiled with debugging symbols on some platforms.
+ * mlpack website can be found at http://mlpack.org.
```
zoq: Maybe it's time to use https as a default?

src/mlpack/core.hpp
```diff
@@ -106,158 +56,11 @@
...
- *   - Arash Abghari <arash.abghari@gmail.com>
- *   - Ayush Chamoli
- *   - Tommi Laivamaa <tommi.laivamaa@protonmail.com>
- *   - Kim SangYeon <sy0814k@gmail.com>
- *   - Niteya Shah <niteya.56@gmail.com>
- *   - Toshal Agrawal <tagrawal1339@gmail.com>
- *   - Dan Timson
- *   - Miguel Canteras <mcanteras@gmail.com>
+ * For the list of contributors to mlpack, see
+ * http://www.mlpack.org/contributors.html.  This library would not be possible
```
zoq: See comment above.

src/mlpack/core.hpp
mirraaj: @zoq Does MlPack website support HTTPS?

rcurtin: Agreed, I hadn't noticed that.  Fixed in bf5506e.

rcurtin:
I merged manually, but Github didn't seem to recognize it.  So I'll go ahead and close this.

PR #1775, opened 2019-03-11
Remove Inline Implimentation

mirraaj:
This PR removes inline implementation of LogProbability function for different distributions as discussed in https://github.com/mlpack/mlpack/pull/1749#discussion_r264014523.

Thanks

### DIFF
--- a/src/mlpack/core/dists/gaussian_distribution.hpp
+++ b/src/mlpack/core/dists/gaussian_distribution.hpp
@@ -103,7 +103,24 @@ class GaussianDistribution
    * @param logProbabilities Output log probabilities for each input
    *     observation.
    */
-  void LogProbability(const arma::mat& x, arma::vec& logProbabilities) const;
+  void LogProbability(const arma::mat& x, arma::vec& logProbabilities) const
+  {
+    // Column i of 'diffs' is the difference between x.col(i) and the mean.
+    arma::mat diffs = x - (mean * arma::ones<arma::rowvec>(x.n_cols));
+
+    // Now, we only want to calculate the diagonal elements of (diffs' * cov^-1 *
+    // diffs).  We just don't need any of the other elements.  We can calculate
+    // the right hand part of the equation (instead of the left side) so that
+    // later we are referencing columns, not rows -- that is faster.
+    const arma::mat rhs = -0.5 * invCov * diffs;
+    arma::vec logExponents(diffs.n_cols); // We will now fill this.
+    for (size_t i = 0; i < diffs.n_cols; i++)
+      logExponents(i) = accu(diffs.unsafe_col(i) % rhs.unsafe_col(i));
+
+    const size_t k = x.n_rows;
+
+    logProbabilities = -0.5 * k * log2pi - 0.5 * logDetCov + logExponents;
+  }
 
   /**
    * Return a randomly generated observation according to the probability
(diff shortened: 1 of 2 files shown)


src/mlpack/core/dists/gaussian_distribution.hpp
```diff
@@ -103,7 +103,24 @@ class GaussianDistribution
    * @param logProbabilities Output log probabilities for each input
    *     observation.
    */
-  void LogProbability(const arma::mat& x, arma::vec& logProbabilities) const;
+  void LogProbability(const arma::mat& x, arma::vec& logProbabilities) const
+  {
+    // Column i of 'diffs' is the difference between x.col(i) and the mean.
+    arma::mat diffs = x - (mean * arma::ones<arma::rowvec>(x.n_cols));
```
zoq: I wonder if ``arma::mat diffs = x.each_col() - mean`` is something that would improve the readability and or the speed.

src/mlpack/core/dists/gaussian_distribution.hpp
```diff
@@ -103,7 +103,24 @@ class GaussianDistribution
...
+    // Now, we only want to calculate the diagonal elements of (diffs' * cov^-1
+    // * diffs).  We just don't need any of the other elements.  We can
+    // calculate the right hand part of the equation (instead of the left side)
+    // so that later we are referencing columns, not rows -- that is faster.
+    const arma::mat rhs = -0.5 * invCov * diffs;
+    arma::vec logExponents(diffs.n_cols); // We will now fill this.
+    for (size_t i = 0; i < diffs.n_cols; i++)
+      logExponents(i) = accu(diffs.unsafe_col(i) % rhs.unsafe_col(i));
+
+    const size_t k = x.n_rows;
```
zoq: Hm, why not just use ``x.n_rows`` instead of ``k``, it's only used once.

src/mlpack/core/dists/gaussian_distribution.hpp
mirraaj: @zoq I think that would improve readability. However, I am not quite sure about speed. Do you wish to know the results? I can try to run tests in my local server and check it out.

src/mlpack/core/dists/gaussian_distribution.hpp
mirraaj: I agree with this. I think we should remove this. Thanks

src/mlpack/core/dists/gaussian_distribution.hpp
mirraaj: Should we leave the implementation as it is?

mirraaj:
@ShikharJ I have merged #1766 with this PR. Thanks

mirraaj:
@rcurtin I think this PR is good to be merged.

src/mlpack/core/dists/gaussian_distribution.hpp
zoq: I would give ``each_col()`` a try.

mirraaj:
@zoq I have added the changes. Please let me know if any other change is needed.

mirraaj:
@rcurtin @zoq @ShikharJ I think this PR is set to be merged. Let me know if you have any other suggestions. Thanks

ShikharJ:
Thanks for the contribution @mirraj.

PR #1776, opened 2019-03-11
Add hard sigmoid

Bk073:
I wanted to make a review of my work. I have added a hard sigmoid function. Thank you so much

### DIFF
--- a/src/mlpack/methods/ann/activation_functions/hard_sigmoid_function.hpp
+++ b/src/mlpack/methods/ann/activation_functions/hard_sigmoid_function.hpp
@@ -1,0 +1,99 @@
+/**
+ * @file hard_sigmoid_function.hpp
+ * @author Bishwa Karki
+ *
+ * Definition and implementation of the hard sigmoid function.
+ *
+ * mlpack is free software; you may redistribute it and/or modify it under the
+ * terms of the 3-clause BSD license.  You should have received a copy of the
+ * 3-clause BSD license along with mlpack.  If not, see
+ * http://www.opensource.org/licenses/BSD-3-Clause for more information.
+ */
+#ifndef MLPACK_METHODS_ANN_ACTIVATION_FUNCTIONS_HARD_SIGMOID_FUNCTION_HPP
+#define MLPACK_METHODS_ANN_ACTIVATION_FUNCTIONS_HARD_SIGMOID_FUNCTION_HPP
+
+#include <mlpack/prereqs.hpp>
+#include <algorithm>
+
+namespace mlpack {
+namespace ann /** Artificial Neural Network. */ {
+
+/**
+ * The hard sigmoid function, defined by
+ *
+ * @f{eqnarray*}{
+ * f(x) &=& \min(1, \max(0, 0.2 * x + 0.5)) \\
+ * f'(x) &=& \left\{
+ *   \begin{array}{lr}
+ *     0.0 & : x={0,1} \\
+ *     0.2
+ *   \end{array}
+ * \right.
+ * @f}
+ */
+class HardSigmoidFunction
+{
+ public:
+  /**
+   * Computes the hard sigmoid function.
+   *
+   * @param x Input data.
+   * @return f(x).
+   */
+  static double Fn(const double x)
+  {
+    return std::min(1.0, std::max(0.0, 0.2 * x + 0.5));
+  }
+
+  /**
+   * Computes the hard sigmoid function.
+   *
+   * @param x Input data.
+   * @param y The resulting output activations.
+   */
+  template<typename InputVecType, typename OutputVecType>
+  static void Fn(const InputVecType& x, OutputVecType& y)
+  {
+    y.set_size(size(x));
+
+    for (size_t i = 0; i < x.n_elem; i++)
+      y(i) = Fn(x(i));
+  }
+
+  /**
+   * Computes the first derivatives of hard sigmoid function.
+   *
+   * @param y Input data.
+   * @return f'(x)
+   */
+  static double Deriv(const double y)
+  {
+    if (y == 0.0 || y == 1.0)
+    {
+      return 0.0;
+    }
+    return 0.2;
+  }
+
+  /**
+   * Computes the first derivatives of the hard sigmoid function.
+   *
+   * @param y Input activations.
+   * @param x The resulting derivatives.
+   */
+  template<typename InputVecType, typename OutputVecType>
+  static void Deriv(const InputVecType& y, OutputVecType& x)
+  {
+    x.set_size(size(y));
+
+    for (size_t i = 0; i < y.n_elem; i++)
+    {
+      x(i) = Deriv(y(i));
+    }
+  }
+}; // class HardSigmoidFunction
+
+} // namespace ann
+} // namespace mlpack
+
+#endif
--- a/src/mlpack/tests/activation_functions_test.cpp
+++ b/src/mlpack/tests/activation_functions_test.cpp
@@ -20,6 +20,7 @@
 #include <mlpack/methods/ann/activation_functions/rectifier_function.hpp>
 #include <mlpack/methods/ann/activation_functions/softplus_function.hpp>
 #include <mlpack/methods/ann/activation_functions/swish_function.hpp>
+#include <mlpack/methods/ann/activation_functions/hard_sigmoid_function.hpp>
 
 #include <boost/test/unit_test.hpp>
 #include "test_tools.hpp"
@@ -580,6 +581,26 @@ BOOST_AUTO_TEST_CASE(SwishFunctionTest)
 
   CheckActivationCorrect<SwishFunction>(activationData, desiredActivations);
   CheckDerivativeCorrect<SwishFunction>(desiredActivations,
+      desiredDerivatives);
+}
+
+/**
+ * Basic test of the hard sigmoid function.
+ */
+BOOST_AUTO_TEST_CASE(HardSigmoidFunctionTest)
+{
+  // Hand-calculated values using Python interpreter.
+  const arma::colvec desiredActivations("0.1 1 1 \
+                                         0 0.7 0.3 \
+                                         0.9 0.5");
+
+  const arma::colvec desiredDerivatives("0.2 0.0 0.0 \
+                                         0.0 0.2 0.2 0.2\
+                                         0.2");
+
+  CheckActivationCorrect<HardSigmoidFunction>(activationData,
+      desiredActivations);
+  CheckDerivativeCorrect<HardSigmoidFunction>(desiredActivations,
       desiredDerivatives);
 }
 
--- a/src/mlpack/methods/ann/layer/base_layer.hpp
+++ b/src/mlpack/methods/ann/layer/base_layer.hpp
@@ -19,6 +19,7 @@
 #include <mlpack/methods/ann/activation_functions/rectifier_function.hpp>
 #include <mlpack/methods/ann/activation_functions/tanh_function.hpp>
 #include <mlpack/methods/ann/activation_functions/softplus_function.hpp>
+#include <mlpack/methods/ann/activation_functions/hard_sigmoid_function.hpp>
 
 namespace mlpack {
 namespace ann /** Artificial Neural Network. */ {
@@ -172,6 +173,16 @@ template <
 using SoftPlusLayer = BaseLayer<
     ActivationFunction, InputDataType, OutputDataType>;
 
+/**
+ * Standard HardSigmoid-Layer using the HardSigmoid activation function.
+ */
+template <
+    class ActivationFunction = HardSigmoidFunction,
+    typename InputDataType = arma::mat,
+    typename OutputDataType = arma::mat
+>
+using HardSigmoidLayer = BaseLayer<
+    ActivationFunction, InputDataType, OutputDataType>;
 
 } // namespace ann
 } // namespace mlpack


src/mlpack/methods/ann/activation_functions/hard_sigmoid_function.hpp
```diff
@@ -0,0 +1,184 @@
...
+ * if x > 2.5 || x < -2.5
+ *    f'(x) &=& 0 \\
+ * else 
+ *    f'(x) &=& 0.2
+ * @f}
+ */
+class HardSigmoidFunction
+{
+ public:
+  // /**
```
zoq: Looks like this is redundant with the one below, except that this one has a parameter description.

src/mlpack/methods/ann/activation_functions/hard_sigmoid_function.hpp
```diff
@@ -0,0 +1,184 @@
...
+  //  * @param x Input data.
+  //  * @return f(x).
+  //  */
+  /**
+   * Computes the hard sigmoid function.
+   * 
+   * @param x 
+   * @return double 
+   */
+
```
zoq: We can remove the extra space here.

src/mlpack/methods/ann/activation_functions/hard_sigmoid_function.hpp
```diff
@@ -0,0 +1,184 @@
...
+  }
+
+  // /**
+  //  * Computes the hard sigmoid function.
+  //  *
+  //  * @param x Input data.
+  //  * @param y The resulting output activation.
+  //  */
+  /**
+   * Computes the hard sigmoid function.
```
zoq: See comment above.

src/mlpack/methods/ann/activation_functions/hard_sigmoid_function.hpp
```diff
@@ -0,0 +1,184 @@
...
+   * 
+   * @param x 
+   * @return double 
+   */
+  
+  static double Fn(const double x)
+  {
+    if (x < -2.5)
```
zoq: What about writing ``std::min(1.0, std::max(0, 0.2 * x + 0.5));``, let me know what you think.

src/mlpack/methods/ann/activation_functions/hard_sigmoid_function.hpp
```diff
@@ -0,0 +1,184 @@
...
+   * @param y
+   * @return double
+   */
+  static double Deriv(const double y)
+  {
+    if(y < -2.5 || y > 2.5)
+    {
+      return 0.0;
+    }
+    else
```
zoq: I guess to save some space we could remove the else here and return right away.

src/mlpack/methods/ann/activation_functions/hard_sigmoid_function.hpp
```diff
@@ -0,0 +1,184 @@
...
+   * @tparam OutputVecType
+   * @param y
+   * @param x
+   */
+  template<typename InputVecType, typename OutputVecType>
+  static void Deriv(const InputVecType& y, OutputVecType& x)
+  {
+    for (size_t i = 0; i < y.n_elem; i++)
+    {
+      if(y[i] < -2.5 || y[i] > 2.5)
```
zoq: We could use the function from above.

src/mlpack/methods/ann/activation_functions/hard_sigmoid_function.hpp
```diff
@@ -0,0 +1,184 @@
...
+        x = 0.0;
+      }
+      else
+      {
+        x = 0.2;
+      }
+    }
+  }
+  // /**
+  //  * Computes the inverse of the hard sigmoid function.
```
zoq: Looks like we can remove the rest here.

src/mlpack/tests/activation_functions_test.cpp
```diff
@@ -243,13 +244,11 @@ void CheckELUDerivativeCorrect(const arma::colvec input,
   ELU<> lrf(1.0);
 
   // Test the calculation of the derivatives using the entire vector as input.
-  arma::colvec derivatives, activations;
+  arma::colvec derivatives;
```
zoq: Not sure I see the reason for the changes here, can you elaborate on that one?

Bk073: I am sorry, that was  changed unknowingly.

src/mlpack/methods/ann/activation_functions/hard_sigmoid_function.hpp
Bk073: Thank you @zoq  this seems clean and efficient. I have updated to this one but 0 is not typecast to double so there would be 0.0

src/mlpack/methods/ann/activation_functions/hard_sigmoid_function.hpp
Bk073: @zoq Thank you

src/mlpack/methods/ann/activation_functions/hard_sigmoid_function.hpp
```diff
@@ -0,0 +1,104 @@
...
+   *
+   * @tparam InputVecType
+   * @tparam OutputVecType
+   * @param x
+   * @param y
+   */
+  template<typename InputVecType, typename OutputVecType>
+  static void Fn(const InputVecType& x, OutputVecType& y)
+  {
+    y = x;
```
ShikharJ: I think we just need to use `resize()` here.

src/mlpack/methods/ann/activation_functions/hard_sigmoid_function.hpp
```diff
@@ -0,0 +1,104 @@
...
+   * @param y
+   * @param x
+   */
+  template<typename InputVecType, typename OutputVecType>
+  static void Deriv(const InputVecType& y, OutputVecType& x)
+  {
+    for (auto& elem : y)
+    {
+      x = Deriv(elem);
+    }
```
ShikharJ: For consistency, please keep the same implementation style as `Fn()`.

src/mlpack/methods/ann/activation_functions/hard_sigmoid_function.hpp
```diff
@@ -0,0 +1,104 @@
...
+  }
+}; // class hard sigmoidFunction
```
ShikharJ: This should be `class HardSigmoidFunction`.

src/mlpack/tests/activation_functions_test.cpp
```diff
@@ -583,4 +584,24 @@ BOOST_AUTO_TEST_CASE(SwishFunctionTest)
       desiredDerivatives);
 }
 
+/**
+ * Basic test of the Hard sigmoid function.
```
ShikharJ: Please convert the above uppercase words to lowercase.

src/mlpack/methods/ann/activation_functions/hard_sigmoid_function.hpp
zoq: Better to use ``set_size``, since it should be much faster.

src/mlpack/methods/ann/activation_functions/hard_sigmoid_function.hpp
Bk073: You mean I should not use auto for getting every instance ?

ShikharJ: I mean that you should make use of the for loop as used in `Fn()`.

Bk073: @ShikharJ Thank you . I have changed accordingly.

ShikharJ: @Bk073 I doesn't seem like you have.

Bk073: @ShikharJ I think now I have.

src/mlpack/methods/ann/activation_functions/hard_sigmoid_function.hpp
```diff
@@ -0,0 +1,101 @@
...
+   * @param x Input data.
+   * @param y The resulting output activation.
+   */
+  template<typename InputVecType, typename OutputVecType>
+  static void Fn(const InputVecType& x, OutputVecType& y)
+  {
+    y.set_size(size(x));
+
+    for (size_t i = 0; i < x.n_elem; i++)
+      y[i] = Fn(x(i));
```
ShikharJ: I think we should keep this as `y(i) = Fn(x(i));` just to be consistent.

src/mlpack/methods/ann/activation_functions/hard_sigmoid_function.hpp
```diff
@@ -0,0 +1,101 @@
...
+  /**
+   * Computes the first derivatives of the hard sigmoid function.
+   *
+   * @param y Input activations.
+   * @param x The resulting derivatives.
+   */
+  template<typename InputVecType, typename OutputVecType>
+  static void Deriv(const InputVecType& y, OutputVecType& x)
+  {
+    for (size_t i = 0; i < y.n_elem; i++)
```
ShikharJ: You forgot the `x.set_size(size(y));`. Also, it should be `x(i) = Deriv(y(i));` inside the loop.

Bk073: @ShikharJ Thank you :)

src/mlpack/methods/ann/activation_functions/hard_sigmoid_function.hpp
```diff
@@ -0,0 +1,103 @@
...
+
+  /**
+   * Computes the first derivatives of hard sigmoid function.
+   *
+   * @param y Input data.
+   * @return f'(x)
+   */
+  static double Deriv(const double y)
+  {
+    if (y < -2.5 || y > 2.5)
```
ShikharJ: I think the implementation here is a bit incorrect, since you're making use of `f(x)` here in place of `x`. So this here should just be:
```
if (y == 0.0 || y == 1.0)
```

Bk073: @ShikharJ This gives me test case error saying difference between input and target output exceeds 0.001%

Bk073: I had tried this before and test case gave the same error.

ShikharJ: That's because your `desiredDerivatives` matrix is wrong.

Bk073: @ShikharJ  Thank you so much

src/mlpack/methods/ann/activation_functions/hard_sigmoid_function.hpp
```diff
@@ -0,0 +1,103 @@
...
+  static double Fn(const double x)
+  {
+    return std::min(1.0, std::max(0.0, 0.2 * x + 0.5));
+  }
+
+  /**
+   * Computes the hard sigmoid function.
+   *
+   * @param x Input data.
+   * @param y The resulting output activation.
```
ShikharJ: Please change `activation` to `activations`.

src/mlpack/methods/ann/activation_functions/hard_sigmoid_function.hpp
```diff
@@ -0,0 +1,103 @@
...
+#include <algorithm>
+
+namespace mlpack {
+namespace ann /** Artificial Neural Network. */ {
+
+/**
+ * The hard sigmoid function, defined by
+ *
+ * @f{eqnarray*}{
+ * if x < -2.5 \\
```
zoq: This is no valid latex code, can you check the code.

src/mlpack/methods/ann/activation_functions/hard_sigmoid_function.hpp
```diff
@@ -0,0 +1,103 @@
...
+ * else \\
+ *    @return x * .2 + .5
+ * endif
+ * if x > 2.5 || x < -2.5
+ *    f'(x) &=& 0 \\
+ * else
+ *    f'(x) &=& 0.2
+ * @f}
+ */
+class HardSigmoidFunction
```
zoq: Can you add another alias to the ``base_layer.hpp``for the hard sigmoid function:

https://github.com/mlpack/mlpack/blob/6faf0b782705c29267fbfeb1eca527b6006e5a04/src/mlpack/methods/ann/layer/base_layer.hpp#L167-L174

src/mlpack/methods/ann/activation_functions/hard_sigmoid_function.hpp
ShikharJ: Ack, should've paid more attention here. What is the standard way of checking for LaTeX code in pull requests?

ShikharJ:
@Bk073 It seems you have also committed the binary files for VSCode? Can you remove them?

Bk073:
@zoq @ShikharJ Please review again I have squashed the commits and made suggested changes.

src/mlpack/methods/ann/activation_functions/hard_sigmoid_function.hpp
zoq: One option might be to look at the latex job output, not ideal but it works.

rcurtin: Yeah, if you have LaTeX installed, `make doc` should give an error.

src/mlpack/methods/ann/activation_functions/hard_sigmoid_function.hpp
```diff
@@ -0,0 +1,99 @@
...
+#include <algorithm>
+
+namespace mlpack {
+namespace ann /** Artificial Neural Network. */ {
+
+/**
+ * The hard sigmoid function, defined by
+ *
+ * @f{eqnarray*}{
+ * f(x) &=& \min(1, max(0, {0.2 * x + 0.5}) \\
```
zoq: Missing ``\`` before max, also no need to use ``{}`` inside the max function. Sorry for being so picky.

src/mlpack/methods/ann/activation_functions/hard_sigmoid_function.hpp
ShikharJ: Hmm, I'll have to pull in the changes for that. I wonder why LaTeX Documentation check shows a valid build though?

zoq: Doxygen is still producing some output it just looks wrong.

Bk073:
@zoq Made suggested changes.

ShikharJ:
Thanks for the contribution @Bk073!

Issue #1661, opened 2019-01-14
`GradientAtrousConvolutionTest` and `GradientTransposedConvolutionTest` fail often; is `Gradient()` implemented right?


rcurtin:
I noticed when working on #1640 that I had to significantly adjust the tolerance of `ANNLayerTest/GradientAtrousConvolutionTest` in order to get it to reliably pass.  Often the difference could be quite large, even larger than 0.1.  I am concerned that this might be an implementation bug.

So, I think that `src/mlpack/methods/ann/layer/atrous_convolution_impl.hpp` needs to be reviewed to ensure that both the implementations of `Evaluate()` and `Gradient()` are correct.

@zoq, @ShikharJ: could I be wrong about this mattering?  If a larger tolerance is okay with both of you and you are both happy with the implementation, we can close this issue without doing anything.

Relevant: #1640 #1777

ShikharJ:
@rcurtin Hmm, the code was merged in a hurry as far as I remember, though I don't think I suspect its correctness. I'll take a thorough look and reach back.

saksham189:
This is quite strange but adding an `IdentityLayer<>` before the convolution layer in the test seems to reduce the failures and works with a tolerance of `1e-2`

saksham189:
I tried debugging this a bit more. Every-time the test fails the `inputTemp` variable inside the `AtrousConvolution` seems to be changed/corrupted. 
The input to the NN is `arma::linspace<arma::colvec>(0, 35, 36)`.
After printing the `inputTemp` variable inside the `Forward` method I get the output below which is what I would expect. 
```
         0    6.0000   12.0000   18.0000   24.0000   30.0000
    1.0000    7.0000   13.0000   19.0000   25.0000   31.0000
    2.0000    8.0000   14.0000   20.0000   26.0000   32.0000
    3.0000    9.0000   15.0000   21.0000   27.0000   33.0000
    4.0000   10.0000   16.0000   22.0000   28.0000   34.0000
    5.0000   11.0000   17.0000   23.0000   29.0000   35.0000
```
Whenever the test fails the `inputTemp` variable inside the `Gradient()` method `inputTemp` has some corrupted values as such below:
```
  3.1050e+169   6.0000e+00   1.2000e+01   1.8000e+01   2.4000e+01   3.0000e+01
  9.8216e+252   7.0000e+00   1.3000e+01   1.9000e+01   2.5000e+01   3.1000e+01
[...]

PR #1777, opened 2019-03-15
Make `GradientTransposedConvolutionLayerTest` robust.

KimSangYeon-DGU:
Because the `ANNLayerTest/GradientTransposedConvolutionLayerTest` is probabilistic test, I noticed it was failed sometimes, so to make it robust, check it five times.

### DIFF
--- a/src/mlpack/tests/ann_layer_test.cpp
+++ b/src/mlpack/tests/ann_layer_test.cpp
@@ -1481,39 +1481,49 @@ BOOST_AUTO_TEST_CASE(SimpleTransposedConvolutionLayerTest)
 BOOST_AUTO_TEST_CASE(GradientTransposedConvolutionLayerTest)
 {
   // Add function gradient instantiation.
-  struct GradientFunction
+  // To make this test robust, check it five times.
+  bool pass = false;
+  for (size_t trial = 0; trial < 5; trial++)
   {
-    GradientFunction()
+    struct GradientFunction
     {
-      input = arma::linspace<arma::colvec>(0, 35, 36);
-      target = arma::mat("1");
+      GradientFunction()
+      {
+        input = arma::linspace<arma::colvec>(0, 35, 36);
+        target = arma::mat("1");
 
-      model = new FFN<NegativeLogLikelihood<>, RandomInitialization>();
-      model->Predictors() = input;
-      model->Responses() = target;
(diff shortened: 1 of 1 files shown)


saksham189:
I think we could increase tolerance for the test if it fails often rather than having more trials.
let me know what you think. Maybe we need to check the implementation.

KimSangYeon-DGU:
> I think we could increase tolerance for the test if it fails often rather than having more trials.
let me know what you think.

Hey @saksham189. Thanks for your suggestion. :) Actually, before opening this PR, I also tested the method to increase the tolerance, and `0.3` is quite robust(right, it is also quite large). However, the `CheckGradient()` in the test sometimes outputs about `1`. Thus, I decided to have more trials.

saksham189:
This https://github.com/mlpack/mlpack/issues/1661 seems related.

ShikharJ:
@KimSangYeon-DGU Was this observation limited to the `OSX` builds, or was it common to other builds as well?

KimSangYeon-DGU:
> Was this observation limited to the OSX builds, or was it common to other builds as well?

Hi @ShikharJ. There are two observations. 

Firstly, in my local computer, Linux(Ubuntu 16.04), this error occurred but the `Linux` builds of Travis CI, it didn't happen.

Lastly, when I'm working #1666, I found this error in `OSX` builds.

ShikharJ:
I see, in that case it may be a sound strategy to go with this implementation. DO you think you wish to take a look into #1661 as well? I couldn't find an issue there, so maybe I'm overlooking something. It'd be nice to have another set of eyes to scrutinize the implementation.

KimSangYeon-DGU:
> Do you think you wish to take a look into #1661 as well?

Is it okay to take some time? Actually, as the GSoC application is approaching, I'm currently spending my time in writing proposal. At the same time, I'm preparing both some extended works from #1666 and my classes as a training assistant in my university.

ShikharJ:
@KimSangYeon-DGU Sure, take as much time as you require.

ShikharJ:
@rcurtin Do you think we should merge these changes in for now?

rcurtin:
@ShikharJ agreed, I think it's reasonable to merge this now to at least fix the unstable builds, and then we can address the deeper issue as we are able. :+1:

ShikharJ:
Thanks for another great contribution @KimSangYeon-DGU.

PR #1779, opened 2019-03-15
fix #1778 removed extra 'given' in comment at line 71

### LINKED ISSUE #1778: grammatical mistake in  comment in mlpack/src/mlpack/core/data/imputer.hpp

greatsharma:


src/mlpack/core/data/imputer.hpp
```diff
@@ -68,7 +68,7 @@ class Imputer
   //! Get the strategy
   const StrategyType& Strategy() const { return strategy; }
 
-  //! Modify the given given strategy (be careful!)
+  //! Modify the given strategy (be careful!)
```
zoq: On a second thought, I think the "be careful" might be confusing, or is there something special about this as compared with the modifier that we use elsewhere?

ShikharJ: I guess something along the lines of why one needs to be careful might be more insightful. But then I have no idea. Maybe @keon could provide some insights on this?

ShikharJ: Makes sense. @zoq Do you think we should change this line, or should we merge as is?

zoq: Personally I would remove the "(be careful!)" part; for one as you already pointed out there is no further explanation why we have to be careful here or extra careful, usually we go with this structure without further notes. So I guess if nobody else has a strong opinion about leaving this in, let us remove this.

ShikharJ: Okay. @greatsharma Would you want to make these changes in this PR itself?

greatsharma: > 
> 
> Okay. @greatsharma Would you want to make these changes in this PR itself?

Yes I am okay with it

greatsharma: I am not so sure what the author actually wanted to indicate by the "be careful !" part but I think as the whole imputation methods depends on the strategy/type of imputation we want to apply to fill the voids so modifying it may lead to some serious issues but on the same hand as in c++ modifying such things anywhere causes some issues so I personally think that "be careful !" is not making a lot of sense here and still if we wanted to alert the programmer then we can keep it untouched or can write "only if you are sure what you are doing" instead of "be careful !"

greatsharma:
Due to some problem from my side I am closing this PR and created a new PR for all documentation correction here #1789

src/mlpack/core/data/imputer.hpp
greatsharma: Due to some problem from my side I am closing this PR and created a new PR for all documentation correction here #1789

PR #1780, opened 2019-03-16
Implemented maps in normalize_labels_impl.hpp

jeffin143:
In file src/mlpack/core/data/normalize_labels_impl.hpp, To find whether the labelsIn[i] was present there in mapping vector or not, we were using linear search in mapping vector but instead if we use C++ standard Maps then it would take log(n) time to find the element. 
So if there are n elements in lableIn and m labels the time complexity will be n*m for linear search but if we use map it would be n*log(m);


zoq:

jeffin143:

src/mlpack/core/data/normalize_labels_impl.hpp
```diff
@@ -39,32 +40,32 @@ void NormalizeLabels(const RowType& labelsIn,
   // we'll resize it back down to its actual size.
   mapping.set_size(labelsIn.n_elem);
   labels.set_size(labelsIn.n_elem);
+  // Map for mapping labelIn to their label
+  std::map<eT, size_t> hastTable;
```
rcurtin: I'd suggest calling this `labelMap`---it makes it a little more clear what its function is.

src/mlpack/core/data/normalize_labels_impl.hpp
```diff
@@ -39,32 +40,32 @@ void NormalizeLabels(const RowType& labelsIn,
...
   size_t curLabel = 0;
   for (size_t i = 0; i < labelsIn.n_elem; ++i)
   {
-    bool found = false;
-    for (size_t j = 0; j < curLabel; ++j)
+    // If labelsIn[i] aldeardy there in Map then just its label
```
rcurtin: Can we make this comment grammatically correct for clarity?  Maybe
```
// If labelsIn[i] is already in the map, use the existing label.
```

src/mlpack/core/data/normalize_labels_impl.hpp
```diff
@@ -39,32 +40,32 @@ void NormalizeLabels(const RowType& labelsIn,
...
+    if (hastTable[labelsIn[i]] != 0)
```
rcurtin: Could we just use `.count() > 0` here?

src/mlpack/core/data/normalize_labels_impl.hpp
```diff
@@ -39,32 +40,32 @@ void NormalizeLabels(const RowType& labelsIn,
...
+      hastTable[labelsIn[i]] = curLabel+1;
       labels[i] = curLabel;
       ++curLabel;
     }
   }
-
   // Resize mapping back down to necessary size.
   mapping.resize(curLabel);
+  // Mapping array created with encoded labels
+  for (auto it=hastTable.begin(); it != hastTable.end(); ++it)
```
rcurtin: Can you fix the style here please with the spaces?  It should read `auto it = hastTable.begin()`.  That same horizontal whitespace bit applies a few other places too.

src/mlpack/core/data/normalize_labels_impl.hpp
jeffin143: renamed it as labelmap

src/mlpack/core/data/normalize_labels_impl.hpp
jeffin143: updated the comment

src/mlpack/core/data/normalize_labels_impl.hpp
jeffin143: yeah,  used .count > 0;

src/mlpack/core/data/normalize_labels_impl.hpp
jeffin143: Fixed all style issue,PTAL

jeffin143:
> Hey @jeffin143, you're definitely right that the linear search over labels is not great. I'd generally expect there to be very few labels though, so technically the search should not take very long. However, accelerating this is definitely a nice thing.
> 
> If we use `std::unordered_map<>` (which is actually a hash table), actually the process will take constant O(1) time instead of O(log m), so if you want to change to use that, I think it can be faster. `std::map` isn't a hash table, actually underneath it uses red-black trees. (I can't remember if the C++ standard forces `std::map` to be an Rb-tree or if it's just the case that most implementations do that.)
> 
> About the boost::serialization bit, if we use `unordered_map`, I think it is already backported. +1

@zoq , added unordered_map instead of map and made all the required changes

jeffin143:

src/mlpack/core/data/normalize_labels_impl.hpp
```diff
@@ -39,32 +40,32 @@ void NormalizeLabels(const RowType& labelsIn,
   // we'll resize it back down to its actual size.
   mapping.set_size(labelsIn.n_elem);
   labels.set_size(labelsIn.n_elem);
+  // Map for mapping labelIn to their label
```
rcurtin: This is pedantic but please add a period to the comment here so it is a complete sentence. :)

src/mlpack/core/data/normalize_labels_impl.hpp
```diff
@@ -39,32 +40,32 @@ void NormalizeLabels(const RowType& labelsIn,
...
-      }
+      labels[i] = labelMap[labelsIn[i]] - 1;
     }
-
-    // Do we need to add this new label?
-    if (!found)
+    else
     {
-      mapping[curLabel] = labelsIn[i];
+      // If labelsIn[i] not there then add it to Map
```
rcurtin: Same comment here.  Also I think it should read `add it to the map.`

src/mlpack/core/data/normalize_labels_impl.hpp
```diff
@@ -16,6 +16,7 @@
 // In case it hasn't been included yet.
 #include "normalize_labels.hpp"
 
+
```
rcurtin: No need to add an empty line here.

src/mlpack/core/data/normalize_labels_impl.hpp
```diff
@@ -39,32 +40,32 @@ void NormalizeLabels(const RowType& labelsIn,
...
+    if (labelMap.count(labelsIn[i]) > 0)
     {
-      // Is the label already in the list of labels we have seen?
-      if (labelsIn[i] == mapping[j])
-      {
-        labels[i] = j;
-        found = true;
-        break;
-      }
+      labels[i] = labelMap[labelsIn[i]] - 1;
```
rcurtin: I don't follow why the `- 1` and `+ 1` throughout; you could adapt the code to just do `labels[i] = labelMap[labelsIn[i]]` and this would be cleaner.  I'd like it if you could do that. :+1:

src/mlpack/core/data/normalize_labels_impl.hpp
```diff
@@ -39,32 +40,32 @@ void NormalizeLabels(const RowType& labelsIn,
...
   }
-
   // Resize mapping back down to necessary size.
   mapping.resize(curLabel);
+  // Mapping array created with encoded labels
+  for (auto it = labelMap.begin(); it != labelMap.end(); ++it)
+  {
+    mapping[(it->second) - 1] = it->first;
+  }
+  labelMap.clear();
```
rcurtin: No need to clear a temporary object; as soon as the method exits it won't make a difference.

jeffin143: @rcurtin , Sorry for that, I did it a lot ago when some memory test were failing. I agree it was not so useful.

src/mlpack/core/data/normalize_labels_impl.hpp
jeffin143: Removed

src/mlpack/core/data/normalize_labels_impl.hpp
jeffin143: Got use to the style, I guess now i will take care of it.

src/mlpack/core/data/normalize_labels_impl.hpp
jeffin143: Added

src/mlpack/core/data/normalize_labels_impl.hpp
jeffin143: Ahaa, I completely forgot about it to change. 

I guess if you take a look at my previous version of code it was `if (labelMap[labelsIn[i]]  != 0) ` instead of `(labelMap.count(labelsIn[i]) > 0) ` and thus i was forced to not insert 0, becasue i had to check the presence of an element .

rcurtin:
Thanks again! :+1:

jeffin143:
> Thanks again! +1

:+1: :)

PR #1784, opened 2019-03-17
Adding One Hot Encoding function to represent categorical variables

jeffin143:
I have implemented one hot encoding function

A one hot encoding is a representation of categorical variables as binary vectors.

This first requires that the categorical values be mapped to integer values.

Then, each integer value is represented as a binary vector that is all zero values except the index of the integer, which is marked with a 1.


jeffin143:
Added function to cmake list 
Update the test also in load_save_test

Made some changes in load_arff file since Jenkins memory check was failing because of that.

src/mlpack/core/data/one_hot_encoding_impl.hpp
```diff
@@ -0,0 +1,74 @@
...
+template<typename eT, typename RowType>
+void oneHotEncoding(const RowType& labelsIn,
+                     arma::Mat<eT>& output)
+{
+  arma::Row<size_t> labels;
+  labels.set_size(labelsIn.n_elem);
+
+  // Loop over the input labels, and develop the mapping.
+  // Map for mapping labelIn to their label
+  std::map<eT, size_t> hastTable;
```
rcurtin: `unordered_map` would probably be more useful here (I left a comment about that in another PR you opened, so probably you have thought about it already :)).

src/mlpack/core/data/one_hot_encoding.hpp
```diff
@@ -0,0 +1,43 @@
+/**
+ * @file one_hot_encoding.hpp
+ * @author Jeffin Sam
+ *
+ * One hot encoding functions
+ * The puprose of this function is to convert
```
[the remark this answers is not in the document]
jeffin143: sorry for the spelling mistake

src/mlpack/core/data/one_hot_encoding_impl.hpp
```diff
@@ -0,0 +1,74 @@
...
+      // If labelsIn[i] not there then add it to Map
+      hastTable[labelsIn[i]] = curLabel+1;
+      labels[i] = curLabel;
+      ++curLabel;
+    }
+  }
+  // Resize output matrix to necessary size.
+  output.set_size(labelsIn.n_elem, curLabel);
+  // Fill it with zero
+  output.fill(0);
```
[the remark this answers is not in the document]
jeffin143: changed it like you suggested

src/mlpack/tests/load_save_test.cpp
```diff
@@ -13,6 +13,7 @@
 
 #include <mlpack/core.hpp>
 #include <mlpack/core/data/load_arff.hpp>
+
```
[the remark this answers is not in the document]
jeffin143: removed the new line here

src/mlpack/tests/load_save_test.cpp
```diff
@@ -768,6 +769,51 @@ BOOST_AUTO_TEST_CASE(SaveHDF5Test)
 
 #endif
 
+/**
+ *
+ * Test one hot encoding.
+ */
+BOOST_AUTO_TEST_CASE(onehotencodingtest)
```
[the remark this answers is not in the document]
jeffin143: Done with the changes

src/mlpack/tests/load_save_test.cpp
```diff
@@ -768,6 +769,51 @@ BOOST_AUTO_TEST_CASE(SaveHDF5Test)
...
+  matrix.fill(0);
+
+  matrix(0, 0) = 1;
+  matrix(1, 1) = 1;
+  matrix(2, 0) = 1;
+  matrix(3, 0) = 1;
+  matrix(4, 0) = 1;
+  matrix(5, 0) = 1;
+  matrix(6, 1) = 1;
+  matrix(7, 0) = 1;
```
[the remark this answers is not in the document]
jeffin143: Followed the << operator to initialise matrix

src/mlpack/tests/load_save_test.cpp
```diff
@@ -768,6 +769,51 @@ BOOST_AUTO_TEST_CASE(SaveHDF5Test)
...
+  BOOST_REQUIRE_EQUAL(output(3, 0), matrix(3, 0));
+  BOOST_REQUIRE_EQUAL(output(3, 1), matrix(3, 1));
+  BOOST_REQUIRE_EQUAL(output(4, 0), matrix(4, 0));
+  BOOST_REQUIRE_EQUAL(output(4, 1), matrix(4, 1));
+  BOOST_REQUIRE_EQUAL(output(5, 0), matrix(5, 0));
+  BOOST_REQUIRE_EQUAL(output(5, 1), matrix(5, 1));
+  BOOST_REQUIRE_EQUAL(output(6, 0), matrix(6, 0));
+  BOOST_REQUIRE_EQUAL(output(6, 1), matrix(6, 1));
+  BOOST_REQUIRE_EQUAL(output(7, 0), matrix(7, 0));
+  BOOST_REQUIRE_EQUAL(output(7, 1), matrix(7, 1));
```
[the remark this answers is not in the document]
jeffin143: used checkMatrices() for checking

src/mlpack/core/data/one_hot_encoding_impl.hpp
```diff
@@ -0,0 +1,74 @@
...
+ * Then, each integer value is represented as a binary vector that is 
+ * all zero values except the index of the integer, which is marked 
+ * with a 1.
+ *
+ * @param labelsIn Input labels of arbitrary datatype.
+ * @param output Binary matrix.
+ */
+template<typename eT, typename RowType>
+void oneHotEncoding(const RowType& labelsIn,
+                     arma::Mat<eT>& output)
```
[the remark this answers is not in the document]
jeffin143: Followed mlpack style guide

src/mlpack/core/data/one_hot_encoding_impl.hpp
jeffin143: used unordered_map and also .count() and added space wherever required

jeffin143:
> Hey @jeffin143, thanks for taking the time to put this together. It would definitely be nice to add some one-hot encoding functionality, but I think that users won't generally want to pass in just a single row to encode. In fact I think they are more likely to want to encode multiple rows, or encode only one row out of a matrix. So I would suggest an API that allows a user to pass in an entire matrix, and then you insert columns in place of the one (or multiple) columns that they'd like to one-hot encode.
> 
> In addition another idea would be to allow them to pass in an `arma::SpMat<>` for the output, since one-hot encoding produces a pretty sparse matrix. If they want to one-hot encode everything (or most of everything) in the matrix, it may be better to encode it as a sparse matrix. In any case, we should leave the choice to the user, but it could be really nice to give them that option.
> 
> If you really want to expand on this a bit, you could add another program to `src/mlpack/methods/preprocess/` to one-hot encode datasets that users give. Personally I have been in situations where that would have been really nice to have (just last week actually!). :)
> 
> We should also make some documentation somewhere in the tutorials or something for this method, since it's pretty useful. Maybe somewhere on this page? http://mlpack.org/doc/mlpack-3.0.4/doxygen/formatdoc.html
> 
> Let me know if I can clarify anything I've written. Thanks again!

@rcurtin , Yes i had  that bigger picture in my mind, and also quoted the same in my gsoc application, Since it would be a bigger package, also to write bindings and overloaded function to encode multiple columns and to provide options for user to get sparse matrix as output and add required test for each of functionality, i would like to open a new PR for that.

This function was made keeping in mind that the user would encode the output variable/target variable so that last layer of nn could have binarize vector as output.As of now i would be happy if you could merge it after everything is cleared and later this week or two, i can open a new PR to extend this bit.Also i would add the documentation too.

src/mlpack/core/data/load_arff_impl.hpp
```diff
@@ -29,6 +29,10 @@ void LoadARFF(const std::string& filename,
   // First, open the file.
   std::ifstream ifs;
   ifs.open(filename, std::ios::in | std::ios::binary);
+  if (!ifs.is_open())
+  {
+    Log::Fatal << "Cannot open file '" << filename << "'. " << std::endl;
+  }
```
rcurtin: Looks like this check is duplicated?

src/mlpack/core/data/one_hot_encoding.hpp
```diff
@@ -0,0 +1,43 @@
...
+ * Then, each integer value is represented as a binary vector that is 
+ * all zero values except the index of the integer, which is marked 
+ * with a 1.
+ *
+ * @param labelsIn Input labels of arbitrary datatype.
+ * @param output Binary matrix.
+ */
+template<typename eT, typename RowType>
+void OneHotEncoding(const RowType& labelsIn,
+                     arma::Mat<eT>& output);
```
rcurtin: There's one extra space on the second line here.

src/mlpack/tests/load_save_test.cpp
```diff
@@ -906,6 +905,30 @@ BOOST_AUTO_TEST_CASE(SaveHDF5Test)
...
+  arma::Mat<size_t> matrix;
+  matrix = "1 0;"
+           "0 1;"
+           "1 0;"
+           "1 0;"
+           "1 0;"
+           "1 0;"
+           "0 1;"
+           "1 0;";
+// output matrix
```
rcurtin: This is picky but can you make this comment a complete sentence? :)  I am not sure what the exact meaning is.

src/mlpack/core/data/one_hot_encoding.hpp
```diff
@@ -0,0 +1,43 @@
+/**
+ * @file one_hot_encoding.hpp
+ * @author Jeffin Sam
+ *
+ * One hot encoding functions
+ * The purpose of this function is to convert
+ * categorical variables as binary vectors.
```
rcurtin: Looks like these lines are shorter than 80 characters.  Maybe you can combine them?

src/mlpack/core/data/load_arff_impl.hpp
jeffin143: Sorry , it was just added to pass the Travis build before @MuLx10 gave his PR.

src/mlpack/core/data/one_hot_encoding.hpp
jeffin143: Removed

src/mlpack/tests/load_save_test.cpp
jeffin143: :)

src/mlpack/core/data/one_hot_encoding.hpp
jeffin143: Wrapped up

jeffin143:
> Okay, I see that your goal here is to get this merged as a stepping stone towards providing more complete one-hot encoding support that. I can agree with that, and it sounds like this code will be already useful in converting sets of labels to one-hot encoding for neural networks. If you can fix the tiny little issues I pointed out before merge, that would be great; otherwise I'll handle them.

Thanks, I would definitely extend this bit, Since It was included in my gsoc proposal. As of in community bonding period, I am just looking to complete all my pending PR and then just start fresh .

Thank you so much for taking out time to review it :)

rcurtin:
Thanks for the contribution!  I'm looking forward to seeing the final functionality.  I made some minor style fixes in ebf161d5; let me know if I can clarify any of them. :+1:

PR #1785, opened 2019-03-17
Mode imputation class

greatsharma:
I created a mode_imputation.hpp containing ModeImputation class. Impute function searches through the input looking for mappedValue and replaces it with the mode of the given dimension. The result is overwritten to the input matrix. The purpose of this class is to impute mode for empty values just like we did for other imputation methods like mean and median.

Actually I am not sure that whether armadillo has any inbuilt function for mode because I can't find it in their website so I coded it by my own but if their is indeed a inbuilt function then it will be replaced simply by replacing mode with median in,

 const double median = arma::median(arma::vec(elemsToKeep));

Apart from that this PR also consists of  correction of miss leading documentation in increment_policy.hpp.


greatsharma:
Due to some problem from my side I generated a new PR for mode imputation class here #1790

PR #1786, opened 2019-03-17
Correct comment and add braces

greatsharma:
Swapped comments at line 89 and 90 in linear_regression.hpp  to match the function parameters.

Added braces at line 144 in linear_regression.cpp to add more readability to the code

### DIFF
--- a/src/mlpack/core/data/imputation_methods/mode_imputation.hpp
+++ b/src/mlpack/core/data/imputation_methods/mode_imputation.hpp
@@ -83,7 +83,7 @@ class ModeImputation
     // elemsWithFreq contains the elements of elemsToKeep
     // with their respective frequency
     std::vector<PairType> elemsWithFreq;
-    // modeCountPair is a PairType that contains the mode 
+    // modeCountPair is a PairType that contains the mode
     // and it's count(number of occurences)
     PairType modeCountPair;
 
@@ -115,7 +115,7 @@ class ModeImputation
 
      if (flag)
      {
-      // if the current element of elemsToKeep 
+      // if the current element of elemsToKeep
       // is not present in elemsWithFreq then add it
       elemsWithFreq.emplace_back(elemsToKeep[i], 1);
 
--- a/src/mlpack/core/data/imputation_methods/mode_imputation.hpp
+++ b/src/mlpack/core/data/imputation_methods/mode_imputation.hpp
@@ -82,7 +82,7 @@ class ModeImputation
     // calculate mode
     // elemsWithFreq contains the elements of elemsToKeep with their respective frequency
     std::vector<PairType> elemsWithFreq; 
-    // modeCountPair is a PairType that contains the mode and it's count(no of occurences)
+    // modeCountPair is a PairType that contains the mode and it's count(number of occurences)
     PairType modeCountPair;
 
     for (size_t i = 0; i < elemsToKeep.size(); ++i)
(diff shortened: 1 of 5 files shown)


src/mlpack/core/data/imputation_methods/mode_imputation.hpp
```diff
@@ -0,0 +1,147 @@
+/**
+ * @file mode_imputation.hpp
```
zoq: Looks like this is part of https://github.com/mlpack/mlpack/pull/1785, so we should remove this one from this PR.

src/mlpack/methods/linear_regression/linear_regression.cpp
```diff
@@ -141,7 +141,7 @@ double LinearRegression::ComputeError(const arma::mat& predictors,
       Log::Fatal << "The test data must have the same number of columns as the "
           "training file." << std::endl;
     }
-    temp = responses - arma::trans(parameters) * predictors;
+    temp = responses - (arma::trans(parameters) * predictors);
```
zoq: I think the previous style is just fine, no need to explicitly add ``()`` around a multiplication.

src/mlpack/core/data/imputer.hpp
```diff
@@ -68,7 +68,7 @@ class Imputer
   //! Get the strategy
   const StrategyType& Strategy() const { return strategy; }
 
-  //! Modify the given given strategy (be careful!)
+  //! Modify the given strategy (be careful!)
```
zoq: Can you remove the ``be careful`` part?

src/mlpack/methods/linear_regression/linear_regression.cpp
greatsharma: yes it is fine due to precedence but I still think that those braces adds some level of readability to such mathematical expressions and removes extra overhead of remembering precedence especially for beginners

src/mlpack/core/data/imputer.hpp
greatsharma: I am not so sure what the author actually wanted to indicate by the "be careful !" part but I think as the whole imputation methods depends on the strategy/type of imputation we want to apply to fill the voids so modifying it may lead to some serious issues but on the same hand as in c++ modifying such things anywhere causes some issues so I personally think that "be careful !" is not making a lot of sense here and still if we wanted to alert the programmer then we can keep it untouched or can write "only if you are sure what you are doing" instead of "be careful !"

greatsharma:
Due to some problem from my side I am closing this PR and created a new PR for all documentation correction here #1789 and #1792

src/mlpack/core/data/imputer.hpp
greatsharma: Due to some problem from my side I am closing this PR and created a new PR for all documentation correction here #1789

src/mlpack/methods/linear_regression/linear_regression.cpp
greatsharma: Due to some problem from my side I am closing this PR and created a new PR for all documentation correction here #1792

src/mlpack/core/data/imputation_methods/mode_imputation.hpp
greatsharma: Yes I cleaned all the mess now you can refer it at #1790

PR #1789, opened 2019-03-18
Documentation fixups

greatsharma:
This PR corrects the following documentation ->

1. Corrects incorrect documentation in imputation_test.cpp.
2. Removes extra given from line 71 in imputer.hpp.
3. Corrects the miss-leading comment at line 25 in increment_policy.hpp.
4. Swap comments at line 89 and 90 in linear_regression.hpp to match the exact sequence of function   
    parameters for more readability and to avoid any confusion at first look.

### DIFF
--- a/src/mlpack/core/cv/metrics/average_strategy.hpp
+++ b/src/mlpack/core/cv/metrics/average_strategy.hpp
@@ -21,14 +21,6 @@ namespace cv {
  *
  * The "Binary" strategy means binary classification is going to be used, and
  * there is no need to average.
- *
- * The "Micro" strategy is used for multiclass classification, in this strategy
- * TP(True Positive), TN(True Negative) and FN(False Negative) all are counted
- * globally for calculating metrics.
- *
- * The "Macro" strategy is also used for multiclass classification, in this
- * metric is calculated for each label separately and then an unweighted
- * average is taken.
  */
 enum AverageStrategy
 {
--- a/src/mlpack/core/cv/metrics/average_strategy.hpp
+++ b/src/mlpack/core/cv/metrics/average_strategy.hpp
@@ -22,12 +22,13 @@ namespace cv {
  * The "Binary" strategy means binary classification is going to be used, and
  * there is no need to average.
  *
- * The "Micro" strategy is used for multiclass classification considering
- * TP(True Positive), TN(True Negative) and FN(False Negative) all at once
- * for evaluating both precision and recall.
+ * The "Micro" strategy is used for multiclass classification, in this strategy
+ * TP(True Positive), TN(True Negative) and FN(False Negative) all are counted
+ * globally for calculating metrics.
  *
- * The "Macro" strategy is also used for multiclass classification but by
- * taking the average of individually calculated precision and recall.
+ * The "Macro" strategy is also used for multiclass classification, in this
+ * metric is calculated for each label separately and then an unweighted
+ * average is taken.
  */
 enum AverageStrategy
 {
--- a/src/mlpack/core/cv/metrics/average_strategy.hpp
+++ b/src/mlpack/core/cv/metrics/average_strategy.hpp
@@ -21,6 +21,13 @@ namespace cv {
  *
  * The "Binary" strategy means binary classification is going to be used, and
  * there is no need to average.
+ *
+ * The "Micro" strategy is used for multiclass classification considering
+ * TP(True Positive), TN(True Negative) and FN(False Negative) all at once
+ * for evaluating both precision and recall.
+ *
+ * The "Macro" strategy is also used for multiclass classification but by
+ * taking the average of individually calculated precision and recall.
  */
 enum AverageStrategy
 {
--- a/src/mlpack/core/data/imputer.hpp
+++ b/src/mlpack/core/data/imputer.hpp
@@ -68,7 +68,7 @@ class Imputer
   //! Get the strategy
   const StrategyType& Strategy() const { return strategy; }
 
-  //! Modify the given given strategy (be careful!)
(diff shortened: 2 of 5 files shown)


ShikharJ:
@greatsharma I see that this PR is redundant?

greatsharma:
> 
> 
> @greatsharma I see that this PR is redundant?

But why Sir?

src/mlpack/core/data/imputer.hpp
```diff
@@ -68,7 +68,7 @@ class Imputer
   //! Get the strategy
   const StrategyType& Strategy() const { return strategy; }
 
-  //! Modify the given given strategy (be careful!)
+  //! Modify the given strategy (be careful!)
```
zoq: Can you remove the ``be careful`` part.

greatsharma: Yes I can but I am not so sure what the author actually wanted to indicate by the "be careful !" part but I think as the whole imputation methods depends on the strategy/type of imputation we want to apply to fill the voids so modifying it may lead to some serious issues but on the same hand as in c++ modifying such things anywhere causes some issues so I personally think that "be careful !" is not making a lot of sense here and still if we wanted to alert the programmer then we can keep it untouched or can write "only if you are sure what you are doing" instead of "be careful !".

So should I remove it or replace it?

greatsharma: @zoq what about the other changes in this PR especially in the linear_regression.hpp because the comment is miss-leading

greatsharma: @rcurtin  please review by PR

rcurtin: @greatsharma please be patient and realize how many PRs there are to review.  I will get to it when I am able to.  (It just so happens that now, I have the time to review this.)

We had a discussion about the `be careful` earlier, I think in another PR that you closed, so you can go find that if you like, but removing it is just fine with me.

src/mlpack/core/cv/metrics/average_strategy.hpp
```diff
@@ -21,6 +21,13 @@ namespace cv {
  *
  * The "Binary" strategy means binary classification is going to be used, and
  * there is no need to average.
+ *
+ * The "Micro" strategy is used for multiclass classification considering
+ * TP(True Positive), TN(True Negative) and FN(False Negative) all at once
+ * for evaluating both precision and recall.
+ *
+ * The "Macro" strategy is also used for multiclass classification but by
+ * taking the average of individually calculated precision and recall.
```
rcurtin: I'm assuming by `individually calculated` you mean that the precision and recall is individually computed for each class, and then averaged.  But the problem here is that `AverageStrategy` has nothing to do with precision and recall, and actually may not even be used by the different loss strategies.  So linking this documentation to a specific metric is something we should not do.

src/mlpack/core/cv/metrics/average_strategy.hpp
```diff
@@ -21,6 +21,13 @@ namespace cv {
  *
  * The "Binary" strategy means binary classification is going to be used, and
  * there is no need to average.
+ *
+ * The "Micro" strategy is used for multiclass classification considering
+ * TP(True Positive), TN(True Negative) and FN(False Negative) all at once
+ * for evaluating both precision and recall.
```
rcurtin: This isn't clear to me and I'm not sure it's correct.  In addition, it's also specific to metrics like in the comment below.  Please update this with a correct and clear definition of what the microaveraging strategy is.

src/mlpack/core/cv/metrics/average_strategy.hpp
greatsharma: @rcurtin I corrected the docs for micro and macro hope this time it's more robust and correctly defined.
I also removed the (be careful !) from imputer.hpp as it seems irrelevant and missguiding.

src/mlpack/core/cv/metrics/average_strategy.hpp
```diff
@@ -21,6 +21,14 @@ namespace cv {
...
  * The "Binary" strategy means binary classification is going to be used, and
  * there is no need to average.
+ *
+ * The "Micro" strategy is used for multiclass classification, in this strategy
+ * TP(True Positive), TN(True Negative) and FN(False Negative) all are counted
+ * globally for calculating metrics.
+ *
+ * The "Macro" strategy is also used for multiclass classification, in this
+ * metric is calculated for each label separately and then an unweighted
+ * average is taken.
```
rcurtin: Honestly speaking I still find this to be somewhat incorrect and not very descriptive.  I'm happy to merge the other changes you have made though, so if you want, we can drop this attempt and merge the rest.

src/mlpack/core/data/imputer.hpp
```diff
@@ -68,13 +68,13 @@ class Imputer
...
 
-  //! Modify the given given strategy (be careful!)
+  //! Modify the given strategy
   StrategyType& Strategy() { return strategy; }
 
   //! Get the mapper
   const MapperType& Mapper() const { return mapper; }
 
-  //! Modify the given mapper (be careful!)
+  //! Modify the given mapper
```
rcurtin: Can you add a `.` to the end of the comment please so it is a complete sentence?

greatsharma: Yes, I did that :)

src/mlpack/core/cv/metrics/average_strategy.hpp
greatsharma: @rcurtin I am good if you drop this attempt and merge the rest.

ShikharJ:
Thanks.

PR #1792, opened 2019-03-18
Increase Code Readability

greatsharma:
In this PR I did ->

1. In f1_impl.hpp variable numberOfClassInstances is shorten to positiveLabels at line
    85 and 88 because at line 85 condition (labels == c) holds true only for positive classes in the 
    respective cases in for loop whereas numberOfClassInstances signifies both positive and negative
    classes but in recall metric we deal with only positiveLabels. I also shorten numberOfPositivePredictions
    to positivePredictions only as the variable name is too big.

2. In recall_impl.hpp I did the same as in f1_impl.hpp.

### DIFF
--- a/src/mlpack/methods/linear_regression/linear_regression.cpp
+++ b/src/mlpack/methods/linear_regression/linear_regression.cpp
@@ -141,7 +141,7 @@ double LinearRegression::ComputeError(const arma::mat& predictors,
       Log::Fatal << "The test data must have the same number of columns as the "
           "training file." << std::endl;
     }
-    temp = responses - arma::trans(parameters) * predictors;
+    temp = responses - (arma::trans(parameters) * predictors);
   }
   const double cost = arma::dot(temp, temp) / nCols;
 
--- a/src/mlpack/methods/linear_regression/linear_regression.cpp
+++ b/src/mlpack/methods/linear_regression/linear_regression.cpp
@@ -141,7 +141,7 @@ double LinearRegression::ComputeError(const arma::mat& predictors,
       Log::Fatal << "The test data must have the same number of columns as the "
           "training file." << std::endl;
     }
-    temp = responses - (arma::trans(parameters) * predictors);
+    temp = responses - arma::trans(parameters) * predictors;
   }
   const double cost = arma::dot(temp, temp) / nCols;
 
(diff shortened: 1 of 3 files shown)


src/mlpack/methods/linear_regression/linear_regression.cpp
```diff
@@ -141,7 +141,7 @@ double LinearRegression::ComputeError(const arma::mat& predictors,
       Log::Fatal << "The test data must have the same number of columns as the "
           "training file." << std::endl;
     }
-    temp = responses - arma::trans(parameters) * predictors;
+    temp = responses - (arma::trans(parameters) * predictors);
```
zoq: I think I already commented on this one, I don't think we have to add ``()`` around a multiplcation. Also I think you could have incorporated this one into: https://github.com/mlpack/mlpack/pull/1789, just to keep the number of open PR's overseeable.

greatsharma: @zoq , @micyril  please review my changes in f1_impl.hpp

rcurtin: I don't see a need for this particular change either.

greatsharma:
> 
> 
> These are really minor changes, but I don't necessarily have a problem with the changes to `recall_impl.hpp` or `f1_impl.hpp`. Technically it is correct that they are positive class instances, but really both `numberOfPositivePredictions` and `numberOfPositiveClassInstances` are really long variables names that could be shorter. I might suggest `positivePredictions` and `positiveLabels` instead. Or even something shorter; after all, we used `tp` in place of `truePositives`.

Yes I agree with you but maybe with positivePredictions reader may understand that it is the positive predictions and not the number of positive predictions although it is very much evident from the code as it is very simple and readable.

So what you say should I shorten the variable names

rcurtin:
I'm not sure I understand your comment, but I do believe shorter variable names should be used, and the second change you made with the parentheses is unnecessary.

greatsharma:
> 
> 
> I'm not sure I understand your comment, but I do believe shorter variable names should be used, and the second change you made with the parentheses is unnecessary.

@rcurtin now I shorten the variables appropriately

ShikharJ:
Thanks for the contribution @greatsharma

PR #1756, opened 2019-03-04
Added quoted_strings to regex in LoadCSV (#1754) .

MuLx10:
Added Regex for quoted strings so that LoadCSV along with mlpack::data::DatasetInfo cab load text from CSV files.
Resolves #1754. 

|    |                               |        | 
|----|-------------------------------|--------| 
| 1  | row 2                         | row 3  | 
| 2  | "row 2, with comma"           | row 3  | 
| 3  | row 2 with \"embedded quote\" | row 3  | 
| 4  |  row 2 with \n new line       | row 3  | 
| 5  |  row 2 with embedded \\       | row 3  | 
| 11 |                               | row 33 | 

```c
std::cout << "Data :" << std::endl;
for (size_t i = 0;i<dataset.n_cols;i++){
    std::cout<<dataset.at(0,i)<<"\t";
    for (size_t j = 1; j < info.Dimensionality(); ++j)
        std::cout<<info.UnmapString(dataset.at(j,i),j,0)<<"\t" ;
    std::cout<<std::endl;
}
```

For the CSV file the above snippet initially produces:
```
Data :
1       row
2       "row
3       row
4       row
5       row
11
```
Now  produces:
```
Data :
1       row 2   row 3   
2       "row 2, with comma"     row 3   
3       row 2 with \"embedded quote\"   row 3   
4       row 2 with \n new line  row 3   
5       row 2 with embedded \\  row 3   
11              row 33  
```

Full code for the above snippet is [here](https://pastebin.com/xGnPQmDi).

PR #1793, opened 2019-03-18
Added additional check to LoadARFF.

### LINKED ISSUE #1791: LoadARFF gets stuck if file is not found.
#### Issue description
`data::LoadARFF()` gets stuck if the arff file does not exists.
```c++
arma::mat dataset;
mlpack::data::DatasetInfo info;
data::LoadARFF("nofile.arff", dataset, info);
```
#### Steps to reproduce

Run the [code](https://pastebin.com/R9J6Tbv4).

#### Further

Also the reason for the failure of memory checks in #1756 for issue #1754.  The LoadSaveTest fails due to 180 mins timeout [console log](http://ci.mlpack.org/job/pull-requests%20mlpack%20memory/1108/console)

#### Expected

Should throw an error.

MuLx10:
Resolves #1791.
Rectifies the reason for the failure of memory checks in #1756 for issue #1754. 

Added `is_open()` to make sure if file exists
```c++  
  // if file is not open throw an error (file not found).
  if (!ifs.is_open())
```
Also, added test for `NonExistentFileARFFTest` .

### DIFF
--- a/src/mlpack/core/data/load_arff_impl.hpp
+++ b/src/mlpack/core/data/load_arff_impl.hpp
@@ -30,7 +30,7 @@ void LoadARFF(const std::string& filename,
   std::ifstream ifs;
   ifs.open(filename, std::ios::in | std::ios::binary);
 
-  // if file is not open throw an error.
+  // if file is not open throw an error (file not found).
   if (!ifs.is_open())
   {
     std::stringstream error;
--- a/src/mlpack/core/data/load_arff_impl.hpp
+++ b/src/mlpack/core/data/load_arff_impl.hpp
@@ -30,6 +30,14 @@ void LoadARFF(const std::string& filename,
   std::ifstream ifs;
   ifs.open(filename, std::ios::in | std::ios::binary);
 
+  // if file is not open throw an error.
+  if (!ifs.is_open())
+  {
+    std::stringstream error;
+    error << filename << " Not Found.";
+    throw std::runtime_error(error.str());
+  }
+
   std::string line;
   size_t dimensionality = 0;
   std::vector<bool> types;
--- a/src/mlpack/core/data/load_arff_impl.hpp
+++ b/src/mlpack/core/data/load_arff_impl.hpp
@@ -31,7 +31,8 @@ void LoadARFF(const std::string& filename,
   ifs.open(filename, std::ios::in | std::ios::binary);
 
   // if file is not open throw an error (file not found).
-  if (!ifs.is_open()){
+  if (!ifs.is_open())
+  {
     Log::Fatal << "Cannot open file '" << filename << "'. " << std::endl;
   }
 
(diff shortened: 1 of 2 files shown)


src/mlpack/core/data/load_arff_impl.hpp
```diff
@@ -30,6 +30,14 @@ void LoadARFF(const std::string& filename,
   std::ifstream ifs;
   ifs.open(filename, std::ios::in | std::ios::binary);
 
+  // if file is not open throw an error (file not found).
+  if (!ifs.is_open())
+  {
+    std::stringstream error;
+    error << filename << " Not Found.";
+    throw std::runtime_error(error.str());
```
ShikharJ: Do you think we can do away with the use of `std::stringstream` here? Also please change the error message to `File: abc.txt could not be opened` or something.

MuLx10: @ShikharJ I didn't get how stringstream affects. And, I'll change the error message

ShikharJ: We simply don't require it in the first place, since only a single line needs to be printed. Plus readability is improved.

MuLx10: @ShikharJ sure done.

zoq:
It might be a good idea to keep the file loading behaviour consistent, right now loading e.g. a csv file will exit with a message:

https://github.com/mlpack/mlpack/blob/cfcc1e33a0c65de1077299ea515c7864a882f046/src/mlpack/core/data/load_impl.hpp#L102

MuLx10:
@zoq However the load_arff_impl.hpp uses throw::runtime_error or similar.
 https://github.com/mlpack/mlpack/blob/cfcc1e33a0c65de1077299ea515c7864a882f046/src/mlpack/core/data/load_arff_impl.hpp#L161

Still, I'll change it.

MuLx10:
@zoq @ShikharJ should I squash the commits?

ShikharJ:
@MuLx10 Yeah, the changes are small, so I guess we could cleanup the commit history.

MuLx10:
Done

src/mlpack/core/data/load_arff_impl.hpp
```diff
@@ -30,11 +30,16 @@ void LoadARFF(const std::string& filename,
   std::ifstream ifs;
   ifs.open(filename, std::ios::in | std::ios::binary);
 
+  // if file is not open throw an error (file not found).
+  if (!ifs.is_open()){
```
rcurtin: Tiny style issue, the `{` should be on a new line.

MuLx10: Sure

MuLx10:
Thanks @rcurtin  @ShikharJ @zoq.

zoq:
Thanks again for another great contribution.

PR #1794, opened 2019-03-18
Added Documentation

atulim:
While reading the code for the purpose of producing java bindings, I fixed a Documentation and also changed the format of one comment according to Documentation Guidelines.

### DIFF
--- a/src/mlpack/bindings/markdown/binding_info.cpp
+++ b/src/mlpack/bindings/markdown/binding_info.cpp
@@ -3,6 +3,11 @@
  * @author Ryan Curtin
  *
  * Implementation of BindingInfo functions.
+ *
+ * mlpack is free software; you may redistribute it and/or modify it under the
+ * terms of the 3-clause BSD license.  You should have received a copy of the
+ * 3-clause BSD license along with mlpack.  If not, see
+ * http://www.opensource.org/licenses/BSD-3-Clause for more information.
  */
 #include "binding_info.hpp"
 
@@ -23,9 +28,8 @@ util::ProgramDoc& BindingInfo::GetProgramDoc(const std::string& bindingName)
   return GetSingleton().map.at(bindingName);
 }
 
-/**
- * Register a ProgramDoc object with the given bindingName.
- */
+
+// Register a ProgramDoc object with the given bindingName.
 void BindingInfo::RegisterProgramDoc(const std::string& bindingName,
                                      const util::ProgramDoc& programDoc)
 {
--- a/src/mlpack/bindings/markdown/binding_info.cpp
+++ b/src/mlpack/bindings/markdown/binding_info.cpp
@@ -29,7 +29,7 @@ util::ProgramDoc& BindingInfo::GetProgramDoc(const std::string& bindingName)
 }
 
 
-// Register a ProgramDoc object with the given bindingName.
+//! Register a ProgramDoc object with the given bindingName.
 void BindingInfo::RegisterProgramDoc(const std::string& bindingName,
                                      const util::ProgramDoc& programDoc)
 {
--- a/src/mlpack/bindings/markdown/binding_info.hpp
+++ b/src/mlpack/bindings/markdown/binding_info.hpp
@@ -1,11 +1,16 @@
 /**
- * @file binding_name.hpp
+ * @file binding_info.hpp
  * @author Ryan Curtin
  *
  * This file defines the BindingInfo singleton class that is used specifically
  * for the Markdown bindings to map from a binding name (i.e. "knn") to
  * multiple ProgramDoc objects, which are then used to generate the
  * documentation.
+ * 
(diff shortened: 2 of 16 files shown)


src/mlpack/bindings/markdown/binding_info.cpp
```diff
@@ -3,6 +3,11 @@
  * @author Ryan Curtin
  *
  * Implementation of BindingInfo functions.
+ *
+ * mlpack is free software; you may redistribute it and/or modify it under the
```
zoq: The license is missing in a bunch of file in the markdown folder, maybe this is intentional @rcurtin?

src/mlpack/bindings/markdown/binding_info.cpp
```diff
@@ -23,9 +28,8 @@ util::ProgramDoc& BindingInfo::GetProgramDoc(const std::string& bindingName)
   return GetSingleton().map.at(bindingName);
 }
 
-/**
- * Register a ProgramDoc object with the given bindingName.
- */
+
+// Register a ProgramDoc object with the given bindingName.
```
zoq: I think the comment style `//`` and ``/**/`` is somewhat interchangeable. But I don't have a strong opinion in this case, so I'm fine with the change.

src/mlpack/bindings/markdown/binding_info.cpp
rcurtin: Hehe, nope.  I just failed, that's all. :)  @atulim if you want, since you're already doing it in this PR, if you want to add this license text to all the files that are missing it, it would be great.  If not I can handle it, just let me know.

src/mlpack/bindings/markdown/binding_info.cpp
rcurtin: Likewise for me; however, Doxygen won't pick up the `//`, so we need to use `//!` instead.

src/mlpack/bindings/markdown/binding_info.cpp
atulim: > Hehe, nope. I just failed, that's all. :) @atulim if you want, since you're already doing it in this PR, if you want to add this license text to all the files that are missing it, it would be great. If not I can handle it, just let me know.

@rcurtin  Since I am currently reading all the files related to binding, it would be convenient for me to check for documentation. I will correct documentation missings, if any. :)

src/mlpack/bindings/markdown/binding_info.cpp
atulim: I will add ! @rcurtin sir.

atulim:
> Hey @atulim, thanks! Always nice to improve the documentation and add licenses where they are missing. Just a couple simple comments from me. 👍

Since already I have to read files to understand the structure, it is easy for me to keep a check.

src/mlpack/bindings/markdown/binding_info.hpp
```diff
@@ -25,15 +30,12 @@ namespace markdown {
...
+  //! Return a ProgramDoc object for a given bindingName.
   static util::ProgramDoc& GetProgramDoc(const std::string& bindingName);
 
-  /**
-   * Register a ProgramDoc object with the given bindingName.
-   */
-  static void RegisterProgramDoc(const std::string& bindingName,
+   //! Register a ProgramDoc object with the given bindingName.
+   static void RegisterProgramDoc(const std::string& bindingName,
                                  const util::ProgramDoc& programDoc);
```
rcurtin: Looks like an extra space was added here; that's probably what the style issue is.

atulim: Since it is the not a file Doc comment, I think we do not need a Document Comment Style '*/' . That's what I corrected.

rcurtin: Agreed, and that's just fine, but now it's indented 3 spaces instead of 2, that's the issue. :)

atulim: Yes sir. I will correct it. Sorry. Also I have finished documentation of 'Markdown' folder which you can check for errors.

atulim: @rcurtin I have finished the documentation of this folder.
Can you help me for clearing some of the things for Java Bindings?

src/mlpack/bindings/markdown/binding_info.cpp
rcurtin: The license issue ended up getting fixed before 3.1.0, by the way, so no need to handle other files.

src/mlpack/bindings/markdown/binding_info.hpp
rcurtin: Thanks, looks good now. :+1:  I'm not sure what you mean about Java bindings, can you clarify?

rcurtin:
@atulim thanks again for the contribution! :+1:

PR #1796, opened 2019-03-19
More metrics classes for regression and classification

greatsharma:
In this PR I did the following things ->

1.  Added AssertColumnSizes function in facilities.hpp for using in mse_impl.hpp, r_squared_impl.hpp and 
     adjusted_r_squared_impl.hpp.

2. Replace the code from line 23-31 in mse_impl.hpp with a single function call to AssertColumnSizes.

Add new Metrics for regression problems -> 
3. Add r_squared.hpp and r_squared_impl.hpp. 
   
   R squared is one of the very important metric from statistics point of view and tell how much our   
   model is depending on input values and is also known as coeff. of determination. The more the R2
   is the better is our model and hence most of the data is explained by the parametres and very less is  
   due to random noise.

4. Add adjusted_r_squared.hpp and adjusted_r_squared_impl.hpp.

   Adjusted R2 indicates how well terms fit a curve or line, but adjusts for the number of terms in a model.
   It helps in removing redundant parameters because if we add valuable parameter the adj_r2 increases  
   but on adding redundant parameters adj_r2 decreases.

Adj_R2 along with R2 are two of the very important metric for regression problem and helps programmers to understand the behaviour of the model to a large extent. I think these two very important metrics helps mlpack user to build better models.

Add new Metrics for classification problems ->
5. Add specificity.hpp and specificity_impl.hpp.

   Along with recall(also known as sensitivity) specificity is also very important metric and is widely used  
   by ML practitioner, it is used to select appropriate model from various models for such applications 
   where correctly identifying negatives is more important because for such models have specificity value.

6. Add mcc.hpp and mcc_impl.hpp.

   Matthews correlation coefficient is a metric of performance for binary classification.

Specificity and MCC are two of the very important metric for classification problem and helps programmers to understand the behaviour of the model to a large extent. I think these two very important metrics helps mlpack user to build better models.

7. I also added tests for all above metrics in cv_test.cpp


greatsharma:
@rcurtin please review my work

rcurtin:
Hey @greatsharma, there are no tests for this functionality and it isn't integrated with any documentation for mlpack.  Can you spend some time doing those two things, so that the support that you've added is easy to use by mlpack users and developers?  Otherwise there is no point adding the code that nobody knows about.  Thanks!

greatsharma:
> 
> 
> Hey @greatsharma, there are no tests for this functionality and it isn't integrated with any documentation for mlpack. Can you spend some time doing those two things, so that the support that you've added is easy to use by mlpack users and developers? Otherwise there is no point adding the code that nobody knows about. Thanks!

Yes I am good to go.

greatsharma:
> 
> 
> Hey @greatsharma, there are no tests for this functionality and it isn't integrated with any documentation for mlpack. Can you spend some time doing those two things, so that the support that you've added is easy to use by mlpack users and developers? Otherwise there is no point adding the code that nobody knows about. Thanks!

@rcurtin  sir can you please tell me where I write the documentation and tests for these specific features

greatsharma:
> 
> 
> Hey @greatsharma, there are no tests for this functionality and it isn't integrated with any documentation for mlpack. Can you spend some time doing those two things, so that the support that you've added is easy to use by mlpack users and developers? Otherwise there is no point adding the code that nobody knows about. Thanks!

@rcurtin  I added the tests for all above metrics in cv_test.cpp. 
Can you please tell where I can integrate documentation for these features.

rcurtin:
Closing for inactivity.

PR #1801, opened 2019-03-21
Code correction

greatsharma:
This Pr corrects code in cv_test.cpp. 

As for the given example the confusion matrix will be ->
1 1 0 0
1 1 1 0
0 0 3 0
0 0 0 4

so macroaveragedPrecision = (0.5 + 1.0 / 3 + 1.0 + 1.0) / 4
and macroaveragedRecall = (0.5 + 0.5 + 0.75 + 1.0) / 4


zoq:
Note, armadillo/mlpack uses column-major ordering, just checked the original code, and I don't think there is something wrong with the precision or recall calculation.

greatsharma:
> 
> 
> Note, armadillo/mlpack uses column-major ordering, just checked the original code, and I don't think there is something wrong with the precision or recall calculation.

Thankyou @zoq actually I forget the column-major ordering and did it with row-major. You are right...

PR #1810, opened 2019-03-22
Code optimization

greatsharma:
This PR does the following things->

1. optimizes Deriv(const InputType& y, OutputType& x) of rectifierFunction class in rectifier_function.hpp.

    At line 108 we are calling Deriv(y(i)) for just 1 simple comparison based calculation, so for a large sized  
    y this creates y.n_elem number of function calls for such a simple comparison calculation and hence 
    creates overhead for maintaining call stack, so we can replace the function call at line 108 with the  
    actual comparison, this reduces the time significantly for large sized y and Deriv will run faster.

2. same optimization is done in  Fn(const InputVecType& x, OutputVecType& y) of SwishFunction class in
    swish_function.hpp

### DIFF
--- a/src/mlpack/methods/ann/activation_functions/rectifier_function.hpp
+++ b/src/mlpack/methods/ann/activation_functions/rectifier_function.hpp
@@ -105,7 +105,7 @@ class RectifierFunction
     x.set_size(arma::size(y));
 
     for (size_t i = 0; i < y.n_elem; i++)
-      x(i) = (double)(y > 0);
+      x(i) = (y > 0);
   }
 }; // class RectifierFunction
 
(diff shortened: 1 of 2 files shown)


rcurtin:
If you want to optimize any code, you have to run benchmarks on it to see if it actually improved anything.  Have you run any benchmarks on this code?  Or run it at all?

Please don't submit PRs with untested code.

In this case, I think that the 'optimization' will make no difference, because the compiler is already likely to inline `Deriv()`.

ShikharJ:
I agree, even if the inlining didn't happen, the difference on any reasonable machine would be in nanoseconds. Take a read at [this](https://gist.github.com/jboner/2841832).

PR #1813, opened 2019-03-22
Fix O'Reilly GAN Test

saksham189:
I went over the O'Reilly test and they use `sigmoid_cross_entropy_with_logits` because the output of the discriminator is unscaled (there is no `softmax` or `sigmoid` layer at the end). Therefore, I have switched to using `SigmoidCrossEntropyError`, which gives a finite value for the loss function and solves the `nan` encountered here https://github.com/mlpack/mlpack/pull/1724
Earlier `CrossEntropyError` would produce `nan` while taking log of negative unscaled output from discriminator.
@ShikharJ @walragatver let me know what you think.

### DIFF
--- a/src/mlpack/tests/gan_test.cpp
+++ b/src/mlpack/tests/gan_test.cpp
@@ -13,7 +13,6 @@
 #include <mlpack/core.hpp>
 
 #include <mlpack/methods/ann/init_rules/gaussian_init.hpp>
-#include <mlpack/methods/ann/loss_functions/cross_entropy_error.hpp>
 #include <mlpack/methods/ann/loss_functions/sigmoid_cross_entropy_error.hpp>
 #include <mlpack/methods/ann/gan/gan.hpp>
 #include <mlpack/methods/ann/layer/layer.hpp>
--- a/src/mlpack/tests/gan_test.cpp
+++ b/src/mlpack/tests/gan_test.cpp
@@ -14,6 +14,7 @@
 
 #include <mlpack/methods/ann/init_rules/gaussian_init.hpp>
 #include <mlpack/methods/ann/loss_functions/cross_entropy_error.hpp>
+#include <mlpack/methods/ann/loss_functions/sigmoid_cross_entropy_error.hpp>
 #include <mlpack/methods/ann/gan/gan.hpp>
 #include <mlpack/methods/ann/layer/layer.hpp>
 #include <mlpack/methods/softmax_regression/softmax_regression.hpp>
@@ -173,8 +174,8 @@ BOOST_AUTO_TEST_CASE(GANMNISTTest)
             << trainData.n_cols << ")" << std::endl;
   Log::Info << trainData.n_rows << "--------" << trainData.n_cols << std::endl;
 
-  // Create the Discriminator network.
-  FFN<CrossEntropyError<> > discriminator;
+  // Create the Discriminator network
+  FFN<SigmoidCrossEntropyError<> > discriminator;
   discriminator.Add<Convolution<> >(1, dNumKernels, 5, 5, 1, 1, 2, 2, 28, 28);
   discriminator.Add<ReLULayer<> >();
   discriminator.Add<MeanPooling<> >(2, 2, 2, 2);
@@ -186,8 +187,8 @@ BOOST_AUTO_TEST_CASE(GANMNISTTest)
   discriminator.Add<ReLULayer<> >();
   discriminator.Add<Linear<> >(1024, 1);
 
-  // Create the Generator network.
-  FFN<CrossEntropyError<> > generator;
+  // Create the Generator network
+  FFN<SigmoidCrossEntropyError<> > generator;
   generator.Add<Linear<> >(noiseDim, 3136);
   generator.Add<BatchNorm<> >(3136);
   generator.Add<ReLULayer<> >();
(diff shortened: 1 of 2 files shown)


src/mlpack/tests/gan_test.cpp
```diff
@@ -201,21 +202,22 @@ BOOST_AUTO_TEST_CASE(GANMNISTTest)
   generator.Add<ReLULayer<> >();
   generator.Add<BilinearInterpolation<> >(28, 28, 56, 56, noiseDim / 4);
   generator.Add<Convolution<> >(noiseDim / 4, 1, 3, 3, 2, 2, 1, 1, 56, 56);
-  generator.Add<TanHLayer<> >();
+  generator.Add<SigmoidLayer<> >();
```
ShikharJ: I don't see why the change here is necessary, it is [recommended](https://github.com/soumith/ganhacks) practice to use `TanH` layer as the last layer of the `Generator`.

saksham189: In the O'Reilly test they use a sigmoid layer at the end in the generator. So, just to be consistent.
But I can revert this change. let me know.

src/mlpack/tests/gan_test.cpp
```diff
@@ -174,7 +175,7 @@ BOOST_AUTO_TEST_CASE(GANMNISTTest)
   Log::Info << trainData.n_rows << "--------" << trainData.n_cols << std::endl;
 
   // Create the Discriminator network
-  FFN<CrossEntropyError<> > discriminator;
+  FFN<SigmoidCrossEntropyError<> > discriminator;
```
ShikharJ: `SigmoidCrossEntropy` is nothing but the standard `CrossEntropy` extended to the case of negative output (multiple classes). We use `CrossEntropy` in almost all of our `GAN` implementations, because most of them make use of `sigmoid` layer at the end of the `Discriminator` and most papers implement GANs this way as well. So we don't even need to worry about negative output. However, in this case, either we can keep in line with O'Reilly's implementation and use `SigmoidCrossEntropy`, or we can try using a `Sigmoid` layer at the end of the `Discriminator` in place of `ReLU` (or introduce a further layer). Probably this is the reason why the other implementations didn't have an issue with convergence.

src/mlpack/tests/gan_test.cpp
ShikharJ: Please do.

saksham189: >  it is recommended practice to use TanH layer as the last layer of the Generator, since we need to keep the images between -1 and 1.

I am a bit confused about this. The training data we use is unnormalised. (it has large positive values and zeros)
Whereas in the O'Reilly test their images are normalized to [0, 1] and they have sigmoid layer at the end of the generator. Do you think we need to normalize the images to [-1, 1] and use the tanH layer at the end?

ShikharJ:
@saksham189 On second thought, I think it might be better to change the loss function to `SigmoidCrossEntropy`, since that would give us the freedom to choose our layers of `Discriminator` more freely. It kind of goes against what's explicitly mentioned in the papers, but if @rcurtin and @zoq don't have an objection with that, we can proceed. Can you try and make those changes to the other (namely `DCGAN`) test as well? I'll run the test and check the output at once, and then maybe we can continue with @walragatver's work?

src/mlpack/tests/gan_test.cpp
ShikharJ: Read my comment [here](https://github.com/mlpack/mlpack/pull/1724#discussion_r263302250). This test in its current form is only there to serve as a measure of logical correctness. You can add normalization for the images, it will slow down your computations a bit, but you can do without them as well. It's the layers that matter the most.

saksham189:
Yes, i have added the changes to the `DCGAN` tests as well.

src/mlpack/tests/gan_test.cpp
```diff
@@ -14,6 +14,7 @@
 
 #include <mlpack/methods/ann/init_rules/gaussian_init.hpp>
 #include <mlpack/methods/ann/loss_functions/cross_entropy_error.hpp>
+#include <mlpack/methods/ann/loss_functions/sigmoid_cross_entropy_error.hpp>
 #include <mlpack/methods/ann/gan/gan.hpp>
```
walragatver: It looks like you have removed use of `cross entropy error` from the file. So you can remove the `cross_entropy_error.hpp` header from file. 
`Note`:This is applicable if it is correct to have all `FFNs` `sigmoid cross entropy error` which will be decided by zoq and rcurtin.

ShikharJ:
@saksham189 Sorry for the late reply, I had my exams. Can you try and resolve the conflicts here? I think we should merge this in.

saksham189:
@ShikharJ I have my last exam on 9th so, I will continue the work after that. Hope that is fine.

ShikharJ:
I'm in the process of obtaining the output of this PR. If anyone has any further comments, please review. I'll merge this in as soon as the results come in (assuming they're good).

ShikharJ:
Okay, so I'm getting some quality results on the full `MNIST` for the standard `GAN`, I'll also run for `DCGAN` and update here. And the execution time on single core seems unfazed, so that's good:

`Standard GAN`
out_mnist

ShikharJ:
Thanks @saksham189 for the contribution!

PR #1815, opened 2019-03-24
Added velocity goal condition in continuous_mountain_car_V2

### LINKED ISSUE #1806: No velocity goal in mountain_car.hpp
#### Issue description

There should also be a velocity goal just like position goal variable as a reference value as and when velocity is updated.

#### Your environment

 * version of mlpack: 3.0.4
 * operating system: Windows 10
 * compiler: g++

#### Steps to reproduce

Compile and execute `mountain_car.hpp`

#### Expected behavior
The velocity goal variable will dictate in which direction and by how much change is allowed in each step.

#### Actual behavior
Due to lack of velocity goal variable controls are not that great.

abhinavsagar:
[Fixes #1806 ](https://github.com/mlpack/mlpack/issues/1806)


src/mlpack/methods/reinforcement_learning/environment/continuous_mountain_car_V2.hpp
```diff
@@ -0,0 +1,218 @@
...
+  /**
+   * Dynamics of Continuous Mountain Car. Get reward and next state based 
+   * on current state and current action.
+   *
+   * @param state The current state.
+   * @param action The current action.
+   * @param nextState The next state.
+   * @return reward, it's always -1.0.
+   */
+  double Sample(const State& state,
```
zoq: To reduce code duplication what about we hold an instance of ``ContinuousMountainCar`` and use ``mountainCar.Sample(...) inside the ``ContinuousMountainCarV2``, ``Sample`` method. Let me know what you think.

abhinavsagar:
@zoq Yeah I don't have any problem with that. Any more changes would you like to suggest?

src/mlpack/methods/reinforcement_learning/environment/continuous_mountain_car_V2.hpp
abhinavsagar: @zoq Yeah I don't have any problem with that. Any more changes would you like to suggest?

zoq: I think we can reuse some other functions from the existing ``ContinuousMountainCar`` class, but besides that I think this looks good.

abhinavsagar: @zoq Can you please review this now. Should I also add tests for these?

zoq: Yes, we should write a test for this one as well.

src/mlpack/methods/reinforcement_learning/environment/continuous_mountain_car_V2.hpp
```diff
@@ -0,0 +1,224 @@
+/**
+ * @file continuous_mountain_car_V2.hpp
+ * @author Abhinav Sagar
+ *
+ * This file is an implementation of Continous Mountain CarV2 task:
```
zoq: Does the MountainCarContinuous-v0 env you reference here include the velocity goal?

src/mlpack/methods/reinforcement_learning/environment/continuous_mountain_car_V2.hpp
```diff
@@ -0,0 +1,224 @@
...
+   * Dynamics of Continuous Mountain Car V2. Get reward based on current state and current
+   * action.
+   *
+   * @param state The current state.
+   * @param action The current action.
+   * @return reward, it's always -1.0.
+   */
+  double Sample(const State& state, const Action& action) const
+  {
+    ContinuousMountainCarV2 ob = ContinuousMountainCarV2(positionMin,
```
zoq: This will call the ``Sample``function from the ``ContinuousMountainCarV2`` class, but I was thinking about using the ``Sample`` function from the ``ContinuousMountainCar`` class since it's the same.

PR #1817, opened 2019-03-26
Overload binarize function

jeffin143:
I have overloaded binarize function to binarize multiple row, The existing one can binarize whole matrix or 1 row.

Also added 2 more test case

### DIFF
--- a/src/mlpack/core/data/binarize.hpp
+++ b/src/mlpack/core/data/binarize.hpp
@@ -13,6 +13,7 @@
 #ifndef MLPACK_CORE_DATA_BINARIZE_HPP
 #define MLPACK_CORE_DATA_BINARIZE_HPP
 
+#include <vector>
 #include <mlpack/prereqs.hpp>
 
 namespace mlpack {
@@ -116,7 +117,7 @@ template<typename T>
 void Binarize(const arma::Mat<T>& input,
               arma::Mat<T>& output,
               const double threshold,
-              vector<int>row)
+              std::vector<int>row)
 {
   output = input;
   for (size_t i = 0; i < row.size(); ++i)
--- a/src/mlpack/core/data/binarize.hpp
+++ b/src/mlpack/core/data/binarize.hpp
@@ -80,10 +80,10 @@ void Binarize(const arma::Mat<T>& input,
               const double threshold,
               const size_t dimension)
 {
-  // stopping invalid memory access and crashes 
+  // stopping invalid memory access and crashes
   if (dimension >= input.n_rows)
   {
-    throw std::invalid_argument("Invalid value for dimension");    
+    throw std::invalid_argument("Invalid value for dimension");
   }
   output = input;
 
--- a/src/mlpack/core/data/binarize.hpp
+++ b/src/mlpack/core/data/binarize.hpp
@@ -79,13 +79,60 @@ void Binarize(const arma::Mat<T>& input,
               const double threshold,
               const size_t dimension)
 {
+  // stopping invalid memory access and crashes 
+  if (dimension >= input.n_rows)
+  {
+    throw std::invalid_argument("Invalid value for dimension");    
+  }
   output = input;
 
   #pragma omp parallel for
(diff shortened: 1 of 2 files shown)


src/mlpack/core/data/binarize.hpp
```diff
@@ -79,13 +80,60 @@ void Binarize(const arma::Mat<T>& input,
...
+ * @param threshold Threshold can by any number.
+ * @param vector of dimension Feature to apply the Binarize function.
+ */
+template<typename T>
+void Binarize(const arma::Mat<T>& input,
+              arma::Mat<T>& output,
+              const double threshold,
+              std::vector<int>row)
+{
+  output = input;
```
zoq: I'm wondering if this is something that is used in practice, in most cases I want to binarize the complete matrix, and not sure I can encounter a case where I wanted to binarize a single row, but maybe this is something that is used?

Would be interesting to get some input from the community on this one.

ShikharJ: I don't know of a case where I'd like to binarize multiple rows. For a single row I understand it might get useful for setting the test labels for binary classification, but that's a guess in itself. I'd be +0 for this change.

jeffin143: @zoq @ShikharJ I agree that, it is not often that you have to binarize multiple rows, but it is also used in discretization of continuous feature values.

You may be right that this overloaded function may not be used by many , but i have encountered occassion where i have to binarize multiple columns and not row depending on different threshold so may be we could over load the function to take multiple threshold to binarize, something like

`void Binarize(const arma::Mat<T>& input,
              arma::Mat<T>& output,
              vector<double> threshold,
              vector<int>row)`

You could find an example [here](https://www.geeksforgeeks.org/sklearn-binarizer-in-python/) , an article in geeksforgeeks, where they have binarized multiple column using different threshold

Would like to here your take on this too.

Also there is [package](https://rdrr.io/github/LudvigOlsen/LRO.utilities/man/binarizer.html) in R,that would binarize multiple columns using same threshold

i would also like to have your input on : the existing overloaded function
 
`void Binarize(const arma::Mat<T>& input,
              arma::Mat<T>& output,
              const double threshold,
              const size_t dimension)`

which Binarizes a row, I also don't know why would anyone do that, I would surely binarize a column and not a row, as @ShikharJ suggested for  setting the test labels for binary classification, Since test labels would be along the column and not along the row.Binarizing a row would make the features go off track because every row would have some features consistency, except the row we binarized.

Maybe we could,change the function to binarize a particular column and not row.

ShikharJ: @jeffin143 In mlpack, each column is a data point. So test labels should be along the row.

jeffin143: > @jeffin143 In mlpack, each column is a data point. So test labels should be along the row.

oh i see.

rcurtin: I can definitely see the utility of binarizing the entire matrix, and that would probably be the "typical" case.  But I think it might still be useful to have a function for just one dimension, since you might want to set a different threshold for each dimension that you are binarizing.  I wasn't able to think of any common cases where you would want to binarize some dimensions but not all of them---but I suppose it's possible someone might want to do that (I am just not totally sure why).

zoq: @jeffin143 Thanks for the links, personally I'm inclined to discard the feature; if somebody depends on it, a simple for loop around the existing:

```c++
void Binarize(const arma::Mat<T>& input,
              arma::Mat<T>& output,
              const double threshold,
              const size_t dimension)
```

function will do the same. Don't have a strong opinion about this, in case anybody likes to get this merged.

jeffin143: @zoq , I agree with you. Closing the PR . Thanks for your input.

zoq: Even if we ended up not merging this in, it was still a good contribution.

PR #1818, opened 2019-03-26
Added a check to avoid invalid access of memory

jeffin143:
The function Impute() gets Dimension as input parameter 

`  void Impute(arma::Mat<T>& input,
              const T& mappedValue,
              const size_t dimension,
              const bool columnMajor = true)`

Now the user may give out of bound dimension, such as greater than the size of matrix and hence to avoid invalid access of memory, and to avoid any crash we could just have a check whether the dimension is inside the range or not.

### DIFF
--- a/.github/ISSUE_TEMPLATE/2-question.md
+++ b/.github/ISSUE_TEMPLATE/2-question.md
@@ -10,9 +10,9 @@ assignees: ''
 <!--
 
 Welcome!  If you have a question you'd like to ask, you can do it here or on the
-mlpack mailing list; see also http://mlpack.org/help.html.
+mlpack mailing list; see also https://mlpack.org/help.html.
 
 If you're looking for how to get involved and contribute, there's no need to
-open an issue---you can see http://www.mlpack.org/involved.html instead.
+open an issue---you can see https://www.mlpack.org/community.html instead.
 
 -->
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,8 +1,9 @@
 # Contributing to mlpack
 
-Anyone is welcome to contribute to mlpack and join the community.  If you would
-like to make improvements to the library or have found a bug that you know how
-to fix, please submit a pull request!
+mlpack is a community-led project; that means that anyone is welcome to
+contribute to mlpack and join the community!  If you would like to make
+improvements to the library, add new features that are useful to you and others,
+or have found a bug that you know how to fix, please submit a pull request!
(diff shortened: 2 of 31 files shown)


ShikharJ:
While I agree with your point of view, I don't feel this is the way to go about it. Checking for input correctness is pure overhead, and we don't want that when we're optimizing for speed.

rcurtin:
I don't mind a check, and I think the overhead should be pretty minimal as compared to the cost of the imputation itself, but I think maybe it would be better to have only one check in `Impute()` in `imputer.hpp`.

jeffin143:
@rcurtin , I guess that would be a better option to go, instead of checking in each impute function, we could only do once in imputer.hpp

jeffin143:
@rcurtin @ShikharJ 
under Armadillo documentation : 
`A std::logic_error exception is thrown if the requested element is out of bounds. The bounds check can be optionally disabled at compile-time to get more speed. `

http://arma.sourceforge.net/docs.html#element_access_bounds_check_note

ShikharJ:
@rcurtin As I mentioned, I agree with the point of view. We should be checking for wrong user input. But I seriously don't think this is the way to go. A better, more scalable approach would be to provide an check abstraction with dual functionality under different compile time flags. Take an example from `SymEngine` [under the topic Object Creation and is_canonical()](https://github.com/symengine/symengine/wiki/Beginner-Contributor-Guide-Design-of-SymEngine). In the `Release` mode, which is set using `CMake`, all the input based checks are turned off for speed. So if you need to check for an issue, simply re-compile by setting `Debug` mode which activates the checks and the input error gets caught. This is a great approach, and frankly what I'd like to see get incorporated into `mlpack` in the long term.

zoq:
@ShikharJ agreed, armadillo does the same thing as @jeffin143 already pointed out, instead of ``SYMENGINE_ASSERT`` it's called``arma_check``. And there is an open issue https://github.com/mlpack/mlpack/issues/891 that is somewhat related.

So, I think in this case, we could close this one in favour of a more general solution.

ShikharJ:
@zoq Nice! So maybe what we need to do is to just implement the checks at places. That should make for a good first issue as well. I agree, we should close this for a more general solution.

jeffin143:
@zoq @ShikharJ , closing this issue. Could you brief me what a general solution maybe so maybe I could work on it.

ShikharJ:
@jeffin143 We'll open an issue for it soon.

PR #1823, opened 2019-03-31
Introduce Memory Sharing Between GAN and FFN

saksham189:
Previously during training the predictors were not being shuffled for the discriminator so, they were visited in the same order while `EvaluateWithGradient` was called on discriminator.

### DIFF
--- a/src/mlpack/methods/ann/gan/gan.hpp
+++ b/src/mlpack/methods/ann/gan/gan.hpp
@@ -308,7 +308,6 @@ class GAN
   //! Modify the matrix of data points (predictors).
   arma::mat& Predictors() { return predictors; }
 
-
   //! Serialize the model.
   template<typename Archive>
   void serialize(Archive& ar, const unsigned int /* version */);
--- a/src/mlpack/methods/ann/gan/gan.hpp
+++ b/src/mlpack/methods/ann/gan/gan.hpp
@@ -210,7 +210,7 @@ class GAN
    * Gradient function for Standard GAN and DCGAN.
    * This function passes the gradient based on which network is being
    * trained, i.e., Generator or Discriminator.
-   * 
+   *
    * @param parameters present parameters of the network.
    * @param i Index of the predictors.
    * @param gradient Variable to store the present gradient.
@@ -228,7 +228,7 @@ class GAN
    * Gradient function for WGAN.
    * This function passes the gradient based on which network is being
    * trained, i.e., Generator or Discriminator.
-   * 
+   *
    * @param parameters present parameters of the network.
    * @param i Index of the predictors.
    * @param gradient Variable to store the present gradient.
@@ -245,7 +245,7 @@ class GAN
    * Gradient function for WGAN-GP.
    * This function passes the gradient based on which network is being
    * trained, i.e., Generator or Discriminator.
-   * 
+   *
    * @param parameters present parameters of the network.
    * @param i Index of the predictors.
    * @param gradient Variable to store the present gradient.
@@ -297,6 +297,17 @@ class GAN
 
   //! Return the number of separable functions (the number of predictor points).
   size_t NumFunctions() const { return numFunctions; }
+
+  //! Get the matrix of responses to the input data points.
+  const arma::mat& Responses() const { return responses; }
+  //! Modify the matrix of responses to the input data points.
+  arma::mat& Responses() { return responses; }
+
+  //! Get the matrix of data points (predictors).
+  const arma::mat& Predictors() const { return predictors; }
+  //! Modify the matrix of data points (predictors).
+  arma::mat& Predictors() { return predictors; }
+
 
   //! Serialize the model.
   template<typename Archive>
--- a/src/mlpack/methods/ann/gan/gan_impl.hpp
+++ b/src/mlpack/methods/ann/gan/gan_impl.hpp
@@ -65,18 +65,17 @@ GAN<Model, InitializationRuleType, Noise, PolicyType>::GAN(
 
   this->discriminator.deterministic = this->generator.deterministic = true;
 
-  responses.set_size(1, predictors.n_cols);
-  responses.ones();
+  this->predictors.set_size(predictors.n_rows, predictors.n_cols + batchSize);
+  this->predictors.cols(0, predictors.n_cols - 1) = predictors;
+  this->discriminator.predictors = arma::mat(this->predictors.memptr(),
+      this->predictors.n_rows, this->predictors.n_cols, false, false);
 
-  this->discriminator.predictors.set_size(predictors.n_rows,
-      predictors.n_cols + batchSize);
-  this->discriminator.predictors.cols(0, predictors.n_cols - 1) = predictors;
-  this->predictors = arma::mat(this->discriminator.predictors.memptr(),
-      predictors.n_rows, predictors.n_cols, false, false);
-  this->discriminator.responses.set_size(1, predictors.n_cols + batchSize);
-  this->discriminator.responses.ones();
-  this->discriminator.responses.cols(predictors.n_cols,
+  responses.set_size(1, predictors.n_cols + batchSize);
+  responses.ones();
+  responses.cols(predictors.n_cols,
       predictors.n_cols + batchSize - 1) = arma::zeros(1, batchSize);
+  this->discriminator.responses = arma::mat(this->responses.memptr(),
+      this->responses.n_rows, this->responses.n_cols, false, false);
 
   numFunctions = predictors.n_cols;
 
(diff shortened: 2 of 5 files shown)


saksham189:
Also, I think that we should have memory sharing between the predictors inside GAN class and predictors inside discriminator.
Let me know what you think @zoq @ShikharJ

src/mlpack/methods/ann/gan/gan_impl.hpp
```diff
@@ -370,7 +370,10 @@ template<
 >
 void GAN<Model, InitializationRuleType, Noise, PolicyType>::Shuffle()
 {
-  math::ShuffleData(predictors, responses, predictors, responses);
```
ShikharJ: Are you sure this doesn't work? It would rather indicate an issue with the `math::ShuffleData` function.

ShikharJ:
Hmm, in that case, I think it makes sense to leave the shuffling as is, and instead make the `discriminator.predictors` an alias of `GAN.predictors`. That way, when we shuffle, it would automatically take care of the issue.

src/mlpack/methods/ann/gan/gan_impl.hpp
```diff
@@ -72,7 +71,8 @@ GAN<Model, InitializationRuleType, Noise, PolicyType>::GAN(
   this->discriminator.predictors.set_size(predictors.n_rows,
       predictors.n_cols + batchSize);
   this->discriminator.predictors.cols(0, predictors.n_cols - 1) = predictors;
-
+  this->predictors = arma::mat(this->discriminator.predictors.memptr(),
```
ShikharJ: I think it should be the other way around. Here you have aliased the `Discriminator` predictors as `GAN` predictors.

src/mlpack/methods/ann/gan/gan_impl.hpp
```diff
@@ -42,7 +42,6 @@ GAN<Model, InitializationRuleType, Noise, PolicyType>::GAN(
     const double multiplier,
     const double clippingParameter,
     const double lambda):
-    predictors(predictors),
```
ShikharJ: Please revert this change. We wish to alias `GAN` predictors as `Discriminator` predictors.

src/mlpack/methods/ann/gan/gan_impl.hpp
saksham189: The discriminator predictors also include the fake data at the end.
So, we would have to do something like this
```
discriminator.predictors.cols(0, numsFunctions- 1) = arma::mat(predictors.memptr(), predictors.n_rows, predictors.n_cols, false, false);
```

I tried testing this. The data gets copied in the appropriate location in discriminator predictors but it doesn't seem like they are sharing memory with the GAN predictors. Maybe because we cannot have a larger matrix that shares memory with multiple different smaller matrix as that would require maintaining multiple pointers for the large matrix.
Do you have any different approach in mind?

ShikharJ: Try changing the size of the `GAN` predictors and responses, to the size of the `Discriminator` counterparts, and instead of directly copying into `discriminator.predictors` at places like this
```
discriminator.predictors.cols(numFunctions, numFunctions + batchSize - 1) =
      boost::apply_visitor(outputParameterVisitor, generator.network.back());
```
do
```
predictors.cols(numFunctions, numFunctions + batchSize - 1) =
      boost::apply_visitor(outputParameterVisitor, generator.network.back());
```

src/mlpack/methods/ann/gan/gan_impl.hpp
```diff
@@ -370,7 +369,10 @@ template<
 >
 void GAN<Model, InitializationRuleType, Noise, PolicyType>::Shuffle()
 {
-  math::ShuffleData(predictors, responses, predictors, responses);
+  arma::uvec ordering = arma::shuffle(arma::linspace<arma::uvec>(0,
+      numFunctions - 1, numFunctions));
+  arma::mat temp = predictors.cols(ordering);
+  predictors.cols(0, numFunctions - 1) = temp;
```
ShikharJ: Hmm, can't we do away with the use of temporary variable here? I think it might be a good optimization.

ShikharJ:
Thanks @saksham189. I'll ideally like to get this merged after we push in the fix for standard `GAN`. I went through the code, and it mostly looks good to me.

src/mlpack/methods/ann/gan/gan_impl.hpp
saksham189: I tried but it seems that discriminator predictors remain unchanged if i assign directly. Not sure why.

ShikharJ: Hmm maybe @rcurtin can help us with the best approach here?

rcurtin: I'm surprised that `predictors = predictors.cols(ordering)` doesn't work.  (Also, is `predictors.n_cols == numFunctions`?  If not my code in the last sentence has to be fixed on the left side to `predictors.cols(0, numFunctions - 1)` I guess.)

In any case, I'm also a bit confused on why we aren't shuffling `responses` here too; I think `ShuffleData()` should work.

Another option, assuming that `predictors.n_cols == numFunctions - 1`, is this:

```
arma::mat temp = predictors.cols(ordering);
predictors = std::move(temp);
```

That at least avoids the copy from `temp` to `predictors`, and I think is roughly how Armadillo would run the operation under the hood anyway.

ShikharJ: There's no need of shuffling responses, because they're all the same (arma::ones). So actually we're saving on computation by not using `ShuffleData()`. Even I think that we should be able to shuffle a few contiguous columns within a range of indices.

saksham189: I was implementing this wrong before which is why it wasn't working. I have removed the use of temporary variable now and `predictors = predictors.cols(ordering)` works as expected.

We are only shuffling the real images here so, that the real images and fake ones produced by the generator remain as distinct groups. The responses for all the real images are same i.e. `1` and therefore do not require shuffling.

rcurtin: Understood!  Thanks for the clarification. :+1:  Glad that you were able to get the code working. :)

saksham189:
Thanks for the review. @zoq  I have made the requested changes.

zoq:

ShikharJ:
@zoq I'll run the tests for this after merging in the O'Reilly GAN fix. Please review when you have some time.

zoq:
This looks good to me, do you think we could add a simple test for the shared memory functionality?

ShikharJ:
Hmm, maybe a simple test for calling `Shuffle()` and then checking if the indices in the discriminator and the GAN match-up.

zoq:
> Hmm, maybe a simple test for calling Shuffle() and then checking if the indices in the discriminator and the GAN match-up.

I like the idea.

src/mlpack/methods/ann/gan/gan.hpp
```diff
@@ -298,6 +298,17 @@ class GAN
...
+  const arma::mat& Responses() const { return responses; }
+  //! Modify the matrix of responses to the input data points.
+  arma::mat& Responses() { return responses; }
+
+  //! Get the matrix of data points (predictors).
+  const arma::mat& Predictors() const { return predictors; }
+  //! Modify the matrix of data points (predictors).
+  arma::mat& Predictors() { return predictors; }
+
+
```
ShikharJ: Please remove the extra space here.

src/mlpack/tests/gan_test.cpp
```diff
@@ -246,4 +246,60 @@ BOOST_AUTO_TEST_CASE(GANMNISTTest)
...
+      GaussianInitialization,
+      std::function<double()> >
+  gan(trainData, generator, discriminator, gaussian, noiseFunction,
+      noiseDim, batchSize, generatorUpdateStep, discriminatorPreTrain,
+      multiplier);
+
+  CheckMatrices(gan.Predictors(), gan.Discriminator().Predictors());
+  gan.Shuffle();
+  CheckMatrices(gan.Predictors(), gan.Discriminator().Predictors());
+}
```
ShikharJ: I guess we can add two checks here where we first check that `trainData` matches with `gan.Predictors()` and after shuffle it does not match with the same.

saksham189: Hey, I have made the changes. A different test is causing the build to fail. It should be irrelevant to the PR.

ShikharJ: No worries, in case this happens in the future, just let me know in the IRC and I can re-run the failing test.

ShikharJ:
Hmm, that failing test isn't passing on a re-build. I'll merge this anyways, if the failure persists, then maybe we can work on fixing it. Thanks for the contribution @saksham189!

PR #1829, opened 2019-04-02
Added support for concat along an axis.

MuLx10:
I have added an axis parameter for `Concat<>` layer to help achieve concatenation along a specific axis. It would help in models which need concat along an arbitrary axis.

Besides, the axis it also needs `inputShape` from the previous layer.

### DIFF
--- a/src/mlpack/methods/ann/layer/concat.hpp
+++ b/src/mlpack/methods/ann/layer/concat.hpp
@@ -61,7 +61,7 @@ class Concat
    * @param model Expose all network modules.
    * @param run Call the Forward/Backward method before the output is merged.
    */
-  Concat(arma::Row<size_t> inputSize,
+  Concat(arma::Row<size_t>& inputSize,
          const size_t axis,
          const bool model = false,
          const bool run = true);
--- a/src/mlpack/methods/ann/layer/concat.hpp
+++ b/src/mlpack/methods/ann/layer/concat.hpp
@@ -62,7 +62,7 @@ class Concat
    * @param run Call the Forward/Backward method before the output is merged.
    */
   Concat(arma::Row<size_t> inputSize,
-         const int axis,
+         const size_t axis,
          const bool model = false,
          const bool run = true);
 
@@ -211,7 +211,10 @@ class Concat
   arma::Row<size_t> inputSize;
 
   //! Parameter which indicates the axis of concatenation.
-  int axis;
+  size_t axis;
+
+  //! Parameter which indicates whether to use the axis of concatenation.
+  bool useAxis;
 
   //! Parameter to store channels
   size_t channels;
--- a/src/mlpack/methods/ann/layer/concat.hpp
+++ b/src/mlpack/methods/ann/layer/concat.hpp
@@ -61,7 +61,7 @@ class Concat
    * @param model Expose all network modules.
    * @param run Call the Forward/Backward method before the output is merged.
    */
-  Concat(std::vector<int> inputSize,
+  Concat(arma::Row<arma::sword> inputSize,
          const int axis = -1,
          const bool model = false,
          const bool run = true);
@@ -208,7 +208,7 @@ class Concat
 
  private:
   //! Parameter which indicates the input size of modules.
-  std::vector<int> inputSize;
+  arma::Row<arma::sword> inputSize;
 
   //! Parameter which indicates the axis of concatenation.
   int axis;
--- a/src/mlpack/methods/ann/layer/concat.hpp
+++ b/src/mlpack/methods/ann/layer/concat.hpp
@@ -213,11 +213,8 @@ class Concat
   //! Parameter which indicates the axis of concatenation.
   int axis;
 
-  //! Parameter to store oldColSize
-  size_t oldColSize;
-
-  //! Parameter to store newColSize
-  size_t newColSize;
+  //! Parameter to store channels
+  size_t channels;
 
   //! Parameter which indicates if the modules should be exposed.
   bool model;
--- a/src/mlpack/methods/ann/layer/concat.hpp
+++ b/src/mlpack/methods/ann/layer/concat.hpp
@@ -61,8 +61,8 @@ class Concat
    * @param model Expose all network modules.
    * @param run Call the Forward/Backward method before the output is merged.
    */
-  Concat(arma::Row<arma::sword> inputSize,
-         const int axis = -1,
+  Concat(arma::Row<size_t> inputSize,
+         const int axis,
          const bool model = false,
          const bool run = true);
 
@@ -208,7 +208,7 @@ class Concat
 
  private:
   //! Parameter which indicates the input size of modules.
-  arma::Row<arma::sword> inputSize;
+  arma::Row<size_t> inputSize;
 
   //! Parameter which indicates the axis of concatenation.
   int axis;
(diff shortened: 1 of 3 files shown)


src/mlpack/methods/ann/layer/concat.hpp
```diff
@@ -194,6 +207,15 @@ class Concat
   void serialize(Archive& /* ar */, const unsigned int /* version */);
 
  private:
+  //! Parameter which indicates the input size of modules.
+  std::vector<int> inputSize;
```
ShikharJ: I'd recommend using an `Armadillo` vector here.

MuLx10: Sure, I'll add it.

src/mlpack/methods/ann/layer/concat.hpp
```diff
@@ -53,6 +53,19 @@ class Concat
...
 
+  /**
+   * Create the Concat object using the specified parameters.
+   *
+   * @param inputSize A vector denoting input size of each layer added.
+   * @param axis Concat axis.
+   * @param model Expose all network modules.
+   * @param run Call the Forward/Backward method before the output is merged.
+   */
+  Concat(arma::Row<arma::sword> inputSize,
```
zoq: ``arma::sword`` is somewhat uncommon in the codebase, what you using ``size_t``instead?

src/mlpack/methods/ann/layer/concat.hpp
```diff
@@ -53,6 +53,19 @@ class Concat
...
+         const int axis = -1,
```
zoq: What about removing the default value, in this case a user is forced to set the value, which is currently done by checking against ``-1``.

src/mlpack/methods/ann/layer/concat.hpp
```diff
@@ -194,6 +207,15 @@ class Concat
   void serialize(Archive& /* ar */, const unsigned int /* version */);
 
  private:
+  //! Parameter which indicates the input size of modules.
+  arma::Row<arma::sword> inputSize;
```
zoq: See comment above.

src/mlpack/methods/ann/layer/concat.hpp
MuLx10: Yes, that would allow me to change to `size_t` instead of `int` or `uword`. However, that would make things difficult for users. I will change it.

src/mlpack/methods/ann/layer/concat.hpp
MuLx10: the inputSize may contain negative numbers, Since input size may be different for layers.
Eg (3,2,4) with (3,5,4) along axis 1, so the inputSize vector contains {3,-1,4}.

MuLx10: I used arma::sword instead of int to comply with mlpack.

rcurtin: We should use `size_t` since that is what is used to represent sizes everywhere else in mlpack.

src/mlpack/methods/ann/layer/concat.hpp
zoq: Hm, the user has to pass the axis value anyway, so not sure I see how this makes things more difficult.

src/mlpack/methods/ann/layer/concat.hpp
MuLx10: Sure, I have added it.

src/mlpack/methods/ann/layer/concat_impl.hpp
```diff
@@ -28,6 +28,27 @@ Concat<InputDataType, OutputDataType, CustomLayers...>::Concat(
...
+      const int axis,
+      const bool model,
+      const bool run) :
+      inputSize(inputSize),
+      axis(axis),
+      model(model),
+      run(run)
+{
+  parameters.set_size(0, 0);
+  channels = 1;
```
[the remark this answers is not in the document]
MuLx10: oh sure.

src/mlpack/methods/ann/layer/concat.hpp
```diff
@@ -53,6 +53,19 @@ class Concat
...
+  /**
+   * Create the Concat object using the specified parameters.
+   *
+   * @param inputSize A vector denoting input size of each layer added.
+   * @param axis Concat axis.
+   * @param model Expose all network modules.
+   * @param run Call the Forward/Backward method before the output is merged.
+   */
+  Concat(arma::Row<size_t> inputSize,
+         const int axis,
```
zoq: For consistency with the rest of the codebase I would like to use `size_t` instead of `int`, and for `-1` we could introduce a boolean `hasAxis`. Let me know what you think.

MuLx10: Done

src/mlpack/methods/ann/layer/concat_impl.hpp
```diff
@@ -54,13 +83,58 @@ void Concat<InputDataType, OutputDataType, CustomLayers...>::Forward(
...
-  // Vertically concatentate output from each layer.
+
+  // Parameter to store dimensions(rowSize).
+  size_t rowSize, oldColSize, newColSize;
   output = boost::apply_visitor(outputParameterVisitor, network.front());
+
+  newColSize = oldColSize = output.n_cols;
+
+  // Axis is specified and useAxis is true.
+  if (useAxis)
```
zoq: Looks like we could move this into the constructor, since we only need this once?

src/mlpack/methods/ann/layer/concat.hpp
```diff
@@ -53,6 +53,19 @@ class Concat
...
 
+  /**
+   * Create the Concat object using the specified parameters.
+   *
+   * @param inputSize A vector denoting input size of each layer added.
+   * @param axis Concat axis.
+   * @param model Expose all network modules.
+   * @param run Call the Forward/Backward method before the output is merged.
+   */
+  Concat(arma::Row<size_t> inputSize,
```
zoq: Should this be `const arma::Row<size_t>& inputSize`?

src/mlpack/methods/ann/layer/concat.hpp
```diff
@@ -194,6 +207,18 @@ class Concat
...
+  //! Parameter which indicates the input size of modules.
+  arma::Row<size_t> inputSize;
+
+  //! Parameter which indicates the axis of concatenation.
+  size_t axis;
+
+  //! Parameter which indicates whether to use the axis of concatenation.
+  bool useAxis;
+
+  //! Parameter to store channels
```
zoq: Missing `.` at the end.

MuLx10: Sorry for that.

src/mlpack/methods/ann/layer/concat.hpp
MuLx10: Actually, I need it as `arma::Row<size_t> inputSize` as a need to pass the values using initialization list `Concat<>({x,y,z},axis) ` .

src/mlpack/methods/ann/layer/concat_impl.hpp
MuLx10: Yeah sure

MuLx10:
Thanks @zoq

zoq:
Thanks for the contribution.

PR #1831, opened 2019-04-03
Documentation fixup

jeffin143:
I have included description about parameters in ballbound.hpp ,cellbound.hpp,hallow_ball_bound.hpp

### DIFF
--- a/src/mlpack/core/tree/cellbound.hpp
+++ b/src/mlpack/core/tree/cellbound.hpp
@@ -228,7 +228,7 @@ class CellBound
   /**
    * Determines if a point is within this bound.
    *
-   * @param point Point to check the codition.
+   * @param point Point to check the condition.
    */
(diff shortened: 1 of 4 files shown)


src/mlpack/core/tree/cellbound.hpp
```diff
@@ -89,6 +89,7 @@ class CellBound
   /**
    * Initializes to specified dimensionality with each dimension the empty
    * set.
+   * @param dimension Dimensionality of bound.
```
ShikharJ: I think there should be an additional `*` containing line above this, just to be consistent.

jeffin143: Done. Thanks for pointing it out

ShikharJ:
Thanks @jeffin143 for the contribution!

PR #1835, opened 2019-04-04
Feature selection based on variance thresholding

jeffin143:
This method removes features with variance below a certain cutoff, the idea is when a feature doesn’t vary much within itself, it generally has very little predictive power.

Thus we could remove those  features.


src/mlpack/core/data/feature_selection.hpp
```diff
@@ -0,0 +1,47 @@
+/**
+ * @file feature_selection_imp.hpp
+ * @author Jeffin Sam
+ *
+ * Feature selction based on variance thresholding.
+ * Motivated by the idea that low variance features contain less
+ * information.
+ * Calculate varience of each feature, then drop features 
+ * with variance below some threshold
```
zoq: Can we reference a paper here?

src/mlpack/core/data/feature_selection.hpp
```diff
@@ -0,0 +1,47 @@
+/**
+ * @file feature_selection_imp.hpp
```
zoq: The filename is wrong.

src/mlpack/core/data/CMakeLists.txt
```diff
@@ -24,6 +24,8 @@ set(SOURCES
   split_data.hpp
   imputer.hpp
   binarize.hpp
+  feature_selection.hpp
+  feature_selection_imp.hpp
```
zoq: For consistency we should go with `impl.hpp`.

src/mlpack/core/data/feature_selection_imp.hpp
```diff
@@ -0,0 +1,73 @@
...
+ * it generally has very little predictive power.
+ * Variance Threshold doesn’t consider the relationship of
+ * features with the target variable.
+ *
+ * @param input Input dataset with actual number of features.
+ * @param threshold Threshold for variance.
+ * @param output Output matrix with lesser number of features.
+ */
+template<typename T>
+void SelectBestFeature(const arma::Mat<T>& input,
```
zoq: The method name might be misleading, since we don't necessary select the best feature. Also we could be more specific, if this is based on the variance what about `VarianceSelection` or something in that direction. Let me know what you think.

src/mlpack/core/data/feature_selection_imp.hpp
```diff
@@ -0,0 +1,73 @@
...
+              const double threshold,
+              arma::Mat<T>& output)
+{
+  // Making sure features have same scale
+  arma::Mat<double> scale = arma::normalise(input);
```
zoq: Picky style comment, we could just use `arma::mat` here.

src/mlpack/core/data/feature_selection_imp.hpp
```diff
@@ -0,0 +1,73 @@
...
+  // Calculate variance of each feature
+  arma::Mat<double> value = arma::var(scale, 0, 1);
```
zoq: Looks like we could inline the `scale` calculation here.

src/mlpack/core/data/feature_selection.hpp
jeffin143: I added this[ paper](https://ieeexplore.ieee.org/document/4959366/references#references) as reference, but also found this [paper](http://cse.vnit.ac.in/people/narendraschaudhari/wp-content/uploads/sites/15/2014/10/054-NaiveBayesian-Yilin.pdf) interesting, but couldn't find out where was the latter one published.

Looking forward for your take on which one should i add.

src/mlpack/core/data/feature_selection.hpp
jeffin143: made apt correction

src/mlpack/core/data/CMakeLists.txt
jeffin143: renamed everywhere

src/mlpack/core/data/feature_selection_imp.hpp
jeffin143: Changed Function name to VarianceSelection

src/mlpack/core/data/feature_selection_imp.hpp
jeffin143: made changes

src/mlpack/core/data/feature_selection_imp.hpp
jeffin143: Yes possibly, made scale inline.

src/mlpack/core/data/feature_selection_impl.hpp
```diff
@@ -35,14 +35,12 @@ namespace data {
  * @param output Output matrix with lesser number of features.
  */
 template<typename T>
-void SelectBestFeature(const arma::Mat<T>& input,
+void VarianceSelection(const arma::Mat<T>& input,
               const double threshold,
               arma::Mat<T>& output)
 {
-  // Making sure features have same scale
-  arma::Mat<double> scale = arma::normalise(input);
```
jeffin143: I should have normalized it using row major ( dim = 1 ), made changes

jeffin143:
This branch was just going havoc, Don't know what was the problem even after resolving conflicts. May be i am not that good with git :( . Deleting this branch .

PR #1836, opened 2019-04-04
Add more Error metrics

greatsharma:
In this PR I added mean absolute error(MAE) and root mean squared error(RMSE), these two errors are very frequently used by machine learning practitioner while dealing with model's accuracy and errors, so having support for these two very common metrics will be very useful and handy for mlpack users.

I also added a function AssetColumnSizes() in facilities.hpp as this function is used by many metric classes. 

I replace code with function call to AssertColumnSizes() in mse_impl.hpp.

I also added tests for both metrics.

### DIFF
--- a/src/mlpack/core/cv/metrics/mse.hpp
+++ b/src/mlpack/core/cv/metrics/mse.hpp
@@ -36,7 +36,8 @@ class MSE
   template<typename MLAlgorithm, typename DataType, typename ResponsesType>
   static double Evaluate(MLAlgorithm& model,
                          const DataType& data,
-                         const ResponsesType& responses);
+                         const ResponsesType& responses,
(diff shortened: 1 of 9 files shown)


src/mlpack/core/cv/metrics/mse.hpp
```diff
@@ -36,7 +36,8 @@ class MSE
   template<typename MLAlgorithm, typename DataType, typename ResponsesType>
   static double Evaluate(MLAlgorithm& model,
                          const DataType& data,
-                         const ResponsesType& responses);
+                         const ResponsesType& responses,
+						              const std::string& callerDescription = "MSE::Evaluate()");
```
jeffin143: Remove spaces

rcurtin:
Closing for inactivity.

Issue #1254, opened 2018-02-17
CLI for ANN


arjunmenon:
Hey
Are ANN classes available from the CLI? From the docs it isn't apparent.
How can we use it otherwise?

zoq:
You are right there doesn't exist an CLI for the ANN classes. You would have to use it from within your C++ code. I think it would be great to have an executable for the network code, however since there is no single architecture it's somewhat complex to provide all the necessary settings and at the same time make it easy to be used.

arjunmenon:
I am not proficient with C++. I was hoping to add a Ruby wrapper to the CLI. So two things, 
1. Is there support planned for a Ruby binding
2. If I have to make an attempt, where do I start.

zoq:
For more information about how to add bindings to other languages please take a look at: http://www.mlpack.org/docs/mlpack-git/doxygen/bindings.html, let us know if we should clarify anything.

rcurtin:
Hi Arjun,

There isn't currently any planned support for Ruby.  If you are interested in writing a Ruby binding generator, it would be great and @zoq has given a link to some useful documentation.  But unfortunately proficiency with C++ is going to be necessary to write this binding generator, so you may want to also study C++ a little more closely before diving too deep.

desai-aditya:
Hi rcurtin,

I was planning to work on this in small pieces. Could I first implement a CLI that achieves this -
https://github.com/mlpack/models/blob/master/Kaggle/DigitRecognizer/src/DigitRecognizer.cpp
and then later improve it?

rcurtin:
We need to wait for @zoq's input also, but my opinion is that any CLI program needs the following functionality:

 * specify a neural network architecture arbitrarily (possibly by loading it from a YAML file or something?  We should probably discuss that part)
 * train on some given training data, with the user able to specify the optimizer and optimizer parameters
 * evaluate a trained model on some given testing data
 * print performance measures on training/test data
 * load an existing model or save a newly trained model

So writing something for a digit recognizer might be a good warmup exercise, but anything that we merge into mlpack should at least have the requirements above, in my opinion.

akhandait:
@zoq, since we don't have any CLI for ANN classes, do you think it will be a good idea to start with a CLI for fully connected ANN with options for all other required parameters?
If yes, I would like to work on it.
@rcurtin I think it will be possible to incorporate all the other functionality mentioned above.
Maybe we could further develop it to support any neural network architecture.

desai-aditya:
[...]

PR #1837, opened 2019-04-04
Create a parser to parse the model from a json

sreenikSS:
The json file should contain the loss function, init type, optimizer details and the network layers corresponding to the architecture in mind. The format of the json is mentioned in the issue https://github.com/mlpack/mlpack/issues/1254

There are a number of flaws, I will try to remove them in subsequent commits. I shall address them in the issue itself.

### DIFF
--- a/src/mlpack/methods/ann/modelParser.cpp
+++ b/src/mlpack/methods/ann/modelParser.cpp
@@ -50,8 +50,6 @@
 #include <boost/foreach.hpp>
 #include <boost/algorithm/string.hpp>
 
-#include "kaggle_utils.hpp"
-
 using namespace mlpack;
 using namespace mlpack::ann;
 using namespace mlpack::optimization;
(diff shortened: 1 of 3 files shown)


zoq:
Thanks for opening the PR, do you mind to change the title of the PR to something more descriptive?

sreenikSS:
Oops, I'm sorry for that. Fixed it

src/mlpack/methods/ann/modelParser.cpp
```diff
@@ -0,0 +1,1132 @@
...
+
+/**
+ * @author Eugene Freyman
+ * Returns the accuracy (percentage of correct answers).
+ * @param predLabels predicted labels of data points.
+ * @param realY real labels (they are double because we usually read them from
+ * CSV file that contain many other double values).
+ * @return percentage of correct answers.
+ */
+double accuracy(arma::Row<size_t> predLabels, const arma::mat& realY)
```
MuLx10: It is better to include the file instead of writing it again

src/mlpack/methods/ann/modelParser.cpp
```diff
@@ -0,0 +1,1132 @@
...
+using namespace mlpack::optimization;
+using namespace arma;
+using namespace std;
+using namespace boost::property_tree;
+
+/**
+ * Implementation of the entire dataset consisting of the training data
+ * and validation data
+ */
+class Dataset
```
MuLx10: Would you mind splitting the file into .hpp and .cpp

src/mlpack/methods/ann/modelParser.cpp
```diff
@@ -0,0 +1,1132 @@
...
+/**
+ * Update the values of a given stl map with that of another map 
+ * corresponding to the keys that are common
+ * 
+ * Keys are of type string and values are of type double
+ * 
+ * @param origParams The map whose values will be updated
+ * @param newParams The map whose values will be used to update origParams
+ */
+void updateParams(map<string, double> &origParams, map<string, double> &newParams)
```
MuLx10: You can use a copy constructor instead or directly assign values to optimizerparam instead of copying in loop.

src/mlpack/methods/ann/modelParser.cpp
sreenikSS: Thanks for the suggestion, but the file is in the *models* repo, so I cannot really include it. Moreover it is basic and simple code, so I am not sure if code repetition can be a big issue here.

src/mlpack/methods/ann/modelParser.cpp
sreenikSS: Sure I will do it. Actually the code has some visible scope of improvement that I am aware of (of course, along with things I am not aware of). I will certainly split it before finally requesting a merge. It might have some redundancy in the include statements as well. I'll clean them up after my exams end (or if I can manage time in between).

src/mlpack/methods/ann/modelParser.cpp
sreenikSS: Thanks for the suggestion. That might be a better solution, though I will have to look into it a bit more. One reason why it does what it does is that I wanted to check for errors. For example, some keys are absolutely necessary, some are not. That validation part is not done yet. I will keep this in mind while implementing the validation process.

src/mlpack/methods/ann/modelParser.cpp
```diff
@@ -0,0 +1,1132 @@
...
+
+/** The final implementation of this file would not have a main method. This is 
+ * merely to ease testing. The following include satement can hence be 
+ * removed later
+*/
+#include <mlpack/core/data/split_data.hpp>
+int main()
+{
+  //arma::mat trainX, trainY;
+  string fileName = "network3.json";
```
zoq: Perhaps you can post this file somewhere? Is the json file autogenerated?

sreenikSS: Sorry for the delay. I had completely missed this comment. 
The json file is user specified. I think it will be a lot clearer if you take a look [here](https://github.com/sreenikSS/Testing/blob/master/Dataset3.json)

rcurtin:
Hey @sreenikSS, still interested in working on this one?  We can let it go stale (or let someone else pick it up) if not. :+1:

sreenikSS:
@rcurtin I'd guess letting it go stale would be fine because as per our previous discussions, this is supposed to be part the new *model converter* repository when it'll be created.

As far as anyone else taking this issue up is concerned, I don't think that will be a good idea because the work is complete (and lying around in my own repo) :)

PR #1841, opened 2019-04-05
Added decay rate hyperparameter to greedy policy.

abhinavsagar:
Added decay rate in `greedy_policy.hpp` file.

### DIFF
--- a/src/mlpack/methods/reinforcement_learning/policy/greedy_policy.hpp
+++ b/src/mlpack/methods/reinforcement_learning/policy/greedy_policy.hpp
@@ -1,6 +1,7 @@
 /**
  * @file greedy_policy.hpp
  * @author Shangtong Zhang
+ * @author Abhinav Sagar
  *
  * This file is an implementation of epsilon greedy policy.
  *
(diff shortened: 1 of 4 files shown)


src/mlpack/methods/reinforcement_learning/policy/greedy_policy.hpp
```diff
@@ -41,13 +42,16 @@ class GreedyPolicy
...
    *        will anneal.
    * @param minEpsilon Epsilon will never be less than this value.
+   * @param decayRate How much to change the model in response to the
+   *        estimated error each time the model weights are updated.
    */
   GreedyPolicy(const double initialEpsilon,
                const size_t annealInterval,
-               const double minEpsilon) :
+               const double minEpsilon,
+               const double decayRate) :
```
zoq: To provide backward compatibility we should set a default value of one, that should also fix the build issue.

src/mlpack/tests/q_learning_test.cpp
```diff
@@ -49,7 +49,7 @@ BOOST_AUTO_TEST_CASE(CartPoleWithDQN)
   model.Add<Linear<>>(128, 2);
 
   // Set up the policy and replay method.
-  GreedyPolicy<CartPole> policy(1.0, 1000, 0.1);
+  GreedyPolicy<CartPole> policy(1.0, 1000, 0.1, 0.99);
```
zoq: Haven't tested it but does this has any visible effect on the test results?

abhinavsagar: @zoq Yes by adding this the change in parameters as they approach the final result is reducing as the iteration proceeds.

zoq:

abhinavsagar:
Thanks.

PR #1842, opened 2019-04-06
change weak learner type of the Adaboost to DecisionTree::ID3DecisionStump

rickllyxu:
In [pull #1577 ](https://github.com/mlpack/mlpack/pull/1577), I have fix the bugs of decision stump. But I was informed that it is better to remove the DecisionStump class entirely and replace it with DecisionTree<..., true>. 

So in this PR, I partly finished this task. Please review the code. 

And I know that I should add corresponding test case for `ID3DecisionStump` and `Adaboost<ID3DecisionStump>` later, and later remove DecisionStump class entirely.


src/mlpack/methods/adaboost/adaboost_model.hpp
```diff
@@ -79,7 +79,7 @@ class AdaBoostModel
   //! Modify the dimensionality of the model.
   size_t& Dimensionality() { return dimensionality; }
 
-  //! Train the model.
+  //! Train the model, treat the data is all of the numeric type.
```
rcurtin: It's nice to add this comment here, but it would be more correct to say `treat the data as all numeric`.

src/mlpack/methods/decision_tree/decision_tree.hpp
```diff
@@ -123,6 +124,18 @@ class DecisionTree :
...
+               MatType data,
+               const data::DatasetInfo& datasetInfo,
+               LabelsType labels,
+               const size_t numClasses,
+               WeightsType weights,
+               const size_t minimumLeafSize = 10,
+               const double minimumGainSplit = 1e-7,
+               const std::enable_if_t<arma::is_arma_type<
+                       typename std::remove_reference<WeightsType>::type>::value>*
+               = 0);
```
rcurtin: Can we add some documentation for this function?  Also, the wrapped line for the SFINAE bit only needs to be indented 4 characters, which should let the `= 0` fit on the line above and clean it up a bit. :)

src/mlpack/methods/decision_tree/decision_tree.hpp
```diff
@@ -150,6 +163,17 @@ class DecisionTree :
...
+  DecisionTree(const DecisionTree& other,
+               MatType data,
+               LabelsType labels,
+               const size_t numClasses,
+               WeightsType weights,
+               const size_t minimumLeafSize = 10,
+               const double minimumGainSplit = 1e-7,
+               const std::enable_if_t<arma::is_arma_type<
+                       typename std::remove_reference<WeightsType>::type>::value>*
+               = 0);
```
rcurtin: Same comment here.

src/mlpack/methods/decision_tree/decision_tree.hpp
```diff
@@ -479,6 +503,7 @@ using DecisionStump = DecisionTree<FitnessFunction,
                                    ElemType,
                                    false>;
 
+typedef DecisionTree<InformationGain, BestBinaryNumericSplit, AllCategoricalSplit, AllDimensionSelect, double, true> ID3DecisionStump;
```
rcurtin: Can we also add a little documentation here for this?  And it should wrap so it is 80 characters per line. :+1:

src/mlpack/methods/decision_tree/decision_tree.hpp
```diff
@@ -150,6 +163,17 @@ class DecisionTree :
...
                     = 0);
 
+  template<typename MatType, typename LabelsType, typename WeightsType>
+  DecisionTree(const DecisionTree& other,
+               MatType data,
+               LabelsType labels,
+               const size_t numClasses,
+               WeightsType weights,
+               const size_t minimumLeafSize = 10,
+               const double minimumGainSplit = 1e-7,
```
rcurtin: If we are using `other` to take parameters, wouldn't `minimumLeafSize` and `minimumGainSplit` be unnecessary?

src/mlpack/methods/decision_tree/decision_tree_impl.hpp
```diff
@@ -157,6 +201,47 @@ DecisionTree<FitnessFunction,
...
+                                   MatType data,
+                                   LabelsType labels,
+                                   const size_t numClasses,
+                                   WeightsType weights,
+                                   const size_t minimumLeafSize,
+                                   const double minimumGainSplit,
+                                   const std::enable_if_t<
+                                           arma::is_arma_type<
+                                                   typename std::remove_reference<
+                                                           WeightsType>::type>::value>*):
```
rcurtin: Probably this could look a little cleaner if you just wrap the first parameter `const DecisionTree& other` to four characters indented on a new line and have the rest match that.

src/mlpack/methods/decision_tree/decision_tree_impl.hpp
```diff
@@ -157,6 +201,47 @@ DecisionTree<FitnessFunction,
...
+        NumericAuxiliarySplitInfo(other),
+        CategoricalAuxiliarySplitInfo(other)  // other info does need to copy
```
rcurtin: I think all that needs to be copied is `minimumLeafSize` and `minimumGainSplit`.

rickllyxu:
Thanks your review. I am more clear to aware the design of the code. I will update it as soon as possible.

rickllyxu:
Thanks your review. I am more clear to aware the design of the code. I will update it as soon as possible.

src/mlpack/methods/decision_tree/decision_tree.hpp
rickllyxu: So if we do not pass the these two parameters,  I think we need to add new private member variables to save `minimumLeafSize` and `minimumGainSplit`.

rickllyxu:
I argue that it should pass `minimumLeafSize` and `minimumGainSplit` parameters when constructing `DecisionTree` class and save them as member variables. Besides, remove these two parameters from `Train()` functions.
How do you think of it?

rickllyxu:
@rcurtin Hi, Ryan, is it possible to abolish some hyper-parameter passing about `Train()` method? 
I think the hyper-parameters assignment of model is the business of the model constructor. 
If in some context, `Train()` has to  take some hyper-parameters to change existing model, why not to re-initial a model with desired hyper-parameters?

So I cancel the transmission of hyper-parameters in `Train()` method.

rickllyxu:
> @rcurtin Hi, Ryan, is it possible to abolish some hyper-parameter passing about `Train()` method?
> I think the hyper-parameters assignment of model is the business of the model constructor.
> If in some context, `Train()` has to take some hyper-parameters to change existing model, why not to re-initial a model with desired hyper-parameters?
> 
> So I cancel the transmission of hyper-parameters in `Train()` method. How do you think of it?

Hello, Ryan, I know that if I save hyper-parameters as model's member variable, there will be many memory waste. But I have to save hyper-parameters as member variables, otherwise, how could I get `other.minimumGainSplit`? 

And If I save hyper-parameters as model's member variable, I have to remove the hyper-paras passing in `Train()` method, otherwise I don't know which hyper-paras should be used when training.

src/mlpack/methods/adaboost/adaboost_model.cpp
```diff
@@ -105,13 +105,13 @@ void AdaBoostModel::Train(const mat& data,
...
-
-    DecisionStump<> ds(data, labels, max(labels) + 1);
-    dsBoost = new AdaBoost<DecisionStump<>>(data, labels, numClasses, ds,
+    ID3DecisionStump ds(data, labels, max(labels) + 1);
+    dsBoost = new AdaBoost<ID3DecisionStump>(data, labels, numClasses, ds,
         iterations, tolerance);
   }
   else if (weakLearnerType == WeakLearnerTypes::PERCEPTRON)
   {
+    delete pBoost;
```
rcurtin: Ooh, nice catch. :+1:

rcurtin:
Sorry for my slow response on this one.  You're right that we are avoiding saving the hyperparameters in the object to save memory usage.  But then how we pass them between Adaboost models is quite tricky.  Do you think that we could use the same strategy as the hyperparameter tuner uses, and pass them as template varargs to the `AdaBoost` constructor and `Train()` methods?

rickllyxu:
> Sorry for my slow response on this one. You're right that we are avoiding saving the hyperparameters in the object to save memory usage. But then how we pass them between Adaboost models is quite tricky. Do you think that we could use the same strategy as the hyperparameter tuner uses, and pass them as template varargs to the `AdaBoost` constructor and `Train()` methods?

Can you tell me the detail about strategy that the hyperparameter tuner uses and how to pass them as template vargars? Can you give me example code?

If we save hyperparas as object member variable, it does cost extra memory. But if we don't, copying the setting of a model really depends on user. 

Besides, in fact, we can refactor `AdaBoost` and its weakleaners, and change following function invocation to others :
```
WeakLearnerType w(other, tempData, labels, numClasses, weights);
```
Which approach do you suggest?

rcurtin:
So, the hyperparameter tuner works using extra parameters to the `Train()` method.  It uses some nice template metaprogramming to pass different values at the right time.  One of the things we could do for AdaBoost is then add a `Train()` method like this:

```
  template<typename... WeakLearnerArgTypes>
  double Train(const MatType& data,
               const size_t numClasses,
               const WeakLearnerType& learner,
               const size_t iterations = 100,
               const double tolerance = 1e-6,
               WeakLearnerArgTypes... args);
```

Later you can use this by calling, e.g.,

```
WeakLearnerType w(tempData, labels, numClasses, weights, args...);
```

(This also means the constructor that takes another learner is unnecessary and we can deprecate it for the perceptron, and avoid adding it to `DecisionTree`.  We might need to modify `Perceptron` a little bit to have the right API that accepts the hyperparameters.  If you want, I can prepare a patch for that.

rickllyxu:
Make sense! But where can we place the `class WeakLearnerArgTypes`? 
I think we can create a patch like this:
```
namespace mlpack {
namespace tree {
class DecisionTreeArgs {
};
// And following is existing code about DecisionTree
```
We can create `ArgDict` in `mlpack::tree` namespace and `mlpack::perceptron` namespace.
How do do think of it?

rcurtin:
I don't think `DecisionTreeArgs` is necessary since `WeakLearnerArgTypes` is a vararg template (i.e. it can be many types).

So given the definition I showed, you should be able to call `Train()` like this for decision trees:

```
model.Train(data, labels, numClasses, decisionTree, 100, 1e-6,
    // now everything else is DecisionTree::Train() parameters
    minimumLeafSize, minimumGainSplit);
```

Then inside `Train()`, `WeakLearnerArgTypes` will be `double, double`, and then we can pass both of those doubles to `DecisionTree::Train()` by passing `args...` (the `...` "unpacks" the varargs).

Does that make sense?  Let me know if I can clarify.

gauravk97:
Hi, I was looking somewhere to start digging in. This looks open.. Can I take it up? If yes, do I start with rickllyxu's code?

rcurtin:
@gauravk97 sure, feel free.  But it may be worth asking @rickllyxu what he thinks, or if he has time to handle this one (I guess not because it went stale and was closed).  In any case it is pretty close; you can read the comments here to see what remains to be done.

rickllyxu:
> Hi, I was looking somewhere to start digging in. This looks open.. Can I take it up? If yes, do I start with rickllyxu's code?

you are welcome.

himanshupathak21061998:
Hey, @rcurtin I want to resurrect this should I add changes in this pr or start with new pr.

jeffin143:
> Hey, @rcurtin I want to resurrect this should I add changes in this pr or start with new pr.

You can probably fetch his branch and then just add your work over it and then open a new PR :), That way you need not redo it from start again

Although it is unto you, whatever you feel comfortable with :)

rcurtin:
@himanshupathak21061998 sounds good to me, I'd love to get this finished and merged.

PR #1843, opened 2019-04-06
Implemetation of Concatenated ReLU Activation

jeffin143:
I tried implementing CRELU Activation, following the research paper by `Wenling Shang, Kihyuk Sohn, Diogo Almeida, Honglak Lee,` in [2016](https://arxiv.org/abs/1603.05201)

This activation scheme preserves both positive and negative phase information while enforcing non-saturated  non-linearity.The unique nature  of CReLU allows a mathematical characterization of convolution layers in terms of reconstruction property.

CReLU has shown to perform very well in Convolutional Neural Networks.CReLU naturally reduces redundancy of learning separate filters that differ in phase only.

I encountered lot of doubts during implementation and have done some tweaking during implementation.So it may be not be fully sound on implementation part. So i won't mind even if takes long for you to review this PR, But i would like to get your input on the code.


src/mlpack/methods/ann/layer/c_relu_impl.hpp
```diff
@@ -0,0 +1,87 @@
...
+template<typename InputDataType, typename OutputDataType>
+template<typename InputType, typename OutputType>
+void CReLU<InputDataType, OutputDataType>::Forward(
+    const InputType&& input, OutputType&& output)
+{
+  // Optimisation needed
+  OutputType temp1;
+  OutputType temp2;
+  Fn(input, temp1);
+  InputType inptemp=-1*input;
```
jeffin143: Could be optimized, Seeking for input

src/mlpack/methods/ann/layer/c_relu_impl.hpp
```diff
@@ -0,0 +1,87 @@
...
+  Fn(inptemp, temp2);
+  // Concat Neg and Pos Relu
+  output = arma::join_cols(temp1,temp2);
```
jeffin143: Concatenation of Relu

src/mlpack/methods/ann/layer/c_relu_impl.hpp
```diff
@@ -0,0 +1,87 @@
...
+template<typename InputDataType, typename OutputDataType>
+template<typename DataType>
+void CReLU<InputDataType, OutputDataType>::Backward(
+    const DataType&& input, DataType&& gy, DataType&& g)
+{
+  DataType derivative;
+  Deriv(input,derivative);
+  DataType temp;
+  temp = gy % derivative;
+  g= temp.rows(0, (input.n_rows/2-1) )-temp.rows(input.n_rows/2,(input.n_rows-1));
```
jeffin143: Backward Propagation, Since the CRELU double the input and hence which backward propagation `w^T*g` throws error due to size miss match, so making it of same size

src/mlpack/methods/ann/layer/c_relu_impl.hpp
```diff
@@ -0,0 +1,87 @@
...
+  * DataType temp2;
+  * Deriv(input, temp1);
+  * DataType inptemp=-1*input;
+  * Deriv(inptemp,temp2);
+  * DataType g1;
+  * DataType g2;
+  * g1 = gy % temp1;
+  * g2 = gy % temp2;
+  * derivative=arma::join_cols(temp1,temp2);
+  * g=arma::join_cols(g1,g2);
```
jeffin143: Some place had CRELU implementation like above

jeffin143:
I made a comparison between LeakyRelu and CRelu, using `Mlpack/models` repo examples .
I took [DigitRecognizer](https://github.com/mlpack/models/blob/master/Kaggle/DigitRecognizer/src/DigitRecognizer.cpp)  program and the dataset and Used the following model 

Model for Leaky Relu
```cpp
  FFN<NegativeLogLikelihood<>, RandomInitialization> model;
  model.Add<Linear<> >(trainX.n_rows,100);
  model.Add<LeakyReLU<> >();
  model.Add<Linear<> >( 100,100);
  model.Add<LeakyReLU<> >();
  model.Add<Linear<> >(100, 10);
  model.Add<LogSoftMax<> >();
```

Screenshot from 2019-04-06 09-01-59

Model For Crelu:
```cpp
  FFN<NegativeLogLikelihood<>, RandomInitialization> model;
  model.Add<Linear<> >(trainX.n_rows,100);
  model.Add<CReLU<> >();
  model.Add<Linear<> >( 200,100);
  model.Add<CReLU<> >();
  model.Add<Linear<> >(200, 10);
  model.Add<LogSoftMax<> >();
```
Screenshot from 2019-04-06 09-01-55

Both gave almost same amount of performance,but may be some amount of tweaking in the implementation part may increase the accuracy.
CRELU ran for 234 sec, and LeakyReLU ran for 199 sec , may be optimizing CRELU could be possible by avoiding copying of data.

Though the following results show that CRELU and LEakyReLU are performing almost same, but the research paper had some good results when it comes to CNN part over the time.

Would like the community to comment on it.

src/mlpack/methods/ann/layer/c_relu.hpp
```diff
@@ -0,0 +1,151 @@
...
+ private:
+  /**
+   * Computes the ReLU function
+   *
+   * @param x Input data.
+   * @return f(x).
+   */
+  double Fn(const double x)
+  {
+    return std::max(x, 0 * x);
```
ShikharJ: I don't understand the rationale behind multiplying a double by 0, and then comparing with itself. Simply use a `0.0`?

src/mlpack/methods/ann/layer/c_relu.hpp
```diff
@@ -0,0 +1,151 @@
+/**
+ * @file c_relu_impl.hpp
+ * @author Jeffin Sam
+ *
+ * Implementation of CReLU layer.
+ * Introduced by,
+ * Wenling Shang, Kihyuk Sohn, Diogo Almeida, Honglak Lee,
+ * "https://arxiv.org/abs/1603.05201", 16th March 2016.
```
zoq: Can you use a similar comment/cite style as used here:

https://github.com/mlpack/mlpack/blob/60fb72eb377901337bca7708e57a038d83226c23/src/mlpack/methods/ann/layer/dropconnect.hpp#L27-L51

src/mlpack/methods/ann/layer/c_relu.hpp
```diff
@@ -0,0 +1,151 @@
...
+ */
+template <
+    typename InputDataType = arma::mat,
+    typename OutputDataType = arma::mat
+>
+class CReLU
+{
+ public:
+  /**
+   * Create the CReLU object using the specified parameters.
```
zoq: There are no specified parameter.

src/mlpack/methods/ann/layer/c_relu_impl.hpp
zoq: Not sure I get the point, do you mean some other framework implemented the method above?

src/mlpack/methods/ann/layer/c_relu_impl.hpp
```diff
@@ -0,0 +1,88 @@
...
+    const InputType&& input, OutputType&& output)
+{
+  // Optimisation needed
+  OutputType temp1;
+  OutputType temp2;
+  Fn(input, temp1);
+  InputType inptemp = -1 * input;
+  Fn(inptemp, temp2);
+  // Concat Neg and Pos Relu
+  output = arma::join_cols(temp1, temp2);
```
zoq: Looks like we can avoid some temporary copies here by using:

```c++
arma::join_cols(std::max(input, 0.0), std::max(-1.0 * input, 0.0));
```

src/mlpack/methods/ann/layer/c_relu_impl.hpp
jeffin143: @zoq , Yes some of the other frameworks did it, But i took the idea from [chainer](https://github.com/chainer/chainer/blob/ccc5469bf318c191583721ec62370e9dbc5a96a0/chainer/functions/activation/crelu.py#L45) framework

src/mlpack/methods/ann/layer/c_relu.hpp
jeffin143: Made changes.

src/mlpack/methods/ann/layer/c_relu.hpp
jeffin143: Made apt changes

src/mlpack/methods/ann/layer/c_relu.hpp
jeffin143: @zoq , Sorry for the error, changed it

src/mlpack/methods/ann/layer/c_relu_impl.hpp
jeffin143: Yup, that would be helpful, Done.

src/mlpack/methods/ann/layer/c_relu.hpp
```diff
@@ -0,0 +1,165 @@
...
+#ifndef MLPACK_METHODS_ANN_LAYER_C_RELU_HPP
+#define MLPACK_METHODS_ANN_LAYER_C_RELU_HPP
+
+#include <mlpack/prereqs.hpp>
+
+namespace mlpack {
+namespace ann /** Artificial Neural Network. */ {
+
+/**
+ *
```
zoq: We can remove the empty line here.

src/mlpack/methods/ann/layer/c_relu.hpp
```diff
@@ -0,0 +1,165 @@
...
+ * A concatenated ReLU has two outputs, one ReLU and one negative ReLU,
+ * concatenated together.In other words, for positive x it produces [x, 0],
```
zoq: Missing space after `.`. Same for the line below.

src/mlpack/methods/ann/layer/c_relu.hpp
```diff
@@ -0,0 +1,165 @@
...
+ * and for negative x it produces [0, x].Because it has two outputs,
+ * CReLU doubles the output dimension.
+ *
+ * Note:
+ * During building of model, The next layer of crelu should have double the size
```
zoq: This sound strange, what about "The CReLU doubles the output size.". And remove the rest?

src/mlpack/methods/ann/layer/c_relu.hpp
```diff
@@ -0,0 +1,165 @@
...
+ * as given input layer since it concatenates the input.
+ *
+ * For more information, see the following.
+ *
+ * @code
+ * @inproceedings{ICML2016,
+ *   title={Understanding and Improving Convolutional Neural Networks
```
zoq: Do you mind to adjust the format here to be consistent with the rest of the codebase?

src/mlpack/methods/ann/layer/c_relu.hpp
```diff
@@ -0,0 +1,165 @@
...
+ *           via Concatenated Rectified Linear Units},
+ *   author = {LWenling Shang, Kihyuk Sohn, Diogo Almeida, Honglak Lee},
+ *   year = {2016}
+ * }
+ * @endcode
+ *
```
zoq: We can remove the extra line here.

src/mlpack/methods/ann/layer/c_relu.hpp
```diff
@@ -0,0 +1,165 @@
...
+  void serialize(Archive& /* ar */, const unsigned int /* version */);
+
+ private:
+  /**
+   * Computes the ReLU function
+   *
+   * @param x Input data.
+   * @return f(x).
+   */
+  double Fn(const double x)
```
zoq: Do we need the methods here?

src/mlpack/tests/activation_functions_test.cpp
```diff
@@ -330,6 +330,54 @@ void CheckPReLUGradientCorrect(const arma::colvec input,
...
 
+/*
+ * Implementation of the CReLU activation function test. The function
+ * is implemented as CReLU layer in the file c_relu.hpp
+ *
+ * @param input Input data used for evaluating the CReLU activation
+ *   function.
+ * @param target Target data used to evaluate the CReLU activation.
+ */
+void CheckCReLUActivationCorrect(const arma::colvec input,
```
zoq: Do you think we can integrate this into the test itself, to reduce the number of lines?

src/mlpack/methods/ann/layer/c_relu.hpp
jeffin143: Removed

src/mlpack/methods/ann/layer/c_relu.hpp
jeffin143: Added

src/mlpack/methods/ann/layer/c_relu.hpp
jeffin143: Changed

src/mlpack/methods/ann/layer/c_relu.hpp
jeffin143: @zoq , Sorry but i couldn't get it, I mean i have taken reference from your comment above, Could you just brief me what is the issue with the format

src/mlpack/methods/ann/layer/c_relu.hpp
jeffin143: Removed

src/mlpack/methods/ann/layer/c_relu.hpp
jeffin143: No, I have removed it.

src/mlpack/tests/activation_functions_test.cpp
jeffin143: @zoq , I just followed the pattern , other test had similar pattern so i just followed it , But i agree you with , we could create only one object and do the required , I made apt changes.

src/mlpack/methods/ann/layer/c_relu.hpp
zoq: Sure, there should be a space right before and after the `=`. Also it would be great if we could align each `=`, we can just add spaces to make it match. I think that improves the overall readability.

jeffin143: @zoq , made changes , Sorry but i didn't know that the style guide was followed for this too, since the dropconnect.hpp didn't follow such, that is why i couldn't get the mistake first hand.

src/mlpack/methods/ann/layer/c_relu_impl.hpp
```diff
@@ -0,0 +1,59 @@
...
+  // Nothing to do here.
+}
+
+template<typename InputDataType, typename OutputDataType>
+template<typename InputType, typename OutputType>
+void CReLU<InputDataType, OutputDataType>::Forward(
+    const InputType&& input, OutputType&& output)
+{
+  output = arma::join_cols(arma::max(input, 0.0 * input), arma::max(
+        (-1 * input), 0.0 * input));
```
zoq: This should be 4 spaces, see https://github.com/mlpack/mlpack/wiki/DesignGuidelines#line-length-and-wrapping for more information.

src/mlpack/methods/ann/layer/c_relu_impl.hpp
```diff
@@ -0,0 +1,59 @@
...
+
+template<typename InputDataType, typename OutputDataType>
+template<typename DataType>
+void CReLU<InputDataType, OutputDataType>::Backward(
+    const DataType&& input, DataType&& gy, DataType&& g)
+{
+  DataType temp;
+  temp = gy % (input >= 0.0);
+  g = temp.rows(0, (input.n_rows / 2 - 1)) - temp.rows(input.n_rows / 2,
+                                            (input.n_rows - 1));
```
zoq: See comment above.

jeffin143: @zoq , Made changes at both the places.

zoq:
Thanks for the contribution.

Issue #1840, opened 2019-04-05
crash due to not checking eig_sym() for failure in PositiveDefiniteConstraint::ApplyConstraint()


conradsnicta:
`arma::eig_sym()` is not checked for failure in `PositiveDefiniteConstraint::ApplyConstraint()`.
This can lead to a crash due to accessing non-existent eigenvalues.

mlpack/src/mlpack/methods/gmm/positive_definite_constraint.hpp
https://github.com/mlpack/mlpack/blob/2326bf8bd0b380c17eed6b5b1c34fbfab1563774/src/mlpack/methods/gmm/positive_definite_constraint.hpp#L44

The code in `PositiveDefiniteConstraint::ApplyConstraint()` assumes that the `eigval` vector has at least one element, which is a false assumption. If `arma::eig_sym()` fails, `eigval` is reset so that it has no elements -- see the [eig_sym() documentation](http://arma.sourceforge.net/docs.html#eig_sym).

Decomposing a non-symmetric matrix with `arma::eig_sym()` is not guaranteed to work. Furthermore, Armadillo 9.400 will have a rudimentary check for symmetry on the input matrix, which makes failure more likely here.

conradsnicta:
Similar problems with eig_sym():
* https://github.com/mlpack/mlpack/blob/master/src/mlpack/core/math/lin_alg.cpp#L88
* https://github.com/mlpack/mlpack/blob/master/src/mlpack/methods/gmm/eigenvalue_ratio_constraint.hpp#L67
* https://github.com/mlpack/mlpack/blob/master/src/mlpack/methods/kernel_pca/kernel_rules/nystroem_method.hpp#L67
* https://github.com/mlpack/mlpack/blob/master/src/mlpack/methods/kernel_pca/kernel_rules/naive_method.hpp#L76
* https://github.com/mlpack/mlpack/blob/master/src/mlpack/tests/gmm_test.cpp#L629

Similar problems with chol():
* https://github.com/mlpack/mlpack/blob/master/src/mlpack/core/dists/gaussian_distribution.cpp#L42
* https://github.com/mlpack/mlpack/blob/master/src/mlpack/methods/gmm/gmm.cpp#L120

chandramoulirajagopalan:
Can I work on this issue ?

zoq:
Sure, feel free to work on the issue.

chandramoulirajagopalan:
[...]

PR #1845, opened 2019-04-07
Fixing the Issue #1840 by a conditional statement to check if eigen values exist or not.

chandramoulirajagopalan:


conradsnicta:
copied from comment https://github.com/mlpack/mlpack/issues/1840#issuecomment-480706675

@chandramoulirajagopalan You may have used an old version of mlpack as the base for your work. For example, why are you modifying the following files ?
src/mlpack/methods/ann/layer/lstm.hpp  
src/mlpack/methods/ann/layer/lstm_impl.hpp  
src/mlpack/tests/ann_layer_test.cpp  
src/mlpack/methods/reinforcement_learning/q_learning_impl.hpp  

Also, it's not clear why you have `#include <iostream>`, and use the `cerr` stream. mlpack already has its own specific streams for logging, such as `Log::Fatal`, `Log::Debug`, `Log::Info`, `Log::Warn`. Example:

    Log::Warn << "message" << std::endl;

PR #1847, opened 2019-04-07
Rectified Predict() in rnn_impl (#1846).

### LINKED ISSUE #1846: RNN: Fails to run Predict() if called before Train().
#### Issue description
```c++
  const size_t rho = 5;
  arma::cube questions = arma::ones(1,2,rho);
  arma::cube answers = arma::ones(1,2,rho);
  
  mlpack::ann::RNN<mlpack::ann::MeanSquaredError<> > model(rho);
  model.Add<mlpack::ann::Linear<> >(1, 10);
  model.Add<mlpack::ann::LSTM<> >(10,10);
  model.Add<mlpack::ann::Linear<> >(10, 1);
  
  model.Parameters().randu();
  model.Predict(questions, answers);
```
Running the above snippet produces :
```bash
error: Mat::submat(): indices out of bounds or incorrectly used
    what():  Mat::submat(): indices out of bounds or incorrectly used
  Aborted (core dumped)
```
However, if the last line in the above snippet is replaced by:
```c++
model.Train(questions, answers);
model.Predict(questions, answers);
```
It works perfectly. 
#### Steps to reproduce

Running the [code](https://pastebin.com/LNafNMUu).

#### Expected behaviour

RNN class methods Predict() and Train() (actually Evalute()) should be consistent.

#### Actual behavior

Fails to run Predict() if called before Train()/ Evaluate().

MuLx10:
```c++
  const size_t rho = 5;
  arma::cube questions = arma::ones(1,2,rho);
  arma::cube answers = arma::ones(1,2,rho);
  
  mlpack::ann::RNN<mlpack::ann::MeanSquaredError<> > model(rho);
  model.Add<mlpack::ann::Linear<> >(1, 10);
  model.Add<mlpack::ann::LSTM<> >(10,10);
  model.Add<mlpack::ann::Linear<> >(10, 1);
  
  model.Parameters().randu();
  model.Predict(questions, answers);
```
Running the above snippet produces :
```bash
error: Mat::submat(): indices out of bounds or incorrectly used
    what():  Mat::submat(): indices out of bounds or incorrectly used
  Aborted (core dumped)
```
However, if the last line in the above snippet is replaced by:
```c++
model.Train(questions, answers);
model.Predict(questions, answers);
```
It works perfectly. 
Evalute method sets RNN::outputSize. When predict is called before Train() RNN::outputSize is 0. 
Thus Mat::submat() fails.This PR makes Predict and Evaluate method of class RNN consistent by adding a test and set for RNN::outputSize .
Closes #1846 .

### DIFF
--- a/src/mlpack/methods/ann/rnn_impl.hpp
+++ b/src/mlpack/methods/ann/rnn_impl.hpp
@@ -153,19 +153,25 @@ void RNN<OutputLayerType, InitializationRuleType, CustomLayers...>::Predict(
     ResetDeterministic();
   }
 
-  arma::mat resultsTemp;
-  Forward(std::move(arma::mat(predictors.slice(0).colptr(0),
-      predictors.n_rows, 1, false, true)));
-  outputSize = boost::apply_visitor(outputParameterVisitor,
-      network.back()).col(0).n_elem;
+  const size_t effectiveBatchSize = std::min(batchSize,
+      size_t(predictors.n_cols));
 
+  Forward(std::move(arma::mat(predictors.slice(0).colptr(0),
+      predictors.n_rows, effectiveBatchSize, false, true)));
+  arma::mat resultsTemp = boost::apply_visitor(outputParameterVisitor,
+      network.back());
+
+  outputSize = resultsTemp.n_rows;
   results = arma::zeros<arma::cube>(outputSize, predictors.n_cols, rho);
+  results.slice(0).submat(0, 0, results.n_rows - 1,
+      effectiveBatchSize - 1) = resultsTemp;
+
   // Process in accordance with the given batch size.
   for (size_t begin = 0; begin < predictors.n_cols; begin += batchSize)
   {
     const size_t effectiveBatchSize = std::min(batchSize,
         size_t(predictors.n_cols - begin));
-    for (size_t seqNum = 0; seqNum < rho; ++seqNum)
+    for (size_t seqNum = !begin; seqNum < rho; ++seqNum)
     {
       Forward(std::move(arma::mat(predictors.slice(seqNum).colptr(begin),
           predictors.n_rows, effectiveBatchSize, false, true)));
--- a/src/mlpack/methods/ann/rnn_impl.hpp
+++ b/src/mlpack/methods/ann/rnn_impl.hpp
@@ -153,6 +153,12 @@ void RNN<OutputLayerType, InitializationRuleType, CustomLayers...>::Predict(
     ResetDeterministic();
   }
 
+  arma::mat resultsTemp;
+  Forward(std::move(arma::mat(predictors.slice(0).colptr(0),
+      predictors.n_rows, 1, false, true)));
+  outputSize = boost::apply_visitor(outputParameterVisitor,
+      network.back()).col(0).n_elem;
+
   results = arma::zeros<arma::cube>(outputSize, predictors.n_cols, rho);
   // Process in accordance with the given batch size.
   for (size_t begin = 0; begin < predictors.n_cols; begin += batchSize)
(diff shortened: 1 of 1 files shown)


src/mlpack/methods/ann/rnn_impl.hpp
```diff
@@ -164,9 +164,16 @@ void RNN<OutputLayerType, InitializationRuleType, CustomLayers...>::Predict(
...
           predictors.n_rows, effectiveBatchSize, false, true)));
 
-      results.slice(seqNum).submat(0, begin, results.n_rows - 1, begin +
-          effectiveBatchSize - 1) = boost::apply_visitor(outputParameterVisitor,
+      arma::mat out = boost::apply_visitor(outputParameterVisitor,
           network.back());
+
+      if (results.n_rows == 0)
+      {
+        results.set_size(out.n_rows, predictors.n_cols, rho);
```
saksham189: I am not in strong favour of this change. We already check and set `outputSize` inside `EvaluateWithGradient` and `Evaluate` methods and adding it inside `Predict` makes it redundant in the code + it is inefficient to have it within 2 for loops here. 
Rather I think we could set `outputSize` once inside the `ResetParameters()` method by creating a `OutputSizeVisitor` and reverse iterating over the layers until we have a non-zero value. This way we can remove the need to check `outputSize == 0` inside every method.
I think we are doing something similar here https://github.com/mlpack/mlpack/pull/1366 which has an implementation of the `outputSizeVisitor`.
Let me know what you think. @MuLx10 @zoq

MuLx10: @saksham189 I agree that checking redundant conditions within two loops is very inefficient. I will rectify that. However, I am not sure that I should remove that from other functions like Evaluate().  @zoq I would like to have your views.

zoq: Sounds like a good idea to me, @MuLx10 feel free to remove the part from the other functions as well.

MuLx10: okay sure

zoq:
@MuLx10 Let me know if I should take a look into the build issue.

MuLx10:
@zoq I figured out the problem. However, I am having  difficulty finding the solution

zoq:
Happy to help, perhaps you can pinpoint me to the issue?

src/mlpack/methods/ann/rnn_impl.hpp
```diff
@@ -153,6 +153,12 @@ void RNN<OutputLayerType, InitializationRuleType, CustomLayers...>::Predict(
     ResetDeterministic();
   }
 
+  arma::mat resultsTemp;
+  Forward(std::move(arma::mat(predictors.slice(0).colptr(0),
```
zoq: What about we move this into the for loop and set the site of `result` at the first step?

MuLx10: @zoq  that's what I did in the [previous commit](https://github.com/mlpack/mlpack/pull/1847/commits/dc3cbf46d95e9a940c6a4510f6ff45e4e54c63c7).

zoq: Ahh, right, I think in this case it's faster to check this at the first step once, instead of doing a complete forward pass; or what about we start with ``begin = 1`` (do the first step outside of the for loop). Let me know what you think.

MuLx10: Yes, that might work. Thanks @zoq

MuLx10:
@mlpack-jenkins

saksham189:
Thanks for fixing this @MuLx10 . I got a bit busy with exams. This looks good to me. I will try and implement my idea in a separate PR.

zoq:

MuLx10:
@saksham189 Yeah sure. @zoq Thanks.

PR #1850, opened 2019-04-08
Fixing crash due to not checking eig_sym() #1840

chandramoulirajagopalan:
Fixing the Issue #1840 by a conditional statement to check if eigen values exist or not.


conradsnicta:
Code still has problems. See https://github.com/mlpack/mlpack/issues/1840#issuecomment-480798003

Issue #1849, opened 2019-04-08
bug in ccov() due to use of internal armadillo classes


conradsnicta:
The `ccov()` function in `mlpack/src/mlpack/core/arma_extend/fn_ccov.hpp` uses internal armadillo classes: `Op` and `Glue`:
https://github.com/mlpack/mlpack/blob/3ad649c7f780abf87193859732f866f2119093ac/src/mlpack/core/arma_extend/fn_ccov.hpp#L8
https://github.com/mlpack/mlpack/blob/3ad649c7f780abf87193859732f866f2119093ac/src/mlpack/core/arma_extend/fn_ccov.hpp#L22

In armadillo 9.400 (under development) these classes are changed, resulting in the following errors when compiling mlpack:

    /usr/include/armadillo_bits/Op_bones.hpp:53:21: error: no class template named ‘traits’ in ‘class arma::op_ccov’

`op_ccov` is defined by mlpack in `src/mlpack/core/arma_extend/op_ccov_proto.hpp`.

I suggest rewriting both forms of `ccov()` so that the functions output a matrix instead of `Op` or `Glue`. When compiling with a C++11 compiler, Armadillo will automatically use the move constructor in the `Mat` class, so the performance impact will be minimal.

The internal `Op` and `Glue` classes are subject to change, as per the [API policy](http://arma.sourceforge.net/docs.html#api_additions).

chandramoulirajagopalan:
Can I work on this issue?

conradsnicta:
@chandramoulirajagopalan - I'd suggest that it's better to finish one PR properly (ie. your work on https://github.com/mlpack/mlpack/issues/1840), and then look at making another one. You seem to be working too fast, which can lead to easy mistakes.

conradsnicta:
@rcurtin I'm too busy make a proper PR, so here is the next best thing.
Attached is `ccov.txt` (because GitHub won't let me attach .hpp files): [ccov.txt](https://github.com/mlpack/mlpack/files/3054801/ccov.txt)
The code is an adapted form of `core/arma_extend/*ccov*.hpp` so that it doesn't use internal armadillo functionality.

The file should be renamed to `ccov.hpp` and placed into `src/mlpack/core/math/`. Then throughout the mlpack codebase, all uses of `arma::ccov(X)` should then be replaced with `mlpack::math::ccov(X)`.

Feel free to adapt, change, etc. License is 3-clause BSD license.

[...]

PR #1855, opened 2019-04-09
refactor ccov() to avoid use of internal armadillo classes

jeffin143:
#1849 Relevant

@conradsnicta @rcurtin , Couldn't resist working on it


src/mlpack/tests/gmm_test.cpp
```diff
@@ -198,7 +200,9 @@ BOOST_AUTO_TEST_CASE(GMMTrainEMMultipleGaussians)
       // Calculate the actual means and covariances because they will probably
       // be different (this is easier to do before we shuffle the points).
       means[i] = arma::mean(data.cols(point, point + counts[i] - 1), 1);
-      covars[i] = ccov(data.cols(point, point + counts[i] - 1), 1 /* biased */);
+      arma::uword norm_type = 1;
+      arma::mat sub = data.cols(point, point + counts[i] - 1);
```
jeffin143: No overload function for `ccov(subview<double> ,int )` and hence had to type case it.

jeffin143:
will fix style issue once #1853 get merged

src/mlpack/tests/gmm_test.cpp
conradsnicta: looks ok

conradsnicta: a slightly more compact way might be to do convert to mat directly in the ccov() call (as shown below), but either way works

    covars[i] = mlpack::math::ccov( mat(data.cols(point, point + counts[i] - 1)), 1 /* biased */);

conradsnicta:
PR looks fine by me, though would probably need to get the okay from @zoq or @rcurtin

src/mlpack/core/math/ccov.hpp
```diff
@@ -0,0 +1,242 @@
+/**
+ * @file ccov.hpp
+ * @author Ryan Curtin
+ * @author Conrad Sanderson
+ *
+ * ccov(X) is same as cov(trans(X)) but without the cost of computing trans(X)
```
zoq: mlpack does use a slightly different coding style as armadillo and since this is no armadillo backport, can you go through the style guidelines and adjust the style.

jeffin143: @zoq , Made appropriate changes.

src/mlpack/core/arma_extend/arma_extend.hpp
```diff
@@ -55,12 +52,6 @@ namespace arma {
   // u64/s64
   #include "hdf5_misc.hpp"
 
-  // ccov()
-  #include "op_ccov_proto.hpp"
-  #include "op_ccov_meat.hpp"
-  #include "glue_ccov_proto.hpp"
-  #include "glue_ccov_meat.hpp"
-  #include "fn_ccov.hpp"
```
rcurtin: Agreed, I think dropping `ccov()` from `arma_extend` is the right thing to do here.

src/mlpack/core/math/ccov.hpp
```diff
@@ -0,0 +1,255 @@
...
+#ifndef MLPACK_CORE_MATH_CCOV_HPP
+#define MLPACK_CORE_MATH_CCOV_HPP
+
+namespace mlpack {
+namespace math /** Miscellaneous math routines. */ {
+
+template<typename eT>
+inline
+arma::Mat<eT>
+ccov(const arma::Mat<eT>& A, const arma::uword norm_type = 0)
```
rcurtin: Since we are making this a part of the mlpack codebase now, can we name this `ColumnCovariance()` instead of `ccov()` for consistency?  Also we can switch `uword` to `size_t` (again for consistency).

src/mlpack/core/math/ccov.hpp
```diff
@@ -0,0 +1,255 @@
...
+}
+
+
+
+template<typename eT>
+inline
+arma::Mat<eT>
+ccov(const arma::Mat<eT>& A,
+     const arma::Mat<eT>& B,
+     const arma::uword norm_type = 0)
```
rcurtin: I don't think we ever use this overload anywhere, so I think that we could just remove it.  I probably wrote this so it would match the functionality of `cov()` but I don't think we need that now.

src/mlpack/core/math/ccov.hpp
```diff
@@ -0,0 +1,255 @@
...
+}
+
+
+
+template<typename T>
+inline
+arma::Mat< std::complex<T> >
+ccov(const arma::Mat< std::complex<T> >& A,
+     const arma::Mat< std::complex<T> >& B,
+     const arma::uword norm_type = 0)
```
rcurtin: Same for this one.

src/mlpack/tests/distribution_test.cpp
```diff
@@ -15,6 +15,7 @@
  * http://www.opensource.org/licenses/BSD-3-Clause for more information.
  */
 #include <mlpack/core.hpp>
+#include <mlpack/core/math/ccov.hpp>
```
rcurtin: I would say, let's include `ccov.hpp` in `mlpack/prereqs.hpp` because it would have previously been included through `arma_extend/`.

src/mlpack/core/math/ccov.hpp
jeffin143: @rcurtin , Done

src/mlpack/core/math/ccov.hpp
jeffin143: @rcurtin , Removed.

src/mlpack/core/math/ccov.hpp
jeffin143: :+1:

src/mlpack/tests/distribution_test.cpp
jeffin143: Added to `mlpack/prereqs.hpp`

jeffin143:
Screenshot from 2019-04-11 19-12-14
@rcurtin , There is some serious issue , I am not able to debug it, 

Log::Fatal is throwing error in unchanged files, may be due to some issues. Tried a lot to trace down the flow and understand what is the issue, but couln't get it.

Log::fatal is throwing error in `/mlpack/core/data/load_model_impl.hpp:55:9` , file i haven't even touched .

src/mlpack/tests/gmm_test.cpp
```diff
@@ -855,7 +858,8 @@ BOOST_AUTO_TEST_CASE(DiagonalGMMTrainEMOneGaussian)
 
     arma::vec actualMean = arma::mean(data, 1);
     arma::vec actualCovar = arma::diagvec(
-        arma::ccov(data, 1 /* biased estimator */));
+        mlpack::math::ColumnCovariance(data,
+        arma::size_t(1) /* biased estimator */));
```
conradsnicta: `arma::size_t` ? did you mean `std::size_t` ?

src/mlpack/core/math/ccov.hpp
```diff
@@ -0,0 +1,111 @@
...
+#include <mlpack/core.hpp>
+#include <mlpack/core/util/log.hpp>
+
+namespace mlpack {
+namespace math /** Miscellaneous math routines. */ {
+
+template<typename eT>
+inline
+arma::Mat<eT>
+ColumnCovariance(const arma::Mat<eT>& A, const arma::size_t norm_type = 0)
```
conradsnicta: `std::size_t`

src/mlpack/tests/gmm_test.cpp
jeffin143: @conradsnicta , Sorry for the big blunder , Shouldn't have used sublime feature :)

src/mlpack/core/math/ccov.hpp
jeffin143: @conradsnicta , Done with the correction, I hope I haven't made any more errors.

src/mlpack/tests/gmm_test.cpp
```diff
@@ -11,7 +11,6 @@
  * http://www.opensource.org/licenses/BSD-3-Clause for more information.
  */
 #include <mlpack/core.hpp>
-#include <mlpack/core/math/ccov.hpp>
```
jeffin143: Adding this line throws unexpected error , snap shot of which i have added. If i remove this Line the build is successful.

jeffin143: Also after long hours of research, I found that `make -j1` throws error, `make -j4` works perfctly alright that means some issue with the order of compilation, but still i am unable to reach a concrete reason.

jeffin143: `ColumnCovariance()`, need `Log::Fatal` and hence include, `mlpack/prereqs.hpp` and thus we cannot inlcude `ccov.hpp` in `mlpack/prereqs.hpp` because it throws an error 

`make -j4` runs parallely and would manage to compile the files in correct order, but while going with `make -j1`, the build fails since it disturbs the order of compilation.

Therfore i have added it in `mlpack/core.hpp`, Along with other `mlpack/data/math/*` , I guess it won't make much difference since core is included at all places and thus, it would be ok to include the file through `core.hpp`

rcurtin: Ah, you are right, sorry---`core.hpp` would be where we have to put it.  Nice debugging :+1:

conradsnicta:
@jeffin143 Attached is a faster implementation of ColumnCovariance(): [faster_ccov.txt](https://github.com/mlpack/mlpack/files/3083855/faster_ccov.txt)

jeffin143:
> @jeffin143 Attached is a faster implementation of ColumnCovariance(): [faster_ccov.txt](https://github.com/mlpack/mlpack/files/3083855/faster_ccov.txt)

thanks @conradsnicta , Made changes :)

rcurtin:
@jeffinsam @conradsand thanks so much! :+1:

PR #1856, opened 2019-04-09
refactor ccov() to avoid use of Op and Glue (for mlpack-3.0.x - lightweight version)

conradsnicta:
This PR is intended for the mlpack-3.x branch, targeting the next patch release (eg. mlpack 3.0.5).
It addresses issue https://github.com/mlpack/mlpack/issues/1849.
It is related to @jeffin143's PR https://github.com/mlpack/mlpack/pull/1855 which is targeting the master branch.

The present PR is less ambitious, as it removes the use of Op and Glue classes but retains use of other internal Armadillo functionality.  As such, this PR is _not_ meant to replace https://github.com/mlpack/mlpack/pull/1855. Instead, it is intended only for the next patch release in the 3.0.x series, keeping changes to the bare minimum.

@rcurtin @zoq


rcurtin:
Note also the discussion of how this affects the release at the bottom of #1846.

rcurtin:
Wait, hang on, this was the wrong PR...

jeffin143:
> Wait, hang on, this was the wrong PR...

Wrong PR, in what Sense ?

rcurtin:
Sorry, I should have actually just deleted the comment.  My error was that I meant to make that approval comment on #1855 but did it here instead. :+1:

rcurtin:
Closed in favor of #1855.  @conradsnicta thanks for your help with this.

PR #1862, opened 2019-04-14
Fixing crash due to not checking eig_sym() by writing a boolean statement. Issue #1840.

chandramoulirajagopalan:


jeffin143:
@chandramoulirajagopalan , isn't #1896 solving the same problem if so , you can keep any one of this open , There are lot of PR , and hence we can do liltle to manage them.

PR #1864, opened 2019-04-16
Python fixes for Travis build

rcurtin:
I was surprised to see that after merging #1816, the Travis build broke.  I was able to reproduce it locally, and after spending some time digging into it, I managed to fix three problems:

 1. Avoid the use of `ravel()` and `reshape()` in `print_input_processing.hpp`.  This is because those methods *may* sometimes copy the matrix, and when this gets passed to C++, the copied memory goes out of scope and there are segfault issues.

 2. Simplify `to_matrix()` handling of Pandas issues to avoid use of `flatten()`, which has to the same problem: `flatten()` may return an object with new memory, which immediately goes out of scope after the `to_matrix()` call, which then causes later allocation/deallocation problems.

 3. Fix a couple tests, which were testing against the wrong matrix's value.

This fix is necessary before 3.1.0, since otherwise we'd be shipping with pretty broken Python support.  I'm still confused why the tests passed before #1816 was merged, but it should work now...

@Yashwants19: take a look and let me know if you disagree with any of the changes.  I've changed some of the code you wrote in previous PRs.  Thanks!

### DIFF
--- a/src/mlpack/bindings/python/print_input_processing.hpp
+++ b/src/mlpack/bindings/python/print_input_processing.hpp
@@ -62,8 +62,6 @@ void PrintInputProcessing(
    *   else:
    *     raise TypeError("'param_name' must have type 'list'!")
    */
-
-
   std::cout << prefix << "# Detect if the parameter was passed; set if so."
       << std::endl;
   if (!d.required)
@@ -190,7 +188,6 @@ void PrintInputProcessing(
    *      raise TypeError("'param_name' must have type 'list'!")
    *
    */
-
   std::cout << prefix << "# Detect if the parameter was passed; set if so."
       << std::endl;
   if (!d.required)
@@ -259,12 +256,9 @@ void PrintInputProcessing(
    *   param_name_tuple = to_matrix(param_name)
    *   if param_name_tuple[0].shape[0] == 1 or
    *       param_name_tuple[0].shape[1] == 1:
-   *     param_name_reshape = param_name_tuple[0].ravel()
-   *     param_name_mat = arma_numpy.numpy_to_mat_s(param_name_reshape,
-   *         param_name_tuple[1])
-   *   else:
-   *     param_name_mat = arma_numpy.numpy_to_mat_s(param_name_tuple[0],
-   *         param_name_tuple[1])
+   *     param_name_tuple[0].shape = (param_name_tuple[0].size,)
+   *   param_name_mat = arma_numpy.numpy_to_mat_s(param_name_reshape,
+   *       param_name_tuple[1])
    *   SetParam[mat](<const string> 'param_name', dereference(param_name_mat))
    *   CLI.SetPassed(<const string> 'param_name')
    *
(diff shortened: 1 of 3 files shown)


src/mlpack/bindings/python/print_input_processing.hpp
```diff
@@ -259,12 +256,9 @@ void PrintInputProcessing(
    *   param_name_tuple = to_matrix(param_name)
    *   if param_name_tuple[0].shape[0] == 1 or
    *       param_name_tuple[0].shape[1] == 1:
-   *     param_name_reshape = param_name_tuple[0].ravel()
-   *     param_name_mat = arma_numpy.numpy_to_mat_s(param_name_reshape,
```
Yashwants19: I think so it must be `param_name_tuple[0]`

src/mlpack/bindings/python/print_input_processing.hpp
```diff
@@ -306,16 +296,11 @@ void PrintInputProcessing(
...
-      std::cout << prefix << "  " << prefix << d.name
-          << "_reshape = np.reshape(("<< d.name << "_tuple[0]), (" << d.name
-          << "_tuple[0].shape[0] , 1))" << std::endl;
-      std::cout << prefix << "    " << d.name << "_mat = arma_numpy.numpy_to_"
-          << GetArmaType<T>() << "_" << GetNumpyTypeChar<T>() << "(" << d.name
-          << "_reshape, " << d.name << "_tuple[1])" << std::endl;
-      std::cout << prefix << "  " << "else:" << std::endl;
-      std::cout << prefix << "    " << d.name << "_mat = arma_numpy.numpy_to_"
+      std::cout << prefix << "    " << d.name << "_tuple[0].shape = (" << d.name
+          << "_tuple[0].shape[0], 1)" << std::endl;
```
Yashwants19: Hi @rcurtin I think so we should also add the same for `to_matrix_with_info()`

Yashwants19: `matrix_with_info` is also using `reshape` for casting 1-D to 2-D matrix

rcurtin: You're right, just haven't had a chance to get to it yet.  I'll address your comments tonight---thanks! :+1:

rcurtin:
Awesome, I think I have everything worked out here and it's ready for merge. :+1:  Thanks @Yashwants19 for the comments.

src/mlpack/bindings/python/tests/test_python_binding.py
```diff
@@ -192,7 +192,7 @@ def testPandasSeriesMatrix(self):
     Test that we can pass pandas.Series as input parameter.
     """
     x = pd.Series(np.random.rand(100))
-    z = copy.copy(x)
```
MuLx10: `copy.copy` does a shallow copy. `copy.deepcopy` is needed for the above. I guess may  be that was the reason.

MuLx10: All set. Good to go.

rcurtin: Thanks, I must have misread the documentation.  Fixed now. :+1:

Issue #1861, opened 2019-04-13
Extracting learned parameters from a QLearning instance


cascala:
Hey folks,

How does one extract the learned parameters from a QLearning instance?

The obvious things don't seem to help me:
* The learningNetwork variable in q_learning.hpp is private...
* and there is no accessor function either

Is there a public member function that I can use?

Thank you!

zoq:
Good point, I think providing a function to access the learned network should solve the issue. I can open a PR, unless you already have something ready?

cascala:
@zoq 
Let me have a look; I will send over a pull request later this week.

Is there any other agent that already does this (so I can mimic the same code and naming style)?

If not, would you prefer that I add an accessor to the parameters only (return type arma::mat), or a copy of the entire trained model (with layers etc.) (return type NetworkType).

zoq:
I would go with the a getter/setter for the complete network, afterwards a user could do `LearningNetwork().Parameters()` to e.g. get the learned parameter. Here is an example:

https://github.com/mlpack/mlpack/blob/00d07ea35a396dfc80f38b4ba6f18b4b6bc4f77e/src/mlpack/methods/ann/gan/gan.hpp#L289-L292

Let me know what you think.

cascala:
Addressed in pull request #1865

rcurtin:
Let's not close the issue until the fix is merged.

zoq:
Fixed in https://github.com/mlpack/mlpack/pull/1865.

PR #1865, opened 2019-04-16
Adding accessors for the network

cascala:
Adding accessors for the network, so users can retrieve the q-learned parameters
Function naming is consistent with AsyncLearning

This is related to #1861


zoq:
Thanks again.

PR #1869, opened 2019-04-19
build using gold linker

jeffin143:
#722 relevant  , Just testing for travis cli build , Reduced significantly in Local System 

Not Ready for Review, Would close it, If it is not improved


jeffin143:
I would be working on it, forever :)

shrit:
@jeffin143 Are you still working on this?
This pull request can improve dramatically the linking time for mlpack by using the golden linker if it exists on the system.
Can I open it ?

jeffin143:
@shrit Please feel free to open a pr , I tried it and it did reduce significantly but then when in Travis ci , it didn't make much difference

CMakeLists.txt
```diff
@@ -168,6 +168,30 @@ if(BUILD_WITH_COVERAGE)
...
+endif()
+
+# Configure CCache if available
+find_program(CCACHE_FOUND ccache)
+if(CCACHE_FOUND)
+        message(STATUS "CCache Found")
+        set_property(GLOBAL PROPERTY RULE_LAUNCH_COMPILE ccache)
+        set_property(GLOBAL PROPERTY RULE_LAUNCH_LINK ccache)
+endif(CCACHE_FOUND)
+
```
shrit, line 192:
```suggestion

```

CMakeLists.txt
```diff
@@ -168,6 +168,30 @@ if(BUILD_WITH_COVERAGE)
...
+
```
shrit, line 193:
```suggestion

```

shrit (commented):
Everything looks fine to me, let us see if it passes the build and what @rcurtin think if there is anything else to add here

jeffin143:
@shrit , do you have the permission to push directly to this branch ? , If you need let me know , also did you try some benchmarks ??

shrit:
@jeffin143 I did not try any benchmark. Golden linker should improve linking.
I think I have the permission to push to your branch directly, by the way, you have done very good work we might not need to do any major changes

rcurtin:
I tried the PR locally building everything. :smile:

```
cmake
    -DBUILD_TESTS=ON
    -DBUILD_CLI_EXECUTABLES=ON
    -DBUILD_PYTHON_BINDINGS=ON
    -DBUILD_GO_BINDINGS=ON
    -DBUILD_JULIA_BINDINGS=ON
    -DJULIA_EXECUTABLE=/opt/julia/bin/julia
    -DBUILD_MARKDOWN_BINDINGS=ON
    -DBUILD_R_BINDINGS=ON
    ..
```

*master* (no gold):

```
real    50m7.983s
user    187m25.218s
sys     6m56.691s
```

*clbuild* (gold):

```
real    51m19.620s
user    190m21.186s
sys     8m0.163s
```

So, apparently no real difference on my system?  (The slight difference is probably noise.)  Can anyone else see any difference for the main build?

I also tried using `gold` for the LSTM stock prediction example: https://raw.githubusercontent.com/mlpack/examples/master/lstm_stock_prediction/lstm_stock_prediction.cpp

*With regular ld*:

```
$ time g++ -O3 -DNDEBUG -Iinclude/ -Llib/ -o lstm_stock_prediction lstm_stock_prediction.cpp -lmlpack -larmadillo -Ideps/ensmallen-2.14.2/include/ -fopenmp
real    1m28.862s
user    1m25.531s
sys     0m1.945s
```

*With gold*:

```
$ time g++ -fuse-ld=gold -O3 -DNDEBUG -Iinclude/ -Llib/ -o lstm_stock_prediction lstm_stock_prediction.cpp -lmlpack -larmadillo -Ideps/ensmallen-2.14.2/include/ -fopenmp

real    1m27.487s
user    1m24.824s
sys     0m1.852s
```

So, maybe this is specific to my system or something?  I know that gold has been benchmarked very well against ld.

(Also, holy crap that takes a long time!  Has gcc gotten slower over time?  Is this our codebase?)

shrit:
@rcurtin I think you need to enable these flags in order to allow multithreading in golden linker:
`-Wl,--threads -Wl,--thread-count,$(nproc)`
I agree that it is hard to see the difference in the examples, but it should be possible on the entire mlpack project.

shrit:
@rcurtin Yes, of course, I will try a local build with this patch and the thread flags.
We will see of these flags will make any considerable differences in compile time.

shrit:
@rcurtin I have downloaded @jeffin143 mlpack branch and added the necessary flags for multi-thread linking.
Here are my numbers:
gold 16 thread
```
make -j16  7445.27s user 308.61s system 1216% cpu 10:37.56 total
make -j16  7449.23s user 300.87s system 1149% cpu 11:14.49 total
```
gold no threading:
```
make -j16  7446.31s user 283.84s system 1159% cpu 11:06.54 total
make -j16  7440.37s user 297.76s system 1198% cpu 10:45.64 total
make -j16  7440.91s user 295.52s system 1160% cpu 11:06.38 total
make -j16  7437.41s user 292.77s system 1184% cpu 10:52.62 total
```
I am adding several tests to have at least have an estimation of the nuances.
I think that the reason behind the segfault on your machine is related to the support of multi-threading. I am afraid that the gold linker installed by default on fedora might not have the support for multi-threading.
Finally, I am not sure that it is improving the linking time,  so either most of the time is wasted on the compilation rather than the linking or the multi-threading linking is not working.

rcurtin:
@shrit hmm, interesting observations.  I think you might be right that most of the time is spent on compilation.  It is possible to benchmark this, if you wanted.  You can use `-ftime-report` in the `CXXFLAGS`; but that will print results for every single file that is compiled.  Still, that can give a pretty good idea of where things are slow.  Last time I did that, a few years ago, it seemed like a good proportion of the time was being spent in the parser and other things related to including lots of other files.  However, I believe that was before we had cotire support, so it could be really different now.  https://stackoverflow.com/questions/3025443/how-to-calculate-gcc-compilation-time

@jeffin143 do you have any experiments showing that gold did give a significant speedup?  If you don't, and we can't seem to get any reliable speedup out of it, maybe we should close this PR and look for improvements elsewhere. :(

jeffin143:
@shrit 
Do you want this to stay open or are we dropping this off ?

shrit:
@jeffin143 I do not see any difference in compile-time since in mlpack most of the compilation time is related to compile-time in g++ and not to linking time.
especially when using neural networks, boost visitor is taking most of it.
We can keep it close, if one-day things change, we can always open it

PR #1870, opened 2019-04-20
Fixed gravity constant

abhinavsagar:


src/mlpack/methods/reinforcement_learning/environment/pendulum.hpp
```diff
@@ -120,7 +120,7 @@ class Pendulum
     double angularVelocity = state.AngularVelocity();
 
     // Define constants which specify our pendulum.
-    const double gravity = 10.0;
+    const double gravity = 9.8;
```
zoq: We tried to replicate the OpenAI gym framework here, which uses 10 as well:

https://github.com/openai/gym/blob/89acef6525918a88a355abcd231bae4b872514a3/gym/envs/classic_control/pendulum.py#L32

so personally I would go with the current value. Let me know what you think.

abhinavsagar: Agreed. I am closing this.

PR #1876, opened 2019-04-25
Class for Scaling of features

jeffin143:
I have implemented following classes helpful for [scaling](https://en.wikipedia.org/wiki/Feature_scaling) of features

- MinMax Scaler class

- MaxAbs Scaler class

- Standard Scaler Class

- Mean Normalization Class

- Zca whitening

- Pca Whitening

Scaling to unit length is already there in armadillo and hence we need not implement it here and simpy call `arma::normalize()` to scale it to unit length.


src/mlpack/core/data/scaler_methods/maxabsscaler.hpp
```diff
@@ -0,0 +1,96 @@
...
+ */
+#ifndef MLPACK_CORE_DATA_MAX_ABS_SCALE_HPP
+#define MLPACK_CORE_DATA_MAX_ABS_SCALE_HPP
+
+#include <mlpack/prereqs.hpp>
+
+namespace mlpack {
+namespace data {
+/**
+ * A simple MaxAbs Scaler class
```
zoq: Would be great if we could add some more informations about what this class does. Also what do you think about adding an example. Similar to:
 https://github.com/mlpack/mlpack/blob/681577d8f95c87412cec6d6fbe8802811fb8405e/src/mlpack/core/data/binarize.hpp#L21-L39

src/mlpack/core/data/scaler_methods/maxabsscaler.hpp
```diff
@@ -0,0 +1,96 @@
...
+ public:
+  /**
+  * Default constructor
+  *
+  */
+  MaxAbsScaler(){}
+  /**
+  * Default Destructor 
+  */
+  ~MaxAbsScaler(){}
```
zoq: No need to define the Destructor here if it doesn't deviate from the standard one.

src/mlpack/core/data/scaler_methods/maxabsscaler.hpp
```diff
@@ -0,0 +1,96 @@
...
+
+  /**
+  * Function to scale Features.
+  *
+  * @param input Datset to scale features.
```
zoq: Should this be "Dataset"?

src/mlpack/core/data/scaler_methods/maxabsscaler.hpp
```diff
@@ -0,0 +1,96 @@
...
+  void InverseTransform(arma::Mat<T>& input)
+  {
+    for (size_t i = 0; i < input.n_rows; i++)
+    {
+      for (size_t j = 0; j < input.n_cols; j++)
+      {
+        input(i, j) *= scale(i);
+      }
+    }
+  }
```
zoq: Can you add an empty line here, to improve the readability?

src/mlpack/core/data/scaler_methods/maxabsscaler.hpp
```diff
@@ -0,0 +1,96 @@
...
+  //! Get the Min row vector.
+  const arma::colvec& ItemMin() const { return itemMin; }
+  //! Get the Max row vector.
+  const arma::colvec& ItemMax() const { return itemMax; }
+  //! Get the Scale row vector.
+  const arma::colvec& Scale() const { return scale; }
```
zoq: Can you add an empty line here, to improve the readability?

src/mlpack/core/data/scaler_methods/maxabsscaler.hpp
```diff
@@ -0,0 +1,96 @@
...
+  * Default Destructor 
+  */
+  ~MaxAbsScaler(){}
+
+  /**
+  * Function to scale Features.
+  *
+  * @param input Datset to scale features.
+  */
+  void Transform(arma::Mat<T>& input)
```
zoq: Do you think we should provide an option to save the result into another matrix?

src/mlpack/core/data/scaler_methods/maxabsscaler.hpp
jeffin143: @zoq , Added , PTAL

src/mlpack/core/data/scaler_methods/maxabsscaler.hpp
jeffin143: Removed

src/mlpack/core/data/scaler_methods/maxabsscaler.hpp
jeffin143: Changes , Typo error

src/mlpack/core/data/scaler_methods/maxabsscaler.hpp
jeffin143: Added , Propagated it to every file

src/mlpack/core/data/scaler_methods/maxabsscaler.hpp
jeffin143: I don't mind it, Changed it accordingly

src/mlpack/core/data/scaler_methods/maxabsscaler.hpp
jeffin143: added

src/mlpack/core/data/scaler_methods/maxabsscaler.hpp
```diff
@@ -0,0 +1,108 @@
...
+ * // Scale the features using
+ * MaxAbsScaler<double> scale;
+ * scale.Tranform(input, output);
+ *
+ * // The input can be retransformed using
+ * scale.InverseTransform(output, input);
+ * @endcode
+ *
+ */
+template <typename T>
```
zoq: Looks like the template parameter isn't used, so perhaps it can be removed?

src/mlpack/core/data/scaler_methods/maxabsscaler.hpp
```diff
@@ -0,0 +1,108 @@
...
+ * arma::Mat<double> output;
+ *
+ * // Scale the features using
+ * MaxAbsScaler<double> scale;
+ * scale.Tranform(input, output);
+ *
+ * // The input can be retransformed using
+ * scale.InverseTransform(output, input);
+ * @endcode
+ *
```
zoq: We can remove the extra line here.

src/mlpack/core/data/scaler_methods/maxabsscaler.hpp
```diff
@@ -0,0 +1,108 @@
...
+ * A simple MaxAbs Scaler class.
+ *
+ * Given an input dataset this class helps you to scale each
+ * feature by its maximum absolute value.
+ *
+ * @code
+ * arma::Mat<double> input = loadData();
+ * arma::Mat<double> output;
+ *
+ * // Scale the features using
```
zoq: Hm, this doesn't really follow the style guidelines, since this isn't a complete sentence. I guess in this case we can remove "using" at end with a point.

src/mlpack/core/data/scaler_methods/maxabsscaler.hpp
```diff
@@ -0,0 +1,108 @@
...
+ * MaxAbsScaler<double> scale;
+ * scale.Tranform(input, output);
+ *
+ * // The input can be retransformed using
```
zoq: What about "Retransform the input."?

src/mlpack/core/data/scaler_methods/maxabsscaler.hpp
```diff
@@ -0,0 +1,108 @@
...
+class MaxAbsScaler
+{
+ public:
+  /**
+  * Function to scale Features.
+  *
+  * @param input Dataset to scale features.
+  * @param output Output matrix with scaled features.
+  */
+  void Transform(arma::Mat<T>& input, arma::Mat<T>& output)
```
zoq: What do you think about remove the template class parameter here and instead use:

```cpp
template<typename MatType>
void Transform(const MatType& input, MatType& output)
```

src/mlpack/core/data/scaler_methods/maxabsscaler.hpp
```diff
@@ -0,0 +1,108 @@
...
+  {
+    output.copy_size(input);
+    itemMin = arma::min(input, 1);
+    itemMax = arma::max(input, 1);
+    scale = arma::max(arma::abs(itemMin), arma::abs(itemMax));
+    // Handline Zeroes in scale vector
```
zoq: Picky style comment; missing point at the end.

src/mlpack/core/data/scaler_methods/maxabsscaler.hpp
jeffin143: @zoq , done

src/mlpack/core/data/scaler_methods/maxabsscaler.hpp
jeffin143: Removed from every file

src/mlpack/core/data/scaler_methods/maxabsscaler.hpp
jeffin143: Sorry , I meant it using the function below , but made changes

src/mlpack/core/data/scaler_methods/maxabsscaler.hpp
jeffin143: Made apt changes everywhere

src/mlpack/core/data/scaler_methods/maxabsscaler.hpp
jeffin143: Added

src/mlpack/core/data/scaler_methods/maxabsscaler.hpp
jeffin143: Made changes , also test too

src/mlpack/core/data/scaler_methods/maxabsscaler.hpp
```diff
@@ -0,0 +1,108 @@
...
+ *
+ * Given an input dataset this class helps you to scale each
+ * feature by its maximum absolute value.
+ *
+ * @code
+ * arma::Mat<double> input = loadData();
+ * arma::Mat<double> output;
+ *
+ * // Scale the features.
+ * MaxAbsScaler<double> scale;
```
zoq: There is no template class parameter, at least not anymore.

src/mlpack/core/data/scaler_methods/maxabsscaler.hpp
```diff
@@ -0,0 +1,108 @@
...
+ * http://www.opensource.org/licenses/BSD-3-Clause for more information.
+ */
+#ifndef MLPACK_CORE_DATA_MAX_ABS_SCALE_HPP
+#define MLPACK_CORE_DATA_MAX_ABS_SCALE_HPP
+
+#include <mlpack/prereqs.hpp>
+
+namespace mlpack {
+namespace data {
+/**
```
zoq: Another picky style comment, can we add an empty line here?

src/mlpack/core/data/scaler_methods/maxabsscaler.hpp
```diff
@@ -0,0 +1,108 @@
...
+  * @param output Output matrix with scaled features.
+  */
+  template<typename MatType>
+  void Transform(const MatType& input, MatType& output)
+  {
+    output.copy_size(input);
+    itemMin = arma::min(input, 1);
+    itemMax = arma::max(input, 1);
+    scale = arma::max(arma::abs(itemMin), arma::abs(itemMax));
+    // Handline Zeroes in scale vector.
```
zoq: Do you mean "zeros"?

src/mlpack/core/data/scaler_methods/maxabsscaler.hpp
```diff
@@ -0,0 +1,108 @@
...
+      if (scale(i) == 0)
+      {
+        scale(i) = 1;
+      }
+    }
+    for (size_t i = 0; i < input.n_rows; i++)
+    {
+      for (size_t j = 0; j < input.n_cols; j++)
+      {
+        output(i, j) = input(i, j) / scale(i);
```
zoq: We could simplify this by using `.each_col()`:

```cpp
output = input.each_col() / scale.
```

maybe we have to use `% 1.0 / scale` here.

A similar approach should work for the other classes as well. I can comment on the other classes as well, if you like.

src/mlpack/core/data/scaler_methods/maxabsscaler.hpp
```diff
@@ -0,0 +1,108 @@
...
+  */
+  template<typename MatType>
+  void InverseTransform(const MatType& input, MatType& output)
+  {
+    output.copy_size(input);
+    for (size_t i = 0; i < input.n_rows; i++)
+    {
+      for (size_t j = 0; j < input.n_cols; j++)
+      {
+        output(i, j) = input(i, j) * scale(i);
```
zoq: We could use `.each_col()` here as well.

src/mlpack/core/data/scaler_methods/meannormalization.hpp
```diff
@@ -0,0 +1,116 @@
...
+ * scale.Tranform(input, output);
+ *
+ * // Retransform the input.
+ * scale.InverseTransform(output, input);
+ * @endcode
+ */
+class MeanNormalization
+{
+ public:
+
```
zoq: We can remove the extra line here.

src/mlpack/core/data/scaler_methods/meannormalization.hpp
```diff
@@ -0,0 +1,116 @@
...
+ *
+ * Given an input dataset this class helps you to normalize each
+ * feature.
+ *
+ * @code
+ * arma::Mat<double> input = loadData();
+ * arma::Mat<double> output;
+ *
+ * // Scale the features.
+ * MeanNormalization<double> scale;
```
zoq: See comment above.

src/mlpack/core/data/scaler_methods/maxabsscaler.hpp
jeffin143: @zoq , That was a really beautiful way of reducing lines. Thanks for pointing it out. I have changed it for other class as well.

Although I didn't undestand the difference between `output = input.each_col() / scale`  and `output = input.each_col() % 1.0 / scale`, but i have used the latter one everywhere.

src/mlpack/core/data/scaler_methods/maxabsscaler.hpp
jeffin143: yes, sorry

src/mlpack/core/data/scaler_methods/maxabsscaler.hpp
jeffin143: Added

src/mlpack/core/data/scaler_methods/maxabsscaler.hpp
jeffin143: My bad, forgot to change documentation.

src/mlpack/core/data/scaler_methods/maxabsscaler.hpp
jeffin143: Adapted.

src/mlpack/core/data/scaler_methods/meannormalization.hpp
jeffin143: Removed.

src/mlpack/core/data/scaler_methods/meannormalization.hpp
jeffin143: Rectified.

src/mlpack/core/data/scaler_methods/maxabsscaler.hpp
zoq: Should be the same.

jeffin143: Ok : ) , Made changes everywhere.

src/mlpack/core/data/scaler_methods/maxabsscaler.hpp
```diff
@@ -0,0 +1,97 @@
...
+    scale = arma::max(arma::abs(itemMin), arma::abs(itemMax));
+    // Handling zeros in scale vector.
+    for (size_t i = 0; i < scale.n_elem; i++)
+    {
+      if (scale(i) == 0)
+      {
+        scale(i) = 1;
+      }
+    }
+    output = input.each_col() % (1.0 / scale);
```
rcurtin: It might make the code a bit longer but you might consider iterating over columns and then multiplying by `1.0 / scale` only if `scale(i) != 0`.  (Also, you can just do `/ scale` instead of `% (1.0 / scale)`.)

Another comment, should we center the data here too?

src/mlpack/core/data/scaler_methods/maxabsscaler.hpp
```diff
@@ -0,0 +1,97 @@
...
+  }
+
+  /**
+  * Function to retrive original dataset.
```
rcurtin: `retrieve`

src/mlpack/core/data/scaler_methods/maxabsscaler.hpp
```diff
@@ -0,0 +1,97 @@
...
+  //! Get the Scale row vector.
+  const arma::colvec& Scale() const { return scale; }
+
+ private:
+  // Vector which holds minimum of each feature.
+  arma::colvec itemMin;
+  // Vector which holds maximum of each feature.
+  arma::colvec itemMax;
+  // Vector which is used to scale up each feature.
+  arma::colvec scale;
```
rcurtin: You can just use `arma::vec` instead of `arma::colvec`.

src/mlpack/core/data/scaler_methods/meannormalization.hpp
```diff
@@ -0,0 +1,102 @@
...
+#include <mlpack/prereqs.hpp>
+
+namespace mlpack {
+namespace data {
+
+/**
+ * A simple Mean Normalization class
+ *
+ * Given an input dataset this class helps you to normalize each
+ * feature.
```
rcurtin: Can we add a comment about how the normalization is performed? :)  Mostly the user will just read this documentation, so if it's not clear on exactly what happens, then the user will probably be confused.

src/mlpack/core/data/scaler_methods/maxabsscaler.hpp
```diff
@@ -0,0 +1,97 @@
...
+#include <mlpack/prereqs.hpp>
+
+namespace mlpack {
+namespace data {
+
+/**
+ * A simple MaxAbs Scaler class.
+ *
+ * Given an input dataset this class helps you to scale each
+ * feature by its maximum absolute value.
```
rcurtin: It might be worth adding a comment with some LaTeX math on exactly the scaling function that will be done here.

src/mlpack/core/data/scaler_methods/meannormalization.hpp
```diff
@@ -0,0 +1,102 @@
...
+ *
+ * // Retransform the input.
+ * scale.InverseTransform(output, input);
+ * @endcode
+ */
+class MeanNormalization
+{
+ public:
+  /**
+  * Function to scale Features.
```
rcurtin: I think this applies throughout; I think you should use `features` here.

src/mlpack/core/data/scaler_methods/minmaxscaler.hpp
```diff
@@ -0,0 +1,123 @@
...
+  // Vector which holds minimum of each feature.
+  arma::colvec itemMin;
+  // Vector which holds maximum of each feature.
+  arma::colvec itemMax;
+  // Scale vector which is used to scale up each feature.
+  arma::colvec scale;
+  // Lower value for range.
+  double scalemin;
+  // Upper value for range.
+  double scalemax;
```
rcurtin: I'd suggest the use of the camelcase `scaleMin` and `scaleMax` to match the rest of the codebase.

src/mlpack/core/data/scaler_methods/minmaxscaler.hpp
```diff
@@ -0,0 +1,123 @@
...
+#include <mlpack/prereqs.hpp>
+
+namespace mlpack {
+namespace data {
+
+/**
+ * A simple MinMax Scaler class
+ *
+ * Given an input dataset this class helps you to scale each
+ * feature to a given range.
```
rcurtin: This comment is more descriptive and I as a user wouuld be able to understand what this does, but if you want, adding some equations to describe the class's functionality could still be helpful.

src/mlpack/core/data/scaler_methods/standardscaler.hpp
```diff
@@ -0,0 +1,94 @@
...
+#include <mlpack/prereqs.hpp>
+
+namespace mlpack {
+namespace data {
+
+/**
+ * A simple Standard Scaler class
+ *
+ * Given an input dataset this class helps you to Standardize features
+ * by removing the mean and scaling to unit variance.
```
rcurtin: Just FYI, this is not the same as whitening, but it reminded me of some of the whitening functionality we have, like `WhitenUsingSVD()` in `src/mlpack/core/math/lin_alg.hpp`.  You might consider adding a whitening scaling too (maybe some other time), because that can sometimes be a useful preprocessing step for some algorithms.

http://ufldl.stanford.edu/wiki/index.php/Data_Preprocessing#PCA.2FZCA_Whitening is not exactly the same but it is similar.

src/mlpack/tests/load_save_test.cpp
```diff
@@ -14,6 +14,10 @@
 #include <mlpack/core.hpp>
 #include <mlpack/core/data/load_arff.hpp>
 #include <mlpack/core/data/map_policies/missing_policy.hpp>
+#include <mlpack/core/data/scaler_methods/minmaxscaler.hpp>
+#include <mlpack/core/data/scaler_methods/maxabsscaler.hpp>
+#include <mlpack/core/data/scaler_methods/standardscaler.hpp>
+#include <mlpack/core/data/scaler_methods/meannormalization.hpp>
```
rcurtin: It might be worth adding another test file here, just to keep things from getting too long.  Technically scaling is not loading/saving. :)

src/mlpack/tests/load_save_test.cpp
```diff
@@ -1727,6 +1731,78 @@ BOOST_AUTO_TEST_CASE(CategoricalNontransposedCSVLoadTest04)
...
+ * Test For MaxAbs Scaler Class.
+ */
+BOOST_AUTO_TEST_CASE(MaxAbsScalerTest)
+{
+    arma::mat dataset = "-1 -0.5 0 1;"
+                        "2 6 10 18;";
+    arma::mat scaleddataset;
+    arma::mat temp;
+    arma::mat scaled = "-1 -0.5 0 1;"
+                       "0.1111111111 0.3333333333 0.55555556 1.0000;";
```
rcurtin: In general, mlpack matrices are column-major (because of Armadillo, which is itself column-major because of FORTRAN), meaning that each *column* is one data point.  In the tests you've written here, it looks like row-major is assumed.  So probably the implementations should all be transposed to work on rows (dimensions) instead of columns (points).

src/mlpack/tests/load_save_test.cpp
```diff
@@ -1727,6 +1731,78 @@ BOOST_AUTO_TEST_CASE(CategoricalNontransposedCSVLoadTest04)
...
+    arma::mat scaleddataset;
+    arma::mat temp;
+    arma::mat scaled = "-0.43750000000 -0.187500000 0.062500000 0.562500000;"
+                       "-0.43750000000 -0.187500000 0.062500000 0.562500000;";
+    data::MeanNormalization scale;
+    scale.Transform(dataset, scaleddataset);
+    scale.InverseTransform(scaleddataset, temp);
+    CheckMatrices(scaleddataset, scaled);
+    CheckMatrices(dataset, temp);
+}
```
rcurtin: You should probably also add some tests where you try and "break" your scalers, by passing in empty data, or passing in data where the scale factor will be 0 for some dimensions, or something like this. :)

src/mlpack/core/data/scaler_methods/maxabsscaler.hpp
rcurtin: Hehe, I see this comment actually contradicts Marcus's comment.  Up to you what you'd like to do.

src/mlpack/tests/load_save_test.cpp
rcurtin: Ooh, another fun one might be passing in the same matrix as input and output and making sure it still works correctly.

src/mlpack/core/data/scaler_methods/CMakeLists.txt
```diff
@@ -0,0 +1,17 @@
+# Define the files we need to compile
+# Anything not in this list will not be compiled into mlpack.
+set(SOURCES
+  minmaxscaler.hpp
+  maxabsscaler.hpp
+  standardscaler.hpp
+  meannormalization.hpp
```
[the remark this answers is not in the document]
jeffin143: Changed file names everywhere :)

src/mlpack/core/data/scaler_methods/maxabsscaler.hpp
```diff
@@ -0,0 +1,97 @@
...
+ *
+ * @code
+ * arma::Mat input = loadData();
+ * arma::Mat output;
+ *
+ * // Scale the features.
+ * MaxAbsScaler scale;
+ * scale.Tranform(input, output);
```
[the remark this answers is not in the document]
jeffin143: Sorry typo.

src/mlpack/core/data/scaler_methods/maxabsscaler.hpp
```diff
@@ -0,0 +1,97 @@
...
+
+/**
+ * A simple MaxAbs Scaler class.
+ *
+ * Given an input dataset this class helps you to scale each
+ * feature by its maximum absolute value.
+ *
+ * @code
+ * arma::Mat input = loadData();
+ * arma::Mat output;
```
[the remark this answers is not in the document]
jeffin143: Made a change, If you are fine with the new documentation, just resolve and if not, just suggest me the change.

src/mlpack/core/data/scaler_methods/maxabsscaler.hpp
jeffin143: > Hehe, I see this comment actually contradicts Marcus's comment. Up to you what you'd like to do.

I have made minor changes, but sticking to it

jeffin143: > It might make the code a bit longer but you might consider iterating over columns and then multiplying by `1.0 / scale` only if `scale(i) != 0`. (Also, you can just do `/ scale` instead of `% (1.0 / scale)`.)

Adapted changes , `/ scale`
 
> Another comment, should we center the data here too?

Not sure, whether  it would be a good choice , I have implemented the standard version which didn't involve any centering of data

The standard[ implementation](https://scikit-learn.org/stable/modules/generated/sklearn.preprocessing.MaxAbsScaler.html#sklearn.preprocessing.MaxAbsScaler)

src/mlpack/core/data/scaler_methods/maxabsscaler.hpp
jeffin143: again typos.

src/mlpack/core/data/scaler_methods/maxabsscaler.hpp
jeffin143: change to `arma::vec`

src/mlpack/core/data/scaler_methods/meannormalization.hpp
jeffin143: Added some more documentation :)

src/mlpack/core/data/scaler_methods/maxabsscaler.hpp
jeffin143: Added, Not sure whether the LaTeX math is correct or not, used an online converter

src/mlpack/core/data/scaler_methods/meannormalization.hpp
jeffin143: changed accordingly.

src/mlpack/core/data/scaler_methods/minmaxscaler.hpp
jeffin143: Changed.

src/mlpack/core/data/scaler_methods/minmaxscaler.hpp
jeffin143: Added. Not sure how good is it.

src/mlpack/core/data/scaler_methods/standardscaler.hpp
jeffin143: I will surely add it post gsoc, or at lease after i am done with string utilities. I have added it to my todo list :)

jeffin143: Also your take on building binding for this ?

src/mlpack/tests/load_save_test.cpp
jeffin143: Added another test file. Have always added test in existing file so was unsure of creating a new one.

src/mlpack/tests/load_save_test.cpp
jeffin143: > You should probably also add some tests where you try and "break" your scalers, by passing in empty data, or passing in data where the scale factor will be 0 for some dimensions, or something like this. :)

Added a test which has empty data, and also which has dimension with scale factor 0

jeffin143: > Ooh, another fun one might be passing in the same matrix as input and output and making sure it still works correctly.

That was a nice suggestion, added accordingly and glad to see that it passes without any issues

src/mlpack/tests/load_save_test.cpp
jeffin143: Umm I am aware of the fact that armadillo is coulmn-major and each column is one data point, and that is why `itemMean = arma::mean(input, 1);` I have take mean row wise

In the test , I have 4 data points with 2 features each, So I guess transposing the matrix wouldn't break my code either because it would simply mean 2 data points with 4 features.

May be I could be wrong, and if it is so, could you please let me know in detail.

jeffin143:
@zoq , could you please once re run Travis test, probably the probabilistic test may have failed, I didn't find any other issue after going through  failed test.

jeffin143:
Umm so I wrote binding for scaler class. Works very smooth for both the cli and python one. Below are some snaps.

**For CLI**
Description : 

Screenshot from 2019-05-23 12-00-30

An example for the api : 
Screenshot from 2019-05-23 11-51-51

**For python**
Description :
Screenshot from 2019-05-23 13-41-36

An example for the api : 
Screenshot from 2019-05-23 17-11-36

I am not sure what to add as test for the binding, could you suggest me some test or potential failure and i would write test for that.

Also the output in case of CLI is 2.5000000e-01, I am not sure this is what user expects, also it is writing  same way in csv file.Is there any way so that i can output 0.25 instead of 2.5000000e-01.The latter one is what the user would like.

The train.csv files contain the same data, which i have used for testing in scaling_test_utility file.

src/mlpack/core/data/scaler_methods/standardscaler.hpp
jeffin143: > Also your take on building binding for this ?

Had some free time up my sleeve, so wrote one , PTAL :)

jeffin143: > I will surely add it post gsoc, or at lease after i am done with string utilities. I have added it to my todo list :)

Ok, So none of my PR had any comments, so had some free time left and thus implemented PCA and ZCA whitening, followed this [article](http://ufldl.stanford.edu/tutorial/unsupervised/PCAWhitening/) to implement that, added a test and also added it to binding.

src/mlpack/core/data/scaler_methods/maxabsscaler.hpp
rcurtin: Sounds good to me, we can leave scaling out here. :+1: Thanks for the fixes.

src/mlpack/core/data/scaler_methods/maxabsscaler.hpp
rcurtin: If it passes the build checks, it's probably fine. :)

src/mlpack/core/data/scaler_methods/minmaxscaler.hpp
rcurtin: Yeah, I think this is a lot clearer on how the functionality works.

src/mlpack/core/data/scaler_methods/standardscaler.hpp
rcurtin: Sounds good, sorry for the slow response on this one.  I'll review the PCA/ZCA implementation now.

src/mlpack/core/data/scaler_methods/whitening.hpp
```diff
@@ -0,0 +1,147 @@
...
+ * arma::mat input;
+ * Load("train.csv", input);
+ * arma::mat output;
+ *
+ * // Scale the features using PCA.
+ * Whitening scale;
+ * scale.PCA(input, output);
+ *
+ * // Scale the features using ZCA.
+ * scale.ZCA(input, ouput);
```
rcurtin: Design-wise, this is different than the other classes because it doesn't have a `Transform()` method.  I'd suggest splitting this into two classes (`PCAWhitening` and `ZCAWhitening`), and providing a `Transform()` method just like the other classes do.

src/mlpack/core/data/scaler_methods/whitening.hpp
```diff
@@ -0,0 +1,147 @@
...
+    itemMean = arma::mean(input, 1);
+    // Centering the Dataset
+    output = (input.each_col() - itemMean);
+    arma::mat diag;
+    // Get eigenvectors and eigenvalues of covariance of input matrix.
+    eig_sym(eigenValues, eigenVectors, mlpack::math::ColumnCovariance(output));
+    // Generate diagonal matrix using 1 / sqrt(eigenvalues + epsilon) for each 
+    // value.
+    for (size_t i = 0; i < eigenValues.n_elem; i++)
+        eigenValues(i) = eigenValues(i) + epsilon;
```
rcurtin: Looks like this doesn't take the square root of the inverse of the eigenvalues.  You could also do that as an Armadillo expression...

```
eigenValues = 1.0 / (arma::sqrt(eigenvalues) + epsilon);
```

src/mlpack/core/data/scaler_methods/whitening.hpp
```diff
@@ -0,0 +1,147 @@
...
+    mlpack::math::VectorPower(eigenValues, -0.5);
+    diag.zeros(eigenValues.n_elem, eigenValues.n_elem);
+    diag.diag() = eigenValues;
+
+    // Our whitening matrix is diag(1 / sqrt(eigenvectors)) * eigenvalues.
+    whiteningMatrix = diag * trans(eigenVectors);
+
+    // Now apply the whitening matrix.
+    output = whiteningMatrix * output;
```
rcurtin: Consider wrapping this all into one expression---it'll be easier for Armadillo to optimize:

```
output = arma::diagmat(eigenValues) * eigenVectors.t() * output;
```

In fact you could even wrap the inversion of the eigenvalues into this too.

src/mlpack/core/data/scaler_methods/whitening.hpp
```diff
@@ -0,0 +1,147 @@
...
+    // ZCA output.
+    output = eigenVectors * output;
```
rcurtin: There's a lot of code duplication here, but I think that you could use the PCA code, and then have an extra overload that returns the eigenvectors, then use that for ZCA.

src/mlpack/core/data/scaler_methods/whitening.hpp
```diff
@@ -0,0 +1,147 @@
...
+  const arma::vec& ItemMean() const { return itemMean; }
+  //! Get the eigenvalues vector.
+  const arma::vec& EigenValues() const { return eigenValues; }
+  //! Get the eigenvector.
+  const arma::mat& EigenVectors() const { return eigenVectors; }
+  //! Get the WhiteningMatrix.
+  const arma::mat& WhiteningMatrix() const { return whiteningMatrix; }
+  //! Get the Regularisation Parameter.
+  const double& Epsilon() const { return epsilon; }
+
```
rcurtin: Should we also add an `InverseTransform()` function?

src/mlpack/methods/preprocess/preprocess_scale_main.cpp
```diff
@@ -0,0 +1,174 @@
...
+#include <mlpack/core/data/scaler_methods/whitening.hpp>
+#include <mlpack/core/data/scaler_methods/standard_scaler.hpp>
+
+PROGRAM_INFO("Scale Data",
+    // Short description.
+    "A utility to perform feature scaling.",
+    // Long description.
+    "This utility takes a dataset and performs feature scaling using one of "
+    "the five scaler methods namely: max_abs_scaler, mean_normalization, "
+    "min_max_scaler ,standard_scaler and whitening. The function takes a "
```
rcurtin: This is pedantic, but for consistency can you put the types of scaling in `''`s?  i.e. `'max_abs_scaler', 'mean_normalization', ...`

src/mlpack/methods/preprocess/preprocess_scale_main.cpp
```diff
@@ -0,0 +1,174 @@
...
+
+// Define parameters for data.
+PARAM_MATRIX_IN_REQ("input", "Matrix containing data.", "i");
+PARAM_MATRIX_OUT("output", "Matrix to save scaled data to.", "o");
+PARAM_STRING_IN("scaler_method", "method to use for scaling, the "
+    "default is standard_scaler.", "a", "standard_scaler");
+PARAM_STRING_IN("whitening_method", "method to use for whitening, the "
+    "default is ZCA.", "w", "ZCA");
+PARAM_DOUBLE_IN("epsilon", "regularization Parameter, should be between"
+  "-1 to 1.", "r", 0.000001);
```
rcurtin: This should be doubly indented since it's a wrapped line.

src/mlpack/methods/preprocess/preprocess_scale_main.cpp
```diff
@@ -0,0 +1,174 @@
...
+
+PARAM_INT_IN("seed", "Random seed (0 for std::time(NULL)).", "s", 0);
+PARAM_INT_IN("min_value", "Starting value of range for min_max_scaler.",
+    "b", 0);
+PARAM_INT_IN("max_value", "Ending value of range for min_max_scaler.",
+    "e", 1);
+
```
rcurtin: Another set of parameters you might consider including is a `PARAM_MODEL_IN()` and `PARAM_MODEL_OUT()` that allows saving of the scalers for reuse, i.e., for inverse scaling (to get back to the original matrix), or for scaling new data with the parameters of the old scaling.  However, I realize that would add a lot of complexity to this, and I'm fine with it as is if you don't want to add that now.

src/mlpack/methods/preprocess/preprocess_scale_main.cpp
```diff
@@ -0,0 +1,174 @@
...
+      ",max_abs_scaler, whitening, standard_scaler or mean_normalization.");
+  // If scaler_method is not set, warn the user.
+  if (!CLI::HasParam("scaler_method"))
+  {
+    Log::Warn << "You did not specify " << PRINT_PARAM_STRING("scaler_method")
+        << ", so it will be automatically set to standard_scaler." << endl;
+  }
+
+  // Load the data.
+  arma::mat input = CLI::GetParam<arma::mat>("input");
```
rcurtin: I'd suggest using `std::move()` here, or having `input` be a reference instead.

src/mlpack/methods/preprocess/preprocess_scale_main.cpp
```diff
@@ -0,0 +1,174 @@
...
+    if (!CLI::HasParam("min_value"))
+    {
+      Log::Warn << "You did not specify " << PRINT_PARAM_STRING("min_value")
+          << ", so it will be automatically set to 0." << endl;
+    }
+    if (!CLI::HasParam("max_value"))
+    {
+      Log::Warn << "You did not specify " << PRINT_PARAM_STRING("max_value")
+          << ", so it will be automatically set to 1." << endl;
+    }
```
rcurtin: No need for these, I think that the default values themselves are good enough (and if the user wants those are documented so they can see how they could set values).

src/mlpack/tests/load_save_test.cpp
rcurtin: Ah, sorry, you're right.  I got myself confused.  Sorry about that.

src/mlpack/core/data/scaler_methods/whitening.hpp
jeffin143: Made appropriate changes, and adapted the same api as others.

src/mlpack/core/data/scaler_methods/whitening.hpp
jeffin143: I agree, Tried to avoid duplication of code, but didn't use overload function, and instead copied the data.

src/mlpack/core/data/scaler_methods/whitening.hpp
jeffin143: I agree with you, Works smoothly. Sometimes my computer behaves in weird ways and after having discussion about the function , i tested it with some inputs and it works smoothly.The same code which i wrote previous worked well.

src/mlpack/methods/preprocess/preprocess_scale_main.cpp
jeffin143: Changed :)

src/mlpack/methods/preprocess/preprocess_scale_main.cpp
jeffin143: Agreed, Made change.

src/mlpack/methods/preprocess/preprocess_scale_main.cpp
jeffin143: Umm, I seriously have no clue how both of these work, and thus would like to study it and then if feasible would implement it, but i would keep that for later. I would learn about `PARAM_MODEL_IN()` after gsoc most probably.

As of now, I would like not to code it since i have no prior knowledge about it.

src/mlpack/methods/preprocess/preprocess_scale_main.cpp
jeffin143: Removed.

src/mlpack/methods/preprocess/preprocess_scale_main.cpp
jeffin143: Done.

jeffin143:
> We should definitely add something in `main_tests/` for the binding, but the key is just that the tests there need to make sure that the options are getting passed correctly---so we don't need to check that the results are exactly right (the tests you already wrote do that) but instead that the options make _something_ happen. So, e.g., tests in `main_tests/` should try passing different scaling strategies to ensure that the output changes, and different parameters for those to make sure the output also changes, and should make sure that unrelated options don't change anything (like if I specify `epsilon` for a scaler that doesn't need it, then the result shouldn't change).
>
Added all the three test according to your suggestion.
 
> For the CLI output format, the code that outputs the matrix in `data::Save()` in turn calls out to Armadillo, but last I checked we were not able to change the format. However, perhaps that has been changed. If you like, you can open a separate issue and we can handle it there. Another option might be submitting a patch upstream to Armadillo so that it doesn't output full precision when it isn't needed.
> 
Ok, as of now we will leave it, but we surely need to look into this issue, I will open an issue, also about submitting patch, i guess i am not so sound with armadillo library :), so i will leave it to you.

> Also, we should document this in `HISTORY.md` so the changelog shows we added it. :)

I will do it.

src/mlpack/core/data/scaler_methods/whitening.hpp
jeffin143: Thanks :) , made apt changes.

jeffin143:
@rcurtin  Just for fun sake : Tried to benchmark the functions
Armadillo is really great, we can see its fast implementation on relatively small dataset.

`c++ > python` :)

Scaler Method | Mlpack | Sklearn 
------------ | ------------- | -------------
MinMaxScaler | 0.001542 | 0.002096
MaxAbsScaler | 0.001190 | 0.001541
StandardScaler |  0.001892 | 0.037366

Screenshot from 2019-05-31 09-31-25


Script i used was : 
```python
from sklearn.datasets import load_digits
from mlpack import preprocess_scale
import time
import numpy as np
from sklearn.preprocessing import MinMaxScaler
from sklearn.preprocessing import MaxAbsScaler
from sklearn.preprocessing import StandardScaler

x = load_digits().data
y = load_digits().data

start_time = time.time()
output = MinMaxScaler().fit(x).transform(x)
end_time = time.time() - start_time
print("sklearn time for minmax scaler is ",end_time)


start_time = time.time()
output = preprocess_scale(input = y, scaler_method = 'min_max_scaler')
end_time = time.time() - start_time
print("mlpack time for minmax scaler is",end_time)

start_time = time.time()
output = MaxAbsScaler().fit(x).transform(x)
end_time = time.time() - start_time
print("sklearn time for MaxAbsScaler scaler is ",end_time)

start_time = time.time()
output = preprocess_scale(input = y, scaler_method = 'max_abs_scaler')
end_time = time.time() - start_time
print("mlpack time for MaxAbsScaler scaler is",end_time)

start_time = time.time()
output = StandardScaler().fit(x).transform(x)
end_time = time.time() - start_time
print("sklearn time for StandardScaler scaler is ",end_time)

start_time = time.time()
output = preprocess_scale(input = y, scaler_method = 'standard_scaler')
end_time = time.time() - start_time
print("mlpack time for StandardScaler is",end_time)
```

PCA whitening and zca whitening were not avialable in sklearn and also mean_normalization wasn't there

src/mlpack/methods/preprocess/preprocess_scale_main.cpp
jeffin143: > Another set of parameters you might consider including is a `PARAM_MODEL_IN()` and `PARAM_MODEL_OUT()` that allows saving of the scalers for reuse, i.e., for inverse scaling (to get back to the original matrix), or for scaling new data with the parameters of the old scaling. However, I realize that would add a lot of complexity to this, and I'm fine with it as is if you don't want to add that now.

ummm, so the PR didn't had any reviews and thus took some time out and implemented, `PARAM_MODEL_IN()` and `PARAM_MODEL_OUT()`, I changed the API a lilttle bit so that both these could be introduced, A function `Fit()` is introduced which could gain information about all the scaling and then the object can be used to transform multiple dataset with same scales and may not be required to calculate the scale again and again

jeffin143:
Not sure , how well I have wrote the binding, if you find that it isn't appropriate to introduce this api, and stick to the old one, let me know and I will revert back to it.

src/mlpack/core/data/scaler_methods/pcawhitening.hpp
```diff
@@ -0,0 +1,136 @@
+/**
+ * @file pcawhitening.hpp
```
rcurtin: Tiny issue, just to match the rest of the codebase I'd suggest `pca_whitening.hpp`. :+1:

src/mlpack/core/data/scaler_methods/pcawhitening.hpp
```diff
@@ -0,0 +1,136 @@
...
+ *
+ * // Retransform the input.
+ * scale.InverseTransform(output, input);
+ * @endcode
+ */
+class PcaWhitening
+{
+ public:
+  /**
+  * A constructor to set the regulatization parameter.
```
rcurtin: `regularization` :)

src/mlpack/core/data/scaler_methods/pcawhitening.hpp
```diff
@@ -0,0 +1,136 @@
...
+  */
+  template<typename MatType>
+  void Fit(const MatType& input)
+  {
+    itemMean = arma::mean(input, 1);
+    // Get eigenvectors and eigenvalues of covariance of input matrix.
+    eig_sym(eigenValues, eigenVectors, mlpack::math::ColumnCovariance(
+        input.each_col() - itemMean));
+    for (size_t i = 0; i < eigenValues.n_elem; i++)
+        eigenValues(i) = eigenValues(i) + epsilon;
```
rcurtin: Actually you can avoid the loop here and just write `eigenValues += epsilon`.

src/mlpack/core/data/scaler_methods/scaling_model.hpp
```diff
@@ -0,0 +1,140 @@
+/**
+ * @file scaling_model.hpp
```
rcurtin: Do you think this file would be more appropriate in `src/mlpack/methods/preprocess/` since that's where it's used?

src/mlpack/core/data/scaler_methods/standard_scaler.hpp
```diff
@@ -0,0 +1,113 @@
...
+  {
+    ar & BOOST_SERIALIZATION_NVP(itemMean);
+    ar & BOOST_SERIALIZATION_NVP(itemStdev);
+  }
+
+ private:
+  // Vector which holds mean of each feature
+  arma::vec itemMean;
+  // Vector which holds standard devation of each feature
+  arma::vec itemStdev;
```
rcurtin: I think this should be `stddev` not `stdev`, so, e.g., `itemStdDev`.

src/mlpack/core/data/scaler_methods/zcawhitening.hpp
```diff
@@ -0,0 +1,137 @@
...
+  */
+  template<typename MatType>
+  void Fit(const MatType& input)
+  {
+    data::PcaWhitening scale(epsilon);
+    scale.Fit(input);
+    // Important to store results, so that i can use in InverseTranform().
+    eigenVectors = scale.EigenVectors();
+    itemMean = scale.ItemMean();
+    eigenValues = scale.EigenValues();
```
rcurtin: Actually you might even be able to simplify this more by just holding a `PcaWhitening` object in this class, so you don't have to extract the eigenvectors, item mean, and eigenvalues.  Up to you what you want to do though.

src/mlpack/methods/preprocess/preprocess_scale_main.cpp
```diff
@@ -0,0 +1,174 @@
...
+  // Make sure the user specified output filenames.
+  RequireAtLeastOnePassed({ "output", "output_model"}, false,
+      "no output will be saved");
+  // Check scaler method.
+  RequireParamValue<std::string>("scaler_method",
+      [](std::string x) { return x == "standard_scaler" || x ==
+      "min_max_scaler" || x == "mean_normalization" || x == "max_abs_scaler" ||
+      x == "whitening";}, true, "scaler_method must be one among min_max_"
+      "scaler, max_abs_scaler, pcawhitening, standard_scaler, zcawhitening"
+      " or mean_normalization.");
```
rcurtin: You could use `RequireParamInSet()` here.

src/mlpack/methods/preprocess/preprocess_scale_main.cpp
```diff
@@ -0,0 +1,174 @@
...
+  else
+  {
+    m->InverseTransform(input, output);
+  }
+
+  // save the output
+  if (CLI::HasParam("output"))
+    CLI::GetParam<arma::mat>("output") = std::move(output);
+
+  CLI::GetParam<ScalingModel*>("output_model") = m;
```
rcurtin: I think we should add `if (CLI::HasParam("output_model"))` here; is there any reason why you didn't?  (Maybe I forgot something?)

src/mlpack/methods/preprocess/preprocess_scale_main.cpp
```diff
@@ -0,0 +1,174 @@
...
+using namespace std;
+
+PROGRAM_INFO("Scale Data",
+    // Short description.
+    "A utility to perform feature scaling.",
+    // Long description.
+    "This utility takes a dataset and performs feature scaling using one of "
+    "the six scaler methods namely: 'max_abs_scaler', 'mean_normalization', "
+    "'min_max_scaler' ,'standard_scaler', 'pcawhitening' and 'zcawhitening'."
+    " The function takes a matrice as " + PRINT_PARAM_STRING("input") +
```
rcurtin: `matrix` :)

src/mlpack/methods/preprocess/preprocess_scale_main.cpp
```diff
@@ -0,0 +1,174 @@
...
+using namespace arma;
+using namespace std;
+
+PROGRAM_INFO("Scale Data",
+    // Short description.
+    "A utility to perform feature scaling.",
+    // Long description.
+    "This utility takes a dataset and performs feature scaling using one of "
+    "the six scaler methods namely: 'max_abs_scaler', 'mean_normalization', "
+    "'min_max_scaler' ,'standard_scaler', 'pcawhitening' and 'zcawhitening'."
```
rcurtin: You might consider naming the PCA and ZCA techniques `pca_whitening` and `zca_whitening`, so they match the underscores of the other methods.

src/mlpack/methods/preprocess/preprocess_scale_main.cpp
```diff
@@ -0,0 +1,174 @@
...
+
+using namespace mlpack;
+using namespace mlpack::util;
+using namespace mlpack::data;
+using namespace arma;
+using namespace std;
+
+PROGRAM_INFO("Scale Data",
+    // Short description.
+    "A utility to perform feature scaling.",
```
rcurtin: It's ok to use two sentences here---this description is what is seen, e.g., here:
http://mlpack.org/doc/mlpack-3.1.1/cli_documentation.html#gmm_train

(i.e. the two sentences under the call definition)

So you could expand on this here.  Here's an idea...

```
A utility to perform feature scaling on datasets using one of six techniques.  Both scaling and inverse scaling are supported, and scalers can be saved and then applied to other datasets.
```

Maybe you can improve on that a little, it's just an idea. :)

src/mlpack/methods/preprocess/preprocess_scale_main.cpp
```diff
@@ -0,0 +1,174 @@
...
+PARAM_DOUBLE_IN("epsilon", "regularization Parameter for pcawhitening,"
+    " or zcawhitening, should be between -1 to 1.", "r", 0.000001);
+
+PARAM_INT_IN("seed", "Random seed (0 for std::time(NULL)).", "s", 0);
+PARAM_INT_IN("min_value", "Starting value of range for min_max_scaler.",
+    "b", 0);
+PARAM_INT_IN("max_value", "Ending value of range for min_max_scaler.",
+    "e", 1);
+PARAM_INT_IN("function", "function to apply,either 0:Transform or 1:Inverse"
+  "Transform", "f", 0)
```
rcurtin: Consider using `PARAM_FLAG_IN()` and maybe calling it `"inverse_scaling"`, so it will apply regular scaling if not specified but inverse scaling if it is.

src/mlpack/tests/main_tests/preprocess_scale_test.cpp
```diff
@@ -0,0 +1,117 @@
...
+static const std::string testName = "PreprocessScale";
+
+#include <mlpack/core/util/mlpack_main.hpp>
+#include <mlpack/methods/preprocess/preprocess_scale_main.cpp>
+
+#include "test_helper.hpp"
+#include <boost/test/unit_test.hpp>
+#include "../test_tools.hpp"
+
+#include <cmath>
```
rcurtin: I think this is already included in `mlpack/core.hpp`---correct me if I'm wrong.

src/mlpack/tests/main_tests/preprocess_scale_test.cpp
```diff
@@ -0,0 +1,117 @@
...
+                    "2 6 10 18;";
+/**
+ * Check that two different scalers give two different output.
+ */
+BOOST_AUTO_TEST_CASE(TwoScalerTest)
+{
+  std::string method = "max_abs_scaler";
+  // Input custom data points.
+  SetInputParam("input", dataset);
+  SetInputParam("scaler_method", std::move(method));
```
rcurtin: Actually if you wanted to simplify this, you could just put `method` in `SetInputParam()` directly: `SetInputParam("scaler_method", "max_abs_scaler")`.  Up to you.

src/mlpack/tests/main_tests/preprocess_scale_test.cpp
```diff
@@ -0,0 +1,117 @@
...
+  SetInputParam("min_value", 2);
+  SetInputParam("max_value", 4);
+  SetInputParam("epsilon", 0.005);
+  mlpackMain();
+  arma::mat output = CLI::GetParam<arma::mat>("output");
+  arma::mat scaled = "-1.18321596 -0.50709255  0.16903085 1.52127766;"
+                   "-1.18321596 -0.50709255  0.16903085 1.52127766;";
```
rcurtin: It might be safer to run `mlpackMain()` once to get the first output, in case the internal implementation of the standard scaler ever changes.

src/mlpack/tests/main_tests/preprocess_scale_test.cpp
```diff
@@ -0,0 +1,117 @@
...
+  CheckMatrices(scaled, output);
+}
+
```
rcurtin: I can think of just a couple more tests to add here---

 1. We should check to make sure a trained scaler model gives the same output in two separate runs.  So, for instance, train a scaler model on a dataset and then save `output` and `output_model`, then run again with the same dataset and `input_model` set to the output model of the last run, and make sure the output of the second run is the same as the output of the first.
 2. Make sure `min_value` and `max_value` make a difference when `min_max_scaler` is used.
 3. Try running in regular mode and inverse mode and make sure that either different results are given, or that we recover the original matrix.
 4. Make sure `epsilon` makes a difference for PCA whitening.
 5. Make sure all six strategies give some kind of output and don't crash.  (Maybe this can be combined with (3)).
 6. Make sure invalid values of `epsilon` cause an exception.
 7. Make sure invalid values of `scaler_method` cause an exception.
 8. Make sure an exception is thrown when `max_value` < `min_value` (or they are equal). :)

It can be kind of tedious to write these tests but if you don't then our users will find when something is wrong and write us bug reports. :)  Better to know that we won't get those now.

src/mlpack/core/data/scaler_methods/pcawhitening.hpp
jeffin143: Done.

src/mlpack/core/data/scaler_methods/pcawhitening.hpp
jeffin143: Thanks for this :)

src/mlpack/core/data/scaler_methods/pcawhitening.hpp
jeffin143: Sorry, Done.

src/mlpack/core/data/scaler_methods/standard_scaler.hpp
jeffin143: I agree with you, changes accordingly.

src/mlpack/core/data/scaler_methods/zcawhitening.hpp
jeffin143: That was a really good idea, Didn't think that way. Thanks for the idea.

src/mlpack/core/data/scaler_methods/scaling_model.hpp
jeffin143: Yes :)

src/mlpack/methods/preprocess/preprocess_scale_main.cpp
jeffin143: i didn't knew about it, Thanks for letting me know.

src/mlpack/methods/preprocess/preprocess_scale_main.cpp
jeffin143: Sorry, I completely missed it.

src/mlpack/methods/preprocess/preprocess_scale_main.cpp
jeffin143: Made apt correction.

src/mlpack/methods/preprocess/preprocess_scale_main.cpp
jeffin143: You have reviewed few of my PR's and hence you might have noticed, I always have this issue with Documentation, I am not so good at it, Hence i have shamelessly copied your statement :)

src/mlpack/tests/main_tests/preprocess_scale_test.cpp
jeffin143: Yes, you are right.

src/mlpack/tests/main_tests/preprocess_scale_test.cpp
jeffin143: `SetInputParam("scaler_method", "max_abs_scaler")` throws error and hence i haven't changed it.

src/mlpack/tests/main_tests/preprocess_scale_test.cpp
jeffin143: I agree with you.

src/mlpack/tests/main_tests/preprocess_scale_test.cpp
jeffin143: > I can think of just a couple more tests to add here---
> 
>     1. We should check to make sure a trained scaler model gives the same output in two separate runs.  So, for instance, train a scaler model on a dataset and then save `output` and `output_model`, then run again with the same dataset and `input_model` set to the output model of the last run, and make sure the output of the second run is the same as the output of the first.
> 
Added this test.
>     2. Make sure `min_value` and `max_value` make a difference when `min_max_scaler` is used.
> 
It was aldeardy there, check test no 2 :)
>     3. Try running in regular mode and inverse mode and make sure that either different results are given, or that we recover the original matrix.
> 
Added this too, Thanks for the suggestion.
>     4. Make sure `epsilon` makes a difference for PCA whitening.
> 
Wrote a test for this too.
>     5. Make sure all six strategies give some kind of output and don't crash.  (Maybe this can be combined with (3)).
> 
I don't think so , we should write one for this because , I have already checked Transform and InverseTransfrom for the functionality, If you still want me to write , I will do so, Do i have to write for all 6 scalers ?
>     6. Make sure invalid values of `epsilon` cause an exception.
> 
Yes, Added test for Invalid epsilon.
>     7. Make sure invalid values of `scaler_method` cause an exception.
> 
Added test for this too.
>     8. Make sure an exception is thrown when `max_value` < `min_value` (or they are equal). :)
> 
Thanks for the idea, Wrote a test accordingly.
> It can be kind of tedious to write these tests but if you don't then our users will find when something is wrong and write us bug reports. :) Better to know that we won't get those now.

I agree with you, No issue with writing test for any of these. If you have some more let me know, I would be happy enough to implement those.

src/mlpack/methods/preprocess/preprocess_scale_main.cpp
jeffin143: I guess you are talking about `PARAM_FLAG()` , because i didn't find `PARAM_FLAG_IN()` anywhere

src/mlpack/methods/preprocess/preprocess_scale_main.cpp
jeffin143: Made changes

src/mlpack/tests/main_tests/preprocess_scale_test.cpp
rcurtin: Awesome, thanks for adding these.

> I don't think so , we should write one for this because , I have already checked Transform and InverseTransfrom for the functionality, If you still want me to write , I will do so, Do i have to write for all 6 scalers ?

The thought here is just to check that the user can specify each of the six options for the strategies without there being some crash.  Sometimes it can happen that the documentation says a certain option should be used but then it is spelled differently in the code. :)  So the check is not that the strategy works; in fact, just running `BOOST_REQUIRE_NO_THROW(mlpackMain())` should be fine for each.

src/mlpack/core/data/scaler_methods/zca_whitening.hpp
```diff
@@ -0,0 +1,120 @@
...
+
+  template<typename Archive>
+  void serialize(Archive& ar, const unsigned int /* version */)
+  {
+    ar & BOOST_SERIALIZATION_NVP(zca);
+  }
+
+ private:
+  // A pointer to PcaWhitening Class.
+  PcaWhitening* zca;
```
rcurtin: Tiny nitpick, but maybe this should be named `pca` for clarity? :)

src/mlpack/tests/main_tests/preprocess_scale_test.cpp
rcurtin: > It was aldeardy there, check test no 2 :)

You're right, I see it now. :)  Must have missed it the first time.

src/mlpack/methods/preprocess/preprocess_scale_main.cpp
rcurtin: No worries! :)  Writing documentation is one of the hardest parts in my opinion.  I'm happy to help out where I can.

src/mlpack/tests/main_tests/preprocess_scale_test.cpp
rcurtin: Oh, you're right, `SetInputParam()` takes an rvalue reference for the second argument, so I guess `std::move()` is needed.  It doesn't really make a difference either way. :)

rcurtin:
@mlpack-bot why aren't you automatically approving this? :)

rcurtin:
It worked!!! Took me like a week to debug that.

rcurtin:
Thanks again for the contribution; I'm really happy to have this merged.  I made some minor style changes in 2d8e1f8e, including changing `PcaWhitening` and `ZcaWhitening` to `PCAWhitening` and `ZCAWhitening` since PCA and ZCA are acronyms. :)  I know it is a little pedantic but I am all about consistency :)

PR #1878, opened 2019-04-25
Cosine distance metric

jeffin143:
I have implemented Cosine Similarity Metric, Cosine similarity is generally used as a metric for measuring distance when the magnitude of the vectors does not matter. 

Post Gsoc, Mlpack will have string utilities and thus Cosine similarity metric could be very useful since, it is used often to evaluate how 2 documents are related.

Optionally We might also want to apply cosine similarity for cases where some properties of the instances make so that the weights might be larger without meaning anything different. Sensor values that were captured in various lengths (in time) between instances could be such an example.

You could find relevance of cosine similarity in this [article](https://cmry.github.io/notes/euclidean-v-cosine), Under Cosine in Action - we could find an implementation of it.

### DIFF
--- a/src/mlpack/core/metrics/CMakeLists.txt
+++ b/src/mlpack/core/metrics/CMakeLists.txt
@@ -7,6 +7,8 @@ set(SOURCES
   lmetric_impl.hpp
   mahalanobis_distance.hpp
   mahalanobis_distance_impl.hpp
+  cosine_similarity.hpp
+  cosine_similarity_impl.hpp
 )
 
 # add directory name to sources
--- a/src/mlpack/core/metrics/cosine_similarity.hpp
+++ b/src/mlpack/core/metrics/cosine_similarity.hpp
@@ -1,0 +1,47 @@
+/***
+ * @file cosine_similarity.hpp
+ * @author jeffin Sam
+ *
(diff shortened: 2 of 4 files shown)


ShikharJ:
I don't know why, but again I feel this is too fundamental to not have been implemented somewhere within `mlpack` already.

jeffin143:
> I don't know why, but again I feel this is too fundamental to not have been implemented somewhere within `mlpack` already.

@ShikharJ , What about `Jaccard similarity`. Would it be good to have that in our codebase ?

rcurtin:
I haven't looked at the code to see what is implemented, but:

(a) cosine similarity is not actually a metric.
(b) we should have it already in `kernels/`.

Maybe there is some functionality difference, so perhaps we can improve that implementation if needed (like I said I haven't looked at the code here).

jeffin143:
> I haven't looked at the code to see what is implemented, but:
> 
> (a) cosine similarity is not actually a metric.
> (b) we should have it already in `kernels/`.
> 
> Maybe there is some functionality difference, so perhaps we can improve that implementation if needed (like I said I haven't looked at the code here).

@rcurtin , Yes its there in kernels , I would go ahead and close it, Look same implementation wise :)

PR #1886, opened 2019-04-28
Added accessors for state and environment 

abhinavsagar:
I have added accessors for state and environment variables in q-learning. Users would like to see the state and environment in which they are.


zoq:

PR #1888, opened 2019-05-03
Dual Optimizer for GANs.

walragatver:
Hi,
This is for the idea in #1437. Currently, I have implemented Dual Optimizers for `StandardGANs` and `DCGANs`. I have started testing the `Orielly test case` on the dual optimizers. I will complete Dual Optimizer for `WGAN` and `WGANGP` soon.

`Note`: At this moment this PR contains a lot of `previous` redundant code. It will be removed after the successful implementation of `WGAN`. 

The `Dual Optimizer` code is present in `gan_impl.hpp` at line `220`


walragatver:
Hi @ShikharJ, @zoq,
This now contains the implementation of `standard GAN` and `WGAN`.

Following are the two codes I have referred.
https://github.com/eriklindernoren/Keras-GAN/blob/master/gan/gan.py
https://github.com/eriklindernoren/Keras-GAN/blob/master/wgan/wgan.py.

Following are some other important notes.
1. I have currently kept `supervised` parameter as `private` so let me know what to do with it.

2. I don't know the `stepsize` of `generator` network of `WGAN` currently I have kept it `0.0001`. Let me know if that's wrong.

walragatver:
Hi @ShikharJ, @zoq,
Sorry for getting impatient. But it would be great if this get's reviewed before the coding period begins. Please review it in your free time.

src/mlpack/tests/gan_test.cpp
```diff
@@ -166,6 +168,10 @@ BOOST_AUTO_TEST_CASE(GANMNISTTest)
 
   trainData = trainData.cols(0, datasetMaxCols - 1);
 
+  // Normalize the input data.
+  trainData = trainData - 127.5;
```
zoq: Is `127.5` equivalent to `arma::mean(trainData, 1)`? Might be more clear to use `mean` instead of a const value.

src/mlpack/methods/ann/gan/gan.hpp
```diff
@@ -95,47 +93,64 @@ class GAN
   //! Move constructor.
   GAN(GAN&&);
 
+  /**
+   * Prepare the network for the given data.
+   * This function won't actually trigger training process.
+   *
+   * @param trainData The real data.
+   */
+  void ResetData(arma::mat trainData);
```
zoq: Do you mind to stick with the `predictors/responses` naming pattern?

src/mlpack/methods/ann/gan/gan.hpp
```diff
@@ -95,47 +93,64 @@ class GAN
...
 
   /**
-   * Evaluate function for the WGAN.
-   * This function gives the performance of the WGAN on the current input.
+   * Train function for WGAN.
    *
-   * @param parameters The parameters of the network.
-   * @param i Index of the current input.
-   * @param batchSize Variable to store the present number of inputs.
+   * @param trainData The real data.
```
zoq: Do you think we could improve the wording here? Not sure "real" is the best term.

src/mlpack/methods/ann/gan/gan_impl.hpp
```diff
@@ -255,111 +225,110 @@ template<
...
 typename std::enable_if<std::is_same<Policy, StandardGAN>::value ||
-                        std::is_same<Policy, DCGAN>::value, double>::type
-GAN<Model, InitializationRuleType, Noise, PolicyType>::
-EvaluateWithGradient(const arma::mat& /* parameters */,
-                     const size_t i,
-                     GradType& gradient,
-                     const size_t /* batchSize */)
+                        std::is_same<Policy, DCGAN>::value, void>::type
+GAN<Model, InitializationRuleType, Noise, PolicyType>::Train(
+    arma :: mat trainData,
```
zoq: Can you remove the extra whitespace here.

src/mlpack/methods/ann/gan/gan_impl.hpp
```diff
@@ -255,111 +225,110 @@ template<
...
+    {
+      noise.set_size(noiseDim, effectiveBatchSize);
+      discriminatorOptimizer.BatchSize() = effectiveBatchSize;
+      discriminatorOptimizer.MaxIterations() = effectiveBatchSize * 2;
+    }
+    noise.imbue( [&]() { return noiseFunction();} );
+    arma::mat fakeImages;
+    generator.Forward(noise, fakeImages);
+
+    discriminatorPredictors = arma::join_rows(
```
zoq: Hm, internally the optimiser could call the Shuffle method, which alters the order of the predictors, responses, which I think has to be in sync with discriminatorPredictors. Correct me if I'm wrong. So I think we either have to adjust the Shuffle method. Let me know what you think.

src/mlpack/tests/gan_test.cpp
walragatver: No, actually as we are using TanHLayer in the generator network our fake images will have pixels in the range of `[-1, 1]` so we need to scale our input images to `[-1, 1]` as well. 
It's mention below as first point in `readme.md` as first point.
https://github.com/soumith/ganhacks

src/mlpack/methods/ann/gan/gan.hpp
walragatver: Yes, I changed it.

src/mlpack/methods/ann/gan/gan.hpp
walragatver: Okay, I have currently kept it as `The data points of real distribution.` Let me know your thoughts regarding it.

src/mlpack/methods/ann/gan/gan_impl.hpp
walragatver: Yes, you are quite correct.
In order to fix this, I have now made `discriminatorOptimizer.shuffle()`  and `generator.Optimizer.shuffle()` as `false` before training. This will prevent shuffling.
We will need to add some `documentation`, that the shuffle parameter's in both the optimizer's will be overrided as `false` before `training`.
Let me know if this is not a correct solution.

rcurtin:

Issue #1887, opened 2019-04-30
Increasing the number of trees in the random forest model does not improve performance


pensivevoice:
#### Issue description

Increasing the number of trees in the random forest model does not improve performance. It is as if all the trees in the forest are equal. Is there a setting to turn on randomness that I am missing? 
[RandomForestNumTreesPotentialBug.txt](https://github.com/mlpack/mlpack/files/3132297/RandomForestNumTreesPotentialBug.txt)

#### Your environment

 * version of mlpack: 3.0.4
 * operating system: Windows 10
 * compiler: Visual Studio 2015
 * version of dependencies (Boost/Armadillo): 1_66_0/7.800.2
 * any other environment information you think is relevant:

#### Steps to reproduce

Create one random forest model with numTrees = 2 and another with numTrees = 10. Keep all other parameters constant. Training performance (e.g. accuracy) does not change. 

#### Expected behavior

Training performance should improve when the number of trees increases since the more complex model is better able to overfit the training data. 

#### Actual behavior

Training performance does not change. 

[RandomForestNumTreesPotentialBug.txt](https://github.com/mlpack/mlpack/files/3132299/RandomForestNumTreesPotentialBug.txt)

MuLx10:
@pensivevoice thanks for opening the issue. I think there might be a bug. I will let you know if I find any.

MuLx10:
I guess #1891 solves the issue.

rcurtin:
Right, thanks for pointing it out @MuLx10.  I should have updated this too.  I ended up having some time to look into this yesterday after an IRC discussion and I think I found what the issue was.  Sorry if I stepped on your toes a bit while you were debugging.

@pensivevoice If you can test that #1891 fixes the issue, it would be really helpful. :+1:

pensivevoice:
Thanks so much for looking into it. Now changing the number of trees makes a difference but not in the expected way. Training performance follows an inverted U shape: it reaches a maximum at some point instead of continue increasing with the number of trees.  Also training performance peaks at a low value. It is not what you would expect from a random forest model with enough trees and a small minimum leaf size.

pensivevoice:
Thanks so much for looking into it. Now changing the number of trees makes a difference but not in the expected way. Training performance follows an inverted U shape: it reaches a maximum at some point instead of continue increasing with the number of trees. Also training performance peaks at a low value. It is not what you would expect from a random forest model with enough trees and a small minimum leaf size.

[...]

PR #1891, opened 2019-05-04
Fix random forest lack of randomness (#1887)

rcurtin:
I had a chance to look into it today and I actually found two bugs in the random forest:

 * The decision tree was not properly initializing the dimension selector in the case of numeric-only data.
 * The random forest was not giving bootstrapped data to the decision trees.

I added a test to check that dimensions are selected randomly, and adapted the other tests (since bootstrapped data no longer gives the same guarantees as we have when we train directly on the full dataset).

### DIFF
--- a/src/mlpack/methods/decision_tree/best_binary_numeric_split_impl.hpp
+++ b/src/mlpack/methods/decision_tree/best_binary_numeric_split_impl.hpp
@@ -30,7 +30,9 @@ double BestBinaryNumericSplit<FitnessFunction>::SplitIfBetter(
 {
   // First sanity check: if we don't have enough points, we can't split.
   if (data.n_elem < (minimumLeafSize * 2))
-    return bestGain;
+    return DBL_MAX;
+  if (bestGain == 0.0)
+    return DBL_MAX; // It can't be outperformed.
 
   // Next, sort the data.
   arma::uvec sortedIndices = arma::sort_index(data);
@@ -38,6 +40,11 @@ double BestBinaryNumericSplit<FitnessFunction>::SplitIfBetter(
   arma::rowvec sortedWeights;
   for (size_t i = 0; i < sortedLabels.n_elem; ++i)
     sortedLabels[i] = labels[sortedIndices[i]];
+
+  // Sanity check: if the first element is the same as the last, we can't split
+  // in this dimension.
+  if (data[sortedIndices[0]] == data[sortedIndices[sortedIndices.n_elem - 1]])
+    return DBL_MAX;
 
   // Only initialize if we are using weights.
   if (UseWeights)
@@ -50,10 +57,72 @@ double BestBinaryNumericSplit<FitnessFunction>::SplitIfBetter(
 
   // Loop through all possible split points, choosing the best one.  Also, force
   // a minimum leaf size of 1 (empty children don't make sense).
-  double bestFoundGain = bestGain;
+  double bestFoundGain = std::min(bestGain + minimumGainSplit, 0.0);
+  bool improved = false;
   const size_t minimum = std::max(minimumLeafSize, (size_t) 1);
-  for (size_t index = minimum; index < data.n_elem - (minimum - 1); ++index)
+
+  // We need to count the number of points for each class.
+  arma::Mat<size_t> classCounts;
+  arma::mat classWeightSums;
+  double totalWeight;
+  double totalLeftWeight = 0.0;
+  double totalRightWeight = 0.0;
+  if (UseWeights)
   {
+    classWeightSums.zeros(numClasses, 2);
+    totalWeight = arma::accu(sortedWeights);
+    bestFoundGain *= totalWeight;
+  }
+  else
+  {
+    classCounts.zeros(numClasses, 2);
+    bestFoundGain *= data.n_elem;
+  }
+
+  // Initialize the counts.
+  // These points have to be on the left.
+  for (size_t i = 0; i < minimum - 1; ++i)
+  {
+    if (UseWeights)
+    {
+      classWeightSums(sortedLabels[i], 0) += sortedWeights[i];
+      totalLeftWeight += sortedWeights[i];
+    }
+    else
+    {
+      ++classCounts(sortedLabels[i], 0);
+    }
+  }
+  // These points have to be on the right.
+  for (size_t i = minimum - 1; i < data.n_elem; ++i)
+  {
+    if (UseWeights)
+    {
+      classWeightSums(sortedLabels[i], 1) += sortedWeights[i];
+      totalRightWeight += sortedWeights[i];
+    }
+    else
+    {
+      ++classCounts(sortedLabels[i], 1);
+    }
+  }
+
+  for (size_t index = minimum; index < data.n_elem - minimum; ++index)
+  {
+    // Update class weight sums or counts.
+    if (UseWeights)
+    {
+      classWeightSums(sortedLabels[index - 1], 1) -= sortedWeights[index - 1];
+      classWeightSums(sortedLabels[index - 1], 0) += sortedWeights[index - 1];
+      totalLeftWeight += sortedWeights[index - 1];
+      totalRightWeight -= sortedWeights[index - 1];
+    }
+    else
+    {
+      --classCounts(sortedLabels[index - 1], 1);
+      ++classCounts(sortedLabels[index - 1], 0);
+    }
+
     // Make sure that the value has changed.
     if (data[sortedIndices[index]] == data[sortedIndices[index - 1]])
       continue;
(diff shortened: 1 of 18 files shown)


gmanlan:
I tested the changes using the sample app, but the trained models are exactly the same every time you run the app. @zoq suggested to add mlpack::math::RandomSeed(time(NULL)); at the beginning of the program and by doing so it now generates different models every time.
So the question would be: shouldn't be the math::random initialization an intrinsic part of Random Forest?

src/mlpack/methods/decision_tree/decision_tree.hpp
```diff
@@ -363,6 +363,10 @@ class DecisionTree :
   //! Modify the child of the given index (be careful!).
   DecisionTree& Child(const size_t i) { return *children[i]; }
 
+  //! Get the split dimension (only meaningful is this is a non-leaf in a
```
zoq: Should this be "if this is a ..."?

src/mlpack/tests/random_forest_test.cpp
```diff
@@ -149,7 +149,7 @@ BOOST_AUTO_TEST_CASE(UnweightedNumericLearningTest)
   size_t rfCorrect = arma::accu(rfPredictions == testLabels);
   size_t dtCorrect = arma::accu(dtPredictions == testLabels);
 
-  BOOST_REQUIRE_GE(rfCorrect, dtCorrect);
+  BOOST_REQUIRE_GE(rfCorrect, dtCorrect * 0.85);
```
zoq: Did the performance in terms of accuracy drop with the changes?

src/mlpack/tests/random_forest_test.cpp
```diff
@@ -483,4 +456,32 @@ BOOST_AUTO_TEST_CASE(RandomForestCategoricalTrainReturnEntropy)
...
+  arma::Row<size_t> l(100);
+  for (size_t i = 0; i < 50; ++i)
+    l(i) = 0;
+  for (size_t i = 50; i < 100; ++i)
+    l(i) = 1;
+
+  bool success = false;
+  size_t trial = 0;
+
+  while (!success && trial < 10)
```
zoq: I think we should clarify the multiple trials here.

src/mlpack/tests/random_forest_test.cpp
gmanlan: Using the sample app, I noticed the accuracy dropped around 10% in average... not sure about this test

zoq: Thanks for the input, didn't expect to see 100% accuracy; I guess if we tweak the model parameter we can probably get the same result.

rcurtin:
I dug in to the problem a bit and exposed several other problems.  I fixed these, and when I finally got everything worked out:

 * Now there is a `--subspace_dim` option to specify the number of dimensions to use in a random forest.
 * Now there is a `--minimum_gain_split` option for random forests.
 * `--print_training_error` is deprecated and changed to `--print_training_accuracy` for the decision tree.
 * There is a very big efficiency improvement, although there is still more that could be done.  This was all I had time for now though.

@gmanlan @pensivevoice can you try this out and let me know if it helps?

src/mlpack/tests/random_forest_test.cpp
rcurtin: Ah, 10 is too many I think, but the trials are still useful.  Comment fixed in cc7e2accb.

src/mlpack/tests/random_forest_test.cpp
rcurtin: Yeah, this was a bit fishy, and after a lot of investigation my conclusions:

 * The last change actually did introduce some problems.  I hope they are fixed now.
 * I'm not sure why the original tests did not fail.  I think it is a "lucky bug".  It's not, in general, guaranteed that a random forest will outperform a single decision tree.  It's very likely, but like I said, not guaranteed.  So I can safely set the tolerance at `0.9 * dtCorrect` and I have no problems when I run many trials.

src/mlpack/methods/random_forest/random_forest.hpp
```diff
@@ -109,39 +128,54 @@ class RandomForest
...
   double Train(const MatType& data,
                const arma::Row<size_t>& labels,
                const size_t numClasses,
-               const size_t numTrees = 50,
-               const size_t minimumLeafSize = 20);
+               const size_t numTrees = 20,
+               const size_t minimumLeafSize = 1,
+               const double minimumGainSplit = 1e-7,
+               DimensionSelectionType dimensionSelector =
+                   DimensionSelectionType());
```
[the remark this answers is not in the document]
rcurtin: Actually that can be inferred by the compiler.  In your particular case you could just write

```
rf = RandomForest<>(dataset, labels,
    numClasses, numTrees, minimumLeafSize, 1e-7, MultipleRandomDimensionSelect(5));
```

src/mlpack/methods/decision_tree/multiple_random_dimension_select.hpp
```diff
@@ -22,16 +22,30 @@ namespace tree {
  *
  * @tparam NumDimensions Number of random dimensions to select.
```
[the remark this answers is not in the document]
rcurtin: Thanks, fixed in d38d3115f.

zoq:

src/mlpack/methods/decision_tree/best_binary_numeric_split_impl.hpp
```diff
@@ -50,47 +57,99 @@ double BestBinaryNumericSplit<FitnessFunction>::SplitIfBetter(
...
+    bestFoundGain *= data.n_elem;
+  }
+
+  // Initialize the counts.
+  // These points have to be on the left.
+  for (size_t i = 0; i < minimum - 1; ++i)
+  {
+    if (UseWeights)
+    {
+      classWeightSums(sortedLabels[i], 0) += sortedWeights[i];
```
zoq: We could move this into the if/else case from above, to make this shorter.

src/mlpack/methods/decision_tree/decision_tree_impl.hpp
```diff
@@ -552,9 +596,8 @@ double DecisionTree<FitnessFunction,
       numClasses,
       UseWeights ? weights.subvec(begin, begin + count - 1) : weights);
   size_t bestDim = datasetInfo.Dimensionality(); // This means "no split".
-  DimensionSelectionType dimensions(datasetInfo.Dimensionality());
-  for (size_t i = dimensions.Begin(); i != dimensions.End();
-       i = dimensions.Next())
+  for (size_t i = dimensionSelector.Begin(); i != dimensionSelector.End();
```
zoq: Maybe we should the same here as well: https://github.com/mlpack/ensmallen/pull/113#pullrequestreview-236418916

src/mlpack/methods/decision_tree/best_binary_numeric_split_impl.hpp
rcurtin: Yeah, agreed.  Done in 7b584bde.  I'm not sure it's actually shorter (didn't check too closely) but it's certainly easier to read.

src/mlpack/methods/decision_tree/decision_tree_impl.hpp
rcurtin: Usually `End()` is just a `const size_t`, but I think paranoia is well warranted. :)  Done in 7b584bde.

rcurtin:
From my end this is ready to go, but I doubt the static code analysis checks are going to finish soon given the Jenkins build queue.  (I also don't believe anything will fail, so I don't think we should wait on them to finish.)  When it's merged I'll release 3.1.1 with this fix and any others that have been merged since the last release.

src/mlpack/methods/random_forest/random_forest_impl.hpp
```diff
@@ -436,38 +456,46 @@ double RandomForest<
   #pragma omp parallel for reduction( + : avgGain)
   for (omp_size_t i = 0; i < numTrees; ++i)
   {
+    Timer::Start("bootstrap");
```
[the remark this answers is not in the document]
rcurtin: I just added these for my own debugging, but I decided to leave them in because they could be interesting.  If you run the binding with `--verbose`, it'll print all the timers collected during the run, which can be nice to figure out exactly where the time is going.  It may be more useful for developers than users, but I've always thought it's reasonably interesting to look at for users too. :)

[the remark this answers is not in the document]
gmanlan: sounds good!

ShikharJ:
Thanks for the contribution, I'll review this today and try to get it merged at the earliest :+1:

src/mlpack/methods/random_forest/random_forest_main.cpp
```diff
@@ -47,7 +47,12 @@ PROGRAM_INFO("Random forests",
     " parameter specifies the minimum number of training points that must fall "
     "into each leaf for it to be split.  The " +
     PRINT_PARAM_STRING("num_trees") +
-    " controls the number of trees in the random forest. If " +
+    " controls the number of trees in the random forest.  The " +
+    PRINT_PARAM_STRING("minimum_gain_split") + " parameter control the minimum "
```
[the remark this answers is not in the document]
rcurtin: Nice catch!  :+1: Are you okay if I merge this, or do you want to take a closer pass first?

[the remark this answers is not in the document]
ShikharJ: Yeah, please go ahead. I'll probably take a look at how mlpack implements Random Forest Algorithm when I am faced with the situation of learning how Random Forests are supposed to work :smile: You might want to resolve the branch conflicts though.

Issue #1874, opened 2019-04-23
Two different versions of the Windows Build Tutorial co-existing (build_windows.html)


gmanlan:
#### Problem location

Website points to: http://mlpack.org/doc/mlpack-3.0.4/doxygen/build_windows.html
Readme.md points to: https://www.mlpack.org/doc/mlpack-git/doxygen/build_windows.html

#### Description of problem

The "build mlpack for Windows from source" documentation is duplicated and is different depending on how you navigate. If you access through the website, you get one version, if you access through the Readme.md, you get a different (more recent?) version.

- NOTE: the apparently more recent version of the tutorial is now confusing because it has "options" or "different paths" to follow during the procedure. For example, the tutorial originally provided cmake cmd line steps but now we are offering cmake GUI steps as well, all in the same place, without proper formatting. This is very confusing. Also, it seems we now support package installation via vcpkg, which again, it is explained without context, under the same page and without even a subtitle. So it would be nice to separate or reformat all this information:
1) Building from source using cmake (the original tutorial)
  1.b) Alternatively using cmake GUI
2) Alternatively, installing via vcpkg

zoq:
Agreed this is somewhat confusing, since we already cover different build/installation it would be great if we could present each in a more structured view.

gmanlan:
This issue is covered with: https://github.com/mlpack/mlpack/pull/1893

PR #1893, opened 2019-05-04
Refactor Windows Build Tutorial

gmanlan:
Based on #1874, the following changes were made:

- Re-organized the tutorial structure so it's not that confusing: installation via vcpkg is introduced as an alternative to the build process with the option to skip it and continue with the build steps.
- Removed all the hardcoded versions of mlpack and dependencies from the paths, so we don't need to unnecessarily maintain the tutorial updated when new versions are released.
- Moved the alternative cmake GUI instructions to an appendix (so one main tutorial path can be followed).
- Added a NOTE about the ENSMALLEN dependency, that must be manually specified if cmake can't automatically download it
- Moved third party/old tutorials to a new section, so users can refer to them in case of issues

### DIFF
--- a/doc/guide/build_windows.hpp
+++ b/doc/guide/build_windows.hpp
@@ -117,7 +117,7 @@ or a different VS version, update the cmake command accordingly.
 cmake -G "Visual Studio 15 2017 Win64" -DBLAS_LIBRARY:FILEPATH="C:/mlpack/mlpack/packages/OpenBLAS.0.2.14.1/lib/native/lib/x64/libopenblas.dll.a" -DLAPACK_LIBRARY:FILEPATH="C:/mlpack/mlpack/packages/OpenBLAS.0.2.14.1/lib/native/lib/x64/libopenblas.dll.a" -DARMADILLO_INCLUDE_DIR="C:/mlpack/armadillo/include" -DARMADILLO_LIBRARY:FILEPATH="C:/mlpack/armadillo/build/Debug/armadillo.lib" -DBOOST_INCLUDEDIR:PATH="C:/boost/" -DBOOST_LIBRARYDIR:PATH="C:/boost/lib64-msvc-14.1" -DDEBUG=OFF -DPROFILE=OFF ..
 @endcode
 
-@note cmake will attempt to automatically download the ENSMALLEN dependency. If for some reason cmake can't download the dependency, you will need to manually download ENSMALLEN from http://ensmallen.org/ and extract it to "C:\mlpack\mlpack\deps\". Then, specify the path to ENSMALLEN using the flag: -DENSMALLEN_INCLUDE_DIR=C:/mlpack/mlpack/deps/ensmallen/include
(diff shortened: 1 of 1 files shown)


doc/guide/build_windows.hpp
```diff
@@ -91,67 +85,81 @@ This tutorial follows the second approach for simplicity.
...
+- You can generate the project using either cmake via command line or GUI. If you prefer to use GUI, refer to the \ref build_windows_appendix "appendix"
+- To use the CMake command line prompt, open the Command Prompt and navigate to "C:\mlpack\mlpack\build"
+- Run cmake:
+
+@code
+cmake -G "Visual Studio 15 2017 Win64" -DBLAS_LIBRARY:FILEPATH="C:/mlpack/mlpack/packages/OpenBLAS.0.2.14.1/lib/native/lib/x64/libopenblas.dll.a" -DLAPACK_LIBRARY:FILEPATH="C:/mlpack/mlpack/packages/OpenBLAS.0.2.14.1/lib/native/lib/x64/libopenblas.dll.a" -DARMADILLO_INCLUDE_DIR="C:/mlpack/armadillo/include" -DARMADILLO_LIBRARY:FILEPATH="C:/mlpack/armadillo/build/Debug/armadillo.lib" -DBOOST_INCLUDEDIR:PATH="C:/boost/" -DBOOST_LIBRARYDIR:PATH="C:/boost/lib64-msvc-14.1" -DDEBUG=OFF -DPROFILE=OFF ..
+@endcode
+
+@note cmake will attempt to automatically download the ENSMALLEN dependency. If for some reason cmake can't download the dependency, you will need to manually download ENSMALLEN from http://ensmallen.org/ and extract it to "C:\mlpack\mlpack\deps\". Then, specify the path to ENSMALLEN using the flag: -DENSMALLEN_INCLUDE_DIR=C:/mlpack/mlpack/deps/ensmallen/include
```
zoq: Is it intentional to write ensmallen in capital letters? When I talk about ensmallen I use the lower case from. My intention is to avoid the same situation that we see with mlpack which is mentioned in different flavors MLpack, MLPACK, mlpack, etc. I can see where MLPACK is coming from but I don't think MLpack is used in the codebase.

gmanlan: Nice catch. I guess it's just a mind trick. I got used to LAPACK, and mlpack is defined as "a machine learning analog to LAPACK", therefore the extrapolation. And ensmallen, well the only reference I found in mlpack's Readme.MD is the cmake flag which is uppercase. My bad. Will fix it soon. Thanks!

rcurtin:
Thanks @gmanlan, this is a really nice improvement and it'll be a lot more maintainable.  During merge I removed the Armadillo version number in c41c8aa01, since Armadillo versions change quickly and usually only the newest is available on the website.  I think the tutorial should still work fine with newer Armadillo versions.

gmanlan:
I agree. Thanks!

PR #1896, opened 2019-05-12
Fixed eig_sym() and chol() functions usage #1840 Issue

chandramoulirajagopalan:


chandramoulirajagopalan:
@zoq need some help with the test cases thing. I am not able to understand what went wrong.  Is it the fatal output which I generated or the multiplication mistakes.

Issue #1772, opened 2019-03-10
Segmentation fault in Python_QuickStart


pvarshney1729:
#### Issue description
I was facing Segmentation Fault (core dumped) error while running the Python QuickStart Code on "http://mlpack.org/docs/mlpack-3.0.4/doxygen/python_quickstart.html". Earlier the issue was with 

`output = mlpack.preprocess_split(input=dataset,
                                 input_labels=labels,
                                 test_ratio=0.3)`

which was resolved from issue #1694 . However the problem still persists, now with the line 

`output = mlpack.random_forest(training=training_set,
                              labels=training_labels,
                              print_training_accuracy=True,
                              num_trees=10,
                              minimum_leaf_size=3)`

#### My environment

 * version of mlpack: 3.0.4
 * operating system: Ubuntu 18.04
 * version of dependencies (Boost/Armadillo): Armadillo

Yashwants19:
I think so you must use git master for correct solution.

pvarshney1729:
Unfortunately, it building mlpack using git master isn't working. Moreover, it is giving Segmentation Fault in the line 
`output = mlpack.preprocess_split(input=dataset, input_labels=labels, test_ratio=0.3)`
also!

Shubhangi100:
@rcurtin I would like to work upon this issue, can I proceed?

rcurtin:
@Shubhangi100, sure, you are more than welcome to help out.  You might be interested in taking a look at the IRC logs from yesterday; I think I was helping someone with a similar problem.

Shubhangi100:
@rcurtin I have read the IRC logs, in my system there is .so file for preprocess_split and it is here '/usr/local/lib/python2.7/dist-packages/mlpack-3.0.5-py2.7-linux-x86_64.egg/mlpack' but still I am getting segmentation fault in 'output=mlpack.preprocess_split(input=dataset,input_labels=labels,test_ratio=0.3)'.
which was already resolved in #1694 . Now, how to proceed further?
'

rcurtin:
[...]

PR #1898, opened 2019-05-17
Python spacing fix

rcurtin:
This is a fix for #1772.

Essentially, what I found is that there was extra spacing for required matrix arguments in the generated `.pyx` files:

```
  if isinstance(copy_all_inputs, bool):
    if copy_all_inputs:
      SetParam[cbool](<const string> 'copy_all_inputs', copy_all_inputs)
      CLI.SetPassed(<const string> 'copy_all_inputs')
  else:
    raise TypeError("'copy_all_inputs' must have type 'bool'!")

  # Detect if the parameter was passed; set if so.
    input_tuple = to_matrix(input, dtype=np.double, copy=CLI.HasParam('copy_all_inputs'))
    if len(input_tuple[0].shape) > 2:
      input_tuple[0].shape = (input_tuple[0].shape[0], 1)
    input_mat = arma_numpy.numpy_to_mat_d(input_tuple[0], input_tuple[1])
    SetParam[arma.Mat[double]](<const string> 'input', dereference(input_mat))
    CLI.SetPassed(<const string> 'input')
    del input_mat
```

I fixed this and added tests to ensure that required matrices and vectors are handled correctly.  Also there was a little error in the quickstart code that produced some amusing output when doing the accuracy computation:

```
341557698 correct out of 30000 (1138525.66%).
```

I don't think our random forest is *that* good. :)

### DIFF
--- a/doc/guide/python_quickstart.hpp
+++ b/doc/guide/python_quickstart.hpp
@@ -99,7 +99,8 @@ output = mlpack.random_forest(input_model=random_forest,
 
 # Now print the accuracy.  The 'probabilities' output could also be used
 # to generate an ROC curve.
-correct = np.sum(output['predictions'] == test_labels)
+correct = np.sum(
+    output['predictions'] == np.reshape(test_labels, (test_labels.shape[0],)))
 print(str(correct) + ' correct out of ' + str(len(test_labels)) + ' (' +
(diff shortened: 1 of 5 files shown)


ShikharJ:
Thanks @rcurtin for the contribution! If you wish, I can send you some mlpack stickers (but wait, I don't have them :thinking: :thinking:).

PR #1899, opened 2019-05-19
Add GMM test for command-line and python bindings 

Yashwants19:
Address #1152


Yashwants19:
Hi @rcurtin can you please help me with some test.

Yashwants19:
Hi @rcurtin this is ready to review. Some test are failing in some ci travis test. Can you please help with those? Thank you :)

src/mlpack/tests/main_tests/gmm_train_test.cpp
```diff
@@ -0,0 +1,377 @@
...
+
+  GMM* gmm1 = CLI::GetParam<GMM*>("output_model");
+
+  arma::uvec sortedIndices = sort_index(gmm->Weights());
+
+  for (size_t k = 0; k < sortedIndices.n_elem; k++)
+    BOOST_REQUIRE(arma::norm(gmm->Component(sortedIndices[k]).Mean() -
+                      gmm1->Component(sortedIndices[k]).Mean()) > 1e-50 ||
+                  arma::norm(gmm->Component(sortedIndices[k]).Covariance() -
+                      gmm1->Component(sortedIndices[k]).Covariance()) > 1e-50);
```
rcurtin: Tiny style issue---since the `BOOST_REQUIRE()` line is wrapped, I think this would be easier to read if we used `{` and `}` for the `for` loop.

src/mlpack/tests/main_tests/gmm_train_test.cpp
```diff
@@ -0,0 +1,377 @@
...
+
+  mlpackMain();
+
+  GMM* gmm1 = CLI::GetParam<GMM*>("output_model");
+
+  BOOST_REQUIRE_EQUAL(gmm1->Gaussians(), (int) 2);
+}
+
+// Ensure that Noise affects the final result.
+BOOST_AUTO_TEST_CASE(GmmTrainNoisetest)
```
rcurtin: I guess this should be `NoiseTest` not `Noisetest`. :)

src/mlpack/tests/main_tests/gmm_train_test.cpp
```diff
@@ -0,0 +1,377 @@
...
+  SetInputParam("gaussians", (int) 2);
+  SetInputParam("trials", (int) 1);
+  SetInputParam("max_iterations", (int)-1); // Invalid.
+
+  Log::Fatal.ignoreInput = true;
+  BOOST_REQUIRE_THROW(mlpackMain(), std::runtime_error);
+  Log::Fatal.ignoreInput = false;
+}
+
+// Ensure that Trials must be greater than 0.
```
rcurtin: We should also test that the number of trials makes a difference.

src/mlpack/tests/main_tests/gmm_train_test.cpp
```diff
@@ -0,0 +1,377 @@
...
+  for (size_t k = 0; k < sortedIndices.n_elem; k++)
+  {
+    arma::mat diagCov(gmm->Component(sortedIndices[k]).Covariance());
+      for (size_t i = 0; i < diagCov.n_rows; i++)
+        for (size_t j = 0; j < diagCov.n_cols; j++)
+          if (i != j && diagCov(i, j) != (double) 0)
+            BOOST_FAIL("Covariance Are Not Diagonal");
+  }
+}
+
```
rcurtin: I can't think of a way to test when `no_force_positive` is given, but I guess the best we can do is just make sure that there is no crash when `no_force_positive` is specified.  (We can't test whether it changes the result, because most of the time, it won't significantly change the result.)

src/mlpack/tests/main_tests/gmm_probability_test.cpp
```diff
@@ -0,0 +1,72 @@
...
+
+// Checking the input and output dimensionality.
+BOOST_AUTO_TEST_CASE(GmmProbabilityDimensionality)
+{
+  arma::mat inputData(5, 10, arma::fill::randu);
+
+  GMM gmm(1, 5);
+  gmm.Train(std::move(inputData), 5);
+
+  arma::mat inputPoints(1, 8, arma::fill::randu);
```
rcurtin: I think your failure is that the dimensionality is wrong here---shouldn't it be `5`?

src/mlpack/tests/main_tests/gmm_train_test.cpp
```diff
@@ -0,0 +1,377 @@
...
+  mlpackMain();
+
+  GMM* gmm = CLI::GetParam<GMM*>("output_model");
+
+  ResetGmmTrainSetting();
+
+  SetInputParam("input", std::move(inputData));
+  SetInputParam("gaussians", (int) 2);
+  SetInputParam("refined_start", true);
+  SetInputParam("percentage", (double) 0.99);
```
rcurtin: `0.99` is probably too high for that code to work right with this dataset, I might suggest trying `0.2` or something instead.

src/mlpack/tests/main_tests/gmm_train_test.cpp
```diff
@@ -0,0 +1,377 @@
...
+    BOOST_REQUIRE(arma::norm(gmm->Component(sortedIndices[k]).Mean() -
+                      gmm1->Component(sortedIndices[k]).Mean()) > 1e-50 ||
+                  arma::norm(gmm->Component(sortedIndices[k]).Covariance() -
+                      gmm1->Component(sortedIndices[k]).Covariance()) > 1e-50);
+}
+
+// Ensure that Percentage affects the final result when refined_start is true.
+BOOST_AUTO_TEST_CASE(GmmTrainPercentageTest)
+{
+  arma::mat inputData(5, 10, arma::fill::randu);
```
rcurtin: You might consider a larger dataset here; more samples can increase the chance that the two models will be different.  I think that this applies across all the tests that check for changes.

src/mlpack/tests/main_tests/gmm_probability_test.cpp
```diff
@@ -0,0 +1,72 @@
...
+
+  SetInputParam("input", std::move(inputPoints));
+  SetInputParam("input_model", &gmm);
+
+  mlpackMain();
+
+  BOOST_REQUIRE_EQUAL(CLI::GetParam<arma::mat>("output").n_cols, 8);
+  BOOST_REQUIRE_EQUAL(CLI::GetParam<arma::mat>("output").n_rows, 1);
+}
```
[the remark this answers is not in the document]
Yashwants19: Hi @rcurtin its strange when we are not passing `input_file` we didn't get any `std::runtime_error` in the test but when we test this from terminal it terminates with `std::runtime_error`.

src/mlpack/tests/main_tests/gmm_generate_test.cpp
```diff
@@ -0,0 +1,93 @@
...
+
+  GMM gmm(1, 5);
+  gmm.Train(inputData, 5);
+
+  SetInputParam("input_model", &gmm);
+
+  Log::Fatal.ignoreInput = true;
+  BOOST_REQUIRE_THROW(mlpackMain(), std::runtime_error);
+  Log::Fatal.ignoreInput = false;
+}
```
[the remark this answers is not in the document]
Yashwants19: Done

src/mlpack/tests/main_tests/gmm_train_test.cpp
Yashwants19: Updated

src/mlpack/tests/main_tests/gmm_train_test.cpp
Yashwants19: Add the test for trials.

src/mlpack/tests/main_tests/gmm_probability_test.cpp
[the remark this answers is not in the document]
rcurtin: Right, the reason for this is that each binding language will enforce the required parameters itself.  So, for command-line bindings, the function `CLI::ParseCommandLine()` will check that all required parameters are passed.  For Python, the required parameters have no default option so Python itself will enforce that.  But for the test bindings, we didn't bother to enforce a check on the required parameters.  I guess, if you wanted, we could do that, but my assumption was that we'd just be sure to pass all the required parameters to the tests.

src/mlpack/tests/main_tests/gmm_probability_test.cpp
```diff
@@ -0,0 +1,87 @@
...
+  arma::mat inputData(5, 10, arma::fill::randu);
+
+  GMM gmm(1, 5);
+  gmm.Train(inputData, 5);
+
+  SetInputParam("input_model", &gmm);
+
+  Log::Fatal.ignoreInput = true;
+  BOOST_REQUIRE_THROW(mlpackMain(), std::runtime_error);
+  Log::Fatal.ignoreInput = false;
```
rcurtin: So, for that reason I described in the other comment, we don't need to include this test---we already know a user will have to pass both `input_model` and `input`.

src/mlpack/tests/main_tests/gmm_probability_test.cpp
```diff
@@ -0,0 +1,87 @@
...
+
+// Checking the input and output dimensionality.
+BOOST_AUTO_TEST_CASE(GmmProbabilityDimensionality)
+{
+  arma::mat inputData(5, 10, arma::fill::randu);
+
+  GMM gmm(1, 5);
+  gmm.Train(std::move(inputData), 5);
+
+  arma::mat inputPoints(1, 5, arma::fill::randu);
```
rcurtin: Shouldn't this have 5 rows, not 1 row?

Yashwants19: Updated

src/mlpack/tests/main_tests/gmm_probability_test.cpp
Yashwants19: I have remove test as suggested.

src/mlpack/tests/main_tests/gmm_train_test.cpp
Yashwants19: I have added the test you can take a look. :)

Yashwants19:
Hi @rcurtin `GmmTrainSamplingsTest` still fails 4 times in 2500 iterations.

src/mlpack/tests/main_tests/gmm_train_test.cpp
rcurtin: Looks good, thanks!

Yashwants19:
Hi @rcurtin can we start ci-travis again for confirmation?
I am testing these tests and I have reached over 750 iterations without any failure. 
Thank You :)

Yashwants19:
Hi @rcurtin from my side this is ready to review (merge). After 2000 iterations I didn't got any failure. If we have to add something more to this PR we can add.
Thanks :)

src/mlpack/tests/main_tests/gmm_train_test.cpp
```diff
@@ -0,0 +1,494 @@
...
+ * @author Yashwant Singh
+ *
+ * Test mlpackMain() of gmm_train_main.cpp.
+ *
+ * mlpack is free software; you may redistribute it and/or modify it under the
+ * terms of the 3-clause BSD license.  You should have received a copy of the
+ * 3-clause BSD license along with mlpack.  If not, see
+ * http://www.opensource.org/licenses/BSD-3-Clause for more information.
+ */
+#include<string>
```
rcurtin: We should probably add a space here between `#include` and `<string>`.

src/mlpack/tests/main_tests/gmm_train_test.cpp
```diff
@@ -0,0 +1,494 @@
...
+    BOOST_FAIL("Unable to load train dataset data_3d_mixed.txt!");
+
+  SetInputParam("input", inputData);
+  SetInputParam("gaussians", (int) 2);
+  SetInputParam("noise", (double) 0.0);
+
+  size_t seed = std::time(NULL);
+  mlpack::math::randGen.seed((uint32_t) seed);
+  srand((unsigned int) seed);
+  arma::arma_rng::set_seed(seed);
```
rcurtin: Now that we've verified that this seems to work reliably, I think we can remove this.  We can use `FixedRandomSeed()` instead.  I can make that change during merge.

src/mlpack/tests/main_tests/gmm_train_test.cpp
```diff
@@ -0,0 +1,494 @@
...
+  SetInputParam("gaussians", (int) 3);
+  SetInputParam("refined_start", true);
+  SetInputParam("samplings", (int) 1000);
+  SetInputParam("trials", (int) 1);
+  SetInputParam("max_iterations", (int) 500);
+
+  size_t seed = std::time(NULL);
+  mlpack::math::randGen.seed((uint32_t) seed);
+  srand((unsigned int) seed);
+  arma::arma_rng::set_seed(seed);
```
rcurtin: Same here, we can use `FIxedRandomSeed()`.

src/mlpack/tests/main_tests/gmm_train_test.cpp
```diff
@@ -0,0 +1,494 @@
...
+  SetInputParam("input", inputData);
+  SetInputParam("gaussians", (int) 2);
+  SetInputParam("refined_start", true);
+  SetInputParam("percentage", (double) 0.01);
+  SetInputParam("samplings", (int) 1000);
+
+  size_t seed = std::time(NULL);
+  mlpack::math::randGen.seed((uint32_t) seed);
+  srand((unsigned int) seed);
+  arma::arma_rng::set_seed(seed);
```
rcurtin: One more place for `FixedRandomSeed()`. :)

src/mlpack/tests/main_tests/gmm_train_test.cpp
```diff
@@ -0,0 +1,494 @@
...
+  SetInputParam("input", inputData);
+  SetInputParam("gaussians", (int) 8);
+  SetInputParam("refined_start", true);
+  SetInputParam("trials", (int) 2);
+  SetInputParam("samplings", (int) 10);
+
+  size_t seed = std::time(NULL);
+  mlpack::math::randGen.seed((uint32_t) seed);
+  srand((unsigned int) seed);
+  arma::arma_rng::set_seed(seed);
```
rcurtin: And I think this is the last one...

src/mlpack/tests/main_tests/gmm_train_test.cpp
```diff
@@ -0,0 +1,494 @@
...
+    BOOST_FAIL("Unable to load train dataset data_3d_mixed.txt!");
+
+  SetInputParam("input", inputData);
+  SetInputParam("gaussians", (int) 2);
+  SetInputParam("tolerance", (double) 1e-8);
+
+  size_t seed = std::time(NULL);
+  mlpack::math::randGen.seed((uint32_t) seed);
+  srand((unsigned int) seed);
+  arma::arma_rng::set_seed(seed);
```
rcurtin: Ah, no, I guess there was one more.

rcurtin:
Actually, when I restarted the build, it looks like one of the tests is still unstable and we should fix that before merge:

```
/home/travis/build/mlpack/mlpack/src/mlpack/tests/main_tests/gmm_train_test.cpp(286): fatal error in "GmmTrainTrialsTest": critical check arma::norm(gmm->Component(sortedIndices[k]).Mean() - gmm1->Component(sortedIndices[k]).Mean()) > 1e-50 || arma::norm(gmm->Component(sortedIndices[k]).Covariance() - gmm1->Component(sortedIndices[k]).Covariance()) > 1e-50 failed
```

Yashwants19:
Hi @ rcurtin: I have resolve the unstable tests.

rcurtin:
@Yashwants19 thanks again for the contribution. :+1:

PR #1901, opened 2019-05-23
Multiple Pole Balancing Environment

favre49:


src/mlpack/methods/reinforcement_learning/environment/multiple_pole_cart.hpp
```diff
@@ -0,0 +1,269 @@
...
+
+   private:
+    //! Locally-stored state data.
+    arma::mat data;
+  };
+
+  /**
+   * Implementation of action of Cart Pole.
+   */
+  enum Action
```
zoq: We can ignore the enum issues that is marked from the Static Analysis Check.

src/mlpack/methods/reinforcement_learning/environment/multiple_pole_cart.hpp
```diff
@@ -0,0 +1,269 @@
...
+   */
+  class State
+  {
+   public:
+    /**
+     * Construct a state instance.
+     * 
+     * @param numPoles The number of poles.
+     */
+    State(size_t numPoles)
```
zoq: For consistency, we should use ``const size_t`` here.

src/mlpack/methods/reinforcement_learning/environment/multiple_pole_cart.hpp
```diff
@@ -0,0 +1,269 @@
...
+    {
+      data = arma::mat(dimension, numPoles, arma::fill::zeros);
```
zoq: Just a note: ``data = arma::zeros(dimension, numPoles)`` works as well.

src/mlpack/methods/reinforcement_learning/environment/multiple_pole_cart.hpp
```diff
@@ -0,0 +1,269 @@
...
+  /**
+   * Implementation of action of Cart Pole.
+   */
+  enum Action
+  {
+    backward,
+    forward,
+
+    // Track the size of the action space.
+    size
```
zoq: Here is something missing.

src/mlpack/methods/reinforcement_learning/environment/multiple_pole_cart.hpp
```diff
@@ -0,0 +1,269 @@
...
+   * @param massCart The mass of the cart.
+   * @param massPole The mass of the pole.
+   * @param length The length of the pole.
+   * @param forceMag The magnitude of the applied force.
+   * @param tau The time interval.
+   * @param thetaThresholdRadians The maximum angle.
+   * @param xThreshold The maximum position.
+   */
+  MultiplePoleCart(const size_t poleNum,
+           const arma::vec poleLengths,
```
zoq: Can you align the parameters with the first one?

zoq: Might be a good idea to use a reference here.

src/mlpack/methods/reinforcement_learning/environment/multiple_pole_cart.hpp
```diff
@@ -0,0 +1,269 @@
...
+   *
+   * @param poleNum The number of poles
+   * @param gravity The gravity constant.
+   * @param massCart The mass of the cart.
+   * @param massPole The mass of the pole.
+   * @param length The length of the pole.
+   * @param forceMag The magnitude of the applied force.
+   * @param tau The time interval.
+   * @param thetaThresholdRadians The maximum angle.
+   * @param xThreshold The maximum position.
```
zoq: The ``doneReward`` parameter is missing.

favre49: Oops, I'll fix that in the CartPole task as well

src/mlpack/methods/reinforcement_learning/environment/multiple_pole_cart.hpp
favre49: I'm not sure I understand? This is the same format for Action as in other environments.

zoq: Ahh, never mind this is inside the enum.

src/mlpack/methods/reinforcement_learning/environment/multiple_pole_cart.hpp
```diff
@@ -0,0 +1,270 @@
...
+      gravity(gravity),
+      massCart(massCart),
+      forceMag(forceMag),
+      tau(tau),
+      thetaThresholdRadians(thetaThresholdRadians),
+      xThreshold(xThreshold),
+      doneReward(doneReward)
+  {
+    if (poleNum != poleLengths.n_elem)
+      Log::Fatal << "The number of lengths should be the same as the number of"
```
zoq: Can we wrap multiple lines with ``{}``to improve the readability.

src/mlpack/methods/reinforcement_learning/environment/multiple_pole_cart.hpp
```diff
@@ -0,0 +1,274 @@
...
+    // Calculate acceleration.
+    double totalForce = action ? forceMag : -forceMag;
+    double totalMass = massCart;
+    for (size_t i = 0; i < poleNum; i++)
+    {
+      double poleOmega = state.AngularVelocity(i);
+      double sinTheta = sin(state.Angle(i));
+      double cosTheta = cos(state.Angle(i));
+      totalForce += (poleMasses[i] * poleLengths[i] * poleOmega * poleOmega *
+          sinTheta) + 0.75 * poleMasses[i] * cosTheta * gravity * sinTheta;
```
manish7294: You could directly use `cosTheta * sinTheta` as `sin(2 * state.Angle(i))/2`, that could be a bit faster in longer computations.

src/mlpack/methods/reinforcement_learning/environment/multiple_pole_cart.hpp
```diff
@@ -0,0 +1,274 @@
...
+      totalMass += poleMasses[i] * (1 - 0.75 * cosTheta * cosTheta);
```
manish7294: and similar trignometric transformation can be applied here.

src/mlpack/methods/reinforcement_learning/environment/multiple_pole_cart.hpp
favre49: Good idea, didn't think of that. I'll change it soon

zoq: Ahh nice catch, I think once this is "fixed" this is ready to be merged.

src/mlpack/methods/reinforcement_learning/environment/multiple_pole_cart.hpp
manish7294: oh, looks like some misunderstanding here, probably because I might have missed out clarifying the exact point,. By this I meant, we could completely avoid calculating `cosTheta` and could just work with `sinTheta`, with probably less computation cost, as you are already computing `sinTheta`. So, maybe if you like you could use (0.25 - 0.75*sinTheta*sinTheta) in place of (1 - 0.75 * cosTheta * cosTheta). Let me know what you think about it.

favre49: Oh yes, good idea. I'll make the changes.

manish7294:
If you can take a look at the static code analysis checks, then It's probably ready to merge.

favre49:
The static code analysis warnings are because we have used enums as a boolean, which has been done in other parts of the RL code as well. zoq agreed that this is fine [here](https://github.com/mlpack/mlpack/pull/1901#discussion_r287084093). I think this is ready for merging?

manish7294:
No, problem then :)

zoq:
Actually there is one last comment, can you add the PR to https://github.com/mlpack/mlpack/blob/master/HISTORY.md?

zoq:

PR #1902, opened 2019-05-25
Make random forest tests more robust

rcurtin:
I was about to release mlpack 3.1.1 with the random forest test fixes, but I found that the tests failed with probability that was too high.  I spent a little time with the tests and I tuned them so that they don't fail now.  Specifically, the categorical dataset tests had significant problems; the reason is that 3 of the 5 dimensions are completely noise; so if the random forest selects sqrt(5) ~= 2 dimensions at each split, the probability of getting only noise is far too high.  This is a really bad case for random forests, and so it can be fixed simply by selecting more dimensions for each split.

I also increased the tolerance margins a little bit, so that the tests have no failures in 1000 trials (on my system at least).  It seems to be a very rare case that the random forest does worse than the decision tree, so I am not worried that there is a problem with the implementation.

Once we merge this I will go ahead and release mlpack 3.1.1 (but I want to wait until then so we don't get lots of "test failed" bug reports :)).

### DIFF
--- a/src/mlpack/tests/random_forest_test.cpp
+++ b/src/mlpack/tests/random_forest_test.cpp
@@ -222,7 +222,8 @@ BOOST_AUTO_TEST_CASE(UnweightedCategoricalLearningTest)
   arma::Row<size_t> testLabels = l.subvec(2000, 3999);
 
   // Train a random forest and a decision tree.
-  RandomForest<> rf(trainingData, di, trainingLabels, 5, 15 /* 15 trees */, 1);
+  RandomForest<> rf(trainingData, di, trainingLabels, 5, 25 /* 25 trees */, 1,
+      1e-7, MultipleRandomDimensionSelect(4));
   DecisionTree<> dt(trainingData, di, trainingLabels, 5, 5);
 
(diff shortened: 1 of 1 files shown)


src/mlpack/tests/random_forest_test.cpp
```diff
@@ -279,7 +280,8 @@ BOOST_AUTO_TEST_CASE(WeightedCategoricalLearningTest)
   arma::Row<size_t> fullLabels = arma::join_rows(trainingLabels, randomLabels);
 
   // Build a random forest and a decision tree.
-  RandomForest<> rf(fullData, di, fullLabels, 5, weights, 15 /* 15 trees */, 1);
+  RandomForest<> rf(fullData, di, fullLabels, 5, weights, 25 /* 20 trees */, 1,
```
favre49: This should be 25 trees in the comment right?

rcurtin:
> I know I'm not a member, but I hope its fine to give this review since i noticed this small mistake.

Of course it's okay! :)  Thanks for pointing out the comment, I'll fix it now. :+1:

PR #1903, opened 2019-05-28
Added support for Loading image

MuLx10:
Added functionality for loading images with the help of [stb_image.h](https://github.com/nothings/stb/blob/master/stb_image.h).
Currently supports `jpg`,`png`,`tga`, `bmp`, `psd`, `gif`, `hdr`, `pic`, `pnm`.

Reamining Task
- [x]  Find stb_image in CMakeList.
- [x]  Change the download url for stb_image.h
- [x] Save image.
- [x] Add documentation.


zoq:
We should update: https://github.com/mlpack/mlpack/blob/master/LICENSE.txt as well.

CMakeLists.txt
```diff
@@ -310,6 +311,29 @@ endif ()
 set(MLPACK_INCLUDE_DIRS ${MLPACK_INCLUDE_DIRS} ${ARMADILLO_INCLUDE_DIRS})
 set(MLPACK_LIBRARIES ${MLPACK_LIBRARIES} ${ARMADILLO_LIBRARIES})
 
+# Download stb_image for image loading.
+if (DOWNLOAD_STB_IMAGE)
+  set(STB_DIR "stb")
+  install(DIRECTORY DESTINATION ${CMAKE_BINARY_DIR}/deps/${STB_DIR})
+  file(DOWNLOAD https://raw.githubusercontent.com/nothings/stb/master/stb_image.h
```
zoq: I think instead of using always the latest version, we should pick a stable and host the file on mlpack, that way we can control which version is going to be included. @MuLx10 @rcurtin Let me know what you think.

CMakeLists.txt
```diff
@@ -392,6 +416,8 @@ find_package(Boost 1.49
       program_options
       unit_test_framework
       serialization
+      system
+      filesystem
```
zoq: I'm wondering if we could use C++11 filesystem support to do the same, boost dependencies was always a critic point.

src/mlpack/core/data/load_image.hpp
```diff
@@ -0,0 +1,144 @@
...
+
+#define STB_IMAGE_IMPLEMENTATION
+#include <stb_image.h>
+
+namespace mlpack {
+namespace data {
+
+/**
+ * Loads a matrix with image. It also supports loading image from 
+ * an entire directory.
```
zoq: Can we provide a simple example here; something like: https://github.com/mlpack/mlpack/blob/584f8533f20a4b94831d6b68ebb100bcbee573b7/src/mlpack/core/data/binarize.hpp#L26-L29

src/mlpack/core/data/load_image.hpp
```diff
@@ -0,0 +1,144 @@
...
+ * - PNG 1/2/4/8/16-bit-per-channel
+ * - TGA (not sure what subset, if a subset)
+ * - BMP non-1bpp, non-RLE
+ * - PSD (composited view only, no extra channels, 8/16 bit-per-channel)
+ * - GIF (*comp always reports as 4-channel)
+ * - HDR (radiance rgbE format)
+ * - PIC (Softimage PIC)
+ * - PNM (PPM and PGM binary only).
+ */
+
```
zoq: Picky style comment, we can remove the extra line here.

src/mlpack/core/data/load_image.hpp
```diff
@@ -0,0 +1,144 @@
...
+class LoadImage
+{
+public:
+  LoadImage();
```
zoq: Either we remove this one or we should add a simple comment here as well.

src/mlpack/core/data/load_image.hpp
```diff
@@ -0,0 +1,144 @@
...
+
+  /**
+   * LoadImage constructor.
```
zoq: Perhaps we can be a little bit more descriptive here, something like "Instantiate the LoadImage object with the given width, height, channels.".

src/mlpack/core/data/load_image.hpp
```diff
@@ -0,0 +1,144 @@
...
+   *
+   * @param width Matrix width for output matrix.
```
zoq: Hm, should this be "for the output matrix"?

src/mlpack/core/data/load_image.hpp
```diff
@@ -0,0 +1,144 @@
...
+   */
+  LoadImage(int width, int height, int channels);
+
+  /**
+   * Checks if the given image filename is supported.
+   *
+   * @param filename Name of the image file.
+   * @return Boolean value indicating success if it is an image.
+   */
+  bool isImageFile(std::string fileName);
```
zoq: Please use upper camel casing for all names, see https://github.com/mlpack/mlpack/wiki/DesignGuidelines#naming-conventions for more details. Also, we might want to use ``const std::string& fileName`` here to avoid a copy.

src/mlpack/core/data/load_image.hpp
```diff
@@ -0,0 +1,144 @@
...
+
```
zoq: Looks like we can remove the extra line here as well.

src/mlpack/core/data/load_image.hpp
```diff
@@ -0,0 +1,144 @@
...
+
+  /**
+   * Load the image file into the given matrix.
+   * Throws exceptions on errors.
```
zoq: Either we throw an exception or we return a boolean, I think armadillo does use the second, so maybe we should do the same here as well.

src/mlpack/core/data/load_image.hpp
```diff
@@ -0,0 +1,144 @@
...
+   * @param fileName Name of the image file.
+   * @param width Width of the image file.
+   * @param height Height of the image file.
+   * @param channels Channels of the image file.
+   * @param outputMatrix Matrix to load into.
+   * @return Boolean value indicating success or failure of load.
+   */
+  bool Load(std::string fileName,
+          arma::Mat<unsigned char>&& outputMatrix,
+          int *width,
```
zoq: Not sure, why to we pass a pointer, instead of using ``const size_t width``?

src/mlpack/tests/image_load_test.cpp
```diff
@@ -0,0 +1,99 @@
...
+/**
+ * Test the images is loaded correctly into the matrix.
+ */
+BOOST_AUTO_TEST_CASE(LoadImageIntoMatrixFromFile)
+{
+  data::LoadImage loader;
+
+  // Matrix to load contents of file into.
+  arma::Mat<unsigned char> img;
+  BOOST_REQUIRE(loader.Load("test_image.png", std::move(img)) == true);
```
zoq: I can't comment on the image file itself, but to keep the repo size low, do you think we could resize the image to reduce the size?

CMakeLists.txt
MuLx10: @zoq Yes using a stable version would be good. But how do I host file on mlpack?

CMakeLists.txt
MuLx10: Yes, I am finding difficulties regarding the same.

CMakeLists.txt
zoq: We can upload the file.

src/mlpack/core/data/load_image.hpp
MuLx10: Okay sure

src/mlpack/core/data/load_image.hpp
MuLx10: Okay sure.

src/mlpack/tests/image_load_test.cpp
MuLx10: I guess it would be better to add in the cmake. Will that be okay?

CMakeLists.txt
MuLx10: Also, the Jenkins build would fail.

src/mlpack/tests/image_load_test.cpp
zoq: The image file? I think it would be fine if we resize the image locally and push the updated version, I guess we could also tar the file as well, not sure that is needed. Right now the file is 463 KB, which I think could be reduced to 50 KB or something in that region, since we don't really care how the image looks like?

CMakeLists.txt
zoq: I think we installed every boost lib, so the Jenkins build is probably fine. But ideally, we can remove the boost filesystem dependency with a C++11 standard function.

zoq: It looks like we need boost filesystem for loading images from a folder, what about we could check for ``std::experimental::filesystem`` or ``std::filesystem`` and provide the functionality only if the feature is supported. Let me know what you think.

src/mlpack/core/data/load_image.hpp
MuLx10: Width, height and channels are returned via a reference in stb_image.

src/mlpack/core/data/load_image.hpp
MuLx10: Okay sure.

CMakeLists.txt
MuLx10: Okay sure

CMakeLists.txt
MuLx10: Where should I upload it?

src/mlpack/core/data/load_image_impl.hpp
```diff
@@ -0,0 +1,208 @@
...
+{
+  unsigned char *image;
+  char fileNameTemp[fileName.size() + 1];
+  strcpy(fileNameTemp, fileName.c_str());
+
+  if (!isImageFile(fileName))
+  {
+    std::ostringstream oss;
+    oss << "File type " << Extension(fileName) << " not supported.\n";
+    oss << "Cuurently it supports ";
```
zoq: Minor spelling issue: "Currently".

src/mlpack/core/data/load_image.hpp
zoq: I see, so what if image is going to be deconstructed, I think we will lose the information.

I see that we need the information later on if we like to reconstruct the image size. But perhaps we should just provide another functions that returns the information? Also, I think this doesn't cover images with different sizes?

src/mlpack/core/data/load_image.hpp
```diff
@@ -0,0 +1,150 @@
...
+  LoadImage();
+
+  /**
+   * Instantiate the LoadImage object with the image width, height, channels..
+   *
+   * @param width Matrix width for the output matrix.
+   * @param height Matrix height for the output matrix.
+   * @param channels Matrix channels for the output matrix.
+   */
+  LoadImage(int width, int height, int channels);
```
zoq: To be consistent with the rest of the codebase, let's use ``const size_t``  here.

src/mlpack/core/data/load_image.hpp
```diff
@@ -0,0 +1,150 @@
...
+
+  /**
+   * Checks if the given image filename is supported.
+   *
+   * @param filename Name of the image file.
+   * @return Boolean value indicating success if it is an image.
+   */
+  bool IsImageFile(std::string& fileName);
```
zoq: Can we use ``const std::string& filename`` here as well, would allow us to use ``IsImageFile("image.png");``

src/mlpack/core/data/load_image.hpp
```diff
@@ -0,0 +1,150 @@
...
+
+  /**
+   * Load the image file into the given matrix.
+   *
+   * @param fileName Name of the image file.
+   * @param outputMatrix Matrix to load into.
+   * @return Boolean value indicating success or failure of load.
+   */
+  bool Load(std::string& fileName, arma::Mat<unsigned char>&& outputMatrix);
```
zoq: See comment above.

src/mlpack/core/data/load_image.hpp
```diff
@@ -0,0 +1,150 @@
...
+   * Load the image file into the given matrix.
+   *
+   * @param fileName Name of the image file.
+   * @param width Width of the image file.
+   * @param height Height of the image file.
+   * @param channels Channels of the image file.
+   * @param outputMatrix Matrix to load into.
+   * @return Boolean value indicating success or failure of load.
+   */
+  bool Load(std::string& fileName,
```
zoq: Might be useful to provide an interface for ``arma::cube, arma::filed``, what do you think?

src/mlpack/core/data/load_image_impl.hpp
```diff
@@ -0,0 +1,202 @@
...
+#define MLPACK_CORE_DATA_LOAD_IMAGE_IMPL_HPP
+
+// In case it hasn't been included yet.
+#include "load_image.hpp"
+
+namespace mlpack {
+namespace data {
+
+LoadImage::LoadImage():
+        matrixWidth(-1),
```
zoq: If we switch to ``size_t`` we could use ``0`` here to get the same functionality.

src/mlpack/core/data/load_image_impl.hpp
```diff
@@ -0,0 +1,202 @@
...
+        matrixHeight(-1),
+        channels(3)
+{
+  fileTypes.clear();
```
zoq: To avoid code duplication, this could use the second constructor for the initialisation.

src/mlpack/core/data/load_image_impl.hpp
```diff
@@ -0,0 +1,202 @@
...
+
+LoadImage::~LoadImage()
+{
+  // Do nothing.
+}
+
+bool LoadImage::IsImageFile(std::string& fileName)
+{
+  // Iterate over all supported file types.
+  for (auto extension : fileTypes){
```
zoq: Hm, not the biggest fan of ``auto`` here, since the type is as simple as ``const std::string&``, might be easier to read, since I don't have to check what type this is.

src/mlpack/core/data/load_image_impl.hpp
```diff
@@ -0,0 +1,202 @@
...
+    std::transform(extension.begin(), extension.end(), extension.begin(),
+                    ::tolower);
```
zoq: The line should be tabbed twice, see https://github.com/mlpack/mlpack/wiki/DesignGuidelines#line-length-and-wrapping for more details.

src/mlpack/core/data/load_image_impl.hpp
```diff
@@ -0,0 +1,202 @@
...
+    std::ostringstream oss;
+    oss << "File type " << Extension(fileName) << " not supported.\n";
+    oss << "Currently it supports ";
+    for (auto extension : fileTypes)
+      oss << " " << extension;
+    oss << std::endl;
+    throw std::runtime_error(oss.str());
+  }
+
+  stbi_set_flip_vertically_on_load(true);
```
zoq: Should this be an option, a user could set?

src/mlpack/core/data/load_image_impl.hpp
```diff
@@ -0,0 +1,202 @@
...
+
+  if (!IsImageFile(fileName))
+  {
+    std::ostringstream oss;
+    oss << "File type " << Extension(fileName) << " not supported.\n";
+    oss << "Currently it supports ";
+    for (auto extension : fileTypes)
+      oss << " " << extension;
+    oss << std::endl;
+    throw std::runtime_error(oss.str());
```
zoq: We might just want to return false here.

src/mlpack/core/data/load_image.hpp
```diff
@@ -0,0 +1,150 @@
...
+#ifndef MLPACK_CORE_DATA_LOAD_IMAGE_HPP
+#define MLPACK_CORE_DATA_LOAD_IMAGE_HPP
+
+#include <string>
+#include <iostream>
+#include <mlpack/core.hpp>
+#include <mlpack/core/util/log.hpp>
+#include <mlpack/prereqs.hpp>
+
+#ifdef HAS_FILESYSTEM
```
zoq: We have to set the flag, similar to https://stackoverflow.com/a/54290906.

jeffin143:
> Added functionality for loading images with the help of [stb_image.h](https://github.com/nothings/stb/blob/master/stb_image.h).
> Currently supports `jpg`,`png`,`tga`, `bmp`, `psd`, `gif`, `hdr`, `pic`, `pnm`.


@MuLx10 Just curious , suppose just like loading images ,can we save too using this.? For suppose we have mat and simply we can it as an image.?

CMakeLists.txt
rcurtin: You can just add it to the mlpack.org repo under files/stb/ or something like this.  Once it's merged there the website will automatically rebuild it and it will be available.

CMakeLists.txt
```diff
@@ -310,6 +311,29 @@ endif ()
 set(MLPACK_INCLUDE_DIRS ${MLPACK_INCLUDE_DIRS} ${ARMADILLO_INCLUDE_DIRS})
 set(MLPACK_LIBRARIES ${MLPACK_LIBRARIES} ${ARMADILLO_LIBRARIES})
 
+# Download stb_image for image loading.
+if (DOWNLOAD_STB_IMAGE)
```
rcurtin: I know this makes things a little more complicated, but we should try to search the system first for STB before downloading it, and just use the system installed version if possible.  For instance, it is available in Debian.  In any case, you can take a look at how `DOWNLOAD_ENSMALLEN` is only used if we can't find `ensmallen.hpp` on the system.  You could do the same for `stb_image.h`.

LICENSE.txt
```diff
@@ -12,6 +12,10 @@ licensed under the Boost Software License, version 1.0.  This code is found in
 src/mlpack/core/boost_backport/ and more details on the licensing are available
 there.
 
+mlpack contains some usage of the source code of stb, which is licensed
```
rcurtin: Technically, `mlpack may contain some usage of the source of stb (depending on build configuration)`---sometimes the user may have it available on their system.  I know that is pedantic but it is good to be clear. :)

src/mlpack/core/data/load_image.hpp
```diff
@@ -0,0 +1,150 @@
...
+   *
+   * @param files A vector containing names of the image file to be loaded.
+   * @param outputMatrix Matrix to load into.
+   * @return Boolean value indicating success or failure of load.
+   */
+  bool Load(std::vector<std::string>& files,
+    arma::Mat<unsigned char>&& outputMatrix);
+
+  /**
+   * Load the image file into the given matrix.
```
rcurtin: It would probably be good to point out that this fails if `HAS_FILESYSTEM` is not available (I guess, that just means C++17 support is not available?).  And it may be better to throw a `std::runtime_error` instead of returning a bool.

src/mlpack/core/data/load_image.hpp
rcurtin: I think, in some cases, it will be `<filesystem>` instead.  You might need to do extra macro checking to see what is available for what compilers (and we should figure out what happens on Windows too because we have lots of Windows users).

src/mlpack/core/data/load_image_impl.hpp
```diff
@@ -0,0 +1,202 @@
...
+LoadImage::~LoadImage()
+{
+  // Do nothing.
+}
+
+bool LoadImage::IsImageFile(std::string& fileName)
+{
+  // Iterate over all supported file types.
+  for (auto extension : fileTypes){
+    std::transform(extension.begin(), extension.end(), extension.begin(),
```
jeffin143: Now , I didn't understand this line, we aldready have a vector , which has names in lower case written (line no 45) in same file , and then again iterating over it's item and converting it to lowercase doesn't make sense to me, may be I would have overlooked something . Sorry If so :)

src/mlpack/core/data/load_image.hpp
MuLx10: Yes, but how do I fill the cube.

src/mlpack/core/data/load_image_impl.hpp
MuLx10: Okay

src/mlpack/core/data/load_image_impl.hpp
MuLx10: OK

src/mlpack/core/data/load_image_impl.hpp
MuLx10: Okay

src/mlpack/core/data/load_image.hpp
zoq: Looks with gcc you also have to add ``-lstdc++fs`` when you want filesystem. But since we have this nice build bots we should be able to cover a bunch of different setups.

zoq: Looks like we could something similar as done here: https://github.com/gulrak/filesystem to check for support:

```c++
#if defined(__cplusplus) && __cplusplus >= 201703L && defined(__has_include) && __has_include(<filesystem>)
#include <filesystem>
namespace fs = std::filesystem;
#else
#include <ghc/filesystem.hpp>
namespace fs = ghc::filesystem;
#endif
```

src/mlpack/core/data/load_image_impl.hpp
MuLx10: But stb require interger pointers

src/mlpack/core/data/load_image.hpp
MuLx10: stb takes integer pointers

zoq:
I like where this Is going, excited to see this merged.

src/mlpack/core/data/load_image.hpp
```diff
@@ -0,0 +1,150 @@
...
+  bool LoadDIR(std::string& dirPath, arma::Mat<unsigned char>&& outputMatrix);
+
+  /**
+   * LoadImage default destructor.
+   */
+  ~LoadImage();
+
+ private:
+  // To store supported image types.
+  std::vector<std::string> fileTypes;
```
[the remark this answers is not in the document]
MuLx10: I guess map would be faster.

src/mlpack/core/data/load_image_impl.hpp
MuLx10: That was a mistake

CMakeLists.txt
MuLx10: Okay

src/mlpack/core/data/load_image.hpp
MuLx10: I tried with size_t and pointer conversion to int while passing to stb functions but there was out of memory error.

zoq: Okay, what about we use ``int`` internally and ``size_t`` for what we expose to the mlpack user base?

src/mlpack/core/data/load_image.hpp
zoq: One ``slice`` is the equivalent of a channel.

src/mlpack/core/data/load_image_impl.hpp
zoq: What I like to achieve here is to get a similar interface for the image loading feature as used for the rest of the codebase. If you like I can see if I can get it to work with ``size_t``.

src/mlpack/core/data/load_image_impl.hpp
```diff
@@ -0,0 +1,210 @@
...
+                   flipVerticalOnLoad,
+                   std::move(outputMatrix.row(i)));
+    Log::Info << "Loaded " << files[i] << std::endl << std::flush;
+  }
+  return status;
+}
+
+bool LoadImage::LoadDIR(const std::string& dirPath,
+                        bool flipVerticalOnLoad,
+                        arma::Mat<unsigned char>&& outputMatrix)
```
zoq: Do you think we can use a reference here, it might be confusing for the user to use a different API when loading images. Let me know what you think.

src/mlpack/core/data/load_image_impl.hpp
```diff
@@ -0,0 +1,210 @@
...
+    throw std::runtime_error(oss.str());
+  }
+
+  arma::Mat<unsigned char> img;
+  int width, height;
+  bool status = Load(files[0],
+                     flipVerticalOnLoad,
+                     std::move(img),
+                     &width, &height, &channels);
+  Log::Info << "Loaded " << files[0] << std::endl << std::flush;
```
zoq: Not sure ``std::flush`` is necessary here, maybe I missed something?

src/mlpack/core/data/load_image.hpp
```diff
@@ -0,0 +1,159 @@
...
+ * terms of the 3-clause BSD license.  You should have received a copy of the
+ * 3-clause BSD license along with mlpack.  If not, see
+ * http://www.opensource.org/licenses/BSD-3-Clause for more information.
+ */
+
+#ifndef MLPACK_CORE_DATA_LOAD_IMAGE_HPP
+#define MLPACK_CORE_DATA_LOAD_IMAGE_HPP
+
+#include <string>
+#include <iostream>
```
zoq: ``string`` and ``iostream`` should come with ``prereqs.hpp``.

src/mlpack/core/data/load_image.hpp
```diff
@@ -0,0 +1,159 @@
...
+#include <mlpack/core.hpp>
+#include <mlpack/core/util/log.hpp>
```
zoq: See comment above.

src/mlpack/core/data/load_image.hpp
```diff
@@ -0,0 +1,159 @@
...
+class LoadImage
+{
+ public:
+  /**
+   * LoadImage default constructor.
+   */
+  LoadImage();
+
+  /**
+   * Instantiate the LoadImage object with the image width, height, channels..
```
zoq: Looks like we can remove the second stop at the end here.

src/mlpack/core/data/load_image.hpp
```diff
@@ -0,0 +1,159 @@
...
+   */
+  LoadImage(int width, int height, int channels);
+
+  /**
+   * Checks if the given image filename is supported.
+   *
+   * @param filename Name of the image file.
+   * @return Boolean value indicating success if it is an image.
+   */
+  bool IsImageFile(const std::string& fileName);
```
zoq: Not sure the function name does express what this function does, since it checks if the image format is supported, perhaps it makes sense to call it ``ImageFormatSupported(...)``?

src/mlpack/core/data/load_image.hpp
```diff
@@ -0,0 +1,159 @@
...
+            int *channels);
+
+  /**
+   * Load the image file into the given matrix.
+   *
+   * @param files A vector containing names of the image file to be loaded.
+   * @param outputMatrix Matrix to load into.
+   * @return Boolean value indicating success or failure of load.
+   */
+  bool Load(std::vector<std::string>& files,
```
zoq: Can we use ``const std::vector<std::string>& files`` here as well?

src/mlpack/core/data/load_image.hpp
```diff
@@ -0,0 +1,159 @@
...
+            arma::Mat<unsigned char>&& outputMatrix,
+            int *width,
+            int *height,
+            int *channels);
+
+  /**
+   * Load the image file into the given matrix.
+   *
+   * @param files A vector containing names of the image file to be loaded.
+   * @param outputMatrix Matrix to load into.
```
zoq: The ``flipVerticalOnLoad`` parameter is missing, also not sure but since we load anyway, ``flipVertical`` might work as well?

MuLx10:
> > Added functionality for loading images with the help of [stb_image.h](https://github.com/nothings/stb/blob/master/stb_image.h).
> > Currently supports `jpg`,`png`,`tga`, `bmp`, `psd`, `gif`, `hdr`, `pic`, `pnm`.
> 
> @MuLx10 Just curious , suppose just like loading images ,can we save too using this.? For suppose we have mat and simply we can save it as an image.?

No, not currently, but I can add support for saving image as well.

src/mlpack/core/data/load_image.hpp
MuLx10: I'll try that.

src/mlpack/core/data/load_image.hpp
MuLx10: Okay sure.

jeffin143:
> > > Added functionality for loading images with the help of [stb_image.h](https://github.com/nothings/stb/blob/master/stb_image.h).
> > > Currently supports `jpg`,`png`,`tga`, `bmp`, `psd`, `gif`, `hdr`, `pic`, `pnm`.
> > 
> > 
> > @MuLx10 Just curious , suppose just like loading images ,can we save too using this.? For suppose we have mat and simply we can save it as an image.?
> 
> No, not currently, but I can add support for saving image as well.

@zoq , will it be good , If we add that support as well, we can visualize what image may be a gan model outputs or may be any other use case

zoq:
Would be nice to save images as well, not sure it should be part of the PR. @MuLx10 do you think this is a simple addition?

MuLx10:
> Would be nice to save images as well, not sure it should be part of the PR. @MuLx10 do you think this is a simple addition?

Small but a bit difficult addition. I will add the support for saving images in another PR.

src/mlpack/core/data/load_image.hpp
```diff
@@ -0,0 +1,161 @@
...
+  LoadImage();
+
+  /**
+   * Instantiate the LoadImage object with the image width, height, channels.
+   *
+   * @param width Matrix width for the output matrix.
+   * @param height Matrix height for the output matrix.
+   * @param channels Matrix channels for the output matrix.
+   */
+  LoadImage(size_t width, size_t height, size_t channels);
```
zoq: Do you mind to use ``const size_t`` for consistency with the rest of the codebase.

src/mlpack/core/data/load_image_impl.hpp
```diff
@@ -0,0 +1,219 @@
...
+  // Iterate over all supported file types.
+  for (auto extension : fileTypes)
+    if (extension == Extension(fileName))
+      return true;
+  return false;
+}
+
+bool LoadImage::Load(const std::string& fileName,
+                      bool flipVertical,
+                      arma::Mat<unsigned char>&& outputMatrix,
```
zoq: Did you see my comment about using a reference and the idea behind ``width, height, channels`` if we load images of different sizes.

MuLx10: @zoq Yes, however, there are some issues associated with resizing of an image via stb.

zoq: Not sure, do we want the option to resize images?

zoq:
@MuLx10 let me know if you need any help or if I should clarify anything.

src/mlpack/core/data/load_image.hpp
```diff
@@ -0,0 +1,161 @@
...
+   * Load the image file into the given matrix.
+   *
+   * @param fileName Name of the image file.
+   * @param flipVertical Flip the image vertical upon loading.
+   * @param outputMatrix Matrix to load into.
+   * @return Boolean value indicating success or failure of load.
+   */
+  bool Load(const std::string& fileName,
+            bool flipVertical,
+            arma::Mat<unsigned char>&& outputMatrix);
```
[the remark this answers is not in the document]
MuLx10: I'll add all the documentation in another PR where I'll add support to save image. It would be a continuation of this PR.

LICENSE.txt
MuLx10: I am not sure what I need to add.

jeffin143: > I am not sure what I need to add.

@MuLx10 , instead of` mlpack contains` --> `mlpack may contain`

src/mlpack/core/data/load_image_impl.hpp
```diff
@@ -0,0 +1,217 @@
...
+  free(image);
+  return true;
+}
+
+bool LoadImage::Load(const std::string& fileName,
+                    bool flipVertical,
+                    arma::Mat<unsigned char>&& outputMatrix)
+{
+  size_t width, height;
+  bool status = Load(fileName,
```
zoq: This looks strange, what about
```cpp
bool status = Load(fileName, flipVertical, std::move(outputMatrix),
    width, height, channels);
```

src/mlpack/core/data/load_image_impl.hpp
```diff
@@ -0,0 +1,217 @@
...
+      flipVertical,
+      std::move(outputMatrix),
+      width, height, channels);
+  if (!status)
+    return status;
+
+  Log::Info << width <<" "<< height <<" "<< channels << std::endl;
```
zoq: Do you think it would be helpful to add some more information here something like:

```cpp
Log::Info << "Image width: " << width <<"  height: "<< height <<" channels: " << channels << std::endl;
```

src/mlpack/core/data/load_image_impl.hpp
```diff
@@ -0,0 +1,217 @@
...
+  if (files.size() == 0)
+  {
+    std::ostringstream oss;
+    oss << "File vector is empty." << std::endl;
+    throw std::runtime_error(oss.str());
+  }
+
+  arma::Mat<unsigned char> img;
+  size_t width, height;
+  bool status = Load(files[0],
```
zoq: See the comment about the other ``Load`` call.

src/mlpack/core/data/load_image_impl.hpp
MuLx10: > > Added functionality for loading images with the help of [stb_image.h](https://github.com/nothings/stb/blob/master/stb_image.h).
> > Currently supports `jpg`,`png`,`tga`, `bmp`, `psd`, `gif`, `hdr`, `pic`, `pnm`.
> 
> @MuLx10 Just curious , suppose just like loading images ,can we save too using this.? For suppose we have mat and simply we can save it as an image.?

Now it can save image as well :smile:

MuLx10: @zoq not sure what you meant.

src/mlpack/tests/image_load_test.cpp
```diff
@@ -0,0 +1,162 @@
...
+  BOOST_REQUIRE_EQUAL(img.n_rows, 50 * 50 * 3); // width * height * channels.
+  BOOST_REQUIRE_EQUAL(img.n_cols, files.size());
+}
+
+/**
+ * Test if the image is saved correctly.
+ */
+BOOST_AUTO_TEST_CASE(SaveImageTest)
+{
+  data::Image loadAndSave;
```
zoq: I guess, we could also compare the image content if we use bmp (no compression). So I was thinking about some random matrix that we can save/load.

zoq:

src/mlpack/core/data/load_image_impl.hpp
```diff
@@ -0,0 +1,305 @@
...
+  }
+  else if ("hdr" == Extension(fileName))
+  {
+    status = stbi_write_hdr(fileName.c_str(), tempWidth, tempHeight,
+        tempChannels, reinterpret_cast<float*>(image));
+  }
+  else if ("jpg" == Extension(fileName))
+  {
+    status = stbi_write_jpg(fileName.c_str(), tempWidth, tempHeight,
+        tempChannels, image, 95);
```
zoq: I think we should make the compression configurable.

src/mlpack/core/data/load_image_impl.hpp
```diff
@@ -0,0 +1,305 @@
...
+    oss << std::endl;
+    throw std::runtime_error(oss.str());
+    return false;
+  }
+  if (inputMatrix.n_cols > 1)
+  {
+    std::cout << "Input Matrix contains more than 1 image." << std::endl;
+    std::cout << "Only the firstimage will be saved!" << std::endl;
+  }
+  unsigned char *image = inputMatrix.memptr();
```
zoq: This should be ``unsigned char* image = inputMatrix.memptr();``.

src/mlpack/core/data/load_image_impl.hpp
```diff
@@ -0,0 +1,305 @@
...
+        tempChannels, image, width*channels);
+  }
+  else if ("bmp" == Extension(fileName))
+  {
+    status = stbi_write_bmp(fileName.c_str(), tempWidth, tempHeight,
+        tempChannels, image);
+  }
+  else if ("tga" == Extension(fileName))
+  {
+    status = stbi_write_tga(fileName.c_str(), tempWidth, tempHeight,
```
zoq: Not sure, but I thought the compiler will convert the type automatically? Or maybe we should just use good old c-type casting.

src/mlpack/core/data/load_image_impl.hpp
```diff
@@ -0,0 +1,305 @@
...
+  return status;
+}
+
+
+bool Image::Save(const std::string& fileName,
+                 size_t width,
+                 size_t height,
+                 size_t channels,
+                 bool flipVertical,
+                 arma::Mat<unsigned char>&& inputMatrix)
```
zoq: Couldn't we  derive the number of channels from ``width, height`` and ``n_elem``?

MuLx10: Yes we can but image shape usually means width, height, channels. However if you want I will change.

src/mlpack/core/data/load_image_impl.hpp
MuLx10: To be on the safe side. However, it shows me an error, I guess that's because my compile is very old.

src/mlpack/core/data/load_image_impl.hpp
MuLx10: Sure

src/mlpack/core/data/load_image_impl.hpp
MuLx10: Okay,  I guess we should make it configurable with a default value. Let me know what you think?

src/mlpack/tests/image_load_test.cpp
MuLx10: Sure.

src/mlpack/core/data/load_image_impl.hpp
zoq: Same if you use ``status = stbi_write_tga(fileName.c_str(), (int) tempWidth, (int) tempHeight, ...)``?

src/mlpack/core/data/load_image_impl.hpp
zoq: Ideally we could derive width, height, channels, from the datatype, which I think is possible if we could use ``arma::cube, or arma::filed``, since most people just want to load/save the object the have. But I think for arma::mat this is difficult. Why not just provide two interface one that allows setting the channels and another one which derives the number of channels from the rest. Does that sound reasonable to you?

zoq:
Just saw your last two commit messages, for further commits it would be helpful if you could use a more descriptive commit message, makes it easier to search through the history and see right away what might have changed.

src/mlpack/core/data/load_image_impl.hpp
MuLx10: Should 
I also add support for cube and field?

src/mlpack/core/data/load_image_impl.hpp
MuLx10: Should I also add 

> Just saw your last two commit messages, for further commits it would be helpful if you could use a more descriptive commit message, makes it easier to search through the history and see right away what might have changed.

Apologies for that.

zoq:
No worries, just something I noticed 😃

src/mlpack/core/data/load_image_impl.hpp
zoq: I think for now ``arma::mat`` is fine.

src/mlpack/core/data/load_image_impl.hpp
zoq: Agreed, that way we can keep the same interface.

src/mlpack/core/data/load_image.hpp
```diff
@@ -0,0 +1,212 @@
...
+   * @param flipVertical Flip the image vertical upon loading.
+   * @param width Width of the image file.
+   * @param height Height of the image file.
+   * @param channels Channels of the image file.
+   * @param outputMatrix Matrix to load into.
+   * @return Boolean value indicating success or failure of load.
+   */
+  bool Load(const std::string& fileName,
+            bool flipVertical,
+            arma::Mat<unsigned char>&& outputMatrix,
```
zoq: Do you think we should go with a simple reference for every output? Might look cleaner to do:

```c++
arma::mat image
Load("test.png", false, image, ...);
```

also I was thinking it might be useful to put the ``flipVertical`` at the end, don't think this option is something that a user should pass each time.

src/mlpack/core/data/load_image_impl.hpp
MuLx10: no typecasting from size_t to int was giving an error . Similarly for char to string.

src/mlpack/core/data/load_image.hpp
```diff
@@ -0,0 +1,212 @@
...
+   * @param inputMatrix Matrix to save images from.
+   * @return Boolean value indicating success or failure.
+   */
+  bool Save(const std::string& fileName,
+            size_t width,
+            size_t height,
+            size_t channels,
+            bool flipVertical,
+            arma::Mat<unsigned char>&& outputMatrix,
+            size_t quality = 95);
```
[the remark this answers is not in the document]
MuLx10: added.

src/mlpack/core/data/load_image.hpp
MuLx10: Using simple reference may require typecasting at various stages. like when I am using `inputmartix.col(i)` it needs to be typecasted to `arma::Mat<uchar>` for every col. Hence, I did not use a simple reference. However, if you want I can add a simple reference.

zoq: Not sure I get the example, can you provide an actual example where you have to cast if you change the interface from ``arma::Mat<unsigned char>&& outputMatrix,`` to ``arma::Mat<unsigned char>& outputMatrix,``? I looked over the code and I think if you just use a reference and remove the move call it should be just fine.

MuLx10: Here https://github.com/mlpack/mlpack/blob/4eae4d4d8faf6a450fd2a78f73f78adc92e9b62d/src/mlpack/core/data/load_image_impl.hpp#L194
This is of type subview and needs to be converted to Mat, else the compiler throws an error.

zoq: Now I see what you mean, what about we do:

```c++
arma::Mat<unsigned char>(outputMatrix.colptr(i), outputMatrix.n_rows, 1, false, true)
```

zoq:
I was thinking instead of providing the functionality as an extra class it would be nice to have this inside the existing ``data::Load``, ``data:.Save`` functions. The only issue I see is that it would have a slightly different interface since we have to get additional information like image width and height. But before I think more about the idea, I'd like to get your thoughts about the idea.

MuLx10:
@zoq I initially thought of the same thing but couldn't, since the input parameters were very different.

zoq:
Hm, I think we could pass some sort of ``DatasetInfo`` object, to get the additional information, I think that would also allow us to load images of different sizes in a row. If you like I can sketch up an interface?

MuLx10:
@zoq Please verify if the interface is correct. 
I shall complete the documentation by today.

zoq:
Do you think we should put the loader functionality into the ``data::Load/data::Save``function? Personally I would only support one way to load/save images.

MuLx10:
@zoq I have trying to do that for a few days.

Do you want me to remove the second API 
```
const std::vector<std::string>& files,
          arma::Mat<eT>& matrix,
          ImageInfo& info,
          const bool fatal,
          const bool transpose
```

I know it is optional but usually, we have to load more than one image and it would be good if the user provides a vector of file names.

zoq:
Sorry for the confusion, I was talking about:

 ```c++
data::Image loadAndSave; 
loadAndSave.Load(..);
```

since now it's also possible to do ``data::Load(...)``without creating an Image instantiation first.

**Comment by MuLx10** [comment]
Oh sorry, I get it..
Is the interface correct. Should I also add the documentation for the new interface.

**Comment by zoq** [comment]
The ``data::Load, data::Save`` interface looks good to me, yes. If we can put the functionality from the ``data::Image`` class directly into the ``data::Load, data::Save`` method I think this is about ready, we have to update the documentation as well as you said.

**Comment by MuLx10** [comment]
I will give one more try to incorporate the functionality in data::Load.

**Comment by zoq** [comment]
Sounds good, thanks!

File: doc/tutorials/image/image.txt [hunk 51]
```diff
@@ -0,0 +1,210 @@
+/*!
+@file image.txt
+@author Mehul Kumar Nirala
+@brief Tutorial for how to load and save images in mlpack.
+
+@page imagetutorial Image Utilities tutorial
+
+@section intro_imagetut Introduction
+
+Image dataset are becoming increasingly popular in deep learning.
```
zoq: Should this be "image datasets"?

doc/tutorials/image/image.txt
```diff
@@ -0,0 +1,210 @@
...
+@author Mehul Kumar Nirala
+@brief Tutorial for how to load and save images in mlpack.
+
+@page imagetutorial Image Utilities tutorial
+
+@section intro_imagetut Introduction
+
+Image dataset are becoming increasingly popular in deep learning.
+
+mlpack currently implements saving and loading of image. It uses [stb/](https://github.com/nothings/stb) to achieve this.
```
zoq: Perhaps we should alter the wording here to something like: "mlpack's image saving/loading functionality is based on [stb/](https://github.com/nothings/stb)." Let me know what you think.

doc/tutorials/image/image.txt
```diff
@@ -0,0 +1,210 @@
...
+ - \ref intro_imagetut
+ - \ref toc_imagetut
+ - \ref model_api_imagetut
+ - \ref imageinfo_api_imagetut
+ - \ref load_api_imagetut
+ - \ref save_api_imagetut
+
+@section model_api_imagetut Model API
+
+Image utilities supports loading and saving of image. Images can also be loaded from a directory if CXX version >= 17 as it requires <filesystem> header to navigate a directory.
```
zoq: Not sure everyone knows what CXX version is, maybe we should go with: "if the compiler supports c++17 as the <filesystem> functionality is required"?

doc/tutorials/image/image.txt
```diff
@@ -0,0 +1,210 @@
...
+
+It supports filetypes "jpg", "png", "tga","bmp", "psd", "gif", "hdr", "pic", "pnm" for loading and "jpg", "png", "tga", "bmp", "hdr" for saving.
+
+The datatype associated is unsigned char to support RGB values in the range 1-255. To feed data into the network typecast of `arma::Mat` may be required. Images are stored in matrix as (width * height * channels, n_images). Therefore imageMatrix.col(0) would be first image if images are loaded in imageMatrix.
```
zoq: Should this be "would be the first image if [...]"?

doc/tutorials/image/image.txt
```diff
@@ -0,0 +1,210 @@
...
+             const bool transpose);
+@endcode
+
+Loading a test image. It also fills up the ImageInfo class object.
+@code
+data::ImageInfo info;
+data::Load("test_image.png", matrix, info, false);
+@endcode
+
+If the size of image is known beforehand, it can be passed into the constructor of ImageInfo to set the size of outputMatrix(the matrix into which the image is loaded).
```
zoq: Might be a good idea to mention the fatal parameter here as well or just use the default value.

zoq:
Let's rerun the travis build.

src/mlpack/core/data/image_info.hpp
```diff
@@ -0,0 +1,106 @@
...
+class ImageInfo
+{
+ public:
+  /**
+   * Instantiate the ImageInfo object with the given image width, height,
+   * channels parameter.
+   *
+   * @param width Image width.
+   * @param height Image height.
+   * @param channels number of channels in the image.
```
zoq: Just noticed, this is missing the ``quality`` parameter.

src/mlpack/core/data/save_impl.hpp
```diff
@@ -283,6 +283,126 @@ bool Save(const std::string& filename,
...
+      status = stbi_write_jpg(filename.c_str(), tempWidth, tempHeight,
+          tempChannels, image, tempQuality);
+    }
+  }
+  catch (std::exception& e)
+  {
+    Timer::Stop("saving_image");
+    if (fatal)
+      Log::Fatal << e.what() << std::endl;
+    else
```
zoq: We can remove the else here, ``Fatal``will exit immediately, so:

```c++
Timer::Stop("saving_image");
if (fatal)
  Log::Fatal << e.what() << std::endl;

Log::Warn << e.what() << std::endl;
return false;
```

will work as well.

src/mlpack/tests/image_load_test.cpp
```diff
@@ -0,0 +1,78 @@
...
+  BOOST_REQUIRE_EQUAL(matrix.n_cols, 1);
+}
+
+/**
+ * Test if the image is saved correctly using API.
+ */
+BOOST_AUTO_TEST_CASE(SaveImageAPITest)
+{
+  data::ImageInfo info;
+  info.Width() = info.Height() = 5;
```
zoq: Not sure, but using the ``ImageInfo`` constructor might be cleaner, what do you think?

src/mlpack/tests/image_load_test.cpp
```diff
@@ -0,0 +1,78 @@
...
+  info.Channels() = 3;
+  info.Quality() = 90;
+
+  srand(time(0));
```
zoq: We can remove the ``srand`` here, this would help us to run the test again and again with the same seed.

src/mlpack/core/data/image_info.hpp
MuLx10: okay.

src/mlpack/core/data/save_impl.hpp
MuLx10: I did add it because it was it the rest of the code.

src/mlpack/tests/image_load_test.cpp
MuLx10: Oh, I guess constructor will be good.

src/mlpack/tests/image_load_test.cpp
MuLx10: okay.

zoq:

src/mlpack/core/data/image_info.hpp
```diff
@@ -0,0 +1,107 @@
...
+ */
+inline bool ImageFormatSupported(const std::string& fileName,
+                                 bool save = false);
+
+class ImageInfo
+{
+ public:
+  /**
+   * Instantiate the ImageInfo object with the given image width, height,
+   * channels parameter.
```
zoq: channels and quality parameter.

src/mlpack/core/data/image_info.hpp
```diff
@@ -0,0 +1,107 @@
...
+  size_t channels;
+
+  // Compression of the image if saved as jpg (0 - 100).
+  size_t quality;
+
+  // Image format.
+  std::string format;
+
+  // Flip the image vertical upon loading/saving.
+  bool flipVertical;
```
zoq: Looks like the parameter is unused.

src/mlpack/core/data/image_info.hpp
```diff
@@ -0,0 +1,107 @@
...
+  size_t width;
+
+  // To store the number of channels in the image.
+  size_t channels;
+
+  // Compression of the image if saved as jpg (0 - 100).
+  size_t quality;
+
+  // Image format.
+  std::string format;
```
zoq: Looks like the parameter is unused.

zoq:
Can you take a look at the warning (e.g. reorder) that are raised by the travis build: https://travis-ci.org/mlpack/mlpack/jobs/558830270

doc/tutorials/image/image.txt
zoq: Can we address the comment?

doc/tutorials/image/image.txt
```diff
@@ -0,0 +1,185 @@
...
+
+@code
+  data::ImageInfo info;
+  std::vector<std::string>> files{"test_image1.bmp","test_image2.bmp"};
+  data::load(files, matrix, info, false, true);
+@endcode
+
+@section save_api_imagetut Save
+
+Save images expects a matrix of type unsigned char in the form (width * height * channels, n_images).
```
zoq: ``(width * height * channels, n_images)`` not sure about the ``n_images`` parameter, maybe ``NumberOfImages`` is more clear?

doc/tutorials/image/image.txt
```diff
@@ -0,0 +1,185 @@
...
+@code
+  data::ImageInfo info;
+  info.width = info.height = 25;
+  info.channels = 3;
+  info.quality = 90;
+  std::vector<std::string>> files{"test_image1.bmp", "test_image2.bmp"};
+  data::Save(files, matrix, info, false, true);
+@endcode
+
+Multiple images are saved according to the vector of file name specified.
```
zoq: Do you mean "accordingly to the vector of filenames"?

src/mlpack/core/data/image_info.hpp
```diff
@@ -0,0 +1,101 @@
...
+/**
+ * Checks if the given image filename is supported.
+ *
+ * @param filename Name of the image file.
+ * @return Boolean value indicating success if it is an image.
+ */
+inline bool ImageFormatSupported(const std::string& fileName,
+                                 const bool save = false);
+
+class ImageInfo
```
zoq: Can we add a description for the ``ImageInfo`` class as well?

MuLx10: Sure.

doc/tutorials/image/image.txt
MuLx10: Oh!  yes.

doc/tutorials/image/image.txt
MuLx10: okay, I'll change

MuLx10:
@zoq I am still getting unused function warnings. Is there a way to suppress it?

zoq:
I see, we could add something like this: https://stackoverflow.com/a/11125034 to the file, what do you think.

MuLx10:
> I see, we could add something like this: https://stackoverflow.com/a/11125034 to the file, what do you think.

I will try it.

zoq:
Another solution might be to provide a minified version, that only contains the functions we need.

zoq:
@MuLx10 Any update about the header warnings?

MuLx10:
Sorry I was working on sentiment analysis. I will rectify it by this week.

CMakeLists.txt
```diff
@@ -78,6 +78,10 @@ if(CMAKE_COMPILER_IS_GNUCC OR "${CMAKE_CXX_COMPILER_ID}" STREQUAL "Clang")
   set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -Wall -Wextra")
 endif()
 
+# To remove unused functions warnings.
+set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wno-unused-function")
```
zoq: That works for me, @rcurtin does this work for you as well?

MuLx10:
@zoq please restart the build, thank you.

zoq:
Restarted.

zoq:

zoq:

zoq:
@MuLx10 Sorry for the Jenkins related comments, I used the PR to figure out an issue with the Static Analysis job.

zoq:
Looks good the, travis test issue isn't caused by this PR.

doc/tutorials/image/image.txt
```diff
@@ -0,0 +1,185 @@
...
+ - \ref intro_imagetut
+ - \ref toc_imagetut
+ - \ref model_api_imagetut
+ - \ref imageinfo_api_imagetut
+ - \ref load_api_imagetut
+ - \ref save_api_imagetut
+
+@section model_api_imagetut Model API
+
+Image utilities supports loading and saving of image.
```
rcurtin: I think this should be `images`. :)  Also, would it be useful here to name what the method is that does the loading?  The description below says what the code in this PR does, but at this point a reader wouldn't know exactly what the name of the utility functions are (or what namespace they're in).

src/mlpack/core/data/load.hpp
```diff
@@ -287,12 +288,33 @@ bool Load(const std::string& filename,
...
+          const bool fatal = false,
+          const bool transpose = true);
+
+template<typename eT>
+bool Load(const std::vector<std::string>& files,
+          arma::Mat<eT>& matrix,
+          ImageInfo& info,
+          const bool fatal = false,
+          const bool transpose = true);
+#endif // HAS_STB.
```
rcurtin: It could be really useful to also have something like

```
#else 
struct ImageInfo { };

template<typename eT>
bool Load(const std::string& filename,
          arma::Mat<eT>& matrix,
          ImageInfo& info,
          const bool fatal = false,
          const bool transpose = true)
{
  throw std::runtime_error("Load(): HAS_STB is not defined, so STB is not available and images cannot be loaded!");
}

template<typename eT>
bool Load(const std::vector<std::string>& files,
          arma::Mat<eT>& matrix,
          ImageInfo& info,
          const bool fatal = false,
          const bool transpose = true)
{
  throw std::runtime_error("Load(): HAS_STB is not defined, so STB is not available and images cannot be loaded!");
}
```

Also it would be useful to add documentation to the top of the definitions here, so that Doxygen will pick it up.

src/mlpack/core/data/save.hpp
```diff
@@ -90,6 +91,22 @@ bool Save(const std::string& filename,
...
+          const bool fatal = false,
+          const bool transpose = true);
+
+template<typename eT>
+bool Save(const std::vector<std::string>& files,
+          arma::Mat<eT>& matrix,
+          ImageInfo& info,
+          const bool fatal = false,
+          const bool transpose = true);
+#endif // HAS_STB.
```
rcurtin: Same comment here as in `Load()` when `HAS_STB` is not defined.

src/mlpack/core/data/save_impl.hpp
rcurtin: Agreed, I think I wrote a lot of the offending code with the `if` and `else`, but the `else` isn't needed.  If I see it in the future I'll change it and open a PR. :)

rcurtin:
Ooh, also, we should update `HISTORY.md`. :)

zoq:
Thanks for looking into the comments; the last commit introduced some style issues, I guess if you can fix those issues as well, we are about ready to merge this in.

zoq:
@MuLx10 let me know if you need any help with the style issues.

zoq:
Nice, I just restarted the travis build since there was an issue with the macOS build, not sure, it is related with the changes from the code. But easy to rerun the test.

src/mlpack/core/data/save.hpp
MuLx10: Okay.

rcurtin: @MuLx10 Thanks---let us know when you're done with further changes for this PR.  I think then we can merge it. :+1:

MuLx10:
@rcurtin @zoq I have to make a few changes, please don't merge the PR now. Thanks.

MuLx10:
@rcurtin @zoq I have added the changes. Once the build is complete, we are good to go.

rcurtin:
@MuLx10 awesome contribution, thanks so much for the hard work on this. :100:

PR #1905, opened 2019-05-29
Fix serialization of glimpse, maxPooling and meanPooling layers.

walragatver:
Hi @zoq, @ShikharJ,
While working on GAN serialization I have found an error in `mean pooling` layer. It was producing an `integer divide by zero` error after serialization at `line 60` in `mean_pooling_impl.hpp`. This is because the parameters  `inputWidth`, `inputHeight`, `outputWidth`, `outputHeight` were not serialize. 
I have also modified serialization of some other layers namely `max_pooling` and `glimpse`. Let me know your thoughts regarding the same.


walragatver:
Sorry, it looks like the `floor` parameter in  `MaxPool` and `MeanPool` is also needed to be serialized. So I have also serialized it.

zoq:
Thanks again.

Issue #66, opened 2014-12-29
fastlib-stl/allknn twice as slow as trunk/allknn


rcurtin:
**Reported by rcurtin on 17 Oct 40820561 21:20 UTC**
The trunk build takes 2177 seconds to calculate the 7 nearest neighbors of every point of a 100k dataset, whereas the fastlib-stl branch takes 3868 seconds to do the same calculation.

From the trunk build:

```
:[ ryan @ greenstreet ]: ~/work/fastlab/trunk/build $ bin/allknn --input_file=/home/ryan/dr1.csv --k=7 --output_file=/home/ryan/dr1-out.csv
[Loaded input data from file /home/ryan/dr1.csv.
[.](.]) Building reference tree...
[Computing 7 nearest neighbors...
[.](.]) Writing results to /home/ryan/dr1-out.csv...
/info/sys/node/name:R greenstreet
/info/sys/arch/name:R x86_64
/info/sys/kernel/name:R Linux
/info/sys/kernel/release:R 2.6.32-trunk-amd64
/info/sys/kernel/build:R %231%20SMP%20Sun%20Jan%2010%2022%3A40%3A40%20UTC%202010
/debug/verbosity_level:D 1
/debug/print_got_heres:D t
/debug/print_warnings:D t
/debug/abort_on_nonfatal:D f
/debug/pause_on_nonfatal:D f
/debug/print_notify_locs:D f
/total_time/real:T 2177.267186
/total_time/user:T 2175.47
/total_time/sys:T 0.82
/output_file:P %2Fhome%2Fryan%2Fdr1-out.csv
/input_file:Q %2Fhome%2Fryan%2Fdr1.csv
/leaf_size:D 20
/tree_building/real:T 4.589253
/tree_building/user:T 4.56
/tree_building/sys:T 0.01
/k:Q 7
/computing_neighbors/real:T 2169.501983
/computing_neighbors/user:T 2167.86
/computing_neighbors/sys:T 0.6899999999999999
/fx/no_docs_nagging:D f
/fx/timing:D f
/fx/rusage:D f
/fx/no_output_types:D f
/fx/silent:D f
```

From the branch build:

```
:[ ryan @ greenstreet ]: ~/work/fastlab/branches/fastlib-stl/build $ bin/allknn --reference_file=/home/ryan/dr1.csv --knns=7
[fx/fx.c:fx__check_mod_type:347: Input parameter "/reference_file" is not documented.
[...]

PR #1908, opened 2019-05-29
[GSoC] [WIP] NeuroEvolution of Augmenting Topologies

favre49:
This is a WIP PR for an implementation of NEAT and its tests. I'll be adding progress updates in the comments.


favre49:
I've just added the test suite. The `BOOST_REQUIRE` lines are still placeholders, since I am not sure what to keep as the test conditions. I've also scaled down the variables in the Double Pole balancing cases, there are two things i wanted to clarify:
- I've scaled the velocity variables by dividing by 100, since we don't have bounds on them. Is this fine?
- If this is fine, should i do the same in all the test cases? I think that would be appropriate, but perhaps we should wait to see what kind of results the final implementation gives on unscaled variables? Let me know what you think.

zoq:
>  I think that would be appropriate, but perhaps we should wait to see what kind of results the final implementation gives on unscaled variables? Let me know what you think.

NEAT should be able to evolve a normalisation step if needed, but I agree it might be useful, to scale the input, perhaps we should add a flag that allows us to train on scaled or unscaled data.

favre49:
Hmmm what exactly do you mean by that?

favre49:
I've added implementations of the Gene classes. Some additional documentation and style fixing needs to be done, I'll do that soon.

zoq:
> Hmmm what exactly do you mean by that?
I don't think scaling is something we have to do, but I also think it could help to accelerate the training process. As you already said, let's keep it in for now and once we have something for testing we can check if it makes a difference.

src/mlpack/methods/neat/gene.hpp
```diff
@@ -0,0 +1,101 @@
...
+
+#include <mlpack/prereqs.hpp>
+
+namespace mlpack{
+namespace neat /** NeuroEvolution of Augmenting Topologies */ {
+
+/**
+ * Connection gene class.
+ */
+class ConnectionGene
```
zoq: Do you think you could draw the overall class connection infrastructure (what class depends on which, connection between classes) on a paper? It doesn't have to be fancy.

favre49: IMG_20190601_193935015_HDR
Do you mean something like this?

zoq: Yes, thanks that is really helpful. So I think you will start with the the implementation of the Genome? You already have the interface for the ConnectionGene. which will be part of the Genome that also holds a list of Genes? If you like I can sketch up a high level interface?

favre49: Yup, I'm working on the implementation of the Genome class right now. In the genome I am planning to maintain two lists of Connection Genes, where one will be an adjacency list to help with evaluating the phenomes and the other is a std::vector storing the genes sorted by innovation ID. As for the interface, besides what we discussed Input() and Output()) what else would the user need? The population would need to use a Mutate() and Crossover() function. I don't think there is much else.

manish7294: Maybe here instead of storing genes as a vector, you can just store pointers to genes sorted by innvoation ID.

favre49: What would be the benefit to this over storing the genes directly? I think if we use pointers, I would have to dynamically allocate the genes, which would really complicate things with having to concern ourselves with memory management. Is there something I'm missing?

manish7294: I intend to find a way to reduce the space usage here. But, you are right about the issue of memory management, I completely missed out this point. And if I recall correctly, we usually don't do pointers here in mlpack.

zoq: You could use ``ptr_vector``, or I guess you can also use move semantics to avoid another copy, but at the end you have to store the object somewhere.

> As for the interface, besides what we discussed Input() and Output()) what else would the user need? 


I was thinking about a more in depth internal class interface, this class will hold this object, this class is called by another class using this interface, etc.

favre49: Oh, I understand. I can't think of much, besides making a class wrapping all the gene storage containers, which makes it easier to pass to the Phenome class? Maybe also a class that stores the "config" of the genome? Like the means for the weights and biases, input and output node counts, maximum complexity for pruning, etc. Once again we wouldn't have to use a lot of parameters to pass basic info to other methods and classes.

zoq: Agreed, we should defiantly think about how we want to load a config, but don't think this has priority right now. If you already have a ``Gene`` class in mind, please feel free to commit something; makes it easier to discuss the interface anyway. Also, personally I would go with a simple container e.g. ``std::vector`` for now @manish7294 do you think that is reasonable?

manish7294: Sure, we can go with vector. And can always at later point of time, see what can be improved.

favre49:
I have added the Genome class as well as the class for evaluation of acyclic networks. Style checks and some documentation still remain. I also have to prevent cycles from forming during mutation if the `isAcyclic` param is set to `true`.

How the acyclic network evaluation works is:
1. Assign depth to all of the nodes. Input and bias nodes are of course depth 0. In the case that multiple depths are found, we choose the maximum depth.
2. Based on these depths, nodes are split into layers.
3. Then, we activate the nodes layer by layer and take the output.

If there are any inefficiencies, please point them out. Another thing I will have to do is make some changes in the test suite, since the genomes now have the activation function as a template parameter.

src/mlpack/methods/neat/acyclic_net.hpp
```diff
@@ -0,0 +1,74 @@
...
+   * 
+   * @param NodeGeneList A vector storing all the node IDs.
+   * @param DirectedGraph A map of maps storing connection genes, whose first
+   *    key is the source ID, and second key is the target ID.
+   * @param actFn The activation function.
+   * @param inputNodeCount The number of input nodes.
+   * @param outputNodeCount The number of output nodes.
+   * @param bias The bias.s
+   */
+  AcyclicNet(std::vector<int>& NodeGeneList,
```
zoq: I'm wondering if it would be more efficient to update the network instead of creating a new one for each evalution. let me know what you think.

favre49: I agree, but I was not able to think of an appropriate way to do this during crossovers, since the entire network structure undergoes a change, unlike mutation where I could update depths for a single node or connection. I'll think about it some more though.

zoq: What we could do is to update the overall network once we do crossovers and update the it in the other case. But ideally we can do both without rebuilding the network, I'll think about it as well.

favre49: The thing is, mutation and crossover occur together during the reproduction step, so under the current system, we don't have to update for mutation either, since we aren't for crossover.

I can however remove the node gene list, since nodes can't be deleted, i just need to store the max node ID. I'm also working on the cyclic network evaluation right now.

zoq: You are right thanks for the clarification.

src/mlpack/methods/neat/genome_impl.hpp
```diff
@@ -0,0 +1,165 @@
...
+  {
+    Log::Fatal << "The input should have the same length as the number of"
+        "input nodes" << std::endl;
+  }
+  input = input;
+}
+
+// Evaluates output based on input.
+template <class ActivationFunction>
+arma::vec Genome<ActivationFunction>::Output()
```
zoq: Might be a good idea to pass the output parameter as input, to avoid a copy.

src/mlpack/methods/neat/genome_impl.hpp
```diff
@@ -0,0 +1,165 @@
...
+      }
+      else
+        directedGraph[i].emplace(j, ConnectionGene(nextInnovID++, 1, i, j));
+    }
+  }
+}
+
+// Loads input.
+template <class ActivationFunction>
+void Genome<ActivationFunction>::Input(const arma::vec& input)
```
zoq: I guess we could remove the `Input` function here and just pass the input once we call the `Output` function. Let me know what you think.

favre49:
I've added the implementation for cyclic networks. I've used the same scheme for activation as in [SharpNEAT](https://github.com/colgreen/sharpneat/blob/master/src/SharpNeatLib/Phenomes/NeuralNets/FastCyclicNetwork/FastCyclicNetwork.cs) for now, but I'm not sure if its correct. Could you check if it's correct or suggest a better scheme for cyclic networks?

src/mlpack/methods/neat/genome_impl.hpp
favre49: Makes sense, I'll change the name to `Evaluate` then.

src/mlpack/methods/neat/cyclic_net.hpp
```diff
@@ -0,0 +1,93 @@
...
+            const size_t nodeCount,
+            const size_t inputNodeCount,
+            const size_t outputNodeCount,
+            const size_t timeStepsPerActivation,
+            const double bias);
+
+  /**
+   * Finds the output of the network from the given input.
+   */
+  arma::vec Evaluate(arma::vec input);
```
manish7294: You should always consider passing by reference, otherwise you will be risking making an unncessary copy.

src/mlpack/methods/neat/cyclic_net_impl.hpp
```diff
@@ -0,0 +1,83 @@
...
+
+    for (size_t i = inputNodeCount; i < nodeCount; i++)
+    {
+      nodeValues[i] = actFn.Fn(inputNodeValues[i]);
+    }
+  }
+
+  arma::vec output(outputNodeCount);
+  for (size_t i = 0; i < output.n_elem; i++)
+    output[i] = putputNodeValues[i + inputNodeCount + 1];
```
manish7294: Typo!

src/mlpack/methods/neat/cyclic_net.hpp
```diff
@@ -0,0 +1,93 @@
...
+  size_t outputNodeCount;
+
+  //! The number of time steps per activation.
+  const size_t timeStepsPerActivation;
+
+  //! Bias.
+  double bias;
+
+  //! The values stored by the nodes.
+  std::vector<double> outputNodeValues;
```
manish7294: Would you like to use arma::vec instead for maintaining consistent with the codebase?

src/mlpack/methods/neat/cyclic_net_impl.hpp
```diff
@@ -0,0 +1,83 @@
...
+    outputNodeValues[i] = input[i - 1];
+
+  for (size_t i = 0; i < timeStepsPerActivation; i++)
+  {
+    for (size_t j = 0; j < nodeCount; j++)
+    {
+      for (auto const &x : directedGraph[j])
+      {
+        double weight = x.second.getWeight();
+        inputNodeValues[x.first] += nodeValues[j] * weight;
```
manish7294: I don't see any declaration of `nodeValues` throughout the cyclic net code. Are you yet to refine this impelementation as it looks kind of rough implementation?

src/mlpack/methods/neat/acyclic_net.hpp
```diff
@@ -0,0 +1,100 @@
...
+   * Finds the output of the network from the given input.
+   */
+  arma::vec Evaluate(arma::vec input);
+
+ private:
+  /**
+   * A digraph containing connection genes sorted by source ID, and then
+   * secondary sorted by target ID.
+   */
+  std::map<size_t, std::map<size_t, ConnectionGene>> directedGraph;
```
manish7294: I am wondering, do we really need to store directedGraph here, I mean we can always pass the graph with evaluate right? Let me know your thoughts.

src/mlpack/methods/neat/cyclic_net_impl.hpp
favre49: Ah yes I've fixed some of these mistakes but am yet to push them. You're right, I'm yet to refine the cyclic net implementation. I plan to do so soon.

src/mlpack/methods/neat/acyclic_net.hpp
favre49: Good idea, I missed that. I'll see where else I can do something like this as well.

src/mlpack/methods/neat/acyclic_net.hpp
```diff
@@ -0,0 +1,83 @@
...
+   */
+  AcyclicNet(const size_t nodeCount,
+             const size_t inputNodeCount,
+             const size_t outputNodeCount,
+             const double bias);
+
+  /**
+   * Finds the output of the network from the given input.
+   */
+  arma::vec Evaluate(arma::vec& input,
```
manish7294: You could actually even pass the ouput as a parameter to evaluate, instead of returning a copy. Your choice.

src/mlpack/methods/neat/cyclic_net.hpp
```diff
@@ -0,0 +1,80 @@
...
+            const size_t inputNodeCount,
+            const size_t outputNodeCount,
+            const size_t timeStepsPerActivation,
+            const double bias);
+
+  /**
+   * Finds the output of the network from the given input.
+   */
+  arma::vec Evaluate(arma::vec& input,
+                     std::map<size_t, std::map<size_t, ConnectionGene>>& directedGraph);
```
manish7294: Please wrap this line, to avoid inconsistency.

src/mlpack/methods/neat/neat_impl.hpp
```diff
@@ -0,0 +1,371 @@
...
+          weightMutationSize, biasMutationProb, biasMutationSize, 
+          nodeAdditionProb, connAdditionProb, isAcyclic);
+  }
+}
+
+template <class TaskType,
+          class ActivationFunction,
+          class SelectionPolicy>
+bool NEAT<TaskType, ActivationFunction, SelectionPolicy>::
+    compareGenome(Genome<ActivationFunction> gen1,
```
manish7294: As I mentioned earlier, always try using references, wherever applicable.

src/mlpack/methods/neat/acyclic_net.hpp
zoq: Agreed, that is also what we do most of the time in other methods.

src/mlpack/methods/neat/acyclic_net_impl.hpp
```diff
@@ -0,0 +1,92 @@
...
+        for (auto const& x : directedGraph[nodeID])
+          nodeValues[x.first] += result * x.second.getWeight();
+      }
+    }
+  }
+
+  // Find the output.
+  arma::vec output(outputNodeCount);
+  for (size_t i = 0; i < output.n_elem; i++)
+    output[i] = nodeValues[i + inputNodeCount + 1];
```
zoq: To me it looks like we could sae an extra copy here if we use ``output`` for the assignments above instead of the ``std::vector``. Let me know what you think.

src/mlpack/methods/neat/cyclic_net_impl.hpp
```diff
@@ -0,0 +1,80 @@
...
+
+    for (size_t i = inputNodeCount; i < nodeCount; i++)
+    {
+      outputNodeValues[i] = ActivationFunction::Fn(inputNodeValues[i]);
+    }
+  }
+
+  arma::vec output(outputNodeCount);
+  for (size_t i = 0; i < output.n_elem; i++)
+    output[i] = outputNodeValues[i + inputNodeCount + 1];
```
zoq: See comment above.

src/mlpack/methods/neat/genome_impl.hpp
```diff
@@ -0,0 +1,349 @@
...
+      continue;
+
+    // Mutate weight.
+    if (arma::randu<double>() < weightMutationProb)
+    {
+      connectionGeneList[i].Mutate(weightMutationSize);
+
+      // Change the weight of the gene in the directed graph as well.
+      directedGraph[connectionGeneList[i].source][connectionGeneList[i].target]
+          .setWeight() = connectionGeneList[i].getWeight();
```
zoq: To be consistent with the rest of the codebase, what about we use:

```c++
const double& Weight() const { return weight; }
double& Weight() { return weight; }
```

src/mlpack/methods/neat/selection_strategies/roulette_selection.hpp
```diff
@@ -0,0 +1,54 @@
...
+        else if (i != fitnesses.n_elem && j != fitnesses.n_elem)
+          break;
+      }
+      if (k == fitnesses.n_elem - 1)
+      {
+        k = 0;
+        prob = 0;
+      }
+    }
+    return std::make_pair(i, j);
```
zoq: Do you think we could use ``arma::vec`` here as well, might be easier to work with the result.

src/mlpack/methods/neat/selection_strategies/roulette_selection.hpp
```diff
@@ -0,0 +1,54 @@
...
+  {
+    size_t i = fitnesses.n_elem, j = fitnesses.n_elem;
+    double totalFitness = arma::accu(fitnesses);
+    double prob = 0;
+    for (size_t k = 0; k < fitnesses.n_elem; k++)
+    {
+      prob += fitnesses[k] / totalFitness;
+      if (arma::randu<double>() < prob)
+      {
+        if (i == fitnesses.n_elem)
```
zoq: Maybe I missed something, but I think this is always true.

src/mlpack/methods/neat/selection_strategies/roulette_selection.hpp
```diff
@@ -0,0 +1,54 @@
...
+ * terms of the 3-clause BSD license.  You should have received a copy of the
+ * 3-clause BSD license along with mlpack.  If not, see
+ * http://www.opensource.org/licenses/BSD-3-Clause for more information.
+ */
+#ifndef MLPACK_METHODS_NEAT_SELECTION_ROULETTE_HPP
+#define MLPACK_METHODS_NEAT_SELECTION_ROULETTE_HPP
+
+#include <mlpack/prereqs.hpp>
+
+class RouletteSelection
```
zoq: Can you write a test for each selection strategy.

src/mlpack/methods/neat/neat_impl.hpp
favre49: Ah yes, i sometimes forget. I'll make the changes, thanks!

src/mlpack/methods/neat/selection_strategies/roulette_selection.hpp
favre49: It ceases to be true the moment `i` is assigned an index for the first parent. After we do that, we can assign a parent to `j`.

src/mlpack/methods/neat/acyclic_net.hpp
favre49: Sounds good, I'll do that.

src/mlpack/methods/neat/selection_strategies/roulette_selection.hpp
favre49: So would that basically be an XOR test with all the different selection strategies?

zoq: I thought we could write some specific tests for the strategies, pass in some defined fitness values and check that the output it correct, the XOR test would be an addition.

zoq:
In the last blog update, you said something about long build times, the build step should only rebuild changes, is that the case on your system as well? Perhaps it makes sense to comment the other methods locally to decrease the build time? Also I think I said this before but you can use multiple cores to build in parallel ``make -jx`` where x is the number of threads you like to use for the build.

zoq:
@favre49 Maybe you can help me here, is the last version you pushed the version that sometimes provides reasonable results?

favre49:
@zoq Sorry for the late response, the newest commit should have no memory issues and also has connection deletion. The fitness is normally about 2.9, which is still subpar.

zoq:
Okay, have to step through the code.

zoq:
Have you locally implement a method to print a Genome? I can do this as well, just wanted to ask first.

favre49:
There is the Parameters() method for that, but I also printed out the directed graph for easier-to-read connectivity data whenever I needed it. Haven't implemented it as a method though.

zoq:
Not sure I can see the printing of the directed graph, maybe it's not pushed? Ideally I can see every connection (enabled/disabled) and the corresponding weight, maybe the input and output as well. But I can get that from the ``Evaluate`` function as well. What I'm currently trying to see if NEAT is able to find a solution just by mutating the weights, which should be possible.

src/mlpack/methods/neat/acyclic_net_impl.hpp
zoq: We should use the activation function for the output as well.

zoq:
If you don't have something like this, let me know, I can implement that function as well.

favre49:
Ah, I have not implemented such a function. I just added print statements locally of whatever I needed.

zoq:
Perhaps something like:

```
Input: 1, 1
Output: 0.93
1 -> 2 (enabled): 0.212
1 -> 3 (enabled): 0.1234
1 -> 3 (enabled): -0.123
1 -> 3 (disabled): -0.123
```

this could work? Something that is easy to read.

zoq:
The first issue I can see is that the activation function isn't used for the last layer (output). Which often results in a output that blows up.

favre49:
Makes sense, I'll make these changes.

favre49:
Alright I've added a Print() function that does what you proposed.

> I can see that the output is really similar for every input around 0.8.

I'll look at this as well

zoq:
Is the printed topology already sorted? So for example:

```
   1.0000
   1.0000
   0.8948
0->3 (enabled) Weight: 0.95126
1->3 (enabled) Weight: 1.64916
2->3 (disabled) Weight: 2.70649
2->4 (enabled) Weight: 0.654655
4->3 (enabled) Weight: 0.839223
1->4 (enabled) Weight: 1.37963
```

will look like:

IMG_0942

and the output is gene 4?

favre49:
No, the output node is always number 3 in this case. Since there is one bias node, two input nodes and one output node,  ID 0 is the bias node, ID 1 and 2 are the input nodes, and ID 3 is the output node. Anything with an ID more than that is a hidden node.

zoq:
Okay, thanks for the clarification, can you print the activation for the genes as well?

favre49:
Yup, Print() does that as well now.

src/mlpack/methods/neat/neat_impl.hpp
```diff
@@ -0,0 +1,447 @@
...
+    size_t minSize = gen1size < gen2size ? gen1size : gen2size;
+    Genome<ActivationFunction>& maxGenome = gen1size > gen2size ? gen1 : gen2;
+    Genome<ActivationFunction>& minGenome = gen1size < gen2size ? gen1 : gen2;
+    while (j < minSize)
+    {
+      size_t innovID1 = maxGenome.connectionGeneList[i].InnovationID();
+      size_t innovID2 = minGenome.connectionGeneList[j].InnovationID();
+      if (innovID2 < innovID1)
+      {
+        if (arma::randu<double>() < 0.5)
```
zoq: Shouldn't the threshold be configurable?

src/mlpack/methods/neat/selection_strategies/rank_selection.hpp
```diff
@@ -0,0 +1,64 @@
...
+   */
+  static void Select(arma::vec& fitnesses, arma::uvec& selection)
+  {
+    selection[0] = fitnesses.n_elem;
+    selection[1] = fitnesses.n_elem;
+    arma::uword pos = 0, size = fitnesses.n_elem;
+    // Choose first genome.
+    while (selection[0] == fitnesses.n_elem)
+    {
+      if (pos >= size)
```
zoq: I'm wondering if it makes performance wise to use the first index if we already iterated over. all.

src/mlpack/tests/neat_test_tools.hpp
```diff
@@ -0,0 +1,251 @@
...
+    error += findError(input1, genome); 
+    error += findError(input2, genome); 
+    error += findError(input3, genome); 
+    error += findError(input4, genome);
+    return 4 - error;
+  }
+
+  double findError(arma::vec& input, Genome<HardSigmoidFunction>& genome)
+  {
+    arma::vec output = genome.Evaluate(input);
```
zoq: To match with the existing codebase, I think it makes sense to allow the user to pass an ``arma::mat``and internally iterate over the columns, instead of doing this outside the method. In this case we could pass the complete input data and use ``arma::accu(arma::pow(...))`` to get the error.

src/mlpack/methods/neat/gene_impl.hpp
```diff
@@ -48,6 +48,17 @@ void ConnectionGene::Mutate(const double mutationSize)
   weight += mutationSize * arma::randn<double>();
 }
 
+template <typename Archive>
+void ConnectionGene::serialize(Archive& ar,
+                                const unsigned int /* version */)
```
zoq: Pedantic style issue number one: the second parameter should align with the first one.

src/mlpack/methods/neat/gene_impl.hpp
```diff
@@ -48,6 +48,17 @@ void ConnectionGene::Mutate(const double mutationSize)
...
+{
+  ar & BOOST_SERIALIZATION_NVP(globalInnovationID);
+  ar & BOOST_SERIALIZATION_NVP(weight);
```
zoq: Should we save the weights only if we initialized the model?

src/mlpack/tests/neat_test_tools.hpp
```diff
@@ -197,79 +213,80 @@ class DPNVTask
...
+    {
+      double sum = 0;
+      for (auto it = wiggleBuffer2.begin(); it != wiggleBuffer2.end(); it++)
+        sum += *it;
+      fitness += timeStep + (10.0 / std::max(1.0, sum));
+    }
+    else
+      fitness += timeStep;
+
+    std::cout <<" Fitness: " << fitness << std::endl;
```
zoq: In case we are going to provide such a function in the final version, we should go for ``Log::Debug``instead of ``std::cout``.

zoq:
Have you removed the Print functionality?

favre49:
Oops must have removed it at some point. I'll add it back along with phased searching later tonight

zoq:
Okay, great, thanks.

favre49:
I have added phased searching. NEAT now performs phased search if `complexityThreshold != 0`. I've done very basic testing, and it seems to be performing fine, except it might be unable to simplify further after a certain number of generations, I will have to investigate this further. If this is the case I might apply a relative complexity ceiling instead of an absolute one.

I also added the Print() function. On the non-Markovian double pole velocity experiment, I have only seen NEAT reach a fitness of about 450, though it should be able to get a fitness above a couple thousand.

zoq:
> I have added phased searching. NEAT now performs phased search if `complexityThreshold != 0`. I've done very basic testing, and it seems to be performing fine, except it might be unable to simplify further after a certain number of generations, I will have to investigate this further. If this is the case I might apply a relative complexity ceiling instead of an absolute one. I'll also implement a test more suited to this, like a many-input multiplexer or the OCR test talked about [here](https://sharpneat.sourceforge.io/phasedsearch.html).

Sounds great, will take a closer look at the code.

> I also added the Print() function. On the non-Markovian double pole velocity experiment, I have only seen NEAT reach a fitness of about 450, though it should be able to get a fitness above a couple thousand.

Great, thanks, should be helpful to track down the issue we see with the double pole velocity env.

src/mlpack/methods/neat/neat_impl.hpp
```diff
@@ -0,0 +1,540 @@
...
+    if (gen == maxGen - 1)
+      break;
+
+    // Set search mode.
+    if (meanComplexity > (double)complexityThreshold)
+      searchMode = 1;
+    else
+      searchMode = 0;
+
+    std::cout << "Maximum fitness " << fitnesses.max() << std::endl;
```
manish7294: Would be better to have them as Log::Info messages.

src/mlpack/methods/neat/neat_impl.hpp
```diff
@@ -0,0 +1,540 @@
...
+  int delta = (int)popSize - (int)arma::accu(speciesSize);
+  if (delta > 0)
+  {
+    size_t i = 0;
+    while (delta > 0)
+    {
+      speciesSize[i++]++;
+      delta--;
+      if(i >= numSpecies)
+        i = 0;
```
manish7294: I guess you could do this a bit faster by mathematically calculating the number of complete repetion and the remaining offset. Though your choice. Same apply for the else if part.

src/mlpack/methods/neat/neat_impl.hpp
```diff
@@ -0,0 +1,540 @@
...
+  }
+
+  // Find the number of elite members in each species.
+  for (size_t i = 0; i < numSpecies; i++)
+  {
+    numElite[i] = std::round(elitismProp * speciesSize[i]);
+    // This ensures that the best genome from every species will be saved.
+    if (numElite[i] == 0)
+      numElite[i] = 1;
+    if (numElite[i] > speciesList[i].size())
```
manish7294: please put this in else if block, otherwise it would be an unccessary check.

src/mlpack/tests/neat_test_tools.hpp
```diff
@@ -0,0 +1,296 @@
...
+class DiscreteRLTask
+{
+ public:
+  DiscreteRLTask(const EnvironmentType& env) : environment(env)
+  { /* Nothing to do here */ }
+
+  double Evaluate(Genome<HardSigmoidFunction>& genome)
+  {
+    // Set the initial state.
+    typename EnvironmentType::State state = environment.InitialSample();
```
zoq: We want to evaluate if a model is all to solve a task using different start conditions, so this should move down into the for loop, so that we start with a new random position instead of using the previous one.

src/mlpack/tests/neat_test_tools.hpp
```diff
@@ -0,0 +1,296 @@
...
+        // Use the current action to get the next state.
+        fitness += environment.Sample(state, action, state);
+
+        // Update the state of the genome for the next step.
+        output = genome.Evaluate(state.Data());
+
+        if (fitness > 10000)
+          break;
+      }
+      total += fitness;
```
zoq: We never reset the ``fitness`` value, so this is already the total fitness, I think we should reset the ``fitness`` parameter before each run, so that ``if (fitness > 10000)`` does work as well.

zoq: Ahh, nevermind I can see you do reset the value.

src/mlpack/methods/neat/neat_impl.hpp
```diff
@@ -0,0 +1,601 @@
...
+{
+  // Translate the genome into points in space.
+  arma::mat data(Genome<ActivationFunction>::nextInnovID, popSize,
+      arma::fill::zeros);
+  for (size_t i = 0; i < popSize; i++)
+  {
+    for (size_t j = 0; j < genomeList[i].connectionGeneList.size(); j++)
+    {
+      size_t innovID = genomeList[i].connectionGeneList[j].InnovationID();
+      data(innovID, i) = genomeList[i].connectionGeneList[j].Weight();
```
zoq: I think we should only account for enabled connections.

src/mlpack/tests/neat_test.cpp
```diff
@@ -44,27 +44,27 @@ BOOST_AUTO_TEST_SUITE(NEATTest)
 /**
  * Test NEAT on the XOR Test.
  */
-BOOST_AUTO_TEST_CASE(NEATXORTest)
-{
-  arma::arma_rng::set_seed_random();
-  XORTask task;
-  NEAT<XORTask, HardSigmoidFunction, RankSelection> model(task, 2, 1, 100, 150, 10);
-  model.FinalFitness() = 3.8;
+// BOOST_AUTO_TEST_CASE(NEATXORTest)
```
zoq: I think as a quick example we should leave the XOR test as well.

favre49: Yup I'll leave it in, but do you think I should also write a short guide?

src/mlpack/tests/neat_test_tools.hpp
```diff
@@ -0,0 +1,155 @@
...
+};
+
+/**
+ * A class that defines the non-Markovian double pole balancing task. The
+ * observation space is comprised of the cart position and the angles of the poles.
+ */
+class DPNVTask
+{
+ public:
+  DPNVTask(const MultiplePoleCart& env) : environment(env)
```
zoq: I'm wondering if the implementation of the task is wrong, I can see there is a test for the env: https://github.com/mlpack/mlpack/blob/c62088ad40c4f9a3839e23a5dbbd5ab55054c7ee/src/mlpack/tests/rl_components_test.cpp#L171 but we haven't really tested if e.g. QLearning is able to solve the env, maybe you tested it?

favre49: I think the learners we have right now aren't compatible with how the multiple pole cart class is structured. For instance, we use a matrix for the state data, and don't have a default constructor, for state and the environment. Maybe we should rewrite this class to allow for compatibility?

I've checked the equations again, and they are right according to the paper. I checked other implementations, and they are using RK4 along with these equations, which I don't know much about. I'll study about that and test that as well.

favre49: We could work around the issue of not having a default constructor for the state by making the multiple pole cart into an only double pole cart method. That way the API will be intact without having an odd exception environment. Moreover, I guess I haven't seen a triple or quadruple pole balancing test before (I made it general since I thought I could). Unless there's a better workaround?

zoq: > We could work around the issue of not having a default constructor for the state by making the multiple pole cart into an only double pole cart method. That way the API will be intact without having an odd exception environment. 

That sounds like a reasonable solution to me.

> Moreover, I guess I haven't seen a triple or quadruple pole balancing test before (I made it general since I thought I could). Unless there's a better workaround?

Yeah, I really liked that.

> I was also thinking, unless it would make running the tests too lengthy, we should add unit tests for all these environments, where we check if one of our learners could solve them, so that we don't run into this problem or the one we had with the pendulum environment. It would make sure the API remains consistent. Let me know what you think.

Had the same idea, we could provide some initial parameter that are almost good enough, that way we could reduce the runtime for the tests.

favre49: I'll implement it locally for now, if NEAT works on it I'll open a PR to fix these issues.

zoq: Sounds good, thanks!

zoq:
Id did some further testing, and it looks like there are some env's that try to maximise the reward and some to minimise the reward, an easy solution would be to use the absolute over the reward; Let me know what you think.

favre49:
Yup, that makes sense to me.

zoq:
> Yup, that makes sense to me.

Should probably be an option, e.g. ``bool absoluteReward``, the default value (``true``?), just in case a user doesn't want to take the absolute value.

favre49:
Actually, I think we should expect the user to ensure that fitness is mostly non-negative. Negative fitnesses don't really work with how species allocation is done, as it could make the species size negative (I handle this in the code by making negative species sizes zero, but it may not be a good idea to have fitnesses to be mostly negative). This seems like something for the user to take care of to me, what do you think?

zoq:
I see your point, so I guess it makes sense to check the output and to print out an error, if the fitness is negative. I tested a bunch of gym env's and some of them return a negative reward, so I had to make sure the reward is reasonable, in some cases taking the absolute reward works, but in some cases it doesn't. So, I agree might be a good idea to let the user handle it.

favre49:
I fixed phased searching, and instead of giving it it's own test, I included it in the XOR task, and checked if the population's mean complexity was at most a unit above the current complexity ceiling as specified by phased searching . This is because it is possible that NEAT is in the process of simplification, in which case the complexity may be in excess of the ceiling (but it can't be too far above). Let me know what you think, if I should change the test or maybe split it into two, one for standalone XOR and one for Phased Searching

zoq:
Personally I would split the test in two, but looking at the current test time:  ``55/148 Test  #66: NEATTest .........................   Passed   24.56 sec`` I'm not sure if the current solution is preferable, we still have some other tests that aren't build around phased searching. Do you have the timings for the XOR test without and with phased searching? (To get a quick result I often use ```time bin/mlpack_test -t NEATTest/NEATXORTest``).

src/mlpack/methods/neat/neat_impl.hpp
```diff
@@ -0,0 +1,669 @@
...
+    // Check termination criteria.
+    if (finalFitness != 0 && fitnesses.max() >= finalFitness)
+      break;
+    else if (gen == maxGen - 1)
+      break;
+
+    // Set search mode.
+    if (meanComplexity > currentComplexityCeiling)
+    {
+      if (searchMode == 0)
```
zoq: I'm wondering if we should implement the search mode as a policy, so that we could add more methods in the future, without touching the NEAT implementation. Let me know what you think.

doc/tutorials/neat/neat.txt
```diff
@@ -0,0 +1,248 @@
...
+
+The genome is also what is used to evaluate an input based on the network it represents. This is done by using the genome's Evaluate() function:
+
+@code
+#include <mlpack/methods/neat/neat.hpp>
+
+arma::vec input = arma::randn<arma::vec>(inputNodeCount);
+
+// Assuming a Genome object called genome has already been instantiated.
+arma::vec output = genome.Evaluate(input);
```
zoq: Do you think we should instantiate a Genome, that way a use could just copy past the lines from the example.

favre49:
I checked, and phased searching has pretty much no effect on performance (probably makes it worse). This is to be expected though, since phased searching should only show it's usefulness in cases where there is the issue of genome bloat, not in easy tests like XOR. I only used this test to make sure phased test controlled complexity correctly.

src/mlpack/methods/neat/neat_impl.hpp
favre49: I had though of this, but I couldn't come up with a general policy implementation since I hadn't heard of any other search types. Moreover in this case it's achievable with just changing probabilities, but in other cases a search mode could require a different kind of mutation, new rules and measures (I'm just theorizing here, but phased searching involved keeping track of complexity metrics and adding a method to measure complexity) and we would have to change the NEAT implementation anyway. Is there a way I could come up with some kind of general policy representation?

zoq:
Okay, thanks for the clarification, I think the current test situation is fine.

src/mlpack/methods/neat/genome.hpp
```diff
@@ -0,0 +1,319 @@
...
+ * 
+ * The genome can undergo three types of structural mutations:
+ * 1. A connection can be split into two, creating a new node.
+ * 2. A new connection can be created between two nodes.
+ * 3. A connection can be deleted.
+ * 
+ * @tparam ActivationFunction The activation function to be used.
+ */
+template <class ActivationFunction = ann::HardSigmoidFunction>
+class Genome
```
zoq: Since we use ``Genome<>`` a lot I think it would make sense from the style perspective, to use ``Genome`` instead, so we could write:

```
template <class ActivationFunction = ann::HardSigmoidFunction>
class GenomeType
{
};

template <class ActivationFunction = HardSigmoidFunction>
using Genome = GenomeType< ActivationFunction>;
```

Let me know what you think.

src/mlpack/methods/neat/neat.hpp
```diff
@@ -0,0 +1,370 @@
...
+       const double elitismProp = 0.1,
+       const double finalFitness = 0,
+       const size_t complexityThreshold = 0,
+       const size_t maxSimplifyGen = 10,
+       const bool isAcyclic = false);
+
+  /**
+   * Trains the model on the task and returns the best Genome.
+   */
+  Genome<ActivationFunction> Train();
```
zoq: To be consistent with the rest of the codebase I think it would make sense to pass a reference of the genome.

src/mlpack/tests/neat_test_tools.hpp
```diff
@@ -0,0 +1,149 @@
+/**
+ * @file neat_test_tools.hpp
```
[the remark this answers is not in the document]
favre49: I don't have an issue with that. The only con would be that separating them is more aesthetically pleasing, and easier to understand. I'll go with your judgement on this.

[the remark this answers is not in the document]
zoq: Personally I find it easier to have them in the same file, as I don't have to open another file to search for the definition.

src/mlpack/tests/neat_test.cpp
```diff
@@ -0,0 +1,77 @@
...
+
+BOOST_AUTO_TEST_SUITE(NEATTest)
+
+/**
+ * Test NEAT and phased searching on the XOR Test. XOR normally performs better
+ * in cyclic cases, this is meant to be a test of the acyclic network.
+ */
+BOOST_AUTO_TEST_CASE(NEATXORTest)
+{
+  arma::arma_rng::set_seed_random();
```
zoq: Can you remove the random seed here, for debugging it would be great to have the same random seed for each run.

src/mlpack/methods/neat/selection_strategies/tournament_selection.hpp
```diff
@@ -0,0 +1,61 @@
...
+class TournamentSelection
+{
+ public:
+  /**
+   * The method that selects two parents out of the population. It returns a
+   * pair of indices of the parents.
+   * 
+   * @param fitnesses A sorted Armadillo vector of fitnesses in descending
+   *    order.
+   * @param contenderNum The number of contenders.
```
zoq: This is missing the ``selection`` vector.

src/mlpack/methods/neat/selection_strategies/tournament_selection.hpp
```diff
@@ -0,0 +1,61 @@
...
+   * @param prob The probability of the fittest candidate being chosen.
+   */
+  static void Select(arma::vec& fitnesses,
```
zoq: Looks like the ``fitnesses`` parameter could be ``const``.

src/mlpack/methods/neat/selection_strategies/roulette_selection.hpp
```diff
@@ -0,0 +1,53 @@
...
+ public:
+  /**
+   * The method that selects parents out of the population. It returns a
+   * pair of indices of the parents.
+   * 
+   * @param fitnesses A sorted Armadillo vector of fitnesses in descending
+   *    order.
+   * @param selection The selected indices.
+   */
+  static void Select(arma::vec& fitnesses, arma::uvec& selection)
```
zoq: See comment below about ``const``.

src/mlpack/methods/neat/selection_strategies/roulette_selection.hpp
```diff
@@ -0,0 +1,53 @@
...
+        if (randNum < prob)
+        {
+          selection[i] = k;
+          prob = 0;
+          break;
+        }
+        k--;
+      }
+      if (k == 0)
+        selection[i] = k;
```
zoq: I guess what could do is to set ``selection[i] = 0`` before the while loop, to avoid the extra check at the end.

zoq:

src/mlpack/methods/neat/acyclic_net_impl.hpp
```diff
@@ -0,0 +1,100 @@
...
+    inputNodeCount(inputNodeCount),
+    outputNodeCount(outputNodeCount),
+    bias(bias)
+{ /* Nothing to do here */ }
+
+// Evaluate a given input.
+template <class ActivationFunction>
+void AcyclicNet<ActivationFunction>::Evaluate(const arma::vec& input,
+                                              arma::vec& output,
+                                              std::map<size_t, std::map<size_t,
```
zoq: Looks like this could be const.

src/mlpack/methods/neat/acyclic_net_impl.hpp
```diff
@@ -0,0 +1,100 @@
...
+                                                ConnectionGene>>& directedGraph,
+                                              std::vector<size_t>& nodeDepths)
```
zoq: Looks like this could be const as well.

src/mlpack/methods/neat/acyclic_net_impl.hpp
```diff
@@ -0,0 +1,100 @@
...
+{
+  arma::vec nodeValues(nodeCount, arma::fill::zeros);
+
+  // Populate the layers.
+  for (size_t i = 0; i < nodeCount; i++)
+  {
+    while (layers.size() < nodeDepths[i] + 1)
```
zoq: Should be faster to call ``reserve`` once with the ``layers.size() - nodeDepths[i] + 1``.

src/mlpack/methods/neat/acyclic_net_impl.hpp
```diff
@@ -0,0 +1,100 @@
...
+  for (size_t i = 1; i <= inputNodeCount; i++)
+    nodeValues[i] = input[i - 1];
+
+  // Activate the layers one by one.
+  for (size_t i = 0; i < layers.size(); i++)
+  {
+    for (size_t j = 0; j < layers[i].size(); j++)
+    {
+      size_t nodeID = layers[i][j];
+      if (nodeID == 0)
```
zoq: Can you comment on the conditions here ``if (nodeID == 0)``, ``else if (nodeID <= inputNodeCount)`` and ``else``, I think it might help to understand the code.

src/mlpack/methods/neat/acyclic_net.hpp
```diff
@@ -0,0 +1,84 @@
...
+#include <mlpack/prereqs.hpp>
+
+namespace mlpack{
+namespace neat /** NeuroEvolution of Augmenting Topologies */{
+/**
+ * A class representation of an acyclic neural network. Genomes are decoded
+ * into this class if the `isAcyclic` parameter is set to `true`. It is used
+ * to evaluate an input to the genome.
+ * 
+ * The steps of the decoding and activating are as follows:
```
zoq: This is really helpful, thanks!

src/mlpack/methods/neat/acyclic_net_impl.hpp
favre49: Yup, I'll go through the codebase and add const wherever I can

src/mlpack/methods/neat/acyclic_net_impl.hpp
```diff
@@ -0,0 +1,106 @@
...
+      // If this is a bias node, we need to add bias to it's connections.
+      if (nodeID == 0)
+      {
+        for (auto const& x : directedGraph[nodeID])
+        {
+          if (x.second.Enabled())
+            nodeValues[x.first] += bias * x.second.Weight();
+        }
+      }
+      // If it is an input node, we need not apply the activation function.
```
zoq: Do you think we should write "we don't need to apply the activation function"?

src/mlpack/tests/neat_test.cpp
```diff
@@ -0,0 +1,210 @@
...
+ * | {0, 1}  |    1     |
+ * | {1, 0}  |    1     |
+ * | {1, 1}  |    0     |
+ *
+ */
+class XORTask
+{
+ public:
+  template<typename GenomeType>
+  double Evaluate(GenomeType& genome)
```
zoq: I think this can be const as well, since we don't change the state of the given genome.

src/mlpack/methods/neat/acyclic_net_impl.hpp
```diff
@@ -0,0 +1,106 @@
...
+  // Activate the layers one by one.
+  for (size_t i = 0; i < layers.size(); i++)
+  {
+    for (size_t j = 0; j < layers[i].size(); j++)
+    {
+      size_t nodeID = layers[i][j];
+      // If this is a bias node, we need to add bias to it's connections.
+      if (nodeID == 0)
+      {
+        for (auto const& x : directedGraph[nodeID])
```
zoq: We might have to change this to ``const auto const& x : directedGraph[nodeID]`` or I guess we could also use a standard for loop and use the index.

favre49: This would give an error, since there would be two const qualifiers

src/mlpack/tests/neat_test.cpp
favre49: Phased searching and cyclic evaluation both change Genome state

src/mlpack/methods/neat/acyclic_net_impl.hpp
zoq: Ohh, right I missed to remove the second const.

src/mlpack/tests/neat_test.cpp
zoq: So if I call ``Evaluate`` on a genome twice, same env settings, I could get different results in the testing phase?

zoq: Let me know if I should clarify this point.

src/mlpack/methods/neat/genome_impl.hpp
```diff
@@ -0,0 +1,523 @@
...
+    }
+  }
+}
+
+// Delete connection.
+template <class ActivationFunction>
+void Genome<ActivationFunction>::DelConnMutation()
+{
+  size_t i = 0;
+  i = arma::randi<arma::uvec>(1, arma::distr_param(0,
```
zoq: You could use:

https://github.com/mlpack/mlpack/blob/12a50a055ba7f69340598329bd146ee37bec110f/src/mlpack/core/math/random.hpp#L106

src/mlpack/methods/neat/genome_impl.hpp
```diff
@@ -0,0 +1,523 @@
...
+
+// Finds complexity of the genome.
+template <class ActivationFunction>
+size_t Genome<ActivationFunction>::Complexity()
+{
+  size_t connCount = 0;
+  for (size_t i = 0 ; i < connectionGeneList.size(); i++)
+  {
+    if (connectionGeneList[i].Enabled())
+      connCount++;
```
zoq: I'm wondering if it makes sense to build the connectivity at the same time as we add/disable the connection. Let me know what you think.

src/mlpack/tests/neat_test.cpp
favre49: Oops sorry, I didn't get a notification for the comment.

Yes, since cyclic networks store their activation values. I could maybe add a reset function which would clear the old activation?

zoq: So it's just used to store temporary results; I don't think this is a problem, the only issue I see is that from a user perspective this could mean that we actually change the object. But I think, it would have a negative effect on the runtime if we make a copy of the genome first. So unless you can see a simple solution, I think we should go with the non-const solution.

src/mlpack/methods/neat/genome_impl.hpp
favre49: Do you mean update the nose depths? I tried this, but I couldn't think of a different way to do it.

zoq: I was thinking we could update the connection count parameter once the genome is created, so if we can a new enabled connection we could increase the counter, if we disable one we could decrease the counter. That way we now at each time what the connection count is. Do you think that is feasible?

favre49: sounds good to me, I'll make that change

zoq: Sounds good I think this could take some time.

zoq:
Let me restart the travis build, maybe this was just a temporary network higup.

zoq:
Do you think we could use the current model to solve a more complicated env something like Mario, I could set it up?

favre49:
Sounds good, that would be a good showcase, if it works out.

zoq:
Okay, will set this up and let you know once there is something to show. This will has to run for a couple of iterations, I tested Q-Learning on Mario as well and if I remember right this took like 200k iterations.

src/mlpack/methods/neat/neat_impl.hpp
```diff
@@ -0,0 +1,669 @@
...
+    HasStartingGenome<Task, std::vector<ConnectionGene>(Task::*)()>::value,
+    void>::type
+NEAT<TaskType, ActivationFunction, SelectionPolicy>::Initialize()
+{
+  std::vector<ConnectionGene> connGeneList = task.startingGenome();
+  size_t maxInnovID = 0, maxNodeID = 0;
+  for (size_t i = 0; i < connGeneList.size(); i++)
+  {
+    if (connGeneList[i].InnovationID() > maxInnovID)
+      maxInnovID = connGeneList[i].InnovationID;
```
zoq: This should be ``InnovationID()``.

src/mlpack/methods/neat/selection_strategies/tournament_selection.hpp
```diff
@@ -0,0 +1,60 @@
...
+                     const size_t contenderNum,
+                     const double prob)
+  {
+    selection.fill(0);
+
+    for (size_t i = 0; i < selection.n_elem; i++)
+    {
+      arma::uvec contenders = arma::randi<arma::uvec>(contenderNum,
+          arma::distr_param(0, fitnesses.n_elem - 1));
+      arma::sort(contenders);
```
zoq: ``arma::sort`` is not inplace.

zoq:
Looks like if I wanted to use https://github.com/Kautenja/gym-super-mario-bros I first have to implement titles support if I don't want to run NEAT with 184320 input nodes.

zoq:
Yes, let's keep this open, will update this with the numbers for the Mario env, soon.

abinezer:
hi mlpack! I am currently exploring neat genetic algos. Anything i could do to help a bit here?

favre49:
Hey! This implementation is done, we wanted to test on super mario bros but got busy with other things. I'm hoping I get to merge this during the summer, if free. If testing the implementation is of interest to you, please feel free to do so :)

rcurtin:
This code touches the ANN code, and there was recently a big refactoring of the ANN code in #2259, so be sure to merge the master branch into your branch here to make sure that nothing will fail if this PR is merged. :+1: (I'm pasting this message into all possibly relevant PRs. I *think* this one is relevant?)

PR #1912, opened 2019-05-30
[WIP] Proximal Policy Optimization

robotcator:
Implementing the [proximal policy optimization](https://arxiv.org/pdf/1707.06347.pdf) algorithm in RL. Adding the basic skeleton.


src/mlpack/methods/ann/loss_functions/surrogate_loss.hpp
```diff
@@ -0,0 +1,62 @@
+/**
+ * @file surrogate_loss.hpp
+ * @author Xiaohong Ji
+ *
```
zoq: This is missing a simple description of the file.

robotcator: > This is missing a simple description of the file.

Yes, I will add a description for the added file.

robotcator:
@zoq Hi, I have a question about how to reuse the template parameter. But I am not very sure the problem is the same as reusing the ReplayType in the PER PR.


I define the PPO class like this, there are actor and critic model. 

```
template<
  typename EnvironmentType,
  typename NetworkType,
  typename UpdaterType,
  typename PolicyType,
  typename ReplayType
>
PPO<
  EnvironmentType,
  NetworkType,
  UpdaterType,
  PolicyType,
  ReplayType
>::PPO(TrainingConfig config,
       NetworkType actor,
       NetworkType critic,
       PolicyType policy,
       ReplayType replayMethod,
       UpdaterType updater,
       EnvironmentType environment) {
}
```

When I create a PPO object, I declare like this:
```
  PPO<Pendulum, decltype(actor), AdamUpdate, decltype(policy)>
      agent(std::move(config), std::move(actor), std::move(critic),
          std::move(policy), std::move(replayMethod));
```
But for the actor and critic model declare like this respectively:
```
FFN<MeanSquaredError<>, GaussianInitialization> critic(MeanSquaredError<>(),
...
  FFN<SurrogateLoss<>, GaussianInitialization> actor(SurrogateLoss<>(0.2),
```
It will get error like this:
```
error: no
      matching constructor for initialization of 'PPO<mlpack::rl::Pendulum, decltype(actor),
      ens::AdamUpdate, decltype(policy)>' (aka 'PPO<mlpack::rl::Pendulum, FFN<SurrogateLoss<>,
      mlpack::ann::GaussianInitialization>, ens::AdamUpdate, GreedyPolicy<mlpack::rl::Pendulum>
      >')
      agent(std::move(config), std::move(actor), std::move(critic),
      ^     ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```
It seems that I am not so famaliar with the template programming or missing something. Hope you can give me some suggestions.

zoq:
In the example above actor and critic are of different types, but the code expects them to be of the same type:

```c++
NetworkType actor,
NetworkType critic,
```

so I think the easiest solution here would be to introduce a new template parameter one type for the actor network and another one for the critic network:

```c++
template<
  typename EnvironmentType,
  typename ActorNetworkType,
  typename CriticNetworkType,
  typename UpdaterType,
  typename PolicyType,
  typename ReplayType
>
PPO<
  EnvironmentType,
  NetworkType,
  UpdaterType,
  PolicyType,
  ReplayType
>::PPO(TrainingConfig config,
       ActorNetworkType actor,
       CriticNetworkType critic,
       PolicyType policy,
       ReplayType replayMethod,
       UpdaterType updater,
       EnvironmentType environment) {
}
```

Let me know if I should clarify anything further.

manish7294:
Agreed, this looks like a clean solution to the problem.

robotcator:
> In the example above actor and critic are of different types, but the code expects them to be of the same type:
> 
> ```c++
> NetworkType actor,
> NetworkType critic,
> ```
> 
> so I think the easiest solution here would be to introduce a new template parameter one type for the actor network and another one for the critic network:
> 
> ```c++
> template<
>   typename EnvironmentType,
>   typename ActorNetworkType,
>   typename CriticNetworkType,
>   typename UpdaterType,
>   typename PolicyType,
>   typename ReplayType
> >
> PPO<
>   EnvironmentType,
>   NetworkType,
>   UpdaterType,
>   PolicyType,
>   ReplayType
> >::PPO(TrainingConfig config,
>        ActorNetworkType actor,
>        CriticNetworkType critic,
>        PolicyType policy,
>        ReplayType replayMethod,
>        UpdaterType updater,
>        EnvironmentType environment) {
> }
> ```
> 
> Let me know if I should clarify anything further.

Thanks, I got it. I am not sure whether I can create a new template parameter, so I make sure here. It seems that all of us agree that create a new parameter.

robotcator:
It seems that I need to rewrite the policy for the continuous task.

robotcator:
The code is a little messy now because I am rusty at the continous task.  

I have a question about that when the model's output is the two dimensions' `colvec`, one for `mean` and the other one for `variance`, but the GaussianDistribution class need to pass `arma` as parameters. The first idea I am going to convert the `colvec` to `arma`, is there any better idea? 
I am not sure I clarify what I said.  If anything else needs to clarify please tell me.

manish7294:
Sorry, can you explain a bit more, what do you mean by `colvec` to `arma` here, isn't `colvec` a dataype of `arma` namespace? Or do you mean something like conversion of column vectors to matrix, or something like that?

src/mlpack/methods/reinforcement_learning/ppo.hpp
```diff
@@ -0,0 +1,154 @@
...
+  //! Convenient typedef for state.
+  using StateType = typename EnvironmentType::State;
+
+  //! Convenient typedef for action.
+  using ActionType = typename EnvironmentType::Action;
+
+  /**
+   * Create the PPO object with given settings.
+   *
+   * If you want to pass in a parameter and discard the original parameter
```
manish7294: I don't think it's needed to be specified here.  Though you can keep it, if you want to.

 Moreover, I would insist on giving a more explanatory description of PPO class above.

robotcator:
> Sorry, can you explain a bit more, what do you mean by `colvec` to `arma` here, isn't `colvec` a dataype of `arma` namespace? Or do you mean something like conversion of column vectors to matrix, or something like that?

Yes, in [line101](https://github.com/mlpack/mlpack/blob/6bed414710b332549e4a44915d6dc4e7cc52917e/src/mlpack/methods/reinforcement_learning/ppo_impl.hpp#L101), the model outputs two data, one for mean and another one for the variance. 

If I the first element `actionValue[0]` is used for mean and the second element `actionValue[1]` is used for variance. The `actionValue[0]` is a `double` type while the `GaussianDistribution` constructor must passed in the `arma matrix` type. So I want to convert the first and second element to the `arma matrix` type.

src/mlpack/methods/reinforcement_learning/ppo_impl.hpp
```diff
@@ -0,0 +1,189 @@
...
+  ActorNetworkType,
+  CriticNetworkType,
+  UpdaterType,
+  PolicyType,
+  ReplayType
+>::Step()
+{
+  // Get the action value for each action at current state.
+  arma::colvec actionValue;
+  ActorNetwork.Predict(state.Encode(), actionValue);
```
manish7294: Here actionValue can't be a `colvec`, `ActorNetwork` expects `arma::mat` type to be pass as output, see https://github.com/mlpack/mlpack/blob/6bed414710b332549e4a44915d6dc4e7cc52917e/src/mlpack/methods/ann/ffn.hpp#L138

So, in this case `actionValue.col(0)` will probably be used for mean and rest of the rows will be used for variance. Hence, you can easily extract mean as a `vec` from `actionValue` and variance as a `mat` by deleting first row of `actionValue`.
Does this clear things up?

robotcator: > Here actionValue can't be a `colvec`, `ActorNetwork` expects `arma::mat` type to be pass as output, see
> 
> https://github.com/mlpack/mlpack/blob/6bed414710b332549e4a44915d6dc4e7cc52917e/src/mlpack/methods/ann/ffn.hpp#L138
> 
> So, in this case `actionValue.col(0)` will probably be used for mean and rest of the rows will be used for variance. Hence, you can easily extract mean as a `vec` from `actionValue` and variance as a `mat` by deleting first row of `actionValue`.
> Does this clear things up?

Okay, why I use `colvec` here because the code in [`q_learning.hpp` ](https://github.com/mlpack/mlpack/blob/master/src/mlpack/methods/reinforcement_learning/q_learning_impl.hpp#L99), so I am not sure. In this situation, `mat` is more suitable.

manish7294: Ohh you are right. Basically the thing is, the functions that expects `mat` as an input can generally also accept `col`. So, here you can just grab values as `colvec` by the logic I mentioned earlier and should be able to pass it to the GaussianDistribution constructor, even though it expects a `mat` type to passed.

zoq: If they don't accept the input, you can also checkout http://arma.sourceforge.net/docs.html#conv_to

robotcator: Hi, I have confused with the `EnvironmentType::Action`, why it can convert to `int` to the `EnvironmentType::Action` when use`CartPole`' in [line](https://github.com/mlpack/mlpack/blob/master/src/mlpack/methods/reinforcement_learning/policy/greedy_policy.hpp#L70).

But when I use the code  to sample action for `Pendulum` task like this:
```
ActionType action = normalDist.Random();
```
the complier output like that:
```
: error: 
      no matching conversion for static_cast from 'arma::vec' (aka 'Col<double>') to 'ActionType' (aka
      'mlpack::rl::Pendulum::Action')
  ActionType action = static_cast<ActionType>(normalDist.Random());
```

zoq: The ``Pendulum`` env does use an array as action, https://github.com/mlpack/mlpack/blob/e2f696cfd5b7ccda2d3af1c7c728483ea6591718/src/mlpack/methods/reinforcement_learning/environment/pendulum.hpp#L85

you can remove the ``[1]`` here and just use action everywhere else to solve the issue. But we should open a PR to fix this in the codebase.

zoq: Hopefully that makes sense, let me know if I should clarify anything.

robotcator: > Hopefully that makes sense, let me know if I should clarify anything.

Sorry for late replay, I still struggle for this part. I remove the `[1]` everywhere, and remain the sample action code.
```
ActionType action = normalDist.Random();
```

It still the same error. 

```
: error: 
      no viable conversion from 'arma::vec' (aka 'Col<double>') to 'ActionType' (aka
      'mlpack::rl::Pendulum::Action')
  ActionType action = normalDist.Random();
             ^        ~~~~~~~~~~~~~~~~~~~
```

I am trying to figure out what's happening.

robotcator: > > Hopefully that makes sense, let me know if I should clarify anything.
> 
> Sorry for late replay, I still struggle for this part. I remove the `[1]` everywhere, and remain the sample action code.
> 
> ```
> ActionType action = normalDist.Random();
> ```
> 
> It still the same error.
> 
> ```
> : error: 
>       no viable conversion from 'arma::vec' (aka 'Col<double>') to 'ActionType' (aka
>       'mlpack::rl::Pendulum::Action')
>   ActionType action = normalDist.Random();
>              ^        ~~~~~~~~~~~~~~~~~~~
> ```
> 
> I am trying to figure out what's happening.

I rewrite the code in this way, it works but not so elegant and extensible.  

```
  ActionType action;
  action.action = normalDist.Random()[0];
```

robotcator: > your best bet

 @manish7294, thank you for your explanation.

robotcator:
Can not pass the CI test right now, need to merge the [pr](https://github.com/mlpack/mlpack/pull/1931).

robotcator:
@zoq I have a question. I find that the action in the `random_replay.hpp`  is [`arma::icolvec`](https://github.com/mlpack/mlpack/blob/master/src/mlpack/methods/reinforcement_learning/replay/random_replay.hpp#L175), it is suitable when the action is discrete. But for the continuous task,  sometimes the action is not `arma::icolvec` type.

zoq:
You are right, it was implemented with a discrete action space in mind, but I think that will change, so if you like to change that one, please feel free.

robotcator:
> You are right, it was implemented with a discrete action space in mind, but I think that will change, so if you like to change that one, please feel free.

I choose the `std::vector` to store the action for the agent. Currently, it is okay for the replay buffer (the corresponding code in the prioritized replay buffer is also need to be changed). I am trying to test whether the `arma::Mat` can serve the same function. 

Also, we need to add the copy constructor for the pendulum's action, Shall we open a new PR?

manish7294:
> I choose the std::vector to store the action for the agent. Currently, it is okay for the replay buffer (the corresponding code in the prioritized replay buffer is also need to be changed). I am trying to test whether the arma::Mat can serve the same function.

std::vector will do the work here. You could also use the arma::mat, but I doubt that makes any difference, and even more using vector you could remain consistent with already existing structure of action. Do you think it's reasonable.

> Also, we need to add the copy constructor for the pendulum's action, Shall we open a new PR?
What do you mean here, can you explain the context a bit more?

zoq:
@robotcator isn't the default copy constructor enough?

src/mlpack/methods/reinforcement_learning/ppo_impl.hpp
```diff
@@ -0,0 +1,186 @@
...
+  tanh.Fn(actionValue.col(0), sigma);
+  softp.Fn(actionValue.col(0), mu);
+
+  distribution::GaussianDistribution normalDist =
+    distribution::GaussianDistribution(sigma, mu);
+
+  // Get the action value for each action at current state.
+  ActorNetwork.Predict(state.Encode(), actionValue);
+
+  OldActorNetwork.Predict(state.Encode(), actionValue);
```
zoq: Please use lower camel casing for all parameters.

src/mlpack/methods/reinforcement_learning/ppo_impl.hpp
```diff
@@ -0,0 +1,186 @@
...
+  tanh.Fn(actionValue.col(0), sigma);
```
zoq: The ``Fn`` function is static, so you can write ``TanhFunction::Fn(actionValue.col(0), sigma)``, same for the ``SoftplusFunction`` function.

robotcator:
> You could also use the arma::mat, but I doubt that makes any difference, and even more using vector you could remain consistent with already existing structure of action. Do you think it's reasonable.

Yes, I tried to use `arma::mat`, it seems that there is some problem. Using `std::vector` is okay, but tit did not support indexing with vector. The code will look like this when sampling.

```
    for (size_t i = 0; i < sampledIndices.n_rows; i ++)
    {
      sampledActions[i] = actions[i];
    }
```

robotcator:
> @robotcator isn't the default copy constructor enough?

Sorry, my mistake. It is the const parameter's problem, not the copy constructor. If the parameter `size` defined as `const`,  the compiler will output error.
```
copy assignment operator of 'Action' is implicitly deleted because field 'size' is of
      const-qualified type 'const int'
    const int size = 1;
```

src/mlpack/methods/reinforcement_learning/ppo_impl.hpp
robotcator: > TanhFunction:

Great, that is what I am looking for.

zoq:
> Sorry, my mistake. It is the const parameter's problem, not the copy constructor. If the parameter `size` defined as `const`, the compiler will output error.
> 
> ```
> copy assignment operator of 'Action' is implicitly deleted because field 'size' is of
>       const-qualified type 'const int'
>     const int size = 1;
> ```

So, what about we remove the const qualifier?

robotcator:
> > Sorry, my mistake. It is the const parameter's problem, not the copy constructor. If the parameter `size` defined as `const`, the compiler will output error.
> > ```
> > copy assignment operator of 'Action' is implicitly deleted because field 'size' is of
> >       const-qualified type 'const int'
> >     const int size = 1;
> > ```
> 
> So, what about we remove the const qualifier?

Yes, I remove the `const` qualifier. Shall I open a new PR?

zoq:
Let's make it a part of this PR.

robotcator:
Here, I have two options to store the action for replay buffer. It seems that both of them will change the original codebase.

One solution is that we use std::vector to store the action. The other solution is to compact the action into one dimension vector, then store it to into arma::mat.

All those two solutions will face the problem that needs to distinguish discrete or continuous action.
Can I add some flag to distinguish?  Such as we use the action to index data in [q_learning.hpp](https://github.com/mlpack/mlpack/blob/master/src/mlpack/methods/reinforcement_learning/q_learning_impl.hpp#L164)

```
target(sampledActions(i), i) = sampledRewards(i) + config.Discount() *
          nextActionValues(bestActions(i), i);
```

src/mlpack/methods/reinforcement_learning/ppo_impl.hpp
```diff
@@ -0,0 +1,216 @@
...
+
+  replayMethod.Sample(sampledStates, sampledActions, sampledRewards,
+      sampledNextStates, isTerminal);
+
+  arma::mat nextActionValues;
+  double values;
+  criticNetwork.Predict(sampledNextStates, nextActionValues);
+  for (size_t i = sampledRewards.n_cols; i > 0; --i)
+  {
+    values = sampledRewards[i-1] + values * config.Discount();
```
zoq: Minor style issue, missing space between: ``[I - 1]``.

src/mlpack/methods/reinforcement_learning/ppo_impl.hpp
```diff
@@ -0,0 +1,216 @@
...
+    discountedRewards.insert_rows(discountedRewards.n_cols, values);
```
zoq: It should be more efficient, if we set the size of ``discountedRewards`` (outside the loop) first and use ``discountedRewards.col(i)`` afterwards.

src/mlpack/methods/reinforcement_learning/ppo_impl.hpp
```diff
@@ -0,0 +1,216 @@
...
+  // Get the action value for each action at current state.
+  arma::mat actionValue;
+  actorNetwork.Predict(state.Encode(), actionValue);
+
+  arma::mat sigma, mu;
+  ann::TanhFunction::Fn(actionValue.col(0), sigma);
+  ann::SoftplusFunction::Fn(actionValue.col(1), mu);
+
+  distribution::GaussianDistribution normalDist =
+    distribution::GaussianDistribution(sigma, mu);
```
zoq: Looks nice, just another minor style issue, this should be tabbed twice so 4 spaces.

src/mlpack/methods/reinforcement_learning/ppo_impl.hpp
```diff
@@ -0,0 +1,216 @@
...
+
+  // Get the action value for each action at current state.
+  actorNetwork.Predict(state.Encode(), actionValue);
+
+  oldActorNetwork.Predict(state.Encode(), actionValue);
+  ann::TanhFunction::Fn(actionValue.col(0), sigma);
+  ann::SoftplusFunction::Fn(actionValue.col(1), mu);
+
+  distribution::GaussianDistribution oldNormalDist =
+    distribution::GaussianDistribution(sigma, mu);
```
zoq: See comment above.

src/mlpack/methods/reinforcement_learning/ppo_impl.hpp
```diff
@@ -0,0 +1,216 @@
...
+  ann::SoftplusFunction::Fn(actionValue.col(1), mu);
+
+  distribution::GaussianDistribution normalDist =
+    distribution::GaussianDistribution(sigma, mu);
+
+  // Get the action value for each action at current state.
+  actorNetwork.Predict(state.Encode(), actionValue);
+
+  oldActorNetwork.Predict(state.Encode(), actionValue);
+  ann::TanhFunction::Fn(actionValue.col(0), sigma);
```
zoq: Nice that we can reuse the function from the ``ann`` namespace.

robotcator:
Sorry push code without test. I will fix.

src/mlpack/methods/reinforcement_learning/ppo_impl.hpp
robotcator: > It should be more efficient, if we set the size of `discountedRewards` (outside the loop) first and use `discountedRewards.col(i)` afterwards.

Sorry for miss this review, it is fixed.

src/mlpack/methods/reinforcement_learning/replay/random_replay.hpp
```diff
@@ -151,7 +154,7 @@ class RandomReplay
    * @param gradients The model's gradients
    */
   void Update(arma::mat /* target */,
-              arma::icolvec /* sampledActions */,
+              std::vector<ActionType> /* sampledActions */,
```
zoq: We should use ``std::vector<ActionType>&`` here.

src/mlpack/methods/reinforcement_learning/replay/prioritized_replay.hpp
```diff
@@ -214,15 +217,15 @@ class PrioritizedReplay
    * @param gradients The model's gradients.
    */
   void Update(arma::mat target,
-              arma::icolvec sampledActions,
+              std::vector<ActionType> sampledActions,
```
zoq: Can we use ``const std::vector<ActionType>& sampledActions`` here?

src/mlpack/methods/ann/loss_functions/surrogate_loss_impl.hpp
```diff
@@ -0,0 +1,49 @@
...
+{
+  // Nothing to do here.
+}
+
+template<typename InputDataType, typename OutputDataType>
+template<typename InputType, typename TargetType>
+double SurrogateLoss<InputDataType, OutputDataType>::Forward(
+  const InputType&& input, const TargetType&& target)
+{
+    return arma::clamp(input, 1-epsilon, 1+epsilon) * target;
```
zoq: Missing space between ``-`` and ``+``.

src/mlpack/methods/reinforcement_learning/replay/prioritized_replay.hpp
robotcator: > Can we use `const std::vector<ActionType>& sampledActions` here?

Yes, I can use `const type &` for `target`, `sampledActions`, `nextActionValues`. The same interface in `random_replay.hpp`.

zoq: Okay, great.

src/mlpack/methods/reinforcement_learning/ppo_impl.hpp
```diff
@@ -115,7 +115,10 @@ void PPO<
   oldNormalDist.Probability(actionValues, oldProb);
   arma::mat ratio =  prob / oldProb;
 
-  arma::mat loss = ratio * advantages;
+  arma::mat surrogateLoss1 = ratio * advantages;
+  arma::mat surrogateLoss2 = arma::clamp(ratio, 1 - config.Epsilon(),
+      1 + config.Epsilon()) * advantages;
+  arma::mat loss = - arma::min(surrogateLoss1, surrogateLoss2);
```
zoq: We might want to inline the ``surrogateLoss1`` -> ``const arma::mat loss = - arma::min(ratio * advantages, surrogateLoss2);``.

robotcator:
Hi, @zoq, can we store the `q_learning` and `ppo`'s training parameters in the same `training_config.hpp`, is that reasonable when we declare a `training_config` instance, only some of the related parameter are initialized? 

Such as training the `ppo` agent, only `Epsilon` for clipping the loss was set while other related to `q_learning`'s parameter are ignored. How to avoid the static code analysis checks problem?

zoq:
> Hi, @zoq, can we store the `q_learning` and `ppo`'s training parameters in the same `training_config.hpp`, is that reasonable when we declare a `training_config` instance, only some of the related parameter are initialized?

Yes, I think that is reasonable, we can definitely adapt the config file.

> Such as training the `ppo` agent, only `Epsilon` for clipping the loss was set while other related to `q_learning`'s parameter are ignored. How to avoid the static code analysis checks problem?

Will look at the static code analysis issues once the latest build finished.

zoq:

robotcator:
> @mlpack-jenkins test this please

Thanks, @zoq. It seems that the static code analysis checks still has problem in my part. Currently, most of the functionality is accomplished, I will explain what I have done. There is a problem with the code, I am debugging it now.

zoq:
Looks like there is one issue left (Static Code Analysis), let's see what the issue is, maybe we can ignore that one.

src/mlpack/methods/reinforcement_learning/training_config.hpp
```diff
@@ -25,7 +25,8 @@ class TrainingConfig
       stepLimit(0),
       explorationSteps(1),
       gradientLimit(40),
-      doubleQLearning(false)
+      doubleQLearning(false),
```
zoq: Looks like if we initialize the ``targetNetworkSyncInterval, stepSize, discount`` parameters in each constructor, we should be fine, in some cases, we don't have a good default value, or they are overwritten at some later anyway, in this case I think it's reasonable to just use ``0`` or something in that direction, to solve the static analysis warning.

robotcator:
Currently, I finished the functionality of PPO. I am debugging the memory access violation problem and finetuning the algorithm. Here, I have some problem to figure out. I write here for discussion.

1: For the loss function, I planned to write a surrogate loss class to calculate the function. I found that we need to pass more than two parameters, so I give up using that class. The implementation is in [here](https://github.com/mlpack/mlpack/blob/de79271971ec406efcbdffa95b27a506f453033d/src/mlpack/methods/reinforcement_learning/ppo_impl.hpp#L124) In the `ppo_test.cpp` file, I think using the empty loss function for the actor model is more reasonable, what do you think of it? @zoq

2: If we want to create a network the same with the other network, can we simply using the assignment like in this? [ppo](https://github.com/mlpack/mlpack/blob/de79271971ec406efcbdffa95b27a506f453033d/src/mlpack/methods/reinforcement_learning/ppo_impl.hpp#L62) and [q_learning](https://github.com/mlpack/mlpack/blob/de79271971ec406efcbdffa95b27a506f453033d/src/mlpack/methods/reinforcement_learning/q_learning_impl.hpp#L54)

zoq:
@robotcator Thanks for the update, I have to look though the code, but I make sure you have an answer later today.

zoq:
> Currently, I finished the functionality of PPO. I am debugging the memory access violation problem and finetuning the algorithm. Here, I have some problem to figure out. I write here for discussion.
> 
> 1: For the loss function, I planned to write a surrogate loss class to calculate the function. I found that we need to pass more than two parameters, so I give up using that class. The implementation is in [here](https://github.com/mlpack/mlpack/blob/de79271971ec406efcbdffa95b27a506f453033d/src/mlpack/methods/reinforcement_learning/ppo_impl.hpp#L124) In the `ppo_test.cpp` file, I think using the empty loss function for the actor model is more reasonable, what do you think of it? @zoq

I looked at the loss calculation and I agree looks reasonable to calculate it in the ``ppo`` class itself, in which case we need an empty loss, but I don't think that is an issue.

> 2: If we want to create a network the same with the other network, can we simply using the assignment like in this? [ppo](https://github.com/mlpack/mlpack/blob/de79271971ec406efcbdffa95b27a506f453033d/src/mlpack/methods/reinforcement_learning/ppo_impl.hpp#L62) and [q_learning](https://github.com/mlpack/mlpack/blob/de79271971ec406efcbdffa95b27a506f453033d/src/mlpack/methods/reinforcement_learning/q_learning_impl.hpp#L54)

Yes, that should call https://github.com/mlpack/mlpack/blob/324d1964b767d1e549072f09fee9094a0771ae1b/src/mlpack/methods/ann/ffn_impl.hpp#L627, just to be sure you can check if ``this->oldActorNetwork. Parameters()`` does return the same as ``learningNetwork.Parameters()``. Let me know if that worked, I can see if I can find the issue.

robotcator:
> > Currently, I finished the functionality of PPO. I am debugging the memory access violation problem and finetuning the algorithm. Here, I have some problem to figure out. I write here for discussion.
> > 1: For the loss function, I planned to write a surrogate loss class to calculate the function. I found that we need to pass more than two parameters, so I give up using that class. The implementation is in [here](https://github.com/mlpack/mlpack/blob/de79271971ec406efcbdffa95b27a506f453033d/src/mlpack/methods/reinforcement_learning/ppo_impl.hpp#L124) In the `ppo_test.cpp` file, I think using the empty loss function for the actor model is more reasonable, what do you think of it? @zoq
> 
> I looked at the loss calculation and I agree looks reasonable to calculate it in the `ppo` class itself, in which case we need an empty loss, but I don't think that is an issue.
> 
> > 2: If we want to create a network the same with the other network, can we simply using the assignment like in this? [ppo](https://github.com/mlpack/mlpack/blob/de79271971ec406efcbdffa95b27a506f453033d/src/mlpack/methods/reinforcement_learning/ppo_impl.hpp#L62) and [q_learning](https://github.com/mlpack/mlpack/blob/de79271971ec406efcbdffa95b27a506f453033d/src/mlpack/methods/reinforcement_learning/q_learning_impl.hpp#L54)
> 
> Yes, that should call
> 
> https://github.com/mlpack/mlpack/blob/324d1964b767d1e549072f09fee9094a0771ae1b/src/mlpack/methods/ann/ffn_impl.hpp#L627
> 
> , just to be sure you can check if `this->oldActorNetwork. Parameters()` does return the same as `learningNetwork.Parameters()`. Let me know if that worked, I can see if I can find the issue.

Thanks for your explanation. Currently, I was stuck on debugging the memory access violation, there are more than one place faces that problem. Let's try again. If I still can not figure it out, I hope we will have a discussion to figure it out together.

zoq:
Just for reference the comment about the ``parameter``was discussed in IRC (https://www.mlpack.org/irc/mlpack.20190727.html).

@robotcator Do you need any help with the distribution?

robotcator:
> Just for reference the comment about the `parameter`was discussed in IRC (https://www.mlpack.org/irc/mlpack.20190727.html).
> 
> @robotcator Do you need any help with the distribution?

Yes, It seems that I figure out what happened now.  I spend several days to implement a normal distribution, so sorry for the delay.

robotcator:
it seems wired that I can compile on my local computer, but the CI failed.

robotcator:
@zoq, Hi, I am facing a problem when how to back through the network. 

I create a network which defines like this. 
```
    FFN<EmptyLoss<>, GaussianInitialization> actor(
        EmptyLoss<>(), GaussianInitialization(0, 0.001));

    actor.Add<Linear<>>(2, 128);
    actor.Add<ReLULayer<>>();
    actor.Add<Linear<>>(128, 2);
```
The output was two dimensions, one is for mean and the other for sigma. I use them to create a normal distribution.  Here I declare two networks -- `actorNetwork` and `oldActorNetwork`. 

```
  arma::mat actionValue, sigma, mu;
  actorNetwork.Forward(sampledStates, actionValue);

  ann::TanhFunction::Fn(actionValue.row(0), mu);
  ann::SoftplusFunction::Fn(actionValue.row(1), sigma);

  ann::NormalDistribution normalDist =
      ann::NormalDistribution(mu.as_col(), sigma.as_col());

  oldActorNetwork.Forward(sampledStates, actionValue);
  ann::TanhFunction::Fn(actionValue.row(0), mu);
  ann::SoftplusFunction::Fn(actionValue.row(1), sigma);

  ann::NormalDistribution oldNormalDist =
      ann::NormalDistribution(mu.as_col(), sigma.as_col());
```
Then I use the obeservation to sample probability to calculate the ratio, finally get the loss. 
```
  normalDist.Probability(observation, prob);
  oldNormalDist.Probability(observation, oldProb);
  // row vector
  arma::mat ratio = (prob / oldProb).as_row();

  arma::mat surrogateLoss = arma::clamp(ratio, 1 - config.Epsilon(),
      1 + config.Epsilon()) % advantages;
  arma::mat loss = - arma::min(ratio % advantages, surrogateLoss);
```
I have some difficult to backward the loss through the network. The idea what I have is to calculate the derivation of `actionValue` (i.e. the output of `actorNetwork`) outside of the network. Then pass it into the backward function, let the gradient back through the network? I am not sure whether is correct, so I share the idea here. Hope that we can find a solution to this problem.

zoq:
What about we extend the ``FFN`` class by another Backward function that takes the error and does continue with the backward step, but without calling the output layer; basically: https://github.com/mlpack/mlpack/blob/301c77d98432f727c60cf5257cb8a0ee8f4ea4fc/src/mlpack/methods/ann/ffn_impl.hpp#L149-L172 but without the output layer step, since we already pass the error.

robotcator:
> What about we extend the `FFN` class by another Backward function that takes the error and does continue with the backward step, but without calling the output layer; basically:
> 
> https://github.com/mlpack/mlpack/blob/301c77d98432f727c60cf5257cb8a0ee8f4ea4fc/src/mlpack/methods/ann/ffn_impl.hpp#L149-L172
> 
> but without the output layer step, since we already pass the error.

Oh, Yes. But what I can not figure it out is that the loss calculated in [line](https://github.com/mlpack/mlpack/blob/9338f2ff25c7086aeff1874d2af0a55818d0341b/src/mlpack/methods/reinforcement_learning/ppo_impl.hpp#L145), how the gradient back through the network of `actorNetwork` defined in [this](https://github.com/mlpack/mlpack/blob/9338f2ff25c7086aeff1874d2af0a55818d0341b/src/mlpack/tests/ppo_test.cpp#L54).  
If I pass it in the [function](https://github.com/mlpack/mlpack/blob/9338f2ff25c7086aeff1874d2af0a55818d0341b/src/mlpack/methods/reinforcement_learning/ppo_impl.hpp#L150), it seems that we skip the part of those [lines](https://github.com/mlpack/mlpack/blob/9338f2ff25c7086aeff1874d2af0a55818d0341b/src/mlpack/methods/reinforcement_learning/ppo_impl.hpp#L121).
Can you help me to understand how the gradient backward?

zoq:
In this case, you have to do some manual calculation:

1. Get the derivative of the normal distribution: https://stackoverflow.com/a/54215856.
2. Backward through the distribution error = gy % derivative; where gy is the loss you calculated (I used https://github.com/mlpack/mlpack/blob/a8dd9f5301015429354973106d39293951bbccac/src/mlpack/methods/ann/layer/base_layer.hpp#L89 as a reference here).
3. Use the error from step two to call the backward function from the two activation functions.
4. Combine the two.
5. Call Backward of the actor-network.


The main idea is to backward through each module and get the new error, since you do some steps outside the defined network we have to do this manually; I can go into more detail if you like, just let me know.

robotcator:
> In this case, you have to do some manual calculation:
> 
> 1. Get the derivative of the normal distribution: https://stackoverflow.com/a/54215856.
> 2. Backward through the distribution error = gy % derivative; where gy is the loss you calculated (I used https://github.com/mlpack/mlpack/blob/a8dd9f5301015429354973106d39293951bbccac/src/mlpack/methods/ann/layer/base_layer.hpp#L89
>     as a reference here).
> 3. Use the error from step two to call the backward function from the two activation functions.
> 4. Combine the two.
> 5. Call Backward of the actor-network.
> 
> The main idea is to backward through each module and get the new error, since you do some steps outside the defined network we have to do this manually; I can go into more detail if you like, just let me know.

Sorry for the late reply. Thank you for your detailed explanation. Now I have a more clear mind of the backpropagation. The core idea is that I need to calculate the gradient, then multiply the error?

In this case, for the calculation of surrogate_loss, like the clamp operations, I also need to calucate the gradient, right?

zoq:
The gradient is only need if you have to update parameters, the surrogate loss doesn't have an trainable parameters so the derivative is all you need, but you are right for each operation you have to implement the backward step, you could take a look at the linear layer or base layer, the base layer is the one used for the activation function, so you can pretty much copy the code from there.

robotcator:
> The gradient is only need if you have to update parameters, the surrogate loss doesn't have an trainable parameters so the derivative is all you need, but you are right for each operation you have to implement the backward step, you could take a look at the linear layer or base layer, the base layer is the one used for the activation function, so you can pretty much copy the code from there.

Thank you, @zoq, your replay is helpful for me to complete the backward part. Can you review that part of code if you have time? I am not very sure of the correct, hoping that I didn't miss anything.

src/mlpack/tests/rl_components_test.cpp
```diff
@@ -227,7 +227,7 @@ BOOST_AUTO_TEST_CASE(RandomReplayTest)
   double reward = env.Sample(state, action, nextState);
   replay.Store(state, action, reward, nextState, env.IsTerminal(nextState));
   arma::mat sampledState;
-  arma::icolvec sampledAction;
+  std::vector<MountainCar::Action> sampledAction;
```
zoq: Can you test PPO on the ``DoublePoleCart`` env, see https://github.com/mlpack/mlpack/blob/12a50a055ba7f69340598329bd146ee37bec110f/src/mlpack/tests/q_learning_test.cpp#L401 for an example as well. Compared with the ``MountainCar`` env it's a more challenging to solve the env.

src/mlpack/methods/reinforcement_learning/ppo_impl.hpp
```diff
@@ -0,0 +1,276 @@
...
+    observation[i] = sampledActions[i].action;
+  }
+  normalDist.Probability(observation, prob);
+  oldNormalDist.Probability(observation, oldProb);
+
+  arma::mat ratio = vectorise((prob / oldProb), 1);
+
+  arma::mat surrogateLoss = arma::clamp(ratio, 1 - config.Epsilon(),
+      1 + config.Epsilon()) % advantages;
+  arma::mat loss = - arma::min(ratio % advantages, surrogateLoss);
```
zoq: Not sure I get this, where is this coming from?

zoq: Don't get me wrong there is nothing wrong with the line, I just don't get where this is coming from.

src/mlpack/tests/rl_components_test.cpp
robotcator: > Can you test PPO on the `DoublePoleCart` env, see
> 
> https://github.com/mlpack/mlpack/blob/12a50a055ba7f69340598329bd146ee37bec110f/src/mlpack/tests/q_learning_test.cpp#L401
> 
> for an example as well. Compared with the `MountainCar` env it's a more challenging to solve the env.

Yes, I know this task. But currently, the PPO implement of mine is only supporting continuous task like `PENDULUM`.

src/mlpack/methods/reinforcement_learning/ppo_impl.hpp
robotcator: > Don't get me wrong there is nothing wrong with the line, I just don't get where this is coming from.

Do you mean the equation of loss calculation? It is the PPO-Clip method. You can refer to [there](https://spinningup.openai.com/en/latest/algorithms/ppo.html#key-equations) or in this [paper](https://arxiv.org/pdf/1707.06347.pdf).

zoq: Exactly, thanks for the reference.

src/mlpack/tests/rl_components_test.cpp
zoq: Should we implement another continuous task, that is more challenging, e.g. what do you think about LunarLander: https://github.com/zippkidd/lunarLander?

src/mlpack/methods/reinforcement_learning/ppo_impl.hpp
```diff
@@ -0,0 +1,279 @@
...
+
+    // backward the gradient
+    arma::mat dratio1 = -loss % (ratio % advantages <= surrogateLoss)
+                        % advantages;
+    arma::mat dsurro = -loss % (ratio % advantages >= surrogateLoss);
+    arma::mat dratio2 = (ratio >= (1 - config.Epsilon())) %
+        (ratio <= (1 + config.Epsilon())) % advantages % dsurro;
+
+    arma::mat dprob = (dratio1 + dratio2) %
+        arma::exp(vectorise(prob - oldProb, 1));
```
zoq: If I haven't missed anything, ``prob - oldProb`` should already be a vector, so ``vectorise`` shouldn't be necessary.

src/mlpack/methods/reinforcement_learning/ppo_impl.hpp
```diff
@@ -0,0 +1,279 @@
...
+
+    arma::mat dmu = (vectorise(observation, 1) - mu) /
```
zoq: See comment above, if you need a rowvec, it's more efficient to just transpose the colvec.

src/mlpack/methods/reinforcement_learning/ppo_impl.hpp
```diff
@@ -0,0 +1,279 @@
...
+        (arma::square(sigma)) % dprob;
+
+    arma::mat dsigma = -1 / sigma +
+        arma::square(vectorise(observation, 1) - mu) / arma::pow(sigma, 3);
+
+    arma::mat dTanh, dSoftP;
+    ann::TanhFunction::Deriv(vectorise(dmu, 1), dTanh);
```
zoq: I don't think you have to use vectorise on ``dmu``, if you do ``dmu.t()`` should do as well.

src/mlpack/tests/rl_components_test.cpp
robotcator: > Should we implement another continuous task, that is more challenging, e.g. what do you think about LunarLander: https://github.com/zippkidd/lunarLander?

Yes, LunarLander is another continuous task, it is more challenging than Pendulum. But I am not sure how many workload of that, I can try to do that.

src/mlpack/methods/reinforcement_learning/ppo_impl.hpp
robotcator: > If I haven't missed anything, `prob - oldProb` should already be a vector, so `vectorise` shouldn't be necessary.

Oh, yes, I miss that.

src/mlpack/methods/reinforcement_learning/ppo_impl.hpp
```diff
@@ -155,20 +156,17 @@ void PPO<
...
-    arma::mat dsigma = -1 / sigma +
-        arma::square(vectorise(observation, 1) - mu) / arma::pow(sigma, 3);
+    arma::mat dmu = (observation.t() - mu) / (arma::square(sigma)) % dprob;
+    arma::mat dsigma = -1.0 / sigma +
+        arma::square(observation.t() - mu) / arma::pow(sigma, 3);
 
     arma::mat dTanh, dSoftP;
-    ann::TanhFunction::Deriv(vectorise(dmu, 1), dTanh);
-    ann::SoftplusFunction::Deriv(vectorise(dsigma, 1), dSoftP);
+    ann::TanhFunction::Deriv(dmu.t(), dTanh);
```
zoq: If you don't transpose ``dmu, dsigma`` you don't have to transpose it in the ``join_cols(..)`` method.

src/mlpack/tests/rl_components_test.cpp
zoq: You can take a look at the repo I referenced above, there are mainly two functions that are interesting:

```c++
void updateStatus(double &velocity, double burnAmount, double &fuelRemaining, double &height)
{
    fuelRemaining = fuelRemaining - burnAmount;
    height = height - ((((velocity - burnAmount) + 5) + velocity)/2);
    velocity = velocity - burnAmount + 5;
}
void touchdown(double &elapsedTime, double &velocity, double &burnAmount, double &height)
{
    double oldvelocity = velocity + burnAmount - 5;
    double oldheight = height + ((velocity + oldvelocity)/2);
    double oldtime = elapsedTime - 1;
    double fraction = 0.0;
    if (burnAmount == 5)
        fraction = oldheight / oldvelocity;
    else
        fraction = (sqrt((oldvelocity*oldvelocity) + (oldheight*(10 - (2 * burnAmount)))) - oldvelocity) / (5 - burnAmount);
    elapsedTime = oldtime + fraction*1.0;
    velocity = oldvelocity + (5 - burnAmount)*fraction *1.0;
    height = oldheight;
}
```

src/mlpack/methods/reinforcement_learning/ppo_impl.hpp
robotcator: > If you don't transpose `dmu, dsigma` you don't have to transpose it in the `join_cols(..)` method.

Oh, yes, such a silly mistake. Fixed.

src/mlpack/tests/rl_components_test.cpp
robotcator: > You can take a look at the repo I referenced above, there are mainly two functions that are interesting:
> 
> ```c++
> void updateStatus(double &velocity, double burnAmount, double &fuelRemaining, double &height)
> {
>     fuelRemaining = fuelRemaining - burnAmount;
>     height = height - ((((velocity - burnAmount) + 5) + velocity)/2);
>     velocity = velocity - burnAmount + 5;
> }
> void touchdown(double &elapsedTime, double &velocity, double &burnAmount, double &height)
> {
>     double oldvelocity = velocity + burnAmount - 5;
>     double oldheight = height + ((velocity + oldvelocity)/2);
>     double oldtime = elapsedTime - 1;
>     double fraction = 0.0;
>     if (burnAmount == 5)
>         fraction = oldheight / oldvelocity;
>     else
>         fraction = (sqrt((oldvelocity*oldvelocity) + (oldheight*(10 - (2 * burnAmount)))) - oldvelocity) / (5 - burnAmount);
>     elapsedTime = oldtime + fraction*1.0;
>     velocity = oldvelocity + (5 - burnAmount)*fraction *1.0;
>     height = oldheight;
> }
> ```

Oh, yes, I read that implementation, it is not so complicated. But I found that it is different from the Lunarlander game in the GYM.
The gym implementation's action is [Action is two real values vector from -1 to +1. First controls main engine, -1..0 off, 0..+1 throttle from 50% to 100% power. The engine can't work with less than 50% power. Second value -1.0..-0.5 fire left engine, +0.5..+1.0 fire right engine, -0.5..0.5 off.](https://github.com/openai/gym/wiki/Leaderboard#lunarlandercontinuous-v2)

In this [implementation](https://github.com/zippkidd/lunarLander/blob/master/LanderSim.cpp), the action is the amount of fuel to use.

robotcator: What do you think of the [continuous mountain car task](https://github.com/mlpack/mlpack/blob/master/src/mlpack/methods/reinforcement_learning/environment/continuous_mountain_car.hpp)?

The implementation in [gym.](https://github.com/openai/gym/blob/master/gym/envs/classic_control/continuous_mountain_car.py)

zoq: The MountainCar env is similar to the Pendulum env, we could test against the ``ContinuousDoublePoleCart`` which is more difficult, not as difficult as the lunar lander env.

> Oh, yes, I read that implementation, it is not so complicated. But I found that it is different from the Lunarlander game in the GYM.
> The gym implementation's action is Action is two real values vector from -1 to +1. First controls main engine, -1..0 off, 0..+1 throttle from 50% to 100% power. The engine can't work with less than 50% power. Second value -1.0..-0.5 fire left engine, +0.5..+1.0 fire right engine, -0.5..0.5 off.
> 
> In this implementation, the action is the amount of fuel to use.
> 

We could go with the OpenAI gym settings.

robotcator: > The MountainCar env is similar to the Pendulum env, we could test against the `ContinuousDoublePoleCart` which is more difficult, not as difficult as the lunar lander env.
> 
> > Oh, yes, I read that implementation, it is not so complicated. But I found that it is different from the Lunarlander game in the GYM.
> > The gym implementation's action is Action is two real values vector from -1 to +1. First controls main engine, -1..0 off, 0..+1 throttle from 50% to 100% power. The engine can't work with less than 50% power. Second value -1.0..-0.5 fire left engine, +0.5..+1.0 fire right engine, -0.5..0.5 off.
> > In this implementation, the action is the amount of fuel to use.
> 
> We could go with the OpenAI gym settings.

Can we implement the env use the setting in [there](https://github.com/zippkidd/lunarLander/blob/master/LanderSim.cpp), then go with the OpenAI gym settings step by step?

zoq: Sounds reasonable, the basic idea is to see if PPO works on more challenging tasks as well as for the current Pendulum env.

zoq: Let me know if you need any help with the implementation.

robotcator: > Let me know if you need any help with the implementation.

Oh, I am implementing the Lunarlander now, can I implement it in this PR?

robotcator: > > Let me know if you need any help with the implementation.
> 
> Oh, I am implementing the Lunarlander now, can I implement it in this PR?

I can not figure out why one of the CI failed.

robotcator: > > > Let me know if you need any help with the implementation.
> > 
> > 
> > Oh, I am implementing the Lunarlander now, can I implement it in this PR?
> 
> I can not figure out why one of the CI failed, can you take a look?

zoq: > Oh, I am implementing the Lunarlander now, can I implement it in this PR?

I think that would be the best idea.

zoq: The appveyor build failure isn't related with the changes from the PR, looks like the latest MSVC version has some issue with loading arff files.

robotcator: With the implementation in [here](https://github.com/zippkidd/lunarLander/blob/master/LanderSim.cpp), I found that we hardly determine the reward of the agent. What's more, I can not define when the agent is solved. 

In the [gym's setting](https://github.com/openai/gym/wiki/Leaderboard), the Lunarlander defines "solving" as getting an average reward of 200 over 100 consecutive trials. 

So I think we can go with the OpenAI's setting was better, what do you think of it? Can you give me some time to figure out the gym's implementation?

zoq: > In the [gym's setting](https://github.com/openai/gym/wiki/Leaderboard), the Lunarlander defines "solving" as getting an average reward of 200 over 100 consecutive trials.

The implementation always implements a single trial, so the user has to run the same model against the env 100 consecutive trials in a row (starting with a need random starting point) to check if the model is able to solve the env.

But you are right a single Lunarlander iteration is solved if you get 200 points.

> So I think we can go with the OpenAI's setting was better, what do you think of it? Can you give me some time to figure out the gym's implementation?

Sure, let me know if you need any help.

src/mlpack/methods/reinforcement_learning/environment/lunarlander.hpp
```diff
@@ -0,0 +1,232 @@
...
+namespace rl{
+
+/**
+ * Implementation of Lunarlander game. Lunarlander is a 2-link pendulum with only the
+ * second joint actuated. Intitially, both links point downwards. The goal is
+ * to swing the end-effector at a height at least the length of one link above
+ * the base. Both links can swing freely and can pass by each other, i.e.,
+ * they don't collide when they have the same angle.
+ */
+class LunarLander
```
zoq: Let me know once you think the lunar lander code is ready or if you need any help from my side.

robotcator: Thank you, I am writing the OpenAI environment setting. The implementation of lunarlander refer to large box2d codes, so I am reading the code the figure out how to implement in C++.

robotcator: This implementation was referring to [this](https://github.com/zippkidd/lunarLander).  I don't it is good to use this implementation because we don't know the detailed definition of reward and etc. To make consistent with current mainstream setting, I am implementing the OpenAI settings.

zoq: I don't mind to implement a variation of the lunar lander env, as we only like to have another  continuous env for testing.

robotcator: Okay, I got it. So, what about simplifying the gym setting, and refer to the [zippkidd](https://github.com/zippkidd/lunarLander) implementation. The implementation will look like that we only control the Lunarlander in the vertical direction. The action will be the amount of fuel to use, we use it to control the speed. When the height is zero, we begin to call the touchdown function to determine the final speed. Then we use the final speed to indicate how well the system landing is.  I will use speed as an indicator to assign reward according to the [piecewise function](https://github.com/zippkidd/lunarLander/blob/master/LanderSim.cpp#L116)

Does this make sense? Can you have a look whenever you have time? 

One more thing, for the simulation discussed in IRC, do you mean the we simulate it by human hands?

zoq: > Does this make sense? Can you have a look whenever you have time?

Sounds good to me, let's see if PPO is able to solve the env.

> One more thing, for the simulation discussed in IRC, do you mean the we simulate it by human hands?

The keyboard agent? I just referenced it, to get familiar with the env, we don't use it for training.

robotcator: Oh, I mean how to set threashold when we define the PPO works like other env? Most of the time, I refer to this [web pages](https://github.com/openai/gym/wiki/Leaderboard). I do not have a good way to get the threashold.

zoq: That depends on the initial start reward, the reward you add if e.g. ``velocity == 0``, do you give a negative reward per action as well? An example would be: initial reward 400, -0.5 per action:

``velocity > 0.0 && velocity < 2.0`` -> reward +100
``velocity >= 2.0 && velocity < 5.0`` -> reward +50
``velocity >= 5.0 && velocity < 10.0`` -> reward -10
``velocity >= 10.0 && velocity < 30.0`` -> reward -30

maybe we can come up with a reward that depends on the velocity, something like reward = -velocity could work, but ideally we add a positive reward (e.g. 100) if ``velocity > 0.0 && velocity < 2.0``.

main engine -> reward -0.3

so I think a reasonable threshold would be 400, the reward we get for a perfect landing should account for the actions (left/right, main engine). We could start with that and train PPO with it and see if the threshold looks reasonable. Let me know what you think.

zoq:
Should we test PPO against the new env?

robotcator:
yes, we need do that and I have writen the test, but I am moving house these days, I will back in 2 days, sorry about that.

robotcator:
It seems that the RL algorithm is very unstable, I spent several days to tuning the parameter for Lunarlander to make it pass the unit test, it is failed. So I decided to add the unit test of the continuous mountain car task on PPO, it works on my computer and gets the reward above 100.
```
Average return: 447.358 Episode return: 447.358
```
But when switch on the other compute, the model with the same parameters was failed. lol.

zoq:
Just to get it right it worked you tested PPO on the mountain car task, which worked, but the same PPO couldn't solve the Lunarlander task?

robotcator:
> Just to get it right it worked you tested PPO on the mountain car task, which worked, but the same PPO couldn't solve the Lunarlander task?

Sorry for the late reply. Actually, yes, it worked on the mountain car task, but the same PPO failed to solve the lunarlander task. I am also investing it why this happened.

zoq:
I don't think the same model necessarily is able to solve the Lunarlander task as well, especially if the task is more complex, but nonetheless we have to figure out what went wrong.

zoq:
@robotcator do you have any updates on this one, happy to help.

robotcator:
> @robotcator do you have any updates on this one, happy to help.

Sorry for missing the massages. I got involved in something else in those days. My problem is that we are hard to figure out whether the model is not carefully tuned or the lunar env is not implemented correctly.

zoq:
No worries, I'll take another look at the env, what I do to check if the env is correct is to manually test some inputs and check if the result is reasonable.

robotcator:
> No worries, I'll take another look at the env, what I do to check if the env is correct is to manually test some inputs and check if the result is reasonable.

Thank you, I will handle the continuous mountain car to make it work. For the lunarlander, do you think if the agent receives the inputs of humans, the behavior of the agent can work as expected, such as fall down according to the speed, then we can say that the agent works?

By the way, from the CI I found that the AdamUpdate has no member named "Update", is there any API changed or something else?

zoq:
> Thank you, I will handle the continuous mountain car to make it work. For the lunarlander, do you think if the agent receives the inputs of humans, the behavior of the agent can work as expected, such as fall down according to the speed, then we can say that the agent works?

Yes, I think that is a reasonable approach.

> By the way, from the CI I found that the AdamUpdate has no member named "Update", is there any API changed or something else?


Right, we can use the same approach as used here: https://github.com/mlpack/mlpack/blob/cf71ecef94a6bb565dcbeb4dfea73928a4ab753f/src/mlpack/methods/reinforcement_learning/worker/n_step_q_learning_worker.hpp#L86-L91 to solve the issue.

robotcator:
@zoq Sorry, I have to say that I may have less time than the summer holidays, but I will try to get it merged in my spare time.

zoq:
Sounds good, hopefully you don't have too much stress.

robotcator:
> Sounds good, hopefully you don't have too much stress.

Oh, I must say sorry for the delay. Also, I think I need to push myself to accomplish this merge request even has limited time.

rcurtin:
@robotcator just let us know if we can help out. :+1:

robotcator:
> @robotcator just let us know if we can help out. 👍

Thanks for your concern. I am too busy the last few weeks, so I am 'disappeared' at that time.

robotcator:
I am confused about the azure pipeline, it that mean it fails at the build stage or test stage?

zoq:
Both, you can click on the jobs to see at which stage it failed; the windows builds are failing at the test stage, the linux builds are failing at the build stage.

The linux build issue is fixed in #https://github.com/mlpack/mlpack/pull/2090.

robotcator:
> Both, you can click on the jobs to see at which stage it failed; the windows builds are failing at the test stage, the linux builds are failing at the build stage.
> 
> The linux build issue is fixed in ##2090.

Great, I got it.

robotcator:
@zoq, Hi, I still have trouble with how to define `works` in the unit test. Just like the test case in [Q_learning_test](https://github.com/mlpack/mlpack/blob/cf71ecef94a6bb565dcbeb4dfea73928a4ab753f/src/mlpack/tests/q_learning_test.cpp#L91), how can I say that the agent 'works' when the average return meets some threshold. How can find the threshold?

zoq:
@robotcator Sorry for the slow response, in this case we just have to manually figure out what the threshold could be, e.g. by manually calculating it, I can take a look into the test and give a threshold.

robotcator:
> @robotcator Sorry for the slow response, in this case we just have to manually figure out what the threshold could be, e.g. by manually calculating it, I can take a look into the test and give a threshold.

Thanks, I fix some bugs in the lunar lander environment. During the debug time, I found that it is a tough task to design an environment, such as how to initialize the initial state and how to design the logic of the game. It is not a good start for the test if we can not follow some commonly used task implementation. 

If we still use the task, I think the random strategy is a good start to get the first baseline reward of the environment. The code is a little bit like this.  

```
  LunarLander env;
  double totalReturn = 0.0;
  double totalNumberGame = 2;
  for (int i = 0; i < totalNumberGame; i ++)
  {
    LunarLander::State state = env.InitialSample();
    LunarLander::State nextState;
    LunarLander::Action action;
    double oneEpochReturn = 0.0;
    while (!env.IsTerminal(state)) 
    {
      action.action[0] = arma::randu();
      double reward = env.Sample(state, action, nextState);
      std::cout << "reward: " << reward << std::endl;
      state = nextState;
      oneEpochReturn += reward;
    }
    totalReturn += oneEpochReturn;
  }
  std::cout << totalReturn / totalNumberGame << std::endl;
```

Here, I am listening to your advice, what do you think about it?

rcurtin:

nishantkr18:
Hey @robotcator @zoq @manish7294 ! Could I work on this PR? As of now, the lunarLander test is failing, is that correct?

src/mlpack/methods/reinforcement_learning/ppo_impl.hpp
```diff
@@ -0,0 +1,342 @@
...
+  arma::mat sampledNextStates;
+  arma::icolvec isTerminal;
+
+  replayMethod.Sample(sampledStates, sampledActions, sampledRewards,
+                      sampledNextStates, isTerminal);
+
+  arma::rowvec discountedRewards(sampledRewards.n_rows);
+  arma::mat nextActionValues;
+  double values = 0.0;
+  criticNetwork.Predict(sampledNextStates, nextActionValues);
```
nishantkr18: @robotcator could you please explain the requirement of `criticNetwork.Predict` here, I think `nextActionValues` is not used anywhere in this part

nishantkr18:
Also, I think renaming `LunarLander` to `LunarLanderContinuous` would be more clear, as `LunarLander` is generally used for discrete action_space

robotcator:
> Also, I think renaming `LunarLander` to `LunarLanderContinuous` would be more clear, as `LunarLander` is generally used for discrete action_space

Yes, you are right. I will update it.

robotcator:
> Hey @robotcator @zoq @manish7294 ! Could I work on this PR? As of now, the lunarLander test is failing, is that correct?

Yes, some of the platform tests are failed somehow.

nishantkr18:
> Yes, some of the platform tests are failed somehow. Great, I think we can work on this PR to make it merged because I don't have enough time to focus on this PR.

Great! I'd be happy to help :) 
I guess we can start with merging changes from master. Also, how could I contribute to this PR? I guess I'd be needing push access to this branch, or maybe there is some other way?

src/mlpack/methods/reinforcement_learning/ppo_impl.hpp
```diff
@@ -0,0 +1,342 @@
...
+  #if ENS_VERSION_MAJOR >= 2
+  actorUpdatePolicy(NULL),
+  #endif
+  criticUpdater(std::move(updater)),
+  #if ENS_VERSION_MAJOR >= 2
+  criticUpdatePolicy(NULL),
+  #endif
+  policy(std::move(policy)),
+  replayMethod(std::move(replayMethod)),
+  environment(std::move(environment))
```
nishantkr18, line 58: I guess, we are never able to actually train the agent as `deterministic` is not initialized, so it takes a garbage value(not False or 0). Also, `totalSteps` is not initialized

src/mlpack/methods/reinforcement_learning/ppo_impl.hpp
```diff
@@ -0,0 +1,342 @@
...
+  replayMethod.Store(state, action, reward, nextState,
+      environment.IsTerminal(nextState));
+
+  // Update current state.
+  state = nextState;
+
+  if (deterministic)
+    return reward;
+
+  return reward;
```
nishantkr18, line 287: Maybe this is redundant?

src/mlpack/methods/reinforcement_learning/ppo_impl.hpp
```diff
@@ -0,0 +1,342 @@
...
+    arma::mat actionParameter, sigma, mu;
+    actorNetwork.Forward(sampledStates, actionParameter);
+
+    ann::TanhFunction::Fn(actionParameter.row(0), mu);
+    ann::SoftplusFunction::Fn(actionParameter.row(1), sigma);
+
+    ann::NormalDistribution normalDist =
+      ann::NormalDistribution(vectorise(mu, 0), vectorise(sigma, 0));
+
+    oldActorNetwork.Forward(sampledStates, actionParameter);
```
nishantkr18, line 176: Is Forward required here? Shouldn't we use Predict?

src/mlpack/methods/reinforcement_learning/ppo_impl.hpp
robotcator: > Maybe this is redundant?

Yes, it is redundant.

src/mlpack/methods/reinforcement_learning/ppo_impl.hpp
robotcator: it seems that it is. I will look at it this problem.

nishantkr18: Actually, I have been working on this PR for a few days now, and have made the required changes including removal of some merge conflicts and other bugs, a PR has been made to this branch with the changes..
Also, I have made changes to the backward gradient, please have a look when free..

Lastly, it would be great if you could provide push access to this branch because I am eager to have my recent changes reviewed by you and the other members! With these changes, PPO finally converges for MountainCar and Pendulum perfectly! As for LunarLander, I guess, we'd need to have a look at the implementation, or tune some hyperparameters..

robotcator: > Actually, I have been working on this PR for a few days now, and have made the required changes including removal of some merge conflicts and other bugs, a PR has been made to this branch with the changes..
> Also, I have made changes to the backward gradient, please have a look when free..
> 
> Lastly, it would be great if you could provide push access to this branch because I am eager to have my recent changes reviewed by you and the other members! With these changes, PPO finally converges for MountainCar and Pendulum perfectly! As for LunarLander, I guess, we'd need to have a look at the implementation, or tune some hyperparameters..

Great, I will give access to you now. For the Lunarlander, the implementation is also need to think twice.

shrit:
@zoq I think we can close this one since #2788 is finishing it right?

Issue #1911, opened 2019-05-30
hoeffding_tree_test.cpp keep warning: moving a temporary object prevents copy elision


robotcator:
#### Issue description

Warning in hoeffding_tree_test.cpp when compiling the mlpack. 

```
mlpack/src/mlpack/tests/main_tests/hoeffding_tree_test.cpp:512:29: warning: 
      moving a temporary object prevents copy elision [-Wpessimizing-move]
  SetInputParam("training", std::move(std::make_tuple(info, inputData)));
                            ^
/Users/robotcator/clone/boost/dev-mlpack/mlpack/src/mlpack/tests/main_tests/hoeffding_tree_test.cpp:512:29: note: 
      remove std::move call here
  SetInputParam("training", std::move(std::make_tuple(info, inputData)));
```

#### Your environment

 * version of mlpack: mlpack master branch 
 * operating system: MacOS Sierra 10.12.6
 * compiler: clang 
 * version of dependencies (Boost/Armadillo): boost 1.68

rcurtin:
I think it's just fine to remove the `std::move()` there like the warning suggests---want to open a PR with that change?  If not I can handle it.

robotcator:
> I think it's just fine to remove the `std::move()` there like the warning suggests---want to open a PR with that change? If not I can handle it.

Okay, done in [PR1914](https://github.com/mlpack/mlpack/pull/1914).

jeffin143:
@robotcator , since it is done , we can close the issue :). Thanks for the issue and fixes

rcurtin:
Fixed in #1914. Thanks! :+1:

PR #1914, opened 2019-05-31
remove move() statement in hoeffding_tree_test.cpp

robotcator:
Like what we said in [issue](https://github.com/mlpack/mlpack/issues/1911)

### DIFF
--- a/src/mlpack/tests/main_tests/hoeffding_tree_test.cpp
+++ b/src/mlpack/tests/main_tests/hoeffding_tree_test.cpp
@@ -66,25 +66,25 @@ BOOST_AUTO_TEST_CASE(HoeffdingTreeOutputDimensionTest)
   size_t testSize = testData.n_cols;
 
   // Input training data.
-  SetInputParam("training", std::move(std::make_tuple(info, inputData)));
+  SetInputParam("training", std::make_tuple(info, inputData));
   SetInputParam("labels", std::move(labels));
 
   // Input test data.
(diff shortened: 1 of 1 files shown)


src/mlpack/tests/main_tests/hoeffding_tree_test.cpp
```diff
@@ -66,25 +66,25 @@ BOOST_AUTO_TEST_CASE(HoeffdingTreeOutputDimensionTest)
...
   // Input test data.
-  SetInputParam("test", std::move(std::make_tuple(info, testData)));
+  SetInputParam("test", std::make_tuple(info, testData));
 
   mlpackMain();
 
   // Check that number of output points are equal to number of input points.
   BOOST_REQUIRE_EQUAL(
-      CLI::GetParam<arma::Row<size_t>>("predictions").n_cols, testSize);
+    CLI::GetParam<arma::Row<size_t>>("predictions").n_cols, testSize);
```
rcurtin: Ah, these are wrapped lines, so they should be doubly-indented (that is, they were correct before the change).  Can you change them back please?  :+1:

robotcator: > Ah, these are wrapped lines, so they should be doubly-indented (that is, they were correct before the change). Can you change them back please? 👍

Okay, it seems that my IDE automatically formats the code style. Now they changed back.

Issue #1910, opened 2019-05-30
Limiting tree depth of random_forest 


yanivbi:
Is there a way to limit the maximum depth of each tree in random_forest?
This is a prevalent option for configuring random forests.

rcurtin:
Hi @yanivbi, I've said to others in the past that `minimum_leaf_size` is similar way to achieve the same goal, but at this point I've said that too many times and I think it is time to add the support.  So, at the moment, there's not a way to limit the maximum depth.

If you are (or anyone else reading this issue is) interested in contributing one, basically what would need to happen is that the constructors for the `DecisionTree` class would need an extra final parameter `maximum_depth`.  Then, if `maximum_depth == 1`, the node should not be allowed to split during training; otherwise, when it splits during training, the children should be trained with maximum depth set to `maximum_depth - 1`.

In any case, if you don't have the time, that's okay, I will get to it eventually (but I am not sure exactly when that will be).

Rajlaxmi:
Hi,
I would like to pick this up.

zoq:
@Rajlaxmi looks like @Yashwants19 already opened a PR, but please feel free to comment on the PR.

rcurtin:
Since #1916 is merged, the issue is fixed.  The `maximum_depth` parameter will be available starting with the next release of mlpack, but if you want "early access" you can always build from the git master branch and it'll be available. :+1: I'll close the issue, but if you have any problems in the future please don't hesitate to open other issues.

PR #1916, opened 2019-06-01
Add new paramter Maximum_depth to DTree And Random Forest

Yashwants19:
Address #1910

### DIFF
--- a/src/mlpack/methods/decision_tree/all_categorical_split.hpp
+++ b/src/mlpack/methods/decision_tree/all_categorical_split.hpp
@@ -47,7 +47,6 @@ class AllCategoricalSplit
    * @param numClasses Number of classes in the dataset.
    * @param minimumLeafSize Minimum number of points in a leaf node for
    *      splitting.
-   * @param maximumDepth Maximum Depth Minimum for the tree.
    * @param classProbabilities Class probabilities vector, which may be filled
    *      with split information a successful split.
    * @param aux Auxiliary split information, which may be modified on a
@@ -63,7 +62,6 @@ class AllCategoricalSplit
       const WeightVecType& weights,
       const size_t minimumLeafSize,
       const double minimumGainSplit,
-      const size_t maximumDepth,
       arma::Col<typename VecType::elem_type>& classProbabilities,
       AuxiliarySplitInfo<typename VecType::elem_type>& aux);
 
--- a/src/mlpack/methods/decision_tree/all_categorical_split.hpp
+++ b/src/mlpack/methods/decision_tree/all_categorical_split.hpp
@@ -47,6 +47,7 @@ class AllCategoricalSplit
    * @param numClasses Number of classes in the dataset.
    * @param minimumLeafSize Minimum number of points in a leaf node for
    *      splitting.
+   * @param maximumDepth Maximum Depth Minimum for the tree.
    * @param classProbabilities Class probabilities vector, which may be filled
    *      with split information a successful split.
    * @param aux Auxiliary split information, which may be modified on a
@@ -62,6 +63,7 @@ class AllCategoricalSplit
       const WeightVecType& weights,
       const size_t minimumLeafSize,
       const double minimumGainSplit,
+      const size_t maximumDepth,
       arma::Col<typename VecType::elem_type>& classProbabilities,
       AuxiliarySplitInfo<typename VecType::elem_type>& aux);
 
--- a/src/mlpack/methods/decision_tree/best_binary_numeric_split_impl.hpp
+++ b/src/mlpack/methods/decision_tree/best_binary_numeric_split_impl.hpp
@@ -25,14 +25,9 @@ double BestBinaryNumericSplit<FitnessFunction>::SplitIfBetter(
     const WeightVecType& weights,
     const size_t minimumLeafSize,
     const double minimumGainSplit,
-    const size_t maximumDepth,
     arma::Col<typename VecType::elem_type>& classProbabilities,
     AuxiliarySplitInfo<typename VecType::elem_type>& /* aux */)
 {
-  // First sanity check: if we have reached maximum depth, we can't split.
-  if (maximumDepth == 1)
-    return DBL_MAX;
-
   // First sanity check: if we don't have enough points, we can't split.
   if (data.n_elem < (minimumLeafSize * 2))
     return DBL_MAX;
--- a/src/mlpack/methods/decision_tree/best_binary_numeric_split_impl.hpp
+++ b/src/mlpack/methods/decision_tree/best_binary_numeric_split_impl.hpp
@@ -25,9 +25,14 @@ double BestBinaryNumericSplit<FitnessFunction>::SplitIfBetter(
     const WeightVecType& weights,
     const size_t minimumLeafSize,
     const double minimumGainSplit,
+    const size_t maximumDepth,
     arma::Col<typename VecType::elem_type>& classProbabilities,
     AuxiliarySplitInfo<typename VecType::elem_type>& /* aux */)
 {
+  // First sanity check: if we have reached maximum depth, we can't split.
+  if (maximumDepth == 1)
+    return DBL_MAX;
+
   // First sanity check: if we don't have enough points, we can't split.
   if (data.n_elem < (minimumLeafSize * 2))
     return DBL_MAX;
--- a/src/mlpack/methods/decision_tree/decision_tree.hpp
+++ b/src/mlpack/methods/decision_tree/decision_tree.hpp
@@ -74,7 +74,7 @@ class DecisionTree :
                const size_t numClasses,
                const size_t minimumLeafSize = 10,
                const double minimumGainSplit = 1e-7,
-               const size_t maximumDepth = 20,
+               const size_t maximumDepth = 0,
                DimensionSelectionType dimensionSelector =
                    DimensionSelectionType());
 
@@ -100,7 +100,7 @@ class DecisionTree :
                const size_t numClasses,
                const size_t minimumLeafSize = 10,
                const double minimumGainSplit = 1e-7,
-               const size_t maximumDepth = 20,
+               const size_t maximumDepth = 0,
                DimensionSelectionType dimensionSelector =
                    DimensionSelectionType());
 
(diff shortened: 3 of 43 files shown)


Yashwants19:
Hi @rcurtin , @zoq : This is ready for review.
Thank you :)

src/mlpack/methods/decision_tree/all_categorical_split.hpp
```diff
@@ -47,6 +47,7 @@ class AllCategoricalSplit
    * @param numClasses Number of classes in the dataset.
    * @param minimumLeafSize Minimum number of points in a leaf node for
    *      splitting.
+   * @param maximumDepth Maximum Depth Minimum for the tree.
```
manish7294: I found this description somewhat doubtful. Maybe a typo?

src/mlpack/tests/random_forest_test.cpp
```diff
@@ -223,7 +223,7 @@ BOOST_AUTO_TEST_CASE(UnweightedCategoricalLearningTest)
 
   // Train a random forest and a decision tree.
   RandomForest<> rf(trainingData, di, trainingLabels, 5, 25 /* 25 trees */, 1,
-      1e-7, MultipleRandomDimensionSelect(4));
+      1e-7, 20, MultipleRandomDimensionSelect(4));
```
manish7294: I think it would be good if we can have a test which checks that the depth of the tree does not exceed actual maximum depth.

yanivbi:
I executed a prediction over an RF model with configured maximum_depth=5. 
The output is:
[INFO ]   input_model_file: shuttle.rf-model-maxD-5.bin
[INFO ]   labels_file: 
**[INFO ]   maximum_depth: 20**
[INFO ]   minimum_gain_split: 0

Is it a print mistake?

src/mlpack/methods/decision_tree/best_binary_numeric_split_impl.hpp
```diff
@@ -25,9 +25,14 @@ double BestBinaryNumericSplit<FitnessFunction>::SplitIfBetter(
     const WeightVecType& weights,
     const size_t minimumLeafSize,
     const double minimumGainSplit,
+    const size_t maximumDepth,
     arma::Col<typename VecType::elem_type>& classProbabilities,
     AuxiliarySplitInfo<typename VecType::elem_type>& /* aux */)
 {
+  // First sanity check: if we have reached maximum depth, we can't split.
+  if (maximumDepth == 1)
+    return DBL_MAX;
```
rcurtin: Honestly I think it would be better to handle this condition at a higher level (in `DecisionTree::Train()` instead).  If we handle it here, then every split policy (including `AllCategoricalSplit` which I don't think you changed) has to have this check in it.  If `maximumDepth == 1` in the regular tree, you can just terminate training and call it a leaf.  We can't do this with `minimumLeafSize`, though, because different split strategies will have different conditions for that (since they may have different numbers of children).

Yashwants19: Now I have just added this for sanity check.

src/mlpack/methods/decision_tree/decision_tree_impl.hpp
```diff
@@ -707,14 +724,15 @@ double DecisionTree<FitnessFunction,
         child->Train<UseWeights>(data, currentChildBegin,
             currentCol - currentChildBegin, datasetInfo, labels, numClasses,
             weights, currentCol - currentChildBegin, minimumGainSplit,
-            dimensionSelector);
+            maximumDepth, dimensionSelector);
```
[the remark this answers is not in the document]
Yashwants19: updated

src/mlpack/methods/decision_tree/all_categorical_split.hpp
Yashwants19: Sorry i will remove this in the next commit.

src/mlpack/tests/random_forest_test.cpp
Yashwants19: I have added the test.

src/mlpack/methods/decision_tree/decision_tree.hpp
```diff
@@ -73,6 +74,7 @@ class DecisionTree :
                const size_t numClasses,
                const size_t minimumLeafSize = 10,
                const double minimumGainSplit = 1e-7,
+               const size_t maximumDepth = 20,
```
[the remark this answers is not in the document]
Yashwants19: Done

src/mlpack/tests/decision_tree_test.cpp
```diff
@@ -582,6 +585,7 @@ BOOST_AUTO_TEST_CASE(PerfectTrainingSetWithWeight)
   weights.ones();
 
   // Minimum leaf size of 1.
+  // Maximum Depth of 1
```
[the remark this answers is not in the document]
Yashwants19: Sorry i will remove this in the next commit.

src/mlpack/methods/decision_tree/best_binary_numeric_split_impl.hpp
rcurtin: Ah, but can we remove it?  I like sanity checks, but the problem with this sanity check is that it requires changing the API of any numeric split policy.  So if anyone wants to implement a new numeric split policy, they now have to handle this `maximumDepth` parameter, which they only need to sanity check.  In my view that's unnecessary overhead and I'd prefer to do the checking at the level of the decision tree only. :+1:

src/mlpack/methods/decision_tree/decision_tree_impl.hpp
```diff
@@ -597,6 +610,7 @@ double DecisionTree<FitnessFunction,
       UseWeights ? weights.subvec(begin, begin + count - 1) : weights);
   size_t bestDim = datasetInfo.Dimensionality(); // This means "no split".
   const size_t end = dimensionSelector.End();
+
   for (size_t i = dimensionSelector.Begin(); i != end;
        i = dimensionSelector.Next())
```
rcurtin: In this function, couldn't we just avoid this part of the code by skipping down to the block of code that calculates class probabilities when `maximumDepth == 1`?  That would make the whole changeset a lot simpler in my opinion, since the only place `maximumDepth` would need to be considered would be in this overload and the non-DatasetInfo overload of `Train<UseWeights>()`.

src/mlpack/tests/decision_tree_test.cpp
```diff
@@ -1213,4 +1217,38 @@ BOOST_AUTO_TEST_CASE(DecisionTreeCategoricalTrainReturnEntropy)
...
+  DecisionTree<> d2(dataset, labels, 2, 10, 1e-7, 0);
+
+  // Now require that we have zero children.
+  BOOST_REQUIRE_EQUAL(d.NumChildren(), 0);
+
+  // Now require that we have zero children.
+  BOOST_REQUIRE_GT(d1.NumChildren(), 0);
+
+  // Now require that we have zero children.
+  BOOST_REQUIRE_GT(d2.NumChildren(), 0);
```
rcurtin: The comment seem inaccurate here.  You might also consider checking deeper in the tree... so for `d1` you might check...

```
BOOST_REQUIRE_EQUAL(d1.NumChildren(), 2);
BOOST_REQUIRE_EQUAL(d1.Child(0).NumChildren(), 0);
BOOST_REQUIRE_EQUAL(d1.Child(0).NumChildren(), 0);
```

and then for `d2`,

```
BOOST_REQUIRE_EQUAL(d2.NumChildren(), 2);
BOOST_REQUIRE_EQUAL(d2.Child(0).NumChildren(), 2);
BOOST_REQUIRE_EQUAL(d2.Child(1).NumChildren(), 2);
```

src/mlpack/tests/main_tests/random_forest_test.cpp
```diff
@@ -186,6 +186,26 @@ BOOST_AUTO_TEST_CASE(RandomForestMinimumLeafSizeTest)
...
+{
+  arma::mat inputData;
+  if (!data::Load("vc2.csv", inputData))
+    BOOST_FAIL("Cannot load train dataset vc2.csv!");
+
+  arma::Row<size_t> labels;
+  if (!data::Load("vc2_labels.txt", labels))
+    BOOST_FAIL("Cannot load labels for vc2_labels.txt");
+
+  SetInputParam("maximum_depth", (int) 0); // Invalid.
```
rcurtin: I think 0 should be valid---that should mean there is no limit on the depth.

Yashwants19: Updated :)

src/mlpack/tests/decision_tree_test.cpp
Yashwants19: I have made the improvement as suggested.

src/mlpack/methods/decision_tree/decision_tree_impl.hpp
Yashwants19: Updated as suggested

Yashwants19:
Hi @rcurtin this ready for review. I have made the changes as suggested.

src/mlpack/tests/decision_tree_test.cpp
```diff
@@ -1213,4 +1216,34 @@ BOOST_AUTO_TEST_CASE(DecisionTreeCategoricalTrainReturnEntropy)
   BOOST_REQUIRE_EQUAL(std::isfinite(entropy), true);
 }
 
+/**
+ * Make sure different Maximum Depth gives different number of childern.
```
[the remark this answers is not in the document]
Yashwants19: Updated :)

src/mlpack/methods/decision_tree/decision_tree_impl.hpp
```diff
@@ -597,127 +610,144 @@ double DecisionTree<FitnessFunction,
...
-    {
-      bestGain = 0.0;
+      // Was there an improvement?  If so mark that it's the new best dimension.
+      bestDim = i;
+      bestGain = dimGain;
+
+      // If the gain is the best possible, no need to keep looking.
+      if (bestGain >= 0.0)
+        break;
     }
```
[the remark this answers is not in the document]
Yashwants19: I have made the suggested changes. :) and also update the `HISTORY.md`

Yashwants19:
Hi @rcurtin I have made the changes as suggested. 
Thank You :)

rcurtin:
@Yashwants19 thanks again.  I made some slight style and spelling fixes in e3989bd4. :+1:

Yashwants19:
Hi @rcurtin now can I create a new PR for the cli-bindings for linear_svm?

rcurtin:
Sure, feel free.  Like I said I'd like to review the Go bindings, but I'm about to be traveling for two weeks so it may be some time until I can get to those.

PR #1917, opened 2019-06-01
bump version to 3.1.1

birm:
A trivial PR, yes. But We probably want the version number to match in the docs.


birm:
Thanks @rcurtin; things are going well, enough for me to at least try to resume contributing, if slowly. I hope things are going well with you too!

PR #1920, opened 2019-06-08
Weight Normalization Layer.

walragatver:
This currently contains the template code of weight norm layer.

### DIFF
--- a/src/mlpack/methods/ann/layer/weight_norm.hpp
+++ b/src/mlpack/methods/ann/layer/weight_norm.hpp
@@ -133,8 +133,8 @@ class WeightNorm
   //! Modify the parameters.
   OutputDataType& Parameters() { return weights; }
 
-  //! Modify the wrapped layer.
-  LayerTypes<CustomLayers...>& Layer() { return wrappedLayer; }
+  //! Get the wrapped layer.
+  LayerTypes<CustomLayers...> const& Layer() { return wrappedLayer; }
 
   /**
    * Serialize the layer.
--- a/src/mlpack/methods/ann/layer/weight_norm.hpp
+++ b/src/mlpack/methods/ann/layer/weight_norm.hpp
@@ -101,7 +101,7 @@ class WeightNorm
                 arma::Mat<eT>&& g);
 
   /**
-   * Calculate the gradient using the output delta and the input activations and
+   * Calculate the gradient using the output delta, input activations and the
    * weights of the wrapped layer.
    *
    * @param input The input activations.
--- a/src/mlpack/methods/ann/layer/weight_norm.hpp
+++ b/src/mlpack/methods/ann/layer/weight_norm.hpp
@@ -84,8 +84,8 @@ class WeightNorm
   void Forward(arma::Mat<eT>&& input, arma::Mat<eT>&& output);
 
   /**
-   * Backward pass through the layer. This function will call the
-   * Backward() function of the wrapped layer.
+   * Backward pass through the layer. This function calls the Backward()
+   * function of the wrapped layer.
    *
    * @param input The input activations.
    * @param gy The backpropagated error.
@@ -200,7 +200,7 @@ class WeightNorm
   //! Locally-stored output parameter visitor module object.
   OutputParameterVisitor outputParameterVisitor;
 
-  // Reset the gradient for all modules that implement the Gradient function.
+  //! Reset the gradient for all modules that implement the Gradient function.
   void ResetGradients(arma::mat& gradient);
 
   //! Locally-stored reset visitor.
--- a/src/mlpack/methods/ann/layer/weight_norm.hpp
+++ b/src/mlpack/methods/ann/layer/weight_norm.hpp
@@ -60,8 +60,12 @@ template <
 class WeightNorm
 {
  public:
-  //! Create the WeightNorm layer object.
-  WeightNorm();
+  /**
+   * Create the WeightNorm layer object.
+   *
+   * @param layer The layer of whoose weights are needed to be normalize.
+   */
+  WeightNorm(LayerTypes<CustomLayers...> layer = LayerTypes<CustomLayers...>());
 
   //! Destructor to release allocated memory.
   ~WeightNorm();
@@ -119,22 +123,6 @@ class WeightNorm
   //! Modify the gradient.
   OutputDataType& Gradient() { return gradient; }
 
-  /*//! Get the input parameter.
-  InputDataType const& InputParameter() const { return inputParameter; }
-  //! Modify the input parameter.
-  InputDataType& InputParameter() { return inputParameter; }*/
-
-  //! Return the model modules.
-  std::vector<LayerTypes<CustomLayers...> >& Model()
-  {
-    if (model)
-    {
-      return network;
-    }
-
-    return empty;
-  }
-
   //! Get the output parameter.
   OutputDataType const& OutputParameter() const { return outputParameter; }
   //! Modify the output parameter.
@@ -145,20 +133,8 @@ class WeightNorm
   //! Modify the parameters.
   OutputDataType& Parameters() { return weights; }
 
-  /*
-   * Add a new module to the model.
-   *
-   * @param args The layer parameter.
-   */
-  template <class LayerType, class... Args>
-  void Add(Args... args);
-
-  /*
-   * Add a new module to the model.
-   *
-   * @param layer The Layer to be added to the model.
-   */
-  void Add(LayerTypes<CustomLayers...> layer);
+  //! Modify the wrapped layer.
+  LayerTypes<CustomLayers...>& Layer() { return wrappedLayer; }
 
   /**
    * Serialize the layer.
@@ -176,23 +152,14 @@ class WeightNorm
   //! Locally-stored delta visitor module object.
   DeltaVisitor deltaVisitor;
 
-  //! Locally-stored empty list of modules.
-  std::vector<LayerTypes<CustomLayers...> > empty;
-
   //! Locally-stored gradient object.
   OutputDataType gradient;
 
-  //! Locally-stored input parameter object.
-  InputDataType inputParameter;
-
-  //! Parameter which indicates if the modules should be exposed.
-  bool model;
-
-  //! Locally-stored network modules.
-  std::vector<LayerTypes<CustomLayers...> > network;
+  //! Locally-stored wrapped layer.
+  LayerTypes<CustomLayers...> wrappedLayer;
 
   //! Locally stored number of elements in the weights of wrapped layer.
-  size_t networkWeightSize;
+  size_t layerWeightSize;
 
   //! Locally-stored output parameter object.
   OutputDataType outputParameter;
--- a/src/mlpack/methods/ann/layer/weight_norm.hpp
+++ b/src/mlpack/methods/ann/layer/weight_norm.hpp
@@ -14,9 +14,13 @@
 
 #include <mlpack/prereqs.hpp>
 #include "layer_types.hpp"
+
 #include "../visitor/delete_visitor.hpp"
 #include "../visitor/delta_visitor.hpp"
 #include "../visitor/output_parameter_visitor.hpp"
+#include "../visitor/reset_visitor.hpp"
+#include "../visitor/weight_size_visitor.hpp"
+#include "../visitor/weight_set_visitor.hpp"
 
 namespace mlpack {
 namespace ann /** Artificial Neural Network. */ {
@@ -77,7 +81,7 @@ class WeightNorm
    * @param output Resulting output activations.
    */
   template<typename eT>
-  void Forward(const arma::Mat<eT>&& input, arma::Mat<eT>&& output);
+  void Forward(arma::Mat<eT>&& input, arma::Mat<eT>&& output);
 
   /**
    * Backward pass through the layer. This function will call the
@@ -101,7 +105,7 @@ class WeightNorm
    * @param gradient The calculated gradient.
    */
   template<typename eT>
-  void Gradient(const arma::Mat<eT>&& input,
+  void Gradient(arma::Mat<eT>&& input,
                 arma::Mat<eT>&& error,
                 arma::Mat<eT>&& gradient);
 
@@ -115,10 +119,10 @@ class WeightNorm
   //! Modify the gradient.
   OutputDataType& Gradient() { return gradient; }
 
-  //! Get the input parameter.
+  /*//! Get the input parameter.
   InputDataType const& InputParameter() const { return inputParameter; }
   //! Modify the input parameter.
-  InputDataType& InputParameter() { return inputParameter; }
+  InputDataType& InputParameter() { return inputParameter; }*/
 
   //! Return the model modules.
   std::vector<LayerTypes<CustomLayers...> >& Model()
@@ -147,14 +151,14 @@ class WeightNorm
    * @param args The layer parameter.
    */
   template <class LayerType, class... Args>
-  void Add(Args... args) { network.push_back(new LayerType(args...)); }
+  void Add(Args... args);
 
   /*
    * Add a new module to the model.
    *
    * @param layer The Layer to be added to the model.
    */
-  void Add(LayerTypes<CustomLayers...> layer) { network.push_back(layer); }
+  void Add(LayerTypes<CustomLayers...> layer);
 
   /**
    * Serialize the layer.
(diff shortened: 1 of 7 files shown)


walragatver:
Hi @zoq, @ShikharJ,
This is ready for review. Just one thing. In the paper, it is mentioned that the `bias term` is `not normalized`. But in order to do so, we will need to introduce a function which will return the `bias` parameter in some of the layers. It looks we will also need a `visitor` for the same. Let me know your thoughts regarding the same. 
Here's the link for the paper. 
https://papers.nips.cc/paper/6114-weight-normalization-a-simple-reparameterization-to-accelerate-training-of-deep-neural-networks.pdf
The note regarding the bias term is in the implementation over here.
https://github.com/TimSalimans/weight_norm/blob/master/nn.py#L267

src/mlpack/methods/ann/layer/weight_norm.hpp
```diff
@@ -0,0 +1,228 @@
...
+  //! Modify the gradient.
+  OutputDataType& Gradient() { return gradient; }
+
+  /*//! Get the input parameter.
+  InputDataType const& InputParameter() const { return inputParameter; }
+  //! Modify the input parameter.
+  InputDataType& InputParameter() { return inputParameter; }*/
+
+  //! Return the model modules.
+  std::vector<LayerTypes<CustomLayers...> >& Model()
```
saksham189: I don't there is need for storing a model and `Add` given that you keep clearing the network. You can pass the single layer directly through the constructor.

walragatver: I thought it would be great to maintain consistency with the API. Also when we pass through the constructor we cannot change the layer inside the `wieghtNorm` wrapper.

saksham189: I believe you can create a getter and setter method for changing the layer. When you use the `Add` method you are not adding anything here which contradicts the purpose.
There is definitely no model or network that you are storing here just a single layer. 
Let me know what you think.

walragatver: Yeah, you are correct. I was earlier thinking about passing from constructor. But later on thought that we may add more than one layers to normalize. But yes I think it would be great to have only one layer.  So I am thinking to go with constructor now. Also I forgot about the getter and setter method as well.
Thanks for looking into it. I will make the changes soon.

saksham189: > But later on thought that we may add more than one layers to normalize. But yes I think it would be great to have only one layer. 

I think this may still be possible using `Sequential` and passing that to the `WeightNorm` constructor.

walragatver: No, Every layer there should be a different `vector and scalar` parameter. So that's not possible using sequential.

walragatver:
@ShikharJ, Thanks for the review. I have made the required changes. I will take care that such silly mistakes do not happen next time.

zoq:
> Hi @zoq, @ShikharJ,
> This is ready for review. Just one thing. In the paper, it is mentioned that the `bias term` is `not normalized`. But in order to do so, we will need to introduce a function which will return the `bias` parameter in some of the layers. It looks we will also need a `visitor` for the same. Let me know your thoughts regarding the same.
> Here's the link for the paper.
> https://papers.nips.cc/paper/6114-weight-normalization-a-simple-reparameterization-to-accelerate-training-of-deep-neural-networks.pdf
> The note regarding the bias term is in the implementation over here.
> https://github.com/TimSalimans/weight_norm/blob/master/nn.py#L267

About the bias term, I wonder if the non-normalisation of the bias term is important. But if that is the case, I agree we could add a method which returns the weights/bias and not the combination of both.

walragatver:
> About the bias term, I wonder if the non-normalisation of the bias term is important. But if that is the case, I agree we could add a method which returns the weights/bias and not the combination of both.

Okay, I will go into more details and will let you know.

src/mlpack/tests/ann_layer_test.cpp
```diff
@@ -2390,6 +2390,126 @@ BOOST_AUTO_TEST_CASE(GradientSequentialLayerTest)
...
+    arma::mat input, target;
+  } function;
+
+  /*
+   * The following is the code similar to CheckGradient() function in
+   * ann_test_tools.hpp.
+   * For weightNorm the weights of wrapped layer are always calculated w.r.t
+   * the scalar and vector parameter before forwarding. Hence, it's gradient
+   * checking only depends on gradients of scalar and vector parameters. Thus,
+   * an offset is required for checking.
```
saksham189: If I understand correctly, then you do not need the gradient for the weights of `wrappedLayer` but only the gradient for the parameters for the weight norm layer.  Could you store the weights of the wrapped layer separately in `wrappedLayerWeights` and avoid calling `Gradient` for the wrappedLayer and computing gradient for its weights.

walragatver: No that's quite incorrect.
I need to calculate gradients of the weight's of the `wrappedLayer` in order to calculate the `gradients` of the parameter of the weight norm layer. Just look into `section 2.1`  in the below link.


> Here's the link for the paper.
> https://papers.nips.cc/paper/6114-weight-normalization-a-simple-reparameterization-to-accelerate-training-of-deep-neural-networks.pdf

saksham189: Yes, I see. But could you store the weights separately in `wrappedLayerWeights` because the `optimizer` would not need to update the weights of the wrapped layer and it should not because there is no use.

walragatver: Yes, you are quite correct about updating the weights. 
But if we need to keep `bias term` separate then we will need to update it using the `optimizer`. 
Hence, I am currently thinking to make the required changes once I finalize whether to keep the `bias` term unaffected.
Thank for the review. :+1:

walragatver:
Hi this is complete now. I have used the visitor in #1956 to address previous issues.

src/mlpack/methods/ann/visitor/bias_set_visitor.hpp
```diff
@@ -0,0 +1,80 @@
...
+
+#include <mlpack/methods/ann/layer/layer_traits.hpp>
+
+#include <boost/variant.hpp>
+
+namespace mlpack {
+namespace ann {
+
+/**
+ * BiasSetVisitor update the module bias parameters given the parameters set.
```
ShikharJ: It should be like `updates the module's bias parameters, given the parameters' set.`.

src/mlpack/methods/ann/visitor/bias_set_visitor.hpp
```diff
@@ -0,0 +1,80 @@
...
+ */
+class BiasSetVisitor : public boost::static_visitor<size_t>
+{
+ public:
+  //! Update the bias parameters given the parameters set and offset.
```
ShikharJ: It should be `parameters' set` here.

src/mlpack/methods/ann/visitor/bias_set_visitor.hpp
```diff
@@ -0,0 +1,80 @@
...
+  BiasSetVisitor(arma::mat&& weight, const size_t offset = 0);
+
+  //! Update the parameters set.
```
ShikharJ: Same here.

src/mlpack/methods/ann/visitor/bias_set_visitor.hpp
```diff
@@ -0,0 +1,80 @@
...
+  template<typename LayerType>
+  size_t operator()(LayerType* layer) const;
+
+ private:
+  //! The parameters set.
```
ShikharJ: And here.

src/mlpack/methods/ann/visitor/bias_set_visitor.hpp
```diff
@@ -0,0 +1,80 @@
...
+  arma::mat&& weight;
+
+  //! The parameters offset.
```
ShikharJ: Same here as well.

src/mlpack/methods/ann/visitor/bias_set_visitor_impl.hpp
```diff
@@ -0,0 +1,94 @@
...
+}
+
+template<typename T>
+inline typename std::enable_if<
+    !HasBiasCheck<T, arma::mat&(T::*)()>::value &&
+    HasModelCheck<T>::value, size_t>::type
+BiasSetVisitor::LayerSize(T* layer) const
+{
+  size_t modelOffset = 0;
+  for (size_t i = 0; i < layer->Model().size(); ++i)
```
ShikharJ: Could you add a newline above this for loop for aesthetics please?

src/mlpack/methods/ann/visitor/bias_set_visitor_impl.hpp
```diff
@@ -0,0 +1,94 @@
...
+inline typename std::enable_if<
+    HasBiasCheck<T, arma::mat&(T::*)()>::value &&
+    HasModelCheck<T>::value, size_t>::type
+BiasSetVisitor::LayerSize(T* layer) const
+{
+  layer->Bias() = arma::mat(weight.memptr() + offset,
+      layer->Bias().n_rows, layer->Bias().n_cols, false, false);
+
+  size_t modelOffset = layer->Bias().n_elem;
+  for (size_t i = 0; i < layer->Model().size(); ++i)
```
ShikharJ: Same here.

src/mlpack/methods/ann/layer/weight_norm.hpp
```diff
@@ -0,0 +1,204 @@
...
+   * @param g The calculated gradient.
+   */
+  template<typename eT>
+  void Backward(const arma::Mat<eT>&& input,
+                arma::Mat<eT>&& gy,
+                arma::Mat<eT>&& g);
+
+  /**
+   * Calculate the gradient using the output delta and the input activations and
+   * weights of the wrapped layer.
```
ShikharJ: This should be like `... using the output delta, input activations and the weights ... `

src/mlpack/methods/ann/layer/weight_norm.hpp
```diff
@@ -0,0 +1,204 @@
...
+  size_t layerWeightSize;
+
+  //! Locally-stored output parameter object.
+  OutputDataType outputParameter;
+
+  //! Locally-stored output parameter visitor module object.
+  OutputParameterVisitor outputParameterVisitor;
+
+  //! Reset the gradient for all modules that implement the Gradient function.
+  void ResetGradients(arma::mat& gradient);
```
ShikharJ: Why is this function made private?

src/mlpack/methods/ann/layer/weight_norm_impl.hpp
```diff
@@ -0,0 +1,168 @@
...
+      false);
+}
+
+template<typename InputDataType, typename OutputDataType,
+         typename... CustomLayers>
+template<typename eT>
+void WeightNorm<InputDataType, OutputDataType, CustomLayers...>::Forward(
+    arma::Mat<eT>&& input, arma::Mat<eT>&& output)
+{
+  // Intialize the non-bias weights of wrapped layer.
```
ShikharJ: This should be `Initialize` here.

src/mlpack/methods/ann/layer/weight_norm.hpp
walragatver: I have made this private taking `line 368` in `ffn.hpp` as a reference. It is also made private so that it is `not accessible directly`.

walragatver:
Hi @ShikharJ, I have made the required changes.

src/mlpack/tests/ann_layer_test.cpp
```diff
@@ -2536,6 +2536,80 @@ BOOST_AUTO_TEST_CASE(GradientSequentialLayerTest)
...
+    GradientFunction()
+    {
+      input = arma::randu(10, 1);
+      target = arma::mat("1");
+
+      model = new FFN<NegativeLogLikelihood<>, NguyenWidrowInitialization>();
+      model->Predictors() = input;
+      model->Responses() = target;
+      model->Add<IdentityLayer<> >();
+
```
saksham189: Could you add a linear layer before the weight norm layer so, that we can check the `Backward` function.

src/mlpack/methods/ann/layer/weight_norm_impl.hpp
```diff
@@ -0,0 +1,168 @@
...
+    boost::apply_visitor(deleteVisitor, wrappedLayer);
+  }
+
+  ar & BOOST_SERIALIZATION_NVP(wrappedLayer);
+  ar & BOOST_SERIALIZATION_NVP(layerWeightSize);
+
+  // If we are loading, we need to initialize the weights.
+  if (Archive::is_loading::value)
+  {
+    weights.set_size(2 * layerWeightSize + 1, 1);
```
saksham189: Here you set a different size as compared to here:
https://github.com/mlpack/mlpack/pull/1920/files#diff-8627a19cad1a77c6ce8149cf66884073R34

src/mlpack/tests/ann_layer_test.cpp
walragatver: Done.

src/mlpack/methods/ann/layer/weight_norm_impl.hpp
walragatver: Thanks done.

ShikharJ:
@walragatver Can you update this to the latest master please?

walragatver:
@ShikharJ, Done.

src/mlpack/methods/ann/layer/weight_norm_impl.hpp
```diff
@@ -0,0 +1,168 @@
...
+}
+
+template<typename InputDataType, typename OutputDataType,
+         typename... CustomLayers>
+template<typename eT>
+void WeightNorm<InputDataType, OutputDataType, CustomLayers...>::Forward(
+    arma::Mat<eT>&& input, arma::Mat<eT>&& output)
+{
+  // Initialize the non-bias weights of wrapped layer.
+  double normVectorParameter = arma::norm(vectorParameter, 2);
```
zoq: We could make this ``const``.

walragatver: Done

src/mlpack/methods/ann/layer/weight_norm.hpp
```diff
@@ -0,0 +1,204 @@
...
+
+  //! Locally-stored weight size visitor.
+  WeightSizeVisitor weightSizeVisitor;
+
+  //! Locally-stored gradients of wrappedLayer.
+  OutputDataType layerGradients;
+
+  //! Locally-stored weights of wrappedLayer.
+  OutputDataType layerWeights;
+}; // class BatchNorm
```
saksham189: `WeightNorm` here

src/mlpack/tests/ann_layer_test.cpp
```diff
@@ -1613,7 +1613,7 @@ BOOST_AUTO_TEST_CASE(GradientBatchNormTest)
 /**
  * VirtualBatchNorm layer numerical gradient test.
  */
-BOOST_AUTO_TEST_CASE(GradientVirtualBatchNormTest)
+/*BOOST_AUTO_TEST_CASE(GradientVirtualBatchNormTest)
```
saksham189: Is there a need to comment out this test?

src/mlpack/methods/ann/layer/weight_norm_impl.hpp
```diff
@@ -0,0 +1,168 @@
...
+      layerWeightSize - biasWeightSize - 1) - gradient[gradient.n_rows - 1] /
+      normVectorParameter * vectorParameter);
+}
+
+template<typename InputDataType, typename OutputDataType,
+         typename... CustomLayers>
+void WeightNorm<InputDataType, OutputDataType, CustomLayers...>::ResetGradients(
+    arma::mat& gradient)
+{
+    boost::apply_visitor(GradientSetVisitor(std::move(gradient), 0),
```
saksham189: remove extra indent

src/mlpack/tests/ann_layer_test.cpp
ShikharJ: Not after we have merged the fix for more than 50 layer types.

walragatver:
While reviewing #1985, I realized that I should also not allow modifying the `wrappedLayer` so I have fixed it.

ShikharJ:
Thanks @walragatver for the contribution.

PR #1922, opened 2019-06-13
Resolve calculation of accuracy

Yashwants19:


rcurtin:
Can you add a description of what the problem is and how you're fixing it to the PR please?  :+1:

Yashwants19:
Hi @rcurtin previously when I was enabling -p for getting prediction and also at same time -L for setting test_labels than I was getting Nan at the place of accuracy. May be because of std::move(predictions)

Yashwants19:
Sorry Earlier I didn't provide the description.

rcurtin:
Nice fix, thanks!  I updated `HISTORY.md` during merge in 7824ae057.

PR #1927, opened 2019-06-17
Support for Bag of words Functionality

jeffin143:
Implemented Bag of Words Encoding, Keeping Dictionary Encoding Template as base

### DIFF
--- a/src/mlpack/core/data/bow_encoding.hpp
+++ b/src/mlpack/core/data/bow_encoding.hpp
@@ -1,0 +1,128 @@
+/**
+ * @file bow.hpp
+ * @author Jeffin Sam
+ *
+ * Definition of Bag of Words encoding functions.
+ *
+ * mlpack is free software; you may redistribute it and/or modify it under the
+ * terms of the 3-clause BSD license.  You should have received a copy of the
+ * 3-clause BSD license along with mlpack.  If not, see
+ * http://www.opensource.org/licenses/BSD-3-Clause for more information.
+ */
+#ifndef MLPACK_CORE_DATA_BOW_ENCODING_HPP
+#define MLPACK_CORE_DATA_BOW_ENCODING_HPP
+
+#include <mlpack/prereqs.hpp>
+#include "mlpack/core/boost_backport/boost_backport_string_view.hpp"
+#include <utility>
+
+namespace mlpack {
+namespace data {
+
+/**
+ * A simple Bow Enocding class
+ */
+class Bow
+{
+ public:
+  /**
+  * A function to create mapping from a given corpus.
+  * 
+  * @param input Corpus of text to encode.
+  * @param tokenizer A function that accepts a boost::string_view as
+  *                  an argument and returns a token.
+  * This can either be a function pointer or function object or a lamda
+  * function. Its return value should be a boost::string_view, a token.
+  *
+  * Definiation of function should be of type
+  * boost::string_view fn()(boost::string_view& str)
+  *
+  */
+  template<typename TokenizerType>
+  void CreateMap(std::string& input, TokenizerType tokenizer);
+  /*
+  * Default Constructor
+  */
+  Bow() {}
+  /*
+  * Copy Constructor.
+  */
+  Bow(const Bow& oldObject);
+  /*
+  * Move Constructor.
+  */
+  Bow(Bow&& oldObject) = default;
+  /*
+  * Move Assignment Operator.
+  */  
+  Bow& operator= (Bow&& oldObject) = default;
+  /*
+  * Assignment Operator.
+  */
+  void operator= (const Bow& oldObject);
+  /**
+  * A function to reset the mapping that is clear all the encodings
+  */
+  void Reset();
+
+  /**
+  * A fucntion to encode given array of strings using a particular delimiter,
+  * providing custom rule for tokenization.
+  *
+  * For example 
+  * Vector is :
+  * [hello@wow, wow@hello@good] would be encoded using '@' as delimiter as 
+  * [1 2 0, 2 1 3] 
+  * The function paddes 0 to maintain same sizes across all the rows
+  * User may also provide their custom tokenization rule.
+  *
+  * @param input Vector of strings.
+  * @param output Output Matrix to store encoded results (sp_mat or mat).
+  * @param tokenizer A function that accepts a boost::string_view as
+  *                  an argument and returns a token.
+  * This can either be a function pointer or function object or a lamda
+  * function. Its return value should be a boost::string_view, a token.
+  *
+  * Definiation of function should be of type
+  * boost::string_view fn()(boost::string_view& str)
+  *
+  */
+  template<typename MatType, typename TokenizerType>
+  void Encode(const std::vector<std::string>& input,
+              MatType& output, TokenizerType tokenizer);
+
+  //! Modify the originalStrings.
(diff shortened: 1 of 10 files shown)


src/mlpack/core/data/bow_encoding.hpp
```diff
@@ -0,0 +1,128 @@
...
+  */
+  Bow(Bow&& oldObject) = default;
+  /*
+  * Move Assignment Operator.
+  */  
+  Bow& operator= (Bow&& oldObject) = default;
+  /*
+  * Assignment Operator.
+  */
+  void operator= (const Bow& oldObject);
```
lozhnikov: I guess the return value should be `Bow&`.

src/mlpack/core/data/bow_encoding.hpp
```diff
@@ -0,0 +1,128 @@
...
+  *                  an argument and returns a token.
+  * This can either be a function pointer or function object or a lamda
+  * function. Its return value should be a boost::string_view, a token.
+  *
+  * Definiation of function should be of type
+  * boost::string_view fn()(boost::string_view& str)
+  *
+  */
+  template<typename TokenizerType>
+  void CreateMap(std::string& input, TokenizerType tokenizer);
```
lozhnikov: Looks like the first argument should be a const reference.

src/mlpack/core/data/bow_encoding_impl.hpp
```diff
@@ -0,0 +1,138 @@
...
+
+void Bow::Reset()
+{
+  mappings.clear();
+  originalStrings.clear();
+}
+
+Bow::Bow(const Bow& oldObject)
+{
+  originalStrings = oldObject.originalStrings;
```
lozhnikov: I think it's better to move this line to the initializer list.

src/mlpack/core/data/bow_encoding_impl.hpp
```diff
@@ -0,0 +1,138 @@
...
+      }
+      token = tokenizer(strView);
+    }
+  }
+  output.zeros(dataset.size(), mappings.size());
+  for (size_t i = 0; i < dataset.size(); ++i)
+  {
+    for (size_t j = 0; j < dataset[i].size(); ++j)
+    {
+      output.at(i, mappings.at(dataset[i][j])) = 1;
```
lozhnikov: Looks like this is the only difference from `DictionaryEncoding` (i.e. this very line). Maybe it's a good idea to merge this classes, how do you think? I mean we can introduce some `EncodingPolicy` which will handle the encoding process.

src/mlpack/core/data/bow_encoding_impl.hpp
jeffin143: Not sure, what does this mean ?

src/mlpack/core/data/bow_encoding_impl.hpp
jeffin143: Yes :)

src/mlpack/core/data/bow_encoding.hpp
jeffin143: Done :)

src/mlpack/core/data/bow_encoding.hpp
jeffin143: Umm, not sure, but i did make the change :)

lozhnikov: It is possible to write
```
int a, b;
a = b = 0;
```

src/mlpack/core/data/bow_encoding_impl.hpp
lozhnikov: I mean
```
Bow::Bow(const Bow& oldObject) :
    originalStrings(oldObject.originalStrings)
{
}
```

src/mlpack/core/data/bow_encoding.hpp
```diff
@@ -0,0 +1,128 @@
...
+#include "mlpack/core/boost_backport/boost_backport_string_view.hpp"
+#include <utility>
+
+namespace mlpack {
+namespace data {
+
+/**
+ * A simple Bow Enocding class
+ */
+class Bow
```
zoq: Is this short for Bag of Words? If so, maybe we should just call the class ``BagOfWords``?

src/mlpack/tests/string_utility_test.cpp
```diff
@@ -0,0 +1,65 @@
...
+#include <boost/test/unit_test.hpp>
+#include "test_tools.hpp"
+
+using namespace mlpack;
+using namespace mlpack::data;
+using namespace std;
+
+BOOST_AUTO_TEST_SUITE(StringUtilityTest);
+
+boost::string_view tokenizer(boost::string_view& str)
```
zoq: Cool, I think we can also just pass this as a lambda function?

jeffin143: This function is already there in other PR, so when it will be rebased , it woudn't make much difference, to reduce line, I just used same function here.

Still if you want, we can go with lamda function. I have used lamda function there, in other PR.

src/mlpack/core/data/bow_encoding.hpp
jeffin143: Yes, Done.

src/mlpack/core/data/bow_encoding_impl.hpp
jeffin143: Thanks :), Done.

src/mlpack/tests/string_utility_test.cpp
zoq: Fine with me.

src/mlpack/core/data/bow_encoding.hpp
```diff
@@ -0,0 +1,128 @@
...
+
+#include <mlpack/prereqs.hpp>
+#include "mlpack/core/boost_backport/boost_backport_string_view.hpp"
+#include <utility>
+
+namespace mlpack {
+namespace data {
+
+/**
+ * A simple Bow Enocding class
```
zoq: perhaps it makes sense to include a simple example here as well, something like:


https://github.com/mlpack/mlpack/blob/897b0cddf6ba23733f701b4679fac08f9f90ebb2/src/mlpack/core/data/binarize.hpp#L26-L34

src/mlpack/core/data/bow_encoding.hpp
```diff
@@ -0,0 +1,128 @@
...
+
+  //! Modify the originalStrings.
+  std::deque<std::string>& OriginalStrings() { return originalStrings; }
+
+  //! Return the originalStrings.
+  const std::deque<std::string>& OriginalStrings() const
+      { return originalStrings; }
+
+  //! Return the Mappings
+  const std::unordered_map<boost::string_view, size_t,
```
zoq: Do you think we should create an alias for the type to shorten this up?

lozhnikov:
@zoq Thanks for the review. Actually we decided to introduce some policy (e.g. some template argument `EncodingPolicy`) in order to combine this PR with the dictionary encoding PR since the only difference is bow_encoding_impl.hpp:105 i.e. the only line. So, I guess we needn't review this PR. It's better to finish with the dictionary encoding PR.

jeffin143:
@lozhnikov , can we reopen this? I will commit the changes today .

Issue #1923, opened 2019-06-14
Backward() implemented incorrectly in BatchNorm Layer.


saksham189:
#### Issue description

Many of the layers use the numerical gradient test to verify the implementation of the `Gradient` function. However, the test does not verify the implementation of `Backward` function because there is no layer before the layer that is being tested.
For example the model used in `GradientBatchNormTest` is:
```
model = new FFN<NegativeLogLikelihood<>, NguyenWidrowInitialization>();
model->Predictors() = input;
model->Responses() = target;
model->Add<IdentityLayer<> >();
model->Add<BatchNorm<> >(10);
model->Add<Linear<> >(10, 2);
model->Add<LogSoftMax<> >();
```
The test passes with this model. However if the model is changed by adding a linear layer before the `BatchNorm` layer, the test will start failing:

```
model = new FFN<NegativeLogLikelihood<>, NguyenWidrowInitialization>();
model->Predictors() = input;
model->Responses() = target;
model->Add<IdentityLayer<> >();
model->Add<Linear<> >(10, 10);
model->Add<BatchNorm<> >(10);
model->Add<Linear<> >(10, 2);
model->Add<LogSoftMax<> >();
```

#### Steps to reproduce

Change the model by adding a linear layer before `BatchNorm` and run the `GradientBatchNormTest` test.

#### Expected behavior

The test should not fail if the `Backward` function had been implemented correctly.

#### Actual behavior

The numerical gradient test fails and the `Backward` function is implemented incorrectly.

Most of the numerical gradient tests do not have a parametric layer being the layer being tested. So the implementation of the `Backward` function is never really tested.

saksham189:
In the following line in `Backward` for BatchNorm layer:
`const arma::mat inputMean = input.each_col() - mean;`
Here input parameter of the function is actually the output of the layer. We should be using the actual input to the layer to fix the implementation.

PR #1930, opened 2019-06-17
Add linear layer in model for numerical gradient tests

saksham189:
I have added a linear layer for all numerical gradient tests behind the layer being tested.
This PR also include a fix in the implementation of `Backward` for BatchNorm layer. 

Closes https://github.com/mlpack/mlpack/issues/1923

### DIFF
--- a/src/mlpack/methods/ann/layer/batch_norm_impl.hpp
+++ b/src/mlpack/methods/ann/layer/batch_norm_impl.hpp
@@ -22,11 +22,11 @@ namespace ann { /** Artificial Neural Network. */
 
 template<typename InputDataType, typename OutputDataType>
 BatchNorm<InputDataType, OutputDataType>::BatchNorm() :
+    size(0),
     eps(1e-8),
     loading(false),
     deterministic(false),
-    count(0),
-    size(0)
+    count(0)
 {
   // Nothing to do here.
 }
--- a/src/mlpack/methods/ann/layer/batch_norm_impl.hpp
+++ b/src/mlpack/methods/ann/layer/batch_norm_impl.hpp
@@ -80,6 +80,7 @@ void BatchNorm<InputDataType, OutputDataType>::Forward(
 
     // Normalize the input.
     output = input.each_col() - mean;
+    inputMean = output;
     output.each_col() /= arma::sqrt(variance + eps);
 
     // Use Welford method to compute the sample variance and mean.
@@ -87,9 +88,9 @@ void BatchNorm<InputDataType, OutputDataType>::Forward(
     {
       count += 1;
 
-      OutputDataType delta = input.col(i) - runningMean;
-      runningMean = runningMean + delta / count;
-      runningVariance += delta % (input.col(i) - runningMean);
+      OutputDataType diff = input.col(i) - runningMean;
+      runningMean = runningMean + diff / count;
+      runningVariance += diff % (input.col(i) - runningMean);
     }
 
     // Reused in the backward and gradient step.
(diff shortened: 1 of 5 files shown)


saksham189:
The `LayerNorm` gradient test was failing. I added the same fix as for `BatchNorm` and it seems to pass after that.

saksham189:
Seems all the numerical gradient tests are passing after adding the linear layer.
@ShikharJ  let me know your thoughts.

ShikharJ:
> The `LayerNorm` gradient test was failing. I added the same fix as for `BatchNorm` and it seems to pass after that.

That's probably because `LayerNorm` directly borrows from `BatchNorm`.

src/mlpack/methods/ann/layer/batch_norm_impl.hpp
```diff
@@ -25,7 +25,8 @@ BatchNorm<InputDataType, OutputDataType>::BatchNorm() :
     eps(1e-8),
     loading(false),
     deterministic(false),
-    count(0)
+    count(0),
+    size(0)
```
ShikharJ: This should be introduced in the same order as the other constructor to prevent compile time warnings.

src/mlpack/methods/ann/layer/batch_norm_impl.hpp
```diff
@@ -80,16 +81,17 @@ void BatchNorm<InputDataType, OutputDataType>::Forward(
...
     for (size_t i = 0; i < input.n_cols; i++)
     {
       count += 1;
 
-      OutputDataType delta = input.col(i) - runningMean;
-      runningMean = runningMean + delta / count;
-      runningVariance += delta % (input.col(i) - runningMean);
+      OutputDataType diff = input.col(i) - runningMean;
+      runningMean = runningMean + diff / count;
+      runningVariance += diff % (input.col(i) - runningMean);
```
ShikharJ: I think we should revert these changes, since they only change the name.

src/mlpack/methods/ann/layer/batch_norm_impl.hpp
```diff
@@ -106,7 +108,6 @@ template<typename eT>
 void BatchNorm<InputDataType, OutputDataType>::Backward(
     const arma::Mat<eT>&& input, arma::Mat<eT>&& gy, arma::Mat<eT>&& g)
 {
-  const arma::mat inputMean = input.each_col() - mean;
```
ShikharJ: I see, we were not using the processed output. I remember facing such an issue in a part implementation of `ELU` as well. Nice catch.

src/mlpack/methods/ann/layer/layer_norm.hpp
```diff
@@ -184,6 +184,9 @@ class LayerNorm
 
   //! Locally-stored normalized input.
   OutputDataType normalized;
+
+  //! Locally-stored input with 0 mean.
```
ShikharJ: Same here.

src/mlpack/methods/ann/layer/layer_norm_impl.hpp
```diff
@@ -23,7 +23,8 @@ namespace ann { /** Artificial Neural Network. */
 template<typename InputDataType, typename OutputDataType>
 LayerNorm<InputDataType, OutputDataType>::LayerNorm() :
     eps(1e-8),
-    loading(false)
+    loading(false),
+    size(0)
```
ShikharJ: Same here.

src/mlpack/methods/ann/layer/batch_norm_impl.hpp
saksham189: I did this change because `delta` is also a class member and hence it is shadowing the class variable. IMO it is best to keep the names separate to avoid mistakes when reading and modifying the code.

ShikharJ:
Thanks @saksham189 for the fix.

PR #1931, opened 2019-06-18
Change the pendulum action type to double

robotcator:
As said in [pr](https://github.com/mlpack/mlpack/pull/1912), remove the [1] in Pendulum's action definition.

### DIFF
--- a/src/mlpack/methods/reinforcement_learning/environment/pendulum.hpp
+++ b/src/mlpack/methods/reinforcement_learning/environment/pendulum.hpp
@@ -82,7 +82,7 @@ class Pendulum
    */
   struct Action
   {
-    double action;
+    double action = 0.0;
     // Storing degree of freedom
     const int size = 1;
   };
--- a/src/mlpack/methods/reinforcement_learning/environment/pendulum.hpp
+++ b/src/mlpack/methods/reinforcement_learning/environment/pendulum.hpp
@@ -82,7 +82,7 @@ class Pendulum
    */
   struct Action
   {
-    double action[1];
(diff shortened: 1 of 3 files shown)


src/mlpack/methods/reinforcement_learning/environment/pendulum.hpp
```diff
@@ -82,7 +82,7 @@ class Pendulum
    */
   struct Action
   {
-    double action[1];
+    double action;
```
zoq: Do you think we should use an enum instead of a struct here as well?

robotcator: > Do you think we should use an enum instead of a struct here as well?

For my understanding, the pendulum is a continuous task, so the action is `double` or `float` type. While other tasks, like `cart_pole`,  the action is discrete. I think the Pendulum can keep using a struct to define the action. What do you think of it?

zoq: Sure, in this case we should set a default value, so that we pass the static analysis check.

zoq: Sure, in this case we should set a default value, so that we pass the static analysis check.

robotcator: > Sure, in this case we should set a default value, so that we pass the static analysis check.

Yes, I am fixing this error.

zoq:
Thanks, that was an easy merge.

PR #1934, opened 2019-06-20
[GSoC] Implement probabilistic KDE error bounds

robertohueso:
Inspired by [this paper](http://papers.nips.cc/paper/3539-fast-high-dimensional-kernel-summations-using-the-monte-carlo-multipole-method.pdf) and corrections made by @rcurtin, this PR adds support for probabilistic relative error bounds when using Gaussian kernel into the existing KDE codebase.


src/mlpack/methods/kde/kde.hpp
```diff
@@ -75,7 +75,12 @@ class KDE
       const double absError = 0,
       KernelType kernel = KernelType(),
       const KDEMode mode = DUAL_TREE_MODE,
-      MetricType metric = MetricType());
+      MetricType metric = MetricType(),
+      const bool monteCarlo = false,
+      const double MCProb = 0.95,
+      const size_t initialSampleSize = 100,
+      const double MCEntryCoef = 3,
+      const double MCBreakCoef = 0.7);
```
rcurtin: We should document these parameters in the comment, and also instead of `MCProb`, to match the style guide, we should do, e.g., `mcProb` or `monteCarloProb`; whichever you like is fine.

src/mlpack/methods/kde/kde.hpp
```diff
@@ -214,6 +219,37 @@ class KDE
...
+  size_t& MCInitialSampleSize() { return initialSampleSize; }
+
+  //! Get Monte Carlo entry coefficient.
+  double MCEntryCoefficient() const { return MCEntryCoef; }
+
+  //! Modify Monte Carlo entry coefficient. (newCoef >= 1).
+  void MCEntryCoefficient(const double newCoef);
+
+  //! Get Monte Carlo break coefficient.
+  double MCBreakCoefficient() const { return MCBreakCoef; }
```
rcurtin: Once you change the names of the parameters, then the names of the functions here can exactly match the name of the variable (i.e. `MCBreakCoef()` for `mcBreakCoef`).

src/mlpack/methods/kde/kde_impl.hpp
```diff
@@ -569,6 +681,11 @@ serialize(Archive& ar, const unsigned int /* version */)
   ar & BOOST_SERIALIZATION_NVP(absError);
   ar & BOOST_SERIALIZATION_NVP(trained);
   ar & BOOST_SERIALIZATION_NVP(mode);
+  ar & BOOST_SERIALIZATION_NVP(monteCarlo);
+  ar & BOOST_SERIALIZATION_NVP(MCProb);
+  ar & BOOST_SERIALIZATION_NVP(initialSampleSize);
+  ar & BOOST_SERIALIZATION_NVP(MCEntryCoef);
+  ar & BOOST_SERIALIZATION_NVP(MCBreakCoef);
```
rcurtin: Technically, this isn't reverse-compatible---KDE models made with earlier versions of mlpack won't have these members, so we should bump the Boost serialization version with a block like

```
//! Set the serialization version of the LSHSearch class.
BOOST_TEMPLATE_CLASS_VERSION(template<typename ...>,
    mlpack::kde::KDE<...>, 1);
```

Then, in `serialize()`, you can use the `version` parameter to detect if this is a version 0 archive (which will not have the Monte Carlo parameters) or a version 1 or newer archive (in which case it will).  You can take a look at `src/mlpack/methods/lsh/lsh_search_impl.hpp` for an example of how this might work. :+1:

src/mlpack/methods/kde/kde_main.cpp
```diff
@@ -48,7 +48,21 @@ PROGRAM_INFO("Kernel Density Estimation",
     "type of tree to use for the dual-tree algorithm with " +
     PRINT_PARAM_STRING("tree") + ". It is also possible to select whether to "
     "use dual-tree algorithm or single-tree algorithm using the " +
-    PRINT_PARAM_STRING("algorithm") + " option."
+    PRINT_PARAM_STRING("algorithm") + " option. This implementation makes "
```
rcurtin: Given how long the section you added is, it may make sense to split the Monte Carlo discussion into a separate paragraph.

src/mlpack/methods/kde/kde_main.cpp
```diff
@@ -68,7 +82,23 @@ PROGRAM_INFO("Kernel Density Estimation",
...
+    "\n\n"+
+    PRINT_CALL("kde", "reference", "ref_data", "query", "qu_data", "bandwidth",
+        0.2, "kernel", "gaussian", "tree", "kd-tree", "rel_error",
+        0.05, "predictions", "out_data", "monte_carlo", "", "mc_probability",
+        0.95, "initial_sample_size", 200, "mc_entry_coef", 3.5, "mc_break_coef",
+        0.6) +
+    "\n\n",
     SEE_ALSO("@knn", "#knn"),
     SEE_ALSO("Kernel density estimation on Wikipedia",
         "https://en.wikipedia.org/wiki/Kernel_density_estimation"),
```
rcurtin: Maybe it's worth adding a `SEE_ALSO()` for Dongryeol's paper too?

src/mlpack/methods/kde/kde_main.cpp
```diff
@@ -108,6 +138,30 @@ PARAM_DOUBLE_IN("abs_error",
...
+                "P",
+                0.95);
+PARAM_INT_IN("initial_sample_size",
+             "Initial sample size for Monte Carlo estimations.",
+             "s",
+             100);
+PARAM_DOUBLE_IN("mc_entry_coef",
+                "Controls how much larger does the amount of node descendants "
+                "has to be compared to the " +
+                PRINT_PARAM_STRING("initial_sample_size")+ " in order to be "
```
rcurtin: Tiny style issue, there should be a space before the `+`. :)

src/mlpack/methods/kde/kde_main.cpp
```diff
@@ -134,6 +193,22 @@ static void mlpackMain()
   ReportIgnoredParam({{ "input_model", true }}, "kernel");
   ReportIgnoredParam({{ "input_model", true }}, "rel_error");
   ReportIgnoredParam({{ "input_model", true }}, "abs_error");
+  ReportIgnoredParam({{ "input_model", true }}, "monte_carlo");
+  ReportIgnoredParam({{ "input_model", true }}, "mc_probability");
+  ReportIgnoredParam({{ "input_model", true }}, "initial_sample_size");
+  ReportIgnoredParam({{ "input_model", true }}, "mc_entry_coef");
+  ReportIgnoredParam({{ "input_model", true }}, "mc_break_coef");
```
rcurtin: Hmmm, should we have the user always specify `--monte_carlo` if they want to do Monte Carlo sampling for search?  Imagine this possibility: a user builds a KDE model with `--monte_carlo` specified, but now they want to run a search with it turned *off*.  I think the design now doesn't support that, but do you think it would be reasonable to change it so that that could be done?  Technically the model only really consists of the tree and its parameters, so at query time it should be feasible to change whether Monte Carlo sampling is done or not.

(That should also be true for `rel_error` and `abs_error`, I guess.  I suppose I missed that when I reviewed the original KDE PR. :))

src/mlpack/methods/kde/kde_model_impl.hpp
```diff
@@ -214,6 +244,26 @@ inline void KDEModel::BuildModel(arma::mat&& referenceSet)
...
+  MCSampleSizeVisitor sampleSizeVisitor(initialSampleSize);
+  boost::apply_visitor(sampleSizeVisitor, kdeModel);
+
+  // Set Monte Carlo entry coefficient.
+  MCEntryCoefVisitor entryCoefficientVisitor(MCEntryCoef);
+  boost::apply_visitor(entryCoefficientVisitor, kdeModel);
+
+  // Set Monte Carlo break coefficient.
+  MCBreakCoefVisitor breakCoefficientVisitor(MCBreakCoef);
+  boost::apply_visitor(breakCoefficientVisitor, kdeModel);
```
rcurtin: Since all these are set together, you might consider writing just one visitor that sets all of the Monte Carlo parameters.  Up to you. :+1:

src/mlpack/methods/kde/kde_model_impl.hpp
```diff
@@ -352,6 +492,11 @@ void KDEModel::serialize(Archive& ar, const unsigned int /* version */)
   ar & BOOST_SERIALIZATION_NVP(absError);
   ar & BOOST_SERIALIZATION_NVP(kernelType);
   ar & BOOST_SERIALIZATION_NVP(treeType);
+  ar & BOOST_SERIALIZATION_NVP(monteCarlo);
+  ar & BOOST_SERIALIZATION_NVP(MCProb);
+  ar & BOOST_SERIALIZATION_NVP(initialSampleSize);
+  ar & BOOST_SERIALIZATION_NVP(MCEntryCoef);
+  ar & BOOST_SERIALIZATION_NVP(MCBreakCoef);
```
rcurtin: Is it necessary to serialize these, if they are already held in the `KDE` object?

src/mlpack/methods/kde/kde_rules.hpp
```diff
@@ -25,6 +25,10 @@ namespace kde {
 template<typename MetricType, typename KernelType, typename TreeType>
 class KDERules
 {
+ private:
+  // SFINAE Check if kernel has bandwidth function.
+  HAS_MEM_FUNC(Bandwidth, HasBandwidth);
```
rcurtin: I think maybe this is extra; I don't see it used. (Correct me if I missed it. :))

src/mlpack/methods/kde/kde_rules_impl.hpp
```diff
@@ -108,20 +122,80 @@ Score(const size_t queryIndex, TreeType& referenceNode)
...
     // Don't explore this tree branch.
     score = DBL_MAX;
   }
+  else if (monteCarlo &&
+           referenceNode.NumDescendants() >= MCAccessCoef * initialSampleSize &&
+           std::is_same<KernelType, kernel::GaussianKernel>::value)
+  {
+    // Calculate reference node depth.
+    if (referenceStat.Depth() == 0)
+      referenceStat.Depth() = CalculateDepth(referenceNode);
```
rcurtin: I don't think there's any issue efficiency-wise with putting this code here, but it might be cleaner if you computed the depth during the initialization of `referenceStat`.  You could just do that in the constructor of `KDEStat`.  That could let you remove the code here and also the `CalculateDepth()` function below (or I guess you could move that into `kde_stat.hpp`).

src/mlpack/methods/kde/kde_rules_impl.hpp
```diff
@@ -108,20 +122,80 @@ Score(const size_t queryIndex, TreeType& referenceNode)
...
+
+      // Increase the sample size.
+      sample.resize(newSize);
+      for (size_t i = 0; i < m; ++i)
+      {
+        // Sample and evaluate random points from the reference node.
+        const size_t randomPoint = math::RandInt(0, numDesc);
+        sample(oldSize + i) =
+            EvaluateKernel(queryIndex, referenceNode.Descendant(randomPoint));
+      }
```
rcurtin: Hmm, I see an acceleration opportunity here, but I don't know how much speedup it would give or if it is worth the amount of work I think it might be.  We're computing the kernel evaluations for `m` points at once, and `m` could be of reasonable size (like 20 or more).  Specifically, `EvaluateKernel` sequentially calls `kernel.Evaluate(metric.Evaluate(query, reference))`, and at least in the `EuclideanDistance` case, `metric.Evaluate(query, reference)` boils down to `arma::norm(query - reference, 2)`.  But that's for just a single point; it may be possible to express the Euclidean distance calculation between *many* query and reference points as a big linear algebra expression and thus Armadillo could make better use of LAPACK or BLAS internally to do that: typically, larger matrix multiplications can be quicker than a series of small vector operations.  (Every situation is different; that may not always apply.)

Anyway, think about it, and let me know what you think.  It may be a lot of work---I haven't fully thought through all the changes.  So if you don't think that you want to do it, it's perfectly okay; I just wanted to suggest the idea.

src/mlpack/methods/kde/kde_rules_impl.hpp
```diff
@@ -108,20 +122,80 @@ Score(const size_t queryIndex, TreeType& referenceNode)
...
+      meanSample = arma::mean(sample);
+      const double stddev = arma::stddev(sample);
+      size_t mThresh =
+          std::ceil(z * stddev * (1 + relError) / (relError * meanSample));
+      mThresh *= mThresh;
```
rcurtin: Correct me if I'm wrong, but shouldn't the `ceil()` come after `mThresh` is squared?

src/mlpack/methods/kde/kde_stat.hpp
```diff
@@ -68,6 +75,7 @@ class KDEStat
   {
     ar & BOOST_SERIALIZATION_NVP(centroid);
     ar & BOOST_SERIALIZATION_NVP(validCentroid);
+    ar & BOOST_SERIALIZATION_NVP(depth);
```
rcurtin: If we do reverse compatibility for `KDE` we'll have to do it here too. :)

src/mlpack/tests/kde_test.cpp
```diff
@@ -818,4 +857,378 @@ BOOST_AUTO_TEST_CASE(NotTrained)
...
+  size_t correctResults = 0;
+  for (size_t i = 0; i < query.n_cols; ++i)
+  {
+    const double resultRelativeError =
+      std::abs((bfEstimations[i] - treeEstimations[i]) / bfEstimations[i]);
+    if (resultRelativeError < relError)
+      ++correctResults;
+  }
+
+  BOOST_REQUIRE_GT(correctResults, 70);
```
rcurtin: Have you tried running this test many times with different random seeds to check if it is robust?  I don't know how likely it is for the Monte Carlo sampling to fail, although based on what I saw when I ran some simulations earlier, getting 70 out of 100 within the bound is probably a good bound that won't fail.

src/mlpack/tests/main_tests/kde_test.cpp
```diff
@@ -405,4 +405,120 @@ BOOST_AUTO_TEST_CASE(KDEMainInvalidRelativeError)
...
+  // Valid between 0 and 1.
+  SetInputParam("mc_break_coef", 0.3);
+  BOOST_REQUIRE_NO_THROW(mlpackMain());
+
+  // Invalid greater than 1.
+  SetInputParam("mc_break_coef", 1.1);
+  BOOST_REQUIRE_THROW(mlpackMain(), std::runtime_error);
+  Log::Fatal.ignoreInput = false;
+}
+
```
rcurtin: Since the Monte Carlo search has a random parameter, it may be worth running twice with `--monte_carlo` specified to ensure that the results are at least a little bit different (due to different random sampling).  This will help test that `--monte_carlo` is actually doing something. :)

src/mlpack/methods/kde/kde_rules_impl.hpp
```diff
@@ -108,20 +122,80 @@ Score(const size_t queryIndex, TreeType& referenceNode)
...
+           referenceNode.NumDescendants() >= MCAccessCoef * initialSampleSize &&
+           std::is_same<KernelType, kernel::GaussianKernel>::value)
+  {
+    // Calculate reference node depth.
+    if (referenceStat.Depth() == 0)
+      referenceStat.Depth() = CalculateDepth(referenceNode);
+
+    // Monte Carlo probabilistic estimation.
+    const double currentAlpha =
+        MCBeta / std::pow(2, referenceNode.Stat().Depth() + 1);
```
rcurtin: What do you think about the idea of "reclaiming" error that Dongryeol mentions in the paper?  Honestly I suspect that is responsible for the largest part of the speedup.  The idea is detailed more in this paper: https://arxiv.org/pdf/1206.6857.pdf in Section 5 (see under Theorem 2).

src/mlpack/methods/kde/kde_rules_impl.hpp
robertohueso: You're right :)

src/mlpack/methods/kde/kde_model_impl.hpp
robertohueso: I think it's a good idea since it makes possible to load a model and get the value of those variables without having to implement visitors for each of them right?

src/mlpack/methods/kde/kde_impl.hpp
robertohueso: There's a small issue with using ```BOOST_TEMPLATE_CLASS_VERSION(...)``` [because of multi argument templates in macros](https://stackoverflow.com/questions/8942912/how-to-pass-multi-argument-templates-to-macros). Anyway I implemented using boost version struct and it works :)

src/mlpack/methods/kde/kde_rules_impl.hpp
robertohueso: I thought about this and I thought I should get the same results. The thing is for ```tree::StandardCoverTree``` I'm getting different depths across reference nodes explored by the single-tree score function, e.g. [diff current implementation (left) vs KDEStat implementation](https://www.diffchecker.com/5WedGrAB) this is specifically from the test ```KDETest/GaussianSingleCoverTreeMonteCarloKDE```. So I thought of making it this way just to be sure depths are right although the code is less clean. In exchange for that you also save time computing depths for query nodes.

src/mlpack/methods/kde/kde_main.cpp
robertohueso: About the first monte_carlo=false by default question: Yes, they need to specify they want to do Monte Carlo sampling. I thought this would be a good default since it produces a non-deterministic output and maybe it's a good tool for users with more expertise. As I don't use KDE everyday I don't know that well if it would be good to use Monte Carlo by default, so if you think it's better then I can quickly change it :)

About the rest of questions. Modifiers are now implemented and users can change parameters for loaded models :)

src/mlpack/methods/kde/kde_main.cpp
```diff
@@ -124,6 +185,11 @@ static void mlpackMain()
   const std::string modeStr = CLI::GetParam<std::string>("algorithm");
   const double relError = CLI::GetParam<double>("rel_error");
   const double absError = CLI::GetParam<double>("abs_error");
+  const bool monteCarlo = CLI::GetParam<bool>("monte_carlo");
+  const double MCProbability = CLI::GetParam<double>("mc_probability");
+  const int initialSampleSize = CLI::GetParam<int>("initial_sample_size");
+  const double MCEntryCoef = CLI::GetParam<double>("mc_entry_coef");
+  const double MCBreakCoef = CLI::GetParam<double>("mc_break_coef");
```
rcurtin: Tiny style issue, we should use `mcEntryCoef` or something like that here. :)

src/mlpack/methods/kde/kde_rules.hpp
```diff
@@ -44,8 +49,13 @@ class KDERules
            arma::vec& densities,
            const double relError,
            const double absError,
+           const double MCProb,
+           const size_t initialSampleSize,
+           const double MCAccessCoef,
+           const double MCBreakCoef,
```
rcurtin: The variables here should also be `mcProb` (or however you want to do it, so long as the first letter is lowercase in accordance with the style guide), and a few other places around the file.

src/mlpack/methods/kde/kde_rules_impl.hpp
rcurtin: Wow, that is strange and unexpected.  I wonder if there is a bug in the cover tree implementation.  I don't have the time to debug it right now, but I'll take a look when I get home.  I think for now we should just leave this code as-is.

src/mlpack/methods/kde/kde_rules_impl.hpp
robertohueso: I'm thinking about how to do it, but in any case, wouldn't it also be nice to modify the implementation of ```src/mlpack/core/metrics/lmetric_impl.hpp ```?
From

```
typename VecTypeA::elem_type LMetric<2, true>::Evaluate(
    const VecTypeA& a,
    const VecTypeB& b)
{
  return sqrt(arma::accu(square(a - b)));
}
```
to
```
typename VecTypeA::elem_type LMetric<2, true>::Evaluate(
    const VecTypeA& a,
    const VecTypeB& b)
{
  return arma::norm(a - b, 2);
}
```

src/mlpack/methods/kde/kde_rules_impl.hpp
robertohueso: I think I made this work (at least for single-trees) but at the moment I can't see that much of a difference in tests so I'm testing whether it's actually working or not :)

robertohueso: Implemented for single-trees in a68d28c8e3ac9ab949eda76469b665104d32912c.
After some tests I think it's robust. Although results depends very much on input parameters, it performs ~25-30% less base cases (compared to the non-reclaiming version) and the amount of estimations out of the relative error bound remain the same :)

src/mlpack/methods/kde/kde_rules_impl.hpp
rcurtin: You're right, that might actually make this a lot easier.  I am not sure what we would need to do for other metrics, though.  If you want to change to `arma::norm(a - b, 2)` though please just do a couple simulations first to make sure that this doesn't affect the runtime.

src/mlpack/methods/kde/kde_model.hpp
```diff
@@ -227,6 +396,22 @@ class KDEModel
...
+  double mcProb;
+
+  //! Size of the initial sample for Monte Carlo estimations.
+  size_t initialSampleSize;
+
+  //! Entry coefficient for Monte Carlo estimations.
+  double mcEntryCoef;
+
+  //! Break coefficient for Monte Carlo estimations.
+  double mcBreakCoef;
```
rcurtin: I see, having these held in `KDEModel` instead of `KDE` can make access a little bit easier, but if we're going to do that we should add a comment pointing out that these are held to simplify the code of accessing or changing these values for the underlying `KDE<>` object.

There is some risk of these things being out of sync, but you are using it carefully in the rest of the code, so I think it is not a problem.

src/mlpack/methods/kde/kde_model_impl.hpp
```diff
@@ -61,6 +81,11 @@ inline KDEModel::KDEModel(KDEModel&& other) :
   other.absError = 0;
   other.kernelType = KernelTypes::GAUSSIAN_KERNEL;
   other.treeType = TreeTypes::KD_TREE;
+  other.monteCarlo = false;
+  other.mcProb = 0.95;
+  other.initialSampleSize = 300;
+  other.mcEntryCoef = 3;
+  other.mcBreakCoef = 0.7;
```
rcurtin: One disadvantage here is that this code is duplicated, and so if the defaults are ever changed in `KDE`, they won't be changed here unless someone remembers to do that.

src/mlpack/methods/kde/kde_rules.hpp
```diff
@@ -105,15 +123,41 @@ class KDERules
...
   MetricType& metric;
 
   //! Instantiated kernel.
   KernelType& kernel;
 
+  //! Whether Monte Carlo estimations are going to be applied.
+  const bool monteCarlo;
+
+  //! Accumulated not used MC alpha values for each query point.
+  arma::vec accumMCAlpha;
```
rcurtin: I think that `alpha` here refers to the absolute error that has been used so far; maybe it would be better to give the variable a name closer than that or add some clarity to the comment?

src/mlpack/methods/kde/kde_rules_impl.hpp
```diff
@@ -108,23 +124,105 @@ Score(const size_t queryIndex, TreeType& referenceNode)
...
+    }
   }
   else
   {
     score = minDistance;
+
+    // If node is going to be exactly computed, reclaim not used alpha.
+    if (kernelIsGaussian && monteCarlo && referenceNode.IsLeaf())
+    {
+        accumMCAlpha(queryIndex) +=
```
rcurtin: Oops, looks like this line is doubly indented.

src/mlpack/methods/kde/kde_rules_impl.hpp
```diff
@@ -108,23 +124,105 @@ Score(const size_t queryIndex, TreeType& referenceNode)
...
+      if (sample.size() < mThresh)
+        m = mThresh - sample.size();
+      else
+        m = 0;
+    }
+
+    if (useMonteCarloPredictions)
+    {
+      // Use Monte Carlo estimation.
+      accumMCAlpha(queryIndex) = 0;
```
rcurtin: Shouldn't this just be the sum of unused error across all nodes?  I think maybe we don't need to modify it at all in this case.  Correct me if I misunderstood.  (Maybe it would be helpful to add a block comment describing what is going on too.)

src/mlpack/methods/kde/kde_rules_impl.hpp
```diff
@@ -200,10 +298,103 @@ Score(TreeType& queryNode, TreeType& referenceNode)
...
+      score = minDistance;
+    }
   }
   else
   {
     score = minDistance;
+
+    // Reclaim not used alpha for Monte Carlo.
+    if (kernelIsGaussian && monteCarlo)
+      queryStat.AccumAlpha() += CalculateAlpha(referenceNode.Stat().Depth());
```
rcurtin: In the code you've written, it looks to me like you are only adding to `AccumAlpha()` when Monte Carlo approximation is not used.  I think this is correct, if the query node (or the query point in the case of single tree) is then evaluated exhaustively with the descendant points of the reference node.  But, the reference node may be pruned by a non-Monte Carlo step (i.e., the difference between minimum and maximum kernel values may be small enough that a "regular" prune can be done).  If that happens, then `AccumAlpha()` will need to be adjusted, as even a regular prune will affect the error that can be used.

Do you think you could write something up (not necessarily in the code, anywhere is fine) detailing the strategy you are using for reclaiming error?  It might be nice to know that we are implementing the right thing by first writing it down, along with either some reasoning for why the result still satisfies the approximation guarantees (perhaps a formal proof would be useful).  That would probably make it a lot easier for me to review too. :+1:

src/mlpack/methods/kde/kde_stat.hpp
```diff
@@ -52,6 +54,18 @@ class KDEStat
...
   }
 
+  //! Get depth of the node.
+  inline size_t Depth() const { return depth; }
+
+  //! Modify depth of the node.
+  inline size_t& Depth() { return depth; }
+
+  //! Get accumulated Monte Carlo alpha of the node.
+  inline double AccumAlpha() const { return accumAlpha; }
```
rcurtin: Do we need to ensure that this is set to 0 before traversal?  Someone may call `KDE::Evaluate()` with a query tree that's already been used.

src/mlpack/methods/kde/kde_rules.hpp
robertohueso: I agree it needs to be clarified. Before doing it, as far as I had understood, I think ```alpha``` refers to the probability of an MC estimation **not** being bounded by the relative error.

Maybe I misunderstood the "reclaiming probability" of [this paper](http://papers.nips.cc/paper/3539-fast-high-dimensional-kernel-summations-using-the-monte-carlo-multipole-method.pdf) but I think it doesn't refer to absolute error (although [the other paper](https://arxiv.org/pdf/1206.6857.pdf) does refer to absolute error).

This is what I have in mind: The bigger ```\alpha_{node}``` is, the more likely the scoring function is to use MC estimation. If some part of the tree is evaluated in any way that is not MC estimation, then its probability has not been used, therefore it can be used in the next scoring function evaluation. The sum of ```\alpha_{node}``` values at different nodes has to be equal to the original alpha input value of the rules (this means, the total probability of MC estimated nodes not being in the relative error bound is the same as the original ```alpha``` input value).

Did I miss or misunderstand anything?

src/mlpack/methods/kde/kde_model_impl.hpp
robertohueso: Solved in 8147e8c33e59d55726ba25a6725722d02e917a75 :)

src/mlpack/methods/kde/kde_rules.hpp
rcurtin: Ah, I see, thanks for putting the time into writing that up.  When I wrote my original comment, I did think what was being referred to was the absolute error reclamation of the "Faster Gaussian Summation: Theory and Experiment".  So, I see now that what you have written is right, and the only thing that remains will be to actually review the code in depth. :)  But it is getting late here (my evening was mostly taken up handling some around-the-house things), so I'll review the code tomorrow over breakfast. :+1:

It might be worth thinking about doing the absolute error reclamation later, but, up to you---I don't think that was in your proposal, so maybe more exciting to move on to other parts. :)  We can always open an issue for it and leave it for later or someone else interested to do.  Let me know what you think.

src/mlpack/methods/kde/kde_rules_impl.hpp
rcurtin: I successfully reproduced this issue with the following code:

```
#include <mlpack/core.hpp>
#include <mlpack/core/tree/cover_tree.hpp>

using namespace mlpack;
using namespace mlpack::tree;
using namespace mlpack::metric;

class DepthStatistic
{
 public:
  DepthStatistic() : depth(0) { }

  template<typename TreeType>
  DepthStatistic(TreeType& node)
  {
    depth = 0;
    TreeType* p = node.Parent();
    while (p != NULL)
    { 
      ++depth;
      p = p->Parent();
    }
  }

  template<typename Archive>
  void serialize(Archive& ar)
  {
    ar & BOOST_SERIALIZATION_NVP(depth);
  }

  const size_t Depth() const { return depth; }

 private:
  size_t depth;
};

int main()
{
  // Create a random dataset and build the tree.
  arma::mat data(2, 3000, arma::fill::randu); // Strangely, the issue does not happen when d=10, n=10k.
  typedef StandardCoverTree<EuclideanDistance, DepthStatistic, arma::mat>
      TreeType;
  TreeType t(std::move(data));

  // Now check each node in the dataset to see if its depth is right.
  std::queue<TreeType*> queue;
  queue.push(&t);

  while (!queue.empty())
  {
    TreeType* node = queue.front();
    queue.pop();

    // Compute depth manually.
    size_t depth = 0;
    TreeType* p = node->Parent();
    while (p != NULL)
    { 
      ++depth;
      p = p->Parent();
    }
    
    if (depth != node->Stat().Depth())
    { 
      Log::Warn << "Depth is different!  We computed " << depth << " now but "
          << node->Stat().Depth() << " when the tree was built!" << std::endl;
    }
    
    for (size_t i = 0; i < node->NumChildren(); ++i)
      queue.push(&node->Child(i));
  }
}
```

I quickly reviewed the cover tree code, and during cover tree construction it is sometimes possible for nodes to be moved around or removed from the tree structure.  Every time this happens, the statistic of a modified child node is rebuilt---but the statistics are not rebuilt recursively!  I bet this is the issue, and tomorrow I'll play with it a bit more and post a patch, which you can just integrate into this PR.

src/mlpack/methods/kde/kde_model_impl.hpp
rcurtin: Clever solution :)

src/mlpack/methods/kde/kde_main.cpp
```diff
@@ -68,10 +83,32 @@ PROGRAM_INFO("Kernel Density Estimation",
...
     "\n"
     "It is possible to select either a reference dataset or an input model "
-    "but not both at the same time.",
+    "but not both at the same time. If an input model is selected and "
+    "parameter values are not set (e.g. " + PRINT_PARAM_STRING("bandwidth") +
+    ") then default parameter values will be used."
+    "\n\n"
+    "In addition to the last program call, it is also possible to activate "
+    "Monte Carlo estimations if a Gaussian kernel is used. The following will "
+    "run KDE using a Monte Carlo estimation when possible. The results will be "
```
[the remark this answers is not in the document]
robertohueso: Sounds good :)

src/mlpack/methods/kde/kde_main.cpp
```diff
@@ -50,6 +50,21 @@ PROGRAM_INFO("Kernel Density Estimation",
...
+    "initial sample size for the Monte Carlo estimation using " +
+    PRINT_PARAM_STRING("initial_sample_size") + ". This implementation will "
+    "only consider a node, as a candidate for the Monte Carlo estimation, if "
+    "its number of descendant nodes is bigger than the initial sample size. "
+    "This can be controlled using a coefficient that will multiply the initial "
+    "sample size and can be set using " + PRINT_PARAM_STRING("mc_entry_coef") +
+    ". To avoid using the same amount of computations an exact approach would "
+    "take, this program recurses the tree whenever a fraction of the amount of "
+    "the node's descendant points have already been computed. This fraction "
+    "is set using " + PRINT_PARAM_STRING("mc_break_coef") + "."
```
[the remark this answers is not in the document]
robertohueso: Sounds also good, thanks! :)

src/mlpack/methods/kde/kde_rules_impl.hpp
```diff
@@ -200,10 +295,103 @@ Score(TreeType& queryNode, TreeType& referenceNode)
...
+    // Calculate reference node depth.
+    if (referenceStat.Depth() == 0)
+      referenceStat.Depth() = CalculateDepth(referenceNode);
+
+    // Monte Carlo probabilistic estimation.
+    // Calculate alpha.
+    const double depthAlpha = CalculateAlpha(referenceNode.Stat().Depth());
+    double alpha = depthAlpha + queryStat.AccumAlpha();
+    if (alpha > 1)
+      alpha = 1;
```
robertohueso: I'm almost sure there's a bug here. If you sum up unused alpha values, it can never be greater than 1. I will work on this.

src/mlpack/methods/kde/kde.hpp
```diff
@@ -257,6 +352,41 @@ class KDE
 } // namespace kde
 } // namespace mlpack
 
+//! Set the serialization version of the KDE class.
+/* TODO FIX
```
rcurtin: I don't know if `TODO FIX` is needed---actually, I think this is not fixable due to the limitations of C macros.  The comment is still useful though. :)

src/mlpack/methods/kde/kde_rules_impl.hpp
```diff
@@ -26,23 +29,35 @@ KDERules<MetricType, KernelType, TreeType>::KDERules(
...
     sameSet(sameSet),
     lastQueryIndex(querySet.n_cols),
     lastReferenceIndex(referenceSet.n_cols),
     baseCases(0),
     scores(0)
 {
-  // Nothing to do.
+  // Initialize accumMCAlpha only if Monte Carlo estimations are available.
+  if (monteCarlo && kernelIsGaussian)
+    accumMCAlpha = std::move(arma::vec(querySet.n_cols, arma::fill::zeros));
```
rcurtin: I think the `std::move()` is superfluous here, and I think return value optimization (RVO) will handle the optimization here (but not if `std::move()` is used).

src/mlpack/methods/kde/kde_rules_impl.hpp
rcurtin: I think I am incorrect on my comment above---if we are doing Monte Carlo sampling here, we are doing so with the entire alpha of `depthAlpha + accumMCAlpha(queryIndex)`.  Thus, the "leftover" should be 0, just like you have it here.

src/mlpack/methods/kde/kde_rules_impl.hpp
```diff
@@ -108,23 +124,102 @@ Score(const size_t queryIndex, TreeType& referenceNode)
...
+      score = DBL_MAX;
+      densities(queryIndex) += numDesc * meanSample;
+    }
+    else
+    {
+      if (referenceNode.IsLeaf())
+      {
+        // Reclaim not used alpha since the node will be exactly computed.
+        accumMCAlpha(queryIndex) +=
+            CalculateAlpha(referenceNode.Stat().Depth());
```
rcurtin: Hmm, but it's possible that non-Monte Carlo approximation may occur at a non-leaf level, meaning that that node's unused probability isn't actually ever added to `accumMCAlpha(queryIndex)`.

src/mlpack/methods/kde/kde_rules_impl.hpp
rcurtin: It's possible that floating point rounding errors could cause it to be greater than 1, but I agree, if it's significantly greater than 1, that would be a problem.

src/mlpack/methods/kde/kde_rules_impl.hpp
```diff
@@ -200,10 +295,103 @@ Score(TreeType& queryNode, TreeType& referenceNode)
...
+      // Use Monte Carlo estimation.
+      queryStat.AccumAlpha() = 0;
+      score = DBL_MAX;
+      for (size_t i = 0; i < queryNode.NumDescendants(); ++i)
+        densities(queryNode.Descendant(i)) += numDesc * means(i);
+    }
+    else
+    {
+      // Reclaim not used alpha for Monte Carlo.
+      queryStat.AccumAlpha() += CalculateAlpha(referenceNode.Stat().Depth());
```
rcurtin: I think the bug may have to do with this---here you are adding to `AccumAlpha()` at every non-approximated reference node.  Let's say I add the alpha for a node to `AccumAlpha()` here, but then I recurse (say there are two children) and those children, too, do not use Monte Carlo approximation.  Then, we will add the alpha for each of the children, but we already know that the sum of the two child alphas is the parent alpha---so, overall, we've added 2*alpha, which is twice as much as we should have.  I think this may be the reason for the too-large alpha values.

src/mlpack/methods/kde/kde_rules_impl.hpp
```diff
@@ -240,6 +428,27 @@ EvaluateKernel(const arma::vec& query, const arma::vec& reference) const
...
+    currentNode = currentNode->Parent();
+  }
+  return depth;
+}
+
+template<typename MetricType, typename KernelType, typename TreeType>
+inline force_inline double KDERules<MetricType, KernelType, TreeType>::
+CalculateAlpha(const size_t depth) const
+{
+  return mcBeta / std::pow(2, depth + 1);
```
rcurtin: Ahh... I realized something frustrating.  For kd-trees, this code works properly because each node has two children, and we "split" the probability equally among them.  But cover trees and other tree types may have arbitrary numbers of children.  Suppose we have a cover tree node with three children... then, each child should receive 1/3 of the probability of the parent, not 1/2.

I think the way to fix this may be a more complex algorithm that stores the computed `alpha` in the statistic, and when recursion is performed, the new `alpha` is computed based on the parent's number of children and the parent's `alpha` value.  Let me know if I can clarify what I mean.

rcurtin: This could also cause the sum of alphas to be too large across trees.  (That will be true for cover trees, most likely, but it shouldn't be true for kd-trees.)

src/mlpack/methods/kde/kde_rules_impl.hpp
rcurtin: Here's a patch you can apply to the branch: https://gist.github.com/071f56263151692ba484e511db859f82

robertohueso: Thanks for the patch! :) I found the same problem with ```RectangleTree``` so I solved it in a similar way in 832c423b825a5cb436f0a21c544ed7804e8fa1ec.

Node's depth is now calculated in ```KDEStat``` (432da7bda5d509e706514715e7d086f8d77beefc) and everything seems to work fine.

src/mlpack/methods/kde/kde_rules_impl.hpp
robertohueso: This should be solved with 253332afaa68faf973c80a2742f0afc4432e65ea. Thanks for pointing this out :)

src/mlpack/methods/kde/kde_rules_impl.hpp
```diff
@@ -240,6 +428,32 @@ EvaluateKernel(const arma::vec& query, const arma::vec& reference) const
...
+    TreeType* parent = node->Parent();
+    if (parent == NULL)
+    {
+      // Base case.
+      stat.MCAlpha() = mcBeta;
+    }
+    else
+    {
+      // Recurse.
+      stat.MCAlpha() = CalculateAlpha(parent) / parent->NumChildren();
```
rcurtin: I think that if we can assume that a parent node is always visited before a child node (which is true for all mlpack traversals), then we can just do `stat.MCAlpha() = parent->Stat().MCAlpha() / parent->NumChildren()`.  Let me know if I overlooked something though.

src/mlpack/methods/kde/kde_rules_impl.hpp
rcurtin: Ah, nice catch, I forgot that that would occur with `RectangleTree` also. :+1:

rcurtin:
I spent some time thinking about how the structure of the cover tree affects the Monte Carlo sampling (and KDE in general).  To be specific, the property of the cover tree that makes this hard is each cover tree node holds one point, and each cover tree node will have a *self-child* that holds the same point.  Thus, there is duplication of points in the tree.  However, `NumDescendants()` does handle this properly---it returns the number of *unique* points that are descendants.  So we can trust that. :)

Where it gets hard is this situation: suppose that we are considering some reference node that happens to be a self-child (i.e. `node.Point(0) == node.Parent()->Point(0)`).  In the single-tree case, the single-tree recursion definition requires that `Score()` be called on a query point and reference node, and if `Score()` does not prune, then `BaseCase()` is called on all combinations of query points and points held in the reference node (which for the cover tree is just that one point).  That means that when we get to this reference node that is a self-child, the base case has already been done with that particular point.  (The implementations of the recursions are smart enough to not call `BaseCase()` twice, but it affects our pruning rule too, which is why I am writing so much about it here.)  Thus, if we compute that we can prune this node and then add `kernelValue * referenceNode.NumDescendants()` to the KDE estimate, we're actually *double-counting* that self-child.  Therefore we should be doing `kernelValue * (referenceNode.NumDescendants() - 1)` in this case.

For Monte Carlo sampling, we need to consider the same situation, and ensure that we don't sample the single point contained in the reference node (since it's already been computed earlier).  Similarly when we add an estimate, we'll need to multiply it by `(referenceNode.NumDescendants() - 1)`.  But again, these extra conditions only apply when we are looking at a reference node that is a self-child of another reference node.

When we expand to the dual-tree case, things get a little more complex.  Before I pointed out that there are cases where we have to multiply by `referenceNode.NumDescendants() - 1` and not just `referenceNode.NumDescendants()` to avoid double-counting.  But for dual-tree it's a bit harder.  In the single-tree case, if we encounter a reference node's self-child, we know the base case has already been computed between our query point and the reference point held by that node.  When it's dual-tree, though, we actually don't know this---we have already computed the base case only if *both* our query node is a self-child *and* our reference node is a self-child.  Thus the condition to check gets a little bit more difficult, but the general idea should be the same.

In order to programmatically check whether or not this situation is happening, you can use this:

```
if (TreeTraits<TreeType>::HasSelfChildren && referenceNode.Point(0) == referenceNode.Parent()->Point(0))
```

(and the same for the query node case).  Anyway, I think that's right, I did it from memory.  It's at least close to right. :)

Let me know what you think.  It's possible I've overlooked a situation or something here. :+1:

src/mlpack/methods/kde/kde_rules_impl.hpp
robertohueso: I think this is fixed for single-trees in f0aa09d651007b0c12f7b01f724f959f776b3e52.

robertohueso: And also for dual-trees in e4d507d4507218d5c2051aa538e9014c587c9dd8 :)

src/mlpack/methods/kde/kde_rules_impl.hpp
robertohueso: Solved with f0aa09d651007b0c12f7b01f724f959f776b3e52 and e4d507d4507218d5c2051aa538e9014c587c9dd8.

src/mlpack/methods/kde/kde_rules_impl.hpp
robertohueso: That's right, but I have issues with Binary trees and Octrees.

Both their traversals work in the way you mentioned but they don't call the ```Score()``` method for the root node, therefore ```rootNode.Stat().MCAlpha() != mcBeta``` instead (if the tree has never been used) will be ```rootNode.Stat().MCAlpha() = 0```.

I can see a few ways of making this work, either:

1. Leave this method as it is.
2. Manually assigning ```rootNode.Stat().MCAlpha()``` before calling the traversal (I think this one would be interesting).
3. Calling ```Score()``` in the root nodes of the trees' traversals (although I tried this one and didn't seem to work very well, I need to have a deeper look).

rcurtin: Huh, I see.  I would imagine that calling `Score()` on the root nodes wouldn't be any problem though---it presumably shouldn't do any pruning.  If you want I can help dig deeper into that, just let me know.

src/mlpack/methods/kde/kde_rules_impl.hpp
rcurtin: Nice.  Actually I thought this code would be really difficult and complex, but now that I read through what you did it turns out to be pretty simple. :+1:

src/mlpack/methods/kde/kde_rules_impl.hpp
robertohueso: The traversals issue should be solved with 3554b2cd9e740432f01a902e0b56d0bca6085e7b and 1dc6df50e21e7d9c95b9a7712d16b24346a9b5bb. ```CalculateAlpha()``` has been improved in a677fa3ce88ac2efb0795b91c6e3e6a26a3a7f00 :)

src/mlpack/methods/kde/kde_stat.hpp
robertohueso: You're right. This should be solved with dd0dca90f555c7f6cc58559581e4bf7d197d0903 and 7ea2abfeefd193f71878ac07d1c743e72a47d3f0 :)

src/mlpack/methods/kde/kde_impl.hpp
```diff
@@ -418,6 +456,16 @@ Evaluate(Tree* queryTree,
                                 "dual-tree");
   }
 
+  // Clean accumulated alpha if Monte Carlo estimations are available.
+  if (monteCarlo && std::is_same<KernelType, kernel::GaussianKernel>::value)
+  {
+    Timer::Start("cleaning_query_tree");
+    KDECleanRules<Tree> cleanRules;
+    DualTreeTraversalType<KDECleanRules<Tree>> cleanTraverser(cleanRules);
```
rcurtin: I'd suggest using two single-tree traversals here; the reason is that a dual-tree traversal will compare every pair of nodes in the query tree and reference tree (`O(N^2)` where each tree has `O(N)` nodes), but a single-tree traversal will fix one tree in `O(N)` time.  So the scaling should be better with two individual single-tree recursions.

rcurtin:
Hey Roberto, this is looking good.  Do you think it is getting close to ready?  From my end it seems like it is pretty close.  Based on my review I think I only see a couple things that need to be handled:

 * Add something to `HISTORY.md` :)
 * The cover tree may need special handling in the rules, like I wrote up here: https://github.com/mlpack/mlpack/pull/1934#issuecomment-510482833
 * If you can provide a couple benchmarks to show the kinds of acceleration we are seeing, we can be sure that nothing is wrong.

But otherwise this seems like it's pretty close to me.  Let me know if I overlooked anything.

src/mlpack/methods/kde/kde_impl.hpp
robertohueso: You're right! And actually we only need to clean the query tree :)

rcurtin: Ah, good point. :+1:

robertohueso:
A couple of benchmarks:
```
Parameters valid for all cases:
[INFO ]   abs_error: 0
[INFO ]   bandwidth: 0.8
[INFO ]   initial_sample_size: 100
[INFO ]   kernel: gaussian
[INFO ]   mc_break_coef: 0.7
[INFO ]   mc_entry_coef: 3
[INFO ]   mc_probability: 0.95
[INFO ]   query_file: thyroid_test.csv
[INFO ]   reference_file: thyroid_train.csv
[INFO ]   rel_error: 0.05
[INFO ]   tree: kd-tree
```
* Dual-tree (With and without Monte Carlo)
```
Using Monte Carlo:
[INFO ] 64050 node combinations were scored.
[INFO ] 418356 base cases were calculated.
[INFO ] Program timers:
[INFO ]   building_query_tree: 0.023515s
[INFO ]   building_reference_tree: 0.026409s
[INFO ]   cleaning_query_tree: 0.000180s
[INFO ]   computing_kde: 2.838616s
[INFO ]   loading_data: 0.019037s
[INFO ]   saving_data: 0.001887s
[INFO ]   total_time: 2.909071s

Without Monte Carlo:
[INFO ] 1441840 node combinations were scored.
[INFO ] 12878996 base cases were calculated.
[INFO ] Program timers:
[INFO ]   building_query_tree: 0.022603s
[INFO ]   building_reference_tree: 0.028030s
[INFO ]   computing_kde: 10.444783s
[INFO ]   loading_data: 0.019791s
[INFO ]   saving_data: 0.002408s
[INFO ]   total_time: 10.516572s
```
* Single-tree (With and without Monte Carlo)
```
Using Monte Carlo:
[INFO ] 24544 node combinations were scored.
[INFO ] 91645 base cases were calculated.
[INFO ] Program timers:
[INFO ]   building_reference_tree: 0.026144s
[INFO ]   computing_kde: 2.349632s
[INFO ]   loading_data: 0.019392s
[INFO ]   saving_data: 0.001897s
[INFO ]   total_time: 2.396106s

Without Monte Carlo:
[INFO ] 2444164 node combinations were scored.
[INFO ] 12878996 base cases were calculated.
[INFO ] Program timers:
[INFO ]   building_reference_tree: 0.027618s
[INFO ]   computing_kde: 12.525483s
[INFO ]   loading_data: 0.020607s
[INFO ]   saving_data: 0.001905s
[INFO ]   total_time: 12.574709s
```
I'm quite excited about this numbers :)

rcurtin:
That looks great!  Do you know what the levels of achieved approximation were for each run?

robertohueso:
For the dual-tree Monte Carlo, 1.6% of the query points are out of the relative error bound and for the single-tree Monte Carlo, 3.6% of the query points are out of the relative error bound. Both are under the 5% I'd expect with 95% Monte Carlo :)

rcurtin:
Right, that sounds good, but I meant more along the lines of comparisons to the default KDE pruning.  i.e., what's the average relative error across query points with and without Monte Carlo sampling?  This can let us understand how much the speedup "cost" in terms of error. :)

robertohueso:
In both dual and single tree without Monte Carlo average relative error is ~0% (i.e. no nodes were pruned by enforcing an strict error bound).
When using Monte Carlo:
* Dual-tree average relative error: ~1.59909691%
* Single-tree average relative error: ~1.869897056%

rcurtin:
Ah, ok, so that is definitely still nice as the regular rules don't allow any pruning at all.  You could also try a smaller bandwidth, so that pruning with the default rules does happen, and then see what happens when Monte Carlo sampling is also added on top.  The more benchmarking we do, the more we can learn about where Monte Carlo sampling works great, which makes it easier to provide good documentation on where the technique should be used---and also see where further improvements might be possible. :+1:

In any case, I'm pretty happy with how this is looking and so I think once the cover tree rules are handled correctly, it's ready for merge.  But, I'm not sure if you had anything else planned, so let me know which direction you'd like to go.

robertohueso:
About the cover tree rules, I think they're already handled in two ways:
1. Using the `newCalculations` variable.
```
if (tree::TreeTraits<TreeType>::FirstPointIsCentroid &&
    (traversalInfo.LastQueryNode() != NULL) &&
    (traversalInfo.LastReferenceNode() != NULL) &&
    (traversalInfo.LastQueryNode()->Point(0) == queryNode.Point(0)) &&
    (traversalInfo.LastReferenceNode()->Point(0) == referenceNode.Point(0)))
{
  // Don't duplicate calculations.
  newCalculations = false;
  lastQueryIndex = queryNode.Point(0);
  lastReferenceIndex = referenceNode.Point(0);
}
```
2. Using the `canReclaimAlpha` variable.
```
// Check if not used alpha can be reclaimed for this combination of nodes.
const bool canReclaimAlpha =
    kernelIsGaussian &&
    monteCarlo &&
    newCalculations &&
    ((queryNode.IsLeaf() && referenceNode.IsLeaf()) ||
     (tree::TreeTraits<TreeType>::HasSelfChildren));
```
There're also tests (i.e. `KDETest/GaussianDualCoverTreeMonteCarloKDE` and `KDETest/GaussianSingleCoverTreeMonteCarloKDE`) implemented.

That said, I'm not sure if (semantically) it would make more sense to have a Tree Trait that was something like
```
//! Whether its traversals evaluate the BaseCase() in non-leaf nodes.
static const bool ComputesBaseCaseInNonLeaf = true;
```
:thinking:

rcurtin:
Ah, thanks for the clarification---I have to get dinner now but I'll try to review and understand the details of the code tonight.

> That said, I'm not sure if (semantically) it would make more sense to have a Tree Trait that was something like
>
> ```
> //! Whether its traversals evaluate the BaseCase() in non-leaf nodes.
> static const bool ComputesBaseCaseInNonLeaf = true;
> ```
>
> :thinking:

Hmmm, typically I've tested this in the past with `node.NumPoints() > 0`, because if both the query node and the reference node hold 1 or more point each, then base cases will occur.  However, if you'd prefer to check that at compile time, then you should feel free to add that, but perhaps a better name would be `PointsOnlyInLeaves`, which would be true for `BinarySpaceTree` and false for `CoverTree` and `RectangleTree`.  I suggest that name just because it's a more clear description of the actual property of the tree that controls whether base cases happen in non-leaves, but let me know what you think.

src/mlpack/tests/kde_test.cpp
robertohueso: Yes, it seems to always have a reasonable amount of fails. Actually, none of the tests I did ever got less than 70 within the bound :)

src/mlpack/methods/kde/kde_main.cpp
robertohueso: Before merging: 
1. What do you think about `--monte_carlo` being deactivated by default?
2. Do you think the rest of defaults are ok? So far, they seem to work well.

src/mlpack/methods/kde/kde_rules.hpp
robertohueso: I agree, it's probably a very good idea to work on the absolute error reclamation, but I think if we want to finish what was on the proposal, then we can open an issue and I can take care of this later :)

src/mlpack/methods/kde/kde_main.cpp
rcurtin: I think deactivated by default probably makes the most sense (although I could be convinced either way), and the rest of the defaults seem good to me. :+1:

src/mlpack/methods/kde/kde_rules.hpp
rcurtin: Sounds good to me. :+1:

src/mlpack/tests/kde_test.cpp
```diff
@@ -671,15 +671,29 @@ BOOST_AUTO_TEST_CASE(EmptyQuerySetTest)
  */
 BOOST_AUTO_TEST_CASE(SerializationTest)
 {
-  // Initial KDE model to me serialized.
+  // Initial KDE model to be serialized.
```
rcurtin: Nice catch :)

src/mlpack/methods/kde/kde_rules_impl.hpp
```diff
@@ -108,23 +131,107 @@ Score(const size_t queryIndex, TreeType& referenceNode)
...
   else
   {
     score = minDistance;
+
+    // If node is going to be exactly computed, reclaim not used alpha for
+    // Monte Carlo estimations.
+    if (kernelIsGaussian &&
+        monteCarlo &&
+        newCalculations &&
+        (referenceNode.IsLeaf() || tree::TreeTraits<TreeType>::HasSelfChildren))
```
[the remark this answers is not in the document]
robertohueso: I think we should keep this line
```(referenceNode.IsLeaf() || tree::TreeTraits<TreeType>::HasSelfChildren)```.
`BaseCase()` will be called for every node that has not been pruned or already computed in CoverTree right?

File: src/mlpack/methods/kde/kde_rules_impl.hpp [hunk 39]
```diff
@@ -108,23 +131,107 @@ Score(const size_t queryIndex, TreeType& referenceNode)
...
+
+      // Accumulated alpha has been used.
+      accumMCAlpha(queryIndex) = 0;
+    }
+    else
+    {
+      // Recurse.
+      score = minDistance;
+
+      if (referenceNode.IsLeaf() || tree::TreeTraits<TreeType>::HasSelfChildren)
```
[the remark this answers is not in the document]
robertohueso: It is exactly as you you wrote, `tree::TreeTraits<TreeType>::HasSelfChildren` means that the tree is a cover tree. But the point is cover trees just have single or dual traversers but not breadth first right?
```
template<typename RuleType>
using BreadthFirstDualTreeTraverser = DualTreeTraverser<RuleType>;
```
This is from `cover_tree.hpp`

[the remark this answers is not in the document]
rcurtin: The default cover tree traverser is somewhat strange: it does a depth-first traversal for the query tree, but it does a breadth-first traversal for the reference tree (this is kind of weird, but remember it can be both of these things at once since we're recursing in both the query tree and the reference tree in a dual-tree recursion).  So for the needs of the `BreadthFirstDualTreeTraverser`, the default `CoverTree::DualTreeTraverser` is sufficient.

src/mlpack/methods/kde/kde_rules_impl.hpp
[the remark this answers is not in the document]
rcurtin: I'm thinking that the intent of this if statement is to only add to `accumMCAlpha` if it is not possible that Monte Carlo approximation will happen at any level deeper than this one.  Correct me if I'm wrong on that.

What you wrote about `BaseCase()` is basically right (if we change the word node to point), but just to restate, `BaseCase()` is called between the query point and every reference point in the reference node (i.e. not all the descendants, just the nodes held in the point, which for the cover tree is just one point, and for the kd-tree is zero points unless it's a leaf), and then recursion happens and `Score()` is called with any reference child nodes and the process repeats.

Anyway, I am not sure I'm answering the question very well; I think I've misunderstood what you mean.  Hopefully the answer is helpful, but if not, let me know what you're thinking and we can get it figured out. :)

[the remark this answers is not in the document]
robertohueso: I think I now understand your point and it's right, cover tree is always a tricky one :)

So basically what this means is cover tree reclaims probability just when a node is pruned or when it gets to its leafs (and at that point it's useless since we're doing  a breadth-first traversal). Thanks for the clarification! :)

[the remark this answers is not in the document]
rcurtin: Right, exactly. :+1:

src/mlpack/methods/kde/kde_rules_impl.hpp
```diff
@@ -200,10 +322,111 @@ Score(TreeType& queryNode, TreeType& referenceNode)
           referenceNode.NumDescendants() * kernelValue;
```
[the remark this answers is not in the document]
robertohueso: You're right. It was probably more difficult to think about it in #1301 because in that PR the rules didn't try to approximate self-child, so this wasn't necessary. Thanks for pointing it out :)

src/mlpack/methods/kde/kde_rules_impl.hpp
```diff
@@ -77,11 +92,19 @@ template<typename MetricType, typename KernelType, typename TreeType>
...
 
+  // Calculate alpha if Monte Carlo is available.
+  if (monteCarlo && kernelIsGaussian)
+    depthAlpha = CalculateAlpha(&referenceNode);
+  else
+    depthAlpha = -1;
+
   if (tree::TreeTraits<TreeType>::FirstPointIsCentroid &&
       lastQueryIndex == queryIndex &&
       traversalInfo.LastReferenceNode() != NULL &&
```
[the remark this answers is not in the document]
robertohueso: Thanks for taking the time to write this :) It's now fixed.

Very good idea reusing the distance computation for minDistance and maxDistance! :+1:

src/mlpack/methods/kde/kde_rules_impl.hpp
```diff
@@ -108,23 +131,107 @@ Score(const size_t queryIndex, TreeType& referenceNode)
...
-    {
       kernelValue = EvaluateKernel(queryIndex, referenceNode.Point(0));
-    }
     else
-    {
-      kde::KDEStat& referenceStat = referenceNode.Stat();
       kernelValue = EvaluateKernel(queryPoint, referenceStat.Centroid());
-    }
 
     densities(queryIndex) += referenceNode.NumDescendants() * kernelValue;
```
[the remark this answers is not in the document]
robertohueso: Hopefully we got that speedup now :)

src/mlpack/methods/kde/kde_rules.hpp
robertohueso: Issue  #1954 open :)

src/mlpack/methods/kde/kde_rules_impl.hpp
robertohueso: Turns out arma::norm doesn't allow to evaluate the norm across one of the dimensions (e.g. columns or rows), this is why it got a bit more messy. I think it would be a good idea to have this feature in Armadillo :)

Sorry about the bad code style, it was just a draft to check performance. I tried something like this:
```c++
size_t firstIndex;
  if (alreadyDidRefPoint0)
    firstIndex = 1;
  else
    firstIndex = 0;

arma::uvec newIndexes(m);

for (size_t i = 0; i < m; ++i)
  newIndexes(i) = referenceNode.Descendant(math::RandInt(firstIndex, refNumDesc));

arma::mat newRefPoints = referenceSet.cols(newIndexes);
arma::mat diff = std::move(newRefPoints).each_col() - querySet.unsafe_col(queryIndex);
arma::rowvec distances = arma::sqrt(arma::sum(arma::square(std::move(diff)), 0));

for (size_t i = 0; i < distances.n_cols; ++i)
  sample(oldSize + i) = kernel.Evaluate(distances(i));
```
after doing some benchmark, it performed worse (i.e. runtime was on average ~8% worse for this test) than the current implementation.

I think there might be better ways to implement this than the one I used (at least more elegant ones) but I'm not sure if it would be possible to get the speedup that would justify making the code a bit more messy. So I think this will stay with the current implementation :)

Anyway, let me know what you think.

robertohueso: I tried using `arma::norm(a - b, 2)` for the `EuclideanDistance` and I think we got a slight speedup :)

Implemented in 5c3dc1a2fe6f397acb072330eeaf81e9f00b687a.

rcurtin: Oh, I see what you mean.  That would indeed be a little messy.  There are two additional optimizations that could be done on the code you wrote, but let me write them, and then I'll write why I think it's not worth pursuing either of them:

 1) for trees that hold contiguous sets of points (the BinarySpaceTree does this), you could avoid making `newIndexes` and actually just pass a first and last index to the matrix (or perhaps even the result of a `.cols()` call to the dataset held by the tree).
 2) one could make a batch call to `kernel.Evaluate()` on the entire distances vector instead of individually calling it for each element.

However, it seems to me that the overhead of doing so way outweighs any speedup we might see, given your initial results.  We'd need to (a) create infrastructure for every tree so that we can get the entire subset of points as a matrix... or as a list of indices... or either and then use `TreeTraits<>` to figure out which to use... (b) redo all the kernels so that they have a batch `Evaluate()`... (c) debug all the metaprogramming errors that might happen... (d) hope the new API makes sense... (e) hope it even gives any speedup.

If you'd seen big speedup, I might say awesome, let me try refactoring some things and let's make it work, but in this particular case I think the effort outweighs the potential speed gain (if you agree).  Thanks for investigating this! :+1:

robertohueso: I agree :) Thanks for proposing the idea.

src/mlpack/methods/kde/kde.hpp
```diff
@@ -28,6 +28,35 @@ enum KDEMode
...
+  static constexpr double mcProb = 0.95;
+
+  //! Initial sample size for Monte Carlo estimations.
+  static constexpr size_t initialSampleSize = 100;
+
+  //! Monte Carlo entry coefficient.
+  static constexpr double mcEntryCoef = 3;
+
+  //! Monte Carlo break coefficient.
+  static constexpr double mcBreakCoef = 0.7;
```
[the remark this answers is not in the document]
robertohueso: I've been doing some testing and this is an example of the KDE computation time (in seconds) of one of the tests:

```
Execution parameters:
abs_error: 0
algorithm: dual-tree
initial_sample_size: 100
kernel: gaussian
mc_entry_coef: 3
mc_probability: 0.95
monte_carlo: true
rel_error: 0.05
tree: kd-tree
```

| Bandwidth\mcBreakCoef | 0.2      | 0.3      | 0.4      | 0.5      | 0.6       | 0.7      | 0.8       |
|----------------------:|----------|----------|----------|----------|-----------|----------|-----------|
|                   0.5 | 6.789707 | 6.266044 | 5.977942 | 6.204131 | 6.275582  | 6.391213 | 6.270569  |
|                   0.8 | 5.146516 | 4.916502 | 6.405031 | 9.853159 | 11.119784 | 2.484530 | 13.522522 |
|                   1.0 | 5.859346 | 1.235331 | 1.224885 | 1.249122 | 1.269203  | 1.254212 | 1.280066  |

This is highly dependant on other parameters, but it seems to follow a similar pattern across different datasets, so a good default would be either 0.3 or 0.4 so let's choose 0.4 :)

robertohueso:
Thanks for the last review! I think everything is now solved and ready for review :)

src/mlpack/methods/kde/kde.hpp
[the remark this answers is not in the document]
rcurtin: Agreed. :+1: Thanks for doing the testing on this one to figure out what is best.

rcurtin:
@mlpack-jenkins build this please

rcurtin:

rcurtin:
Agreed, this all seems ready to me except for the static code analysis error.  This is the one about the `RectangleTree` copy constructor.  Did you want me to provide a quick patch for that?

Any of the failing tests on Travis appear to be in ANNLayerTest, so #1953 is trying to solve that but we don't need to consider it here.

robertohueso:
If it doesn't take too much time for you, it would be nice. Otherwise I can also try to fix it :)

rcurtin:
:+1: let me see if I can do it in a moment here once I handle a few other emails.

rcurtin:
https://gist.github.com/d8fd23ded2bda55277689265362c224f should handle the `RectangleTree` fixes.  Let me know if you see anything that can be improved. :+1:  Otherwise I think that this is ready, let's just make sure there are no CI problems before approval.

src/mlpack/tests/knn_test.cpp
```diff
@@ -245,6 +245,7 @@ BOOST_AUTO_TEST_CASE(TrainTreeTest)
   BOOST_REQUIRE_EQUAL(neighbors.n_cols, baselineNeighbors.n_cols);
   BOOST_REQUIRE_EQUAL(distances.n_rows, baselineDistances.n_rows);
   BOOST_REQUIRE_EQUAL(distances.n_cols, baselineDistances.n_cols);
+  BOOST_REQUIRE_EQUAL(oldFromNewReferences.size(), distances.n_cols);
```
rcurtin: It's a little irritating but I guess to fix the static code analysis issue we can replace `i < distances.n_cols` with `i < oldFromNewReferences.size()`, since those should be the same anyway.

robertohueso: It is :/ It got fixed with f35b22d38f3e94a16011e9caa1c5a6aaa849db39.

rcurtin:
Awesome, happy to have this merged. :+1:

PR #1935, opened 2019-06-21
Add CLI and python bindngs for linear_svm

Yashwants19:
Address #1633


Yashwants19:
Hi @rcurtin, @zoq : This is ready for review.

src/mlpack/methods/linear_svm/linear_svm_main.cpp
```diff
@@ -0,0 +1,417 @@
...
+PARAM_MATRIX_IN("test", "Matrix containing test dataset.", "T");
+PARAM_UROW_IN("test_labels", "Matrix containing test labels.", "A");
+PARAM_UROW_OUT("predictions", "If test data is specified, this matrix is where "
+    "the predictions for the test set will be saved.", "P");
+PARAM_MATRIX_OUT("score", "If test data is specified, this "
+    "matrix is where the class score for the test set will be saved.",
+    "p");
+
+size_t NumberOfClasses(const size_t numClasses,
+                                const arma::Row<size_t>& labels);
```
zoq: Picky style comment, the second parameter should align with the first one.

src/mlpack/methods/linear_svm/linear_svm_main.cpp
```diff
@@ -0,0 +1,417 @@
...
+    #endif
+  }
+  if (CLI::HasParam("test"))
+  {
+    // Get the test dataset, and get predictions.
+    testSet = std::move(CLI::GetParam<arma::mat>("test"));
+    arma::Row<size_t> predictions;
+    size_t trainingDimensionality;
+
+    // Set the dimensionality according to fitintercept
```
zoq: Missing stop at the end.

src/mlpack/methods/linear_svm/linear_svm_main.cpp
```diff
@@ -0,0 +1,417 @@
...
+          << CLI::GetPrintableParam<arma::mat>("test") << "'." << endl;
+      CLI::GetParam<arma::Row<size_t>>("predictions") = std::move(predictions);
+    }
+  }
+
+  CLI::GetParam<LinearSVM<>*>("output_model") = model;
+}
+
+size_t NumberOfClasses(const size_t numClasses,
+                                const arma::Row<size_t>& labels)
```
zoq: See comment above.

src/mlpack/methods/linear_svm/linear_svm_main.cpp
Yashwants19: I have made the changes as suggested :).

src/mlpack/methods/linear_svm/linear_svm_main.cpp
Yashwants19: Updated.

src/mlpack/methods/linear_svm/linear_svm_main.cpp
Yashwants19: Done.

src/mlpack/methods/linear_svm/linear_svm.hpp
```diff
@@ -119,6 +119,20 @@ class LinearSVM
...
+   * Initialize the Linear SVM without performing training.  Default
+   * value of lambda is 0.0001.  Be sure to use Train() before calling
+   * Classify() or ComputeAccuracy(), otherwise the results may be meaningless.
+   *
+   * @param numClasses Number of classes for classification.
+   * @param lambda L2-regularization constant.
+   * @paran delta Margin of difference between correct class and other classes.
+   * @param fitIntercept add intercept term or not.
+   */
+  LinearSVM(const size_t numClasses = 0,
```
rcurtin: Instead of saying `be sure to use Train() before calling Classify() or ComputeAccuracy()`, it seems to me like we could just check in those functions whether or not the model has 0 classes, and issue an error if so.

src/mlpack/methods/linear_svm/linear_svm_main.cpp
```diff
@@ -0,0 +1,417 @@
...
+#include "linear_svm.hpp"
+
+#include <ensmallen.hpp>
+
+using namespace std;
+using namespace mlpack;
+using namespace mlpack::svm;
+using namespace mlpack::util;
+
+PROGRAM_INFO("LinearSVM is an L2-regularized support vector machine model",
```
rcurtin: The first argument is just the name of the program, not a full sentence, so `L2-regularized linear support vector machine` or something like this would be fine.

src/mlpack/methods/linear_svm/linear_svm_main.cpp
```diff
@@ -0,0 +1,417 @@
...
+    // Short description.
+    "An implementation of LinearSVM for multiclass classification. "
+    "Given labeled data, a model can be trained and saved for "
+    "future use; or, a pre-trained model can be used to classify new points.",
+    // Long description.
+    "An implementation of LinearSVM using either the "
```
rcurtin: I would suggest changing `An implementation of LinearSVM using` to `An implementation of linear SVMs that uses either L-BFGS or parallel SGD (stochastic gradient descent) to train the model.`, since the optimizers are only used for training. :+1:

Also, note that the typical user won't be familiar with the internal C++ classes like `LinearSVM` or `ParallelSGD`, so maybe it is better to use their "logical names", i.e. `linear SVM` and `parallel SGD`.

src/mlpack/methods/linear_svm/linear_svm_main.cpp
```diff
@@ -0,0 +1,417 @@
...
+    " The trained LinearSVM model may be saved using the " +
+    PRINT_PARAM_STRING("output_model") + " output parameter."
+    "\n\n"
+    "The training data, if specified, may have class labels as its last "
+    "dimension.  Alternately, the " + PRINT_PARAM_STRING("labels") + " "
+    "parameter may be used to specify a separate matrix of labels."
+    "\n\n"
+    "When a model is being trained, there are many options.  L2 regularization "
+    "(to prevent overfitting) can be specified with the " +
+    PRINT_PARAM_STRING("lambda") + " option, and The number of classes can be "
```
rcurtin: `The` -> `the`

src/mlpack/methods/linear_svm/linear_svm_main.cpp
```diff
@@ -0,0 +1,417 @@
...
+    "manually specified with the " + PRINT_PARAM_STRING("number_of_classes") +
+    "and if an intercept term is not desired in the model, the " +
+    PRINT_PARAM_STRING("no_intercept") + " parameter can be specified."
+    "Margin of difference between correct class and other classes can "
+    "be specified with the" + PRINT_PARAM_STRING("delta") + "option."
+    "The optimizer used to train the model can be specified with the " +
+    PRINT_PARAM_STRING("optimizer") + " parameter.  Available options are "
+    "'psgd' (stochastic gradient descent) and 'lbfgs' (the L-BFGS optimizer).  "
```
rcurtin: It might be worth writing `parallel stochastic gradient descent` not just `stochastic gradient descent` here since there are lots of variations of SGD.

src/mlpack/methods/linear_svm/linear_svm_main.cpp
```diff
@@ -0,0 +1,417 @@
...
+    "number of allowed iterations, and the " +
+    PRINT_PARAM_STRING("tolerance") + " parameter specifies the tolerance for "
+    "convergence.  For the ParallelSGD optimizer, the " +
+    PRINT_PARAM_STRING("step_size") + " parameter controls the step size taken "
+    "at each iteration by the optimizer. If the "
+    "objective function for your data is oscillating between Inf and 0, the "
+    "step size is probably too large.  There are more parameters for the "
+    "optimizers, but the C++ interface must be used to access these."
+    "\n\n"
+    "We can only use ParallelSGD when OPENMP is specified.  "
```
rcurtin: Is this true?  I think that parallel SGD will still work fine when OpenMP isn't available, it will just use one core.  Correct me if I'm wrong.

src/mlpack/methods/linear_svm/linear_svm_main.cpp
```diff
@@ -0,0 +1,417 @@
...
+    "For ParallelSGD, an iteration refers to a single point. So to take"
+    " a single pass over the dataset with ParallelSGD, " +
+    PRINT_PARAM_STRING("max_iterations") +
+    " should be set to the number of points in the dataset."
```
rcurtin: Most users will expect, when using an SGD-like optimizer, to instead just specify the number of passes over the data.  So, I would suggest that instead of doing one-iteration-per-point here, either we add an `epochs` option which is used for parallel SGD instead of `max_iterations` (meaning `max_iterations` is only used for L-BFGS), or do some processing so that `max_iterations` refers to the number of passes when SGD is used.

src/mlpack/methods/linear_svm/linear_svm_main.cpp
```diff
@@ -0,0 +1,417 @@
...
+    "'psgd').", "O", "lbfgs");
+PARAM_DOUBLE_IN("tolerance", "Convergence tolerance for optimizer.", "e",
+    1e-10);
+PARAM_INT_IN("max_iterations", "Maximum iterations for optimizer (0 indicates "
+    "no limit).", "n", 10000);
+#ifdef HAS_OPENMP
+PARAM_DOUBLE_IN("step_size", "Step size for ParallelSGD optimizer.", "s", 0.01);
+PARAM_FLAG("shuffle", "Don't shuffle the order in which data points are "
+    "visited for ParallelSGD.", "S");
+#endif
```
rcurtin: I think it should be okay to enable these even if `HAS_OPENMP` is not set.  If we use preprocessor macros to enable or disable some options, it actually makes the documentation really hard for the website, since we would need some way to clearly specify which conditions are needed for a certain option to exist (which we currently don't have, and preferably I'd like to avoid creating :)).

src/mlpack/methods/linear_svm/linear_svm_main.cpp
```diff
@@ -0,0 +1,417 @@
...
+
+size_t NumberOfClasses(const size_t numClasses,
+                       const arma::Row<size_t>& labels);
+
+static void mlpackMain()
+{
+  // Collect command-line options.
+  const double lambda = CLI::GetParam<double>("lambda");
+  const double delta = CLI::GetParam<double>("delta");
+  const bool intercept = CLI::HasParam("no_intercept") ? true : false;
```
rcurtin: No need for the `? true : false` here; `HasParam()` already returns a `bool`.

src/mlpack/methods/linear_svm/linear_svm_main.cpp
```diff
@@ -0,0 +1,417 @@
...
+      true, "lambda must be positive or zero");
+
+  // Number of Classes must be Non-Negative
+  RequireParamValue<int>("number_of_classes", [](int x) { return x >= 0; },
+                         true, "number of classes must be greater than or "
+                         "equal to 0 (equal to 0 in case of unspecified.)");
+
+  // Delta must be positive.
+  RequireParamValue<double>("delta", [](double x) { return x >= 0.0; }, true,
+      "Margin of difference between correct class and other classes");
```
rcurtin: I think we need an actual error here, maybe adding ` must be nonnegative` to the existing message is sufficient.

src/mlpack/methods/linear_svm/linear_svm_main.cpp
```diff
@@ -0,0 +1,417 @@
...
+    }
+
+    // The initial predictors for y, Nx1.
+    labels = arma::conv_to<arma::Row<size_t>>::from(
+        trainingSet.row(trainingSet.n_rows - 1));
+    trainingSet.shed_row(trainingSet.n_rows - 1);
+  }
+
+  const size_t numClasses = NumberOfClasses(
+       (size_t) CLI::GetParam<int>("number_of_classes"), labels);
```
rcurtin: I see that the `NumberOfClasses()` method is a way of counting the unique labels in the dataset.  That nicely handles the situation where the user passes the labels, e.g., `1, 2, 3` instead of `0, 1, 2`.  But, the `LinearSVM` class actually requires that we have labels from `0` to `numClasses - 1`.  Therefore, regardless of the labels we get, we should use `data::NormalizeLabels()` in order to convert what we got to the range `0` to `numClasses - 1`.  As a nice benefit, after doing that, we can get `numClasses` as `CLI::GetParam<int>("number_of_classes") == 0 ? arma::max(labels) + 1 : CLI::GetParam<int>("number_of_classes")` or something like this.

src/mlpack/methods/linear_svm/linear_svm_main.cpp
```diff
@@ -0,0 +1,417 @@
...
+
+  // Load the model, if necessary.
+  LinearSVM<>* model;
+  if (CLI::HasParam("input_model"))
+    model = CLI::GetParam<LinearSVM<>*>("input_model");
+  else
+  {
+    model = new LinearSVM<>;
+  }
```
rcurtin: If we use braces for the `else`, we should also use braces for the `if` (or, we should use braces for neither).  Sorry for the picky style comment. :)

src/mlpack/methods/linear_svm/linear_svm_main.cpp
```diff
@@ -0,0 +1,417 @@
...
+      CLI::GetParam<arma::mat>("score") = std::move(score);
+    }
+
+    model->Classify(testSet, predictions);
+
+    // Calculate accuracy, if desired.
+    if (CLI::HasParam("test_labels"))
+    {
+      arma::Row<size_t> testLabels =
+        std::move(CLI::GetParam<arma::Row<size_t>>("test_labels"));
```
rcurtin: This line should be indented twice. :+1:

src/mlpack/methods/linear_svm/linear_svm_main.cpp
```diff
@@ -0,0 +1,417 @@
...
+
+      if (testSet.n_cols != testLabels.n_elem)
+      {
+        Log::Fatal << "Test data given with " << PRINT_PARAM_STRING("test")
+            << " has " << testSet.n_cols << " points, but labels in "
+            << PRINT_PARAM_STRING("test_labels") << " have "
+            << testLabels.n_elem << " labels!" << endl;
+      }
+
+      vector<size_t> bingoLabels(numClasses, 0);
```
rcurtin: Maybe it's clearer to call this `correctClassCounts` instead of `bingoLabels`?  Up to you.

src/mlpack/tests/main_tests/linear_svm_test.cpp
```diff
@@ -0,0 +1,893 @@
...
+
+  SetInputParam("training", std::move(trainData));
+  SetInputParam("labels", std::move(trainLabels));
+  SetInputParam("test", std::move(testData));
+
+  // Training the model.
+  mlpackMain();
+
+  // Get the output predictions of the test data.
+  const arma::Row<size_t> &testLabels =
```
rcurtin: This should be `arma::Row<size_t>& testLabels`, not `arma::Row<size_t> &testLabels`.

src/mlpack/tests/main_tests/linear_svm_test.cpp
```diff
@@ -0,0 +1,893 @@
...
+ **/
+BOOST_AUTO_TEST_CASE(LinearSVMCheckDimOfTestData)
+{
+  arma::mat trainData;
+  if (!data::Load("iris.csv", trainData))
+    BOOST_FAIL("Cannot load test dataset iris.csv!");
+
+  arma::mat testData;
+  if (!data::Load("iris_test.csv", testData))
+    BOOST_FAIL("Cannot load test dataset iris_test.csv!");
```
rcurtin: I think the dimensionalities are actually the same here for these files, but the key is that you're not specifying labels, so `mlpackMain()` assumes the last dimension is the labels, thus causing the model to be trained with `trainData.n_rows - 1` rows.  That isn't obvious from just looking at the code here---can you add a comment to point this out?

src/mlpack/tests/main_tests/linear_svm_test.cpp
```diff
@@ -0,0 +1,893 @@
...
+  SetInputParam("optimizer", std::string("hello"));
+
+  // Tolerance is negative. It should throw a runtime error.
+  Log::Fatal.ignoreInput = true;
+  BOOST_REQUIRE_THROW(mlpackMain(), std::runtime_error);
+  Log::Fatal.ignoreInput = false;
+}
+
+/**
+  * Ensuring changing Maximum number of iterations s the output model.
```
rcurtin: Should the `s` here be `changes`?

src/mlpack/tests/main_tests/linear_svm_test.cpp
```diff
@@ -0,0 +1,893 @@
...
+                ->Parameters());
+
+  // Reset the settings.
+  bindings::tests::CleanMemory();
+  CLI::ClearSettings();
+  CLI::RestoreSettings(testName);
+
+  SetInputParam("training", std::move(trainData));
+  SetInputParam("labels", std::move(trainLabels));
+  SetInputParam("number_of_classes", int(2));
```
rcurtin: I think this test is unsafe---if we specify `number_of_classes` less than the actual number of classes, then there are likely to be invalid memory accesses.  On the other hand, if it is specified to be *more* than the actual number of classes, it should work fine without invalid memory accesses, but the parameters associated with the extra classes may take any value.

src/mlpack/tests/main_tests/linear_svm_test.cpp
```diff
@@ -0,0 +1,893 @@
...
+
+  // Both solutions should be not equal.
+  CheckMatricesNotEqual(parameters1, parameters2);
+}
+
+/**
+ * The test is only compiled if the user has specified OpenMP to be
+ * used.
+ */
+#ifdef HAS_OPENMP
```
rcurtin: We should still be able to run this test without OpenMP.

src/mlpack/tests/main_tests/linear_svm_test.cpp
```diff
@@ -0,0 +1,893 @@
...
+  arma::Row<size_t> trainLabels = "1 0 1";
+
+  SetInputParam("training", trainData);
+  SetInputParam("labels", trainLabels);
+  SetInputParam("optimizer", std::string("psgd"));
+  SetInputParam("number_of_classes", int(2));
+  SetInputParam("shuffle", bool(false));
+
+  // First solution.
+  mlpackMain();
```
rcurtin: Note that, for all `psgd` tests, there are two issues that will give different results when they aren't expected:

 1. The random seed.  I think there is no option for `seed` in the program, but we should add one and use `FixedRandomSeed()` for these tests, as parallel SGD may depend on random components (primarily when `shuffle = true`).

 2. The number of cores.  Since parallel SGD allows individual updates in potentially different orders, we should restrict the tests to run on one core only for these tests that are checking if something is different between two runs.

src/mlpack/methods/linear_svm/linear_svm.hpp
Yashwants19: I have added this part. Please take a look.
Sorry for the delay.

src/mlpack/methods/linear_svm/linear_svm_main.cpp
Yashwants19: Updated.

src/mlpack/methods/linear_svm/linear_svm_main.cpp
Yashwants19: Updated.

src/mlpack/methods/linear_svm/linear_svm_main.cpp
Yashwants19: Please check about the number of cores. I m pretty new in OPENMP.

src/mlpack/methods/linear_svm/linear_svm_main.cpp
Yashwants19: Added new parameter `epochs`.

src/mlpack/methods/linear_svm/linear_svm_main.cpp
Yashwants19: Done.

src/mlpack/methods/linear_svm/linear_svm_main.cpp
Yashwants19: Done

src/mlpack/tests/main_tests/linear_svm_test.cpp
Yashwants19: Add a comment for this.

src/mlpack/tests/main_tests/linear_svm_test.cpp
Yashwants19: Removed.

src/mlpack/tests/main_tests/linear_svm_test.cpp
Yashwants19: I have made the changes as suggested. Is this approach is correct for one core?

Yashwants19:
Hi @rcurtin This is ready for review. :)

src/mlpack/methods/linear_svm/linear_svm_impl.hpp
```diff
@@ -48,10 +48,35 @@ LinearSVM<MatType>::LinearSVM(
     delta(delta),
     fitIntercept(fitIntercept)
 {
+  if (numClasses == 1)
```
rcurtin: Shouldn't this be `<= 1` not `== 1`?  Further, we should put the check in `Train()`, since it is valid to create a `LinearSVM` object without training it.

src/mlpack/methods/linear_svm/linear_svm_impl.hpp
```diff
@@ -48,10 +48,35 @@ LinearSVM<MatType>::LinearSVM(
...
+  {
+    std::ostringstream oss;
+    oss << "LinearSVM dataset has 0 number of classes!";
```
rcurtin: No need for an `ostringstream` here since the error message is just a plain string.

src/mlpack/methods/linear_svm/linear_svm_main.cpp
```diff
@@ -0,0 +1,419 @@
...
+    "those things at once.  In addition, this program allows classification on "
+    "a test dataset (specified with the " + PRINT_PARAM_STRING("test") + " "
+    "parameter) and the classification results may be saved with the " +
+    PRINT_PARAM_STRING("predictions") + " output parameter."
+    " The trained linear SVM model may be saved using the " +
+    PRINT_PARAM_STRING("output_model") + " output parameter."
+    "\n\n"
+    "The training data, if specified, may have class labels as its last "
+    "dimension.  Alternately, the " + PRINT_PARAM_STRING("labels") + " "
+    "parameter may be used to specify a separate matrix of labels."
```
rcurtin: Should that be `vector` not `matrix`?

src/mlpack/methods/linear_svm/linear_svm_main.cpp
```diff
@@ -0,0 +1,419 @@
...
+    "\n\n"
+    "When a model is being trained, there are many options.  L2 regularization "
+    "(to prevent overfitting) can be specified with the " +
+    PRINT_PARAM_STRING("lambda") + " option, and the number of classes can be "
+    "manually specified with the " + PRINT_PARAM_STRING("number_of_classes") +
+    "and if an intercept term is not desired in the model, the " +
+    PRINT_PARAM_STRING("no_intercept") + " parameter can be specified."
+    "Margin of difference between correct class and other classes can "
+    "be specified with the" + PRINT_PARAM_STRING("delta") + "option."
```
rcurtin: Don't forget spaces here: `...with the "` and `" option."`. :)

src/mlpack/methods/linear_svm/linear_svm_main.cpp
```diff
@@ -0,0 +1,419 @@
...
+    "of predictors, X).", "t");
+PARAM_UROW_IN("labels", "A matrix containing labels (0 or 1) for the points "
+    "in the training set (y).", "l");
+
+// Optimizer parameters.
+PARAM_DOUBLE_IN("lambda", "L2-regularization parameter for training.", "L",
+    0.0001);
+PARAM_DOUBLE_IN("delta", "Margin of difference between correct class and other "
+    "classes.", "d", 1.0);
+PARAM_INT_IN("number_of_classes", "Number of classes for classification; if "
```
rcurtin: I think for consistency with the rest of the codebase, we should call this `num_classes`.

src/mlpack/methods/linear_svm/linear_svm_main.cpp
```diff
@@ -0,0 +1,419 @@
...
+  RequireAtLeastOnePassed({ "output_model", "predictions", "score"},
+      false, "no output will be saved");
+
+  ReportIgnoredParam({{ "test", false }}, "predictions");
+  ReportIgnoredParam({{ "test", false }}, "score");
+  ReportIgnoredParam({{ "test", false }}, "test_labels");
+
+  // Max Iterations needs to be positive.
+  RequireParamValue<int>("max_iterations", [](int x) { return x >= 0; },
+      true, "max_iterations must be positive or zero");
```
rcurtin: We can just use `non-negative` in place of `positive of zero`. :+1:

src/mlpack/methods/linear_svm/linear_svm_main.cpp
```diff
@@ -0,0 +1,419 @@
...
+    // The initial predictors for y, Nx1.
+    rawLabels = arma::conv_to<arma::Row<size_t>>::from(
+        trainingSet.row(trainingSet.n_rows - 1));
+    trainingSet.shed_row(trainingSet.n_rows - 1);
+  }
+
+  data::NormalizeLabels(rawLabels, labels, mappings);
+
+  // Load the model, if necessary.
+  LinearSVM<>* model;
```
rcurtin: The use of `NormalizeLabels` makes this a little bit tricky.  Consider the following use of the program:

1) a user trains a model passing in data with labels `1, 2, 3` (these will be turned into `0, 1, 2` by `NormalizeLabels()`)
2) now a user saves the model and uses it in a separate call to get some predictions...

but those predictions have labels `0, 1, 2`, not `1, 2, 3`, so they don't match the original labels.  In addition, if the user uses a test set and `test_labels`, the test labels are not mapped in the same way.

So, I think it may be necessary to create a `LinearSVMModel` that also holds the mappings, and then applies the mappings from any saved model to any labels (input or output).

Let me know if I can clarify what I mean. :+1:

src/mlpack/methods/linear_svm/linear_svm_main.cpp
```diff
@@ -0,0 +1,419 @@
...
+    }
+    else if (optimizerType == "psgd")
+    {
+      const double step_size = CLI::GetParam<double>("step_size");
+      const bool shuffle = !CLI::HasParam("shuffle");
+      const size_t maxIt = epochs * trainingSet.n_cols;
+
+      ens::ConstantStep decayPolicy(step_size);
+
+      #ifndef HAS_OPENMP
```
rcurtin: There shouldn't be a need for this check here; regardless of whether or not OpenMP is available, `ens::ParallelSGD` is still available and will still work (even with one thread).  It may be worthwhile to print a warning if OpenMP is not in use and parallel SGD was chosen, like `Log::Warn << "Using parallel SGD, but OpenMP support is not available!" << std::endl;` or something.

src/mlpack/methods/linear_svm/linear_svm_main.cpp
```diff
@@ -0,0 +1,419 @@
...
+      ens::ParallelSGD<ens::ConstantStep> psgdOpt(maxIt, trainingSet.n_cols,
+        tolerance, shuffle, decayPolicy);
+      #endif
+
+      // This optimizer is only compiled if OpenMP is used.
+      #ifdef HAS_OPENMP
+      ens::ParallelSGD<ens::ConstantStep> psgdOpt(maxIt, std::ceil(
+        (float)trainingSet.n_cols / omp_get_max_threads()), tolerance, shuffle,
```
rcurtin: There should be a space between the `(float)` and `trainingSet.n_cols`.  For the thread share size, perhaps you can do it like this:

```
#ifdef HAS_OPENMP
size_t threads = omp_get_max_threads();
#else
size_t threads = 1;
#endif
```

src/mlpack/methods/linear_svm/linear_svm_main.cpp
```diff
@@ -0,0 +1,419 @@
...
+    {
+      Log::Fatal << "Test data dimensionality (" << testSet.n_rows << ") must "
+          << "be the same as the dimensionality of the training data ("
+          << trainingDimensionality << ")!" << endl;
+    }
+
+    // Save class score, if desired.
+    if (CLI::HasParam("score"))
+    {
+      Log::Info << "Calculating class score of points in '"
```
rcurtin: No need for a `'` here around the output of `GetPrintableParam()`.  In the cases where a filename is printed, I think it is already enclosed in `''`s, and when it isn't, it just prints the size of the matrix, which looks a bit odd in quotes. :)

src/mlpack/tests/main_tests/linear_svm_test.cpp
```diff
@@ -0,0 +1,866 @@
+/**
+  * @file logistic_regression_test.cpp
```
rcurtin: Incorrect filename. :)

src/mlpack/tests/main_tests/linear_svm_test.cpp
```diff
@@ -0,0 +1,866 @@
...
+#include <mlpack/methods/linear_svm/linear_svm_main.cpp>
+#include <mlpack/core/util/mlpack_main.hpp>
+#include "test_helper.hpp"
+
+#include <boost/test/unit_test.hpp>
+#include "../test_tools.hpp"
+
+using namespace mlpack;
+
+
```
rcurtin: There's an extra line here---we can remove it.

src/mlpack/tests/main_tests/linear_svm_test.cpp
```diff
@@ -0,0 +1,866 @@
...
+  const arma::Row<size_t>& testLabels =
+      CLI::GetParam<arma::Row<size_t>>("predictions");
+
+  // Output predictions size must match the test data set size.
+  BOOST_REQUIRE_EQUAL(testLabels.n_rows, 1);
+  BOOST_REQUIRE_EQUAL(testLabels.n_cols, testSize);
+}
+
+/**
+  * Ensuring that the labels size is checked.
```
rcurtin: This is a really pedantic comment (sorry about that) but the line should only be indented two spaces, not three. :+1:

src/mlpack/methods/linear_svm/linear_svm_main.cpp
```diff
@@ -0,0 +1,419 @@
...
+      lbfgsOpt.MinGradientNorm() = tolerance;
+
+      Log::Info << "Training model with L-BFGS optimizer." << endl;
+
+      // This will train the model.
+      model->Train(trainingSet, labels, numClasses, lbfgsOpt);
+    }
+    else if (optimizerType == "psgd")
+    {
+      const double step_size = CLI::GetParam<double>("step_size");
```
rcurtin: The variable should be named `stepSize` here (no underscores).

src/mlpack/tests/main_tests/linear_svm_test.cpp
```diff
@@ -0,0 +1,866 @@
...
+BOOST_AUTO_TEST_CASE(LinearSVMLabelsRepresentationTest)
+{
+  arma::mat trainData1({{1.0, 2.0, 3.0}, {1.0, 4.0, 9.0}, {0, 1, 1}});
+  arma::mat testData({{4.0, 5.0}, {1.0, 6.0}});
+
+  SetInputParam("training", std::move(trainData1));
+  SetInputParam("test", testData);
+
+  // The first solution.
+  mlpackMain();
```
rcurtin: Since the initial point for the optimization is random, should we set a fixed random seed for both calls here?

src/mlpack/tests/main_tests/linear_svm_test.cpp
rcurtin: Thanks. :+1:

src/mlpack/tests/main_tests/linear_svm_test.cpp
```diff
@@ -0,0 +1,866 @@
...
+  // Both solutions should be not equal.
+  CheckMatricesNotEqual(parameters1, parameters2);
+}
+
+/**
+ * The test is only compiled if the user has not specified OpenMP to be
+ * used and we test these case in one core.
+ */
+#ifndef HAS_OPENMP
+/**
```
rcurtin: It shouldn't be necessary to check for OpenMP here---the test should work both with and without OpenMP.

src/mlpack/tests/main_tests/linear_svm_test.cpp
rcurtin: Yep, this should work fine.  But the last thing to do is ensure that the test only runs on one core. :+1:

src/mlpack/tests/main_tests/linear_svm_test.cpp
```diff
@@ -0,0 +1,866 @@
...
+  SetInputParam("epochs", int(10));
+
+  // First solution.
+  mlpack::math::FixedRandomSeed();
+  mlpackMain();
+
+  // Get the parameters of the output model obtained after first training.
+  const arma::mat parameters1 =
+      std::move(CLI::GetParam<LinearSVM<>*>("output_model")
+                ->Parameters());
```
rcurtin: This could probably look a little cleaner like this:

```
const arma::mat parameters1 = std::move(
    CLI::GetParam<LinearSVM<>*>("output_model")->Parameters();
```

I think that will fit in 80 characters... (not 100% sure though)

src/mlpack/methods/linear_svm/linear_svm_main.cpp
```diff
@@ -0,0 +1,419 @@
...
+#include "linear_svm.hpp"
+
+#include <ensmallen.hpp>
+
+using namespace std;
+using namespace mlpack;
+using namespace mlpack::svm;
+using namespace mlpack::util;
+
+PROGRAM_INFO("Linear SVM is an L2-regularized support vector machine",
```
rcurtin: This is better, but the `is an` makes it more of a sentence than a title.  I might suggest `Linear SVM: L2-regularized support vector machines`.

src/mlpack/methods/linear_svm/linear_svm_main.cpp
rcurtin: Yeah, as far as I can tell, it should work just fine without OpenMP, and it will only use one core.

src/mlpack/methods/linear_svm/linear_svm_main.cpp
```diff
@@ -0,0 +1,419 @@
...
+    }
+    if (CLI::HasParam("shuffle"))
+    {
+      Log::Warn << PRINT_PARAM_STRING("shuffle") << " ignored because "
+          << "optimizer type is not 'psgd'." << std::endl;
+    }
+    if (CLI::HasParam("epochs"))
+    {
+      Log::Warn << PRINT_PARAM_STRING("epochs") << " ignored because "
+          << "optimizer type is not 'psgd'." << std::endl;
```
rcurtin: Similarly, if `max_iterations` is specified but the optimizer is `psgd`, we should issue a warning.

Yashwants19: I have added this.

src/mlpack/methods/linear_svm/linear_svm_main.cpp
Yashwants19: Done

src/mlpack/tests/main_tests/linear_svm_test.cpp
Yashwants19: Updated

src/mlpack/tests/main_tests/linear_svm_test.cpp
Yashwants19: Removed this. Without OPEN_MP how can we ensure that it must run on one core because this OPEN_MP it will go `omp_get_max_threads()`.

src/mlpack/tests/main_tests/linear_svm_test.cpp
Yashwants19: Added

src/mlpack/methods/linear_svm/linear_svm_main.cpp
Yashwants19: Updated.

src/mlpack/tests/main_tests/linear_svm_test.cpp
Yashwants19: Resolved. :)

src/mlpack/tests/main_tests/linear_svm_test.cpp
Yashwants19: Sorry about that.

src/mlpack/methods/linear_svm/linear_svm_main.cpp
Yashwants19: Removed extra `'`. :)

src/mlpack/methods/linear_svm/linear_svm_main.cpp
Yashwants19: Added This.

src/mlpack/methods/linear_svm/linear_svm_main.cpp
Yashwants19: Removed.

src/mlpack/methods/linear_svm/linear_svm_main.cpp
Yashwants19: I have added a `LinearSVMModel` please take a look.

src/mlpack/methods/linear_svm/linear_svm_main.cpp
Yashwants19: Updated.

src/mlpack/methods/linear_svm/linear_svm_main.cpp
Yashwants19: Updated.

src/mlpack/methods/linear_svm/linear_svm_main.cpp
Yashwants19: Resolved.

src/mlpack/methods/linear_svm/linear_svm_impl.hpp
Yashwants19: Updated Please take a look. :)

src/mlpack/methods/linear_svm/linear_svm_impl.hpp
Yashwants19: Resolved and added this part in `Train()` also.

src/mlpack/tests/main_tests/linear_svm_test.cpp
Yashwants19: Hi @rcurtin I made the changes as suggested.

src/mlpack/methods/linear_svm/linear_svm_impl.hpp
```diff
@@ -48,10 +48,29 @@ LinearSVM<MatType>::LinearSVM(
     delta(delta),
     fitIntercept(fitIntercept)
 {
+  if (numClasses <= 1)
+  {
+    std::string error = "LinearSVM dataset has 0 number of classes!";
```
rcurtin: Actually you could just put the string error directly into the constructor for `std::invalid_argument`. :)

In addition, it's valid to create a `LinearSVM` object without training it, so this check should only be done in `Train()`.  Maybe it is better to wait to initialize the weights until `Train()` is called too.

src/mlpack/methods/linear_svm/linear_svm_main.cpp
```diff
@@ -0,0 +1,440 @@
...
+    "Optionally, the model can be used to predict the labels for another "
+    "matrix of data points, if " + PRINT_PARAM_STRING("test") + " is "
+    "specified.  The " + PRINT_PARAM_STRING("test") + " parameter can be "
+    "specified without the " + PRINT_PARAM_STRING("training") + " parameter, "
+    "so long as an existing linear SVM model is given with the " +
+    PRINT_PARAM_STRING("input_model") + " parameter.  The output predictions "
+    "from the linear SVM model may be saved with the " +
+    PRINT_PARAM_STRING("predictions") + " parameter." +
+    "\n\n"
+    "As an example, to train a LinaerSVM model on the data '" +
```
rcurtin: Oops, `LinearSVM`. :)

src/mlpack/methods/linear_svm/linear_svm_main.cpp
```diff
@@ -0,0 +1,440 @@
...
+  {
+    if (CLI::HasParam("max_iterations"))
+    {
+      Log::Warn << PRINT_PARAM_STRING("max_iterations") << " ignored because "
+          << "optimizer type is not 'lbfgs'." << std::endl;
+    }
+  }
+
+  // Step Size must be positive.
+  RequireParamValue<double>("step_size", [](double x) { return x >= 0.0; },
```
rcurtin: Should this be `>` instead of `>=`?

src/mlpack/methods/linear_svm/linear_svm_main.cpp
```diff
@@ -0,0 +1,440 @@
...
+  {
+    model = new LinearSVMModel();
+  }
+
+  // Now, do the training.
+  if (CLI::HasParam("training"))
+  {
+    data::NormalizeLabels(rawLabels, labels, model->mappings);
+    numClasses = CLI::GetParam<int>("num_classes") == 0 ?
+        arma::max(labels) + 1 : CLI::GetParam<int>("num_classes");
```
rcurtin: We can avoid the `arma::max(labels)` by using the mappings we just produced: `model->mappings.n_elem` should be equivalent.

src/mlpack/methods/linear_svm/linear_svm_main.cpp
```diff
@@ -0,0 +1,440 @@
...
+    {
+      numClasses = model->svm.NumClasses();
+    }
+    // Get the test dataset, and get predictions.
+    testSet = std::move(CLI::GetParam<arma::mat>("test"));
+    arma::Row<size_t> predictions;
+    size_t trainingDimensionality;
+
+    // Set the dimensionality according to fitintercept.
+    if (intercept && optimizerType == "lbfgs")
```
rcurtin: Why is this specific to L-BFGS being used?  Shouldn't we be able to fit the intercept for PSGD also?

src/mlpack/methods/linear_svm/linear_svm_main.cpp
```diff
@@ -0,0 +1,440 @@
...
+    "linear svm model.", "M");
+
+// Testing.
+PARAM_MATRIX_IN("test", "Matrix containing test dataset.", "T");
+PARAM_UROW_IN("test_labels", "Matrix containing test labels.", "A");
+PARAM_UROW_OUT("predictions", "If test data is specified, this matrix is where "
+    "the predictions for the test set will be saved.", "P");
+PARAM_MATRIX_OUT("score", "If test data is specified, this "
+    "matrix is where the class score for the test set will be saved.",
+    "p");
```
rcurtin: In other programs this is called `probabilities`; can we change the name of this to match?

src/mlpack/tests/main_tests/linear_svm_test.cpp
```diff
@@ -0,0 +1,853 @@
...
+                        "0 2 1;"
+                        "1 0 2;"
+                        "0 1 0";
+
+  arma::Row<size_t> trainLabels = "0 0 0";
+
+  SetInputParam("training", std::move(trainData));
+  SetInputParam("labels", std::move(trainLabels));
+
+  // Step size for optimizer is negative. It should throw a runtime error.
```
rcurtin: I think this comment is inaccurate. :)  This is true in a couple other places too.

src/mlpack/tests/main_tests/linear_svm_test.cpp
```diff
@@ -0,0 +1,853 @@
...
+  arma::Row<size_t> trainLabels;
+  if (!data::Load("iris_labels.txt", trainLabels))
+    BOOST_FAIL("Cannot load test dataset iris_labels.txt!");
+
+  SetInputParam("training", trainData);
+  SetInputParam("labels", trainLabels);
+  SetInputParam("lambda", double(0.001));
+
+  // First solution.
+  mlpackMain();
```
rcurtin: This test should use a fixed random seed also, since the random parameters will differ.  Correct me if I'm wrong but I believe this will apply to every test where we are checking for different results with two different training parameters.

src/mlpack/tests/main_tests/linear_svm_test.cpp
rcurtin: Probably what's needed is just an ifdef, and if OpenMP is available, then a function to set the number of threads to 1 can be called.

src/mlpack/tests/main_tests/linear_svm_test.cpp
```diff
@@ -0,0 +1,853 @@
...
+                        "0 1 0";
+
+  arma::Row<size_t> trainLabels = "1 0 1";
+
+  SetInputParam("training", trainData);
+  SetInputParam("labels", trainLabels);
+  SetInputParam("optimizer", std::string("lbfgs"));
+
+  // First solution.
+  mlpackMain();
```
rcurtin: Here, too, we will need a fixed random seed to ensure that both optimizers start from the same initial point.

rcurtin:
Also for the failing Travis test, I bet using smaller numbers of epochs can help.  Maybe it is converging too early.

src/mlpack/methods/linear_svm/linear_svm_main.cpp
Yashwants19: Yes this is strange but here we get inf and zero for prediction and scores when we use fit the intercept for psgd.

src/mlpack/methods/linear_svm/linear_svm_main.cpp
Yashwants19: Updated.

src/mlpack/tests/main_tests/linear_svm_test.cpp
Yashwants19: Updated all the comments :).

src/mlpack/tests/main_tests/linear_svm_test.cpp
Yashwants19: Done.

src/mlpack/tests/main_tests/linear_svm_test.cpp
Yashwants19: Done.

src/mlpack/methods/linear_svm/linear_svm_main.cpp
Yashwants19: Updated.

src/mlpack/methods/linear_svm/linear_svm_impl.hpp
Yashwants19: Updated.

src/mlpack/methods/linear_svm/linear_svm_main.cpp
Yashwants19: Done. :)

src/mlpack/methods/linear_svm/linear_svm_main.cpp
Yashwants19: Resolved.

src/mlpack/methods/linear_svm/linear_svm_main.cpp
```diff
@@ -0,0 +1,446 @@
...
+      const size_t maxIt = epochs * trainingSet.n_cols;
+
+      ens::ConstantStep decayPolicy(stepSize);
+
+      #ifdef HAS_OPENMP
+        #ifdef BUILD_TESTS
+        size_t threads = 1;
+        #else
+        size_t threads = omp_get_max_threads();
+        #endif
```
rcurtin: I see that the aim here is to use only one thread when we are testing, but this connects the testing code to the binding, and we should aim to keep these as separate as possible.  Why not just use `omp_set_num_threads(1);` in the tests when we need to specify only one thread?  I think that should work.

src/mlpack/methods/linear_svm/linear_svm_main.cpp
rcurtin: I see.  What happens when only one thread is used when we train with PSGD and fit the intercept?  I think I understand why this is happening, but need to know what happens in the singlethreaded case before I am sure.

Yashwants19:
Hi @rcurtin I am really sorry I am out of town right now. It will take around 1 week or less to resolve these comments.

Sorry

rcurtin:
@Yashwants19 no need to be sorry, there is no hurry at all! :)  Enjoy your trip.

src/mlpack/methods/linear_svm/linear_svm_main.cpp
Yashwants19: Sorry for the slow work.
I have updated this you can take a look. :)

Yashwants19:
I really don't initially when I was working with `psgd` and `intercept` it was not working that great.
But today when I tried with both single-core and multiple cores it worked.

Yashwants19:
Hi, @rcurtin When we enable `intercept` for both single and multiple cores it gives matrix incompatible error.

Yashwants19:
Hi @rcurtin this is ready for review.

Yashwants19:
Hi @rcurtin I think so we should finish this. 
If you have some time please review this.
Thank you

rcurtin:
Hey @Yashwants19 I totally agree, I have just been busy with other things and haven't really been able to look into this.  Basically, we need to figure out what is going on with parallel SGD and why this causes problems with the intercepts.  Once we can resolve that then I think it's ready to merge.  I hope that maybe I'll have some more time to look into this more comprehensively later this week.

rcurtin:
@Yashwants19 I'm not able to reproduce the error that you keep mentioning.  Can you give me details on how to reproduce it so I can debug it?  Thanks!

Yashwants19:
Hi @rcurtin as you can see in Travis Build psgd got some matrix incompatible error.

rcurtin:
I was able to debug this pretty easily with gdb.  Here's the diff to fix the issue:

```
diff --git a/src/mlpack/methods/linear_svm/linear_svm_function_impl.hpp b/src/mlpack/methods/linear_svm/linear_svm_function_impl.hpp
index 4e8a751ce..0656631c2 100644
--- a/src/mlpack/methods/linear_svm/linear_svm_function_impl.hpp
+++ b/src/mlpack/methods/linear_svm/linear_svm_function_impl.hpp
@@ -320,7 +320,7 @@ void LinearSVMFunction<MatType>::Gradient(
   {
     scores = parameters.rows(0, dataset.n_rows - 1).t()
         * dataset.cols(firstId, lastId)
-        + arma::repmat(parameters.row(dataset.n_rows).t(), 1, dataset.n_cols);
+        + arma::repmat(parameters.row(dataset.n_rows).t(), 1, batchSize);
   }
 
   arma::mat margin = scores - (arma::repmat(arma::ones(numClasses).t()
```

You can see that the `repmat` call incorrectly had `dataset.n_cols` as a parameter instead of `batchSize`.

Yashwants19:
You are great @rcurtin. Thanks for the help now this ready for complete review. Thank you :)

src/mlpack/methods/linear_svm/linear_svm_main.cpp
```diff
@@ -0,0 +1,443 @@
...
+        Log::Fatal << "Test data given with " << PRINT_PARAM_STRING("test")
+            << " has " << testSet.n_cols << " points, but labels in "
+            << PRINT_PARAM_STRING("test_labels") << " have "
+            << testLabels.n_elem << " labels!" << endl;
+      }
+
+      numClasses = CLI::GetParam<int>("num_classes") == 0 ?
+        model->mappings.n_elem : CLI::GetParam<int>("num_classes");
+      vector<size_t> correctClassCounts(numClasses, 0);
+      vector<size_t> labelSize(numClasses, 0);
```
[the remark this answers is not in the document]
Yashwants19: Updated

src/mlpack/methods/linear_svm/linear_svm_main.cpp
```diff
@@ -0,0 +1,443 @@
...
+    }
+
+    model->svm.Classify(testSet, predictedLabels);
+    data::RevertLabels(predictedLabels, model->mappings, predictions);
+
+    // Calculate accuracy, if desired.
+    if (CLI::HasParam("test_labels"))
+    {
+      arma::Row<size_t> testLabels =
+          std::move(CLI::GetParam<arma::Row<size_t>>("test_labels"));
```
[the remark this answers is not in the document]
Yashwants19: I have added this part.

src/mlpack/methods/linear_svm/linear_svm_main.cpp
```diff
@@ -0,0 +1,443 @@
...
+    1e-10);
+PARAM_INT_IN("max_iterations", "Maximum iterations for optimizer (0 indicates "
+    "no limit).", "n", 10000);
+PARAM_DOUBLE_IN("step_size", "Step size for parallel SGD optimizer.",
+    "s", 0.01);
+PARAM_FLAG("shuffle", "Don't shuffle the order in which data points are "
+    "visited for parallel SGD.", "S");
+PARAM_INT_IN("epochs", "Maximum number of full epochs over dataset for "
+    "psgd", "E", 50);
+PARAM_INT_IN("seed", "Random seed.  If 0, 'std::time(NULL)' is used.", "r", 0);
```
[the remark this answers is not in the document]
Yashwants19: Updated.

src/mlpack/methods/linear_svm/linear_svm_main.cpp
```diff
@@ -0,0 +1,443 @@
...
+
+// Training parameters.
+PARAM_MATRIX_IN("training", "A matrix containing the training set (the matrix "
+    "of predictors, X).", "t");
+PARAM_UROW_IN("labels", "A matrix containing labels (0 or 1) for the points "
+    "in the training set (y).", "l");
+
+// Optimizer parameters.
+PARAM_DOUBLE_IN("lambda", "L2-regularization parameter for training.", "L",
+    0.0001);
```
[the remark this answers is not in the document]
Yashwants19: Updated.

Yashwants19:
I will definitely open a PR for benchmarking afterward. 
Thank You.

src/mlpack/methods/linear_svm/linear_svm_main.cpp
```diff
@@ -0,0 +1,449 @@
...
+      if (testSet.n_cols != testLabels.n_elem)
+      {
+        Log::Fatal << "Test data given with " << PRINT_PARAM_STRING("test")
+            << " has " << testSet.n_cols << " points, but labels in "
+            << PRINT_PARAM_STRING("test_labels") << " have "
+            << testLabels.n_elem << " labels!" << endl;
+      }
+
+      numClasses = CLI::GetParam<int>("num_classes") == 0 ?
+        model->mappings.n_elem : CLI::GetParam<int>("num_classes");
```
rcurtin: Oops, this line should be doubly indented.  I can handle it during merge unless you get to it before me. :)

Yashwants19: Job done :)

rcurtin:
Thanks again, awesome contribution. :+1:

Yashwants19:
Hi @rcurtin Now looking forward to complete go bindings PR. And then I would like to work on Extra trees.

rcurtin:
Sounds good, I'll focus my review efforts there as I am able to. :+1:

Yashwants19:
No problem and thank you for the help.

PR #1941, opened 2019-06-25
Made continuous RL environments' API consistent

favre49:
I also added an isTerminal() method for Pendulum environment


src/mlpack/methods/reinforcement_learning/environment/pendulum.hpp
```diff
@@ -93,13 +93,17 @@ class Pendulum
...
    * @param maxTorque Maximum torque.
    * @param dt The differential value.
+   * @param angleThreshold The region about the upright position where the
+   *    state is considered terminal.
    */
   Pendulum(const double maxAngularVelocity = 8,
            const double maxTorque = 2.0,
-           const double dt = 0.05) :
+           const double dt = 0.05,
+           const double angleThreshold = M_PI/6) :
```
zoq: Is that some default value you selected? Also there is a missing space between the ``/`` operator.

favre49: Yes, is this too lenient? perhaps `M_PI/12` is better?

zoq: I was mainly just wondering where this value comes from, but I agree ``M_PI/12`` might be better.

favre49: Also, should I update HISTORY.MD? We didn't do it for PR #1931 so I'm not sure when we do this?

zoq: Good point, I forgot about the HISTORY document, would be great to update the file as well.

manish7294: I wonder why do we need to impose constraint on pendulum using some angleThreshold, Isn't it better to just have terminals at `-PI `& `PI`, otherwise we won't ever be exploring with those states? Moreover this threshold still seems to be a bit large to me. Does it make sense?

zoq: The idea here is that everything below the threshold would result in a terminal state. So we should choose something small, the issue I see here is that we don't terminate if we are above the threshold. So I guess it would make sense to introduce a maximal number of steps? I guess in this case we could also use the maximal number of steps as the terminal check?

favre49: > the issue I see here is that we don't terminate if we are above the threshold

I'm not sure I understand, have I made a mistake in my terminal condition?

>  So I guess it would make sense to introduce a maximal number of steps?

I agree. I think if we do that for the Pendulum environment, we should do this for all environments.

manish7294: > I'm not sure I understand, have I made a mistake in my terminal condition?
No, there is notthing wrong there. As per my understanding Marcus wants to highlight the point that, as of now we don't terminate even when we reaches the terminal state, we just now have a function to lookup whether we arein terminal state or not. So, It could be a good to introduce maximum number of steps to act as a factor for deciding termination point.

> I agree. I think if we do that for the Pendulum environment, we should do this for all environments.
 Sounds good. If we all agree on this, than maybe the change can just be done within this PR.

zoq: > > the issue I see here is that we don't terminate if we are above the threshold
> 
> I'm not sure I understand, have I made a mistake in my terminal condition?

No issue with the. terminal condition, just an issue at which point we stop, but that applies for all tasks.

>I agree. I think if we do that for the Pendulum environment, we should do this for all environments.
> > Sounds good. If we all agree on this, than maybe the change can just be done within this PR.

Fine with me, but I think the default should either be no limit or a pretty high number. What do you think?

favre49: I think we could make the default value 0, in which case we won't make a check for termination.

favre49: Do you think we should maybe add a `Log::Info` stating the reason for termination of the episode, since now there are two possibilities?

favre49:
I fixed the multiple pole cart environments, they were meant to have one column more in `data`. I've changed the for loops appropriately as well.

I also made the changes we talked about in the Pendulum environment. I also noticed the environment does not include a `doneReward`, so I included that as well. If how I have implemented this in the Pendulum environment is fine, I will extend it to the other environments as well.

Also, since the change in the `Action` struct would qualify as non-backwards compatible, should I make a change in HISTORY.md to include these in 4.0.0?

src/mlpack/methods/reinforcement_learning/environment/pendulum.hpp
zoq: Good idea.

src/mlpack/methods/reinforcement_learning/environment/pendulum.hpp
```diff
@@ -93,13 +93,24 @@ class Pendulum
...
+   * @param maxTimeSteps The number of time steps after which the episode
+   *    terminates. If the value is 0, there is no limit.
    */
   Pendulum(const double maxAngularVelocity = 8,
            const double maxTorque = 2.0,
-           const double dt = 0.05) :
+           const double dt = 0.05,
+           const double angleThreshold = M_PI / 12,
+           const double doneReward = 0.0,
+           const size_t maxTimeSteps = 0) :
```
zoq: Not sure about the name, do you think ``maxSteps`` or ``maxIterations`` could work as well?

zoq:
About the backwards compatible, I think that is reasonable, don't see an easy workaround to support both.

src/mlpack/methods/reinforcement_learning/environment/pendulum.hpp
```diff
@@ -140,6 +154,15 @@ class Pendulum
         -maxAngularVelocity), maxAngularVelocity);
     nextState.Theta() = theta + newAngularVelocity * dt;
 
+    // Check if the episode has terminated
+    bool done = IsTerminal(nextState);
+
+    // Do not reward the agent if time ran out.
+    if (done && maxTimeSteps != 0 && timeStepsPerformed >= maxTimeSteps)
```
manish7294: Should we terminate the moment we reached maximum iterations allowed? Let me know your thoughts on this?

manish7294: Oh, I see you are doing the same. I thought done is being taken from angle terminal condition. Ignore the above comment.

src/mlpack/tests/rl_components_test.cpp
```diff
@@ -37,7 +37,7 @@ BOOST_AUTO_TEST_SUITE(RLComponentsTest)
  */
 BOOST_AUTO_TEST_CASE(SimplePendulumTest)
 {
-  const Pendulum task = Pendulum();
+  Pendulum task = Pendulum();
```
zoq: Might be a good idea to test terminates after the maximal steps as well, we could use a super low value, to keep the runtime low.

favre49:
I extended what I did with `Pendulum` to other environments, and added tests as well. Besides that, I noticed some environments don't have `doneReward`, so I added that. Finally, I removed the `typedef` from `Acrobot` seeing as this PR will remove backwards compatibility anyway.

zoq:
Great, thanks for looking into it.

src/mlpack/methods/reinforcement_learning/environment/acrobot.hpp
```diff
@@ -192,20 +204,33 @@ class Acrobot
...
   {
+    timeStepsPerformed = 0;
     return State((arma::randu<arma::colvec>(4) - 0.5) / 5.0);
   }
 
   /**
    * This function checks if the acrobot has reached the terminal state.
    *
    * @param state The current State.
+   * @return true if state is a terminal state, otherwise false.   *
```
zoq: We can remove the ``*`` at the end.

src/mlpack/methods/reinforcement_learning/environment/pendulum.hpp
```diff
@@ -93,13 +93,25 @@ class Pendulum
    * @param maxAngularVelocity Maximum angular velocity.
    * @param maxTorque Maximum torque.
    * @param dt The differential value.
+   * @param angleThreshold The region about the upright position where the
+   *    state is considered terminal.
+   * @param doneReward The reward recieved by the agent on success.
+   * @param maxTimeSteps The number of time steps after which the episode
```
zoq: Can you change every occurrence of ``maxTimeSteps`` with ``maxSteps``?

src/mlpack/methods/reinforcement_learning/environment/pendulum.hpp
```diff
@@ -140,6 +155,15 @@ class Pendulum
         -maxAngularVelocity), maxAngularVelocity);
     nextState.Theta() = theta + newAngularVelocity * dt;
 
+    // Check if the episode has terminated
+    bool done = IsTerminal(nextState);
+
+    // Do not reward the agent if time ran out.
+    if (done && maxSteps != 0 && timeStepsPerformed >= maxSteps)
```
zoq: Since we are going with ``maxSteps`` I think it makes sense to call this ``stepsPerformed``?

src/mlpack/methods/reinforcement_learning/environment/pendulum.hpp
```diff
@@ -183,6 +208,32 @@ class Pendulum
...
+   * This function checks if the pendulum has reaches a terminal state
+   * 
+   * @param state desired state.
+   * @return true if state is a terminal state, otherwise false.
+   */
+  bool IsTerminal(const State& state) const
+  {
+    if (maxSteps != 0 && timeStepsPerformed >= maxSteps)
+    {
+      Log::Info << "Episode terminated due to the maximum number of time steps"
```
zoq: Should we go with ``number of steps``?

src/mlpack/methods/reinforcement_learning/environment/pendulum.hpp
```diff
@@ -183,6 +208,32 @@ class Pendulum
...
+        state.Theta() < -M_PI + angleThreshold)
+    {
+      Log::Info << "Episode terminated due to agent succeeding.";
+      return true;
+    }
+    return false;
+  }
+
+  //! Get the number of time steps performed
+  size_t TimeStepsPerformed() const { return timeStepsPerformed; }
```
zoq: See comment above.

src/mlpack/tests/rl_components_test.cpp
```diff
@@ -37,7 +37,7 @@ BOOST_AUTO_TEST_SUITE(RLComponentsTest)
  */
 BOOST_AUTO_TEST_CASE(SimplePendulumTest)
 {
-  const Pendulum task = Pendulum();
+  Pendulum task = Pendulum(8, 2, 0.05, M_PI/12, 0, 5);
```
zoq: Picky style comment, missing space between ``/``.

src/mlpack/tests/rl_components_test.cpp
```diff
@@ -57,15 +65,24 @@ BOOST_AUTO_TEST_CASE(SimplePendulumTest)
  */
 BOOST_AUTO_TEST_CASE(SimpleContinuousMountainCarTest)
 {
-  const ContinuousMountainCar task = ContinuousMountainCar();
+  ContinuousMountainCar task = ContinuousMountainCar(-1.2, 0.6, 0.45, -0.07,
```
zoq: Might be a good idea to introduce a method to set the maximal number of steps? That way we could use the default constructor.

src/mlpack/methods/reinforcement_learning/environment/cart_pole.hpp
```diff
@@ -104,7 +104,9 @@ class CartPole
    * @param tau The time interval.
    * @param thetaThresholdRadians The maximum angle.
    * @param xThreshold The maximum position.
-   * @param doneReward Reward recieved on termination.
+   * @param doneReward Reward recieved by agent on success.
```
manish7294: picky comment!, I think we should keep the description of doneReward as "Reward recieved by agent on termination"

favre49: I made this change since now the termination of an episode can happen because the maximum number of steps have been taken (which could be a success in cases like cart pole but a failure because in other cases) or any other termination case we have included. I changed the description to make it more clear, since it may not be evident to the user.

manish7294: Still I think termination word deals with both the cases unbiasedly. Though, no worries, you can keep it that way. :)

manish7294:
Great work, glad to merge this in.

PR #1943, opened 2019-06-27
Minor Optimisation in Backward computation

saksham189:
`inputMean = input.each_row() - mean;`
Now if we take `arma::mean(-2 * inputMean, 0)` that would be 0 so we, can remove that from the computation.


ShikharJ:
Thanks for the contribution @saksham189.

Issue #1942, opened 2019-06-25
MSI Installer sample-ml-app is missing data/german.csv test file


gmanlan:
#### Issue description

The MSI installer bundles the sample-ml-app, but it refers to a sample dataset (german.csv) which is missing in the C:\Program Files\mlpack\examples\sample-ml-app\sample-ml-app\data directory.

#### Your environment

 * version of mlpack: 3.1.1
 * operating system: Windows
 * compiler: VS2017

#### Steps to reproduce

Download the MSI, install, check the dataset directory

#### Expected behavior

It should contain the CSV dataset so the sample app can use it

#### Actual behavior

The CSV dataset file is missing

gmanlan:
CSV is being copied using a post-build event.

gmanlan:
The post-build event to copy the CSV is included in the sample project, but:
1) the dataset CSV is not included with the mlpack files at C:\Program Files\mlpack\mlpack\tests
2) the post-build event refers to the source code path and not the installer path

rcurtin:
Thanks for the report---I'm hoping #1948 will fix this.  I'll be able to test it once AppVeyor generates the MSI.

gmanlan:
Thanks @rcurtin. I'm not 100% sure, but I believe the build failed because the cp path is missing the directory 'examples' between staging and sample-ml-app:
C:\projects\mlpack\dist\win-installer\staging\   **examples\**   sample-ml-app\sample-ml-app\data\

rcurtin:
I verified with #1948 that the data does get copied, and I built the sample ML app in Visual Studio and ran it successfully, so the data does install to the right place.  So I'll go ahead and close this, but feel free to reopen if I missed something. :)

PR #1948, opened 2019-07-06
Copy data for sample Windows installer

rcurtin:
This is intended to handle #1942.  I think it will work, but I need to test using the generated AppVeyor artifact first.  The goal is just to have `german.csv` in the `sample-ml-app\data\` directory upon install.


rcurtin:
I just verified that this works by running the compiled ml-sample-app from inside Visual Studio, so I'll go ahead and merge it.

PR #1949, opened 2019-07-10
Julia bindings

rcurtin:
This PR adds Julia bindings to mlpack.  Here's an example usage:

```
               _
   _       _ _(_)_     |  Documentation: https://docs.julialang.org
  (_)     | (_) (_)    |
   _ _   _| |_  __ _   |  Type "?" for help, "]?" for Pkg help.
  | | | | | | |/ _` |  |
  | | |_| | | | (_| |  |  Version 1.1.0 (2019-01-21)
 _/ |\__'_|_|_|\__'_|  |  Official https://julialang.org/ release
|__/                   |

julia> using mlpack

julia> x = rand(100, 10);

julia> y = mlpack.pca(x, verbose=true, new_dimensionality=5)
[INFO ] Performing PCA on dataset...
[INFO ] 67.3709% of variance retained (5 dimensions).
5×10 Array{Float64,2}:
  1.13254   -0.637608  -0.630083   0.152717   -1.18094   -0.835606   0.641949   2.10899   -1.4188     0.666852
 -0.451976   0.536181   0.114842   1.383       0.446358  -2.19528    0.117253   0.653234   0.787656  -1.39127 
  1.3781    -0.557872   1.18812   -0.0910189  -1.43166   -0.200452  -1.18344   -0.135198   1.27368   -0.240258
 -1.25992    0.877698  -0.492323  -0.504153   -1.41801   -0.257386   0.507117   0.284061   1.23223    1.03068 
  0.261067  -0.926367   0.448858  -1.42664     0.271981  -0.323317   1.86706   -0.06633    0.634037  -0.740351
```

And another one with a model:

```
               _
   _       _ _(_)_     |  Documentation: https://docs.julialang.org
  (_)     | (_) (_)    |
   _ _   _| |_  __ _   |  Type "?" for help, "]?" for Pkg help.
  | | | | | | |/ _` |  |
  | | |_| | | | (_| |  |  Version 1.1.0 (2019-01-21)
 _/ |\__'_|_|_|\__'_|  |  Official https://julialang.org/ release
|__/                   |

julia> using mlpack

julia> using CSV

julia> x = convert(Array{Float64, 2}, CSV.read("/home/ryan/src/mlpack/build/vc2.csv"));

julia> y = convert(Array{Int64, 2}, CSV.read("/home/ryan/src/mlpack/build/vc2_labels.txt"))[:, 1] .+ 1;

julia> _, model, _ = mlpack.perceptron(training=x, labels=y, verbose=true, points_are_rows=true)
[INFO ] Training perceptron on dataset '6x206 matrix' with labels in '1x206 matrix' for a maximum of 1000 iterations.
(Int64[], Ptr{Nothing} @0x000000000223e010, Int64[])

julia> _, _, predictions = mlpack.perceptron(input_model=model, verbose=true, points_are_rows=true, test=x)
[INFO ] Using saved perceptron from PerceptronModel model at 0x223e010.
[INFO ] Classifying dataset '6x206 matrix'.
([3, 3, 3, 3, 3, 3, 3, 3, 3, 3  …  3, 3, 3, 3, 3, 3, 3, 3, 3, 3], Ptr{Nothing} @0x000000000223e010, [3, 3, 3, 3, 3, 3, 3, 3, 3, 3  …  3, 3, 3, 3, 3, 3, 3, 3, 3, 3])

julia> sum(y .== predictions)
128
```

(I guess that perceptron model wasn't very good.  But it still worked!)

There's a lot of code here, so if you're looking to read or review the code here is a guide to how it's laid out:

 * All the Julia-related code is in `src/mlpack/bindings/julia/`.
 * In order to make a Julia binding, we have to make a `.jl` file which internally calls out to mlpack functions that we have compiled into a standalone `.so`.  Here is an example: https://gist.github.com/90d917e4fe394a51efff02b15d83bd02
 * But to call the functions in those standalone `.so`s (which contain CLI functionality and also `mlpackMain()`), we have to generate a `.h` and `.cpp` file.  Code to generate those files can be found in `julia_method.h.in` and `julia_method.cpp.in`, and we simply use CMake's `configure_file()` to get those set up correctly.
 * Each `.jl` is generated using the program `generate_jl.cpp.in`, which, like the others, is just configured with CMake's `configure_file()`.
 * The real bulk of work for printing the `.jl` is in `print_jl.cpp`.  You can follow that file through to see what happens.
 * There's not much special about how each type is handled in the Julia-C++ conversions.  Specifically, we use Julia's built-in `Array{Float64, 2}` to represent `arma::mat`, and for categorical matrices we use `Tuple{Array{Bool, 1}, Array{Float64, 2}}` which maybe isn't pretty but it works.  For holding mlpack models, we just have a `Ptr{Nothing}`.
 * The Julia bindings are integrated with the Markdown bindings, so when `make markdown` is run, it also generates Markdown sections for Julia bindings too.  This should work with the existing website generation scripts.
 * Travis will build and test the Julia bindings.

If you want to build these and run them, you can follow this procedure:

 * Download Julia from julialang.org (or get it through your package manager, but I had nothing but nightmares trying to use the Debian-packaged Julia) and unpack into, e.g., `/opt/julia-1.1.1/`.
 * Configure mlpack with `-DBUILD_JULIA_BINDINGS=ON -DJULIA_EXECUTABLE=/opt/julia-1.1.1/bin/julia` (optionally `-DBUILD_MARKDOWN_BINDINGS=ON` if you want to see that output).
 * Build with `make julia`, and if you want `make test` will test the Julia bindings also.
 * Once the Julia bindings are built, then you can run them by (from the build directory) `cd src/mlpack/bindings/julia/mlpack/ && JULIA_PROJECT=$PWD /opt/julia-1.1.1/bin/julia` then you can do `using mlpack` and play with it.
 * Note that documentation is available, so if you do `? mlpack.pca` while in the Julia REPL you'll get the documentation.

I realized while writing this PR description that there are still a couple of little things I need to do.

 - [x] write a quickstart tutorial
 - [x] when exceptions are thrown, these should not cause Julia to crash
 - [x] set up a local copy of the website documentation with Julia bindings available
 - [x] fix documentation where sometimes not all matrices show up in the initialization

The last comment is how distribution will work.  Unfortunately we can't build directly from the mlpack repository, because the Julia package registry will require a repository like `mlpack/mlpack.jl`.  Thus, once this is integrated, we'll have to set up a simple `mlpack/mlpack.jl` repo that has a `build.jl` file so that when someone does `Pkg.add("mlpack")` or `Pkg.build("mlpack")`, mlpack will be downloaded and built with the instructions above.  Fortunately, we should be able to keep the contents of that repository quite minimal.

I know this is a big PR, and the automatic binding code is rather complex, so I'm happy to clarify anything.


rcurtin:
Ok, unfortunately it took a while for me to get to this but now it is ready.  I built a temporary copy of the binding documentation that one can find here:

http://mlpack.ratml.org/doc/mlpack-git/julia_documentation.html

but note that page doesn't show the Julia quickstart; however, when we merge it it will be there.  (This has to do with the way that I built this temporary documentation.)

I'll work on fixing the style issues and anything else that comes up, but otherwise this is ready for review.

src/mlpack/bindings/julia/test_julia_binding_main.cpp
```diff
@@ -0,0 +1,189 @@
+/**
+ * @file julia_binding_test.cpp
```
Yashwants19: Hey @rcurtin do we require this file.?

rcurtin: I'm not sure what you mean; can you explain further?

Yashwants19: I think so here we have two main.cpp files.

Yashwants19: One Julia folder and one in tests folder.

rcurtin: Oh!  Nice catch, thanks.  I removed the duplicate in 80c9a8d.

src/mlpack/bindings/julia/mlpack/cli.jl
```diff
@@ -0,0 +1,364 @@
...
+                     vector::Vector{Int64})
+  ccall((:CLI_SetParamVectorInt, library), Nothing, (Cstring, Ptr{Int64},
+      Int64), paramName, Base.pointer(vector), size(vector, 1));
+end
+
+function CLISetParam(paramName::String,
+                     matWithInfo::Tuple{Array{Bool, 1}, Array{Float64, 2}},
+                     pointsAsRows::Bool)
+  ccall((:CLI_SetParamMatWithInfo, library), Nothing, (Cstring, Ptr{Bool},
+      Ptr{Float64}, Int64, Int64, Bool), paramName,
```
Yashwants19: I am not familiar we Julia but can we add a sanity check for dimensions of bool array must be equal to float array or Julia will itself check this.?

.travis.yml
```diff
@@ -35,6 +35,18 @@ matrix:
         - cmake . && make && sudo make install && cd ..
         - sudo cp .travis/config.hpp /usr/include/armadillo_bits/config.hpp
 
+    - os: linux
+      dist: xenial
```
zoq: I'm wondering if it makes sense to use travis stages to distinguish the builds; not sure can we just add ``stage: Julia``

src/mlpack/bindings/julia/strip_type.hpp
```diff
@@ -0,0 +1,57 @@
...
+    const size_t loc = cppType.find("<>");
+    if (loc != std::string::npos)
+    {
+      // Convert it from "<>".
+      cppType.replace(loc, 2, "");
+    }
+    else
+    {
+      // Let's just replace the '<' and '>' with a valid '_' character.
+      while (cppType.find("<") != std::string::npos)
```
zoq: Not sure I see the reason to use ``find`` on the string first, as ``cppType.replace(cppType.begin(), cppType.end(), "<", "_");`` should do the same?

doc/guide/julia_quickstart.hpp
```diff
@@ -0,0 +1,192 @@
...
+mlpack.
+
+@section julia_quickstart_movierecs Using mlpack for movie recommendations
+
+In this example, we'll train a collaborative filtering model using mlpack's
+<tt><a href="https://www.mlpack.org/doc/mlpack-git/julia_documentation.html#cf">cf()</a></tt> method.  We'll train this on the MovieLens dataset from
+https://grouplens.org/datasets/movielens/, and then we'll use the model that we
+train to give recommendations.
+
+You can copy-paste this code directly into Python to run it.
```
Yashwants19: Hi, @rcurtin I think so it must be `You can copy-paste this code directly into Julia to run it.`

.travis.yml
rcurtin: I think you're right, but let me try that in a different PR since it would be a total change to how the Travis build works.

src/mlpack/bindings/julia/strip_type.hpp
rcurtin: Which overload of `std::string::replace()` is that?  When I read the documentation at http://www.cplusplus.com/reference/string/string/replace/ I can't figure out which one it is.  Maybe I overlooked it?

doc/guide/julia_quickstart.hpp
rcurtin: Nice catch, thanks.  Fixed in be48606c.

src/mlpack/bindings/julia/strip_type.hpp
zoq: You are right, I was talking about: ``std::replace``, so ``std::replace(cppType.begin(), cppType.end(), "<", "_");``.

.travis.yml
zoq: Sure makes sense.

src/mlpack/bindings/julia/strip_type.hpp
rcurtin: Ah, got it; I refactored to use `std::replace()` in 74620db49.  Thanks for the suggestion, it is a lot nicer. :+1:

src/mlpack/bindings/julia/CMakeLists.txt
```diff
@@ -0,0 +1,158 @@
+if (BUILD_JULIA_BINDINGS)
+  ## We need to check here if Julia is even available.  Although actually
+  ## technically, I'm not sure if we even need to know!  For the tests though we
+  ## do.  So it's probably a good idea to check.
+  find_package(Julia 0.7.0)
+  if (NOT JULIA_FOUND)
```
zoq: Should we print out a warning or something like that, since the user ask to build the julia bindings?

src/mlpack/bindings/julia/default_param_impl.hpp
```diff
@@ -0,0 +1,147 @@
...
+    return "zeros(Int64, 0, 0)";
+  }
+  else
+  {
+    return "zeros(0, 0)";
+  }
+}
+
+/**
+ * Return the default value of a model option (always "None").
```
zoq: Or "nothing" I guess.

src/mlpack/bindings/julia/get_printable_type_impl.hpp
```diff
@@ -0,0 +1,109 @@
...
+  if (std::is_same<T, bool>::value)
+    return "Bool";
+  else if (std::is_same<T, int>::value)
+    return "Int";
+  else if (std::is_same<T, double>::value)
+    return "Float64";
+  else if (std::is_same<T, std::string>::value)
+    return "String";
+  else
+    throw std::invalid_argument("unknown parameter type" + data.cppType);
```
zoq: Should there be a space right after type?

src/mlpack/bindings/julia/get_printable_type_impl.hpp
```diff
@@ -0,0 +1,109 @@
...
+  else if (std::is_same<T, arma::rowvec>::value)
+    return "Float64 vector-like";
+  else if (std::is_same<T, arma::Row<size_t>>::value)
+    return "Int64 vector-like";
+  else if (std::is_same<T, arma::vec>::value)
+    return "Float64 vector-like";
+  else if (std::is_same<T, arma::Col<size_t>>::value)
+    return "Int64 vector-like";
+  else
+    throw std::invalid_argument("unknown Armadillo type" + data.cppType);
```
zoq: See comment above.

src/mlpack/bindings/julia/julia_option.hpp
```diff
@@ -0,0 +1,118 @@
...
+    data.loaded = false;
+
+    // Only "verbose" will be persistent.
+    if (identifier == "verbose")
+      data.persistent = true;
+    else
+      data.persistent = false;
+    data.cppType = cppName;
+
+    // Every parameter we'll get from Python will have the correct type.
```
zoq: Do you mean Julia?

src/mlpack/bindings/julia/CMakeLists.txt
rcurtin: Ah, good point, fixed in dda0fc9a.  @Yashwants19 and I were talking about how to do that for Go bindings in #1884.  @Yashwants19: take a look at dda0fc9a and you could adapt the same strategy for the Go bindings.

Basically this means that if the user did not specify `-DBUILD_JULIA_BINDINGS=ON` and Julia is not found, then configuration will complete successfully but Julia bindings will not be built.  If the user *does* specify `-DBUILD_JULIA_BINDINGS=ON` and Julia is *not* found, then a fatal error will be issued during configuration.

src/mlpack/bindings/julia/default_param_impl.hpp
rcurtin: Oops, fixed in eee1173a.

src/mlpack/bindings/julia/get_printable_type_impl.hpp
rcurtin: Yes, nice catch; fixed in eee1773.

src/mlpack/bindings/julia/get_printable_type_impl.hpp
rcurtin: Also fixed in eee1773a.

src/mlpack/bindings/julia/julia_option.hpp
rcurtin: Right, there was a good amount of code that I based on the Python implementation.  I went through and changed any references from Python to Julia as needed in eee1173a.

rcurtin:
I'll merge this at the same time mlpack 3.2.2 is released.

src/mlpack/bindings/julia/julia_util.cpp
```diff
@@ -0,0 +1,497 @@
...
+}
+
+/**
+ * Get the memory pointer for a matrix parameter.
+ * Note that this will assume that whatever is calling will take ownership of
+ * the memory!
+ */
+double* CLI_GetParamMat(const char* paramName)
+{
+  arma::access::rw(CLI::GetParam<arma::mat>(paramName).mem_state) = 1;
```
Vasniktel: I am curious whether we should take into account preallocated storage in `arma::mat` like we do in `CLI_GetParamUMat` .

rcurtin: Nice catch, you're right!  Fixed in 004e104be. :)

src/mlpack/bindings/julia/julia_util.cpp
```diff
@@ -0,0 +1,510 @@
...
+/**
+ * Get the memory pointer for a column vector parameter.
+ * Note that this will assume that whatever is calling will take ownership of
+ * the memory!
+ */
+double* CLI_GetParamCol(const char* paramName)
+{
+  arma::access::rw(CLI::GetParam<arma::vec>(paramName).mem_state) = 1;
+  return CLI::GetParam<arma::vec>(paramName).memptr();
+}
```
Vasniktel: This might not be a big issue but still... The same situation as in #1949 - `arma::vec` might have its memory inlined since it inherits from `arma::Mat`. The same applies to `GetParamUCol`, `GeParamRow`, `GetParamURow` and `GetParamMatWithInfoPtr`

rcurtin: You're right.  I thought that I checked all the overloads but apparently not ... fixed in 08ae85244.

rcurtin:
Since 3.2.2 is released, I'm going to make the Julia build work with Azure pipelines, and once that's successful, I'll go ahead and merge this PR.

src/mlpack/bindings/julia/get_julia_type.hpp
```diff
@@ -0,0 +1,136 @@
...
+template<typename T>
+inline std::string GetJuliaType(
+    const typename std::enable_if<!util::IsStdVector<T>::value>::type* = 0,
+    const typename std::enable_if<!arma::is_arma_type<T>::value>::type* = 0,
+    const typename std::enable_if<!std::is_same<T,
+        std::tuple<data::DatasetInfo, arma::mat>>::value>::type* = 0,
+    const typename std::enable_if<!data::HasSerialize<T>::value>::type* = 0)
+{
+  return "unknown_"; // This will cause an error most likely...
+}
```
Vasniktel: This function gets called for model types which causes this:
```
`input_model::unknown_`: Input AdaBoost model.
```
I fixed the same bug in java bindings by adding
```
const typename std::enable_if<!std::is_same<T,
        std::tuple<data::DatasetInfo, arma::mat>>::value>::type* = 0
```
to the function that is meant to be called with serializable types (the last one in his file)

rcurtin: This overload should only be called for classes that:

 * aren't one of the specific types in the template specifications below: `bool`, `int`, `size_t`, `double`, `std::string`
 * aren't `std::vector<T>`
 * aren't Armadillo types
 * aren't `std::tuple<data::DatasetInfo, arma::mat>`
 * don't have a `serialize()` method

So, actually, this should never be called.  As far as I can reason, adding `const typename std::enable_if<!std::is_same<T, std::tuple<data::DatasetInfo, arma::mat>>::value>::type*` shouldn't be necessary for the overload that gets models anyway, since `std::tuple<data::DatasetInfo, arma::mat>` won't have a `serialize()` member.

At least, I've certainly never seen any issues with this code in the generated Julia bindings.  Maybe the Java bindings are different---I haven't looked at them yet. :+1:

rcurtin:
Ok, I have this building correctly now and it is ready to merge.  There will have to be a follow-up PR after I get mlpack's Julia package set up correctly in a separate repository (so that installation from Julia is easy)---I'll have to update the installation/documentation details for Julia.

@zoq the changes I had to make modified the Azure pipelines somewhat.  Do you want to take a look at the modifications I made there before I merge this and make sure I didn't break anything or change the functionality in a bad way?  Now we use `ctest` directly, but I'm not sure if everything is right.  It seems that all the tests run, and it does give correct output when tests fail.  Let me know what you think.

zoq:
Looks good to me, I guess I used the wrong settings for ctest, but great that you got it working. Excited to see the Julia bindings merged.

rcurtin:
Scary to click that merge button since this is so much code, but I did it. :)  Next I have to do some work to set things up from the Julia side.  Ideally we should be able to do

```
julia> import Pkg; Pkg.add("mlpack")
julia> using mlpack
```

but this is likely to mean setting up an mlpack/mlpack.jl repository with the right build scripts.  I'll start investigating that.

PR #1951, opened 2019-07-14
Double pole cart environments fix and RL tests

favre49:
I've fixed the double pole cart environments which I tested on NEAT. I'll also be adding QLearning tests for all the environments soon.

### DIFF
--- a/src/mlpack/methods/reinforcement_learning/environment/continuous_double_pole_cart.hpp
+++ b/src/mlpack/methods/reinforcement_learning/environment/continuous_double_pole_cart.hpp
@@ -183,7 +183,7 @@ class ContinuousDoublePoleCart
 
   /**
    * This is the ordinary differential equations required for estimation of
-   * nextState through RK4 method.
+   * next state through RK4 method.
    *
    * @param state The current state.
    * @param action The action taken.
--- a/src/mlpack/methods/reinforcement_learning/environment/continuous_double_pole_cart.hpp
+++ b/src/mlpack/methods/reinforcement_learning/environment/continuous_double_pole_cart.hpp
@@ -1,8 +1,8 @@
 /**
- * @file continuous_multiple_pole_cart.hpp
+ * @file continuous_double_pole_cart.hpp
  * @author Rahul Ganesh Prabhu
  *
- * This file is an implementation of Continuous Multiple Pole Cart Balancing
+ * This file is an implementation of Continuous Double Pole Cart Balancing
  * Task.
  *
  * mlpack is free software; you may redistribute it and/or modify it under the
@@ -11,8 +11,8 @@
  * http://www.opensource.org/licenses/BSD-3-Clause for more information.
  */
 
-#ifndef MLPACK_METHODS_RL_ENVIRONMENT_CONTINUOUS_MULTIPLE_POLE_CART_HPP
-#define MLPACK_METHODS_RL_ENVIRONMENT_CONTINUOUS_MULTIPLE_POLE_CART_HPP
+#ifndef MLPACK_METHODS_RL_ENVIRONMENT_CONTINUOUS_DOUBLE_POLE_CART_HPP
+#define MLPACK_METHODS_RL_ENVIRONMENT_CONTINUOUS_DOUBLE_POLE_CART_HPP
 
 #include <mlpack/prereqs.hpp>
 
@@ -20,13 +20,13 @@ namespace mlpack {
 namespace rl {
 
 /**
- * Implementation of Continuous Multiple Pole Cart Balancing task.
+ * Implementation of Continuous Double Pole Cart Balancing task.
  */
-class ContinuousMultiplePoleCart
+class ContinuousDoublePoleCart
 {
  public:
   /**
-   * Implementation of the state of Continuous Multiple Pole Cart. The state is 
+   * Implementation of the state of Continuous Double Pole Cart. The state is 
    * expressed as a vector (position, velocity, angle, angular velocity, angle,
    * angular velocity)
    */
@@ -84,7 +84,7 @@ class ContinuousMultiplePoleCart
   };
 
   /**
-   * Implementation of action of Continuous Multiple Pole Cart.
+   * Implementation of action of Continuous Double Pole Cart.
    */
   struct Action
   {
@@ -94,7 +94,7 @@ class ContinuousMultiplePoleCart
   };
 
   /**
-   * Construct a Multiple Pole Cart instance using the given constants.
+   * Construct a Double Pole Cart instance using the given constants.
    *
    * @param m1 The mass of the first pole.
    * @param m2 The mass of the second pole.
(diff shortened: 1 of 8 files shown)


favre49:
It looks like we already have Q learning set up for all discrete environments, and we don't have policy gradient methods set up to test continuous environments. So I've written the test for the double pole balancing environment. Q learning takes quite some time to build up an average return greater than 30, so instead, I say the environment has been solved if it is able to successively cross a step size close to the maximum number of steps, since that seems like a stronger condition.

I was also thinking that I should rename the MultiplePoleCart environment to DoublePoleCart, I'll do that soon.

src/mlpack/methods/reinforcement_learning/environment/continuous_double_pole_cart.hpp
```diff
@@ -0,0 +1,374 @@
...
+#ifndef MLPACK_METHODS_RL_ENVIRONMENT_CONTINUOUS_DOUBLE_POLE_CART_HPP
+#define MLPACK_METHODS_RL_ENVIRONMENT_CONTINUOUS_DOUBLE_POLE_CART_HPP
+
+#include <mlpack/prereqs.hpp>
+
+namespace mlpack {
+namespace rl {
+
+/**
+ * Implementation of Continuous Double Pole Cart Balancing task.
```
zoq: Can we add some more information about the environment, reference the existing Cart Pole env, and say that this is an extend could be a solution. What do you think?

src/mlpack/methods/reinforcement_learning/environment/continuous_double_pole_cart.hpp
```diff
@@ -0,0 +1,374 @@
...
+    double Position() const { return data[0]; }
+    //! Modify the position of the cart.
+    double& Position() { return data[0]; }
+
+    //! Get the velocity of the cart.
+    double Velocity() const { return data[1]; }
+    //! Modify the velocity of the cart.
+    double& Velocity() { return data[1]; }
+
+    //! Get the angle of the $i^{th}$ pole with the vertical.
```
zoq: Not sure I get the vertical point.

src/mlpack/methods/reinforcement_learning/environment/continuous_double_pole_cart.hpp
```diff
@@ -0,0 +1,374 @@
...
+   * @param forceMag The magnitude of the applied force.
+   * @param tau The time interval.
+   * @param thetaThresholdRadians The maximum angle.
+   * @param xThreshold The maximum position.
+   * @param doneReward Reward recieved by agent on success.
+   * @param maxSteps The number of steps after which the episode
+   *    terminates. If the value is 0, there is no limit.
+   */
+  ContinuousDoublePoleCart(const double m1 = 0.1,
+                             const double m2 = 0.01,
```
zoq: Minor style issue, the parameter should align with the first one.

src/mlpack/methods/reinforcement_learning/environment/continuous_double_pole_cart.hpp
```diff
@@ -0,0 +1,374 @@
...
+                State& nextState)
+  {
+    // Update the number of steps performed.
+    stepsPerformed++;
+
+    arma::vec dydx(6, arma::fill::zeros);
+    for (size_t i = 0; i < 2; i++)
+    {
+      dydx[0] = state.Velocity();
+      dydx[2] = state.AngularVelocity(1);
```
zoq: Looks like the ``state.AngularVelocity`` and ``state.AngularVelocity`` is missing the ``I``parameter.

src/mlpack/methods/reinforcement_learning/environment/continuous_double_pole_cart.hpp
```diff
@@ -0,0 +1,374 @@
...
+      dydx[4] = state.AngularVelocity(2);
+    }
+    Dsdt(state, action, dydx);
+    RK4(state, action, dydx, nextState);
+
```
zoq: Minor style issue, we can remove the empty line here.

src/mlpack/methods/reinforcement_learning/environment/continuous_double_pole_cart.hpp
```diff
@@ -0,0 +1,374 @@
...
+    totalMass += m2 * (0.25 + 0.75 * sinTheta2 * sinTheta2);
+
+    // Calculate acceleration.
+    double xAcc = totalForce / totalMass;
+
+    // Calculate angular acceleration.
+    double angAcc1 = -0.75 * (xAcc * cosTheta1 + gravity * sinTheta1) / l1;
+    double angAcc2 = -0.75 * (xAcc * cosTheta2 + gravity * sinTheta2) / l2;
+
+    dydx[1] = xAcc;
```
zoq: We could also use ``dydx[1]``, ``dydx[3]`` and ``dydx[5]`` right away, to store the values.

src/mlpack/methods/reinforcement_learning/environment/continuous_double_pole_cart.hpp
```diff
@@ -0,0 +1,374 @@
...
+    /**
+     * When done is false, it means that the cartpole has fallen down.
+     * For this case the reward is 1.0.
+     */
+    return 1.0;
+  }
+
+  /**
+   * This is the ordinary differential equations required for estimation of
+   * nextState through RK4 method.
```
zoq: In the ``RK4`` you called it next state, to be consistent we should use this here as well.

src/mlpack/methods/reinforcement_learning/environment/continuous_double_pole_cart.hpp
```diff
@@ -0,0 +1,374 @@
...
+   * @param action The action to be applied.
+   * @param dydx The differential.
+   * @param nextState The next state.
+   */
+  void RK4(const State& state,
+           const Action& action,
+           arma::vec& dydx,
+           State& nextState)
+  {
+    double hh = tau * 0.5;
```
zoq: I might use ``const`` here just to make it clear this isn't going to be changed.

src/mlpack/methods/reinforcement_learning/environment/continuous_double_pole_cart.hpp
zoq: I was talking about "estimation of nextState through RK4 method." and "to estimate the next state based on given ordinary differential equation.".

src/mlpack/methods/reinforcement_learning/environment/double_pole_cart.hpp
```diff
@@ -214,15 +213,11 @@ class DoublePoleCart
...
+    dydx[1] = totalForce / totalMass;
 
     // Calculate angular acceleration.
-    double angAcc1 = -0.75 * (xAcc * cosTheta1 + gravity * sinTheta1) / l1;
-    double angAcc2 = -0.75 * (xAcc * cosTheta2 + gravity * sinTheta2) / l2;
-
-    dydx[1] = xAcc;
-    dydx[3] = angAcc1;
-    dydx[5] = angAcc2;
+    dydx[3] = -0.75 * (xAcc * cosTheta1 + gravity * sinTheta1) / l1;
```
zoq: ``xAcc``should be `` dydx[1]`` now.

favre49: Yup, that's on line 216.

favre49: Oops, I think I saw what you mean

zoq:
Thanks again for looking into the issue.

PR #1952, opened 2019-07-15
Add padding layer

saksham189:


src/mlpack/methods/ann/layer/convolution.hpp
```diff
@@ -324,6 +278,9 @@ class Convolution
   //! Locally-stored transformed gradient parameter.
   arma::cube gradientTemp;
 
+  //! Locally-stored padding layer.
+  LayerTypes<> padding;
```
walragatver: I think this should be a pointer.

src/mlpack/methods/ann/layer/padding.hpp
```diff
@@ -0,0 +1,109 @@
...
+{
+ public:
+  /**
+   * Create the Padding object using the specified number of output units.
+   *
+   * @param padW Padding width of the input.
+   * @param padH Padding height of the input.
+   */
+  Padding(const size_t padW = 0,
+          const size_t padH = 0);
```
walragatver: The layer looks good to me. Can you just add one more option which will take `valid` or `same` as an option and then pad accordingly?

src/mlpack/methods/ann/layer/convolution.hpp
saksham189: yes it is already a pointer.

src/mlpack/methods/ann/layer/padding.hpp
saksham189: I think that `Same` and `valid` are options of the convolutional layers. The padding layer has nothing to do with it. The dimensions cannot be the `same` after padding.
I think it would be easier to implement this in a separate PR after this is merged.

walragatver: No, I don't think so. For having `same` padding `TensorFlow` uses a trick. 
Here are the links you can refer to.
https://www.tensorflow.org/api_docs/python/tf/pad

For `same` padding to make always possible:
https://stackoverflow.com/questions/37674306/what-is-the-difference-between-same-and-valid-padding-in-tf-nn-max-pool-of-t

For the same padding, if the rows to come out `odd` instead of `even`, then they add an extra row or column at the end. So if you wish to implement the way it is implemented in the first link you just will need to add a vector of p[n, 2] as an argument where n is the dimension and p[i, 0] specifies the pad to add at start and p[i, 1] specifies the pad to add at the end of the dimension.

Or what you do is just take `valid` and `same` as an argument and handle everything inside the constructor. It's totally up to you.
But yes to make some modification for sure.

src/mlpack/methods/ann/layer/convolution.hpp
walragatver: Okay, I didn't know about it. Thanks

src/mlpack/methods/ann/layer/padding.hpp
saksham189: > Or what you do is just take valid and same as an argument and handle everything inside the constructor. It's totally up to you.

No, this will not work. The dimensions cannot be the same after padding so, it is illogical to pass the argument `same` to the Padding constructor.

> For the same padding, if the rows to come out odd instead of even, then they add an extra row or column at the end.

Alright, I think this could be done. Basically, you want flexibility to have different padding on different sides (to allow for odd padding).

walragatver: Yes

src/mlpack/methods/ann/layer/convolution.hpp
zoq: You are right, this is already a pointer, but II'm wondering, since we already know this would be a Padding layer, we could just use the ``Padding`` type here.

src/mlpack/methods/ann/layer/convolution_impl.hpp
```diff
@@ -119,7 +122,13 @@ void Convolution<
 
   if (padW != 0 || padH != 0)
   {
-    Pad(inputTemp, padW, padH, inputPaddedTemp);
+    inputPaddedTemp.set_size(inputTemp.n_rows + padW * 2,
+        inputTemp.n_cols + padH * 2, inputTemp.n_slices);
+    for (size_t i = 0; i < inputTemp.n_slices; ++i)
+    {
+      boost::apply_visitor(ForwardVisitor(std::move(inputTemp.slice(i)),
```
zoq: If we use the ``Padding`` type, see comment above, we could just call ``padding.Forward(..)`` without the visitor, let me know what you think.

src/mlpack/methods/ann/layer/padding_impl.hpp
```diff
@@ -0,0 +1,69 @@
...
+
+template<typename InputDataType, typename OutputDataType>
+template<typename eT>
+void Padding<InputDataType, OutputDataType>::Backward(
+    const arma::Mat<eT>&& /* input */,
+    const arma::Mat<eT>&& gy,
+    arma::Mat<eT>&& g)
+{
+  g = gy.submat(padW, padH, padW + nRows - 1,
+        padH + nCols - 1);
```
zoq: Minor style issue, should be 4 spaces.

src/mlpack/methods/ann/layer/padding_impl.hpp
```diff
@@ -0,0 +1,69 @@
...
+template<typename InputDataType, typename OutputDataType>
+template<typename eT>
+void Padding<InputDataType, OutputDataType>::Forward(
+    const arma::Mat<eT>&& input, arma::Mat<eT>&& output)
+{
+  nRows = input.n_rows;
+  nCols = input.n_cols;
+  output = arma::zeros(input.n_rows + padW * 2, input.n_cols + padH * 2);
+  output.submat(padW, padH, padW + input.n_rows - 1,
+        padH + input.n_cols - 1) = input;
```
zoq: See comment below.

ShikharJ:
@saksham189 What is the status on this one?

saksham189:
I can continue the work on this after commenting out a layer. @ShikharJ

ShikharJ:
@saksham189 Please feel free to do that. We can worry about the layer limit later on.

src/mlpack/methods/ann/layer/layer_types.hpp
```diff
@@ -175,7 +176,7 @@ using LayerTypes = boost::variant<
     DropConnect<arma::mat, arma::mat>*,
     Dropout<arma::mat, arma::mat>*,
     AlphaDropout<arma::mat, arma::mat>*,
-    ELU<arma::mat, arma::mat>*,
+    // ELU<arma::mat, arma::mat>*,
```
zoq: Can you comment the ``ReinforceNormal`` layer instead as it's mainly used the the glimpse network?

ShikharJ:
We just merged in the changes for supporting additional layers, @saksham189 can you update this to the latest master please?

src/mlpack/methods/ann/layer/padding.hpp
```diff
@@ -0,0 +1,119 @@
...
+  //! Modify the output parameter.
+  OutputDataType& OutputParameter() { return outputParameter; }
+
+  //! Get the delta.
+  OutputDataType const& Delta() const { return delta; }
+  //! Modify the delta.
+  OutputDataType& Delta() { return delta; }
+
+  /**
+   * Serialize the layer
```
ShikharJ: Punctuation at the end of the sentence please.

src/mlpack/methods/ann/layer/padding_impl.hpp
```diff
@@ -0,0 +1,76 @@
...
+template<typename eT>
+void Padding<InputDataType, OutputDataType>::Forward(
+    const arma::Mat<eT>&& input, arma::Mat<eT>&& output)
+{
+  nRows = input.n_rows;
+  nCols = input.n_cols;
+  output = arma::zeros(input.n_rows + padWLeft + padWRight,
+      input.n_cols + padHTop + padHBottom);
+  output.submat(padWLeft, padHTop, padWLeft + input.n_rows - 1,
+      padHTop + input.n_cols - 1) = input;
```
ShikharJ: Please re-use the `nRows` and `nCols` variables initiated above for efficiency.

ShikharJ:
Thanks @saksham189 for the contribution.

PR #1956, opened 2019-07-25
Bias Visitor.

walragatver:
Hi,
I have added a `bias_set_visitor`. It helps to set the bias parameters only. It also returns the number of bias terms in the given layer.

### DIFF
--- a/src/mlpack/methods/ann/visitor/bias_set_visitor.hpp
+++ b/src/mlpack/methods/ann/visitor/bias_set_visitor.hpp
@@ -1,0 +1,80 @@
+/**
+ * @file bias_set_visitor.hpp
+ * @author Toshal Agrawal
+ *
+ * This file provides an abstraction for the Bias() function for different
+ * layers and automatically directs any parameter to the right layer type.
+ *
+ * mlpack is free software; you may redistribute it and/or modify it under the
+ * terms of the 3-clause BSD license.  You should have received a copy of the
+ * 3-clause BSD license along with mlpack.  If not, see
+ * http://www.opensource.org/licenses/BSD-3-Clause for more information.
+ */
+#ifndef MLPACK_METHODS_ANN_VISITOR_BIAS_SET_VISITOR_HPP
+#define MLPACK_METHODS_ANN_VISITOR_BIAS_SET_VISITOR_HPP
+
+#include <mlpack/methods/ann/layer/layer_traits.hpp>
+
+#include <boost/variant.hpp>
+
+namespace mlpack {
+namespace ann {
+
+/**
+ * BiasSetVisitor update the module bias parameters given the parameters set.
+ */
+class BiasSetVisitor : public boost::static_visitor<size_t>
+{
+ public:
+  //! Update the bias parameters given the parameters set and offset.
+  BiasSetVisitor(arma::mat&& weight, const size_t offset = 0);
+
+  //! Update the parameters set.
+  template<typename LayerType>
+  size_t operator()(LayerType* layer) const;
+
+ private:
+  //! The parameters set.
+  arma::mat&& weight;
+
+  //! The parameters offset.
+  const size_t offset;
+
+  //! Do not update the bias parameters if the module doesn't implement the
+  //! Bias() or Model() function.
+  template<typename T>
+  typename std::enable_if<
+      !HasBiasCheck<T, arma::mat&(T::*)()>::value &&
+      !HasModelCheck<T>::value, size_t>::type
+  LayerSize(T* layer) const;
+
+  //! Update the bias parameters if the module implements the Model() function.
+  template<typename T>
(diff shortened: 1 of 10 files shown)


src/mlpack/methods/ann/visitor/bias_set_visitor_impl.hpp
```diff
@@ -0,0 +1,94 @@
...
+ * mlpack is free software; you may redistribute it and/or modify it under the
+ * terms of the 3-clause BSD license.  You should have received a copy of the
+ * 3-clause BSD license along with mlpack.  If not, see
+ * http://www.opensource.org/licenses/BSD-3-Clause for more information.
+ */
+#ifndef MLPACK_METHODS_ANN_VISITOR_BIAS_SET_VISITOR_IMPL_HPP
+#define MLPACK_METHODS_ANN_VISITOR_BIAS_SET_VISITOR_IMPL_HPP
+
+// In case it hasn't been included yet.
+#include "bias_set_visitor.hpp"
```
zoq: I know there isn't a test for every visitor, at least there isn't one that only tests the visitor. In this case it's not used, at least not right now, so I think it makes sense to write a simple test for the visitor.

walragatver:
Hi @zoq,
Sorry for the slow response. You were quite correct about adding tests. The implementation had some flaws in it. 
In order to test this visitor I have used it in #1920. I only required this visitor to implement `weight_norm` as per the requirement. Please review this in your free time.

zoq:
Thanks for the update, using it for the weight normalisation is good, but I think it would be still a good idea if we could write a simple test just for the visitor. We could create a simple Linear layer set some static weights and check if the visitor returns the correct bias. Let me know what you think.

walragatver:
Hi @zoq, I have added the test in a new file namely `ann_visitor_test.cpp`. Let me know what you think.

src/mlpack/tests/ann_visitor_test.cpp
```diff
@@ -0,0 +1,55 @@
...
+
+#include <boost/test/unit_test.hpp>
+#include "test_tools.hpp"
+
+using namespace mlpack;
+using namespace mlpack::ann;
+
+BOOST_AUTO_TEST_SUITE(ANNVisitorTest);
+
+BOOST_AUTO_TEST_CASE(BiasSetVisitorTest)
```
zoq: If you can add a super simple description what this test does, I think this is ready.

walragatver: Hi @zoq. Done.

src/mlpack/methods/ann/visitor/bias_set_visitor.hpp
```diff
@@ -0,0 +1,80 @@
...
+
+#include <mlpack/methods/ann/layer/layer_traits.hpp>
+
+#include <boost/variant.hpp>
+
+namespace mlpack {
+namespace ann {
+
+/**
+ * BiasSetVisitor update the module bias parameters given the parameters set.
```
ShikharJ: I think it should be `updates` here.

src/mlpack/methods/ann/visitor/bias_set_visitor_impl.hpp
```diff
@@ -0,0 +1,94 @@
...
+}
+
+template<typename T>
+inline typename std::enable_if<
+    !HasBiasCheck<T, arma::mat&(T::*)()>::value &&
+    HasModelCheck<T>::value, size_t>::type
+BiasSetVisitor::LayerSize(T* layer) const
+{
+  size_t modelOffset = 0;
+  for (size_t i = 0; i < layer->Model().size(); ++i)
```
ShikharJ: Can you please add a newline above this for loop for aesthetics?

src/mlpack/methods/ann/visitor/bias_set_visitor_impl.hpp
```diff
@@ -0,0 +1,94 @@
...
+inline typename std::enable_if<
+    HasBiasCheck<T, arma::mat&(T::*)()>::value &&
+    HasModelCheck<T>::value, size_t>::type
+BiasSetVisitor::LayerSize(T* layer) const
+{
+  layer->Bias() = arma::mat(weight.memptr() + offset,
+      layer->Bias().n_rows, layer->Bias().n_cols, false, false);
+
+  size_t modelOffset = layer->Bias().n_elem;
+  for (size_t i = 0; i < layer->Model().size(); ++i)
```
ShikharJ: Same here?

walragatver:
Hi @ShikharJ, I have done the required changes.

ShikharJ:
Build failure seems unrelated to the code.

ShikharJ:
Thanks @walragatver for the contribution.

PR #1814, opened 2019-03-23
Implement Encoding Class with Dictionary Encoding Policy

jeffin143:
Added a function to Encode the Documents as Dictionary Encoding

Here we simply assign a word (or a character) to a numeric index and treat the data set as categorical.

Working on a test

PR #1960, opened 2019-07-30
[Ready] String encoding fixes

lozhnikov:
The PR absorbs #1814.

- [x] The API is stable.
  - [x] Benchmarks for `boost::optional` as the output of the tokenizers are done.
  - [x] The tokenizers don't return `boost::optional` (not needed).
- [x] The string encoding policies API is stable.
- [x] All algorithms work fine.
- [x] Everything looks great.


lozhnikov:
@jeffin143 The string encoding policies API is stable, so you can fix your implementation of the bag of words algorithm and Tf-Idf.

lozhnikov:
# Background

Currently each tokenizer has a special method (`IsTokenEmpty()`) that returns `true` if the given token is empty. Hence we can't use lambda expressions as tokenizers since the tokenizers return various value types. Modern C++ has an obvious solution: `std::optional`. I tried to investigate if we can use `boost::optional` as the return value type of the tokenizers (i.e. the tokenizer returns `nullopt` if the next token is empty).

Source code: [boost-optional-tests.zip](https://github.com/mlpack/mlpack/files/3449020/boost-optional-tests.zip)

I used the same compilers, the same compiler options, the same benchmark options and the same input data as [here](https://github.com/mlpack/mlpack/pull/1814#issuecomment-514687037).

**All figures show the time of the described tests**.

**The figures related to the same tests have the same scale**.

# Benchmarks
## Tokenizer comparison. Part 1.
### These tokenizers split the given string using the given set of delimiters

There are three tests for two different tokenizers.
- The first tokenizer returns `boost::string_view`.
- The second tokenizer returns `boost::optional<boost::string_view>`.

The first test just splits a number of strings into tokens. The second test does the same and saves the result into `std::vector`. The third test extracts tokens and saves them into `std::unordered_set`.

### g++ 9.1.1

g++_tokenizers_benchmark

### clang++ 8.0

clang++_tokenizers_benchmark

### clang++ 8.0 and libc++

clang++_libc++_tokenizers_benchmark

## Tokenizer comparison. Part 2.
### The tokenizers below extract a character at a time

There are three tests for two different tokenizers.
- The first tokenizer returns `int`.
- The second tokenizer returns `boost::optional<int>`.

The first test just splits a number of strings into tokens. The second test does the same and saves the result into `std::vector`. The third test extracts tokens and saves them into `std::unordered_set`.

### g++ 9.1.1

g++_char_extract_tokenizers_benchmark

### clang++ 8.0

clang++_char_extract_tokenizers_benchmark

### clang++ 8.0 and libc++

clang++_libc++_char_extract_tokenizers_benchmark

## Encode() tests. Part 1.
### First consider the `Encode()` function that accepts a tokenizer which splits the given string using the given set of delimiters

There are two tests for two different tokenizers.
- The first tokenizer returns `boost::string_view`.
- The second tokenizer returns `boost::optional<boost::string_view>`.

The first test encodes a number of strings in two passes:
1. Pass 1: Extract all tokens and add them to a dictionary (`std::unordered_map`).
2. Pass 2: Encode all tokens using the dictionary and save them somewhere.

The second test encodes a number of strings in one pass:
1. For all tokens: extract a token, add it to a dictionary (`std::unordered_map`), encode the token and save the result.

### g++ 9.1.1

g++_encode_benchmark

### clang++ 8.0

clang++_encode_benchmark

### clang++ 8.0 and libc++

clang++_libc++_encode_benchmark

## Encode() tests. Part 2.
### Now consider the `Encode()` function that accepts a tokenizer which extracts a character at a time

There are two tests for two different tokenizers.
- The first tokenizer returns `int`.
- The second tokenizer returns `boost::optional<int>`.

The first test encodes a number of strings in two passes:
1. Pass 1: Extract all tokens and add them to a dictionary (`std::array`).
2. Pass 2: Encode all tokens using the dictionary and save them somewhere.

The second test encodes a number of strings in one pass:
1. For all tokens: extract a token, add it to a dictionary (`std::array`), encode the token and save the result.

### g++ 9.1.1

g++_char_encode_benchmark

### clang++ 8.0

clang++_char_encode_benchmark

### clang++ 8.0 and libc++

clang++_libc++_char_encode_benchmark

# Conclusion

- The very first figure (Tokenizers, part 1, g++ 9.1.1) looks like an anomaly. I think `g++` optimized something out, but I didn't find out why it happens.
- We can neglect the performance difference between tokenizers that return `boost::string_view` and `boost::optional<boost::string_view>`.
- The tokenizer that returns `boost::optional<int>` works much slower than the one that returns `int`. However we should take into account that the dictionary encoding algorithm is pretty simple in case of encoding individual characters. @jeffin143 Perhaps I will repeat the benchmark as soon as Jeffin implements other algorithms.
- I think it is reasonable to keep the present API i.e. the `Encode()` function has the `TokenizerType`  template argument  and each tokenizer has the `IsTokenEmpty()` method. This is rather robust approach, and we can switch to `boost::optional` without external changes anytime.

Thanks for the attention!

lozhnikov:
AppVeyor throws the following error:
`multiple versions of a defaulted special member functions are not allowed`.

Looks like it's a MSVS bug (see [stackoverflow](https://stackoverflow.com/questions/36657243/c-multiple-versions-of-a-defaulted-special-member-functions-error-in-msvc-2)). Old versions of Visual Studio (however, MSVS2017 build failed too) doesn't compile the following code:
```c++
class Foo
{
  template<typename ...Args>
  Foo(Args&& ...args);

  // We has to declare explicitly the copy-constructor since the parameter-pack
  // constructor hides it.
  Foo(const Foo&) = default;
};
```
@rcurtin, @zoq Is there an easy fix (except adding the implementation of the required constructors)?

The constructors are pretty simple, but `= default` looks better than the explicit implementation.

rcurtin:
@lozhnikov I took a look... not too surprising that MSVS fails something that it shouldn't even after they said they fixed it.  We've definitely had a lot of those issues before.  I agree that it's uglier but I think your solution in 2a5464e is probably unfortunately the right way to go...

`std::optional` is C++17, which might be a little hard.  I don't have too much of a problem upgrading so long as we can make sure that the newest mlpack is still compilable on recent Ubuntu LTS versions and supported RHEL versions (RHEL7 is current but from mid-2015, probably most departments have been able to transition from RHEL6; RHEL8 is just becoming stable now but we would want to wait for a couple years before having that be required to compile mlpack).  However I haven't taken a look into where exactly compiler support is on each of those distros (and how safe any of these features are to actually try to use on MSVS or other compilers...).

lozhnikov:
@rcurtin Thanks for the response! Yes, `std::optional` is a C++17 feature. We can't use it until Ubuntu LTS and RHEL support modern compilers. However, we can use `boost::optional`. If I am not mistaken Boost has the `optional` class since Boost 1.57.

lozhnikov:
@rcurtin , @zoq I think the PR is ready. Currently it contains only dictionary encoding. The `Tf-Idf` algorithm (#1969) isn't ready yet. I am going to perform some benchmarks as soon as Jeffin finishes `Tf-Idf`. So, it'll take some time.

Do we want to see this PR in the upcoming release? I am not quite sure that it has any sense without `Tf-Idf`.

zoq:
Agreed, this makes more sense in combination with ``Tf-Idf``.

src/mlpack/core/data/string_encoding_impl.hpp
```diff
@@ -0,0 +1,208 @@
+/**
+ * @file string_encoding_impl.hpp
+ * @author Jeffin Sam
```
zoq: I think this was a team effort, so maybe it would be fair to list @lozhnikov as author as well?

src/mlpack/core/data/tokenizers/split_by_any_of.hpp
```diff
@@ -0,0 +1,113 @@
...
+
+ private:
+  /**
+   * The function finds the first character in the given string view equal to 
+   * any of the delimiters and returns the position of the character or 
+   * boost::string_view::npos if no such character is found.
+   *
+   * @param str String where to find the character.
+   */
+  size_t FindFirstDelimiter(boost::string_view str) const
```
zoq: I'm wondering if it makes sense to use ``const boost::string_view& str`` here?

src/mlpack/core/data/string_encoding_policies/dictionary_encoding_policy.hpp
```diff
@@ -0,0 +1,114 @@
...
+  * The function initializes the output matrix.
+  *
+  * @param output Output matrix to store the encoded results (sp_mat or mat).
+  * @param datasetSize The number of strings in the input dataset.
+  * @param maxNumTokens The maximum number of tokens in the strings of the 
+                        input dataset.
+  * @param dictionarySize The size of the dictionary (not used).
+  */
+  template<typename MatType>
+  static void InitMatrix(MatType& output,
```
zoq: Would probably use ``const size_t`` just to be more clear that this parameter isn't going to be changed.

src/mlpack/core/data/string_encoding_impl.hpp
lozhnikov: I agree.

src/mlpack/core/data/string_encoding_policies/dictionary_encoding_policy.hpp
lozhnikov: Fixed everywhere.

src/mlpack/core/data/tokenizers/split_by_any_of.hpp
lozhnikov: I think the present implementation could be a bit faster since the size of `boost::string_view` is equal to 16 bytes and the copy-constructor is trivial.

src/mlpack/core/data/tokenizers/split_by_any_of.hpp
```diff
@@ -0,0 +1,114 @@
...
+
+ private:
+  /**
+   * The function finds the first character in the given string view equal to 
+   * any of the delimiters and returns the position of the character or 
+   * boost::string_view::npos if no such character is found.
+   *
+   * @param str String where to find the character.
+   */
+  size_t FindFirstDelimiter(const boost::string_view str) const
```
lozhnikov: @zoq Thanks for the review. I pass the argument without reference since I think the present implementation could be a bit faster since the size of `boost::string_view` is equal to 16 bytes and the copy-constructor is trivial.

zoq: Thanks for the clarification, makes sense.

rcurtin: I made a comment on the exact same thing but then I found your response here.  Sounds good to me.  Even if it is not actually faster, with only 16 bytes, it definitely will not be a bottleneck. :+1:

src/mlpack/tests/CMakeLists.txt
```diff
@@ -110,7 +111,6 @@ add_executable(mlpack_test
   timer_test.cpp
   tree_test.cpp
   tree_traits_test.cpp
-  scaling_test.cpp
```
rcurtin: Any reason to remove this test?  Or did I miss something?

src/mlpack/tests/string_encoding_test.cpp
```diff
@@ -0,0 +1,441 @@
...
+
+//! Common input for some tests.
+static vector<string> stringEncodingInput = {
+    "mlpack is an intuitive, fast, and flexible C++ machine learning library "
+    "with bindings to other languages. ",
+    "It is meant to be a machine learning analog to LAPACK, and aims to "
+    "implement a wide array of machine learning methods and functions "
+    "as a \"swiss army knife\" for machine learning researchers.",
+    "In addition to its powerful C++ interface, mlpack also provides "
+    "command-line programs and Python bindings."
```
rcurtin: Might be worth making things a bit harder by adding some Unicode characters, just to make sure there aren't any issues with those.  (I doubt it, I think all the code is fine as is, but never hurts to check. :))

src/mlpack/tests/CMakeLists.txt
lozhnikov: Nice catch! Looks like an incorrect merge.

src/mlpack/core/data/string_encoding_dictionary.hpp
```diff
@@ -0,0 +1,347 @@
...
+#include <mlpack/prereqs.hpp>
+#include <mlpack/core/boost_backport/boost_backport_string_view.hpp>
+#include <unordered_map>
+#include <deque>
+#include <array>
+
+namespace mlpack {
+namespace data {
+
+/*
```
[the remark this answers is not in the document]
lozhnikov: Yes, you're right. I didn't see it.

src/mlpack/core/data/string_encoding_dictionary.hpp
```diff
@@ -0,0 +1,347 @@
...
+ * http://www.opensource.org/licenses/BSD-3-Clause for more information.
+ */
+#ifndef MLPACK_CORE_DATA_STRING_ENCODING_DICTIONARY_HPP
+#define MLPACK_CORE_DATA_STRING_ENCODING_DICTIONARY_HPP
+
+#include <mlpack/prereqs.hpp>
+#include <mlpack/core/boost_backport/boost_backport_string_view.hpp>
+#include <unordered_map>
+#include <deque>
+#include <array>
```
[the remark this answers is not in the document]
lozhnikov: `std::array` and `std::deque` aren't used. However, `std::unordered_map` is used a few times. Probably it's better to leave as is in order to minimize the prepocessing time.

src/mlpack/core/data/string_encoding_dictionary.hpp
```diff
@@ -0,0 +1,347 @@
...
+  //! Modify the mapping.
+  MapType& Mapping() { return mapping; }
+
+  /**
+   * Serialize the class to the given archive.
+   */
+  template<typename Archive>
+  void serialize(Archive& ar, const unsigned int /* version */)
+  {
+    ar & BOOST_SERIALIZATION_NVP(mapping);
```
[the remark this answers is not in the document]
lozhnikov: Fixed.

src/mlpack/tests/string_encoding_test.cpp
```diff
@@ -0,0 +1,441 @@
...
+template<>
+void CheckDictionaries(const StringEncodingDictionary<int>& expected,
+                       const StringEncodingDictionary<int>& obtained)
+{
+  // MapType is equal to std::arry<size_t, 256>.
+  using MapType = typename StringEncodingDictionary<int>::MapType;
+
+  const MapType& expectedMapping = expected.Mapping();
+  const MapType& mapping = obtained.Mapping();
+
```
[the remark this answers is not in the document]
lozhnikov: Fixed.

src/mlpack/core/data/string_encoding_dictionary.hpp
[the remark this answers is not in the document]
lozhnikov: Fixed.

src/mlpack/core/data/string_encoding_impl.hpp
```diff
@@ -0,0 +1,209 @@
...
+    encodingPolicy(std::forward<ArgTypes>(args)...)
+{
+}
+
+template<typename EncodingPolicyType, typename DictionaryType>
+StringEncoding<EncodingPolicyType, DictionaryType>::StringEncoding(
+    EncodingPolicyType encodingPolicy) :
+    encodingPolicy(std::move(encodingPolicy))
+{
+}
```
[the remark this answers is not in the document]
lozhnikov: Fixed everywhere.

src/mlpack/tests/CMakeLists.txt
lozhnikov: Fixed.

src/mlpack/tests/string_encoding_test.cpp
lozhnikov: You're right, I'll work on it.

jeffin143:
s:keep-open

rcurtin:
@lozhnikov it seems like this one got overlooked---I didn't see your responses to the PR comments until now.  Thanks for handling that.  Should I go ahead and merge this or did you have anything else that you wanted to handle before then?

lozhnikov:
@rcurtin I fixed a few issues you pointed out. However I think I had forgotten about a feature called `n-gram` (https://scikit-learn.org/stable/modules/generated/sklearn.feature_extraction.text.TfidfVectorizer.html). I think it could be quite useful. I'd like to make sure our API allows to implement it. Then I'll merge the PR. Unfortunately I hadn't time to dig into it.

rcurtin:
Ok, sounds good.  I think it's okay if we merge this now and then add the n-gram support in a separate PR, but however you want to do it sounds good to me. :+1:

rcurtin:
Hey there @lozhnikov, just wanted to check in on this one and see what you think.  I'm still excited about it, but I think we've all been busy over the past months.  Do you think that we should merge as-is, and then open an issue for the n-gram support?  Let me know how I can help out. :+1:

lozhnikov:
# N-gram bag of words encoders comparison

I implemented a number of tests. **Sources:** [ngram-tests.zip](https://github.com/mlpack/mlpack/files/4246042/ngram-tests.zip)
As usual I used *Les Miserables* (in plain text) as the input data.

I implemented the following algorithms:
1. **Naive**. It means that each n-gram is a string (ordinary `std::string`) which contains space separated tokens. The algorithm requires a special tokenizer. Other classes weren't changed.
2. **Deque**. The algorithm looks like the previous one. The tokenizer returns a `std::string`, but it uses a `std::deque` in order to reduce the number of token extractions. It works like a conveyor.
3. **Vector**. Each n-gram is a `std::vector<boost::string_view>`. It requires a special dictionary. The dictionary key is a `std::vector<boost::string_view>`. Unfortunately since each `string_view` refers to a string we have to store `std::unordered_set<boost::string_view>` and `std::deque<std::string>` in order to avoid `string_view` invalidation. The set guarantees that we store each token only once.
4. **Vector V2**. The algorithm is similar to the previous one. Again, each n-gram is a `std::vector<boost::string_view>`. The dictionary differs a bit. The dictionary key is a `std::vector<size_t>` where each element of the vector is the ID of the corresponding token. Unfortunately we have to store a map in order to get token IDs (`std::unordered_map<boost::string_view, size_t>`).
5. **Vector V3**. Again, each n-gram is a `std::vector<boost::string_view>`. The dictionary key is a `std::vector<boost::string_view>`. In order to avoid `string_view` invalidation we add each token into a `std::deque<std::string>` and into the dictionary as a vector of length 1.
6. **Static**. The algorithm is the same as **Vector**. The dictionary key is a `std::array<boost::string_view, N>`.
7. **StrView**. Each n-gram is an ordinary `boost::string_view`. The algorithm doesn't ignore the delimiters, therefore two identical n-grams divided by different delimiters are mapped to different numbers.
8. **StrView V2**. Again each n-gram is an ordinary `boost::string_view`, but the hash function and the equality operator ignore the delimiters.

### n-gram range: (1, 1)
g++_encode_benchmark-(1,1)

### n-gram range: (1, 3)
g++_encode_benchmark-(1,3)

### n-gram range: (1, 5)
g++_encode_benchmark-(1,5)

### n-gram range: (1, 7)
g++_encode_benchmark-(1,7)

### n-gram range: (3, 3)
g++_encode_benchmark-(3,3)

### n-gram range: (3, 5)
g++_encode_benchmark-(3,5)

### n-gram range: (3, 7)
g++_encode_benchmark-(3,7)

### n-gram range: (5, 5)
g++_encode_benchmark-(5,5)

### n-gram range: (5, 7)
g++_encode_benchmark-(5,7)

### n-gram range: (7, 7)
g++_encode_benchmark-(7,7)

## Conclusion

1. I didn't expect, but an ordinary `std::string` works well with n-grams. I thought `std::vector<boost::string_view>` (or `std::array<boost::string_view, N>`) should work better but it's not the case.
2. Unfortunately, `std::deque` doesn't make any significant difference.
3. The last algorithm (**StrView V2**) sometimes outperforms the **Naive** approach. But the difference isn't significant.

Anyway, it is possible to implement the n-grams support in terms of the API.

lozhnikov:
@rcurtin Sorry for the slow response. I promised to sort it out in October-November, but things didn't go as planned.

I've done a small investigation. The n-gram support doesn't break the API. Perhaps it's a good idea to make `MapType` a template parameter in `StringEncodingDictionary`, but we can do it later since it's an internal part of the API. And I'm not 100% sure it's really needed.

I think we should move the n-gram support into another PR otherwise we'll never merge this one:). I'm a bit surprised that an ordinary `std::string` works fine with n-grams. I thought `std::array<boost::string_view>` should outperform ordinary strings like `string_view`s do. But it's not the case. I'd like to dig into it a bit more.

jeffin143:
You are a hero @lozhnikov :)

lozhnikov:
@jeffin143 Thanks:)

@rcurtin I think the PR is ready now. I added a couple of unicode symbols (several emojis) to the test. Looks like everything works fine. However, we don't have a proper unicode support without wide characters. It's easy to add the wide strings support, but I'm not sure it's needed since multibyte encodings like UTF-8 are more popular than fixed-size Unicode. Anyway, this PR is ready. So, let's wait for the results of the automated checks.

src/mlpack/tests/CMakeLists.txt
rcurtin: Hmm, maybe I overlooked something but it seems like this test is still removed in the diff.

src/mlpack/tests/string_encoding_test.cpp
```diff
@@ -0,0 +1,524 @@
...
+  for (auto& keyValue : dictionary.Mapping())
+  {
+    keysCount[keyValue.second]++;
+
+    BOOST_REQUIRE_EQUAL(keysCount[keyValue.second], 1);
+  }
+
+  arma::mat expected = {
+    { 1,  2, 0, 0, 0 },
```
rcurtin:
```suggestion
    { 1, 2, 0, 0, 0 },
```
Super pedantic comment. :)

lozhnikov: I've already added `scaling_test.hpp`. See line 75. Now the lines are in alphabetic order.

lozhnikov:
> Really cool results with the n-gram experiments; I was surprised that the naive approach performed so well! Anyway, I do agree, perhaps better to put that in a different PR. :)

Looks like other approaches overcomplicate the dictionary. I think they increase the number of lookup operations. I found out recently that `std::string` stores short strings (up to 15 or 16 bytes if I am not mistaken) on the stack. It's interesting. I didn't know.

> Everything looks great to me, just a couple tiny comments I found when passing through the diff. Should we merge this once it passes the CI tests?

I removed the space. Yes, I think we can merge it.

src/mlpack/tests/string_encoding_test.cpp
rcurtin: Weird, I think that I was looking at the wrong diff---it looks like I wasn't seeing all changes.  It seems fine, sorry about that. :+1:

rcurtin:
Perfect, sounds good.  The AppVeyor build might take a little while since there are so many PRs building these days, but I imagine it will pass, then we can merge it. :+1:

PR #1961, opened 2019-07-30
Add CGAN

saksham189:


saksham189:
@ShikharJ  currently for generating samples from GAN we do `gan.Generator().Forward(noise, samples);`. However for cgan we also need to pass the labels and concatenate them. Do you think we should add a method `Forward(noise, labels, samples)` specific to CGAN for generating samples.

zoq:
> @ShikharJ currently for generating samples from GAN we do `gan.Generator().Forward(noise, samples);`. However for cgan we also need to pass the labels and concatenate them. Do you think we should add a method `Forward(noise, labels, samples)` specific to CGAN for generating samples.

Personally I like the idea, makes is easier to generate samples.

saksham189:
@ShikharJ this is ready for review.

src/mlpack/methods/ann/gan/gan_policies.hpp
```diff
@@ -78,6 +78,10 @@ class WGAN { /* Nothing to do here */ };
  */
 class WGANGP { /* Nothing to do here */ };
 
+
+
+class CGAN { /* Nothing to do here */ };
```
ShikharJ: Can this be put in alphabetic order?

saksham189:
> Thanks for the contribution, I'd also appreciate if you could provide the output of the GAN as well to check if everything works well.

Hey, @ShikharJ thanks for the review. I think that Toshal has a PR open regarding fixing the gradient of the GAN so, we should test that first and then proceed to testing this. 
Also I think that the training would take many days if I do it locally on my machine. Would you be able to provide me access to the server that Toshal was using for testing this.

rcurtin:

PR #1962, opened 2019-07-31
[GSoC] [WIP] Subspace tree

robertohueso:
This PR is the implementation of a new data structure called "Subspace Tree". This data structure is defined in the section 3 of the paper [Fast High-dimensional Kernel Summations Using the Monte Carlo Multipole Method](http://papers.nips.cc/paper/3539-fast-high-dimensional-kernel-summations-using-the-monte-carlo-multipole-method.pdf).

The idea is to build a space-partitioning tree with orthogonal bases associated with each node to enable dimensionality reduction within the tree.

This tree could be useful to improve performance on some algorithms like KDE.

### DIFF
--- a/src/mlpack/core/math/lin_alg.cpp
+++ b/src/mlpack/core/math/lin_alg.cpp
@@ -43,7 +43,7 @@ void mlpack::math::VectorPower(arma::vec& vec, const double power)
 void mlpack::math::Center(const arma::mat& x, arma::mat& xCentered)
 {
   // Get the mean of the elements in each row.
-  arma::vec rowMean = arma::sum(x, 1) / x.n_cols;
+  arma::vec rowMean = arma::mean(x, 1);
 
   xCentered = x - arma::repmat(rowMean, 1, x.n_cols);
 }
--- a/src/mlpack/methods/kde/kde_impl.hpp
+++ b/src/mlpack/methods/kde/kde_impl.hpp
@@ -61,7 +61,9 @@ KDE(const double relError,
     const double mcProb,
     const size_t initialSampleSize,
     const double mcEntryCoef,
-    const double mcBreakCoef) :
+    const double mcBreakCoef,
+    const bool pca,
+    const double pcaVarRetained) :
     kernel(kernel),
     metric(metric),
     referenceTree(nullptr),
@@ -72,7 +74,9 @@ KDE(const double relError,
     trained(false),
     mode(mode),
     monteCarlo(monteCarlo),
-    initialSampleSize(initialSampleSize)
+    initialSampleSize(initialSampleSize),
+    pca(pca),
+    pcaVarRetained(pcaVarRetained) // TODO check 0 < var < 1.
 {
   CheckErrorValues(relError, absError);
   MCProb(mcProb);
@@ -106,7 +110,9 @@ KDE(const KDE& other) :
     mcProb(other.mcProb),
     initialSampleSize(other.initialSampleSize),
     mcEntryCoef(other.mcEntryCoef),
-    mcBreakCoef(other.mcBreakCoef)
+    mcBreakCoef(other.mcBreakCoef),
+    pca(other.pca),
+    pcaVarRetained(other.pcaVarRetained)
 {
   if (trained)
   {
@@ -152,7 +158,9 @@ KDE(KDE&& other) :
     mcProb(other.mcProb),
     initialSampleSize(other.initialSampleSize),
     mcEntryCoef(other.mcEntryCoef),
-    mcBreakCoef(other.mcBreakCoef)
+    mcBreakCoef(other.mcBreakCoef),
+    pca(other.pca),
+    pcaVarRetained(other.pcaVarRetained)
 {
   other.kernel = std::move(KernelType());
   other.metric = std::move(MetricType());
@@ -168,6 +176,8 @@ KDE(KDE&& other) :
   other.initialSampleSize = KDEDefaultParams::initialSampleSize;
   other.mcEntryCoef = KDEDefaultParams::mcEntryCoef;
   other.mcBreakCoef = KDEDefaultParams::mcBreakCoef;
+  other.pca = KDEDefaultParams::pca;
+  other.pcaVarRetained = KDEDefaultParams::pcaVarRetained;
 }
 
 template<typename KernelType,
@@ -214,6 +224,8 @@ operator=(KDE other)
   this->initialSampleSize = other.initialSampleSize;
   this->mcEntryCoef = other.mcEntryCoef;
   this->mcBreakCoef = other.mcBreakCoef;
+  this->pca = other.pca;
+  this->pcaVarRetained = other.pcaVarRetained;
 
   return *this;
 }
@@ -368,6 +380,9 @@ Evaluate(MatType querySet, arma::vec& estimations)
                                   "referenceSet dimensions don't match");
     }
 
+    if (pca && std::is_same<KernelType, kernel::GaussianKernel>::value)
+      ComputePCA(*referenceTree);
+
     Timer::Start("computing_kde");
 
     // Evaluate.
@@ -384,6 +399,7 @@ Evaluate(MatType querySet, arma::vec& estimations)
                               metric,
                               kernel,
                               monteCarlo,
+                              pca,
                               false);
 
     // Create traverser.
@@ -482,6 +498,7 @@ Evaluate(Tree* queryTree,
                             metric,
                             kernel,
                             monteCarlo,
+                            pca,
                             false);
 
   // Create traverser.
@@ -551,6 +568,7 @@ Evaluate(arma::vec& estimations)
                             metric,
                             kernel,
                             monteCarlo,
+                            pca,
                             true);
 
   if (mode == DUAL_TREE_MODE)
@@ -720,6 +738,8 @@ serialize(Archive& ar, const unsigned int version)
     ar & BOOST_SERIALIZATION_NVP(initialSampleSize);
     ar & BOOST_SERIALIZATION_NVP(mcEntryCoef);
     ar & BOOST_SERIALIZATION_NVP(mcBreakCoef);
+    ar & BOOST_SERIALIZATION_NVP(pca);
+    ar & BOOST_SERIALIZATION_NVP(pcaVarRetained);
   }
   else if (Archive::is_loading::value)
   {
@@ -728,6 +748,8 @@ serialize(Archive& ar, const unsigned int version)
     initialSampleSize = KDEDefaultParams::initialSampleSize;
     mcEntryCoef = KDEDefaultParams::mcEntryCoef;
     mcBreakCoef = KDEDefaultParams::mcBreakCoef;
+    pca = KDEDefaultParams::pca;
+    pcaVarRetained = KDEDefaultParams::pcaVarRetained;
   }
 
   // If we are loading, clean up memory if necessary.
--- a/src/mlpack/methods/kde/kde_impl.hpp
+++ b/src/mlpack/methods/kde/kde_impl.hpp
@@ -923,6 +923,23 @@ ComputePCA(Tree& rootNode)
       stat.EigVec().shed_cols(newDimension, normalizedEigVal.size() - 1);
       stat.EigVal().shed_rows(newDimension, normalizedEigVal.size() - 1);
     }
+
+    // Pre-cache computations that will be needed for the KDE evaluation.
+    stat.MinKernelRecon() = DBL_MAX;
+    stat.MaxKernelRecon() = DBL_MIN;
+    for (size_t i = 0; i < node->NumDescendants(); ++i)
+    {
+      const arma::vec& rPoint = node->Dataset().unsafe_col(i);
+      const arma::vec rProj = stat.EigVec().t() * (rPoint - stat.Mean());
+      const arma::vec rRecon = stat.EigVec() * rProj + stat.Mean();
+      const double rReconKernelValue = kernel.Evaluate(metric.Evaluate(rRecon,
+                                                                       rPoint));
+
+      if (rReconKernelValue > stat.MaxKernelRecon())
+        stat.MaxKernelRecon() = rReconKernelValue;
+      if (rReconKernelValue < stat.MinKernelRecon())
+        stat.MinKernelRecon() = rReconKernelValue;
+    }
   }
   Timer::Stop("computing_dimensionality_reduction");
 }
--- a/src/mlpack/methods/kde/kde_impl.hpp
+++ b/src/mlpack/methods/kde/kde_impl.hpp
@@ -763,6 +763,172 @@ void KDE<KernelType,
          TreeType,
          DualTreeTraversalType,
          SingleTreeTraversalType>::
+ComputePCA(Tree& rootNode, const double varToRetain)
+{
+  // TODO check method parameters.
+  Timer::Start("computing_dimensionality_reduction");
+  std::stack<Tree*> nodes;
+  KDEStackRules<Tree> reverseRules(nodes);
+  SingleTreeTraversalType<KDEStackRules<Tree>> reverseTraverser(reverseRules);
+  reverseTraverser.Traverse(0, rootNode);
+
+  while(!nodes.empty())
+  {
+    Tree* node = nodes.top();
+    nodes.pop();
+    KDEStat& stat = node->Stat();
+
+    // Compute base.
+    if (node->IsLeaf())
+    {
+      const size_t numPoints = node->NumPoints();
+      const size_t firstPoint = node->Point(0);
+      arma::mat pointsInTheNode =
+          node->Dataset().cols(firstPoint, firstPoint + numPoints - 1);
+      math::Center(pointsInTheNode, pointsInTheNode);
+
+      // Calculate base mean.
+      stat.Mean() = arma::mean(pointsInTheNode, 1);
+
+      // Right singular values. Unused.
+      arma::mat v;
+
+      // Singular value decomposition.
+      if (pointsInTheNode.n_rows > pointsInTheNode.n_cols)
+        arma::svd_econ(stat.EigVec(), stat.EigVal(), v, pointsInTheNode, "l");
+      else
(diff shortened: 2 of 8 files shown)


robertohueso:
So far I have been trying to implement this as a `StatisticType` for existing trees, since the orthogonal bases are computed bottom->up, but I'm mostly convinced that this would be quite inconvenient.
The most obvious issue for me is `StatisticType` is built for all trees using the constructor:
```c++
template<typename TreeType>
StatisticType(TreeType& node);
```
This means:
* Computing orthogonal base, even if the method doesn't have dimensionality reduction activated.
* No parameter selection for the dimensionality reduction method (e.g. amount of variance to preserve in PCA).

I'm also thinking that these issues could be solved by creating a bottom->up traversal for all trees, but at the moment none of these exist.

Does anyone has any thoughts on this? :smile:

src/mlpack/core/math/lin_alg.cpp
```diff
@@ -43,7 +43,7 @@ void mlpack::math::VectorPower(arma::vec& vec, const double power)
 void mlpack::math::Center(const arma::mat& x, arma::mat& xCentered)
 {
   // Get the mean of the elements in each row.
-  arma::vec rowMean = arma::sum(x, 1) / x.n_cols;
+  arma::vec rowMean = arma::mean(x, 1);
```
rcurtin: Hehe, this is a nice improvement.  Thanks. :)

src/mlpack/methods/kde/kde_stat.hpp
```diff
@@ -48,6 +49,128 @@ class KDEStat
...
+          arma::svd(newEigVec, newEigVal, rightEigVal, newEigenProblem);
+
+          arma::mat upsilon(U.n_rows, U.n_cols + v.n_cols);
+          upsilon.cols(0, U.n_cols - 1) = U;
+          upsilon.cols(U.n_cols, upsilon.n_cols - 1) = v;
+
+          // Update node's orthogonal base.
+          mean = (1 / (double) P) * (N * mean0 + M * mean1);
+          eigVal = newEigVal;
+          eigVec = upsilon * newEigVec;
```
rcurtin: I haven't read the eigenspaces merging paper yet so I can't comment on this, but I'll review it more closely later.

src/mlpack/methods/kde/kde_stat.hpp
```diff
@@ -48,6 +49,128 @@ class KDEStat
     {
       validCentroid = false;
     }
+
+    // Build orthogonal base if possible.
+    if (tree::TreeTraits<TreeType>::BinaryTree)
```
rcurtin: I see that this is specific to binary trees, and the reason seems to be that the subspace merging is done for two subspaces at a time only.  Maybe you have already thought of doing this (the PR is a WIP after all) but it seems to me that you could handle a node with three children by merging the subspaces of child 0 and 1, and then subsequently merging that resulting subspace with the subspace of child 2.

In any case, probably best to get it working just with the kd-tree first. :)

src/mlpack/methods/kde/kde_stat.hpp
```diff
@@ -101,12 +233,18 @@ class KDEStat
...
+      ar & BOOST_SERIALIZATION_NVP(mean);
     }
     else if (Archive::is_loading::value)
     {
       mcBeta = -1;
       mcAlpha = -1;
       accumAlpha = -1;
+      eigVec = arma::mat();
+      eigVal = arma::vec();
+      mean = arma::vec();
```
rcurtin: I think `.clear()` could also be used here.

src/mlpack/methods/pca/decomposition_policies/exact_svd_method.hpp
```diff
@@ -38,12 +38,12 @@ class ExactSVDPolicy
...
-             arma::mat& transformedData,
-             arma::vec& eigVal,
-             arma::mat& eigvec,
-             const size_t /* rank */)
+  static void Apply(const arma::mat& data,
+                    const arma::mat& centeredData,
+                    arma::mat& transformedData,
+                    arma::vec& eigVal,
+                    arma::mat& eigvec,
+                    const size_t /* rank */)
```
rcurtin: Ah, yes, this can be static, nice catch.

src/mlpack/methods/kde/kde_rules_impl.hpp
```diff
@@ -76,16 +78,58 @@ double KDERules<MetricType, KernelType, TreeType>::BaseCase(
...
+    const arma::vec qProj = stat.EigVec().t() * (qPoint - stat.Mean());
+    const arma::vec qRecon = stat.EigVec() * qProj + stat.Mean();
+    const double kernelValue1 = EvaluateKernel(qPoint, qRecon);
+
+    // Base case for each reference node.
+    const arma::vec& rPoint = referenceSet.col(referenceIndex);
+    const arma::vec rProj = stat.EigVec().t() * (rPoint - stat.Mean());
+    const arma::vec rRecon = stat.EigVec() * rProj + stat.Mean();
+
+    const double maxMetricError = metric.Evaluate(rRecon, rPoint);
```
rcurtin: There are two sources of error: the reference point reconstruction error, which you have here as `maxMetricError` and then the query point reconstruction error (i.e., `metric.Evaluate(qRecon, qPoint)`).  So when we compute `maxProjKernel`, we should use `kernel.Evaluate(metricProjValue - maxMetricError - maxQueryMetricError)` (also need to make sure that's nonnegative), and for `minProjKernel` we should use `kernel.Evaluate(metricProjValue + maxMetricError + maxQueryMetricError)`.

I suspect that all the extra overhead of computing these projections will actually not help save any runtime; but, if we can get this to work, we can then figure out the best way to incorporate it into the tree pruning rules, and that is where we can make it fast. :)

src/mlpack/methods/kde/kde_rules_impl.hpp
```diff
@@ -76,16 +78,58 @@ double KDERules<MetricType, KernelType, TreeType>::BaseCase(
...
+    densities(queryIndex) += kernel.Evaluate(distance);
+  }
 
   ++baseCases;
   lastQueryIndex = queryIndex;
   lastReferenceIndex = referenceIndex;
-  traversalInfo.LastBaseCase() = distance;
-  return distance;
+  traversalInfo.LastBaseCase() = 0; // distance;
+  return 0; // distance;
```
rcurtin: Hmm, are you sure about this change?  I thought other parts of the code depended on `LastBaseCase()`.

src/mlpack/methods/kde/kde_rules_impl.hpp
robertohueso: I've tried this but `(metricProjValue - maxMetricError - maxQueryMetricError) < 0` in every case (using `KDETest/GaussianSingleKDTreePCA`). This means the error bound is large so I tried retaining a higher amount of variance when performing PCA (e.g. 0.999999999) but the error bound is still too big, there's barely any difference with retaining 0.9 of the variance.

Are you sure we need to take into account query point reconstruction error? Note that I'm using `kernel(q, r) = kernel(q, qRecon) * kernel(qProj, rProj)`. I tried not taking into account query reconstruction error as well but almost same results :-1: .

And I don't think it's worth the overhead in the base case but I think it's the simplest place to code a first implementation and avoid issues that could come from working with trees, so that's why :)

src/mlpack/methods/kde/kde_rules_impl.hpp
robertohueso: This is just to try the PCA implementation. It shouldn't have any effect for KD-Trees (which is the one I'm using for tests right now) but this will go back to normal when it's fully implemented :)

src/mlpack/methods/kde/kde_rules_impl.hpp
rcurtin: Oh, sorry, you're right---query reconstruction error is indeed accounted for with `kernelValue1`.  Sorry that I overlooked that.

So, if I understood correctly, you are saying that `metricProjValue - maxMetricError` is less than 0 for every single point (and that does not consider query reconstruction error).  And I think this can be true, but I also think we can be clever enough to work around this.

Take a look at Figure 1 (right side) in the paper.  We have `q`, `r`, `q_proj`, and `r_proj`, and our goal is to approximate `d(q, r)` quickly by projecting both `q` and `r` down to a lower-dimensional subspace where `q_proj` and `r_proj` are these projections.  By precomputing these projections and by the fact that the lower-dimensional distance computation is faster, our hope is that the extra error induced by these projections is small enough that we see speedup while still getting a kernel density estimate that matches our bounds.  (I think we're both on the same page about this, but for some reason I wrote it anyway.  Maybe because it is getting late.)

Specifically, for the sake of computing `K(q, r)`, we need to know `d(q, r)`---or an approximation thereof.  Since we are using the Gaussian kernel with the Euclidean distance, and since the vector `q -> q_proj` (also known as `q_recon`) and `q_proj -> r_proj` are orthogonal, it is true that `d(q, r_proj)^2 = d(q, q_proj)^2 + d(q_proj, r_proj)^2`.  This equality (not bound) here is how Dongryeol is able to decompose the distance computation in Eq. (3).  In fact, the "approximately equals" can be replaced with "exactly equals" if the left hand side is `e^{- || q - r_proj ||^2 / (2h^2)}`.  (I didn't double check that but I'm pretty sure.  Feel free to check...)

That means that the only error in the distance computation will come from the reference point reconstruction error.  Here, we don't have the nice orthogonality that we could use before, and so the best we can do is use the triangle inequality to say

```
d(q, r) <= d(q, r_proj) + d(r, r_proj)
```

where note that `d(r, r_proj)` is also equal to the (non-squared) norm of `r_recon`.  Similarly, we can use the triangle inequality in the other direction to say

```
d(q, r) >= | d(q, r_proj) - d(r, r_proj) |.
```

The important thing to notice here is that these bounds are on `d(q, r)`... but what we are using in the code is actually a bound on `d(q_proj, r)`.  Specifically, the code that uses `metricProjValue - maxMetricError` is equivalent to the bound `d(q_proj, r) > d(q_proj, r_proj) - d(r_proj, r)`; then, we later add in the orthogonal distance `d(q, q_proj)` via the computed kernel value `kernelValue1`.

So the first thing I notice is that we've actually forgotten part of the triangle inequality.  While it is true that `max(metricProjValue - maxMetricError, 0)` *is* a valid bound on `d(q_proj, r)`, we can actually do a bit better, because by the triangle inequality we should be able to just use `| metricProjValue - maxMetricError |` (or, `| d(q_proj, r_proj) - d(r, r_proj) |`).

The next observation is that the derivation we are using actually does not take into account the `d(q, q_proj)` term.  If we do that, we get a slightly different set of bounds on the distance:

```
d(q, r) <= sqrt(d(q, q_proj)^2 + d(q_proj, r_proj)^2) + d(r, r_proj)
d(q, r) >= | sqrt(d(q, q_proj)^2 + d(q_proj, r_proj)^2) - d(r, r_proj) |
```

For the upper bound, this doesn't really make *too* much difference, but I think that it can for the lower bound.  However, it's getting too late for me to think that through fully, so I'll give it another shot tomorrow.  Still though, I think that there are two things you can try:

(1) use `abs()` instead of `max(..., 0)` for the lower bound and see how that does
(2) use the more complex bounding strategy and don't compute `kernelValue1` directly

It's also possible to test whether the bounds are valid just by seeing if the actual kernel value is between `minK` and `maxK`---that should expose if something is horribly wrong in our thinking. :)

robertohueso: I think your whole reasoning is correct :+1: 

Same issues when using (1), I will try (2) later tonight :)
Differences between the real value and the PCA calculated value are actually really big, e.g.
```
real vs pca
1.59032e-20 vs 9.4963e-14
1.1172e-19 vs 1.71036e-15
1.04255e-19 vs 9.75098e-14
4.10788e-16 vs 2.60875e-13.
2.66177e-19 vs 2.68426e-14
```
Also checked and real values are between `minK` and `maxK`, e.g.
```
1.59032e-20 -> 1.07165e-30 --- 1.96156e-10
1.1172e-19 -> 7.95175e-34 --- 5.08942e-10
1.04255e-19 -> 2.0276e-32 --- 9.9817e-11
4.10788e-16 -> 1.41317e-30 --- 8.89619e-11
2.66177e-19 -> 3.00812e-32 --- 2.65021e-10
```
So I guess the problem here is the range between `minK` and `maxK` is too large.

It might be possible that I misunderstood how PCA works, so an important thing to note here is that this only happens when some eigenvector is removed from the eigenvectors matrix (line 921 in `kde_impl.hpp`). If we retain 100% of variance then everything works fine.
Let's say my dataset points have 50 dimensions and PCA is computed in a leaf node of a KD-Tree, this means the eigenvector matrix will have dimension 50(rows)x(number of points in the node)(columns), so for a node that contains 8 points, dimension of the eigenvector matrix will be 50x8 (which means projections of points will have 8 dimensions). When we choose to retain 95% of variance this could mean that the last 2 columns are removed so the eigenvector matrix will be 50x6 and points projections have 6 dimensions. Is this right?

rcurtin: Oh, sorry, maybe I should have looked more closely at the PCA code.  I'm about to go to bed so I don't have a chance now (I can do it tomorrow) but what you described is transposed---if our dataset has 50 dimensions, the goal of using PCA is to keep all 8 points, but project them onto a lower subspace.  For instance, if the majority of the variance of the data is contained in only a few dimensions, then, perhaps, only 5 dimensions are required to capture 95% of the variance; thus, after projection, we end up with 8 points in 5 dimensions, instead of 50.

When we do PCA, one way is to compute the eigendecomposition of the covariance matrix (which will be 50x50).  It sounds based on what you wrote like you are doing eigendecomposition of the data points, which is a slightly different thing.  (Actually you can use SVD on the data points themselves, but depending on the size of the data matrix, it may be slower or quicker, and the math to get the principal components is just a bit different.)

I'll take a better look at the code in the morning and hopefully be able to provide some better input then.

robertohueso: Some updates: After implementing (2) it all seemed to work fine :tada: but then I tried to change dimension of the datasets as a test. It works fine for ~50 dimensions datasets but whenever we go to higher dimensionality (e.g. ~300 dimensions) I get a few estimations out of bound, and although they're usually few, this means there's still something wrong with the algorithm, so I'm trying to debug this.

src/mlpack/methods/kde/kde_impl.hpp
```diff
@@ -749,6 +771,179 @@ serialize(Archive& ar, const unsigned int version)
...
+      const size_t numPoints = node->NumPoints();
+      arma::mat pointsInTheNode(node->Dataset().n_rows, numPoints);
+      for (size_t i = 0; i < numPoints; ++i)
+        pointsInTheNode.col(i) = node->Dataset().col(node->Point(i));
+
+      // Amount of points for which the base has been calculated.
+      stat.PointsInTheBase() = numPoints;
+
+      // Calculate base mean.
+      stat.Mean() = arma::mean(pointsInTheNode, 1);
```
rcurtin: I think we also have to subtract `stat.Mean()` from `pointsInTheNode`, so that when we take the SVD of `pointsInTheNode`, it is on centered data.

src/mlpack/methods/kde/kde_impl.hpp
```diff
@@ -749,6 +771,179 @@ serialize(Archive& ar, const unsigned int version)
...
+
+      // Right singular values. Unused.
+      arma::mat v;
+
+      // Singular value decomposition.
+      if (pointsInTheNode.n_rows > pointsInTheNode.n_cols)
+        arma::svd_econ(stat.EigVec(), stat.EigVal(), v, pointsInTheNode, "l");
+      else
+        arma::svd(stat.EigVec(), stat.EigVal(), v, pointsInTheNode);
```
rcurtin: Ok---I see the algorithm you are using here is SVD (not the eigendecomposition) and as far as I can tell this does correctly recover the eigenvectors for the covariance matrix.

src/mlpack/methods/kde/kde_rules_impl.hpp
rcurtin: When you run on 300 dimensions (or more), are you doing any dimensionality reduction with PCA?  Also, how far out of bounds are the estimations?  (If you can print output or show me how to run it that could be helpful. :+1:)

src/mlpack/methods/kde/kde_rules_impl.hpp
```diff
@@ -76,16 +78,58 @@ double KDERules<MetricType, KernelType, TreeType>::BaseCase(
...
+    const KDEStat& stat = traversalInfo.LastReferenceNode()->Stat();
+
+    const arma::vec& qPoint = querySet.col(queryIndex);
+    const arma::vec qProj = stat.EigVec().t() * (qPoint - stat.Mean());
+    const arma::vec qRecon = stat.EigVec() * qProj + stat.Mean();
+
+    // Base case for each reference node.
+    const arma::vec& rPoint = referenceSet.col(referenceIndex);
+    const arma::vec rProj = stat.EigVec().t() * (rPoint - stat.Mean());
+    const arma::vec rRecon = stat.EigVec() * rProj + stat.Mean();
```
rcurtin: As far as I can tell, these computations are also correct assuming that `stat.EigVec()` holds one eigenvector in each column.  (I am not sure if that is true... I think it is, but maybe it is worth double-checking.)

rcurtin:
@robertohueso up to you what you want to do with this one---from our discussions, I think that it would be necessary for the subspace tree to work with MC sampling, and I think that some computational tricks may be needed to cache some results to make it fast.  I think it's still possible we could see some speedup with this though.  In any case, I just wanted to write my thoughts down and say I am happy to help if you want to keep working on it.  If not that's okay also. :)

ShikharJ:
@rcurtin Should we keep this closed?

rcurtin:
@ShikharJ up to Roberto; @robertohueso if you'd like to keep working on it I'm here and happy to help, but I don't have enough time to keep it going myself.  I'll leave it closed for now but we can reopen it if desired.  (And if not that's perfectly ok!) :+1:

PR #1969, opened 2019-08-05
Adding Bow and TfIdf Encoding Policy 

jeffin143:
This branch is based on PR1960


src/mlpack/tests/string_encoding_test.cpp
```diff
@@ -0,0 +1,578 @@
...
+  arma::mat output;
+  TfIdfEncoding<CharExtract::TokenType> encoder;
+
+  // Passing a empty string to encode characters
+  encoder.Encode(input, output, CharExtract());
+  arma::mat target = {
+    { 0.0440, 0.0440, 0.0440, 0, 0 },
+    { 0, 0, 0, 0, 0.0587 },
+    { 0.0185, 0.0556, 0, 0.0371, 0 }
+  };
```
jeffin143: The above test is correct, since I have manually, created it, but I am unsure of this output, I will manually calculate and update you

src/mlpack/core/data/string_encoding_policies/tf_idf_encoding_policy.hpp
```diff
@@ -65,6 +72,12 @@ class TfIdfEncodingPolicy
                          size_t dictionarySize)
   {
     output.resize(datasetSize, std::vector<OutputType> (dictionarySize,0));
+    std::cout<<"std::vector<char> v;\n";
+    for(auto it=idfdict.begin();it!=idfdict.end();it++)
+    {
+      std::cout<<it->first<<" "<<it->second<<"\n";
+    }
+    std::cout<<"\n";
```
jeffin143: Screenshot from 2019-08-05 22-42-38

Am I missing Something ?, The tfidf Values Are changing for same dataset, I don't know how !

src/mlpack/tests/string_encoding_test.cpp
lozhnikov: I think it's better to calculate the output with `scikit-learn` or something else.

src/mlpack/core/data/string_encoding_policies/tf_idf_encoding_policy.hpp
```diff
@@ -0,0 +1,174 @@
...
+  {
+    // Important since Mapping starts from 1 whereas allowed column value is 0.
+    // std::cout<<"divide "<<(tokenCount[row][value - 1] / row_size[row])<<"\n";
+    // std::cout<<"output size "<<output.n_rows<<" idfdict value "<<value-1<<" is "<<idfdict[value-1]<<"\n";
+    // std::cout<<"multipy "<<(output.n_rows / idfdict[value-1])<<" "<<std::log10(output.n_rows / idfdict[value-1])<<"\n";
+    output(row, value-1) = (tokenCount[row][value - 1] / row_size[row]) * 
+        std::log10(output.n_rows / idfdict[value-1]);
+    // std::cout<<"output at "<<row<<" "<<value-1<<" "<<output(row, value-1)<<"\n";
+  }
+
```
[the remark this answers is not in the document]
lozhnikov: Probably you didn't reset the number of tokens.

src/mlpack/tests/string_encoding_test.cpp
```diff
@@ -0,0 +1,628 @@
...
+  for(int i=0;i<expected.size();i++)
+  {
+    for(int j=0;j<expected[i].size();j++)
+    {
+      cout<<expected[i][j]<<" ";
+    }
+    std::cout<<std::endl;
+  }
+  BOOST_REQUIRE(output == expected);
+}
```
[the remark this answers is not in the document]
lozhnikov: The exact comparison doesn't work well with floating point types.
You should specify some `eps > 0` e.g. `eps = 1e-12` and verify elementwise that
```
std::fabs(output[i] - expected[i]) < eps

File: src/mlpack/core/data/string_encoding_policies/tf_idf_encoding_policy.hpp (hunk 2 above)
Comment (lozhnikov) [reply src/mlpack/core/data/string_encoding_policies/tf_idf_encoding_policy.hpp]: Did you run the `Encode()` function sequentially 2 times? Probably you didn't reset the token counters.

File: src/mlpack/core/data/string_encoding_policies/tf_idf_encoding_policy.hpp [hunk 5]
```diff
@@ -0,0 +1,174 @@
...
+      row_size.push_back(0);
+      tokenCount.push_back(std::unordered_map<size_t, double>());
+    }
+    tokenCount.back()[value-1]++;
+    if(tokenCount.back()[value-1]==1)
+      idfdict[value-1]++;
+    row_size.back()++;
+  }
+ private:
+  static std::vector<std::unordered_map<size_t, double>> tokenCount;
```
lozhnikov: The `TfIdf` policy shouldn't contain static variables.

src/mlpack/core/data/string_encoding_policies/tf_idf_encoding_policy.hpp
jeffin143: @lozhnikov  Yes, The issue was with not reseting the token counters, Since they were static data members !

src/mlpack/core/data/string_encoding_policies/tf_idf_encoding_policy.hpp
jeffin143: Yes, I removed it .

src/mlpack/tests/string_encoding_test.cpp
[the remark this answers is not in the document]
jeffin143: @lozhnikov , will `BOOST_REQUIRE_CLOSE(a,b,eps)` will work?

src/mlpack/tests/string_encoding_test.cpp
jeffin143: Yes, I calculated it and it came to be same. I have updated the values.

src/mlpack/tests/string_encoding_test.cpp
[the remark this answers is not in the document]
lozhnikov: Yes, sure.

src/mlpack/core/data/string_encoding_policies/tf_idf_encoding_policy.hpp
```diff
@@ -0,0 +1,162 @@
...
+                       size_t value)
+  {
+    if(row>=tokenCount.size())
+    {
+      row_size.push_back(0);
+      tokenCount.push_back(std::unordered_map<size_t, double>());
+    }
+    tokenCount.back()[value-1]++;
+    if(tokenCount.back()[value-1]==1)
+      idfdict[value-1]++;
```
lozhnikov: Looks like you can move `idfdict[value-1]++;` to the previous `if` body.

src/mlpack/tests/string_encoding_test.cpp
```diff
@@ -0,0 +1,640 @@
...
+    "mlpack is an intuitive, fast, and flexible C++ machine learning library "
+    "with bindings to other languages. ",
+    "It is meant to be a machine learning analog to LAPACK, and aims to "
+    "implement a wide array of machine learning methods and functions "
+    "as a \"swiss army knife\" for machine learning researchers.",
+    "In addition to its powerful C++ interface, mlpack also provides "
+    "command-line programs and Python bindings."
+};
+
+static vector<string> stringEncodingInputSmall = {
```
lozhnikov: I think it's okay for the debug purpose. But we should switch to `stringEncodingInput` as soon as the algorithms are done.

src/mlpack/tests/string_encoding_test.cpp
```diff
@@ -0,0 +1,640 @@
...
+  CheckDictionaries(encoder.Dictionary(), binaryEncoder.Dictionary());
+
+  xmlEncoder.Encode(stringEncodingInput, xmlOutput, tokenizer);
+  textEncoder.Encode(stringEncodingInput, textOutput, tokenizer);
+  binaryEncoder.Encode(stringEncodingInput, binaryOutput, tokenizer);
+
+  CheckMatrices(output, xmlOutput, textOutput, binaryOutput);
+}
+
+BOOST_AUTO_TEST_CASE(BowEncodingTest)
```
lozhnikov: I suggest to rename the test: `BagOfWordsEncodingTest`.

src/mlpack/tests/string_encoding_test.cpp
```diff
@@ -0,0 +1,640 @@
...
+    { 0, 0, 0, 0, 1, 1, 1, 0 },
+    { 0, 1, 1, 1, 0, 0, 0, 1 }
+  };
+  CheckMatrices(output, expected);
+}
+
+/**
+ * Test the one pass modification of the Bag of Words encoding algorithm.
+ */
+BOOST_AUTO_TEST_CASE(OnePassBowEncodingTest)
```
lozhnikov: `OnePassBagOfWordsEncodingTest`

src/mlpack/tests/string_encoding_test.cpp
```diff
@@ -0,0 +1,640 @@
...
+    { 0, 1, 1, 1, 0, 0, 0, 1 }
+  };
+
+  BOOST_REQUIRE(output == expected);
+}
+
+/**
+* Test Bag of Words encoding for characters using lamda function.
+*/
+BOOST_AUTO_TEST_CASE(BowEncodingIndividualCharactersTest)
```
lozhnikov: `BagOfWordsEncodingIndividualCharactersTest`

src/mlpack/tests/string_encoding_test.cpp
```diff
@@ -0,0 +1,640 @@
...
+  };
+
+  CheckMatrices(output, target);
+}
+
+/**
+ * Test the one pass modification of the Bag of Words encoding algorithm
+ * in case of individual character encoding.
+ */
+BOOST_AUTO_TEST_CASE(OnePassBowEncodingIndividualCharactersTest)
```
lozhnikov: The same issue.

src/mlpack/tests/string_encoding_test.cpp
```diff
@@ -0,0 +1,640 @@
...
+    keysCount[keyValue.second]++;
+    // Every token should be mapped only once
+    BOOST_REQUIRE_EQUAL(keysCount[keyValue.second], 1);
+  }
+  arma::mat expected = {
+    { 0.1193, 0.0440, 0.0440, 0.0440, 0, 0, 0, 0 },
+    { 0, 0, 0, 0, 0.1590, 0.1590, 0.1590, 0 },
+    { 0, 0.0440, 0.0440, 0.0440, 0, 0, 0, 0.1193 }
+  };
+  CheckMatrices(output, expected, 1e-01);
```
lozhnikov: It's not enough. The double floating point precision around 1 is about `1e-16`. So, you can take `eps = 1e-12`.

src/mlpack/tests/string_encoding_test.cpp
```diff
@@ -0,0 +1,640 @@
...
+  }
+
+  vector<vector<double>> expected = {
+    { 0.11928, 0.0440228, 0.0440228, 0.0440228, 0, 0, 0, 0 },
+    { 0, 0, 0, 0, 0.15904, 0.15904, 0.15904, 0 },
+    { 0, 0.044028, 0.0440228, 0.0440228, 0, 0, 0, 0.11928 }
+  };
+  for(size_t i=0;i<expected.size();i++)
+    for(size_t j=0;j<expected[i].size();j++)
+      BOOST_REQUIRE_CLOSE(expected[i][j], output[i][j], 1e-01);
```
lozhnikov: Again, `eps = 1e-12`.

src/mlpack/tests/string_encoding_test.cpp
```diff
@@ -0,0 +1,640 @@
...
+  TfIdfEncoding<CharExtract::TokenType> encoder;
+
+  // Passing a empty string to encode characters
+  encoder.Encode(input, output, CharExtract());
+  arma::mat target = {
+    { 0.0352, 0, 0.0704, 0, 0 },
+    { 0, 0, 0.0503, 0.0503, 0.0682 },
+    { 0.0587, 0, 0, 0.0587, 0 }
+  };
+  CheckMatrices(output, target, 1e-01);
```
lozhnikov: `1e-12`.

src/mlpack/tests/string_encoding_test.cpp
```diff
@@ -0,0 +1,640 @@
...
+  // Passing a empty string to encode characters
+  encoder.Encode(input, output, CharExtract());
+  vector<vector<double>> expected = {
+    { 0.0352, 0, 0.0704, 0, 0 },
+    { 0, 0, 0.0503, 0.0503, 0.0682 },
+    { 0.0587, 0, 0, 0.0587, 0 }
+  };
+  for(size_t i=0;i<expected.size();i++)
+    for(size_t j=0;j<expected[i].size();j++)
+      BOOST_REQUIRE_CLOSE(expected[i][j], output[i][j], 1e-01);
```
lozhnikov: `1e-12`

src/mlpack/core/data/string_encoding_policies/tf_idf_encoding_policy.hpp
jeffin143: That will run only once for a row, and hence i am not sure, that will give us correct output,Did i miss something ?

lozhnikov: The previous condition holds true once for each row too.

jeffin143: > That will run only once for a row, and hence i am not sure, that will give us correct output,Did i miss something ?

The above thing was for the previous if

`idfdict[value-1]++;`: I want it to run once every row for every unique token, and hence i need to compare it on the basis of token, And that is why i am checking whether the value is 1, since it should run only once every row

src/mlpack/tests/string_encoding_test.cpp
jeffin143: also isn't 1e-12 too strict ?
I have to type 12 digits after decimal

can't we make it
1e-10 ?

lozhnikov: I don't see any difference between `1e-10` and `1e-12`. There is no need to type it. You can write the result of `scikit-learn` to a file using the desired format.

src/mlpack/core/data/string_encoding_policies/tf_idf_encoding_policy.hpp
lozhnikov: I see. You're right.

src/mlpack/tests/string_encoding_test.cpp
jeffin143: Ok , I have used the following [article](https://towardsdatascience.com/tfidf-for-piece-of-text-in-python-43feccaa74f8) to write the implemetation

src/mlpack/tests/string_encoding_test.cpp
```diff
@@ -0,0 +1,640 @@
...
+    keysCount[keyValue.second]++;
+    // Every token should be mapped only once
+    BOOST_REQUIRE_EQUAL(keysCount[keyValue.second], 1);
+  }
+  arma::mat expected = {
+    { 0.1192803136799, 0.0440228147639, 0.0440228147639, 0.0440228147639, 0, 0, 0, 0 },
+    { 0, 0, 0, 0, 0.1590404182399, 0.1590404182399, 0.1590404182399, 0 },
+    { 0, 0.0440228147639 , 0.0440228147639 , 0.0440228147639 , 0, 0, 0, 0.1192803136799 }
+  };
+  CheckMatrices(output, expected, 1e-10);
```
lozhnikov: Why does your result differ from `scikit-learn`?

I tried the following test
```python
from sklearn.feature_extraction.text import TfidfVectorizer

vectorizer = TfidfVectorizer()

text = [
    "hello how are you",
    "i am good",
    "Good how are you"
]

X = vectorizer.fit_transform(text)

print(X)
```
The output was
```
  (0, 3)	0.6046521283053111
  (0, 4)	0.4598535287588349
  (0, 1)	0.4598535287588349
  (0, 5)	0.4598535287588349
  (1, 0)	0.7959605415681652
  (1, 2)	0.6053485081062916
  (2, 4)	0.5
  (2, 1)	0.5
  (2, 5)	0.5
  (2, 2)	0.5
```

src/mlpack/tests/string_encoding_test.cpp
```diff
@@ -0,0 +1,677 @@
...
+       0.014458219839989772, 0.014458219839989772, 0.014458219839989772,
+       0.014458219839989772, 0.014458219839989772, 0.014458219839989772, 0,
+       0, 0, 0, 0, 0, 0, 0, 0, 0, 0  },
+    {  0.011739417270378749, 0, 0, 0, 0, 0, 0, 0.011739417270378749, 0, 0, 0,
+       0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0,
+       0.031808083647977492, 0.031808083647977492, 0.031808083647977492,
+       0.031808083647977492, 0.031808083647977492, 0.031808083647977492,
+       0.031808083647977492, 0.031808083647977492, 0.031808083647977492,
+       0.031808083647977492, 0.031808083647977492  }
+  };
```
lozhnikov: Looks like these values are also incorrect.

Try the following example:
```python
from sklearn.feature_extraction.text import TfidfVectorizer
vectorizer = TfidfVectorizer(norm=None)
#vectorizer = TfidfVectorizer(norm='l1')
#vectorizer = TfidfVectorizer(norm='l2')

text = [
    "mlpack is an intuitive, fast, and flexible C++ machine learning           +++library "
    "with bindings to other languages. ",
    "It is meant to be a machine learning analog to LAPACK, and aims to "
    "implement a wide array of machine learning methods and functions "
    "as a \"swiss army knife\" for machine learning researchers.",
    "In addition to its powerful C++ interface, mlpack also provides "
    "command-line programs and Python bindings."
]
 
X = vectorizer.fit_transform(text)

print(X.toarray())
```

src/mlpack/tests/string_encoding_test.cpp
jeffin143: > Why does your result differ from `scikit-learn`?
> 
> I tried the following test
> 
> ```python
> from sklearn.feature_extraction.text import TfidfVectorizer
> 
> vectorizer = TfidfVectorizer()
> 
> text = [
>     "hello how are you",
>     "i am good",
>     "Good how are you"
> ]
> 
> X = vectorizer.fit_transform(text)
> 
> print(X)
> ```
> 
> The output was
> 
> ```
>   (0, 3)	0.6046521283053111
>   (0, 4)	0.4598535287588349
>   (0, 1)	0.4598535287588349
>   (0, 5)	0.4598535287588349
>   (1, 0)	0.7959605415681652
>   (1, 2)	0.6053485081062916
>   (2, 4)	0.5
>   (2, 1)	0.5
>   (2, 5)	0.5
>   (2, 2)	0.5
> ```

Ok This was very strange enough , I though Sklearn implemented the basic version of tfidf
but it is not so, They are using some other variant

https://en.wikipedia.org/wiki/Tf%E2%80%93idf#Term_frequency_2

`tfidf score = tf_value * idf_value`
idf_value doesn't have different variants

But tf_value has 4 types of variants, namely 

-     Boolean "frequencies": tf(t,d) = 1 if t occurs in d and 0 otherwise;
 

-     term frequency adjusted for document length : ft,d ÷ (number of words in d)

   

-     logarithmically scaled frequency: tf(t,d) = log (1 + ft,d);[6]

    

-     augmented frequency, to prevent a bias towards longer documents, e.g. raw frequency divided by the raw frequency of the most occurring term in the document:

I have implemented the second variant and thus there is difference in calculation 

Also I have noticed , there are different base of for idf values 
some implementation uses base 2, some used base e and some used base 10, but the Wikipedia page and standard implementation which most of them use is base 10

https://en.wikipedia.org/wiki/Tf–idf#Example_of_tf–idf

lozhnikov: I think we should implement the same algorithm as in `scikit-learn` by default since it's on of the most known packages. Probably it's a good idea to implement other techniques as well. It's possible since we can pass arguments to the policy constructor.

lozhnikov: Try the following formulas: [https://scikit-learn.org/stable/modules/feature_extraction.html#tfidf-term-weighting](https://scikit-learn.org/stable/modules/feature_extraction.html#tfidf-term-weighting).

jeffin143: I am not sure, we will be able to provide support for normalization since The output matrix is created index by index , thus I guess we cannot support normalisation

lozhnikov: It's possible. We can add a policy function and call it at the end of each line. But I am not sure that it's needed. Probably it's better to write a number of scaler classes than transform the encoder to a huge combine. Let's do without the normalization.

jeffin143: > Try the following formulas: https://scikit-learn.org/stable/modules/feature_extraction.html#tfidf-term-weighting.

Ok so I have added
Following variation 
  bool binary;
  bool sublinear_tf;
  bool raw_count;
  bool term_frequency;

and each of these can be coupled with   bool smooth_idf;
so we 8 different types 
Do i have to write test for all the 8 ? for both arma::mat and vector<vector<double>>
and also do i have to write it for both charextract and stringInput 

If we are taking all of these then we have 4*2*2*2 (24 test alone for tf-idf)

jeffin143:
I ll take care of style checks after we are done with the algorithms and also add documentation and fix-up other documentation error.

src/mlpack/tests/string_encoding_test.cpp
jeffin143: Done

src/mlpack/core/data/string_encoding_policies/bow_encoding_policy.hpp
```diff
@@ -0,0 +1,144 @@
...
+                        input dataset.
+  * @param dictionarySize The size of the dictionary (not used).
+  */
+  template<typename OutputType>
+  static void InitMatrix(std::vector<std::vector<OutputType> >& output,
+                         size_t datasetSize,
+                         size_t /*maxNumTokens*/,
+                         size_t dictionarySize)
+  {
+    output.resize(datasetSize, std::vector<OutputType> (dictionarySize,0));
```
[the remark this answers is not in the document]
jeffin143: Done

src/mlpack/core/data/string_encoding_policies/bow_encoding_policy.hpp
```diff
@@ -0,0 +1,144 @@
...
+ * @author Jeffin Sam
+ *
+ * Definition of the BagOfWordsEncodingPolicy class.
+ *
+ * mlpack is free software; you may redistribute it and/or modify it under the
+ * terms of the 3-clause BSD license.  You should have received a copy of the
+ * 3-clause BSD license along with mlpack.  If not, see
+ * http://www.opensource.org/licenses/BSD-3-Clause for more information.
+ */
+#ifndef MLPACK_CORE_DATA_BAG_OF_WORDS_ENCODING_POLICY_HPP
```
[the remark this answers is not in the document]
jeffin143: Changed it.

lozhnikov:
@jeffin143 
I suggest to rename it to `TfIdfBinarySmoothIdfTest`. Could you also use the camel case for the names?

src/mlpack/core/data/string_encoding_policies/tf_idf_encoding_policy.hpp
```diff
@@ -0,0 +1,196 @@
...
+      idf = std::log((output.n_rows + 1) / (1 + idfdict[value-1])) + 1;
+    else
+      idf = std::log(output.n_rows / idfdict[value - 1]) + 1;
+    if (sublinear_tf)
+      tf = tokenCount[row][value - 1] / row_size[row];
+    else if (term_frequency)
+      tf = std::log(tokenCount[row][value - 1]) + 1;
+    else if (binary)
+      tf = tokenCount[row][value - 1] > 0 ? 1 : 0;
+    else
```
lozhnikov: If one excludes the other we should use `enum class` instead of `bool`.

Probably we should use a template parameter instead since the comparison is performed inside the loop.

src/mlpack/core/data/string_encoding.hpp
```diff
@@ -0,0 +1,197 @@
...
+ * @tparam EncodingPolicyType Type of the encoding algorithm itself.
+ * @tparam DictionaryType Type of the dictionary.
+ */
+template<typename EncodingPolicyType,
+         typename DictionaryType>
+class StringEncoding
+{
+ public:
+  /**
+   * Pass the given arguments to the policy constructor and cunstruct
```
zoq: I think this should be "construct" instead of "cunstruct".

src/mlpack/core/data/string_encoding_dictionary.hpp
```diff
@@ -0,0 +1,334 @@
...
+{
+ public:
+  //! A convenient alias for the internal type of the map.
+  using MapType = std::unordered_map<Token, size_t>;
+
+  //! The type of the token that the dictionary stores.
+  using TokenType = Token;
+
+  /**
+   * The function returns true if the dictionary caontains the given token.
```
zoq: Should this be "contains" instead of "caontains"?

src/mlpack/core/data/string_encoding_dictionary.hpp
```diff
@@ -0,0 +1,334 @@
...
+  /**
+   * The function adds the given token to the dictionary and assigns a label
+   * to the token.
+   *
+   * @param token The given token.
+   */
+  template<typename T>
+  void AddToken(T&& token)
+  {
+    size_t size = mapping.size();
```
zoq: I think there is no need to create an alias here, if it's only used once?

src/mlpack/core/data/string_encoding_dictionary.hpp
```diff
@@ -0,0 +1,334 @@
...
+
+    return *this;
+  }
+
+  //! Standard move assignment operator.
+  StringEncodingDictionary& operator=(
+      StringEncodingDictionary&& other) = default;
+
+  /**
+   * The function returns true if the dictionary caontains the given token.
```
zoq: See comment above.

src/mlpack/core/data/string_encoding_policies/bag_of_words_encoding_policy.hpp
```diff
@@ -0,0 +1,144 @@
...
+  * @param output Output matrix to store the encoded results (sp_mat or mat).
+  * @param value The encoded token.
+  * @param row The row number at which the encoding is performed.
+  * @param col The row token number at which the encoding is performed.
+  */
+  template<typename MatType>
+  static void Encode(MatType& output, size_t value, size_t row, size_t /*col*/)
+  {
+    // Important since Mapping starts from 1 whereas allowed column value is 0.
+    output(row, value-1) = 1;
```
zoq: Pedantic style issue missing space between the ``-``.

src/mlpack/core/data/string_encoding_policies/bag_of_words_encoding_policy.hpp
```diff
@@ -0,0 +1,144 @@
...
+  * @param value The encoded token.
+  * @param row The row number at which the encoding is performed.
+  * @param col The row token number at which the encoding is performed.
+  */
+  template<typename OutputType>
+  static void Encode(std::vector<std::vector<OutputType> >& output, size_t value,
+                     size_t row, size_t /*col*/)
+  {
+    // Important since Mapping starts from 1 whereas allowed column value is 0.
+    output[row][value-1] = 1;
```
zoq: See comment above.

src/mlpack/core/data/string_encoding_policies/bag_of_words_encoding_policy.hpp
```diff
@@ -0,0 +1,144 @@
...
+  * the encoded token to the ouput.
+  * Overload function to accepted vector<vector<size_t>> as output type.
+  *
+  * @param output Output matrix to store the encoded results.
+  * @param value The encoded token.
+  * @param row The row number at which the encoding is performed.
+  * @param col The row token number at which the encoding is performed.
+  */
+  template<typename OutputType>
+  static void Encode(std::vector<std::vector<OutputType> >& output, size_t value,
```
zoq: Another pedantic style issue, would be great if we could put each parameter on it's extra line; as I read the code I was wondering where is value coming from.

src/mlpack/core/data/string_encoding_policies/bag_of_words_encoding_policy.hpp
```diff
@@ -0,0 +1,144 @@
...
+  }
+
+  /**
+  * Empty function, Important for tf-idf encoding policy
+  *
+  * @param row The row number at which the encoding is performed.
+  * @param numToken The count of token parsed till now.
+  * @param value The encoded token.
+  */
+  static void PreprocessToken(size_t /*row*/, size_t /*numTokens*/,
```
zoq: See comment above.

src/mlpack/core/data/string_encoding_policies/tf_idf_encoding_policy.hpp
```diff
@@ -0,0 +1,196 @@
...
+  * @param value The encoded token.
+  * @param row The row number at which the encoding is performed.
+  * @param col The row token number at which the encoding is performed.
+  */
+  template<typename MatType>
+  void Encode(MatType& output, size_t value, size_t row, size_t /*col*/)
+  {
+    // Important since Mapping starts from 1 whereas allowed column value is 0.
+    double idf, tf;
+    if(smooth_idf)
```
zoq: Please use camel casing for all names.

zoq:
Just took a first look over the code, really liked the first part.

src/mlpack/core/data/string_encoding.hpp
lozhnikov: It's already resolved in #1960.

src/mlpack/core/data/string_encoding_dictionary.hpp
lozhnikov: It's already resolved in #1960.

src/mlpack/core/data/string_encoding_dictionary.hpp
lozhnikov: The main reason is the following: I am not quite sure which part of the equality is performed first.
```c++
std::map<int, int> m;

/* m[0] adds (0, 0) to the dictionary and increases the size of the dictionary.
 * The result of the expression depends on the moment 
 * of the execution of m.size().
 */
m[0] = m.size();
```

src/mlpack/core/data/string_encoding_dictionary.hpp
lozhnikov: It's already resolved in #1960.

src/mlpack/core/data/string_encoding_policies/bag_of_words_encoding_policy.hpp
```diff
@@ -0,0 +1,149 @@
...
+#include <mlpack/core/data/string_encoding_policies/policy_traits.hpp>
+#include <mlpack/core/data/string_encoding.hpp>
+
+namespace mlpack {
+namespace data {
+/**
+ * Definition of the BagOfWordsEncodingPolicy class.
+ *
+ * BagOfWords is used as a helper class for StringEncoding.
+ * The encoder create a vector of all the unique token and then assigns
```
lozhnikov: Small typo: `creates`.

src/mlpack/core/data/string_encoding_policies/bag_of_words_encoding_policy.hpp
```diff
@@ -0,0 +1,149 @@
...
+  * @param numToken The count of token parsed till now.
+  * @param value The encoded token.
+  */
+  static void PreprocessToken(size_t /*row*/,
+                              size_t /*numTokens*/,
+                              size_t /*value*/) { }
+};
+
+/**
+ * The specialization provides some information about the dictionary encoding
```
lozhnikov: The comment is out of date: `about the bag of words encoding policy`.

src/mlpack/core/data/string_encoding_policies/tf_idf_encoding_policy.hpp
```diff
@@ -0,0 +1,227 @@
...
+  * subinerTf : logarthimic weighting scheme (log(tf) + 1) 
+  * 
+  */
+  enum tfTypes
+  {
+    rawCount,
+    binary,
+    sublinearTf,
+    termFrequency,
+  };
```
lozhnikov: It's not mentioned in the mlpack style guide. But we use the following style in the rest of the code:
```
enum TfTypes
{
  RAW_COUNT,
  BINARY,
  SUBLINEAR_TF,
  TERM_FREQUENCY,
};
```

lozhnikov: Probably it's better to use `enum class` instead of `enum` since `enum class` can't be initialized from an integer without `static_cast`.

src/mlpack/core/data/string_encoding_policies/tf_idf_encoding_policy.hpp
```diff
@@ -0,0 +1,227 @@
...
+
+  TfIdfEncodingPolicy(size_t tfType = 0, bool smooth_idf = true) :
+                      tfType(tfType), smooth_idf(smooth_idf)
```
lozhnikov: Could you break the line after the comma?

src/mlpack/core/data/string_encoding_policies/tf_idf_encoding_policy.hpp
```diff
@@ -0,0 +1,227 @@
...
+  * 
+  */
+  enum tfTypes
+  {
+    rawCount,
+    binary,
+    sublinearTf,
+    termFrequency,
+  };
+
```
lozhnikov: Could you add the description of the method?

src/mlpack/core/data/string_encoding_policies/tf_idf_encoding_policy.hpp
```diff
@@ -0,0 +1,227 @@
...
+              size_t row,
+              size_t /*col*/)
+  {
+    // Important since Mapping starts from 1 whereas allowed column value is 0.
+    double idf, tf;
+    if (smooth_idf)
+      idf = std::log((output.n_rows + 1) / (1 + idfdict[value - 1])) + 1;
+    else
+      idf = std::log(output.n_rows / idfdict[value - 1]) + 1;
+    if (tfType == tfTypes::termFrequency)
```
lozhnikov: I suggest to insert an empty line between the conditions.

src/mlpack/core/data/string_encoding_policies/tf_idf_encoding_policy.hpp
```diff
@@ -0,0 +1,227 @@
...
+      tf = tokenCount[row][value - 1] / row_size[row];
+    else if (tfType == tfTypes::sublinearTf)
+      tf = std::log(tokenCount[row][value - 1]) + 1;
+    else if (tfType == tfTypes::binary)
+      tf = tokenCount[row][value - 1] > 0 ? 1 : 0;
+    else
+      tf = tokenCount[row][value - 1];
+    output(row, value-1) =  tf * idf;
```
lozhnikov: And an empty line here.

src/mlpack/core/data/string_encoding_policies/tf_idf_encoding_policy.hpp
```diff
@@ -0,0 +1,227 @@
...
+      idf = std::log(output.size() / idfdict[value - 1]) + 1;
+    if (tfType == tfTypes::termFrequency)
+      tf = tokenCount[row][value - 1] / row_size[row];
+    else if (tfType == tfTypes::sublinearTf)
+      tf = std::log(tokenCount[row][value - 1]) + 1;
+    else if (tfType == tfTypes::binary)
+      tf = tokenCount[row][value - 1] > 0 ? 1 : 0;
+    else
+      tf = tokenCount[row][value - 1];
+    output[row][value-1] =  tf * idf;
```
lozhnikov: See the comments above.

src/mlpack/core/data/string_encoding_policies/tf_idf_encoding_policy.hpp
```diff
@@ -0,0 +1,227 @@
...
+  }
+
+  /**
+   * Serialize the class to the given archive.
+   */
+  template<typename Archive>
+  void serialize(Archive& /* ar */, const unsigned int /* version */)
+  {
+    // Nothing to serialize.
```
lozhnikov: This is not the case. Could you add the Tf-Idf serialization test as well.

src/mlpack/core/data/string_encoding_policies/tf_idf_encoding_policy.hpp
```diff
@@ -0,0 +1,227 @@
...
+  void PreprocessToken(size_t row,
+                       size_t /*numTokens*/,
+                       size_t value)
+  {
+    if (row >= tokenCount.size())
+    {
+      row_size.push_back(0);
+      tokenCount.push_back(std::unordered_map<size_t, double>());
+    }
+    tokenCount.back()[value-1]++;
```
lozhnikov: I suggest to add spaces around the operator.

src/mlpack/core/data/string_encoding_policies/tf_idf_encoding_policy.hpp
```diff
@@ -0,0 +1,227 @@
...
+  /**
+   * Indicates if the policy is able to encode the token at once without 
+   * any information about other tokens as well as the total tokens count.
+   */
+  static const bool onePassEncoding = false;
+};
+
+template<typename TokenType>
+using TfIdfEncoding = StringEncoding<TfIdfEncodingPolicy,
+                                          StringEncodingDictionary<TokenType>>;
```
lozhnikov: Could you correct the indent?

src/mlpack/tests/string_encoding_test.cpp
```diff
@@ -0,0 +1,1017 @@
...
+  CheckMatrices(output, xmlOutput, textOutput, binaryOutput);
+}
+
+BOOST_AUTO_TEST_CASE(BagOfWordsEncodingTest)
+{
+  using DictionaryType = StringEncodingDictionary<boost::string_view>;
+
+  arma::mat output;
+  BagOfWordsEncoding<SplitByAnyOf::TokenType> encoder;
+  SplitByAnyOf tokenizer(" ");
```
lozhnikov: The input has some punctuation as well as spaces.
```
SplitByAnyOf tokenizer(" ,.");
```

src/mlpack/tests/string_encoding_test.cpp
```diff
@@ -0,0 +1,1017 @@
...
+ * Test the one pass modification of the Bag of Words encoding algorithm.
+ */
+BOOST_AUTO_TEST_CASE(OnePassBagOfWordsEncodingTest)
+{
+  using DictionaryType = StringEncodingDictionary<boost::string_view>;
+
+  vector<vector<size_t>> output;
+  BagOfWordsEncoding<SplitByAnyOf::TokenType> encoder(
+      (BagOfWordsEncodingPolicy()));
+  SplitByAnyOf tokenizer(" ");
```
lozhnikov: See the comment above.

src/mlpack/tests/string_encoding_test.cpp
```diff
@@ -0,0 +1,1017 @@
...
+    {  1, 0, 0, 0, 0, 1, 0, 1, 0, 0, 0, 0, 0, 1, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0,
+       0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1  }
+  };
+  CheckMatrices(output, expected);
+}
+
+/**
+ * Test the one pass modification of the Bag of Words encoding algorithm.
+ */
+BOOST_AUTO_TEST_CASE(OnePassBagOfWordsEncodingTest)
```
lozhnikov: It's not the one pass modification. It's exactly the same algorithm, but the output is a vector.

src/mlpack/tests/string_encoding_test.cpp
```diff
@@ -0,0 +1,1017 @@
...
+  };
+
+  CheckMatrices(output, target);
+}
+
+/**
+ * Test the one pass modification of the Bag of Words encoding algorithm
+ * in case of individual character encoding.
+ */
+BOOST_AUTO_TEST_CASE(OnePassBagOfWordsEncodingIndividualCharactersTest)
```
lozhnikov: See the comment above.

src/mlpack/tests/string_encoding_test.cpp
```diff
@@ -0,0 +1,1017 @@
...
+ * Test the Tf-Idf Encoding using rawcount type and smoothidf as true,
+ * which is the deafult values used for algorithim.
+ */
+BOOST_AUTO_TEST_CASE(RawCountSmoothIdfEncodingTest)
+{
+  using DictionaryType = StringEncodingDictionary<boost::string_view>;
+
+  arma::mat output;
+  TfIdfEncoding<SplitByAnyOf::TokenType> encoder;
+  SplitByAnyOf tokenizer(" ");
```
lozhnikov: See the comment above.

src/mlpack/tests/string_encoding_test.cpp
```diff
@@ -0,0 +1,1017 @@
...
+       1.69314718055995, 1.69314718055995, 1.69314718055995  }
+  };
+  CheckMatrices(output, expected, 1e-12);
+}
+
+/**
+ * Test the one pass modification of the TfIdf encoding algorithm, using rawcount
+ * as type of tf and smoothIdf as true.
+ */
+BOOST_AUTO_TEST_CASE(OnePassRawCountSmoothIdfEncodingTest)
```
lozhnikov: It's not the one pass modification. It's exactly the same algorithm. But the output is a vector.

src/mlpack/tests/string_encoding_test.cpp
```diff
@@ -0,0 +1,1017 @@
...
+{
+  using DictionaryType = StringEncodingDictionary<boost::string_view>;
+
+  vector<vector<double>> output;
+  TfIdfEncoding<SplitByAnyOf::TokenType> encoder(
+      (TfIdfEncodingPolicy()));
+  SplitByAnyOf tokenizer(" ");
```
lozhnikov: See the comment above.

src/mlpack/tests/string_encoding_test.cpp
```diff
@@ -0,0 +1,1017 @@
...
+    { 1.2876820724517808, 1, 0, 1.2876820724517808, 0 }
+  };
+  CheckMatrices(output, target, 1e-12);
+}
+
+/**
+ * Test the one pass modification of the Tf-Idf encoding algorithm
+ * in case of individual character encoding using default values.
+ */
+BOOST_AUTO_TEST_CASE(OnePassRawCountSmoothIdfEncodingIndividualCharactersTest)
```
lozhnikov: It's not the one pass modification. It's exactly the same algorithm. But the output is a vector.

src/mlpack/tests/string_encoding_test.cpp
```diff
@@ -0,0 +1,1017 @@
...
+ * Test the Tf-Idf Encoding using rawcount type and smoothidf as false.
+ */
+BOOST_AUTO_TEST_CASE(TfIdfRawCountEncodingTest)
+{
+  using DictionaryType = StringEncodingDictionary<boost::string_view>;
+
+  arma::mat output;
+  TfIdfEncoding<SplitByAnyOf::TokenType> encoder(
+      (TfIdfEncodingPolicy(0, false)));
+  SplitByAnyOf tokenizer(" ");
```
lozhnikov: See the comment above.

src/mlpack/tests/string_encoding_test.cpp
```diff
@@ -0,0 +1,1017 @@
...
+       2.09861228866811, 2.09861228866811, 2.09861228866811  }
+  };
+  CheckMatrices(output, expected, 1e-12);
+}
+
+/**
+ * Test the one pass modification of the TfIdf encoding algorithm, with rawcount
+ * as type, but with smoothidf as false.
+ */
+BOOST_AUTO_TEST_CASE(OnePassTfIdfRawCountEncodingTest)
```
lozhnikov: It's not the one pass modification. It's exactly the same algorithm. But the output is a vector.

src/mlpack/tests/string_encoding_test.cpp
```diff
@@ -0,0 +1,1017 @@
...
+  };
+  CheckMatrices(output, target, 1e-12);
+}
+
+/**
+ * Test the one pass modification of the Tf Idf encoding algorithm
+ * in case of individual character encoding, using raw count as type,
+ * and smoothidf as false.
+ */
+BOOST_AUTO_TEST_CASE(OnePassRawcountEncodingIndividualCharactersTest)
```
lozhnikov: It's not the one pass modification. It's exactly the same algorithm. But the output is a vector.

src/mlpack/tests/string_encoding_test.cpp
```diff
@@ -0,0 +1,1017 @@
...
+    { 1.2876820724517808, 1, 0, 1.2876820724517808, 0 }
+  };
+  CheckMatrices(output, target, 1e-12);
+}
+
+/**
+ * Test TFIDF encoding for characters using lamda function, using binary
+ * weighting scheme for tf and smoothidf as true.
+ */
+BOOST_AUTO_TEST_CASE(OnePassBnarySmoothIdfEncodingIndividualCharactersTest)
```
lozhnikov: It's not the one pass modification. It's exactly the same algorithm. But the output is a vector.

src/mlpack/tests/string_encoding_test.cpp
```diff
@@ -0,0 +1,1017 @@
...
+       1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0  },
+    {  1, 0, 0, 0, 0, 1, 0, 1, 0, 0, 0, 0, 0, 1, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0,
+       0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1  }
+  };
+
+  BOOST_REQUIRE(output == expected);
+}
+
+/**
+* Test Bag of Words encoding for characters using lamda function.
```
lozhnikov: There are no lambda expressions.

src/mlpack/tests/string_encoding_test.cpp
```diff
@@ -0,0 +1,1017 @@
...
+       1.69314718055995, 1.69314718055995, 1.69314718055995, 1.69314718055995,
+       1.69314718055995, 1.69314718055995, 1.69314718055995  }
+  };
+  for (size_t i = 0; i < expected.size(); i++)
+    for (size_t j = 0; j < expected[i].size(); j++)
+      BOOST_REQUIRE_CLOSE(expected[i][j], output[i][j], 1e-12);
+}
+
+/**
+ * Test TFIDF encoding for characters using lamda function, using rawcount as tf
```
lozhnikov: There are no lambda expressions.

src/mlpack/tests/string_encoding_test.cpp
```diff
@@ -0,0 +1,1017 @@
...
+       2.09861228866811, 2.09861228866811, 2.09861228866811, 2.09861228866811,
+       2.09861228866811, 2.09861228866811, 2.09861228866811  }
+  };
+  for (size_t i = 0; i < expected.size(); i++)
+    for (size_t j = 0; j < expected[i].size(); j++)
+      BOOST_REQUIRE_CLOSE(expected[i][j], output[i][j], 1e-12);
+}
+
+/**
+ * Test TFIDF encoding for characters using lamda function, using rawcount as
```
lozhnikov: There are no lambda expressions.

src/mlpack/tests/string_encoding_test.cpp
```diff
@@ -0,0 +1,1017 @@
...
+    { 0, 2, 2.8109302162163288, 2.8109302162163288, 2.0986122886681100 },
+    { 1.4054651081081644, 1, 0, 1.4054651081081644, 0 }
+  };
+  for (size_t i = 0; i < expected.size(); i++)
+    for (size_t j = 0; j < expected[i].size(); j++)
+      BOOST_REQUIRE_CLOSE(expected[i][j], output[i][j], 1e-12);
+}
+
+/**
+ * Test TFIDF encoding for characters using lamda function, using binary
```
lozhnikov: There are no lambda expressions.

src/mlpack/tests/string_encoding_test.cpp
```diff
@@ -0,0 +1,1017 @@
...
+  arma::mat target = {
+    { 1.2876820724517808, 1, 1.2876820724517808, 0, 0 },
+    { 0, 1, 1.2876820724517808, 1.2876820724517808, 1.6931471805599454 },
+    { 1.2876820724517808, 1, 0, 1.2876820724517808, 0 }
+  };
+  CheckMatrices(output, target, 1e-12);
+}
+
+/**
+ * Test TFIDF encoding for characters using lamda function, using binary
```
lozhnikov: There are no lambda expressions.

src/mlpack/tests/string_encoding_test.cpp
```diff
@@ -0,0 +1,1017 @@
...
+    { 0, 1, 1.2876820724517808, 1.2876820724517808, 1.6931471805599454 },
+    { 1.2876820724517808, 1, 0, 1.2876820724517808, 0 }
+  };
+  for (size_t i = 0; i < expected.size(); i++)
+    for (size_t j = 0; j < expected[i].size(); j++)
+      BOOST_REQUIRE_CLOSE(expected[i][j], output[i][j], 1e-12);
+}
+
+/**
+ * Test TFIDF encoding for characters using lamda function, using binary
```
lozhnikov: There are no lambda expressions.

src/mlpack/tests/string_encoding_test.cpp
```diff
@@ -0,0 +1,1017 @@
...
+  arma::mat target = {
+    { 1.4054651081081644, 1, 1.4054651081081644, 0, 0 },
+    { 0, 1, 1.4054651081081644, 1.4054651081081644, 2.0986122886681100 },
+    { 1.4054651081081644, 1, 0, 1.4054651081081644, 0 }
+  };
+  CheckMatrices(output, target, 1e-12);
+}
+
+/**
+ * Test TFIDF encoding for characters using lamda function, using sublinear
```
lozhnikov: There are no lambda expressions.

src/mlpack/tests/string_encoding_test.cpp
```diff
@@ -0,0 +1,1017 @@
...
+    { 1.2876820724517808, 1.6931471805599454, 2.1802352704293200, 0, 0 },
+    { 0, 1.6931471805599454, 2.1802352704293200, 2.1802352704293200,
+      1.6931471805599454 },
+    { 1.2876820724517808, 1, 0, 1.2876820724517808, 0 }
+  };
+  CheckMatrices(output, target, 1e-12);
+}
+
+/**
+ * Test TFIDF encoding for characters using lamda function, using sublinear
```
lozhnikov: There are no lambda expressions.

src/mlpack/tests/string_encoding_test.cpp
```diff
@@ -0,0 +1,1017 @@
...
+    { 1.4054651081081644, 1.6931471805599454, 2.3796592851687173, 0, 0 },
+    { 0, 1.6931471805599454, 2.3796592851687173, 2.3796592851687173,
+      2.0986122886681100 },
+    { 1.4054651081081644, 1, 0, 1.4054651081081644, 0 }
+  };
+  CheckMatrices(output, target, 1e-12);
+}
+
+/**
+ * Test TFIDF encoding for characters using lamda function, using term
```
lozhnikov: There are no lambda expressions.

src/mlpack/tests/string_encoding_test.cpp
```diff
@@ -0,0 +1,1017 @@
...
+    { 0.2575364144903562, 0.4, 0.5150728289807124, 0, 0 },
+    { 0, 0.2857142857142857, 0.3679091635576516, 0.3679091635576516,
+      0.2418781686514208 },
+    { 0.4292273574839269, 0.3333333333333333, 0, 0.4292273574839269, 0 }
+  };
+  CheckMatrices(output, target, 1e-12);
+}
+
+/**
+ * Test TFIDF encoding for characters using lamda function, using Term
```
lozhnikov: There are no lambda expressions.

src/mlpack/tests/string_encoding_test.cpp
```diff
@@ -0,0 +1,1017 @@
...
+/**
+ * Test the one pass modification of the Bag of Words encoding algorithm.
+ */
+BOOST_AUTO_TEST_CASE(OnePassBagOfWordsEncodingTest)
+{
+  using DictionaryType = StringEncodingDictionary<boost::string_view>;
+
+  vector<vector<size_t>> output;
+  BagOfWordsEncoding<SplitByAnyOf::TokenType> encoder(
+      (BagOfWordsEncodingPolicy()));
```
lozhnikov: You don't need to call the policy constructor explicitly. I added this code once in order to test the constructor. You can use the parameter-pack constructor i.e. pass the policy arguments directly to the encoder (in this case there are no arguments).
```
BagOfWordsEncoding<SplitByAnyOf::TokenType> encoder;
```
There are the same issues below. I didn't point them out.

src/mlpack/tests/string_encoding_test.cpp
```diff
@@ -0,0 +1,1017 @@
...
+ * Test the one pass modification of the TfIdf encoding algorithm, with rawcount
+ * as type, but with smoothidf as false.
+ */
+BOOST_AUTO_TEST_CASE(OnePassTfIdfRawCountEncodingTest)
+{
+  using DictionaryType = StringEncodingDictionary<boost::string_view>;
+
+  vector<vector<double>> output;
+  TfIdfEncoding<SplitByAnyOf::TokenType> encoder(
+      (TfIdfEncodingPolicy(0, false)));
```
lozhnikov: You don't need to call the policy constructor explicitly. I added this code once in order to test the constructor. You can use the parameter-pack constructor i.e. pass the policy arguments directly to the encoder.
```
TfIdfEncoding<SplitByAnyOf::TokenType> encoder(0, false);
```
There are the same issues below. I didn't point them out.

However, I think it's reasonable to call the policy constructor once for the Tf-Idf policy in order to test the encoder constructor.

src/mlpack/tests/string_encoding_test.cpp
```diff
@@ -0,0 +1,1066 @@
...
+  CheckMatrices(output, xmlOutput, textOutput, binaryOutput);
+}
+
+BOOST_AUTO_TEST_CASE(BagOfWordsEncodingTest)
+{
+  using DictionaryType = StringEncodingDictionary<boost::string_view>;
+
+  arma::mat output;
+  BagOfWordsEncoding<SplitByAnyOf::TokenType> encoder;
+  SplitByAnyOf tokenizer(" ");
```
lozhnikov: The same comment.
```
SplitByAnyOf tokenizer(" ,.");

File: src/mlpack/tests/string_encoding_test.cpp [hunk 66]
```diff
@@ -0,0 +1,1066 @@
...
+    {  1, 0, 0, 0, 0, 1, 0, 1, 0, 0, 0, 0, 0, 1, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0,
+       0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1  }
+  };
+  CheckMatrices(output, expected);
+}
+
+/**
+* Test the one pass modification of the Bag of Words encoding algorithm.
+*/ 
+BOOST_AUTO_TEST_CASE(OnePassBagOfWordsEncodingTest)
```
lozhnikov: Again the same comment: It's not the one pass modification. It's exactly the same algorithm, but the output is a vector.

@jeffin143 Did you read my comments?

src/mlpack/tests/string_encoding_test.cpp
```diff
@@ -0,0 +1,1066 @@
...
+/**
+ * Serialization test for the TF-IDF encoding algorithm with
+ * the SplitByAnyOf tokenizer.
+ */
+BOOST_AUTO_TEST_CASE(SplitByAnyOfTfIdfEncodingSerialization)
+{
+  using EncoderType = TfIdfEncoding<SplitByAnyOf::TokenType>;
+
+  EncoderType encoder;
+  SplitByAnyOf tokenizer(" ");
```
lozhnikov: Again,
```
SplitByAnyOf tokenizer(" ,.");
```

lozhnikov:
What's wrong with the AppVeyor build?

src/mlpack/tests/string_encoding_test.cpp
jeffin143: Sorry @lozhnikov , I was Rectifying it chuck by Chunk, and that's why I didn't address all of them, I was debugging serialisation test morning and thus made some commits and then I was addressing your comments, I also mentioned that in the commit message as  `next to resolve comments` , I will address all of these in an hour or so and notify you.

jeffin143: Sorry @lozhnikov , I was Rectifying it chuck by Chunk, and that's why I didn't address all of them, I was debugging serialisation test morning and thus made some commits and then I was addressing your comments, I also mentioned that in the commit message as  `next to resolve comments` , I will address all of these in an hour or so and notify you.

src/mlpack/core/data/string_encoding_policies/tf_idf_encoding_policy.hpp
```diff
@@ -0,0 +1,233 @@
...
+    double idf, tf;
+    if (smoothIdf)
+      idf = std::log((output.n_rows + 1) / (1 + idfdict[value - 1])) + 1;
+    else
+      idf = std::log(output.n_rows / idfdict[value - 1]) + 1;
+
+    if (tfType == static_cast<size_t>(TfTypes::TERM_FREQUENCY))
+      tf = tokenCount[row][value - 1] / row_size[row];
+    else if (tfType == static_cast<size_t>(TfTypes::SUBLINEAR_TF))
+      tf = std::log(tokenCount[row][value - 1]) + 1;
```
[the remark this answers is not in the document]
jeffin143: https://github.com/scikit-learn/scikit-learn/blob/1495f69242646d239d89a5713982946b8ffcf9d9/sklearn/feature_extraction/text.py#L1309-L1310

is it, I am not sure

jeffin143:
> What's wrong with the AppVeyor build?

[00:21:04] c:\projects\mlpack\src\mlpack\methods\linear_svm\linear_svm_function_impl.hpp(295): warning C4244: 'argument': conversion from 'const arma::uword' to 'const double', possible loss of data [C:\projects\mlpack\build\src\mlpack\tests\mlpack_test.vcxproj]
[00:21:04] c:\projects\mlpack\src\mlpack\methods\linear_svm\linear_svm_function_impl.hpp(345): error C2668: 'arma::size': ambiguous call to overloaded function [C:\projects\mlpack\build\src\mlpack\tests\mlpack_test.vcxproj]
[00:21:04]   c:\projects\mlpack\armadillo-7.800.2\include\armadillo_bits/fn_size.hpp(38): note: could be 'const arma::SizeMat arma::size<arma::mat>(const T1 &)' [found using argument-dependent lookup]

Not sure is the issue with my PR or something else

src/mlpack/core/data/string_encoding_policies/tf_idf_encoding_policy.hpp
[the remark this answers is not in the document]
lozhnikov: You're right.

lozhnikov:
> Not sure is the issue with my PR or something else

I agree. Looks like it's not your fault.

src/mlpack/tests/string_encoding_test.cpp
```diff
@@ -0,0 +1,1111 @@
...
+  TfIdfEncoding<CharExtract::TokenType>
+      encoder(TfIdfEncodingPolicy::TfTypes::BINARY, false);
+
+  // Passing a empty string to encode charactersrawcountsmoothidftrue
+  encoder.Encode(input, output, CharExtract());
+  arma::mat target = {
+    { 1.4054651081081644, 1, 1.4054651081081644, 0, 0 },
+    { 0, 1, 1.4054651081081644, 1.4054651081081644, 2.0986122886681100 },
+    { 1.4054651081081644, 1, 0, 1.4054651081081644, 0 }
+  };
```
lozhnikov: I wonder how did you get these values? `scikit-learn` doesn't support the binary mode.

src/mlpack/tests/string_encoding_test.cpp
```diff
@@ -0,0 +1,1111 @@
...
+      encoder(TfIdfEncodingPolicy::TfTypes::TERM_FREQUENCY, true);
+
+  // Passing a empty string to encode charactersrawcountsmoothidftrue
+  encoder.Encode(input, output, CharExtract());
+  arma::mat target = {
+    { 0.2575364144903562, 0.4, 0.5150728289807124, 0, 0 },
+    { 0, 0.2857142857142857, 0.3679091635576516, 0.3679091635576516,
+      0.2418781686514208 },
+    { 0.4292273574839269, 0.3333333333333333, 0, 0.4292273574839269, 0 }
+  };
```
lozhnikov: See the comment above.

src/mlpack/tests/string_encoding_test.cpp
```diff
@@ -0,0 +1,1111 @@
...
+  TfIdfEncoding<CharExtract::TokenType>
+      encoder(TfIdfEncodingPolicy::TfTypes::TERM_FREQUENCY, false);
+
+  // Passing a empty string to encode charactersrawcountsmoothidftrue
+  encoder.Encode(input, output, CharExtract());
+  arma::mat target = {
+    { 0.2810930216216329, 0.4, 0.5621860432432658, 0, 0 },
+    { 0, 0.2857142857142857, 0.4015614594594755, 0.4015614594594755,
+      0.2998017555240157 },
+    { 0.4684883693693881, 0.3333333333333333, 0, 0.4684883693693881, 0 }
```
lozhnikov: See the comment above.

src/mlpack/tests/string_encoding_test.cpp
```diff
@@ -0,0 +1,1111 @@
...
+/**
+ * Serialization test for the TF-IDF encoding algorithm with
+ * the SplitByAnyOf tokenizer.
+ */
+BOOST_AUTO_TEST_CASE(SplitByAnyOfTfIdfEncodingSerialization)
+{
+  using EncoderType = TfIdfEncoding<SplitByAnyOf::TokenType>;
+
+  EncoderType encoder;
+  SplitByAnyOf tokenizer(" ");
```
lozhnikov: The same comment:
```c++
SplitByAnyOf tokenizer(" ,.\"");
```

src/mlpack/tests/string_encoding_test.cpp
```diff
@@ -0,0 +1,1111 @@
...
+  };
+
+  CheckMatrices(output, target);
+}
+
+/**
+ * Test the Bag of Words encoding algorithm in case of individual
+ * character encoding, storing resulting in vector<vector<vectorType>>.
+ */
+BOOST_AUTO_TEST_CASE(OnePassBagOfWordsEncodingIndividualCharactersTest)
```
lozhnikov: Again, the `BagOfWords` policy doesn't support the `one pass` modification.

src/mlpack/tests/string_encoding_test.cpp
```diff
@@ -0,0 +1,1111 @@
...
+    { 0.4684883693693881, 0.3333333333333333, 0, 0.4684883693693881, 0 }
+  };
+  CheckMatrices(output, target, 1e-12);
+}
+
+/**
+ * Serialization test for the TF-IDF encoding algorithm with
+ * the CharExtract tokenizer.
+ */
+BOOST_AUTO_TEST_CASE(CharExtractTfIdfEncodingSerialization)
```
lozhnikov: I think we don't need it since we have the `SplitByAnyOf` test.

src/mlpack/core/data/string_encoding_policies/tf_idf_encoding_policy.hpp
```diff
@@ -0,0 +1,228 @@
...
+ * retrieval, that has also found good use in document classification.
+ * The goal of using tf-idf is to scale down the impact of tokens that occur
+ * very frequently in a given corpus while using the raw frequencies of a token
+ * and that are hence empirically less informative than features that occur in
+ * a small fraction of the training corpus.
+ * TfIdfEncodingPolicy is used as a helper class for StringEncoding.
+ * The encoder assigns a tf-idf number to each unique token and treat 
+ * the dataset as categorical. The tokens are labeled in the order of their
+ * occurrence in the input dataset.
+ */
```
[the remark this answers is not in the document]
jeffin143: Yes sure, I thought it ok to use, Anyways I will remove it :)

src/mlpack/tests/string_encoding_test.cpp
jeffin143: I test it manually for a smaller input and then when got correct values I just escalated for bigger values and saved it in a csv file

jeffin143: Also there is an option to use binary mode , ` binary : boolean (default=False)
If True, all non-zero term counts are set to 1. This does not mean outputs will have only 0/1 values, only that the tf term in tf-idf is binary. (Set idf and normalization to False to get 0/1 outputs.) ` , refered from sklearn documentation.

src/mlpack/tests/string_encoding_test.cpp
jeffin143: I agree with you, sklearn doesn't support term frequency mode, so I had to calculate it manually, for small input , the previous year case, and then got them corrected and then wrote to a CSV file for a bigger input.

src/mlpack/tests/string_encoding_test.cpp
jeffin143: Yes, but you forgot to serialise the size , data member for overload dictionary_encoding class for int type and hence it was not working smoothly , I have corrected that and add the line , so may be this could come in handy in future, still if you want I will remove it.

src/mlpack/tests/string_encoding_test.cpp
lozhnikov: > Also there is an option to use binary mode

Ok, I'll try to reproduce.

src/mlpack/tests/string_encoding_test.cpp
lozhnikov: Yes, I see. Actually, I wrote the test (I mean `CharExtractDictionaryEncodingSerialization`), but I didn't check the size in `CheckDictionaries<int>()`.

I think the best way is to fix `CheckDictionarie4s<int>()`:
```c++
template<>
void CheckDictionaries(const StringEncodingDictionary<int>& expected,
                       const StringEncodingDictionary<int>& obtained)
{
  // MapType is equal to std::arry<size_t, 256>.
  using MapType = typename StringEncodingDictionary<int>::MapType;

  BOOST_REQUIRE_EQUAL(expected.Size(), obtained.Size());

  const MapType& expectedMapping = expected.Mapping();
  const MapType& mapping = obtained.Mapping();

  // And so on...
```

src/mlpack/tests/string_encoding_test.cpp
lozhnikov: Ok

lozhnikov:
I think the PR is almost ready. I am going to write benchmarks and fix the implementation (if needed). Probably I'll open a PR.

lozhnikov:
mlpack-bot, you are too fast:)

rcurtin:
If I can help out just let me know---this is a PR I am excited about. :+1:

lozhnikov:
@rcurtin I am going to try n-grams first since it may change the API.

src/mlpack/core/data/string_encoding.hpp
```diff
@@ -89,6 +89,22 @@ class StringEncoding
...
+   * Set minimum value for ngram.
+   */
+  void SetminNgram(int value);
+
+  //! Return the maxNgram
+  const int& MaxNgram() const { return maxNgram; }
+
+  //! Return the maxNgram
+  const int& MinNgram() const { return minNgram; }
+
```
jeffin143: Function to set range of ngrams, Since I didn't want to break the api

src/mlpack/core/data/string_encoding_impl.hpp
```diff
@@ -71,23 +75,26 @@ void StringEncoding<EncodingPolicyType, DictionaryType>::CreateMap(
...
+    boost::string_view strView(input);
+    auto token = tokenizer(strView, ngram);
 
-    token = tokenizer(strView);
+    static_assert(
+        std::is_same<typename std::remove_reference<decltype(token)>::type,
+                     typename std::remove_reference<typename DictionaryType::
+                        TokenType>::type>::value,
+        "The dictionary token type doesn't match the return value type "
+        "of the tokenizer.");
```
jeffin143: We should probably remove this check outside the loop, since it is costly to check multiple times, But I am not sure how

src/mlpack/core/data/tokenizers/char_extract.hpp
```diff
@@ -36,15 +36,20 @@ class CharExtract
...
-
+    int count = 0;
+    while(ngram--)
+    {
+      retval = retval * pow(10,count) + (static_cast<unsigned char>(str[count]));
+      count++;
+      if(count>=str.length())
+        break;
+    }
     str.remove_prefix(1);
```
jeffin143: We need to find a better formula to differentiate between `abc` and `cba` when ngram is 3, taking 3 at a time

src/mlpack/tests/string_encoding_test.cpp
```diff
@@ -1079,5 +1079,98 @@ BOOST_AUTO_TEST_CASE(SplitByAnyOfTfIdfEncodingSerialization)
...
+
+  // Checking that everything is mapped to different numbers
+  std::unordered_map<size_t, size_t> keysCount;
+  for (auto& keyValue : dictionary.Mapping())
+  {
+    keysCount[keyValue.second]++;
+    // Every token should be mapped only once
+    BOOST_REQUIRE_EQUAL(keysCount[keyValue.second], 1);
+  }
+
```
jeffin143: Test verified using sklearn ngram  for range (4,5)

sklearn output

Screenshot from 2019-11-08 15-19-38

Mlpack output

Screenshot from 2019-11-08 15-19-32

src/mlpack/tests/string_encoding_test.cpp
```diff
@@ -1079,5 +1079,98 @@ BOOST_AUTO_TEST_CASE(SplitByAnyOfTfIdfEncodingSerialization)
...
+  arma::mat expected = {
+    {  1.51082562376599070, 1.9162907318741551, 1.51082562376599070, 0, 0, 0,
+       0, 0, 0, 0, 0, 0 },
+    {  0, 0, 0, 1.9162907318741551, 1.9162907318741551, 1.9162907318741551, 
+       1.9162907318741551, 0, 0, 0, 0, 0 },
+    {  1.51082562376599070, 0, 0, 0, 0, 0, 0, 1.9162907318741551,
+       1.9162907318741551, 1.9162907318741551, 0, 0 },
+    {  0, 0, 1.51082562376599070, 0, 0, 0, 0, 0, 0, 0 , 1.9162907318741551,
+       1.9162907318741551 } 
+  };
```
jeffin143: This is the output if the range has same values (4,4)
Mlpack output

Screenshot from 2019-11-07 22-58-30

Python output

Screenshot from 2019-11-07 22-58-34

jeffin143:
@lozhnikov , also this one should I remove the ngram support and merge this one too
Since it has different policy  , that are completed

lozhnikov:
> @lozhnikov , also this one should I remove the ngram support and merge this one too
> Since it has different policy , that are completed

If I remember right I wanted to fix several style issues first. I'll open a PR on the basis of this one like I did with #1960.

jeffin143:
In support of #2251 , closing this

Issue #1928, opened 2019-06-17
mlpack 3.1.1 fails to compile with gcc 9.1.0 and USE_OPENMP=yes


janbraun:
#### Issue description

I cannot compile mlpack 3.1.1 with openmp support when using gcc 9.1.0.
Compiling without openmp support (USE_OPENMP=no) and gcc 9.1.0 works fine.

Compilation with openmp support and clang 8.0.0 also works.

#### Your environment

 * version of mlpack: 3.1.1
 * operating system: Gentoo Linux
 * compiler: gcc 9.1.0
 * version of dependencies (Boost/Armadillo): boost 1.65, armadillo 9.200.6
 * any other environment information you think is relevant:

#### Steps to reproduce

 1. Unpack sources
 2. Change to source dir, create build dir and change to that.
 3. Call cmake with -DUSE_OPENMP=yes ```cmake -G Ninja -DCMAKE_INSTALL_PREFIX=/usr -DBUILD_CLI_EXECUTABLES=yes -DDEBUG=no -DARMA_EXTRA_DEBUG=no -DBUILD_MARKDOWN_BINDINGS=no -DMATHJAX=yes -DUSE_OPENMP=yes -DPROFILE=no -DBUILD_PYTHON_BINDINGS=no -DBUILD_TESTS=no -DMATHJAX_JS_PATH=/usr/share/mathjax/MathJax.js ..```
 4. Call `ninja`

#### Expected behavior

Compilation completes

#### Actual behavior

Compilation fails for src/mlpack/methods/det/CMakeFiles/mlpack_det.dir/det_main.cpp.o

Call ```/usr/bin/c++  -DARMA_NO_DEBUG -DBOOST_TEST_DYN_LINK -DHAS_OPENMP -DNDEBUG -I../src -Isrc/mlpack/.. -fdiagnostics-color -fdiagnostics-color -Wall -Wextra -ftemplate-depth=1000 -O3 -fopenmp   -DBINDING_TYPE=BINDING_TYPE_CLI -std=gnu++11 -MD -MT src/mlpack/methods/det/CMakeFiles/mlpack_det.dir/det_main.cpp.o -MF src/mlpack/methods/det/CMakeFiles/mlpack_det.dir/det_main.cpp.o.d -o src/mlpack/methods/det/CMakeFiles/mlpack_det.dir/det_main.cpp.o -c ../src/mlpack/methods/det/det_main.cpp```

Output
 ```
[...]

PR #1970, opened 2019-08-06
[Ready] gcc-9 build fixes

lozhnikov:
gcc 9.1.1 throws errors like
```
src/mlpack/methods/det/dt_utils_impl.hpp:189:3: error: ‘folds’ not specified in enclosing ‘parallel’
src/mlpack/methods/det/dt_utils_impl.hpp:192:33: error: ‘testSize’ not specified in enclosing ‘parallel’
src/mlpack/methods/det/dt_utils_impl.hpp:194:61: error: ‘cvData’ not specified in enclosing ‘parallel’
```

These variables are used inside `omp parallel for default(none)` clause and they are not marked as `shared`. Looks like the previous versions of gcc didn't throw any errors since the variables are declared as `const`.

### DIFF
--- a/src/mlpack/methods/det/dt_utils_impl.hpp
+++ b/src/mlpack/methods/det/dt_utils_impl.hpp
@@ -184,8 +184,7 @@ DTree<MatType, TagType>* Trainer(MatType& dataset,
   // intmax_t because size_t is not yet supported by their OpenMP
   // implementation. omp_size_t is the appropriate type according to the
   // platform.
-  #pragma omp parallel for default(none) \
-      shared(prunedSequence, regularizationConstants)
+  #pragma omp parallel for shared(prunedSequence, regularizationConstants)
   for (omp_size_t fold = 0; fold < (omp_size_t) folds; fold++)
   {
     // Break up data into train and test sets.


lozhnikov:
Looks like it's quite an issue. The static analyzer tells:
```
src/mlpack/methods/det/dt_utils_impl.hpp:189:61: error: ‘folds’ is predetermined ‘shared’ for ‘shared’
```

rcurtin:
Awesome, thanks for looking into this.  I think it is the same thing as reported in #1928.  I hadn't managed to reproduce it (don't have gcc 9 installed, though I guess it would be easy to set up...).  What happens if we just drop the `default(none)`?

If that doesn't work, maybe (unfortunately) some `#if`s to check the OpenMP version (or compiler version?) might be needed.

lozhnikov:
Dropping `default(none)` works fine. However, I think checking the OpenMP version is a more delicate way than dropping `default(none)'.

lozhnikov:
It's not a bug. The reason is described in detail here [https://www.gnu.org/software/gcc/gcc-9/porting_to.html](https://www.gnu.org/software/gcc/gcc-9/porting_to.html).

The OpenMP 3.1 specifications ([https://www.openmp.org/specifications/](https://www.openmp.org/specifications/)) state the following rule:

**Data-sharing Attribute Rules for Variables Referenced in a Construct**
- Variables with const-qualified type having no mutable member are shared.

The OpenMP 4.0 (and later) specifications don't have this rule.

lozhnikov:
How about the following fix:
```c++
  #if _OPENMP >= 201307
    #pragma omp parallel for default(none) \
        shared(prunedSequence, regularizationConstants, folds, testSize, \
               cvData, minLeafSize, maxLeafSize, useVolumeReg)
  #else
    #pragma omp parallel for default(none) \
        shared(prunedSequence, regularizationConstants)
  #endif
```
?

lozhnikov:
No, looks like it doesn't work since gcc 7.3.0 provides the same OpenMP version (namely, 4.5) as gcc 9.1 (see [http://ci.mlpack.org/job/pull-requests-mlpack-static-code-analysis/2430/console](http://ci.mlpack.org/job/pull-requests-mlpack-static-code-analysis/2430/console)). Hence the previous versions of gcc won't work with the fix.

Looks like the previous versions of gcc didn't fully support it.

In this case I think it's better to remove the `default(none)` instead of write various compiler workarounds.

rcurtin:
Agreed, if we can get it working without `default(none)` I think that is better than various compiler workarounds.  The `_OPENMP` idea was really nice, so it's quite unfortunate that it's a compiler behavior change. :(

lozhnikov:
The PR is ready.

rcurtin:
Thanks again; I made some changes to HISTORY.md to point out this fix in 565617b7.

Issue #1863, opened 2019-04-15
mlpack_test.exe fails for some tests when running on Windows [3.1.1 release]


gmanlan:
When building mlpack x64 from source, some tests fail when executed in a local Windows 10 system. These errors are diverse so if required, we can split them into different submissions.

**Build configuration:**
- mlpack 3.1.1 release or master branch
- Boost 1_65_1 (lib64-msvc-14.1)
- armadillo-9.300.2
- OpenBLAS built using msvc - branch develop (April 2019)
- Visual Studio 15 2017 Win64
- x64
 
**Tests Errors:**
1) unknown location(0): fatal error: in "ConvolutionalNetworkTest/VanillaNetworkTest": memory access violation
**Possible workaround:** manually unzip the .arm dataset from mnist_first250_training_4s_and_9s.tar, then the test will pass

2) unknown location(0): fatal error: in "LoadSaveTest/CaseTest": class std::runtime_error: Parse error at line 13 token 0: "TA".
Context: this is trying to parse the file casecheck.arff

3) parallel_sgd_test.cpp(100): fatal error: in "ParallelSGDTest/GeneralizedRosenbrockTest": absolute value of result{-nan(ind)} exceeds 1e-08

4) unknown location(0): fatal error: in "SdpPrimalDualTest/SmallLovaszThetaSdp": memory access violation occurred at address 0xfcfc4000, while attempting to  write to an inaccessible (or protected) address. sdp_primal_dual_test.cpp(278): last checkpoint: "SmallLovaszThetaSdp" test entry

5) input stream error tests/serialization.hpp(235): fatal error: in "SerializationTest/LayerNormSerializationTest": critical check success == true has failed [false != true]

6) main_tests/decision_tree_test.cpp(103): fatal error: in "DecisionTreeMainTest/DecisionTreeCategoricalOutputDimensionTest": Cannot load train dataset braziltourism.arff!

I made an effort to go over open/closed issues and although I found some references, I couldn't find exact matches with these errors. If somebody has seen some of this issues before, please link them here.

I have also checked one of the last AppVeyor builds and the tests seem to be passing fine, so I'm trying to understand if these failures are because:
- Something is wrong with my local system?
- Boost version? (AppVeyor build script is using 1.60)
- Armadillo version? (AppVeyor build script is using 7.800.2)
- Visual Studio version? (for some reason AppVeyor build script is using Visual Studio 14 2015 Win64 for both the Jobs? If so, the VS2017 job is useless?)

[...]

PR #1971, opened 2019-08-07
Use correct Visual Studio version in AppVeyor

rcurtin:
@gmanlan pointed out in #1863 that the AppVeyor build configuration doesn't actually use Visual Studio 2017 like it's supposed to.  This is an attempt to fix that.  (And if it works, I'll also add Visual Studio 2019 to the build matrix.)


rcurtin:
I haven't figured out how to use PR drafts yet, but in any case, I think this one is ready for review. :)

PR #1973, opened 2019-08-11
Resolve appVeyor issues

Yashwants19:
While working on go bindings appVeyor fails. In this pr I am trying to resolve the issue.

### DIFF
--- a/src/mlpack/methods/linear_svm/linear_svm_function_impl.hpp
+++ b/src/mlpack/methods/linear_svm/linear_svm_function_impl.hpp
@@ -285,7 +285,7 @@ void LinearSVMFunction<MatType>::Gradient(
   }
   else
   {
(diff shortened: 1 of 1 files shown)


rcurtin:
Ah, synchronicity, I just had the same issue in #1971. :+1:  Looks like an ADL failure on the part of the Visual Studio compiler...

Yashwants19:
Oh This seems to duplicate of #1971 should I closed this PR. By the way I got the same issue when I was working on cli-bindings for LinearSVM.

zoq:
This also fixes potential issues in the gradient step, so I think it is reasonable to merge this one and merge #1971 afterwards.

rcurtin:
Agreed, #1971 doesn't actually have all the fixes that you made here, so let's merge this one.  Thanks! :+1:

PR #1975, opened 2019-08-11
Add STB image directory to include directories.

rcurtin:
This PR fixes a report by @jeffin143 in IRC that mlpack doesn't build correctly after downloading STB.  I reproduced the issue locally and we found that the STB header directory just needs to be added to the list of mlpack include directories.

I also updated the build documentation to describe the `DOWNLOAD_STB_IMAGE` CMake variable and updated the list of dependencies (to include ensmallen also, as well as STB).

@MuLx10 take a look and let me know what you think, or if I overlooked something.


CMakeLists.txt
```diff
@@ -344,6 +344,7 @@ if (NOT STB_IMAGE_FOUND)
         install(FILES ${CMAKE_BINARY_DIR}/deps/${STB_DIR}/stb_image.h DESTINATION ${CMAKE_INSTALL_INCLUDEDIR})
         install(FILES ${CMAKE_BINARY_DIR}/deps/${STB_DIR}/stb_image_write.h DESTINATION ${CMAKE_INSTALL_INCLUDEDIR})
         add_definitions(-DHAS_STB)
+        set(MLPACK_INCLUDE_DIRS ${MLPACK_INCLUDE_DIRS} ${CMAKE_BINARY_DIR}/deps/${STB_DIR}/)
```
zoq: I think this is a duplication of line 339.

rcurtin: You're right, thanks. :+1:

PR #1976, opened 2019-08-12
Reduce style related warnings

zoq:
Reduce style related warnings in the build process; example: https://travis-ci.org/mlpack/mlpack/jobs/570648288

### DIFF
--- a/src/mlpack/core/data/scaler_methods/pca_whitening.hpp
+++ b/src/mlpack/core/data/scaler_methods/pca_whitening.hpp
@@ -88,7 +88,7 @@ class PCAWhitening
     if (eigenValues.is_empty() || eigenVectors.is_empty())
     {
       throw std::runtime_error("Call Fit() before Transform(), please"
-        " refer documentation.");
+          " refer documentation.");
     }
     output.copy_size(input);
     output = (input.each_col() - itemMean);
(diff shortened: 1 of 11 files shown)


src/mlpack/core/data/scaler_methods/pca_whitening.hpp
```diff
@@ -88,7 +88,7 @@ class PCAWhitening
     if (eigenValues.is_empty() || eigenVectors.is_empty())
     {
       throw std::runtime_error("Call Fit() before Transform(), please"
-        " refer documentation.");
+          " refer documentation.");
```
rcurtin: I guess this should say `refer to the` if you want to fix that also. :)

src/mlpack/tests/augmented_rnns_tasks_test.cpp
```diff
@@ -95,7 +95,8 @@ class HardCodedSortModel
   void Train(arma::field<arma::mat>& predictors,
              arma::field<arma::mat>& labels)
   {
-    assert(predictors.n_elem == labels.n_elem);
+    const bool check = predictors.n_elem == labels.n_elem;
+    assert(check == true);
```
rcurtin: Interesting, I guess this one is a little tricker---I wonder if just using `Log::Assert()` could fix the unused variable warning.

src/mlpack/tests/augmented_rnns_tasks_test.cpp
```diff
@@ -95,7 +95,7 @@ class HardCodedSortModel
   void Train(arma::field<arma::mat>& predictors,
              arma::field<arma::mat>& labels)
   {
-    assert(predictors.n_elem == labels.n_elem);
+    Log::Assert(predictors.n_elem == labels.n_elem);
```
ShikharJ: I guess `Log` isn't declared here.

rcurtin:
The build failure was just an AppVeyor timeout, so I saw no issue with merging. :)

PR #1977, opened 2019-08-14
Fix Markdown binding issue

rcurtin:
While skimming through various emails I noticed that the nightly job that builds the mlpack-git Markdown documentation was failing.  That means that https://mlpack.org/doc/mlpack-git/cli_documentation.html will 404 until we get this fix incorporated (I suppose that I could rebuild after hand-patching, but I'd prefer to just merge this and then the problem goes away on its own).

But, this Markdown binding generation failure didn't actually trigger any build failure or anything like this.  So I did some digging and came up with these changes:

1) Make any programs executed by CMake cause a fatal error if those programs themselves fail.

2) Remove use of `PRINT_PARAM_STRING()` macro inside of documentation for an individual program option.  The reason for this is that `PRINT_PARAM_STRING()` was only meant to be used as part of the long-form program documentation, which actually takes a `std::function()` not a string.  This interacts badly with the Markdown bindings, because any use of `PRINT_PARAM_STRING()` requires a language to be set, and it isn't at global construction time when the global object corresponding to an individual program option is constructed.

Admittedly, that restriction of `PRINT_PARAM_STRING()` isn't documented anywhere...

@robertohueso let me know what you think---I worked around the issue by just referring to the logical name of the option instead of the actual parameter name.  I guess I didn't realize it would be a problem during review. :)

### DIFF
--- a/CMake/RunProgram.cmake
+++ b/CMake/RunProgram.cmake
@@ -5,4 +5,9 @@
 #
 #   PROGRAM: the program to run to.
 #   OUTPUT_FILE: the file to store the output in.
-execute_process(COMMAND ${PROGRAM} OUTPUT_FILE ${OUTPUT_FILE})
+execute_process(COMMAND ${PROGRAM} OUTPUT_FILE ${OUTPUT_FILE}
+    ERROR_VARIABLE err)
+
+if (err)
+  message(FATAL_ERROR "Fatal error running ${PROGRAM}: ${err}!")
+endif ()
--- a/src/mlpack/core/util/mlpack_main.hpp
+++ b/src/mlpack/core/util/mlpack_main.hpp
@@ -36,11 +36,43 @@
 #include <mlpack/bindings/cli/cli_option.hpp>
 #include <mlpack/bindings/cli/print_doc_functions.hpp>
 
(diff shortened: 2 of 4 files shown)


rcurtin:
> Perhaps a line to inform about this in the docs of PRINT_PARAM_STRING() would be nice to have :)

Yeah, you're right, I added documentation in all the relevant places.  I should have done that instead of mentioning in the PR description that I didn't. :)

zoq:
Hm, looks like the mlpack bot missed this one?

rcurtin:
Yeah, I am debugging it now.  mlpack-bot currently uses the `author_association` field with a review to determine whether or not it counts towards the 2 approval limit, but this is actually not correct, because anyone on any mlpack team has the same author association.  So I now have to go get the list of team members for the "Contributors" team and verify against that... hope to have it sorted out by later today.

zoq:
Okay, sounds like a lot of fun, do you like to keep this open for testing or should we merge it now?

rcurtin:
Let's leave it open for testing, so I can know when I have the issue solved. :)

rcurtin:
Victory! :)

PR #1978, opened 2019-08-14
Work Around to add more layers.

walragatver:
Looks like this is complete. Just let me know what you think.

### DIFF
--- a/src/mlpack/methods/ann/layer/layer_types.hpp
+++ b/src/mlpack/methods/ann/layer/layer_types.hpp
@@ -163,6 +163,10 @@ template<typename InputDataType,
 >
 class MultiplyMerge;
 
+using MoreTypes = boost::variant<
+        VirtualBatchNorm<arma::mat, arma::mat>*
+>;
+
 template <typename... CustomLayers>
 using LayerTypes = boost::variant<
     Add<arma::mat, arma::mat>*,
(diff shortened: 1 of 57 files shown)


zoq:
Wow, that was fast.

src/mlpack/methods/ann/layer/layer_types.hpp
```diff
@@ -225,7 +229,7 @@ using LayerTypes = boost::variant<
     Sequential<arma::mat, arma::mat, true>*,
     Subview<arma::mat, arma::mat>*,
     // VRClassReward<arma::mat, arma::mat>*,
-    VirtualBatchNorm<arma::mat, arma::mat>*,
+    MoreTypes,
```
zoq: Do you think we should move more layers to ``MoreTypes`` as this might open up the possibility to add more custom layers.

walragatver: Yes you are quite correct. We can move them. Just let me know how many we should move. If we move all of them then it will be 50 * 50 = 2500 layers.

zoq: I guess for now we can move 10 or so into ``MoreTypes``.

walragatver: Done.

zoq: Great, thanks, let's wait for the build jobs.

ShikharJ:
The builds are failing because no entry to the recently merged `BiasSetVisitor` has been made.

walragatver:
Hi @ShikharJ, Thanks for looking into the error. I have made the changes.

ShikharJ:
Thanks @walragatver for the contribution!

PR #1979, opened 2019-08-15
Fix intermittently failing Epanechnikov kernel KDE test

rcurtin:
Frequently, the test `KDETest/EpanechnikovCoverSingleKDETest` fails:

http://ci.mlpack.org/job/docker%20mlpack%20nightly%20build/armadillo_version=armadillo-6.500.5,boost_version=boost_1_49_0,buildmode=debug,compiler_version=gcc-7.2.0,label=docker/lastCompletedBuild/testReport/

```
[Exception] - difference{37.9501%} between bfEstimations[i]{0.6164662373649441} and treeEstimations[i]{0.99350104568814213} exceeds 8%
 == [File] - /home/jenkins/workspace/docker mlpack nightly build/armadillo_version/armadillo-6.500.5/boost_version/boost_1_49_0/buildmode/debug/compiler_version/gcc-7.2.0/label/docker/repo/src/mlpack/tests/kde_test.cpp
 == [Line] -233
```

I spent some time digging into this and this was kind of a funny bug to dig into.  I first determined that some of the prunes were invalid---it was estimated that the maximum kernel and minimum kernel for a query point and a reference node were both 0... but the actual kernel values were quite large!  So, this was confusing, until I noticed that the minimum distance between the query point and reference node was being computed as a negative value!

At this moment it all fell into place: this bug *only* occurs with a few types of kernels, most specifically *not* the Gaussian kernel, because if you give the Gaussian kernel a negative distance, it computes a very large value.  So it is only kernels that are symmetric about 0 that have this problem.  Further, the bug was only in the cover tree computation, so that narrows down the bug further.

Inside the cover tree, the minimum distance between a point and a node was being computed as

```
d(q, r) - furthestDescendantDistance
```

where `r` is the point at the center of the reference node.  However, we need to make sure this is not negative!  So the fix is quite easy.

We can thank me circa 2013 for this bug: https://github.com/mlpack/mlpack/commit/8a58c9fa98c3702625dea665be928539d1443500
:)

I also fixed this situation where it occurs a couple of times in the KDE rules, and everything seems to work well now.  I ran the test with the cover tree and Epanechnikov kernel in both single-tree and dual-tree mode 1000+ times and saw no failures, so I think the issue is resolved.

@robertohueso let me know what you think or if I overlooked anything. :)

### DIFF
--- a/src/mlpack/methods/kde/kde_rules_impl.hpp
+++ b/src/mlpack/methods/kde/kde_rules_impl.hpp
@@ -115,7 +115,8 @@ Score(const size_t queryIndex, TreeType& referenceNode)
     // Don't duplicate calculations.
     alreadyDidRefPoint0 = true;
     const double furthestDescDist = referenceNode.FurthestDescendantDistance();
-    minDistance = traversalInfo.LastBaseCase() - furthestDescDist;
+    minDistance = std::max(traversalInfo.LastBaseCase() - furthestDescDist,
+        0.0);
     maxDistance = traversalInfo.LastBaseCase() + furthestDescDist;
   }
   else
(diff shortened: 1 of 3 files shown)


src/mlpack/methods/kde/kde_rules_impl.hpp
```diff
@@ -115,7 +115,8 @@ Score(const size_t queryIndex, TreeType& referenceNode)
     // Don't duplicate calculations.
     alreadyDidRefPoint0 = true;
     const double furthestDescDist = referenceNode.FurthestDescendantDistance();
-    minDistance = traversalInfo.LastBaseCase() - furthestDescDist;
+    minDistance = std::max(traversalInfo.LastBaseCase() - furthestDescDist,
+        0.0);
```
robertohueso: Super minor style question here, should this be lined up with the previous argument `traversalInfo.LastBaseCase() - furthestDescDist` or does this just apply to method declarations?

Issue #1954, opened 2019-07-21
Reclaim not used absolute error in KDE


robertohueso:
This issue came out of the review of #1934.

The current KDE implementation provides the possibility to set an absolute/relative error tolerance. This accelerates computations by pruning nodes whose computation would not be very relevant for the estimation (you can find a more detailed explanation [here](https://arxiv.org/pdf/1304.4327.pdf) in section 8).

[This paper](https://arxiv.org/pdf/1206.6857.pdf), in section 5 after "Theorem 2", mentions how current implementation is not taking full advantage of error tolerance. The main idea is that, when a traverser was not able to prune and has to evaluate the `BaseCase()` for some combination of points, there's an amount of error tolerance that has not been used and could potentially be used later to try and prune even more.

As an example, this has been already implemented for `alpha` when using probabilistic error bounds in #1934.

rcurtin:
A PR is open as #1984.

rcurtin:
#1984 is merged, so I will go ahead and close this.

PR #1984, opened 2019-08-19
[GSoC] Reclaim unused KDE error tolerance

robertohueso:
This PR is an implementation of #1954.

### DIFF
--- a/src/mlpack/methods/kde/kde_rules_impl.hpp
+++ b/src/mlpack/methods/kde/kde_rules_impl.hpp
@@ -84,6 +84,9 @@ double KDERules<MetricType, KernelType, TreeType>::BaseCase(
                                           referenceSet.col(referenceIndex));
   densities(queryIndex) += kernel.Evaluate(distance);
 
+  // Update accumulated relative error tolerance for single-tree pruning.
+  accumError(queryIndex) += relError * kernel.Evaluate(distance);
+
   ++baseCases;
   lastQueryIndex = queryIndex;
   lastReferenceIndex = referenceIndex;
@@ -134,10 +137,12 @@ Score(const size_t queryIndex, TreeType& referenceNode)
   const double bound = maxKernel - minKernel;
 
   // Error tolerance of the current query point and reference node.
-  const double errorTolerance =
-      (absError + relError * minKernel) / referenceSet.n_cols;
+  const double absErrorTol = absError / referenceSet.n_cols;
+  const double relErrorTol = relError * minKernel;
+  const double errorTolerance = absErrorTol + relErrorTol;
+  const double pointAccumErrorTol = accumError(queryIndex) / refNumDesc;
 
-  if (bound <= accumError(queryIndex) + errorTolerance)
+  if (bound <= 2 * (errorTolerance + pointAccumErrorTol))
   {
     // Estimate values.
     double kernelValue;
@@ -157,7 +162,7 @@ Score(const size_t queryIndex, TreeType& referenceNode)
     score = DBL_MAX;
 
     // Update accumulated unused error tolerance.
-    accumError(queryIndex) = std::max(accumError(queryIndex) - bound, 0.0);
+    accumError(queryIndex) -= (bound - 2 * errorTolerance);
 
     // Store not used alpha for Monte Carlo.
     if (kernelIsGaussian && monteCarlo)
@@ -251,9 +256,9 @@ Score(const size_t queryIndex, TreeType& referenceNode)
     // Recurse.
     score = minDistance;
 
-    // Update accumulated unused error tolerance.
+    // Update accumulated unused absolute error tolerance.
     if (referenceNode.IsLeaf())
-      accumError(queryIndex) += errorTolerance;
+      accumError(queryIndex) += refNumDesc * absError / referenceSet.n_cols;
 
     // If node is going to be exactly computed, reclaim not used alpha for
     // Monte Carlo estimations.
@@ -333,11 +338,13 @@ Score(TreeType& queryNode, TreeType& referenceNode)
   const double bound = maxKernel - minKernel;
 
   // Error tolerance of the current nodes combination.
-  const double errorTolerance =
-      (absError + relError * minKernel) / referenceSet.n_cols;
+  const double absErrorTol = absError / referenceSet.n_cols;
+  const double relErrorTol = relError * minKernel;
+  const double errorTolerance = absErrorTol + relErrorTol;
+  const double pointAccumErrorTol = queryStat.AccumError() / refNumDesc;
 
   // If possible, avoid some calculations because of the error tolerance.
-  if (bound <= queryStat.AccumError() + errorTolerance)
+  if (bound <= 2 * (errorTolerance + pointAccumErrorTol))
   {
     // Auxiliary variables.
     double kernelValue;
@@ -367,7 +374,7 @@ Score(TreeType& queryNode, TreeType& referenceNode)
     score = DBL_MAX;
 
     // Update accumulated unused error tolerance.
-    queryStat.AccumError() = std::max(queryStat.AccumError() - bound, 0.0);
+    queryStat.AccumError() -= (bound - 2 * errorTolerance);
 
     // Store not used alpha for Monte Carlo.
     if (kernelIsGaussian && monteCarlo)
@@ -482,7 +489,10 @@ Score(TreeType& queryNode, TreeType& referenceNode)
 
     // Update accumulated unused error tolerance.
     if (referenceNode.IsLeaf() && queryNode.IsLeaf())
-      queryStat.AccumError() += errorTolerance;
+    {
+      queryStat.AccumError() += (refNumDesc * absError / referenceSet.n_cols) +
+                                (relError * refNumDesc * minKernel);
+    }
 
     // If node is going to be exactly computed, reclaim not used alpha for
     // Monte Carlo estimations.
--- a/src/mlpack/methods/kde/kde_rules_impl.hpp
+++ b/src/mlpack/methods/kde/kde_rules_impl.hpp
@@ -146,10 +146,14 @@ Score(const size_t queryIndex, TreeType& referenceNode)
   const double minKernel = kernel.Evaluate(maxDistance);
   const double bound = maxKernel - minKernel;
 
-  // Error tolerance of the current query point and reference node.
+  // Error tolerance of the current combination of query point and reference
+  // node.
   const double relErrorTol = relError * minKernel;
   const double errorTolerance = absErrorTol + relErrorTol;
 
+  // We relax the bound for pruning by accumError(queryIndex), so that if there
+  // is any leftover error tolerance from the rest of the traversal, we can use
+  // it here to prune more.
   double pointAccumErrorTol;
   if (alreadyDidRefPoint0)
     pointAccumErrorTol = accumError(queryIndex) / (refNumDesc - 1);
@@ -169,7 +173,8 @@ Score(const size_t queryIndex, TreeType& referenceNode)
     // Don't explore this tree branch.
     score = DBL_MAX;
 
-    // Update accumulated unused error tolerance.
+    // Subtract used error tolerance or add extra available tolerace from this
+    // prune.
     if (alreadyDidRefPoint0)
       accumError(queryIndex) -= (refNumDesc - 1) * (bound - 2 * errorTolerance);
     else
@@ -267,7 +272,7 @@ Score(const size_t queryIndex, TreeType& referenceNode)
     // Recurse.
     score = minDistance;
 
-    // Update accumulated unused absolute error tolerance.
+    // Add accumulated unused absolute error tolerance.
     if (referenceNode.IsLeaf())
     {
       if (alreadyDidRefPoint0)
@@ -352,10 +357,14 @@ Score(TreeType& queryNode, TreeType& referenceNode)
   const double minKernel = kernel.Evaluate(maxDistance);
   const double bound = maxKernel - minKernel;
 
-  // Error tolerance of the current nodes combination.
+  // Error tolerance of the current combination of query node and reference
+  // node.
   const double relErrorTol = relError * minKernel;
   const double errorTolerance = absErrorTol + relErrorTol;
 
+  // We relax the bound for pruning by queryStat.AccumError(), so that if there
+  // is any leftover error tolerance from the rest of the traversal, we can use
+  // it here to prune more.
   const double pointAccumErrorTol = queryStat.AccumError() / refNumDesc;
 
   // If possible, avoid some calculations because of the error tolerance.
@@ -376,7 +385,8 @@ Score(TreeType& queryNode, TreeType& referenceNode)
     // Prune.
     score = DBL_MAX;
 
-    // Update accumulated unused error tolerance.
+    // Subtract used error tolerance or add extra available tolerace from this
+    // prune.
     queryStat.AccumError() -= refNumDesc * (bound - 2 * errorTolerance);
 
     // Store not used alpha for Monte Carlo.
@@ -490,7 +500,7 @@ Score(TreeType& queryNode, TreeType& referenceNode)
     // Recurse.
     score = minDistance;
 
-    // Update accumulated unused error tolerance.
+    // Add accumulated unused error tolerance.
     if (referenceNode.IsLeaf() && queryNode.IsLeaf())
       queryStat.AccumError() += refNumDesc * 2 * errorTolerance;
 
--- a/src/mlpack/methods/kde/kde_rules_impl.hpp
+++ b/src/mlpack/methods/kde/kde_rules_impl.hpp
@@ -55,6 +55,9 @@ KDERules<MetricType, KernelType, TreeType>::KDERules(
     baseCases(0),
     scores(0)
 {
+  // Initialize accumError.
+  accumError = arma::vec(querySet.n_cols, arma::fill::zeros);
+
   // Initialize accumMCAlpha only if Monte Carlo estimations are available.
   if (monteCarlo && kernelIsGaussian)
     accumMCAlpha = arma::vec(querySet.n_cols, arma::fill::zeros);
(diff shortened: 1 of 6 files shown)


robertohueso:
This PR implementation doesn't only reclaim absolute error tolerance but also relative error tolerance.

After 1000+ tests with different random seeds I only got the error fixed in #1979 (which should be solved once it's merged), so I think this is ready for review :)

src/mlpack/methods/kde/kde_rules_impl.hpp
```diff
@@ -130,7 +133,11 @@ Score(const size_t queryIndex, TreeType& referenceNode)
   const double minKernel = kernel.Evaluate(maxDistance);
   const double bound = maxKernel - minKernel;
 
-  if (bound <= (absError + relError * minKernel) / referenceSet.n_cols)
+  // Error tolerance of the current query point and reference node.
+  const double errorTolerance =
+      (absError + relError * minKernel) / referenceSet.n_cols;
+
+  if (bound <= accumError(queryIndex) + errorTolerance)
```
rcurtin: It may be worth adding a comment for clarity here; something like

```
// We relax the bound for pruning by accumError(queryIndex), so that if there
// is any "leftover" error from the rest of the traversal, we can use it in this check.
```

or maybe something more elaborate, I am not sure, but I suspect someone reading this code will wonder what is going on here. :)

src/mlpack/methods/kde/kde_rules_impl.hpp
```diff
@@ -149,6 +156,9 @@ Score(const size_t queryIndex, TreeType& referenceNode)
     // Don't explore this tree branch.
     score = DBL_MAX;
 
+    // Update accumulated unused error tolerance.
+    accumError(queryIndex) = std::max(accumError(queryIndex) - bound, 0.0);
```
rcurtin: Hmm, I think it is possible to do a little more pruning here.  What this code does (if I read it right) is subtracts all the error that was "used" by this node (as represented by `bound`) from `accumError(queryIndex)`.  But actually we have a little extra error as given by `errorTolerance`.

So, if we do this instead, does it work?

```
accumError(queryIndex) -= std::max(bound - errorTolerance, 0.0);
```

That instead subtracts any "extra" error used above and beyond the original error tolerance.

rcurtin: In fact, to follow onto this, it may be possible to do

```
accumError(queryIndex) -= (bound - errorTolerance);
```

because if `bound - errorTolerance` is negative, this means that the bound we had was *within* the error tolerance, and we have some extra error tolerance that we can use.

However, I have not worked out whether or not this will actually work in practice.

src/mlpack/methods/kde/kde_rules_impl.hpp
```diff
@@ -130,7 +133,11 @@ Score(const size_t queryIndex, TreeType& referenceNode)
   const double minKernel = kernel.Evaluate(maxDistance);
   const double bound = maxKernel - minKernel;
 
-  if (bound <= (absError + relError * minKernel) / referenceSet.n_cols)
+  // Error tolerance of the current query point and reference node.
+  const double errorTolerance =
+      (absError + relError * minKernel) / referenceSet.n_cols;
```
rcurtin: I just realized, I think that maybe this should be `referenceNode.NumDescendants()` not `referenceSet.n_cols`.  The reason is this:

For simplicity let's consider relative error only and also single-tree only.  Our goal for KDE is to ensure that, overall,

```
| K(p_q) - K'(p_q) | < \epsilon * K(p_q)
```

where `K(p_q)` is the true kernel density estimate for `q` and `K'(q)` is our estimate.  Also, `K'(p_q) = k'(p_q, p_r0) + k'(p_q, p_r1) + ... + k'(p_q, p_rn)` where `k'(q, r_i)` is our kernel function estimate between the points `p_q` and `p_ri`.

We do this by making sure that the following is true for each point:

```
| k(p_q, p_ri) - k'(p_q, p_ri) | <= \epsilon * k(p_q, p_ri)
```

(I will finish this comment, but from a different device.  That means I have to hit "add single comment" now, and I'll add more later.)

rcurtin: In the single tree case, where we have a node `N_r`, we simply need to ensure that this is true for each descendant point `r_di`.  Given `k_max` and `k_min` as the maximum and minimum kernel evaluations between `p_q` and any `r_di`, we have that

```
| k(p_q, r_di) - k'(p_q, r_di) | <= | k_max - k_min |
```

for any `k'(p_q, r_di)` and actually if we are assuming that we will take `k'(p_q, r_di) = (k_max - k_min) / 2` we can get an even tighter bound:

```
| k(p_q, r_di) - k'(p_q, r_di) | <= | k_max - k_min | / 2.
```

and then, we also have that `k_min <= k(p_q, r_di)`.  So, we can plug this into both sides of our per-point bound at the bottom of our last comment to get

```
| k_max - k_min | / 2 <= \epsilon * k_min.
| k_max - k_min | <= \epsilon * k_min / 2.
```

This actually gives a very different pruning rule for relative error than what we have implemented, and while I was doing this derivation I realized that it will have a different form for absolute error.  I will try and derive that later.

rcurtin: IMG_20190820_161221

I derived it again for absolute error, and also found that the tests all pass when I change the code.  So I think that it is correct.  I haven't yet figured out what the "combined form" of the bound is though (maybe it's easy).

In fact, as you can see, the bound for relative error came out even tighter than I thought---instead of dividing by `referenceNode.NumDescendants()`, we can multiply by 2!

rcurtin: Finally, when I derived the "combined bound", I get the following:

We require that our kernel density estimate `K'(q)` satisfies

```
| K(q) - K'(q) | <= \epsilon_abs + \epsilon_rel * K(q).
```

Let's consider only a single node, in which case we require that for a reference node N_r,

```
\sum_{p_ri \in N_r} | k(q, p_ri) - k'(q, p_ri) | <= (| N_r | / | S_r |) \epsilon_abs + \epsilon_rel (\sum_{p_ri \in N_r} | k(q, p_ri) |
| N_r | | k_max(q, N_r) - k_min(q, N_r) | / 2 <= (| N_r | / | S_r |) \epsilon_abs + \epsilon_rel * | N_r | k_min(q, N_r)
| k_max(q, N_r) - k_min(q, N_r) | <= 2 \epsilon_abs / | S_r | + 2 \epsilon_rel k_min(q, N_r).
```

So, this means that the single-tree bound should be

```
bound <= (2.0 * absError / referenceSet.n_cols) + 2.0 * relError * minKernel
```

and similarly for the dual-tree bound.

So, I guess, at the end of this long journey:

(1) our bounds for KDE pruning were very loose, and we can do much better!
(2) the original error stems from the Tree-Independent Dual-Tree Algorithms paper
(3) the `referenceSet.n_cols` term is indeed part of the bound, but only for the absolute error part
(4) there is an extra factor of 2 in the bounding we can take advantage of
(5) the relative error bound term doesn't need to depend on `referenceSet.n_cols` at all

Anyway, if anything about this is unclear, just let me know.  We should make these changes, but it's up to you if we do it in this PR or another one.  But any benchmarking we do of the error reclamation should be done after these changes are made, since they'll make a huge performance difference (I think). :+1:
```

**Comment by rcurtin** [comment]
Sorry for so many comments in one night. :)  I thought about how the bounds should relax for both relative and absolute error, considering just the case where we exhaustively computed the KDE estimate for all the points in one node (N_r1) and want to prune another node (N_r2).

Let's take the absolute-error-only case first, because that's a little bit easier.

In this case, we require that our final kernel density estimate `K(q)` is such that

```
K(q) - K'(q) <= \epsilon.
```

and, the way we enforce this is by allowing each point to have no more than `\epsilon / | S_r |` points (where `S_r` is the reference set).  So, if we exhaustively computed the kernel density estimate exactly for every descendant point of `N_r1`, then in our error budget we have not used `| N_r1 | * \epsilon / | S_r |` error.  We could use this in our next pruning decision, where instead of pruning with the rule

```
| k_max(q, N_r2) - k_min(q, N_r2) | <= 2 \epsilon / |S_r|
```

we have an extra absolute budget of `C = | N_r1 | * \epsilon / | S_r |` for the entire node, or, `C / | N_r2 |` per point.  That gives a pruning rule of

```
| k_max(q, N_r2) - k_min(q, N_r2) | <= 2 (\epsilon / |S_r| + C / | N_r2 |).
```

So, basically, with absolute error only, we can cache `C = | N_r1 | * \epsilon / | S_r |` as the unused budget from a node where we computed exhaustively, and prune with `2 (\epsilon / | S_r | + C / | N_r2 |)` on the right hand side.

Now let's consider the relative error case.  In this setting we wish to compute the kernel density estimate such that

```
K(q) - K'(q) <= \epsilon K(q).
```

If we computed all the reference points in N_r1 exhaustively, this means that the error budget that we did not use is `C = \epsilon * \sum_{i \in N_r1} k(q, r_1)` (where the `i \in N_r1` just means "all descendant points of N_r1", because plaintext is too hard to use the notation from the tree-independent dual-tree algorithms paper :)).

Our pruning rule for N_r2 would normally be this:

```
| k_max(q, N_r2) - k_min(q, N_r2) | <= 2 \epsilon k_min(q, N_r2).
```

but for each point we have an extra allowable error of `C / | N_r2 |`.  This gives a modified pruning rule of

```
| k_max(q, N_r2) - k_min(q, N_r2) | <= 2 (\epsilon k_min(q, N_r2) + C / | N_r2 |).
```

Note that this kind of looks like a "combined" absolute and relative error, because that's exactly what it is!  Our relative error is for the node at hand, but our leftover error from before is no longer relative---it can now be treated as absolute.

Finally, we can now put these two things together for the "combined" absolute and relative error case that the code works for.  Assuming that `C` is again used for any unused error budget from `N_r1`, we get the pruning rule of

```
| k_max(q, N_r2) - k_min(q, N_r2) | <= 2 (\epsilon_abs / | S_r | + C / | N_r2 | + \epsilon_rel k_min(q, N_r2)
```

So, I think the general strategy of what you wrote is just the same, but I think that we can just adjust slightly how we are computing `accumError`, then when we use it, we just need to divide it by `referenceNode.NumDescendants()`.

As always please double-check my math to see if anything is wrong with it.  It's possible I overlooked something or made an error. :)  And let me know if anything was unclear.  I was a little bit lazy with the notation in places, but hope it still makes sense.

File: src/mlpack/methods/kde/kde_rules_impl.hpp (hunk 3 above)
Comment (rcurtin) [reply src/mlpack/methods/kde/kde_rules_impl.hpp]: Last comment for the night, I promise---

I ran the code with the rederived bound and it seems to work.  But there are some places in `kde_main_test.cpp` where we have to replace the `relError` in `BOOST_REQUIRE_CLOSE()` with `relError * 100` (just like in `kde_test.cpp`).

File: src/mlpack/methods/kde/kde_rules_impl.hpp [hunk 4]
```diff
@@ -110,7 +117,7 @@ Score(const size_t queryIndex, TreeType& referenceNode)
   if (tree::TreeTraits<TreeType>::FirstPointIsCentroid &&
       lastQueryIndex == queryIndex &&
       traversalInfo.LastReferenceNode() != NULL &&
-      traversalInfo.LastReferenceNode()->Point(0) == referenceNode.Point(0))
+      lastReferenceIndex == referenceNode.Point(0))
```
rcurtin: I didn't find that this made a difference---did you observe that `traversalInfo.LastReferenceNode().Point(0)` was sometimes different than `lastReferenceIndex`?

src/mlpack/methods/kde/kde_rules_impl.hpp
```diff
@@ -125,22 +132,38 @@ Score(const size_t queryIndex, TreeType& referenceNode)
...
     minDistance = r.Lo();
     maxDistance = r.Hi();
+
+    // Check if we are a self-child.
+    if (tree::TreeTraits<TreeType>::HasSelfChildren &&
+        referenceNode.Parent() != NULL &&
+        referenceNode.Parent()->Point(0) == referenceNode.Point(0))
+    {
+      alreadyDidRefPoint0 = true;
+    }
```
rcurtin: In this case, technically we have already done the base case between the query point and `referenceNode.Point(0)`, but I couldn't think of an easy way to cache that and pass it down the tree.  So it's unfortunate but it seems there is some duplicated computation here.  (I don't think it's worth solving in this PR, but I at least wanted to point it out.)

src/mlpack/methods/kde/kde_rules_impl.hpp
```diff
@@ -125,22 +132,38 @@ Score(const size_t queryIndex, TreeType& referenceNode)
...
+  const double errorTolerance = absErrorTol + relErrorTol;
+
+  // We relax the bound for pruning by accumError(queryIndex), so that if there
+  // is any leftover error tolerance from the rest of the traversal, we can use
+  // it here to prune more.
+  double pointAccumErrorTol;
+  if (alreadyDidRefPoint0)
+    pointAccumErrorTol = accumError(queryIndex) / (refNumDesc - 1);
+  else
+    pointAccumErrorTol = accumError(queryIndex) / refNumDesc;
```
rcurtin: If you wanted, you could rewrite this like this:

```
const double pointAccumErrorTol = accumError(queryIndex) /
    ((alreadyDidRefPoint0) ? (refNumDesc - 1) : refNumDesc);
```

It's up to you which you think is more readable. :+1:

src/mlpack/methods/kde/kde_rules_impl.hpp
```diff
@@ -150,6 +173,13 @@ Score(const size_t queryIndex, TreeType& referenceNode)
     // Don't explore this tree branch.
     score = DBL_MAX;
 
+    // Subtract used error tolerance or add extra available tolerace from this
+    // prune.
+    if (alreadyDidRefPoint0)
+      accumError(queryIndex) -= (refNumDesc - 1) * (bound - 2 * errorTolerance);
+    else
+      accumError(queryIndex) -= refNumDesc * (bound - 2 * errorTolerance);
```
rcurtin: The same ternary operator trick could be used here, but again, up to you how you want to handle it.

src/mlpack/tests/kde_test.cpp
```diff
@@ -135,12 +135,12 @@ BOOST_AUTO_TEST_CASE(KDETreeAsArguments)
  */
 BOOST_AUTO_TEST_CASE(GaussianKDEBruteForceTest)
 {
-  arma::mat reference = arma::randu(2, 200);
-  arma::mat query = arma::randu(2, 60);
+  arma::mat reference = arma::randu(2, 3000);
+  arma::mat query = arma::randu(2, 100);
```
rcurtin: Now that we've confirmed everything works, we can probably reduce the dataset sizes here a bit to save some time; what do you think?

src/mlpack/tests/main_tests/kde_test.cpp
```diff
@@ -88,7 +88,7 @@ BOOST_AUTO_TEST_CASE(KDEGaussianRTreeResultsMain)
 
   // Check whether results are equal.
   for (size_t i = 0; i < query.n_cols; ++i)
-    BOOST_REQUIRE_CLOSE(kdeEstimations[i], mainEstimations[i], relError);
+    BOOST_REQUIRE_CLOSE(kdeEstimations[i], mainEstimations[i], 100 * relError);
```
rcurtin: What's surprising about this is that it highlights how loose our bounds were before: when the tests passed with `relError`, it means that KDE was achieving relative error 100 times tighter than what was asked for.  I imagine that the speedup from the new pruning rules and bounds is very significant!

src/mlpack/methods/kde/kde_rules_impl.hpp
robertohueso: Yes, let's say we are using a binary tree:
1. The function that updates LastReferenceNode() is score.
Our tree:
```
    ...
   /
   A
 /   \
B     C
```
Lets say we are going to traverse A.
The traversal behaviour is:
```
Score(B) -> LastNode = B
Score(C) -> LastNode = C

if (Score(B) < Score(C))
  Traverse(B) -> LastReferenceIndex = Whatever there's in B.
...
```
And actually this change fixed `kernelMin < kernelEstimationValue < kernelMax` not being true :)

src/mlpack/methods/kde/kde_rules_impl.hpp
robertohueso: To be honest I'm not a big fan of the ternary operator, I feel like it's less readable :sweat_smile:.

src/mlpack/tests/kde_test.cpp
robertohueso: You're right. Done :)

rcurtin:
I did a little extra benchmarking for absolute error.  I used again the covertype dataset with absolute error of 0.1 and a bandwidth for the Gaussian kernel of 500.

```
            time      base cases      scored nodes

old         15682.0s  214.79B         19.74B
bound       15429.9s  210.01B         19.38B
new         11468.0s  154.96B         16.88B
with-mc'    12991.6s  154.30B         16.80B
```

I wrote `with-mc'` not `with-mc` because I am trying to tune the parameters to see if I can get any speedup.  For that trial I tried `--mc_entry_coef 8 --initial_sample_size 25 --mc_break_coef 0.15`.  I'll try some more as time goes on and see what happens. :)

rcurtin:
I'm really happy to have this merged.  If you want to keep benchmarking, I think maybe we can figure out where this code is fast (and where it's slow), and maybe see the way to further improvements, if you like. :+1:

rcurtin:
@robertohueso I did just a little bit more benchmarking using Monte Carlo sampling and was able to get some improvements.  As I swept `initial_sample_size`, I saw significant changes in runtime:

```
iss = 5      6928.089s
iss = 15    6222.226s
iss = 25   13367.768s
```

so by tuning the parameters, it seems like better performance can be found.  But that probably differs for every dataset.

